PR #11475, opened 2026-07-09
fix(operator): use global admission for namespace-restricted installs

sttts:
## Summary

- remove the separate admission webhooks, webhook Service, certificate resources, and webhook RBAC from namespace-restricted installations
- run webhook registration, certificate management, and CA injection only in the cluster-wide operator
- publish the namespace-restricted operator's effective feature gates and service-account principal on its existing ownership Lease
- make the global admission stack resolve feature gates per request namespace and authorize the namespace-restricted controller from that Lease
- keep the existing Lease-based reconciliation exclusion unchanged

Namespace-restricted mode remains development/test only and unsupported for production. Running different operator versions in parallel is strongly discouraged. The cluster-wide operator must be the same version or newer and ship the newest APIs.

## Validation

- `go test ./cmd/... ./internal/features/... ./internal/namespace_scope/... ./internal/webhook/...`
- Helm `make test` (30 tests passed)
- Helm `make lint`
- Helm `make generate-helm-docs`
- pre-commit hooks

Linear: [DYN-3412](https://linear.app/nvidia/issue/DYN-3412/dynamorelease130operatorwebhook-multi-operator-cluster-wide-defaulter)

### DIFF
--- a/deploy/helm/charts/platform/README.md
+++ b/deploy/helm/charts/platform/README.md
@@ -49,7 +49,7 @@ No action is required for most upgrades — the operator's built-in cert-control
 
 ---
 
-## ⚠️ Important: Cluster-Wide and Namespace-Restricted Deployment
+## ⚠️ DEVELOPMENT AND TESTING ONLY: Namespace-Restricted Deployment
 
 Deploy one cluster-wide operator per cluster. It owns the Custom Resource Definitions (CRDs),
 conversion webhook, and conversion certificate authority (CA).
@@ -57,9 +57,9 @@ conversion webhook, and conversion certificate authority (CA).
 > [!WARNING]
 > Namespace-restricted mode is only for development and testing. It is not supported for production.
 
-A namespace-restricted operator owns reconciliation, validation, and mutation only in its target
-namespace. Its Lease makes the cluster-wide operator skip that namespace. Install it alongside an
-existing cluster-wide operator with CRD management disabled:
+A namespace-restricted operator reconciles only its target namespace and serves no webhooks. Its
+Lease makes the cluster-wide reconcilers skip that namespace and provides feature gates to global
+admission. Install it alongside an existing cluster-wide operator with CRD management disabled:
 
 ```bash
 helm install dynamo-test dynamo-platform-${RELEASE_VERSION}.tgz \
@@ -105,7 +105,7 @@ Kubernetes: `>=1.30.0-0`
 | dynamo-operator.etcdAddr | string | `""` | etcd server address for an external etcd instance. Only needed when using external etcd without the bundled subchart. Format: `http://hostname:2379` or `https://hostname:2379` |
 | dynamo-operator.modelExpressURL | string | `""` | URL for the Model Express server if not deployed by this helm chart. This is ignored if Model Express server is installed by this helm chart (global.model-express.enabled is true). |
 | dynamo-operator.namespaceRestriction | object | `{"enabled":false,"lease":{"duration":"30s","renewInterval":"10s"},"targetNamespace":null}` | DEVELOPMENT AND TESTING ONLY: Namespace-restricted mode is not supported for production. Use cluster-wide mode for production deployments. |
-| dynamo-operator.namespaceRestriction.enabled | bool | `false` | DEVELOPMENT AND TESTING ONLY: Enable namespace-restricted reconciliation and admission. Not supported for production. |
+| dynamo-operator.namespaceRestriction.enabled | bool | `false` | DEVELOPMENT AND TESTING ONLY: Enable namespace-restricted reconciliation. Not supported for production. |
 | dynamo-operator.namespaceRestriction.targetNamespace | string | `nil` | DEVELOPMENT AND TESTING ONLY: Target namespace. Defaults to the Helm release namespace. |
 | dynamo-operator.namespaceRestriction.lease | object | `{"duration":"30s","renewInterval":"10s"}` | DEVELOPMENT AND TESTING ONLY: Namespace ownership Lease settings. |
 | dynamo-operator.namespaceRestriction.lease.duration | string | `"30s"` | DEVELOPMENT AND TESTING ONLY: Namespace ownership Lease duration. |
@@ -152,7 +152,7 @@ Kubernetes: `>=1.30.0-0`
 | dynamo-operator.webhook.failurePolicy | string | `"Fail"` | Webhook failure policy controls how Kubernetes handles requests when the webhook is unavailable. 'Fail' (recommended for production) rejects requests if the webhook cannot be reached, ensuring strict validation. 'Ignore' allows requests through if the webhook is unavailable, providing availability over validation guarantees. |
 | dynamo-operator.webhook.podCheckpointRestoreFailurePolicy | string | `"Ignore"` | Failure policy for the Pod CREATE checkpoint-restore mutating webhook. Defaults to Ignore so a webhook outage falls back to cold-start instead of blocking workload Pods. |
 | dynamo-operator.webhook.timeoutSeconds | int | `10` | Timeout in seconds for webhook validation calls. If the webhook doesn't respond within this time, the request will be handled according to the failurePolicy. |
-| dynamo-operator.webhook.namespaceSelector | object | `{}` | Unsupported; must remain empty. Helm configures global admission for the cluster-wide operator and target-namespace admission for restricted operators. |
+| dynamo-operator.webhook.namespaceSelector | object | `{}` | Unsupported; must remain empty. Admission and conversion webhooks are always global and owned by the cluster-wide operator. Installation fails if set. |
 | dynamo-operator.webhook.certManager.enabled | bool | `false` | Whether to use cert-manager for automatic certificate management. Requires cert-manager to be installed in the cluster. When enabled, cert-manager will provision and rotate certificates instead of the operator's built-in cert-controller. |
 | dynamo-operator.webhook.certManager.certificate.duration | string | `"8760h"` | Certificate duration for webhook certificates managed by cert-manager (e.g., "8760h" for 1 year). cert-manager will automatically renew the certificate before it expires. |
 | dynamo-operator.webhook.certManager.certificate.renewBefore | string | `"360h"` | Time before certificate expiration to trigger renewal (e.g., "360h" for 15 days). cert-manager will attempt to renew the certificate when this threshold is reached. |
--- a/deploy/helm/charts/platform/README.md.gotmpl
+++ b/deploy/helm/charts/platform/README.md.gotmpl
@@ -49,7 +49,7 @@ No action is required for most upgrades — the operator's built-in cert-control
 
 ---
 
-## ⚠️ Important: Cluster-Wide and Namespace-Restricted Deployment
+## ⚠️ DEVELOPMENT AND TESTING ONLY: Namespace-Restricted Deployment
 
 Deploy one cluster-wide operator per cluster. It owns the Custom Resource Definitions (CRDs),
 conversion webhook, and conversion certificate authority (CA).
@@ -57,9 +57,9 @@ conversion webhook, and conversion certificate authority (CA).
 > [!WARNING]
 > Namespace-restricted mode is only for development and testing. It is not supported for production.
 
-A namespace-restricted operator owns reconciliation, validation, and mutation only in its target
-namespace. Its Lease makes the cluster-wide operator skip that namespace. Install it alongside an
-existing cluster-wide operator with CRD management disabled:
+A namespace-restricted operator reconciles only its target namespace and serves no webhooks. Its
+Lease makes the cluster-wide reconcilers skip that namespace and provides feature gates to global
+admission. Install it alongside an existing cluster-wide operator with CRD management disabled:
 
 ```bash
 helm install dynamo-test dynamo-platform-${RELEASE_VERSION}.tgz \
--- a/deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
+++ b/deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- if not .Values.namespaceRestriction.enabled }}
 ---
 # Role to manage the webhook certificate secret (create placeholder, read, update)
 apiVersion: rbac.authorization.k8s.io/v1
@@ -117,3 +118,4 @@ subjects:
 - kind: ServiceAccount
   name: {{ include "dynamo-operator.fullname" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/deploy/helm/charts/platform/components/operator/values.yaml
+++ b/deploy/helm/charts/platform/components/operator/values.yaml
@@ -33,7 +33,7 @@ env: []
 
 # -- DEVELOPMENT AND TESTING ONLY: Namespace-restricted mode is not supported for production. Use cluster-wide mode for production deployments.
 namespaceRestriction:
-  # -- DEVELOPMENT AND TESTING ONLY: Enable namespace-restricted reconciliation and admission. Not supported for production.
+  # -- DEVELOPMENT AND TESTING ONLY: Enable namespace-restricted reconciliation. Not supported for production.
   enabled: false
   # -- DEVELOPMENT AND TESTING ONLY: Target namespace. Defaults to the Helm release namespace.
   targetNamespace: ""
@@ -219,8 +219,8 @@ webhook:
   # Timeout for webhook calls (in seconds)
   timeoutSeconds: 10
 
-  # -- Unsupported; must remain empty. Helm configures global admission for the
-  # cluster-wide operator and target-namespace admission for restricted operators.
+  # -- Unsupported; must remain empty. Admission and conversion webhooks are always
+  # global and owned by the cluster-wide operator. Installation fails if set.
   namespaceSelector: {}
 
   # cert-manager integration (optional, alternative to built-in cert-controller)
--- a/deploy/helm/charts/platform/tests/namespace_restriction_webhook_rbac_test.yaml
+++ b/deploy/helm/charts/platform/tests/namespace_restriction_webhook_rbac_test.yaml
@@ -24,28 +24,14 @@ tests:
               - patch
         documentIndex: 2
 
-  - it: prevents a namespace-restricted operator from patching CRDs
+  - it: renders no webhook RBAC for a namespaced operator
     template: charts/dynamo-operator/templates/webhook-rbac.yaml
     set:
       dynamo-operator.namespaceRestriction.enabled: true
       dynamo-operator.upgradeCRD: false
     asserts:
       - hasDocuments:
-          count: 4
-      - notContains:
-          path: rules
-          content:
-            apiGroups:
-              - apiextensions.k8s.io
-            resources:
-              - customresourcedefinitions
-            verbs:
-              - get
-              - list
-              - watch
-              - update
-              - patch
-        documentIndex: 2
+          count: 0
 
   - it: omits cluster-scoped admission and CRD rules from the namespace-restricted manager Role
     template: charts/dynamo-operator/templates/manager-rbac.yaml
--- a/deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
+++ b/deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
@@ -40,63 +40,25 @@ tests:
           path: webhooks[3].namespaceSelector
         documentIndex: 1
 
-  - it: scopes namespace-restricted admission to the target namespace
+  - it: renders no admission configuration for a namespaced installation
     template: charts/dynamo-operator/templates/webhook-configuration.yaml
     set:
       dynamo-operator.namespaceRestriction.enabled: true
-      # Exercise a YAML boolean-like namespace to ensure the rendered selector stays a string.
-      dynamo-operator.namespaceRestriction.targetNamespace: "true"
       dynamo-operator.upgradeCRD: false
     asserts:
       - hasDocuments:
-          count: 2
-      - equal:
-          path: webhooks[0].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 0
-      - equal:
-          path: webhooks[1].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 0
-      - equal:
-          path: webhooks[2].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 0
-      - equal:
-          path: webhooks[3].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 0
-      - equal:
-          path: webhooks[4].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 0
-      - equal:
-          path: webhooks[0].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 1
-      - equal:
-          path: webhooks[1].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 1
-      - equal:
-          path: webhooks[2].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 1
-      - equal:
-          path: webhooks[3].namespaceSelector.matchLabels["kubernetes.io/metadata.name"]
-          value: "true"
-        documentIndex: 1
+          count: 0
 
-  - it: renders a webhook service for a namespace-restricted installation
+  - it: renders no webhook service for a namespaced installation
     template: charts/dynamo-operator/templates/webhook-service.yaml
     set:
       dynamo-operator.namespaceRestriction.enabled: true
       dynamo-operator.upgradeCRD: false
     asserts:
       - hasDocuments:
-          count: 1
+          count: 0
 
-  - it: renders webhook certificates for a namespace-restricted installation
+  - it: renders no webhook certificates for a namespaced installation
     template: charts/dynamo-operator/templates/webhook-certificates.yaml
     set:
       dynamo-operator.namespaceRestriction.enabled: true
@@ -104,4 +66,4 @@ tests:
       dynamo-operator.webhook.certManager.enabled: true
     asserts:
       - hasDocuments:
-          count: 4
+          count: 0
--- a/deploy/operator/cmd/main.go
+++ b/deploy/operator/cmd/main.go
@@ -309,23 +309,36 @@ func main() {
 	setupLog.Info("Initializing observability metrics")
 	observability.InitMetrics()
 
-	// Set up webhook certificate management.
-	// A direct (non-cached) client is needed because the manager's cache isn't started yet.
-	directClient, err := client.New(mgr.GetConfig(), client.Options{Scheme: crdScheme})
+	gates, err := features.New(mainCtx, mgr, operatorCfg)
 	if err != nil {
-		setupLog.Error(err, "unable to create direct client for cert management")
+		setupLog.Error(err, "unable to resolve operator feature gates")
 		os.Exit(1)
 	}
-	certMgr, err := internalcert.NewCertManager(directClient, &operatorCfg.Server.Webhook)
-	if err != nil {
-		setupLog.Error(err, "unable to create cert manager")
-		os.Exit(1)
+	runtimeConfig.Gate = gates
+
+	var operatorPrincipal string
+	if sa, ns := os.Getenv("POD_SERVICE_ACCOUNT"), os.Getenv("POD_NAMESPACE"); sa != "" && ns != "" {
+		operatorPrincipal = fmt.Sprintf("system:serviceaccount:%s:%s", ns, sa)
 	}
-	// Auto mode runs one synchronous certificate refresh with the direct client,
-	// then registers the cert-controller with the not-yet-started manager.
-	if err = certMgr.SetupAndRunOnce(mainCtx, mgr); err != nil {
-		setupLog.Error(err, "failed to setup webhook certificate management")
-		os.Exit(1)
+
+	// Only the cluster-wide operator owns webhook certificates.
+	var directClient client.Client
+	var certMgr *internalcert.CertManager
+	if isClusterWide {
+		directClient, err = client.New(mgr.GetConfig(), client.Options{Scheme: crdScheme})
+		if err != nil {
+			setupLog.Error(err, "unable to create direct client for cert management")
+			os.Exit(1)
+		}
+		certMgr, err = internalcert.NewCertManager(directClient, &operatorCfg.Server.Webhook)
+		if err != nil {
+			setupLog.Error(err, "unable to create cert manager")
+			os.Exit(1)
+		}
+		if err = certMgr.SetupAndRunOnce(mainCtx, mgr); err != nil {
+			setupLog.Error(err, "failed to setup webhook certificate management")
+			os.Exit(1)
+		}
 	}
 
 	// Initialize namespace scope mechanism
@@ -345,6 +358,8 @@ func main() {
 			operatorVersion,
 			operatorCfg.Namespace.Scope.LeaseDuration.Duration,
 			operatorCfg.Namespace.Scope.LeaseRenewInterval.Duration,
+			gates,
+			operatorPrincipal,
 		)
 		if err != nil {
 			setupLog.Error(err, "unable to create namespace scope marker lease manager")
@@ -411,13 +426,6 @@ func main() {
 		setupLog.Error(err, "unable to register resource counter")
 		os.Exit(1)
 	}
-
-	gates, err := features.New(mainCtx, mgr, operatorCfg)
-	if err != nil {
-		setupLog.Error(err, "unable to resolve operator feature gates")
-		os.Exit(1)
-	}
-	runtimeConfig.Gate = gates
 
 	dockerSecretRetriever := secrets.NewDockerSecretIndexer(mgr.GetAPIReader(), restrictedNamespace)
 	// refresh whenever a secret is created/deleted/updated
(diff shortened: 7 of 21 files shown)


deploy/helm/charts/platform/README.md
```diff
@@ -115,16 +120,13 @@ The chart includes built-in validation to prevent all operator conflicts:
...
 | dynamo-operator.modelExpressURL | string | `""` | URL for the Model Express server if not deployed by this helm chart. This is ignored if Model Express server is installed by this helm chart (global.model-express.enabled is true). |
-| dynamo-operator.namespaceRestriction | object | `{"enabled":false,"lease":{"duration":"30s","renewInterval":"10s"},"targetNamespace":null}` | DEPRECATED: Namespace-restricted mode is deprecated and will be removed in a future release. Use cluster-wide mode (the default) instead. Do not enable this for new deployments. |
+| dynamo-operator.namespaceRestriction | object | `{"enabled":false,"targetNamespace":null}` | DEPRECATED: Namespace-restricted mode is deprecated and will be removed in a future release. Use cluster-wide mode (the default) instead. Do not enable this for new deployments. |
 | dynamo-operator.namespaceRestriction.enabled | bool | `false` | DEPRECATED: Do not enable for new deployments. Namespace-restricted mode is deprecated. |
 | dynamo-operator.namespaceRestriction.targetNamespace | string | `nil` | DEPRECATED: Only used in namespace-restricted mode, which is deprecated. |
```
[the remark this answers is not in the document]
sttts: We should keep it deprecated.

deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
```diff
@@ -0,0 +1,22 @@
+suite: namespace restriction webhooks
```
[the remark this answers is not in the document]
sttts: Fixed in cd6c90795f: added the required SPDX header. The copyright check now has all three affected Helm test files covered.

deploy/operator/internal/namespace_scope/marker_watcher.go
```diff
@@ -0,0 +1,134 @@
...
+	marker, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		w.logger.V(1).Info("Ignoring non-ConfigMap namespace exclusion event", "type", fmt.Sprintf("%T", obj))
+		return nil
+	}
+	if marker.Name != MarkerName {
+		return nil
+	}
+	return marker
+}
```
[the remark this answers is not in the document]
sttts: Obsolete after the design revision: the ConfigMap marker implementation was removed. The current branch retains the existing Lease-based reconciliation claims, so there is no MarkerWatcher or self-created ConfigMap exclusion path.

deploy/helm/charts/platform/tests/namespace_restriction_webhook_rbac_test.yaml
```diff
@@ -0,0 +1,16 @@
+suite: namespace restriction webhook RBAC
```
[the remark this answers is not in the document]
sttts: Fixed in cd6c90795f: added the required SPDX header. The copyright check now has all three affected Helm test files covered.

deploy/helm/charts/platform/components/operator/values.yaml
```diff
@@ -30,21 +30,23 @@ featureGates:
...
-  # -- DEPRECATED: Do not enable for new deployments. Namespace-restricted mode is deprecated.
+  # -- Enable development/test-only namespace-restricted reconciliation.
   enabled: false
-  # -- DEPRECATED: Only used in namespace-restricted mode, which is deprecated.
+  # -- Explicitly serve validation for the target namespace from this namespaced operator. CRD schema/CEL validation always applies; defaulting, mutation, and conversion remain cluster-wide. Disabled by default.
+  runNamespacedValidation: false
+  # -- Reconciliation target for development/test-only namespace-restricted mode. Defaults to the release namespace.
   targetNamespace: ""
-  # -- DEPRECATED: Only used in namespace-restricted mode, which is deprecated.
+  # -- Reconciliation ownership claim settings for development/test-only namespace-restricted mode.
```
sttts: put DEVELOPMENT AND TESTING ONLY: on each of them.

deploy/helm/charts/platform/README.md
```diff
@@ -43,53 +43,82 @@ The Dynamo Platform Helm chart deploys the complete Dynamo Kubernetes Platform i
 
 ### Webhooks are now mandatory (v1.0.0+)
 
-The `webhook.enabled` Helm value has been removed. Admission webhooks are now a required component of the operator and cannot be disabled. This change aligns with the upcoming addition of CRD conversion webhooks, which are mandatory for multi-version API support.
+The `webhook.enabled` Helm value has been removed. Admission and conversion webhooks are
+required and cannot be disabled on the cluster-wide operator. Namespaced operators never
+serve webhooks; they use the cluster-wide operator's global API implementation.
```
sttts: not quite right. Admission validation can be served for dev and testing.

deploy/helm/charts/platform/README.md
```diff
@@ -43,53 +43,82 @@ The Dynamo Platform Helm chart deploys the complete Dynamo Kubernetes Platform i
...
 
-- ✅ **Recommended**: Deploy one cluster-wide operator per cluster
-- ❌ **Not Recommended**: Multiple cluster-wide operators in the same cluster
+Namespaced operators reconcile only their target namespace and renew a namespace-scoped
+Lease that makes the cluster-wide operator's reconcilers hold off. By default they do not
+serve admission. CRD schema and CEL validation still apply, while defaulting, mutation,
+and conversion are always served globally by the cluster-wide operator.
 
-### Multiple Namespace-Scoped Operators (Advanced)
+### Installing a Namespaced Development Operator
```
sttts: DEVELOPMENT AND TESTING ONLY:

deploy/helm/charts/platform/README.md
```diff
@@ -115,18 +144,19 @@ The chart includes built-in validation to prevent all operator conflicts:
...
-| dynamo-operator.namespaceRestriction.lease.duration | string | `"30s"` | DEPRECATED: Lease duration for namespace-restricted mode, which is deprecated. |
-| dynamo-operator.namespaceRestriction.lease.renewInterval | string | `"10s"` | DEPRECATED: Lease renew interval for namespace-restricted mode, which is deprecated. |
-| dynamo-operator.gpuDiscovery | object | `{"enabled":true}` | DEPRECATED: GPU discovery for namespace-scoped operators is deprecated along with namespace-restricted mode. |
-| dynamo-operator.gpuDiscovery.enabled | bool | `true` | DEPRECATED: Only relevant when namespaceRestriction is enabled, which is deprecated. |
+| dynamo-operator.namespaceRestriction | object | `{"enabled":false,"lease":{"duration":"30s","renewInterval":"10s"},"runNamespacedValidation":false,"targetNamespace":null}` | DEVELOPMENT AND TESTING ONLY: Namespace-restricted mode is not supported for production. Use cluster-wide mode (the default) for production deployments. |
+| dynamo-operator.namespaceRestriction.enabled | bool | `false` | Enable development/test-only namespace-restricted reconciliation. |
```
sttts: DEVELOPMENT AND TESTING ONLY:  for each

deploy/helm/charts/platform/README.md.gotmpl
```diff
@@ -43,53 +43,82 @@ The Dynamo Platform Helm chart deploys the complete Dynamo Kubernetes Platform i
 
 ### Webhooks are now mandatory (v1.0.0+)
 
-The `webhook.enabled` Helm value has been removed. Admission webhooks are now a required component of the operator and cannot be disabled. This change aligns with the upcoming addition of CRD conversion webhooks, which are mandatory for multi-version API support.
+The `webhook.enabled` Helm value has been removed. Admission and conversion webhooks are
+required and cannot be disabled on the cluster-wide operator. Namespaced operators never
+serve webhooks; they use the cluster-wide operator's global API implementation.
```
sttts: oops, this is the template. Everything written about on README.md applies here to the tmpl.

deploy/operator/api/config/v1alpha1/types.go
```diff
@@ -152,20 +152,24 @@ type LeaderElectionConfiguration struct {
 
 // NamespaceConfiguration determines operator namespace mode.
 type NamespaceConfiguration struct {
-	// Deprecated: Namespace-restricted mode is deprecated and will be removed in a future release.
-	// Use cluster-wide mode (leave Restricted empty) instead.
+	// Restricted enables development/test-only namespace-restricted reconciliation.
```
sttts: DEVELOPMENT AND TESTING ONLY:  for each field.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment_handler.go
```diff
@@ -116,81 +92,62 @@ func (h *DynamoComponentDeploymentHandler) validateUpdate(
...
+	logger.Info("validate delete", "name", deployment.Name, "namespace", deployment.Namespace)
+
+	// No special validation needed for deletion
 	return nil, nil
 }
 
 // RegisterWithManager registers the webhook with the manager.
-// The handler is automatically wrapped with LeaseAwareValidator to add namespace exclusion logic
-// and ObservedValidator to add metrics collection.
 func (h *DynamoComponentDeploymentHandler) RegisterWithManager(mgr manager.Manager) error {
```
sttts: why remove all the comments?

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
```diff
@@ -193,8 +193,6 @@ func (h *DynamoGraphDeploymentHandler) validateDelete(
 }
 
 // RegisterWithManager registers the webhook with the manager.
-// The handler is automatically wrapped with LeaseAwareValidator to add namespace exclusion logic
-// and ObservedValidator to add metrics collection.
 func (h *DynamoGraphDeploymentHandler) RegisterWithManager(mgr manager.Manager) error {
 	h.registerWithManager(
```
sttts, line 197: why remove all the comments?

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest_handler.go
```diff
@@ -125,12 +125,8 @@ func (h *DynamoGraphDeploymentRequestHandler) ValidateDelete(ctx context.Context
 }
 
 // RegisterWithManager registers the webhook with the manager.
-// The handler is automatically wrapped with LeaseAwareValidator to add namespace exclusion logic.
 func (h *DynamoGraphDeploymentRequestHandler) RegisterWithManager(mgr manager.Manager) error {
-	// Wrap the handler with lease-aware logic for cluster-wide coordination
 	leaseAwareValidator := internalwebhook.NewLeaseAwareValidator(h, internalwebhook.GetExcludedNamespaces())
-
-	// Wrap with metrics collection
```
sttts, line 126: why remove all the comments?

deploy/operator/internal/webhook/validation/dynamomodel_handler.go
```diff
@@ -120,12 +120,8 @@ func (h *DynamoModelHandler) ValidateDelete(ctx context.Context, obj runtime.Obj
 }
 
 // RegisterWithManager registers the webhook with the manager.
-// The handler is automatically wrapped with LeaseAwareValidator to add namespace exclusion logic.
 func (h *DynamoModelHandler) RegisterWithManager(mgr manager.Manager) error {
-	// Wrap the handler with lease-aware logic for cluster-wide coordination
 	leaseAwareValidator := internalwebhook.NewLeaseAwareValidator(h, internalwebhook.GetExcludedNamespaces())
-
-	// Wrap with metrics collection
```
sttts, line 128: why remove all the comments?

deploy/helm/charts/platform/README.md.gotmpl
sttts: Applied the same webhook wording and development/testing-only heading to , then regenerated .

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
sttts: Restored the LeaseAwareValidator and ObservedValidator explanatory comments unchanged.

deploy/operator/api/config/v1alpha1/types.go
sttts: Prefixed every namespace configuration field comment, including lease duration and renewal, with  and regenerated the API reference.

deploy/helm/charts/platform/README.md
sttts: Updated the heading to .

deploy/helm/charts/platform/README.md
sttts: Corrected this: namespaced operators may serve validation with ; conversion, defaulting, and mutation remain cluster-wide.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment_handler.go
sttts: Restored the LeaseAwareValidator and ObservedValidator explanatory comments unchanged.

deploy/helm/charts/platform/components/operator/values.yaml
sttts: Prefixed every namespace-restricted and GPU-discovery Helm value with  in both the component and platform values files.

deploy/helm/charts/platform/README.md
sttts: Regenerated the README table after prefixing every namespace-restricted and GPU-discovery field description with .

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest_handler.go
sttts: Restored the LeaseAwareValidator and metrics-wrapper explanatory comments unchanged.

deploy/operator/internal/webhook/validation/dynamomodel_handler.go
sttts: Restored the LeaseAwareValidator and metrics-wrapper explanatory comments unchanged.

deploy/operator/internal/namespace_scope/lease_watcher.go
```diff
@@ -59,23 +59,51 @@ func NewLeaseWatcher(config *rest.Config) (*LeaseWatcher, error) {
...
+// admission to the namespaced operator. Leases created by older operators have
+// no ownership label and are treated as namespaced because those operators
+// always served validation.
+func (lw *LeaseWatcher) NamespacedValidationOwner(namespace string) bool {
+	lease, exists := lw.activeLease(namespace)
 	if !exists {
 		return false
 	}
 
+	owner, found := lease.Labels[ValidationOwnerLabel]
```
sttts: [P1] Do not trust a namespace-local Lease as validation authority

Every Lease with the well-known name reaches this code because `isNamespaceScopeMarker` checks only `lease.Name`; the owner label is also ordinary, forgeable metadata, and this compatibility branch even transfers validation when it is absent. A namespace user that can create/update Leases can therefore publish a fresh `RenewTime`/duration (with no label or `namespaced`) without installing a validating webhook. The cluster-wide `LeaseAwareValidator` then returns allow and all Go validation is bypassed. This is the same trust-boundary problem that was identified for the removed ConfigMap marker; retaining the Lease does not authenticate the claimant. Validation ownership needs a claim ordinary tenant RBAC cannot forge (for example, a protected coordination resource/capability), rather than a name or label on a tenant-writable object.

deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
```diff
@@ -64,9 +65,8 @@ subjects:
   name: {{ include "dynamo-operator.fullname" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
 ---
-# ClusterRole for patching webhook configurations and CRDs (cluster-scoped resources).
-# These are cluster-scoped resources that cannot be granted via a namespace-scoped Role,
-# so they are always provided as a dedicated ClusterRole regardless of namespace-restriction mode.
+# ClusterRole for injecting CA bundles. Namespaced validation may patch only its
```
sttts: [P1] Scope namespaced CA injection to the exact webhook configuration

The comment says "only its labeled" configuration, but the rendered namespaced ClusterRole grants `get/list/watch/update/patch` on every `ValidatingWebhookConfiguration` in the cluster; RBAC does not enforce label selectors. This leaves the namespaced operator service account able to rewrite the global operator's or another tenant's admission policy. The implementation also selects only `part-of=dynamo-operator` plus release namespace, so two releases colocated in one namespace will actually patch each other's CA bundles. For the namespaced branch, render `resourceNames` for the exact `<fullname>-<release-namespace>-validating` object and have the injector `Get`/patch that object instead of listing all configurations.

deploy/operator/internal/controller_common/namespace_filter.go
```diff
@@ -0,0 +1,31 @@
...
+
+// WithNamespaceExclusion prevents queued requests for namespaces managed by a
+// namespaced operator from reaching the wrapped reconciler.
+func WithNamespaceExclusion(delegate reconcile.Reconciler, runtimeConfig *RuntimeConfig) reconcile.Reconciler {
+	return &namespaceFilteringReconciler{delegate: delegate, runtimeConfig: runtimeConfig}
+}
+
+func (r *namespaceFilteringReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if ShouldSkipReconciliation(ctx, r.runtimeConfig, req.Namespace) {
+		return ctrl.Result{}, nil
```
sttts: [P1] Re-enqueue work when reconciliation ownership returns

Returning an empty result permanently forgets an already queued request. The event predicate also drops object updates while the Lease is active, while `LeaseWatcher` only changes its in-memory map on expiry/deletion and never enqueues the affected namespace's resources. If the namespaced operator crashes and an object update arrives before the Lease expires, neither operator processes it; after expiry, the cluster-wide operator does not resume until some unrelated later event happens. The ownership transition needs to fan out reconciles on Lease release/expiry (or retain bounded requeues here until the claim clears) so failover actually converges.

deploy/operator/internal/namespace_scope/lease_manager.go
```diff
@@ -195,24 +217,35 @@ func (lm *LeaseManager) createOrUpdateLease(ctx context.Context) error {
...
-	lm.logger.V(1).Info("Refreshed namespace scope marker lease")
+	lm.logger.V(1).Info("Refreshed namespace reconciliation lease")
 	return nil
 }
 
+func (lm *LeaseManager) validationOwner() string {
+	if lm.runNamespacedValidation {
+		return ValidationOwnerNamespaced
+	}
+	return ValidationOwnerClusterWide
```
sttts: [P1] Keep validation local until the cluster-wide peer understands this label

With the new default (`runNamespacedValidation=false`), the namespaced chart renders no validating configuration and writes `cluster-wide` here. A pre-PR cluster-wide operator does not interpret this label: its `LeaseAwareValidator` still calls `Contains(namespace)` and skips validation for every active Lease. Consequently, deploying the newer namespaced operator against an older cluster-wide operator silently leaves the namespace with no Go validator at all. That is exactly the development flow described as acceptable when testing newer namespaced code. Add a capability/acknowledgement handshake before returning validation to the cluster-wide peer, keep serving namespaced validation until such an acknowledgement exists, or fail startup when the peer is incompatible.

sttts (commented):
Static code review only, as requested; I did not run tests, builds, linters, or formatters. I traced the Helm resources, RBAC, certificate setup, Lease transitions, admission dispatch, and controller handoff end to end. The ownership split is directionally sound, but the four inline findings below leave validation or reconciliation without a trustworthy, deterministic owner, so I do not think this is ready to merge yet.

deploy/operator/internal/namespace_scope/lease_manager.go
sttts: Double validation is not unsafe. So can we make it so that not knowing the label leads to exactly that?

deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
sttts: Fixed in 3b18a320e8. The namespaced ClusterRole now restricts update/patch to the exact `<fullname>-<release-namespace>-validating` resourceName, and the CA injector Gets and patches that exact configured object instead of listing by labels. Capability discovery uses a separate read-only get/list/watch rule; it cannot mutate another release’s VWC.

deploy/operator/internal/controller_common/namespace_filter.go
sttts: Fixed in 3b18a320e8. Excluded reconciles now return a bounded 10s requeue, and the event predicate no longer drops events merely because a namespace currently has a claim. Requests therefore remain live and resume after Lease deletion/expiry; focused tests cover both behaviors.

deploy/operator/internal/namespace_scope/lease_manager.go
sttts: Fixed in 3b18a320e8 for both skew directions. A new cluster-wide operator treats an unlabeled/unknown Lease as non-delegating, so an older namespaced validator may run in parallel. Conversely, every new namespaced install keeps a scoped validation fallback and delegates only after observing `nvidia.com/dynamo-operator-validation-lease-capability=v1` on a protected, global VWC. An older cluster-wide install does not advertise that capability, so the local validator stays active. Unknown peers therefore produce at worst duplicate validation, not a gap.

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
```diff
@@ -170,6 +171,13 @@ func (h *DynamoGraphDeploymentHandler) ValidateDelete(ctx context.Context, obj r
 	return h.validateDelete(ctx, obj, nvidiacomv1beta1.DynamoGraphDeploymentGVK)
 }
 
+func (h *DynamoGraphDeploymentHandler) effectiveGroveGate(ctx context.Context) bool {
```
sttts: keep it small. Could be groveEnabled(ctx)

deploy/operator/internal/webhook/validation/dynamographdeployment_test.go
```diff
@@ -1697,15 +1697,16 @@ func TestDynamoGraphDeploymentValidator_Validate(t *testing.T) {
...
 			var (
 				warnings []string
 				err      error
 			)
 			if tt.oldDeployment == nil {
-				warnings, err = handler.ValidateCreate(ctx, currentBeta)
+				warnings, err = validator.ValidateCreate(ctx, currentBeta)
 			} else {
-				warnings, err = handler.ValidateUpdate(ctx, oldBeta, currentBeta)
+				warnings, err = validator.ValidateUpdate(ctx, oldBeta, currentBeta)
```
sttts: why not handler?

deploy/operator/internal/namespace_scope/lease_watcher.go
sttts: Resolved by the final design: the cluster-wide FeatureAwareValidator always invokes the underlying validator, and the Lease only supplies effective feature gates. A tenant-created Lease can no longer disable Go validation. The final cleanup is in e9b996a59c.

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
sttts: Fixed in e9b996a59c. The method is now groveEnabled(ctx); the fallback field was renamed defaultGroveEnabled to avoid a Go field/method name collision.

deploy/operator/internal/webhook/validation/dynamographdeployment_test.go
sttts: Fixed in e9b996a59c. Removed the no-op validator alias and invoke handler directly.

deploy/operator/internal/webhook/validation/dynamocheckpoint_handler.go
```diff
@@ -76,18 +78,27 @@ func (h *DynamoCheckpointHandler) ValidateDelete(ctx context.Context, obj runtim
...
 		WithRecoverPanic(true)
 	mgr.GetWebhookServer().Register(dynamoCheckpointWebhookPath, webhook)
 	return nil
 }
 
-func validateDynamoCheckpointGMSSnapshot(ckpt *nvidiacomv1alpha1.DynamoCheckpoint) error {
+func validateDynamoCheckpointGMSSnapshot(ctx context.Context, ckpt *nvidiacomv1alpha1.DynamoCheckpoint) error {
+	gmsSnapshotEnabled := os.Getenv(consts.DynamoOperatorAllowGMSSnapshotEnvVar) == "1"
+	if gates, ok := features.FromContext(ctx); ok {
+		gmsSnapshotEnabled = gates.GMSSnapshot
```
sttts: this looks risky. The getenv should run in the feature package where the flag is. This code here now will override the env all the time, no?!

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
```diff
@@ -45,9 +46,9 @@ const (
 // DynamoGraphDeploymentHandler is a handler for validating DynamoGraphDeployment resources.
 // It is a thin wrapper around DynamoGraphDeploymentValidator.
 type DynamoGraphDeploymentHandler struct {
-	mgr               manager.Manager
-	operatorPrincipal string
-	groveEnabled      bool
+	mgr                 manager.Manager
+	operatorPrincipal   string
+	defaultGroveEnabled bool
```
sttts: why is this not coming from the gate? The bools should come from the gates in all cases. The gates must be adapted to give the right values.

deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
```diff
@@ -116,3 +115,35 @@ subjects:
...
+  labels:
+  {{- include "dynamo-operator.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - validatingwebhookconfigurations
+  verbs:
+  - get
+  - list
```
sttts: Why is this new?

deploy/operator/internal/webhook/validation/dynamocheckpoint_handler.go
sttts: Fixed in d4ef61bc5b. The environment variable and os.Getenv now live in internal/features via features.FromEnvironment(). Admission reads only the effective Gates from context, and the separate ValidateGMSSnapshotGateEnabled boolean path was removed; checkpoint validation now accepts Gates in every caller.

deploy/operator/internal/webhook/validation/dynamographdeployment_handler.go
sttts: Fixed in d4ef61bc5b. Admission now has one feature path: FeatureAwareValidator attaches the effective Gates, handlers require that context, and DGD/DGDR validators receive Gates directly. The Grove fallback field and DGDR cluster-wide/GPU booleans were removed.

deploy/helm/charts/platform/components/operator/templates/webhook-rbac.yaml
sttts: The namespaced release cannot know the cluster-wide release's ValidatingWebhookConfiguration name, so startup discovers the compatible global configuration by capability label and requires list. The adjacent get permission was unused; removed it in 250c7fb94a so this reader now has list only.

deploy/operator/internal/namespace_scope/validation_capability_watcher.go
```diff
@@ -0,0 +1,60 @@
...
+	for idx := range configurations.Items {
+		if isClusterWideAdmissionCapability(&configurations.Items[idx]) {
+			return nil
+		}
+	}
+	return errors.New("no cluster-wide Dynamo validating webhook supports namespaced admission feature gates")
+}
+
+func isClusterWideAdmissionCapability(configuration *admissionregistrationv1.ValidatingWebhookConfiguration) bool {
+	if configuration.Labels[partOfLabel] != partOfValue ||
```
julienmancuso: this is only done once at startup, and this is static label helm metadata.
There is an edge case when the cluster wide operator is rolled back to old implementation or deleted. in this case, the label disappears, but the namespaced process keeps renewing its lease.
The oldglobal LeaseAwareValidator skips that namespace, leaving no Go validation.

There is also an upgrade window where Helm can apply the label while the old pod is still serving. 
But this would need a live, continuously monitored capability handshake which would release the lease

deploy/operator/cmd/main.go
```diff
@@ -650,41 +657,47 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := registerWebhookHandlers(mgr, operatorCfg, runtimeConfig, operatorVersion); err != nil {
-		setupLog.Error(err, "failed to register webhooks")
-		os.Exit(1)
+	if restrictedNamespace == "" {
```
julienmancuso: Once namespaced instances stop registering validation here, their DGD updates are handled by the global validator, whose operatorPrincipal identifies only the cluster-wide operator SA. The namespaced DynamoGraphDeploymentScalingAdapter controller updates DGD replicas using its local operator SA, so CanModifyDGDReplicas rejects the update as an unauthorized direct replica change. This leaves DGDSA reconciliation retrying indefinitely. Global validation needs a trusted, namespace-aware way to recognize the active namespaced operator principal, or the scaling update needs another authorization mechanism.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -126,7 +124,7 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeployment(
 		hasIntraPodFailover(&dgd.Spec),
 	)...)
 
-	grovePathway, grovePathwayRequirement := grovePathwayForDynamoGraphDeployment(v.groveEnabled, dgd)
+	grovePathway, grovePathwayRequirement := grovePathwayForDynamoGraphDeployment(v.gates.Grove, dgd)
```
julienmancuso: Scenario:
- cluster-wide operator: grove enabled
- namespaced operator: grove explicitly disabled
- user creates a normal DGD without minAvailable

The global defaulter uses the global Grove value and injects minAvailable: 1. The global validator then resolves the namespace-local Grove value to false and rejects minAvailable as grove-only...
So the mutating webhook creates the validation error itself. Removing the admission-time minAvailable default and leaving the controller’s effective default would resolve this cleanly

deploy/operator/internal/webhook/mutation/pod_checkpoint_restore_handler.go
```diff
@@ -61,9 +60,6 @@ func (h *PodCheckpointRestoreMutator) Handle(ctx context.Context, req admission.
 	if h.config == nil || !h.config.Checkpoint.Enabled {
 		return admission.Allowed("checkpoint disabled")
 	}
-	if excluded := internalwebhook.GetExcludedNamespaces(); excluded != nil && excluded.Contains(req.Namespace) {
-		return admission.Allowed("namespace excluded")
```
julienmancuso, line 65: another feature enablement mismatch scenario where checkpoint is disabled in cluster-wide but enabled in namespaced mode : the namespace controller creates restore-candidate Pods, while the global mutator returns “checkpoint disabled” and leaves them uninjected

deploy/operator/internal/namespace_scope/lease_manager.go
```diff
@@ -195,21 +214,25 @@ func (lm *LeaseManager) createOrUpdateLease(ctx context.Context) error {
...
-		lm.logger.Info("Created namespace scope marker lease")
+		lm.logger.Info("Created namespace reconciliation lease")
 		return nil
 	}
 
 	// Lease exists, update it
 	existingLease.Spec.HolderIdentity = &lm.holderIdentity
 	existingLease.Spec.LeaseDurationSeconds = &leaseDurationSeconds
 	existingLease.Spec.RenewTime = &now
+	if existingLease.Annotations == nil {
```
julienmancuso, line 234: 2 namespaced operator would blindly overwrite the same lease and start reconciling.
deleting 1 of them would also delete the shared lease allowing the cluster wide operator to resume in parallel with the remaining namespaced operator.

deploy/operator/internal/namespace_scope/validation_capability_watcher.go
sttts: We do not need to support downgrading the cluster-wide operator to the old skip behavior while namespaced operators are running, or at least in best-effort manner; this is dev/test only. The cluster-wide upgrade must complete before starting a new namespaced install, and temporary duplicate validation for existing installs is fine. In the supported steady state global validation always runs and the Lease only supplies gates, so no live handshake is needed.

deploy/operator/internal/webhook/mutation/pod_checkpoint_restore_handler.go
sttts: I have added the same gate logic to defaulting. I think it makes it even simple because uniform, less surprises, no effort.

deploy/operator/internal/namespace_scope/validation_capability_watcher.go
sttts: Removed the capability label and watcher. Global validation now always runs; the Lease only supplies per-namespace gates. Downgrade to the old skip behavior is outside the supported dev/test setup.

deploy/operator/cmd/main.go
sttts: Fixed for 1.3: the active namespaced operator principal is published on the scope Lease and accepted by global DGD validation. A synthetic RBAC subresource can replace this in a follow-up.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
sttts: Fixed: defaulting now resolves the same namespace gates from the Lease as validation, so a Grove-disabled namespace does not receive minAvailable.

deploy/operator/internal/webhook/mutation/pod_checkpoint_restore_handler.go
sttts: Fixed: checkpoint Pod mutation now uses the same namespace-aware feature resolver. There is no separate global checkpoint boolean in this path.

deploy/operator/internal/namespace_scope/lease_manager.go
sttts: Fixed for the supported setup: the scope Lease manager runs only on the controller-runtime leader and no longer deletes the Lease on shutdown. A successor renews it or it expires. Multiple Helm releases in one namespace remain unsupported.

deploy/operator/internal/namespace_scope/lease_manager_runnable.go
```diff
@@ -0,0 +1,39 @@
...
+
+package namespace_scope
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type leaderElectedLeaseManager struct {
```
sttts: is this new? Why is this not yet necessary on main?

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler_test.go
```diff
@@ -44,7 +44,10 @@ func admissionCtx(op admissionv1.Operation, kind schema.GroupVersionKind) contex
 			},
 		},
 	})
-	return features.WithGate(ctx, features.Defaults())
+}
+
+func newDGDDefaulterForTest(operatorVersion string, groveEnabled bool) *DGDDefaulter {
```
sttts: useless function

deploy/operator/internal/webhook/common.go
```diff
@@ -190,6 +204,14 @@ func CanModifyDGDReplicas(operatorPrincipal string, userInfo authenticationv1.Us
 	}
 
 	parts := strings.Split(username, ":")
+	if len(parts) == 4 && namespaceOperatorPrincipal != nil {
+		if principal, found := namespaceOperatorPrincipal(parts[2]); found && principal == username {
```
julienmancuso, line 208: this looks up the lease using the service account namespace from the username.
When namespaceRestriction.targetNamespace differs from the Helm release namespace, the lease is in the target namespace but the service account is not

deploy/operator/cmd/main.go
```diff
@@ -526,44 +534,34 @@ func main() {
...
-	if err != nil {
-		setupLog.Error(err, "unable to create CA bundle injector")
-		os.Exit(1)
-	}
-	if operatorCfg.Server.Webhook.CertProvisionMode == configv1alpha1.CertProvisionModeAuto {
-		if isClusterWide {
-			err = caInjector.InjectAll(mainCtx)
-		} else {
-			err = caInjector.InjectAdmission(mainCtx)
+	if isClusterWide {
```
julienmancuso, line 537: P1 from codex :

[P1] Global checkpoint mutation ignores namespace-local checkpoint configuration.
The new global registration resolves only feature gates before constructing handlers with the cluster-wide [operatorCfg](https://github.com/ai-dynamo/dynamo/blob/b46814b2d261c4f8b8602d465f2fd00698b4760a/deploy/operator/cmd/main.go#L539-L541). The checkpoint mutator still reads h.config.Checkpoint.Storage and EffectiveSeccompProfile() from that global configuration. A namespaced release enabling checkpoint with PVC storage or different seccomp settings therefore runs mutation with the wrong configuration, potentially falling back to DaemonSet storage and admitting the restore Pod unchanged. The Lease snapshot must include the admission-relevant checkpoint configuration, or mutation must otherwise resolve it per namespace.

deploy/operator/cmd/main.go
```diff
@@ -526,44 +534,34 @@ func main() {
...
+		internalwebhook.SetExcludedNamespaces(nil)
+		internalwebhook.SetFeatureResolver(features.NewResolver(gates, leaseWatcher.AdmissionGateSnapshot))
+		internalwebhook.SetNamespaceOperatorPrincipalResolver(leaseWatcher.OperatorPrincipal)
+		if err := registerWebhookHandlers(mgr, operatorCfg, operatorVersion, operatorPrincipal, gates); err != nil {
+			setupLog.Error(err, "failed to register webhooks")
+			os.Exit(1)
 		}
+
+		caInjector, err := internalcert.NewCABundleInjector(directClient, operatorCfg)
```
atchernych: nit, non-blocking: the scope check that protects the shared CRD is entirely positional — it's this `if isClusterWide` block. `ensureCRDConversionCA` / `patchCRDConversionCA` in `internal/cert/cert.go` still don't know their own scope, so nothing at the callee stops a future caller from reaching the CRD patch from a namespace-restricted process.

Worth noting the same file already does the callee-side check for the admission path: `webhookLabels()` filters the VWC/MWC list by `operatorNamespaceLabel: i.namespace`, so those writes are ownership-scoped regardless of who calls them. The CRD path is the one mutation there that targets purely by name.

Four lines at the top of `ensureCRDConversionCA` would make it symmetric:

```go
if i.cfg.Namespace.Restricted != "" {
    return fmt.Errorf("refusing to patch cluster-scoped CRD conversion CA: operator is namespace-restricted to %q", i.cfg.Namespace.Restricted)
}
```

Startup path only, no runtime cost, and it makes the guarantee unit-testable in `cert_test.go` rather than only in `main.go`, which has no tests. The Helm RBAC gate already covers this in practice — this is just so a regression fails with a clear message instead of an RBAC denial at an arbitrary moment.

PR #11492, opened 2026-07-09
fix(vllm-omni): encode i2v video as VP9/webm

nv-tusharma:
## Summary

Encode vLLM-Omni image-to-video output as **VP9 in a webm container** instead of MP4/H.264.

VP9/webm is already provided by the in-tree ffmpeg build, so this keeps the runtime image minimal — no extra packages are needed. And because `libvpx-vp9` is a software encoder, it also works on GPUs without NVENC (e.g. A100/H100), not just the L4 CI runner.

## Changes

- **`omni/output_formatter.py`** — `_encode_video` defaults to and only accepts `webm`, routing frames through `encode_to_video_bytes` (`libvpx-vp9`) instead of `diffusers.export_to_video` (`libx264`, which isn't in the runtime ffmpeg). Float pipeline frames are coerced to the uint8 imageio requires.
- **`common/utils/video_utils.py`** — `normalize_video_frames` returns an `ndarray` so the dtype coercion works.
- **`deps/requirements.vllm.txt`** — comment updated to reflect imageio VP9/webm encoding against the source-built ffmpeg.
- **`tests/omni/test_output_formatter.py`** — unit tests updated to webm (coercion + reject-non-webm; existing output-format tests flipped to webm).

No Dockerfile change is required — the runtime ffmpeg already ships `libvpx-vp9` and the webm muxer.

## Validation (RTX 6000 Ada, runtime image)

- Real 832×480 encode produces valid VP9: `Video: vp9 (Profile 0), yuv420p, 832x480, 16 fps` in a webm container.
- Unit tests: 7/7 pass.
- `py_compile` + `ruff`: clean.

Closes [🎫 OPS-7447](https://linear.app/nvidia/issue/OPS-7447/ci-failure-post-merge-ci-pipeline-vllm-runtime-test-cuda130-amd64)

Closes OPS-7608


components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -141,11 +142,16 @@ async def _encode_video(
...
-                video_bytes = tmp.read()
+            # Encode via the in-tree helper (codec="h264_nvenc") instead of
+            # diffusers.export_to_video: the latter lets imageio default to
+            # codec="libx264", which the vllm-runtime image deliberately purges
+            # for LGPL compliance (see container/templates/vllm_runtime.Dockerfile).
+            # The image ships an ffmpeg built with only h264_nvenc/libvpx_vp9
+            # encoders, so the libx264 default fails with "No valid H.264 encoder
+            # was found", breaking test_omni_serve_deployment[omni_i2v].
+            video_bytes = await asyncio.to_thread(
+                encode_to_video_bytes, frame_list, fps, output_format
```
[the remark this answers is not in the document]
nv-tusharma: The CUDA omni_i2v job runs on NVIDIA L4 (g6, pinned in the prod-tester-amd-gpu-v2 NodePool), which has NVENC — not A100/H100. h264_nvenc is the vllm-runtime image's only built-in H.264 encoder and already the codec used by trtllm and sglang video on this same runner, so this aligns omni with existing usage. The omni_i2v xpu variant's mp4 path is a pre-existing limitation (the nvenc ffmpeg is cuda-gated in vllm_runtime.Dockerfile) and is not regressed by this change.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -141,11 +142,16 @@ async def _encode_video(
...
+            )
```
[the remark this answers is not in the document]
nv-tusharma: The CUDA omni_i2v job runs on NVIDIA L4 (g6, pinned in the prod-tester-amd-gpu-v2 NodePool), which has NVENC — not A100/H100. h264_nvenc is the vllm-runtime image's only built-in H.264 encoder and already the codec used by trtllm and sglang video on this same runner, so this aligns omni with existing usage. The omni_i2v xpu variant's mp4 path is a pre-existing limitation (the nvenc ffmpeg is cuda-gated in vllm_runtime.Dockerfile) and is not regressed by this change.

ayushag-nv (approved):

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -142,11 +143,12 @@ async def _encode_video(
...
-                suffix=f".{output_format}", delete=True
-            ) as tmp:
-                await asyncio.to_thread(export_to_video, frame_list, tmp.name, fps)
-                video_bytes = tmp.read()
+            # Use the in-tree encoder helper (h264_nvenc, numpy frames in) instead
+            # of diffusers.export_to_video, whose libx264 default is not built into
+            # the vllm-runtime image's ffmpeg. See video_utils.encode_to_video_bytes.
+            video_bytes = await asyncio.to_thread(
+                encode_to_video_bytes, np.asarray(frame_list), fps, output_format
+            )
```
dmitry-tokarev-nv: `encode_to_video_bytes` -> `imageio.imwrite` requires **uint8 [0,255]** frames, whereas the `export_to_video` you're replacing tolerated float `[0,1]` (and PIL) and scaled internally. `normalize_video_frames` does no dtype coercion, so if the Wan stage emits float frames this will fail or distort. Not caught by the unit tests (both `normalize_video_frames` and `encode_to_video_bytes` are mocked) — only by the GPU post-merge run.

Optional hardening below restores the float->uint8 tolerance `export_to_video` gave you (no-op when frames are already uint8). Only needed if the pipeline can emit float; confirm the stage output dtype first.

```suggestion
            # Use the in-tree encoder helper (h264_nvenc, numpy frames in) instead
            # of diffusers.export_to_video, whose libx264 default is not built into
            # the vllm-runtime image's ffmpeg. See video_utils.encode_to_video_bytes.
            frames_np = np.asarray(frame_list)
            if np.issubdtype(frames_np.dtype, np.floating):
                # export_to_video accepted float [0,1]; imageio/h264_nvenc need uint8.
                frames_np = (frames_np * 255.0).round().clip(0, 255).astype(np.uint8)
            video_bytes = await asyncio.to_thread(
                encode_to_video_bytes, frames_np, fps, output_format
            )
```

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -142,11 +143,12 @@ async def _encode_video(
         try:
             start_time = time.time()
             frame_list = normalize_video_frames(images)
```
dmitry-tokarev-nv: Minor (not blocking): `normalize_video_frames` returns `list(frames)`, then this stacks it back with `np.asarray(...)`, forcing a full contiguous copy of the clip (~100 MB+ at 97x832x480x3) even when the pipeline already handed over one ndarray. omni is now the only caller of `normalize_video_frames` (`common/utils/video_utils.py:65`), so having it return the ndarray directly would drop the array->list->array round-trip. Its docstring ("for export_to_video") is now stale too.

dmitry-tokarev-nv (commented):
Reviewed the diff plus the surrounding paths (`video_utils.encode_to_video_bytes`, the trtllm video handler, `vllm_runtime.Dockerfile` / `requirements.vllm.txt`, and the omni output-formatter test).

**Overall:** clean, minimal fix that aligns omni i2v with the encoder convention the rest of the codebase already uses (`encode_to_mp4`, trtllm `video_handler.py`), and swapping the temp-file dance for in-memory bytes is a nice simplification. It's a strict improvement — `libx264` was never in the image, so nothing regresses. Non-blocking notes below; the real gate is the `vllm-runtime / Test cuda13.0` post-merge run.

### Correctness
- Worth confirming before relying on this in production: `encode_to_video_bytes` -> `imageio.imwrite` requires **uint8 [0,255]** frames (cf. `test_video_utils.make_frames` and the trtllm caller's documented `(B,T,H,W,C) uint8` contract). The removed `export_to_video` tolerated float `[0,1]`/PIL and scaled internally; `imageio` does not, and `normalize_video_frames` does no dtype coercion. If the Wan diffusion stage can emit float frames, mp4 encoding will fail or distort. Not exercised by any CPU test (both `normalize_video_frames` and `encode_to_video_bytes` are mocked), so it rests entirely on the GPU post-merge e2e. Inline suggestion added.

### On the CodeRabbit NVENC comment
- Accurate but mis-scoped as a blocker. A100/H100-class datacenter GPUs genuinely ship no NVENC engine, so `h264_nvenc` fails there — but that's a pre-existing, codebase-wide assumption (`encode_to_mp4`, `encode_to_video_bytes`, trtllm handler), not introduced here, and not a regression (libx264 was never available, so omni i2v mp4 was already broken on every GPU). This fix makes it work on NVENC-capable GPUs (T4/A10/L4/L40S/RTX — presumably the CI runner and the intended target). If A100/H100 i2v ever becomes a requirement, that's a separate issue for all video paths; note `encode_to_video_bytes` already has a software/universal `libvpx-vp9` path for webm, but `_encode_video` hard-rejects non-mp4, so that would be the natural portable fallback.

### Conventions / minor
- `requirements.vllm.txt` comment fix is correct and welcome — the old "no `IMAGEIO_FFMPEG_EXE` needed" text was wrong post-change; it IS set (`vllm_runtime.Dockerfile:259`) and now IS used.
- FYI, pre-existing / out of scope: the same outdated "no `IMAGEIO_FFMPEG_EXE` needed" narrative also lives in the checked-in rendered Dockerfiles (`vllm-runtime-cuda13.0-amd64-rendered.Dockerfile:786`, `-xpu-...:720`), but those are already drifted from the template (they lack the whole `IMAGEIO_FFMPEG_EXE` block it has) — separate cleanup.

### Tests
- Mock update is correct and patch coverage is 100%, but the real encode / `np.asarray` / dtype path is GPU-only — CPU tests only validate response packaging (url/b64/status).

components/src/dynamo/vllm/omni/output_formatter.py
nv-tusharma: Applied. Confirmed the omni Wan stage carries no in-repo dtype guarantee (unlike trtllm's uint8 contract), so added the float->uint8 coercion to restore export_to_video's old behavior, plus a CPU test covering it.

components/src/dynamo/vllm/omni/output_formatter.py
nv-tusharma: Done — normalize_video_frames now returns the ndarray directly (dropped the list<->array round-trip) and the docstring is updated.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -141,12 +142,15 @@ async def _encode_video(
...
-            with tempfile.NamedTemporaryFile(
-                suffix=f".{output_format}", delete=True
-            ) as tmp:
-                await asyncio.to_thread(export_to_video, frame_list, tmp.name, fps)
-                video_bytes = tmp.read()
+            frames = normalize_video_frames(images)
+            # Encode via encode_to_video_bytes (h264_nvenc); diffusers.export_to_video
+            # defaults to libx264, absent from the vllm-runtime image's ffmpeg. It also
+            # scaled float frames internally, so coerce to the uint8 imageio requires.
+            if np.issubdtype(frames.dtype, np.floating):
```
dmitry-tokarev-nv, line 150: Heads-up — this new `frames.dtype` access will fail the 3 `TestDiffusionFormatterVideoOutputFormat` tests as currently written. Their shared `_patches()` helper (test file ~L562-566) still:

1. patches `dynamo.vllm.omni.output_formatter.export_to_video` — which this PR removes from the module, so `patch(...)` raises `AttributeError` at `with`-entry (no `create=True`); and
2. mocks `normalize_video_frames` -> `[MagicMock()]` (a list) — so `frames.dtype` here raises `AttributeError`, is caught by the `except`, and returns `status="failed"` / `data=[]`, making the tests' `result["data"][0]` assertions `IndexError`.

Both live in `_patches()`; fix:

```python
_patch(
    "dynamo.vllm.omni.output_formatter.normalize_video_frames",
    return_value=np.zeros((1, 2, 2, 3), dtype=np.uint8),  # ndarray, not a list
),
_patch(
    "dynamo.vllm.omni.output_formatter.encode_to_video_bytes",
    return_value=b"fake-mp4-bytes",
),
```

(+ a module-level `import numpy as np` — currently only imported locally in the newer test.) These carry the module `unit`/`pre_merge` marks, so they go red on the next CI run. Re-applying only the lost `export_to_video` -> `encode_to_video_bytes` rename isn't enough — issue #2 needs the ndarray mock too.

dmitry-tokarev-nv (commented):
Before this lands, one direction change: rather than standardizing the omni video path on `h264_nvenc` baked into the runtime image, let's prefer the vp9/`webm` path.

`encode_to_video_bytes` already supports `libvpx-vp9`. It's a software encoder, so it runs across all GPU SKUs — including the NVENC-less ones raised above — and keeps the image's built-in encoder surface small. `h264_nvenc` both ties this feature to specific hardware and grows what the base image ships by default, which we'd rather not commit to here.

Concretely:
- Default omni video output to `webm`/`libvpx-vp9`, and relax the mp4-only guard in `_encode_video` so vp9 is a first-class path.
- If an mp4/h264 path is ever genuinely required for a deployment, prefer bringing it in at runtime rather than baking it into the image.

Net: the default image stays on the more portable, lower-footprint encoder and isn't pinned to the nvenc path.

components/src/dynamo/vllm/omni/output_formatter.py
nv-tusharma: Fixed as suggested — `_patches()` now returns an ndarray from `normalize_video_frames` and patches `encode_to_video_bytes`. Also moved the codec to VP9/webm (already in the in-tree ffmpeg, so no extra base-image packages, and it works on GPUs without NVENC); the 3 tests plus new webm coverage pass in the runtime image.

dmitry-tokarev-nv (approved):
LGTM. Fix the pre-commit issue.
FWIW this overlaps a bit with other PR that touch codecs #11836

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
```diff
@@ -234,6 +234,43 @@ async def test_error_returns_failed_status(self):
...
+        assert chunk["status"] == "completed"
+        assert chunk["data"][0]["output_format"] == "webm"
+
+    @pytest.mark.asyncio
+    async def test_non_webm_output_format_rejected(self):
+        f = _make_diffusion_formatter()
+        with pytest.raises(ValueError, match="webm"):
+            await f._encode_video(
+                [MagicMock()], "req-1", fps=16, output_format="mp4"
+            )
```
dmitry-tokarev-nv: `pre-commit` hunk 1 of 2 — `black` collapses this call to a single line (77 chars + indent, fits in 88):

```suggestion
            await f._encode_video([MagicMock()], "req-1", fps=16, output_format="mp4")
```

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
```diff
@@ -523,20 +560,24 @@ async def test_audio_data_carries_output_format_url_path(self):
 
 
 class TestDiffusionFormatterVideoOutputFormat:
-    """_encode_video always sets VideoData.output_format='mp4'."""
+    """_encode_video always sets VideoData.output_format='webm'."""
 
     def _patches(self):
+        import numpy as np
         from unittest.mock import patch as _patch
```
dmitry-tokarev-nv: `pre-commit` hunk 2 of 2 — `isort` wants the third-party import after the stdlib one, separated by a blank line:

```suggestion
        from unittest.mock import patch as _patch

        import numpy as np
```

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
```diff
@@ -234,6 +234,43 @@ async def test_error_returns_failed_status(self):
...
+        import numpy as np
+
+        f = _make_diffusion_formatter()
+        float_frames = np.zeros((4, 8, 8, 3), dtype=np.float32)
+        captured = {}
+
+        def fake_encode(frames, fps, output_format):
+            captured["dtype"] = frames.dtype
+            captured["output_format"] = output_format
+            return b"webm-bytes"
```
dmitry-tokarev-nv: The fixture is all-zeros, so this test can't distinguish `frames * 255.0` from no scaling at all — `0.0` maps to `0` either way. It pins the dtype but not the conversion it exists to protect. A mid-range value makes it a real regression test:

```suggestion
        float_frames = np.full((4, 8, 8, 3), 0.5, dtype=np.float32)
        captured = {}

        def fake_encode(frames, fps, output_format):
            captured["dtype"] = frames.dtype
            captured["max"] = int(frames.max())
            captured["output_format"] = output_format
            return b"webm-bytes"
```

…then add `assert captured["max"] == 128` next to the existing dtype assertion (0.5 × 255 = 127.5 → `.round()` → 128, which also pins the round-vs-truncate choice).

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -129,24 +130,28 @@ async def _encode_video(
...
-            frame_list = normalize_video_frames(images)
-            with tempfile.NamedTemporaryFile(
-                suffix=f".{output_format}", delete=True
-            ) as tmp:
-                await asyncio.to_thread(export_to_video, frame_list, tmp.name, fps)
-                video_bytes = tmp.read()
+            frames = normalize_video_frames(images)
+            # imageio requires uint8; the diffusion pipeline emits float frames.
+            if np.issubdtype(frames.dtype, np.floating):
+                frames = (frames * 255.0).round().clip(0, 255).astype(np.uint8)
```
dmitry-tokarev-nv: Non-blocking, same class as the array round-trip you already dropped: this chain allocates three full-size float temporaries (`* 255.0`, `.round()`, `.clip()`) before the uint8 cast. At the default 832×480×97×3 float32 that's ~460 MB each — roughly 1.4 GB of transient allocation in the same process as the diffusion model. One copy is unavoidable (`normalize_video_frames` can hand back the pipeline's own array, so scaling in place would corrupt upstream state), but the rest can run in place on that copy:

```suggestion
            # imageio requires uint8; the diffusion pipeline emits float frames.
            if np.issubdtype(frames.dtype, np.floating):
                # One float copy (never in place — normalize_video_frames may
                # return the pipeline's own array), then round/clip on that copy.
                scaled = frames * 255.0
                np.rint(scaled, out=scaled)
                np.clip(scaled, 0, 255, out=scaled)
                frames = scaled.astype(np.uint8)
```

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -129,24 +130,28 @@ async def _encode_video(
...
+        # codec packages in the base image, and as a software encoder it also works
+        # on GPUs without NVENC (A100/H100).
+        output_format = output_format or "webm"
         response_format = response_format or "url"
         if response_format not in ("url", "b64_json"):
             raise ValueError(
                 f"Unsupported response_format: {response_format!r}; expected 'url' or 'b64_json'"
             )
-        if output_format != "mp4":
+        if output_format != "webm":
```
dmitry-tokarev-nv: Worth a conscious call, not a blocker: this `raise` sits outside the `try`, so a client that explicitly sends `output_format: "mp4"` now gets an exception propagating out of `format()`, where before the narrowing it produced the ordinary `status="failed"` response envelope (mp4 reached the encoder and failed there). Nothing regresses in practice — mp4 never actually encoded, since libx264 isn't in the runtime ffmpeg — but the two format checks are the only failure path in this method that doesn't come back as a normal response. Moving them inside the `try` would make a bad request behave like every other error here.

dmitry-tokarev-nv (commented):
Re-reviewed after the responses — no new commits since my approval (`4ef5b99b`, Jul 20), so this pass covers the three reply threads plus what's still outstanding.

**All three of my earlier comments are correctly applied** — resolving those threads:

- float→uint8 coercion (`output_formatter.py:149-151`): the implementation is actually stricter than the `export_to_video` behavior it replaces (`.round().clip(0, 255)` vs a bare `.astype(uint8)`), so no wrap on values slightly above 1.0.
- `normalize_video_frames` now returns the ndarray directly, with the annotation and docstring updated; omni is still the only caller, and trtllm's `frames_to_numpy` path is untouched.
- `_patches()` fixed, and the 4-tuple unpack in the three tests still matches.

I also verified the VP9/webm direction end to end: the in-tree ffmpeg is built with `--enable-encoder=h264_nvenc,libvpx_vp9` and `--enable-muxer=mov,mp4,matroska,webm` (`container/templates/wheel_builder.Dockerfile:348-350`), and imageio 2.37's plugin priority for `.webm` is `['FFMPEG', 'pyav']` — so it resolves through `IMAGEIO_FFMPEG_EXE=/usr/local/bin/ffmpeg` (`container/templates/vllm_runtime.Dockerfile:259`), i.e. the mechanism the new `requirements.vllm.txt` comment describes is the one actually in play, not PyAV's bundled copy. `lib/llm/src/protocols/openai/videos.rs:47,62` types `output_format` as a plain `String` on both request and response, so `"webm"` round-trips with no Rust-side change needed.

**Blocking:** `pre-commit` is still red → `pre-merge-status-check` red. Two mechanical hunks, both flagged inline below; `pre-commit run --all-files` fixes them.

**Doc/reference updates this needs** (all outside the diff, so no inline anchor):

- `docs/backends/vllm/vllm-omni.md:128` — the `/v1/videos` sample response still shows `file:///tmp/dynamo_media/videos/req-abc123.mp4`; it's `.webm` now.
- `docs/backends/vllm/vllm-omni.md:289` — same for the `--media-output-http-url` example (`videos/req-id.mp4`).
- `components/src/dynamo/common/protocols/video_protocol.py:81` — the hint reads "e.g. 'mp4', 'mjpeg'"; worth noting that vLLM-Omni accepts only `webm` (the model is shared with trtllm, which still emits mp4).
- `components/src/dynamo/vllm/tests/omni/test_omni_stage_router.py:445` still round-trips `output_format: "mp4"` through the router. It passes (the router doesn't encode), but it's now the one place in-tree asserting a value the formatter rejects.

**Coverage note:** `omni_i2v` is `gpu_1` + `post_merge` (`tests/serve/test_vllm_omni.py:145-146`), so no pre-merge job exercises the real encode — the RTX 6000 Ada run in the description is the only functional evidence until this lands. `VideoGenerationPayload.response_handler` does assert `status == "completed"` (`tests/utils/payloads.py:2033`), so a broken encode fails loudly rather than passing silently; nothing checks the bytes are actually a webm, though — an `endswith(".webm")` plus a non-trivial size check would close that cheaply.

Approval stands — the only must-fix is pre-commit.

components/src/dynamo/vllm/omni/output_formatter.py
dmitry-tokarev-nv: Confirmed on `4ef5b99b` — `_patches()` now returns an ndarray from `normalize_video_frames` and patches `encode_to_video_bytes`, and the 4-tuple unpack in all three tests still matches. Resolving.

components/src/dynamo/vllm/omni/output_formatter.py
nv-tusharma: Applied — one float copy, then `np.rint`/`np.clip` with `out=` and the uint8 cast; the transient temporaries are gone.

components/src/dynamo/vllm/omni/output_formatter.py
nv-tusharma: Done — both validations now sit inside the `try`, so a bad request returns the normal `status="failed"` envelope instead of propagating.

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
nv-tusharma: Applied — fixture is `0.5` now with `assert max == 128`, so it pins the round-to-128 scaling, not just the dtype.

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
nv-tusharma: Fixed — stdlib import first, blank line, then numpy.

components/src/dynamo/vllm/tests/omni/test_output_formatter.py
nv-tusharma: This test now asserts the `status="failed"` envelope (checks moved into the try), and black 23.1.0 leaves the line as-is.

dmitry-tokarev-nv (commented):
Re-reviewed `888f91e5` (`bf13153d` + the main merge). **All five inline comments from the last pass are applied and verified — resolving those threads.**

- **In-place scaling** (`output_formatter.py:150-156`): landed as suggested, and critically it keeps the scale off the caller's array — `frames * 255.0` makes the single copy, then `np.rint`/`np.clip` reuse it via `out=`. Since `normalize_video_frames` can hand back the pipeline's own buffer, that distinction was the part that mattered.
- **Validations moved inside the `try`** (`:138-146`): a bad `output_format`/`response_format` now comes back as the standard `status="failed"` envelope. I checked the `except` branch doesn't reference `start_time` (bound at `:147`, after the checks), so there's no `UnboundLocalError` on that path, and no other test expected the old raise.
- **Coercion test** now fills `0.5` and asserts `max == 128`, so it pins round-to-nearest rather than just the dtype.
- **isort** ordering fixed. The `black` item resolved itself once the test became a failed-envelope assertion — `chunk = await f._encode_video(...)` is 90 columns, so it correctly stays exploded.
- **`pre-commit` is green**, and `pre-merge-status-check` with it. That was the only blocker.

`vllm-runtime / Test` (amd64 + arm64), Multi-GPU, and all four Compliance audits are green. `deploy-operator` / `deploy-status-check` are red, but this PR touches no operator, chart, or deploy code — that reads as the usual infra-side failure class, worth a re-run rather than a code change.

**One item from my last review body is still open**, and its path moved under you when main's docs reorg came in with the merge, so here's the current location:

`docs/fern/features/diffusion/text-to-video/README.md:54` — inside the vLLM-Omni tab (`agg_omni_video.sh`), the sample response still advertises the old container:

> …`"data": [{"url": "file:///tmp/dynamo_media/videos/req-abc123.mp4"}]}`

That's the same `_encode_video` path this PR changes, so it should read `req-abc123.webm`. The `output.mp4` lines further down (`:349`, `:352`) are in the FastVideo tab and are correct as-is, and `docs/fern/features/diffusion/image-to-video/README.md` has no response-extension example, so nothing to change there.

Still optional, unchanged from last time: `components/src/dynamo/common/protocols/video_protocol.py:81` hints "e.g. 'mp4', 'mjpeg'", and `components/src/dynamo/vllm/tests/omni/test_omni_stage_router.py:445` round-trips `output_format: "mp4"` through the router (passes — the router doesn't encode).

Approval stands; the doc line is the only thing I'd still fold in before merge.

sttts (approved):

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -129,24 +130,33 @@ async def _encode_video(
...
+                    f"Unsupported output_format: {output_format!r}; only 'webm' is supported"
+                )
             start_time = time.time()
-            frame_list = normalize_video_frames(images)
-            with tempfile.NamedTemporaryFile(
-                suffix=f".{output_format}", delete=True
-            ) as tmp:
-                await asyncio.to_thread(export_to_video, frame_list, tmp.name, fps)
-                video_bytes = tmp.read()
+            frames = normalize_video_frames(images)
```
jh-nv, line 148: We run the normalization, scaling and some other operations on the main thread. For video data, IIUC, it could be quite some data, should we consider moving them to a thread as we do for encode_to_video_bytes() ?

nv-tusharma:
closing stale PR

PR #11503, opened 2026-07-09
perf(kv-router): share request block ownership chains

jthomson04:
## Summary

- replace per-request `Vec<(SequenceHash, Arc<()>)>` ownership with one tail `Arc` into an immutable persistent block chain
- reuse a live prompt tail directly and allocate nodes only for a missing suffix
- release only the uniquely owned suffix, stopping at the first shared node
- preserve prompt-membership events, fractional accounting, output-block behavior, and compile-time `Send + Sync` guarantees
- destroy long unique chains iteratively to avoid recursive-drop stack overflow at block size 1

## Details:

This keeps Arc-based liveness while removing per-block `Weak::upgrade`/Arc cloning on shared prompt acquisition and full-chain scans on release. Each `SequenceHash` is trusted to identify one prefix ancestry and depth, matching the KV indexer's engine-to-router contract; this PR does not add runtime recurrence validation.

The review follow-up documents that contract, makes the output-hash debug assertion side-effect-free, and covers fractional cleanup of a unique suffix above a shared prefix. Public APIs, event schemas, routing behavior, and PromptRegistry behavior are unchanged.

Start with `lib/kv-router/src/sequences/block_tracker.rs`, especially `acquire_prompt`, `release`, and `RequestBlockChain::drop`. Then inspect `lib/kv-router/src/sequences/single.rs` for the request-lifecycle integration.

## Related Issues

## Validation

- `cargo fmt --all --check`
- `cargo test -p dynamo-kv-router block_tracker --lib` — 6 passed
- `cargo test -p dynamo-kv-router --lib` — 726 passed, 2 ignored
- `cargo test -p dynamo-bench --test active_sequences_trace` — 9 passed

### Frontend profile

Matched frontend profiles were collected from baseline and candidate builds at `e7ba0adbb99c4fed9fdf77971dbc2e41cf031458`:

- frontend on CPUs 0-1; KV routing; block size 1; fastokens; 1 GiB tokenizer cache; frontend-only jemalloc
- eight SGLang FIFO mockers and AIPerf on CPUs 2-23
- AgentX/Weka trace; speedup ratio 1,000,000; concurrency 256
- 40-second frontend-only `perf record -e cycles:u -F 99 --call-graph dwarf` capture after a 40-second settle

The inclusive BlockTracker lifecycle fell from 25.36% (806/3178 samples, 95% CI 23.88%-26.90%) to 17.04% (511/2999 samples, 95% CI 15.74%-18.43%). Weighted-cycle share fell from 26.21% to 17.59%, and frontend CPU per completed request fell from 14.249 ms to 13.665 ms.

Profiled throughput was diagnostic only and remained effectively flat (14.379 vs 14.250 req/s). The mocker side saturated individual cores, so this is not presented as a maximum frontend-capacity result. Both comparison arms used the same AIPerf 0.8.0 client on the non-frontend CPU partition.

## Summary by CodeRabbit

* **New Features**
  * Improved request tracking for prompt and output blocks, with smoother handling of shared prefixes and extended sequences.
  * Added more reliable support for long-running request chains without risking stack overflow.

* **Bug Fixes**
  * Fixed block cleanup so released request data is removed more consistently.
  * Improved active-block counting for fractional/shared blocks.
  * Strengthened consistency checks to better detect invalid request state.

### DIFF
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -5,47 +5,181 @@ use dynamo_tokens::SequenceHash;
 use rustc_hash::FxHashMap;
 use std::sync::{Arc, Weak};
 
+/// One node in a persistent request block chain.
+///
+/// Sequence hashes identify their lineage, so a live tail transitively owns the
+/// entire prompt prefix through these parent references.
 #[derive(Debug)]
-pub(super) struct BlockAcquire {
-    pub(super) rc: Arc<()>,
-    pub(super) became_present_on_worker: bool,
+struct BlockNode {
+    hash: SequenceHash,
+    depth: usize,
+    parent: Option<Arc<BlockNode>>,
+}
+
+/// The single strong block owner retained by a request.
+///
+/// This type intentionally does not implement `Clone`. New request owners must
+/// be acquired through [`BlockTracker::acquire_prompt`], which also maintains
+/// the weak block index and membership transitions.
+#[derive(Debug, Default)]
+pub(super) struct RequestBlockChain {
+    tail: Option<Arc<BlockNode>>,
+    prompt_depth: usize,
+}
+
+impl RequestBlockChain {
+    fn new(tail: Option<Arc<BlockNode>>, prompt_depth: usize) -> Self {
+        Self { tail, prompt_depth }
+    }
+
+    #[cfg(any(test, debug_assertions))]
+    fn nodes_from_tail(&self) -> impl Iterator<Item = &BlockNode> {
+        std::iter::successors(self.tail.as_deref(), |node| node.parent.as_deref())
+    }
+
+    #[cfg(test)]
+    pub(super) fn prompt_hashes(&self) -> impl Iterator<Item = SequenceHash> + '_ {
+        self.nodes_from_tail()
+            .filter(|node| node.depth <= self.prompt_depth)
+            .map(|node| node.hash)
+    }
+}
+
+impl Drop for RequestBlockChain {
+    fn drop(&mut self) {
+        let Some(mut current) = self.tail.take() else {
+            return;
+        };
+
+        // A block-size-1 prompt can contain tens of thousands of nodes. Unwrap
+        // an exclusively owned suffix iteratively so dropping the parent chain
+        // cannot overflow the stack.
+        loop {
+            match Arc::try_unwrap(current) {
+                Ok(mut node) => match node.parent.take() {
+                    Some(parent) => current = parent,
+                    None => break,
+                },
+                Err(shared) => {
+                    drop(shared);
+                    break;
+                }
+            }
+        }
+    }
 }
 
 #[derive(Debug, Default)]
 pub(super) struct BlockTracker {
-    pub(super) unique_blocks: FxHashMap<SequenceHash, Weak<()>>,
-    pub(super) fractional_blocks: FxHashMap<SequenceHash, f64>,
+    unique_blocks: FxHashMap<SequenceHash, Weak<BlockNode>>,
+    fractional_blocks: FxHashMap<SequenceHash, f64>,
 }
 
 impl BlockTracker {
-    pub(super) fn touch_block(&mut self, block: &SequenceHash) -> BlockAcquire {
-        if let Some(weak) = self.unique_blocks.get(block)
-            && let Some(rc) = weak.upgrade()
-        {
-            return BlockAcquire {
-                rc,
-                became_present_on_worker: false,
-            };
+    /// Acquire one strong tail for a prompt and return the first newly present
+    /// prompt index, if any.
+    pub(super) fn acquire_prompt(
+        &mut self,
+        sequence: &[SequenceHash],
+    ) -> (RequestBlockChain, Option<usize>) {
+        let Some(&tail_hash) = sequence.last() else {
+            return (RequestBlockChain::default(), None);
+        };
+
+        if let Some(tail) = self.upgrade_live_node(tail_hash) {
+            Self::debug_assert_lineage(&tail, sequence);
+            return (RequestBlockChain::new(Some(tail), sequence.len()), None);
         }
 
-        let rc = Arc::new(());
-        self.unique_blocks.insert(*block, Arc::downgrade(&rc));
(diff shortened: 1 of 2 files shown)


lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -5,47 +5,181 @@ use dynamo_tokens::SequenceHash;
...
+    pub(super) fn acquire_prompt(
+        &mut self,
+        sequence: &[SequenceHash],
+    ) -> (RequestBlockChain, Option<usize>) {
+        let Some(&tail_hash) = sequence.last() else {
+            return (RequestBlockChain::default(), None);
+        };
+
+        if let Some(tail) = self.upgrade_live_node(tail_hash) {
+            Self::debug_assert_lineage(&tail, sequence);
```
jthomson04, line 96: The lineage invariant this fast path relies on is not enforced where client-supplied hashes enter. `normalize_hashes` / `normalize_for_reservation` (`lib/kv-router/src/services/selection/input.rs`) forward wire-provided `sequence_hashes` into `acquire_prompt` without verifying the chain, so a client submitting e.g. `[4]` while another request holds `[1,2,3,4]` gets:

- **debug builds**: this `assert!` fires — a remotely-triggerable panic on syntactically valid input to the selection HTTP service
- **release builds**: the request silently adopts the depth-4 node, transitively pinning blocks 1–3, and on release the `depth <= prompt_depth` filter means hash 4's removal is never reported to the membership delta — a permanent PromptRegistry desync

The old position-agnostic tracker handled such input gracefully (the `vec![4]` → `vec![5]` test change in `single.rs` reflects the tightened contract). Suggest validating at the boundary rather than weakening the tracker: `normalize_hashes` already receives both `block_hashes` and `sequence_hashes`, so it can cheaply verify `seq[i+1] == compute_next_sequence_hash(seq[i], block[i+1])` and reject with `BadRequest`. At minimum, document the precondition on `acquire_prompt`.

Boundary validation would also subsume the degenerate duplicate-hash-within-one-sequence case, where the suffix-build loop overwrites the earlier map entry and `release` reports the hash twice.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -5,47 +5,181 @@ use dynamo_tokens::SequenceHash;
...
-    pub(super) fn try_remove_block(&mut self, block: &SequenceHash) -> bool {
-        if let Some(weak) = self.unique_blocks.get(block)
-            && weak.strong_count() == 0
-        {
-            self.unique_blocks.remove(block);
-            self.fractional_blocks.remove(block);
-            return true;
+    /// Append a unique output block to an existing request chain.
+    pub(super) fn append_output(&mut self, chain: &mut RequestBlockChain, hash: SequenceHash) {
+        debug_assert!(
```
jthomson04, line 132: `upgrade_live_node` mutates the maps (removes stale entries on failed upgrade), so this `debug_assert!` makes debug and release builds diverge in map state. Practically harmless today, but consider a non-mutating check (e.g. a `&self` helper doing `self.unique_blocks.get(&hash).and_then(Weak::upgrade).is_none()`) to keep asserts side-effect-free.

jthomson04 (commented):
Local review of the chain-ownership redesign, traced against the old per-block refcounting semantics (full share, prefix share, suffix extension, interleaved frees, outputs on shared tails, fractional decay ordering). Behavior is equivalent on all internally-hashed paths. Verified at `fbd12cbd8`: `cargo check -p dynamo-kv-router --lib --release` compiles (the debug-only helpers are correctly cfg-gated), 725/725 lib tests pass, 9/9 `active_sequences_trace` tests pass.

Points that hold up well:

- Stop-at-first-shared-node is required for correctness in both `release` and `set_unique_suffix_fractional` — below a directly shared node, `Arc::strong_count == 1` no longer means request-unique (the node is transitively shared through the child's `parent` field) — and both walks apply it consistently.
- The dead-node set computed in `release` exactly matches what the chain `Drop` deallocates, since both observe the same strong counts under exclusive `&mut` access with no interleaving.
- The iterative `Drop` is sound: `Arc::try_unwrap` + `parent.take()` means every node drops with `parent == None`, so no recursive drop path remains.
- `add_output_block` always returning `Some` fixes a latent misreport: `multi_worker.rs` mapped `None` to `RequestNotFound`, which the old code could also return on a (vanishingly unlikely) random-hash collision.

Main pre-merge concern (in code outside this diff): the lineage invariant the design relies on — a given hash can never appear at two different depths — is structurally guaranteed for internally computed hashes (`compute_next_sequence_hash` chains parent + block hash), but `normalize_hashes` / `normalize_for_reservation` in `lib/kv-router/src/services/selection/input.rs` forward client-supplied `sequence_hashes` into `acquire_prompt` without validating the chain. Details in the inline comment on `acquire_prompt`.

Minor note: `debug_assert_lineage` walks the full sequence even on the fully-shared fast path, so debug-profile acquires stay O(n) — fine for tests, just worth knowing when profiling debug builds.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -57,50 +191,128 @@ impl BlockTracker {
...
         assert_eq!(tracker.active_blocks(), 1);
 
-        drop(second.rc);
-        assert!(tracker.try_remove_block(&1));
+        assert_eq!(tracker.release(second), vec![1]);
+        assert_eq!(tracker.active_blocks(), 0);
+    }
+
+    #[test]
+    fn release_only_removes_unique_branch_suffix() {
```
jthomson04, line 277: Coverage suggestion: no test releases a chain that has fractional entries above a shared node — e.g. A `[1,2,3]` with a fraction set on 3 while B holds `[1,2]`, then A frees. That is the one path where `release` must clean `fractional_blocks` for a strict suffix while stopping at the shared boundary. Cheap to add alongside this test.

jthomson04:
Regarding the lineage invariant raised by both automated reviews: callers are responsible for maintaining it. As with the KV indexer, BlockTracker assumes each SequenceHash uniquely identifies the complete prefix, including its ancestry and depth, and does not revalidate the chain.

jthomson04:
Closing in favor of #11508

PR #11508, opened 2026-07-09
perf(kv-router): use arena block ownership counts

jthomson04:
## Summary

- Replace per-request, per-block `Arc` ownership with a persistent request block chain backed by a typed `SlotMap` arena.
- Track request-tail and child-parent ownership with checked, non-atomic `NonZeroU32` counts. Keep request chains private and non-`Clone`, validate arena generations and hashes on every index hit, and reconstruct all ownership counts in debug/test consistency checks.
- Preserve eager removal from the hash and fractional indexes, fractional accounting, active hashes, prompt-membership events, output hashes, replica behavior, routing semantics, and public interfaces.
- Keep release iterative, including 65,536-node request chains. This is Phase 1 only; the optional stale-index/deferred-deletion optimization is not included.

This draft targets `main` directly and includes the persistent-chain change from #11503 as well as the arena conversion.

### Main to arena profile

The retained main capture was pinned at `e7ba0adbb99c4fed9fdf77971dbc2e41cf031458`; the arena capture used `0d758f56fd51d0d90db2e22d84f522ade4cf1872`. Both used the AgentX/Weka trace, frontend CPUs `0-1`, eight SGLang FIFO mockers plus AIPerf on CPUs `2-23`, KV block size 1, speedup ratio 1,000,000, concurrency 256, fastokens, a 1 GiB tokenizer cache, frontend jemalloc, and a 40-second 99 Hz DWARF user-space cycles capture after settling.

| Metric | Main | Arena | Change |
|---|---:|---:|---:|
| BlockTracker inclusive weighted cycles | 26.21% | 15.04% | -42.62% |
| BlockTracker inclusive samples | 806/3,178 (25.36%) | 438/2,955 (14.82%) | -41.56% |
| Sample-level 95% Wilson interval | 23.88%-26.90% | 13.59%-16.15% | non-overlapping |

These are single, non-contemporaneous captures. The intervals quantify perf sample uncertainty, not run-to-run variance. CPU time per request is not compared across these two captures because capture-window accounting was corrected in the newer harness.

For the contemporaneous Phase 1 Arc-chain-to-arena comparison, BlockTracker weighted cycles fell from 18.27% to 15.04% (-17.66%), frontend CPU time per capture-window completion fell from 65.215 ms to 59.175 ms (-9.26%), and peak frontend RSS fell from 4,410,300 KiB to 4,335,552 KiB (-1.69%). Diagnostic throughput was 14.368 versus 14.188 requests/s; the mocker-side partition remained limiting, so this fixed-concurrency profiled throughput is not a frontend-capacity result.

One fresh-process `active_sequences_bench` run per arm at equivalent chain length measured:

| Metric | Arc chain | Arena | Change |
|---|---:|---:|---:|
| Logical block visits/s | 98,482,783 | 148,245,979 | +50.53% |
| Free p50 | 192.7 us | 110.5 us | -42.66% |
| Free p99 | 3,016.9 us | 1,751.8 us | -41.93% |
| Peak RSS | 7,779,496 KiB | 7,596,536 KiB | -2.35% |

## Validation

- `cargo fmt --all --check`
- Focused BlockTracker tests: 16 passed
- Focused ActiveSequences tests: 11 passed
- `cargo test -p dynamo-kv-router`: 737 passed, 2 ignored
- `cargo test -p dynamo-bench --test active_sequences_trace`: 9 passed
- Randomized 10,000-step ownership-model test
- SlotMap stale-generation and wrong-hash tests
- Checked ownership-overflow test
- 65,536-node explicit-release and worker-teardown tests
- One fresh-process microbenchmark and one frontend perf capture per arm

## Summary by CodeRabbit

* **Performance & Reliability**
  * Improved request block tracking for prompt reuse and output generation.
  * Added more efficient cleanup of long request sequences.
  * Strengthened handling of shared, fractional, stale, and invalid blocks.

* **Bug Fixes**
  * Improved block ownership and lifecycle management.
  * Added safeguards against inconsistent block state and overflow conditions.

* **Tests**
  * Expanded coverage for request lifecycle behavior, concurrency, and long-chain teardown.

### DIFF
--- a/lib/kv-router/Cargo.toml
+++ b/lib/kv-router/Cargo.toml
@@ -58,6 +58,7 @@ parking_lot = { workspace = true }
 rmp-serde = { workspace = true }
 
 rustc-hash = "2.1.1"
+slotmap = "1"
 
 # standalone-indexer (optional)
 axum = { workspace = true, optional = true }
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -127,9 +127,6 @@ impl BlockTracker {
             "random output hash unexpectedly collided with a live block"
         );
 
-        self.nodes.reserve(1);
-        self.unique_blocks.reserve(1);
-
         let parent = chain.tail;
         let depth = parent.map_or(1, |node_id| {
             self.nodes
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -5,47 +5,181 @@ use dynamo_tokens::SequenceHash;
 use rustc_hash::FxHashMap;
 use std::sync::{Arc, Weak};
 
+/// One node in a persistent request block chain.
+///
+/// Sequence hashes identify their lineage, so a live tail transitively owns the
+/// entire prompt prefix through these parent references.
 #[derive(Debug)]
-pub(super) struct BlockAcquire {
-    pub(super) rc: Arc<()>,
-    pub(super) became_present_on_worker: bool,
+struct BlockNode {
+    hash: SequenceHash,
+    depth: usize,
+    parent: Option<Arc<BlockNode>>,
+}
+
+/// The single strong block owner retained by a request.
+///
+/// This type intentionally does not implement `Clone`. New request owners must
+/// be acquired through [`BlockTracker::acquire_prompt`], which also maintains
+/// the weak block index and membership transitions.
+#[derive(Debug, Default)]
+pub(super) struct RequestBlockChain {
+    tail: Option<Arc<BlockNode>>,
+    prompt_depth: usize,
+}
+
+impl RequestBlockChain {
+    fn new(tail: Option<Arc<BlockNode>>, prompt_depth: usize) -> Self {
+        Self { tail, prompt_depth }
+    }
+
+    #[cfg(any(test, debug_assertions))]
+    fn nodes_from_tail(&self) -> impl Iterator<Item = &BlockNode> {
+        std::iter::successors(self.tail.as_deref(), |node| node.parent.as_deref())
+    }
+
+    #[cfg(test)]
+    pub(super) fn prompt_hashes(&self) -> impl Iterator<Item = SequenceHash> + '_ {
+        self.nodes_from_tail()
+            .filter(|node| node.depth <= self.prompt_depth)
+            .map(|node| node.hash)
+    }
+}
+
+impl Drop for RequestBlockChain {
+    fn drop(&mut self) {
+        let Some(mut current) = self.tail.take() else {
+            return;
+        };
+
+        // A block-size-1 prompt can contain tens of thousands of nodes. Unwrap
+        // an exclusively owned suffix iteratively so dropping the parent chain
+        // cannot overflow the stack.
+        loop {
+            match Arc::try_unwrap(current) {
+                Ok(mut node) => match node.parent.take() {
+                    Some(parent) => current = parent,
+                    None => break,
+                },
+                Err(shared) => {
+                    drop(shared);
+                    break;
+                }
+            }
+        }
+    }
 }
 
 #[derive(Debug, Default)]
 pub(super) struct BlockTracker {
-    pub(super) unique_blocks: FxHashMap<SequenceHash, Weak<()>>,
-    pub(super) fractional_blocks: FxHashMap<SequenceHash, f64>,
+    unique_blocks: FxHashMap<SequenceHash, Weak<BlockNode>>,
+    fractional_blocks: FxHashMap<SequenceHash, f64>,
 }
 
 impl BlockTracker {
-    pub(super) fn touch_block(&mut self, block: &SequenceHash) -> BlockAcquire {
-        if let Some(weak) = self.unique_blocks.get(block)
-            && let Some(rc) = weak.upgrade()
-        {
-            return BlockAcquire {
-                rc,
-                became_present_on_worker: false,
-            };
+    /// Acquire one strong tail for a prompt and return the first newly present
+    /// prompt index, if any.
+    pub(super) fn acquire_prompt(
+        &mut self,
+        sequence: &[SequenceHash],
+    ) -> (RequestBlockChain, Option<usize>) {
+        let Some(&tail_hash) = sequence.last() else {
+            return (RequestBlockChain::default(), None);
+        };
+
+        if let Some(tail) = self.upgrade_live_node(tail_hash) {
+            Self::debug_assert_lineage(&tail, sequence);
+            return (RequestBlockChain::new(Some(tail), sequence.len()), None);
         }
 
-        let rc = Arc::new(());
-        self.unique_blocks.insert(*block, Arc::downgrade(&rc));
-        BlockAcquire {
-            rc,
-            became_present_on_worker: true,
+        let mut parent = None;
+        let mut first_new_idx = 0;
+
+        // Prompt liveness is prefix-closed. Search backward for the deepest
+        // live prefix, then construct only the missing suffix.
+        for idx in (0..sequence.len().saturating_sub(1)).rev() {
+            if let Some(node) = self.upgrade_live_node(sequence[idx]) {
+                Self::debug_assert_lineage(&node, &sequence[..=idx]);
+                parent = Some(node);
+                first_new_idx = idx + 1;
+                break;
+            }
         }
+
+        for (idx, &hash) in sequence[first_new_idx..].iter().enumerate() {
+            let node = Arc::new(BlockNode {
+                hash,
+                depth: first_new_idx + idx + 1,
+                parent,
+            });
+            self.unique_blocks.insert(hash, Arc::downgrade(&node));
+            parent = Some(node);
+        }
+
+        (
+            RequestBlockChain::new(parent, sequence.len()),
+            Some(first_new_idx),
+        )
     }
 
-    pub(super) fn try_remove_block(&mut self, block: &SequenceHash) -> bool {
-        if let Some(weak) = self.unique_blocks.get(block)
-            && weak.strong_count() == 0
-        {
-            self.unique_blocks.remove(block);
-            self.fractional_blocks.remove(block);
-            return true;
+    /// Append a unique output block to an existing request chain.
+    pub(super) fn append_output(&mut self, chain: &mut RequestBlockChain, hash: SequenceHash) {
+        debug_assert!(
+            self.upgrade_live_node(hash).is_none(),
+            "random output hash unexpectedly collided with a live block"
+        );
+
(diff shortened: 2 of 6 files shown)


lib/kv-router/Cargo.toml
```diff
@@ -58,6 +58,7 @@ parking_lot = { workspace = true }
 rmp-serde = { workspace = true }
 
 rustc-hash = "2.1.1"
+slotmap = "1"
```
jthomson04, line 61: **Blocking (CI):** `lib/bindings/kvbm/Cargo.lock` also locks `dynamo-kv-router` and wasn't regenerated with `slotmap`, so `rust-clippy (lib/bindings/kvbm)` fails under `--locked` (`cannot update the lock file ... because --locked was passed`). Regenerate that lockfile (e.g. `cargo check` in `lib/bindings/kvbm`) and commit it — the other red checks are fail-fast fallout from this one.

Nit: `rustc-hash` sets precedent for crate-local deps, but consider workspace-managing `slotmap` if other crates are likely to pick it up.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -3,49 +3,231 @@
...
+        (
+            RequestBlockChain::new(parent, sequence.len()),
+            Some(first_new_idx),
+        )
+    }
+
+    /// Append a unique output block to an existing request chain.
+    pub(super) fn append_output(&mut self, chain: &mut RequestBlockChain, hash: SequenceHash) {
+        debug_assert!(
+            !self.unique_blocks.contains_key(&hash),
```
jthomson04, line 136: Non-blocking, just noting the trade-off: if a random output hash ever collided with a live block (~n/2^64 per append), this only fires in debug builds; in release the index insert below replaces the live entry and the eventual `release` of the shadowed node hits the hard `"live block node is missing from the hash index"` panic. The old code degraded gracefully (shared the `Arc`). Fail-stop on a astronomically-unlikely corruption is a defensible choice — flagging so it's a conscious one.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -3,49 +3,231 @@
...
+            "random output hash unexpectedly collided with a live block"
+        );
+
+        self.nodes.reserve(1);
+        self.unique_blocks.reserve(1);
```
jthomson04: Trivial: these `reserve(1)` calls add nothing over the plain inserts below — both collections grow amortized on insert anyway.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -3,49 +3,231 @@
...
+                    .expect("request chain references a missing block node")
+                    .incoming = NonZeroU32::new(incoming - 1)
+                    .expect("shared block ownership count cannot become zero");
+                break;
+            }
+
+            let node = self
+                .nodes
+                .get(node_id)
+                .expect("request chain references a missing block node");
```
jthomson04, line 189: Optional perf/simplification in the loop this PR profiled: each removed node does three slot lookups (`get` for `incoming`, this second `get` for fields, then `remove`), and the shared-stop path does `get` + `get_mut`. A single `get_mut` up front covers both branches:

```rust
let node = self
    .nodes
    .get_mut(node_id)
    .expect("request chain references a missing block node");
let incoming = node.incoming.get();
if incoming > 1 {
    node.incoming = NonZeroU32::new(incoming - 1)
        .expect("shared block ownership count cannot become zero");
    break;
}
let (hash, depth, parent) = (node.hash, node.depth, node.parent);
```

Slotmap gets are cheap (index + version check), so feel free to skip if the codegen difference doesn't show up.

lib/kv-router/src/sequences/single.rs
```diff
@@ -272,22 +239,14 @@ impl ActiveSequences {
             return PromptMembershipDelta::default();
         };
 
-        let _ = request_state.expected_output_tokens;
+        let RequestState {
+            blocks,
+            expected_output_tokens,
+            ..
+        } = request_state;
+        let _ = expected_output_tokens;
```
jthomson04: Nit: the `..` in the destructuring already discards the remaining fields, so binding `expected_output_tokens` just to `let _` it is redundant — `let RequestState { blocks, .. } = request_state;` suffices (or keep the old one-line `let _ = request_state.expected_output_tokens;` if the intent is to document the deliberately-unused field).

jthomson04 (commented):
## Review summary

Verified the ownership algebra by hand across the key scenarios (shared full prompts, shorter/longer prefix reuse, branching, output appends transferring the tail edge, release-in-either-order, fractional marking): the `incoming` edge accounting is correct, and `release`/`set_unique_suffix_fractional` are behaviorally equivalent to the old `Arc::strong_count` scheme under the lineage-hash invariant (counts are monotonically non-increasing with depth, so single-owner blocks always form a chain suffix). The non-`Clone`, `#[must_use]`, consumed-by-value `RequestBlockChain` makes double-release unrepresentable — a real design improvement over the `Arc` scheme. Test coverage is the strongest part of the PR: the randomized differential test with full per-step edge reconstruction, adversarial stale-generation/wrong-hash/overflow panics, and 65k iterative-release/teardown tests.

**Blocking (CI):** `rust-clippy (lib/bindings/kvbm)` fails with `cannot update the lock file lib/bindings/kvbm/Cargo.lock because --locked was passed` — that lockfile also pins `dynamo-kv-router` and is missing the `slotmap` entry (root and python lockfiles were updated). The other red checks are fail-fast fallout. See inline comment on `Cargo.toml`.

**Semantic tightening to sign off on:** the lineage contract is now load-bearing. The old tracker tolerated arbitrary per-block hashes; the new one requires each `SequenceHash` to identify exactly one ancestry and depth (prefix-closed liveness, backward scan stops at the deepest live prefix). This holds in-tree since `SequenceHash` is chained via `compute_next_sequence_hash`, and it's the same trust the KV indexer already places in these hashes — the `vec![4]` → `vec![5]` test change is exactly an old contract violation being cleaned up. Worth a conscious ack from reviewers since violations now panic in debug builds and silently mis-report membership events in release builds.

`add_output_block` returning `Some` unconditionally for a live request is a fix, not a regression: the `multi_worker.rs` caller maps `None` to `RequestNotFound`, which previously conflated an (improbable) output-hash collision with a missing request.

Per `sequences/CLAUDE.md`, this rewrites the block-ownership core of the write model — PeaBrane should sign off. Benchmark numbers are included as the guardrails require.

Remaining inline comments are optional polish.

jthomson04:
A note on collisions: Output block hashes are randomly generated, so collisions are theoretically possible but negligibly unlikely. The previous implementation relied on the same uniqueness assumption and had no retry or fallible append path. This PR preserves that behavior and retains the debug assertion to catch invariant violations during development. Adding production collision handling would introduce complexity disproportionate to the practical risk.

PeaBrane (commented):
AI-assisted neutral review of the arena ownership conversion. The ownership accounting looks sound for valid lineage; the inline comments cover two collision paths that remain production hazards.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -3,49 +3,228 @@
...
+    /// that invariant and does not validate the hash recurrence in production.
+    pub(super) fn acquire_prompt(
+        &mut self,
+        sequence: &[SequenceHash],
+    ) -> (RequestBlockChain, Option<usize>) {
+        let Some(&tail_hash) = sequence.last() else {
+            return (RequestBlockChain::default(), None);
+        };
+
+        if let Some(tail) = self.live_node_id(tail_hash) {
```
PeaBrane, line 80: **AI review:** `live_node_id` keys only on `SequenceHash`, while lineage validation is disabled in release builds. Public reservation and slot-tracker paths accept supplied hashes, so `[1,9]` followed by `[2,9]` adopts the wrong chain and suppresses its store; `[H,H]` overwrites the first index entry and later panics during release. Please reject inconsistent/repeated lineage before mutation or use production identity containing depth and parent ancestry. Depth alone does not cover same-depth, different-parent aliases.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -3,49 +3,228 @@
...
+                .checked_add(1)
+                .expect("block chain depth overflowed")
+        });
+        let node_id = self.nodes.insert(BlockNode {
+            hash,
+            depth,
+            parent,
+            incoming: NonZeroU32::MIN,
+        });
+        let previous = self.unique_blocks.insert(hash, node_id);
```
PeaBrane, line 155: **AI review:** The output-collision guard is debug-only, so release builds overwrite an existing live index entry. Releasing the shadowed node later hits a hard assertion or missing-entry panic. Even though UUID collisions are unlikely, make append fallible and retry the candidate before mutating the arena.

PeaBrane (approved):
The positional lineage key makes `BlockTracker` distinguish colliding lineages, but CRTC/`PromptRegistry` membership and inverse lookups remain keyed by raw `SequenceHash`. This is an existing deficiency rather than one introduced by the arena conversion. `main` is already collision-incorrect in the opposite direction: the raw-hash tracker aliases the colliding blocks behind one refcount, so releasing one lineage may emit no removal while another colliding owner remains. That hides the lifetime transition from `PromptRegistry` and leaves stale membership.

This PR separates those ownership lifetimes correctly, but the resulting raw-hash removal is ambiguous downstream. It resolves to whichever node most recently populated the inverse-index entry and may therefore remove the wrong lineage.

In either case, the practical consequence should mostly be incorrect derived read/routing state rather than corruption of the arena’s core ownership accounting, which is why I do not view this as blocking the optimization. Could we note this existing limitation somewhere so the new collision handling is not interpreted as end-to-end collision safety?

PeaBrane (commented):
Follow-up after looking more closely at the actual collision domain: I think I overstated this concern. The inverse lookup is scoped per `(worker_id, dp_rank)`, so an accidental collision requires two distinct lineages with the same 64-bit sequence hash to be simultaneously active on the same worker. Under the birthday bound, that takes roughly 609 million live hashes for a 1% collision probability and 5.06 billion for 50%; realistic per-worker active-cache populations should be far smaller. Unless caller-controlled or adversarial hashes are explicitly in scope, the existing uniqueness assumption seems reasonable. Given that, I think it would be kosher to revert the recent collision-hardening follow-up commit(s) and keep this PR focused on the arena optimization rather than carrying complexity for a negligible risk that is not handled end-to-end elsewhere anyway.

jthomson04:
Discussed offline with @PeaBrane, as well as threads above. Given that collision risk is scoped to active blocks within a single worker, weakening collision resistance in exchange for significantly better perf is a reasonable tradeoff.

jthomson04:
/ok to test [0a67b14](https://github.com/ai-dynamo/dynamo/commit/0a67b1427758e34dc017b0f23a32b08ad33611ac)

PR #11511, opened 2026-07-10
fix(llm): parse config.json with JSON5 when deriving max context length

rmccorm4:
## Summary

- Cherry-pick #11445 to `release/1.3.0`.
- Parse Hugging Face `config.json` with JSON5 when deriving maximum context length, so unrelated non-finite literals such as `Infinity` no longer prevent model registration.
- Preserve the existing top-level → `text_config` → tokenizer fallback order and per-field error behavior.

Fixes DIS-2378

## Root cause

The max-context probe pre-parsed the entire model config with strict `serde_json`. A bare non-finite literal anywhere in the file caused parsing to fail before Dynamo could read the relevant `max_position_embeddings` field.

## Original PR

- Main PR: #11445
- Merge commit: `80c4cf56c9ee487d0da641c9fafab4053ff4d73a`
- Backport commit: `9c13be8459dc7e631216bad4ff4283cc96abc8fa`

## Test plan

- [x] `cargo test --locked -p dynamo-llm --lib --no-default-features model_card::ownership_tests::architectural_context_` (3 passed)
- [x] Source and backport patch IDs match
- [ ] Release-branch CI passes

### DIFF
--- a/lib/llm/src/model_card.rs
+++ b/lib/llm/src/model_card.rs
@@ -50,6 +50,47 @@ fn extract_hf_special_tokens(hf: &HfTokenizer) -> Vec<String> {
     out
 }
 
+/// serde `deserialize_with` that maps an explicitly-present value -- *including
+/// an explicit JSON `null`* -- to `Some`. Paired with `#[serde(default)]` (which
+/// supplies `None` only when the key is absent), this distinguishes "field
+/// missing" from "field present but null/invalid". A plain `Option<Value>` would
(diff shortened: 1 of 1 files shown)


rmccorm4:
vLLM CI failures appear to be a pre-existing failure introduced on another PR

I think this PR might've broken something, but for some reason the filters on github CI triggers are wrong and this change didn't trigger vllm tests: https://github.com/ai-dynamo/dynamo/pull/11374

@nv-nmailhot @tanmayv25

PR #11515, opened 2026-07-10
perf(kv-router): remove prompt membership hash lookup

jthomson04:
Stacked on #11508.

## Summary

- replace PromptRegistry's per-worker `SequenceHash -> Arc<Node>` mutation lookup with move-only full-path store/remove records
- walk compressed trie paths from the root, preserving compressed edges, worker cutoffs, prefix-identity splits, per-node locking, and stale-child cleanup
- reconstruct the complete prompt path during SlotMap BlockTracker release while the arena lineage is still live
- remove sequence-side `WorkerLookup`, node-local `edge_index`, stale lookup repair, and lookup diagnostics without changing replica wire events or CRTC/indexer lookups
- mutate the worker table under its write lock, then release it before request-index cleanup, trie teardown, and load removal; publish notifications after derived state is cleaned up

## Validation

The active-sequences A/B was pinned to #11508 SHA `0d758f56fd51d0d90db2e22d84f522ade4cf1872`; both arms used separate optimized binaries, jemalloc, and CPUs 0-1. One fresh-process warmup per arm was discarded, followed by ten measured runs per arm in alternating order. Every run had valid generation, exact IDs, per-worker FIFO ordering, all operations completed successfully, and empty final state. `kept_up` was intentionally not a gate for this saturated cell.

Configuration: `mooncake_trace_1000.jsonl`, block size 2, trace length factor 512, trace duplication factor 1, 8 unique workers, worker duplication factor 1, 8 operation lanes, 1,048,576 GPU blocks, 1 ms benchmark duration, seed 42.

| Metric | #11508 median | Candidate median | Change |
|---|---:|---:|---:|
| Logical block visits/s | 146.18M | 322.56M | +120.67% |
| Logical operations/s | 14.50K | 32.00K | +120.67% |
| Project-and-add p50 | 119.10 us | 55.61 us | -53.31% |
| Project-and-add p99 | 2.361 ms | 1.042 ms | -55.88% |
| Add p50 | 110.34 us | 48.22 us | -56.30% |
| Add p99 | 2.359 ms | 1.040 ms | -55.90% |
| Free p50 | 112.66 us | 47.49 us | -57.84% |
| Free p99 | 1.795 ms | 0.770 ms | -57.13% |
| Peak RSS | 8,047,824 KiB | 8,045,600 KiB | -0.03% |

The implementation was validated after rebasing onto #11508 head `20f60ce59f3924eced1706c006ac8927bd6ba198`. It is now based on #11508 head `59c9ef9ef5d97dbbd567da4567af0c084092b49f`, whose intervening net change is documentation-only.

- `cargo fmt --all --check`
- `cargo test -p dynamo-kv-router` (741 passed, 2 ignored)
- `cargo test -p dynamo-bench --test active_sequences_trace` (9 passed)

New focused coverage includes a 5,000-step lineage-consistent randomized prefix-union model, concurrent first-root insertion and cross-worker compressed-edge splits, cutoff splits before/at/after the boundary, shortening and re-extension, DP-rank isolation, idempotent removal, pinned-child cleanup, topology teardown, replica replay, expiry, and 65,536-node iterative release/destruction.


lib/kv-router/src/sequences/prompt_membership_trie.rs
```diff
@@ -338,298 +212,218 @@ impl PromptMembershipTrie {
...
+            };
+            drop(node);
+            current = next;
         }
-        worker_lookup.clear();
-        drop(worker_lookup);
+    }
 
-        for node in nodes {
+    pub(super) fn remove_worker(&self, worker: WorkerWithDpRank) {
```
jthomson04, line 418: **Topology teardown now walks the entire trie under `workers.write()`.** Previously teardown was O(worker's block count) via the lookup and ran outside the lock; now `remove_worker` is O(total trie nodes) while every request mutation is blocked on the worker-table write lock. `apply_topology_change_without_cleanup` also loops per removed worker, so a `reconcile` that drops R workers does R full BFS passes.

Topology changes are rare and holding the lock is what closes the same-identity re-registration race, so this is likely the right trade — but consider a single BFS taking the set of removed workers, and be aware of the pause this introduces during large scale-downs.

lib/kv-router/src/sequences/prompt_registry.rs
```diff
@@ -213,27 +212,26 @@ impl PromptRegistry {
...
-        lookup: &Arc<parking_lot::RwLock<WorkerLookup>>,
         delta: PromptMembershipDelta,
         load: WorkerLoadSnapshot,
     ) {
-        self.apply_membership_delta_and_load_without_cleanup(worker, lookup, delta, load);
+        self.apply_membership_delta_and_load_without_cleanup(worker, delta, load);
         self.maybe_cleanup();
     }
 
     pub(super) fn apply_membership_delta_and_load_without_cleanup(
```
jthomson04, line 228: **Load-bearing ordering contract is undocumented.** Correctness now strictly requires (a) `removes` applied before `stores`, mirroring tracker mutation order within a delta — guaranteed today because `force_expiry()` runs before `acquire_prompt` in `add_request_with_prefill_tracking` — and (b) per-worker serialized generation + application under the slot's `sequences.write()`. The old lookup-repair machinery tolerated drift; the new design doesn't. A future refactor that batches deltas differently (e.g. a store followed by a remove of the same path, applied remove-then-store) would silently corrupt membership.

Suggest a doc comment here (or on `PromptMembershipDelta`) stating both invariants.

lib/kv-router/src/sequences/prompt_membership_trie.rs
```diff
@@ -338,298 +212,218 @@ impl PromptMembershipTrie {
...
 
-                {
-                    let guard = node.read();
-                    let Some(&pos) = guard.edge_index.get(&parent_hash) else {
-                        continue;
-                    };
-                    if !guard.covers_pos(worker, pos) {
-                        worker_lookup.remove(&parent_hash);
+        let mut current = {
+            let mut root = self.root.write();
```
jthomson04, line 252: nit: this takes `root.write()` on every store, momentarily serializing all workers' stores even when the first child already exists (the common case). A `root.read()` fast path with write-upgrade-on-miss would remove that. The A/B numbers say it doesn't matter at 8 lanes, so optional.

lib/kv-router/src/sequences/prompt_membership_trie.rs
```diff
@@ -338,298 +212,218 @@ impl PromptMembershipTrie {
...
-                &mut worker_lookup,
-                worker,
-                hash,
-                LookupRepairDirection::TowardHead,
-            ) {
-                Some(node) => node,
-                None => continue,
+        let mut current = {
+            let root = self.root.read();
+            let Some(node) = root.children.get(&path[0]).cloned() else {
```
jthomson04, line 370: nit: this returns silently while every other impossible-state path in store/remove emits `tracing::warn!`. A missing root child on removal implies the same kind of inconsistency — consistent diagnostics would help if the invariants ever break in production.

lib/kv-router/src/sequences/prompt_membership_trie.rs
```diff
@@ -338,298 +212,218 @@ impl PromptMembershipTrie {
...
-                        worker,
-                        last_hash,
-                        LookupRepairDirection::TowardTail,
-                    ) {
-                        current_parent = resolved;
+            let path_ends = match_len == remaining.len();
+            let edge_ends = match_len == edge_len;
+            if !edge_ends {
+                if path_ends {
+                    if path_pos + match_len > new_suffix_start {
```
jthomson04, line 296: nit: this guard is always true — the entry assert guarantees `new_suffix_start < path.len()`, and in this branch `path_pos + match_len == path.len()`. Could be a `debug_assert!` or dropped.

jthomson04 (commented):
## Review

**Verdict: LGTM.** Traced the new mutation algorithms end-to-end (store/remove walks, split translation, release-path reconstruction, lock ordering) and found no correctness bugs. Independently re-ran the 108 `sequences::` tests at this head — all pass. Findings below: one design trade-off worth a conscious sign-off, one documentation request, and three nits (all inline).

### What was verified

- **Path/boundary arithmetic.** `release()` removes a contiguous chain suffix (the walk stops at the first node with `incoming > 1`), so `remove_from = prompt_depth − removed_count` is exact, and the two `assert_eq!`s cross-check the reconstruction. Empty-removal → `None`, full-removal (`retained = None`) → `remove_from = 0` both hold. Shared break-nodes above `prompt_depth` (output blocks) are correctly filtered by the `depth <= prompt_depth` guard.
- **Ordering invariants.** The registry applies `removes` before `stores` within a batched delta; this matches tracker mutation order because `force_expiry()` runs before `acquire_prompt` in `add_request_with_prefill_tracking`. See the inline doc request.
- **Concurrency.** All delta generation *and* trie application happen under `workers.read()` + the slot's `sequences.write()` (multi_worker and replica_sync), so per-worker mutations are serialized — required, since path records encode tracker state at generation time. Cross-worker races are handled by splits keeping the prefix in place (node identity stable), child Arcs cloned only while holding the parent's write lock, and the sweep's `strong_count == 2` gate pinning in-flight nodes. The `clear_children_if_unreachable` interleavings traced (concurrent store-through vs. last-full-worker drop) all converge to a correct reachable state.
- **Topology race fix.** Moving `apply_worker_topology_change_locked` inside `workers.write()` genuinely closes the window where a same-identity re-registration could interleave with the old worker's trie teardown, while publisher I/O correctly stays outside the lock.
- **No dangling references** to the removed APIs; remaining `WorkerLookup` hits are the untouched indexer/CRTC subsystem, as the description promises.

### Quality / performance / tests / security

- **Quality:** consistent with house style — fail-fast asserts on arena invariants, `tracing::warn!` + early-return on defensive paths (correctly skipping mutation rather than corrupting). The new code is substantially simpler than what it replaces (drops `edge_index`, `RemoveOutcome`, `LookupRepairDirection`, and all repair logic). Hidden win: detached subtrees are now actually freed — the old lookup maps' Arcs kept dead nodes alive.
- **Performance:** benchmark methodology is rigorous (pinned base SHA, alternating measured runs, medians of 10, jemalloc, CPU pinning). `release()` now does an O(prompt_depth) lineage walk per free and stores/removes walk from the root — both amortized away by eliminating per-block lookup maintenance, per the −56% free p50.
- **Tests:** strong. The 5,000-step randomized model validates full trie state (`worker_hashes` plus overlap depths per query shape) against a naive reference at every step, including DP-rank isolation and `remove_worker` resets. Split boundary translation is tested before/at/after the cut; idempotent removal and pinned-child cleanup are covered; the block-tracker reference model now checks positional `remove_from`. One gap: the only multithreaded test is two concurrent stores — a stress test mixing concurrent `store_path`/`remove_path`/`remove_worker` would directly exercise the `clear_children_if_unreachable` and truncate-vs-store races that were only verifiable by inspection.
- **Security:** no exposure — internal data-structure change with no new external input surface; the trusted-lineage contract for `SequenceHash` is unchanged.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -485,17 +474,27 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
+    /// Apply request-index and registry state while the caller holds
+    /// `workers.write()`. This excludes in-flight mutations and same-identity
+    /// re-registration until removed membership is fully drained.
+    fn apply_worker_topology_change_locked(&self, change: &WorkerTopologyChange) {
         for removed in &change.removed {
             self.request_index.remove_worker_requests(removed.worker);
+        }
+        self.prompt_registry
+            .apply_topology_change_without_cleanup(change);
+    }
```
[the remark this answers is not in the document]
jthomson04: Addressed on the rewritten main-based head. Worker-table mutation is again limited to workers.write(); teardown runs after releasing it and drains only the removed worker's weak node set, so request mutations no longer wait on a full-trie BFS.

lib/kv-router/src/sequences/prompt_membership_trie.rs
jthomson04: Addressed by removing the full-trie scan entirely. Each worker now tracks only compressed trie nodes carrying its membership, so reconcile teardown scales with the removed workers' node footprints and remains outside workers.write().

lib/kv-router/src/sequences/prompt_registry.rs
jthomson04: Documented both invariants on apply_membership_delta_and_load_without_cleanup: generation and application stay serialized under the slot sequence lock, and removals are applied before stores because expiry precedes acquisition.

lib/kv-router/src/sequences/prompt_membership_trie.rs
jthomson04: Implemented the root read fast path. store_path takes root.write() only after a read miss and rechecks before inserting.

lib/kv-router/src/sequences/prompt_membership_trie.rs
jthomson04: Implemented. A missing root child on removal now emits a warning; child disappearance caused by the same truncation intentionally unlinking descendants remains a normal quiet stop.

lib/kv-router/src/sequences/prompt_membership_trie.rs
jthomson04: Removed the redundant path-ending guard; that branch now extends worker coverage directly.

jthomson04:
@dcoapp recheck

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -461,7 +448,9 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
     ) -> Result<(), WorkerTopologyError> {
         let change = {
             let mut table = self.workers.write();
-            table.reconcile(self.block_size, ranges.into_iter().collect())?
+            let change = table.reconcile(self.block_size, ranges.into_iter().collect())?;
+            self.apply_worker_topology_change_locked(&change);
```
PeaBrane: AI-assisted, non-blocking performance note: this runs while `workers.write()` is held, and `apply_topology_change_without_cleanup` performs a full-trie `remove_worker` traversal once per removed rank. A large reconciliation is therefore approximately `O(removed ranks × trie nodes)` inside the global mutation lock. Since topology changes are a cold path and the steady-state benchmark does not exercise scale-down, I would accept this statically if large-reconcile latency is not operationally important. Otherwise, could we run a targeted scale-down benchmark? If it is material, a single bulk traversal or compact per-worker node inventory/generation scheme would avoid repeated sweeps.

lib/kv-router/src/sequences/prompt_membership_trie.rs
PeaBrane, line 252: AI-assisted optional scalability note: every store briefly takes the root write lock, but the current eight-lane benchmark includes this behavior and remains strongly positive, so I do not consider it blocking. If substantially higher mutation concurrency matters, a read fast path followed by `root.write()` plus a recheck on miss—or a wider lane-count sweep—would show whether this is worth optimizing.

lib/kv-router/src/sequences/prompt_registry.rs
```diff
@@ -213,27 +212,26 @@ impl PromptRegistry {
...
         &self,
         worker: WorkerWithDpRank,
-        lookup: &Arc<parking_lot::RwLock<WorkerLookup>>,
         delta: PromptMembershipDelta,
         load: WorkerLoadSnapshot,
     ) {
         for remove in delta.removes {
-            self.membership.remove_chain(worker, lookup, &remove.hashes);
+            self.membership
```
PeaBrane, line 235: Static documentation request: the path boundaries remain correct only while delta generation and application are serialized under the worker slot’s `sequences.write()` lock, and removals must be applied before stores because expiry precedes acquisition. The current callers and ordering satisfy both invariants, but the types do not encode them. Could we document that contract here? No benchmark is needed for this one.

PeaBrane (approved):
AI-assisted review. I found no correctness blocker and am approving. The inline notes distinguish one cold-path topology concern that may merit targeted scale-down measurement, one optional higher-concurrency optimization already covered favorably by the eight-lane benchmark, and one static documentation request.

PeaBrane (commented):
AI-assisted note for posterity: #11515 makes the derived `PromptRegistry` read projection substantially more collision-tolerant. Membership mutations now carry the complete path, and the raw `SequenceHash → Node` inverse lookup is removed. If two distinct lineages reach the registry, a store or removal follows the intended path instead of resolving through whichever raw-hash entry was installed most recently.

This does not attempt end-to-end collision prevention. The registry still depends on `BlockTracker` preserving the lineage and emitting the correct path and boundary. The current tracker is keyed by raw `SequenceHash`, so a collision can be coalesced or hidden before any membership event reaches the derived read model. In that case, `PromptRegistry` cannot recover the missing identity and may faithfully reflect already-malformed authoritative state.

As discussed in #11508, the accepted sequence-side contract treats these 64-bit collisions as an extremely unlikely birthday-bound case within the relevant live set. This remains consistent with the existing raw-`SequenceHash` identity used by the sequence-side/CRTC paths rather than introducing PLH-based end-to-end validation here.

So this is not a regression: it removes one downstream source of collision ambiguity, but intentionally does not claim complete collision safety. End-to-end prevention would additionally require collision-aware or validated identity in the authoritative tracker, which remains out of scope.

PR #11520, opened 2026-07-10
fix(lora): make disaggregated prefill lifecycle-aware

biswapanda:
## Summary

- register and serve LoRA lifecycle endpoints on legacy vLLM prefill workers
- make initial prefill registration metadata-only and activate adapters from the inference-time `LoRARequest`
- publish prefill adapter model cards with configured LoRA capacity while keeping decode and aggregated workers eager
- deduplicate discovered endpoints by physical pod and use stable pod/workload identity for lifecycle calls
- add a narrow rolling-upgrade fallback for the exact legacy missing-handler response, scoped to Kubernetes-ready vLLM prefill workers covered by a capable prefill in the same topology group
- unregister adapter discovery before engine removal to prevent transient base-model fallback during unload

## Root cause

The operator discovered duplicate service endpoints and did not call the LoRA lifecycle API on prefill workers. As a result, disaggregated deployments could publish only the decode adapter model card, leaving the `DynamoModel` NotReady and the frontend without a complete adapter worker set. Eager prefill loading would also consume a vLLM LoRA slot before the adapter was used.

This change makes prefill registration publish the adapter topology and cache its deterministic ID/path without eagerly adding it to the engine. The request path supplies the adapter path to vLLM for lazy activation.

Linear: https://linear.app/nvidia/issue/DYN-3408

## Validation

- `go list ./... | grep -v /e2e | xargs go test`
- `go vet ./api/v1alpha1 ./internal/modelendpoint ./internal/controller`
- `go test -race ./internal/modelendpoint ./internal/controller`
- changed Python source and tests pass `python3 -m py_compile`
- GCP Kubernetes validation in namespace `dyn-3408-lora` using exactly four GPUs (2 prefill + 2 decode):
  - a fresh LoRA `DynamoModel` reached 2/2 Ready
  - `/v1/models` listed the base model and adapter
  - adapter chat through the frontend returned HTTP 200
  - prefill logs showed adapter-path resolution at request time, after readiness
  - delete/recreate unloaded and republished both worker roles cleanly

## Summary by CodeRabbit

* **New Features**
  * Improved LoRA adapter lifecycle support for disaggregated deployments.
  * Prefill workers can defer adapter activation until needed, while decode workers continue to preload adapters.
  * Prefill workers now expose LoRA load, unload, and listing operations when enabled.
  * Model registrations now advertise available LoRA capacity.
  * Added compatibility handling for legacy prefill workers during rolling upgrades.

* **Bug Fixes**
  * Improved endpoint readiness tracking and deduplication.
  * Redirects and oversized error responses are now handled safely.
  * LoRA failures provide more reliable cleanup and status reporting.

* **Documentation**
  * Clarified LoRA endpoint readiness behavior during lifecycle operations and upgrades.

### DIFF
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -1948,6 +1948,15 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                 self._lora_load_locks[lora_name] = lock
             return lock
 
+    def _preload_lora_into_engine(self) -> bool:
+        """Whether lifecycle registration should eagerly activate the adapter.
+
+        Prefill keeps the downloaded adapter metadata and supplies its path in
+        the inference-time ``LoRARequest``. Decode and aggregated workers must
+        be immediately ready to generate and therefore continue to preload.
+        """
+        return self.config.disaggregation_mode != DisaggregationMode.PREFILL
+
     async def load_lora(self, request=None):
         """
         Load a LoRA adapter dynamically into the vLLM's AsyncLLM engine.
@@ -2043,36 +2052,43 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                             }
                             return
 
-                    try:
-                        await self.engine_client.add_lora(
-                            LoRARequest(
-                                lora_name=lora_name,
-                                lora_int_id=lora_id,
-                                lora_path=lora_path,
+                    # Initial prefill registration is metadata-only. A hot
+                    # swap must still replace any lazily activated old adapter
+                    # atomically before the prefix cache is reset.
+                    preload_into_engine = (
+                        self._preload_lora_into_engine() or is_hot_swap
+                    )
+                    if preload_into_engine:
+                        try:
+                            await self.engine_client.add_lora(
+                                LoRARequest(
+                                    lora_name=lora_name,
+                                    lora_int_id=lora_id,
+                                    lora_path=lora_path,
+                                )
                             )
-                        )
-                    except Exception as e:
-                        if is_hot_swap and old_info is not None:
-                            try:
-                                await self.engine_client.add_lora(
-                                    LoRARequest(
-                                        lora_name=lora_name,
-                                        lora_int_id=old_info.id,
-                                        lora_path=old_info.path,
+                        except Exception as e:
+                            if is_hot_swap and old_info is not None:
+                                try:
+                                    await self.engine_client.add_lora(
+                                        LoRARequest(
+                                            lora_name=lora_name,
+                                            lora_int_id=old_info.id,
+                                            lora_path=old_info.path,
+                                        )
                                     )
-                                )
-                            except Exception as rollback_error:
-                                self.loaded_loras.pop(lora_name, None)
-                                logger.exception(
-                                    f"Rollback failed for LoRA {lora_name}: "
-                                    f"{rollback_error}"
-                                )
-                        yield {
-                            "status": "error",
-                            "message": f"Failed to add LoRA '{lora_name}': {e}",
-                            "lora_name": lora_name,
-                        }
-                        return
+                                except Exception as rollback_error:
+                                    self.loaded_loras.pop(lora_name, None)
+                                    logger.exception(
+                                        f"Rollback failed for LoRA {lora_name}: "
+                                        f"{rollback_error}"
+                                    )
+                            yield {
+                                "status": "error",
+                                "message": f"Failed to add LoRA '{lora_name}': {e}",
+                                "lora_name": lora_name,
+                            }
+                            return
 
                     # Track the LoRA
                     self.loaded_loras[lora_name] = LoRAInfo(id=lora_id, path=lora_path)
@@ -2095,7 +2111,8 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                             rolled_back = "tracking only"
                             if old_info is not None:
                                 try:
-                                    await self.engine_client.remove_lora(lora_id)
+                                    if preload_into_engine:
+                                        await self.engine_client.remove_lora(lora_id)
                                     await self.engine_client.add_lora(
                                         LoRARequest(
                                             lora_name=lora_name,
@@ -2219,12 +2236,14 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                                 f"Failed to publish LoRA {lora_name} ModelDeploymentCard: {e}"
                             )
 
-                            # Rollback: remove the LoRA from the engine to maintain consistency
+                            # Roll back engine state when this worker preloaded;
+                            # prefill only needs to discard the cached metadata.
                             try:
-                                logger.debug(
-                                    f"Rolling back: removing LoRA '{lora_name}' from engine"
-                                )
-                                await self.engine_client.remove_lora(lora_id)
+                                if preload_into_engine:
+                                    logger.debug(
+                                        f"Rolling back: removing LoRA '{lora_name}' from engine"
+                                    )
+                                    await self.engine_client.remove_lora(lora_id)
                                 self.loaded_loras.pop(lora_name, None)
                                 logger.debug(
                                     f"Successfully rolled back LoRA '{lora_name}'"
@@ -2299,14 +2318,11 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
 
                     logger.debug(f"Unloading LoRA adapter: {lora_name}")
                     lora_id = lora.id
-                    lora_path = lora.path
 
-                    await self.engine_client.remove_lora(lora_id)
-
-                    # Remove from tracking
-                    del self.loaded_loras[lora_name]
-
-                    # Unregister the LoRA model from the model registry
+                    # Stop advertising the adapter before mutating engine or
+                    # tracking state. Otherwise requests can still route here
+                    # after _resolve_lora_request has forgotten the adapter and
+                    # silently execute against the base model.
                     if self.generate_endpoint is not None:
                         logger.debug(
                             f"Unregistering LoRA '{lora_name}' ModelDeploymentCard"
@@ -2323,32 +2339,6 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                             logger.exception(
                                 f"Failed to unregister LoRA {lora_name} ModelDeploymentCard: {e}"
                             )
-
-                            # Rollback: re-add the LoRA to the engine to maintain consistency
-                            try:
-                                logger.debug(
-                                    f"Rolling back: re-adding LoRA '{lora_name}' to engine"
-                                )
-                                await self.engine_client.add_lora(
-                                    LoRARequest(
-                                        lora_name=lora_name,
-                                        lora_int_id=lora_id,
-                                        lora_path=lora_path,
-                                    )
-                                )
-                                # Re-add to tracking
-                                self.loaded_loras[lora_name] = LoRAInfo(
-                                    id=lora_id, path=lora_path
-                                )
-                                logger.debug(
-                                    f"Successfully rolled back LoRA '{lora_name}'"
-                                )
-                            except Exception as rollback_error:
-                                logger.exception(
-                                    f"Failed to rollback LoRA {lora_name}: {rollback_error}"
-                                )
-
-                            # Return error status since unregistration failed
                             yield {
                                 "status": "error",
                                 "message": f"Failed to unregister LoRA '{lora_name}' from discovery registry: {str(e)}",
@@ -2359,6 +2349,12 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                         logger.debug(
                             f"Cannot unregister LoRA '{lora_name}': generate_endpoint={self.generate_endpoint}"
                         )
+
+                    # Discovery no longer routes new requests here. If engine
+                    # removal fails, keep tracking so a retry can remove it
+                    # without ever falling back to the base model.
+                    await self.engine_client.remove_lora(lora_id)
+                    del self.loaded_loras[lora_name]
 
                     logger.info(
                         f"Successfully unloaded LoRA adapter: {lora_name} with ID {lora_id}"
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -1014,8 +1014,9 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
         # metadata-only, but vLLM activates a prefill adapter lazily when an
         # inference request supplies its LoRARequest.
         self._engine_loaded_loras: set[str] = set()
-        # Fixed stripes serialize same-name lifecycle and lazy-request admission
-        # without retaining one asyncio.Lock per adapter ever seen.
+        # Requests and load/unload operations for the same adapter share a lock,
+        # so an unload cannot race with vLLM's lazy adapter activation. A fixed
+        # number of shared locks bounds memory as adapter names come and go.
         self._lora_load_locks = [asyncio.Lock() for _ in range(_LORA_LOCK_STRIPES)]
         self._paused: bool = False
         self._weight_version: str = "initial"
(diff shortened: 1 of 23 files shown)


components/src/dynamo/vllm/llm_engine.py
```diff
@@ -1197,25 +1218,29 @@ async def unload_lora(self, body: dict) -> dict:
...
-                    return {
-                        "status": "error",
-                        "message": f"Failed to remove LoRA '{lora_name}' from engine: {str(e)}",
-                        "lora_name": lora_name,
-                    }
+                # Prefill lifecycle registration is metadata-only. Only
+                # remove adapters this process explicitly preloaded; asking
+                # vLLM to remove a metadata-only adapter fails and leaves the
+                # discovery and local lifecycle state out of sync.
+                if lora_name in self._engine_loaded_loras:
```
julienmancuso, line 1277: [P1] Evict request-loaded prefill adapters before dropping lifecycle metadata

A prefill registration starts metadata-only, but the first inference request passes this adapter's `LoRARequest` to vLLM. In the pinned vLLM 0.24 path, `set_active_adapters`/`_apply_adapters` loads a missing adapter on demand. That request path never adds the name to `_engine_loaded_loras`, so this guard skips `remove_lora` and then deletes `loaded_loras`. The engine can retain the old deterministic adapter ID, and deleting/recreating the same adapter name can therefore reuse stale weights. Track request-time activation or query vLLM state during unload (treating not-found as idempotent), and apply the same fix to the legacy path in `handlers.py`.

deploy/operator/internal/modelendpoint/client.go
```diff
@@ -86,41 +94,78 @@ func (c *Client) LoadLoRA(
...
 	for _, result := range results {
-		endpoints[result.Index] = result.Value
-		if result.Value.Ready {
+		candidate := candidates[result.Index]
+		endpoint := &endpoints[result.Index]
+		if result.Value.usedUnavailableFallback {
+			// A legacy prefill can be non-serving only when another capable
+			// vLLM prefill in the same runtime topology published the adapter card.
+			_, covered := capableVLLMPrefillGroups[candidate.LoRAFallbackGroup]
+			endpoint.Ready = candidate.LoRAFallbackGroup != "" && covered && candidate.KubernetesReady
```
julienmancuso: [P1] Preserve the existing per-endpoint `Ready` contract

This sets `EndpointInfo.Ready=true` after this specific endpoint returned the legacy HTTP 500 and did not load or publish the adapter, based on coverage by a different pod. That changes the public meaning of `status.endpoints[*].ready` and inflates `readyEndpoints` with a pod that is not itself ready to serve the adapter; existing monitoring and automation can receive a false positive after an operator upgrade. Keep per-endpoint readiness false and represent topology-level rolling coverage separately (for example, an optional fallback field or model-level condition) when deciding aggregate `DynamoModel` readiness.

julienmancuso (commented):
Two P1 backward-compatibility findings from the operator upgrade review.

biswapanda:
@coderabbitai Addressed the outside-diff Black formatting observation. The test was reformatted, and the current pre-commit check passes.

components/src/dynamo/vllm/llm_engine.py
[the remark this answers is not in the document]
biswapanda: Addressed in 85ae20abb3 and refined in fbcbb4bf5f. Prefill lifecycle registration stays metadata-only; unload removes only adapters known to have reached vLLM, and a missing-adapter response is idempotent. Added unified-path regression coverage.

components/src/dynamo/vllm/handlers.py
```diff
@@ -2259,6 +2249,12 @@ async def unload_lora(self, request=None):
                             f"Cannot unregister LoRA '{lora_name}': generate_endpoint={self.generate_endpoint}"
                         )
 
+                    # Discovery no longer routes new requests here. If engine
+                    # removal fails, keep tracking so a retry can remove it
+                    # without ever falling back to the base model.
+                    await self.engine_client.remove_lora(lora_id)
```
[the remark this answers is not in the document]
biswapanda: Addressed in 85ae20abb3 and refined in fbcbb4bf5f. The legacy path now follows the same metadata-only/request-activation tracking and idempotent removal behavior. Added legacy regression coverage.

deploy/operator/internal/controller/dynamo_model_controller.go
```diff
@@ -373,6 +481,34 @@ func (r *DynamoModelReconciler) getEndpointCandidates(
...
+			r.Recorder.Event(model, corev1.EventTypeWarning, "ComponentDiscoveryFailed", err.Error())
+			return nil, nil, err
+		}
+		graphs := &v1beta1.DynamoGraphDeploymentList{}
+		if err := r.List(ctx, graphs, client.InNamespace(model.Namespace)); err != nil {
+			logs.Error(err, "Failed to list graph deployments for LoRA lifecycle classification")
+			r.Recorder.Event(model, corev1.EventTypeWarning, "GraphDiscoveryFailed", err.Error())
+			return nil, nil, err
+		}
+		candidates = markLoRAManagementUnavailableFallbackEligible(candidates, components.Items, graphs.Items)
```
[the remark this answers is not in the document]
biswapanda: Reviewed. This is an informational control-plane scaling tradeoff, not a correctness regression. The lookup remains scoped to LoRA reconciliation so fallback eligibility can be derived reliably; DCD/DGD watches now requeue affected LoRA models when that eligibility changes.

deploy/operator/internal/controller/dynamo_model_controller.go
```diff
@@ -373,6 +481,34 @@ func (r *DynamoModelReconciler) getEndpointCandidates(
...
+	}
```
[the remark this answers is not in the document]
biswapanda: Addressed in 85ae20abb3. SetupWithManager now watches DynamoComponentDeployment and DynamoGraphDeployment changes and maps them to LoRA DynamoModels in the same namespace. The mapping is covered by controller unit tests.

components/src/dynamo/vllm/llm_engine.py
biswapanda: Addressed in fbcbb4bf5f. Both unified and legacy handlers record an adapter when its LoRARequest is handed to vLLM, then remove it during unload; a not-loaded response is treated as idempotent. Added regression coverage for request activation and missing-adapter removal.

deploy/operator/internal/modelendpoint/client.go
biswapanda: Addressed in fbcbb4bf5f. EndpointInfo.Ready remains true only when that endpoint loaded the adapter. Rolling-upgrade coverage is now represented separately through loraFallbackCovered and loraFallbackCoveredEndpoints, while model readiness uses the effective aggregate. Regenerated CRDs and API reference in 8fe42654a9.

deploy/operator/internal/controller/dynamo_model_controller.go
```diff
@@ -266,10 +398,47 @@ func (r *DynamoModelReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				GenericFunc: func(e event.GenericEvent) bool { return false },
 			}),
 		).
+		// The vLLM-prefill fallback classification depends on the owning
+		// component or graph deployment. Reconcile the affected namespace's
+		// LoRA models when either workload definition changes instead of
+		// waiting for an unrelated EndpointSlice update.
```
sttts: this has some bad architectural taste that a child reconciler needs the parent. Reconcilers should be a one-way route.

deploy/operator/internal/controller/dynamo_model_controller.go
```diff
@@ -373,6 +542,34 @@ func (r *DynamoModelReconciler) getEndpointCandidates(
...
+				logs.Error(err, "Failed to get endpoint pod for LoRA lifecycle classification", "podName", candidates[i].PodName)
+				return nil, nil, err
+			}
+			candidates[i] = withPodIdentity(candidates[i], pod)
+		}
+
+		components := &v1beta1.DynamoComponentDeploymentList{}
+		if err := r.List(ctx, components, client.InNamespace(model.Namespace)); err != nil {
+			logs.Error(err, "Failed to list component deployments for LoRA lifecycle classification")
+			r.Recorder.Event(model, corev1.EventTypeWarning, "ComponentDiscoveryFailed", err.Error())
```
sttts: this might be preexisting as a pattern: creating an event for a either plumbing error or a transient problem is an anti-pattern.

biswapanda: Agreed, broader Event anti-pattern predates this PR through EndpointDiscoveryFailed. 
That legacy occurrence should remain outside this patch for release-oriented fix.

deploy/operator/internal/controller/dynamo_model_controller.go
biswapanda: Addressed in 8896d25485. The DynamoModel reconciler no longer watches, lists, or reads DynamoComponentDeployment/DynamoGraphDeployment parents; the added RBAC and namespace-wide mapper are removed as well. Fallback eligibility now flows one way from the discovered Pod’s existing component/topology labels and vLLM command, and fails closed when the child does not prove eligibility. The focused controller tests, full non-E2E operator suite, and the four-GPU GCP lifecycle test pass.

deploy/operator/internal/controller/dynamo_model_controller.go
biswapanda: Addressed in 8896d25485. The PR-added ComponentDiscoveryFailed and GraphDiscoveryFailed transient Events were removed together with the parent-resource reads. I intentionally left the pre-existing EndpointDiscoveryFailed pattern unchanged to keep this release cherry-pick scoped to newly added code.

components/src/dynamo/vllm/tests/test_vllm_lora.py
```diff
@@ -641,6 +751,57 @@ async def test_unload_lora_engine_removal_failure_after_unregister(monkeypatch):
...
+    engine.loaded_loras = {"adapterA": LoRAInfo(id=123, path="/cache/adapter")}
+    engine._engine_loaded_loras = {"adapterA"}
+    engine.engine_client.remove_lora.side_effect = RuntimeError("adapter not found")
+
+    result = await engine.unload_lora({"lora_name": "adapterA"})
+
+    assert result["status"] == "success"
+    unregister.assert_awaited_once()
+    assert "adapterA" not in engine.loaded_loras
+    engine.engine_client.remove_lora.assert_not_awaited()
```
julienmancuso: [P1] Assert the idempotent removal was awaited

This setup puts `adapterA` in `_engine_loaded_loras`, so `unload_lora` necessarily executes `await self.engine_client.remove_lora(123)`. The configured `RuntimeError("adapter not found")` is deliberately swallowed as an idempotent removal, but `AsyncMock` still records the await; therefore `assert_not_awaited()` always fails and the vLLM unit suite cannot pass. This should assert `assert_awaited_once_with(123)` instead.

biswapanda: Fixed in f73cc880b5. The idempotent not-found path still awaits `remove_lora(123)`; the test now asserts that await rather than asserting it was not awaited.

julienmancuso (approved):

components/src/dynamo/vllm/tests/test_vllm_lora.py
biswapanda: fixed

components/src/dynamo/vllm/llm_engine.py
```diff
@@ -500,6 +511,7 @@ async def generate(
         # model resolves to None. With LoRA enabled, an unknown adapter name
         # raises rather than silently falling back to the base model.
         lora_request = self._resolve_lora_request(request.get("model"))
+        self._track_lora_request_activation(lora_request)
```
sttts: [P2] Serialize lazy request admission with unload

The adapter is added to `_engine_loaded_loras` before vLLM actually admits the async generator's request. An unload can run while `add_request` is suspended, call `remove_lora` before the adapter exists, clear `_engine_loaded_loras` and `loaded_loras`, and return success; the request can then resume and lazily load the captured path, leaving an untracked adapter in vLLM. Please serialize same-name request admission with lifecycle removal, mirror the fix in the legacy handler, and add a gated concurrency test for both paths.

sttts (commented):
Code-only inline finding.

components/src/dynamo/vllm/llm_engine.py
biswapanda: Addressed in 1382c91ee2. Unified and legacy request paths now hold the deterministic per-adapter lifecycle stripe through the first vLLM generator iteration, so unload cannot remove lifecycle tracking before lazy admission completes. If unload wins while a request is waiting, the request re-resolves under the same lock and fails rather than admitting a stale adapter path. Added `gpu_0`-gated concurrency coverage for both paths.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1009,10 +1010,13 @@ def __init__(
...
-        # Per-LoRA locks to prevent concurrent load operations for the same LoRA
-        self._lora_load_locks: dict[str, asyncio.Lock] = {}
-        # Guard lock-map access in case handlers are invoked from multiple threads.
-        self._lora_load_locks_guard = threading.Lock()
+        # Adapters known to have been handed to vLLM. Prefill registration is
+        # metadata-only, but vLLM activates a prefill adapter lazily when an
+        # inference request supplies its LoRARequest.
+        self._engine_loaded_loras: set[str] = set()
+        # Fixed stripes serialize same-name lifecycle and lazy-request admission
+        # without retaining one asyncio.Lock per adapter ever seen.
```
tzulingk: nit: sorry, I don't understand the comments, could you rephrase it, so that people who is not that familiar with the codes could understand it easily?

components/src/dynamo/vllm/handlers.py
```diff
@@ -1939,14 +1943,66 @@ def _resolve_lora_request(self, model_name: str | None) -> LoRARequest | None:
...
+        if lora_request is None or self._preload_lora_into_engine():
+            self._track_lora_request_activation(lora_request)
+            async for result in create_generator(lora_request):
+                yield result
+            return
+
+        lock = self._get_lora_lock(lora_request.lora_name)
+        async with lock:
+            # A pending request may have waited behind an unload. Re-resolve
+            # under the lock instead of admitting its stale path afterwards.
```
tzulingk: what does the stale-path refer to? sorry, I am not familiar with lora codes

components/src/dynamo/vllm/handlers.py
```diff
@@ -1939,14 +1943,66 @@ def _resolve_lora_request(self, model_name: str | None) -> LoRARequest | None:
...
+            if admitted_lora_request is None:
+                raise ValueError(
+                    f"unknown model or LoRA adapter: '{lora_request.lora_name}'"
+                )
+            generator = create_generator(admitted_lora_request)
+            self._track_lora_request_activation(admitted_lora_request)
+            try:
+                first_result = await anext(generator)
+            except StopAsyncIteration:
+                return
```
tzulingk, line 1998: should we issue some warning messages instead of directly return from exception?

tzulingk (changes requested):

components/src/dynamo/vllm/handlers.py
biswapanda: Addressed in 963cb74b63. Rephrased the lock comment in plain terms: requests and load/unload operations for the same adapter share a lock so unload cannot race with lazy vLLM activation, while a fixed lock count bounds memory.

components/src/dynamo/vllm/handlers.py
biswapanda: Addressed in 963cb74b63. The comment now states explicitly that while a request waits, the adapter may be unloaded or reloaded at a different filesystem path, so the request looks it up again while holding the lifecycle lock.

components/src/dynamo/vllm/handlers.py
biswapanda: Addressed in 963cb74b63. The rejection path now logs a warning identifying the adapter that was unloaded before vLLM admission, and the existing regression test asserts that warning.

sttts (approved):

tzulingk (approved):

peilii (approved):

jthomson04 (approved):

Issue #8749, opened 2026-04-27
DEP (light): Worker self-hosted metadata files


nnshah1:
> **Updated 2026-04-28** to reflect the final design after review. The original `Vec<ArtifactRef>` proposal has been replaced by a single additive field (`CheckedFile.size: Option<u64>`) on the existing typed-enum slots. Implementation: PR [#8754](https://github.com/ai-dynamo/dynamo/pull/8754).

### Authors

@tanmayv25, @grahamking, @nicolasnoble, @biswapanda, @Kaonael

### Area

frontend

### Summary

Enable workers to self-host the model metadata files the frontend needs for preprocessing (`config.json`, `tokenizer.json` / `tokenizer.model`, `tokenizer_config.json`, `chat_template.{jinja,json}`, `generation_config.json`).

Workers serve those files at `GET /v1/metadata/{model_slug}/{filename}` on the existing `system_status_server`, and rewrite the per-file `CheckedFile` URI on the MDC to that route. The MDC's typed-enum `CheckedFile` slots are the single source of identity — every slot already carries a blake3 checksum and a `path: Either<PathBuf, Url>` capable of holding any URI scheme. The only new field is an additive `CheckedFile.size: Option<u64>` that doubles as the per-file fetch cap and the new/legacy discriminant.

### Motivation

The frontend today resolves metadata files via one of two paths: a local read (when frontend and worker share a filesystem, or both pods happen to have the same path on disk) or `hub::from_hf(source_path)` (which transparently uses ModelExpress as a regional cache when `MODEL_EXPRESS_URL` is set, otherwise hits HuggingFace directly). Both paths fail for **private or custom models that aren't published anywhere reachable** — ModelExpress can't serve a file that doesn't exist upstream, and HF can't either. They also fail when `source_path` isn't a valid HF slug (e.g. `--served-model-name=cosmos-reason1-7b`): the HF download errors and the model never registers with the frontend. `curl /v1/models` returns empty data; the failure is visible in frontend pod logs, but `/v1/models` itself gives operators no signal that registration was attempted.

[...]

PR #5102, opened 2025-12-27
fix: preserve original model path for frontend config downloads

yurekami:
## Overview
Fixes the frontend model download bug where using `--served-model-name` causes config download to fail because the customized name doesn't match any HuggingFace repository.

### Root Cause
When `--served-model-name` is used, `set_name()` overwrites `display_name` with the custom name. Later, `download_config()` uses `display_name` to fetch from HuggingFace, but the custom name isn't a valid HuggingFace repository path.

### Solution
1. **Add `source_model` field** to `ModelDeploymentCard` - preserves the original HuggingFace repository path
2. **Update `set_name()`** - automatically saves the original `display_name` to `source_model` before overwriting (addresses CodeRabbit's review feedback)
3. **Update `download_config()`** - uses `source_model` when available, falls back to `display_name` for backward compatibility
4. **Initialize `source_model: None`** in `from_repo_checkout()` - for local paths where no HF download is needed

### Key Improvement
The `set_name()` fix ensures that whenever `--served-model-name` is called, the original model path is automatically preserved. This is more robust than requiring callers to manually set `source_model`.

## Where to start
- `lib/llm/src/model_card.rs` - All changes are in this file

## Related Issues
Fixes #4748

## Test Plan
1. Deploy a model with `--served-model-name custom-name`
2. Verify frontend can still download config files from the original HuggingFace repository
3. Verify backward compatibility when `source_model` is not set

---

**Note:** This PR is based on the fork's main branch. There may be merge conflicts with upstream due to recent changes. If so, please use "Update with merge" or squash-merge after resolving conflicts in `lib/llm/src/model_card.rs`.

## Release Notes

* **Improvements**
  * Enhanced model download reliability through improved handling of model source path resolution
  * Strengthened tracking of model origins for better resource management

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #4748, opened 2025-12-04
[BUG]: Frontend pod tries to download a model via huggingface with the value of arg --served-model-name (and it fails)


MisakiNakagawan:
### Describe the Bug

Hi, I deployed a VLM model in Dynamo with huggingface model, adding an argument "--served-model-name".
All the pod deployed successfully. However, when I use API, it fails. It seems that the frontend pod tries to fetch the model from huggingface with the value of --served-model-name.

Does frontend pod need to download a model? Even if so, I guess it use "--model" value, not "--served-model-name" value.

Thanks in advance!

### Steps to Reproduce

1. Apply the yaml file below  e.g. `kubectl apply -f dynamo/examples/backends/vllm/deploy/vlm_agg.yaml -n dynamo-system`
2. Wait until all pod is READY
3. Get the svc name and port-forward  e.g. `kubectl port-forward svc/llm-vllm-agg-llmfrontend 8000:8000 -n dynamo-system`
4. `curl localhost:8000/v1/models`
5. get the log of the frontend pod  e.g.  `kubectl logs <frontend pod> -n dynamo-system`
```
### yaml file for kubectl apply
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0

apiVersion: nvidia.com/v1alpha1
kind: DynamoGraphDeployment
metadata:
  name: vlm-vllm-agg
spec:
  services:
    VLMFrontend:
      dynamoNamespace: vlm-vllm-agg
      componentType: frontend
      replicas: 1
      extraPodSpec:
        mainContainer:
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.6.1
    VLMVllmDecodeWorker:
      envFromSecret: hf-token-secret
      dynamoNamespace: vlm-vllm-agg
      componentType: worker
      replicas: 1
      resources:
        limits:
          gpu: "1"
      extraPodSpec:
        mainContainer:
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.6.1
          workingDir: /workspace/examples/backends/vllm
          command:
            - python3
            - -m
            - dynamo.vllm
          args:
            - --model=nvidia/Cosmos-Reason1-7B
            - --served-model-name=cosmos-reason1-7b
          startupProbe:
            httpGet:
              path: /health
              port: 9090
            initialDelaySeconds: 120
[...]

PR #11522, opened 2026-07-10
fix(local_model): keep local directory as source_path when a custom model name is set

nicole-lihui:
## Summary

When `LocalModelBuilder` loads a local directory with a custom served name and no explicit `source_path`, the source falls back to that name. If metadata self-hosting is disabled or unavailable, `attach()` can then publish bogus `hf://<served-name>/...` metadata URLs.

Preserve the canonical local directory before applying a different served name. Explicit local or Hugging Face sources still take precedence. Default names, including an explicitly supplied name equal to the canonical path, retain the existing absent-source representation and MDC checksum.

The change is confined to `LocalModelBuilder` and regression tests in `local_model.rs`. Metadata transport, frontend resolution, the checksum algorithm, and model-weight distribution are unchanged. Renamed local models acquire an explicit source and therefore a different MDC checksum; this PR does not introduce mixed-checksum WorkerSet admission.

## Validation

Validated on macOS after rebasing onto `8612fc179`:

- `cargo test -p dynamo-llm --no-default-features --lib metadata_publish_tests`: 3 passed. Covers actual `attach()` publication and frontend discovery with a custom name, unchanged default/same-path names, and explicit local/HF source precedence.
- `cargo clippy -p dynamo-llm --no-default-features --lib --tests --no-deps -- -D warnings`
- `git diff --check upstream/main`
- `pre-commit run --files lib/llm/src/local_model.rs`

Full Linux/CUDA CI requires separate validation on the updated PR head.

## Related Issues

- Related to #4748
- Follow-up to #5102
- Related metadata transport design: #8749
- Part of #14253

### DIFF
--- a/lib/bindings/python/rust/llm/entrypoint.rs
+++ b/lib/bindings/python/rust/llm/entrypoint.rs
@@ -724,6 +724,9 @@ pub fn make_engine<'p>(
     pyo3_async_runtimes::tokio::future_into_py(py, async move {
         if let Some(model_path) = args.model_path.clone() {
             let local_path = if model_path.exists() {
+                // Preserve the on-disk path as source_path so a custom
+                // --model-name does not become the HF repo id fallback.
+                builder.source_path(model_path.clone());
                 model_path
             } else {
                 // Mocker only needs tokenizer, not weights
--- a/lib/bindings/python/rust/llm/entrypoint.rs
+++ b/lib/bindings/python/rust/llm/entrypoint.rs
@@ -724,9 +724,6 @@ pub fn make_engine<'p>(
     pyo3_async_runtimes::tokio::future_into_py(py, async move {
         if let Some(model_path) = args.model_path.clone() {
             let local_path = if model_path.exists() {
-                // Preserve the on-disk path as source_path so a custom
-                // --model-name does not become the HF repo id fallback.
-                builder.source_path(model_path.clone());
                 model_path
             } else {
                 // Mocker only needs tokenizer, not weights
--- a/lib/llm/src/local_model.rs
+++ b/lib/llm/src/local_model.rs
@@ -382,6 +382,10 @@ impl LocalModelBuilder {
         // path of the downloaded model.
         if let Some(source_path) = self.source_path.take() {
             card.set_source_path(source_path);
+        } else {
+            // Local snapshot with a custom display name must still record the
+            // directory as the metadata/weight source, not display_name().
+            card.set_source_path(model_path.clone());
         }
         // The served model name defaults to the full model path.
         // This matches what vllm and sglang do.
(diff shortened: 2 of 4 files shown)


lib/llm/src/model_card.rs
```diff
@@ -696,6 +696,20 @@ fn parse_hf_uri(uri: &str) -> anyhow::Result<(String, String)> {
...
 
+fn local_model_path_overlay_uri(
+    filename: &str,
+    local_model_path: Option<&Path>,
+) -> anyhow::Result<Option<String>> {
+    let Some(prefix) = local_model_path else {
+        return Ok(None);
+    };
+    let local = prefix.join(filename);
+    if local.exists() {
```
jthomson04: [P2] This chooses any same-named local file before checking its checksum. A stale overlay then fails later in `resolve_uri`, and the valid `hf://` artifact is never tried. Pass the expected `CheckedFile` into this helper and only return the overlay when its checksum matches; otherwise continue to the original URI/fallback.

lib/bindings/python/rust/llm/entrypoint.rs
```diff
@@ -690,6 +690,9 @@ pub fn make_engine<'p>(
     pyo3_async_runtimes::tokio::future_into_py(py, async move {
         if let Some(model_path) = args.model_path.clone() {
             let local_path = if model_path.exists() {
+                // Preserve the on-disk path as source_path so a custom
+                // --model-name does not become the HF repo id fallback.
+                builder.source_path(model_path.clone());
```
jthomson04: [P2] This call is redundant with `LocalModelBuilder::build()` and changes the result for relative paths: `--model-path ./model` remains `./model` instead of the canonical directory. Delete the local-branch assignment; the remote branch still needs its explicit Hugging Face source.

lib/llm/src/local_model.rs
```diff
@@ -851,6 +846,62 @@ fn harvest_extra_files(
...
+    fn sample_model_dir() -> PathBuf {
+        Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/data/sample-models/TinyLlama_v1.1")
+    }
+
+    /// Regression test: a local model dir plus a custom `--model-name` must
+    /// keep the directory as `source_path`. Before the fix, `source_path()`
+    /// fell back to the custom display name, and `attach()` published
+    /// `hf://<custom-name>/...` URIs pointing at a nonexistent HF repo.
+    #[tokio::test]
+    async fn build_local_model_custom_name_keeps_directory_source_path() {
```
jthomson04: [P2] This test only exercises `build()`, despite describing the `attach()` rewrite that caused the regression. Add a focused test that would fail if `attach()` publishes `hf://<custom-name>/...` again.

jthomson04 (commented):
Three P2 findings from the static review.

lib/llm/src/local_model.rs
nicole-lihui: agreed, attach() is the real gap. Needs a real Endpoint/DistributedRuntime test

lib/llm/src/model_card.rs
nicole-lihui: Good catch! Thanks for spotting that.

lib/llm/src/local_model.rs
nicole-lihui: Added a unit test (checked_file_uri_rejects_custom_model_name_as_hf_source in model_card.rs) and a real attach()→etcd→discovery integration test (test_attach_publishes_directory_source_path_not_custom_name in http_metrics.rs). Verified both against a local kind cluster (etcd + NATS) — confirmed they fail without the fix and pass with it.

nicole-lihui:
@jthomson04  I rebased and narrowed this PR to the LocalModelBuilder source_path fix with regression coverage; the broader design is tracked in #14253. Could you re-review it and, if it looks good, post `/ok to test eef4f4d`?

PR #11526, opened 2026-07-10
fix(sglang): warm every prefill DP rank before serving

weireweire:
## Summary

Dynamo embedded SGLang prefill warmup currently sends one fake-transfer request through automatic load balancing. In a multinode DP deployment, this warms only one DP rank and leaves the remaining ranks to compile request-time kernels after the worker registers.

Using one Engine API batch of DP-size requests is not sufficient for DP-attention deployments. SGLang intentionally takes the sequential tokenization path when DP attention is enabled, so the API batch is emitted as individual scheduler dispatches. With the total-tokens balancer, load-snapshot refreshes can overwrite speculative dispatch counters before the scheduler publishes the new load; an observed DP8 startup routed two warmup requests to DP0 and none to DP7.

Send one concurrent warmup request per DP rank, route each request explicitly with `routed_dp_rank`, assign a unique bootstrap room, and drain every stream before registration. The public `routed_dp_rank` type remains scalar.

## Validation

- All repository pre-commit hooks and PR checks pass.
- A two-node DP8/TP8/EP8 startup logged exactly one warmup prefill request on each of DP0 through DP7 within a 1 ms window.
- Concurrently warming all eight ranks took 135.12 seconds, essentially unchanged from the 135.40-second single-rank embedded warmup.
- The 377-request AIPerf warmup completed with zero errors in 1109.42 seconds, down from 1268.64 seconds for the single-rank baseline: 159.22 seconds, or 12.6%, faster.
- A full 30-minute agentic replay passed with 16,383 measured requests, zero request errors, and a 96.84% theoretical prefix-cache hit rate versus 96.78% for the baseline.
- The change only runs before worker registration. The full replay is treated as correctness and production-path no-regression evidence; single-run throughput differences are not attributed to this warmup change.

similar fix in sglang: https://github.com/sgl-project/sglang/pull/30748

### DIFF
--- a/components/src/dynamo/sglang/_disagg.py
+++ b/components/src/dynamo/sglang/_disagg.py
@@ -14,9 +14,9 @@
   `dist_init_addr` into a runtime-data key that non-leader workers can use
   to discover the leader endpoint.
 
-* :func:`warmup_prefill_engine` — drive one request through the disagg
-  path with SGLang's `FAKE_BOOTSTRAP_HOST` so the first real request
-  doesn't pay the JIT/CUDA-graph compile cost.
+* :func:`warmup_prefill_engine` — drive one request per DP rank through
+  the disagg path with SGLang's `FAKE_BOOTSTRAP_HOST` so the first real
+  request doesn't pay the JIT/CUDA-graph compile cost.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


ishandhanani (approved):

weireweire:
@ishandhanani could you please trigger ci again?

ishandhanani (approved):

PR #11447, opened 2026-07-08
feat(router): add ThunderAgent admission strategy

ishandhanani:
PR #11434 defines the minimal policy-class admission API, and PR #11530 hosts its lifecycle plus a live request-progress capability. This stacked PR implements session-aware ThunderAgent in `dynamo-admission-control`. ThunderAgent retains quiescent sessions behind a 30-minute inactivity lease, preserving native performance without requiring clients to send a terminal-session request.

CLOSES: DYN-3231

### How This Was Implemented

- Keep the ThunderAgent algorithm, state, configuration, capacity accounting, and tests in the separate `dynamo-admission-control` crate.
- Reuse the policy queue and scheduler actor from the lower PRs for deferred storage, exact-worker readiness, cancellation, lifecycle correlation, and reconciliation.
- Retain the read-only `RequestProgress` capability from PR #11530 and sample it during admission/reconciliation; the response path performs one relaxed atomic update and sends no per-output actor messages.
- Implement sticky placement, capacity pressure, pause/mark/resume, timeout fallback, same-session serialization, and 30-minute inactivity retention inside the strategy.
- Register configured built-ins explicitly at LLM composition without replacing caller-provided strategies for the same policy class.

<details>
<summary>Walkthrough</summary>

#### Stack and ownership

```text
#11434 admission API
   -> #11530 queue host, lifecycle, live progress
      -> #11447 ThunderAgent strategy and composition
```

```mermaid
flowchart LR
  R["Tracked request"] --> H["Policy-class admission host"]
  H -->|"bypass / ready"| Q["Policy queue"]
  H -->|"defer"| D["Deferred queue entry"]
  D -->|"MakeReady"| Q
  Q --> S["Worker selector"]
  S --> W["Engine"]
  W -->|"atomic progress update"| P["RequestProgress"]
  W -->|"dispatch / finish"| H
  H --> T["dynamo-admission-control"]
  P -->|"read on reconcile"| T
  T -->|"30-minute inactivity"| X["Expire quiescent program"]
```

The queue actor owns request payloads and invokes a per-class strategy before ordinary worker selection. The strategy may bypass admission, make a request ready with optional placement, or defer it. It never owns scheduling requests, queue order, worker slots, or final selection.

#### State and ordering

[...]

PR #11434, opened 2026-07-08
feat(router): define queue admission strategy API

ishandhanani:
This defines the minimal public contract for pluggable policy-class queue admission strategies. It keeps request payloads and scheduler ownership private while giving independently packaged Rust strategies typed request facts, live worker eligibility, admission decisions, and lifecycle vocabulary.

CLOSES: DYN-3231

### How This Was Implemented

- Adds an object-safe `PolicyClassAdmissionStrategy` with one admission method, one lifecycle-event method, and an optional reconciliation interval.
- Exposes an opaque host-issued ID, optional session identity, full logical context size, and a retained read-only `WorkerEligibility` capability.
- Defines `Bypass | Ready | Defer`, exact-or-any placement, `Dispatched | Completed | Aborted | Reconcile`, and `MakeReady` without exposing `SchedulingRequest`.
- Makes `queue_admission` configuration strategy-neutral and documents the ownership and extension rules in root `CLAUDE.md`.

<details>
<summary>Walkthrough</summary>

#### Mental model

The router host will own request payloads, queues, lifecycle delivery, worker selection, and booking. A strategy owns only algorithmic state and communicates through this typed contract; this PR defines that boundary without implementing the host or a concrete algorithm.

```mermaid
sequenceDiagram
    participant H as "Router admission host"
    participant S as "PolicyClassAdmissionStrategy"
    H->>S: "admit(borrowed request facts)"
    S-->>H: "Bypass, Ready(placement), or Defer"
    H->>S: "Dispatched / Completed / Aborted / Reconcile"
    S-->>H: "zero or more MakeReady actions"
```

#### State and ordering

The router assigns a copyable `AdmissionId`; strategies use it instead of client-controlled request identity. `Bypass` opts out of lifecycle handling, `Ready(Any)` preserves ordinary routing, and `Ready(Exact(worker))` adds a constraint without replacing selector validation, scoring, or booking. Deferred work may retain the live eligibility capability and requests reconsideration through `MakeReady`; duplicate and unknown actions are harmless by contract.

#### Request lifecycle

A future host calls `admit` once per tracked request. A bypassed request receives no strategy events; a ready or deferred request receives at most one accepted dispatch and exactly one `Completed` or `Aborted` terminal event while the host remains alive. `Completed { context_tokens }` supplies the authoritative final logical context, while `Aborted` commits no new context. Reconciliation is an opportunity delivered by the host, not a strategy-owned thread.

#### Boundaries and limitations

[...]

PR #11527, opened 2026-07-10

ishandhanani:
### Summary

This stacked PR builds on #11434 by implementing the generic policy-class admission host and a typed live request-progress capability. It lets session-level strategies observe in-flight context growth without allocating request IDs or sending best-effort progress commands through the serialized scheduler actor.

CLOSES: DYN-3231

### How This Was Implemented

- Pair a strategy-retained `RequestProgress` reader with a response-stream-owned `RequestProgressUpdater` backed by a monotonic relaxed atomic.
- Carry the updater only for requests that entered a real admission lifecycle; bypassed and admission-disabled requests retain the ordinary scheduling path.
- Let the queue actor own one admission record across pending, selected, dispatched, and terminal states, with idempotent dispatch and terminal notifications.
- Retract deferred or ready queue entries immediately when their scheduling future is dropped, including queue-counter cleanup and one abort event.
- Keep strategy construction explicit through injectable policy-class strategy maps; this PR adds no ThunderAgent policy or policy-specific dependency.

<details>

#### Mental model

The queue actor creates a reader/writer pair when a strategy admits a tracked request. The strategy retains the reader, while ownership of the writer crosses the scheduling response into the existing response-stream guard.

```mermaid
flowchart LR
  Q -->|"write capability in scheduling response"| G["RequestGuard"]
  G -->|"atomic update at block boundary"| S
  G -->|"dispatched / completed / aborted"| Q
  Q -->|"lifecycle event"| S
```

#### State and ordering

`AdmissionRequest::context_tokens` remains the immutable initial context. `AdmissionRequest::progress` exposes the latest observed logical context; updater writes use `fetch_max`, so delayed or concurrent observations cannot move the value backward. Relaxed ordering is sufficient because this capability communicates only the numeric latest value and does not publish related memory.

The queue keeps one request-ID keyed lifecycle record whose worker is absent while pending and present after selection. Dispatch is emitted at most once, terminal completion or abort removes the record, and duplicate or stale notifications are ignored. Strategy actions are still applied actor-locally through the existing policy queue.

#### Request lifecycle

On success, the queue books scheduler state before delivering the response and then transfers the updater to `RequestGuard`. The guard publishes cumulative context only when output crosses an accounting-block boundary; normal stream completion publishes the authoritative observed total and emits `Completed`, while cancellation or backend failure emits `Aborted` without committing context.

Before response delivery, a cancellation guard owns the admission lifecycle. Dropping the scheduling future sends a cancellation command, removes either deferred or ready queue storage, subtracts aggregate and class counters, and emits one abort; after response delivery, the response-stream guard owns cleanup instead.

This is stacked on #11434 and intentionally contains no ThunderAgent algorithm, `session_final` behavior, or `dynamo-admission-control` dependency. The upstream PR targets `main` because #11434's head exists only in a fork; the displayed diff will include #11434 until that PR merges.

The capability and queue lifecycle are frontend-local. A third-party program table alone is not sufficient for multi-frontend correctness: distributed ThunderAgent would additionally need globally unique admission identities, per-session fencing or versioned CAS, stale-safe idempotent terminal updates, global capacity accounting, owner-routed wakeups, and frontend-death recovery. The separate O(retained sessions) strategy scan identified in #11447 is also outside this substrate change.

</details>

### Validation

- `cargo test -p dynamo-kv-router --lib --quiet` (639 passed)
- `cargo clippy -p dynamo-kv-router --lib --tests --benches -- -D warnings`
- `cargo check -p dynamo-llm --lib`
- `git diff --check`
- Not run: the `dynamo-llm` unit-test binary exceeded the available filesystem capacity while compiling test-only native dependencies; the LLM library check and clippy pass completed successfully.

### Benchmark Results

Release Criterion benchmark on the local AMD EPYC 9254 host. The prior actor-command baseline was measured separately on #11447 and included request-ID allocation, bounded-channel transport, actor lookup, and strategy dispatch.

| Operation | Time | Comparison |
|---|---:|---:|
| Atomic progress update | 2.525 ns | 87-222x lower than the accepted actor command |
| Atomic progress read | 0.486 ns | N/A |
| Prior accepted actor progress command | 220-560 ns | Baseline |

### DIFF
--- a/lib/kv-router/benches/policy_queue.rs
+++ b/lib/kv-router/benches/policy_queue.rs
@@ -14,7 +14,7 @@ use criterion::{
 use dynamo_kv_router::RouterQueuePolicy;
 use dynamo_kv_router::protocols::WorkerWithDpRank;
 use dynamo_kv_router::scheduling::{
-    PolicyProfile, PolicyQueue, QueueSnapshot, RouterPolicyConfig, WorkerPlacement,
+    PolicyProfile, PolicyQueue, QueueSnapshot, RequestProgress, RouterPolicyConfig, WorkerPlacement,
 };
 
 #[derive(Debug, Clone, Copy)]
@@ -218,6 +218,22 @@ fn bench_drain_shared(c: &mut Criterion) {
     group.finish();
 }
 
+fn bench_request_progress(c: &mut Criterion) {
+    let mut group = c.benchmark_group("request_progress");
+    let (progress, updater) = RequestProgress::new(0);
+    let mut context_tokens = 0usize;
+    group.bench_function("update", |b| {
+        b.iter(|| {
+            context_tokens += 1;
+            updater.update_context_tokens(black_box(context_tokens));
+        });
+    });
+    group.bench_function("read", |b| {
+        b.iter(|| black_box(progress.context_tokens()));
+    });
+    group.finish();
+}
+
 criterion_group! {
     name = benches;
     config = Criterion::default()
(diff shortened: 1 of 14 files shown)


ishandhanani:
Recreated as https://github.com/ishandhanani/dynamo/pull/2 so the PR can target #11434's fork-only head branch directly.

ishandhanani:
Superseded by upstream stacked PR #11530, based on `pull-request/11434`.

Issue #9741, opened 2026-05-19
operator/api: retire temporary v1alpha1 conversion shims for 1.3


sttts:
## Summary
Track the 1.3 cleanup for temporary v1alpha1 conversion compatibility shims. This should be broad enough to collect related conversion-removal TODOs as subtasks while keeping the current DGDR downgrade TODO visible.

## Current TODO
- `deploy/operator/api/v1alpha1/dynamographdeploymentrequest_conversion.go`: remove the legacy DGDR annotation downgrade shims after 1.2 when downgrade compatibility no longer matters.
- `deploy/operator/api/CONVERSION.md`: update the legacy DGDR exception guidance once those shims are removed.

## Subtasks
- [ ] Confirm the 1.3 compatibility policy for dropping 1.2 downgrade shims.
- [ ] Remove legacy DGDR annotation constants and related read/write helpers that are no longer needed.
- [ ] Remove or simplify legacy DGDR conversion tests and old-converter oracle coverage where it only exists for downgrade compatibility.
- [ ] Update `CONVERSION.md` to remove the temporary legacy DGDR exception language.
- [ ] Search for any other temporary v1alpha1 conversion TODOs that should be bundled into the same 1.3 cleanup.
- [ ] Check the DGD annotations used to avoid rollouts to DCDs, some hash over the spec. It think we were still stamping the old annotation on every object, plus the new v2 one, but believe the v1 annotation until the next spec change.

## References
- TODO: `deploy/operator/api/v1alpha1/dynamographdeploymentrequest_conversion.go`
- Conversion policy: `deploy/operator/api/CONVERSION.md`

sttts:
## Proposed implementation plan

I looked through the DGDR conversion code, the DGD/DCD structural conversion paths, and the DGD/DCD reconcilers. The cleanup should separate rollback-only compatibility from conversion fidelity that is still required while v1alpha1 and v1beta1 are both served.

### 1. Remove DGDR legacy downgrade annotations

Target: `deploy/operator/api/v1alpha1/dynamographdeploymentrequest_conversion.go`.

Remove the temporary Dynamo 1.0/1.1 annotation compatibility layer:

[...]

PR #11529, opened 2026-07-10
refactor(operator): lazily converge worker hashes to v2

sttts:
## Summary

- stop computing the v1alpha1 worker hash for normal dual-hash or canonical state; retain it only for incomplete pre-dual-hash/`legacy` migrations
- preserve unchanged Dynamo 1.2 worker generations by treating `current-worker-hash` as their opaque active identity and `current-worker-hash-v2` as the matching v2 fingerprint
- preserve the dual-hash bridge when stale `InProgress` status is completed, preventing the following reconcile from starting a second rollout
- converge to a canonical v2-only generation only after a real v2 worker-spec rollout completes
- keep leftover DCDs labeled `nvidia.com/dynamo-worker-hash=legacy` from changing the active generation or triggering another rollout

This targets `main` for Dynamo 1.4. It deliberately avoids an operator-upgrade-induced worker rollout and is not intended for the 1.3 release branch.

Part of #9741

Linear: [DGH-895](https://linear.app/nvidia/issue/DGH-895/operatorapi-retire-temporary-v1alpha1-conversion-shims-for-13)

## Validation

- `go test ./api/...`
- `KUBEBUILDER_ASSETS=deploy/operator/bin/k8s/1.30.0-darwin-arm64 go test ./internal/...`
- `go vet ./api/... ./internal/...`
- `git diff --check`
- repository pre-commit hooks


deploy/operator/internal/controller/dynamographdeployment_controller_test.go
```diff
@@ -4237,7 +4237,7 @@ func Test_reconcileDynamoComponentsDeployments(t *testing.T) {
 				}),
 				betaDCD(t, &v1alpha1.DynamoComponentDeployment{
 					ObjectMeta: metav1.ObjectMeta{
-						Name:      "test-dgd-decode-e1f2a6fe",
+						Name:      "test-dgd-decode-1b69c0d3",
```
sttts, line 4242: A note for the reviewer: this is intentionally the v2-named DCD for a fresh 1.4 DGD. The adapted test setup now models the real startup ordering: the DGD canonical v2 annotation and matching DCD and pod worker-hash labels already exist before reconcileDynamoComponentsDeployments runs. Existing 1.2 DGDs retain their v1-named DCD when v2 is unchanged, as covered by TestLegacyAlphaHashCompatibility_NoOpUpgradeUsesExistingWorkerGeneration; this fixture change does not represent an upgrade rollout.

deploy/operator/internal/controller/dynamographdeployment_controller_test.go
```diff
@@ -4593,6 +4593,24 @@ func Test_reconcileDynamoComponentsDeployments(t *testing.T) {
 				Spec: tt.dgdSpec,
 			})
 
+			// Full reconciliation initializes a fresh DGD with its v2 hash before calling this helper.
```
sttts: this whole block I don't understand. Doesn't it fake the actual testing??! Why do we teak the fixtures?

sttts: Agreed. Mutating every fixture after construction obscured what the test was actually exercising. I removed that block. Only the two adapted fresh-v2 cases now declare the canonical DGD annotation and matching worker hash labels directly in their fixture data; the other cases are untouched. The targeted test passes in 564845b2d5.

deploy/operator/internal/controller/dynamographdeployment_rollingupdate.go
```diff
@@ -55,16 +55,23 @@ func (h workerGenerationHashes) contains(hash string) bool {
...
 	if err != nil {
 		return workerGenerationHashes{}, fmt.Errorf("failed to compute v2 worker hash: %w", err)
 	}
 
+	current := r.currentWorkerHashes(dgd)
+	v1Hash := v2Hash
+	if current.v2 != "" {
+		v1Hash = current.v1
+	}
+	if current.v2 == "" && current.v1 != "" && current.v1 != v2Hash {
```
sttts, line 68: A fresh or completed canonical generation has current.v1 set to the old v2 hash and current.v2 empty. After a worker-spec change, this condition computes the new legacy v1 hash; downstream selection rolls to that identity and completion restores dual-hash bridge state instead of remaining canonical v2. Keep legacy computation confined to proven pre-dual migration state or the explicit legacy sentinel, and add a regression covering fresh canonical initialization -> spec mutation -> new v2 selection -> canonical completion.

sttts (commented):
Static code review at 564845b2. One blocking state-machine issue remains. The annotation contract comments should also be updated because current-worker-hash now carries canonical v2 after convergence.

deploy/operator/internal/controller/dynamographdeployment_rollingupdate.go
sttts: Fixed in 34281ace11. A v1-only state is treated as pre-dual only when its value equals the computed legacy hash or is the explicit legacy sentinel. Existing 1.2 bridge state remains dual and unchanged, so upgrade does not roll workers. After a real v2 rollout completes, the canonical DGD state is v2-only: current-worker-hash is removed and current-worker-hash-v2 stores the active DCD worker-hash value. The lifecycle test covers fresh v2-only initialization, spec mutation, new-v2 selection, and v2-only completion. Validation: go test ./internal/... -count=1 and go vet ./internal/....

julienmancuso (approved):

PR #11530, opened 2026-07-10

ishandhanani:
Stacked on #11434, this implements the generic policy-class admission host and a typed live request-progress capability. It lets strategies observe in-flight logical context without allocating request IDs or sending best-effort progress commands through the serialized scheduler actor.

CLOSES: DYN-3231

### How This Was Implemented

- Pair a strategy-retained `RequestProgress` reader with a response-stream-owned `RequestProgressUpdater` backed by a monotonic relaxed atomic.
- Carry the updater only for requests that entered a real admission lifecycle; bypassed and admission-disabled requests retain the ordinary scheduling path.
- Let the queue actor own one admission record across pending, selected, dispatched, and terminal states, with idempotent dispatch and terminal notifications.
- Retract deferred or ready queue entries immediately when their scheduling future is dropped, including queue-counter cleanup and one abort event.
- Keep strategy construction explicit through injectable policy-class strategy maps; this PR adds no concrete admission policy.

<details>

#### Mental model

The queue actor creates a reader/writer pair when a strategy admits a tracked request. The strategy retains the reader, while ownership of the writer crosses the scheduling response into the existing response-stream guard.

```mermaid
flowchart LR
  Q -->|"write capability"| G["RequestGuard"]
  G -->|"atomic block-boundary update"| S
  G -->|"dispatched / completed / aborted"| Q
  Q -->|"lifecycle event"| S
```

#### State and ordering

`AdmissionRequest::context_tokens` remains the immutable initial context. `AdmissionRequest::progress` exposes the latest observed logical context; updater writes use `fetch_max`, so delayed or concurrent observations cannot move the value backward. Relaxed ordering is sufficient because the capability communicates only the numeric latest value and does not publish related memory.

The queue keeps one request-ID-keyed lifecycle record whose worker is absent while pending and present after selection. Dispatch is emitted at most once, terminal completion or abort removes the record, and duplicate or stale notifications are ignored. Strategy actions remain actor-local and reuse the existing policy queue.

#### Request lifecycle

On success, the queue books scheduler state before delivering the response and transfers the updater to `RequestGuard`. The guard publishes cumulative context only when output crosses an accounting-block boundary; normal stream completion publishes the authoritative observed total and emits `Completed`, while cancellation or backend failure emits `Aborted` without committing context.

Before response delivery, a cancellation guard owns the admission lifecycle. Dropping the scheduling future sends a cancellation command, removes deferred or ready queue storage and counters, and emits one abort; after response delivery, the response-stream guard owns cleanup instead.

This contains no ThunderAgent algorithm, client-specific terminal-session behavior, or `dynamo-admission-control` dependency. The capability and queue lifecycle are frontend-local; distributed admission would additionally require globally unique identities, per-session fencing, stale-safe idempotent terminal updates, global capacity accounting, owner-routed wakeups, and frontend-death recovery.

</details>

### Validation

- `cargo test -p dynamo-kv-router --lib --quiet` (639 passed)
- `cargo clippy -p dynamo-kv-router --lib --tests --benches -- -D warnings`
- `cargo check -p dynamo-llm --lib`
- `git diff --check`

### Benchmark Results

Release Criterion benchmark on an AMD EPYC 9254. The prior actor-command baseline included request-ID allocation, bounded-channel transport, actor lookup, and strategy dispatch.

| Operation | Time | Comparison |
|---|---:|---:|
| Atomic progress update | 2.525 ns | 87-222x lower than the accepted actor command |
| Atomic progress read | 0.486 ns | N/A |
| Prior accepted actor progress command | 220-560 ns | Baseline |

### DIFF
--- a/lib/kv-router/src/scheduling/queue.rs
+++ b/lib/kv-router/src/scheduling/queue.rs
@@ -688,6 +688,15 @@ impl<
             (class_index, Some(snapshot))
         };
 
+        if let Some(request_id) = request.mode.tracked_request_id()
+            && self.tracked_admissions.contains_key(request_id)
+        {
+            request.respond(Err(KvSchedulerError::BookingFailed(format!(
+                "request {request_id} already has an active admission"
+            ))));
+            return false;
+        }
+
         let mut admission = if request.mode.is_tracked() && self.admission.has_strategy(class_index)
         {
             let allowed_worker_ids = request.allowed_worker_ids.clone();
@@ -877,7 +886,17 @@ impl<
     }
 
     fn handle_cancelled(&mut self, request_id: &str) -> bool {
-        self.handle_finished(request_id, RequestOutcome::Aborted)
+        let rollback_booking = self
+            .tracked_admissions
+            .get(request_id)
+            .is_some_and(|tracked| tracked.worker.is_some() && !tracked.dispatched);
+        if rollback_booking
+            && let Err(error) = self.slots.free(&request_id.to_owned(), Instant::now())
+        {
+            tracing::error!(%request_id, %error, "Failed to roll back cancelled scheduler booking");
+        }
+        let made_ready = self.handle_finished(request_id, RequestOutcome::Aborted);
+        rollback_booking || made_ready
     }
 
     fn handle_finished(&mut self, request_id: &str, outcome: RequestOutcome) -> bool {
@@ -2169,6 +2188,75 @@ policy_classes:
             }
             Vec::new()
         }
+    }
+
+    #[tokio::test]
+    async fn cancellation_after_response_send_rolls_back_booking() {
+        let state = Arc::new(StdMutex::new(GateState::default()));
+        let (queue, slots) = make_queue_with_admission_strategy(Box::new(ReadyGate {
+            state: Arc::clone(&state),
+        }));
+        let (mut request, response) = make_request("cancelled-after-handoff", 64);
+        request.policy_class = Some("agents".to_owned());
+        let cancellation = queue
+            .cancellation_guard(Some("cancelled-after-handoff"))
+            .unwrap();
+
+        queue.enqueue(request).await;
+        assert_eq!(
+            slots
+                .active_request_counts()
+                .get(&WorkerWithDpRank::new(0, 0))
+                .copied(),
+            Some(1)
+        );
+
+        drop(cancellation);
+        tokio::time::timeout(Duration::from_secs(1), async {
+            while state.lock().unwrap().aborted.len() != 1 {
+                tokio::task::yield_now().await;
+            }
+        })
+        .await
+        .expect("cancellation did not abort admission");
+
+        slots.assert_completely_drained(decay_now());
+        assert_eq!(state.lock().unwrap().aborted, vec![AdmissionId::new(0)]);
+        drop(response);
+    }
+
+    #[tokio::test]
+    async fn duplicate_request_id_does_not_replace_active_admission() {
+        let state = Arc::new(StdMutex::new(GateState::default()));
+        let (queue, slots) = make_queue_with_admission_strategy(Box::new(ReadyGate {
+            state: Arc::clone(&state),
+        }));
+        let (mut first, first_response) = make_request("duplicate", 64);
+        first.policy_class = Some("agents".to_owned());
+        queue.enqueue(first).await;
+        first_response.await.unwrap().unwrap();
+
+        let (mut duplicate, duplicate_response) = make_request("duplicate", 64);
+        duplicate.policy_class = Some("agents".to_owned());
+        queue.enqueue(duplicate).await;
+        assert!(matches!(
+            duplicate_response.await.unwrap(),
+            Err(KvSchedulerError::BookingFailed(message))
+                if message == "request duplicate already has an active admission"
+        ));
+        assert_eq!(
+            slots
+                .active_request_counts()
+                .get(&WorkerWithDpRank::new(0, 0))
+                .copied(),
+            Some(1)
+        );
+        assert!(state.lock().unwrap().aborted.is_empty());
+
+        queue.finish("duplicate", RequestOutcome::Aborted).await;
+        queue.update().await;
+        assert_eq!(state.lock().unwrap().aborted, vec![AdmissionId::new(0)]);
+        slots.free(&"duplicate".to_owned(), decay_now()).unwrap();
     }
 
     struct BypassGate {
--- a/lib/kv-router/src/scheduling/queue.rs
+++ b/lib/kv-router/src/scheduling/queue.rs
@@ -20,9 +20,9 @@ use super::policy_config::{PolicyClassConfig, PolicyProfile};
 use super::policy_queue::{PolicyQueue, QueueSnapshot};
 use super::prefill_load::{PrefillLoadEstimator, effective_prefill_tokens};
 use super::queue_admission::{
-    AdmissionAction, AdmissionDecision, AdmissionTicket, PolicyClassAdmissionController,
-    PolicyClassAdmissionStrategies, RequestProgressUpdater, WorkerEligibility,
-    WorkerEligibilitySnapshot, WorkerPlacement,
+    AdmissionAction, AdmissionDecision, AdmissionTicket, ClassAdmissionAction,
+    PolicyClassAdmissionController, PolicyClassAdmissionStrategies, RequestProgressUpdater,
+    WorkerEligibility, WorkerEligibilitySnapshot, WorkerPlacement,
 };
 use super::selector::{DefaultWorkerSelector, WorkerSelector};
 use super::types::{
@@ -892,7 +892,12 @@ impl<
 
     fn handle_cancelled(&mut self, request_id: &str) -> bool {
         let Some(tracked) = self.tracked_admissions.get(request_id).copied() else {
-            return false;
+            let request_id = request_id.to_owned();
+            let rollback_booking = self.slots.request_worker(&request_id).is_some();
+            if rollback_booking && let Err(error) = self.slots.free(&request_id, Instant::now()) {
+                tracing::error!(%request_id, %error, "Failed to roll back cancelled scheduler booking");
+            }
+            return rollback_booking;
         };
         let rollback_booking = tracked.worker.is_some() && !tracked.dispatched;
         if rollback_booking
@@ -968,7 +973,7 @@ impl<
 
     fn apply_admission_actions(
         &mut self,
-        actions: impl IntoIterator<Item = super::queue_admission::ClassAdmissionAction>,
+        actions: impl IntoIterator<Item = ClassAdmissionAction>,
     ) -> bool {
         let mut made_ready = false;
         let mut actions: VecDeque<_> = actions.into_iter().collect();
@@ -1037,13 +1042,19 @@ impl<
         for class_index in std::mem::take(&mut self.cancelled_ready_classes) {
             let tracked_admissions = &self.tracked_admissions;
             let cancelled = self.pending.take_if_in_class(class_index, |queued| {
-                queued.admission.is_some()
-                    && queued.request.response_is_closed()
+                let Some(admission) = queued.admission.as_ref() else {
+                    return false;
+                };
+                queued.request.response_is_closed()
                     && queued
                         .request
                         .mode
                         .tracked_request_id()
-                        .is_some_and(|request_id| !tracked_admissions.contains_key(request_id))
+                        .is_some_and(|request_id| {
+                            tracked_admissions
+                                .get(request_id)
+                                .is_none_or(|tracked| tracked.ticket != admission.ticket)
+                        })
             });
             for entry in cancelled {
                 self.subtract_pending_counters(class_index, entry.snapshot());
@@ -1136,12 +1147,13 @@ impl<
             let request = queued.request;
             // A ready cancellation removes lifecycle state before its batched
             // heap cleanup; an intervening capacity update may pop it first.
-            if admission.is_some()
-                && request
-                    .mode
-                    .tracked_request_id()
-                    .is_some_and(|request_id| !self.tracked_admissions.contains_key(request_id))
-            {
+            if admission.as_ref().is_some_and(|admission| {
+                request.mode.tracked_request_id().is_some_and(|request_id| {
+                    self.tracked_admissions
+                        .get(request_id)
+                        .is_none_or(|tracked| tracked.ticket != admission.ticket)
+                })
+            }) {
                 continue;
             }
             tracing::debug!(
@@ -2382,6 +2394,33 @@ policy_classes:
         slots.free(&"unmanaged".to_owned(), decay_now()).unwrap();
     }
 
+    #[tokio::test]
+    async fn cancellation_after_bypassed_handoff_rolls_back_booking() {
+        let events = Arc::new(AtomicUsize::new(0));
+        let (queue, slots) = make_queue_with_admission_strategy(Box::new(BypassGate {
+            events: Arc::clone(&events),
+        }));
+        let (mut request, response) = make_admission_request("bypassed-handoff", 64);
+        request.policy_class = Some("agents".to_owned());
+        let cancellation = queue.cancellation_guard(Some("bypassed-handoff")).unwrap();
+
+        queue.enqueue(request).await;
+        assert_eq!(
+            slots
+                .active_request_counts()
+                .get(&WorkerWithDpRank::new(0, 0))
+                .copied(),
+            Some(1)
+        );
+
+        drop(cancellation);
+        queue.update().await;
+
+        slots.assert_completely_drained(decay_now());
+        assert_eq!(events.load(Ordering::Relaxed), 0);
+        drop(response);
+    }
+
     #[derive(Default)]
     struct FinishReleaseGate {
         first: Option<AdmissionId>,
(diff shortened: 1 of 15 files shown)


lib/kv-router/src/scheduling/queue.rs
```diff
@@ -488,73 +672,180 @@ impl<
...
                 .profile
                 .resolve_class_index(request.policy_class.as_deref(), snapshot.uncached_tokens);
-            let class = self.profile.class(class_index);
-            let should_queue = self.should_queue(class_index, class, || {
-                Self::all_workers_prefill_busy_with(&active_tokens, &workers, class, eligibility)
+            (class_index, Some(snapshot))
+        };
+
+        if let Some(request_id) = request.mode.admission_request_id()
+            && self.tracked_admissions.contains_key(request_id)
```
PeaBrane, line 696: [AI] I’m treating one active request per request ID as a strong router invariant, so I’m not asking for generation-keyed cancellation here. Could we document that invariant in the repository’s `CLAUDE.md` or scheduling guidance? It is doing real correctness work in this lifecycle, and making it explicit should prevent future automated reviews from repeatedly flagging request-ID-only cancellation as an unsupported ABA case.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -350,26 +358,32 @@ impl KvPushRouter {
...
+                        failed |= response_item_failed(&item);
                         guard.on_item(&item).await;
                         yield item;
                     }
                 }
-            }
+            };
 
-            guard.finish().await;
+            if completed && !failed {
```
PeaBrane, line 382: [AI] A successful admission is reported `Completed` only after the backend stream returns `None`. We yield the terminal `Stop`/`EoS`/`Length` item first, so a consumer that treats that item as completion and drops without polling EOF causes `RequestCleanup::drop` to send `Aborted`. That loses authoritative context despite backend success. Could the guard retain successful-terminal state before yielding so dropping after the final item still completes the admission lifecycle?

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -488,73 +672,180 @@ impl<
...
+                    (RequestAdmission { ticket, progress }, decision)
+                })
+        } else {
+            None
         };
+        let mut deferred = false;
+        if let Some((request_admission, decision)) = admission.as_ref() {
+            match decision {
+                AdmissionDecision::Bypass => admission = None,
+                AdmissionDecision::Ready(placement) => {
```
PeaBrane, line 752: [AI] For family-resolved requests, the class and queue snapshot are computed before strategy admission. If `Ready(Exact)` pins a different worker, enqueue reuses the best-cache snapshot from the pre-pin candidate set. Deferred `MakeReady(Exact)` similarly moves the immutable entry without rekeying it. This can choose the wrong cache bucket and mischarge cached-token counters, WSPT priority, and DRR cost. Could exact placement trigger recomputation/reclassification, or explicitly separate the classification snapshot from the queue-cost snapshot?

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -583,6 +874,197 @@ impl<
...
+        {
+            self.tracked_admissions.remove(request_id);
+            self.subtract_pending_counters(entry.class_index(), entry.snapshot());
+            return self.abort_admission(tracked.ticket);
+        }
+
+        // Ready heaps do not have an admission-ID index. Batch cancellations so
+        // one reconcile scans each affected class once instead of once per drop.
+        self.tracked_admissions.remove(request_id);
+        self.cancelled_ready_classes
```
PeaBrane, line 925: [AI] This removes lifecycle state but leaves the ready heap entry and its counters until reconciliation. With the default 60-second interval and no admission action producing runnable work, canceled entries can keep `has_backlog` true and consume queue limits, potentially causing stale `QueueRejected`. The PR description also says ready-entry retraction and counter cleanup are immediate. Can accounting be released immediately, perhaps using an indexed removal or an accounted tombstone whose storage is reclaimed later?

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -65,6 +76,59 @@ enum AdmissionCommand {
...
+        let command = AdmissionCommand::Cancelled { request_id };
+        match self.admission_tx.try_send(command) {
+            Ok(()) | Err(mpsc::error::TrySendError::Closed(_)) => {}
+            Err(mpsc::error::TrySendError::Full(command)) => {
+                let Ok(handle) = tokio::runtime::Handle::try_current() else {
+                    tracing::warn!("No tokio runtime for queued admission cancellation");
+                    return;
+                };
+                let admission_tx = self.admission_tx.clone();
+                handle.spawn(async move {
```
PeaBrane, line 126: [AI] The actor channel is bounded, but this full-channel fallback creates one Tokio task per canceled request waiting on that same channel. A slow strategy callback plus a cancellation storm can therefore retain unbounded tasks, sender clones, and request-ID strings precisely while the actor is overloaded. Could cancellation use a bounded cleanup channel or coalesced tombstone mechanism instead?

lib/llm/src/kv_router/scheduler.rs
```diff
@@ -83,6 +114,16 @@ where
         let profile = kv_router_config
             .policy_profile(model_name)
             .map_err(|error| KvSchedulerError::InitFailed(error.to_string()))?;
+        let strategy_recheck_interval = admission_strategies
```
PeaBrane, line 117: [AI] Taking the global minimum means one 10 ms strategy drives `Reconcile` for every strategy and global queue housekeeping at 100 Hz. Whenever pending or deferred work exists, this can also recheck all worker lanes at that cadence. Could the controller track per-strategy deadlines and perform global draining only when cleanup or returned actions require it?

PeaBrane (commented):
AI-assisted review focused on admission lifecycle semantics, queue accounting, and overhead. The underlying DRR algorithm remains unchanged; these comments cover new lifecycle edges and cost inputs.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -122,11 +124,16 @@ impl KvPushRouter {
...
+                        .finish(&context_id, RequestOutcome::Aborted)
+                        .await
+                {
                     tracing::warn!(
                         request_id = %context_id,
                         %error,
-                        "Failed to free scheduler state after cancellation during worker selection"
+                        "Failed to finish scheduler state after cancellation during worker selection"
                     );
                 }
```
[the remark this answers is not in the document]
PeaBrane: Thanks, but I don’t think the reported warning/error behavior holds on the current head. `ActiveSequencesMultiWorker::free` explicitly treats `RequestNotFound` as an idempotent repeat and returns `Ok`, emitting only a debug trace. The cleanup paths are redundant, but the second free does not reach the warning/error branches cited here. I’m treating this as a false positive.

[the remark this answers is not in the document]
PeaBrane: Thanks, but I don’t think the reported warning/error behavior holds on the current head. `ActiveSequencesMultiWorker::free` explicitly treats `RequestNotFound` as an idempotent repeat and returns `Ok`, emitting only a debug trace. The cleanup paths are redundant, but the second free does not reach the warning/error branches cited here. I’m treating this as a false positive.

lib/kv-router/src/scheduling/queue.rs
PeaBrane: Stepping back, I think these cancellation issues may point to one lifecycle design rather than separate fixes.

Could `SchedulingCancellationGuard` and `RequestGuard` be successive owners of a single `AdmissionLease`, backed by shared cancel/terminal state, with the actor remaining the only owner of queue accounting, booking rollback, and strategy callbacks?

Dropping either phase’s lease would atomically mark `CancelRequested` and enqueue the ticket into a one-per-active dirty set, then issue a coalesced wake-up—coalescing the wake, not the identities.

The actor would drain dirty tickets, release ready/backlog/request-token accounting immediately, and mark the heap entry already-accounted for lazy reclamation. Candidate selection and `has_backlog` would need to ignore those tombstones, and eventual heap removal must not decrement twice.

That would unify the pre/post-selection handoff, remove task-per-cancellation overflow and duplicate cleanup authority, and close the 60-second stale-accounting window. It would still preserve the two RAII scopes—the same lease simply moves across the scheduling-response handoff.

Since these issues share the same ownership boundary, I’d prefer to fold this lifecycle shape into the current change rather than address them as separate fixes.

PeaBrane:
For shared context, here is a general mental checklist for cancellation and cleanup duties, ordered by correctness impact:

| Priority | Component/state | Current status | Assessment |
|---:|---|---|---|
| 1 | Worker scheduler booking | **Done well** | Selected-but-not-dispatched cancellation calls `slots.free()`. Duplicate cleanup is redundant but safely idempotent. |
| 2 | Logical queue accounting | **Partial** | Deferred cancellation retracts counters immediately. Ready cancellation leaves counters and `has_backlog` charged until reconciliation—up to 60 seconds by default. |
| 3 | Admission strategy state | **Partial, strong foundation** | Strategy callbacks are actor-serialized and the ticket is normally terminated once. However, successful terminal output can still be reported as `Aborted`, so exactly-once delivery does not guarantee the correct outcome. |
| 4 | Admission lifecycle record | **Partial, strong foundation** | One actor-owned record and idempotent dispatch/terminal handling are good. The pre/post-selection cleanup paths remain separate, and successful-terminal state does not survive the final yield. |
| 5 | Queue storage | **Partial** | Deferred entries have indexed immediate removal. Ready heaps have no admission-ID index and rely on later reconciliation; there is no immediately accounted tombstone yet. |
| 6 | Progress/terminal context | **Partial** | Monotonic `fetch_max` progress is good. Normal EOF commits authoritative context, but dropping after the successful terminal item incorrectly aborts without committing it. |
| 7 | Wake and transport machinery | **Not adequately handled** | The main channel is bounded, but saturation spawns one waiting Tokio task per cancellation. There is no coalesced wake plus retained dirty-ticket mechanism. |
| 8 | Metrics and tracing | **Partial** | Ordinary paths update bookkeeping, but metrics inherit stale ready counters and incorrect `Completed`/`Aborted` classification from the lifecycle issues above. |

ishandhanani:
Follow-up performance validation for the request-progress/admission host:

### Admission disabled vs stacked base

Three interleaved A/B pairs with a fresh frontend and four fresh mock workers per sample. Release builds, KV block size 64, concurrency 256, 256 warmup + 2,048 measured requests, ~60.5k input tokens and 8 output tokens.

| Arm | Throughput | Request p50 / p99 | TTFT p50 / p99 | Router scheduling p50 / p99 |
|---|---:|---:|---:|---:|
| #11434 base (`e2962f8dfd`) | 339.1 req/s | 558 / 2,769 ms | 497 / 2,769 ms | 0.177 / 138.5 ms |
| #11530, admission disabled (`e9d51f0ace`) | 349.0 req/s | 540 / 2,408 ms | 491 / 2,201 ms | 0.180 / 103.0 ms |

Per-run throughput was `[339.1, 357.2, 302.7]` for the base and `[349.0, 356.1, 328.9]` for this PR. This supports **no measurable end-to-end regression when admission is disabled**. Router scheduling p50 differs by 0.003 ms; tail/mean values varied more between samples than between arms.

### Enabled strategy ceiling

A benchmark-only actor test scheduled 262,144 requests/sample at concurrency 256. Arm order rotated, the first complete rotation was discarded, and the table reports medians of the remaining six samples. The temporary code was removed after measurement.

| Arm | Throughput | Scheduling p50 | Scheduling p99 |
|---|---:|---:|---:|
| Disabled | 277,504 req/s | 772 µs | 1,442 µs |
| Always-Bypass | 226,514 req/s (-18.4%) | 1,176 µs | 1,578 µs |
| Always-Ready | 220,493 req/s (-20.5%) | 1,161 µs | 1,691 µs |

So admission is not free in isolation, but the enabled ceiling remains above 220k req/s—over 600x the measured mocker E2E rate.

### Concurrent progress + cancellation/deferred churn

Five samples per arm, 262,144 deferred enqueue/cancel lifecycles/sample at concurrency 256:

| Arm | Churn throughput | p50 | p99 | Progress traffic |
|---|---:|---:|---:|---:|
| No progress | 559,018 req/s | 408 µs | 821 µs | 0 |
| 4 writers + 4 readers | 313,160 req/s | 787 µs | 944 µs | 44.7M atomic ops/s |

The progress arm intentionally saturates the same eight-CPU pool as the actor, so the throughput reduction is CPU competition rather than lock/queue collapse. Even under that stress, cancellation/deferred churn sustains 313k req/s with sub-1 ms p99.

My conclusion: the evidence supports “no measurable overhead when disabled,” while acknowledging the ~18–21% isolated actor-ceiling cost for enabled strategies. I do not see a performance change justified by these numbers.

ishandhanani:
Follow-up on Peabrane's lifecycle review and the performance concern:

All six actionable review areas are addressed in the local follow-up, but these commits are **not pushed** and this PR is currently closed. I am leaving the review threads unresolved until the code is visible on GitHub.

The implementation now:
- documents request-ID uniqueness and rejects duplicate active admission IDs;
- transfers one `AdmissionLease` from selection into the response stream;
- records a successful terminal item before yielding it;
- immediately retracts canceled deferred, ready, and legacy queued requests with actor-owned accounting;
- recomputes exact-placement classification, cached-token accounting, WSPT priority, and DRR cost;
- uses a lock-free dirty queue plus one coalesced actor wake, with per-strategy reconcile deadlines;
- skips both lease creation and cleanup polling when router queueing is disabled.

I reran a full old/new ABBA matrix against PR head `e9d51f0ace`: two ABBA cycles per cell, four samples/revision, fresh process per sample, CPUs 0-7, 131,072 lifecycles/sample. The final no-queue control used 1,048,576 lifecycles/sample.

| Pattern | Progress | Concurrency | Capacity | Throughput | p99 |
|---|---|---:|---:|---:|---:|
| Default/no queue | Off | 1 | 65,536 | +4.0% | 5→5 us |
| Default/no queue | Off | 256 | 65,536 | +3.7% | -0.7% |
| Deferred | Off / On | 256 | 65,536 | +33.9% / +76.3% | -17.7% / -93.6% |
| Ready/rollback | Off / On | 256 | 65,536 | +34.2% / +18.5% | -24.0% / -0.9% |
| Mixed | Off / On | 256 | 65,536 | +27.9% / +11.0% | -18.8% / -7.1% |
| Mixed | Off / On | 32 | 65,536 | +0.1% / +9.6% | +5.2% / -12.3% |
| Forced saturation | Off / On | 256 | 1 | +82.9% / +55.9% | -46.9% / -36.8% |

Current passed all 56 matrix samples. Old passed 55/56 and reproduced the capacity-one pending-entry leak. The previously observed mixed c32/progress-on p99 increase did not repeat; p99 improved 12.3% in this ABBA run. The default/no-queue path now performs neither lease work nor cleanup polling, and its long control shows no regression.

Validation: router tests 655/655; strict router clippy, rustfmt, and diff checks pass.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Addressed locally. I added the invariant to `lib/kv-router/src/scheduling/CLAUDE.md`: one request ID identifies at most one active scheduler request and cannot be reused before terminal cleanup. The actor also rejects a duplicate active admission ID, with focused duplicate/reuse tests. I am leaving this thread unresolved until the follow-up is pushed.

lib/llm/src/kv_router/push_router.rs
ishandhanani: Addressed locally. Selection and streaming now own successive phases of one `AdmissionLease`. The response wrapper marks the lease completed before yielding a successful Stop/EoS/Length terminal item, so dropping immediately after that item reports `Completed` with the latest context rather than `Aborted`. Focused terminal-drop coverage passes. Leaving unresolved until pushed.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Addressed locally. Both immediate `Ready(Exact)` and deferred `MakeReady(Exact)` recompute the worker-specific queue snapshot after placement, then reclass/rekey cached-token counters, WSPT priority, and DRR cost before enqueue/release. Exact-placement and deferred-rekey tests cover the behavior. Leaving unresolved until pushed.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Addressed locally. Cleanup is actor-owned and retracts deferred entries directly; ready cancellations are batched by affected class and removed with one heap rebuild, with pending/class counters released in the same actor turn. Legacy queued requests use the same immediate accounting path. The full ABBA matrix ended with zero pending state in every current sample. Leaving unresolved until pushed.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Addressed locally using the unified lifecycle you suggested. One transferable `AdmissionLease` queues the request identity/outcome into a lock-free dirty queue; an `AtomicBool` coalesces only the wake. The actor remains the sole owner of queue accounting, booking rollback, and strategy callbacks, and bounded-batch fallback draining preserves progress when the channel is saturated. Current passed every capacity-one ABBA sample; old reproduced a pending-entry leak. Leaving unresolved until pushed.

lib/llm/src/kv_router/scheduler.rs
ishandhanani: Addressed locally. The controller now tracks a separate next-reconcile deadline per strategy and only invokes strategies that are due. Global queue rechecks retain their own deadline and run only when due or when cleanup/actions make work ready. The focused deadline test and full router suite pass. Leaving unresolved until pushed.

ishandhanani:
Replacement PR targeting `main`: #11615. The queue API dependency is now merged, so the implementation has been replayed directly onto current `main`; the review discussion and ABBA results here remain relevant context.

PR #11539, opened 2026-07-10
perf(runtime): avoid copying MessagePack event payloads

jthomson04:
## Summary

- Decode MessagePack event-envelope payloads as slices of their received `Bytes` frame, avoiding the receive-side `Vec<u8>` copy.
- Use an allocation-free identity decoder in deduplication and ZMQ publisher framing.
- Preserve the public event-envelope API, typed-event behavior, and wire format.

## Performance

A local AgentX/Weka c256 frontend profile on the equivalent envelope-decoding change reduced `Codec::decode_envelope` inclusive samples from 5.07% to 0.16% (-96.8%) and eliminated the envelope payload `VecVisitor` subtree (4.70% to 0.00%). Overall serialization fell from 14.55% to 9.48%.

The comparison used a retained control rather than a fresh interleaved A/B, and the mocker/client partition was CPU-saturated. Treat its CPU/request change (11.818 ms to 11.679 ms) as diagnostic, not a capacity claim.

## Validation

- `cargo fmt --all --check`
- `cargo test -p dynamo-runtime msgpack_codec --lib`
- Local AgentX/Weka on-CPU profile: block size 1, frontend CPUs 0-1, eight SGLang FIFO mockers plus AIPerf on CPUs 2-23, 40-second DWARF capture at 99 Hz

## Summary by CodeRabbit

* **Performance**
  * Improved event processing by extracting publisher and sequence metadata without fully decoding message topics and payloads.
  * Reduced unnecessary memory allocation during event decoding.
  * Decoded payloads can now reference the original encoded message data.

* **Reliability**
  * Added validation to reject malformed or out-of-bounds encoded envelopes.
  * Improved handling of duplicate event detection and message publishing metadata.

### DIFF
--- a/lib/runtime/src/transports/event_plane/codec.rs
+++ b/lib/runtime/src/transports/event_plane/codec.rs
@@ -3,9 +3,9 @@
 
 //! Event plane codec for serializing/deserializing envelopes and payloads.
 
-use anyhow::Result;
+use anyhow::{Result, anyhow, bail};
 use bytes::Bytes;
-use serde::{Serialize, de::DeserializeOwned};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
 
 use super::EventEnvelope;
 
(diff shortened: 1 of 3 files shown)


ishandhanani (approved):

jthomson04:
/ok to test [d99263c](https://github.com/ai-dynamo/dynamo/commit/d99263cd45866eee7cef582b91b4b60dff4dcc38)

PR #11540, opened 2026-07-10
feat(bindings): expose Client.instance_tcp_addresses() to Python

### LINKED ISSUE #11543: [FEATURE]: Expose instance TCP transport addresses on Client (Python bindings)
### Feature request

Expose a Python API on `Client` that returns the TCP transport addresses of an endpoint's currently-registered instances, e.g.:

```python
client = await endpoint.client()
addrs = client.instance_tcp_addresses()
# ["10.0.0.12:9401/ns/component/endpoint", ...]
```

The binding would read the runtime `Client`'s watched instance set (`client.instances()`) and return each instance's `TransportType::Tcp` address string. Non-TCP transports (e.g. NATS) are skipped. Like `instance_ids()`, the result is a snapshot; callers can pair it with `wait_for_instances()` to block until instances exist.

### Describe the problem you're encountering

From Python there is currently no dynamo API to learn *where* an endpoint's peer instances are running (their node addresses). `Client.instance_ids()` only returns opaque instance IDs.

A concrete use case: in a disaggregated deployment, a worker wants to run a startup RDMA connectivity check against its peer workers, which requires the peer node IPs. Today the only way to get them is to read the discovery/etcd registry layout directly and parse the registered instance records, which couples user code to dynamo's internal registry schema and can break on upgrades.

Environment: any deployment using the Python bindings (`dynamo._core.Client`); observed on Linux with recent `main`.

### Describe alternatives you've tried

- Reading the etcd/discovery registry directly from Python and parsing the instance JSON to extract the transport address. This works but depends on the internal key layout and record schema, which are not a public contract.
- Passing peer addresses out-of-band via deployment configuration (env vars / operator-rendered config). This duplicates information dynamo already tracks and goes stale when instances are rescheduled.

Implementation is proposed in PR #11540.

jyizheng:
## Overview:

Expose a new Python binding, `Client.instance_tcp_addresses()`, that returns a dict mapping instance id to the TCP transport address of each of an endpoint's currently-registered instances.

## Details:

- Adds `Client.instance_tcp_addresses()` in `lib/bindings/python/rust/lib.rs`. It reads the runtime `Client`'s watched instance set (`client.instances()`) and returns a dict of `instance_id -> TransportType::Tcp` address string (keys match `instance_ids()`); non-TCP transports (e.g. NATS) are skipped.
- Adds the corresponding type stub and docstring in `lib/bindings/python/src/dynamo/_core.pyi`.

This lets a Python caller reach a peer worker's node via a dynamo API instead of reading the discovery/etcd registry layout directly — for example, a startup RDMA connectivity check that needs peer node IPs. Like `instance_ids()`, the result is a snapshot of the watched instance set; callers can pair it with `wait_for_instances()` to block until instances exist.

- `lib/bindings/python/rust/lib.rs` — the new `instance_tcp_addresses()` method on `Client`.
- `lib/bindings/python/src/dynamo/_core.pyi` — the matching stub.

## Related Issues

- Closes #11543


lib/bindings/python/rust/lib.rs
```diff
@@ -1352,6 +1352,27 @@ impl Client {
...
+    /// dynamo API instead of reading the discovery/etcd registry layout
+    /// directly. Non-TCP transports (e.g. NATS) are skipped. This is a snapshot
+    /// of the watched instance set (like ``instance_ids``); combine with
+    /// ``wait_for_instances`` if you need to block until instances exist.
+    fn instance_tcp_addresses(&self) -> Vec<String> {
+        self.router
+            .client
+            .instances()
+            .into_iter()
+            .filter_map(|instance| match instance.transport {
```
michaelfeil, line 1363: Should this not be a MAP/dict for instances?

michaelfeil (commented):
quick quesiton

lib/bindings/python/rust/lib.rs
jyizheng: Good point — changed in f5cc94848: `instance_tcp_addresses()` now returns a dict mapping `instance_id -> TCP transport address` (keys match `instance_ids()`), instead of a bare list. The `_core.pyi` stub is updated to `Dict[int, str]` accordingly.

lib/bindings/python/rust/lib.rs
```diff
@@ -1352,6 +1352,29 @@ impl Client {
         self.router.client.instance_ids()
     }
 
+    /// Get the TCP transport addresses of this endpoint's current instances,
```
michaelfeil: docstring should be shorter

michaelfeil:
https://github.com/ai-dynamo/dynamo/blob/cfdc6302b64a771ab38242af41a3284430328d2d/lib/bindings/python/tests/test_request_plane_python_payload.py#L62 maybe add a small usage of this python function in an existing test. Perhaps call the function here, to make sure it survives the next minor versions of dynamo.

michaelfeil:
LGTM, thanks for contributing

lib/bindings/python/rust/lib.rs
jyizheng: fixed

lib/bindings/python/rust/lib.rs
jyizheng: fixed

jyizheng:
> https://github.com/ai-dynamo/dynamo/blob/cfdc6302b64a771ab38242af41a3284430328d2d/lib/bindings/python/tests/test_request_plane_python_payload.py#L62
> 
> maybe add a small usage of this python function in an existing test. Perhaps call the function here, to make sure it survives the next minor versions of dynamo.

UT added

lib/bindings/python/tests/test_request_plane_python_payload.py
```diff
@@ -61,6 +61,12 @@ async def request_plane_client(runtime):
     client = await endpoint.client()
     await client.wait_for_instances()
 
+    # This fixture runs on the TCP request plane, so every instance must
+    # expose a TCP transport address keyed by its instance id.
+    tcp_addresses = client.instance_tcp_addresses()
+    assert set(tcp_addresses) == set(client.instance_ids())
```
michaelfeil: actually, this only works for TCP... .I can also be None. But if its not None, you can run these assertions.

michaelfeil (commented):
another comment

lib/bindings/python/tests/test_request_plane_python_payload.py
jyizheng: fixed

nnshah1:
Thanks for pushing on this @jyizheng — the peer-address need in #11543 is real. One suggestion on the shape: rather than `instance_tcp_addresses()` returning `Dict[int, str]`, could we expose the structured model the runtime already has?

`instance_tcp_addresses()` filters to TCP (silently drops NATS instances) and flattens `host:port/<id>/<endpoint>` into a single string the caller has to re-parse — which is the same "coupled to an internal string layout" problem the issue set out to avoid.

Proposed alternative: `Client.instances() -> List[InstanceInfo]`, mirroring the runtime `Instance`/`TransportType` 1:1 — each instance exposes `instance_id`/`namespace`/`component`/`endpoint`/`device_type` plus `transport {kind, address}` (kind = `"tcp"`/`"nats_tcp"`). It keeps all transports and makes id/namespace/etc first-class fields. Your TCP list stays a one-liner:

```python
addrs = [i.transport.address for i in client.instances() if i.transport.kind == "tcp"]
```

I put it up as #11617 (builds, clippy clean, request-plane test passing) so you can look at the concrete shape side by side — not trying to race your PR, just opening the direction for discussion. WDYT? Happy to go either way.

PR #11074, opened 2026-06-30
feat(epp): reconcile standalone worker topology with the embedded selector

atchernych:
## Overview

Adds the worker topology adapter for standalone EPP. Raw vLLM pods do not publish a `ModelRuntimeConfig`, so the adapter converts each ready worker into a `WorkerRegistration` using resolved pod endpoints and environment-derived defaults, then reconciles the desired worker set with the in-process `Selector`.

## Behavior

- Runs an initial reconciliation when spawned, then reconciles on `PodDiscovery` changes.
- Reuses the stable `worker_id` produced by `PodDiscovery` so each EPP replica builds the same catalog.
- Registers the milestone-1 single-rank topology: DP size 1 in `Selector`, with the worker's KV-event endpoint assigned to rank 0.
- Propagates model name, block size, optional KV capacity and batching limits, HTTP endpoint, and optional replay endpoint.
- Cancels the background reconciliation task when `TopologyAdapter` is dropped.
- Removes the unused `stable_routing_id` field from the EPP `WorkerRegistration` contract.
- Fixes the feature-gated `kvbm-offload` test imports required by the Pre Merge Rust test.

## Lifecycle

`Selector` owns an embedded `SelectionService` and shares the EPP process lifecycle. When an EPP replica restarts, it constructs a fresh selector and the topology adapter performs an immediate reconciliation before waiting for pod changes. There is no external selector fleet or selector-fleet watch; optional peer discovery manages only the replication peer set.

## Scope

This PR exports `TopologyAdapter` and its registration defaults. It does not yet wire the adapter into the standalone EPP request path; request-path wiring remains in the follow-up PR.

## Validation

- `cargo fmt -- --check`
- `cargo test --locked -p dynamo-ext-proc`: 82 unit tests and 1 integration test passed
- `cargo clippy --locked -p dynamo-ext-proc --no-deps --all-targets -- -D warnings`
- `cargo test --locked -p dynamo-mocker --features replay-bench,kvbm-offload replay_forces_g1_to_g2 -- --nocapture`: 2 passed
- `cargo test --locked -p dynamo-bench --test mooncake_trace --features mocker-kvbm-offload g2 -- --nocapture`: 1 passed, 18,340 events processed

## Related Issues

None specified.

PR #11541, opened 2026-07-10
feat(ext-proc): embed SelectionService in standalone EPP

atchernych:
#### Overview:

This PR adds the embedded, runtime-free selection layer used by the standalone EPP. The ext-proc binary links `dynamo-kv-router`'s `SelectionService` and calls its typed Rust API in-process, so standalone mode does not require the Dynamo runtime or a separate selector Deployment.

It also adds optional multi-EPP active-load replication. When `DYN_EPP_PEER_SERVICE` is configured, each EPP discovers sibling replicas from its own Service's EndpointSlices and subscribes to their replica-sync publishers over ZMQ. Without that setting, the selector remains fully local to one EPP replica.

This is PR 4 in the standalone EPP train and is based on `epp-selector-reflector`. It provides the selector and peer-discovery primitives; the topology adapter and request/lifecycle wiring land in the stacked follow-up PRs.

#### Details:

- Add an in-process `Selector` wrapper and the typed `WorkerRegistration`, `SelectRequest`, `SelectResponse`, and `OverlapSummary` interfaces used by the reflector -> topology adapter -> request path.
- Build `SelectionService` directly from the standard `DYN_ROUTER_*` configuration and expose worker reconciliation, readiness, and `select_and_reserve` without JSON or HTTP on the selection hot path.
- Keep the milestone-1 worker contract aggregated and single-rank: routing group `default`, DP size `1`, and no standalone multi-rank selection surface.
- Make worker reconciliation idempotent. Only `Schedulable` upserts are cached as converged; `Incomplete` workers remain uncached, log their unschedulable reasons, and are retried on a later reconciliation.
- Resolve the effective per-model queueing policy through `KvRouterConfig::queueing_enabled(model_name)`. Standalone startup fails early when queueing is enabled but `DYN_EPP_MAX_NUM_BATCHED_TOKENS` is missing or zero; worker reconciliation uses the same predicate.
- Enable multi-EPP replication by the presence of `DYN_EPP_PEER_SERVICE`:
  - resolve a consistent named TCP `replica-agg` port from the Service's labeled EndpointSlices before building the selector;
  - require `POD_IP` so an EPP does not register itself as a peer;
  - watch EndpointSlices with reflector backoff and register/deregister sibling ZMQ peers as membership changes;
  - hold the peer-readiness latch false until the initial EndpointSlice LIST and first peer reconciliation complete, then retain the last-known peer set across transient watch failures;
  - replicate live admission, prefill-complete, and free events. This PR does not add an active-reservation snapshot protocol for a joining replica.
- Add the `dynamo-kv-router/standalone-selection` feature to ext-proc, export/default the selection request types needed by typed callers, and give `WorkerRequest::default()` the same `default` model/routing-group values as deserialization.
- Install libzmq build and runtime packages in the ext-proc image and add the required `tokio-util` dependency.

#### Testing:

- Worker reconciliation retries identical registrations that remain `Incomplete`.
- Queueing detection covers synthetic thresholds and effective threshold-free policies.
- Peer discovery covers named-port validation, TCP protocol enforcement, address-family filtering, initial reconciliation, peer-set changes, and terminating/serving transitions.
- Replica-sync coverage exercises peer registration and remote reservation/load release through lifecycle events.

#### Where should the reviewer start?

1. `deploy/inference-gateway/ext-proc/src/selector.rs` for the embedded selector boundary, milestone-1 worker mapping, queueing validation, and reconciliation semantics.
2. `deploy/inference-gateway/ext-proc/src/peer_discovery.rs` for Service port resolution, EndpointSlice reflection, readiness, and ZMQ peer membership.
3. `lib/kv-router/src/scheduling/config.rs` and `lib/kv-router/src/services/selection/` for the shared queueing predicate and additive typed-API/default changes.
4. `deploy/inference-gateway/ext-proc/Cargo.toml` and `Dockerfile` for feature and libzmq packaging changes.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- Relates to #11661
- Recreates #11371, which GitHub marked merged during a stacked-rebase even though it did not reach `main`.
- Stacked follow-ups: topology adapter #11074, request-path wiring #11542, and reservation lifecycle #11155.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11541" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added standalone worker selection with registration, reconciliation, reservation, and overlap information.
  * Added automatic peer discovery and synchronization for replicated selector instances.
  * Added readiness reporting for worker availability and peer synchronization.
  * Added configurable queueing detection based on model routing policies.

* **Bug Fixes**
  * Improved handling of incomplete worker registrations so they are retried and remain unavailable until valid.

* **Improvements**
  * Added more consistent default handling for worker and prompt selection requests.


deploy/inference-gateway/ext-proc/src/peer_discovery.rs
```diff
@@ -0,0 +1,285 @@
...
+    for ip in &removed {
+        let endpoint = format!("tcp://{}", authority(ip, sync_port));
+        if let Err(e) = service.deregister_replica_peer(endpoint.clone()).await {
+            tracing::debug!(%endpoint, error = %e, "deregister_replica_peer failed");
+        }
+    }
+    if !added.is_empty() || !removed.is_empty() {
+        tracing::info!(added = ?added, removed = ?removed, total = live.len(), "EPP peer set changed");
+    }
+    *known = live;
```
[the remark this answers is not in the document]
atchernych: register_replica_peer ultimately does a zmq_connect, which is fire-and-forget: it records the endpoint and the ZMQ I/O thread (re)connects in the background, so a transient network blip doesn't surface as a registration error — that's the reconnect path we already rely on for not-yet-ready siblings. The only real error variants are InvalidEndpoint (deterministic, and endpoints here are always well-formed IP:port) and Unavailable (terminal — the peer-manager actor is gone, so per-peer retry can't help). Also, the reconcile loop is edge-triggered on EndpointSlice changes with no periodic tick, so "retry next cycle" wouldn't provide timely retries anyway. Marking the live set as known is correct for these failure modes.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Select a worker for a prompt and book its load in one operation. Takes the
+    /// request by value so per-request fields (`token_ids`, `model_name`, ...) are
+    /// moved into the core request rather than cloned on the hot path.
+    pub async fn select_and_reserve(&self, req: SelectRequest) -> Result<SelectResponse> {
+        let core_req = CoreSelectAndReserveRequest {
+            model_name: req.model_name,
+            tenant_id: DEFAULT_TENANT.to_string(),
+            selection_id: req.selection_id,
+            reservation_id: None,
```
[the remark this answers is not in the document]
atchernych: This is a stacked PR. select_and_reserve intentionally lands here as the booking primitive; the release path the comment asks for — threading reservation_id through SelectResponse and adding Selector::free_reservation / Selector::prefill_complete, wired to EppRouter::on_request_complete / on_prefill_complete — is implemented in the follow-up PR epp-selector-reservations (#11155). No load-leak exists in the combined stack.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,310 @@
...
+        // Upsert new or changed workers.
+        for (worker_id, reg) in desired {
+            if current.get(worker_id) == Some(reg) {
+                continue;
+            }
+            self.service
+                .upsert_worker(Self::worker_request(reg))
+                .await
+                .map_err(|e| anyhow!("upsert_worker failed: {e}"))?;
+            current.insert(*worker_id, reg.clone());
```
tmonty12: **[P1] Do not cache `Ok(Incomplete)` as reconciled**

`upsert_worker()` can return `Ok(record)` with lifecycle `Incomplete`, but this unconditionally inserts the desired registration into `current`. Every later identical snapshot then skips the worker, even though it never became schedulable. Please insert into `current` only when the returned lifecycle is `Schedulable`; surface `not_schedulable_reasons` for `Incomplete` so it remains visibly unconverged. Add a regression test showing that `Ok(Incomplete)` is not cached as reconciled.

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
```diff
@@ -0,0 +1,406 @@
...
+        }
+    });
+
+    let store_for_wait = store.clone();
+    match tokio::time::timeout(INITIAL_LIST_TIMEOUT, store_for_wait.wait_until_ready()).await {
+        Ok(Ok(())) => tracing::info!("EPP peer EndpointSlice initial LIST sync complete"),
+        Ok(Err(e)) => {
+            tracing::warn!(error = %e, "EPP peer EndpointSlice writer dropped before initial LIST")
+        }
+        Err(_) => tracing::warn!(
```
tmonty12: **[P1] Initial peer discovery must gate readiness**

When `DYN_EPP_PEER_SERVICE` is configured, readiness should remain false until the EndpointSlice reflector completes its first full LIST and the initial peer-set reconciliation runs. The current 30-second timeout warns and returns success, allowing local selector readiness before the EPP has an authoritative peer set. Please keep discovery retrying while startup readiness remains false, then latch the initial-sync condition after success. Subsequent transient watch failures should retain the last-known peers and must not clear readiness.

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
```diff
@@ -0,0 +1,406 @@
...
+    let want_ipv6 = is_ipv6(self_ip);
+    let mut ips = peer_ips(store.state().iter().map(|s| s.as_ref()), want_ipv6);
+    ips.remove(self_ip);
+    ips
+}
+
+/// Format `host:port`, bracketing IPv6 literals (`fd00::1` -> `[fd00::1]`) so the
+/// resulting `tcp://` endpoint stays valid on dual-stack clusters.
+fn authority(ip: &str, port: u16) -> String {
+    if ip.contains(':') {
```
tmonty12, line 281: **[P1] Enable IPv6 on the ZMQ sockets**

Bracketing IPv6 literals is necessary but insufficient: the shared ZMQ socket configuration never enables `ZMQ_IPV6`, whose default is false. I verified that a default socket fails to bind `tcp://[::1]:*` with `No such device`, while enabling `ZMQ_IPV6` succeeds. IPv6-only EPPs therefore cannot bind or connect the replica plane. Please enable IPv6 for both PUB and SUB sockets and add an actual IPv6 loopback delivery test.

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    for slice in slices {
+        slice_count += 1;
+        let slice_name = slice.metadata.name.as_deref().unwrap_or("<unnamed>");
+        let mut matches = slice
+            .ports
+            .as_deref()
+            .unwrap_or_default()
+            .iter()
+            .filter(|port| port.name.as_deref() == Some(REPLICA_AGG_PORT_NAME));
```
tmonty12: **[P2] Enforce the named Service/TCP contract**

DEP #11661 requires the configured peer Service and a named TCP `replica-agg` port. This resolver validates only labeled EndpointSlices and ignores `EndpointPort.protocol`, so a stale/manually managed slice or an explicit UDP/SCTP port can pass before the selector binds and dials `tcp://`. Please verify that the Service exists, accept only absent/default or `TCP` protocol, and add missing-Service and explicit-UDP tests.

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
```diff
@@ -0,0 +1,406 @@
...
+    slices: impl Iterator<Item = &'a EndpointSlice>,
+    want_ipv6: bool,
+) -> BTreeSet<String> {
+    let mut ips = BTreeSet::new();
+    for slice in slices {
+        if slice.address_type.eq_ignore_ascii_case("IPv6") != want_ipv6 {
+            continue;
+        }
+        for endpoint in &slice.endpoints {
+            if endpoint
```
tmonty12: **[P2] Retain terminating peers while they are draining**

`terminating=true` means the pod has entered termination, not that its in-flight ext-proc streams have finished. Disconnecting immediately can miss final PrefillComplete/Free events and leave phantom peer load until expiry. Please retain a terminating endpoint while `serving=true`, and remove it when `serving=false` or the endpoint disappears. Add coverage for a terminating-but-serving peer completing an in-flight reservation.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -778,6 +778,14 @@ impl KvRouterConfig {
         self.router_predicted_ttl_secs.is_some()
     }
 
+    /// True when the effective scheduling policy requires each worker to report
+    /// `max_num_batched_tokens` (queue-threshold admission or a router policy
+    /// config is active). Discovery-time callers can validate the value up front
+    /// rather than letting workers silently become unschedulable later.
+    pub fn requires_max_num_batched_tokens(&self) -> bool {
```
tmonty12: **[P2] Derive queueing from the effective model policy**

This checks raw configuration presence rather than the effective policy. An applicable threshold-free policy does not enable queueing even when `router_policy_config` is set, and `router_queue_threshold` only applies to the synthetic fallback when no model/root profile matches. Could we replace this with a fallible, model-aware `KvRouterConfig::queueing_enabled(model_name)` that resolves `policy_profile(model_name)` and returns `classes().iter().any(PolicyClassConfig::queueing_enabled)`? Please use the same method in `SelectionCore::reconcile_worker` and EPP startup so both enforce the current `queueing → max_num_batched_tokens` invariant from one policy predicate.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,310 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! In-process, runtime-free worker selector.
+//!
+//! Standalone mode runs the runtime-free selection service **in-process**: the
```
tmonty12: nit: also long comment here, can we reduce and keep more generic

deploy/inference-gateway/ext-proc/src/selector.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
atchernych: will add

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
atchernych: This will require changes in  lib/kv-router/src/services/common/zmq.rs
The fix is small but That file is owned by the routing team and is used by the standalone-indexer and slot-tracker paths, so it's out of scope for this PR. Lets do a follow-up with them

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
atchernych: will fix the protocol issue. The service check is not needed. A missing Service already fails (no Service -> no labeled EndpointSlices ->  "peer Service has no EndpointSlices" error

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
atchernych: Will drop on  terminating && serving != Some(true)

deploy/inference-gateway/ext-proc/src/selector.rs
atchernych: will do

deploy/inference-gateway/ext-proc/src/peer_discovery.rs
tmonty12: Thanks—the terminating-but-serving behavior is now correct. Keeping this open for the requested end-to-end coverage: retain a draining peer, deliver its final `PrefillComplete`/`Free` over replica sync, and assert the remote reservation/load is released. The current tests cover only the EndpointSlice filtering predicate.

atchernych: added

lib/kv-router/src/scheduling/config.rs
atchernych: i made this update but in general I would like to get this run by Rudy.

PeaBrane (commented):
AI-assisted review comments for human evaluation. These are neutral static-analysis observations; please validate the assumptions and severity against the intended standalone-EPP scope. I did not duplicate the existing IPv6/ZMQ thread, which already covers that concern. No approval or request-changes verdict is intended.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+            stable_routing_id: reg.stable_routing_id.clone(),
+            ..Default::default()
+        }
+    }
+
+    pub async fn reconcile(&self, desired: &HashMap<u64, WorkerRegistration>) -> Result<()> {
+        let mut current = self.current.lock().await;
+
+        // Upsert new or changed workers.
+        for (worker_id, reg) in desired {
```
PeaBrane: AI-assisted observation — please evaluate: `desired` carries the worker ID twice. This loop keys `current` and stale deletion by the map key, while `worker_request(reg)` registers `reg.worker_id`. If they differ, removing the entry deletes the outer ID (and accepts `NotFound`) while the inner ID can remain schedulable. Would it be safer to validate equality or derive bookkeeping exclusively from `reg.worker_id`?

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+                    worker_id = *worker_id,
+                    lifecycle = ?record.lifecycle,
+                    reasons = ?record.not_schedulable_reasons,
+                    "Worker upserted but not schedulable; leaving uncached to retry on the next reconcile"
+                );
+            }
+        }
+
+        // Delete workers that are no longer desired.
+        let stale: Vec<u64> = current
```
PeaBrane: AI-assisted observation — please evaluate: only `Schedulable` upserts enter `current`, but stale detection also scans only `current`. A worker first stored by `SelectionService` as `Incomplete` can disappear from `desired` without `delete_worker` ever being called, leaving its catalog lifecycle `Incomplete`. Consider tracking applied worker IDs separately from the schedulable convergence cache and adding an `{incomplete} → {empty}` regression test.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    pub async fn reconcile(&self, desired: &HashMap<u64, WorkerRegistration>) -> Result<()> {
+        let mut current = self.current.lock().await;
+
+        // Upsert new or changed workers.
+        for (worker_id, reg) in desired {
+            if current.get(worker_id) == Some(reg) {
+                continue;
+            }
+            let record = self
```
PeaBrane, line 235: AI-assisted performance observation — please evaluate against expected pool sizes: reconciliation awaits each upsert serially, and each successful core upsert republishes a full scheduler-config map. Initial convergence of W workers is therefore roughly O(W²) config cloning/notifications, plus serialized listener setup. A batch or deferred-publish path may be preferable if InferencePools can be large.

deploy/inference-gateway/ext-proc/src/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+impl Selector {
+    pub async fn new(cfg: &EppStandaloneConfig) -> Result<Self> {
+        let kv_router_config = kv_router_config_from_dynamo_env();
+        let cancel = CancellationToken::new();
+
+        // If queueing is enabled, we need to validate that the max_num_batched_tokens is set.
+        // Done once at startup to avoid validating on every reconcile.
+        let queueing_enabled = kv_router_config
+            .queueing_enabled(Some(&cfg.model_name))
+            .map_err(|e| anyhow!("resolving router policy for model {}: {e}", cfg.model_name))?;
```
PeaBrane, line 114: AI-assisted test-coverage observation — please evaluate: this startup guard is the user-visible queueing failure mode, but the current selector test always supplies `max_num_batched_tokens`, while the config test covers only the synthetic threshold. Could we add model-specific policy cases for queueing enabled + missing max → startup error, and threshold-free + missing max → startup success?

deploy/inference-gateway/ext-proc/src/selector.rs
tmonty12: Fixed in fd59c9a0b3: reconciliation now tracks all introduced worker IDs separately from the schedulable cache, deletes stale Incomplete workers, and includes regression coverage.

deploy/inference-gateway/ext-proc/src/selector.rs
tmonty12: Fixed in fd59c9a0b3: added model-specific startup tests for queueing with missing capacity and threshold-free policy with missing capacity.

deploy/inference-gateway/ext-proc/src/selector.rs
tmonty12: Deferred: avoiding repeated full-map publication requires a batched or deferred-publish SelectionService API; the ext-proc wrapper cannot eliminate it while calling single-worker upserts.

deploy/inference-gateway/ext-proc/src/selector.rs
tmonty12: Fixed in fd59c9a0b3: reconcile now accepts a registration slice, derives bookkeeping keys from worker_id, and rejects duplicate IDs before mutation.

PeaBrane (approved):

nv-anants (approved):

PR #11072, opened 2026-06-30
feat: PR 3 Worker Pod Discovery for the EPP and standalone Selector Router Integration

atchernych:
## Overview

Discover standalone EPP workers from the InferencePool targeted by the gateway, so both components use the same pool membership and target port. This PR provides discovery only; wiring the discovered workers into the embedded SelectionService/EPP occurs in later PRs in the train.

## Details

- Watch the configured GAIE `InferencePool` and derive its backend Pod `matchLabels` selector and HTTP target port into `PoolState`.
- Watch Pods in the namespace through kube-rs' reflector store and maintain an incremental index of Ready, selector-matching workers rather than rebuilding all workers for every Pod event.
- Rebuild the index after a completed reflector relist. A valid Pool selector/port edit observed during `Init`/`InitApply` is deferred to `InitDone`, so the rebuilt index is derived from the completed Pod set and latest PoolState.
- Clear the index immediately when the Pool is deleted or becomes invalid, preventing routing with a stale selector or port.
- Publish discovery readiness and indexed workers/endpoints for the later topology-adapter and router PRs.
- Add coverage for a Pool edit interleaved with a Pod relist.

## Where should the reviewer start?

- `deploy/inference-gateway/ext-proc/src/pod_discovery.rs` — reflector event loop, incremental worker index, and Pool-change/relist interaction.
- `deploy/inference-gateway/ext-proc/src/inference_pool.rs` — InferencePool watch and PoolState parsing/validation.

## Related Issues

**🚫 This PR is NOT linked to an issue**:

- [x] Confirmed — no related issue

## Summary by CodeRabbit

* **New Features**
  * Added automatic discovery of ready inference workers from Kubernetes pods.
  * Workers are matched to their configured inference pool and exposed with HTTP, key-value events, and optional replay endpoints.
  * Added live readiness tracking as pool and pod information becomes available.
  * Added support for IPv4 and IPv6 worker endpoints.

* **Bug Fixes**
  * Invalid or incomplete inference pool configurations no longer leave stale worker state.
  * Unready, malformed, unmatched, or terminating pods are excluded from routing.

PR #11070, opened 2026-06-30
feat: PR 1: EPP config for the EPP and standalone Selector Router Integration

atchernych:
## Summary

This PR adds the first configuration-only slice of standalone EPP mode.

- Adds `DYN_EPP_MODE`: `dynamo` remains the default existing runtime-backed mode; `standalone` selects the future selector-only path.
- Adds `EppStandaloneConfig` to parse and validate the standalone environment contract.
- Treats `DYN_MODEL_NAME` as the served/catalog identity used to group discovered workers, not as an offline-tokenizer input.
- Fails fast in standalone mode after parsing and logging the validated config. It does not yet construct a selector, discover workers, or route requests.
- Leaves the existing Dynamo-runtime EPP path unchanged.

### Standalone configuration

`DYN_EPP_MODE=standalone` requires:

- `DYN_EPP_INFERENCE_POOL_NAME`: the `InferencePool` this EPP backs.
- `POD_NAMESPACE`: namespace supplied through the downward API.
- `DYN_MODEL_NAME`: served/catalog model identity.
- `DYN_KV_CACHE_BLOCK_SIZE`: positive KV-cache block size that must match the inference engine.

Optional settings:

- `DYN_EPP_SELECTION_INDEXER_THREADS`: positive selector indexer thread count; defaults to `4`.
- `DYN_EPP_PEER_SERVICE`: peer Service identity. Future selector wiring resolves its named `replica-agg` port from EndpointSlices; this PR does not configure a replication port.
- `DYN_EPP_KV_EVENT_PORT`: positive KV-event ZMQ port; defaults to `5557`.
- `DYN_EPP_KV_EVENT_REPLAY_PORT`: optional replay endpoint port; unset by default and must match the worker endpoint when provided.
- `DYN_EPP_TOTAL_KV_BLOCKS`: optional per-worker total KV-block count.
- `DYN_EPP_MAX_NUM_BATCHED_TOKENS`: optional per-worker batching limit; when set, it must be greater than zero.

There is no `DYN_DATA_PARALLEL_SIZE` or `DYN_EPP_PEER_SYNC_PORT` configuration in this milestone.

### Scope boundary

This milestone only establishes the mode switch and validates the configuration contract. Standalone startup intentionally exits with `DYN_EPP_MODE=standalone is not yet supported` after parsing the config. Offline tokenization, Kubernetes discovery, EndpointSlice port resolution, selector construction, KV-router-policy validation, replication, and request routing remain follow-on work.

## Where should the reviewer start?

[...]

Issue #10321, opened 2026-06-04
DEP: Runtime-Free Router and Gateway On-Ramp


PeaBrane:
## Table of Contents

- [Overview](#user-content-overview)
  - [Area](#user-content-area)
  - [Summary](#user-content-summary)
  - [Motivation](#user-content-motivation)
- [Design](#user-content-design)
  - [Proposal](#user-content-proposal)
  - [Worker Identity, Tenant Isolation, and Schedulability](#user-content-worker-identity-tenant-isolation-and-schedulability)
  - [Selection Input and Admission Sequencing](#user-content-selection-input-and-admission-sequencing)
  - [Worker Topology and Platform Boundary](#user-content-worker-topology-and-platform-boundary)
  - [Engine Metadata Production](#user-content-engine-metadata-production)
- [Interfaces](#user-content-interfaces)
  - [HTTP API](#user-content-http-api)
  - [In-Process Python API](#user-content-in-process-python-api)
  - [Existing Runtime-Backed Path](#user-content-existing-runtime-backed-path)
- [Scope and Rollout](#user-content-scope-and-rollout)
  - [V1 Scope](#user-content-v1-scope)
  - [Phased Migration](#user-content-phased-migration)
- [Requirements](#user-content-requirements)
  - [MUST](#user-content-must)
  - [SHOULD](#user-content-should)
- [Open Questions](#user-content-open-questions)
- [References](#user-content-references)

<a id="overview"></a>

## Overview

<a id="area"></a>

### Area

router, gateway

<a id="summary"></a>

### Summary

Migrate Dynamo routing toward a runtime-free router core and service API that can serve both Gateway/EPP selector-service integration and HTTP push-router/frontend use cases, while preserving the existing Dynamo-runtime-backed frontend/router path during migration. V1 is a practical runtime-free on-ramp; router-local parity and behaviors that require frontend, gateway, or worker coordination follow in later phases.

<a id="motivation"></a>

### Motivation

The Dynamo router is one of the easiest Dynamo capabilities for users to want, but today the cleanest production paths often make router adoption feel like Dynamo runtime adoption. Existing vLLM/SGLang/TRT-LLM users frequently want a low-risk first proof: keep their current workers and request plane, add smarter KV/load-aware routing, measure TTFT/throughput, and roll back cleanly.

[...]

PR #11542, opened 2026-07-10
feat(ext-proc): wire standalone EPP routing and reservation lifecycle

atchernych:
## Overview

Wires the previously landed standalone EPP components into the ext-proc request path. With `DYN_EPP_MODE=standalone`, ext-proc now constructs an in-process `EppRouter` from the standalone configuration, vLLM render client, InferencePool-driven worker discovery, embedded `SelectionService`, and topology adapter.

For each aggregated inference request, the EPP renders the prompt for routing tokens, selects and reserves a Ready worker, and returns the destination endpoint to Envoy. The raw vLLM worker receives the original request body and performs its own tokenization.

Standalone mode does not require `DistributedRuntime`, etcd, NATS, or a separate selector service.

### Startup and readiness

- Build `EppRouter` from the validated standalone configuration.
- Start worker discovery, the embedded selector, topology reconciliation, and optional peer discovery.
- Publish gRPC health as serving after the worker reflector is synchronized and, when replication is configured, the initial peer reconciliation completes.
- Treat a valid empty pool as ready while returning 503 per request until a Ready worker is available.

### Request routing

- Buffer the request body once into shared `Bytes`.
- Resolve request priority using header-over-body precedence:
  - `x-dynamo-request-priority`
  - `x-dynamo-request-strict-priority`
  - `nvext.agent_hints`
- Forward the unchanged chat-completions body to the configured vLLM `/v1/chat/completions/render` endpoint.
- Constrain selection to an Envoy-provided destination subset when present, including exact endpoints and bare IPv4/IPv6 addresses.
- Call the embedded `SelectionService` through its typed `select_and_reserve` API.
- Refuse stale selections when the chosen worker is no longer resolvable as Ready.
- Route body-less requests to any Ready worker, respecting the destination subset.

## Reservation lifecycle

- Mint a unique reservation UUID for every successful pick instead of using the client-controlled `x-request-id`.
- Carry the reservation ID in the per-stream request context, avoiding a shared request-ID map and isolating streams that reuse the same request ID.
- Release a reservation if selection is canceled or dropped before the server adopts it.
- Mark prefill complete on the first non-empty response chunk without blocking first-token forwarding.
- Free the reservation when the ext-proc stream completes.
- Keep lifecycle operations idempotent for duplicate or late callbacks.

## Failure handling and backpressure

- Cancel an in-progress pick when Envoy closes the ext-proc stream.
- Classify renderer failures as:
  - 400/422 → invalid client request
  - 429/503 or transport unavailable → retryable 503
  - timeout → 504
  - other upstream or response-contract failures → 502
- Bound concurrent `pick()` operations with `DYN_EPP_MAX_INFLIGHT_REQUESTS`, defaulting to `1024`.
- Shed excess work immediately with 503 instead of queueing request bodies and render calls behind the limit.

## Scope

- Aggregated serving with raw vLLM workers only.
- Workers re-tokenize the forwarded request; routing token IDs are not injected into the request body.
- Disaggregated serving remains follow-up work.
- Peer readiness establishes discovery and connectivity for future load deltas. This PR does not add active-reservation snapshot replication, which is outside the current `SelectionService` contract.

## Validation

- `cargo fmt -p dynamo-ext-proc -- --check`
- `git diff --check`
- `cargo test -p dynamo-ext-proc --locked`
  - 97 unit tests
  - 1 integration test
  - doc tests
- `cargo clippy -p dynamo-ext-proc --all-targets --locked -- -D warnings`

1. `deploy/inference-gateway/ext-proc/src/epp_router.rs` — standalone assembly, tokenization, selection, reservation ownership, and load shedding.
2. `deploy/inference-gateway/ext-proc/src/server.rs` — per-stream lifecycle, disconnect cancellation, routing responses, and callback handling.
3. `deploy/inference-gateway/ext-proc/src/selector.rs` — typed embedded-selector boundary and lifecycle operations.
4. `lib/llm/src/protocols/common/extensions.rs` — shared request-priority contract and precedence.

## Related

- Relates to #11661.
- Recreates #11075, which GitHub marked merged during a stacked-rebase even though it did not reach `main`.
- Preceded by:
  - #11070 — standalone configuration
  - #11827 — vLLM render client
  - #11072 — InferencePool worker discovery
  - #11541 — embedded `SelectionService`
  - #11074 — worker topology adapter

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11542" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added standalone endpoint routing with worker selection, destination-subset support, and readiness-aware health reporting.
  * Added request priority controls through `x-dynamo-request-priority` and `x-dynamo-request-strict-priority` headers.
  * Improved request routing lifecycle and cancellation handling for more reliable stream processing.

* **Bug Fixes**
  * Added clearer handling for tokenizer unavailable, timeout, and upstream failures.
  * Prevented routing to workers that become unavailable during selection.

* **Documentation**
  * Documented request priority headers, precedence rules, accepted values, and fallback behavior.


deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,260 @@
...
+            });
+        }
+
+        let body = std::str::from_utf8(&req.body)
+            .map_err(|e| PickError::TokenizationFailed(format!("request body not UTF-8: {e}")))?;
+        let (tokens, priority_jump, strict_priority) = self
+            .tokenize(body)
+            .map_err(|e| PickError::TokenizationFailed(e.to_string()))?;
+
+        let select_req = SelectRequest {
```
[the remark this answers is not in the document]
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,260 @@
...
+        req: &RequestInfo,
+        _endpoints: &[Endpoint],
+    ) -> Result<PickResult, PickError> {
+        if !self.reflector_ready.load(Ordering::Acquire) {
+            return Err(PickError::RoutingFailed(
+                "pod reflector cache not ready".to_string(),
+            ));
+        }
+
+        let allowed = self.reflector.ready_worker_ids();
```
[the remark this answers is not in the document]
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,260 @@
...
+        };
+
+        // Atomic select-and-reserve: books the request's load on the chosen
+        // worker. Releasing the booking over the request lifecycle
+        // (`on_request_complete` / `on_prefill_complete`) is wired in a follow-up.
+        let resp = self
+            .selector
+            .select_and_reserve(select_req)
+            .await
+            .map_err(|e| PickError::RoutingFailed(e.to_string()))?;
```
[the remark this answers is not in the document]
atchernych: this is addressed in the later branch epp-selector-reservations

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        // The reflector is the source of truth for the routable address (the
+        // pod IP can change between registration and pick). Fall back to the
+        // selector-reported endpoint, stripping the URL scheme for Envoy.
+        let endpoint = self
+            .reflector
+            .resolve_endpoint(resp.worker_id)
+            .unwrap_or_else(|| strip_scheme(&resp.endpoint).to_string());
```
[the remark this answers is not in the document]
atchernych: Will implement the minimal consistency fix (this branch - NoEndpoints; epp-selector-reservations branch - free-on-that-path)

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        // Mint an EPP-side reservation id (not the gateway request id): the
+        // booking key can't collide with a reused `x-request-id` and stays
+        // EPP-known, so it's releasable even if the reserve response is lost.
+        // Recorded before the call; the error arm below cleans up a failed reserve.
+        let reservation_id = uuid::Uuid::new_v4().to_string();
+        self.reservations
+            .lock()
+            .unwrap()
+            .insert(req.request_id.clone(), reservation_id.clone());
```
tmonty12: **[P1] Keep reservation ownership separate from `x-request-id`.** `req.request_id` is client/upstream controlled when that header is present, so concurrent requests can reuse it. If A and B use the same value, B overwrites this entry; A's first-token/completion callback can then mutate or free B's reservation while A's booking is left behind. Mint an unconditional per-ext-proc-stream lifecycle ID and use it directly as the selector reservation/callback token; keep `x-request-id` only for observability. That should also let us delete this global map and mutex.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+            model_name: self.model_name.clone(),
+            reservation_id: reservation_id.clone(),
+            token_ids: tokens,
+            allowed_worker_ids: Some(allowed),
+            priority_jump: (priority_jump > 0.0).then_some(priority_jump),
+            strict_priority: (strict_priority > 0).then_some(strict_priority),
+            cache_salt,
+        };
+
+        let resp = match self.selector.select_and_reserve(select_req).await {
```
tmonty12, line 316: **[P1] Cancel queued selection when the ext-proc stream closes.** `server.rs:463-469` awaits `pick()` inline, so while this call is queued it does not observe `tx.closed()` or inbound cancellation. The abandoned request retains its `AdmissionLease`/queue position, can later book capacity for a request Envoy has already dropped, and is cleaned only if selection eventually returns. Make the ext-proc boundary race selection against stream closure (for example, a biased `tokio::select!` on `tx.closed()` and the pick future) so dropping the selection future triggers the queue's existing RAII retraction. Cancellation must also remove any local lifecycle state created before this await; eliminating the request-ID map in the adjacent finding makes that substantially safer.

deploy/inference-gateway/ext-proc/src/main.rs
```diff
@@ -162,19 +150,38 @@ async fn main() -> Result<()> {
...
+    if standalone {
+        let selector_cfg = EppStandaloneConfig::from_env()?;
+        tracing::info!(
+            inference_pool_name = %selector_cfg.inference_pool_name,
+            model_name = %selector_cfg.model_name,
+            block_size = selector_cfg.block_size,
+            "Initializing standalone selector mode (no Dynamo runtime)..."
+        );
+        let router = dynamo_ext_proc::EppRouter::from_selector(selector_cfg).await?;
+        let ready = router.reflector_ready();
```
tmonty12: **[P1] Make readiness live and InferencePool-aware.** This flag represents only completion of the Pod LIST, and `serve()` changes health to `SERVING` in one direction. A missing, malformed, or deleted `PoolState` can therefore empty routing state while health remains serving; conversely, `wait_for_selector_ready()` delays a valid empty pool while waiting for a worker. Standalone readiness should be `pod_store_synced && pool_state.is_some()` and should drive both upgrades and downgrades of gRPC health. A valid pool with zero ready endpoints remains ready; requests can still receive 503 until capacity appears.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+                    ),
+                ],
+                ..Default::default()
+            });
+        }
+
+        let (tokens, priority_jump, strict_priority, body_salt) = self
+            .tokenize(&req.body)
+            .await
+            .map_err(|e| PickError::TokenizationFailed(e.to_string()))?;
```
tmonty12: **[P1] Preserve typed vLLM Render failures across the picker boundary.** `VllmRenderClient` already distinguishes unavailable, timeout, upstream-status, and invalid-response failures, but this stringifies all of them as `TokenizationFailed`, which `server.rs:733-736` turns into HTTP 400. Reserve 400/422 for malformed client input; map renderer unavailability to 503, timeout to 504, and renderer 5xx/invalid responses to 502 or 503 as appropriate. Keep semantic error variants through `PickError` and avoid returning raw renderer internals to the client.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        // KV cache-isolation namespace: the `x-tenant-id` header overrides the
+        // body's `cache_salt`, matching the frontend's header-routing precedence.
+        let cache_salt = tenant_id_header(&req.headers).or(body_salt);
```
tmonty12: **[P1] Descope `cache_salt` until standalone can carry it end to end.** This applies `x-tenant-id`/`nvext.cache_salt` to selector hashing, but `server.rs:304-333` forwards the raw vLLM request body; unlike the Dynamo frontend flow, those inputs are not translated into the engine request. The selector namespace can therefore diverge from the backend KV-event namespace. For M1, remove standalone tenant/cache-salt extraction and `SelectRequest.cache_salt` wiring.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    /// Tokenize a chat-completions request body for routing. Returns
+    /// `(token_ids, priority_jump, strict_priority, cache_salt)`, where
+    /// `cache_salt` is the body-derived KV cache namespace (`nvext.cache_salt`,
+    /// with a legacy top-level fallback) via Dynamo's public precedence rules.
+    async fn tokenize(&self, request_body: &[u8]) -> Result<(Vec<u32>, f64, u32, Option<String>)> {
+        let request: dynamo_llm::types::openai::chat_completions::NvCreateChatCompletionRequest =
+            serde_json::from_slice(request_body)?;
+        let priority_jump = extract_priority_jump(&request);
```
tmonty12: **[P1] Use one public canonical priority resolver.** These local helpers read only `nvext.agent_hints`, so `x-dynamo-request-priority` and `x-dynamo-request-strict-priority` do not get their documented header-over-body precedence. Put a public, transport-neutral resolver next to the existing constants and `apply_header_routing_overrides` in `lib/llm/src/protocols/common/extensions.rs`; have it accept the body hints plus already-looked-up raw header values so it works with both `HeaderMap` and Envoy protobuf headers. A valid header, including zero or negative priority, must override the body; malformed/missing values should fall back independently, and `latency_sensitivity` should be used only when no effective priority exists. Reuse it from each EPP/preprocessor entry point instead of duplicating this policy.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+) -> u32 {
+    request
+        .nvext
+        .as_ref()
+        .and_then(|n| n.agent_hints.as_ref())
+        .and_then(|h| h.strict_priority)
+        .unwrap_or(0)
+}
+
+#[cfg(test)]
```
tmonty12, line 466: **[P1] Add lifecycle coverage for the capacity this PR books.** The current tests cover tenant parsing and subset matching only. Add async tests for successful reserve/final free, selection failure, stale selected worker, queued disconnect/cancellation, first-token prefill completion exactly once, and concurrent duplicate `x-request-id` values proving that one stream cannot mutate or free another stream's booking.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+        req: &RequestInfo,
+        _endpoints: &[Endpoint],
+    ) -> Result<PickResult, PickError> {
+        if !self.reflector_ready.load(Ordering::Acquire) {
+            return Err(PickError::RoutingFailed(
+                "pod reflector cache not ready".to_string(),
+            ));
+        }
+
+        let mut allowed = self.reflector.ready_worker_ids();
```
tmonty12: **[P2] Avoid materializing every ready worker on every request.** `ready_worker_ids()` allocates and fills an O(worker-count) `HashSet` on the hot path; subset routing then allocates another candidate set/result set and repeatedly resolves/clones endpoints. Keep the ordinary path allocation-free by letting the selector own eligibility or by using a generation-backed shared snapshot, and build/use an indexed filter only when a non-empty candidate subset is present.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+                        "x-dynamo-routing-mode".to_string(),
+                        "aggregated".to_string(),
+                    ),
+                ],
+                ..Default::default()
+            });
+        }
+
+        let (tokens, priority_jump, strict_priority, body_salt) = self
+            .tokenize(&req.body)
```
tmonty12: **[P2] Keep request-body ownership as `bytes::Bytes`.** The ext-proc path currently accumulates a `Vec<u8>`, copies it into `RequestInfo`, sends another owned body to reqwest, and copies the unchanged request again for forwarding/protobuf chunks. Freeze once at request EOS, pass cheap `Bytes` clones to Render and forwarding, and allocate only when an actual JSON body mutation is required. In this standalone path `token_ids` is `None`, so forwarding the original request should not require another full-body copy.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,393 @@
...
+            self.reservations.lock().unwrap().remove(&req.request_id);
+            if let Err(cleanup) = self.selector.free_reservation(&reservation_id).await {
+                tracing::debug!(request_id = %req.request_id, %reservation_id, error = %cleanup, "reservation cleanup after stale selection");
+            }
+            return Err(PickError::NoEndpoints);
+        };
+
+        Ok(PickResult {
+            endpoint,
+            headers: vec![
```
tmonty12: **[P2] Emit only the GAIE destination endpoint required for routing.** `build_request_header_response` already derives the destination mutation from `PickResult.endpoint`; `x-dynamo-worker-instance-id` and `x-dynamo-routing-mode` add a standalone backend-header contract that aggregated routing does not need. Remove both here and in the body-less path unless a concrete downstream consumer and compatibility contract are established.

tmonty12 (commented):
Reviewed the standalone EPP request wiring against the current head and the milestone-1 DEP contract. I am leaving the remaining lifecycle, readiness, error-mapping, cache-isolation, priority, coverage, and request-hot-path findings inline.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: yes, it should be there, will fix.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: fixed

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: We cannot let the selector own eligibility. The allowed set is not redundant bookkeeping; it's a freshness constraint. We may have issues during scale down for example.
The 2nd suggestion is not sufficient.
I will add a reflector method that filters by subset in a single borrow() with no per-worker string clones, and only run it when candidate_subset is non-empty. 
Trim the ordinary path cheaply: we can keep passing the reflector's ready set but pre-size it (HashSet::with_capacity(endpoints.len())) so it's one allocation, no rehashing, and read it from a single borrow.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: 3 redundant copies (RequestInfo, render, forward-to_vec). Will remove by the Bytes plumbing.

1 serialization copy to send to Envoy. This one is unavoidable.

1 assignment copy at the proto field. This is removable, but via a crate-wide codegen change so will be in the follow-up.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: x-dynamo-routing-mode is shared for agg and disag. The disag is not yet implemented, are you sure you will not need it?

deploy/inference-gateway/ext-proc/src/epp_router.rs
tmonty12: Follow-up after re-review at fe2342e: racing pick against tx.closed() handles the abandoned-await path, and I am fine treating the closed scheduler response receiver as the queued-work cancellation contract for this PR. One adapter-side handoff race remains: the scheduler can book and send its response immediately before the biased tx.closed() branch wins. server.rs then drops pick() before storing booking_id/body_routed, so final cleanup is skipped. Please keep this scoped to the EPP adapter: arm an EPP-owned reservation cleanup guard when reservation_id is minted; if pick is dropped, have the guard schedule idempotent free_reservation(reservation_id), and disarm it only when the successful pick is immediately adopted by RequestContext. Please add a deterministic test for disconnect after booking but before context adoption.

deploy/inference-gateway/ext-proc/src/main.rs
tmonty12: Follow-up after re-review at fe2342e: the live pool-aware health upgrades and downgrades are fixed. One startup issue remains: EppRouter construction still calls wait_for_selector_ready(), which waits up to 30 seconds for an actual Ready worker. A valid, synchronized InferencePool with zero endpoints should complete startup and report serving; individual picks can return 503 until capacity appears. Please remove worker availability from the constructor wait and add an empty-pool startup/readiness test.

deploy/inference-gateway/ext-proc/src/epp_router.rs
tmonty12: Follow-up after re-review at fe2342e: preserving typed Render failures and mapping unavailable/timeout/upstream failures to 503/504/502 is good. One boundary issue remains: status.is_client_error() still maps every renderer 4xx to client HTTP 400. Please use an explicit mapping: preserve known request-validation statuses such as 400/413/422, map renderer 429 to 503, and map dependency auth/path/config failures such as 401/403/404 and other unexpected renderer 4xx responses to 502. Add focused status-mapping tests and continue keeping raw renderer details out of the client response.

deploy/inference-gateway/ext-proc/src/epp_router.rs
tmonty12: I think that selector owning eligibility makes the most sense, i.e. just passing `None` for `allowed_worker_ids`. There is a very small gap - order of single ms - between the reflector receiving a pod event and the SelectionService asynchronously receiving this update. We already are dealing will staleness in the system as Pod readiness can take substantially longer so I don't think this warrants the code/space complexity for having the reflector pass the allowed worker ids.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: added

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,481 @@
...
+        })
+    }
+
+    /// Overall EPP readiness for the gRPC health signal: the pod reflector is
+    /// ready (workers synced + pool resolved) AND, in replicated mode, the peer
+    /// set has finished its initial sync. Polled by the health mirror in `main`.
+    pub fn is_ready(&self) -> bool {
+        compute_ready(
+            self.reflector_ready.load(Ordering::Acquire),
+            self.peer_ready.as_ref().map(|p| p.load(Ordering::Acquire)),
```
tmonty12, line 108: [P1] Peer readiness only proves EndpointSlice/socket reconciliation. Replica sync publishes future deltas and has no active-load snapshot, so a joining/restarted EPP can advertise SERVING while missing existing requests. Gate readiness on a state bootstrap/snapshot and cover rolling restart.

deploy/inference-gateway/ext-proc/src/server.rs
```diff
@@ -492,7 +537,13 @@ impl<P: EndpointPicker> ExternalProcessor for ExtProcServer<P> {
...
                             {
                                 ctx.prefill_complete_signaled = true;
-                                picker.on_prefill_complete(&ctx.request_id).await;
+                                // Hand back the picker's own booking id (falls
+                                // back to request_id if it booked nothing).
+                                let booking_id = ctx
+                                    .booking_id
+                                    .clone()
+                                    .unwrap_or_else(|| ctx.request_id.clone());
+                                picker.on_prefill_complete(&booking_id).await;
```
tmonty12: [P1] This await is on the first-token forwarding path. Standalone `prefill_complete` can block on the admission queue send/ack with no timeout, stalling TTFT indefinitely. Make bookkeeping fail-open with a bounded timeout or dispatch it off-path.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,481 @@
...
+            serde_json::from_slice(&request_body).map_err(TokenizeError::InvalidBody)?;
+        let resolved = resolve_request_priority(
+            hints.nvext.as_ref().and_then(|n| n.agent_hints.as_ref()),
+            priority_header.as_deref(),
+            strict_priority_header.as_deref(),
+        );
+        // Moves the `Bytes` into reqwest (zero-copy) rather than copying.
+        let token_ids = self
+            .renderer
+            .render_chat(request_body)
```
tmonty12, line 135: [P2] Every standalone request performs a render HTTP call, but ext-proc admits one task per HTTP/2 stream and only bounds TCP connections. Add a shared in-flight limit/load shedding around render calls to prevent unbounded task and request growth.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,481 @@
...
+            }
+            TokenizeError::Render(e) => {
+                tracing::warn!(request_id, error = %e, "vLLM render failed");
+                match e {
+                    VllmRenderError::Unavailable { .. } => PickError::TokenizerUnavailable,
+                    VllmRenderError::Timeout { .. } => PickError::TokenizerTimeout,
+                    // A 4xx from the renderer means the client's payload was
+                    // rejected (client error → 400); a 5xx is the renderer's
+                    // fault (→ 502), as is a response that breaks its contract.
+                    VllmRenderError::UpstreamStatus { status, .. } if status.is_client_error() => {
```
tmonty12: [P2] `status.is_client_error()` also includes 401/403/404/429; these are auth, configuration, or overload failures, not invalid inference payloads. Map only renderer validation statuses such as 400/422 to 400 and preserve the rest as upstream errors.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,481 @@
...
+/// Overall EPP health: pod readiness AND, when replicated (`peer_ready = Some`),
+/// the initial peer sync. `None` means no replication → pod readiness alone.
+fn compute_ready(pod_ready: bool, peer_ready: Option<bool>) -> bool {
+    pod_ready && peer_ready.unwrap_or(true)
+}
+
+/// True if a scheme-less `ip:port` endpoint is covered by an Envoy subset,
+/// matching either the full `ip:port` or the bare `ip`.
+fn endpoint_in_subset(endpoint: &str, candidates: &HashSet<&str>) -> bool {
+    let ip = endpoint.split(':').next().unwrap_or("");
```
tmonty12: [P2] `split(':')` breaks bracketed IPv6 (`[fd00::10]:8000` becomes `[fd00`). Parse the authority as `SocketAddr` and compare normalized IPs; add bare-IPv6 subset coverage.

deploy/inference-gateway/ext-proc/src/server.rs
```diff
@@ -745,7 +815,12 @@ impl ExtProcError {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::sync::atomic::{AtomicU32, Ordering};
+    use std::collections::HashSet;
```
tmonty12, line 834: [P2] This PR grows `server.rs` from 958 to 1,220 lines. Extract the test module at minimum; preferably move the per-stream lifecycle state machine into a focused module so the shared protocol code remains scannable.

tmonty12 (commented):
Concise follow-up on the rebased head; inline findings only.

deploy/inference-gateway/ext-proc/src/server.rs
tmonty12: Confirmed. `server.rs` already had its inline test module at 958 lines before this PR, and latest `main` has 130 Rust files over 1,000 lines with inline `#[cfg(test)]` modules (including `queue.rs` at 4,585, `protocols.rs` at 2,153, selection core at 1,643, and sibling `epp.rs` at 1,195). Keeping these focused tests colocated is consistent with repository precedent, so I am withdrawing this concern.

deploy/inference-gateway/ext-proc/src/epp_router.rs
```diff
@@ -0,0 +1,481 @@
...
+    candidates.contains(endpoint) || candidates.contains(ip)
+}
```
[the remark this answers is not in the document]
tmonty12: Fixed in c72cafc132. Bare candidate IPs are normalized once as `IpAddr`, worker endpoints are parsed as `SocketAddr`, and coverage now includes exact and bare IPv4/IPv6 matching plus port mismatches.

deploy/inference-gateway/ext-proc/src/epp_router.rs
tmonty12: Fixed in c72cafc132. Bare candidate IPs are normalized once as `IpAddr`, worker endpoints are parsed as `SocketAddr`, and coverage now includes exact and bare IPv4/IPv6 matching plus port mismatches.

deploy/inference-gateway/ext-proc/src/server.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: Is gating readiness on a load snapshot is the right fix? Active load is transient, self-healing state and readiness on it contradicts the idea that  readiness reflects stable prerequisites, not volatile state rule and would delay replicas entering rotation during a rolling restart (worse availability). It's also not a correctness bug: requests still route to valid workers; the only effect is brief load imbalance bounded by request-drain time, and in a rolling restart it's a single cold replica among warm peers.

@sttts  to comment

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: In practice the gateway's own per-upstream stream ceiling  provides some implicit cap, but will add the global in-flight semaphore with load-shedding similar to how llm-d does it.

deploy/inference-gateway/ext-proc/src/epp_router.rs
atchernych: will fix

deploy/inference-gateway/ext-proc/src/epp_router.rs
sttts: I agree with Anna's description of the impact:

> “Active load is transient, self-healing state” and “the only effect is brief load imbalance bounded by request-drain time.”

Requests still route to valid workers. What is missing on a new replica is the active-load bookkeeping for requests which started before it connected. This is temporary and disappears when these requests drain. During a rolling restart, it is one cold replica among otherwise warm replicas.

## 1. A consistent startup would need snapshot recovery

If we want a stronger guarantee, connecting the sockets is not enough. We need snapshot plus live-event recovery, for example ZeroMQ's [Clone Pattern / CHP](https://rfc.zeromq.org/spec/12/):

1. subscribe to live events,
2. request the active-request snapshot,
3. buffer live events during the snapshot,
4. receive the snapshot sequence,
5. apply buffered events above that sequence, and
6. recover again if there is a gap.

`AddRequest` would add snapshot state, `MarkPrefillCompleted` would update it, and `Free` would remove it.

This is a real recovery protocol, not an additional readiness check.

## 2. We already have most of the recovery pattern in the KV-cache code

The KV-cache indexer already implements the same main pieces:

- The worker serves event ranges or a tree dump: [worker_query_endpoint.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/kv_router/indexer/recovery/worker_query_endpoint.rs#L22-L153).
- Live events are buffered during initial or gap recovery and drained after the recovered sequence: [worker_query_state.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/kv_router/indexer/recovery/worker_query_state.rs#L72-L210).
- Sequence gaps trigger reset/fencing and another recovery: [worker_query.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/kv_router/indexer/recovery/worker_query.rs#L670-L729).

We could reuse this design for active requests. It would still be a shared recovery feature, not a small EPP-only change.

## 3. Frontend/router also starts with incomplete active-load state

With `router_replica_sync` enabled, frontend/router creates an empty tracker, starts its subscriber in the background, and returns it to the scheduler without a snapshot or catch-up barrier: [sequence.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/kv_router/sequence.rs#L383-L476).

Its direct-ZMQ path detects initial and later sequence gaps, but only records metrics and continues applying events: [direct_zmq.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/kv_router/sequence/direct_zmq.rs#L97-L127), [direct_zmq_fan_in.rs](https://github.com/ai-dynamo/dynamo/blob/01cb043a2c577d196918a981856643a5537724fb/lib/llm/src/direct_zmq_fan_in.rs#L464-L478).

EPP should at least provide the same observability: per-publisher sequence tracking plus initial-gap, later-gap, and out-of-order metrics.

## Conclusion

For this PR, I think EPP should keep the same best-effort contract as frontend/router. Readiness means peer discovery and socket setup completed, so future events can flow.

The missing state only affects the load estimate, is temporary, and self-heals as existing requests drain. EPP does not need a stronger consistency guarantee than frontend/router, but it should have the same gap metrics.

If we want consistent active-load startup, we should add snapshot recovery once for both paths, using the existing KV-cache recovery code as the base.

tmonty12: Agreed—deferring this concern. DEP #10321 intentionally keeps request/slot-tracker state replica-local at startup and defines lifecycle replication as best-effort. SelectionService does not expose request-state snapshot/recovery APIs, so adding bootstrap here would require a SelectionService-level consistency protocol outside this EPP integration’s scope. We can track that separately if stronger cross-replica consistency becomes a requirement.

tmonty12 (approved):

sttts:
Follow-ups from the review; I do not think either should block this PR:

- #12267 tracks per-publisher sequence/gap/out-of-order metrics for the EPP replica lifecycle stream. This keeps the current best-effort contract and only brings EPP observability in line with frontend/router; snapshot recovery remains a separate shared concern.
- #12268 tracks admission before streamed request-body buffering plus a cumulative body-size limit. This is defense-in-depth; gateway stream ceilings already provide an operational bound.

For this PR, requests still route to valid workers and both residual concerns are appropriately handled as follow-up work.

sttts (approved):

ishandhanani (approved):

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/pull/11542/commits/01cb043a2c577d196918a981856643a5537724fb

PR #11548, opened 2026-07-10
feat(router): add provider-keyed tracking hashes

jthomson04:
## Summary

- add opt-in `keyed-xxh3-v1` identities for router-derived active tracking, backed by a zeroized 32-byte provider key and provider-managed key ID
- preserve public primary-indexer hashes, the legacy public Rust hashing API, precomputed-hash inputs, replica envelopes, slot-tracker APIs, KV events, standalone indexer APIs, and all existing wire formats
- make environment parsing fallible for startup callers, reject invalid keyed configuration before router initialization, and retain the non-fallible configuration helper as a compatibility wrapper with a clear panic
- require a nonempty model identity for keyed embedded routers, share the default routing-group constant across embedded and selection paths, and log only the algorithm and key ID when replica synchronization initializes
- avoid calculating discarded public block hashes for reservations and use one canonical normal/Eagle complete-block counter across deterministic and random tracking paths
- centralize raw-token tracking identity policy behind one public context method used by embedded, replay, lifecycle, and standalone-selection paths; keep the lower-level reusable-hash operation crate-private
- document the flags, environment variables, trusted-plane boundary, rotation procedure, and XXH3's non-PRF limitation

### Selection behavior change

Token-bearing standalone selection requests now honor `router_assume_kv_reuse=false` by using random tracking identities, matching the embedded router. Public block hashes still drive indexer lookups, so this changes active-load tracking behavior without changing indexer identity.

### Trust and rotation boundary

Precomputed sequence hashes remain trusted opaque inputs. Producers are responsible for using the deployment's configured tracking hash function and maintaining lineage uniqueness. Replica synchronization also transports opaque hashes: it logs the local algorithm and key ID at initialization but does not negotiate or carry epoch metadata.

Keyed tracking therefore assumes a trusted internal plane and identically configured hash producers. Seeded XXH3 is not a PRF or MAC. Rotation is a coordinated hard cut: change the key and key ID together, restart all producers, and flush or recreate derived tracker state. This PR does not add authentication, mixed-epoch detection, negotiation, dual reads, or live reload.

## Benchmarks

Criterion, 100 samples per arm:

| Complete blocks | Public XXH3 | Keyed XXH3 | Benchmark-only keyed BLAKE3 | Keyed XXH3 overhead |
| 32 | 108.90 ns | 360.99 ns | 6.392 us | +252.09 ns (+231%) |
| 128 | 470.50 ns | 968.71 ns | 25.499 us | +498.21 ns (+106%) |

The keyed-XXH3 regression is sustained above 5%. It comes from the required keyed scope derivation plus a second canonical per-block hashing pass; public primary-indexer block hashes cannot be reused as keyed tracking hashes. Keyed XXH3 remains about 17.7x faster than the keyed-BLAKE3 block-and-chain comparison at 32 blocks and 26.3x faster at 128 blocks. The BLAKE3 arm is benchmark-only and is not exposed as a production algorithm.

## Validation

- rebased onto `origin/main` at `ee1d3b35f9`; `git range-diff` verified both PR commits were preserved after the final replay
- audited main's live-admission progress, active-request expiry, replica session-affinity, and DC CKF relay changes; live-admission reuses the centralized active-tracking identities, while CKF/KV-event identities intentionally remain public
- `git diff --check refs/remotes/origin/main..HEAD`
- `cargo test -p dynamo-kv-router --features standalone-selection,standalone-slot-tracker --lib` (913 passed, 2 ignored)
- `cargo test -p dynamo-llm --lib` (1,729 passed, 5 ignored)
- `cargo check --manifest-path lib/bindings/python/Cargo.toml --features select-service,slot-tracker`
- `cargo check -p dynamo-kv-router --features standalone-selection,standalone-slot-tracker --benches`

## Follow-up

[DEP 11971](https://github.com/ai-dynamo/dynamo/issues/11971) is the existing draft architecture gate for this work. It records the precomputed-trust, replica-negotiation, threat-model, and `IndexerDomainId`/tracking-domain relationship concerns that require architectural review; this PR does not modify the DEP.

## Summary by CodeRabbit

* **New Features**
  * Added configurable router tracking hashes, including public and keyed modes.
  * Added support for provider-managed tracking keys and key identifiers through CLI, environment variables, and Python APIs.
  * Keyed tracking preserves public indexer hashes while providing scoped sequence identities.

* **Bug Fixes**
  * Invalid router tracking configuration now prevents startup with clear initialization errors.

* **Documentation**
  * Documented configuration, validation requirements, hash behavior, and key rotation procedures.

### DIFF
--- a/lib/kv-router/src/scheduling/config.rs
+++ b/lib/kv-router/src/scheduling/config.rs
@@ -15,7 +15,9 @@ use crate::protocols::{
     BlockHashOptions, LocalBlockHash, complete_block_count, compute_block_hash_for_seq,
     compute_seq_hash_for_block,
 };
-use crate::tracking_hash::{TrackingHashAlgorithm, TrackingHashContext, TrackingHashScope};
+use crate::tracking_hash::{
+    TrackingHashAlgorithm, TrackingHashContext, TrackingHashScope, validate_tracking_hash_options,
+};
 
 const fn default_track_prefill_tokens() -> bool {
     true
@@ -627,15 +629,12 @@ pub struct KvRouterConfig {
     pub router_track_prefill_tokens: bool,
 
     /// Hash algorithm used for router-derived active-sequence identities.
-    #[serde(default)]
     pub router_tracking_hash: TrackingHashAlgorithm,
 
     /// File containing the 32-byte provider key used by keyed tracking mode.
-    #[serde(default)]
     pub router_tracking_key_file: Option<PathBuf>,
 
     /// Provider-managed epoch identifier mixed into keyed tracking scope derivation.
-    #[serde(default)]
     pub router_tracking_key_id: Option<String>,
 
     /// Optional model for estimating effective prompt-side prefill load over time.
@@ -845,29 +844,11 @@ impl TryFrom<KvRouterConfigSerde> for KvRouterConfig {
 }
 
 fn validate_kv_router_config(config: &KvRouterConfig) -> Result<(), String> {
-    match config.router_tracking_hash {
-        TrackingHashAlgorithm::PublicXxh3V1 => {
-            if config.router_tracking_key_file.is_some() || config.router_tracking_key_id.is_some()
-            {
-                return Err(
-                    "router tracking key options require router_tracking_hash=keyed-xxh3-v1"
-                        .to_string(),
-                );
-            }
-        }
-        TrackingHashAlgorithm::KeyedXxh3V1 => {
-            if config.router_tracking_key_file.is_none() {
-                return Err("keyed-xxh3-v1 requires router_tracking_key_file".to_string());
-            }
-            let valid_key_id = config
-                .router_tracking_key_id
-                .as_deref()
-                .is_some_and(|value| !value.is_empty() && value.trim() == value);
-            if !valid_key_id {
-                return Err("keyed-xxh3-v1 requires a nonempty router_tracking_key_id".to_string());
-            }
-        }
-    }
+    validate_tracking_hash_options(
+        config.router_tracking_hash,
+        config.router_tracking_key_file.is_some(),
+        config.router_tracking_key_id.as_deref(),
+    )?;
     if config.router_track_output_blocks && !config.router_track_active_blocks {
         return Err(
             "router_track_output_blocks requires router_track_active_blocks=true".to_string(),
--- a/lib/kv-router/src/scheduling/config.rs
+++ b/lib/kv-router/src/scheduling/config.rs
@@ -1156,13 +1156,13 @@ fn random_sequence_hashes(num_blocks: usize) -> Vec<u64> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::identity::RoutingPartitionRef;
     use crate::protocols::{BlockExtraInfo, BlockMmObjectInfo, compute_seq_hash_for_block};
     use std::collections::HashMap;
 
     fn test_tracking_scope(block_size: u32) -> TrackingHashScope<'static> {
         TrackingHashScope {
-            model_name: "model",
-            routing_group: "default",
+            partition: RoutingPartitionRef::new("model", "default"),
             block_size,
         }
     }
--- a/lib/kv-router/src/scheduling/config.rs
+++ b/lib/kv-router/src/scheduling/config.rs
@@ -11,7 +11,10 @@ use std::time::Duration;
 use derive_builder::Builder;
 use serde::{Deserialize, Serialize};
 
-use crate::protocols::{BlockHashOptions, LocalBlockHash};
+use crate::protocols::{
+    BlockHashOptions, LocalBlockHash, complete_block_count, compute_block_hash_for_seq,
+    compute_seq_hash_for_block,
+};
 use crate::tracking_hash::{TrackingHashAlgorithm, TrackingHashContext, TrackingHashScope};
 
 const fn default_track_prefill_tokens() -> bool {
@@ -94,8 +97,24 @@ pub fn apply_deprecated_overlap_score_weight_override(
 }
 
 /// Build a [`KvRouterConfig`] from defaults and standard Dynamo environment variables.
+///
+/// # Panics
+///
+/// Panics when `DYN_ROUTER_TRACKING_HASH` is not a supported algorithm. Startup
+/// paths should use [`try_kv_router_config_from_dynamo_env`] to report the error.
 pub fn kv_router_config_from_dynamo_env() -> KvRouterConfig {
-    let config = kv_router_config_from_lookup(|key| env::var(key).ok());
+    try_kv_router_config_from_dynamo_env()
+        .unwrap_or_else(|error| panic!("invalid Dynamo router environment configuration: {error}"))
+}
+
+/// Build a [`KvRouterConfig`] from standard Dynamo environment variables.
+pub fn try_kv_router_config_from_dynamo_env() -> Result<KvRouterConfig, String> {
+    let config = kv_router_config_from_lookup(|key| env::var(key).ok())?;
+    log_env_config(&config);
+    Ok(config)
+}
+
+fn log_env_config(config: &KvRouterConfig) {
     tracing::info!(
         overlap_score_credit = config.overlap_score_credit,
         overlap_score_credit_decay = config.overlap_score_credit_decay,
@@ -119,10 +138,11 @@ pub fn kv_router_config_from_dynamo_env() -> KvRouterConfig {
         router_predicted_ttl_secs = ?config.router_predicted_ttl_secs,
         "KvRouterConfig initialized (DYN_* env overrides applied)"
     );
-    config
 }
 
-fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvRouterConfig {
+fn kv_router_config_from_lookup(
+    get_env: impl Fn(&str) -> Option<String>,
+) -> Result<KvRouterConfig, String> {
     fn parse_f64(get_env: &impl Fn(&str) -> Option<String>, key: &str) -> Option<f64> {
         get_env(key).and_then(|value| value.parse().ok())
     }
@@ -180,7 +200,7 @@ fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvR
         config.router_track_prefill_tokens = value;
     }
     if let Some(value) = get_env("DYN_ROUTER_TRACKING_HASH") {
-        config.router_tracking_hash = value.parse().unwrap_or(TrackingHashAlgorithm::Invalid);
+        config.router_tracking_hash = value.parse()?;
     }
     if let Some(value) = get_env("DYN_ROUTER_TRACKING_KEY_FILE") {
         config.router_tracking_key_file = Some(value.into());
@@ -227,7 +247,7 @@ fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvR
         config.router_predicted_ttl_secs = Some(value);
     }
 
-    config
+    Ok(config)
 }
 
 fn apply_deprecated_overlap_score_weight_override_option(
@@ -844,15 +864,8 @@ fn validate_kv_router_config(config: &KvRouterConfig) -> Result<(), String> {
                 .as_deref()
                 .is_some_and(|value| !value.is_empty() && value.trim() == value);
             if !valid_key_id {
-                return Err(
-                    "keyed-xxh3-v1 requires a nonempty router_tracking_key_id".to_string(),
-                );
+                return Err("keyed-xxh3-v1 requires a nonempty router_tracking_key_id".to_string());
             }
-        }
-        TrackingHashAlgorithm::Invalid => {
-            return Err(
-                "router_tracking_hash must be public-xxh3-v1 or keyed-xxh3-v1".to_string(),
-            );
         }
     }
     if config.router_track_output_blocks && !config.router_track_active_blocks {
@@ -1067,7 +1080,54 @@ impl KvRouterConfig {
     /// - `None` if `router_track_active_blocks` is false
     /// - Random hashes if `router_track_active_blocks` is true but `router_assume_kv_reuse` is false
     /// - Actual sequence hashes if both are true
+    /// # Panics
+    ///
+    /// Panics in keyed mode because the legacy interface has no initialized
+    /// [`TrackingHashContext`]. Keyed callers must use
+    /// [`Self::compute_seq_hashes_for_tracking_with_context`].
     pub fn compute_seq_hashes_for_tracking(
+        &self,
+        tokens: &[u32],
+        block_size: u32,
+        config_override: Option<&RouterConfigOverride>,
+        hash_options: BlockHashOptions<'_>,
+        precomputed_block_hashes: Option<&[LocalBlockHash]>,
+    ) -> Option<Vec<u64>> {
+        assert_eq!(
+            self.router_tracking_hash,
+            TrackingHashAlgorithm::PublicXxh3V1,
+            "compute_seq_hashes_for_tracking cannot be used with keyed tracking; initialize a TrackingHashContext and call compute_seq_hashes_for_tracking_with_context"
+        );
+
+        if !self.router_track_active_blocks {
+            return None;
+        }
+
+        let num_blocks = complete_block_count(
+            tokens.len(),
+            block_size,
+            hash_options.is_eagle.unwrap_or(false),
+        );
+        if num_blocks == 0 {
+            return Some(Vec::new());
+        }
+
+        if self.assume_kv_reuse(config_override) {
+            let block_hashes = match precomputed_block_hashes {
+                Some(block_hashes) => block_hashes,
+                None => {
+                    let computed = compute_block_hash_for_seq(tokens, block_size, hash_options);
+                    return Some(compute_seq_hash_for_block(&computed));
+                }
+            };
+            Some(compute_seq_hash_for_block(block_hashes))
+        } else {
+            Some(random_sequence_hashes(num_blocks))
+        }
+    }
+
+    /// Compute sequence hashes with a router-initialized tracking-hash context.
+    pub fn compute_seq_hashes_for_tracking_with_context(
         &self,
         tracking_hash: &TrackingHashContext,
         scope: TrackingHashScope<'_>,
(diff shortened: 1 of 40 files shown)


lib/kv-router/src/services/selection/input.rs
```diff
@@ -123,23 +160,40 @@ impl PromptRequest {
...
+        tracking.scope.block_size,
         BlockHashOptions {
             block_mm_infos,
             lora_name,
             cache_namespace,
             is_eagle: Some(is_eagle),
         },
     );
-    let sequence_hashes = compute_seq_hash_for_block(&block_hashes);
+    let sequence_hashes = if tracking.assume_kv_reuse {
```
jthomson04: **Behavior change worth calling out in the PR description**: at the merge base, the selection service always computed real sequence hashes here — `router_assume_kv_reuse=false` was effectively ignored on this path. This hunk makes the standalone service honor it with randomized tracking identities (matching the embedded router). It's a reasonable consistency fix and `disabled_kv_reuse_keeps_public_indexer_hashes_and_randomizes_tracking` pins it, but it's a behavior change independent of keyed hashing, so it should be mentioned in the summary (or split out) for anyone bisecting a tracking-accuracy shift.

lib/kv-router/src/services/selection/input.rs
```diff
@@ -73,17 +80,17 @@ impl PromptRequest {
 
     pub(super) fn normalize_for_reservation(
```
jthomson04, line 80: **Keyed-mode hole to settle in the DEP**: the hash-bearing path (no `token_ids`) still books tracking slots with caller-supplied public-domain `sequence_hashes`, even when the service runs in keyed mode. Consequences: (a) identical content arriving via tokens vs. precomputed hashes won't dedup in active-block accounting, and (b) any caller that can send precomputed hashes can inject arbitrary tracking identities regardless of the key. If the selection API is trusted-plane-only that may be acceptable, but the DEP should state it explicitly — or keyed mode should warn/reject precomputed-hash requests.

lib/kv-router/src/tracking_hash.rs
```diff
@@ -0,0 +1,527 @@
...
+    pub model_name: &'a str,
+    pub routing_group: &'a str,
+    pub block_size: u32,
+}
+
+/// Runtime tracking-hash state. Secret bytes are intentionally excluded from
+/// serialization and debug output.
+pub struct TrackingHashContext {
+    algorithm: TrackingHashAlgorithm,
+    key_id: Option<Box<str>>,
```
jthomson04, line 96: **No cross-replica key-consistency check**: replica sync exchanges tracking hashes as opaque u64s, so replicas with mismatched keys or `key_id`s (e.g., a rolling restart mid-rotation) silently diverge — dedup quietly degrades with no error. Since `key_id` exists to manage epochs, consider carrying it (or a key-check value like `blake3::keyed_hash(key, b"key-check")`) in the replica-sync handshake, or at minimum logging `key_id` at sync establishment.

jthomson04 (commented):
Automated review (Claude Code). **Verdict: solid, careful implementation — no correctness blockers found.** Domain separation is done right, the public path is pinned bit-compatible, and startup fails fast on bad config. Inline comments cover the specific findings; items without a good anchor are below.

**Verified while reviewing:**
- No cross-domain identity leakage: slot freeing is by `request_id`, `RoutingDecisionHashes` (public-domain) only feeds decision recording/replay, and `get_overlap_scores` now uses `block_hashes_for_indexer`, so indexer matching stays public. The two hash domains combine only arithmetically, never by identity match.
- Keyed framing's empty-string normalization matches the public path's "empty = absent" rule, so both modes agree on identity granularity.
- Test coverage is strong: pinned keyed vectors, scope-sensitivity matrix, key-loading failure modes without byte leakage, config validation, Python CLI/env flow, and a criterion bench.

**Not anchored inline:**
- **Docs**: no `docs/` update for `--router-tracking-hash`/`--router-tracking-key-file`/`--router-tracking-key-id` or the `DYN_ROUTER_TRACKING_*` env vars — should land alongside the DEP before ready-for-review.
- **Public Rust API break**: `compute_seq_hashes_for_tracking` and the `normalize_for_*` signatures changed; fine if that matches repo policy, but out-of-tree consumers will need updates.
- **Minor efficiency**: `normalize_for_reservation`'s token path in keyed mode computes public block hashes and then discards them (`NormalizedReservation` keeps only sequence hashes + ISL) — that pass is skippable in keyed mode.

**Before marking ready for review:** mention the `assume_kv_reuse` selection-service behavior change in the description, settle the precomputed-hash-path and XXH3-seed-recovery stances in the DEP, and add the flag docs.

lib/kv-router/src/tracking_hash.rs
```diff
@@ -0,0 +1,527 @@
...
+    provider_key: Option<[u8; KEY_SIZE]>,
```
jthomson04: Minor hardening: the provider key sits in memory for the process lifetime without zeroization on drop — the `zeroize` crate is cheap insurance for a long-lived secret.

lib/kv-router/src/tracking_hash.rs
```diff
@@ -0,0 +1,527 @@
...
+                );
+                compute_seq_hash_for_block_with_seed(&block_hashes, chain_seed)
+            }
+            TrackingHashAlgorithm::Invalid => {
+                unreachable!("invalid tracking hash algorithms cannot construct a context")
+            }
+        }
+    }
+
+    fn derive_seeds(
```
jthomson04, line 231: **Threat-model note for the DEP**: keyed BLAKE3 protects the provider key itself, but the per-scope XXH3 seeds are the effective secret for identity unlinkability, and XXH3 makes no PRF/MAC claim — an observer with tracking hashes for chosen token sequences can plausibly recover a scope's seed. If tracking hashes stay inside the trusted plane (replica sync, internal state) that's fine, but the DEP should say so explicitly. A truncated keyed-BLAKE3-per-block variant is worth one benchmark line as the conservative comparison, given the keyed path already pays ~2.5× vs public.

lib/llm/src/kv_router.rs
```diff
@@ -256,6 +258,8 @@ where
     ) -> Result<Self> {
         let kv_router_config = kv_router_config.unwrap_or_default();
         kv_router_config.validate()?;
+        let tracking_hash = TrackingHashContext::from_config(&kv_router_config)?;
+        let tracking_model_name = model_name.clone().unwrap_or_default();
```
jthomson04: In keyed mode an absent `model_name` yields an empty-string scope component, weakening scope separation between routers that share a key. Consider failing (or at least warning) at construction when `router_tracking_hash=keyed-xxh3-v1` and `model_name` is `None`.

lib/llm/src/kv_router.rs
```diff
@@ -384,6 +390,14 @@ where
         self.is_eagle
     }
 
+    fn tracking_hash_scope(&self) -> TrackingHashScope<'_> {
+        TrackingHashScope {
+            model_name: &self.tracking_model_name,
+            routing_group: "default",
```
jthomson04: Worth a one-line comment that this constant must match the selection service's default routing group — keyed identities only align across router types if both derive the same scope.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -168,6 +170,15 @@ fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvR
     if let Some(value) = parse_bool(&get_env, "DYN_ROUTER_TRACK_PREFILL_TOKENS") {
         config.router_track_prefill_tokens = value;
     }
+    if let Some(value) = get_env("DYN_ROUTER_TRACKING_HASH") {
+        config.router_tracking_hash = value.parse().unwrap_or(TrackingHashAlgorithm::Invalid);
```
jthomson04: `KvRouterConfig` derives `Serialize` field-by-field, and `TrackingHashAlgorithm::Invalid` is `#[serde(skip)]` — so serializing an env-parsed-but-not-yet-validated config errors at runtime. The window is tiny since validation runs at startup, but parsing at the validation boundary (keeping the raw string here) would avoid a sentinel variant in the domain enum entirely.

lib/kv-router/src/services/selection/core/mod.rs
```diff
@@ -108,21 +110,43 @@ impl SelectionCore {
         indexer_threads: usize,
         cancel_token: CancellationToken,
     ) -> Self {
-        Self::new_inner(kv_router_config, indexer_threads, cancel_token, None, true)
+        Self::try_new_local(kv_router_config, indexer_threads, cancel_token)
+            .expect("selection tracking hash configuration must be valid")
```
jthomson04: `new_local` is referenced as public API in `docs/components/router/standalone-selection.md`, and it now panics on invalid tracking config. Fail-fast at startup is defensible, but consider pointing the docs at `try_new_local` (or deprecating `new_local`).

jthomson04 (commented):
## Code review

### Correctness — verified, no blocking bugs found

Cross-checked the riskiest interactions against the branch:

- **Serde wire compat**: `KvRouterConfigSerde` has container-level `#[serde(default)]`, so configs serialized before this PR deserialize fine with the three new fields defaulting.
- **Legacy API panic trap**: `compute_seq_hashes_for_tracking` now asserts public mode. Every in-repo production caller (3 in `lib/llm/src/kv_router.rs`, 3 in `lib/mocker`) is migrated to the `_with_context` variant in this diff, so the panic is only reachable by out-of-tree consumers who opt into keyed mode — and it's documented with a `# Panics` section and pinned by a `should_panic` test.
- **Random-identity consistency across select→reserve**: with `assume_kv_reuse=false`, token-bearing selections now produce random tracking hashes. `PendingSelection` stores the already-normalized `sequence_hashes`, so the reservation replay path reuses the exact identities booked at selection time — no mismatch.
- **Keyed scope covers the public seed's discriminators**: public mode mixes `cache_namespace`/`lora_name` into the seed via `block_hash_seed` → `compute_salt_hash`; the keyed scope digest frames both plus model, routing group, block size, Eagle flag, and key ID — a strict superset. Canonical block bytes (tokens, MM hashes, Eagle windows) are preserved via `compute_block_hash_for_seq_with_seed`. Framing is injective (tagged + length-prefixed, fixed field order).
- **Eagle block-count fix**: the legacy path previously generated `tokens.len() / block_size` random hashes even in Eagle mode; `complete_block_count` now matches the actual `(len−1)/stride` window count. Subtle behavior change (one fewer random hash for Eagle + `--no-router-assume-kv-reuse`) but a correctness fix, pinned by `randomized_reservation_uses_canonical_complete_block_count`.

### Code quality

Strong overall: `try_` naming for fallible variants matches the codebase, compat wrappers panic loudly with documented `# Panics` sections, `Debug` redacts the key, secret bytes are `Zeroizing`, and the pinned hash-vector tests protect wire stability. Docs are unusually honest about the XXH3-not-a-PRF limitation and rotation semantics.

Minor issues:

- **Duplicated validation logic**: the key-file/key-id consistency rules (including the `trim() == value` rule) live in both `validate_kv_router_config` (`scheduling/config.rs`) and `TrackingHashContext::from_config` (`tracking_hash.rs`). If one drifts, a config could validate but fail context construction (or vice versa). Extract a shared helper, or have `from_config` call `validate_config()` first and drop its own re-checks.
- **Inert serde attributes**: `#[serde(default)]` on the three new `KvRouterConfig` fields does nothing — deserialization goes entirely through `try_from = "KvRouterConfigSerde"`. Harmless but misleading; drop them or comment why they're kept.
- **Inconsistent env strictness**: invalid `DYN_ROUTER_TRACKING_HASH` now fails startup while an invalid `DYN_ROUTER_QUEUE_THRESHOLD` is still silently ignored (`.parse().ok()`). Strict is right for key config; consider a follow-up making the other vars at least log on parse failure.
- **Benchmark `unsafe`**: the `u32`→`u8` slice cast in `benches/tracking_hash.rs` copies the `hash_block_no_mm` idiom, but in a benchmark-only BLAKE3 reference arm the safe `to_le_bytes` loop would cost nothing meaningful.

### Behavior changes / risks

- **Standalone selection + `assume_kv_reuse=false`**: token-bearing requests switch from deterministic to random tracking identities. Intentional and documented, and it aligns standalone with embedded semantics — but deployments relying on deterministic tracking-hash dedup across duplicate in-flight prompts in that mode lose prompt-membership coalescing. Worth an explicit release-note callout, not just docs.
- **Mixed-epoch blindness**: replica sync transports opaque hashes with no epoch negotiation; a partially rotated fleet silently miscounts active load. Docs state this clearly and prescribe hard-cut rotation, and the builder logs algorithm + key ID at replica-sync init — the right observability breadcrumb. Acceptable for an experimental flag; also the strongest argument for the deferred DEP.
- **CLI/env interaction edge**: in `SelectServiceCli`, `--router-tracking-hash public-xxh3-v1` can't override an env-set key file (validation rejects key options in public mode). Fail-fast is defensible; just be aware you can't "downgrade" via CLI alone.

### Performance

- Default path is bit-identical and pays only an enum compare per request (`assert_eq!` in `compute_seq_hashes_for_tracking_with_context` — could be `debug_assert_eq!`, though the cost is noise).
- Keyed overhead (+252 ns @ 32 blocks) is honestly benchmarked and negligible against request latency; the second hashing pass is unavoidable since public block hashes can't seed keyed identities.
- Nice incidental win: `overlap_scores` now uses `block_hashes_for_indexer` and skips computing sequence hashes it previously discarded.

### Security

- Key handling is careful: fixed 32-byte read into `Zeroizing`, redacted `Debug`, error messages verified by test not to leak the path, key ID (not key) in logs.
- Seeded XXH3 is not a MAC — a party on the internal plane who can observe tracking hashes and choose tokens could plausibly recover seeds. Docs say exactly this and scope the mechanism to a trusted plane. Optional hardening: warn at startup if the key file is group/world-readable (Unix mode check).
- Precomputed `sequence_hashes` remain trusted and unlabeled in keyed mode — documented as the producer's responsibility. This is the main architectural question for the deferred DEP.

### Test coverage

Comprehensive: env parsing, mode-consistency validation, pinned keyed vectors (scope framing, MM, Eagle), scope-sensitivity matrix, key-loading failure modes, redaction/zeroize, public bit-compatibility, legacy-vs-context equivalence, Eagle block counts, Python CLI/env flow-through. Minor gaps: no test for the `SelectServiceCli` override-merge path, and no keyed end-to-end embedded-router test (understandably heavy).

### Verdict

No blocking defects. Main follow-ups: dedupe the validation logic, drop the inert serde attrs, and add a release-note callout for the `assume_kv_reuse=false` selection change. Since the PR defers precomputed-trust and replica-negotiation questions to an outstanding DEP, keeping it in draft until that lands is the right call.

PeaBrane (commented):
Architectural follow-up for the DEP/future work: this tracking-hash domain is closely related to the recently introduced `IndexerDomainId`, but it should not be scoped by `PoolId` or `DcId`.

`IndexerDomainId` already expresses cache compatibility and routing isolation, while `TrackingHashScope` independently reconstructs model, routing-group, and block-size identity and adds request-local `cache_salt`/LoRA/Eagle inputs. Longer-term, please evaluate making `IndexerDomainId` the parent static identity for a tracking domain, rather than maintaining two independent notions of compatible scope. `PoolId` should remain separate because it represents DC-local placement; including it would make equivalent requests hash differently across cooperating pools.

Could this relationship be recorded explicitly in the pending DEP and left as a TODO near `TrackingHashScope` or seed derivation so it is not lost? This need not block the PR.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -167,6 +191,15 @@ fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvR
     if let Some(value) = parse_bool(&get_env, "DYN_ROUTER_TRACK_PREFILL_TOKENS") {
         config.router_track_prefill_tokens = value;
     }
+    if let Some(value) = get_env("DYN_ROUTER_TRACKING_HASH") {
```
PeaBrane, line 208: AI-assisted review: `DYN_ROUTER_ASSUME_KV_REUSE` is documented as a standard router environment variable, but this parser never reads it, and the standalone CLI has no equivalent flag. Both standalone entry points build their config through this helper, so deployment-wide standalone selection remains at `assume_kv_reuse=true`; only per-request overrides or direct Rust construction can reach the newly supported random-tracking path. Please parse the env var and add a regression test, and consider a CLI flag for parity.

lib/kv-router/src/tracking_hash.rs
```diff
@@ -0,0 +1,612 @@
...
+                    compute_seq_hash_for_block(&compute_block_hash_for_seq(
+                        tokens,
+                        scope.block_size,
+                        options,
+                    ))
+                }
+            }
+            TrackingHashAlgorithm::KeyedXxh3V1 => {
+                let (block_seed, chain_seed) = self.derive_seeds(scope, options);
+                let block_hashes = compute_block_hash_for_seq_with_seed(
```
PeaBrane: AI-assisted review (non-blocking performance): keyed mode materializes a full `Vec<LocalBlockHash>` and immediately rescans it into the sequence vector. At one million tokens with block size 16, this is roughly 0.5 MB of avoidable transient allocation per concurrently hashing request, while the benchmark stops at 128 blocks. Consider fusing block hashing with chain construction, or at least extending the benchmark to long-context sizes.

lib/kv-router/src/services/selection/input.rs
```diff
@@ -73,22 +79,18 @@ impl PromptRequest {
...
-            );
-            return Ok(NormalizedReservation {
-                sequence_hashes: normalized.sequence_hashes,
-                isl_tokens: normalized.isl_tokens,
-            });
+                tracking,
+            ));
         }
 
         let sequence_hashes = self.sequence_hashes.as_ref().ok_or_else(|| {
```
PeaBrane, line 96: AI-assisted review (test follow-up): this no-token path intentionally accepts caller-supplied sequence hashes unchanged in keyed mode. That is a documented external trust/compatibility boundary, but the new keyed tests cover only token-derived requests. Please add keyed selection and explicit-reservation cases that pin this pass-through behavior so a future refactor does not rekey or reject trusted inputs.

PeaBrane (approved):
✅ Approved from a code-review perspective. The following inline notes are AI-assisted, non-blocking follow-ups. The outstanding DEP remains the process/architecture gate already called out in the PR.

lib/kv-router/src/scheduling/config.rs
jthomson04: Addressed in [8ed55adffc](https://github.com/ai-dynamo/dynamo/commit/8ed55adffc51b1a122b1e1806a99682cbf290725). The shared environment helper now parses `DYN_ROUTER_ASSUME_KV_REUSE` with the existing boolean parser and logs the resolved value. The standalone selection CLI also has mutually exclusive `--router-assume-kv-reuse` and `--no-router-assume-kv-reuse` flags: no flag preserves the environment-derived value, while an explicit flag takes precedence. The environment and CLI cases are covered by focused regression tests.

lib/kv-router/src/services/selection/input.rs
jthomson04: Addressed in [8ed55adffc](https://github.com/ai-dynamo/dynamo/commit/8ed55adffc51b1a122b1e1806a99682cbf290725). I added one combined keyed hash-only regression test covering both selection and explicit reservation. It pins caller-supplied block hashes on the selection path and caller-supplied sequence hashes on both paths, including the bit-preserving signed-to-unsigned interpretation of `-1` as `u64::MAX`. The trusted-input compatibility boundary remains unchanged.

lib/kv-router/src/tracking_hash.rs
jthomson04: Addressed in [8ed55adffc](https://github.com/ai-dynamo/dynamo/commit/8ed55adffc51b1a122b1e1806a99682cbf290725). Keyed mode now folds each canonical block hash directly into the rolling sequence hash, eliminating the transient `Vec<LocalBlockHash>` while sharing the multimodal and Eagle hashing path. The Criterion matrix now includes 62,500 blocks (one million tokens at block size 16). Relative to the two-pass baseline, the fused keyed arm improved by 26.4% at 128 blocks (958.36 ns to 702.68 ns) and 34.3% at 62,500 blocks (396.58 us to 261.01 us).

saturley-hall (approved):
`zeroize` is MIT or Apache-2.0 and meeting our permissive dependency requirement

PR #11552, opened 2026-07-10
feat(sglang): add cache-salt-aware routing

jthomson04:
## Summary

- Forward the first non-empty cache salt from `routing.cache_salt`, top-level `nvext.cache_salt`, or `extra_args.nvext.cache_salt` to SGLang's base-model generation paths.
- Preserve unsalted compatibility with older/custom SGLang builds and return a clear error when a salted request reaches an `Engine.async_generate` implementation without `cache_salt` support.
- Decode SGLang's typed metadata map in positional `BlockStored` slot 7 while preserving vLLM's string `lora_name` and `extra_keys` fallback behavior.
- Add SGLang to the existing two-worker cache-salt isolation E2E harness.

This draft depends on [sgl-project/sglang#30827](https://github.com/sgl-project/sglang/pull/30827). Before merge, the SGLang Python/runtime-image pins and lockfile must be updated to the first released build containing that change, and the GPU E2E test must be rerun against that image.

## Validation

- `cargo test -p dynamo-kv-router zmq_wire::tests` (26 passed)
- `cargo fmt --all --check`
- Black 23.1.0, isort 5.12.0, and Ruff 0.5.2 checks on all changed Python files
- Python bytecode compilation for all changed SGLang component and router E2E modules
- GPU E2E not run locally; pending the released SGLang runtime image described above

- `components/src/dynamo/sglang/request_utils.py` for request extraction and capability handling
- `lib/kv-router/src/zmq_wire/deserialize.rs` for positional event decoding and namespace precedence
- `tests/router/test_router_e2e_with_sglang.py` for the isolation harness integration

## Related Issues

### DIFF
--- a/components/src/dynamo/sglang/request_utils.py
+++ b/components/src/dynamo/sglang/request_utils.py
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import Mapping
+from typing import Any, Optional
+
+from dynamo.sglang._compat import filter_supported_async_generate_kwargs
+
+
+def request_cache_salt(request: Mapping[str, Any]) -> Optional[str]:
+    """Return the first non-empty cache salt in routing-authoritative order."""
+    extra_args = request.get("extra_args")
+    sources = (
+        request.get("routing"),
+        request.get("nvext"),
+        extra_args.get("nvext") if isinstance(extra_args, Mapping) else None,
+    )
+    for source in sources:
+        if not isinstance(source, Mapping):
+            continue
+        cache_salt = source.get("cache_salt")
+        if cache_salt is None or cache_salt == "":
+            continue
+        if not isinstance(cache_salt, str):
+            raise ValueError("cache_salt must be a string")
+        return cache_salt
+    return None
+
+
+def cache_salt_kwargs(engine: Any, request: Mapping[str, Any]) -> dict[str, str]:
+    """Build the required SGLang Engine kwarg for a cache-salted request."""
+    cache_salt = request_cache_salt(request)
+    if cache_salt is None:
+        return {}
+
+    kwargs = filter_supported_async_generate_kwargs(engine, {"cache_salt": cache_salt})
+    if "cache_salt" not in kwargs:
+        raise ValueError(
+            "cache_salt requires an SGLang version whose "
+            "Engine.async_generate accepts cache_salt"
+        )
+    return kwargs
--- a/components/src/dynamo/sglang/request_utils.py
+++ b/components/src/dynamo/sglang/request_utils.py
@@ -36,7 +36,7 @@ def cache_salt_kwargs(engine: Any, request: Mapping[str, Any]) -> dict[str, str]
     kwargs = filter_supported_async_generate_kwargs(engine, {"cache_salt": cache_salt})
     if "cache_salt" not in kwargs:
         raise ValueError(
-            "cache_salt requires an SGLang version whose "
-            "Engine.async_generate accepts cache_salt"
+            "cache_salt requires Engine.async_generate to accept cache_salt "
+            "and expose an inspectable signature"
         )
     return kwargs
--- a/lib/kv-router/src/zmq_wire/deserialize.rs
+++ b/lib/kv-router/src/zmq_wire/deserialize.rs
@@ -10,7 +10,10 @@ use serde::de::{self, IgnoredAny, MapAccess, SeqAccess, Visitor};
 use crate::protocols::BlockExtraInfo;
 
 use super::extra_keys::{extra_keys_to_block_mm_infos, extra_keys_to_cache_namespace};
-use super::filter::{BlockStoredTrailingField, KvCacheEventMetadata, KvCacheEventTrailingField};
+use super::filter::{
+    BlockStoredPositionSeven, BlockStoredTrailingField, KvCacheEventMetadata,
+    KvCacheEventTrailingField,
+};
 use super::types::{BlockHashValue, ExtraKeyItem, KvTokenIds, RawKvEvent};
 
 /// Our producers use msgspec with `tag=True` and `array_like=True`, which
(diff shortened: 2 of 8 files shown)


tests/router/test_router_e2e_with_sglang.py
```diff
@@ -282,6 +283,33 @@ def test_sglang_kv_router_basic(
     )
 
 
+@pytest.mark.e2e
+@pytest.mark.model(MODEL_NAME)
+@pytest.mark.pre_merge
```
jthomson04, line 288: **Merge-ordering risk (medium — acknowledged in the PR body, restating so it isn't lost).** This `pre_merge` test fails deterministically on any runtime image without sgl-project/sglang#30827 — the worker raises `ValueError` for the salted request, so it's guaranteed red, not flaky. The pin/lockfile bump should land in this PR, or this PR should not merge until it exists; otherwise every subsequent PR's pre-merge CI breaks.

An alternative is a capability-based skip (probe `async_generate` for `cache_salt` and `pytest.skip` on old builds), which makes the test image-tolerant at the cost of silently not running — given the isolation semantics, bumping the pin in the same PR seems preferable.

components/src/dynamo/sglang/request_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+def cache_salt_kwargs(engine: Any, request: Mapping[str, Any]) -> dict[str, str]:
+    """Build the required SGLang Engine kwarg for a cache-salted request."""
+    cache_salt = request_cache_salt(request)
+    if cache_salt is None:
+        return {}
+
+    kwargs = filter_supported_async_generate_kwargs(engine, {"cache_salt": cache_salt})
+    if "cache_salt" not in kwargs:
+        raise ValueError(
```
jthomson04, line 38: **Misleading error when signature inspection fails (low).** `_get_async_generate_supported_kwarg_names` returns `frozenset()` when `inspect.signature` raises (`_compat.py:63-68`), so on such a build this reports "requires an SGLang version whose Engine.async_generate accepts cache_salt" even if the installed version does support it. Failing closed is correct, but the message sends an operator hunting the wrong cause. Consider logging the inspection failure at warning (not debug) when it blocks a *required* kwarg, or folding "or its signature could not be inspected" into the message.

components/src/dynamo/sglang/request_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        extra_args.get("nvext") if isinstance(extra_args, Mapping) else None,
+    )
+    for source in sources:
+        if not isinstance(source, Mapping):
+            continue
+        cache_salt = source.get("cache_salt")
+        if cache_salt is None or cache_salt == "":
+            continue
+        if not isinstance(cache_salt, str):
+            raise ValueError("cache_salt must be a string")
```
jthomson04, line 25: **Nits:** this `cache_salt must be a string` branch has no test — cheap to add.

Also a note on backend drift: vLLM's `_apply_nvext_cache_salt` checks only the two nvext sources and uses truthiness (a non-string salt passes through into an f-string); this helper adds `routing` first and raises on non-strings. The behaviors agree in practice because the frontend sets both fields from one value, but vLLM's own docstring warns about exactly this kind of helper drift. Consider hoisting `request_cache_salt` into `dynamo.common` in a follow-up and pointing both backends at it.

lib/kv-router/src/zmq_wire/deserialize.rs
```diff
@@ -182,7 +185,15 @@ impl<'de> Visitor<'de> for RawKvEventVisitor {
                 // Position 5 was lora_id in older formats; consume and discard for compat.
                 let _lora_id: Option<u64> = seq.next_element()?.unwrap_or(None);
                 let medium: Option<String> = seq.next_element()?.unwrap_or(None);
-                let lora_name: Option<String> = seq.next_element()?.unwrap_or(None);
+                let position_seven: Option<BlockStoredPositionSeven> =
```
jthomson04, line 188: **Map-form KV events silently drop the salt (low).** Only `visit_seq` handles the slot-7 metadata; `visit_map` is unchanged. The module comment says producers use msgspec `array_like=True`, so this is fine today — but if SGLang ever emits map-form events, `cache_salt` would be dropped silently, and a silent drop here is a silent isolation break. A one-line comment in `visit_map` noting the deliberate omission would prevent someone assuming it's handled.

lib/kv-router/src/zmq_wire/filter.rs
```diff
@@ -7,6 +7,18 @@ use serde::Serialize;
...
 use crate::protocols::BlockExtraInfo;
 
+#[derive(Debug, Deserialize)]
+pub(super) struct BlockStoredMetadata {
+    pub(super) cache_salt: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub(super) enum BlockStoredPositionSeven {
```
jthomson04, line 17: **Note (no action needed):** with `cache_salt` optional and unknown fields ignored, *any* future map at slot 7 deserializes as `Metadata`, nulling `lora_name`. That leniency is probably intended (forward-compat, and SGLang LoRA names could later ride the same map) — just be aware slot 7's string form is now effectively frozen as vLLM-only.

Related test gap: there's no case for a metadata map with a missing/nil `cache_salt` (should fall through to the extra-keys fallback / `None`).

jthomson04 (commented):
## Review summary

The design is correct and carefully aligned with the existing vLLM salt plumbing on both the Python and Rust sides. Key integration points verified against `main`:

- **Wire key is correct.** `RoutingHints.cache_namespace` is `#[serde(rename = "cache_salt")]` (`lib/llm/src/protocols/common/preprocessor.rs`), so reading `routing["cache_salt"]` first matches what the frontend actually sends, and the frontend is authoritative (it already applies nvext-over-legacy precedence and empty-string fallback in `extensions.rs::request_cache_salt`).
- **Disagg-safe.** `BaseWorkerHandler._get_input_param` is the shared path for both decode and prefill, and the prefill `inner_request` carries `routing` (it reads `routing.priority`/`dp_rank` a few lines later) — salted requests hash consistently on both sides of a disaggregated deployment.
- **Fail-closed is the right posture.** Raising when the engine lacks `cache_salt` support is correct for an isolation feature — silently dropping the salt would mean cross-tenant KV-cache sharing. `ValueError` from the generate path is the established pattern (the top-logprobs gate does the same).
- **Rust precedence is right and tested.** Typed metadata beats the tagged `dynamo-cache-salt:` extra-keys fallback; the vLLM string-at-slot-7 path has a regression test; the untagged-enum-in-sequence pattern is already proven in this visitor (`BlockStoredTrailingField`).
- **Performance is fine.** The signature probe is `lru_cache`d; per-request cost is a couple of dict ops, only when a salt is present.

Test coverage is good overall: unit tests cover precedence, empty-string fallback, the unsupported-engine error, and both worker/unified integration paths; Rust tests cover the three slot-7 shapes; the E2E test reuses the proven isolation harness with markers matching its siblings.

Specific findings are inline. **Verdict:** looks good to merge once the SGLang pin bump is attached (see the inline comment on the `pre_merge` marker); the rest are optional follow-ups.

---
*Review posted by Claude Code on behalf of @jthomson04.*

jthomson04:
Refreshing PR as https://github.com/ai-dynamo/dynamo/pull/14803

PR #11560, opened 2026-07-10
fix(tests): prevent port-forward EADDRINUSE flake in TRTLLM checkpoint test

nv-rinig:
## Summary

- On port-forward retry, create a fresh `pod.portforward(local_port=0)` object instead of calling `stop()+start()` on the existing one. kr8s retains the previously-assigned port in the object, so restarting would try to bind the same port still in TIME_WAIT — causing `OSError: [Errno 98] address already in use` in the background thread.
- Add `pytest.PytestUnhandledThreadExceptionWarning` to `filterwarnings` in `pyproject.toml` as a safety net. The existing filter only covered `PytestUnraisableExceptionWarning` (`sys.unraisablehook`); background thread exceptions go through `threading.excepthook` and a separate warning type, which is what caused the `ExceptionGroup: multiple thread exception warnings` failure in CI.

## Root cause

`tests/deploy/test_dynamocheckpoint.py::test_dgd_checkpoint_restore_deploy` failed in post-merge run [29126250839](https://github.com/ai-dynamo/dynamo/actions/runs/29126250839) despite inference returning HTTP 200 both before and after checkpoint restore. The actual fix is the retry-path change in `managed_deployment.py`; the `pyproject.toml` change is a belt-and-suspenders guard for teardown races.

## Test plan

- [ ] Re-run `TRTLLM DynamoCheckpoint Deploy Test` on the post-merge pipeline
- [ ] Confirm no regression in other deploy tests (vLLM/SGLang DynamoCheckpoint, agg/disagg/router)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved port-forward retry handling to recover more reliably after failed attempts.
  * Prevented port-forward shutdown errors from interrupting subsequent retries.
* **Tests**
  * Reduced noise from expected background-thread warnings during port-forward teardown.

### DIFF
--- a/tests/deploy/test_dynamocheckpoint.py
+++ b/tests/deploy/test_dynamocheckpoint.py
@@ -25,6 +25,15 @@ from tests.utils.managed_deployment import (
 
 logger = logging.getLogger(__name__)
 
+# kr8s port-forward teardown runs in background threads; on pod termination it
+# can surface expected OSErrors (e.g. EADDRINUSE for a local port still in
+# TIME_WAIT) via threading.excepthook. Under filterwarnings=error those would
+# fail this live-cluster test, so scope the suppression to this module only
+# rather than globally hiding unrelated background-thread crashes.
+pytestmark = pytest.mark.filterwarnings(
+    "ignore::pytest.PytestUnhandledThreadExceptionWarning"
+)
+
 TRANSIENT_K8S_EXCEPTIONS = (
     aiohttp.ClientError,
     asyncio.TimeoutError,
--- a/tests/utils/managed_deployment.py
+++ b/tests/utils/managed_deployment.py
@@ -1557,6 +1557,26 @@ class ManagedDeployment:
                     continue
                 try:
                     port_forward.stop()
+                except Exception as e:
+                    self._logger.debug(
+                        f"Error stopping port forward for pod {pod.name}: {e}"
+                    )
+                # kr8s' sync stop() can return before the background thread has
+                # fully torn down (or even finished starting), so we can't assume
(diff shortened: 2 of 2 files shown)


galletas1712 (approved):

pyproject.toml
```diff
@@ -210,6 +210,8 @@ filterwarnings = [
     "ignore:.*unclosed event loop.*:ResourceWarning",
     # unraisable exception warnings
     "ignore:.*Exception ignored in.*:pytest.PytestUnraisableExceptionWarning",
+    # background thread exceptions from kr8s port-forward teardown (OSError EADDRINUSE etc.)
+    "ignore::pytest.PytestUnhandledThreadExceptionWarning",
     # pynvml deprecation, temporary until upstream migrates to nvidia-ml-py
```
dmitry-tokarev-nv, line 218: This filter matches ALL background-thread exceptions across every test in the suite — it will silently swallow unrelated thread crashes. Scope it to the specific error pattern instead:

```suggestion
    # background thread OSError from kr8s port-forward teardown (EADDRINUSE in TIME_WAIT)
    "ignore:.*address already in use.*:pytest.PytestUnhandledThreadExceptionWarning",
```

dmitry-tokarev-nv (changes requested):
small suggestion

pyproject.toml
nv-rinig: thanks dmitry, instead of making it global i moved it to the module where the test is actually run. please take a look!

dmitry-tokarev-nv (approved):

galletas1712 (approved):

KrishnanPrash (approved):

PR #11566, opened 2026-07-11
fix(gms): preserve runtime tensor aliases

galletas1712:
## Summary

This revision intentionally replaces the existing GMS Torch metadata format. There are no production users that require schema compatibility, so the reader no longer carries a legacy decoder, schema versions, synthetic storage/object group IDs, or generic helpers such as `strict_int` and `strict_int_list`.

The new contract is storage-first:

- write one typed MessagePack `StorageManifest` per PyTorch `StorageImpl`
- describe each distinct tensor object's dtype, shape, stride, byte offset, `requires_grad`, and module bindings inside that storage manifest
- represent exact aliases as multiple bindings on one tensor object
- represent distinct views or tensors sharing storage as sibling tensor objects in one manifest
- use `ModuleTensorKind`, a string-backed enum, instead of literal strings or string matching
- keep the GMS allocation ID and storage base offset only in the outer metadata envelope
- treat the writer's tensor/storage graph as authoritative when constructing a fresh reader; reader placeholder identity and placeholder storage topology are not part of the contract

This is a deliberately breaking metadata change. Old GMS Torch metadata must be regenerated by loading the model through the write path again.

The implementation was also simplified in response to review:

- removed the one-field displaced-owner wrapper and retained owners in collision-safe sentinel-backed model state
- removed the one-use materialized-object record in favor of a local typed tuple
- renamed `Slot` to `ModuleTensorBinding`, `_DiscoveredObject` to `_DiscoveredTensor`, and the reader target to `_ResolvedTensorDestination`
- renamed the private pipeline stages and added concise stage docstrings plus a top-level write/read/rebind flow diagram
- separated policy-free discovery from GMS allocation selection and validation, so unrelated unsupported non-parameter tensors are ignored while every selected storage is validated before side effects
- kept `_DiscoveredTensor` and `_ResolvedTensorDestination` as named records because they represent real identity and installation boundaries rather than generic containers
- kept the storage lifecycle together in `module.py`; splitting it would add private coupling among discovery, publication, materialization, rebinding, and TensorRT-LLM relocation, while `tensor.py` already separates the schema and raw-storage operations
- use PyTorch `UntypedStorage.clone()` / `copy_()` directly for complete storage envelopes instead of temporary byte-tensor copy wrappers
- make storage-to-GMS matching return allocation-envelope data rather than mutating discovery records with optional state
- remove custom slot enumeration and one-use cloning/template/swap helpers where stdlib or direct local control flow is clearer
- consolidate overlapping tests around behavioral boundaries: 46 test functions became 41 while retaining malformed input, topology, rollback, atomicity, publication, backend, and real CUDA cases

Even after adding focused review regressions, the cleanup series removes a net 222 lines.

## Before and after

The examples below show the decoded structures. The new manifest is MessagePack on the wire.

### Before: one metadata value per tensor slot

```json
{
  "shape": [4, 8],
  "dtype": "torch.float16",
  "stride": [8, 1],
  "tensor_type": "tensor_attr",
  "storage": {
    "schema_version": 1,
    "storage_group_id": 3,
    "object_group_id": 7,
    "storage_base_offset": 4096,
    "storage_nbytes": 1024,
    "storage_offset": 16,
    "buffer_persistent": false
  }
}
```

Every slot repeated tensor and storage fields. The reader reconstructed aliases and shared storage by joining integer group IDs while preserving the legacy per-tensor `offset_bytes` meaning.

### After: one manifest per storage

Outer GMS metadata envelope:

```json
{
  "key": "torch.module.storage/0",
  "allocation_id": "<gms-allocation-id>",
  "offset_bytes": 4096,
  "value": "<MessagePack StorageManifest>"
}
```

Conceptually decoded manifest:

```json
{
  "nbytes": 1024,
  "objects": [
    {
      "dtype": "float16",
      "shape": [4, 8],
      "stride": [8, 1],
      "storage_offset_bytes": 32,
      "requires_grad": false,
      "bindings": [
        {"path": "runtime_cache", "kind": "attribute"},
        {"path": "runtime_cache_alias", "kind": "attribute"}
      ]
    },
    {
      "dtype": "float16",
      "shape": [2, 8],
      "stride": [16, 1],
      "storage_offset_bytes": 32,
      "requires_grad": false,
      "bindings": [
        {"path": "runtime_view", "kind": "nonpersistent_buffer"}
      ]
    }
  ]
}
```

There are no embedded allocation IDs and no join IDs. The structure directly expresses the source storage graph. `kind` remains a lowercase string on the wire, but typed decoding produces a `ModuleTensorKind` enum and rejects unknown values.

## Lifecycle and ownership

The writer discovers complete `StorageImpl` groups without applying GMS policy, selects the groups owned by GMS, validates the complete selected set, publishes their manifests, and rebinds mutable non-parameter-only storage before commit. Parameter-containing storage remains one shared read-only mapping so parameter aliases and views retain their original storage topology. Tensor swaps are preflighted before mutation. TensorRT-LLM relocation copies each complete untracked storage once and reconstructs all of its tensor objects with their original layouts.

The fresh reader imports each allocation once, reconstructs each storage once, then installs the source bindings. Missing parameter, buffer, and direct-attribute bindings may be created; conflicting destinations fail closed. vLLM performs required post-load processing once on the meta model and propagates failure rather than returning a partially initialized reader.

Failure paths release writer leases best-effort. Imported CUDA mappings are freed only after synchronization proves that asynchronous copies are no longer using them.

## What this PR handles

| Flow | Role of this PR |
| --- | --- |
| First disk/checkpoint load into GMS | Publishes the storage manifests after backend loading and post-processing. |
| New process cold start or restart from existing GMS state | Creates a fresh model shell and materializes the published tensor/storage graph. |
| Initial shadow-engine attachment that uses the GMS read path | Uses the same fresh-reader materialization path. |
| Shadow-engine sleep/wake or failover after tensors are already materialized | Does not rerun the manifest reader. The existing GMS VA unmap/remap lifecycle preserves the already-created tensor objects. |
| GMS allocation snapshot to disk and restore | The snapshot layer copies allocations and metadata. Because allocation identity lives only in the outer envelope, restore can remap that identity without decoding or rewriting the opaque manifest. This PR does not add snapshot orchestration itself. |
| CRIU/process snapshot restore | Does not rebuild the Python module from these manifests; the process object graph is restored and GMS remaps its preserved virtual addresses. |

Backend state hidden outside normal `nn.Module` tensor bindings is still owned by the backend's meta-model/post-load lifecycle. This PR does not attempt to serialize arbitrary Python or backend runtime state.

## Guards retained

The remaining checks protect concrete correctness boundaries for storage selected by a GMS operation: typed MessagePack fields and enum values, canonical dtypes, storage bounds and alignment, non-overlapping allocation ranges, unique destination bindings, reserved internal-state collisions, supported tensor/Parameter ownership for `torch.utils.swap_tensors`, and safe CUDA cleanup. Compatibility-only and placeholder-topology guards were removed.

## Validation

The branch was rebased onto current `main`; the GMS VMMDevice abstraction and its singleton-reset integration coverage were retained during conflict resolution.

Passed locally:

```bash
files=$(git diff HEAD^ --name-only -- '*.py')
ruff format --check $files
ruff check $files
python3 -m compileall -q $files
git diff HEAD^ --check
```

Independent production, test-coverage, final-code, and validation reviews approved the refactor. The consolidated suite retains the distinct schema, malformed payload, alias/shared-storage, source-authoritative installation, rollback, swap atomicity, publication ordering, backend cleanup, and real CUDA integration behaviors.

In the previous full CI run:

- all 31 selected tests in the GMS GPU shard passed, including the new shared-storage, alias, remap, materialization, and TensorRT-LLM relocation coverage
- the expanded GMS publication tests passed in the sequential runtime shard
- vLLM, SGLang, and TensorRT-LLM runtime test pipelines passed
- the nine reported failures reduced to seven jobs affected by unrelated, non-hermetic AIConfigurator API/bincode dependency skew and two downstream aggregate gates
- three additional stale replay failures were already fixed on `main` and are included by this rebase

A new CI run is required for HEAD `87f501b354`.

Runtime and GPU tests were not run in the local checkout because it has no `.venv` and the available Python environment lacks `pytest`, `torch`, and `msgspec`. Full backend and snapshot E2E validation remains pending.

AI assistance was used to investigate, implement, review, and validate this change. The submitting human must review and understand every changed line before marking it ready for review.

## Summary by CodeRabbit

- **New Features**
  - Added reliable saving and restoration of shared, aliased, and non-persistent tensors.
  - Preserved tensor layouts, offsets, data types, gradients, parameter subclasses, and tied weights during model migration.
  - Added support for reconstructing runtime tensor aliases after deferred writes.

- **Bug Fixes**
  - Improved cleanup and recovery when publishing, importing, or materializing tensors fails.
  - Prevented incomplete mappings and stale pending writes after errors.
  - Improved handling of shared storage during import and remapping.

- **Tests**
  - Expanded coverage for storage manifests, aliasing, validation, failure handling, and publication ordering.

### DIFF
--- a/lib/gpu_memory_service/client/torch/module.py
+++ b/lib/gpu_memory_service/client/torch/module.py
@@ -14,10 +14,17 @@ from __future__ import annotations
 import copyreg
 import logging
 import weakref
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Iterator, Tuple
 
 import torch
-from gpu_memory_service.client.torch.tensor import GMSTensorSpec, TensorMetadata
+from gpu_memory_service.client.torch.tensor import (
+    GMSTensorSpec,
+    TensorMetadata,
+    _storage_from_pointer,
+    _tensor_from_storage,
+    _validate_layout,
+)
 
 if TYPE_CHECKING:
     from gpu_memory_service.client.memory_manager import GMSClientMemoryManager
@@ -136,13 +143,15 @@ def _resolve_module_attr(
     parts = qualified_name.split(".")
     mod = root
     for p in parts[:-1]:
-        if hasattr(mod, p):
-            mod = getattr(mod, p)
-        elif hasattr(mod, "__getitem__"):
-            try:
-                mod = mod[int(p)] if p.isdigit() else mod[p]
-            except Exception:
+        if isinstance(mod, torch.nn.Module) and p in mod._modules:
+            mod = mod._modules[p]
+        elif isinstance(mod, torch.nn.Module) and p in mod.__dict__:
+            value = mod.__dict__[p]
+            if not isinstance(value, (list, tuple)):
                 raise AttributeError(f"Cannot resolve {p!r} in {qualified_name!r}")
+            mod = value
+        elif isinstance(mod, (list, tuple)) and p.isdigit():
+            mod = mod[int(p)]
         else:
             raise AttributeError(f"Cannot resolve {p!r} in {qualified_name!r}")
     return mod, parts[-1]
@@ -257,9 +266,7 @@ def _validate_parameter_swap(
     _make_parameter_replacement(parameter, probe, name=name)
 
 
-def _tensor_source(
-    spec: GMSTensorSpec,
-) -> tuple[str, int, tuple[int, ...], tuple[int, ...], torch.dtype,]:
+def _tensor_source(spec: GMSTensorSpec):
     return (
         spec.allocation_id,
         spec.offset_bytes,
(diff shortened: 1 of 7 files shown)


lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,650 @@
...
-    mod = root
-    for p in parts[:-1]:
-        if hasattr(mod, p):
-            mod = getattr(mod, p)
-        elif hasattr(mod, "__getitem__"):
-            try:
-                mod = mod[int(p)] if p.isdigit() else mod[p]
-            except Exception:
-                raise AttributeError(f"Cannot resolve {p!r} in {qualified_name!r}")
+class _ReboundTensorOwners:
```
galletas1712: Do we need a whole class for this?

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,650 @@
...
+    def __init__(self) -> None:
+        self.tensors: list[torch.Tensor] = []
+
+
+@dataclass
```
galletas1712, line 48: same here - do we need a whole class for this?

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,650 @@
...
+
+@dataclass(frozen=True)
+class _StoredManifest:
+    allocation_id: str
+    storage_base_offset: int
+    manifest: StorageManifest
+
+
+@dataclass(frozen=True)
+class _SlotTarget:
```
galletas1712: I feel like "Slot" isn't a very good name since it's not immediately obvious/descriptive

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,650 @@
...
+                values = list(container)
+                values[self.index] = tensor
+                self.module.__dict__[self.attr] = tuple(values)
 
 
-# =============================================================================
-# Public API - Registration and Materialization
-# =============================================================================
+@dataclass(frozen=True)
+class _MaterializedObject:
```
galletas1712: Do we also need a separate class here?

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,650 @@
...
+    tensor: torch.Tensor
+    targets: tuple[_SlotTarget, ...]
+
+
+def _iter_module_slots(
```
galletas1712: as a general comment, we need docstrings that describe briefly (what exactly each of these functions is doing, since they are quite large). Maybe we should also split out this module.py file into different files, but not too sure about that. Generally the naming of each function isn't very good and I don't have a good sense of what each function does. Also, maybe let's have a brief ASCII diagram/explainer at the top ofthe file showing how the different flows work.

lib/gpu_memory_service/client/torch/tensor.py
```diff
@@ -1,228 +1,129 @@
...
 
+import msgspec
 import torch
 
-if TYPE_CHECKING:
-    from gpu_memory_service.client.memory_manager import GMSClientMemoryManager
+STORAGE_MANIFEST_PREFIX = "torch.module.storage/"
 
+SlotKind = Literal[
+    "parameter",
```
galletas1712: I don't like how these are literals. This should be an enum. Let's not do string matching anywhere

lib/gpu_memory_service/client/torch/module.py
galletas1712: Removed `_ReboundTensorOwners`. Displaced source tensors now live in sentinel-backed model state as a plain list, so there is no one-field wrapper class and pre-existing user attributes still fail collision validation.

lib/gpu_memory_service/client/torch/module.py
galletas1712: Renamed this to `_DiscoveredTensor` and added a concise docstring. I kept the record because it is the identity boundary between one Python tensor plus its exact aliases and the separate grouping of distinct tensor objects by `StorageImpl`; replacing it with anonymous tuples made that distinction harder to follow.

lib/gpu_memory_service/client/torch/module.py
galletas1712: Renamed `Slot` to `ModuleTensorBinding`, renamed the manifest field to `bindings`, and renamed the resolved reader-side record to `_ResolvedTensorDestination`. The names now distinguish serialized module bindings from resolved installation destinations.

lib/gpu_memory_service/client/torch/module.py
galletas1712: Removed `_MaterializedObject`. Materialization now uses a local typed `(tensor, destinations)` tuple. `_ResolvedTensorDestination` remains because it owns destination identity, collision detection, and installation behavior.

lib/gpu_memory_service/client/torch/module.py
galletas1712: Added a concise write/read/writer-rebind ASCII flow at the top, one-line docstrings for the major stages, and renamed the private pipeline functions around discovery, mapping, manifest construction, destination resolution, cloning, and swapping. I kept the lifecycle in `module.py`: discovery/rebinding are shared with TensorRT-LLM and read/write share the manifest and identity rules, so splitting now would add private cross-module coupling; `tensor.py` already separates schema/raw-storage operations.

lib/gpu_memory_service/client/torch/tensor.py
galletas1712: Replaced the literal union with Python-3.10-compatible `ModuleTensorKind(str, Enum)` and changed every construction/comparison to enum members. MessagePack still stores lowercase string values, while typed decoding returns enum members and rejects unknown values; there is no binding-kind string matching left.

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,670 @@
...
+                "GMS inference storage does not support autograd tensor "
+                f"{tensor_object.bindings[0].path!r}"
+            )
+
+        storage = tensor.untyped_storage()
+        if storage.nbytes() == 0:
+            raise RuntimeError(
+                "GMS module manifests do not support zero-byte StorageImpls: "
+                f"{tensor_object.bindings[0].path!r}"
+            )
```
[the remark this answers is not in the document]
galletas1712: Fixed in `87f501b354`. Discovery is now policy-free: it only records tensor identity, bindings, and shared storage. Each operation then matches/selects the storage it owns and validates the complete selected set before metadata publication, cloning, mapping, or tensor swaps. Unmapped empty/autograd non-parameter tensors are ignored again, while an unmapped registered parameter still fails closed. The focused tests cover both the ignored-unmapped case and selected-invalid failure before publication or rebind.

lib/gpu_memory_service/integrations/vllm/model_loader.py
```diff
@@ -361,9 +380,6 @@ def _create_meta_model(vllm_config, model_config) -> torch.nn.Module:
         with meta_device:
             model = initialize_model(vllm_config=vllm_config, model_config=model_config)
 
-    try:
-        process_weights_after_loading(model, model_config, meta_device)
-    except Exception as e:
-        logger.debug("[GMS] Post-processing on meta tensors: %s", e)
+    process_weights_after_loading(model, model_config, meta_device)
```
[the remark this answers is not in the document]
galletas1712: This remains intentionally fail-closed. vLLM post-load processing is required to produce a valid reader model; swallowing an arbitrary hook failure can return a partially initialized model whose backend state does not match its materialized weights. If a supported hook is not meta-compatible, the clean fix is to make that hook/meta-model path compatible rather than treat every exception as success. The previous vLLM backend pipeline passed; the updated head still requires the normal backend-matrix rerun.

lib/gpu_memory_service/integrations/common/utils.py
```diff
@@ -158,9 +157,16 @@ def finalize_gms_write(
...
+    try:
+        stats = prepare_gms_write(allocator, model)
+        rebound_bytes = rebind_nonparameter_tensors(allocator, model)
+        publish_gms_write(allocator)
+    except BaseException:
+        try:
+            allocator.close(best_effort=True)
+        except BaseException:
+            logger.exception("[GMS] Failed to release failed write finalization")
+        raise
```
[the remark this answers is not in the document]
galletas1712: This ordering is intentional for the shared backend path. Rebinding is part of preparing the writer model for a usable committed layout, so it must succeed before publication makes that layout visible. Publishing first would allow a terminal swap failure to leave committed metadata for a writer that could not complete finalization. Swap destinations are preflighted before mutation, and displaced source owners are retained on the model. SGLang uses the same contract rather than a backend-specific branch; its previous runtime pipeline passed and the updated head requires the normal rerun.

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,670 @@
...
-# Public API - Registration and Materialization
-# =============================================================================
+    intervals: dict[str, list[tuple[int, int]]] = {}
+    for discovered_storage, allocation_id, start in located:
+        intervals.setdefault(allocation_id, []).append(
+            (start, start + discovered_storage.storage.nbytes())
+        )
+    for allocation_id, allocation_intervals in intervals.items():
+        ordered = sorted(allocation_intervals)
+        for previous, current in zip(ordered, ordered[1:], strict=False):
```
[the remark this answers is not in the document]
galletas1712: Fixed in `87f501b354`: both adjacent interval checks now use `itertools.pairwise`.

lib/gpu_memory_service/client/torch/module.py
```diff
@@ -1,162 +1,670 @@
...
+
+def _free_imported_mappings(
+    manager: "GMSClientMemoryManager",
+    imported: list[int],
+) -> None:
+    for va in reversed(imported):
+        try:
+            manager.free_va(va)
+        except BaseException:
+            logger.exception("[GMS] Failed to release imported mapping at %#x", va)
```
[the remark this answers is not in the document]
galletas1712: I am keeping `BaseException` here because this is not normal teardown. `_free_imported_mappings` is called only from an already-active `except BaseException` rollback path, before installation begins, followed by a bare `raise`. Cleanup is best-effort: a cleanup-time `KeyboardInterrupt`, `SystemExit`, or cancellation must not replace the original materialization failure or stop rollback of the remaining mappings. Catching `Exception` here would lose that property. An independent re-review checked this caller context and approved the existing catch.

lib/gpu_memory_service/tests/test_gms_write_publication.py
```diff
@@ -76,66 +162,954 @@ def prune(_allocator, *, referenced_allocation_ids):
...
+    ]
+    replacements = {
+        id(first): torch.full_like(first, 20),
+        id(second): torch.full_like(second, 30),
+    }
+    first_storage = first.untyped_storage()._cdata
+    second_storage = second.untyped_storage()._cdata
+    second_ref = weakref.ref(second)
+
+    with pytest.raises(RuntimeError, match="second.*weakref"):
```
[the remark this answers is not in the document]
galletas1712: Fixed in `87f501b354`: the pattern is now the explicit raw regex `r"second.*weakref"`. I also fixed the same issue in the newly added parameter regression pattern, and the focused RUF043 check passes.

lib/gpu_memory_service/tests/test_torch_integration.py
```diff
@@ -72,6 +80,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
 
 
+class _AliasedRuntimeTensor(torch.nn.Module):
+    def __init__(self, tensor: torch.Tensor) -> None:
+        super().__init__()
+        self.register_buffer("runtime", tensor, persistent=False)
+        self.direct = tensor
+
+
+def test_trtllm_moves_complete_shared_parameter_storage_once(running_gms):
```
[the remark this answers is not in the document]
galletas1712: Fixed in `87f501b354`: both `running_gms` integration tests now have `@pytest.mark.timeout(120)`, matching the existing convention in this file.

lib/gpu_memory_service/tests/test_torch_integration.py
```diff
@@ -355,6 +468,89 @@ def test_live_gms_tensor_survives_unmap_and_remap(running_gms):
...
+    writer_model.data_view = writer_model.data_weight.data
+    writer_model.register_parameter(
+        "view_weight", torch.nn.Parameter(view_tensor, requires_grad=False)
+    )
+    writer_model.strided = writer_model.view_weight[2:10:2]
+    writer_model.transposed = writer_model.view_weight.reshape(3, 4).T
+    register_module_tensors(writer, writer_model)
+    assert writer.commit()
+    del writer_model, data_tensor, view_tensor, data_storage, view_storage
+    writer.close()
```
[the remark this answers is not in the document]
galletas1712: Fixed in `87f501b354`: the writer now uses `GMSClientMemoryManager` as a context manager, with a `finally` that drops live model/tensor/storage references before manager cleanup. This releases the RW lease and mappings even when setup, publication, or commit fails.

galletas1712:
Also addressed the two review-summary items that did not have inline threads in `87f501b354`: the public manifest types now have concise schema-focused docstrings, and parameter reconstruction no longer wraps arbitrary construction/programming failures in a generic `RuntimeError`, preserving the original exception and traceback.

PR #11568, opened 2026-07-11
perf(kv-router): streamline block release cleanup

jthomson04:
## Overview

Reduce redundant work in the default `BlockTracker::release` path. Active-prompt cleanup walks millions of blocks in capacity-stress workloads, so repeated arena accesses and unnecessary fractional-map probes directly affect free-service latency and drain throughput.

## Summary

- inspect each block node once with mutable access before deciding whether to decrement shared ownership or remove an exclusive node
- remove exclusive arena nodes once and destructure their fields for the remaining cleanup
- retain the single-lookup hash-index removal while making its redundant hash-to-node identity check debug-only
- skip fractional-map removal when fractional accounting is inactive

There are no public API changes or intended behavior changes.

## Details

The shared path now performs one mutable arena lookup instead of a read followed by a mutable lookup. The exclusive path removes the validated arena node directly instead of reading it again before removal. The old `Entry`-based hash-index removal was already a single hash lookup; changing it to `remove` is a simplification, while the hot-path difference is that the redundant identity assertion is debug-only. Release builds already trust `SequenceHash` uniqueness during insertion, where collision checks are also debug-only. A debug-mode regression test now pins the identity assertion and its diagnostic message.

The change preserves release ordering, shared-prefix ownership, output-chain compatibility, fractional accounting when enabled, and iterative deep-chain cleanup.

`BlockTracker::release` and `release_rejects_hash_index_alias` in `lib/kv-router/src/sequences/block_tracker.rs`.

## Related Issues

## Validation

- `cargo test -p dynamo-kv-router --lib` (742 passed, 2 ignored)
- one baseline arm followed by one candidate arm of `active_sequences_bench`, using CPUs 0-1, jemalloc, block size 2, trace-length factor 512, eight workers, eight operation lanes, 1,048,576 GPU blocks, a 1 ms offered schedule, and seed 42

The baseline binary was built from commit `28649cabdb1b75889928fae73cf15a336f9c453f`. The candidate binary was built from that exact source with this release-path patch applied; the measured patch was subsequently committed as `4139d7641fbdcba7204df5b9c886733146e286e5`. Follow-up commit `82ca3a9064aac358c9c496e14e67f2b0f5264d9c` applies the reviewer's semantically equivalent direct-borrow form and adds the debug-only test, so the benchmark was not rerun.

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Logical block visits/s | 221.2M | 288.4M | +30.4% |
| Logical operations/s | 21,942 | 28,616 | +30.4% |
| Free service p50 | 47.1 us | 43.3 us | -8.0% |
| Free service p99 | 1,327.4 us | 1,090.3 us | -17.9% |
| Free service p999 | 4,332.8 us | 3,272.5 us | -24.5% |
| Drain time | 135.7 ms | 103.8 ms | -23.5% |

Both arms completed exactly 1,000 adds, prefill completions, and frees; 10,079,232 release block visits; and 30,237,696 total logical block visits. Exact IDs, per-worker FIFO ordering, generator validity, and empty final state all passed with no failure reasons. Both arms reported the expected `kept_up=false` result under the 1 ms capacity-stress schedule. This single-run comparison is directional.

### DIFF
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -67,6 +67,10 @@ impl BlockTracker {
     ///
     /// # Collision scope
     ///
+    /// Collision handling is outside the active-sequence tracker's scope: it
+    /// assumes distinct live lineages have distinct [`SequenceHash`] values and
+    /// does not detect, prevent, or recover from true hash collisions.
+    ///
     /// `SequenceHash` is a probabilistic 64-bit identifier, so collisions are
     /// possible but very unlikely. A collision matters here only when both
     /// blocks are live in the same `WorkerWithDpRank` tracker: each worker and
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -174,22 +174,14 @@ impl BlockTracker {
         let mut current = tail;
 
         while let Some(node_id) = current {
-            let shared = {
-                let node = self
-                    .nodes
-                    .get_mut(node_id)
-                    .expect("request chain references a missing block node");
-                let incoming = node.incoming.get();
-                if incoming > 1 {
-                    node.incoming = NonZeroU32::new(incoming - 1)
-                        .expect("shared block ownership count cannot become zero");
-                    true
-                } else {
-                    false
-                }
-            };
-
-            if shared {
+            let node = self
+                .nodes
+                .get_mut(node_id)
+                .expect("request chain references a missing block node");
+            let incoming = node.incoming.get();
+            if incoming > 1 {
+                node.incoming = NonZeroU32::new(incoming - 1)
+                    .expect("shared block ownership count cannot become zero");
                 retained = Some(node_id);
                 break;
             }
@@ -740,6 +732,19 @@ mod tests {
         tracker.unique_blocks.insert(2, node_id);
         let _ = tracker.contains_block(&2);
         drop(chain);
+    }
+
+    #[cfg(debug_assertions)]
+    #[test]
+    #[should_panic(expected = "block hash index references a different live node")]
+    fn release_rejects_hash_index_alias() {
+        let mut tracker = BlockTracker::default();
+        let (first, _) = tracker.acquire_prompt(&[1]);
+        let (_second, _) = tracker.acquire_prompt(&[2]);
+        let aliased_node_id = tracker.node_id_for(2);
+
+        tracker.unique_blocks.insert(1, aliased_node_id);
+        let _ = tracker.release(first);
     }
 
     #[test]
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -6,7 +6,6 @@ use rustc_hash::FxHashMap;
 #[cfg(debug_assertions)]
 use rustc_hash::FxHashSet;
 use slotmap::{SlotMap, new_key_type};
-use std::collections::hash_map::Entry;
 use std::num::NonZeroU32;
 
 new_key_type! {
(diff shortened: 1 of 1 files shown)


lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -175,45 +174,45 @@ impl BlockTracker {
...
+                    true
+                } else {
+                    false
+                }
+            };
+
+            if shared {
                 retained = Some(node_id);
                 break;
             }
```
jthomson04: Nit: the `shared` bool indirection can go — under NLL the mutable borrow ends at its last use, so this is equivalent and more direct:

```suggestion
            let node = self
                .nodes
                .get_mut(node_id)
                .expect("request chain references a missing block node");
            let incoming = node.incoming.get();
            if incoming > 1 {
                node.incoming = NonZeroU32::new(incoming - 1)
                    .expect("shared block ownership count cannot become zero");
                retained = Some(node_id);
                break;
            }
```

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -175,45 +174,45 @@ impl BlockTracker {
...
-                }
-                Entry::Vacant(_) => panic!("live block node is missing from the hash index"),
+            let indexed_node_id = self
+                .unique_blocks
+                .remove(&hash)
+                .expect("live block node is missing from the hash index");
+            debug_assert_eq!(
+                indexed_node_id, node_id,
+                "block hash index references a different live node"
+            );
```
jthomson04, line 208: Making the trade-off here explicit (fine to keep, just noting it): in a release build, if the index ever aliases — `unique_blocks[hash]` pointing at a different live node, the ~n²/2⁶⁵ collision case documented on `acquire_prompt` — the old `Entry` path panicked immediately at the first release. Now the wrong entry is removed silently, and the panic surfaces later, when the *other* node is released, as `"live block node is missing from the hash index"` — which points away from the real cause. In the interim `contains_block`/`acquire_prompt` would misreport that block as absent.

Since insertion-side collision detection (`debug_assert_new_hashes`, the `previous.is_none()` asserts) was already debug-only, this backstop was never load-bearing in release, so the downgrade seems reasonable.

Optional: none of the existing `should_panic` tests cover this message. A debug-mode test in the style of `stale_generation_in_index_is_rejected` — acquire two chains, alias one hash to the other's node id via `tracker.unique_blocks.insert`, then release — would pin the `debug_assert_eq!` behavior.

jthomson04 (commented):
Reviewed the release-path refactor — traced both the shared and exclusively-owned paths against the old code and the surrounding invariants (`assert_consistent`, the randomized reference-model test). Observable behavior of `release` is unchanged in all legitimate states, and the perf claims are plausible for a loop this hot: one fewer slotmap lookup per node on both paths, and skipping the `fractional_blocks` hash-and-probe when fractional accounting is inactive (likely the biggest win in capacity-stress workloads).

Two notes beyond the inline comments:

- Description nit: the old `Entry`-based hash-index removal was already single-lookup, so "remove each live hash-index entry with one lookup" isn't a change in itself — the actual win there is dropping the release-mode identity assert.
- "both built from `28649cab…`" in the validation section reads ambiguously — presumably baseline = that commit and candidate = that commit + this patch? Worth clarifying. A second run per arm would also firm up the noisier p999/drain numbers, though the direction is clear.

Per `lib/kv-router/src/sequences/CLAUDE.md`, make sure PeaBrane is on the review since this is inside the write model's hottest loop (the benchmark-numbers requirement is already satisfied).

lib/kv-router/src/sequences/block_tracker.rs
jthomson04: Applied the direct NLL-compatible form in follow-up commit 82ca3a9064. The shared boolean and artificial scope are gone; the mutable borrow ends after its last use before the exclusive-node removal.

lib/kv-router/src/sequences/block_tracker.rs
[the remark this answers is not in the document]
jthomson04: Confirmed. The PR Validation section includes the before and after active-sequence benchmark, exact topology, workload invariants, and provenance. Logical block-visit throughput increased from 221.2M/s to 288.4M/s, while free-service p50, p99, and p999 decreased by 8.0%, 17.9%, and 24.5%. The single-run result is explicitly labeled directional.

lib/kv-router/src/sequences/block_tracker.rs
jthomson04: Added release_rejects_hash_index_alias under cfg(debug_assertions) in follow-up commit 82ca3a9064. It aliases one live hash entry to another node, releases the original chain, and pins the expected block hash index references a different live node panic.

PeaBrane (approved):
✅

PR #11569, opened 2026-07-11
perf(llm): reduce ITL metrics overhead

jthomson04:
## Overview

Reduce the cost of publishing per-token inter-token latency (ITL) metrics while preserving the existing Prometheus schema and Tower tracing behavior.

## Summary

- Buffer ITL observations in a request-local `prometheus::LocalHistogram`, flushing every 64 samples and when the request collector is dropped.
- Keep output-token counters and worker last-ITL gauges real-time.
- Clone worker-type metadata only when it is first latched, while keeping the existing six-field latch logic centralized in `set_worker_info`.
- Leave request spans, streaming-body instrumentation, endpoint plumbing, protocols, and configuration unchanged.

## Details

The global histogram previously performed synchronization for every output token. Each request now observes into a local histogram and periodically publishes the accumulated count, sum, and buckets to the same global series. Active requests may have at most 63 unpublished ITL samples; completion, error, cancellation, timeout, disconnect, and ordinary drop all destroy the request collector and flush the remainder.

### On-CPU profile

The comparison used 40-second `cycles:u` captures at 99 Hz with DWARF call graphs. Both captures contain 8,177 samples.

| Cycle-weighted on-CPU stack | Baseline | Optimized | Reduction |
| --- | ---: | ---: | ---: |
| ITL histogram path | 0.407% | 0.114% | **71.9%** |
| Overall response-metrics path | 1.180% | 0.691% | **41.5%** |

The benchmark used one frontend on CPUs 0-1; eight default-vLLM mock workers and AIPerf on CPUs 2-23; the AgentX/Weka trace at concurrency 256; block size 16; 262,144 KV blocks per worker; fastokens; a 1 GiB tokenizer cache; and jemalloc. It ran a 32-request warmup, 120-second measured phase, 40-second settle, and 40-second profile capture.

End-to-end values are regression guardrails rather than the primary result: throughput was 212.26 versus 215.42 requests/second (+1.49%), frontend CPU was 9.329 versus 9.178 ms/request (-1.62%), and p95 TTFT was 1209.79 versus 1187.79 ms (-1.82%). Both arms reported zero errors or cancellations.

Runtime reconciliation confirmed that no ITL observations were lost:

`25,161,507 output tokens - 26,162 completed streams = 25,135,345 ITL observations`

Start in `lib/llm/src/http/service/metrics.rs` at `ResponseMetricCollector`. The production change and its histogram/latching tests are contained in that file.

## Validation

- `cargo fmt --all --check`
- `cargo test -p dynamo-llm --lib` — 1,635 passed, 5 ignored
- `cargo clippy -p dynamo-llm --all-targets -- -D warnings`
- AgentX/Weka metrics-only benchmark and on-CPU profile described above
- Exact Prometheus count/sum/bucket and runtime ITL reconciliation checks

## Related Issues

### DIFF
--- a/lib/llm/src/http/service/metrics.rs
+++ b/lib/llm/src/http/service/metrics.rs
@@ -520,14 +520,14 @@ pub enum ErrorType {
 pub struct ResponseMetricCollector {
     metrics: Arc<Metrics>,
     model: String,
-    // Per-model metric handles resolved once at construction. The collector lives for a
-    // single request and `model` is fixed, so caching these avoids re-hashing the `model`
-    // label via `with_label_values` on every chunk (and, for ITL, on every output token —
-    // it was previously resolved inside a `for _ in 0..num_tokens` loop). Each handle
-    // shares the underlying metric with its vec, so observations are equivalent.
+    // Per-model metric handles cached for the request. Most are resolved at construction;
+    // ITL is resolved lazily on its first observation so requests that never produce ITL
+    // do not allocate a local histogram. Caching avoids re-hashing the `model` label on
+    // every chunk or output token. Each handle shares the underlying metric with its vec,
+    // so observations are equivalent.
     output_tokens_counter: prometheus::IntCounter,
     time_to_first_token: prometheus::Histogram,
-    inter_token_latency: prometheus::local::LocalHistogram,
+    inter_token_latency: Option<prometheus::local::LocalHistogram>,
     itl_pending_tokens: u64,
     input_sequence_length: prometheus::Histogram,
     cached_tokens: prometheus::Histogram,
@@ -1547,10 +1547,6 @@ impl ResponseMetricCollector {
         // per-chunk / per-token hot path in `observe_response` does no label hashing.
         let output_tokens_counter = metrics.output_tokens_counter.with_label_values(&[&model]);
         let time_to_first_token = metrics.time_to_first_token.with_label_values(&[&model]);
-        let inter_token_latency = metrics
-            .inter_token_latency
-            .with_label_values(&[&model])
-            .local();
         let input_sequence_length = metrics.input_sequence_length.with_label_values(&[&model]);
         let cached_tokens = metrics.cached_tokens.with_label_values(&[&model]);
         let images_per_request = metrics.images_per_request.with_label_values(&[&model]);
@@ -1561,7 +1557,7 @@ impl ResponseMetricCollector {
             model,
             output_tokens_counter,
             time_to_first_token,
-            inter_token_latency,
+            inter_token_latency: None,
             itl_pending_tokens: 0,
             input_sequence_length,
             cached_tokens,
@@ -1646,6 +1642,17 @@ impl ResponseMetricCollector {
             metrics.decode_dp_rank,
             decode_worker_type,
         );
+    }
+
+    fn local_inter_token_latency(&mut self) -> &mut prometheus::local::LocalHistogram {
+        let metrics = &self.metrics;
+        let model = self.model.as_str();
+        self.inter_token_latency.get_or_insert_with(|| {
+            metrics
+                .inter_token_latency
+                .with_label_values(&[model])
+                .local()
+        })
     }
 
     /// Observe the current output sequence length
@@ -1772,14 +1779,20 @@ impl ResponseMetricCollector {
             let itl = response_duration.as_secs_f64() / num_tokens as f64;
             self.itl_sum_secs += itl * num_tokens as f64;
             self.itl_count += num_tokens as u64;
-            // Handle resolved once at construction — the observe loop no longer re-hashes
-            // the `model` label on every output token.
-            for _ in 0..num_tokens {
-                self.inter_token_latency.observe(itl);
-            }
             self.itl_pending_tokens = self.itl_pending_tokens.saturating_add(num_tokens as u64);
-            if self.itl_pending_tokens >= ITL_LOCAL_FLUSH_TOKENS {
-                self.inter_token_latency.flush();
+            let should_flush = self.itl_pending_tokens >= ITL_LOCAL_FLUSH_TOKENS;
+            {
+                // Resolve the request-local histogram on the first ITL only, then reuse
+                // it without re-hashing the model label for each output token.
+                let histogram = self.local_inter_token_latency();
+                for _ in 0..num_tokens {
+                    histogram.observe(itl);
+                }
+                if should_flush {
+                    histogram.flush();
+                }
+            }
+            if should_flush {
                 self.itl_pending_tokens = 0;
             }
 
@@ -1820,7 +1833,9 @@ impl ResponseMetricCollector {
 
 impl Drop for ResponseMetricCollector {
     fn drop(&mut self) {
-        self.inter_token_latency.flush();
+        if let Some(histogram) = &self.inter_token_latency {
+            histogram.flush();
+        }
         self.itl_pending_tokens = 0;
 
         if !self.detokenize_latency_total.is_zero() && self.detokenize_count_total > 0 {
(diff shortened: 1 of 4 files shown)


lib/llm/src/http/service/metrics.rs
```diff
@@ -1554,9 +1563,11 @@ impl ResponseMetricCollector {
         ResponseMetricCollector {
             metrics,
             model,
+            span: tracing::Span::current(),
```
jthomson04: **Span-field loss on streaming `/v1/completions` (should fix before merge).**

This retains `Span::current()` at construction, but `completions`, `completions_single`, and `completions_batch` carry `#[tracing::instrument(skip_all)]` — INFO-level spans that are **enabled** under the default `DYN_LOG=info` filter. For those routes the retained span is the handler's `#[instrument]` span, not the `http-request` span. In `Drop`, `span.record("input_tokens"/"output_tokens"/"ttft_ms"/"avg_itl_ms"/"prefill_worker_id"/"decode_worker_id")` **silently no-ops** because those fields are declared only on `make_inference_request_span`'s span.

Before this PR, for *streaming* requests the collector dropped inside a body poll while `TraceLayer` had the `http-request` span entered, so the records landed on the right span. Net regression: streaming `/v1/completions` (single and batch) loses token counts, TTFT/ITL, and worker IDs from the request span in log context and OTLP export, at default settings. `chat_completions` has no `#[instrument]`, which is presumably why runtime validation (done on chat) passed. The Anthropic messages and `/v1/responses` handlers have `level = "debug"` instrument spans, so they hit this only under `DYN_LOG=debug`. Non-streaming completions/embeddings were already no-op at base (collector dropped inside the handler span), so those aren't regressions.

Suggested fixes, in order of preference:
1. Insert `InferenceRequestSpan` into the **request** extensions in `trace_inference_request` (before `next.run`) and have `create_response_collector` callers pass the canonical span down, falling back to `Span::current()`.
2. Or remove the `#[tracing::instrument(skip_all)]` attributes from the completions/embeddings handlers — they carry no fields and add little beyond a name.

Also worth a regression test: assert the collector's retained span is the `request_span`-target span when constructed under a nested handler span.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -60,12 +64,56 @@ async fn echo_request_id_header(
     response
 }
 
+#[derive(Clone)]
+struct InferenceRequestSpan(tracing::Span);
+
+/// Create the inference request span around handler execution without wrapping
+/// the response body. Streaming owners retain the span explicitly so body-time
+/// telemetry can use it without entering/exiting the span on every frame poll.
+async fn trace_inference_request(
```
jthomson04: **Minor:** `TraceLayer::new_for_http()` also emitted `DefaultOnRequest`'s `"started processing request"` DEBUG event per inference request; the new middleware doesn't. Mitigated by `InflightGuard`'s `"request received"` INFO event, but it's a small deviation from the "events unchanged" claim — worth a line in the PR description or restoring.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -60,12 +64,56 @@ async fn echo_request_id_header(
...
+    }
+
+    if let ClassifiedResponse::Ready(Err(failure_class)) =
+        ServerErrorsAsFailures::default().classify_response(&response)
+    {
+        let mut on_failure = DefaultOnFailure::default();
+        span.in_scope(|| on_failure.on_failure(failure_class, start.elapsed(), &span));
+    }
+
+    response.extensions_mut().insert(InferenceRequestSpan(span));
```
jthomson04: **Nit:** the layer ordering is now load-bearing — `track_inflight_inference` (outer) depends on `trace_inference_request` (inner) inserting this extension. A one-line comment here or at the `.layer(...)` call sites would protect against someone reordering the layers. (Leaving the extension in the outgoing response is harmless since extensions aren't serialized, but removing it in `track_inflight_inference` would also make the handoff explicit.)

lib/llm/src/http/service/metrics.rs
```diff
@@ -1620,6 +1631,27 @@ impl ResponseMetricCollector {
         }
     }
 
+    fn set_worker_info_from_metrics(&mut self, metrics: &LLMMetricAnnotation) {
```
jthomson04, line 1625: **Nit:** this duplicates the six-field latch logic in `set_worker_info`. Consider implementing one in terms of the other so a future field only needs to be added in one place.

lib/llm/src/http/service/openai.rs
```diff
@@ -3197,33 +3206,38 @@ async fn video_stream(
...
-            "--frame\r\nContent-Type: image/jpeg\r\nContent-Length: {}\r\n\r\n",
-            jpeg_bytes.len()
-        );
-        let mut chunk = Vec::with_capacity(header.len() + jpeg_bytes.len() + 2);
-        chunk.extend_from_slice(header.as_bytes());
-        chunk.extend_from_slice(&jpeg_bytes);
-        chunk.extend_from_slice(b"\r\n");
-        Some(Ok::<Bytes, std::convert::Infallible>(Bytes::from(chunk)))
+    let request_span = tracing::Span::current();
+    let frame_span = request_span.clone();
```
jthomson04: **Nit:** `frame_span` *is* the request span, kept only to satisfy the closure move — the naming suggests a per-frame child span. A rename (e.g. `span_for_frames`) or comment would help. Also, the per-frame `Span::clone` does `clone_span`/`try_close` subscriber calls on every frame; cheap relative to enter/exit, but hoisting into an `Arc<Span>` would eliminate it if MJPEG frame rates matter.

jthomson04 (commented):
Reviewed against the vendored tower-http 0.6.8 and prometheus 0.14 sources. The parity claims hold up well:

- The new `"http response sent"` event is identical to the old `on_response` closure (message, `status`/`latency_ms` fields, levels, module target).
- Using the literal `DefaultOnFailure`/`DefaultOnEos` types preserves the exact `"response failed"`/`"end of stream"` events; `on_eos` did fire at end-of-stream in the old code (the `Ready` classification branch keeps it), so the emulation in `track_inflight_inference` is true parity.
- Body-error `on_failure` was already dead code with `ServerErrorsAsFailures` (`classify_eos` is `None` in the `Ready` branch), so dropping it loses nothing.
- Middleware ordering is correct (`trace_inference_request` inner, `track_inflight_inference` outer), and drain rejections bypassed `TraceLayer` before too.
- `InflightGuard::Drop` already enters its retained span, so `"request completed"` stays correlated.
- `LocalHistogram` is `RefCell`-based (`Send`, sufficient here) and auto-flushes on `Drop` in prometheus 0.14, so the explicit flush is safe defense-in-depth.
- `set_worker_info` already latched, so `set_worker_info_from_metrics` only moves the clone behind the latch — no behavior change.

One real issue to address before merge (span-field loss on streaming `/v1/completions` — see inline comment on `metrics.rs`), plus a few minor notes inline.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -60,12 +64,56 @@ async fn echo_request_id_header(
...
+            "http response sent"
+        );
+    }
+
+    if let ClassifiedResponse::Ready(Err(failure_class)) =
+        ServerErrorsAsFailures::default().classify_response(&response)
+    {
+        let mut on_failure = DefaultOnFailure::default();
+        span.in_scope(|| on_failure.on_failure(failure_class, start.elapsed(), &span));
+    }
```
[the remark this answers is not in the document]
jthomson04: Addressed in 282ad9e766 by removing the custom inference tracing implementation entirely. Inference routes now use the existing Tower TraceLayer unchanged, so this PR no longer adds a custom duplicate-event path. The cumulative PR diff is metrics-only.

lib/llm/src/http/service/metrics.rs
jthomson04: Addressed in 282ad9e766 by removing ResponseMetricCollector span retention and restoring the existing Tower body instrumentation and ambient span recording. The final PR has no tracing changes, so completions span behavior matches the base implementation. Validation: cargo test -p dynamo-llm --lib (1,635 passed, 5 ignored) and clippy with -D warnings.

lib/llm/src/http/service/service_v2.rs
jthomson04: Addressed in 282ad9e766 by restoring the existing inference TraceLayer unchanged, including DefaultOnRequest and its request-start event. The final PR contains no service_v2 tracing changes.

lib/llm/src/http/service/service_v2.rs
jthomson04: Addressed in 282ad9e766 by removing the custom middleware pair and response-extension span handoff. There is no longer a load-bearing middleware order introduced by this PR.

lib/llm/src/http/service/metrics.rs
jthomson04: Addressed in 282ad9e766. set_worker_info_from_metrics now conditionally clones worker-type strings only when unlatch, then delegates all six assignments to the existing set_worker_info implementation. The first-value latch regression test passes, along with the full dynamo-llm library suite.

lib/llm/src/http/service/openai.rs
jthomson04: Addressed in 282ad9e766 by reverting the OpenAI tracing plumbing. MJPEG streaming now matches the base implementation, and this PR no longer adds a per-frame Span clone.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -82,10 +130,24 @@ async fn track_inflight_inference(
...
+        .map(|request_span| request_span.0.clone());
+    let mut body = body.into_data_stream();
+    let mut on_eos = request_span.map(|span| (DefaultOnEos::default(), Instant::now(), span));
     // Keep the permit alive until the full response body, including streams,
     // finishes or is dropped.
-    let stream = body.into_data_stream().map(move |result| {
+    let stream = futures::stream::poll_fn(move |cx| {
         let _permit = &permit;
+        let result = Pin::new(&mut body).poll_next(cx);
+        if matches!(result, Poll::Ready(None))
```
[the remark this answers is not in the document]
jthomson04: Addressed in 282ad9e766 by removing the custom poll_fn tracing middleware altogether. The final PR has no service_v2 changes, and cargo test -p dynamo-llm --lib plus clippy with -D warnings both pass.

ishandhanani (approved):

PR #11570, opened 2026-07-11
fix(frontend): reject over-budget max tokens

sytianhe:
## Overview:

Reject requests during frontend preprocessing when an explicit output-token budget plus prompt tokens exceeds the model context length. This aligns Dynamo request validation with backend behavior such as vLLM returning a client error for over-budget requests.

## Summary

- Add frontend validation for explicit `max_tokens` / `max_completion_tokens` budgets after prompt tokenization.
- Preserve existing omitted-token behavior by only rejecting explicitly requested budgets.
- Cover the Responses API handoff so `max_output_tokens` reaches the same chat budget field before preprocessing.

## Details:

The budget check runs after Dynamo has an effective prompt token count. Text-only requests use the tokenized prompt length. Multimodal requests use the expanded multimodal prompt length when routing metadata provides one, and defer to backend validation when Dynamo only has unexpanded image placeholder tokens.

The completion path performs the same validation after tokenization and skips prompt-embedding requests, where token length is not represented by `token_ids`.

- `lib/llm/src/preprocessor.rs`: `validate_explicit_max_tokens_budget` and its call sites in chat-like and completion preprocessing.
- `lib/llm/src/http/service/openai.rs`: Responses API conversion test for `max_output_tokens`.

## Validation

- `cargo fmt --check`
- `git diff --check upstream/main...HEAD`
- `cargo test -p dynamo-llm test_validate_explicit_max_tokens_budget --lib` could not run to completion locally on macOS because `dynamo-llm` fails to compile before tests on Linux-gated imports: `dynamo_memory::numa`, `nix::fcntl::fallocate`, and `OFlag::O_DIRECT`.
- `cargo test -p dynamo-llm test_responses_max_output_tokens_reaches_chat_budget_field --lib` hit the same local macOS compile gate before tests.

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11570" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured maximum output token settings are consistently applied across Responses API requests.
  * Added validation to prevent requested prompt and completion tokens from exceeding the model’s context limit.
  * Requests that exceed the available context now return a clear error describing the configured limit and token usage.

### DIFF
--- a/lib/llm/src/preprocessor.rs
+++ b/lib/llm/src/preprocessor.rs
@@ -318,6 +318,38 @@ impl OpenAIPreprocessor {
         Some(context_length.saturating_sub(prompt_len as u32))
     }
 
+    fn validate_explicit_max_tokens_budget(
+        prompt_len: usize,
+        max_tokens: Option<u32>,
+        context_length: u32,
+    ) -> Result<()> {
+        let Some(max_tokens) = max_tokens else {
+            return Ok(());
+        };
+        if context_length == 0 {
+            return Ok(());
+        }
+
+        let max_len = context_length as usize;
+        let completion_len = max_tokens as usize;
+        let requested_tokens = prompt_len.saturating_add(completion_len);
+        if requested_tokens > max_len {
+            return Err(DynamoError::builder()
+                .error_type(ErrorType::InvalidArgument)
+                .message(format!(
+                    "This model's maximum context length is {} tokens. \
+                     However, you requested {} tokens \
+                     ({} in the messages, {} in the completion). \
+                     Please reduce the length of the messages or completion.",
+                    max_len, requested_tokens, prompt_len, completion_len,
+                ))
+                .build()
+                .into());
+        }
+
+        Ok(())
+    }
+
     /// Prompt length for sizing the omitted-`max_tokens` cap. Prefers the
     /// MM-expanded length; a 0 (serde-default / absent) counts as missing. With
     /// images but no expanded length, defers to the backend (`None`); text-only
@@ -943,6 +975,13 @@ impl OpenAIPreprocessor {
             has_images,
             preprocessed.token_ids.len(),
         );
+        if let Some(prompt_len) = effective_prompt_len {
+            Self::validate_explicit_max_tokens_budget(
+                prompt_len,
+                preprocessed.stop_conditions.max_tokens,
+                self.context_length,
+            )?;
+        }
         if preprocessed.stop_conditions.max_tokens.is_none()
             && let Some(prompt_len) = effective_prompt_len
             && let Some(max_tokens) =
@@ -3540,6 +3579,13 @@ impl
             .await?;
 
         let mut common_request = builder.build()?;
+        if common_request.prompt_embeds.is_none() {
+            Self::validate_explicit_max_tokens_budget(
+                common_request.token_ids.len(),
+                common_request.stop_conditions.max_tokens,
+                self.context_length,
+            )?;
+        }
         attach_agent_context_from_context(&mut common_request, &context);
 
         let trace_state = crate::request_trace::build_request_end_trace_state(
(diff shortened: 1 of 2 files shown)


rmccorm4:
/ok to test [140790b](https://github.com/ai-dynamo/dynamo/pull/11570/commits/140790bf4372b273a4156a76707b4b3b816f32dc)

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `e88be5d` | unsigned |
| `5054735` | unsigned |

lib/llm/src/preprocessor.rs
```diff
@@ -318,6 +318,38 @@ impl OpenAIPreprocessor {
         Some(context_length.saturating_sub(prompt_len as u32))
     }
 
+    fn validate_explicit_max_tokens_budget(
```
rmccorm4, line 321: @biswapanda looking for your thoughts on:
1. should this be on or off by default
2. should we have a flag to toggle the behavior in the event the default setting is not desired
3. how do we relate this to sglang args like `--allow-auto-truncate`

lib/llm/src/preprocessor.rs
```diff
@@ -943,6 +975,13 @@ impl OpenAIPreprocessor {
             has_images,
             preprocessed.token_ids.len(),
         );
+        if let Some(prompt_len) = effective_prompt_len {
+            Self::validate_explicit_max_tokens_budget(
```
rmccorm4, line 979: @biswapanda thoughts on whether should this be called at a different spot (here) or bundled into similar existing validation around here: https://github.com/ai-dynamo/dynamo/blob/aad60fe1c5501496e9245888fd76826f93525f7f/lib/llm/src/preprocessor.rs#L1767-L1795?

lib/llm/src/preprocessor.rs
biswapanda: Initially, this appeared to involve two distinct cases. There is one caveat: when SGLang is configured with `--allow-auto-truncate`, Case 1 may be transformed into Case 2 because SGLang first truncates the input prompt to fit within the model’s context window.

1.  input_tokens > context window
2.  input_tokens + requested_output_tokens > context window

biswapanda: My guiding principles ~

1. No surprises for end-user:   The default behavior should minimize unexpected differences, particularly when users switch between backends.

2. Preserve backend capabilities: Dynamo frontend should allow users to take advantage of the capabilities provided by each backend rather than imposing unnecessary frontend restrictions.

Backend behavior currently differs:

   * vLLM rejects requests whose combined input and requested output exceed the context window.
   * TensorRT-LLM can clamp the requested output length to the remaining context capacity.
   * SGLang can clamp the output length and, when configured with `--allow-auto-truncate`, can also truncate an oversized input prompt.

biswapanda: 1. Option A: Delegate the decision to the backend

The simplest approach is for the frontend to reject only requests where the input prompt itself exceeds the model context window, unless the backend is explicitly configured to support prompt truncation.

For requests where the prompt fits but the requested output exceeds the remaining context capacity, Dynamo would preserve the requested value and allow the backend to reject or clamp it according to its native behavior.

2. Option B: Expose an explicit frontend policy

For greater user control, introduce a frontend argument such as:

```text
--output-overflow-policy=[reject|clamp|backend]
```

The default would be `backend`, which is equivalent to Option A.

The policies would behave as follows:

* **`reject`**: Apply strict vLLM-style behavior. Reject the request when:

  ```text
  input_tokens + requested_output_tokens > context_length
  ```

* **`clamp`**: Reduce the requested output length in the Dynamo frontend so that it fits within the remaining context capacity:

  ```text
  effective_output_tokens =
      context_length - input_tokens
  ```

* **`backend`**: Preserve the explicitly requested output length and allow the selected backend to decide whether to reject or clamp the request.

Prompt overflow should be controlled separately, for example:

```text
--prompt-overflow-policy=[reject|backend]
```

Here:

* `reject` rejects prompts that exceed the admissible input length.
* `backend` forwards the request so that backends with explicit prompt-truncation support, such as SGLang with `--allow-auto-truncate`, can handle it.

Keeping prompt-overflow and output-overflow policies separate avoids conflating two different operations:

* Prompt overflow requires removing input content.
* Output overflow can often be handled safely by reducing the generation budget.

rmccorm4:
Related https://github.com/ai-dynamo/dynamo/pull/12092

Aphoh:
Reconciled in #12092. I incorporated this PR's useful legacy Completions validation path and Responses API max_output_tokens handoff regression, with co-author credit to Tianhe in commit 8628dadd2b.

The remaining behavioral difference is the backend policy discussed in this review thread: this PR validates unconditionally against runtime_config.context_length, which can override native clamp/truncate behavior. #12092 instead has SGLang publish strict_request_token_limit only when total-token validation is enabled and --allow-auto-truncate is disabled; the published value also subtracts SGLang's reserved speculative-token budget. The shared frontend preprocessor rejects only when that explicit engine contract is present and otherwise defers to the backend. It also defers prompt embeddings and multimodal inputs whose expanded length is not exact.

That makes #12092 the consolidated version of this fix while preserving the coverage added here. Thanks for identifying the Completions and Responses surfaces.

rmccorm4:
Superseded by https://github.com/ai-dynamo/dynamo/pull/12092

PR #10481, opened 2026-06-09
feat(router): add GMS placement protocol [GMS 16/18]

hutm:
## Summary

Define the placement events and protocol types needed for the router to reason about GMS-resident KV content and decode-to-decode transfer.

## Scope

Adds protocol surfaces and compatibility wiring only; indexing policy and publication/orchestration are separate follow-up PRs.

## Stack

Position **16/18** in the GMS-managed KV review train. This PR is based on `review/gms-v2-latehost-08-2-2-host-tier-transport-backends`.

Parent: #10450  
Tracking: #10457

## Review migration

This existing PR is updated in place so its comments and reviews remain attached. Line comments may appear outdated where code moved during the rebase; they remain visible and should be dispositioned against the current diff. Previous approvals should be revalidated.

The train is rebased on `main` at `aeda23b483831df2f75b697b8ccf65f4ffd9f3d6`. The reordered functionality is preserved while each PR boundary is independently buildable and lintable: compatibility call sites and the engine-facing placement adapter now appear at first use; missing type imports and test markers were added; repository-pinned formatting was applied; one unused constant was removed; and every commit carries a DCO sign-off. The complete range passes `git diff --check` and the full pre-commit suite.

## Validation

Passed `cargo check -p dynamo-kv-router -p dynamo-llm --all-targets` at this stack boundary before the exact latest-main rebase.

Issue #10457, opened 2026-06-09
Add optional router event-plane placement for decode-decode KV transfer


hutm:
## Problem

For decode-decode KV transfer, router can make better placement decisions if it knows where reusable KV exists. However, router should not connect to GMS on the request path or learn GMS internals.

## Proposal

Publish GMS placement metadata through the existing KV event plane. Add optional router/indexer support that consumes placement events and uses them for decode-decode routing decisions when enabled. Keep the feature disabled by default and isolate it late in the PR train.

## Scope

- Placement protocol/config surfaces.
- Router/indexer placement lookup and policy.
- Worker-side placement publication.
- E2E router coverage for vLLM, SGLang, and TensorRT-LLM configurations.
- Documentation of feature flags and default-off behavior.

## Acceptance Criteria

- Router does not open request-path connections to GMS.
- Placement metadata flows over the event plane.
- Decode-decode routing remains optional and disabled by default.
- Existing router/worker contract remains valid for standard Dynamo deployments.

## Parent Issue

- #10450 Umbrella: Make Dynamo inference workers restartable with GMS-managed KV

## Current Review PRs

- #10481 — placement protocol
- #10482 — placement index and policy
- #10483 — publication and decode-to-decode orchestration

hutm:
Relevant PRs:

- #10481 router: add GMS placement protocol surfaces
- #10482 router: add GMS placement indexer policy
- #10483 router: publish GMS placement from workers
- #10484 router: add GMS placement E2E coverage

hutm:
PR-train migration update: Router placement remains the final optional phase; close only after all three PRs merge and default-off behavior is verified. The issue remains open; no acceptance criterion is being waived by the review-stack rewrite.

hutm:
This child tracker is deprecated together with #10450. The active replacement is the deliberately MVP-only vLLM tracker #11600. Any subset required for that local vLLM failover loop is listed explicitly there; all broader functionality is deferred to future milestones. Closing this ticket as superseded/deferred, not completed. Existing PR links and review history remain untouched.

PR #10368, opened 2026-06-05
feat(kv-router): Route vLLM CPU KV events to HostPinned and count lower-tier applies

Change72:
## What

This PR makes Dynamo's KV router consume and observe vLLM native CPU-offload KV events.

Changes:

- Treat vLLM's `medium="CPU"` as an alias for the existing `HostPinned` tier, in the shared
  `StorageTier::from_kv_medium` normalization layer (so the alias also applies to the
  `kv_consolidator` consumer — verified consistent: `CPU` → `HostPinned`).
- Preserve existing `CPU_PINNED` and `CPU_TIER1` behavior.
- Add wire tests for placeholder CPU payloads and full self-describing CPU payloads.
- Count lower-tier `Stored` / `Removed` / `Cleared` applies in `kv_cache_events_applied`.
- Wire the metrics handle into lazily-created lower-tier indexers used by the `dynamo-llm` router
  assembly path.

## Where should the reviewer start?

1. `lib/kv-router/src/protocols.rs` — the `medium="CPU"` -> `StorageTier::HostPinned` alias.
2. `lib/kv-router/src/zmq_wire/tests.rs` — focused wire tests for placeholder and full CPU
   `BlockStored` payloads.
3. `lib/kv-router/src/indexer/lower_tier.rs` — lower-tier worker now increments
   `kv_cache_events_applied` using the same `(event_type, status)` labels as the primary indexer.
4. `lib/kv-router/src/indexer/lower_tier_indexers.rs`,
   `lib/kv-router/src/indexer/local.rs`, and `lib/llm/src/kv_router/indexer/mod.rs` — metrics
   handles are threaded into lazily-created lower-tier indexers.

## Why

vLLM's native `OffloadingConnector` publishes CPU-tier KV events with `medium="CPU"`. Dynamo
previously did not classify that string as `HostPinned`, so the event could not be routed to the
lower-tier indexer as intended.

After adding the alias, the lower-tier path also needed observability. The first metrics patch made
`LowerTierIndexer::worker` increment the same `kv_cache_events_applied` counter as the primary
device-tier indexer, but the `dynamo-llm` router assembly path still constructed lower-tier
indexers without a metrics handle. The final wiring patch passes the shared metrics handle through
`LocalKvIndexer` and `LowerTierIndexers::new_with_metrics`, including lazily-created
HostPinned/Disk/External indexers.

This is why there are two metrics commits:

1. Apply-path support: count lower-tier events when a lower-tier indexer has metrics.
[...]

PR #11571, opened 2026-07-11
feat(kv-router): route vLLM STORAGE KV events to the Disk tier with locality gating

Change72:
## Summary

Route vLLM native-storage KV events to the router's **Disk lower tier** instead of letting them pollute the device (G1) primary radix tree, and consume the new per-event `locality` field with a fail-closed policy for non-local events. This is a full rewrite of the earlier draft, scoped down to the minimum needed for correct routing.

Upstream wire contract this builds on:
- **vLLM #48281** — adds optional `locality: "LOCAL" | "REMOTE"` on `BlockStored` / `BlockRemoved` (shipped in v0.26.0); absent means unspecified and vLLM never infers it.
- **vLLM #48123** — collapses native FS/OBJ offload into a single `STORAGE` medium (on `main`, v0.27+).

We map only `STORAGE`. The older `FS` / `OBJ` mediums (vLLM #47923, shipped in v0.26.0) are superseded by `STORAGE` upstream; they stay unrecognized here and keep today's fallback behavior. Hardening dynamo's handling of unrecognized media (fail-closed instead of defaulting to Device) is a separate follow-up PR — it is a pre-existing latent issue on `main`, independent of this change.

## Behavior

| medium | locality | behavior |
|---|---|---|
| absent / `GPU` | absent | Device tier, worker-local (unchanged) |
| `CPU` (and legacy `CPU_PINNED` / `CPU_TIER1`) | absent | HostPinned, worker-local (unchanged, #10368) |
| legacy `CPU_TIER2` / `DISK` / `NVME` (consolidator republish) | absent | Disk, worker-local (unchanged) |
| `STORAGE` | absent or `LOCAL` | **Disk, worker-local (new)** |
| any | `REMOTE` or unknown locality | **dropped (new; reserved for a future shared index)** |

## Design notes

- **`None` locality = local.** dynamo has no shared/global index yet (#10457), so `REMOTE` events have no consumer and are dropped. Treating absent locality as local matches legacy CPU-offload events, which never carried the field and have always been worker-local. A genuinely shared FS mount must be declared `locality=REMOTE` on the vLLM side (its documented contract); mislabeling only biases routing toward one worker (suboptimal, not incorrect).
- **Normalizer isolation.** Lower-tier events carry no `extra_keys` / `cache_salt`, so they are routed straight past the `ZmqEventNormalizer` — they must not mutate per-group metadata or the salted cache-namespace propagation chain, nor are they in scope of the SW/SSM group filter. (Direct regression test included.)
- **Consolidators stay G1-only.** Both ZMQ ingresses (`kvbm-consolidator` and the LLM `kv_consolidator`) drop recognized non-device media. The consolidator's Disk tier denotes KVBM G2/G3 ownership; vLLM-native `STORAGE` is a separate system and must not be conflated with KVBM-owned blocks (KVBM offload arrives via its own source).
- **Routability.** On vLLM 0.27+ with `self_describing_kv_events` (#48679), rich LOCAL/absent `STORAGE` blocks carry full `token_ids` / `block_size`, so they flow through the existing token-derived lower-tier indexer and are immediately routable. Placeholder events (`block_size=0`, empty tokens — all of v0.26) are filtered to an empty block list by `create_stored_blocks`, a harmless index no-op.

## Scope

Relative to the previous draft this drops `Placement::shared` / `PlacementOwner::Shared` / `resolve_placement` / `worker_local_event_tier`, and leaves `services/indexer/listener.rs`, `kv_router/publisher/event_processor.rs`, and `kv_consolidator/publisher.rs` untouched — `convert_event` only ever yields a local-worker placement or drops, so the listener's existing local-only handling is unaffected. Unrecognized-media fail-closing is deliberately deferred to a follow-up (see Summary). Complements #11930 (lower-tier removal) and does not overlap the GMS placement line (#10481–10483 / #10457): no `Placement` / owner changes here.

~45 lines of production code; the rest is tests.

## Validation

All commands run on a Linux workstation (cargo 1.95):

```
cargo test -p dynamo-kv-router --lib                                # ok, 0 failed
cargo test -p dynamo-kv-router --features standalone-indexer --lib  # ok, 0 failed
cargo test -p dynamo-llm --lib kv_router::publisher::tests::event_processor_tests  # ok, 0 failed
cargo test -p dynamo-llm --lib block_manager::kv_consolidator       # ok, 0 failed
cargo test -p kvbm-consolidator                                     # ok, 0 failed
cargo fmt --all -- --check                                          # clean
cargo clippy --no-deps --all-targets -- -D warnings                 # clean (0 warnings)
```

The new G1-only ingress tests (`block_manager::kv_consolidator::subscriber::tests::process_event_tracks_only_g1_device_events` and `ingress::zmq_subscriber::tests::process_event_tracks_only_g1_device_events`) and the new `zmq_wire::tests::*` cases (locality decode symmetry, nil/legacy locality, mixed-case → Unknown, locality/tier gating, placeholder no-op, normalizer non-pollution) all pass.

The Python contract test (`components/src/dynamo/vllm/tests/test_vllm_kv_events_api.py`) was updated to assert the `locality` field but not executed here — no vLLM environment is available on the workstation. The assertions are `_has_locality`-guarded (same pattern as the existing `_has_kv_cache_spec_sliding_window` checks), so they are inert against vLLM builds without the field.

## AI assistance

Implemented with AI assistance (Cursor). All design decisions, the scope reduction, and the review were human-directed.

## Summary by CodeRabbit

- **New Features**
  - Added storage locality support for block events, including local, remote, and unknown locations.
  - Added access to storage-medium information for event processing.

- **Bug Fixes**
  - Remote and unknown-locality events are now safely excluded from placement processing.
  - Lower-tier storage events no longer affect GPU tracking or namespace metadata.
  - CPU tier-two storage is now correctly classified as disk storage.

- **Tests**
  - Expanded coverage for locality handling, legacy event formats, storage classification, and filtering behavior.


dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `4f8ca4f` | unsigned |
| `8872753` | unsigned |

lib/kv-router/src/zmq_wire/mod.rs
```diff
@@ -127,6 +127,20 @@ impl ZmqEventNormalizer {
             return Err(ZmqEventFilterReason::IgnoredEvent);
         }
 
+        // Hash-only lower-tier events (STORAGE -> Disk, and External) carry no
```
jthomson04, line 142: Note for future work: current vLLM self-describing offload events set `extra_keys=None`. Because Dynamo derives `cache_salt` / `cache_namespace` from `extra_keys`, rich `STORAGE` blocks are hashed unsalted and cannot match cache-salted query hashes. I don't think this PR can fix the missing upstream metadata, but could we document cache-salted native storage as unsupported here and track the vLLM follow-up?

lib/kv-router/src/protocols.rs
```diff
@@ -440,7 +440,7 @@ impl StorageTier {
         match medium {
             "GPU" | "DEVICE" => Some(Self::Device),
             "CPU" | "CPU_PINNED" | "CPU_TIER1" => Some(Self::HostPinned),
-            "CPU_TIER2" | "DISK" | "NVME" => Some(Self::Disk),
+            "CPU_TIER2" | "DISK" | "NVME" | "STORAGE" => Some(Self::Disk),
```
jthomson04, line 443: This mapping also enables `STORAGE` events in the standalone indexer's direct ZMQ path. vLLM documents that self-describing offload chunks may re-announce shared block hashes and consumers must reference-count the matching store/remove announcements. The embedded LLM publisher has `EventDedupFilter`, but the standalone listener feeds the set-based lower-tier index directly, so the first remove can delete a block that another offload chunk still references. This is already a limitation for native CPU offload; `STORAGE` inherits it here. Do we intend standalone indexer to support these native offload events? If so, it should share the reference-counting behavior; otherwise we should document the limitation.

lib/kv-router/src/zmq_wire/convert.rs
```diff
@@ -22,13 +22,29 @@ pub fn convert_event(
...
+        } => (medium.as_deref(), *locality),
+        RawKvEvent::AllBlocksCleared => (None, None),
         RawKvEvent::Ignored => return None,
     };
+
+    // No consumer exists for a shared/global index yet (dynamo #10457), so
+    // REMOTE and any unrecognized locality fail closed. Absent or LOCAL keeps
+    // the event worker-local, matching legacy CPU-offload events that never
+    // carried a locality field.
+    if matches!(locality, Some(Locality::Remote | Locality::Unknown)) {
```
jthomson04, line 43: Could we reject `REMOTE` / unknown locality in `preprocess_with_reason` instead? As written, the LLM ZMQ listener records the event as accepted and allocates `next_event_id` before conversion returns `None`. The next accepted local event then appears to have an input gap and incorrectly increments `engines_dropped_events`. Both stores and removes are already dropped; moving the same gate earlier would classify them as intentionally filtered and avoid the false warning/metric.

jthomson04 (commented):
Three non-blocking comments on native `STORAGE` behavior and observability. The cache-salt item is a note for upstream/future work; the standalone item asks whether to support or document the inherited limitation; the locality item addresses false dropped-event accounting.

lib/kv-router/src/zmq_wire/mod.rs
Change72: documented in [4688306](https://github.com/ai-dynamo/dynamo/pull/11571/commits/4688306c4a49b14cc3b4dccd427af7a55d373b55)

lib/kv-router/src/protocols.rs
Change72: out of this PR scope; being addressed by draft #11930 (chunk-owner tracking in the lower-tier indexer, which the standalone path shares)

lib/kv-router/src/zmq_wire/convert.rs
Change72: fixed in [1c19413](https://github.com/ai-dynamo/dynamo/pull/11571/commits/1c19413e0fd88fee68cd755243ad961f0e61a303).

jthomson04 (approved):

PeaBrane (approved):

PR #11574, opened 2026-07-11
perf(runtime): avoid copying received ZMQ frames

jthomson04:
## Overview:

Eliminate receive-side copies in the runtime ZMQ socket pump while preserving wire format, validation, broadcast, shutdown, and backpressure behavior.

The previous path copied every multipart frame into `Vec<Vec<u8>>`, then cloned the payload vector again before `Frame::decode`. In the saturated 16-frontend AgentX profile, those copies accounted for 7.37% of sampled frontend CPU. This change eliminated 63.5% of ZMQ receive-pump CPU time per completed request in that profile.

## Details:

- Process received `tmq::Multipart` frames directly.
- Borrow the publisher and sequence frames without allocation.
- Move the payload `tmq::Message` into `Bytes::from_owner`, allowing the existing `Frame::decode` path to retain a zero-copy slice.
- Keep payload extraction panic-free after the existing four-frame validation.
- Declare `bytes = "1.9"` as the workspace minimum because `Bytes::from_owner` was introduced in 1.9.0.
- Cover inline and heap-backed message ownership, clones, slices, thread transfer, delayed drop, and continuation after malformed messages.

This does not change queue sizing, HWM behavior, retry or drop policy, batching, topology, or any public API.

Start with `ZmqSubTransport::start_socket_pump` and `ZmqMessageOwner` in `lib/runtime/src/transports/event_plane/zmq_transport.rs`. The ownership and malformed-message tests are in the same module. The only other change is the workspace `bytes` minimum in the root `Cargo.toml`.

## Related Issues

## Validation:

- `git diff --check`
- `cargo test -p dynamo-runtime transports::event_plane::zmq_transport::tests --lib` — 4 passed
- `cargo build --release -p dynamo-runtime`
- `cargo metadata --locked --format-version 1 --no-deps` in the root workspace, `lib/bindings/python`, `lib/runtime/examples`, and `lib/bindings/kvbm`

All four locked metadata checks passed without lockfile changes.

The performance profiles below used `main` with #11568 grafted on top because that was the frontend profile under investigation. This ZMQ receive change was the only A/B difference; this PR itself targets `main` and does not depend on #11568.

### Matched zero-drop AgentX profile

Single matched on-CPU A/B with four frontends sharing CPUs 0-1, four mock workers on CPUs 2-23, KV routing with replica sync, block size 16, 262144 blocks per mocker, mocker speedup 1000000, fastokens, 128 MiB tokenizer cache per frontend, and AgentX concurrency 2. Both arms were unsaturated and reported zero subscriber lag, request errors, and cancellations.

The change eliminated **11.2% of ZMQ receive-pump CPU time per completed request**. The pump's share of sampled frontend CPU fell from 5.24% to 4.44%, a 0.80 percentage-point reduction, and sampled receive-pump `memcpy` CPU was eliminated entirely.

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Throughput | 33.084 req/s | 33.062 req/s | -0.07% |
| Frontend CPU utilization | 33.9% | 34.8% | +0.9 pp |
| ZMQ receive-pump CPU/request | 4.011M sampled cycles | 3.562M sampled cycles | **-11.2%** |
| ZMQ receive-pump `memcpy` CPU/request | 595K sampled cycles | 0 | **-100%** |
| Mean RSS/frontend | 0.396 GiB | 0.390 GiB | -1.4% |
| Peak aggregate frontend RSS | 1.780 GiB | 1.698 GiB | -4.6% |

The matched state at capture start was also close: 391,920 vs. 393,007 successful stored KV events, 205.18M vs. 206.00M input tokens, and 193.50M vs. 193.94M tokenizer-cached tokens.

### Saturated AgentX profile

A separate single A/B used 16 frontends on CPUs 0-1 and AgentX concurrency 256. The change eliminated **63.5% of ZMQ receive-pump CPU time per completed request**, including 99.8% of sampled receive-pump `memcpy` CPU. The pump's share of sampled frontend CPU fell from 16.18% to 6.11%, a 10.07 percentage-point reduction, while throughput increased 3.4%.

The candidate also consumed more event traffic in this overloaded regime, so the aggregate RSS increase observed there is not treated as a matched memory comparison; the zero-drop profile above is the memory validation.

### DIFF
--- a/lib/runtime/src/transports/event_plane/zmq_transport.rs
+++ b/lib/runtime/src/transports/event_plane/zmq_transport.rs
@@ -21,7 +21,7 @@ use bytes::Bytes;
 use futures::{SinkExt, StreamExt};
 use std::sync::{Arc, OnceLock};
 use tmq::{
-    AsZmqSocket, Context, Multipart, SocketBuilder,
+    AsZmqSocket, Context, Message, Multipart, SocketBuilder,
     publish::{Publish, publish},
     subscribe::{Subscribe, subscribe},
 };
@@ -67,8 +67,16 @@ where
         .set_rcvtimeo(ZMQ_RCVTIMEOUT_MS)
 }
 
-fn multipart_message(multipart: Multipart) -> Vec<Vec<u8>> {
-    multipart.into_iter().map(|frame| frame.to_vec()).collect()
+/// Keeps a received ZMQ message alive for as long as any derived `Bytes` exists.
+///
+/// `Bytes::from_owner` obtains the message data pointer only after moving this
+/// owner into stable storage, so this also supports libzmq's inline messages.
+struct ZmqMessageOwner(Message);
+
+impl AsRef<[u8]> for ZmqMessageOwner {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
 }
 
 /// ZMQ PUB transport for publishing events.
@@ -280,8 +288,8 @@ impl ZmqSubTransport {
                     break;
                 };
 
-                let frames = match result {
-                    Ok(frames) => multipart_message(frames),
+                let mut frames = match result {
+                    Ok(frames) => frames,
                     Err(error) => {
                         tracing::error!(error = %error, "ZMQ receive error in socket pump");
                         break;
@@ -304,8 +312,7 @@ impl ZmqSubTransport {
                     );
                     continue;
                 }
-                let publisher_id =
-                    u64::from_be_bytes(publisher_id_bytes.as_slice().try_into().unwrap());
+                let publisher_id = u64::from_be_bytes(publisher_id_bytes[..].try_into().unwrap());
 
                 let sequence_bytes = &frames[2];
                 if sequence_bytes.len() != 8 {
@@ -315,7 +322,7 @@ impl ZmqSubTransport {
                     );
                     continue;
                 }
-                let sequence = u64::from_be_bytes(sequence_bytes.as_slice().try_into().unwrap());
+                let sequence = u64::from_be_bytes(sequence_bytes[..].try_into().unwrap());
 
                 tracing::trace!(
                     publisher_id = publisher_id,
@@ -323,7 +330,10 @@ impl ZmqSubTransport {
                     "Socket pump received ZMQ message"
                 );
 
-                let frame_bytes = Bytes::from(frames[3].clone());
+                let frame_message = frames
+                    .pop_back()
+                    .expect("four-frame multipart must contain a payload");
+                let frame_bytes = Bytes::from_owner(ZmqMessageOwner(frame_message));
                 match Frame::decode(frame_bytes) {
                     Ok(frame) => {
                         let _ = broadcast_tx.send(frame.payload);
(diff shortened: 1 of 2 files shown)


lib/runtime/src/transports/event_plane/zmq_transport.rs
```diff
@@ -315,15 +322,18 @@ impl ZmqSubTransport {
...
                 tracing::trace!(
                     publisher_id = publisher_id,
                     sequence = sequence,
                     "Socket pump received ZMQ message"
                 );
 
-                let frame_bytes = Bytes::from(frames[3].clone());
+                let frame_message = frames
+                    .pop_back()
+                    .expect("four-frame multipart must contain a payload");
```
jthomson04: Nit: `let Some(frame_message) = frames.pop_back() else { continue };` would make the pump structurally panic-free. As written the `expect` is unreachable (guarded by the frame-count check above), but a panic here would silently kill the pump task since its `JoinHandle` is never awaited. Take it or leave it.

lib/runtime/src/transports/event_plane/zmq_transport.rs
```diff
@@ -315,15 +322,18 @@ impl ZmqSubTransport {
...
+                let frame_bytes = Bytes::from_owner(ZmqMessageOwner(frame_message));
```
jthomson04, line 336: `Bytes::from_owner` only exists since bytes 1.9.0, but the workspace `Cargo.toml` declares `bytes = { version = "1" }`. The lockfile pins 1.11.1 so CI is fine, but the declared floor no longer reflects the true minimum — suggest bumping the workspace constraint to `"1.9"`.

jthomson04 (commented):
## Review

**Verdict: correct and well-executed.** The zero-copy mechanism checks out — I verified the soundness-critical claims against the vendored dependency sources rather than taking them on faith.

### Correctness (verified against dependency sources)

- **Inline (VSM) message safety** — libzmq stores small messages inline in the 64-byte `zmq_msg_t`, so the data pointer moves if the struct moves. The `ZmqMessageOwner` doc comment claims `Bytes::from_owner` takes the pointer only after the owner reaches stable storage. Confirmed in bytes 1.11.1 (`bytes.rs:269-283`): the owner is `Box::into_raw`'d first, and `as_ref()` is called on the heap-pinned value. The internal safety comment there shows this is a deliberate, Miri-validated design choice, so it's unlikely to regress silently — and the new small-message test would catch it if it did.
- **`Send` bound** — `Bytes::from_owner` requires `T: Send`. `zmq::Message` is auto-`Send` because `zmq_msg_t` is `[c_uchar; 64]` in zmq-sys 0.12. Semantically sound too: libzmq permits closing a message from any thread (atomic refcount on heap content, no-op for inline), which is the standard cross-thread ownership-transfer pattern for ZMQ bindings.
- **`Multipart` usage** — tmq 0.5.0's `Multipart` provides `pop_back()` and `Index<usize>`, and `Message` derefs to `[u8]`, so `publisher_id_bytes[..].try_into()` is guarded by the preceding length check. `pop_back()` after the `len() != 4` check correctly yields frame 3.
- **Behavior preservation** — all five reject paths (`continue`) and the receive-error `break` are unchanged; broadcast, warning, and shutdown semantics are intact. Bonus: the topic frame (frame 0), which was copied and never used before, is now not copied at all.
- **Zero-copy end to end** — `Frame::decode` on a `Bytes` input uses `copy_to_bytes` (= `split_to`, a refcounted slice), so the broadcast payload aliases the original zmq message with no copies.

### Tests

- The lifetime test exercises both the inline (18-byte) and heap (64 KiB) message regimes, and the `as_ptr()` equality assertion pins the zero-copy property so a regression to copying would fail the test, not just slow things down.
- The malformed-message test covers all five reject paths, and the readiness-anchor loop is a proper slow-joiner mitigation rather than a bare sleep. The `inproc://` endpoint is valid because the base branch's `shared_zmq_context()` gives pub and sub the same context.

### Suggestions (minor, inline)

1. Bump the `bytes` workspace constraint from `"1"` to `"1.9"` — `Bytes::from_owner` only exists since 1.9.0 (see inline comment).
2. Optional: a panic-free alternative to the `.expect()` on `pop_back()` (see inline comment).

### Risks

- **Memory retention shape** — each broadcast payload now pins the whole zmq message (payload + 5-byte header + owner box) instead of a tight payload copy. Since the payload is essentially the whole message, overhead is a few bytes per in-flight message, and the RSS measurements confirm no regression. Only theoretical concern: a subscriber holding a tiny sub-slice long-term would pin the full message — no current consumer does this.
- **Reliance on a bytes implementation detail** — the post-boxing pointer capture is documented in bytes' internal safety comments, not its public API contract. The added test is the right mitigation.

Meta: the PR labels (`documentation`, `backend::vllm`, `backend::sglang`, `multimodal`, `frontend`) look auto-applied and don't match a single runtime Rust file — worth fixing if labels drive release notes.

ishandhanani (approved):

dmitry-tokarev-nv (approved):
LGTM. claude also approved

PR #10767, opened 2026-06-17
refactor(planner): Clean up planner core, abstract out peripheral logic

sachalmalick:
## Overview:

Hello Hongkuan...

So this PR is the first step in my effort to clean up Planner Core and the logic around how we initiate and validate deployments.

The heart of this new design pattern is that we abstract all of what I call the "peripherals" outside of Planner Core.  This is the trajectory that Planner was already headed with the connector interface, this takes it a step further by

a) Also abstracting out observation (FPM and Traffic metrics) behind a generic interface for a `TrafficMetricsProvider` and `FpmProvider`
b) Letting PlannerCore be a consumer of a new interface called `PlannerEnvironment` which lets Planner not have to own any specific connector logic at all!

PlannerCore, just needs to know a PlannerEnvironment will be provided to me which I can use to
1) Making scaling decisions
2) Know the current deployment state (number of workers and their configurations)
3) Ask for metrics it needs, Fpm or Traffic

So in addition to this being cleaner, this also sets up Planner long term for a wider "total addressable market" / wide adoption.

Planner is a brilliant state of the art prediction and decision engine, but it need not be environment specific at all.  

Someone can go ahead and write your own FpmProvider that will just read events from the cloud if you want! Or a PlannerConnector that makes scaling actions through a generic API call rather than to the Kubernetes management API.   

## Related Issues

> ⚠️ **This section is required.** Choose one path below and delete the other.

**🔗 This PR is linked to an issue:**

- Closes #

**🚫 This PR is NOT linked to an issue**:
- [x] Confirmed — no related issue

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/10767" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Restructured planner architecture to use protocol-based interfaces instead of abstract base classes for connectors.
  * Separated planner initialization concerns: deployment state management, metrics collection, and scaling operations now handled via a dedicated environment abstraction layer.
  * Refactored native planner core to delegate lifecycle and diagnostics to the planner environment.

* **New Features**
  * Added factory functions for constructing connectors, environments, and planners from configuration.
  * Introduced runtime namespace binding for dynamic worker endpoint resolution.

PR #11578, opened 2026-07-12
fix(planner): resolve aggregate workers by component type

### LINKED ISSUE #11491: Planner: aggregate (mode=agg) DGDs not resolved via agg_worker_k8s_name in startup/validation/scaling
## Summary

The SLA planner's decode-role component-name resolution uses `decode_worker_k8s_name` (e.g. vLLM `VllmDecodeWorker`) for `mode=agg`, but aggregate v1beta1 DGDs declare a single worker named `agg_worker_k8s_name` (e.g. `VllmWorker`) with `type: worker`. `get_component_from_type_or_name` requires the explicit fallback name to match a component exactly, so for a real `VllmWorker`/`type: worker` DGD the decode role resolves to nothing — the planner can fail at startup/validation and does not scale the aggregate worker.

This is **pre-existing on `main`**, independent of the power-annotation work in #9683. Before #9683, `agg_worker_k8s_name` had no planner-runtime consumers at all (only the profiler's DGD generator referenced it).

## Affected paths (all pre-existing; not modified by #9683)

- `core/base.py` — `_async_init` → `validate_deployment(decode_component_name=defaults.decode_worker_k8s_name, ...)`
- `connectors/kubernetes.py` — `validate_deployment` / `get_model_name`
- `monitoring/dgd_services.py` — `_resolve_dgd_service` (populates `decode_worker_info.k8s_name`)
- worker counts — `get_actual_worker_counts(decode_component_name=self.decode_worker_info.k8s_name)`
- scaling — `core/adapters.py` `AggPlanner._apply_effects` (`component_name=self.decode_worker_info.k8s_name`)

## Proposed fix

Make decode-role name resolution agg-aware at its source: for `mode == "agg"`, resolve `agg_worker_k8s_name` (falling back to `decode_worker_k8s_name` for backends that don't define an agg name). Applying this in `_resolve_dgd_service` (which populates `decode_worker_info.k8s_name`), plus the `_async_init` `validate_deployment` call and `get_model_name`, fixes startup validation, worker counts, and scaling in one place.

#9683 already applies this pattern locally in the power-annotation sweep (`_run_power_annotation_sweep`) and can serve as a reference.

## Testing

Requires real aggregate integration coverage (built `dynamo._core` plus a cluster / dev pod), since the resolution feeds live startup, validation, and scaling.

## Context

Surfaced during review of #9683 (Phase-1 power infrastructure); deferred there to keep that PR scoped to the power feature.

waynehacking8:
## Summary

- Resolve aggregate planner workers by their DGD component type instead of backend-specific names.
- Treat the sole generic `type: worker` component as the planner's decode component, regardless of its name.
- Keep the existing explicit-name fallback when multiple generic workers make type-only selection ambiguous.

This fixes #11491 and lets #9848 rename vLLM workers without carrying planner-side aliases for every historical name.

## Validation

- Current-main A/B through `KubernetesConnector.validate_deployment`: an aggregate DGD with an arbitrary worker name fails before this change and passes after it.
- `test_kubernetes_connector.py`: 75 passed.
- Ruff check and format check passed for both changed files.

Live-cluster startup and scaling were not run because they require a real aggregate deployment.

### DIFF
--- a/components/src/dynamo/planner/monitoring/dgd_services.py
+++ b/components/src/dynamo/planner/monitoring/dgd_services.py
@@ -210,6 +210,15 @@ def get_component_from_type_or_name(
         component_names = [name for name, _ in matching_components]
         raise DuplicateSubComponentError(component_type.value, component_names)
 
+    if not matching_components and component_type == SubComponentType.DECODE:
+        generic_workers = [
+            (name, component)
+            for name, component in components.items()
+            if get_component_type(component) == V1BETA1_GENERIC_WORKER_COMPONENT_TYPE
+        ]
+        if len(generic_workers) == 1:
+            matching_components = generic_workers
+
     if not matching_components and component_name in components:
         component = components[component_name]
         if not _can_use_explicit_component_name(component, component_type):
--- a/components/src/dynamo/planner/tests/unit/test_kubernetes_connector.py
+++ b/components/src/dynamo/planner/tests/unit/test_kubernetes_connector.py
@@ -378,6 +378,55 @@ def test_get_service_name_from_v1beta_worker_type_by_name(kubernetes_connector):
     assert service.number_replicas() == 2
 
 
+def test_get_service_name_from_unique_v1beta_worker_type_for_decode(
+    kubernetes_connector,
+):
+    deployment = _deployment(_component("arbitrary-name", "worker", replicas=2))
+
+    service = get_component_from_type_or_name(deployment, SubComponentType.DECODE)
+
+    assert service.name == "arbitrary-name"
+    assert service.number_replicas() == 2
+
+
+def test_get_service_name_from_multiple_v1beta_workers_by_name(
+    kubernetes_connector,
+):
+    deployment = _deployment(
+        _component("prefill-name", "worker"),
+        _component("decode-name", "worker"),
+    )
+
+    with pytest.raises(SubComponentNotFoundError):
+        get_component_from_type_or_name(deployment, SubComponentType.DECODE)
+
+    service = get_component_from_type_or_name(
+        deployment, SubComponentType.DECODE, "decode-name"
+    )
+    assert service.name == "decode-name"
+
+
+@pytest.mark.asyncio
+async def test_validate_deployment_agg_worker_by_type(
+    kubernetes_connector, mock_kube_api
+):
+    mock_kube_api.get_graph_deployment.return_value = _deployment(
+        _component("Frontend", "frontend", replicas=1),
+        _component(
+            "arbitrary-name",
+            "worker",
+            replicas=2,
+            args=["--model", "Qwen/Qwen3-8B"],
+        ),
+    )
+
+    await kubernetes_connector.validate_deployment(
+        decode_component_name="stale-default-name",
+        require_prefill=False,
+        require_decode=True,
+    )
+
+
 def test_get_service_name_from_sub_component_type_not_found(kubernetes_connector):
     deployment = _deployment(_component("test-component-decode", "decode", replicas=3))
     with pytest.raises(SubComponentNotFoundError) as exc_info:


waynehacking8:
Applied in 77f8bba. GPU-count resolution now gets the same explicit names (dynamo-review-agent's catch), and resolve_worker_info forwards decode_component_name to the model-name fallback (Devin's first point -- in practice that path only fires when the DGD args carry no model at all, since get_worker_info's arg-parsing fallback already used the name, but forwarding it makes the residual case fail with the truthful error). Devin's second point actually cut the other way once I checked the recipes: the repo's own vllm agg examples are split between VllmWorker (agg_failover, agg_gms) and VllmDecodeWorker (agg, agg_router, agg_kvbm), so a single static name would regress one shape -- resolution now takes ordered candidates (agg name, then decode name), which also covers the sglang/trtllm agg recipes via their decode names.

waynehacking8:
Rebased over #10767, which moved deployment bootstrapping out of core/base.py into the environment facade -- the candidate derivation now lives in PlannerEnvironmentImpl._decode_component_name (environment/base.py) and the connector-side threading is unchanged. Re-ran the planner unit suites against current main (identical pass/fail set to main in my environment) and updated the PR body to match the new structure.

Jont828:
@waynehacking8 It seems like your PR overlaps with mine #9848. I'm shortening the vllm renames but have a problem with making it backwards compatible. Maybe we can try to collab on getting both of these in together?

waynehacking8:
Happy to collab -- I think the two compose cleanly. This PR makes the planner's decode-role resolution try ordered name candidates read from the backend defaults, so your rename flows through automatically once the defaults say prefill/decode/worker, and the candidate list is also a natural home for the legacy fallback sttts asked about (try "worker" then "VllmWorker", so pre-rename DGDs keep resolving on the planner side -- it does not help the operator/override side of your change). If yours lands first I'll rebase mine over it.

Jont828:
I think my PR is blocked on having backwards compatibility so yours would probably be a better candidate to merge first. I think we shouldn't rely on hard coded name lookups and instead have the selector choose based on the component type. That way, if we decide to change the name again, we'll need to maintain `<new name>`, `worker`, and `VllmWorker` too. IIRC the name lookup is a fallback behavior, so using it means that something fell through on the selection logic. Ideally the user should be able to name their prefill/decode/worker anything they want. The component type `prefill` or `decode` is a better indicator imo.

Do you want to take this? Or could I PR into your branch or something?

Jont828:
@waynehacking8 Also are you on the [CNCF Slack](https://www.cncf.io/membership-faq/#how-do-i-join-cncfs-slack)? Could ping you there too in case we need to chat/brainstorm some more.

waynehacking8:
@Jont828 Yes, I took this. I replaced the component name candidate plumbing with a type based fallback: a single generic `type: worker` now resolves as decode, while multiple generic workers still use the existing explicit name disambiguation. I rebased and force pushed the two file version; the connector suite passes 75/75, so #9848 no longer needs historical planner aliases.

waynehacking8:
@kaim-eng The planner fix is now reduced to a type-based fallback for the single generic worker, with 75/75 connector tests passing; this resolves #11491 without coupling the planner to backend-specific names. When you have time, could you review the rebuilt head?

Jont828:
@waynehacking8 Thanks for doing this! Btw, @sttts left a [comment](https://github.com/ai-dynamo/dynamo/pull/9848#discussion_r3603294033) on my PR for renaming "VllmPrefillWorker" -> "prefill".

> docs apparently still say VllmPrefillWorker. This matters for DGDR overrides. We should better make overrides silently support some normalization?

Does this component based matching work for overrides as well? Meaning if someone forgot to update the `VllmPrefillWorker` name, would it be silently dropped?

waynehacking8:
No. This PR only changes planner lookup after the DGD exists. DGDR overrides are merged earlier by exact `spec.services` key in `apply_dgd_overrides`. If the generated service is `prefill`, an override keyed by `VllmPrefillWorker` is skipped and a warning is logged. So it is not silent, but it is not normalized either. #9848 should update the override names or add compatibility in the override merge path; #11578 does not cover that.

sttts (approved):

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/commit/4229039109cc0159cac398eb445058f21a063124

hhzhang16 (approved):

PR #11587, opened 2026-07-13
ci: Split Dynamo CI port ranges by workload

VincyZhang:
## Overview:

Split the dynamic frontend port off from the serve/system pool so host-network CI can isolate frontend, worker/system, KV event, router, NIXL, and FPM allocations more cleanly.

## Details:

- Introduced a dedicated `DynamoPortRange.FRONTEND` base for the dynamic serve frontend.
- Kept serve worker/system ports and KV event ports on the serve range.
- Kept router, NIXL, and FPM on their own disjoint bases.
- Increased the allocator random start offset to 500 and widened the gaps between ranges to leave headroom for sequential scans.

- `tests/utils/constants.py` for the updated port base layout.
- `tests/conftest.py` for the serve fixture allocation changes.
- `tests/utils/port_utils.py` for the allocator offset change.

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11587" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test reliability by allocating service, router, metrics, and side-channel ports dynamically from dedicated, non-overlapping ranges.
  * Reduced port collisions during concurrent test execution and containerized runs.
  * Expanded port allocation probing to better handle busy ports.

* **Tests**
  * Updated fault-tolerance, migration, router, and benchmarking scenarios to use the shared port allocation strategy.

### DIFF
--- a/tests/fault_tolerance/migration/test_sglang.py
+++ b/tests/fault_tolerance/migration/test_sglang.py
@@ -15,7 +15,7 @@ import shutil
 
 import pytest
 
-from tests.utils.constants import FAULT_TOLERANCE_MODEL_NAME
+from tests.utils.constants import DynamoPortRange, FAULT_TOLERANCE_MODEL_NAME
 from tests.utils.managed_process import ManagedProcess
 from tests.utils.payloads import check_models_api
 from tests.utils.port_utils import allocate_port, deallocate_port
@@ -95,7 +95,7 @@ class DynamoWorkerProcess(ManagedProcess):
         disagg_mode: str | None = None,
     ):
         self.worker_id = worker_id
-        self.system_port = allocate_port(9100)
+        self.system_port = allocate_port(DynamoPortRange.SERVE.value)
         self.bootstrap_port: int | None = None
         self.prefill_port: int | None = None
         self.disagg_mode = disagg_mode
--- a/tests/fault_tolerance/migration/test_sglang.py
+++ b/tests/fault_tolerance/migration/test_sglang.py
@@ -124,7 +124,7 @@ class DynamoWorkerProcess(ManagedProcess):
             command.append("--skip-tokenizer-init")
         else:
             # Disaggregated
-            self.bootstrap_port = allocate_port(12340)
+            self.bootstrap_port = allocate_port(DynamoPortRange.BOOTSTRAP.value)
             request.addfinalizer(lambda port=self.bootstrap_port: deallocate_port(port))
             command.extend(
                 [
@@ -139,7 +139,7 @@ class DynamoWorkerProcess(ManagedProcess):
                 ]
             )
             if disagg_mode == "prefill":
-                self.prefill_port = allocate_port(20000)
+                self.prefill_port = allocate_port(DynamoPortRange.PREFILL.value)
                 request.addfinalizer(
                     lambda port=self.prefill_port: deallocate_port(port)
                 )
--- a/tests/fault_tolerance/migration/test_sglang.py
+++ b/tests/fault_tolerance/migration/test_sglang.py
@@ -15,7 +15,7 @@ import shutil
 
 import pytest
 
-from tests.utils.constants import DynamoPortRange, FAULT_TOLERANCE_MODEL_NAME
+from tests.utils.constants import FAULT_TOLERANCE_MODEL_NAME, DynamoPortRange
 from tests.utils.managed_process import ManagedProcess
 from tests.utils.payloads import check_models_api
 from tests.utils.port_utils import allocate_port, deallocate_port
--- a/tests/fault_tolerance/migration/test_sglang.py
+++ b/tests/fault_tolerance/migration/test_sglang.py
@@ -96,6 +96,7 @@ class DynamoWorkerProcess(ManagedProcess):
     ):
         self.worker_id = worker_id
         self.system_port = allocate_port(DynamoPortRange.SERVE.value)
+        request.addfinalizer(lambda port=self.system_port: deallocate_port(port))
         self.bootstrap_port: int | None = None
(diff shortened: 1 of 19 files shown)


tests/vllm_self_benchmark/test_self_benchmark_gpu.py
```diff
@@ -199,7 +199,7 @@ def __init__(
         self.frontend_port = frontend_port
 
         # Allocate a per-worker system port like the cancellation test does.
-        self.system_port = allocate_port(9100)
+        self.system_port = allocate_port(DynamoPortRange.SERVE.value)
```
dmitry-tokarev-nv, line 204: Port is allocated here but never freed — no `request.addfinalizer` is registered for either `system_port` or `fpm_port` in this class, unlike the analogous ports in the cancellation test classes this PR already fixed.

Suggestion:
```python
self.system_port = allocate_port(DynamoPortRange.SERVE.value)
request.addfinalizer(lambda port=self.system_port: deallocate_port(port))
```
(and the same for `fpm_port` below)

tests/fault_tolerance/migration/test_sglang.py
```diff
@@ -124,6 +125,7 @@ def __init__(
         else:
             # Disaggregated
```
dmitry-tokarev-nv, line 126: The finalizer is now wired up (good), but the base `12340` is a magic number not in `DynamoPortRange`. Without a named constant it's invisible to the range-isolation scheme introduced by this PR and could still collide with other allocations.

Suggestion: add a `BOOTSTRAP = 20000` (or similar disjoint value) entry to `DynamoPortRange` and use it here:
```python
self.bootstrap_port = allocate_port(DynamoPortRange.BOOTSTRAP.value)
```

tests/fault_tolerance/migration/test_sglang.py
```diff
@@ -138,6 +140,9 @@ def __init__(
             )
             if disagg_mode == "prefill":
                 self.prefill_port = allocate_port(20000)
```
dmitry-tokarev-nv: Same issue: `20000` is not in `DynamoPortRange`, so this allocation sits outside the documented isolation scheme. Consider adding a `PREFILL` entry and using it here:
```python
self.prefill_port = allocate_port(DynamoPortRange.PREFILL.value)
```

dmitry-tokarev-nv (changes requested):
see inline comments with suggestions

tests/vllm_self_benchmark/test_self_benchmark_gpu.py
VincyZhang: Good catch! Fixed in latest commit

tests/fault_tolerance/migration/test_sglang.py
VincyZhang: Added `DynamoPortRange.PREFILL` and `DynamoPortRange.BOOTSTRAP`. 
Changed all hard-coded port assignments that had been overlooked in the repository to the corresponding DynamoPortRange.

tests/fault_tolerance/migration/test_sglang.py
VincyZhang: Added `DynamoPortRange.PREFILL` and `DynamoPortRange.BOOTSTRAP`. 
Changed all hard-coded port assignments that had been overlooked in the repository to the corresponding DynamoPortRange.

tests/utils/constants.py
```diff
@@ -32,6 +32,25 @@ class DefaultPort(IntEnum):
...
+    NIXL, and FPM workloads.
+    """
+
+    FRONTEND = 22500
+    SERVE = 24000
+    BOOTSTRAP = 24600
+    PREFILL = 25200
+    NIXL = 25800
+    ROUTER = 27000
+    FPM = 28500
```
dmitry-tokarev-nv, line 51: **Retracting this suggestion — it's a regression, good catch.** Widening every gap to a uniform 1000 actually *shortens* `NIXL→ROUTER` (1200→1000) and `ROUTER→FPM` (1500→1000), and the second one breaks a real case.

With this PR's new `_START_PORT_RANDOM_OFFSET_MAX = 500`, worst-case reach of an allocation off a base is:
- `allocate_ports` (max_retries=100): `base + 500 + 99 = base + 599` → min safe gap **600**
- `allocate_contiguous_ports` (max_retries=500, block size `k`): `base + 500 + 499 + (k-1) = base + 998 + k` → for `k=2`, `base + 1000` → min safe gap **1001**

`ROUTER` is used with `allocate_contiguous_ports` at `block_size = data_parallel_size`, and there's a live `data_parallel_size=2` test in both the sglang and vllm router suites, so `k=2` is exercised:
- Original `ROUTER=27000 → FPM=28500` (gap 1500): worst block at dp=2 tops out at **28000**, 500 below FPM → **safe**.
- Suggested `ROUTER=28000 → FPM=29000` (gap 1000): worst block tops out at **29000 = FPM base exactly** → collision.

And my original premise was wrong too: the 600-port gaps on `SERVE`/`BOOTSTRAP`/`PREFILL`/`NIXL` are **not** unsafe — those bases are only ever allocated non-contiguously (reach `base+599`), so 600 is exactly the right minimum. The values already in this PR are correctly tuned; **keep them as-is.**

(Optional, non-blocking: a short comment on the enum noting which bases are used with `allocate_contiguous_ports` — currently only `ROUTER` and `FPM` — would stop a future edit from "simplifying" the gaps and reintroducing this collision.) Sorry for the noise.

dmitry-tokarev-nv (approved):

PR #11532, opened 2026-07-10
feat(vllm): complete exact MM routing for generate

furionw:
_Based on #11588._

## Why

The frontend can project renderer-provided image identity into KV-routing tokens, but vLLM workers must reconstruct the native Generate request and emit the same identity in KV events. Without that parity, request-side hashes and event-side hashes diverge, so repeated multimodal requests remain cold even when generation succeeds. This completes the worker half for aggregated and P/D serving while preserving existing chat, completions, custom-encoder, and conditional-disaggregation behavior. Cache salts are tagged only at the vLLM boundary, keeping caller-visible namespaces unchanged while making event hashing exact.

## What Change

- Execute native Generate envelopes in aggregated and P/D vLLM workers.
- Normalize opaque multimodal identifiers and cache salts in KV events.
- Preserve native sampling, logprob, expert, and transfer metadata.

```mermaid
classDiagram
    DecodeWorkerHandler --|> BaseWorkerHandler
    PrefillWorkerHandler --|> BaseWorkerHandler
    DecodeWorkerHandler ..> GenerateEnvelopeAdapter : rebuilds native requests
    PrefillWorkerHandler ..> GenerateEnvelopeAdapter : rebuilds native requests
    KvEventPublisher --> RawKvEventVisitor : streams vLLM cache events
    RawKvEventVisitor ..> BlockExtraInfo : restores MM and namespace identity

    note for GenerateEnvelopeAdapter "Reconstructs prompts, sampling, salts, and metadata"
    note for KvEventPublisher "Carries the resolved image token into normalization"
    note for RawKvEventVisitor "Separates LoRA, cache namespace, and image identifiers"
```

## Test Plan

- `cargo test -p dynamo-llm` (84), `cargo test -p dynamo-kv-router` (48), and backend Python tests (143).
- Focused vLLM Generate tests (14), clippy, fmt, pre-commit, and sidecar all-features check.
- Local two-worker Qwen3.5-2B E2E with `qiwa-dev-vllm-x86-08-18:latest`: salted and unsalted suites each sent 10 successful Generate/derender requests; cold images split across workers at `0/4`, and all four warm repeats returned to the matching worker at `3/4` overlap.

PR #11588, opened 2026-07-13
feat(frontend): add exact MM routing for generate

furionw:
## Why

`/inference/v1/generate` receives pretokenized multimodal prompts, so normal frontend preprocessing cannot attach image identity for KV routing. Token-only hashes make different images with identical placeholder layouts look equivalent and reduce cache-aware routing accuracy. This derives routing-only identity from valid renderer metadata while leaving execution tokens unchanged; malformed or ambiguous metadata safely falls back to token-only routing. The selected Generate endpoint, block size, and LoRA identity come from one WorkerSet so request hashing cannot mix worker metadata. #11532 completes the paired worker normalization and end-to-end validation.

## What Change

- Project renderer multimodal identities into routing-only Generate tokens.
- Select Generate engine, block size, and LoRA identity atomically.
- Reject ambiguous placeholder layouts without changing execution payloads.

```mermaid
classDiagram
    GenerateHandler --> ModelManager : requests Generate selection
    ModelManager --> Model : delegates selection
    Model --> WorkerSet : selects ready Generate endpoint
    WorkerSet ..> GenerateEngineSelection : supplies routing metadata
    GenerateHandler ..> PreprocessedRequest : builds routing projection

    note for GenerateHandler "Validates renderer metadata and projects image identity"
    note for GenerateEngineSelection "Keeps engine, block size, and LoRA identity atomic"
    note for PreprocessedRequest "Separates execution tokens from routing-only tokens"
```

## Test Plan

- `cargo test -p dynamo-llm` (84), clippy with warnings denied, fmt, and pre-commit.
- Malformed, sparse, LoRA, overlapping-placeholder, and stray-image-token fallback coverage.
- Stacked #11532 local two-worker Qwen3.5-2B E2E: salted and unsalted suites each sent 10 successful Generate/derender requests; cold images split across workers at `0/4`, and all four warm repeats returned to the matching worker at `3/4` overlap.


biswapanda (approved):
Rebased onto current main; verified exact-MM routing capability isolation, malformed metadata fallback, targeted Generate tests, clippy, formatting, and DCO.

biswapanda (approved):

lib/llm/src/http/service/generate.rs
```diff
@@ -204,6 +205,223 @@ impl<'a> VllmTitoEnvelope<'a> {
...
+        .as_object()
+        .ok_or("features must be a JSON object")?;
+    let mm_hashes = features
+        .get("mm_hashes")
+        .and_then(serde_json::Value::as_object)
+        .ok_or("features.mm_hashes must be a JSON object")?;
+    let mm_placeholders = features
+        .get("mm_placeholders")
+        .and_then(serde_json::Value::as_object)
+        .ok_or("features.mm_placeholders must be a JSON object")?;
```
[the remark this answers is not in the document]
biswapanda: Fixed in 5df0396324: a features object without mm_hashes now returns token-only routing without an error/debug fallback. Added coverage for unrelated features and malformed present mm_hashes.

biswapanda:
Rebase/repair summary:

- rebased the two intended commits onto current `main`;
- isolated `vllm_exact_mm_routing` in model-card compatibility so mixed-capability workers cannot share a first-card `WorkerSet`;
- fixed token-only fallback when `features` is present without `mm_hashes`;
- added both worker-arrival-order and malformed-feature regressions.

Validation: `cargo test -p dynamo-llm --lib generate` (71 passed), focused watcher test, `cargo clippy -p dynamo-llm --lib --tests -- -D warnings`, formatting, DCO. Auto-merge is enabled after required CI and code-owner approval.

biswapanda (approved):
Cross-PR capability and WorkerSet/LoRA contracts verified; all review threads are addressed and resolved.

lib/kv-router/src/protocols.rs
```diff
@@ -76,6 +76,20 @@ pub fn pad_value_for_mm_hash(mm_hash: u64) -> u32 {
...
+/// Preserve vLLM's canonical 64-character hex-digest mapping for compatibility,
+/// and hash shorter opaque identifiers emitted by external renderers with XXH3.
+pub fn hash_mm_identifier(identifier: &str) -> Option<u64> {
+    if identifier.is_empty() {
+        return None;
+    }
+    if identifier.len() == 64 && identifier.chars().all(|c| c.is_ascii_hexdigit()) {
+        return u64::from_str_radix(&identifier[..16], 16).ok();
+    }
+    Some(xxh3::xxh3_64(identifier.as_bytes()))
```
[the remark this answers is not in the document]
biswapanda: This contract is completed by the stacked backend PR #11532 before the capability is advertised. #11588 gates this path on `vllm_exact_mm_routing`; current main workers do not advertise it, so opaque routing remains inert after this PR alone. The rebased #11532 worker delta updates the real ZMQ `extra_keys` path to hash offset-tagged opaque identifiers with the same `hash_mm_identifier` function and includes an opaque-identifier regression. Keeping the event-side implementation in the owning backend PR avoids enabling an incomplete worker contract.

lib/llm/src/discovery/model.rs
```diff
@@ -713,9 +723,27 @@ impl Model {
...
+                    supports_exact_mm_routing: ws
+                        .card()
+                        .runtime_config
+                        .runtime_data
+                        .get(VLLM_EXACT_MM_ROUTING_CAPABILITY)
+                        .and_then(serde_json::Value::as_bool)
+                        .unwrap_or(false),
+                })
+        })
+        .ok_or_else(|| self.engine_error(self.has_generate_engine()))
```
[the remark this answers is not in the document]
biswapanda: Confirmed intentional: Generate has no per-request LoRA selector, so adapter identity comes from the atomically selected adapter-specific WorkerSet/model card. Routing hints and the returned engine therefore cannot disagree.

biswapanda (approved):
Cross-PR capability and WorkerSet/LoRA contracts verified; all review threads are addressed and resolved.

biswapanda (approved):
Final head includes the linear multimodal-range sweep and deterministic complexity regression; 72 Generate tests, clippy, formatting, and DCO are green.

lib/kv-router/src/protocols.rs
```diff
@@ -76,6 +76,20 @@ pub fn pad_value_for_mm_hash(mm_hash: u64) -> u32 {
...
+}
```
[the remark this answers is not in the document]
biswapanda: Confirmed as an intentionally stacked contract: #11588 does not advertise `vllm_exact_mm_routing`, so this path remains inert on current workers. The immediately following backend PR #11532 adds `extra_keys_to_block_mm_infos_with_lora`, hashes offset-tagged opaque identifiers through `hash_mm_identifier`, and tests that real event parser path before enabling the capability. No unmatched opaque routing is exposed between merges.

biswapanda (approved):
follow up stacked PR #11532

lib/llm/src/http/service/generate.rs
```diff
@@ -212,12 +468,36 @@ fn preprocessed_from_generate(
...
+                    target: "mm_routing",
+                    reason,
+                    "invalid /generate multimodal routing metadata; using token-only routing"
+                );
+                None
+            }
+        }
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
furionw: Confirmed. `vllm_exact_mm_routing` now explicitly guarantees that the worker's KV-event normalizer has a resolved image-token ID and uses the same multimodal identifier mapping as the frontend. In stacked backend PR #11532, the publisher is configured from `resolve_image_token_id(...)`, and the capability is advertised only when that same resolver succeeds; workers without it stay on token-only routing. I also updated the parity coverage to exercise the real vLLM `extra_keys` event parser.

lib/llm/src/discovery/model.rs
```diff
@@ -671,6 +681,33 @@ impl Model {
...
+                    kv_cache_block_size: ws.card().kv_cache_block_size,
+                    lora_name: ws.card().lora.as_ref().map(|lora| lora.name.clone()),
+                    supports_exact_mm_routing: ws
+                        .card()
+                        .runtime_config
+                        .runtime_data
+                        .get(VLLM_EXACT_MM_ROUTING_CAPABILITY)
+                        .and_then(serde_json::Value::as_bool)
+                        .unwrap_or(false),
+                })
```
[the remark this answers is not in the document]
furionw: Fixed in da35cabaaf. `ModelRuntimeConfig::supports_runtime_capability` now accepts both JSON `true` and Dynamo's canonical truthy strings. `WorkerSet` delegates to that helper, and Generate selection plus the model-card checksum use the same interpretation. Added regressions for boolean/string true and false values, selection, and checksum equivalence/isolation.

lib/kv-router/src/zmq_wire/extra_keys.rs
```diff
@@ -68,11 +70,9 @@ pub fn extra_keys_to_block_mm_infos(
...
                 .iter()
                 .filter_map(|key| match key {
-                    ExtraKeyItem::Hash(hash)
-                    | ExtraKeyItem::HashWithSignedOffset((hash, _))
-                    | ExtraKeyItem::HashWithUnsignedOffset((hash, _)) => {
-                        parse_mm_hash_from_extra_key(hash)
-                    }
+                    ExtraKeyItem::Hash(hash) => parse_mm_hash_from_extra_key(hash),
+                    ExtraKeyItem::HashWithSignedOffset((hash, _))
+                    | ExtraKeyItem::HashWithUnsignedOffset((hash, _)) => hash_mm_identifier(hash),
```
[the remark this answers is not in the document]
furionw: Fixed in da35cabaaf by restoring canonical-only parsing for offset-bearing `extra_keys` in this parent PR, so #11588 no longer changes worker event hashes before the capability is advertised. Canonical request/event parity still exercises the real parser, while opaque frontend projection remains covered separately. Opaque worker-event normalization stays in stacked #11532 alongside publication of `vllm_exact_mm_routing`; the remaining old-frontend/new-worker rollout effect is bounded cache misses, never incorrect reuse.

lib/llm/src/http/service/generate.rs
```diff
@@ -205,6 +206,268 @@ impl<'a> VllmTitoEnvelope<'a> {
...
+                .get(run_index)
+                .or_else(|| worker_objects.last())
+                .copied();
+            if worker_hash != Some(expected_hash) {
+                return Err(
+                    "sparse multimodal layout cannot be normalized exactly by worker events",
+                );
+            }
+        }
+    }
```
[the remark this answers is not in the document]
furionw: The current head already guards this exact case. `generate_mm_routing_info` infers the image token from declared embed positions, compares its total occurrence count with the declared count, and falls back to token-only routing when any matching token lies outside the placeholder ranges. The `undeclared_image_token_disables_exact_routing` regression covers a same-ID token before the declared range. No additional change is needed for this finding.

lib/llm/src/http/service/generate.rs
```diff
@@ -240,9 +557,11 @@ fn preprocessed_from_generate(
             ..Default::default()
         })
         .output_options(Default::default())
+        .mm_routing_info(mm_routing_info)
         .routing(Some(crate::protocols::common::preprocessor::RoutingHints {
             dp_rank: data_parallel_rank,
             expected_output_tokens: max_tokens,
+            lora_name,
             cache_namespace: cache_salt,
             // `priority_jump` is a boost-only scheduler input. Preserve penalties
```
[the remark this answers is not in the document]
furionw: Verified as a pre-existing cache-salt contract issue, outside this exact-MM change. The merge base already copies `cache_salt` into both `RoutingHints.cache_namespace` and `extra_args.vllm_tito`, and this PR does not change that path; the vLLM worker helper currently tags only `nvext` sources. Exact-MM routing therefore neither introduces nor worsens the mismatch. I am keeping this PR scoped to multimodal routing.

lib/llm/src/http/service/generate.rs
```diff
@@ -313,11 +632,17 @@ async fn handler_generate(
...
+        kv_cache_block_size,
+        lora_name,
+    } = match state
         .manager()
-        .get_generate_engine_for_capability(&model, VLLM_INFERENCE_V1_GENERATE_CAPABILITY)
-    {
-        Ok(engine) => engine,
+        .get_generate_engine_for_capability_with_routing(
+            &model,
+            VLLM_INFERENCE_V1_GENERATE_CAPABILITY,
```
jthomson04, line 640: [P1] Gate this projection on an exact-MM capability

`vllm_inference_v1_generate` only says that the endpoint exists. It does not guarantee that the KV-event publisher resolved the image token or applies the same MM identifier mapping. This reintroduces the condition from [the earlier resolved thread](https://github.com/ai-dynamo/dynamo/pull/11588#discussion_r3714542575): ordinary Generate workers can enter exact projection and silently miss MM event hashes. Restore a dedicated `vllm_exact_mm_routing` gate.

lib/llm/src/http/service/generate.rs
```diff
@@ -213,12 +511,31 @@ fn preprocessed_from_generate(
...
 ) -> anyhow::Result<PreprocessedRequest> {
     let sampling = &request.sampling_params;
     let max_tokens = sampling.max_tokens();
     let min_tokens = sampling.min_tokens();
     let ignore_eos = sampling.ignore_eos();
     let routing_priority = dynamo_routing_priority(request.priority);
+    // With vLLM's default `enable_tower_connector_lora=false`, the vision tower
+    // and connector stay on base weights, so MM identifiers are adapter-invariant.
+    // `lora_name` separately salts the LM KV hashes below, allowing exact MM+LoRA
+    // routing. If tower/connector LoRA is enabled, vLLM scopes MM identifiers by
```
jthomson04: [P2] Disable exact routing for tower/connector LoRA

This describes a known hash mismatch, but the code still constructs `mm_routing_info`. In this mode vLLM scopes the worker-side MM identifier by adapter, while the frontend hashes the unscoped renderer identifier. Salting the block with `lora_name` does not repair the pad-token mismatch. Add a capability/config guarantee or fall back to token-only routing.

lib/llm/src/http/service/generate.rs
```diff
@@ -205,6 +206,303 @@ impl<'a> VllmTitoEnvelope<'a> {
...
+            .token_ids
+            .iter()
+            .filter(|token_id| **token_id == image_token_id)
+            .count()
+            != declared_image_tokens
+    {
+        return Err("image tokens must be covered by multimodal placeholder ranges");
+    }
+
+    // vLLM's current event normalizer associates MM objects with contiguous
```
jthomson04: [P2] Share the worker MM normalization

This is a second implementation of the worker run-order and clamping rules. The earlier sparse-mask and stray-token defects show how easily these copies drift. Move the normalization into `dynamo-kv-router` and use the same helper for request and event hashes.

lib/llm/src/http/service/generate.rs
```diff
@@ -205,6 +206,303 @@ impl<'a> VllmTitoEnvelope<'a> {
...
 
+type MmPlaceholderRange = (usize, usize, u64, Option<Vec<bool>>);
+
+#[inline]
+fn intersecting_mm_ranges<'a>(
+    ranges: &'a [MmPlaceholderRange],
+    block_start: usize,
+    block_end: usize,
+    first_intersecting: &mut usize,
+    mut on_examined: impl FnMut(),
```
jthomson04: [P3] Remove the test hook from this production helper

`on_examined` exists only for one unit test; the real call passes `|| {}` and invokes it in both scan loops. Remove the callback and keep a behavioral sparse-range test.

lib/llm/src/discovery.rs
```diff
@@ -3,7 +3,7 @@
 
 mod model;
 pub(crate) mod readiness;
-pub use model::Model;
+pub use model::{GenerateEngineSelection, Model};
```
jthomson04: [P3] Keep this selection type internal

`GenerateEngineSelection` is only used by the frontend handler, but this re-export makes its fields part of the crate public API. Make the type and selection methods `pub(crate)` and drop this public re-export.

lib/llm/src/http/service/generate.rs
```diff
@@ -901,6 +1228,428 @@ mod tests {
         assert!(envelope.get("token_ids").is_none());
     }
 
+    #[test]
```
jthomson04, line 1250: [P3] Reduce these parser tests

This adds 15 tests and more than 400 lines for one private parser, with many cases differing only in malformed input shape. Keep the three important behavior tests: valid parity, safe fallback, and a sparse or ambiguous layout.

jthomson04 (commented):
Six inline findings.

lib/llm/src/http/service/generate.rs
furionw: Addressed in 3db4a59b6c without restoring the broad `vllm_exact_mm_routing` preference flag. Exact projection now requires the selected WorkerSet to publish its resolved `vllm_routing_image_token_id`, and the frontend sends only per-request marker-form `dynamo_mm_routing_hashes`. This parent deliberately has no producer for the token metadata, so it remains dormant alone; stacked backend #11532 must consume the marked hashes and publish the resolved token in the same change. Missing or mismatched tokens fall back to token-only routing, and WorkerSets are checksum-isolated by token ID.

lib/llm/src/http/service/generate.rs
furionw: Addressed in 3db4a59b6c. The vLLM worker publishes `enable_tower_connector_lora` as targeted runtime metadata, selection reads it atomically with the engine and routing metadata, and adapter requests fall back to token-only MM routing when it is enabled. Missing and false remain checksum-equivalent; true is isolated.

lib/llm/src/http/service/generate.rs
furionw: Addressed in 3db4a59b6c. `normalize_mm_token_runs` is now a pure helper in `dynamo-kv-router`, and both request projection and KV-event conversion use it. The helper has direct run-order and clamping coverage.

lib/llm/src/http/service/generate.rs
furionw: Addressed in 3db4a59b6c. The production callback was removed from `intersecting_mm_ranges`; sparse-range behavior remains covered through the parser behavior tests.

lib/llm/src/discovery.rs
furionw: Addressed in 3db4a59b6c. `GenerateEngineSelection`, its fields, and both routing-aware selection methods are now crate-internal, and the public re-export was removed.

lib/llm/src/http/service/generate.rs
furionw: Addressed in 3db4a59b6c. Multimodal parser coverage is consolidated into exactly three behavior tests: valid request/event parity with payload preservation, sparse or ambiguous layouts, and invalid or unsupported safe fallback.

lib/llm/src/http/service/generate.rs
```diff
@@ -205,6 +206,303 @@ impl<'a> VllmTitoEnvelope<'a> {
...
+    for pair in ranges.windows(2) {
+        let (_, previous_end, previous_hash, _) = &pair[0];
+        let (next_offset, _, next_hash, _) = &pair[1];
+        if previous_end > next_offset {
+            return Err("multimodal placeholder ranges must not overlap");
+        }
+        if previous_end == next_offset && previous_hash != next_hash {
+            return Err("adjacent multimodal placeholders must share an identifier");
+        }
+    }
```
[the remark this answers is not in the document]
furionw: Addressed in 3db4a59b6c. Placeholder input order is now preserved; out-of-offset-order metadata is rejected and falls back safely instead of being sorted into a potentially different worker object order. The fallback case is covered in the consolidated invalid-metadata test.

components/src/dynamo/vllm/main.py
```diff
@@ -744,8 +744,15 @@ async def register_vllm_model(
         runtime_config, config.engine_args, worker_type, dp_range
     )
     runtime_config.context_length = vllm_config.model_config.max_model_len
+    tower_connector_lora_enabled = bool(
+        vllm_config.lora_config and vllm_config.lora_config.enable_tower_connector_lora
```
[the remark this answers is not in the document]
furionw: Addressed in fef092fdf5. This access was introduced by the PR, so model registration now uses `getattr(..., False)` and preserves the expected default-disabled behavior when a supported vLLM LoRAConfig does not expose the field. The focused pre-commit suite passes.

lib/llm/src/http/service/generate.rs
```diff
@@ -205,6 +205,295 @@ impl<'a> VllmTitoEnvelope<'a> {
...
+            block_tokens,
+            image_token_id,
+            &worker_objects,
+        )
+        .map(|(tokens, _)| tokens)
+        .ok_or("multimodal block must contain a routing hash")?;
+        if normalized_tokens != expected_tokens {
+            return Err("sparse multimodal layout cannot be normalized exactly by worker events");
+        }
+        routing_token_ids.extend(normalized_tokens);
```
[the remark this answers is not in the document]
furionw: No code change is needed here because the pinned vLLM worker establishes this ordering before emitting KV events. `InputProcessor` calls `argsort_mm_positions(decoder_mm_positions)` and constructs `request.mm_features` in ascending `mm_position.offset`; `_gen_mm_extra_hash_keys` then walks that ordered list and appends each per-block `(identifier, offset)` in the same order. vLLM's `test_generate_block_hash_extra_keys` also asserts a block containing two items emits `(("hash1", 0), ("hash2", 10))`. The frontend independently rejects renderer placeholders that are not offset-ordered, so both sides consume the same ordering contract. This is a pre-existing vLLM invariant, not a new best-effort assumption in the Dynamo event parser.

jthomson04 (approved):

lib/llm/src/http/service/generate.rs
furionw: Thank you so much for your review, @jthomson04 

We don't have much production usage currently so I wouldn't worry about compatibility. Hence, I opt'd in for simplicity by not introducing the new flag. 

some AI content below ----

Removed the token/capability handshake in `be4c67cd02177f0007715a33adb18a8b48df96a5`.

For this rollout we are not concerned about mixed frontend/worker version compatibility; `/generate` plus KV events is treated as a homogeneous single-version stack, so that compatibility guard is deliberately out of scope. The frontend derives the image placeholder token from the rendered request and projects the routing sequence. The worker still resolves its local image token for KV-event normalization and consumes the frontend-provided routing markers. A mismatched stack could reduce cache-affinity quality, but it does not alter execution token IDs or vLLM native cache keys.

E2E passed on the exact pushed frontend SHA, with the stacked worker consumer changes at local combined head `3d8d867a6d2c599d78242c59d43482eac70f221c`:

- image `nvcr.io/nvstaging/ai-dynamo/vllm-runtime:qiwa-dev-vllm-amd64-08-25` (`sha256:b12498d5d459e2de53b7dcf2afb1c04fa658a3e6ce350d12c10ac61d0f01f019`)
- `Qwen/Qwen3.5-2B`, two aggregated workers, KV events enabled, no forced block size
- pure chat baseline: HTTP 200 and correct red-left/blue-right response
- unsalted and salted Generate suites: cold images split across different workers at 0/4 overlap; all four warm repeats per image returned to the corresponding cold worker and met the dynamic overlap threshold
- all Generate/derender calls returned HTTP 200 with correct red/blue output
- no metadata/token-only fallback, unresolved image token, stale event, source mismatch, or block-size mismatch warnings

PR #11590, opened 2026-07-13
KLXHY-RM-3252:feat(kv-router) decode length pooling, memory guard, co…

tobeprozy:
…nfigurable balance metric

Add decode-side routing controls for length-skewed workloads:
- pool_long_isl_threshold + pool_short_dp_ranks: split decode dp ranks into short/long pools by request ISL
- decode_block_guard_frac: hard memory admission guard per dp rank

Change-Id: I6bdc218ce108a0e9751a762a9e97b354b079a24c

## Details:

Add decode-side routing controls for length-skewed workloads:
- \`pool_long_isl_threshold\` + \`pool_short_dp_ranks\`: split decode dp ranks into short/long pools by request ISL, preventing long requests from crowding out short ones
- \`decode_block_guard_frac\`: hard memory admission guard per dp rank; router stops sending to a dp rank when projected memory usage exceeds \`guard_frac * total_kv_blocks\`
- \`max_kv_cache_tokens\`: override worker-reported KV cache size for guard calculations
- Decode worker cost function now uses batch-size (\`active_requests\`) instead of block count for more even load distribution across dp ranks
- New env vars: \`DYN_ROUTER_DECODE_BLOCK_GUARD_FRAC\`, \`DYN_ROUTER_POOL_LONG_ISL_THRESHOLD\`, \`DYN_ROUTER_POOL_SHORT_DP_RANKS\`, \`DYN_ROUTER_MAX_KV_CACHE_TOKENS\`

All new fields default to their off/disabled values so existing deployments are unaffected.

## Validation

- \`cargo check -p dynamo-kv-router\` passes
- Python config args, PyO3 bindings, and CLI flags expose all new options
- Existing unit tests in selector, queue, and sequences pass
- Merge-conflict resolution preserves main's \`SchedulingRequest\` structure (\`worker_loads\` projection model), cancel-rollback semantics, and \`router_predicted_ttl_secs\` predict-on-route feature"

### DIFF
--- a/components/src/dynamo/common/configuration/groups/kv_router_args.py
+++ b/components/src/dynamo/common/configuration/groups/kv_router_args.py
@@ -52,6 +52,10 @@ _KV_ROUTER_FIELDS: tuple[str, ...] = (
     "shared_cache_multiplier",
     "shared_cache_type",
     "router_predicted_ttl_secs",
+    "decode_block_guard_frac",
+    "pool_long_isl_threshold",
+    "pool_short_dp_ranks",
+    "max_kv_cache_tokens",
 )
 
 _DEPRECATED_OVERLAP_WEIGHT_MESSAGE = (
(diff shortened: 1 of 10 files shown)


PeaBrane (commented):
Thank you for the contribution and for exploring decode-side memory admission and length-aware routing here.

We are currently working through a router flow-control change that would hook decode pressure into the unified queue-admission path, so queueing and rejection share one mechanism rather than introducing a separate selector-side hard guard. That direction overlaps with `decode_block_guard_frac` in this PR.

Please hold off on merging this for now while we settle that integration. We will follow up with the intended direction and which parts of this work we may be able to reuse. Thanks again for putting this together.

PR #11597, opened 2026-07-13
fix(operator): isolate namespaced admission and CRD ownership

sttts:
## Summary

- keep validation and mutation separated: cluster-wide handlers skip namespaces with active ownership Leases, while namespace-restricted webhook configurations select only their target namespace
- make defaulting Lease-aware and keep conversion, CRD application, and conversion CA ownership exclusively cluster-wide
- reject namespace-restricted CRD upgrades and custom webhook namespace selectors; document the required --skip-crds installation flag
- classify namespace-restricted mode as development/test only and document version and lifecycle constraints

This deliberately avoids a feature-gate or principal protocol in the Lease.

## 1.3 and post-1.3 plan

| Target | PR | Admission model |
| --- | --- | --- |
| 1.3 | This PR ([#11597](https://github.com/ai-dynamo/dynamo/pull/11597)) | Keep separate global and namespaced webhooks. Global handlers stand down for namespaces with an active ownership Lease. |
| Post-1.3 | [#11475](https://github.com/ai-dynamo/dynamo/pull/11475) | Move to one global admission stack. Namespaced operators publish their complete effective feature gates on the ownership Lease for global admission to resolve. |

#11475 remains the candidate for the more involved post-1.3 admission model and compatibility work.

Linear: https://linear.app/nvidia/issue/DYN-3412

## Validation

- source .venv/bin/activate && make test (deploy/operator)
- make test (deploy/helm/charts/platform)
- make lint (deploy/helm/charts/platform)
- fern check
- fern docs broken-links
- pre-commit hooks on both commits

## Summary by CodeRabbit

* **New Features**
  * Namespace-restricted operator deployments can target a specified namespace while coexisting with a cluster-wide operator.
  * Admission webhooks now automatically target the appropriate namespaces, and restricted deployments avoid cluster-scoped CRD management.

* **Bug Fixes**
  * Added validation to prevent unsupported CRD upgrade and custom webhook selector configurations.
  * Improved admission handling to prevent restricted operators from processing excluded namespaces.

* **Documentation**
  * Updated installation, configuration, and troubleshooting guidance to clarify that namespace-restricted mode is for development and testing only.


deploy/helm/charts/platform/components/operator/templates/webhook-configuration.yaml
```diff
@@ -53,7 +53,7 @@ webhooks:
   {{- else if .Values.namespaceRestriction.enabled }}
   namespaceSelector:
     matchLabels:
-      kubernetes.io/metadata.name: {{ .Release.Namespace }}
+      kubernetes.io/metadata.name: {{ default .Release.Namespace .Values.namespaceRestriction.targetNamespace }}
```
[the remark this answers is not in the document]
sttts: Fixed all occurrences and covered a namespace named `true` in the Helm tests.

docs/kubernetes/webhooks.md
```diff
@@ -335,61 +320,45 @@ helm install dynamo-platform . -n <namespace> -f values.yaml
...
 
 ```
 Cluster:
 ├─ Operator A (cluster-wide, namespace: platform-system)
-│  └─ Validates all namespaces EXCEPT team-a
+│  └─ Skips reconciliation and admission in team-a
 └─ Operator B (namespace-restricted, namespace: team-a)
-   └─ Validates only team-a namespace
+   └─ Reconciles, validates, and mutates only team-a
 ```
```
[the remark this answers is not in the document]
sttts: Fixed with a `text` fence.

deploy/helm/charts/platform/components/operator/templates/webhook-configuration.yaml
```diff
@@ -53,7 +53,7 @@ webhooks:
   {{- else if .Values.namespaceRestriction.enabled }}
   namespaceSelector:
     matchLabels:
-      kubernetes.io/metadata.name: {{ .Release.Namespace }}
+      kubernetes.io/metadata.name: {{ default .Release.Namespace .Values.namespaceRestriction.targetNamespace | quote }}
```
julienmancuso, line 56: [P1] Require a Kubernetes version that supplies this namespace label

The chart still advertises Kubernetes v1.20+, but `kubernetes.io/metadata.name` is only auto-applied starting in Kubernetes 1.21 (`NamespaceDefaultLabelName`, GA in 1.22). On a supported 1.20 cluster, this selector matches no namespace. Once the ownership Lease is active, the cluster-wide handlers skip the namespace as well, leaving validation, defaulting, and mutation silently bypassed.

Please either raise and enforce the minimum Kubernetes version for namespace-restricted mode (1.22 is the safe GA boundary), or use a selector label that the chart requires to be present on the target namespace. Update both the README template and generated README accordingly.

Reference: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/#namespacedefaultlabelname

julienmancuso (commented):
One remaining P1 from my review.

deploy/helm/charts/platform/components/operator/templates/webhook-configuration.yaml
sttts: Oh, we should not try to support <1.22. That was like 2020 or so.

sttts: Agreed. We will raise and enforce the minimum to Kubernetes 1.33 in both the platform and operator charts, and align the documentation. The existing 1.20 chart claim is stale and already conflicts with the 1.24 documentation. More importantly, our current CRDs use `optionalOldSelf` / CRD validation ratcheting, which is GA in 1.33. This also safely covers the `kubernetes.io/metadata.name` label used here. I would not raise the general minimum to 1.34 only for DRA v1, since Dynamo can run without DRA and that feature already has its own 1.34 requirement.

julienmancuso (approved):

deploy/helm/charts/platform/Chart.yaml
```diff
@@ -20,6 +20,7 @@ maintainers:
 description: A Helm chart for NVIDIA Dynamo Platform.
 type: application
 version: 1.3.0
+kubeVersion: ">=1.32.0-0"
```
dmitry-tokarev-nv: What specific API or Kubernetes feature requires 1.32? Before this PR the chart had no `kubeVersion` constraint and the docs listed v1.24+. The `kubernetes.io/metadata.name` namespace-selector label stabilised in 1.21; the Coordination/Lease API in 1.17 — neither needs 1.32.

`kubeVersion` in `Chart.yaml` is enforced hard by Helm at install time; clusters running 1.28–1.31 (roughly two years of releases) would get a flat rejection. If this comes from the contributor's local cluster version rather than a real API dependency, please drop it or lower it to match the actual requirement — and add a comment on this line explaining what the floor protects against.

deploy/helm/charts/platform/components/operator/Chart.yaml
```diff
@@ -28,6 +28,7 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 version: 1.3.0
+kubeVersion: ">=1.32.0-0"
```
dmitry-tokarev-nv: Same question as the platform `Chart.yaml` above — why specifically 1.32? Please document the reason (ideally as a comment here) or lower the floor to the minimum version the operator actually needs.

deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
```diff
@@ -0,0 +1,106 @@
...
+        documentIndex: 1
+      - notExists:
+          path: webhooks[3].namespaceSelector
+        documentIndex: 1
+
+  - it: scopes namespace-restricted admission to the target namespace
+    template: charts/dynamo-operator/templates/webhook-configuration.yaml
+    set:
+      dynamo-operator.namespaceRestriction.enabled: true
+      dynamo-operator.namespaceRestriction.targetNamespace: "true"
```
dmitry-tokarev-nv, line 48: `"true"` is a valid namespace name (DNS label) but reads as a boolean and will confuse future readers. Consider `"test-ns"` or `"tenant-a"`.

dmitry-tokarev-nv (changes requested):
Two issues that are hard to pin to a single line:

**`--skip-crds` gap (NOTES.txt)**

Installing in namespace-restricted mode requires both `upgradeCRD=false` (now enforced by the new Helm validation) *and* `--skip-crds` (unenforced). Helm processes the `crds/` directory before template rendering, so no template can detect a missing `--skip-crds` flag. A user who sets `upgradeCRD=false` but forgets `--skip-crds` will silently install or upgrade cluster-wide CRDs from the chart's bundled `crds/` directory; the `crd-apply` init container is suppressed, but the Helm-managed CRDs still land. Suggest adding a reminder in `templates/NOTES.txt` under the namespace-restricted banner.

**Failureolicy risk window — admission recovery path**

`dynamo-operator.md` correctly notes that a dead namespace-restricted Pod can block admission and says "Recover or uninstall." The full sequence matters for ops:
1. Namespace-restricted Pod dies.
2. Lease expires after ~30 s → cluster-wide **reconciliation** resumes.
3. But the namespace-restricted `ValidatingWebhookConfiguration` / `MutatingWebhookConfiguration` still exist and point to the dead Pod's Service.
4. **Admission for the target namespace continues to fail** until `helm uninstall` is run.

Lease expiry restores reconciliation, not admission. `helm uninstall` is the only path that removes the webhook configs and fully restores cluster-wide admission. A sentence to that effect would prevent misunderstanding during incidents.

deploy/helm/charts/platform/Chart.yaml
sttts: CRD CEL features forces into 1.30 at least. Changing the minimum to that. `optionalOldSelf` and ratcheting are not available in 1.29 and before.

sttts: done

deploy/helm/charts/platform/components/operator/Chart.yaml
sttts: done

deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
sttts: the "true" was intentional to check proper YAML type handling. Added a comment.

sttts:
> `dynamo-operator.md` correctly notes that a dead namespace-restricted Pod can block admission and says "Recover or uninstall." The full sequence matters for ops:
> 
> 1. Namespace-restricted Pod dies.
> 2. Lease expires after ~30 s → cluster-wide **reconciliation** resumes.
> 3. But the namespace-restricted `ValidatingWebhookConfiguration` / `MutatingWebhookConfiguration` still exist and point to the dead Pod's Service.
> 4. **Admission for the target namespace continues to fail** until `helm uninstall` is run.
> 
> Lease expiry restores reconciliation, not admission. `helm uninstall` is the only path that removes the webhook configs and fully restores cluster-wide admission. A sentence to that effect would prevent misunderstanding during incidents.

Clarified.

dmitry-tokarev-nv (approved):
lgtm. all concerns were addressed.

deploy/helm/charts/platform/README.md
```diff
@@ -115,18 +101,18 @@ The chart includes built-in validation to prevent all operator conflicts:
...
+| dynamo-operator.namespaceRestriction.enabled | bool | `false` | DEVELOPMENT AND TESTING ONLY: Enable namespace-restricted reconciliation and admission. Not supported for production. |
+| dynamo-operator.namespaceRestriction.targetNamespace | string | `nil` | DEVELOPMENT AND TESTING ONLY: Target namespace. Defaults to the Helm release namespace. |
+| dynamo-operator.namespaceRestriction.lease | object | `{"duration":"30s","renewInterval":"10s"}` | DEVELOPMENT AND TESTING ONLY: Namespace ownership Lease settings. |
+| dynamo-operator.namespaceRestriction.lease.duration | string | `"30s"` | DEVELOPMENT AND TESTING ONLY: Namespace ownership Lease duration. |
+| dynamo-operator.namespaceRestriction.lease.renewInterval | string | `"10s"` | DEVELOPMENT AND TESTING ONLY: Namespace ownership Lease renewal interval. |
+| dynamo-operator.gpuDiscovery | object | `{"enabled":true}` | DEVELOPMENT AND TESTING ONLY: GPU discovery settings for namespace-restricted operators. |
+| dynamo-operator.gpuDiscovery.enabled | bool | `true` | DEVELOPMENT AND TESTING ONLY: Enable GPU discovery in namespace-restricted mode. |
```
rmccorm4, line 115: If these are only for namespace-restricted mode, why don't they have the same naming scheme like `dynamo-operator.namespaceRestriction.gpuDiscovery`?

julienmancuso: Conceptually, nesting it under namespaceRestriction would be cleaner. However, gpuDiscovery predates this PR as a top-level operator value, and moving it would break existing values files and --set dynamo-operator.gpuDiscovery.enabled=... usage. This PR only updates namespace ownership/admission behavior, so I’d keep the existing path and handle a values migration separately.

rmccorm4 (approved):
Approving to unblock docs for release, no glaring issues, but I'm not that in the loop on recent changes to the restricted namespace option that was previously slated for deprecation but no longer seems to be

deploy/operator/internal/namespace_scope/lease_manager.go
carlory, line 19: Hi @sttts ,  is the deprecation  message still valid?

sttts: We have technically degraded the feature from production to dev+test only. The deprecation applied to the former.

PR #11604, opened 2026-07-13
fix(responses): preserve raw reasoning output

Stelath:
## Overview

Preserves backend `reasoning_content` in Dynamo’s `/v1/responses` API. Non-streaming responses now use `reasoning_text` content instead of `summary_text`, and streaming responses emit the complete `v1/responses` lifecycle following the Open AI API spec.

## Details

- Emits `response.reasoning_text.delta` with the corresponding added/done events.
- Preserves reasoning, text, and tool-call ordering, including interleaved reasoning spans.
- Includes reasoning items in `response.completed.output`.
- Replays `content` while retaining `summary` as a compatibility fallback.
- Preserves the optimized borrowed serializer from #10498.

- `lib/llm/src/protocols/openai/responses/stream_converter.rs`
- `lib/llm/src/protocols/openai/responses/mod.rs`

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Improved reasoning output to preserve the original reasoning content when available.
  * Added structured reasoning segments that support interleaved text and tool-call activity.
  * Reasoning output is now provided only when explicitly requested.
  * Streaming responses now emit reasoning text events with improved ordering and support for multiple reasoning segments.

* **Bug Fixes**
  * Improved completion, error, truncation, and serialization handling for streamed reasoning content.

### DIFF
--- a/lib/llm/src/protocols/openai/responses/stream_converter.rs
+++ b/lib/llm/src/protocols/openai/responses/stream_converter.rs
@@ -16,16 +16,18 @@ use axum::response::sse::Event;
 use dynamo_protocols::types::responses::{
     AssistantRole, FunctionToolCall, InputTokenDetails, Instructions, OutputContent, OutputItem,
     OutputMessage, OutputMessageContent, OutputStatus, OutputTextContent, OutputTokenDetails,
-    Response, ResponseCompletedEvent, ResponseContentPartAddedEvent, ResponseContentPartDoneEvent,
-    ResponseCreatedEvent, ResponseFailedEvent, ResponseFunctionCallArgumentsDeltaEvent,
+    ReasoningItem, ReasoningTextContent, Response, ResponseCompletedEvent,
+    ResponseContentPartAddedEvent, ResponseContentPartDoneEvent, ResponseCreatedEvent,
+    ResponseFailedEvent, ResponseFunctionCallArgumentsDeltaEvent,
     ResponseFunctionCallArgumentsDoneEvent, ResponseInProgressEvent, ResponseOutputItemAddedEvent,
-    ResponseOutputItemDoneEvent, ResponseStreamEvent, ResponseTextDeltaEvent,
-    ResponseTextDoneEvent, ResponseTextParam, ResponseUsage, ServiceTier, Status,
-    TextResponseFormatConfiguration, ToolChoiceOptions, ToolChoiceParam, Truncation,
+    ResponseOutputItemDoneEvent, ResponseReasoningTextDeltaEvent, ResponseReasoningTextDoneEvent,
+    ResponseStreamEvent, ResponseTextDeltaEvent, ResponseTextDoneEvent, ResponseTextParam,
+    ResponseUsage, ServiceTier, Status, TextResponseFormatConfiguration, ToolChoiceOptions,
+    ToolChoiceParam, Truncation,
 };
 use serde::{
     Serialize,
-    ser::{SerializeMap, Serializer},
+    ser::{SerializeMap, SerializeSeq, Serializer},
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


lib/llm/src/protocols/openai/responses/stream_converter.rs
```diff
@@ -271,55 +399,10 @@ impl ResponseStreamConverter {
...
-                    ResponseReasoningSummaryTextDeltaEvent {
-                        sequence_number: self.next_seq(),
-                        item_id: self.reasoning_item_id.clone(),
-                        output_index: self.reasoning_output_index,
-                        summary_index: 0,
-                        delta: reasoning.to_string(),
-                    },
-                );
-                events.push(self.make_sse_event(&reasoning_delta));
+                self.append_reasoning_delta(reasoning, events);
```
rmccorm4, line 412: Could we complete any pending tool call before opening a resumed reasoning item here? For a `reasoning → tool call → reasoning` stream, the tool call currently remains in progress until EOF, so its `done` event is emitted after the later reasoning item’s `done` event. The final output sorting hides this, but the live SSE order is wrong. Please also add a test that checks the event order.

lib/llm/src/protocols/openai/responses/stream_converter.rs
```diff
@@ -54,13 +53,10 @@ pub struct ResponseStreamConverter {
...
     accumulated_text: String,
-    // Reasoning summary tracking
-    reasoning_item_id: String,
-    reasoning_started: bool,
-    reasoning_done: bool,
-    reasoning_output_index: u32,
-    reasoning_output_status: Option<OutputStatus>,
-    accumulated_reasoning: String,
+    // Ordered reasoning spans. A new item is opened if reasoning resumes after
+    // text or a tool call, preserving interleaved model output.
```
rmccorm4: This comment says reasoning can resume after text, but the `!message_started` guard below intentionally drops reasoning once visible text has started. Could we update the comment to describe the actual after-tool behavior? If post-text reasoning is intended, that would require splitting the message item.

lib/llm/src/protocols/openai/responses/stream_converter.rs
```diff
@@ -133,6 +148,119 @@ impl ResponseStreamConverter {
...
+                    part: OutputContent::ReasoningText(ReasoningTextContent {
+                        text: String::new(),
+                    }),
+                });
+            events.push(self.make_sse_event(&part_added));
+        }
+
+        let state_index = self
+            .active_reasoning_index
+            .expect("reasoning state must be active after opening it");
```
rmccorm4: Can we avoid the `expect` here? The active index is established immediately above, so carrying it in a local would keep this production path panic-free and make the invariant clearer.

lib/llm/src/protocols/openai/responses/stream_converter.rs
Stelath: Oh yeah sorry this snuck in from another change I believe

lib/llm/src/protocols/openai/responses/stream_converter.rs
Stelath: Good point, I think a `match` here to hand the value out should work

lib/llm/src/protocols/openai/responses/stream_converter.rs
Stelath: Done 👍

rmccorm4:
/ok to test [37e7be8](https://github.com/ai-dynamo/dynamo/pull/11604/commits/37e7be81fd886c9965109d6ffb32212e7ba63c8f)

rmccorm4:
/ok to test [1330d39](https://github.com/ai-dynamo/dynamo/pull/11604/commits/1330d39a5c6892d0150d6ac78d08350128e6ed88)

rmccorm4 (approved):

PR #10517, opened 2026-06-10
feat(trtllm): plumb conversation_id to engine for conversation-aware ADP routing

nv-yna:
## Summary

In disaggregated serving, TRT-LLM's `ConversationAwareADPRouter` pins each conversation to an attention-DP rank (for KV locality) by reading `disaggregated_params.conversation_id`. Native `trtllm-serve` populates this; the Dynamo TRT-LLM workers did **not** (**gap #4**), so the engine's conversation-affinity router never engaged and CTX prefill load scattered across ranks → prefill queue → high TTFT tail.

This plumbs `conversation_id` end-to-end so the engine router engages.

## Changes

- `RoutingHints` gains `conversation_id`, seeded by the OpenAI preprocessor from `nvext.session_control.session_id` (`lib/llm`).
- Both the **unified** worker (`llm_engine.py`) and the **legacy** handler (`request_handlers/handler_base.py`) set `disaggregated_params.conversation_id` on the engine request, on **prefill and decode**.
- Gated by `DYN_ENGINE_CONV_AFFINITY=1`: when set, the worker stops force-feeding `attention_dp_rank` — an explicit rank is honored *before* conversation affinity in the engine and would bypass `ConversationAwareADPRouter`.
- `dp_rank.conversation_id_from_request` helper + unit test; one-shot INFO log to confirm the runtime invariants without enabling debug.

## Validation — Lyris GB300, DSV4-Pro, 3×CTX-DEP4 + 1×GEN-DEP8, c460, perf-off

With the engine config `kv_cache_routing_conversation_affinity: true` (25-min run; 55-min full in progress):

| metric | before (kv-aware) | **this PR (conv-affinity)** | native trtllm-serve |
|---|---|---|---|
| TTFT p95 (s) | 8.39 | **4.84** | 4.81 |
| TTFT p99 (s) | 11.57 | 7.12 | 7.00 |
| TTFT p50 (s) | 2.96 | 2.92 | 2.76 |
| SSE p25 (tok/s) | 58.3 | 58.6 | 62.36 |
| req/s | 46.3 | 49.6 | 51.56 |

**TTFT p95 dropped ~42%, matching native.** Runtime-confirmed (one-shot INFO): `conversation_id` set + `attention_dp_rank` not forced, on both PREFILL and DECODE.

## Notes / dependencies

[...]

PR #11609, opened 2026-07-13
feat(trtllm): engine-owned conversation-aware ADP routing

krishung5:
_Port of #11288 (from `nv-yna:conv-aware-adp-routing`), opened while @nv-yna is OOO. Rebased on latest `main`; only merge conflict was orthogonal (`cache_salt` on main + `conv_kwargs` on this branch — both kept in the `generate_async` call). Original commit authored by @nv-yna._

---

## What this PR does

Adds opt-in support for TensorRT-LLM's **engine-owned conversation-affinity ADP routing**. When the engine's `attention_dp_config.kv_cache_routing_conversation_affinity` is enabled, the TRT-LLM worker:
1. derives the conversation id from the frontend-forwarded **`agent_context.session_id`**,
2. passes it as **`ConversationParams`** to `generate_async`, and
3. **stops forcing `attention_dp_rank`**, so the engine's `ConversationAwareADPRouter` pins the conversation to a DP rank (sticky, load-balanced first turn).

When the engine policy is **off**, Dynamo-owned strict DP-rank forcing is **byte-for-byte unchanged**.

## Why this shape (design notes for review)

This is the main-appropriate re-implementation of the DSV4 feature branch's conversation-aware routing (`321567ee83`, #10517), following @PeaBrane's review guidance ("avoid duplicate routing plumbing; derive the id from session identity in the engine handlers"):

- **No Rust changes.** On `main` the session identity already reaches the worker via the serialized `PreprocessedRequest.agent_context` field (`lib/llm/src/protocols/common/preprocessor.rs:272`). The feature branch's approach added a *second* copy (`RoutingHints.conversation_id`) — dropped here entirely.
- **`agent_context.session_id`, not the old `session_control`** — `main` diverged; `session_control` doesn't exist here. Uses `session_id` (the active reasoning/tool chain), not `parent_session_id`.
- **Gated on the resolved engine config, not a `DYN_*` env var** — a Dynamo-side switch could disagree with the engine (e.g. suppress the rank when no engine router exists). The engine config is the single source of truth.
- **`ConversationParams`, not `disaggregated_params.conversation_id`** — the feature branch mutated disaggregation metadata (specific to a custom wheel). Upstream shipped a dedicated `ConversationParams` argument (`generate_async(..., conversation_params=...)`), which also covers ordinary aggregated requests (the old path only worked for P/D).
- **No `exp-H`/`gap #4` instrumentation** — all experimental one-shot logging stripped for production.

## Files
- `components/src/dynamo/trtllm/conversation_affinity.py` — new helper: session-id extraction, engine-config gate, `ConversationParams` builder (guarded import).
- `components/src/dynamo/trtllm/llm_engine.py` — unified path: resolve gate at init; suppress rank + pass `ConversationParams` when enabled.
- `components/src/dynamo/trtllm/request_handlers/handler_base.py` — legacy path: same, gate resolved lazily.
- `components/src/dynamo/trtllm/tests/test_conversation_affinity.py` — wheel-independent unit tests for the helpers.

## Testing
- `ruff check` / `ruff format --check` clean; helper unit tests pass (6 passed, 1 skipped where the API is absent).
- No Rust diff → no `cargo` exposure.
- **Deferred to the dependency-bump follow-up:** GPU integration (sticky rank across turns, no-id fallback, aggregated + P/D) and the mocked-`generate_async` handler matrix.

## Open questions for review
1. Which released TRT-LLM version/image will carry the post-rc20 API? (sole hard blocker)
2. Frontend per-rank accounting vs engine-owned placement (`push_router.rs` still stamps the frontend rank) — needs an e2e accounting pass; must not reintroduce Rust conversation plumbing.
3. Fail-startup on incompatible-wheel-with-affinity-enabled — currently yes; confirm desired.

### DIFF
--- a/components/src/dynamo/trtllm/conversation_affinity.py
+++ b/components/src/dynamo/trtllm/conversation_affinity.py
@@ -1,0 +1,72 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Helpers for engine-owned conversation-aware ADP routing (opt-in).
+
+When TensorRT-LLM's ``attention_dp_config.kv_cache_routing_conversation_affinity`` is
+enabled, the engine's ``ConversationAwareADPRouter`` pins a conversation to an
+attention-DP rank (sticky, load-balanced first turn). Dynamo must then:
+
+  1. forward the stable conversation id via ``ConversationParams``, and
+  2. NOT force ``attention_dp_rank`` — an explicit rank is honored *before* affinity
+     in the engine and would bypass it.
+
+The conversation id is the frontend-forwarded ``agent_context.session_id`` — no new
+routing plumbing is added on the Rust side (see the serialized
+``PreprocessedRequest.agent_context`` field). When affinity is off, callers keep their
+existing Dynamo-owned DP-rank forcing unchanged.
+
+``ConversationParams`` and ``kv_cache_routing_conversation_affinity`` require a
+TensorRT-LLM build newer than 1.3.0rc20; on older wheels the import is absent and
+``CONVERSATION_PARAMS_AVAILABLE`` is False.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any, Optional
+
+try:  # Requires a TensorRT-LLM release newer than 1.3.0rc20.
+    from tensorrt_llm.llmapi import ConversationParams
+except ImportError:  # pragma: no cover - depends on installed wheel
+    ConversationParams = None  # type: ignore[assignment]
+
+CONVERSATION_PARAMS_AVAILABLE: bool = ConversationParams is not None
+
+
+def session_id_from_request(request: Mapping[str, Any]) -> Optional[str]:
+    """Return the stable conversation/session id the frontend forwards as
+    ``agent_context.session_id``, or ``None`` if absent, blank, or malformed.
+
+    Uses ``session_id`` (the active reasoning/tool chain), not ``parent_session_id``.
+    """
+    agent_context = request.get("agent_context")
+    if not isinstance(agent_context, dict):
+        return None
+    session_id = agent_context.get("session_id")
+    if isinstance(session_id, str) and session_id.strip():
+        return session_id.strip()
+    return None
+
+
+def engine_conversation_affinity_enabled(llm: Any) -> bool:
+    """Whether the resolved engine config enables conversation-affinity ADP routing.
+
+    Reads ``llm.args.attention_dp_config.kv_cache_routing_conversation_affinity``.
+    Only a genuine boolean ``True`` counts as enabled; tolerates a config model or a
+    plain dict, and defaults to False when the field is absent (e.g. older wheels).
+    """
+    adp_cfg = getattr(getattr(llm, "args", None), "attention_dp_config", None)
+    if adp_cfg is None:
+        return False
+    if isinstance(adp_cfg, dict):
+        return adp_cfg.get("kv_cache_routing_conversation_affinity") is True
+    return getattr(adp_cfg, "kv_cache_routing_conversation_affinity", None) is True
+
+
+def conversation_params_for(session_id: Optional[str]) -> Optional[Any]:
+    """Build ``ConversationParams`` for ``session_id``, or ``None`` when there is no id
+    (the engine router then applies its no-id balancing fallback). Requires the API."""
+    if session_id is None or ConversationParams is None:
+        return None
+    return ConversationParams(conversation_id=session_id)
--- a/components/src/dynamo/trtllm/llm_engine.py
+++ b/components/src/dynamo/trtllm/llm_engine.py
@@ -61,6 +61,12 @@ from dynamo.common.utils.structural_tag import serialize_structural_tag
 from dynamo.llm import KvEventPublisher, ModelInput
 from dynamo.trtllm.args import parse_args
 from dynamo.trtllm.constants import DisaggregationMode
+from dynamo.trtllm.conversation_affinity import (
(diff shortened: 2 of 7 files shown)


PeaBrane (commented):
The rank-ownership note is the main design concern. The remaining observations are minor, and API/version-sequencing comments are probably not important for this PR.

components/src/dynamo/trtllm/llm_engine.py
```diff
@@ -838,16 +861,27 @@ async def _generate_started(
...
-            else None
-        )
+        # In conversation-affinity mode let the engine's ConversationAwareADPRouter pick the
+        # attention-DP rank from the conversation id; do NOT force a rank (an explicit rank is
+        # honored before affinity and would bypass it). The rank is suppressed even without an
+        # id so the engine's no-id balancing fallback runs. Otherwise honour the router's DP
+        # rank decision; without it TRT-LLM picks its own rank and KV events land on the wrong
+        # publisher.
+        conversation_params = None
+        if self._conversation_affinity:
```
PeaBrane, line 885: Design note for human evaluation: consider solving this on the TensorRT-LLM router side. When a new conversation arrives with an explicit `attention_dp_rank`, `ConversationAwareADPRouter` could honor that rank and record the initial `conversation_id -> rank` binding; existing conversations would continue using their recorded binding. The current upstream ordering honors explicit placement before affinity without recording the binding, so forcing only the first request is not sufficient as-is.

If that change cannot land now and this PR needs to unblock, please leave a prominent TODO/NOTE here explaining the Dynamo-versus-engine rank-accounting mismatch and open/link a tracking issue for the TensorRT-LLM follow-up. Human reviewers should decide whether that deferral is acceptable for the intended rollout.

components/src/dynamo/trtllm/llm_engine.py
```diff
@@ -184,6 +190,10 @@ def __init__(
         self._metrics_thread: Optional[threading.Thread] = None
         self._kv_events_thread: Optional[threading.Thread] = None
         self._attention_dp_size: int = 1
+        # Engine conversation-affinity gate; real value resolved in initialize().
+        # Default False so generate() is safe if reached before initialize()
+        # (e.g. unit tests that drive generate directly).
+        self._conversation_affinity: bool = False
```
PeaBrane, line 196: AI review note (minor; please evaluate): an existing trace-propagation fixture constructs `TrtllmLLMEngine` through `__new__`, so it does not initialize this field and a later affinity read can raise `AttributeError`. This likely only requires the fixture to set `_conversation_affinity = False`.

components/src/dynamo/trtllm/request_handlers/handler_base.py
```diff
@@ -1096,11 +1106,33 @@ async def _generate_locally_impl(
         # Build trace headers for distributed tracing
         trace_headers = context.trace_headers()
 
+        # Resolve the engine-owned conversation-affinity gate once (lazily; the engine is
+        # initialized by first request). When on, the engine's ConversationAwareADPRouter
+        # picks the attention-DP rank from the conversation id, so we must NOT force a rank.
+        if self._conversation_affinity is None:
```
PeaBrane, line 1112: AI review note (minor; probably not important for this PR): this API compatibility check occurs on the first request and outside the method’s `try` block, rather than during worker setup. That gives incompatible wheels different failure timing between the legacy and unified paths. Human reviewers may want to confirm that this is acceptable for the staged rollout.

components/src/dynamo/trtllm/conversation_affinity.py
```diff
@@ -0,0 +1,72 @@
...
+  1. forward the stable conversation id via ``ConversationParams``, and
+  2. NOT force ``attention_dp_rank`` — an explicit rank is honored *before* affinity
+     in the engine and would bypass it.
+
+The conversation id is the frontend-forwarded ``agent_context.session_id`` — no new
+routing plumbing is added on the Rust side (see the serialized
+``PreprocessedRequest.agent_context`` field). When affinity is off, callers keep their
+existing Dynamo-owned DP-rank forcing unchanged.
+
+``ConversationParams`` and ``kv_cache_routing_conversation_affinity`` require a
```
PeaBrane, line 19: AI review note (minor; probably not important for this PR): this requires a post-rc20 TensorRT-LLM API while the repository and container still pin rc20. The PR already describes a dependency-bump follow-up, so this sequencing may be intentional; please evaluate whether the follow-up is tracked clearly enough.

components/src/dynamo/trtllm/tests/test_conversation_affinity.py
```diff
@@ -0,0 +1,100 @@
...
+    assert engine_conversation_affinity_enabled(_Llm(_Args(None))) is False
+    assert engine_conversation_affinity_enabled(_Llm(None)) is False
+    assert engine_conversation_affinity_enabled(object()) is False
+
+
+def test_conversation_params_for_none_id_is_none():
+    assert conversation_params_for(None) is None
+
+
+@pytest.mark.skipif(
```
PeaBrane, line 93: AI review note (minor; please evaluate): these tests cover the helpers but not the changed unified and legacy `generate_async` call shapes. A small mocked matrix could catch wiring regressions, though this can reasonably be follow-up work if the API rollout is deliberately staged.

PeaBrane (approved):
✅

components/src/dynamo/trtllm/llm_engine.py
krishung5: Added a TODO comment.

components/src/dynamo/trtllm/tests/test_conversation_affinity.py
krishung5: Done in [6ab25d7](https://github.com/ai-dynamo/dynamo/pull/11609/commits/6ab25d74b55319179e7fc31539454d8ad1b91e40)

richardhuo-nv (approved):

PR #11141, opened 2026-07-01
feat(sglang): support multiple served model names

fengyuli-dev:
## Overview:

Support advertising a model under multiple served names, for the **SGLang** backend. A worker may pass several names via `--served-model-name` (whitespace- or comma-separated). The first is the primary served name; the rest become aliases. The same WorkerSet is registered under every name, and requests for any alias are canonicalized to the primary at the HTTP entrypoint so engine routing, Prometheus metrics, and the OpenAI `response.model` stay consistent.

This PR adds the shared alias core (ModelManager, watcher, HTTP, model card) plus the SGLang wiring. vLLM and TRT-LLM support follow in separate, independent PRs that build on this core.

## Details:

- **Rust core**: `ModelDeploymentCard.aliases`; `ModelManager` `alias_to_primary` map with `register_alias` / `unregister_alias_if_empty` / `resolve_canonical_name` + collision guards; watcher registers aliases before the engine is visible and tears them down when empty. Aliases are excluded from the MDC checksum so an alias-only change doesn't force a worker drain.
- **HTTP**: resolve alias → primary in every handler (composed with the existing `metric_model_for()` unknown-model cardinality guard) and canonicalize the cancellation-metric label; `/v1/models` reports the primary's `context_window` for alias entries.
- **Binding**: `register_model` gains an optional `model_aliases` argument.
- **SGLang**: split a packed `--served-model-name` into primary + aliases via a shared `split_served_model_names` helper, plumbed through `register_model`.

Unit tests added for `ModelManager` (Rust) and SGLang arg parsing (Python).

## Where should the reviewer start?

- `lib/llm/src/discovery/model_manager.rs` — the alias map + collision guards.
- `lib/llm/src/discovery/watcher.rs` — alias registration/teardown ordering.
- `lib/llm/src/http/service/openai.rs` — canonicalization + metric labels + `/v1/models`.

## Related Issues

**🚫 This PR is NOT linked to an issue**:
- [x] Confirmed — no related issue

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11141" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple model aliases when registering a served model name.
  * Model requests and listings now recognize aliases and route to the canonical model name.

[...]

PR #11611, opened 2026-07-13
feat(vllm): support multiple served model names

fengyuli-dev:
## Overview:

Follow-up to #11141 (SGLang multi-served-model-name), applying the same support
to the vLLM backend so a single vLLM worker can advertise a primary served name
plus additional aliases.

## Details:

vLLM's `--served-model-name` is `nargs="+"`, but the Dynamo wrapper previously
rejected more than one name (`"We do not support multiple model names."`). Now
the first name is the primary served name and any additional names are
registered as aliases for the same worker, reusing the shared
`split_served_model_names` helper and passing them through
`register_model(model_aliases=...)` — the same frontend path SGLang uses.
Alias→primary canonicalization, global name reservation, and teardown ownership
already live in the frontend (from #11141); this PR only wires the vLLM worker
into it.

- `args.py`: `update_dynamo_config_with_engine` splits the served names into
  primary + aliases instead of rejecting multiples (direct attribute access, no
  defensive `getattr`).
- `backend_args.py`: add `served_model_aliases: Optional[List[str]]`.
- `main.py`: pass `model_aliases=config.served_model_aliases or None` to
  `register_model`.

`components/src/dynamo/vllm/args.py` — `update_dynamo_config_with_engine`, where
`--served-model-name` is split into primary + aliases.

## Validation

- `black` / `isort` / `flake8` and `pre-commit` clean.
- Unit tests pass (`components/src/dynamo/vllm/tests/test_vllm_unit.py`):
  `test_parse_args_splits_served_model_name_into_aliases`,
  `test_parse_args_splits_comma_packed_served_model_name`,
  `test_parse_args_without_served_model_name_has_no_aliases`.
- End-to-end on 1×H100 with Qwen3-0.6B
  (`--served-model-name Qwen/Qwen3-0.6B alias-one alias-two`):
  - `/v1/models` lists the primary + both aliases.
  - Chat completions to the primary and to each alias return
    `response.model == "Qwen/Qwen3-0.6B"` (canonicalized) with non-empty content.
  - An unknown model name returns 404.
  - Frontend logs confirm `Registering model alias ... alias="alias-one"/"alias-two"`
    and `Adding worker set to model=alias-one/alias-two`.

## Related Issues

### DIFF
--- a/components/src/dynamo/vllm/args.py
+++ b/components/src/dynamo/vllm/args.py
@@ -22,6 +22,7 @@ from dynamo.common.configuration.groups.runtime_args import (
     DynamoRuntimeArgGroup,
     DynamoRuntimeConfig,
 )
+from dynamo.common.configuration.utils import split_served_model_names
 from dynamo.common.utils.runtime import parse_endpoint
 from dynamo.vllm.backend_args import DynamoVllmArgGroup, DynamoVllmConfig
 from dynamo.vllm.constants import DisaggregationMode
@@ -169,13 +170,12 @@ def update_dynamo_config_with_engine(
 ) -> None:
     """Update dynamo_config fields from engine_config and worker flags."""
 
-    if getattr(engine_config, "served_model_name", None) is not None:
-        served = engine_config.served_model_name
-        if len(served) > 1:
-            raise ValueError("We do not support multiple model names.")
-        dynamo_config.served_model_name = served[0]
-    else:
-        dynamo_config.served_model_name = None
+    # vLLM's --served-model-name is nargs="+"; each token may itself pack
+    # several comma-separated names. The first is the primary served name; any
+    # remaining names are registered as aliases for the same worker.
+    served_names = split_served_model_names(engine_config.served_model_name)
+    dynamo_config.served_model_name = served_names[0] if served_names else None
+    dynamo_config.served_model_aliases = served_names[1:]
 
     # Capture user-provided --endpoint before defaults overwrite it
     user_endpoint = dynamo_config.endpoint
--- a/components/src/dynamo/vllm/main.py
+++ b/components/src/dynamo/vllm/main.py
@@ -756,6 +756,7 @@ async def register_vllm_model(
         worker_type=worker_type,
         needs=needs,
         ignore_weights=should_register_model_ignore_weights(config),
+        model_aliases=config.served_model_aliases or None,
         # Advertise LoRA capacity on the BASE card so the frontend can place the first
         # adapter onto an idle worker. Decode, aggregated, and prefill workers all serve
         # lifecycle registration; embeddings still do not.
--- a/components/src/dynamo/vllm/backend_args.py
+++ b/components/src/dynamo/vllm/backend_args.py
@@ -9,7 +9,7 @@ import argparse
 import logging
 import os
 import warnings
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 from dynamo.common.configuration.arg_group import ArgGroup
 from dynamo.common.configuration.config_base import ConfigBase
@@ -446,6 +446,10 @@ class DynamoVllmConfig(ConfigBase):
 
     # GMS shadow mode
     gms_shadow_mode: bool = False
+
+    # Extra served names beyond the primary, parsed from --served-model-name.
+    # None (not []) since ConfigBase copies class defaults by reference.
+    served_model_aliases: Optional[List[str]] = None
 
     # Benchmark / self-profiling
     benchmark_mode: Optional[str] = None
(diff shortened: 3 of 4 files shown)


components/src/dynamo/vllm/main.py
```diff
@@ -752,6 +752,7 @@ async def register_vllm_model(
         worker_type=worker_type,
         needs=needs,
         ignore_weights=should_register_model_ignore_weights(config),
+        model_aliases=config.served_model_aliases or None,
```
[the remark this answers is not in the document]
fengyuli-dev: This matches the merged SGLang alias PR (#11141) — both wire the documented dynamo.{vllm,sglang} path, and the shared backend-common/unified_main path carries no alias field for either backend, so probably let's not widen this PR.

components/src/dynamo/vllm/args.py
```diff
@@ -169,13 +170,12 @@ def update_dynamo_config_with_engine(
...
-            raise ValueError("We do not support multiple model names.")
-        dynamo_config.served_model_name = served[0]
-    else:
-        dynamo_config.served_model_name = None
+    # vLLM's --served-model-name is nargs="+"; each token may itself pack
+    # several comma-separated names. The first is the primary served name; any
+    # remaining names are registered as aliases for the same worker.
+    served_names = split_served_model_names(engine_config.served_model_name)
+    dynamo_config.served_model_name = served_names[0] if served_names else None
+    dynamo_config.served_model_aliases = served_names[1:]
```
tmonty12, line 178: `served_model_aliases` is parsed here, but it is only consumed by the legacy registration path. `python -m dynamo.vllm.unified_main` builds the shared `WorkerConfig` without aliases, and the Rust `build_local_model()` consequently registers only the primary name. That makes this feature silently unavailable on the unified backend. Can we carry aliases through the shared WorkerConfig/PyO3/backend-common boundary and call `LocalModelBuilder::model_aliases()`? This should be shared rather than vLLM-specific: unified SGLang has the same gap. Please add a unified registration test asserting the resulting model card contains the aliases; the new parser-only tests do not cover the failure mode.

tmonty12 (commented):
P1: unified alias registration

components/src/dynamo/vllm/args.py
fengyuli-dev: Done — carried aliases through the shared boundary rather than the vLLM-specific path:
- `WorkerConfig` gains a `model_aliases` field (Python dataclass + PyO3 + `backend-common`), populated once in `from_runtime_config` from `served_model_aliases`, so both vLLM and SGLang unified paths pick it up.
- `build_local_model()` now calls `LocalModelBuilder::model_aliases()`.
- Added a `backend-common` test asserting the built model card carries the aliases, plus Python coverage for the `from_runtime_config` plumbing.

Verified end-to-end on `python -m dynamo.vllm.unified_main` with Qwen3-0.6B: `/v1/models` lists the primary + both aliases and each alias canonicalizes to the primary. `deb5176`

tmonty12 (commented):
P2: missing public typing stub for model_aliases

lib/bindings/python/rust/backend.rs
```diff
@@ -319,6 +319,7 @@ impl WorkerConfig {
         route_to_encoder = false,
         media_decoder = None,
         media_fetcher = None,
+        model_aliases = Vec::new(),
```
tmonty12: This makes `model_aliases` a public `_core.backend.WorkerConfig` keyword, but the checked-in stub at `lib/bindings/python/src/dynamo/_core.pyi:3262-3290` still ends the constructor at `media_fetcher`. The unified worker now passes this keyword, so mypy/pyright report a false unsupported-keyword error and IDE autocomplete omits it, despite the compiled binding accepting it at runtime. Please add `model_aliases: List[str] = ...` after `media_fetcher` in that stub (and preferably a binding-signature test).

tmonty12 (commented):
P2: public typing contract for model_aliases

tmonty12 (approved):

lib/bindings/python/rust/backend.rs
fengyuli-dev: added

fengyuli-dev:
> /ok to test [7bef958](https://github.com/ai-dynamo/dynamo/commit/7bef9581ad693b1713182f1fef19c5839360db38)

Pipeline still saying "review required"

nnshah1 (approved):
@tanmayv25 , @connorcarpenter15 , @alec-flowers for viz

fengyuli-dev:
> @tanmayv25 , @connorcarpenter15 , @alec-flowers for viz

Still saying: 
> Waiting on code owner review from ai-dynamo/dynamo-backend-vllm-codeowners.

connorcarpenter15 (approved):

PR #11612, opened 2026-07-13
docs(profiler): clarify infeasible SLA behavior

tedzhouhk:
## Summary

Fixes [DYN-2396](https://linear.app/nvidia/issue/DYN-2396/dynamorelease100-dgdr-tc-92-infeasible-sla-targets-silently-accepted).

DYN-2396 was previously addressed by adding an `SLA is unverified` warning to the naive fallback path. Follow-up QA verification found that the profiler documentation still described different behavior for the AIC-supported rapid path: when every AIC experiment returns no SLA-feasible configuration, profiling fails instead of relaxing the SLA and generating a DGD.

This PR:

- documents that rapid profiling fails when every AIC experiment returns no SLA-feasible configuration
- clarifies that rapid profiling does not relax the SLA or generate a DGD in this case
- distinguishes thorough profiling, which selects the closest measured configuration when sweeping produced usable results
- aligns the profiler guide and troubleshooting guidance with current runtime behavior; this PR does not change runtime behavior

## Validation

- `PYTHONPATH="$PWD/lib/bindings/kvbm/python" .venv/bin/pre-commit run --files docs/components/profiler/profiler-guide.md`
- `NODE_OPTIONS=--experimental-global-webcrypto fern check`
- `NODE_OPTIONS=--experimental-global-webcrypto fern docs broken-links`
- `git diff --check`


PeaBrane (approved):
✅

PR #11615, opened 2026-07-13

ishandhanani:
This replaces [#11530](https://github.com/ai-dynamo/dynamo/pull/11530) directly on `main` now that its queue admission API dependency [#11434](https://github.com/ai-dynamo/dynamo/pull/11434) has merged. The old PR was closed automatically when copy-pr-bot deleted its temporary `pull-request/11434` base; its review discussion and detailed ABBA results remain the historical record.

CLOSES: DYN-3231

### How This Was Implemented

- Pair a strategy-retained `RequestProgress` reader with a response-stream-owned updater backed by a monotonic relaxed atomic.
- Track one actor-owned admission lifecycle across pending, selected, dispatched, completed, aborted, and cancelled states.
- Transfer cleanup ownership from the scheduling future to the response guard exactly once, with coalesced actor cleanup for dropped leases.
- Reclassify exact queue placement after admission decisions and support per-strategy deadlines.
- Keep the default path isolated: when queue admission is not configured, no admission lease is created and cleanup polling is skipped.
- Rebase the implementation onto `main`'s merged `QueueAdmissionConfig::SessionAware` API and cancellation-token lifecycle.

<details>

```mermaid
flowchart LR
  Q -->|"write capability"| G["RequestGuard"]
  G -->|"monotonic progress"| S
  G -->|"completed / aborted"| Q
  Q -->|"lifecycle event"| S
```

The queue actor creates the progress pair only for a request that enters an admission lifecycle. The strategy retains the reader; the updater crosses the scheduling response into the existing response guard. Before handoff, the scheduling future owns cancellation. After handoff, the response guard owns terminal cleanup.

Progress writes use `fetch_max`, so delayed observations cannot move logical context backward. Relaxed ordering is sufficient because the capability communicates only the numeric latest value.

No concrete admission algorithm or distributed admission protocol is added here.

</details>

### Validation

- `cargo test -p dynamo-kv-router --lib --no-fail-fast` — 771 passed, 2 ignored
- `cargo test -p dynamo-llm --lib --no-fail-fast` — 1,647 passed, 5 ignored
- `cargo clippy -p dynamo-kv-router -p dynamo-llm --lib --tests -- -D warnings`
- `git diff --check`

### Benchmark Results

Two-cycle release ABBA tests used four samples per arm, a fresh process per sample, and CPUs 0-7. The default-path comparison is current `main` (`fa58175170`) versus this PR (`40c792e7f0`) with admission unconfigured; positive throughput and negative p99 deltas are better.

| Admission | Concurrency | Main throughput | PR throughput | Delta | Main p99 | PR p99 | Result |
|---|---:|---:|---:|---:|---:|---:|---|
| Off | 1 | 432,962 req/s | 437,230 req/s | +1.0% | 5 us | 5 us | No regression |
| Off | 256 | 87,844 req/s | 91,990 req/s | +4.7% | 10,923 us | 10,549 us (-3.4%) | No regression |

The separate opt-in comparison ran the PR with admission off versus a minimal always-ready session-aware strategy. The on arm includes dispatch notification, one progress update, completed lease drop, and actor-owned terminal cleanup; compare within each row because this was a separate ABBA block.

| Concurrency | Off throughput | On throughput | Delta | Off p99 | On p99 | Effect of enabling admission |
|---:|---:|---:|---:|---:|---:|---|
| 1 | 421,395 req/s | 254,961 req/s | -39.5% | 5 us | 6 us | Lower synthetic actor ceiling; +1 us p99 |
| 256 | 116,784 req/s | 210,946 req/s | +80.6% | 7,409 us | 2,022 us (-72.7%) | Batched actor cleanup wins under saturation |

The merge therefore shows no measurable regression when admission is off. Enabling admission adds fixed lifecycle work at concurrency 1, while its batched cleanup improves saturated throughput; the earlier [112-run adversarial matrix](https://github.com/ai-dynamo/dynamo/pull/11530#issuecomment-4961430893) separately covers deferred, ready, mixed, progress, and forced-capacity patterns.

## Summary by CodeRabbit

* **New Features**
  * Added admission-aware scheduling strategies with deferred request handling.
  * Added request progress tracking, including live context-token updates.
  * Added lifecycle controls for marking requests dispatched, completed, or aborted.
  * Added admission-aware routing and configurable strategy startup support.

* **Bug Fixes**
  * Improved cancellation, streaming completion, and cleanup behavior.
  * Corrected queue accounting and priority handling for deferred requests.

* **Tests**
  * Added coverage for admission lifecycles, deferred scheduling, progress tracking, and stream outcomes.


lib/llm/src/kv_router/push_router/selection.rs
```diff
@@ -75,7 +77,7 @@ impl KvPushRouter {
     async fn select_best_match(&self, args: BestMatchArgs<'_>) -> Result<WorkerSelection, Error> {
         let outcome = self
             .chooser
-            .find_best_match_details_with_policy_class(
+            .find_best_match_details_with_admission(
```
ishandhanani, line 80: im not sure if we needed to change this api name

ishandhanani: Kept the public find_best_match_details_with_policy_class API unchanged. The push router uses the crate-private find_best_match_details_with_admission variant because it also needs the progress updater and admission lease; existing external callers remain on the original API.

lib/llm/src/kv_router/push_router/request_guard.rs
```diff
@@ -20,29 +21,52 @@ use crate::{
...
-        {
+    async fn finish(&mut self, outcome: RequestOutcome) {
+        let result = if let Some(mut lease) = self.admission_lease.take() {
+            if let RequestOutcome::Completed { context_tokens } = outcome {
+                lease.mark_completed(context_tokens);
+            }
+            // AdmissionLease reports the terminal outcome to the scheduler actor.
+            // The scheduler actor remains the sole owner of booking and queue cleanup.
+            drop(lease);
+            Ok(())
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 17fc6c45b4. Abort now explicitly resets the lease outcome to Aborted, with completed-then-aborted regression coverage.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -583,6 +966,276 @@ impl<
...
+                if self.slots.request_worker(&owned_request_id).is_some() {
+                    if let Err(error) = self.slots.free(&owned_request_id, Instant::now()) {
+                        tracing::error!(%request_id, %error, "Failed to release dropped scheduler booking");
+                    }
+                    made_ready = true;
+                }
+            }
+
+            let Some(tracked) = tracked else {
+                unmanaged_request_ids.insert(request_id);
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 17fc6c45b4. unmanaged_request_ids now uses the standard HashSet; the same commit documents that AdmissionRequest::new creates fixed progress.

tedzhouhk (approved):

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -448,22 +678,67 @@ impl<
...
+                    let made_ready = self.handle_enqueue(request, block_hashes)
+                        | (drain_cleanup && self.drain_cleanup());
+                    if made_ready {
+                        self.handle_update(None).await;
+                    }
                     let _ = ack_tx.send(());
                 }
                 AdmissionCommand::Update { worker, ack_tx } => {
                     self.handle_update(worker).await;
+                    if drain_cleanup && self.drain_cleanup() {
```
PeaBrane, line 811: **[AI][blocker] Cleanup can overtake the lifecycle command it represents.** The lease publishes its outcome through the side `SegQueue`, independently of FIFO actor commands. An older `Update` can decide to drain cleanup, await scheduling work, and then consume cancellation for an `Enqueue` that entered the actor channel during that await but has not been processed yet. `drain_cleanup` treats it as unmanaged, finds no storage, and forgets it; the later `Enqueue` can then install dead deferred or ready state. The same ordering permits a terminal event to be processed before an already-enqueued `Dispatched`. Please retain early cancellation until the matching lifecycle generation is observed, or otherwise put these transitions under one sequenced authority. Checking only for a closed response during enqueue would not fix the `Dispatched` ordering case.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -488,73 +763,185 @@ impl<
...
+        if let Some((request_admission, decision)) = admission.as_ref() {
+            match decision {
+                AdmissionDecision::Bypass => admission = None,
+                AdmissionDecision::Ready(placement) => {
+                    if let Err(error) = apply_admission_placement(&mut request, *placement) {
+                        request.respond(Err(error));
+                        return self.abort_admission(request_admission.ticket);
+                    }
+                    if matches!(placement, WorkerPlacement::Exact(_)) {
+                        // Classification stays in the family selected above; queue cost and
```
PeaBrane: **[AI][blocker] Exact placement does not reclassify the physical policy class.** `class_index` was resolved from the pre-strategy best-cache snapshot. Clearing the snapshot recomputes cached-token accounting and scheduling cost for the exact worker, but the request is still enqueued in the original class. With multiple uncached-ISL buckets it can therefore inherit the wrong DRR quantum, queue limits, busy threshold, queue policy, and backlog isolation. Deferred `MakeReady(Exact)` has the same problem because it calls `make_ready` with `class_action.class_index`. Please model the strategy-owner class separately from the physical queue class and rekey the latter after exact placement. The current exact tests use a single `all` bucket and do not cover this transition.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -350,26 +346,35 @@ impl KvPushRouter {
...
 
                     item = response_stream.next() => {
                         let Some(item) = item else {
-                            break;
+                            break true;
                         };
+                        failed |= response_item_failed(&item);
                         guard.on_item(&item).await;
+                        if !failed && response_item_completed(&item) {
+                            guard.mark_completed_terminal();
```
PeaBrane: **[AI][blocker] Successful completion remains reversible and consumer-poll dependent.** This marks the lease `Completed` before yielding the successful terminal item. If the consumer drops immediately, drop reports `Completed`; if it polls once more and cancellation or a late error wins, `abort()` calls `mark_aborted()` and overwrites the outcome. The same delivered `Stop`/`EoS`/`Length` response can therefore commit or discard strategy context depending only on another poll. If a successful terminal item is the commit point, completion should be sticky and the wrapper should terminate after that yield resumes.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -583,6 +970,276 @@ impl<
...
+                    .entry(tracked.ticket.class_index)
+                    .or_default()
+                    .insert(tracked.ticket.id);
+            }
+            made_ready |= self.finish_admission(tracked.ticket, outcome);
+        }
+
+        // One heap rebuild per affected class keeps cancellation storms O(classes * queue),
+        // while counters are released in the same actor turn as the lifecycle event.
+        for (class_index, tickets) in ready_by_class {
```
PeaBrane, line 1157: **[AI][warning] Removing managed ready entries does not redrive scheduling.** This loop retracts storage and counters but never sets `made_ready`, so the `Cleanup` command does not call `handle_update`. If a canceled shared-queue head was hiding a dispatchable request behind it, that request remains parked until an unrelated worker update or periodic reconciliation—up to 60 seconds by default. The unmanaged-removal branch does set `made_ready`, which highlights the inconsistency. `handle_finished` has the same omission for pre-selection ready entries.

lib/kv-router/src/scheduling/policy_queue.rs
```diff
@@ -540,11 +577,135 @@ impl<T> PolicyQueue<T> {
...
+        class.pending = retained;
+
+        removed.extend(
+            class
+                .deferred
+                .extract_if(|_, entry| remove_sequences.contains(&entry.enqueue_seq))
+                .map(|(_, entry)| entry),
+        );
+
+        class.ready_by_worker.retain(|_, ready| {
```
PeaBrane, line 743: **[AI][warning] A steady cancellation can rebuild the entire class on every actor turn.** `take_if_in_class` first scans all entries, allocates a removal set, then drains and reinserts every retained element from the shared heap and every worker heap. This is roughly `O(N log N)` serialized scheduler work for one cancellation. Coalescing helps bursts already accumulated together, but not a steady stream where the actor drains one cancellation per turn. An admission-ticket index with accounted tombstones/indexed removal would contain this cost; at minimum, rebuilding retained heaps with linear heapification would avoid repeated insertion cost.

lib/kv-router/src/scheduling/local.rs
```diff
@@ -457,6 +518,27 @@ where
...
+    /// Release request state and report its terminal admission outcome.
+    pub async fn finish(
+        &self,
+        request_id: &str,
+        outcome: RequestOutcome,
+    ) -> Result<(), SequenceError> {
+        let request_id = request_id.to_string();
+        let worker = self.slots.request_worker(&request_id);
+        let result = self.slots.free(&request_id, Instant::now());
+        self.queue.finish(&request_id, outcome).await;
```
PeaBrane: **[AI][warning] `finish` can half-complete cleanup if its future is canceled.** Worker booking is freed synchronously before this bounded-channel send. If the task is canceled while `send().await` waits for capacity, the unsent `Finished` command is dropped: the slot is released, but actor lifecycle state and any strategy reservation remain live. The response-guard lease path avoids this function, but this is public and documented as the admission-managed cleanup API. Prefer making `finish` consume/disarm the lease or submit one actor-owned transaction that performs both cleanup duties.

PeaBrane (commented):
AI-assisted review. The lease direction is strong, but I found three correctness blockers and three warnings.

Architectural recommendations to evaluate before patching the findings individually:

- Maintain one actor-owned, monotonic lifecycle per admission. Retain early cancellation until the matching admission exists, acknowledge `Dispatched` processing, and perform worker-slot and admission cleanup in one actor transition.
- Represent the admission strategy’s owner class separately from the mutable physical queue class, allowing exact placement to reclassify queue storage without losing strategy callback ownership.

These recommendations are scoped to the admission/queue lifecycle. They should not require a router-wide redesign or change DRR itself; equivalent designs preserving ordering and atomicity are fine.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: [AI] Fixed in 4c463dd297. The admission lease now travels inside the bounded EnqueueLeased command and is armed only after the actor establishes lifecycle ownership. Cancellation before or during the acknowledgement therefore drops actor-owned state after Enqueue; generation matching prevents stale cleanup from touching a reused request ID, and the lease records dispatch synchronously so cleanup emits Dispatched before the terminal event. The batched SegQueue cleanup path and bounded backpressure are preserved.

lib/kv-router/src/scheduling/local.rs
ishandhanani: [AI] Fixed in 4c463dd297. Removed the public cancelable finish API chain from SchedulerQueue, LocalScheduler, KvScheduler, and KvRouter. AdmissionLease is now the sole admission-managed terminal cleanup path, so cancellation cannot free the worker slot without also retaining actor lifecycle cleanup.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: [AI] Fixed in 8021bbc799. Admission strategy ownership is now separate from the physical queue class. Immediate and deferred exact placements recompute the exact-worker snapshot, reclassify/migrate the queue entry, and move physical accounting; multi-bucket tests cover both paths.

lib/llm/src/kv_router/push_router.rs
ishandhanani: [AI] Fixed in 8021bbc799. After yielding a successful terminal item, the wrapper now terminates as soon as that yield resumes, so later polling, cancellation, or a late stream error cannot reverse the completed outcome.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: [AI] Fixed in 8021bbc799. Managed ready cancellation now detects removal of a ready lane head and immediately redrives when the replacement head is dispatchable. The regression test covers a canceled blocking head exposing the next request; no-op redrives under saturation are avoided.

lib/kv-router/src/scheduling/policy_queue.rs
ishandhanani: [AI] Fixed in 8021bbc799. Shared and exact-worker retained entries are collected into vectors and rebuilt with BinaryHeap::from, changing heap reconstruction from repeated O(N log N) insertion to linear heapification.

lib/llm/src/kv_router/push_router/selection.rs
```diff
@@ -75,7 +77,7 @@ impl KvPushRouter {
...
                 Some(args.context_id),
                 args.routing_parts.token_ids,
                 args.routing_parts.block_mm_infos,
```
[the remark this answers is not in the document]
ishandhanani: [AI] Confirmed intentional. select_best_match serves lifecycle-owning push requests, including explicit or affinity-pinned requests, so update_states=true must use TrackedWithAdmission to receive the same admission progress and lease lifecycle. Query-only calls keep update_states=false and still select ScheduleMode::QueryOnly. No code change needed.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -350,26 +346,39 @@ impl KvPushRouter {
...
                 }
-            }
+            };
 
-            guard.finish().await;
+            if completed && !failed {
+                guard.finish().await;
+            } else {
+                guard.abort().await;
+            }
```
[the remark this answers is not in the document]
ishandhanani: Intentional: natural EOF is the fallback successful-completion signal, preserving the existing stream semantics. `AdmissionEvent::Completed` is defined as the response stream ending normally. A successful `finish_reason` permits early completion but is not required; cancellation or annotated/typed errors report `Aborted`.

lib/kv-router/src/scheduling/policy_queue.rs
```diff
@@ -540,11 +598,172 @@ impl<T> PolicyQueue<T> {
...
+        entry.class_index = target_class_index;
+        entry.snapshot = snapshot;
+        let target = &mut self.classes[target_class_index];
+        entry.priority.policy_score = OrderedFloat(queue_policy_score(
+            target.config.queue_policy,
+            snapshot,
+            arrival_offset_secs,
+            priority_jump,
+        ));
+        add_stats(&mut target.stats, snapshot);
```
PeaBrane, line 658: **[AI][warning] Exact reclassification bypasses the destination class’s queue limits.** `make_ready` transfers the entry and accounting into `target`, but unlike `enqueue_with_state` it never runs `queue_rejection` against that target. A roomy cached class can therefore move deferred requests into an already-full uncached class, bypassing its request/token cap. Could migration check target capacity before removing the source entry and return a defined outcome—remain deferred versus reject/abort? A full-destination test would pin the contract.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -52,19 +58,160 @@ struct QueuedRequest {
...
+        self.dirty.push(cleanup);
+        !self.pending.swap(true, AtomicOrdering::AcqRel)
+    }
+
+    fn drain(&self) -> Vec<AdmissionCleanupEntry> {
+        if !self.pending.load(AtomicOrdering::Acquire) {
+            return Vec::new();
+        }
+
+        let mut dirty = Vec::new();
```
PeaBrane, line 137: **[AI][note] Please document the cleanup fairness tradeoff.** This deliberately drains until global quiescence so no coalesced wake is lost, but sustained cancellation traffic can keep the actor in `drain` and delay enqueue/update/reconcile work. I’m okay leaving batching out of this PR; a NOTE that this is intentionally unbounded actor work would keep the assumption visible.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -52,19 +58,160 @@ struct QueuedRequest {
...
+impl AdmissionLease {
+    pub fn mark_completed(&mut self, context_tokens: usize) {
+        self.outcome = RequestOutcome::Completed { context_tokens };
+    }
+
+    pub fn mark_aborted(&mut self) {
+        self.outcome = RequestOutcome::Aborted;
+    }
+
+    pub fn mark_dispatched(&mut self) {
```
PeaBrane, line 188: **[AI][warning] The public dispatch lifecycle currently requires two calls.** This method records only the lease fallback; a live `AdmissionEvent::Dispatched` also requires awaiting `LocalScheduler::mark_dispatched(request_id)`. The push router correctly does both, but external callers can reasonably interpret this lease method as the complete notification. Could the lease become the single dispatch authority, or could the required two-step protocol be documented here and on `SchedulingResponse`?

lib/llm/src/kv_router/push_router.rs
```diff
@@ -350,26 +346,39 @@ impl KvPushRouter {
...
                     item = response_stream.next() => {
                         let Some(item) = item else {
-                            break;
+                            break true;
                         };
+                        failed |= response_item_failed(&item);
                         guard.on_item(&item).await;
+                        let completed_terminal = !failed && response_item_completed(&item);
+                        if completed_terminal {
+                            guard.mark_completed_terminal();
```
PeaBrane: **[AI][note] Could we pin the terminal-yield cancellation regression at stream level?** The control flow now correctly marks completion before yielding and exits when resumed. A test that receives `Stop`, `EoS`, or `Length` and then drops without polling EOF would protect the exact previous failure boundary.

PeaBrane (approved):
AI-assisted re-review. The earlier lifecycle and cancellation blockers are addressed, so I’m approving. I left one queue-limit semantics question and several non-blocking API/documentation/test notes inline.

lib/kv-router/src/scheduling/policy_queue.rs
ishandhanani: Queue limits are intentionally admission-time backpressure gates for incoming requests, not strict occupancy invariants. This reclassification moves an already accepted queued request rather than admitting a new one; as with worker-count shrink, the class can remain above its current per-worker limit and subsequent arrivals reject until usage drains. Enforcing the destination limit here would require defining rejected `MakeReady` feedback and terminal semantics for the strategy, so I’m keeping the existing contract in this PR.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Documented in `af3754ee7b`. The note now states that draining to quiescence preserves the coalesced-wake handoff, can delay actor commands during a large burst of already-armed lease drops, and that any bounded/interleaved alternative must preserve wake correctness and be benchmarked. Batching and bounded backpressure are unchanged.

lib/kv-router/src/scheduling/queue.rs
ishandhanani: Leaving the two-call protocol unchanged in this PR. The push router is the only production caller and correctly records the lease fallback before awaiting the bounded actor notification. Making the lease the single authority would require a generation-tagged dispatch command plus cancellation, saturation, and request-ID reuse coverage across the scheduler layers, so that is better handled as a focused follow-up rather than widening this approved PR.

lib/llm/src/kv_router/push_router.rs
ishandhanani: Added in `af3754ee7b`. The stream-level regression test receives each successful terminal reason (`Stop`, `EoS`, and `Length`), drops immediately without polling EOF, and verifies admission observes `Dispatched` followed by `Completed` rather than `Aborted`.

nv-tusharma (approved):

PR #11616, opened 2026-07-13
feat(admission): add session-aware admission policy

ishandhanani:
This stacks on #11787's generic class-local admission-policy host and replaces #11447 directly in `ai-dynamo/dynamo`. It adds Dynamo's built-in Session-Aware Admission Control policy, inspired by ThunderAgent, without per-output actor messages on the response path.

CLOSES: DYN-3231

### Summary

- Add a built-in session-aware policy with sticky worker affinity, same-session serialization, device-capacity gating, retained-context accounting, and bounded inactive-session cleanup.
- Keep the abstraction split explicit: the KV-router queue owns requests and lifecycle; an opaque policy returns `Bypass`, `Ready`, or `Defer`; final worker selection remains in the router.
- Release session accounting when the explicit final signal arrives, then forward that terminal request through the normal request/response path. Normal transport EOF behavior is unchanged outside admission-controlled traffic.
- Preserve the existing single-bucket/default-family behavior. Multi-bucket policy ownership and decoded tool-dispatch shadow events remain follow-ups.

<details>

```mermaid
flowchart LR
  R["Tracked request"] --> Q["Policy-class queue"]
  Q --> P["Session-Aware policy"]
  P -->|"bypass / ready / defer"| Q
  Q --> W["Worker selection"]
  W --> G["Response lifecycle guard"]
  G -->|"progress + terminal event"| Q
  Q --> P
```

Sessionless requests bypass the policy. Session requests retain a logical context footprint, preserve an eligible sticky worker through transient overload, serialize concurrent turns, and roll back state on cancellation or dispatch failure. Reconciliation uses per-worker eligibility and capacity, marks only enough running context to reach the pause target, and fails open when usable capacity metadata is absent.

An explicit class requires `x-dynamo-meta-policy-class: <class-name>` on each HTTP request. A default family with one `all` bucket remains supported for deployments that apply the policy to all traffic; attaching Session-Aware state to one class in a multi-bucket family is rejected to avoid splitting a session's program table.

</details>

### Validation

- `cargo test -p dynamo-admission-control --lib` — 42 passed
- `cargo test -p dynamo-kv-router --lib` — 775 passed, 2 ignored
- `cargo test -p dynamo-llm --lib kv_router::push_router::tests` — 6 passed
- `cargo check -p dynamo-llm`
- `cargo clippy -p dynamo-admission-control -p dynamo-kv-router -p dynamo-llm --all-targets -- -D warnings`
- `git diff --check`
- Matched MiniMax c256, 2×TP4, no HiCache/offload, 10-minute warmup + 3,456-second measured window at effective pause/resume thresholds 0.90/0.85:
  - Rust Session-Aware: 185.78 valid turns/min, 693.36 generated tok/s, 3.680 valid trials/min, 96.08% device KV reuse.
  - Dynamo Python ThunderAgent: 204.13 valid turns/min, 733.96 generated tok/s, 4.375 valid trials/min, 89.46% device KV reuse.
  - Both matched runs recorded zero measured stream cancellations, publish failures, disconnects, or request rejections.

## Summary by CodeRabbit

* **New Features**
  * Added session-aware admission control for queue scheduling.
  * Added configurable pause, resume, retention, and reconciliation settings.
  * Added worker-capacity-aware scheduling and session retention behavior.
  * Built-in strategy registration now supports session-aware queue policies.
  * Queue rechecks can respond to admission strategy reconciliation intervals.

* **Documentation**
  * Added configuration guidance, behavior details, and policy ownership documentation.
  * Documented known limitations and operational considerations.


lib/kv-router/src/scheduling/local.rs
```diff
@@ -443,7 +443,7 @@ where
             priority_jump,
             strict_priority,
             policy_class,
-            session_id: None,
+            admission_session: None,
```
ishandhanani: can we instead call this session_context ?

ishandhanani:
replacing with future pluggable impl

PR #11617, opened 2026-07-13
feat(bindings): expose Client.instances() with transport object model

nnshah1:
## Summary

Counter-proposal to #11540. Rather than `instance_tcp_addresses() -> Dict[int, str]` (TCP-only, address flattened to an opaque string), expose the runtime's instance/transport model to Python as read-only snapshot objects:

- `Client.instances() -> List[Instance]` — snapshot of the watched instance set, like `instance_ids()` (pair with `wait_for_instances()`)
- `Instance` — wraps a runtime `Instance`; attributes `instance_id`, `namespace`, `component`, `endpoint`, `transport`, `device_type`; `str(i)` yields `"namespace/component/endpoint/instance_id"`
- `TransportType` — wraps the runtime `TransportType`; attributes `kind` (`"tcp"` / `"nats_tcp"`) and `address`

Keeps **all** transports (NATS included) instead of dropping non-TCP, and exposes `instance_id`/`namespace`/`component`/`endpoint` as first-class attributes rather than re-parsing an address string. Names mirror the runtime types (parallel construction).

The peer-address use case (e.g. a startup readiness / RDMA connectivity check) becomes a caller-side one-liner:

```python
addrs = [i.transport.address for i in client.instances() if i.transport.kind == "tcp"]
```

## Scope

**Bindings-only.** The new pyclasses wrap runtime values through already-public API (`Client::instances()`, public `Instance` fields, `TransportType::address()`, `Display`). No changes to `lib/runtime` or any core crate.

## Credit

Builds directly on the approach, use case, and original PR by **@jyizheng** (#11540). This PR reshapes that contribution into a snapshot object model; @jyizheng is credited as co-author on the commit.

## Details

- `lib/bindings/python/rust/lib.rs` — `Instance` and `TransportType` pyclasses wrapping the runtime value with `#[getter]` delegation; `Client.instances()` moves each owned runtime `Instance` into a wrapper (no construction-time clone); both classes registered.
- `lib/bindings/python/src/dynamo/_core.pyi` — matching stubs.
- `lib/bindings/python/tests/test_request_plane_python_payload.py` — asserts `instances()` ids match `instance_ids()` and transport details are populated (validation wrapped in `try/finally` so fixture cleanup always runs).

## Relationship to #11540

Alternative shape for the same goal (#11543). Opening as a comparison artifact for discussion.

## Validation

- `cargo check`, `cargo fmt --check`, `cargo clippy` — clean
- `maturin develop --uv` builds + installs; `from dynamo._core import Instance, TransportType` works
- `pytest tests/test_request_plane_python_payload.py` — passes

## Related Issues

Relates to #11543, alternative to #11540

— Neelay + 🤖

## Summary by CodeRabbit

* **New Features**
  * Added `Client.instances()` to retrieve available endpoint instance snapshots.
  * Exposed instance details including identifiers, namespace, component, endpoint, device type, and transport metadata.
  * Added transport information with transport type and address fields.

* **Tests**
  * Added validation covering instance listings, identifiers, endpoints, and transport metadata.

### DIFF
--- a/lib/bindings/python/src/dynamo/_core.pyi
+++ b/lib/bindings/python/src/dynamo/_core.pyi
@@ -254,8 +254,10 @@ class TransportType:
     not a stable parse target.
     """
 
-    kind: str
-    address: str
+    @property
+    def kind(self) -> str: ...
+    @property
+    def address(self) -> str: ...
 
 class Instance:
     """
@@ -264,13 +266,20 @@ class Instance:
     ``"namespace/component/endpoint/instance_id"``.
     """
 
-    instance_id: int
-    namespace: str
-    component: str
-    endpoint: str
-    transport: TransportType
-    device_type: Optional[str]
-    """Device type, e.g. "cpu" or "cuda", or None if unspecified."""
+    @property
+    def instance_id(self) -> int: ...
+    @property
+    def namespace(self) -> str: ...
+    @property
+    def component(self) -> str: ...
+    @property
+    def endpoint(self) -> str: ...
+    @property
+    def transport(self) -> TransportType: ...
+    @property
+    def device_type(self) -> Optional[str]:
+        """Device type, e.g. "cpu" or "cuda", or None if unspecified."""
+        ...
 
 class Client:
     """
--- a/lib/bindings/python/src/dynamo/_core.pyi
+++ b/lib/bindings/python/src/dynamo/_core.pyi
@@ -246,6 +246,32 @@ class PyAsyncRequestStream:
     def __aiter__(self) -> "PyAsyncRequestStream": ...
     async def __anext__(self) -> JsonLike: ...
 
+class TransportType:
+    """
+    A read-only view of an instance's transport, wrapping the runtime
+    ``TransportType``. ``kind`` is the transport variant ("tcp" / "nats_tcp")
+    and ``address`` is its (transport-specific) address. The address format is
+    not a stable parse target.
+    """
+
+    kind: str
+    address: str
+
+class Instance:
+    """
+    A read-only view of a single registered instance of an endpoint, wrapping a
+    snapshot of the runtime ``Instance``. ``str(instance)`` yields
+    ``"namespace/component/endpoint/instance_id"``.
+    """
+
+    instance_id: int
+    namespace: str
+    component: str
+    endpoint: str
+    transport: TransportType
+    device_type: Optional[str]
+    """Device type, e.g. "cpu" or "cuda", or None if unspecified."""
+
 class Client:
     """
     A client capable of calling served instances of an endpoint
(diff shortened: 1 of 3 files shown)


michaelfeil (commented):
LGTM on high level

jyizheng:
LGTM.

nnshah1:
@coderabbitai full review

nnshah1:
@devin-ai-integration review this PR

nnshah1:
@dynamo-review-agent please re-review — this PR was substantially reworked since the approval on the first commit (`6dc1f0256`): the data model moved to a Tier-3 wrapper (pyclasses holding the runtime `Instance`/`TransportType` and delegating via getters) and the public classes were renamed `InstanceInfo`→`Instance`, `TransportInfo`→`TransportType`. Current head is `f6c159be`. Bindings-only; no core changes.

lib/bindings/python/tests/test_request_plane_python_payload.py
```diff
@@ -59,13 +59,30 @@ async def request_plane_client(runtime):
...
+        instances = client.instances()
+        assert {i.instance_id for i in instances} == set(client.instance_ids())
+        # On the TCP request plane every instance exposes the identity fields of
+        # the served endpoint plus a populated "tcp" transport address.
+        for instance in instances:
+            assert instance.namespace == "direct-python-msgpack"
+            assert instance.component == "backend"
+            assert instance.endpoint == "generate"
+            assert instance.device_type in (None, "cpu", "cuda")
+            assert instance.transport.kind == "tcp"
```
sttts: do we need another test for nats?

nnshah1: Good call — added `test_client_instances_snapshot` parametrized over `[tcp, nats]` (788da2cadf), so the `nats_tcp` transport branch is exercised too. Also moved the assertions out of the shared fixture into that dedicated test. Both planes pass locally.

— Neelay + 🤖

lib/bindings/python/src/dynamo/_core.pyi
```diff
@@ -246,6 +246,32 @@ class PyAsyncRequestStream:
...
 
+class TransportType:
+    """
+    A read-only view of an instance's transport, wrapping the runtime
+    ``TransportType``. ``kind`` is the transport variant ("tcp" / "nats_tcp")
+    and ``address`` is its (transport-specific) address. The address format is
+    not a stable parse target.
+    """
+
+    kind: str
```
tedzhouhk: nit: These fields are backed by read-only getters on `#[pyclass(frozen)]`, but annotating them as class attributes makes type checkers treat them as writable. Could we model `kind`/`address` and the `Instance` fields below as `@property` methods instead, so assignments rejected at runtime are also rejected statically?

tedzhouhk (approved):

lib/bindings/python/src/dynamo/_core.pyi
nnshah1: Good catch — done in 5bb5ab2b53: modeled `kind`/`address` and the `Instance` fields as read-only `@property` getters in the stub, so the `frozen` read-only contract is enforced statically too. Thanks!

— Neelay + 🤖

PR #11626, opened 2026-07-13
feat(http): Python-loadable frontend route extensions

### LINKED ISSUE #11863: DEP (light): Python-loadable frontend route extensions
## Summary

A supported way for an external Python package to add HTTP routes to the stock `python -m dynamo.frontend` — no custom binary, no from-source build — via a typed `FrontendRoute` API and an opt-in entry-point loader.

## Motivation

Downstream products (e.g. NIM) need to add management/health routes — such as a `/v1/health/ready` that reflects **live** model readiness — served on the frontend's own port alongside the inference API, without forking or rebuilding the frontend. Today that requires a custom binary. A first-class, opt-in extension surface enables it while the stock frontend keeps full ownership of arg parsing, routing, and every built-in feature. Since this ships a public Python API plus a new entry-point plugin group that third-party wheels register against, it warrants a durable design record.

## Proposal

- **Public surface:** `dynamo.llm.FrontendRoute(method, path, handler)` (constructor-validated); handlers receive a read-only, live-state context (model-manager / discovery snapshot), so responses reflect models registering/deregistering at runtime rather than a startup snapshot.
- **Discovery:** opt-in by name via `--frontend-route-extension NAME` / `DYN_FRONTEND_ROUTE_EXTENSIONS` from a `dynamo.frontend_routes` entry-point group (not arbitrary import), so nothing loads unless an installed package registered it *and* the operator named it.
- **Merge semantics:** extension routes merge through the existing frontend system-router path with `(method, path)` duplicate rejection, so an extension cannot shadow a built-in route; HTTP-only.
- **Execution:** synchronous handlers, returning JSON (⇒ 200) or `(status, body)`, run on a blocking thread so a slow extension can't pin an async worker.
- **Additive:** the existing serving path is unchanged; with no extension registered there is no runtime effect.
- **Implementation:** PR #11626.

## References

- PR #11626 (implementation + design discussion).

cc @aniskumar-nv (PR author / implementer)

aniskumar-nv:
## Overview

Add a **Python-loadable** path for the frontend's route-extension hook, so an external Python package can register additional HTTP routes on the stock `python -m dynamo.frontend` — with no custom binary and no from-source build for the consumer.

This is the wheel-consumable variant of the generic route-extension hook: the Rust core is the same as the Rust-only version, and this PR adds a thin, typed PyO3 layer plus an opt-in loader so the routes can be supplied from Python.

## Details

- `dynamo.llm.FrontendRoute(method, path, handler)` — a typed class restricted to **static `GET` routes**. The constructor rejects non-`GET` methods, non-static paths (path params `{...}`, wildcards `*`, `:`/`*`-prefixed segments, whitespace), `async def` handlers, and non-callables — so a misconfigured route is a clean error at construction, never an axum `Router` panic. Path validation is a single shared core function reused by the Rust builder and the Python bindings.
- Handlers receive a live `dynamo.llm.FrontendExtensionContext` exposing current frontend state — `has_any_ready_model()`, `serving_ready_display_names()`, `is_model_ready_to_serve(model)`, `model_display_names()`, `is_ready()`, `is_cancelled()` — so responses reflect models registering / deregistering / swapping at runtime rather than a startup snapshot.
- Extensions are loaded **opt-in** via `--frontend-route-extension VALUE` / `DYN_FRONTEND_ROUTE_EXTENSIONS`, where `VALUE` is either a name registered under the `dynamo.frontend.routes` entry-point group (preferred) or a direct `module:function` path (a registered name wins; the path fallback applies only when the value is not a name and contains `:`). Repeated values are de-duplicated (first-seen order).
- Extension routes merge through the **existing** system-router path with **method+path** duplicate-route rejection, so a consumer cannot shadow a built-in route. OpenAPI operations are merged per path, so multiple methods on one path (e.g. a built-in `GET`+`POST`, or an extension `GET` on a built-in path) all appear in `/openapi.json`.
- Handlers are synchronous and return JSON (⇒ `200`) or a `FrontendResponse(status, body)` for a custom status. They run on a **dedicated bounded worker pool**, kept off tokio's shared `spawn_blocking` pool (which the inference hot path uses for tokenization), with a bounded queue that sheds `503` when saturated and a per-request timeout; a job whose caller already gave up is skipped before it runs. So a slow, hung, or heavy extension can't starve inference.
- Extensions are HTTP-only: mode compatibility is validated before any provider code is loaded, and non-HTTP inputs reject extensions with a clear error.

Additive and HTTP-only: the existing `run_input` path is unchanged, and the stock frontend keeps ownership of arg parsing, routing, and every feature — the consumer only adds routes.

- `lib/bindings/python/rust/llm/frontend_routes.rs` — the `FrontendRoute` / `FrontendResponse` / `FrontendExtensionContext` PyO3 types, the per-request handler bridge, and the dedicated extension pool.
- `components/src/dynamo/frontend/main.py` + `frontend_args.py` — the entry-point loader and the `--frontend-route-extension` flag.
- `lib/llm/src/http/service/frontend_extension.rs` and `service_v2.rs` — the generic Rust hook and builder (the same core used by the Rust-only variant).

## Example

```python
# nim_llm/dynamo/frontend_routes.py
from dynamo.llm import FrontendRoute, FrontendExtensionContext, FrontendResponse

def _health_ready(ctx: FrontendExtensionContext):
    ready = ctx.has_any_ready_model()  # live state, not a startup snapshot
    body = {"object": "health.response",
            "message": "ready" if ready else "not ready",
            "status": "ready" if ready else "not ready"}
    return body if ready else FrontendResponse(503, body)

def nim_management_routes():
    return [FrontendRoute("GET", "/v1/health/ready", _health_ready)]
```

```toml
[project.entry-points."dynamo.frontend.routes"]
nim-management = "nim_llm.dynamo.frontend_routes:nim_management_routes"
```

```bash
python -m dynamo.frontend --http-port 8000 --frontend-route-extension nim-management
```

## Validation

- **Rust:** `cargo check -p dynamo-llm` + the bindings crate build; unit tests cover route construction/validation (static-`GET`, `:`/`*`/param rejection, within-set duplicate), method+path dedup, and OpenAPI operation merge (both methods survive on a shared path); `rustfmt --edition 2024` and clippy clean.
- **Python:** `pre-commit` (isort/black/flake8/ruff/mypy) passes; construction tests (non-`GET`, non-static paths, `async` handler, non-callable, invalid status) and standalone e2e tests for the loader + extension routes (entry-point and `module:function`, duplicate rejection).
- **End-to-end:** a real branch-built wheel in a single-container NIM-D deployment (TinyLlama) — all six NIM management routes serve on `:8000`, `/v1/health/ready` returns `FrontendResponse(503)` until a model is ready-to-serve then `200`, `/v1/models` and `/v1/chat/completions` unaffected, the dedicated extension pool threads are present, and no panics.

## Related Issues

Closes #11863


components/src/dynamo/frontend/main.py
```diff
@@ -88,6 +91,65 @@ def setup_sglang_engine_factory(
...
+    for route in routes:
+        if not isinstance(route, SystemRoute):
+            raise TypeError(
+                f"System route extension '{extension_name}' returned unsupported route "
+                f"object {route!r}; expected dynamo.llm.SystemRoute"
+            )
+    return routes
+
+
+def load_system_route_extensions(extension_names: list[str]) -> list[SystemRoute]:
```
rmccorm4: Can you share a minimal example python extension implementation you would add and how you would specify it on the CLI?

aniskumar-nv: Sure.  @rmccorm4 

1. Extension module in nim-llm:
```
# nim_llm/dynamo/frontend_routes.py
from dynamo.llm import FrontendRoute, FrontendRouteContext


def _health_live(ctx: FrontendRouteContext):
    # liveness: process is up, no model dependency
    return {"object": "health.response", "message": "live", "status": "live"}


def _health_ready(ctx: FrontendRouteContext):
    # readiness reflects LIVE frontend state, not startup config
    ready = ctx.has_any_ready_model()
    body = {
        "object": "health.response",
        "message": "ready" if ready else "not ready",
        "status": "ready" if ready else "not ready",
    }
    return body if ready else (503, body)   # dict -> 200; (status, body) -> that status


def _version(ctx: FrontendRouteContext):
    # static image metadata — reuse NIM's existing response builder
    from nim_llm.nimlib_impl import get_version_endpoint_response
    return get_version_endpoint_response()


def nim_management_routes():
    # factory: returns a FrontendRoute or an iterable of FrontendRoute
    return [
        FrontendRoute("GET", "/v1/health/live", _health_live),
        FrontendRoute("GET", "/v1/health/ready", _health_ready),
        FrontendRoute("GET", "/v1/version", _version),
        # /v1/metadata, /v1/manifest, /v1/license follow the same pattern (nimlib_impl).
        # /v1/models is intentionally NOT registered — the frontend already serves it.
    ]
```

2. Register as entry-point:
```
[project.entry-points."dynamo.frontend_routes"]
nim-management = "nim_llm.dynamo.frontend_routes:nim_management_routes"
```
3. Select it on the CLI (how NIM launches the frontend):
```
python -m dynamo.frontend \
  --router-mode kv --http-port 8000 \
  --frontend-route-extension nim-management
# equivalently: DYN_FRONTEND_ROUTE_EXTENSIONS="nim-management" python -m dynamo.frontend ...
```

components/src/dynamo/frontend/main.py
```diff
@@ -88,6 +91,67 @@ def setup_sglang_engine_factory(
...
+            )
+        provider = matches[0].load()
+        if not callable(provider):
+            raise TypeError(
+                f"Frontend route extension '{extension_name}' entry point must load a callable"
+            )
+        routes.extend(_normalize_frontend_routes(extension_name, provider()))
+
+    return routes
+
```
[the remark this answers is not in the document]
aniskumar-nv: Can you review again? It is fixed in latest commits

components/src/dynamo/frontend/main.py
```diff
@@ -295,14 +359,21 @@ def signal_handler():
 
     e = EntrypointArgs(EngineType.Dynamic, **kwargs)
     engine = await make_engine(runtime, e)
+    frontend_route_extensions = load_frontend_route_extensions(
+        config.frontend_route_extensions
+    )
+    if frontend_route_extensions and (config.interactive or config.kserve_grpc_server):
+        raise ValueError(
+            "frontend route extensions are only supported by HTTP frontend mode"
+        )
```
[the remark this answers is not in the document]
aniskumar-nv: Can you review again? It is fixed in latest commits.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,251 @@
...
+        Err(err) => {
+            tracing::error!(error = %err, "Python system route extension failed");
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(json!({"error": "system route extension failed"})),
+            )
+                .into_response()
+        }
+    }
+}
```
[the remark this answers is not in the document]
aniskumar-nv: Can you review again? It is fixed in latest commits

aniskumar-nv:
Pushed `4e29e74` — **fix: clone the Python route handler under the GIL**.

While testing this PR end-to-end with a real Python extension (NIM's management
routes) built against the wheel, the frontend panicked on first setup:

```
thread 'tokio-runtime-worker' panicked at rust/llm/frontend_routes.rs:144:
pyo3_async_runtimes.RustPanic: rust future panicked
```

Root cause: the per-route setup cloned the handler `PyObject` with a bare
`PyObject::clone` on a tokio worker thread that doesn't hold the GIL, which pyo3
panics on. Fix: clone once under the GIL via `Py::clone_ref` and share it into
the per-request axum handlers through an `Arc` (GIL-free clone).

The existing Rust unit tests register **Rust-native** handlers, so this path was
never exercised — it only fails with an actual Python handler. Worth noting for
coverage: a test that registers a Python-callable route would have caught it.

Full end-to-end validation (single-container, all six extension routes serving on
:8000, `/v1/health/ready` transitioning 503→200 on live model state, `/v1/models`
still served by the frontend) passed after the fix.

components/src/dynamo/frontend/main.py
```diff
@@ -50,6 +52,7 @@
 logger = logging.getLogger(__name__)
 
 MIN_INITIAL_WORKERS_ENV = "DYN_ROUTER_MIN_INITIAL_WORKERS"
+FRONTEND_ROUTE_ENTRYPOINT_GROUP = "dynamo.frontend_routes"
```
rmccorm4: Does this entrypoint group name have any impact/effect on end user ability to pass an arbitrary module? Does it require they structure or name their modules in a certain way?

components/src/dynamo/frontend/frontend_args.py
```diff
@@ -345,6 +346,19 @@ def add_arguments(self, parser) -> None:
             help="Dump config to the specified file path.",
         )
 
+        add_argument(
+            g,
+            flag_name="--frontend-route-extension",
```
rmccorm4, line 351: Can you add a succinct doc `docs/components/frontend/python-route-extensions.md` in https://github.com/ai-dynamo/dynamo/tree/main/docs/components/frontend that briefly explains this feature and how to use it with a minimal example?

For example, some kind of in-line
```python
<some_hello_world_python_route>
```
and
```bash
python -m dynamo.frontend --frontend-route-extension <hello_world_extension>
```
and
```bash
curl localhost:8000/hello_world
# <hello world!>
```

and link this doc in the table of high level frontend README: https://github.com/ai-dynamo/dynamo/tree/main/docs/components/frontend#next-steps

components/src/dynamo/frontend/frontend_args.py
rmccorm4, line 1: Can you add an e2e test that tests this feature? You can use some references such as this one: https://github.com/ai-dynamo/dynamo/blob/main/tests/frontend/test_vllm_prepost_integration.py

but you can replace these [frontend args](https://github.com/ai-dynamo/dynamo/blob/81b73698c6d687321e436f654b3c89f69483ff79/tests/frontend/test_vllm_prepost_integration.py#L194-L204) with the newly added frontend arg pointing to a custom route module, and have a much simpler version of this [test case](https://github.com/ai-dynamo/dynamo/blob/81b73698c6d687321e436f654b3c89f69483ff79/tests/frontend/test_vllm_prepost_integration.py#L220) that just calls the custom route and validates the expected output

components/src/dynamo/frontend/main.py
aniskumar-nv: It was a deliberate security choice. `--frontend-route-extension` takes an entry-point name, not a module path. It's resolved only against providers already registered under the fixed `dynamo.frontend_routes` group in installed packages - so it can't import an arbitrary module (passing `foo.bar:baz` just fails as an unknown name, listing the available ones). This keeps loading opt-in and explicit rather than "import whatever the CLI points at."

components/src/dynamo/frontend/frontend_args.py
aniskumar-nv: ACK, addressed in commit 9e3db67beb

components/src/dynamo/frontend/frontend_args.py
aniskumar-nv: ACK, addressed in commit 7abcc7c2c4

components/src/dynamo/frontend/main.py
```diff
@@ -88,6 +91,82 @@ def setup_sglang_engine_factory(
...
+    for route in routes:
+        if not isinstance(route, FrontendRoute):
+            raise TypeError(
+                f"Frontend route extension '{extension_name}' returned unsupported route "
+                f"object {route!r}; expected dynamo.llm.FrontendRoute"
+            )
+    return routes
+
+
+def load_frontend_route_extensions(extension_names: list[str]) -> list[FrontendRoute]:
```
nnshah1, line 175: A couple of notes on the extension-loading design — both non-blocking, but this is the cheap pre-merge window to settle them.

**1. Also accept a `module:function` path, not just registered entry-point names?**

Right now `--frontend-route-extension` / `DYN_FRONTEND_ROUTE_EXTENSIONS` only resolve against names registered in the entry-point group. Would you be open to *also* accepting a `module:function` path, with registered names taking precedence? i.e. look the name up in the group first; if there's no match and the argument is unambiguously a path (contains `:`), fall back to `importlib.import_module` + `getattr`.

The motivation: the entry-point-only restriction reads as a trust gate, but these flags are set by whoever controls the frontend's launch (the DGD / pod spec author) — and that actor already controls the process, so requiring a pre-registered name doesn't really raise the security bar against them. What it costs is the convenient ad-hoc case: a dev or a one-off deployment pointing directly at a provider that isn't packaged as a published plugin.

Supporting both is ~6 lines and keeps the ergonomics of named plugins (short names + the nice `Available: …` error) while unlocking direct paths. Gating the fallback on the `:` form avoids a typo'd name silently becoming an import attempt, and the existing HTTP-mode check still runs before anything is imported.

Is names-only a deliberate curation choice (e.g. you want the plugin surface to stay enumerable via `entry_points()`), or just the natural first cut? If deliberate, what's the concern with the path fallback?

**2. The entry-point group name (`dynamo.frontend_routes`).**

This looks like the first `dynamo.*` entry-point group in the repo, so it effectively sets the convention for future plugin surfaces — and it's a public contract the moment consumers register against it, so worth pinning down pre-merge.

Two small things: it mixes a dot and an underscore (group names conventionally use dots as separators — `console_scripts`, `flake8.extension`, `airflow.plugins`), and it's flat, so it doesn't scale into a family. Would something hierarchical read better? e.g. **`dynamo.ext.frontend.routes`** — `ext` marks it an extension point, `frontend` the component, `routes` the surface — which extends cleanly to future groups like `dynamo.ext.frontend.middleware` or `dynamo.ext.planner.policies`. A lighter variant without the explicit marker would be `dynamo.frontend.routes`.

— Neelay + 🤖

lib/llm/src/http/service/frontend_route.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl FrontendRouteContext {
+    pub(crate) fn new(state: Arc<State>) -> Self {
+        Self { state }
+    }
+
+    /// Return the shared Axum state for routes that need to install handlers
+    /// with Router::with_state.
+    pub fn state_clone(&self) -> Arc<State> {
```
nnshah1: **The extension contract differs by language — worth deciding if that's intended, and possibly renaming this type**

`FrontendRouteContext` gives native Rust extensions `state_clone() -> Arc<State>` (and `manager_clone()` / `discovery()`), so a Rust extension can reach the full internal `State` and do anything a built-in route can — the test extensions use exactly this (`context.state_clone()` + `with_state`). The Python `FrontendRouteContext` pyclass, by contrast, exposes only the read-only accessors (`has_any_ready_model()`, `serving_ready_display_names()`, etc.) with no `state_clone`. So the *effective capability* of an extension depends on which language wrote it.

If the extension surface is meant to be a supported, language-agnostic contract, that asymmetry is worth closing. Note it can only be closed by **narrowing the Rust side** — you can't meaningfully widen Python (an `Arc<State>` can't cross PyO3 usefully, and there's no `with_state` analogue), so the two converge only if Rust extensions are also confined to the read-only accessors, with the surface grown by typed accessors as needs arise. That's also the philosophy this repo already documents for the admission-strategy boundary ("expose changing system state through a narrow read-only capability instead of copying actor state"). The cost is small here — only the native test extensions use `state_clone()`, and there are no external consumers yet.

If instead the intent is just "let a consumer add a health route," the asymmetry is benign and not worth the ceremony — but then it's worth a doc line saying the Rust and Python surfaces intentionally differ.

Separately, a naming thought while this is pre-merge and has no consumers: `FrontendRouteContext` doesn't convey that it's the *extension-facing* surface (vs. the internal `State` that built-ins use), and "Route" over-narrows it — the object is live frontend state, not anything route-specific (routes are just the first extension kind that consume it). Something like **`FrontendExtensionContext`** scopes it correctly. I'd avoid folding `State` into the name (overpromises "this is the state" when it's deliberately a narrowed view). Cheap to rename now, breaking later.

Both non-blocking — just design calls that are cheapest to settle before this lands.

— Neelay + 🤖

components/src/dynamo/frontend/main.py
aniskumar-nv: ACK. Addressed in 798f6c36ef

We can now accept a module:function path, updated the docs and tests as well.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,270 @@
...
+        for route in routes.iter() {
+            docs.push(RouteDoc::new(route.method.clone(), route.path.clone()));
+            let path = route.path.clone();
+            // Clone the Python handler under the GIL (`Py::clone_ref` requires it)
+            // and share it into the per-request handlers via `Arc`, whose clone is
+            // GIL-free — the axum handler closures run on tokio workers that do not
+            // hold the GIL, so a bare `PyObject::clone` there would panic.
+            let handler = Arc::new(Python::with_gil(|py| route.handler.clone_ref(py)));
+            let route_context = context.clone();
+            router = match route.method {
```
nnshah1: **Duplicate method+path among Python extension routes panics at startup instead of returning the clean duplicate-route error**

`frontend_route_extensions_from_py` folds *all* providers' routes into a **single** `FrontendRouteSet` (line 131), whose one axum `Router` is built by repeated `Router::route` in this loop. On axum 0.7.9, adding two handlers for the same method+path **panics** rather than erroring:
```
Overlapping method route. Handler for `GET /foo` already exists
```
That panic fires while the extension callback runs — at `extension(FrontendRouteContext::new(...))` in `service_v2.rs:1080` — which is **before** `append_route_docs` (`:1085`) ever sees the docs. So the duplicate-route guard the PR built (and tests via two separate native extensions) is **bypassed** for the Python path: the frontend aborts on a Rust panic at startup instead of failing with the intended `duplicate HTTP route registered: ...` message.

I reproduced it by running a single `FrontendRouteSet` with a repeated `GET` route through `HttpService::builder().add_frontend_route_extension(...).build()` — `build()` panics with the message above.

This is realistic because **all providers collapse into that one set**: two independently-installed extensions that each register, say, `GET /metrics/custom` — or one provider that returns the same route twice — collide even though neither knows about the other. (Note: shadowing a *built-in* like `/health` is still caught cleanly, since built-in docs live in a separate set; only extension-vs-extension dups hit the panic.)

Suggested fix: dedup/validate **within** the Python route set before building the router — e.g. in `frontend_route_extension_from_routes`, track seen `(method, path)` and return a clean `PyValueError` on a collision (mirroring `append_route_docs`), or check the route isn't already present before `router.route(...)`. Worth a test for the two-providers-same-route case.

— Neelay + 🤖

lib/llm/src/http/service/frontend_route.rs
nnshah1: Related to the type name — the module `http/service/frontend_route.rs` reads as if it owns frontend routes, but it's the extension *mechanism*; the actual routes live in the sibling modules (`health`, `metrics`, …), and it's the only module here with a `frontend_` prefix. Worth renaming for the same reason. Three internally-consistent schemes, depending on how general you want the surface to be (and since nothing's published yet, the flag/env/group names can move to match whichever you pick):

| current | Opt 1 (generalize) | Opt 2 (terse) | Opt 3 (explicit) |
|---|---|---|---|
| mod `frontend_route` | `frontend_extension` | `route_extension` | `frontend_route_extension` |
| `FrontendRouteContext` | `FrontendExtensionContext` | `RouteExtensionContext` | `FrontendRouteExtensionContext` |
| `FrontendRouteSet` | `FrontendRouteSet` (stays) | `RouteExtensionSet` | `FrontendRouteExtensionSet` |
| `FrontendRouteExtension` | (stays) | `RouteExtension` | (stays) |
| `add_frontend_route_extension` | (stays) | `add_route_extension` | (stays) |

- **Opt 1** keeps `Route` only on the types that genuinely produce routes, while the module and context generalize — best if non-route extensions (middleware, lifecycle hooks) might follow.
- **Opt 2** is the tersest but still clear — drops the redundant `Frontend` (the module already lives under the frontend's `http::service`) and reads uniformly as "route extension."
- **Opt 3** is the most explicit but verbose (`FrontendRouteExtensionContext`/`...Set`).

All cheap now, pre-consumers.

— Neelay + 🤖

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Good catch! Fixed in commit 9aec1a3022

lib/llm/src/http/service/frontend_route.rs
aniskumar-nv: ACK, going ahead with `Opt 1` and narrowing the Rust context in commit fc78168b6f

rmccorm4:
/ok to test [9fa95c4](https://github.com/ai-dynamo/dynamo/pull/11626/commits/9fa95c4e07be3173ea33f578c50d3adfcda6e734)

lib/bindings/python/src/dynamo/llm/__init__.py
```diff
@@ -13,6 +13,8 @@
 from dynamo._core import FpmDirectPublisher as FpmDirectPublisher
 from dynamo._core import FpmEventRelay as FpmEventRelay
 from dynamo._core import FpmEventSubscriber as FpmEventSubscriber
+from dynamo._core import FrontendExtensionContext as FrontendExtensionContext
```
rmccorm4, line 16: *Comment generated with AI assistance*

**Blocker — the shipped typing surface does not include the new public API**

These exports are absent from `lib/bindings/python/src/dynamo/_core.pyi`, and that stub still declares the old three-argument `run_input` even though `components/src/dynamo/frontend/main.py` now passes `frontend_route_extensions` as a fourth argument. The modern `importlib.metadata.EntryPoints` branch also type-checks incorrectly because the fallback calls `.get()` after a `hasattr(..., "select")` check that mypy does not narrow. The mandatory [mypy job](https://github.com/ai-dynamo/dynamo/actions/runs/29539521443/job/87759844119) reports four errors across these two files.

Because the package ships `py.typed`, this is also a downstream compatibility issue rather than only an internal CI failure. Please add stubs for `FrontendExtensionContext` and `FrontendRoute`, add the optional fourth `run_input` parameter, and use `.select(group=...)` directly for the supported Python >=3.10 floor (or explicitly cast a legacy mapping branch).

tests/frontend/test_frontend_route_extension.py
```diff
@@ -0,0 +1,156 @@
...
+
+    with DynamoFrontendProcess(
+        request,
+        frontend_port=0,
+        extra_args=["--frontend-route-extension", extension],
+        extra_env={
+            "PYTHONPATH": pythonpath,
+            # In-process discovery: the frontend boots standalone (no etcd/NATS/
+            # worker) since the custom route does not depend on inference.
+            "DYN_DISCOVERY_BACKEND": "mem",
```
rmccorm4: *Comment generated with AI assistance*

**Blocker — all three new E2E cases fail under the CI environment**

Both subprocess paths inherit CI's ambient `NATS_SERVER` while leaving the event plane implicit. `DistributedRuntime` treats that combination as a legacy NATS opt-in even with `mem` discovery and the default TCP request plane, so the two positive cases poll for 60 seconds and the negative case exits for the wrong reason. The exact-head [parallel test job](https://github.com/ai-dynamo/dynamo/actions/runs/29539521443/job/87759844286) reports all three failures as `Failed to connect to NATS: Connection refused`.

The narrow fix is to keep this test NATS-free and pin the complete standalone configuration for **both** commands:

```python
standalone_args = [
    "--discovery-backend", "mem",
    "--request-plane", "tcp",
    "--event-plane", "zmq",
]
```

Pass `*standalone_args` in `DynamoFrontendProcess.extra_args` and in the duplicate-route `subprocess.run` command, and remove the `DYN_DISCOVERY_BACKEND` env override. Explicit `zmq` prevents an ambient `NATS_SERVER` from enabling NATS, so no runtime-services fixture is needed. Also check `frontend.is_running()` inside the readiness loop and fail with `frontend.read_logs()` when it exits; that turns startup failures into roughly one-second failures instead of two 60-second polls.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,277 @@
...
+#[pymethods]
+impl PyFrontendRoute {
+    #[new]
+    pub fn new(py: Python<'_>, method: String, path: String, handler: PyObject) -> PyResult<Self> {
+        if !handler.bind(py).is_callable() {
+            return Err(PyTypeError::new_err(
+                "FrontendRoute handler must be callable",
+            ));
+        }
+        if !path.starts_with('/') {
```
rmccorm4: *Comment generated with AI assistance*

**High — path validation still permits Axum startup panics**

The constructor validates only a leading slash, and the later `Router::route` call is infallible/panicking. With the locked Axum 0.8.4, malformed forms such as `/{} ` and `/src/{*rest}/x` panic; structurally equivalent routes such as `/items/{id}` and `/items/{name}` also conflict even though the current `(method, path)` string set considers them distinct. This is separate from the exact duplicate case already addressed.

Please validate the complete accepted path grammar before router construction and reject structurally conflicting parameter patterns with `PyValueError`. Given that handlers currently cannot observe path parameters, initially restricting this surface to static path segments would also be a coherent, lower-risk contract. Add regression tests asserting malformed paths never unwind through Rust.

rmccorm4 (commented):
*Comment generated with AI assistance*

Eight inline review findings are attached.

lib/llm/src/http/service/frontend_extension.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/// Routes and route documentation returned by a frontend route extension.
+pub struct FrontendRouteSet {
+    route_docs: Vec<RouteDoc>,
+    router: Router,
+}
+
+impl FrontendRouteSet {
+    pub fn new(route_docs: Vec<RouteDoc>, router: Router) -> Self {
```
rmccorm4: *Comment generated with AI assistance*

**High — `FrontendRouteSet` cannot enforce consistency between its router and documentation**

`FrontendRouteSet::new(Vec<RouteDoc>, Router)` accepts two independent representations. `service_v2` validates only `route_docs()` before merging the opaque router, so a caller can provide empty docs with an actual `/health` route (bypassing collision validation and then panicking during merge), serve a hidden route absent from OpenAPI, or advertise a route that returns 404.

Please make raw construction crate-private and expose an atomic typed registration/builder API that records each method/path in both the router and its `RouteDoc`. That removes the representational invariant instead of relying on every caller to keep the two values synchronized.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+}
+
+fn call_python_frontend_route_inner(
+    py: Python<'_>,
+    handler: &PyObject,
+    context: RsFrontendExtensionContext,
+) -> PyResult<(StatusCode, Value)> {
+    let py_context = Py::new(py, PyFrontendExtensionContext { inner: context })?;
```
rmccorm4, line 284: *Comment generated with AI assistance*

**High — the handler ABI discards the entire HTTP request**

The Axum closures install no request extractors, and this call supplies only `FrontendExtensionContext`. Consequently handlers registered for POST/PUT/PATCH/DELETE cannot inspect the body, headers, query parameters, path captures, or even the request method; a natural two-argument handler is accepted at registration and fails with a 500 only when called. Per-route authentication is also impossible.

Before this public callback signature ossifies, please add a bounded immutable Dynamo-owned request value as a second argument (or a per-call request view on the context). Otherwise restrict and document the initial surface as static GET-only routes.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,277 @@
...
+    let result = handler.call1(py, (py_context,))?;
+    normalize_route_response(py, result)
+}
+
+fn normalize_route_response(py: Python<'_>, result: PyObject) -> PyResult<(StatusCode, Value)> {
+    let bound = result.bind(py);
+    if bound.hasattr("__await__")? {
```
rmccorm4: *Comment generated with AI assistance*

**Medium — ordinary `async def` handlers are rejected only on first traffic**

The constructor accepts any callable, including coroutine functions. The frontend therefore starts successfully with an invalid provider, then this runtime check turns the first request into a generic 500; the created coroutine is also left unclosed and can emit an unawaited-coroutine warning. That does not match the documentation's claim that async handlers are rejected.

Please reject detectable coroutine functions during `FrontendRoute` construction/provider loading, retain this runtime awaitable guard for callable wrappers, and explicitly close an awaitable rejected by the fallback path.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,277 @@
...
+        return Err(PyTypeError::new_err(
+            "async FrontendRoute handlers are not supported; return JSON synchronously",
+        ));
+    }
+
+    if let Ok(tuple) = bound.downcast::<PyTuple>()
```
rmccorm4: *Comment generated with AI assistance*

**Medium — valid two-element JSON tuples are misclassified as status responses**

Every two-element tuple is reserved as `(status_code, body)`, so a handler returning `("a", "b")`—a JSON-serializable sequence under the documented return contract—fails `u16` extraction and becomes a 500. Checking only whether element zero is an integer would still leave integer-leading JSON tuples ambiguous.

Please use an explicit response wrapper/type for status overrides and let ordinary tuples follow the JSON-body serialization path.

docs/components/frontend/python-route-extensions.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Handler contract
+
+- **Signature:** `handler(ctx: FrontendExtensionContext)` — synchronous. Async handlers are rejected.
+- **Return:** a JSON-serializable value (implies `200`), or a `(status_code, body)` tuple to set the status.
+- **Live state:** `FrontendExtensionContext` exposes the current frontend state so responses reflect models registering/draining at runtime — e.g. `ctx.has_any_ready_model()`, `ctx.serving_ready_display_names()`, `ctx.is_model_ready_to_serve(name)`, `ctx.is_ready()`.
+
+## Notes
+
+- Select multiple extensions by repeating `--frontend-route-extension` (or via a space/comma-separated `DYN_FRONTEND_ROUTE_EXTENSIONS`). Names are de-duplicated.
```
rmccorm4: *Comment generated with AI assistance*

**Low — the page is missing from Fern navigation, and the documented comma syntax is unsupported**

The new page is not listed under Frontend in `docs/index.yml`; the repository documentation style guide states that unlisted pages are unreachable in the generated site. Separately, `DYN_FRONTEND_ROUTE_EXTENSIONS` is parsed through the shared whitespace-only `value.split()`, so `alpha,beta` is resolved as one extension name rather than two.

Please add this page to the Frontend navigation and either support splitting on commas plus whitespace with coverage, or document the environment variable as whitespace-separated only.

aniskumar-nv: Fixed in commit 8a7fd8013d — added the page under Frontend in `docs/index.yml`, and documented `DYN_FRONTEND_ROUTE_EXTENSIONS` as whitespace-separated only.

lib/llm/src/http/service/frontend_extension.rs
aniskumar-nv: Fixed in commit 8a7fd8013d — raw `FrontendRouteSet` construction is now crate-private; routes are registered atomically via `FrontendRouteSet::builder().get(path, handler)`, which records each path in both the router and its `RouteDoc` so the two can't drift.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Fixed in commit 8a7fd8013d — restricted and documented the initial surface as static GET-only. Non-GET methods are rejected at `FrontendRoute` construction, so the callback signature no longer implies request-body access it can't provide.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Fixed in commit 8a7fd8013d — coroutine functions are now rejected at `FrontendRoute` construction via `inspect.iscoroutinefunction`. The runtime awaitable guard is retained for callable wrappers and now explicitly `close()`s a rejected awaitable.

tests/frontend/test_frontend_route_extension.py
aniskumar-nv: Fixed in commit 8a7fd8013d — both commands now pin `--discovery-backend mem --request-plane tcp --event-plane zmq`, the `DYN_DISCOVERY_BACKEND` override and ambient `NATS_SERVER` are dropped, and the readiness loop fast-fails via `frontend.is_running()`. All three E2E cases pass NATS-free.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Fixed in commit 8a7fd8013d — restricted the initial surface to static paths. `validate_static_path` rejects path params `{...}`, wildcards `*`, and whitespace/control chars with `PyValueError` before router construction, with regression tests asserting malformed paths never unwind through Rust.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Fixed in commit 8a7fd8013d — status overrides now use an explicit `FrontendResponse(status, body)` wrapper; ordinary tuples follow the JSON-body path (serialized as JSON arrays), so `("a", "b")` is no longer misclassified.

lib/bindings/python/src/dynamo/llm/__init__.py
aniskumar-nv: Fixed in commit 8a7fd8013d — added `FrontendExtensionContext`/`FrontendRoute`/`FrontendResponse` stubs and the fourth `frontend_route_extensions` param on `run_input` to `_core.pyi`, and switched to `EntryPoints.select(group=...)` directly for the py>=3.10 floor. The mypy job is green.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,270 @@
...
+    })
+}
+
+async fn call_python_frontend_route(
+    handler: Arc<PyObject>,
+    context: RsFrontendRouteContext,
+) -> Response {
+    // Run the synchronous Python handler on a blocking thread so a slow
+    // extension cannot pin a tokio worker and reduce request concurrency.
+    let outcome = tokio::task::spawn_blocking(move || {
```
nnshah1: **Extension handlers share the bounded blocking pool with per-request tokenization — a slow/hung handler can degrade inference, not just its own route**

The handler runs via `spawn_blocking` on the process's tokio blocking pool. That's the **same** pool the inference hot path uses for the tokenizer BPE encode (`preprocessor.rs:1981` / `encode_batch` `:2009`), and the pool is sized to **`num_cores` by default** (`config.rs:393` → `.max_blocking_threads(...)`), not tokio's 512. So extension handlers and per-request tokenization contend for a small, shared pool.

Three properties compound:
- **No timeout** on the handler/request (the only timeouts here are graceful-shutdown drain).
- **`spawn_blocking` is non-cancellable** — if the client disconnects or the request future is dropped, the handler keeps running to completion on its thread; nothing reclaims it.
- Python handlers are additionally **GIL-serialized** among themselves.

The consequence is that the blast radius exceeds the "trusted extension" framing: a slow, hung, or merely heavy extension handler doesn't just fail its own route — it holds a scarce blocking thread indefinitely, and enough concurrent ones exhaust the pool and **starve the tokenize stage of real inference requests**, raising their latency. I confirmed the two mechanics in isolation: with a bounded pool, saturating tasks delay a new `spawn_blocking` (the "tokenizer") until a slot frees, and dropping the join handle does not stop the blocking work.

Worth considering:
- Run extension handlers on a **separate bounded pool** (or a `Semaphore`-capped subset) so they can't starve tokenization — this is the only option that actually protects the inference path.
- A `tokio::time::timeout` around the handler would at least free the *request* and return 503 on a hang — but note it won't cancel the blocking work, so it doesn't help pool exhaustion by itself.
- At minimum, document a handler-duration/non-blocking contract, since the failure mode lands on inference latency rather than the extension.

— Neelay + 🤖

nnshah1: Following up on the fix direction: I'd treat "extension routes must not contend with the inference path for the blocking pool" as a design goal, not just a doc note. These are system-tier routes, but unlike `/health`/`/metrics` (trivial, near-instant) they run arbitrary Python on the same pool per-request tokenization uses — so a "keep handlers lightweight" note is a request, not a guarantee.

Since tokio's blocking pool is per-runtime and not selectable per call, isolating it means either a **dedicated bounded executor** for extension handlers (separate runtime or a managed K-thread pool, result bridged back via a oneshot) so extension work never touches the tokenizer's pool — the strongest form — or at minimum a **`Semaphore` cap** of N permits around the `spawn_blocking`, guaranteeing tokenization always retains `pool_size − N` threads. Either bounds the blast radius so a slow/hung handler can't starve inference tokenization.

— Neelay + 🤖

aniskumar-nv: Agreed on treating it as a design goal — fixed in daaf3e8cb9 with the strongest form you described: a **dedicated bounded worker pool**, not the semaphore or a doc note.

Handlers now run on a dedicated 2-thread pool (GIL-serialized, so small suffices), off tokio's shared `spawn_blocking` pool entirely, with the result bridged back via a `oneshot` — so extension work never touches the tokenizer's pool. Bounded queue sheds `503` when full; 30s per-request timeout.

Verified e2e: 80 concurrent slow handlers → 14 shed immediately (`503`, queue full), 66 hit the 30s timeout (`503`); normal routes stay sub-ms; a panicking handler doesn't kill a worker (`catch_unwind`). The docs note is supplementary guidance, not the isolation mechanism.

Residual (your earlier #2): `spawn_blocking` isn't cancellable, so a timed-out handler runs to completion on its dedicated thread — confined there, never touching inference.

lib/llm/src/http/service.rs
```diff
@@ -33,10 +34,11 @@ pub mod realtime;
 pub mod service_v2;
 
 pub use axum;
+pub use frontend_route::{FrontendRouteContext, FrontendRouteExtension, FrontendRouteSet};
 pub use metrics::Metrics;
 
 /// Documentation for a route
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
nnshah1, line 44: Nit: the `PartialEq, Eq, Hash` derives added here are unused. `append_route_docs` keys its dedup map on `(axum::http::Method, String)`, not on `RouteDoc`, and there's no `HashSet<RouteDoc>` / `HashMap<RouteDoc, _>` / `==` on `RouteDoc` anywhere in `lib/`. `Debug` and `Clone` are used; the other three are dead.

Two ways to resolve: drop `PartialEq, Eq, Hash`, or — probably tidier — key the dedup `HashMap` on `RouteDoc` itself and drop the parallel `(Method, String)` tuple, which would put the derives to use.

— Neelay + 🤖

lib/bindings/python/rust/llm/frontend_routes.rs
nnshah1: One more structural thought (discussion, not a blocker): the natural home for that isolation might be a dedicated **extension sub-router** on the same port, rather than folding extension routes into `system_router`.

Right now extension routes are merged into `system_router` alongside `/health` and `/metrics`, which then gets a single `TraceLayer` — so there's no place to attach extension-specific policy without also affecting the built-in system routes. If extensions instead had their own `axum::Router` merged into the main app (same port, same reachability), it becomes the clean seam to hang:
- a `tower` **`ConcurrencyLimitLayer` / `LoadShedLayer`** scoped to extensions — which caps in-flight extension requests and thereby bounds how many `spawn_blocking` tasks they can launch (the semaphore-cap protection for the tokenizer pool, as a router layer);
- a **`TimeoutLayer`** for extensions only;
- a distinct request span / error handling for extension traffic.

(To be clear this is the *same-port* sub-router idea — not a separate server/port like `rl_router`, since extensions do want to live on the main API surface.)

Not asking for a redesign in this PR necessarily — just flagging that a dedicated extension router is probably where the resource-isolation guardrails belong, if/when they're added.

— Neelay + 🤖

lib/llm/src/http/service.rs
aniskumar-nv: Good catch — fixed in fb250067bc. Keyed the dedup `HashSet` on `RouteDoc` directly and dropped the parallel `(Method, String)` tuple, so route identity has a single source of truth (and the derives are now used). Behavior-identical since `RouteDoc` is exactly `{method, path}`; both duplicate-rejection tests still pass.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Agreed — a dedicated extension sub-router (same port, its own `tower` layers, mirroring `inference_router`) is the right home for HTTP-level guardrails, and it keeps that policy off the built-in `/health` / `/metrics` routes.

Since the core isolation is already handled at the thread level in this PR (dedicated bounded pool off the shared `spawn_blocking` pool, so the `ConcurrencyLimitLayer` is now largely redundant for tokenizer protection), I'd like to keep this PR focused and land the sub-router refactor as a follow-up: #11843. Captured your `ConcurrencyLimitLayer` / `LoadShedLayer` / `TimeoutLayer` + distinct-span points there.

rmccorm4:
/ok to test [fb25006](https://github.com/ai-dynamo/dynamo/pull/11626/commits/fb250067bcf8cfe335164a3442928a995e40dad3)

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,270 @@
...
+            docs.push(RouteDoc::new(route.method.clone(), route.path.clone()));
+            let path = route.path.clone();
+            // Clone the Python handler under the GIL (`Py::clone_ref` requires it)
+            // and share it into the per-request handlers via `Arc`, whose clone is
+            // GIL-free — the axum handler closures run on tokio workers that do not
+            // hold the GIL, so a bare `PyObject::clone` there would panic.
+            let handler = Arc::new(Python::with_gil(|py| route.handler.clone_ref(py)));
+            let route_context = context.clone();
+            router = match route.method {
+                Method::GET => router.route(
```
nnshah1: Handlers receive only the `FrontendRouteContext` — these closures take no extractors, so a handler never sees the request **body, query, headers, or path params**. That makes `POST`/`PUT`/`PATCH`/`DELETE` functionally identical to `GET`; the per-method `match` implies a capability that isn't there. Is that intentional MVP scope?

If request access is intended (or likely soon), it's a contained addition and cheapest now, pre-consumers, since it's a handler-signature change: add raw axum extractors to the closures (`Method`, `OriginalUri` for path+query, `HeaderMap`, `Bytes` for the body — all infallible, so no forced parse), thread the owned values into `call_python_frontend_route`, and inside the existing `spawn_blocking` + `with_gil` build a small `FrontendRequest` (pyclass or dict: method/path/query/headers/body) passed as `handler.call1((ctx, request))`. Extraction stays on the async side; Python construction stays in the blocking closure. Worth inheriting the existing body-size limit (`DYN_HTTP_BODY_LIMIT_MB`).

— Neelay + 🤖

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,270 @@
...
+    context: RsFrontendRouteContext,
+) -> PyResult<(StatusCode, Value)> {
+    let py_context = Py::new(py, PyFrontendRouteContext { inner: context })?;
+    let result = handler.call1(py, (py_context,))?;
+    normalize_route_response(py, result)
+}
+
+fn normalize_route_response(py: Python<'_>, result: PyObject) -> PyResult<(StatusCode, Value)> {
+    let bound = result.bind(py);
+    if bound.hasattr("__await__")? {
```
nnshah1: An `async def` handler is rejected here (`hasattr("__await__")`), but only *after* `handler.call1(...)` above has already produced the coroutine — which is then dropped un-awaited, so Python emits `RuntimeWarning: coroutine '…' was never awaited` alongside the 500 on every async-handler call. Detecting `inspect.iscoroutinefunction` / `__await__` on the *handler* before invoking would avoid creating the coroutine at all.

Separately (not required): a small typed response — e.g. `FrontendResponse(status, body)` — would be an alternative to the positional `(status, body)` tuple parsing just below; explicit, and sidesteps the tuple-vs-list distinction. Purely a preference — the current form is fine.

— Neelay + 🤖

lib/llm/src/http/service/frontend_route.rs
```diff
@@ -0,0 +1,88 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nnshah1: Nit: brand-new file with a `2024-2026` copyright range — should be the current year only (the sibling new file `frontend_routes.rs` correctly uses `2026`).

— Neelay + 🤖

lib/llm/src/http/service/frontend_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Register a `GET` route, recording it in both the router and the docs.
+    pub fn get<H, T>(mut self, path: impl Into<String>, handler: H) -> Self
+    where
+        H: Handler<T, ()>,
+        T: 'static,
+    {
+        let path = path.into();
+        self.route_docs
+            .push(RouteDoc::new(Method::GET, path.clone()));
+        self.router = self.router.route(&path, get(handler));
```
rmccorm4, line 125: *Comment generated with AI assistance*

**High — route registration remains a panicking configuration API**

The validation added on the Python side is not enforced by this public Rust builder, and it still has a concrete Python-side hole: `validate_static_path` accepts segments beginning with `:`, so `FrontendRoute("GET", "/:id", handler)` reaches this call and locked Axum 0.8.4 panics with `Path segments must not start with ':'`. Native callers can also panic by registering the same `GET` path twice because both `get` and `FrontendRouteExtension` are infallible.

Please centralize static-path validation and per-set deduplication in a fallible core builder, reuse that validation from Python, and let the extension callback return `Result<FrontendRouteSet>` so invalid extension configuration becomes a clean startup error. Add coverage for a colon-prefixed segment and duplicate native `GET` registration.

lib/bindings/python/rust/llm/frontend_routes.rs
```diff
@@ -0,0 +1,344 @@
...
+    (status, Json(json!({ "error": message }))).into_response()
+}
+
+async fn call_python_frontend_route(
+    handler: Arc<PyObject>,
+    context: RsFrontendExtensionContext,
+) -> Response {
+    let (tx, rx) = tokio::sync::oneshot::channel();
+    let job: ExtensionJob = Box::new(move || {
+        let result = Python::with_gil(|py| call_python_frontend_route_inner(py, &handler, context));
```
rmccorm4, line 236: *Comment generated with AI assistance*

**High — callbacks that time out while queued still execute and poison the extension pool**

The existing thread notes that an already-running callback cannot be preempted. This is a separate, avoidable case: timing out drops `rx`, but the closure remains in the `sync_channel`, and the worker later invokes it unconditionally even if it never started before its caller received `503`. With two workers and 64 queued 30-second callbacks, those 64 abandoned jobs can consume roughly another 16 minutes of pool capacity; because the executor is process-global, the backlog can also affect a subsequent frontend run in the same process.

Please carry cancellation/deadline state with each job and check `tx.is_closed()`, deadline expiry, and service cancellation immediately before acquiring the GIL. A deterministic test should fill both workers, let a queued request time out, release the workers, and verify the abandoned handler is skipped and fresh work succeeds.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -888,7 +894,39 @@ static HTTP_SVC_ANTHROPIC_PATH_ENV: &str = "DYN_HTTP_SVC_ANTHROPIC_PATH";
 pub(super) static VLLM_ENABLE_INFERENCE_V1_GENERATE_ENV: &str =
     "DYN_VLLM_ENABLE_INFERENCE_V1_GENERATE";
 
+fn append_route_docs(
+    all_docs: &mut Vec<RouteDoc>,
+    seen_routes: &mut HashSet<RouteDoc>,
+    route_docs: Vec<RouteDoc>,
+) -> Result<()> {
+    for route_doc in route_docs {
+        if let Some(existing) = seen_routes.get(&route_doc) {
```
rmccorm4, line 903: *Comment generated with AI assistance*

**Medium — a cross-method path collision can silently remove the extension route from OpenAPI**

`RouteDoc` equality includes both method and path, so extension `GET /v1/chat/completions` is accepted alongside the built-in `POST /v1/chat/completions`, and Axum serves both. However, `generate_openapi_spec` inserts a complete `PathItem` keyed only by path. Because extension docs are appended before inference docs, the later POST replaces the GET, leaving a served extension route absent from `/openapi.json`. This also contradicts the extension documentation's promise that a duplicate path is rejected.

Please either reject any extension path that overlaps a built-in path regardless of method, or merge operations into an existing OpenAPI `PathItem`. Add a regression test that registers an extension `GET` on a built-in POST path and verifies either a clean build error or both operations in the generated spec.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Both addressed: `async def` is now rejected at **construction** via `inspect.iscoroutinefunction` — before any `call1`, so no coroutine is ever created (no un-awaited-coroutine warning); the `__await__` runtime guard remains for the sync-callable-returns-a-coroutine case. And `FrontendResponse(status, body)` is now the explicit status wrapper — ordinary tuples serialize as JSON.

lib/llm/src/http/service/frontend_extension.rs
aniskumar-nv: Fixed in 91416642a5. Added a shared `validate_extension_route_path` — rejects leading `:`/`*` segments plus params/wildcards/whitespace — that both the core builder and the Python bindings call, so there's a single source of truth and no drift. The builder is now fallible: `FrontendRouteSetBuilder::get` returns `Result`, dedups within the set, and the extension callback returns `Result<FrontendRouteSet>`, so an invalid path (`/:id`) or a duplicate native `GET` is a clean startup error instead of an axum panic. Coverage: Rust unit tests for colon/`*`/param/within-set-dup, plus `/:id` and `/*rest` added to the Python construction test.

lib/llm/src/http/service/frontend_route.rs
aniskumar-nv: Fixed in 91416642a5 — the file is now `frontend_extension.rs` (renamed) with copyright `2026`; also aligned the new docs page to `2026`.

lib/llm/src/http/service/service_v2.rs
aniskumar-nv: Good catch — and it turned out to be deeper than extensions, so I fixed it by merging OpenAPI operations. Details, since the "why" matters here:

**First, is it pre-existing?** Yes. The built-in `/busy_threshold` router registers **both GET and POST** on one path (`busy_threshold.rs`), and `generate_openapi_spec` inserts a `PathItem` keyed by path only (`openapi_docs.rs`), so on `main` today one of those two methods is already dropped from `/openapi.json` — independent of this PR (our commits don't touch either file).

**That rules out the "reject any path overlap" option** — it would break that legitimate built-in GET+POST `/busy_threshold`. So I took your merge option instead: `generate_openapi_spec` now merges operations into an existing `PathItem` (`PathItem::merge_operations`) rather than overwriting. That fixes both the extension case and the pre-existing built-in one — every method on a path now appears in the spec.

Route dedup stays **method+path** (so an extension still can't shadow a built-in route — e.g. an extension `GET /health` is rejected), and I updated the docs to say a duplicate **method+path** is rejected. Added a regression test asserting GET+POST on one path both survive in the generated spec. Fixed in 91416642a5.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Fixed in 91416642a5. The job now checks `tx.is_closed()` before acquiring the GIL — a job whose caller already timed out or disconnected (dropped `rx`) is skipped instead of running to completion, so abandoned queued jobs no longer consume pool capacity. Already-running jobs still can't be preempted (`spawn_blocking` is non-cancellable), but they're confined to the dedicated pool and never touch inference.

On the test: a deterministic unit test against the process-global pyo3 executor is awkward (cdylib linking + shared global state), so this is validated behaviorally via the saturation/timeout e2e rather than a dedicated unit test — happy to add one if you'd prefer.

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Resolved by the GET-only restriction — there's no longer a per-method `match` (non-GET is rejected at construction), so no method implies a capability that isn't there. Request-body/extractor access is deliberately out of scope for this GET-only MVP; captured as a follow-up direction.

nnshah1: just double checking is it difficult to add the ability to access headers, etc. - if its small effort I think it may be worth it for future proofing - 

@rmccorm4 - defer to you here.

components/src/dynamo/frontend/main.py
nnshah1: I tend to think we could allow arbitrary via cli - I don't think security is an issue here as the cli that is launched is assumed protected - but defer

lib/bindings/python/rust/llm/frontend_routes.rs
nnshah1: small nit: we have a env variable to determine the overload signal (503 vs 529) - if possible lets use the same here.

aniskumar-nv: Good call — done in 1fa15c8433. The pool-saturation shed now signals overload via `SanitizedError::Overloaded.status()` (the shared `DYN_HTTP_OVERLOAD_STATUS_CODE`, default 529) instead of a hardcoded 503 — the same overload signal as the rest of the frontend's admission-control rejections, reusing the existing public path so it adds no new API surface. (The per-request timeout stays 503 — that's a stuck-handler condition, not overload.)

lib/bindings/python/rust/llm/frontend_routes.rs
aniskumar-nv: Quick effort read (deferring the decision to @rmccorm4) — small and **non-breaking**, ~100 lines:

- **`lib/bindings/python/rust/llm/frontend_routes.rs`** — add a `FrontendRequest` pyclass (read-only `method`/`path`/`query`/`headers`); the axum closure gains infallible extractors (`Method`, `OriginalUri`, `HeaderMap`), and the owned values are threaded into `call_python_frontend_route` and built into the request inside the existing `spawn_blocking` + `with_gil`.
- **Handler call** — `handler.call1((ctx, request))`, passed as an optional 2nd arg only to handlers that declare it (arity-detect via `inspect.signature`), so nim-llm's current `(ctx)` handlers keep working untouched — no consumer churn.
- **`lib/bindings/python/rust/lib.rs` + `_core.pyi`** — register the pyclass and add its type stub.
- **docs + tests** — document the 2-arg form; add construction/e2e coverage for request access.

Cheapest now, pre-consumers.

@rmccorm4 — land it in v1, or track as a follow-up (fold into #11843)? Your call / views.

nnshah1 (approved):
Approving from my review. The correctness/robustness and design items I raised have been addressed well:

- **Blocking-pool contention** → a dedicated bounded worker pool for extension handlers (not the shared tokenizer pool), with overload shed via the shared `DYN_HTTP_OVERLOAD_STATUS_CODE`.
- **Duplicate-route panic** → shared `validate_extension_route_path` enforced by both the Rust builder and the Python bindings.
- **Response contract** → typed `FrontendResponse` (positional-tuple ambiguity gone; tuples now serialize as JSON arrays).
- **Async handlers** → rejected at construction (no un-awaited-coroutine warning).
- **Timed-out/disconnected callers** → job skips execution before acquiring the GIL.
- **`RouteDoc` dedup** keyed on `RouteDoc` directly; module renamed to `frontend_extension.rs`; copyright fixed.

Nice work turning these around, @aniskumar-nv. Deferring the **final gating review to @rmccorm4** — his pass still has open items (typing stubs, E2E test coverage, the residual path-validation/panic thread) that should close before merge.

— Neelay + 🤖

nnshah1 (approved):
Approving from my review. The correctness/robustness and design items I raised have been addressed well:

- **Blocking-pool contention** → a dedicated bounded worker pool for extension handlers (not the shared tokenizer pool), with overload shed via the shared `DYN_HTTP_OVERLOAD_STATUS_CODE`.
- **Duplicate-route panic** → shared `validate_extension_route_path` enforced by both the Rust builder and the Python bindings.
- **Response contract** → typed `FrontendResponse` (positional-tuple ambiguity gone; tuples now serialize as JSON arrays).
- **Async handlers** → rejected at construction (no un-awaited-coroutine warning).
- **Timed-out/disconnected callers** → job skips execution before acquiring the GIL.
- **`RouteDoc` dedup** keyed on `RouteDoc` directly; module renamed to `frontend_extension.rs`; copyright fixed.

Nice work turning these around, @aniskumar-nv. Deferring the **final gating review to @rmccorm4** — his pass still has open items (typing stubs, E2E test coverage, the residual path-validation/panic thread) that should close before merge.

— Neelay + 🤖

rmccorm4 (approved):

rmccorm4:
/ok to test [b4c337c](https://github.com/ai-dynamo/dynamo/pull/11626/commits/b4c337c94a954390ba6dc230be1fd36192f6d23b)

PR #11607, opened 2026-07-13
chore(container): remove opencv-python-headless from TRT-LLM images (OPS-7625)

dmitry-tokarev-nv:
## Summary

`opencv-python-headless` is preinstalled in the TRT-LLM base via TRT-LLM's own `requirements.txt`, and its wheel vendors a set of media libraries (`libav*` under `opencv_python_headless.libs/`). Nothing in Dynamo imports cv2, and upstream removed the preinstall and made cv2 an optional video extra ([NVIDIA/TensorRT-LLM#16206](https://github.com/NVIDIA/TensorRT-LLM/pull/16206)) with its `cv2` imports already function-local in 1.3.x. Remove it from Dynamo's TRT-LLM images to slim them and drop the unused vendored media libraries.

- **Uninstall in the `runtime_full` stage**, via the **system interpreter** (`/usr/bin/python3 -m pip uninstall`). This is load-bearing: with `VIRTUAL_ENV=/opt/dynamo/venv` set image-wide, a plain `pip uninstall` resolves to the venv and exits 0 **without touching** the system-site install. Guards fail the build if `cv2` or its vendored `.libs` survive.
- **Mirror the deletion in the `pre_runtime` whiteout** — the squash `COPY --from=runtime_full / /` cannot represent deletions, so without this the package would leak back into the final image.
- **Regression test** (`tests/dependencies/test_no_opencv.py`, trtllm/pre_merge/post_merge/gpu_0) keeps a base-image bump from silently reintroducing it.

Only affects TRT-LLM images; nothing in Dynamo imports cv2. See OPS-7625 for background. Relates to the media-stack cleanup in #11628 / #11632.

## Validation

Verified against `nvcr.io/nvstaging/ai-dynamo/tensorrtllm-runtime:1.3.0-rc.7` (bundles `tensorrt_llm` 1.3.0rc19) on a GPU host:

- `/usr/bin/python3 -m pip uninstall -y --break-system-packages opencv-python-headless` → **Successfully uninstalled**; the vendored `.libs` dir gone; `import cv2` → `ModuleNotFoundError`.
- Confirmed the plain venv-`pip` no-op trap (refuses with "outside environment /opt/dynamo/venv" yet exits 0) — the reason the uninstall must use the system interpreter.
[...]

PR #11628, opened 2026-07-13
build(container): restrict the in-tree ffmpeg to a narrow media-codec allowlist (OPS-7665)

dmitry-tokarev-nv:
## Summary

The in-tree ffmpeg already restricted encoders/muxers, but it still shipped ffmpeg's **full default decoder set** (291 video + 212 audio decoders) in the `libav*.so` copied into every runtime image. Narrow it to the media formats we actually build and use.

This switches the `wheel_builder` ffmpeg configure to an **allowlist**: blanket `--disable-decoders/--disable-demuxers/--disable-parsers/--disable-protocols` plus exactly the components the two real paths need:

- **encode CLI** (imageio): rawvideo demux/decode over a pipe → `h264_nvenc`/`libvpx_vp9`
- **Rust `media-ffmpeg` VideoDecoder**: VP8/VP9 in mp4/webm/mkv

Image decode does not use ffmpeg (it goes through the Rust `image` crate), so no still-image decoders are enabled. HW **NVDEC** (`h264_cuvid`/`hevc_cuvid`) can be re-added explicitly if a decode feature ever needs H.264/H.265 (not wired today — noted in the media README).

Also bumps **ffmpeg 8.1 → 8.1.2** (an upstream maintenance release) to pick up security fixes and trim the decode surface further.

Test fixtures are re-encoded to **VP9-in-mp4**; filenames and every Rust/Python test that consumes them are unchanged.

Pairs with the compliance-side gate in #11632. See OPS-7665 for background.

## Validation

- **Built ffmpeg 8.1.2 with this exact allowlist** — `configure` accepts every component token (caught one bug: `webm` is not a separate demuxer token; `matroska` covers it) and the build succeeds LGPL-2.1.
- **Decoded all five regenerated fixtures** through that restricted build to raw YUV420p; byte counts match `frames × w × h × 1.5` exactly (2×2/10f, 320×240/1f/10f/100f, 3840×2160/10f).
- `container/render.py` renders the sglang runtime (the `enable_media_ffmpeg=true` consumer) with the new configure block intact.
- Fixtures committed as Git LFS; `native_packages.yaml` ffmpeg version bumped to match.

Linear: OPS-7665


dmitry-tokarev-nv:
Harrison — thanks for the SGLang codec removal. It lines up exactly with the rc.10 codec audit I ran: SGLang had the most codec sources of the three images (in-tree ffmpeg + a vendored `ai_dynamo_runtime.libs` copy + PyAV `av.libs` + `decord2.libs` + `opencv`), so removing ffmpeg + purging those wheels is the right call. A few follow-ups on this PR:

**CI test failures (root-caused):**
- `sglang-runtime / Test` → `test_sglang_deployment[video_e_pd_qwen]` fails with `No module named 'decord'`. SGLang's video-multimodal serving path still imports `decord`, which your commit (correctly) removed. The test needs to be removed/skipped — SGLang video decode is intentionally gone now.
- `trtllm-runtime / Test` → `test_deployment[video_diffusion]` fails: `h264_nvenc` encode→mp4 "Broken pipe", empty ffmpeg stderr. This is the diffusion video **encode** path. I can't tell from the log whether it's a regression from my decoder allowlist or pre-existing NVENC flakiness (the empty stderr leans toward the latter). It needs a GPU repro — build the allowlist ffmpeg and run the exact command: `ffmpeg -f rawvideo -i - -vcodec h264_nvenc -pix_fmt yuv420p out.mp4`. If it *is* the allowlist, the likely fix is adding `--enable-parser=h264` (a parser, not a codec — no royalty impact) so the mp4 muxer can pull SPS/PPS from the h264_nvenc stream. I'm happy to run this down if you'd like.

**Open review nits from my original commits:**
- `OPS-7665` appears in `container/templates/wheel_builder.Dockerfile:303` and `lib/llm/src/preprocessor/media/README.md:56` — Devin/CodeRabbit flag it (repo bans Linear IDs in source). Should drop both.
- `lib/llm/src/preprocessor/media/decoders/video.rs:318-320` — the ffmpeg regen command in the docstring is missing the output filename (CodeRabbit).

**Devin's `--disable-bsfs`/VP9 comment is a non-issue:** I validated VP9-in-mp4 decode against the exact allowlist build (which has `--disable-bsfs`) — all fixtures decoded to the correct frame counts. The `vp9_superframe` BSF isn't needed for the mov-demuxer + vp9-decoder + vp9-parser path.

Sibling PRs: #11607 (opencv removal) and #11632 (the compliance gate, which should merge last). Glad to take any of the above off your plate — just say the word.

dmitry-tokarev-nv:
Addressed the review comments on my commits (pushed `c45909b182`, `642de0e3e1`):

- **Removed the internal ticket references** from `wheel_builder.Dockerfile` and the media `README.md` (Devin/CodeRabbit).
- **video.rs docstring** ffmpeg regen command now includes the output filename so it's executable (CodeRabbit).
- **trtllm `video_diffusion` encode failure** (`h264_nvenc` → mp4, broken pipe): my decoder allowlist had dropped the `h264` **parser**, which the mp4 muxer needs to package the `h264_nvenc` bitstream. Added it back (`--enable-parser=vp8,vp9,h264`). A parser is not a codec — no software H.264 encode/decode is reintroduced; `h264_nvenc` stays the sanctioned HW encoder. Caveat: the failure log had an empty ffmpeg stderr, which can also indicate an NVENC/env flake, so the trtllm GPU test is the real confirmation — but the parser is a correct, zero-risk addition regardless.
- **Devin's `--disable-bsfs`/VP9 concern**: validated a non-issue — I decoded all VP9-in-mp4 fixtures with the exact allowlist build (which has `--disable-bsfs`); the `vp9_superframe` BSF isn't needed for the mov-demuxer + vp9-decoder + vp9-parser path.

**Still open (not a review comment — a consequence of the SGLang codec removal):** the SGLang `video_agg_qwen` and `video_e_pd_qwen` serve tests fail with `No module named 'decord'`. SGLang no longer has any video decoder (decord/av/ffmpeg all removed), so these two video-multimodal tests can't pass and need to be skipped or removed. Happy to push a skip with a clear reason if you'd like, @hsaturleyhal — flagging since it's your removal + implies SGLang drops video-multimodal input.

dmitry-tokarev-nv:
Added the **vLLM codec cleanup** (`eaabac23c6`) so #11632's gate can pass on vllm-runtime:

- Removes the video-**decode** wheels inherited from the `vllm-openai` base — `av` (PyAV), `opencv-python(-headless)`, `decord`/`decord2`, `torchcodec`, and **`PyNvVideoCodec`** — each of which bundles a full software ffmpeg (PyAV/decord also ship GPL libx264/libx265). Dynamo's vLLM component imports none of them.
- **Keeps** the in-tree LGPL ffmpeg + `imageio-ffmpeg` for the omni HW video-**encode** path (`h264_nvenc`, sanctioned).
- Guards fail the build if `cv2`/`av` survive.

Lower risk than the SGLang removal: **no vLLM serve/deployment test decodes video** (the `video_url` references in the vLLM tests are request-parser unit mocks + omni *output* formatting, not decode), so this shouldn't add a CI test failure. It does drop software video-**decode** input from vLLM — the same compliance trade-off as SGLang.

Note: the #11632 scan also flagged `PyNvVideoCodec` — if it's present on the **trtllm** image too (it's new since my rc.10 audit), trtllm will need the same strip. Worth a check when trtllm's scan runs.

dmitry-tokarev-nv:
Combined into #11836 (with the vLLM codec cleanup added on top). Keeping this open for the review thread + CI. **Review comments addressed** (OPS refs removed, video.rs regen command fixed, Devin `--disable-bsfs`/VP9 validated as a non-issue). **Open CI items:** `sglang Test` still fails on the removed `decord` (`video_agg_qwen`/`video_e_pd_qwen` need a skip — @hsaturleyhal, your call); `trtllm`/`vllm Test` are running (the `h264` parser fix should clear the trtllm encode failure).

dmitry-tokarev-nv:
superseeded by https://github.com/ai-dynamo/dynamo/pull/11836

PR #11408, opened 2026-07-08
feat(sglang): add native gRPC sidecar

connorcarpenter15:
## Overview:

Add a standalone Dynamo SGLang sidecar that communicates with SGLang through its native gRPC server. The sidecar supports aggregated serving and prefill/decode disaggregation while keeping the SGLang engine in a separate process.

This currently targets the gRPC disaggregation support in [sgl-project/sglang#30440](https://github.com/sgl-project/sglang/pull/30440).

## Details:

- Add the `dynamo-sglang-sidecar` Rust binary and library.
- Temporarily vendor SGLang's native gRPC contract from merge commit `cc7d6659fd68694797892d0d863b2549a5b61b69` and generate the client locally without a build-time network dependency.
- Document the vendored contract's source commit and upstream SHA-256.
- Remove the vendored copy once SGLang includes `sglang/srt/grpc/sglang.proto` in a release wheel; Dynamo can then pin and install that wheel as a build dependency and compile its packaged proto.
- Map Dynamo requests to SGLang generation parameters, including guided decoding, logprobs, LoRA, and parser metadata.
- Support prefill and decode worker modes and the bootstrap metadata required for KV transfer.
- Preserve bootstrap room precision across the sidecar boundary.
- Add aggregated and disaggregated launch scripts.
- Register the new crate in the Dynamo workspace.

### Validation

- Built the SGLang native gRPC extension from PR #30440 at `cf2aab82a3c5342f6bc89b7bae47a078ace1149c`.
- Ran an end-to-end disaggregated smoke test on two NVIDIA H200 GPUs with `Qwen/Qwen3-0.6B`.
- Verified both prefill and decode sidecars started, NIXL initialized with the UCX backend, and a routed OpenAI chat request returned HTTP 200 with a non-empty completion.
- SLURM job `3023079` completed with exit code `0:0`.
- Verified the imported proto matched the documented upstream SHA-256 before adding local headers and repository formatting, and confirmed the resulting protobuf descriptor is unchanged.
- Compiled the vendored contract with `protoc --experimental_allow_proto3_optional`.
- Ran `cargo fmt --package dynamo-sglang-sidecar -- --check`.
- Ran `cargo check -p dynamo-sglang-sidecar` through the local checkout source path before this build-only vendoring change.

## Where should the reviewer start?

- `lib/sglang-sidecar/src/engine.rs` for request handling, worker registration, and disaggregated execution.
- `lib/sglang-sidecar/src/client.rs` for the native SGLang gRPC client.
[...]

PR #11636, opened 2026-07-14
refactor(sglang): use Rust executable for sidecar

connorcarpenter15:
## Summary

- move the SGLang sidecar crate to `lib/backend/sglang-sidecar`
- keep its existing sidecar terminology:
  - package/executable: `dynamo-sglang-sidecar`
  - Rust library: `dynamo_sglang_sidecar`
  - engine type: `SglangSidecarEngine`
- keep the sidecar as a standalone Rust executable composed with `dynamo_backend_common::run`
- remove the Python native launcher binding and `dynamo.sglang_sidecar` module
- have the proposed upstream SGLang integration launch and supervise the executable directly
- leave wheel and container distribution of the executable to a follow-up PR

## Why

Follow-up to #11408 and #11623.

The sidecar is already a process boundary: it implements Dynamo's `LLMEngine` contract and talks to SGLang's native gRPC server. SGLang can supervise that Rust executable directly without importing a Python module or routing the lifecycle through a Python/GIL binding.

Transport, discovery, request conversion, health gating, cancellation, disaggregation, and the unified Dynamo worker lifecycle remain in the sidecar crate.

## Interfaces

Build and run the standalone sidecar from the Dynamo workspace:

```bash
cargo build --release -p dynamo-sglang-sidecar
./target/release/dynamo-sglang-sidecar \
    --sglang-endpoint http://127.0.0.1:30001
```

Proposed SGLang-managed interface after sgl-project/sglang#31076 accepts an executable instead of a Python module:

```bash
python3 -m sglang.launch_server \
    <args> \
    --grpc-port 30001 \
    --sidecar-executable /path/to/dynamo-sglang-sidecar
```

The SGLang supervisor should start:

```bash
/path/to/dynamo-sglang-sidecar \
    --sglang-endpoint http://127.0.0.1:30001
```

after native gRPC is listening. It should supervise the executable directly (or install parent-death handling in a child and then `execvp` it), propagate unexpected exits, send `SIGTERM` on shutdown, and force-kill the process tree only after a configurable timeout. The default timeout should exceed Dynamo's 5-second release grace plus 30-second worker shutdown timeout; 40 seconds is recommended.

## Packaging scope

This PR does not add the executable to a wheel or runtime image. Distribution will be designed and reviewed separately.

## Validation

- `cargo test -p dynamo-sglang-sidecar` — 22 unit tests, the executable-contract test, and doc tests pass
- `cargo check --manifest-path lib/bindings/python/Cargo.toml --locked --lib`
- root Cargo metadata with `--locked`
- Cargo fmt and launch-script syntax checks
- stale-name scan confirms the removed Python launcher and prior naming are absent from sidecar source, scripts, metadata, logs, and documentation

GPU end-to-end serving, the proposed SGLang executable supervisor, and executable distribution remain follow-up integration work.


.github/codeowners/areas.yaml
```diff
@@ -89,10 +89,10 @@ areas:
   github_team: '@ai-dynamo/dynamo-backend-sglang-codeowners'
   path_globs:
   - components/src/dynamo/sglang/
-  - components/src/dynamo/sglang_sidecar/
   - components/src/dynamo/sglang/request_handlers/
   - components/src/dynamo/sglang/tests/
   - examples/backends/sglang/
+  - lib/sglang-remote/
```
tanmayv25: how about `lib/backend/sglang-remote`?

tanmayv25: Similarly, we would have `lib/backend/trtllm-remote`  and `lib/backend/vllm-remote`. They will be remote backend sidecar implementations.

components/src/dynamo/sglang/sidecar.py
tanmayv25, line 1: Curious why do we need python launcher for sglang-remote backend? We are using rust executable in remote_agg.sh script right?

lib/bindings/python/rust/backend.rs
tanmayv25, line 1: same here. Do we need this in bindings? It doesn't look like we need python launcher right?

.github/codeowners/areas.yaml
connorcarpenter15: I like this change. Implemented.

components/src/dynamo/sglang/sidecar.py
connorcarpenter15: This is more for external use and not for Dynamo. @ishandhanani is working on a [sidecar launcher within SGLang](https://github.com/sgl-project/sglang/pull/31076) that would allow us to run the sidecar in the SGLang container rather than as a separate additional container, simplifying the deployment flow.

The Python bindings are here so the sidecar can be passed in to this launcher as a Python module.

lib/bindings/python/rust/backend.rs
connorcarpenter15: See above.

ishandhanani (approved):

nv-anants (approved):

components/src/dynamo/common/backend/tests/test_backend_bindings.py
```diff
@@ -45,6 +45,10 @@ def test_module_exposes_expected_classes():
         assert hasattr(backend, name), f"missing {name} on dynamo._core.backend"
 
 
+def test_module_exposes_sglang_sidecar_launcher():
+    assert callable(backend._run_sglang_sidecar)
```
connorcarpenter15: Remove this test case.

components/src/dynamo/sglang/sidecar.py
```diff
@@ -0,0 +1,21 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""SGLang-managed launcher for Dynamo's native sidecar."""
+
+import sys
+
+from dynamo._core import backend as _backend
+from dynamo.runtime.logging import configure_dynamo_logging
+
+_run_sglang_sidecar = _backend._run_sglang_sidecar
```
connorcarpenter15: Remove this alias.

components/src/dynamo/sglang/tests/test_sidecar.py
connorcarpenter15, line 1: Remove this file, tests aren't needed for this.

lib/backend/sglang-sidecar/src/args.rs
```diff
@@ -120,7 +120,20 @@ pub fn normalize_endpoint(raw: &str) -> Result<String, String> {
...
+    fn parses_sglang_managed_executable_args() {
+        let args = Args::try_parse_from([
+            "dynamo-sglang-sidecar",
+            "--sglang-endpoint",
+            "http://127.0.0.1:30001",
+        ])
+        .unwrap();
+
+        assert_eq!(args.sglang_endpoint, "http://127.0.0.1:30001");
+    }
```
connorcarpenter15: Remove test, unnecessary.

lib/backend/sglang-sidecar/src/engine.rs
```diff
@@ -63,6 +63,10 @@ impl SglangSidecarEngine {
         }
         .map_err(|err| client::invalid_arg(err.to_string()))?;
 
+        Self::from_parsed_args(args)
+    }
+
+    pub fn from_parsed_args(args: Args) -> Result<(Self, WorkerConfig), DynamoError> {
```
connorcarpenter15: Remove this change, we don't need another helper function.

lib/backend/sglang-sidecar/src/main.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Entry point for the `dynamo-sglang-sidecar` binary.
+//!
+//! Mirrors the mocker backend: parse the process CLI with clap, bootstrap-discover
+//! the engine in `from_parsed_args` (building the
+//! [`WorkerConfig`](dynamo_backend_common::WorkerConfig) `run` needs
+//! synchronously), then hand the engine to the shared runtime harness.
```
connorcarpenter15: Remove slop comment.

lib/backend/sglang-sidecar/README.md
```diff
@@ -0,0 +1,71 @@
...
+    --grpc-port 30001 \
+    --sidecar dynamo.sglang.sidecar
+```
+
+The entry point configures Dynamo logging when `main()` runs, then calls the
+private `dynamo._core.backend._run_sglang_sidecar(argv)` binding. The binding
+prepends the executable name expected by clap, releases the GIL, and runs the
+same unified worker lifecycle as the standalone executable.
+
+## SGLang-managed executable contract
```
connorcarpenter15: Remove any mention of the executable contract for now. We're only adding the module contract to SGLang for now.

lib/bindings/python/rust/backend.rs
```diff
@@ -1634,3 +1663,23 @@ fn py_err_to_dynamo(err: PyErr) -> DynamoError {
...
+                "http://127.0.0.1:30001".to_string(),
+            ]),
+            vec![
+                SGLANG_SIDECAR_PROGRAM_NAME.to_string(),
+                "--sglang-endpoint".to_string(),
+                "http://127.0.0.1:30001".to_string(),
+            ]
+        );
+    }
+}
```
connorcarpenter15: Remove slop test.

alec-flowers (approved):
Reviewed with Connor live

tanmayv25 (approved):

PR #11639, opened 2026-07-14
fix(snapshot): preserve NCCL P2P setting

galletas1712:
## Summary

- Preserve an explicitly configured `NCCL_P2P_DISABLE` value in Snapshot mode instead of always rewriting it to `0`.

Snapshot deployments may need to force NCCL through the Socket transport, for example while validating CRIU/cuda-checkpoint restore independently of GPU P2P. Overriding `NCCL_P2P_DISABLE=1` inside the process prevents the deployment from selecting that policy. The other Snapshot safety overrides remain unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot capture configuration now preserves an explicitly configured `NCCL_P2P_DISABLE` value instead of overriding it.
  * Default behavior remains unchanged when the setting is not provided.

* **Tests**
  * Added coverage to verify environment settings are preserved and configured as expected during snapshot capture.

### DIFF
--- a/components/src/dynamo/common/snapshot/lifecycle.py
+++ b/components/src/dynamo/common/snapshot/lifecycle.py
@@ -114,14 +114,7 @@ def configure_snapshot_capture_env() -> None:
         )
     os.environ["NCCL_CUMEM_ENABLE"] = "0"
 
-    nccl_p2p_disable = os.environ.get("NCCL_P2P_DISABLE")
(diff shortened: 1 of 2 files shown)


mohammedabdulwahhab (approved):

galletas1712:
/ok to test [e049c18](https://github.com/ai-dynamo/dynamo/commit/e049c18e1387295564cdd7eeef1ab3f11e36c23c)

PR #11641, opened 2026-07-14
perf(router): prototype unified active prompt tracker

jthomson04:
## Overview

This is Prototype B from the compressed block-tracking study. It replaces the per-worker prompt trackers and separate `PromptMembershipTrie` with one purpose-built, root-sharded compressed trie that owns prompt lifecycle and membership state.

This remains a draft for evaluation. The initial benchmark showed a large throughput increase, but routing projection tail latency exceeded the acceptance threshold, so the prototype is not yet a promotion candidate.

## Summary

- Add a private compressed-path arena with generational node IDs and stable request handles across edge splits and recompression.
- Add a global `UnifiedPromptTracker` with per-worker edge coverage, terminal multiplicity, fractional weights, direct routing-overlap queries, and worker teardown.
- Keep prefill and output state worker-local while sourcing prompt totals and overlap directly from the unified tracker.
- Remove the prompt-membership delta pipeline and the separately maintained membership trie from the active-sequence path.
- Preserve expiry, replica replay, load projection, output-block accounting, and worker-topology behavior.

## Details

Each first prompt hash selects an independently locked root shard. Nodes store compressed hash edges plus per-worker lifecycle metadata. Request endpoints and branch points are materialized explicitly, splits retain the suffix node ID, and unary recompression retains the child node ID so live request handles survive structural changes.

`BlockTracker` is reduced to a worker/request-local facade for output blocks. Prompt acquisition, release, fractional suffix accounting, overlap lookup, and active prompt totals are handled by `UnifiedPromptTracker`. Empty root-shard shells intentionally remain until tracker teardown.

This prototype does not reuse or modify CRTC, synthesize KV events, or add request ownership to the KV indexer.

1. `lib/kv-router/src/sequences/unified_prompt_tracker.rs`
2. `lib/kv-router/src/sequences/compressed_path_arena.rs`
3. `lib/kv-router/src/sequences/block_tracker.rs`
4. `lib/kv-router/src/sequences/prompt_registry.rs`
5. `lib/kv-router/src/sequences/multi_worker.rs`

## Related Issues

## Validation

- `cargo test -p dynamo-kv-router --lib` — 717 passed, 2 ignored
- Active-sequences benchmark: one discarded warmup plus nine measured runs per arm, eight workers, eight operation lanes, block size 2, trace-length factor 512, and 3,000 exact completed events per run.
  - Prototype B median throughput: 2,573,143,913 logical block visits/s, +670.52% over baseline; paired bootstrap 95% CI [+652.41%, +918.21%].
  - All 9/9 runs drained successfully and the deterministic lifecycle/projection transcript was byte-identical across arms.
  - Project p99/p999 regressed from 11.9/46.7 us to 24.2/180.2 us, so Prototype B does **not** meet the latency acceptance gate.
  - Add p99/p999 improved from 1024.2/2771.3 us to 59.9/116.1 us; free p99/p999 improved from 786.5/1451.3 us to 26.8/101.2 us.
  - These exploratory benchmark results were collected with one-off structural counters enabled. Those counters and unconditional lock-wait timing were removed from this PR; the algorithm and lifecycle semantics are unchanged.

No AgentX or full-frontend profile was run for this prototype comparison.

## Summary by CodeRabbit

* **New Features**
  * Unified prompt tracking across workers improves consistency when requests share prompt prefixes.
  * Prompt overlap and active-block calculations now use a centralized view.
  * Added support for fractional ownership of unique prompt suffixes.

* **Bug Fixes**
  * Improved cleanup and lifecycle handling when requests expire, are released, or workers change.
  * Preserved shared prompt ownership correctly across request releases.

### DIFF
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -1,290 +1,82 @@
 // SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
+
 use dynamo_tokens::SequenceHash;
 use rustc_hash::FxHashMap;
-#[cfg(debug_assertions)]
-use rustc_hash::FxHashSet;
-use slotmap::{SlotMap, new_key_type};
-use std::num::NonZeroU32;
 
-new_key_type! {
-    struct BlockNodeId;
-}
+use super::unified_prompt_tracker::{UnifiedPromptTracker, UnifiedRequestHandle};
+use crate::protocols::WorkerWithDpRank;
 
-/// One node in a persistent request block chain.
-///
-/// Sequence hashes identify their lineage. Parent IDs are generational arena
-/// keys, while `incoming` counts direct request-tail and child-to-parent edges.
-#[derive(Debug)]
-struct BlockNode {
-    hash: SequenceHash,
-    depth: usize,
-    parent: Option<BlockNodeId>,
-    incoming: NonZeroU32,
-}
-
-/// The single block-chain tail retained by a request.
-///
-/// This type intentionally does not implement `Clone`. New request owners must
-/// be acquired through [`BlockTracker::acquire_prompt`], which also maintains
-/// the arena ownership counts and block index.
 #[derive(Debug, Default)]
 #[must_use = "request block chains must be retained by a request and released through BlockTracker"]
 pub(super) struct RequestBlockChain {
-    tail: Option<BlockNodeId>,
-    prompt_depth: usize,
+    prompt: UnifiedRequestHandle,
+    output_hashes: Vec<SequenceHash>,
 }
 
-impl RequestBlockChain {
-    fn new(tail: Option<BlockNodeId>, prompt_depth: usize) -> Self {
-        Self { tail, prompt_depth }
-    }
-}
-
-#[derive(Debug, PartialEq, Eq)]
-pub(super) struct ReleasedPromptPath {
-    pub(super) path: Vec<SequenceHash>,
-    pub(super) remove_from: usize,
-}
-
-#[derive(Debug, Default)]
+/// Per-worker facade for request-local output state. Prompt ownership lives
+/// exclusively in the shared [`UnifiedPromptTracker`].
+#[derive(Debug)]
 pub(super) struct BlockTracker {
-    nodes: SlotMap<BlockNodeId, BlockNode>,
-    unique_blocks: FxHashMap<SequenceHash, BlockNodeId>,
-    fractional_blocks: FxHashMap<SequenceHash, f64>,
+    worker: WorkerWithDpRank,
+    prompts: Arc<UnifiedPromptTracker>,
+    output_blocks: FxHashMap<SequenceHash, f64>,
 }
 
 impl BlockTracker {
-    /// Acquire one request tail for a prompt and return the first newly present
-    /// prompt index, if any.
-    ///
-    /// # Lineage contract
-    ///
-    /// Each [`SequenceHash`] must identify exactly one prefix ancestry and
-    /// depth. As with the KV indexer, this tracker trusts callers to maintain
-    /// that invariant and does not validate the hash recurrence in production.
-    ///
-    /// # Collision scope
-    ///
-    /// Collision handling is outside the active-sequence tracker's scope: it
-    /// assumes distinct live lineages have distinct [`SequenceHash`] values and
-    /// does not detect, prevent, or recover from true hash collisions.
-    ///
-    /// `SequenceHash` is a probabilistic 64-bit identifier, so collisions are
-    /// possible but very unlikely. A collision matters here only when both
-    /// blocks are live in the same `WorkerWithDpRank` tracker: each worker and
-    /// DP rank owns a separate `BlockTracker` and hash index. Assuming uniform
-    /// hashes, an active set of `n` blocks has collision probability of roughly
-    /// `n * (n - 1) / 2^65`. Replica sync may reproduce the same collision for
-    /// that worker on peer routers, but cannot alias arena nodes across workers.
-    pub(super) fn acquire_prompt(
-        &mut self,
-        sequence: &[SequenceHash],
-    ) -> (RequestBlockChain, Option<usize>) {
-        let Some(&tail_hash) = sequence.last() else {
-            return (RequestBlockChain::default(), None);
-        };
-
-        if let Some(tail) = self.live_node_id(tail_hash) {
-            self.debug_assert_lineage(tail, sequence);
-            self.increment_incoming(tail);
-            return (RequestBlockChain::new(Some(tail), sequence.len()), None);
+    pub(super) fn new(worker: WorkerWithDpRank, prompts: Arc<UnifiedPromptTracker>) -> Self {
+        prompts.ensure_worker(worker);
+        Self {
+            worker,
+            prompts,
+            output_blocks: FxHashMap::default(),
         }
-
-        let mut parent = None;
-        let mut first_new_idx = 0;
-
-        // Prompt liveness is prefix-closed. Search backward for the deepest
-        // live prefix, then construct only the missing suffix.
-        for idx in (0..sequence.len().saturating_sub(1)).rev() {
-            if let Some(node_id) = self.live_node_id(sequence[idx]) {
-                self.debug_assert_lineage(node_id, &sequence[..=idx]);
-                parent = Some(node_id);
-                first_new_idx = idx + 1;
-                break;
-            }
-        }
-
-        let missing = sequence.len() - first_new_idx;
-        self.nodes.reserve(missing);
-        self.unique_blocks.reserve(missing);
-        self.debug_assert_new_hashes(&sequence[first_new_idx..]);
-
-        // Adding the first new child contributes one new incoming edge to the
-        // reused prefix. Subsequent construction transfers the provisional
-        // request-tail edge into a child-to-parent edge without changing it.
-        if let Some(parent_id) = parent {
-            self.increment_incoming(parent_id);
-        }
-
-        for (idx, &hash) in sequence[first_new_idx..].iter().enumerate() {
-            let node_id = self.nodes.insert(BlockNode {
-                hash,
-                depth: first_new_idx + idx + 1,
-                parent,
-                incoming: NonZeroU32::MIN,
-            });
-            let previous = self.unique_blocks.insert(hash, node_id);
-            debug_assert!(
-                previous.is_none(),
-                "new block unexpectedly replaced a live index entry"
-            );
-            parent = Some(node_id);
-        }
-
-        (
-            RequestBlockChain::new(parent, sequence.len()),
-            Some(first_new_idx),
-        )
     }
 
-    /// Append a unique output block to an existing request chain.
+    pub(super) fn acquire_prompt(&mut self, sequence: &[SequenceHash]) -> RequestBlockChain {
+        RequestBlockChain {
+            prompt: self.prompts.acquire(self.worker, sequence),
+            output_hashes: Vec::new(),
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    fn apply_total_delta(&self, worker: WorkerWithDpRank, delta: f64) {
+        if delta == 0.0 {
+            return;
+        }
+        let mut totals = self.worker_totals.write();
+        let total = totals.entry(worker).or_insert(0.0);
+        *total += delta;
+        assert!(
+            *total >= -f64::EPSILON,
```
jthomson04, line 448: **Production panic from float drift.** This is a release-mode `assert!`, and `f64::EPSILON` (≈2.2e-16) is an *absolute* tolerance. Totals are maintained incrementally (`± edge.len() * fraction`), and `decay_fraction` from `add_output_block` is an arbitrary float, so individual updates round (e.g. `4.0 − 4·0.9 − 4·0.1 ≈ −2.8e-17` already). With totals in the thousands of blocks, accumulated rounding error sits orders of magnitude above epsilon, and when a worker drains to zero the residual can land negative → router panic under normal fractional workloads.

The old design was immune because it recomputed `count − 1.0 + frac` from scratch per query. Suggest clamping to 0 with a magnitude-scaled tolerance (or `debug_assert` + clamp). The exact-`0.0` comparisons in `is_empty()` have the same drift problem (test/bench flakiness).

lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+pub(super) struct UnifiedRequestHandle {
+    shard: Option<Arc<RootShard>>,
+    tail: Option<CompressedNodeId>,
+    prompt_depth: usize,
+}
+
+/// Authoritative request-lifecycle and cross-worker prompt-membership trie.
+#[derive(Debug, Default)]
+pub(super) struct UnifiedPromptTracker {
+    roots: RwLock<FxHashMap<SequenceHash, Arc<RootShard>>>,
```
jthomson04, line 35: **Unbounded growth of the root-shard map.** Entries here are never removed (the empty-shell retention is documented as intentional), so every distinct first-block hash ever seen leaks a permanent `Arc<RootShard>` (lock + empty maps). Diverse traffic produces unbounded distinct first hashes over a router's lifetime, and the old trie's cleanup path is now a no-op stub. Needs either GC of empty shards or an explicit memory-bound argument before promotion.

lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+    worker_totals: RwLock<FxHashMap<WorkerWithDpRank, f64>>,
```
jthomson04, line 36: `apply_total_delta` takes this **global** write lock on every acquire/release/fraction update *while still holding the shard's arena write lock*, coupling all shards through one lock and lengthening the shard critical section. Per-worker atomic/padded cells (or at least dropping the arena guard first) would decouple this.

lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub(super) fn acquire(
+        &self,
+        worker: WorkerWithDpRank,
+        sequence: &[SequenceHash],
+    ) -> UnifiedRequestHandle {
+        let Some(&first_hash) = sequence.first() else {
+            return UnifiedRequestHandle::default();
+        };
+        let shard = self.root_shard(first_hash);
+        let mut arena = shard.arena.write();
```
jthomson04, line 53: **Hot-shard serialization is structural, and it's what failed the latency gate.** Shard = first block hash, so a fleet-wide shared system prompt maps *every request on every worker* onto this one write lock — previously these mutations only serialized per worker slot. Projections take `arena.read()` on the same lock, and parking_lot's RwLock is write-preferring, so readers queue behind the mutation stream. That is a sufficient explanation for the projection p99/p999 regression (11.9→24.2 µs, 46.7→180.2 µs) — and the benchmark itself used a shared `[1, 2, …]` root. Splitting shards deeper than the first hash, or making projection reads lock-free (seqlock/epoch snapshot per shard), looks like a prerequisite for promotion.

lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+            .unwrap_or_else(|| panic!("unified prompt {kind} underflowed"));
+        if *count == 0 {
+            entries.remove(&worker);
+            true
+        } else {
+            false
+        }
+    }
+
+    pub(super) fn set_unique_suffix_fractional(
```
jthomson04, line 210: **No direct test coverage.** This is the trickiest new logic — per-worker "incoming" computed from terminals + covered-children, the fraction/recompression equality interplay, and the incremental totals math — and `unified_prompt_tracker`'s test module never calls it. Nothing tests recompression being *blocked* by a fraction mismatch, or acquiring through a node that already carries a fraction.

lib/kv-router/src/sequences/unified_prompt_tracker.rs
```diff
@@ -0,0 +1,624 @@
...
+                node.metadata.coverage.remove(&worker);
+                node.metadata.terminals.remove(&worker);
+                node.metadata.fractions.remove(&worker);
+            }
+            Self::compact_after_worker_removal(&mut arena);
+        }
+        self.worker_totals.write().remove(&worker);
+    }
+
+    fn compact_after_worker_removal(arena: &mut CompressedPathArena<UnifiedNodeMetadata>) {
```
jthomson04, line 355: **O(n²) compaction under the shard write lock.** Each loop iteration does a full `arena.nodes.iter().find_map(...)` scan to remove one leaf or perform one merge, so worker teardown on a large arena is quadratic — and it stalls all routing on that shard for the duration. Collecting candidates in one pass and cascading from parents would make this linear.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -1,865 +1,139 @@
...
-            hash,
-            depth,
-            parent,
-            incoming: NonZeroU32::MIN,
-        });
-        let previous = self.unique_blocks.insert(hash, node_id);
-        debug_assert!(
-            previous.is_none(),
-            "new output unexpectedly replaced a live index entry"
+        assert!(
```
jthomson04, line 46: This collision check was promoted from `debug_assert!` (old tracker) to a release-mode `assert!`: a random 64-bit output-hash collision now panics production instead of being tolerated. Probability is negligible (~n²/2⁶⁵ per worker), but it's a behavior change the PR description doesn't call out — worth making deliberate either way.

lib/kv-router/src/sequences/compressed_path_arena.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub(super) children: FxHashMap<SequenceHash, CompressedNodeId>,
+    pub(super) metadata: M,
+}
+
+/// A single-owner compressed path forest with stable generational handles.
+#[derive(Debug)]
+pub(super) struct CompressedPathArena<M> {
+    pub(super) nodes: SlotMap<CompressedNodeId, CompressedPathNode<M>>,
+    pub(super) roots: FxHashMap<SequenceHash, CompressedNodeId>,
+    pub(super) splits: u64,
```
jthomson04, line 25: `splits` and `merges` are updated (`saturating_add`) but never read — leftovers from the one-off structural counters the PR description says were removed. Delete the fields and the two increment sites.

lib/kv-router/src/sequences/prompt_registry.rs
```diff
@@ -76,24 +60,6 @@ impl WorkerLoadSnapshot {
...
-    // technically UB under Rust/LLVM if a writer mutates the same memory at the
-    // same time: volatile is not atomic and does not by itself legalize a data
-    // race. See https://github.com/Amanieu/seqlock/issues/2#issuecomment-473606523.
-    //
-    // `WorkerLoadSnapshot` is a small `Copy` value with no references or drop
-    // state. The sequence check is what makes the read logically race-free for
-    // this slot: a reader either observes a stable whole snapshot or retries.
-    // In practice this is the same seqlock/READ_ONCE pattern the crate is
-    // designed for, while avoiding the per-worker RwLock read on every request.
     load: SeqLock<WorkerLoadSnapshot>,
```
jthomson04, line 63: The SeqLock soundness comment that used to live here (the volatile-read UB caveat and why the sequence check makes the pattern sound for this `Copy` payload) was deleted without any change to the code it documents. Same for the `WorkerLoadProjection`/`WorkerLoadSnapshot` field docs. The eventual-consistency WARNING is rightly obsolete now that membership is authoritative, but the rest should be restored — it's load-bearing safety context.

lib/kv-router/src/sequences/prompt_registry.rs
```diff
@@ -252,7 +172,6 @@ impl PromptRegistry {
     pub(super) fn maybe_cleanup(&self) {
```
jthomson04, line 172: `maybe_cleanup` is now an empty stub in non-test builds (only the test-only `cleanup_attempts` counter remains) but is still called from the hot path. Either remove it and its call sites or leave a comment explaining what future cleanup is expected to live here.

lib/kv-router/src/sequences/topology.rs
```diff
@@ -86,12 +88,15 @@ pub(super) struct WorkerSlot {
...
-        } else {
-            ActiveSequences::new_without_expiry(block_size)
-        };
+    fn new(
+        worker: WorkerWithDpRank,
+        block_size: usize,
+        expiry_enabled: bool,
+        prompts: Arc<UnifiedPromptTracker>,
+    ) -> Self {
+        let expiry_duration = expiry_enabled.then_some(std::time::Duration::from_secs(300));
```
jthomson04, line 97: This hardcodes the 5-minute expiry while `EXPIRY_DURATION` in `single.rs` became `#[cfg(test)]`-only. Keep one shared named constant so the two can't drift.

jthomson04 (commented):
Self-review pass over Prototype B (Claude-assisted). Inline comments carry the specific findings; summary below.

## What looks solid

- The handle-stability invariants are the hard part of this design and they hold up under tracing: splits retain the suffix node ID, unary merges retain the child ID, and the terminals-nonempty guard prevents a live handle's tail from ever being merged away, so `path_from_root(tail)` always reconstructs exactly `prompt_depth` blocks.
- Lock ordering (`roots` → shard `arena` → `worker_totals`) is consistent across all paths; no deadlock.
- Overlap semantics are equivalent to the old membership trie (coverage is whole-edge and prefix-closed), and making membership authoritative-at-mutation removes the documented eventual-consistency read-skew window — a real improvement over the delta pipeline.
- Acquire/release/fraction delta accounting balances to zero in every interleaving I traced — mathematically. Floating-point is another matter (see inline comment on `apply_total_delta`).

## Test coverage — the weakest part of the PR

The deleted suites were the strongest ones, and the replacements are much thinner:

- The 10k-step randomized differential test against a reference model, the 65k-deep chain test, dp-rank isolation, worker-removal registry, and naive-projection comparison tests are all gone.
- `set_unique_suffix_fractional` on the unified tracker has no direct test at all — and it is the trickiest new code (per-worker "incoming" from terminals + covered children, fraction/recompression equality interplay, incremental totals math). Nothing tests recompression being blocked by a fraction mismatch, or acquiring through a node that already carries a fraction.
- Recommendation: port the randomized reference-model test to `UnifiedPromptTracker` — drive multi-worker acquire/release/fraction interleavings against a naive per-worker hash-set + fraction-map model, asserting both `active_block_weight` and `compute_overlap_depths` at every step. That test would likely have caught the float-drift issue flagged inline.

## Verdict

Fine as an evaluation prototype, but before any promotion: fix the `f64::EPSILON` assert (production crash risk that also invalidates fractional-workload benchmarks), decide shard GC for the root map, address the hot-shard write serialization (the structural cause of the projection p99/p999 regression that failed the latency gate), and restore fuzz-style differential coverage.

jthomson04:
Superseded by #11644

PR #11644, opened 2026-07-14
perf(kv-router): compress block tracker prompt paths

jthomson04:
## Summary

- Replace each worker's per-block prompt arena and reverse hash map with compressed radix-tree edges while preserving stable request handles across splits.
- Keep the arena split-only during request lifetimes: release removes dead suffixes but does not eagerly recompress retained unary paths.
- Use the trusted `SequenceHash` lineage contract to check the boundary hash before scanning a compressed edge during prompt acquisition.
- Keep active-block load accounting O(1) with running prompt and output totals, and avoid reconstructing a prompt path when release removes no prompt blocks.
- Preserve prompt-membership deltas, fractional accounting, output bookkeeping, expiry, replica replay, and routing behavior. `PromptMembershipTrie`, `PromptRegistry`, CRTC, KV events, and public APIs are unchanged.

## Validation

- `cargo test --locked -p dynamo-kv-router --lib block_tracker` (13 passed)
- `cargo test --locked -p dynamo-kv-router --lib` (741 passed, 2 ignored)
- `cargo check --locked --release -p dynamo-kv-router`
- `cargo clippy --locked -p dynamo-kv-router --lib -- -D warnings`
- Optimized Python bindings build and matched AgentX frontend end-to-end run

### End-to-end AgentX block-size-1 profile

The comparison repeats the original accepted methodology exactly: all 336 AgentX/Weka entries, concurrency 256, 32 warmups, 120 measured seconds, eight block-size-1 mock workers, frontend pinned to CPU 0, mockers and AIPerf on CPUs 1-23, fastokens with a 1 GiB tokenizer cache, jemalloc, and a 40-second `cycles:u`/99 Hz/DWARF frontend capture after 40 seconds of settling.

| Metric | Original baseline | This PR | Change |
|---|---:|---:|---:|
| Input tokens/s | 1,146,733 | 1,143,492 | -0.28% |
| Mean request latency | 15,268 ms | 15,215 ms | -0.35% |
| Frontend process CPU | 61.02% | 39.48% | **-35.30%** |
| Frontend CPU 0 aggregate busy | 66.47% | 41.11% | **-38.16%** |
| TTFT p50 | 487 ms | 398 ms | **-18.28%** |
| TTFT p95 | 2,365 ms | 1,587 ms | **-32.91%** |
| TTFT p99 | 3,054 ms | 2,216 ms | **-27.45%** |
| Frontend cycles with active-sequence frames | 25.79% | 2.04% | **-23.75 pp** |
| Absolute sampled active-sequence cycle weight | 24.67B | 1.11B | **-95.51%** |
| Total sampled frontend cycle weight | 95.68B | 54.27B | **-43.28%** |

The old per-block `HashMap::insert`, `RawTable::remove_entry`, and `BlockTracker::live_node_id` leaves disappear from the candidate's top self-time list. End-to-end input throughput and mean latency are effectively unchanged while frontend CPU, TTFT, and active-sequence tracking time fall substantially.

Both arms remain overloaded profiles rather than clean capacity measurements, and this is one matched run per arm. Event-plane lag warnings fell from 231 to 103 and skipped messages from 86,349 to 38,877. In the same run, output tokens/s was 1.43% lower and request-latency p99 was 1.09% higher.

### DIFF
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -103,12 +103,20 @@ impl BlockTracker {
             let (edge_len, match_len) = {
                 let node = &self.arena.nodes[node_id];
                 let remaining = &sequence[path_pos..];
-                let match_len = node
-                    .edge
-                    .iter()
-                    .zip(remaining)
-                    .take_while(|(left, right)| left == right)
-                    .count();
+                let probe_len = node.edge.len().min(remaining.len());
+                debug_assert!(probe_len > 0, "compressed path probe cannot be empty");
+                // A SequenceHash commits to its full prefix, so equality at
+                // the probe boundary implies equality before it under the
+                // trusted lineage contract documented above.
+                let match_len = if node.edge[probe_len - 1] == remaining[probe_len - 1] {
+                    probe_len
+                } else {
+                    node.edge
+                        .iter()
+                        .zip(remaining)
+                        .position(|(left, right)| left != right)
+                        .unwrap_or(probe_len)
+                };
                 (node.edge.len(), match_len)
             };
             assert!(match_len > 0, "compressed path selected a mismatched edge");
@@ -214,7 +222,6 @@ impl BlockTracker {
             tail_node.metadata.terminal_owners != 0 || !tail_node.children.is_empty()
         };
         if tail_remains_live {
-            self.recompress_from(tail);
             return None;
         }
 
@@ -231,11 +238,9 @@ impl BlockTracker {
 
         let mut removed_hashes = 0;
         let mut current = Some(tail);
-        let mut retained = None;
         while let Some(node_id) = current {
             let node = &self.arena.nodes[node_id];
             if node.metadata.terminal_owners != 0 || !node.children.is_empty() {
-                retained = Some(node_id);
                 break;
             }
             let parent = node.parent;
@@ -248,10 +253,6 @@ impl BlockTracker {
 
         if self.arena.nodes.is_empty() {
             self.prompt_total = 0.0;
-        }
-
-        if let Some(node_id) = retained {
-            self.recompress_from(node_id);
         }
 
         debug_assert!(removed_hashes > 0, "dead prompt tail removed no hashes");
@@ -309,25 +310,6 @@ impl BlockTracker {
                 .nodes
                 .values()
                 .any(|node| node.edge.contains(hash))
-    }
-
-    fn recompress_from(&mut self, mut node_id: CompressedNodeId) {
-        loop {
-            let child_id = {
-                let node = &self.arena.nodes[node_id];
-                if node.metadata.terminal_owners != 0 || node.children.len() != 1 {
-                    return;
-                }
-                node.children.values().next().copied().unwrap()
-            };
-            if self.arena.nodes[node_id].metadata.fraction
-                != self.arena.nodes[child_id].metadata.fraction
-            {
-                return;
-            }
-            self.arena.merge_parent_into_child(node_id, child_id);
-            node_id = child_id;
-        }
     }
 
     #[cfg(any(test, debug_assertions))]
@@ -514,7 +496,7 @@ mod tests {
     }
 
     #[test]
-    fn branch_release_recompresses_while_preserving_survivor_handle() {
+    fn branch_release_preserves_split_and_survivor_handle() {
         let mut tracker = BlockTracker::default();
         let (left, _) = tracker.acquire_prompt(&[1, 2, 3]);
         let (right, first_new) = tracker.acquire_prompt(&[1, 2, 4]);
@@ -522,6 +504,7 @@ mod tests {
 
         assert_eq!(first_new, Some(2));
         assert_eq!(tracker.release(left), released(&[1, 2, 3], 2));
+        assert_eq!(tracker.arena.nodes.len(), 2);
         assert!(tracker.arena.nodes.contains_key(right_tail));
         tracker.assert_consistent([&right]);
         assert_eq!(tracker.release(right), released(&[1, 2, 4], 0));
@@ -569,7 +552,7 @@ mod tests {
     }
 
     #[test]
-    fn fraction_mismatch_prevents_recompression() {
+    fn fraction_mismatch_preserves_split_and_totals() {
         let mut tracker = BlockTracker::default();
         let (longer, _) = tracker.acquire_prompt(&[1, 2, 3, 4]);
         let longer_tail = longer.prompt_tail.unwrap();
@@ -591,7 +574,7 @@ mod tests {
     }
 
     #[test]
-    fn equal_fractional_edges_recompress_and_preserve_child_handle() {
+    fn equal_fractional_edges_remain_split_and_preserve_child_handle() {
         let mut tracker = BlockTracker::default();
         let (longer, _) = tracker.acquire_prompt(&[1, 2, 3, 4]);
         let longer_tail = longer.prompt_tail.unwrap();
@@ -601,7 +584,7 @@ mod tests {
         assert_eq!(tracker.arena.nodes.len(), 2);
         assert_eq!(tracker.prompt_total, 2.0);
         assert_eq!(tracker.release(shorter), None);
-        assert_eq!(tracker.arena.nodes.len(), 1);
+        assert_eq!(tracker.arena.nodes.len(), 2);
         assert!(tracker.arena.nodes.contains_key(longer_tail));
         assert_eq!(tracker.prompt_total, 2.0);
         tracker.assert_consistent([&longer]);
--- a/lib/kv-router/src/sequences/block_tracker.rs
+++ b/lib/kv-router/src/sequences/block_tracker.rs
@@ -26,6 +26,10 @@ impl BlockNodeMetadata {
             .checked_add(1)
             .expect("block ownership count overflowed");
     }
+
+    fn edge_weight(self, edge_len: usize) -> f64 {
+        edge_len as f64 * self.fraction.unwrap_or(1.0)
+    }
 }
 
 /// The compressed prompt tail plus request-local output hashes.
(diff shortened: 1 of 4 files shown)


lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +56,335 @@ pub(super) struct ReleasedPromptPath {
...
+                .remove(&hash)
+                .expect("request output hash is missing from active bookkeeping");
+        }
+
+        let Some(tail) = chain.prompt_tail else {
+            assert_eq!(chain.prompt_depth, 0, "empty prompt retained a depth");
+            return None;
+        };
+
+        let path_ids = self.arena.path_from_root(tail);
```
jthomson04, line 228: `release` now builds the full O(prompt_depth) path `Vec` (walk + hash copy) on **every** release, then discards it when `removed_hashes == 0` — the common case for shared/duplicate prompts. The old code returned `None` before constructing the path.

Since the path must be captured before removal, you can decrement the terminal first and check whether the tail will actually die (`terminal_owners == 0 && children.is_empty()`); if not, recompress and return `None` without building the path. For long contexts (128k tokens → ~2k hashes) this is a real per-release allocation.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +56,335 @@ pub(super) struct ReleasedPromptPath {
...
+            if incoming != 1 {
                 break;
             }
-            self.fractional_blocks.insert(node.hash, fraction);
             current = node.parent;
+            self.arena.nodes[node_id].metadata.fraction = Some(fraction);
         }
     }
 
     pub(super) fn active_blocks(&self) -> usize {
```
jthomson04, line 301: This went from ~O(1 + #fractional) to O(#arena nodes + #output blocks), and it's called via `worker_load_snapshot()` on every mutation — including once per appended output block — plus replica sync. The `output_blocks.values().sum()` term scales with total in-flight output blocks per worker. This is almost certainly the 11.9 → 14.2 µs projection p99 regression in the PR description.

Suggestion: maintain running `prompt_total`/`output_total` f64 sums updated incrementally at acquire/release/append/`set_unique_suffix_fractional` (splits and merges don't change totals; the fractional walk adjusts by `edge.len() * (new - old)`). That restores O(1) and would likely clear the all-latency gate this draft currently fails.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +56,335 @@ pub(super) struct ReleasedPromptPath {
...
+            .arena
+            .nodes
+            .values()
+            .map(|node| node.edge.len() as f64 * node.metadata.fraction.unwrap_or(1.0))
+            .sum::<f64>();
+        let output_blocks = self.output_blocks.values().sum::<f64>();
+        (prompt_blocks + output_blocks).round() as usize
     }
 
     pub(super) fn contains_block(&self, hash: &SequenceHash) -> bool {
```
jthomson04, line 306: `contains_block` is now an O(total live hashes) linear scan that looks like a cheap lookup. Its only production caller is inside a `debug_assert!` in `single.rs`, so nothing is broken today, but the signature invites a future hot-path caller. Consider gating it `#[cfg(any(test, debug_assertions))]` or documenting the cost.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +56,335 @@ pub(super) struct ReleasedPromptPath {
...
-    /// `SequenceHash` is a probabilistic 64-bit identifier, so collisions are
-    /// possible but very unlikely. A collision matters here only when both
-    /// blocks are live in the same `WorkerWithDpRank` tracker: each worker and
-    /// DP rank owns a separate `BlockTracker` and hash index. Assuming uniform
-    /// hashes, an active set of `n` blocks has collision probability of roughly
-    /// `n * (n - 1) / 2^65`. Replica sync may reproduce the same collision for
-    /// that worker on peer routers, but cannot alias arena nodes across workers.
+    /// Acquire a request terminal and return the first block that became newly
+    /// present for this worker.
     pub(super) fn acquire_prompt(
```
jthomson04, line 85: The old doc contract here spelled out the lineage (prefix-closed liveness, one ancestry per hash) and collision-scope assumptions the tracker trusts callers for. The compressed walk relies on exactly the same assumptions (root/child lookup by first hash). Worth keeping a condensed version — that trust boundary is worth having in writing.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +56,335 @@ pub(super) struct ReleasedPromptPath {
...
-        self.live_node_id(*hash).is_some()
+        self.output_blocks.contains_key(hash)
+            || self
+                .arena
+                .nodes
+                .values()
+                .any(|node| node.edge.contains(hash))
+    }
+
+    fn recompress_from(&mut self, mut node_id: CompressedNodeId) {
```
jthomson04: The fraction-mismatch guard here has no test coverage — the randomized model never calls `set_unique_suffix_fractional`, and no unit test hits the path where a release wants to merge a unary parent whose fraction differs from its child's.

Suggested test: A acquires `[1,2,3,4]`; B acquires `[1,2]` (split); A marks fractional; B releases → assert no merge and correct `active_blocks`; A releases → full path removed. A merge-with-equal-`Some(f)` case would also be cheap to add.

lib/kv-router/src/sequences/compressed_path_arena.rs
```diff
@@ -0,0 +1,292 @@
...
+/// A single-owner compressed path forest with generational node handles.
+///
+/// The existing node ID always stays with the suffix during a split. During a
+/// unary merge the child ID survives. Those two rules keep request-held tail
+/// handles valid across unrelated insertions and removals.
+#[derive(Debug)]
+pub(super) struct CompressedPathArena<M> {
+    pub(super) nodes: SlotMap<CompressedNodeId, CompressedPathNode<M>>,
+    pub(super) roots: FxHashMap<SequenceHash, CompressedNodeId>,
+    pub(super) splits: u64,
```
jthomson04: `splits`/`merges` are written but never read anywhere in the tree (the `dead_code` lint is silenced by `derive(Debug)`). Leftover prototype instrumentation — remove them or wire them into stats.

jthomson04 (commented):
## Review

Replaces the per-block slotmap arena and `unique_blocks` reverse index with a compressed radix tree; output blocks move to a flat map. Handle stability comes from two rules — splits keep the existing node ID on the suffix, unary merges keep the child ID — so a request's `prompt_tail` always ends at the same absolute depth.

**Verification:** ran `cargo test --locked -p dynamo-kv-router --lib` on the head locally — 738 passed (10/10 in `block_tracker`).

**Correctness:** traced the invariants against the old per-block semantics and they hold:
- Only `terminal_owners == 0` nodes are merged away or leaf-removed, so no live handle can dangle; mid-edge terminals are impossible (an acquire ending mid-edge forces a split first).
- Node-granularity fractional marking is equivalent to the old per-block walk (mid-edge blocks always have exactly one implicit incoming edge), and the pre-existing quirk that fractional marks persist across a later second acquire is carried over exactly.
- `recompress_from` runs before the `removed_hashes == 0` early return, correctly handling the terminal 1→0-on-a-unary-node case, and the fraction-equality guard blocks merges that would corrupt accounting.
- The 10k-step randomized reference-model test with shared-prefix prompts subsumes most of the deleted ordering tests.

**Main asks (inline):** avoid building the full prompt path on no-op releases, restore O(1) `active_blocks()` via running totals (likely converts the projection p99 regression into a win and clears the all-latency gate), drop the dead `splits`/`merges` counters, and add a test for the fraction-mismatch merge guard.

Process note per `sequences/CLAUDE.md`: benchmark numbers are included ✓; the write-model change and the acknowledged projection regression still want PeaBrane's sign-off.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -48,460 +60,401 @@ pub(super) struct ReleasedPromptPath {
...
+                .insert_root(sequence.to_vec(), BlockNodeMetadata::terminal());
+            self.prompt_total += sequence.len() as f64;
+            return (RequestBlockChain::new(Some(tail), sequence.len()), Some(0));
+        };
 
-        let mut parent = None;
-        let mut first_new_idx = 0;
+        let mut path_pos = 0;
+        loop {
+            let (edge_len, match_len) = {
```
PeaBrane, line 103: `acquire_prompt` now walks and compares the entire shared prefix. An exact duplicate or one-block extension of a long live prompt moves from the old tail/deepest-prefix hash lookup to O(prompt length). The aggregate AgentX result is positive, but could we add a focused benchmark for long duplicate and short-extension acquisitions so this reuse-heavy case is explicit?

lib/kv-router/src/sequences/compressed_path_arena.rs
```diff
@@ -0,0 +1,286 @@
...
+                .nodes
+                .get_mut(node_id)
+                .expect("compressed split node is missing");
+            assert!(
+                split_at > 0 && split_at < node.edge.len(),
+                "compressed split must be inside the edge"
+            );
+            let old_parent = node.parent;
+            let old_first = node.edge[0];
+            let suffix = node.edge.split_off(split_at);
```
PeaBrane, line 101: Ending a request inside a long edge makes `split_off` allocate and copy the entire suffix; releasing that short-lived request immediately recompresses by copying it back. The old per-block tracker handled this endpoint with an ownership increment/decrement. Could we benchmark repeated shorter-prefix acquire/release churn, or otherwise document the accepted trade-off?

PeaBrane (approved):
✅ AI-reviewed. Correctness and lifecycle invariants look sound. Approving with two non-blocking performance follow-ups for reuse-heavy prompt shapes.

lib/kv-router/src/sequences/block_tracker.rs
jthomson04: Updated to check the boundary hash before committing to walking the entire node.

lib/kv-router/src/sequences/compressed_path_arena.rs
jthomson04: Removed the recompression logic

jthomson04:
/ok to test [fd4bfc6](https://github.com/ai-dynamo/dynamo/pull/11644/commits/fd4bfc6febc1df3c8095ccdc1872768686320273)

jthomson04:
/ok to test [5c45032](https://github.com/ai-dynamo/dynamo/commit/5c4503220cd0230d2a162478a54ecd9fb53f5303)

jthomson04:
/ok to test [cbc4abd](https://github.com/ai-dynamo/dynamo/commit/cbc4abdcbd64448ab3090d494f2d50f80e0dd6c2)

Issue #10205, opened 2026-06-02
LWS/DS orchestrator support: complete LWS, add DisaggregatedSet


panpan0000:
The LWS pathway in Dynamo is already complete — the DCD controller has full LWS resource generation (`reconcileLeaderWorkerSetResources` + `generateLeaderWorkerSet` + `IsLeaderWorkerSetReady` at `dynamocomponentdeployment_controller.go:306`/`555`/`448`), and a DGD going through the LWS pathway produces LWS resources via the chain DGD → DCDs → LWS. RBAC for `leaderworkersets` is full CRUD.

This issue tracks the work needed to add DisaggregatedSet as a third orchestrator option. It is split into three independently shippable PRs.

## What is in place

- LWS as a multinode orchestrator (DCD-level and DGD-level paths both work; LWS resources are created and reconciled).
- Runtime detection (`DetectLWSAvailability`, `RuntimeConfig.LWSEnabled`) and `nvidia.com/enable-grove: "false"` opt-in.
- `LWSMultinodeDeployer` (`internal/dynamo/lws.go`) for hostname/rank expansion.

## What is missing

DisaggregatedSet (https://github.com/kubernetes-sigs/lws/tree/main/disaggregatedset, module `sigs.k8s.io/disaggregatedset`) is not supported. DS owns N LWSes per resource (one per role) and adds an N-dimensional coordinated rolling update, per-role headless service, and a single `DisaggregatedSet.status.roleStatuses[]` status surface. It is co-designed with llm-d and is the natural fit for Dynamo's prefill/decode disagg shape.

### Phase 1: DisaggregatedSet orchestrator

Add DS as a third orchestrator option alongside Grove and LWS.

- `deploy/operator/go.mod`: add `sigs.k8s.io/disaggregatedset`.
- `internal/consts/consts.go`: `MultinodeDeploymentTypeDisaggregatedSet`, `DisaggregatedSetGVR`.
- `api/v1alpha1/dynamographdeployment_types.go` + `api/v1beta1/common.go`: add `DisaggregatedSet` to `ComponentKind` (kept for future per-component use; Phase 1 uses the deployment-level annotation).
- `internal/dynamo/disaggregatedset.go` (new): `MultinodeDeployer`, `GenerateDisaggregatedSet`, `CheckDisaggregatedSetReady`. Each role embeds a full `LeaderWorkerSetSpec` (DS API at `disaggregatedset/api/v1alpha1/disaggregatedset_types.go:43`).
[...]

PR #11645, opened 2026-07-14
feat(operator): add DisaggregatedSet pathway

panpan0000:
## Summary

- detect `disaggregatedset.x-k8s.io/v1` with its required LWS API independently from the existing LWS + Volcano pathway
- route workloads with explicit precedence: Grove first, opt-in DisaggregatedSet second, and the standard DCD pathway as fallback
- render eligible multinode worker roles into one DisaggregatedSet while retaining regular DCDs for non-selected components
- reuse the existing LWS leader/worker pod-template rendering path
- emit schema-valid LWS role specs with explicit rollout/restart policies and render every role with the stable DGD-level Kubernetes discovery identity
- normalize desired DisaggregatedSets through an API-server dry run so reconciliation converges after CRD defaulting, including defaults added by compatible future CRDs
- reconcile/adopt component and model Services, preserve shared model-Service ownership, and remove stale DGD-owned component Services only after the replacement is ready
- restore component and shared model Service ownership to replacement DCDs before deleting a stale DisaggregatedSet during fallback; recreate a missing shared model Service under its selected replacement DCD before cleanup
- coordinate checkpoint startup gating across every selected DS role, keeping legacy DCDs serving until checkpoint readiness and DS cutover
- aggregate readiness and replica status across target and draining child LWS revisions; use `status.roleStatuses` directly when the installed DS API exposes it and require exact replica convergence
- watch DS/LWS label and controller-owner changes used by routing, readiness, and event mapping
- preserve unmanaged DS labels, annotations, and non-controller owner references while applying desired operator metadata during reconciliation
- validate the DS API's 10-role limit and use stable hash truncation so upstream `<set>-<revision>-<role>` child names always fit the 63-character DNS label limit
- migrate between DS and Grove/DCD pathways without deleting the serving workload before its replacement is ready
- close lifecycle gaps identified in review: annotation-only pathway changes now reconcile, stale Grove PodCliqueSets are removed after DS cutover, DS restart coalescing only runs when DS wins routing, restart annotations survive consecutive requests after DCD cleanup, non-selected multinode roles require LWS, and reparented Services trigger DGD reconciliation
- complete coalesced sequential DS restarts without waiting on an already-restarted sibling role, and accept a shared model Service already owned by any replacement DCD controlled by the same DGD
- add generated RBAC, envtest fixtures, unit coverage, and Kubernetes documentation

Workload routing is ordered, not mutually exclusive: `Grove > opt-in DS > DCD`. On a Grove-enabled cluster, selecting DS requires both `nvidia.com/enable-grove: "false"` and `nvidia.com/enable-disaggregatedset: "true"`; installing the DS API alone never changes existing workload routing.

## Compatibility and scope

- requires LWS `v0.9.0` or newer; Dynamo currently builds and tests against `v0.9.0`
- treats `status.roleStatuses` as authoritative when accompanied by generation-observation evidence and otherwise reads target-revision child LeaderWorkerSets; if neither source proves readiness, cutover remains pending and the previous workload stays serving
- keeps DS as a supported, explicit per-DGD opt-in below the default Grove pathway; the annotations are the per-DGD routing opt-in

The Dynamo platform chart intentionally does not add a synthetic `disaggregatedset.enabled` value. DS is supplied by an external LWS installation and detected at runtime.

## Validation

- `KUBEBUILDER_ASSETS="$PWD/bin/k8s/1.30.0-darwin-arm64" go test ./internal/controller ./internal/controller_common/... ./cmd/... -count=1`
  - controller envtest suite passed, including API-server-backed DCD-to-DS-to-DCD cutover, component and shared model Service ownership restoration, recreation of a missing DGD-owned shared model Service before DS fallback cleanup, checkpoint coordination, child-revision draining, metadata preservation, bounded child names, role limits, stale Service cleanup, annotation routing, Grove cleanup and precedence, multinode guard, and consecutive sequential restart regressions
- real pinned LWS v0.9.0 CRD validation/defaulting test:
  - `KUBEBUILDER_ASSETS="$PWD/bin/k8s/1.30.0-darwin-arm64" go test ./internal/controller -run '^TestDisaggregatedSetRealLWSCRDValidationAndConvergence$' -count=1 -v`
  - initial creation passes the real schema and an identical second reconciliation returns `modified=false`
- focused unit regressions cover explicit LWS defaults, DGD-level ServiceAccount rendering, DS/LWS metadata predicates, exact parent replica convergence, and safe DGD-to-DCD owner transfer including refusal to steal foreign-owned resources
- `go vet ./internal/controller ./internal/controller_common/... ./cmd/...`
- `GOTOOLCHAIN=go1.26.3 go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.62.2 run --timeout=5m`
- `uvx pre-commit run --files <changed files>`
- current controller package coverage: 71.4%; the new convergence and ownership helpers are exercised by API-server-backed tests
- `make manifests`
- `make generate-helm-docs`
- CI-equivalent container linter:
  - `docker buildx build --platform linux/amd64 --target linter --build-context snapshot=../snapshot --build-context operator=../operator .`
- CI-equivalent full operator test target:
  - `docker buildx build --platform linux/amd64 --target tester --build-context snapshot=../snapshot --build-context operator=../operator .`
- Live Kubernetes validation on OrbStack with LWS v0.9.0:
  - a generated two-role DS created two child LWS objects and four Ready pods
  - every pod used the existing DGD-level Kubernetes discovery ServiceAccount
  - coordinated rollout moved both roles from revision `d65d7965` to `b47ae5a3`, then removed the old revision
  - confirmed v0.9.0 leaves the parent DS status empty, exercising the child-LWS readiness/watch compatibility requirement
- on head `7854e10ff5`, focused DisaggregatedSet regressions and the controller DisaggregatedSet test set pass; lightweight CI is rerunning after the latest lint-only commit
- stacked Draft #12352 runs the real live-cluster suite with LWS v0.9.0 and Volcano v1.14.0; two immediate consecutive runs passed 7/7 (48.1s and 49.6s), covering DS/LWS readiness, metadata watch, restart completion, DS-to-DCD fallback, deleted model-Service recreation, DCD readiness, and Service ownership handoff

## Follow-up testing

Draft #12352 implements the dependent phase-2 live-cluster E2E suite and remains intentionally outside this feature PR. It adds pinned dependency preparation plus `make test-e2e-disaggregatedset` for ordinary kubeconfig-backed clusters; it must not merge before this PR.

## Review focus

- `deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go`
- `deploy/operator/internal/controller/dynamographdeployment_controller.go`
- `deploy/operator/internal/controller/dynamocomponentdeployment_controller.go`

## Related issues

Relates to #10205

Addresses the correctness findings in https://github.com/yankay/dynamo/issues/4


deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,811 @@
...
+			return nil
+		}
+		return fmt.Errorf("failed to get stale DisaggregatedSet %s/%s: %w", dgd.Namespace, disaggregatedSetName(dgd), err)
+	}
+	if !isControlledByBetaDGD(ds, dgd) {
+		return fmt.Errorf("refusing to delete DisaggregatedSet %s/%s because it is not controlled by DynamoGraphDeployment %s/%s", dgd.Namespace, disaggregatedSetName(dgd), dgd.Namespace, dgd.Name)
+	}
+	if err := r.Delete(ctx, ds); err != nil {
+		return fmt.Errorf("failed to delete stale DisaggregatedSet %s/%s: %w", dgd.Namespace, disaggregatedSetName(dgd), err)
+	}
```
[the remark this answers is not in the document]
panpan0000: Fixed in b13b631dcb: Delete now ignores a NotFound race, with a focused unit test.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,811 @@
...
+		return ReconcileResult{}, err
+	}
+	resources = append(resources, syncedDSResource)
+
+	dsReady, _, _ := checkDisaggregatedSetReadiness(syncedDS, selection)
+	for _, key := range sortedDCDKeys(dynamoComponentsDeployments) {
+		dcd := dynamoComponentsDeployments[key]
+		if _, selected := selection.componentToRole[key]; selected {
+			continue
+		}
```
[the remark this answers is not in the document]
panpan0000: Fixed in d6b5a6e6a1: checkpoint startup policies are now applied to all generated DCD specs before DS role rendering, including selected roles. Added coverage for WaitForCheckpoint gating a selected role to zero replicas.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -464,6 +440,92 @@ func (r *DynamoGraphDeploymentReconciler) reconcileResources(ctx context.Context
...
+	logger.Info("Reconciling Dynamo components deployments", "hasMultinode", hasMultinode, "lwsEnabled", r.RuntimeConfig.LWSEnabled)
+	return r.reconcileDynamoComponentsDeployments(ctx, dynamoDeployment, restartState, checkpointInfos)
+}
+
+func (r *DynamoGraphDeploymentReconciler) deleteDisaggregatedSetOnLegacyPath(ctx context.Context, dgd *nvidiacomv1beta1.DynamoGraphDeployment) error {
+	if !r.RuntimeConfig.DisaggregatedSetEnabled {
+		return nil
+	}
+	return r.deleteDisaggregatedSetIfExists(ctx, dgd)
+}
```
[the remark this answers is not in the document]
panpan0000: Fixed in b13b631dcb: Grove/DCD replacement reconciliation now runs first; the old DS is deleted only after the replacement reports DGDStateSuccessful. Pending/error paths preserve the serving DS, and the ordering has a focused unit test.

docs/backends/sglang/sglang-diffusion.md
```diff
@@ -23,7 +23,7 @@ If you see a CuDNN version mismatch error on startup (`cuDNN frontend 1.8.1 requ
 
 Diffusion Language Models generate text through iterative refinement rather than autoregressive token-by-token generation. The model starts with masked tokens and progressively replaces them with predictions, refining low-confidence tokens each step.
 
-LLM diffusion is auto-detected: when `--dllm-algorithm` is set, the worker automatically uses `DiffusionWorkerHandler` without needing a separate flag. For more details on diffusion algorithms, see the [SGLang Diffusion Language Models documentation](https://github.com/sgl-project/sglang/blob/main/docs/supported_models/text_generation/diffusion_language_models.md).
+LLM diffusion is auto-detected: when `--dllm-algorithm` is set, the worker automatically uses `DiffusionWorkerHandler` without needing a separate flag. For more details on diffusion algorithms, see the [SGLang Diffusion documentation](https://docs.sglang.io/docs/sglang-diffusion).
```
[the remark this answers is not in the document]
panpan0000: Resolved by reverting the unrelated SGLang documentation change. This PR is now scoped to the DS pathway.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,811 @@
...
+}
+
+func checkDisaggregatedSetReadiness(ds *unstructured.Unstructured, selection disaggregatedSetSelection) (bool, string, map[string]nvidiacomv1beta1.ComponentReplicaStatus) {
+	statuses := make(map[string]nvidiacomv1beta1.ComponentReplicaStatus, len(selection.componentToRole))
+	roleStatuses := disaggregatedSetRoleStatuses(ds)
+	notReadyReasons := []string{}
+	for componentName, roleName := range selection.componentToRole {
+		desiredReplicas := selection.desiredReplicas[componentName]
+		componentStatus := nvidiacomv1beta1.ComponentReplicaStatus{
+			ComponentKind:  nvidiacomv1beta1.ComponentKindLeaderWorkerSet,
```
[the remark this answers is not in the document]
panpan0000: Keeping ComponentKindLeaderWorkerSet intentionally: per-component status reports the concrete child workload resources, which are LeaderWorkerSets; there is no ComponentKindDisaggregatedSet enum. The v0.9 compatibility path now also reports the actual target-revision child LWS names.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -464,6 +440,92 @@ func (r *DynamoGraphDeploymentReconciler) reconcileResources(ctx context.Context
...
+	checkpointInfos map[string]*checkpoint.CheckpointInfo,
+) (ReconcileResult, error) {
+	logger := log.FromContext(ctx)
+
+	if r.isGrovePathway(dynamoDeployment) {
+		logger.Info("Reconciling Grove resources", "hasMultinode", hasMultinode, "lwsEnabled", r.RuntimeConfig.LWSEnabled)
+		if err := r.deleteDisaggregatedSetOnLegacyPath(ctx, dynamoDeployment); err != nil {
+			return ReconcileResult{}, err
+		}
+		return r.reconcileGroveResources(ctx, dynamoDeployment, restartState, checkpointInfos)
```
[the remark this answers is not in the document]
panpan0000: Keeping Grove priority intentionally, consistent with the existing orchestrator-selection contract and docs. On clusters with Grove, users opt into DS by setting both enable-disaggregatedset=true and enable-grove=false; this is documented in the PR's Kubernetes docs.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1093 @@
...
+	}
+	original := current.DeepCopy()
+	if current.GetLabels() == nil {
+		current.SetLabels(map[string]string{})
+	}
+	maps.Copy(current.GetLabels(), desired.GetLabels())
+	if current.GetAnnotations() == nil && len(desired.GetAnnotations()) > 0 {
+		current.SetAnnotations(map[string]string{})
+	}
+	maps.Copy(current.GetAnnotations(), desired.GetAnnotations())
```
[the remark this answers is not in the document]
panpan0000: Fixed in d9498fbd41. `syncDisaggregatedSet` now merges the copied label and annotation maps and explicitly writes them back with `SetLabels`/`SetAnnotations`. The regression test now verifies both preservation of unmanaged metadata and application of desired metadata, including re-reading the persisted object from the client.

deploy/operator/cmd/main.go
```diff
@@ -435,6 +435,8 @@ func main() {
 
 	setupLog.Info("Detecting LWS availability...")
 	lwsDetected := commonController.DetectLWSAvailability(mainCtx, mgr)
+	setupLog.Info("Detecting DisaggregatedSet availability...")
+	runtimeConfig.DisaggregatedSetEnabled = commonController.DetectDisaggregatedSetAvailability(mainCtx, mgr)
```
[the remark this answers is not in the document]
panpan0000: Keeping API-based auto-detection intentionally for this PR. DisaggregatedSet remains a per-DGD opt-in pathway: API presence alone cannot select it, and `nvidia.com/enable-disaggregatedset: "true"` is always required (with Grove disabled when Grove is available). A global tri-state override would expand the operator config API, generated configuration, Helm surface, and tests without changing the current safe routing behavior, so that should be a separate follow-up if an operational kill switch is required.

panpan0000:
@yankay   @athreesh  please kindly help to review , thank you very much

panpan0000:
Resolved the latest `main` conflicts and aligned DS capability detection with the new unified operator feature-gate architecture. Re-ran operator controller/envtests, full non-e2e Go packages, `go vet`, and pre-commit successfully. The repository e2e package only failed because the configured local Kubernetes API was not running (`127.0.0.1:26443 connection refused`). Latest head: `3dd6d41232`; please run `/ok to test 3dd6d41232` for full CI.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -2793,6 +2872,23 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
 		}))
 	}
+	if r.RuntimeConfig.Gate.Enabled(features.DisaggregatedSet) {
+		ctrlBuilder = ctrlBuilder.Owns(newDisaggregatedSetObject(), builder.WithPredicates(predicate.Funcs{
```
julienmancuso, line 3003: **[P1] Watch pathway annotation changes**

The primary DGD watch above filters updates with `GenerationChangedPredicate`, but both DS and Grove routing read metadata annotations. Kubernetes does not increment `metadata.generation` for annotation-only updates, so adding `nvidia.com/enable-disaggregatedset: "true"` and `nvidia.com/enable-grove: "false"` to a stable DGD may never enqueue reconciliation; the migration can stall until an unrelated child event. Please also compare these routing annotations in the predicate.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1098 @@
...
+
+	nonSelectedResources, err := r.reconcileDisaggregatedSetNonSelectedDCDs(
+		ctx, dynamoDeployment, dynamoComponentsDeployments, selection,
+	)
+	if err != nil {
+		return ReconcileResult{}, err
+	}
+	resources = append(resources, nonSelectedResources...)
+
+	if dsReady {
```
julienmancuso, line 360: **[P1] Clean up Grove after DS cutover**

Once DS is ready, this only deletes selected DCDs. If the source pathway was Grove, the DGD-owned `PodCliqueSet` is not among them and no DS-path code deletes it, so the old Grove workload remains running and consuming GPUs after DS reports success (and may continue matching the same Services). Please delete the owned PCS after DS readiness using the same replacement-first ordering.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1098 @@
...
+		return ReconcileResult{}, fmt.Errorf("failed to build rolling update context: %w", err)
+	}
+
+	existingRestartAnnotations, err := r.getExistingRestartAnnotationsDCD(ctx, dynamoDeployment)
+	if err != nil {
+		logger.Error(err, "failed to get existing restart annotations")
+		return ReconcileResult{}, fmt.Errorf("failed to get existing restart annotations: %w", err)
+	}
+
+	dynamoComponentsDeployments, err := dynamo.GenerateDynamoComponentsDeployments(
```
julienmancuso, line 300: **[P1] Preserve sequential restart semantics**

`restartState` is rendered into per-component pod annotations before all selected roles are combined into one DS. LWS computes one revision for the entire role list and creates a new child LWS for every role whenever any role template changes. Under Dynamo's default `Sequential` restart, restarting role A therefore rolls all selected roles; advancing to B creates another revision and rolls them all again. That violates the requested order and repeatedly restarts roles. Please treat the selected DS roles as one restart unit or reject/fallback sequential restarts.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -425,27 +399,30 @@ func (r *DynamoGraphDeploymentReconciler) reconcileResources(ctx context.Context
 		}
 	}
 
-	// return error early if Grove and LWS is not available for multinode
-	if !r.isGrovePathway(dynamoDeployment) && hasMultinode && !r.RuntimeConfig.Gate.Enabled(features.LWS) {
+	useDisaggregatedSet, disaggregatedSetFallbackReason := r.shouldUseDisaggregatedSet(dynamoDeployment)
+
+	// return error early if Grove, DS, and LWS are all unavailable for multinode
+	if !r.isGrovePathway(dynamoDeployment) && hasMultinode && !r.RuntimeConfig.Gate.Enabled(features.LWS) && !useDisaggregatedSet {
```
julienmancuso, line 407: **[P2] Guard non-DS multinode components**

A successful DS selection bypasses the LWS availability check for the whole graph. A valid graph can contain two selected multinode workers plus a multinode frontend or planner; with the DS API present but the LWS gate disabled (for example, no Volcano), the non-selected component is later emitted as a DCD, and the DCD controller chooses a plain Deployment, silently dropping its multinode topology. Please require LWS support for every remaining multinode component before bypassing this guard.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -2793,6 +2872,23 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
 		}))
 	}
+	if r.RuntimeConfig.Gate.Enabled(features.DisaggregatedSet) {
+		ctrlBuilder = ctrlBuilder.Owns(newDisaggregatedSetObject(), builder.WithPredicates(predicate.Funcs{
+			CreateFunc:  func(ce event.CreateEvent) bool { return false },
+			DeleteFunc:  func(de event.DeleteEvent) bool { return true },
+			UpdateFunc:  func(ue event.UpdateEvent) bool { return disaggregatedSetStatusChanged(ue.ObjectOld, ue.ObjectNew) },
+			GenericFunc: func(ge event.GenericEvent) bool { return true },
+		})).Watches(
```
julienmancuso, line 3008: **[P2] Watch reparented Services**

DS cutover reparents component/model Services from their DCD to the DGD. The DCD controller's `Owns(&corev1.Service{})` watch no longer maps those objects after reparenting, while this DGD controller does not own or watch Services. A later Service deletion or drift will therefore remain unreconciled until an unrelated event. Please add a DGD Service ownership/watch for the reparented resources.

julienmancuso (changes requested):
I found five behavioral gaps in the current DisaggregatedSet implementation. The focused DS tests, full controller envtest suite, feature-gate tests, `go vet`, and `git diff --check` pass at `3dd6d41232`; the comments below cover interactions not exercised by those tests.

deploy/operator/internal/controller/dynamographdeployment_controller.go
panpan0000: Fixed in fec36f7fd8. The primary DGD predicate now ORs generation changes with normalized comparisons of the Grove and DisaggregatedSet routing annotations. Added `TestWorkloadRoutingAnnotationsChanged`.

deploy/operator/internal/controller/dynamographdeployment_controller.go
panpan0000: Fixed in fec36f7fd8. DS selection now refuses to bypass the multinode-orchestrator guard when LWS is unavailable and any multinode component remains outside the selected DS roles. Added coverage for both the guarded and LWS-supported cases.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in fec36f7fd8. After the replacement DS is current and ready, the DS path now deletes the DGD-owned Grove PodCliqueSet before retiring selected DCDs. The cleanup is ownership-checked, NotFound-idempotent, and covered by `TestDeleteGrovePodCliqueSetOnDisaggregatedSetPath`.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in fec36f7fd8 by treating selected DS roles as one restart unit. When a sequential restart reaches any selected role, all selected roles receive the same restart annotation, so LWS creates one whole-set revision; later selected-role steps reuse that revision instead of rolling the set again. DS readiness already aggregates all selected roles before the sequence advances. Added `TestCoalesceDisaggregatedSetRestartState`.

deploy/operator/internal/controller/dynamographdeployment_controller.go
panpan0000: Fixed in fec36f7fd8. The DGD controller now owns/watches Services directly; creation events are suppressed, while update/delete/generic events enqueue reconciliation for DS-reparented component and model Services. Added predicate coverage in `TestDGDOwnedServiceEventPredicate`.

panpan0000:
> I found five behavioral gaps in the current DisaggregatedSet implementation. The focused DS tests, full controller envtest suite, feature-gate tests, `go vet`, and `git diff --check` pass at `3dd6d41232`; the comments below cover interactions not exercised by those tests.

Thanks a lot for review @julienmancuso and all resoved.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -425,27 +399,33 @@ func (r *DynamoGraphDeploymentReconciler) reconcileResources(ctx context.Context
...
 	restartState := dynamo.DetermineRestartState(dynamoDeployment, restartStatus)
-
-	var result ReconcileResult
-	if r.isGrovePathway(dynamoDeployment) {
-		logger.Info("Reconciling Grove resources", "hasMultinode", hasMultinode, "lwsEnabled", r.RuntimeConfig.Gate.Enabled(features.LWS))
-		result, err = r.reconcileGroveResources(ctx, dynamoDeployment, restartState, checkpointInfos)
-	} else {
-		logger.Info("Reconciling Dynamo components deployments", "hasMultinode", hasMultinode, "lwsEnabled", r.RuntimeConfig.Gate.Enabled(features.LWS))
-		result, err = r.reconcileDynamoComponentsDeployments(ctx, dynamoDeployment, restartState, checkpointInfos)
+	if useDisaggregatedSet {
```
julienmancuso: **[P1] Only coalesce on the active DS path**

`useDisaggregatedSet` means DS is eligible, but Grove can still win routing precedence. With both gates enabled and only the DS annotation set, this coalesces the restart state before `reconcileWorkloadResources` selects Grove, causing a sequential Grove restart to roll multiple roles together. Please gate coalescing on the actual DS pathway (for example, also require `!r.isGrovePathway(dynamoDeployment)`) and add coverage for Grove precedence.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1143 @@
...
+) (ReconcileResult, error) {
+	resources := []Resource{}
+	logger := log.FromContext(ctx)
+
+	rollingUpdateCtx, err := r.buildRollingUpdateContext(ctx, dynamoDeployment)
+	if err != nil {
+		return ReconcileResult{}, fmt.Errorf("failed to build rolling update context: %w", err)
+	}
+
+	existingRestartAnnotations, err := r.getExistingRestartAnnotationsDCD(ctx, dynamoDeployment)
```
julienmancuso, line 285: **[P1] Preserve DS restart annotations across requests**

After DS cutover, selected DCDs are deleted, so this lookup cannot recover their existing restart annotations. On a subsequent sequential restart whose first component is outside the DS, the desired DS drops the previous restart annotation and creates a new whole-set revision before the DS roles' turn; reaching the first DS role then rolls the set again. Please preserve annotations from the current DS role templates (or otherwise keep them stable) and test two consecutive restart IDs with a non-DS component first.

julienmancuso (changes requested):
Re-reviewing `fec36f7fd8`, four of the five original findings are resolved, but the sequential-restart fix still has two blocking lifecycle gaps. The full controller suite, `go vet`, and `git diff --check` pass; the scenarios below are not covered by the added tests.

deploy/operator/internal/controller/dynamographdeployment_controller.go
panpan0000: Fixed in `64246ccb4e`. Restart-state preparation now calls `coalesceDisaggregatedSetRestartStateForPath`, which leaves the state unchanged whenever Grove is the active pathway and only coalesces when DS actually wins routing. Added `TestDisaggregatedSetRestartCoalescingRespectsGrovePrecedence` covering both gates with Grove precedence, followed by explicit Grove opt-out selecting DS.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in `64246ccb4e`. The DS reconcile now reads restart annotations from the current DS role templates and overlays them on the legacy-DCD fallback map, so selected roles retain their prior restart ID after cutover has removed their DCDs. Added `TestDisaggregatedSetRestartAnnotationsSurviveConsecutiveRequests`: an existing `restart-1` DS receives `restart-2` with a non-DS frontend first, keeps both DS roles at `restart-1`, then moves both roles to `restart-2` exactly once when the DS restart unit is reached. Full controller/envtest, `go vet`, and CI-version golangci-lint pass.

panpan0000:
Thanks for the re-review—both remaining sequential-restart lifecycle gaps are fixed in `64246ccb4e` with regression coverage for Grove precedence and consecutive restart IDs, and the latest `main`-synced head `88b846d6b3` passes the full controller suite, `go vet`, `golangci-lint`, and Pre Merge.

panpan0000:
A defensive follow-up is now in `1e3b1251c2`: stale ready-looking `roleStatuses` without generation-observation evidence fall back to target-revision child LWS readiness, and DS auto-detection now also requires the LWS API; focused tests, the full controller suite, `go vet`, `golangci-lint`, and `git diff --check` pass.

julienmancuso (approved):

panpan0000:
Thank you @yankay , you suggestions are addressed and replied in thread https://github.com/yankay/dynamo/issues/4

athreesh:
/ok to test [017d94d](https://github.com/ai-dynamo/dynamo/pull/11645/commits/017d94d8bc3f1d442209d4520a1601cb3ef8f9be)

panpan0000:
Follow-up hardening is pushed in [`a499a69d7c`](https://github.com/ai-dynamo/dynamo/commit/a499a69d7ca5bb4355f888983d0dabcbec62c842).

An API-server-backed regression now reproduces the narrow DS→DCD handoff case where the DGD-owned shared model Service disappears after replacement DCDs are Ready but before ownership transfer. Fallback now recreates that Service under the selected replacement DCD before deleting the stale DisaggregatedSet. The focused regression fails before the fix and passes after it; the full controller/internal controller-common/cmd tests, `go vet`, pre-commit, and `git diff --check` also pass.

I also clarified the validation boundary in the PR description. The current OrbStack check is a manual real-LWS smoke, not a committed full operator E2E. A separate follow-up PR should mirror `deploy/operator/test/e2e/dgdr/`, add `make test-e2e-disaggregatedset`, and cover live DGD-driven DS creation, annotation/watch transitions, DS→DCD ownership, and restart revisions without expanding this PR further.

Please run `/ok to test a499a69d7c` for full CI on the new head.

panpan0000:
Phase-2 live-cluster E2E is now open as Draft #12352 (stacked on this PR).

Running it against OrbStack with real LWS v0.9.0 + Volcano v1.14.0 found two controller issues, both fixed here in focused commits:

- `cf043609ce`: complete a coalesced DisaggregatedSet sequential restart instead of leaving the second role permanently `Restarting`
- `1a33ed81f1`: accept a shared model Service already controlled by a sibling replacement DCD owned by the same DGD

The full lifecycle suite now passes twice consecutively, 7/7 each run: real DS/LWS creation and readiness, metadata propagation, restart, annotation removal, DS→DCD fallback, deleted model-Service recreation, replacement readiness, and Service ownership handoff.

athreesh:
/ok to test [7854e10](https://github.com/ai-dynamo/dynamo/pull/11645/commits/7854e10ff5b5663fa73c8067e20d8ae1e826f9b4)

athreesh (approved):
docs lgtm

athreesh:
/ok to test [84e63e5](https://github.com/ai-dynamo/dynamo/pull/11645/commits/84e63e5e4b22595f87b5e7e767929aaeaab8b6e4)

panpan0000:
Pushed a fix for the Operator Integration compile failure on the latest head (8763efe7de).

Pre-merge operator checks are green now. 

Could you please re-run full CI with `/ok to test 8763efe7de`?

panpan0000:
Adding context for reviewers: follow-up live-cluster e2e coverage is in #12352, which depends on this PR and stays unmerged until this one lands. I also reran the DisaggregatedSet live tests on a real Kubernetes cluster after the latest fixes; both consecutive runs passed 7/7 (54.8s and 49.9s).

julienmancuso:
Now that #12283 has merged (associated with #12035), the DGD controller has a composition-first architecture. I think this PR should be reworked on top of current `main` rather than resolving the old monolithic-controller conflicts mechanically.

The key design point is that DisaggregatedSet should become a third complete graph-level workload program:

- Grove program
- DisaggregatedSet program
- Component/DCD program

DisaggregatedSet is a graph-level program because one DisaggregatedSet can coordinate multiple DGD components while unselected components remain ordinary DCDs. LWS itself is not another DGD program: it remains a workload mechanism used by DCD and by the DisaggregatedSet implementation.

The outer DGD reconciler should only select the program, invoke it, and persist its returned status/events. The DisaggregatedSet program should own its sequencing and compose concrete collaborators for:

- selecting and validating participating components and role names;
- rendering roles using the existing read-only `dcdWorkloadRenderer`;
- synchronizing the DisaggregatedSet resource;
- reconciling unselected components as DCDs;
- reconciling Services and ownership handoff;
- resolving readiness across the DisaggregatedSet, target/draining LWS revisions, and unselected DCDs;
- grouping selected roles as one provider-specific restart unit;
- performing DisaggregatedSet-specific compatibility cleanup;
- returning complete authoritative DGD status, including `Ready`.

In particular, the DisaggregatedSet implementation should not construct or call a complete DCD reconciler. The renderer extraction previously included in this PR is also no longer needed because #12283 introduced the narrower `dcdWorkloadRenderer`, including multinode pod-template rendering.

The existing large DisaggregatedSet controller file should ideally be decomposed by responsibility rather than moving all its methods unchanged onto another large type. Nested reconcilers should have narrow dependencies and must not write DGD status directly. Status changes and queued events should be returned through the workload-program result and persisted once by the outer controller.

Restart coalescing for selected roles should remain DisaggregatedSet-specific. It should not alter the common restart flow to introduce generic provider lifecycle behavior. Similarly, DCD/Grove/DisaggregatedSet transitions and Service ownership handoff should be implemented as narrow provider-specific compatibility cleanup, not as a universal zero-downtime cutover abstraction.

I suggest rebuilding the implementation from current `main`, using the existing PR as the behavioral specification. The current tests should be retained and reorganized around the new owners, especially coverage for mixed DS/DCD graphs, checkpoint gating, restart grouping, target/draining LWS readiness, Service adoption, metadata preservation, and DCD → DS → DCD transitions.

This should preserve the behavior and feature scope already implemented here while making DisaggregatedSet coherent with the architecture now present on `main`.

deploy/operator/internal/controller/dynamographdeployment_controller.go
sttts: Cross-PR note for this mutable routing contract: [ai-dynamo/dynamo#12769 discussion](https://github.com/ai-dynamo/dynamo/pull/12769#discussion_r3749289437).

#12769 introduces the operator-owned `nvidia.com/grove-worker-hash-suffix-enabled` annotation as a durable compatibility epoch: once a DGD has moved to suffixed Grove runtime namespaces, it must not silently return to the legacy unsuffixed behavior. Its current update defaulter can lose that marker across `Grove true → false → true`, because both pathway transitions return before operator-owned state is preserved.

This PR is a concrete consumer of that edge case: it intentionally watches `enable-grove`, uses `enable-grove=false` to select DS on Grove-enabled clusters, and supports Grove/DCD/DS handoffs. If #12769 lands, the pathway-transition contract and tests here should retain operator-owned compatibility markers across those handoffs (or #12769 should fix the invariant before merge). Conversely, making `enable-grove` immutable would conflict with the supported routing model in this PR.

Not raising a separate blocker on #11645 here; linking the threads so the two API contracts stay aligned.

panpan0000:
rework moved to https://github.com/ai-dynamo/dynamo/pull/12625

tmonty12:
Closing, implementation of DisaggSet support refactored in this PR: https://github.com/ai-dynamo/dynamo/pull/12625

PR #11646, opened 2026-07-14
feat(kv-router): fence dead workers from candidate/affinity set

nnshah1:
## The problem, in one paragraph

When a worker dies, discovery removes it — but the KV router kept selecting it. Cache affinity made it worse: a request whose prefix was cached on the dead worker was pinned to it, so the router kept choosing a worker that was already gone. That produced the `instance_id not found` / `Connection refused` burst in the DIS-2382 incident, and cycling frontends didn't help because they came back and read the same stale view.

## The fix, in one sentence

**Ask the router's own discovery snapshot whether a worker is available, instead of maintaining a separate list of dead ones.**

The first revision kept a fenced-worker map with a 30-second TTL, mirrored from discovery. Review asked whether the existing `instance_avail` could do the job. It can, and it's strictly better — so this revision deletes the mirror.

`RoutingInstances::reconcile_discovered` already rebuilds the routable set from `discovered_ids`, in the same task that observes the removal. Reading it directly removes three problems the mirror had:

| Mirror problem | Gone because |
|---|---|
| A request could see stale candidates *and* an empty fence | There is no mirror to lag behind |
| The 30s TTL could re-enable a worker that was still absent | There is no TTL |
| Every selection took a lock, read the clock, and allocated a set | One `Arc` clone off a cached snapshot |

It also fixes a bug by deletion: the old code fenced removed **prefill** workers on the **decode** client, so departed prefill workers stayed selectable. Each client now tracks its own discovery — 8 call sites of manual mirroring are gone.

## The one rule to know

Availability is an **inclusion** set, the opposite of the overloaded set:

- `Some(set)` is authoritative — an **empty set rejects every candidate**
- `None` means this client has discovered nothing yet: *no information*, not *everything is dead*

That second rule is load-bearing. Without it, a client that hasn't received its first discovery update would reject all routing.

And unlike overload — which is transient and deliberately ignored by cache affinity — unavailability is absolute. **A dead worker is never selected, even for a cache hit.** That is the behavior the incident needed.

## Validation

`dynamo-runtime` 502 · `dynamo-kv-router` 828 · `dynamo-llm` 1903 · `dynamo-mocker` 700 — 0 failures. `cargo fmt --check` and `clippy --all-targets` clean.

Regressions: an unavailable worker is rejected on rank validation; it is rejected on the affinity path that ignores overload; an empty available set rejects everything; and a lifecycle test with **distinct decode and prefill clients** — an affinity-pinned prefill worker removed while the scheduler's candidate map is still stale is rejected, decode availability is provably undisturbed, and restoring it is effective immediately with no TTL.

**Not yet run:** `dynamo-frontend-benchmark` with a churn phase, and dyntest `--set fault=kill`. Both being readied separately. Worth noting the rework changes what the benchmark shows — the hot-path lock and allocation no longer exist, so it is regression-confirmation rather than a measurement of fence overhead.

`dynamo-runtime` gains a `testing` feature for the white-box helpers the cross-crate test needs. Dev-dependencies only; `resolver = "3"` keeps it out of normal builds.

## Related

Sibling: #11647.

— Neelay + 🤖


nnshah1:
## Graham-style code review

Core routing + runtime surface — reviewed strictly. The design is sound: the fence-vs-overload asymmetry (fenced rejected on *every* path incl. affinity; overload ignorable on the pin path) is the right model, and the `filter.rs` wiring + the two behavior tests capture exactly that. Parallel construction with the existing overloaded-worker plumbing is clean.

### High — addressed in 7c98986
- **`unfence_present_instances` was dead code.** No call sites, and it could never match: a re-registered worker gets a fresh lease id, so a fenced id is never in the "present" set. The `fenced_instances` doc also claimed *"Cleared when the same id re-registers,"* which the code never did. Removed the method; corrected the doc to state the fence is `FENCE_TTL`-bounded only.

### Medium (not yet changed — open to a follow-up)
- **`fenced_instance_ids()` prunes under a write lock on the selection hot path.** It's called on every `select_worker` via the provider closure; whenever the fenced set is non-empty (the full 30s window after *any* worker death) each selection takes the `write()` lock to `retain()`, serializing concurrent selections during exactly the failover window. The sibling `overloaded_instance_ids()` is read-only. Suggest moving TTL pruning off the read path (prune in `fence_instances_removed`, or a periodic sweep) and keeping the getter read-only.

### Low / style
- `std::sync::RwLock` / `std::time::Instant` are fully-qualified inline, inconsistent with the file's existing `StdMutex` alias convention — import them.
- `FencedWorkerProvider` is a byte-identical alias of `OverloadedWorkerProvider`; the distinct name + doc earns its keep (parallel construction) — noting it's intentional, not an oversight.
- The `fenced_instances` / `FENCE_TTL` rationale comments are load-bearing but a touch long; could trim to the one sentence that matters.

nnshah1:
Addressed the bot + review findings in `0f5ca71` (and earlier `7c98986`/`790ac96`):
- **Dead `unfence_present_instances` / 30s rejoin block** (Devin + CodeRabbit): removed the unused method and corrected the `fenced_instances` doc — the fence is `FENCE_TTL`-bounded, and a re-registered worker gets a fresh lease id so it was never re-matchable anyway.
- **`fenced_instance_ids()` write-lock on the selection hot path** (Devin + CodeRabbit): now read-only — expired entries are filtered out of the snapshot and pruning moved to the (rare) `fence_instances_removed` write path.
- **Internal the internal ticket IDs ticket IDs in comments** (Devin + CodeRabbit): stripped from all source comments per `.ai/linear-ticket-refs.md`.

nnshah1:
Both remaining findings addressed in `8e9f441`:

- **Fence missing from the admission-policy availability snapshot** (CodeRabbit, queue.rs:1436): confirmed real — `handle_enqueue` filtered only overload out of `available`, so a policy could return `Ready(Exact(fenced_worker))`. Now drops fenced workers from availability too, parallel to overload; structural legality is unchanged and final selection still applies the fence.

- **Transient availability drop fences a live worker for the TTL** (Devin, worker_monitor.rs:838): confirmed real — `runtime_configs` is `instance_avail ∩ configs`, and `instance_avail` transiently drops then reconciles a live worker under the *same* id (`report_instance_down`). Re-introduced and **wired** `unfence_present_instances` at each removal site with the present membership, so a returning worker is un-fenced immediately; the TTL stays as the backstop for ids that never return. (This intentionally restores the method removed in `7c98986` — it was never wired, and the transient-drop path is exactly where it's needed.)

lib/llm/src/kv_router.rs
```diff
@@ -282,6 +283,12 @@ where
         let overloaded_worker_provider: OverloadedWorkerProvider =
             Arc::new(move || client_for_overload.overloaded_instance_ids());
 
+        // fenced (dead) workers, populated by the worker monitor on
+        // discovery removal and read at selection to reject them everywhere.
+        let client_for_fence = client.clone();
+        let fenced_worker_provider: FencedWorkerProvider =
+            Arc::new(move || client_for_fence.fenced_instance_ids());
```
PeaBrane: Could we build hard eligibility directly from this router Client’s current `instance_avail` state, as Jacky suggested, instead of maintaining a second TTL-based mirror?

`instance_avail` already excludes discovery-removed and locally penalty-boxed workers. In the current flow, that snapshot changes first and `KvWorkerMonitor` copies the removal into `fenced_instances` asynchronously, so a request can still observe stale candidates plus an empty fence. Independent runtime-config snapshots can also clear a newer fence, and the 30-second TTL can re-enable a worker that remains absent if candidate convergence stalls.

I suggest an availability provider sampled at admission and final selection. It should expose a cheap cached set-like snapshot—such as `Arc<HashSet<WorkerId>>`—rather than converting the existing `Vec` per request. An empty provided set must mean “reject every candidate”; `None` should mean only “this path has no provider.” Overload should remain separate because it has different affinity semantics.

lib/llm/src/discovery/worker_monitor.rs
```diff
@@ -1092,8 +1109,15 @@ impl WorkerLoadMonitor for KvWorkerMonitor {
                             }
                             overloaded_tracker.remove_workers(&removed_workers);
                             client.clear_overloaded_instances_for_removed(&removed_workers);
+                            client.fence_instances_removed(&removed_workers);
```
PeaBrane: This callback is driven by `prefill_instances_rx`, but it fences and unfences `client`, which is the decode Client. The prefill scheduler reads the separately attached prefill Client, so the removed prefill worker remains selectable until the merged runtime-config watcher catches up.

Please apply these updates to the attached prefill Client. Reusing the exact router Client’s `instance_avail` snapshot would eliminate this manual decode/prefill mirroring entirely.

lib/runtime/src/component/client.rs
```diff
@@ -642,6 +658,60 @@ impl Client {
...
+            fenced.remove(id);
+        }
+    }
+
+    /// Snapshot of currently fenced workers, or `None` when none are live
+    /// (matching the overloaded-set convention so selection can skip the check
+    /// cheaply). Read-only: expired entries are filtered out of the snapshot but
+    /// pruning is left to [`fence_instances_removed`], so this stays on a read
+    /// lock — it is called on every selection while the fence set is non-empty.
+    pub fn fenced_instance_ids(&self) -> Option<HashSet<u64>> {
```
PeaBrane: Expired entries are filtered from the returned set but remain in `fenced_instances` until a later removal write. After the last removal, every scheduling request can therefore keep taking the read lock, reading the clock, and scanning stale entries indefinitely; while fences are live, each call also allocates a new `HashSet`.

The direct-availability design above removes this state. If the fence map remains, it should publish/cache an immutable snapshot or actually remove expired entries without imposing permanent request-path work.

PeaBrane (commented):
The central eligibility change looks good: hard unavailability overrides cache affinity, and admission plus final selection now agree. I also agree with Jacky that `instance_avail` should probably be the source of truth rather than introducing a second TTL-based state machine.

I left two correctness comments around that design:

1. The asynchronously mirrored fence still leaves ordering windows and can be cleared by an older independent watcher.
2. The immediate prefill-removal path currently updates the decode Client.

The direct `RoutingEligibility` tests are useful, but please add a lifecycle regression using distinct decode/prefill Clients: remove an affinity-pinned prefill worker while scheduler candidates remain stale, verify reselection, then restore it and verify immediate eligibility.

Before merging, please also run the repository’s `dynamo-frontend-benchmark` workflow if it has not already been run, and record the reproducibility details and results in the PR description for posterity. In particular:

- Compare the exact base and PR SHAs.
- Fully stop/restart and drain discovery between runs.
- Interleave the A/B arms and report medians from at least three measured runs, excluding the cold first run.
- Keep CPU isolation, frontend/mocker cores, block size, tokenizer, allocator, concurrency, workload, and warmup identical.
- Recompute throughput from the raw `profile_export.jsonl` artifacts.
- Include a worker-removal/churn phase so the new fencing path is active; a steady-state run with no removed worker will not exercise the new lock/allocation path.
- Report throughput, request latency, frontend CPU utilization, post-removal error behavior, commands/configuration, and artifact locations.

If this benchmark has already been completed, adding those details and results to the PR description is sufficient.

lib/runtime/src/component/client.rs
```diff
@@ -642,6 +656,32 @@ impl Client {
...
+    /// selected on any path. `Some` is therefore authoritative — an empty set
+    /// means "reject every candidate", not "no constraint".
+    ///
+    /// Returns `None` only when this client has discovered nothing at all, which
+    /// is absence of information rather than evidence that every worker is gone.
+    /// Callers that route from a separately-supplied worker set must not have it
+    /// emptied by a client that has yet to see its first discovery update.
+    pub fn available_instance_ids(&self) -> Option<Arc<HashSet<u64>>> {
+        let snapshot = self.routing_instances.snapshot();
+        if snapshot.discovered_ids().is_empty() {
```
PeaBrane, line 693: Known-empty discovery is collapsed into “no availability source” here. After this Client has previously discovered workers, removing the last one also makes `discovered_ids()` empty, so this returns `None`; the scheduler treats `None` as no constraint and can select that departed worker from the stale candidate map. The new lifecycle test keeps `PREFILL_PEER`, so it does not exercise this transition. Please track whether the first discovery snapshot has been observed: return `None` only while uninitialized, return `Some(empty)` for an authoritative zero-worker snapshot, and add the last-worker removal case to the lifecycle test.

PeaBrane (approved):
✅ The direct-availability refactor addresses the original fence-mirroring concerns. One last-worker edge remains inline.

kthui (approved):

lib/runtime/src/component/client.rs
hhzhang16: +1 -- There's no distinguishing here between "discovery has not initialized yet" and "discovery authoritatively removed the last worker"

PeaBrane:
This PR is being superseded by #11647; please consider closing it once the fold is pushed there.

I reconciled the final diff and review threads before making that call. The direct Client-availability design is the part to carry forward: hard routability remains distinct from transient overload and per-request migration exclusions, and unavailable workers override cache affinity without weakening explicit pins, allowlists, or routing constraints. The older TTL/mirrored-fence concerns are obsolete under that final design.

The fold also needs to resolve the remaining real review finding here: availability must distinguish an uninitialized Client (`None`) from an authoritative last-worker removal (`Some(empty)`).

lib/llm/src/kv_router.rs
PeaBrane: Addressed by the fold in #11647 at 9ee6bb4503. KV selection now samples a cached Arc<HashSet> directly from the exact router Client as an independent hard-availability inclusion set. The TTL and asynchronously mirrored fence state are not carried forward; transient overload and request-local migration exclusions remain separate.

lib/llm/src/discovery/worker_monitor.rs
PeaBrane: Superseded by the direct design folded into #11647 at 9ee6bb4503. There is no decode/prefill fence callback mirror: each KV scheduler provider closes over its own router Client. The lifecycle regression uses distinct decode and prefill Clients and verifies that prefill removals do not alter decode availability.

lib/runtime/src/component/client.rs
PeaBrane: Superseded by #11647 at 9ee6bb4503. The TTL map, expiry scan, and request-path RwLock are gone. Client publishes a cached immutable Arc<HashSet>; selection clones the Arc and checks membership, while rebuilding the set only on discovery or inhibition changes. The #11647 description records the focused release churn measurement.

lib/runtime/src/component/client.rs
PeaBrane: Addressed in #11647 at 9ee6bb4503. The routing snapshot now carries an initialization bit: startup before any discovered worker returns None, but once discovery has initialized, removing the last worker returns authoritative Some(empty). The lifecycle regression explicitly exercises partial removal, last-worker removal with stale scheduler candidates, and restoration using distinct decode and prefill Clients.

nnshah1:
Superseded by main. The DIS-2404 behavior this PR introduced — a dead/stale worker is never selected, including on the cache-affinity path — is now enforced by the `instance_avail` subsystem (`lib/runtime/src/component/client.rs`: watch channel + `report_instance_down` + periodic reconciliation), with the refactored `kv_router/routing_host` selection consuming an already-availability-filtered candidate set. Regression coverage lives in `routing_host/tests.rs` (`builtin_direct_fallback_stays_disabled_for_affinity` and the `stale_worker` affinity cases), which assert an affinity-pinned dead worker is rejected and the pin cleared. Closing as no longer needed; the sibling migration fix #11647 merged separately.

PR #11647, opened 2026-07-14
fix(migration): migrate stalled workers + release booking on re-dispatch

nnshah1:
## Summary

This PR makes worker-local failures recoverable without weakening routing constraints or the public booking contract.

- Adds `WorkerOverloaded` for one-worker capacity rejection; `ResourceExhausted` remains terminal pool/constrained-capacity exhaustion.
- Classifies the existing worker-capacity ACK as `WorkerOverloaded` and keeps both overload classes client-visible as HTTP 529.
- Records the selected worker in frontend-local migration state before exposing an immediate or streamed failure.
- Releases the failed scheduler booking synchronously before a retry can begin.
- Excludes previously failed workers on later embedded KV-routing attempts while preserving caller allowlists and required taints.
- Stops migration when the currently failing phase has an explicit request pin. Affinity-derived sticky routing may still invalidate and rebind.
- Folds #11646's final direct-`Client` availability design into KV selection without carrying over its iterative TTL/fence machinery.
- Keeps global hard availability, transient overload, and request-local migration exclusions as independent filters.
- Keeps ordinary and standalone duplicate request IDs strict: duplicates conflict regardless of worker; there is no implicit cross-worker public `/add` rebind.
- Retains internal `free_if_worker` cleanup as a defensive ownership check.
- Treats a fully busy eligible LoRA replica set as terminal `ResourceExhausted`.

## Behavioral contract

- Migration is opt-in with `migration_limit > 0` (default: `0`).
- One request has one serialized migration manager.
- A failed booking is released before the error becomes observable to that manager.
- Embedded KV routing carries frontend-local failed-worker exclusions across attempts.
- Split/standalone routing adds no new wire protocol; it preserves routing constraints and explicit pins but remains best-effort and may select the same worker again from global state.
- Explicit backend pins block migration for every phase. Prefill/decode pins block migration only when that phase fails.
- Session-affinity pins are not explicit request pins and may reassign after invalidation, but globally unavailable workers are rejected on every path.
- The direct availability provider returns `None` only before its `Client` has ever discovered a worker. Once initialized, removal of the last worker is authoritative `Some(empty)`.
- Hard availability intersects caller allowlists and required taints; it never widens constraints. Standalone routing deliberately attaches no hard-availability provider.
- Public slot-tracker duplicate IDs return `409` regardless of worker; the original booking remains unchanged.

## Validation

- Runtime, KV-router, LLM, and mocker clippy passes with warnings denied.
- Repository-wide Rust formatting and diff checks pass.
- Focused regressions cover worker-capacity ACK classification, HTTP 529 compatibility, hard-pin retry suppression, accumulated failed-worker exclusions, constraint-preserving affinity reassignment, synchronous cleanup-before-error visibility, duplicate-ID conflicts, and a real two-worker overload A→B migration.
- Direct-availability regressions cover hard unavailability overriding affinity's overload bypass, unpinned and explicitly pinned selection, independent decode/prefill `Client` state, partial worker removal, last-worker removal as `Some(empty)`, and immediate restoration.
- A temporary release-mode churn harness (removed before commit) ran five trials per point. Median request-side snapshot read plus membership check was 7.38 ns/op at 1 worker and 6.56 ns/op at 1,024 workers (2,000,000 operations/trial). Median membership rebuild plus read was 267.30 ns/op at 1 worker and 26,754.48 ns/op at 1,024 workers (20,000 operations/trial). Reads remain O(1); discovery-churn updates remain O(workers) and are not per-token work.
- The older full frontend A/B benchmark request on #11646 targeted the discarded TTL/mirrored-fence implementation. It is explicitly reconciled here rather than rerun against code that is not being folded; the retained direct-snapshot mechanism is measured above.
- Fern configuration and broken-link checks pass locally.

The exact error-specific routing integration for `StreamIncomplete` is intentionally not added; it uses the same streamed-failure recording and exclusion path covered by the shared routing and cleanup regressions.

## Related Issues

- Related: #11646


nnshah1:
## Graham-style code review

Tight, well-scoped diff — two files, two real behaviors, guard test included. The `multi_worker.rs` retry correctly `drop(table)`s the read lock before `self.free()` takes the write lock, and bounds the retry with `released_stale_booking` (no infinite loop). Clean.

### Low / Medium
- **`let _ = self.free(&request_id, decay_now);` silently discards a `Result<(), SequenceError>`.** If the release fails, the `continue` re-hits `try_insert_request` → `DuplicateRequest` again → back to the original bug, silently. The booking was just observed to exist so it should succeed, but at least `debug!` the error or add a one-line comment justifying the discard.
- **Adding `ErrorType::ResponseTimeout` to `MIGRATABLE` is a semantic decision worth calling out in the PR body.** A response-inactivity timeout can mean slow-but-alive, so failover risks duplicate execution if the original later produces output. The comment notes push_router quarantines on the same signal, which mitigates it — worth making that trade-off explicit in the description for the human reviewer, since it's the load-bearing correctness claim.

Nice tests — behavior-focused and not over-enumerated.

nnshah1:
Addressed in `6254056`:
- **Release the captured stale worker, not a fresh lookup** (CodeRabbit, Major): the re-dispatch path now frees `existing_worker` directly via a targeted `free_stale_booking`, and the index mapping is dropped atomically only if it still points at that worker (`RequestIndex::remove_request_if_worker`), so a concurrent re-dispatch can't cause the wrong worker to be freed. The result is now logged on failure instead of discarded.
- **Internal the internal ticket IDs ticket IDs in comments** (Devin + CodeRabbit): stripped per `.ai/linear-ticket-refs.md`.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1137,8 +1173,24 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
             if let Err(existing_worker) =
                 self.request_index
-                    .try_insert_request(request_id.clone(), worker, lora_name)
+                    .try_insert_request(request_id.clone(), worker, lora_name.clone())
             {
+                // a same-id re-dispatch to a DIFFERENT worker is a
+                // migration failover. Release the stale booking on the old
+                // (failed) worker and retry once so the new worker can be booked,
+                // instead of failing stream recreation with DuplicateRequest.
+                if existing_worker != worker && !released_stale_booking {
```
PeaBrane: Could this be an explicit `transfer_request` (or `rebind_request`) operation rather than teaching generic `add_request` that “same ID on another worker means migration”? A useful contract would be:

- verify the current owner is exactly the expected old worker A;
- remove A’s booking and install the request on B as one guarded operation;
- move the request index from A to B;
- publish one `TransferRequest(A → B)` replica event;
- fail or no-op if local ownership is no longer A.

That lets ordinary `add_request` keep returning `DuplicateRequest` for an existing ID—including public slot-tracker callers—while migration gets the explicit handoff it needs. I think a dedicated `ActiveSequenceEventData::TransferRequest` carrying the old worker plus the existing add payload is cleaner than coordinating separate `Free(A)` and `Add(B)` events: peers can apply the ownership and load transition as one operation. The old stream’s cleanup still needs to be disarmed/transferred or conditioned on A, but I do not think this requires a generation-based ownership protocol.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1137,8 +1173,24 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
+                    drop(table);
+                    released_stale_booking = true;
+                    // Free the specific stale worker the duplicate maps to, not
+                    // whatever a fresh lookup resolves — see free_stale_booking.
+                    if let Err(err) =
+                        self.free_stale_booking(&request_id, existing_worker, decay_now)
```
PeaBrane: One lifecycle hole appears to remain after this rebind: the old stream’s `RequestCleanup` is still armed. Once the retry books B and replaces/drops the old stream, that cleanup calls `free(request_id)`; because the index now resolves the ID to B, it can release the replacement booking. Could the transfer also disarm/transfer the old cleanup lease, or make cleanup conditional on its captured worker A so it cannot free B?

lib/kv-router/src/sequences/request_maps.rs
```diff
@@ -65,6 +65,24 @@ impl RequestIndex {
...
+        &self,
+        request_id: &RequestId,
+        worker: WorkerWithDpRank,
+    ) -> bool {
+        let removed = self
+            .request_to_worker
+            .remove_if(request_id, |_, mapped| *mapped == worker)
+            .is_some();
+        if removed {
+            self.request_to_lora.remove(request_id);
```
PeaBrane, line 80: Non-blocking soft ping: worker ownership and LoRA accounting live in separate maps. After the conditional worker removal succeeds, a concurrent rebind can install B’s worker/LoRA state before this line, allowing the stale cleanup to remove B’s LoRA entry. The underlying non-atomicity is pre-existing and this path only gives it another place to surface, so I’m okay deferring consolidation to a follow-up.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1869,6 +1921,47 @@ mod tests {
         assert_eq!(active_request_count(&sequences, worker), 1);
     }
 
+    // a migration re-dispatch of the same request-id to a DIFFERENT
+    // worker must release the stale booking on the old (failed) worker and
+    // re-book to the new one, instead of failing with DuplicateRequest — which
+    // broke stream recreation on failover.
+    #[tokio::test]
+    async fn same_id_redispatch_to_new_worker_releases_stale_booking() {
```
PeaBrane: This direct sequential test is useful, but could we also cover the observable contracts around the handoff?

- exercise `Migration → KvPushRouter → RequestGuard` and verify old-stream cleanup cannot release B;
- verify an ordinary cross-worker duplicate remains `DuplicateRequest` (and the public API remains a conflict), with the success case using the explicit transfer operation;
- exercise replica sync through `TransferRequest(A → B)` and verify the peer removes A’s booking, installs B’s booking, and updates both worker loads.

If correctness relies on exactly one migration state machine and serialized redispatches per request, please document or reference that invariant as well. Under it, further owner-aware replica hardening is optional rather than required here.

PeaBrane (commented):
Thanks for tackling the stalled-worker handoff. I left two correctness concerns around request-ID-based rebinding and cleanup, plus a few non-blocking invariant and coverage follow-ups. The core slot tracker looks like the right layer, but I think migration should use an explicit transfer operation while ordinary adds preserve their existing duplicate-ID contract.

kthui (approved):
the `llm/src/migration.rs` looks good to me!

PeaBrane:
Redoing another round of review with a relaxed invariant. Ignore my previous review

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1137,8 +1173,24 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
+                    .try_insert_request(request_id.clone(), worker, lora_name.clone())
             {
+                // a same-id re-dispatch to a DIFFERENT worker is a
+                // migration failover. Release the stale booking on the old
+                // (failed) worker and retry once so the new worker can be booked,
+                // instead of failing stream recreation with DuplicateRequest.
+                if existing_worker != worker && !released_stale_booking {
+                    drop(table);
+                    released_stale_booking = true;
+                    // Free the specific stale worker the duplicate maps to, not
```
PeaBrane: This helper correctly captures old worker A, but worker ownership is lost in the old stream’s cleanup path. `RequestCleanup` stores only the request ID, so after this handoff books B, dropping A’s stream calls `chooser.free(R)`, resolves the current owner B, and removes B’s booking.

Could we promote this into a reusable `free_if_worker(R, A)`, capture the selected worker in `RequestGuard`/`RequestCleanup`, and use it for finish, abort, and drop cleanup? A missing request or ownership mismatch should be a harmless no-op. This completes the existing implicit-migration design without needing a transfer API.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1137,8 +1173,24 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
+                    // whatever a fresh lookup resolves — see free_stale_booking.
+                    if let Err(err) =
+                        self.free_stale_booking(&request_id, existing_worker, decay_now)
+                    {
+                        tracing::debug!(%request_id, ?existing_worker, %err, "failed to release stale booking on migration re-dispatch");
+                    }
+                    continue;
```
PeaBrane: In admission-managed mode, the old stream still owns its `RequestLifecycleLease` while `RetryManager` creates the replacement. The queue actor therefore rejects the same request ID as an active admission before execution reaches this slot-level retry.

If migration is supported with admission policies, the old ticket needs to be retired and acknowledged before rescheduling, or the actor needs a ticket-aware replacement path. Otherwise, could the PR explicitly scope this behavior to non-admission mode?

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -814,6 +814,41 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
...
+        ) {
+            Ok(()) | Err(SequenceError::RequestNotFound { .. }) => {}
+            Err(err) => return Err(err),
+        }
+        self.request_index
+            .remove_request_if_worker(request_id, worker);
+        self.enqueue_publish_event(ActiveSequenceEvent {
+            request_id: request_id.clone(),
+            worker,
+            data: ActiveSequenceEventData::Free,
```
PeaBrane, line 843: The replica `Free` handler’s request-ID-only behavior predates this PR, but this handoff makes it concretely reachable: this code intentionally emits `Free(R, A)` while ownership moves from A to B, so the event’s worker is now an ownership constraint rather than informational metadata.

`apply_replica_event` currently ignores `event_worker` and frees whichever worker is currently indexed by R. If the replica has already applied `Add(R, B)`, it can incorrectly free B.

Could replica cleanup mirror `free_stale_booking`: free R specifically from `event_worker`, remove the index only if it still maps `R → event_worker`, and preserve a newer `R → B` mapping? The existing `replica_sync_free_uses_canonical_worker_for_collapsed_load` test locks in the old behavior and should be revised. No new event type is needed.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1137,8 +1173,24 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
             };
             if let Err(existing_worker) =
                 self.request_index
-                    .try_insert_request(request_id.clone(), worker, lora_name)
+                    .try_insert_request(request_id.clone(), worker, lora_name.clone())
             {
+                // a same-id re-dispatch to a DIFFERENT worker is a
```
PeaBrane: Since this intentionally makes cross-worker duplication an implicit migration, could we update the standalone selection and slot-tracker contracts?

```text
same ID + same worker       → 409 DuplicateRequest
same ID + different worker → state-changing rebind
```

The current docs say all duplicate IDs conflict. They should also warn that retrying an ambiguous `/add` against a different worker is not idempotent. This is documentation of the accepted design, not a request for an explicit transfer API.

PeaBrane (commented):
Re-reviewing under the intended implicit-migration contract: same-ID/different-worker adds perform a serialized rebind, while cleanup is worker-targeted. This keeps the implementation small—no transfer API, transfer event, or generation protocol is needed. The five comments below supersede my earlier hidden review.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -1869,6 +1921,47 @@ mod tests {
...
+            worker,
+            lora_name: None,
+        };
+
+        sequences.add_request(req(old_worker), decay_now).unwrap();
+        assert_eq!(active_request_count(&sequences, old_worker), 1);
+
+        // Migration failover: same request-id re-dispatched to a new worker.
+        sequences
+            .add_request(req(new_worker), decay_now)
```
PeaBrane: This verifies the direct sequential slot handoff, but not the lifecycle boundary where the current bug appears. Could we add regressions covering:

- B remains booked when A’s old `RequestGuard` cleanup arrives;
- legacy and lifecycle/admission cleanup modes;
- replica `Free(R, A)` cannot remove `R → B`;
- public API behavior: same-worker duplicate remains `409`, while a different-worker duplicate rebinds successfully?

No simultaneous-migration stress test is needed under the serialized-redispatch invariant.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: @nnshah1 One remaining documentation follow-up: the slot-tracker page is updated, but `standalone-selection.md` still says every duplicate booking conflicts with `409`. Could you update that contract as well?

PeaBrane (approved):
✅

lib/kv-router/src/services/slot_tracker/registry.rs
```diff
@@ -741,6 +741,52 @@ mod tests {
...
     }
 
+    /// The public `/add` contract after same-id rebind landed: a duplicate on the
+    /// *same* worker is still a conflict (`409` at the HTTP layer), while a
+    /// duplicate on a *different* worker rebinds instead of failing.
+    ///
+    /// The rebind is what makes migration failover work; the conflict is what
+    /// keeps accidental id reuse an error.
+    #[tokio::test]
+    async fn duplicate_add_conflicts_on_same_worker_and_rebinds_across_workers() {
```
hhzhang16: This pins the public cross-worker rebind, but not the documented late-cleanup guarantee. `/free` carries only `request_id` and calls `tracker.free(request_id)`, so after `/add(R,A)` → `/add(R,B)`, a late `/free(R)` from A resolves the current owner and releases B.

Is the standalone contract that superseded attempts can never emit lifecycle cleanup? If so, the docs should not claim cleanup is worker-targeted. Otherwise `/free` needs worker/attempt ownership and should use `free_if_worker`. Could this test cover the three-call sequence?

lib/llm/src/migration.rs
```diff
@@ -405,6 +420,41 @@ mod tests {
...
+
+        let stream_incomplete = DynamoError::builder()
+            .error_type(ErrorType::Backend(BackendError::StreamIncomplete))
+            .message("stream ended before completion")
+            .build();
+        assert!(
+            is_migratable(&stream_incomplete),
+            "StreamIncomplete (truncated stream from departed worker) must be migratable"
+        );
+    }
```
hhzhang16, line 494: `StreamIncomplete` is migratable here, but it is absent from push-router `is_inhibited`, and `RetryManager` carries no failed-worker exclusion—only `migration_attempt`. Before discovery/#11646 fencing catches up, KV or session affinity can select the same source worker again.

Should `StreamIncomplete` quarantine the worker, or should migration carry an attempt-local source-worker exclusion? Please add a routing-level regression proving A’s `StreamIncomplete` retries on B rather than merely testing `is_migratable()`.

The PR description says the migratable set aligns with is_inhibited, but it does not for StreamIncomplete.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Resolved by the final design: the implicit cross-worker rebind was removed. Ordinary `add_request` preserves `DuplicateRequest` for any existing ID, including public slot-tracker callers. Migration now synchronously releases the failed booking before retrying, so no transfer operation or replica transfer event is needed.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Addressed. `RequestCleanup` now captures the selected worker and uses `free_if_worker`; streamed failures also await `guard.abort()` before yielding the error to Migration. A late cleanup therefore cannot release a different worker's booking.

lib/kv-router/src/sequences/request_maps.rs
PeaBrane: Closing as the accepted pre-existing/out-of-scope case. The final design removed implicit same-ID rebinding, and request IDs remain globally unique. This PR no longer creates the concurrent rebind window described here, so consolidating the worker/LoRA maps would be disproportionate.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Resolved under the final invariant. There is no transfer operation now. Coverage includes a real `Migration -> KvPushRouter` A-to-B retry, cleanup-before-error visibility, worker-targeted cleanup mismatch as a no-op, and cross-worker duplicate conflict. The serialized migration/release-before-retry invariant is documented.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Addressed as suggested. `free_if_worker(R, A)` is reusable through the scheduler, and `RequestGuard`/`RequestCleanup` captures A for finish, abort, and drop cleanup. Missing requests and ownership mismatches are harmless.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Resolved by the current lifecycle boundary. `RequestLifecycleLease` covers only admission before handoff and is disarmed before the scheduling response returns. The long-lived booking belongs to `RequestGuard`, whose abort is awaited before Migration can schedule the replacement, so no old admission ticket overlaps the retry.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Superseded by removing implicit cross-worker rebind and the expanded replica ownership contract. The failed local booking is released synchronously before retry, and replica events retain their canonical request-owner behavior. The concrete `Add(B)`-before-late-`Free(A)` handoff introduced by the earlier design no longer exists.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Superseded and documented for the final contract: duplicate reservation/request IDs conflict with `409` regardless of target worker, and the original booking is unchanged. There is no state-changing public rebind or ambiguous `/add` idempotency exception.

lib/kv-router/src/sequences/multi_worker.rs
PeaBrane: Resolved for the final design. Rebind-specific public and replica cases were removed with that contract. The retained regressions cover synchronous cleanup before retry, defensive worker-targeted cleanup, duplicate conflict across workers, and the actual embedded A-to-B migration path.

lib/kv-router/src/services/slot_tracker/registry.rs
PeaBrane: Superseded by removing public cross-worker rebind. `/add(R, B)` now conflicts while R is booked on A, so the three-call late-free scenario cannot transfer ownership to B. Public `/free` remains request-ID based, and the docs no longer claim worker-targeted public cleanup.

lib/llm/src/migration.rs
PeaBrane: The production gap is addressed with frontend-local `MigrationState`: every streamed failure records the selected worker before the error is exposed, and later unpinned KV selections exclude all recorded workers while preserving constraints. We are intentionally not adding a separate `StreamIncomplete`-specific routing integration; the shared exclusion mechanism and stream cleanup ordering have focused regressions. Closing with that coverage decision accepted.

ishandhanani (approved):

PR #11460, opened 2026-07-09
feat(llm): route requests through encode workers

connorcarpenter15:
## Overview:

Add the backend-agnostic encoder routing hop that sends multimodal requests to a surface-less Encode worker before aggregated or prefill execution.

## Summary

- add an optional round-robin `EncoderRouter` to the preprocessed request pipeline
- activate and deactivate encoder routing through model discovery
- scope encoder endpoints and consumers by model namespace
- preserve passthrough behavior when no encoder is configured or available

## Details:

This is stack PR 1 of 4 for [DIS-2034](https://linear.app/nvidia/issue/DIS-2034/vllm-add-separate-encode-worker-in-unified-backend). It builds on the merged encoder-result contract in #10969 and Encode role/discovery foundation in #10970.

The router consumes the Encode worker's terminal `encoder_result`, attaches it to the downstream request, and carries the worker trace link forward. Discovery retains endpoint state across consumer rebuilds and deactivates the hop when Encode workers leave.

## Where should the reviewer start?

- `lib/llm/src/kv_router/encoder_router.rs`
- `lib/llm/src/discovery/model_manager.rs`
- `lib/llm/src/discovery/watcher.rs`
- `lib/llm/src/entrypoint/input/common.rs`

## Validation

- focused Rust tests: 4 passed
  - encoder router registration and waiter cleanup
  - model-namespace isolation
  - object-shaped Encode terminal consumption
  - rejection of terminals without `encoder_result`
- `cargo fmt --all -- --check`
- `git diff --check`
- full-stack GPU validation is recorded in #11449

## Stack

1. **This PR:** generic encoder routing and discovery
2. #11461 — unified vLLM Encode engine
3. #11462 — unified vLLM encoder handoff consumer
4. #11449 — launchers, GPU profiles, and documentation

## Related Issues

**🔗 This PR is linked to an issue:**

- [DIS-2034: vLLM: Add separate encode worker in unified backend](https://linear.app/nvidia/issue/DIS-2034/vllm-add-separate-encode-worker-in-unified-backend)

## Summary by CodeRabbit

* **New Features**
  * Added support for encoder-based routing in multimodal inference flows.
  * Improved handling for encode-only workers, including automatic routing setup and reactivation when needed.
  * Added a new routing stage so requests can pass through encoding before prefill when applicable.

* **Bug Fixes**
  * Better cleanup when workers are removed, preventing stale routing state.
  * Improved routing behavior across namespaces so encoder readiness is tracked more reliably.

PR #11205, opened 2026-07-03
fix:  Fix guided decoding for force-reasoning parsers

indrajit96:
## Overview:

Fix guided decoding for force-reasoning parsers when `tool_choice=required` or a named tool is used.
This PR **detects the generated stream shape and preserves both reasoning and structured tool calls**.

Closes DIS-2306

## Details:

vLLM/Sglang may return either bare guided JSON or reasoning</think>JSON. 
Dynamo previously bypassed force-reasoning parsers unconditionally, which could drop `reasoning_content` or leak reasoning and JSON into content.

For allowlisted force-reasoning parsers, required and named tool calls can produce two output shapes:

| Backend output | Dynamo behavior |
|---|---|
| `{...}` or `[{...}]` | Bypass reasoning parsing and send the JSON directly to the tool-call jail. |
| `reasoning</think>{...}` | Extract `reasoning_content`, then send the remaining JSON to the tool-call jail. |

The decision is made once from the first non-whitespace streaming content and
retained for the rest of the response. This preserves the existing bare-JSON
behavior while allowing reasoning and structured tool calls to coexist.
Non-force parsers continue to use their existing reasoning-parsing path.

### SGLang engine integration

Dynamo calls SGLang's low-level `async_generate()` API, bypassing the SGLang
OpenAI layer that normally derives the per-request reasoning state. This
change:

- Allows native and Dynamo reasoning parsers together: SGLang controls when
  guided JSON begins, while Dynamo constructs `reasoning_content`.
- Propagates `require_reasoning=true` for required or named tool choices when
  thinking is enabled, across aggregated and disaggregated engine paths.
- Preserves request-level thinking disablement and drops the optional argument
  for older SGLang versions that do not support it.

Plain chat, `tool_choice=auto`, `tool_choice=none`, and explicitly disabled
thinking are unchanged. The signal is scoped to required and named tool-call
guided decoding, not arbitrary `response_format` constraints.

### End-to-end validation

Tests used aggregated Dynamo deployments with the native vLLM reasoning parser,
Dynamo reasoning parser, and Dynamo tool-call parser configured.

The following modes were exercised with raw requests, nonstreaming responses,
and SSE captures:

- Plain reasoning.
- Optional tool choice.
- Required tool choice.
- Named tool choice.
- Required streaming.
- Named streaming.
- Thinking enabled, disabled, and default where supported.

Validated models:

Core force-reasoning validation:

- Kimi K2.6 (`kimi_k25`)
[...]

PR #8901, opened 2026-04-30
fix: deepseek v4 inter-leaved tool and reasoning parser

biswapanda:
#### Overview:

Fixes DeepSeek V4 tool-continuation reasoning parsing without regressing Kimi K2.5. In the failing DSv4 flow, the request ends with a `role=tool` message and the DSv4 formatter has already seeded `<think>` into the next assistant prompt. The model can then stream reasoning text followed by `</think>` without an opening `<think>` token in the completion. If the postprocessor clears `prompt_injected_reasoning` for all tool-continuation turns, the closing tag is treated as normal content and leaks to the client.

This PR keeps DSv4 in injected-reasoning mode for that path, while preserving the Kimi K2.5 behavior introduced by `5671d8e374021ca1aeb399fa9729285903ea6645`: direct post-tool final answers for `kimi_k25` should remain normal `content`, not be mislabeled as `reasoning_content`.

#### Details:

- Removes the broad `last_is_tool` suppression that caused the DSv4 `</think>` leak.
- Reintroduces post-tool reasoning suppression specifically for `kimi_k25`, where the parser is force-reasoning and direct tool follow-up answers must bypass reasoning parsing.
- Adds a DSv4 stream regression test for `user -> assistant(tool_call) -> tool -> assistant` where the completion starts inside reasoning and emits `</think>`.
- Adds a Kimi K2.5 regression test for the same conversation shape where the final direct answer must stay in `content`.
- Keeps `tool_choice=required` / named guarded from reasoning parsing for guided-decoding JSON outputs.

#### Where should the reviewer start?

- `lib/llm/src/preprocessor.rs` — parser-specific post-tool reasoning gate.
- `lib/llm/tests/postprocessor_parsing_stream.rs` — DSv4 `</think>` leak regression and Kimi K2.5 non-regression coverage.
- `lib/llm/tests/test_reasoning_parser.rs` — parser-level DSv4 injected-reasoning tests.

#### Validation:

Ran with `CUDARC_CUDA_VERSION=12090` because local CUDA 13.2 is rejected by `cudarc 0.19.3` during build.

- `cargo test -p dynamo-llm --no-default-features --test postprocessor_parsing_stream tool_continuation -- --nocapture`
- `cargo test -p dynamo-llm --no-default-features --test test_reasoning_parser deepseek_v4_post_tool -- --nocapture`
- `cargo test -p dynamo-llm --no-default-features --test test_reasoning_parser kimi_k25 -- --nocapture`
[...]

PR #11653, opened 2026-07-14
fix: Fix multi-turn reasoning in Kimi 2.5/2.6

indrajit96:
## Summary

Fix Kimi K2.6 post-tool reasoning and make request-level thinking controls consistent across prompt rendering and response parsing.

Dynamo's existing Kimi compatibility guard skipped reasoning parsing whenever the last input was a tool result. K2.6 can reason after a tool result and emit an implicit reasoning body followed by `</think>`, so skipping the parser leaked
private reasoning and the closing marker into `content` while `reasoning_content` remained empty.

The initial fix exposed a second issue: the chat template, SGLang gate, and postprocessor interpreted `thinking`, `enable_thinking`, omitted values, strings, and numbers differently. For example, Jinja treats the non-empty string
`"false"` as true while `Value::as_bool()` rejects it. This could prompt the model to reason while disabling the parser.

## Details

- Normalize the thinking argument once before prompt rendering and postprocessing.
- Accept both `thinking` and `enable_thinking` and normalize boolean, string, and numeric spellings to canonical booleans.
- Treat omitted thinking as enabled for the `kimi_k25` parser, matching the Kimi K2.5 and K2.6 template default.
- Write the normalized value back to both request keys so the template, SGLang reasoning gate, and Dynamo postprocessor consume the same value.
- Parse Kimi post-tool reasoning whenever effective thinking is enabled; an explicit false value still disables reasoning.
- Cover streaming `</think>` markers split across chunk boundaries.
- Keep stream-shape buffering out of this change. Marker-free K2.5 post-tool answers remain a documented model deviation rather than delaying every response until `</think>` or end-of-stream.

Supported equivalent values include:

| Enabled | Disabled |
|---|---|
| `true`, `"true"`, `"TRUE"`, `"yes"`, `"on"`, `1` | `false`, `"false"`, `"no"`, `"off"`, `0` |

Before:

```json
{
  "content": "The tool returned 15°C and cloudy.</think>The current weather in London is 15°C and cloudy.",
  "reasoning_content": null
}
```

After:

```json
{
  "content": "The current weather in London is 15°C and cloudy.",
  "reasoning_content": "The tool returned 15°C and cloudy."
}
```

## Validation

- Postprocessor stream tests: 45/45 passed on the latest head.
- Focused preprocessor unit tests: 28/28 passed, including the parameterized normalization test.
- Focused GB200 K2.6 multiturn validation: 12/12 passed across `consume_prior_tool_result` and `e2e_parallel_weather_final_answer`, streaming and nonstreaming.
- Omitted-thinking K2.6 post-tool validation passed in both streaming and nonstreaming modes with correctly separated reasoning and content.
- Full 38-case K2.6 run had no reasoning-marker leaks. The non-passing cases were model decision, token-budget, or generation-timeout outcomes rather than response-parser leakage.
- K2.5 non-buffering experiment: 98/100 responses produced the expected reasoning/answer split; two produced follow-up tool calls; none produced a marker-free direct answer.

## Where Should the Reviewer Start?

- `lib/llm/src/preprocessor.rs` — normalizes request-level thinking and applies the effective reasoning mode consistently.
- `lib/llm/tests/postprocessor_parsing_stream.rs` — covers K2.6 explicit and omitted thinking, post-tool reasoning, and split closing-marker boundaries.
- `lib/llm/PREPROCESSOR_CASES.md` — documents the normalized Kimi behavior and the marker-free K2.5 limitation.
- `docs/reasoning/README.md` — documents the model-level reasoning opt-out limitation.

## Related Issues


lib/llm/src/preprocessor.rs
```diff
@@ -2062,22 +2062,29 @@ impl OpenAIPreprocessor {
...
+        // compatibility behavior unless the request explicitly enables thinking:
+        // Kimi K2.6 may then emit implicit reasoning followed by `</think>`, which
+        // must be parsed to avoid leaking reasoning and the close tag into content.
         let last_is_tool = matches!(
             request.inner.messages.last(),
             Some(ChatCompletionRequestMessage::Tool(_))
         );
+        let kimi_thinking_explicitly_enabled = request
+            .chat_template_args
+            .as_ref()
```
keivenchang: This gate requires explicit `thinking: true`, but elsewhere kimi effective-thinking is ON when omitted (`!= Some(false)` in #11205). Do default-config K2.6 users keep the leak — deliberate or accident?

lib/llm/src/preprocessor.rs
```diff
@@ -2062,22 +2062,29 @@ impl OpenAIPreprocessor {
...
-        // injected-reasoning signal is required to avoid leaking the close tag.
+        // mislabels that marker-free answer as reasoning_content. Preserve that
+        // compatibility behavior unless the request explicitly enables thinking:
+        // Kimi K2.6 may then emit implicit reasoning followed by `</think>`, which
+        // must be parsed to avoid leaking reasoning and the close tag into content.
         let last_is_tool = matches!(
             request.inner.messages.last(),
             Some(ChatCompletionRequestMessage::Tool(_))
         );
+        let kimi_thinking_explicitly_enabled = request
```
rmccorm4: I think both Kimi K2.5 and K2.6 are thinking by default, even when this thinking=true flag is omitted, it would only be disabled by thinking=false -- do we need to handle the case when thinking flag is omitted?

https://github.com/vllm-project/vllm/blob/cdaa40d2a8c9297349669fc6e7db6dcb282df0b8/vllm/parser/kimi_k2.py#L52-L170

Separately, I think Kimi K2.7 is force-reasoning, regardless of request thinking flag. I'm calling this out here for awareness, but we may not reasonably be able to address this in this PR - let me know.

keivenchang (dismissed):
Ran the stream suite (45/45) and preprocessor units (66/66) on this head; legacy suppression + new thinking=true tests pass, and the split-marker coverage is the right shape. One question inline on the default-omitted case.

lib/llm/src/preprocessor.rs
```diff
@@ -2062,22 +2062,29 @@ impl OpenAIPreprocessor {
...
+            .chat_template_args
+            .as_ref()
+            .and_then(|args| args.get("thinking"))
+            .and_then(serde_json::Value::as_bool)
+            == Some(true);
```
KrishnanPrash: If a client sends the value as a string ("thinking": "true") instead of a bool, as_bool returns None, the gate reads "off", suppression stays on, and the reasoning + </think> leak straight back into content. I ran the current fix locally against five request variants (identical streamed model output each time, only the thinking value differs):

| Trial | thinking value | Type | reasoning_content | </think> leaks? |
|-------|------------------|------|---------------------|-------------------|
| A | true (+ tool_choice=required) | bool | parsed correctly | No |
| B | "true" | string | empty | **Yes** |
| C | "false" | string | empty | **Yes** |
| D | true | bool | parsed correctly | No |
| E | omitted | — | empty | Yes (intended legacy behavior) |

KrishnanPrash: Could do something like this:
```
fn thinking_truthy(args: Option<&HashMap<String, Value>>) -> Option<bool> {
    match args?.get("thinking")? {
        Value::Bool(b) => Some(*b),
        Value::String(s) => match s.to_ascii_lowercase().as_str() {
            "true" | "1"            => Some(true),
            "false" | "0" | ""      => Some(false),
            _                        => None,
        },
        Value::Number(n) => Some(n.as_f64().is_some_and(|v| v != 0.0)),
        _ => None,
    }
}
// call site:
let kimi_thinking_explicitly_enabled =
    Self::thinking_truthy(request.chat_template_args.as_ref()) == Some(true);
```

KrishnanPrash (approved):
Minor comment boolean/string handling

keivenchang:
For the `thinking`-flag half — normalize once, so `true` = `"true"` = `"yes"` = `"on"` = `1`:

1. **Compose the two helpers that already exist** — `dynamo_runtime::config::is_truthy` (`lib/runtime/src/config.rs:425`) for strings + `dynamo_renderer::thinking_bool_from_args` (`renderer/src/lib.rs:66`, already called twice in this file; its only gap is `as_bool()`-only). Same shape as @KrishnanPrash's `thinking_truthy`, homed where the key already lives.
2. **Write the normalized bool back into the render context** before rendering — Jinja treats string `"false"` as truthy, so template and gate diverge unless both consume the same value.
3. **Single-reader**: 3 call sites currently read the raw key with 3 different rules (`== Bool(false)`, `.as_bool() == Some(true)`, `!= Some(false)`) — that's where the omitted/string bugs came from. Omitted -> Kimi default ON.
4. **One parametrized spelling test** asserting gate decision and rendered prompt agree.

Side note: dynamo has 8+ hand-rolled truthy parsers with divergent sets (`lora`/`agents` miss `on`; `lora/source.rs` accepts only `true`) — separate cleanup ticket, not this PR. Composes with the stream-shape buffering.

indrajit96:
> Minor comment boolean/string handling

Thanks. Agreed on coercing the request value. I implemented the same idea in 8eae4e4229, but reused Dynamo’s existing truthy handling instead of adding another parser:

> Suggestion for the rework (the `thinking` flag half — separate from the stream-shape buffering): normalize the flag **once**, so `true` = `"true"` = `"yes"` = `"on"` = `1` and there is no ambiguity anywhere downstream.
> 
> **1. Don't write a new helper — the two pieces already exist. Compose them.**
> 
> * `dynamo_runtime::config::is_truthy` (`lib/runtime/src/config.rs:425`) is the canonical string-truthy, already importable from `preprocessor.rs`:
>   ```rust
>   pub fn is_truthy(val: &str) -> bool {
>       matches!(val.to_lowercase().as_str(), "1" | "true" | "on" | "yes")
>   }
>   ```
> * `dynamo_renderer::thinking_bool_from_args` (frontend-crates `renderer/src/lib.rs:66`) is the existing shared thinking-arg reader — **already called twice in this file**, and it reads both key spellings (`thinking`, `enable_thinking`). Its one gap is exactly the bug under discussion: it only does `as_bool()`, so the string `"true"` falls through to `None`.
> 
> The clean fix: extend `thinking_bool_from_args` to coerce `Value::String` via the `is_truthy` set (and `Value::Number` != 0), then have this PR's gate call it instead of hand-reading the raw key. One owner for "thinking arg -> bool". (It lives in frontend-crates, so that's a small renderer patch + version bump — or a local shim here until the bump lands, deleted after.)
> 
> Building on @KrishnanPrash's `thinking_truthy` idea — this is the same shape, just homed in the helper that already owns the key.
> 
> **2. Keep template and parser in sync — this is the subtle one.** Jinja truthiness treats any non-empty string as truthy, so the _string_ `"false"` renders a thinking-ON prompt while a parser-side `.as_bool()` gate reads it as not-enabled: prompt says "reason", parser says "don't", and the leak returns through the back door. Fix: after normalizing, **write the normalized bool back into the render context** (overwrite `chat_template_kwargs.thinking` before the template runs). Then the template and every gate consume the same value by construction and cannot diverge.
> 
> **3. Make the raw key single-reader.** Today three call sites read `thinking` with three different rules — `== Bool(false)` in `is_reasoning_disabled_by_request`, `.as_bool() == Some(true)` in this PR's gate, `!= Some(false)` in #11205's `sglang_effective_reasoning_enabled` — which is exactly where the omitted/string bugs came from. After this change, the raw key should be read in exactly one place (the extended helper); everything else consumes the returned bool. Omitted -> Kimi default is thinking ON (both K2.5 and K2.6, per Ryan).
> 
> **4. Test that pins it:** parametrize one test over `true, "true", "TRUE", "yes", "on", 1` (all must behave identically to bool `true`) and `false, "false", "no", "off", 0` (identical to bool `false`) plus omitted (= Kimi default ON), asserting the gate decision and the rendered prompt agree in every spelling.
> 
> Side note while researching this: dynamo has at least 8 hand-rolled truthy parsers with _divergent_ accepted sets (`lora/config.rs:149` and `protocols/agents.rs:105` accept `true|1|yes` but not `on`; `lora/source.rs:200` accepts only `true`; `runtime/config.rs`/`fpm_trace`/`local_model`/`nvtx` accept the full set). So `=on` works for some flags and silently not others — same bug class one layer down. Worth a separate cleanup ticket; no need to grow this PR.
> 
> This closes the spelling ambiguity entirely; the K2.5-vs-K2.6 stream-shape question (marker-free direct answer vs `reasoning` + close marker) still needs the buffering you're building — the two fixes compose.

Thanks for the detailed rework suggestion. I followed this desing in 8eae4e4229.

One implementation detail: I used the local-shim option to avoid a renderer crate change and version bump in this PR. 

1. Uses dynamo_runtime::config::is_truthy for strings, supports nonzero numbers, and recognizes both thinking and enable_thinking.
2. Normalizes the value before rendering and writes the resulting boolean to both keys, keeping Jinja and the parser gates synchronized.
 3. Is the only place that interprets unnormalized request values. Downstream gates use thinking_bool_from_args against the canonical booleans.
 4. Defaults omitted Kimi thinking to enabled.

The parameterized test covers true, "true", "TRUE", "yes", "on", 1, false, "false", "no", "off", 0, and omitted. It asserts that prompt rendering, the SGLang gate, and the postprocessor gate agree for every spelling.

I agree the other hand-rolled truthy parsers should be consolidated separately rather than expanding this PR.

For the stream-shape half, I ultimately left buffering out. 
On the non-buffering PR build, 98/100 K2.5 post-tool requests produced the expected reasoning/answer split, two produced follow-up tool calls, **and none produced a marker-free direct answer**. 
Given the hot-path latency cost, we are treating marker-free K2.5 output as a known model deviation instead of buffering every response.

lib/llm/src/preprocessor.rs
indrajit96: Thanks for catching this. 
The raw `.as_bool()` gate has been removed in
`8eae4e4229`. The request value is now normalized before prompt rendering and postprocessing, and the normalized boolean is written to both `thinking` and `enable_thinking`.

| Trial | Request value | Normalized value | Behavior after the fix | Expected result |
|---|---|---:|---|---|
| A | `true` with `tool_choice=required` | `true` | Template and parser enable reasoning | Reasoning is parsed; no leak |
| B | `"true"` | `true` | String is accepted by the canonical truthy rules | Reasoning is parsed; no leak |
| C | `"false"` | `false` | Template and parser both disable reasoning | Model should not emit reasoning |
| D | `true` | `true` | Template and parser enable reasoning | Reasoning is parsed; no leak |
| E | omitted | `true` for `kimi_k25` | Kimi's default-enabled behavior is applied consistently | Reasoning is parsed; no leak |

lib/llm/src/preprocessor.rs
indrajit96: fixed in 8eae4e4229. 
Omitted Kimi thinking now normalizes to true before rendering and parsing, so all gates use the same default. 
I also validated the omitted K2.6 post-tool case on GB200 in streaming and non-streaming modes; both split reasoning and content correctly with no marker leakage.

lib/llm/src/preprocessor.rs
indrajit96: fixed in 8eae4e4229. 
Omitted thinking now normalizes to true for Kimi; only explicit false disables it.
FOr `K2.7` added a note in docs part of this PR

lib/llm/PREPROCESSOR_CASES.md
```diff
@@ -69,22 +69,23 @@ behavior.
...
+Thinking arguments are normalized to a JSON boolean before prompt
+rendering so the template, SGLang guided-output gate, and postprocessor
+use the same value. The accepted enabled values are `true`, `"true"`,
+`"yes"`, `"on"`, and nonzero numbers. Kimi defaults an omitted value to
+enabled; `thinking=false` and equivalent falsey values disable PRE.2.
+
+Kimi tool-continuation turns therefore honor prompt-injected reasoning
+for both explicit and default-enabled thinking. K2.5 can rarely return a
+marker-free answer despite a thinking prompt; without output buffering,
+that model deviation is classified as reasoning content.
```
keivenchang, line 83: Non-blocking: if this ever stops being rare, nonstreaming is fixable for free — responses aggregate before returning, so an EOF reclassify (no `</think>` seen + zero content emitted -> move reasoning to content) costs no latency. Only streaming needs buffering; this line would then say "streaming" specifically.

lib/llm/src/preprocessor.rs
```diff
@@ -539,6 +536,45 @@ impl OpenAIPreprocessor {
         default_enabled
     }
 
+    fn normalize_thinking_arg(
```
keivenchang, line 539: Minor: with conflicting keys (`{"thinking": "true", "enable_thinking": false}`) loop order makes `thinking` win and overwrite the explicit false — worth a comment or test case to pin. Docs also omit string `"1"` from the accepted values.

keivenchang (approved):
Re-reviewed 8eae4e4 — looks good. Normalization is right: one helper owns the key, write-back keeps template and gates in sync, the spelling test pins it. Ran suites on this head (45/45 stream, 67/67 unit); probed K2.5 marker-free post-tool — swallowed into reasoning_content as documented, acceptable if rare. Two non-blocking notes inline.

rmccorm4 (approved):
Approving for dynamo-docs change

lib/llm/src/preprocessor.rs
```diff
@@ -2120,11 +2139,9 @@ impl OpenAIPreprocessor {
         // Try to parse reasoning content only if parser is configured.
         let should_parse_reasoning = self.runtime_config.reasoning_parser.is_some()
             && !reasoning_disabled_by_request
-            && !suppress_reasoning_after_tool
             && !skip_reasoning_for_guided_json;
         let should_strip_disabled_reasoning_start = reasoning_disabled_by_request
```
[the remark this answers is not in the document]
indrajit96: Correct—this is intentional. The old unconditional post-tool suppression fixed marker-free K2.5 answers but broke K2.6’s default reasoning output. We tested 100 K2.5 post-tool requests without buffering: 98 split correctly, two made follow-up tool calls, and none produced a marker-free answer. We accepted the documented K2.5 edge case rather than buffering every response; explicit thinking=false still disables parsing.

lib/llm/src/preprocessor.rs
```diff
@@ -501,7 +498,7 @@ impl OpenAIPreprocessor {
             Some("deepseek_v3_2" | "deepseek_v4" | "deepseek-v4" | "deepseekv4") => {
                 Self::deepseek_renderer_reasoning_enabled(chat_template_args, true)
             }
-            Some("gemma4" | "gemma-4") => bool_arg("enable_thinking") == Some(true),
+            Some("gemma4" | "gemma-4") => thinking_enabled == Some(true),
 
             // SGLang's Mistral reasoner is active only for a concrete effort.
             Some("mistral") => Self::mistral_reasoning_enabled(chat_template_args),
```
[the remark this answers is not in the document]
indrajit96: This is intentional. thinking and enable_thinking are treated as aliases by the shared reader. In production, normalization runs before the only call path and writes both canonical keys, so behavior cannot depend on which spelling the client used. The helper is private; only tests call it directly without normalization. The normalization test covers both key spellings.

lib/llm/PREPROCESSOR_CASES.md
indrajit96: Agreed. 
The current wording describes today’s behavior for both paths. 
If marker-free output becomes common, we can reclassify at EOF for non-streaming with no latency cost and limit buffering to streaming. 
Given 0/100 marker-free samples, I’m leaving this unchanged for now.

lib/llm/src/preprocessor.rs
indrajit96: Makes sense to add a comment, good catch.
  - Documented that thinking wins conflicting aliases.
  - Added a conflicting-key regression test.
  - Added string "1" normalization coverage.

glamr-agent:
Template: existing-request
Review request: https://github.com/ai-dynamo/dynamo/pull/11653
Engine: vllm

## User intent

This is a request-changes re-entry for the Dynamo issue about Kimi K2.6 post-tool reasoning parsing. The caller still wants the issue fixed without creating a duplicate Kimi parser implementation. The prior review agreed that upstream PR https://github.com/ai-dynamo/dynamo/pull/11653 appears to resolve the Kimi behavior, but it requested an unqualified passing local `02-rust-cargo-check` record because the latest recorded clippy run failed on `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs:1678` with `unknown lint: clippy::manual_option_zip`.

## Non-goals

Do not reimplement or duplicate the Kimi K2.6 parser/preprocessor fix that PR #11653 already merged. Do not add a Kimi K2.6 model-name allowlist, do not change Kimi tool-call parsing, do not download or run real Kimi weights, do not consume remote GPU/Slurm/compute-session resources, and do not run Docker or Kubernetes validation. The only allowed new source correction in this re-entry is the minimal non-behavioral Rust lint-compatibility fix needed to make the planned `cargo clippy -p dynamo-llm --lib -- -D warnings` command pass locally.

## Discovery

I read the normalized caller brief at `/sandbox/job/artifacts/wi-20260723T020417Z-1/input.md`, the compute environment at `/sandbox/job/artifacts/wi-20260723T020417Z-1/compute-env.md`, the selected template at `/workflow/agents/01-agency/templates/existing-request.md`, the planner instructions at `/workflow/agents/01-agency/planner/goal.md`, and the relevant learnings under `/workflow/agents/learnings/`. The sandbox declares `Engine: vllm`, has one local NVIDIA A100-SXM4-80GB, has CUDA 12.9, and has no Docker socket. The available validation should therefore remain local; multi-GPU, Docker, Kubernetes, and remote compute recipes are out of scope.

I read all prior packets in `/sandbox/job/artifacts/wi-20260723T020417Z-1`: the existing `plan.md`, `change.md`, `change.diff`, `change-validation.md`, `review.md`, and the recorded validation registry/log listing. The prior plan correctly selected `Template: existing-request` and `Review request: https://github.com/ai-dynamo/dynamo/pull/11653`. The prior `change.md` and `change-validation.md` materialized PR #11653 and showed that the Kimi post-tool parser behavior is already present, but `change-validation.md` recorded `## Verdict: fail` because `02-rust-cargo-check` failed its clippy section.

I read the requested-changes review at `/sandbox/job/artifacts/wi-20260723T020417Z-1/review.md`. Its only blocking finding is the failed planned recipe: `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/logs/2026-07-23T02-40-09.988Z-cargo-d620.log` records `cargo clippy -p dynamo-llm --lib -- -D warnings` exiting 101 on `#[allow(clippy::manual_option_zip)]` in `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs:1678`. The reviewer explicitly allowed either fixing/guarding that incompatible lint in the checkout or validating with a clippy toolchain where the lint is known, but required the latest recorded `cargo clippy -p dynamo-llm --lib -- -D warnings` row to exit 0.

I inspected the current source under `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/.claude/worktrees/agent-ab61183d6a78a5f51` without modifying the repository. In `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/preprocessor.rs`, `OpenAIPreprocessor::postprocessor_parsing_stream` computes `should_parse_reasoning` at lines 2167-2170 from the configured parser, request-level disablement, and guided-output bypass; the old `last_is_tool` / `suppress_reasoning_after_tool` gate is absent. The same function forwards `prompt_injected_reasoning` into `parse_reasoning_content_from_stream_inner` at lines 2201-2207. `normalize_thinking_arg` at lines 539-568 defaults omitted `kimi_k25` thinking to enabled, and `is_reasoning_disabled_by_request` at lines 2998-3005 disables Kimi reasoning only for explicit false.

I inspected `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/tests/postprocessor_parsing_stream.rs`. The focused Kimi fixtures at lines 580-602 and 608-628 cover the reported post-tool shape: final input message is a tool result, the Kimi-compatible `kimi_k25` parser starts from `prompt_injected_reasoning=true`, output contains reasoning followed by a split `</think>` marker and final content, and the assertions require separated `reasoning_content` and visible `content` with no leaked close tag. I also inspected `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/PREPROCESSOR_CASES.md`, whose table line 146 documents that `kimi_k25` is ON for tool continuations when thinking is enabled or omitted and OFF only when disabled.

I surveyed related history and review requests. `git log --since=2026-06-01 -- lib/llm/src/preprocessor.rs lib/llm/tests/postprocessor_parsing_stream.rs lib/llm/PREPROCESSOR_CASES.md lib/llm/src/discovery/model_manager.rs` shows PR #11653 (`73ddd6c3e451f64840666b3914418a69284b585d`) as the Kimi fix already merged into the current checkout, PR #11728 as its release-branch cherry-pick, PR #8901 as the earlier DeepSeek V4 post-tool reasoning context cited in the issue, and PR #11460 as the encode-worker change that introduced the `manual_option_zip` allow in `model_manager.rs`. `gh pr view 11653 --repo ai-dynamo/dynamo` reports state `MERGED`, title `fix: Fix multi-turn reasoning in Kimi 2.5/2.6`, head SHA `5c568898268cc587e0f28e7e4faddf7e36026167`, merge commit `73ddd6c3e451f64840666b3914418a69284b585d`, merge parent/base commit `e8c43ff85109589bf00c5aea372df08aac32a13d`, base branch `main`, review decision `APPROVED`, merged at `2026-07-15T19:20:47Z`, and four changed files: `docs/reasoning/README.md`, `lib/llm/PREPROCESSOR_CASES.md`, `lib/llm/src/preprocessor.rs`, and `lib/llm/tests/postprocessor_parsing_stream.rs`. `gh pr view 11728 --repo ai-dynamo/dynamo` reports the same four files cherry-picked to `release/1.3.0`. `gh pr view 8901 --repo ai-dynamo/dynamo` confirms the earlier merged DeepSeek V4 parser change touched the same preprocessor/test surface.

I inspected the clippy failure site at `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs:1674-1688`. The code intentionally avoids `Option::zip` because `consumer.take()` must not run until an Encode endpoint exists. The active sandbox toolchain is `rustc 1.93.1` / `clippy 0.1.93`, while `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/rust-toolchain.toml` declares Rust `1.96.1`. That explains the failure shape: the source carries an allow for a newer Clippy lint name, but this sandbox’s pinned `RUSTUP_TOOLCHAIN=1.93.1` treats `clippy::manual_option_zip` as unknown and `-D warnings` escalates the unknown-lint warning to an error.

## Chosen approach

Continue under the `existing-request` contract for PR #11653. The Kimi parser behavior should remain the upstream PR #11653 implementation; the printer must not author a second Kimi parser/preprocessor solution and must not change the focused Kimi tests except as already present in the materialized PR #11653 diff.

For this re-entry, the printer should address only the review-requested validation blocker. Prefer a minimal non-behavioral source correction in `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs` around line 1678 so the intentional manual `Option` composition remains accepted by both the active older Clippy and the repo-declared newer Clippy. A suitable correction is either to guard the lint attribute locally with `unknown_lints`, or to rewrite the small `state.endpoint` / `state.consumer.take()` block as an explicit `match` that preserves the lazy `take()` behavior and no longer needs `#[allow(clippy::manual_option_zip)]`. The change must preserve the current behavior: do not drop the waiting consumer when no endpoint exists, and send `(sender, endpoint)` only when both are present.

The printer should record clearly in `change.md` that PR #11653 is still the request under assessment and that any additional diff outside PR #11653 is limited to the clippy-compatibility fix requested by the prior review. `change.diff` should remain nonempty and must make it easy for the reviewer to distinguish the materialized PR #11653 Kimi fix from the small validation unblocker in `model_manager.rs`.

## Rejected alternatives

One alternative is to implement a new Kimi K2.6-specific parser, model-name exception, or post-tool branch. That is rejected because PR #11653 already implements the requested behavior by relying on `prompt_injected_reasoning` and normalized thinking controls rather than duplicating parser logic or hardcoding K2.6.

Another alternative is to ignore the clippy failure because it is outside PR #11653’s Kimi diff. That is rejected because `02-rust-cargo-check` was a planned recipe and the reviewer explicitly made an unqualified passing clippy record a blocking requested change.

Another alternative is to skip local clippy and cite upstream CI. That is rejected by the `existing-request` template: upstream CI is context only, while the validator must produce local recorded recipe evidence.

Another alternative is to rely only on installing or switching to Rust/Clippy 1.96.1. That may also be valid if the validator can make the final recorded command a green `cargo clippy -p dynamo-llm --lib -- -D warnings`, but it is less deterministic in this sandbox because `RUSTUP_TOOLCHAIN=1.93.1` is currently active and only toolchain `1.93.1` is installed. A small compatibility fix at the lint site directly addresses the reviewer finding and does not touch the Kimi parser surface.

A final alternative is to run an end-to-end Kimi K2.6 vLLM smoke with real weights. That is rejected because the issue’s proof obligation is the Rust preprocessor/postprocessor behavior under a known stream shape, which the focused local fixtures exercise directly; the environment has one A100 and vLLM, but real Kimi weights and multi-turn model sampling are unnecessary for this re-entry and would consume resources without addressing the clippy blocker.

## Validation strategy

Use Recipe 05 to inspect the existing request and the new minimal unblocker. The validator should confirm PR #11653 metadata, immutable head SHA `5c568898268cc587e0f28e7e4faddf7e36026167`, merge commit `73ddd6c3e451f64840666b3914418a69284b585d`, and the four PR files listed above. It should trace the Kimi claims again through `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/preprocessor.rs`, `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/tests/postprocessor_parsing_stream.rs`, and `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/PREPROCESSOR_CASES.md`, and it should also inspect `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs` to confirm any new change is limited to clippy compatibility and preserves the lazy `consumer.take()` behavior. The parallel-pattern scan should verify that `suppress_reasoning_after_tool` remains absent and no new Kimi post-tool suppression was introduced.

Use Recipe 02 for the required Rust proof on crate `dynamo-llm`. The validator must run the recipe through the recorder and must produce a latest recorded `cargo clippy -p dynamo-llm --lib -- -D warnings` row with exit 0. Do not mark this recipe passed with a qualified or ignored clippy failure. Do not set `CARGO_TARGET_DIR`; use the sandbox’s existing job-scoped value. The expected local commands are `cargo check -p dynamo-llm --lib`, `cargo clippy -p dynamo-llm --lib -- -D warnings`, the focused Kimi behavior test `cargo test -p dynamo-llm --test postprocessor_parsing_stream postprocessor_parsing_stream_kimi -- --nocapture`, the DeepSeek post-tool control `cargo test -p dynamo-llm --test postprocessor_parsing_stream postprocessor_parsing_stream_deepseek_v4_tool_continuation_keeps_injected_reasoning -- --nocapture`, `cargo test -p dynamo-llm --lib`, and `cargo fmt --check -p dynamo-llm`. If the printer chose a source-level lint guard, the exact clippy command should pass under the active toolchain; if it chose the toolchain route instead, the validator must record the Rust/Clippy versions and still make the final clippy row unambiguously green.

Do not select Recipe 06 because Docker is unavailable and no image-build claim is being made. Do not select Recipe 07, 08, or 09 because this re-entry is about a local Rust parser validation blocker, not a GPU serving behavior claim; Recipe 08 would also be blocked by the one-GPU environment. Do not select Recipe 10 because there is no performance claim.

```validation-recipes
05-code-inspection
02-rust-cargo-check
```

## Required deliverables

This work item must produce `/sandbox/job/artifacts/wi-20260723T020417Z-1/plan.md`, `/sandbox/job/artifacts/wi-20260723T020417Z-1/change.md`, `/sandbox/job/artifacts/wi-20260723T020417Z-1/change.diff`, `/sandbox/job/artifacts/wi-20260723T020417Z-1/change-validation.md`, and `/sandbox/job/artifacts/wi-20260723T020417Z-1/review.md`. The printer should materialize PR #11653 under the existing-request contract and add only the review-requested clippy-compatibility unblocker if necessary. The validator should record both planned recipes and must end with a passing `02-rust-cargo-check` record, especially an exit-0 clippy row. The reviewer should verify that no duplicate Kimi parser implementation was authored and that the requested validation blocker is resolved.

glamr-agent:
# Change

Branch: `fix/issue-1--3936cd44d51e`
Commit: `bbdbb0e7b65505d1894d029a9b49575e48f5aa74`
Branch point: `c785ab8d6b0f54ee9d24ccb1304b8df75f28936b` (`main`)
Review request: https://github.com/ai-dynamo/dynamo/pull/11653
Template: `existing-request`

## Summary

This printer pass updates the actual provided checkout at `/home/sandbox/workspace/wi-20260723T020417Z-1/repo`, not an SDK-created worktree. PR #11653 remains the existing request under assessment; I did not duplicate or alter the Kimi parser implementation.

The only repository diff on `fix/issue-1--3936cd44d51e` is the reviewer-requested clippy unblocker in `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs`.

## Per-file changes

- `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs`: removed the incompatible `#[allow(clippy::manual_option_zip)]` and rewrote the occupied encoder-activation branch as an explicit `match` on `state.endpoint.as_ref()`. The rewrite preserves the lazy `state.consumer.take()` behavior: the waiting consumer is only taken after an Encode endpoint exists, and `(sender, endpoint)` is produced only when both are present.

## Safety constraints inspected

- Confirmed the checkout is `/home/sandbox/workspace/wi-20260723T020417Z-1/repo` and the active branch is `fix/issue-1--3936cd44d51e`.
- Confirmed the branch diff against `main` is limited to `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs`.
- Kept the Kimi preprocessor/parser/test surface untouched in this correction; PR #11653 remains the existing Kimi behavior fix under review.
- Avoided Docker, Kubernetes, remote compute, GPU downloads, and any duplicate Kimi model-name/parser implementation per the plan non-goals.

## Build and lint results

From `/home/sandbox/workspace/wi-20260723T020417Z-1/repo`:

- `cargo fmt --check -p dynamo-llm` — passed, exit 0.
- `cargo clippy -p dynamo-llm --lib -- -D warnings` — passed, exit 0.
- `cargo check -p dynamo-llm --lib` — passed, exit 0.

## Notes

A prior printer pass had already created commit `bbdbb0e7b65505d1894d029a9b49575e48f5aa74` in an SDK-created worktree. Per the correction, I detached that old worktree from the branch and checked out `fix/issue-1--3936cd44d51e` directly in `/home/sandbox/workspace/wi-20260723T020417Z-1/repo`, so the actual checkout now carries the committed diff and is clean.

glamr-agent:
# Change validation

Template: `existing-request`
Review request: https://github.com/ai-dynamo/dynamo/pull/11653
Hardware target: local validation sandbox only — vLLM engine variant, one NVIDIA A100-SXM4-80GB, CUDA 12.9, no Docker socket. The selected recipes were local code inspection and Rust cargo checks; I did not use Docker, Kubernetes, remote compute, Slurm, or GPU-serving resources.

## Evidence narrative

I recorded the planned recipes through `/workflow/agents/customs/evidence.ts` against the actual checkout at `/home/sandbox/workspace/wi-20260723T020417Z-1/repo` on branch `fix/issue-1--3936cd44d51e`. The active local toolchain in the recorded inspection was `rustc 1.93.1` / `clippy 0.1.93`.

For `05-code-inspection`, the final comprehensive recorded run is `validation/logs/2026-07-23T03-35-05.381Z-bash-6351.log`. It confirms PR #11653 is the existing merged request titled `fix: Fix multi-turn reasoning in Kimi 2.5/2.6`, head SHA `5c568898268cc587e0f28e7e4faddf7e36026167`, merge commit `73ddd6c3e451f64840666b3914418a69284b585d`, state `MERGED`, review decision `APPROVED`, and four PR files: `docs/reasoning/README.md`, `lib/llm/PREPROCESSOR_CASES.md`, `lib/llm/src/preprocessor.rs`, and `lib/llm/tests/postprocessor_parsing_stream.rs`. The same recorded run shows the local `change.diff` is separate and limited to `lib/llm/src/discovery/model_manager.rs`.

The code inspection traced the Kimi behavior through `lib/llm/src/preprocessor.rs`: `should_parse_reasoning` is controlled by parser presence, request-level disablement, and guided-output bypass, then `prompt_injected_reasoning` is passed into `parse_reasoning_content_from_stream_inner`. The Kimi-specific thinking normalization defaults omitted `kimi_k25` thinking to enabled and disables Kimi reasoning only for an explicit false. The focused tests in `lib/llm/tests/postprocessor_parsing_stream.rs` cover Kimi post-tool continuations for explicit thinking and omitted thinking, including split `</think>` output, and the DeepSeek V4 post-tool control remains present. The pattern scan found no source `suppress_reasoning_after_tool` gate; the only `last_is_tool` hit is documentation/test-table context in `lib/llm/PREPROCESSOR_CASES.md`.

The same inspection verified the minimal clippy-compatibility unblocker in `lib/llm/src/discovery/model_manager.rs`: the incompatible `clippy::manual_option_zip` allow is gone, and the occupied encoder-activation branch now explicitly matches `state.endpoint.as_ref()`. That preserves the lazy behavior the comment required: `state.consumer.take()` occurs only in the `Some(endpoint)` arm, so the waiting consumer is not dropped when the Encode endpoint is absent. The recorded PR status check reports zero `CHANGES_REQUESTED` reviews and all required PR checks passing.

For `02-rust-cargo-check`, I recorded the planned Rust commands locally. The critical reviewer requirement is satisfied by the latest recorded `cargo clippy -p dynamo-llm --lib -- -D warnings` row, `validation/logs/2026-07-23T03-34-16.336Z-cargo-da00.log`, which exits 0 under the active local Clippy and finishes without the previous unknown-lint error. The same recipe also recorded `cargo check -p dynamo-llm --lib` green in `validation/logs/2026-07-23T03-27-42.625Z-cargo-ed6a.log`, the Kimi focused stream tests green in `validation/logs/2026-07-23T03-28-17.739Z-cargo-2f13.log`, the DeepSeek V4 post-tool control green in `validation/logs/2026-07-23T03-31-07.408Z-cargo-8688.log`, `cargo test -p dynamo-llm --lib` green in `validation/logs/2026-07-23T03-31-12.351Z-cargo-4e75.log`, and `cargo fmt --check -p dynamo-llm` green in `validation/logs/2026-07-23T03-33-17.377Z-cargo-2f7b.log`.

The generated evidence report now clears both planned recipes from `validation/registry.jsonl`; no disposition was needed.

## Verdict: pass — both planned recipes have fresh green recorded evidence, and the latest recorded required clippy command exits 0

glamr-agent:
> 🤖 **Automated AI review — advisory.** An AI agent's judgment of
> whether this change is logically sound on the evidence; not a merge
> authorization. CI, customs, and a human reviewer hold that.

## Assessment: sound

### Findings

No findings.

### Evidence audit

- The committed change is on a real non-default branch in the target checkout. `/sandbox/job/artifacts/wi-20260723T020417Z-1/change.md` records branch `fix/issue-1--3936cd44d51e` and commit `bbdbb0e7b65505d1894d029a9b49575e48f5aa74`; `git status --short --branch` for `/home/sandbox/workspace/wi-20260723T020417Z-1/repo` shows `## fix/issue-1--3936cd44d51e` with no uncommitted files, and `git rev-parse HEAD` returns the recorded SHA.
- The local diff is limited to the prior requested validation unblocker. `/sandbox/job/artifacts/wi-20260723T020417Z-1/change.diff` changes only `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs`; at lines 1680-1685 the occupied activation path now matches `state.endpoint.as_ref()` and calls `state.consumer.take()` only in the `Some(endpoint)` arm, preserving the lazy behavior that avoids dropping the waiter when no Encode endpoint exists.
- The change does not author a duplicate Kimi parser implementation. The only branch diff is the model-manager clippy compatibility rewrite, while the existing-request evidence in `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/logs/2026-07-23T03-35-05.381Z-bash-6351.log` records PR #11653 as the merged, approved Kimi fix with files `docs/reasoning/README.md`, `lib/llm/PREPROCESSOR_CASES.md`, `lib/llm/src/preprocessor.rs`, and `lib/llm/tests/postprocessor_parsing_stream.rs`.
- The prior blocking clippy request is addressed. The latest recorded row for `cargo clippy -p dynamo-llm --lib -- -D warnings` is `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/logs/2026-07-23T03-34-16.336Z-cargo-da00.log`; `validation/registry.jsonl` records exit 0, and the log ends with `Finished dev profile` rather than the previous `unknown lint: clippy::manual_option_zip` failure.
- Both planned validation recipes pass in the customs evidence report: `02-rust-cargo-check` and `05-code-inspection` are `validated` in the report generated from `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/registry.jsonl`. The validator's own `/sandbox/job/artifacts/wi-20260723T020417Z-1/change-validation.md` verdict is an unqualified `pass`.
- The validated rows cite logs that show execution and passing results. Besides the latest clippy log, `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/logs/2026-07-23T03-35-05.381Z-bash-6351.log` shows the inspection command executed against branch `fix/issue-1--3936cd44d51e`, commit `bbdbb0e7b65505d1894d029a9b49575e48f5aa74`, Rust `1.93.1`, and Clippy `0.1.93`, and it confirms `manual_option_zip` is removed and no `suppress_reasoning_after_tool` source gate remains. The focused Kimi and DeepSeek cargo-test logs also show `test result: ok`.
- The validator used the local sandbox pathway described by `/sandbox/job/artifacts/wi-20260723T020417Z-1/compute-env.md`. The recorded commands in `/sandbox/job/artifacts/wi-20260723T020417Z-1/validation/registry.jsonl` are local `bash`, `gh`, and `cargo` commands only; there is no Docker, Kubernetes, SSH, Slurm, remote compute, or GPU-serving validation usage.
- No new test infrastructure was introduced by this printer pass; the change only rewrites an existing Rust branch in `/home/sandbox/workspace/wi-20260723T020417Z-1/repo/lib/llm/src/discovery/model_manager.rs`.

PR #10366, opened 2026-06-05
fix(parsers): reasoning parser to recover Harmony analysis-channel tool calls

nv-kmcgill53:
#### Overview

GPT-OSS 120B emits ~47% of real tool calls on the `analysis` Harmony channel instead of the canonical `commentary` channel. Before this PR those calls were silently dropped — the analysis-channel payload landed in `reasoning_text` and the Harmony jail never saw a structured envelope to parse.

This PR recovers them. The recipient (`to=functions.*`) is the definitive signal of a tool call regardless of channel label.

**Deliberate reversal of #10111's `REASONING.batch.3.e` contract.**

#### Concrete before / after

Model output that triggers the bug (analysis channel, bare `code` constraint):

```
<|channel|>analysis to=functions.grep code<|message|>{"pattern":"foo"}<|call|>
```

| | before | after |
|---|---|---|
| `reasoning_text` | `{"pattern":"foo"}` | `` |
| `normal_text` | `` | `<\|channel\|>analysis to=functions.grep code<\|message\|>{"pattern":"foo"}<\|call\|>` |

Before: args leaked into `reasoning_text`, jail saw empty `normal_text`, `tool_calls = []`.
After: complete envelope in `normal_text`, jail parses it, `tool_calls = [{name: "grep", arguments: {...}}]`.

Mixed case — analysis call + commentary call + final answer:

```
<|channel|>analysis to=functions.search <|constrain|>json<|message|>{"q":"x"}<|call|>
<|start|>assistant<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"city":"SF"}<|call|>
<|start|>assistant<|channel|>final<|message|>It is sunny.
```

| | before | after |
|---|---|---|
| `reasoning_text` | `{"q":"x"}` | `` |
| `normal_text` | commentary only (search call dropped) | both envelopes present |

#### How the reasoning→TC handoff works

The GPT-OSS reasoning parser splits raw model output into `reasoning_text` (chain-of-thought) and `normal_text`. For directed tool calls, the parser places the complete Harmony envelope into `normal_text` as an internal handoff to the tool-calling jail. The jail calls `parse_tool_calls_harmony_complete` on that content, emits structured `tool_calls`, and strips the markup — it never surfaces in the final assistant `content`.

[...]

PR #9778, opened 2026-05-20
fix(llm): support GPT-OSS tool calling from Codex on Dynamo

Chokoyo:
#### Overview

Two fixes found while running GPT-OSS through Dynamo with Codex using the OpenAI Responses API path.

Codex/Claude Code sends tool schemas that expose two Dynamo-side gaps for GPT-OSS/Harmony tool calling:

1. The Harmony terminal tool-call marker \`<|call|>\` was being treated as a hidden EOS token and removed before the Harmony parser could see it.
2. GPT-OSS chat templates use Python/Jinja-style \`.items()\` iteration, which can conflict with JSON Schema properties also named \`items\`.

#### Concrete before / after: the \`<|call|>\` EOS problem

GPT-OSS model output for a tool call:

```
<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"city":"SF"}<|call|>
```

\`<|call|>\` is simultaneously the Harmony tool-call terminator AND one of the model's EOS token IDs. Dynamo adds all EOS IDs to \`stop_token_ids_hidden\`, which causes the decoder to stop on \`<|call|>\` AND strip it from output.

**Before** — Harmony parser receives a truncated envelope:

```
<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"city":"SF"}
```

No \`<|call|>\` terminator → parser drops the call → \`tool_calls = []\`, tool call silently lost.

**After** — \`<|call|>\` is moved from \`stop_token_ids_hidden\` to \`stop_token_ids_visible\`. Decoder still stops on it, but now includes it in output. Harmony parser receives the complete envelope:

```
<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>{"city":"SF"}<|call|>
```

Parser sees the terminator → \`tool_calls = [{name: "get_weather", arguments: {"city": "SF"}}]\`.

##### Fix 1: Harmony stop-token visibility

\`remove_tool_parser_end_tokens_from_hidden_stops\` finds each parser's \`tool_call_end_tokens\`, encodes them with the model tokenizer, and moves any that land in the hidden-EOS set to \`stop_token_ids_visible\` instead. Gated on \`has_tools && tool_choice != "none"\` — no change for non-tool requests.

\`backend.rs\` gets a new \`VisibleStopTokenDetected\` stop trigger that stops generation and includes the token in output (symmetric with \`HiddenStopTokenDetected\` which strips it).

[...]

PR #11671, opened 2026-07-14
fix(trtllm): handle visible stop tokens in unified backend

Chokoyo:
## Overview:

Apply TRT-LLM visible and hidden stop-token handling consistently in both the legacy and unified backend paths.

TRT-LLM's `ignore_eos` behavior disables both `end_id` and stop words. Without this normalization, the unified backend can consume an overlapping EOS/visible-stop token before Dynamo sees it. GPT-OSS Harmony's `<|call|>` token is the known example, but the issue applies to any model with overlapping EOS and visible-stop tokens.

## Details:

- Extract the existing TRT-LLM stop-condition normalization into a shared `apply_stop_conditions_to_sampling_params` helper.
- Use the helper from both the legacy request handler and `TrtllmLLMEngine`.
- When visible stop tokens are present:
  - set `ignore_eos=True`, which TRT-LLM maps to `end_id=-1` and an empty stop-word list;
  - remove visible tokens from engine-side stop IDs;
  - let Dynamo's decoder remain the authoritative stopping layer.
- Preserve existing hidden stop-token merging and `min_tokens` handling.
- Add unit coverage for visible and hidden stop-token normalization.
- Add a regression test covering the unified backend request path.

This follows up on the original TRT-LLM workaround in #9778. The related Harmony parser recovery was added in #10366.

- `components/src/dynamo/trtllm/utils/request_utils.py`
- `components/src/dynamo/trtllm/llm_engine.py`
- `components/src/dynamo/trtllm/tests/test_trtllm_trace_propagation.py`

## Related Issues

- DIS-2219

## Summary by CodeRabbit

* **New Features**
  * Improved handling of stop conditions during text generation.
  * Visible stop tokens are now managed by the application, while hidden stop tokens continue to support engine-level stopping.
  * Minimum token requirements are applied consistently when configured.

* **Bug Fixes**
  * Corrected stop-token and end-of-sequence behavior across generation pathways.
  * Added coverage for visible and hidden stop-token scenarios.

### DIFF
--- a/components/src/dynamo/trtllm/llm_engine.py
+++ b/components/src/dynamo/trtllm/llm_engine.py
@@ -75,6 +75,7 @@ from dynamo.trtllm.utils.disagg_utils import (
     DisaggregatedParamsCodec,
 )
 from dynamo.trtllm.utils.request_utils import (
+    apply_stop_conditions_to_sampling_params,
     request_cache_salt,
     stored_event_cache_salt,
 )
@@ -855,11 +856,7 @@ class TrtllmLLMEngine(LLMEngine):
             elif self.max_seq_len is not None:
                 sampling_params.max_tokens = max(1, self.max_seq_len - len(token_ids))
 
-        # TODO: mirror visible/hidden stop-token handling from the disagg
-        # path (handler_base.py) into a shared helper. See PR #9778.
-        ignore_eos = stop_conditions.get("ignore_eos")
-        if ignore_eos:
-            sampling_params.ignore_eos = ignore_eos
+        apply_stop_conditions_to_sampling_params(sampling_params, stop_conditions)
 
         # In conversation-affinity mode let the engine's ConversationAwareADPRouter pick the
         # attention-DP rank from the conversation id; do NOT force a rank (an explicit rank is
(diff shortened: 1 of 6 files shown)


connorcarpenter15:
Overall LGTM, just one agent comment: 

1. Generation-only policy leaks into PREFILL.  
   [`apply_stop_conditions_to_sampling_params`](https://github.com/ai-dynamo/dynamo/blob/9a7b8f73b56e9eec08f7ec389068914388c98da3/components/src/dynamo/trtllm/llm_engine.py#L848-L859) runs after PREFILL forces `max_tokens = 1`. The helper also applies caller `min_tokens`, hidden stops, and `ignore_eos`—settings relevant only to workers producing output. Dynamo already has the canonical [`is_generation_stage`](https://github.com/ai-dynamo/dynamo/blob/9a7b8f73b56e9eec08f7ec389068914388c98da3/components/src/dynamo/common/backend/engine.py#L501-L512) boundary.

   Gate the helper to AGGREGATED/DECODE, mirror that in the legacy path, and add a PREFILL regression test. Don’t add mode awareness inside the helper; the call sites already own that decision.

connorcarpenter15 (approved):

tanmayv25 (approved):

PR #11672, opened 2026-07-14
refactor(operator): centralize feature gates

sttts:
## Summary

- introduce a single operator-wide `features.Gate` interface and documented gate definitions
- centralize configuration, environment, and Kubernetes API discovery into one resolved gate snapshot
- use the same gate source in controllers, generators, validators, defaulters, and mutators instead of copied booleans

This is the feature-gate foundation for the namespace-scoped admission follow-up in #11475.

Related: DYN-3412

## Validation

- `make test`
- `make lint`
- `go vet ./...`
- `git diff --check origin/main...HEAD`

## Review note

The review findings are addressed: Istio desired state remains distinct from API availability, and the complete gate registry is covered by a consistency test.

## Summary by CodeRabbit

* **New Features**
  * Added centralized feature-gate management for optional capabilities, including checkpoints, GPU discovery, schedulers, service mesh, and resource allocation.
  * Feature settings now consistently control controller, webhook, admission, and workload behavior.
  * Added compatibility gating to preserve supported vLLM multiprocessing behavior across upgrades.

* **Bug Fixes**
  * Improved handling when optional cluster APIs are unavailable.
  * Replaced environment-variable-based behavior for GMS and snapshot validation with managed feature settings, providing more predictable admission and reconciliation results.


deploy/operator/internal/features/gates.go
```diff
@@ -0,0 +1,386 @@
...
+		return Gates{}, err
+	}
+
+	log.FromContext(ctx).Info("Resolved operator feature gates",
+		"grove", gates.Grove,
+		"lws", gates.LWS,
+		"volcano-scheduler", gates.VolcanoScheduler,
+		"kai-scheduler", gates.KaiScheduler,
+		"dra", gates.DRA,
+		"istio", gates.Istio,
```
sttts: don't hardcode, make this via loop.

deploy/operator/internal/features/gates.go
```diff
@@ -0,0 +1,386 @@
...
+		return g.VolcanoScheduler
+	case DRA:
+		return g.DRA
+	case Istio:
+		return g.Istio
+	case GPUDiscovery:
+		return g.GPUDiscovery
+	default:
+		panic(fmt.Sprintf("unknown feature gate %q", name))
+	}
```
sttts, line 351: use reflection.

deploy/operator/internal/features/gates.go
sttts: Done.

deploy/operator/internal/features/gates.go
sttts: Done.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -2642,6 +2644,10 @@ func generateAdapterName(dgdName, componentName string) string {
 	return fmt.Sprintf("%s-%s", dgdName, strings.ToLower(componentName))
 }
 
+func (r *DynamoGraphDeploymentReconciler) istioState() (enabled, available bool) {
+	return r.Config.ServiceMesh.IsEnabled(), r.RuntimeConfig.Enabled(features.Istio)
+}
```
sttts: This has smell. Why two sources. Was ist `r.Config.ServiceMesh.IsEnabled()`. Why another source. Shouldn we only look at the gate? Is some logic missing in the gate?

deploy/operator/internal/webhook/common.go
```diff
@@ -70,6 +71,36 @@ type LeaseAwareDefaulter struct {
...
+
+type gateDefaulter struct {
+	defaulter admission.Defaulter[runtime.Object]
+	gate      features.Gate
+}
+
+type gateHandler struct {
+	handler admission.Handler
+	gate    features.Gate
+}
```
sttts: these have smell as they break interface checks.

Wondering, can we just make feature.FromContext fail hard, and put the gate into the context? Less pumbing, same safety.

sttts: Done. The gate is injected with Webhook.WithContextFunc, the interface wrappers are gone, and a missing gate panics.

deploy/operator/internal/controller/dynamographdeployment_controller.go
sttts: Done. Istio enablement now only comes from the gate; API detection remains only for cleanup.

julienmancuso:
General question : How should this package relate to the existing internal/featuregate package? They serve different scopes : operator-wide capabilities versus per-resource compatibility, but the nearly identical names make it look like competing feature-gate systems. Should we consolidate them under one package, or rename/document the existing package as origin-version compatibility gates?

sttts:
> General question : How should this package relate to the existing internal/featuregate package? They serve different scopes : operator-wide capabilities versus per-resource compatibility, but the nearly identical names make it look like competing feature-gate systems. Should we consolidate them under one package, or rename/document the existing package as origin-version compatibility gates?

very good question. Looking into.

julienmancuso:
P1 found by codex, this is a pre-existing issue, so it's fine if we postpone the fix to a follow up MR :

[P1] The new Checkpoint gate does not actually disable checkpoint creation.
The gate is documented as enabling “checkpoint creation and restore” and is populated from checkpoint.enabled in [features/gates.go](https://github.com/ai-dynamo/dynamo/blob/79b24095a9f94cd588929666cac46554490d8453/deploy/operator/internal/features/gates.go#L42-L52), but several primary paths never check it:
[reconcileCheckpoints](https://github.com/ai-dynamo/dynamo/blob/79b24095a9f94cd588929666cac46554490d8453/deploy/operator/internal/controller/dynamographdeployment_controller.go#L1954-L1975) still ensures storage and creates/resolves DynamoCheckpoint resources regardless of features.Checkpoint.
The [DynamoCheckpoint admission handler](https://github.com/ai-dynamo/dynamo/blob/79b24095a9f94cd588929666cac46554490d8453/deploy/operator/internal/webhook/validation/dynamocheckpoint_handler.go#L47-L54) checks only the GMS/Snapshot sub-gate.
The [CheckpointReconciler](https://github.com/ai-dynamo/dynamo/blob/79b24095a9f94cd588929666cac46554490d8453/deploy/operator/internal/controller/dynamocheckpoint_controller.go#L220-L226) likewise checks only GMSSnapshot, then proceeds to create the checkpoint Job.
Consequently, with the default checkpoint.enabled=false, a DGD checkpoint specification or manually created DynamoCheckpoint can still create PVCs, CRs, and Jobs. Only selected workload-generation and pod-restore paths honor the new gate.
This behavior is partly pre-existing, but this PR formalizes Checkpoint as an operator-wide gate while preserving the bypass. I would block on either enforcing it consistently at admission/controller boundaries or narrowing/removing the gate’s stated contract.

julienmancuso (approved):

deploy/operator/internal/controller_common/runtime.go
```diff
@@ -17,22 +17,13 @@
...
-	KaiSchedulerEnabled bool
-	// VolcanoSchedulerEnabled indicates whether Dynamo should inject Volcano scheduler settings into Grove PodCliqueSets
-	VolcanoSchedulerEnabled bool
-	// DRAEnabled indicates whether Dynamic Resource Allocation (resource.k8s.io/v1) is available
-	DRAEnabled bool
-	// IstioEnabled is the resolved Istio DestinationRule reconciliation state
-	// after merging config overrides with API auto-detection.
-	IstioEnabled bool
+	// Gate contains the resolved operator features.
+	Gate features.Gates
```
hhzhang16, line 26: Can `Gate` be a pointer `*features.Gates`? Right now the zero value of `features.Gates` looks like a valid "all disabled" configuration. Using a pointer would make it obvious if `Gate` wasn't initialized, causing an immediate failure instead of silently disabling all features.

deploy/operator/internal/features/gates.go
```diff
@@ -0,0 +1,379 @@
...
+				return true
+			}
+		}
+		return false
+	}
+	return false
+}
+
+// Enabled reports whether name is enabled.
+func (g Gates) Enabled(name Name) bool {
```
hhzhang16, line 344: Any reason to use reflection here instead of an explicit switch? The switch is a bit easier to follow and avoids coupling runtime behavior to the JSON tags.

sttts: Just that I want to make it easy for ppl to add flags, and not understand the mechanics.

deploy/operator/internal/controller_common/runtime.go
sttts: Actually it should be the interface type here.

Btw, I like that thinking about immediate failure, fail hard and early 👍

hhzhang16 (approved):

PR #11678, opened 2026-07-14
fix(kv-router): detach dead radix nodes from their parent on removal

Shang-Pin:
## Overview:

`RadixTree` never frees the memory of removed blocks: `apply_removed` (and worker
removal / `Cleared`) clears worker state and lookup entries, but a node whose
workers are all gone is never detached from its parent's `children` map — nodes
have no parent link, and `clear_children_if_unreachable` only clears the node's
*own* children. Every unique stored chain therefore permanently retains its
`edge` Vec and `edge_index` hashbrown map as an unreachable-but-alive "zombie"
under the immortal root.

On a production disaggregated frontend serving unique-heavy traffic (~30 rps),
this leaked ~1 GB/h of router memory indefinitely: jeprof showed 9.6 GB of live
allocations after 11 h, 98% under `RadixTree::apply_event_with_counters` →
`RadixBlock::for_blocks`, while the tracked-blocks gauge stayed healthily
bounded (350–870k, oscillating with load) — the tree forgets blocks but never
frees their nodes. The arithmetic closes: ~1M requests × ~140 blocks/chain ×
~65 B/block ≈ the measured heap. An A/B with only this fix reverted reproduced
the ~1 GB/h growth; with the fix, the same frontend held a flat ~1 GB working
set over a 3-day soak.

The leak affects every removal path, including the TTL prune (it funnels
through the same `apply_removed`); kv-events mode is hit hardest because
nothing else bounds the tree. Hot shared prefixes are unaffected (dead nodes
under them get revived by the next store) — it is the unique-suffix tail that
accumulates, so severity scales with prompt uniqueness.

## Details:

- Add a `Weak` parent back-link to `RadixBlock` (no `Rc` cycles), wired at all
  node creation/move sites: `apply_stored`'s two child-creation sites, and
  `split_node` (the new suffix node, plus re-pointing the children it inherits).
- New `detach_if_dead`: once a node holds no workers and has no children,
  remove it from its parent's `children` map, cascading upward while ancestors
  become empty; stops at the root (dangling parent). Guarded by `Rc::ptr_eq`
  against slot reuse.
- Called from the three removal paths: `apply_removed`,
  `remove_or_clear_worker_blocks` (also covers `Cleared`), and
  `remove_worker_dp_rank`.
- Behavior-preserving otherwise: nodes still alive for another worker (or with
  live children) are untouched; a re-stored chain after detach recreates
  cleanly.
- 5 regression tests: full-chain detach, cascade through split ancestors,
  multi-worker sparing, worker-removal detach, store-after-detach recreation.

`lib/kv-router/src/indexer/radix_tree.rs` — `detach_if_dead` and its three call
sites; then the parent wiring in `split_node` (the inherited-children
re-pointing is the one non-obvious spot).

## Related Issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed inactive routing branches after workers are removed, preventing stale cache data from being retained.
  * Improved cleanup of nested and split routing paths.
  * Preserved routing nodes that are still in use by other workers.
  * Ensured removed paths can be recreated correctly when needed.

### DIFF
--- a/lib/kv-router/src/indexer/radix_tree.rs
+++ b/lib/kv-router/src/indexer/radix_tree.rs
@@ -3,7 +3,11 @@
 
 //! Single-threaded compressed radix tree for KV cache routing.
 
-use std::{cell::RefCell, collections::VecDeque, rc::Rc};
+use std::{
+    cell::RefCell,
+    collections::VecDeque,
+    rc::{Rc, Weak},
+};
 
 use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 
@@ -20,6 +24,12 @@ pub(crate) struct RadixBlock {
     children: FxHashMap<LocalBlockHash, SharedRadixBlock>,
     /// Once a node has children it is never eligible for leaf extension again.
     internal: bool,
+    /// Back-link to the parent node (dangling for the root). Needed so a node
+    /// whose blocks have all been removed can be detached from the parent's
+    /// `children` map; without it dead nodes stay allocated forever, retaining
+    /// their `edge` Vec and `edge_index` map (unbounded memory growth on
+    /// unique-heavy traffic even though the lookup tables stay bounded).
+    parent: Weak<RefCell<RadixBlock>>,
 }
 
 impl RadixBlock {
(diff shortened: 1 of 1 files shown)


dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `8537539` | unsigned |

PR #11682, opened 2026-07-14
feat(router): configure session headers and simulate hybrid routing

biswapanda:
## Summary

- Make the frontend's router-local session-affinity header configurable with `--router-session-header-key` or `DYN_ROUTER_SESSION_HEADER_KEY`, while retaining `X-Dynamo-Session-ID` as the default.
- Validate custom HTTP header names and propagate the selected header through the Python binding and the OpenAI, Anthropic, and Generate HTTP paths. Custom affinity headers affect routing only; canonical agent-session metadata remains unchanged.
- Move the shared Rendezvous/HRW scoring primitive into `dynamo-kv-router` with domain separation and stable worker/DP-rank targets, while preserving LoRA's legacy score encoding.
- Extend offline replay with deterministic local, modulo, HRW, KV-overlap-bounded, and local-LRU + HRW + KV-cost-bounded policies. The replay compares round-robin, KV-only, session-only, and hybrid session + KV-aware routing under complete, partial, and missing KV-event visibility.

## Motivation

Deployments often already carry a conversation identifier such as `X-Correlation-ID`. Allowing the frontend to consume that header avoids rewriting requests at ingress and lets related turns remain on the same worker and DP rank. The replay additions quantify the cache-reuse, load-distribution, and topology-change tradeoffs before selecting a production hybrid policy.

## End-to-end performance validation

The A/B ran entirely in the `biswa-session-aware-routing-dev-gpu-pod` dev GPU pod in cluster `nv-prd-dgxc.teleport.sh-dynamo-gcp-dev-02`, namespace `bis-rl-3`, at commit `c1218abd0dac6f50886ab8482368bd8074605ce7`.

Setup:

- 4 NVIDIA GB200 GPUs; four standalone aggregated vLLM workers, one GPU per worker
- `Qwen/Qwen3-30B-A3B-Thinking-2507`, prefix caching enabled
- Weka-derived trace: 7 conversations, 32 turns, 1,356,436 prompt tokens, 5,909 completion tokens
- AIPerf concurrency 7, sequential conversation replay, seed 42
- Both arms used `--router-mode kv --router-session-header-key X-Correlation-ID`
- KV-aware baseline: session-affinity TTL disabled
- Session + KV-aware arm: `--router-session-affinity-ttl-secs 1200`
- Each arm started fresh frontend and worker processes

| Metric | KV-aware | Session + KV-aware | Delta |
| --- | ---: | ---: | ---: |
| Cached-token ratio | 72.22% | 73.60% | +1.38 pp |
| Worker stickiness | 96.0% (1 switch) | 100.0% (0 switches) | +4.0 pp |
| Run duration | 135.11 s | 123.43 s | -8.65% |
| Request throughput | 0.2369 req/s | 0.2593 req/s | +9.46% |
| Output throughput | 43.74 tok/s | 47.88 tok/s | +9.46% |
| Prompt throughput | 10,039.80 tok/s | 10,989.95 tok/s | +9.46% |
| p50 latency | 10.513 s | 10.257 s | -2.44% |
| p90 latency | 25.441 s | 25.791 s | +1.38% |
| p99 latency | 41.095 s | 41.099 s | effectively flat |

This is a single, small-trace directional A/B rather than a statistically powered benchmark. It shows the expected routing behavior—zero worker transitions with session affinity—and a modest cache-reuse increase; the throughput result should be confirmed on a larger Weka trace and across repeated runs.

## Validation

- `cargo test -p dynamo-kv-router rendezvous --lib` — 3 passed
- `cargo test -p dynamo-mocker session_affinity --lib` — 22 passed, 2 report-only tests ignored
- `cargo test -p dynamo-llm session_affinity_header_name_defaults_and_validates --lib` — 1 passed
- `git diff --check`
- End-to-end 4-GPU A/B described above; all 32 requests completed without errors


jthomson04 (changes requested):
blocking this for now based on some benchmarking methodology concerns communicated over slack

PR #11688, opened 2026-07-14
test(trtllm): cap mm_overlap KV pool with token lever, not byte budget (OPS-7716)

dmitry-tokarev-nv:
## Summary

`tests/mm_router/test_mm_router_e2e.py::test_trtllm_mm_overlap_all` flakes on the pre_merge trtllm GPU job: the TRT-LLM worker exits code 1 during startup and the health check never goes ready.

**Root cause (contention + a greedy default, not a per-test KV budget):**
- The test was the only TRT-LLM KV test using the byte-based `requested_trtllm_vram_gib` → `max_gpu_total_bytes`. That override is `deep_update`-merged over TRT-LLM's default `free_gpu_memory_fraction=0.9`, so the worker greedily grows its footprint toward the whole 22 GiB card.
- Co-scheduled `profiled=0` "filler" unit tests bypass the parallel scheduler's VRAM gates and add ~5 GiB of CUDA context. The sum crosses 22 GiB during the worker's startup allocation → OOM-shaped exit.
- It is contention-driven: the **identical** config passes at 8 concurrent slots and fails at 16 (slot count is runner-CPU dependent). Verified across four CI runs (2 pass @8 slots, ≥3 fail @16 slots).

**Fix:** switch to `requested_trtllm_kv_tokens(32768)` — the token-based lever every other passing TRT-LLM KV serve/router test already uses. It sets `KvCacheConfig.max_tokens`, which TRT-LLM sizes deterministically and which takes precedence over the greedy fraction (also sidestepping a `deep_update` clobber that drops `free_gpu_memory_fraction`). The worker footprint becomes a fixed `KV(32768 tokens) + weights` regardless of free memory at launch. 32768 tokens = 1024 blocks ≈ 2× the worst-case single request (~520 blocks), so overlap-reuse assertions keep headroom. `profiled_vram_gib` set to 14.0 (conservative) pending a measured standalone peak.

## Validation

- Analyzed the GPU memory timeline + per-worker scheduling across 4 CI runs; confirmed the 8-vs-16-slot A/B and that the worker (not a co-scheduled model) is the allocator during its own startup spike.
- Standalone peak-VRAM probe pending on a GPU box to (a) confirm the token cap drops the peak vs the default/byte-cap paths and (b) finalize `profiled_vram_gib`.

Supersedes the earlier byte-budget tweak on this branch (which the analysis showed would not bind under contention).

Linear: OPS-7716

### DIFF
--- a/tests/mm_router/test_mm_router_e2e.py
+++ b/tests/mm_router/test_mm_router_e2e.py
@@ -281,8 +281,8 @@ def _send_request_get_overlap(
 
 
 @pytest.mark.pre_merge
-@pytest.mark.profiled_vram_gib(20.0)
-@pytest.mark.requested_trtllm_vram_gib(20.0)
+@pytest.mark.profiled_vram_gib(18.0)
+@pytest.mark.requested_trtllm_vram_gib(16.0)
 @pytest.mark.timeout(1800)
 def test_trtllm_mm_overlap_all(start_trtllm_mm_services, predownload_models):
     """Run all TRT-LLM MM overlap scenarios under one worker startup."""
--- a/tests/mm_router/test_mm_router_e2e.py
+++ b/tests/mm_router/test_mm_router_e2e.py
@@ -121,7 +121,10 @@ class TRTLLMWorkerProcess(ManagedProcess):
             health_check_urls=[
                 (f"http://localhost:{system_port}/health", _check_ready)
             ],
-            timeout=900,
+            # Worker load + engine warmup is ~150-180s with the token-capped KV
+            # pool; 360 keeps ~2x margin for slow runners without bloating the
+            # test-level timeout budget.
+            timeout=360,
             straggler_commands=["-m dynamo.trtllm"],
             log_dir=_prepare_log_dir(request, "trtllm-worker"),
             **_COMMON_PROCESS_KWARGS,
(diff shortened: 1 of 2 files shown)


dmitry-tokarev-nv:
## Follow-ups surfaced by this investigation

This PR fixes the one test (deterministic token cap), but the investigation surfaced three separate issues worth tracking. Filed for visibility:

| Issue | Priority | What |
|-------|----------|------|
| [OPS-7718](https://linear.app/nvidia/issue/OPS-7718) | Medium | **Systemic root cause.** GPU-parallel scheduler treats `profiled=0` tests as free fillers (`pytest_parallel_gpu.py:394`), but many touch CUDA (~0.3–0.5 GiB each). At 16 slots that's ~5 GiB of untracked memory that tips a co-scheduled worker's startup peak over the card. Same config passes @8 slots, fails @16. This PR patches the symptom for one test; OPS-7718 is the class. |
| [OPS-7719](https://linear.app/nvidia/issue/OPS-7719) | Medium | **Diagnosability.** On worker crash, `display_output=False` + block-buffered file stdout loses the tail, and `_log_tail_on_error` doesn't reach CI capture — so the actual OOM/error string was never captured and root cause had to be inferred from the GPU memory timeline. Fix: run workers unbuffered + surface the tail to stdout. |
| [OPS-7717](https://linear.app/nvidia/issue/OPS-7717) | Low | **Latent bug.** `deep_update` (`trtllm_utils.py:18`) clobbers the `KvCacheConfig` object with a bare override dict, dropping `free_gpu_memory_fraction` (and any other fields). Harmless today since defaults align, but any single-field `kv_cache_config` override silently discards the rest. |

All three link back to OPS-7716.

tests/mm_router/test_mm_router_e2e.py
```diff
@@ -281,9 +284,22 @@ def _send_request_get_overlap(
...
+# lets the worker's startup footprint balloon toward the full card. On a 22 GiB
+# GPU shared with concurrently-scheduled fillers that pushed a ~20 GiB peak, that
+# tipped over into an OOM-shaped worker exit during startup (OPS-7716). 32768
+# tokens (1024 blocks) is ~2x the worst-case single request (~520 blocks /
+# 16640 tokens per THREE_IMAGE_TOTAL_BLOCKS_RANGE), leaving ample reuse headroom.
+@pytest.mark.profiled_vram_gib(12.0)  # measured device peak ~11 GiB w/ token cap
+@pytest.mark.requested_trtllm_kv_tokens(32768)
+# ~3x the measured deterministic runtime (~311s w/ the token cap, at 8 and 16
+# slots). Clears the worst-case fixture startup budget (worker 360 + router 240
+# + frontend 240) plus the ~150s of scenarios.
```
nv-anants, line 301: this adds up to 990 - is 960 timeout sufficient? 😅

nv-anants (approved):

PR #11689, opened 2026-07-14
fix(operator): stop bundling CRDs in the Helm subchart

### LINKED ISSUE #11618: [BUG]: Standalone operator subchart exceeds Kubernetes Secret size limit
### Describe the Bug

Direct installation of the internal `deploy/helm/charts/platform/components/operator` Helm subchart exceeds Kubernetes' 1 MiB Secret size limit when Helm uses its default Secret storage driver.

This does **not** affect the published and documented `dynamo-platform` chart. The Platform chart from `v1.3.0-rc1` installs successfully, including with its default dependencies; its stored Helm release was about 34 KiB in the reproduction environment. The operator subchart is currently packaged by the Platform chart Makefile for local Helm testing, not published as the supported installation artifact.

The operator subchart contains 4.2-4.3 MiB of CRD YAML. Although its packaged archive is about 511-531 KiB, Helm stores the complete standalone release object as gzipped, base64-encoded data in a Kubernetes Secret. The resulting Secret exceeds 1,048,576 bytes.

`--skip-crds` does not avoid the standalone subchart failure. It prevents Helm from applying files from `crds/`, but those files remain part of the stored standalone release object.

### Steps to Reproduce

1. Create a Kubernetes cluster. This was reproduced with Kind and Kubernetes 1.30.13.
2. Package the internal operator subchart:

   ```bash
   helm package deploy/helm/charts/platform/components/operator
   ```

3. Attempt a direct installation, including `--skip-crds`:

   ```bash
   helm install dynamo-operator ./dynamo-operator-1.3.0.tgz \
     --namespace dynamo-system \
     --create-namespace \
     --skip-crds
   ```

### Expected Behavior

Either the internal operator subchart can be installed directly for development and testing, or the repository clearly prevents or documents direct installation as unsupported.

### Actual Behavior

The direct installation fails while creating Helm's release Secret:

```text
Error: INSTALLATION FAILED: create: failed to create: Secret "sh.helm.release.v1.dynamo-operator.v1" is invalid: data: Too long: must have at most 1048576 bytes
```

### Environment

- Helm 3.17.1 with the default Secret storage driver
- Kind with Kubernetes 1.30.13
- Reproduced from:
  - `v1.2.1`: packaged operator subchart 511 KiB, CRD YAML 4.2 MiB
  - `v1.3.0-rc1`: packaged operator subchart 528 KiB, CRD YAML 4.3 MiB
  - `release/1.3.0`: packaged operator subchart 529 KiB, CRD YAML 4.3 MiB
  - `main`: packaged operator subchart 531 KiB, CRD YAML 4.3 MiB

### Additional Context

This predates and is not caused by #11597. It was discovered while performing Helm installation smoke tests for that PR. The first test installed the internal operator subchart directly; a follow-up test confirmed that the actual `dynamo-platform` release artifact is not affected.

Removing `crds/` from a temporary operator subchart reduced it to 19 KiB and allowed direct installation. The cluster-wide operator's existing `crd-apply` init container then installed all nine CRDs successfully.

Potential resolutions:

1. Explicitly treat the operator chart as an internal-only subchart and prevent or document direct installation.
2. Stop embedding CRDs in the operator subchart and make `crd-apply` the single automatic CRD installation and upgrade path.
3. Reduce generated CRD schema size, with explicit size headroom.

julienmancuso:
## Summary

Fixes #11618.

Remove generated CRDs from the internal operator Helm subchart. The CRDs remain generated under `deploy/operator/config/crd/bases`, are bundled in the operator image, and are installed or upgraded by the `crd-apply` init container.

This change:

- Reduces the packaged operator chart from approximately 530 KiB to 19 KiB.
- Prevents standalone releases from exceeding Kubernetes' 1 MiB Helm Secret limit.
- Stops `make manifests` from copying CRDs into the operator chart.
- Updates Tilt to apply CRDs from their canonical generated location.
- Removes obsolete `--skip-crds` guidance.
- Adds a chart check preventing CRDs from being reintroduced.
- Updates Helm and Kubernetes documentation to describe the image-based CRD lifecycle.

Existing CRDs are not deleted during upgrades. Namespace-restricted operators continue to rely on CRDs managed by the cluster-wide operator.

## Validation

- Ran `helm lint` against the operator chart.
- Ran `helm lint` against the packaged platform chart.
- Verified the operator package contains no `crds/` entries.
- Verified `helm show crds` returns no operator CRDs.
- Verified the platform package contains no operator CRDs.
- Verified cluster-wide rendering includes the `crd-apply` init container.
- Verified namespace-restricted rendering with `upgradeCRD=false` omits `crd-apply`.
- Ran the new `verify-operator-chart-crds` Make target.
- Ran `git diff --check`.

Not run:

- `helm unittest` because the plugin was unavailable locally.
- Fern documentation checks because the Fern CLI was unavailable locally.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11689" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * CRDs are now applied by the operator before startup, improving consistency across installations.
  * Namespace-restricted deployments can rely on CRDs managed by a cluster-wide operator.

* **Documentation**
  * Updated Helm, Kubernetes, development, and webhook guidance for the new CRD workflow.
  * Clarified that namespace-restricted installations require `dynamo-operator.upgradeCRD=false`.
  * Removed outdated `--skip-crds` instructions.


deploy/helm/charts/platform/Makefile
```diff
@@ -69,10 +69,17 @@ helm-unittest-install: ## Install helm-unittest plugin if not present
 charts-prep: ## Package operator subchart for local testing
 	@helm package components/operator -d charts/ 2>/dev/null
 
+.PHONY: verify-operator-chart-crds
+verify-operator-chart-crds: ## Verify operator CRDs are managed from the image, not the Helm chart
+	@if helm show crds components/operator | grep -q .; then \
+		echo "ERROR: operator CRDs must be managed by crd-apply from the operator image, not bundled in the Helm chart"; \
+		exit 1; \
+	fi
```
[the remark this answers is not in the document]
julienmancuso: fixed

dmitry-tokarev-nv (commented):
Automated review of the CRD-unbundling change. Direction is correct, and the new `verify-operator-chart-crds` guard is properly wired into CI — `pr.yaml:203` runs `make -C deploy/helm/charts/platform lint test`, and the guard is a prerequisite of both `lint` and `test`, so a reintroduced `crds/` directory fails pre-merge. `helm unittest` also runs there, covering the check the author couldn't run locally.

Three inline notes below. The substantive one is the `upgradeCRD=false` cluster-wide behavior change (documented as intended — just confirm no internal recipe/automation relied on the old chart-native initial CRD install).

Out-of-diff nit (can't anchor inline): `deploy/CONTRIBUTING.md:51` still tells contributors to bump the CRD chart version when editing `deploy/helm/charts/crds/templates/*.yaml` — that path doesn't exist on this branch, and the premise (chart-managed CRDs) is exactly what this PR removes. Good moment to fix or drop it.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -395,13 +395,16 @@ runs:
       shell: bash
       run: |
         echo "::group::Wait for Dynamo operator CRDs inside vCluster"
-        # CRDs live in the chart's crds/ directory so Helm installs them before any pod
-        # starts. Waiting for Established is independent of vCluster syncer pod-Ready
-        # latency, which is unreliable on some clusters (e.g. AWS with LimitRange QoS mismatch).
+        # The crd-apply init container installs CRDs asynchronously after Helm returns.
+        # Wait for the resource to exist before checking its Established condition.
```
dmitry-tokarev-nv: `wait --for=create` was added in kubectl v1.31, so this step now requires kubectl ≥ 1.31. The action's `kubectl_version` default (`v1.32.13`) is safe, but the input is overridable — a caller pinning an older kubectl would now fail here rather than at the previous `--for=condition=established` call. A one-line note documents the new floor:

```suggestion
        # The crd-apply init container installs CRDs asynchronously after Helm returns.
        # Wait for the resource to exist before checking its Established condition.
        # `wait --for=create` requires kubectl >= 1.31 (this action defaults to v1.32.13).
```

deploy/helm/README.md
```diff
@@ -23,7 +23,12 @@ The following Helm chart is available for the Dynamo Kubernetes Platform:
...
\ No newline at end of file
+The cluster-wide operator manages Custom Resource Definitions (CRDs) automatically. CRD manifests
+are generated under [`deploy/operator/config/crd/bases`](../operator/config/crd/bases/) and bundled
+in the operator image.
+
+- **Initial installation and upgrades**: The operator Deployment's `crd-apply` init container applies
+  CRDs from the operator image using server-side apply before the manager starts.
+- **External management**: Set `dynamo-operator.upgradeCRD=false` when another process manages CRDs.
+  Namespace-restricted operators require this setting and use the CRDs managed by the cluster-wide
+  operator.
```
dmitry-tokarev-nv: This is the one behavior change worth calling out. Previously Helm's native `crds/` install populated CRDs on **every** `helm install` regardless of `upgradeCRD` (that's why namespace-restricted mode also needed `--skip-crds`). With the `crds/` directory gone, the `crd-apply` init container — gated on `and (not .Values.namespaceRestriction.enabled) .Values.upgradeCRD` in `deployment.yaml` — is the sole installer. So a **cluster-wide** install with `upgradeCRD=false` now installs **no CRDs at all**, where before it still got them from the chart on first install. Suggest making that explicit here so nobody trips over it:

```suggestion
- **External management**: Set `dynamo-operator.upgradeCRD=false` when another process manages CRDs.
  The chart no longer ships a `crds/` directory, so with this set it installs no CRDs itself — a
  cluster-wide operator started with `upgradeCRD=false` requires the CRDs to be applied out-of-band
  before it can reconcile. Namespace-restricted operators require this setting and use the CRDs
  managed by the cluster-wide operator.
```

deploy/helm/charts/platform/components/operator/values.yaml
```diff
@@ -16,9 +16,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-# Whether the cluster-wide operator applies CRDs via an init container.
-# This must be false for namespace-restricted installations, which must also
-# be installed with Helm's --skip-crds flag.
+# Whether the cluster-wide operator applies CRDs from its image through the crd-apply init container.
+# This must be false for namespace-restricted installations, which use the CRDs
+# managed by the cluster-wide operator.
```
dmitry-tokarev-nv: Follow-on from the README note: this help text (and the identical block in `deploy/helm/charts/platform/values.yaml`, which feeds the generated README table) frames `upgradeCRD=false` as being "for namespace-restricted installations" but doesn't mention that a **cluster-wide** install with `upgradeCRD=false` now installs no CRDs at all. Worth a caveat so the meaning is clear from `helm show values` (update both files if you take this):

```suggestion
# Whether the cluster-wide operator applies CRDs from its image through the crd-apply init container.
# When false, the chart installs no CRDs; a cluster-wide operator will not reconcile until CRDs are
# applied out-of-band. This must be false for namespace-restricted installations, which use the CRDs
# managed by the cluster-wide operator.
```

dmitry-tokarev-nv (approved):
LGTM. claude posted non-blocking comments if you'd like to review them

sttts (approved):

.github/actions/setup-dynamo-operator/action.yml
julienmancuso: fixed

deploy/helm/README.md
julienmancuso: fixed

deploy/helm/charts/platform/components/operator/values.yaml
julienmancuso: fixed

deploy/helm/charts/platform/Makefile
```diff
@@ -69,10 +69,23 @@ helm-unittest-install: ## Install helm-unittest plugin if not present
 charts-prep: ## Package operator subchart for local testing
 	@helm package components/operator -d charts/ 2>/dev/null
 
+.PHONY: verify-operator-chart-crds
+verify-operator-chart-crds: ## Verify operator CRDs are managed from the image, not the Helm chart
+	@tmp="$$(mktemp)"; \
+	trap 'rm -f "$$tmp"' EXIT; \
```
mc-nv, line 75: What is the purpose of this `trap` command for?

julienmancuso: It registers cleanup for the temporary file when the recipe shell exits, whether successfully or after an error. The file lets us check helm show crds’ exit status separately from grep, avoiding the original pipeline issue where a Helm failure could be masked.

mc-nv: Thanks for the explanation.  Manipulation with `tmp` location could  redundant in favor of output interception.

Issue #11673, opened 2026-07-14
DEP: KV Cache Controller


mkhazraee:
### Area

kv-memory

### Summary

This DEP describes a plan to improve the KV Block Manager (KVBM) by making better use of capabilities already present in the KV router and by simplifying the cache-management boundary. The resulting component is called the KV Cache Controller throughout this DEP. 

Existing systems such as the Dynamo KV Router demonstrate that global cache awareness and request routing can be achieved efficiently. Because the router already tracks KV cache placement and uses it in routing decisions, it is a natural component to work hand in hand with the cache controller rather than duplicate that global view elsewhere. The router can also incorporate harness-level context, such as relationships among sub-agent requests, and pass relevant hints to the KV Cache Controller. Continued improvements to this coordination layer—such as the [Flash Indexer](https://docs.nvidia.com/dynamo/dev/digest/flash-indexer#7-benchmarks)—make that boundary increasingly practical at scale. The KV Cache Controller can therefore focus on local cache policy, residency, and data movement while reusing the router's global inventory and request-scoped hints.

The KV Cache Controller will be developed in the open as part of Dynamo. A public repository will be available soon, with the design and implementation evolving through community collaboration. This issue is for discussions and feedback from the community.

### Motivation

Inference engines managing KV caches face a hierarchy of storage tiers — HBM (G1) → Host DRAM / controller-attached pool (G2) → Local SSD (G3) → Object Store (G4) — across both single nodes and large clusters. Current approaches suffer from one of two failure modes: tight GPU coupling, where external entities manage GPU-side KV blocks and compete for kernel launch bandwidth, requiring continuous synchronization with the engine; or naive simplicity, where eviction happens reactively under peak memory pressure with no admission filtering and no cross-node coordination. Neither scales cleanly. In multi-node disaggregated serving, the problem compounds: prefix caches are distributed, routing decisions affect hit rates, and coordination without a clear authority creates either duplicated state or missed reuse.

### Goals

**Non-Intrusiveness:** The KV Cache Controller must not put blocking work on the engine's hot path or interfere with memory the framework uses for execution. Cache data that the framework depends on must retain a local, fast-access guarantee, whether the framework holds the bytes directly or delegates that guarantee to the cache layer. Framework-facing calls either read local metadata, enqueue work, or complete asynchronously, while internal caching remains transparent to the engine.

[...]

PR #11695, opened 2026-07-15
feat: Add compact router hints for remote KV reuse

oandreeva-nv:
## Overview:

This PR adds an experimental router-generated KV reuse hint path for backend workers that advertise router-hint support through runtime metadata.

The KV router can attach a compact advisory `router_hint` to selected backend requests under:

```
extra_args.kv_transfer_params.router_hint
```

Dynamo still selects the target worker using the existing scheduler path. The hint is attached only after target selection, and only when the router sees a better non-target source prefix. The target backend remains responsible for deciding what to fetch, validating availability, and falling back safely if the hinted blocks cannot be used.

Current hint shape:

```json
{
  "source_control_endpoint": "tcp://host:port",
  "block_hashes": [123, 456]
}
```

`source_control_endpoint` is required in the request hint. Backend integrations that advertise router-hint support are responsible for publishing a usable source-control endpoint. If a backend advertises `router_hint` capability but no advertisable endpoint can be derived, worker registration fails.

### 1. Runtime capability gating

Router hints are enabled by discovered worker capability, not by a user-facing router flag.

A worker opts in by publishing:

```text
ModelRuntimeConfig.runtime_data["router_hint"] = true
```

Backend integrations that can be selected as hint sources also publish:

```text
ModelRuntimeConfig.runtime_data["router_hint_source_control_endpoint"] = "tcp://host:port"
```

The capability bit is a backend integration contract: Dynamo uses it as an opt-in guard for hint preparation and attachment, while the backend integration that sets it is responsible for ensuring the selected engine consumes the hint correctly. Source selection also requires a source-control endpoint; missing or inconsistent metadata causes hint construction to be skipped instead of panicking.

`WorkerConfigLike` exposes `supports_router_hints()` and `router_hint_source_control_endpoint()` so source selection stays backend-agnostic. vLLM is the first consumer; SGLang and TensorRT-LLM can opt in later by publishing the same runtime metadata instead of adding backend-specific branches in `KvRouter`.

### 2. Worker capability advertisement

Backend integrations opt in by publishing router-hint runtime metadata on `ModelRuntimeConfig.runtime_data`.

Dynamo's vLLM integration publishes this through `enable_router_hint_support(...)` during base model registration when a secondary tier advertises `router_hint` in `router_capabilities`.

There are two `router_hint` keys in different namespaces:

- request payload: `extra_args.kv_transfer_params.router_hint`
- worker capability: `ModelRuntimeConfig.runtime_data["router_hint"] = true`

`source_control_endpoint` is required in the request hint. Backend integrations that advertise router-hint support are responsible for publishing a usable source-control endpoint under:

```text
ModelRuntimeConfig.runtime_data["router_hint_source_control_endpoint"]
```

This keeps the capability contract internally consistent. A worker that advertises `router_hint` may be selected as a hint source, and the request hint requires a `source_control_endpoint`. If that endpoint cannot be derived during registration, the backend integration fails worker registration instead of publishing partial metadata. The router still checks endpoint presence during source selection as a defensive fail-closed guard for stale or inconsistent runtime metadata.

`WorkerConfigLike` exposes `supports_router_hints()` and `router_hint_source_control_endpoint()` so the router remains backend-agnostic. vLLM is the first consumer; SGLang and TensorRT-LLM can opt in later by publishing the same runtime metadata instead of adding backend-specific branches in `KvRouter`.

### 3. Router-side hint construction

Adds `lib/kv-router/src/router_hint.rs` with:

- `RouterHint`: serialized request hint attached to backend requests
- `RouterHintRootCandidates`: retained root-aligned source candidates prepared during tiered lookup
- `best_source(...)`: selects the eligible source with the longest root prefix that beats the target prefix

When any registered worker advertises router-hint support, the primary/device lookup retains the root-aligned external KV hash prefix it already matched. The lower-tier lookup uses the existing per-worker continuation walk and optionally records the external child hashes it traverses. Dynamo combines the device prefix with lower-tier extensions to build root-aligned candidates without running a second lower-tier root walk.

After the scheduler selects the target worker, `KvRouter` attempts hint construction without changing that routing decision. A hint is attached only when:

- the selected target advertises router-hint support
- the selected source is different from the selected target
- the selected source advertises router-hint support
- the selected source has a source-control endpoint
- the selected source has a longer root-prefix match than the target's local cached prefix

If the target is the best matching owner, if no better non-target source remains, or if the selected source metadata is incomplete, the request proceeds without a router hint.

### 4. Request pass-through to vLLM

Dynamo attaches hints under `extra_args.kv_transfer_params.router_hint`. If existing `kv_transfer_params` is malformed (`null`, scalar, or array), the router replaces it with a valid object containing the hint instead of silently dropping the hint. Existing object-valued `kv_transfer_params` are preserved and extended.

Dynamo's vLLM handler forwards only the allowlisted `router_hint` key from incoming `kv_transfer_params` into vLLM `SamplingParams`. Existing trusted/internal `sampling_params.extra_args["kv_transfer_params"]` keys are preserved, while unrelated request-provided connector keys are dropped.

### 5. Current limitation

- Router hints are currently prepared only for the local primary indexer path. Remote primary indexer mode does not yet propagate router-hint chain-retention metadata through the remote indexer protocol, so that path logs a warning and proceeds without router hints.

- vLLM is the only supported backend at the moment.

Start with the hint contract and final request mutation:

- `lib/kv-router/src/router_hint.rs`
- `lib/llm/src/protocols/common/preprocessor.rs`

Then review the router decision path:

- `lib/llm/src/kv_router.rs`
- `lib/llm/src/kv_router/indexer/lookup.rs`

Then review how lookup preserves enough data for that decision:

- `lib/kv-router/src/indexer/lower_tier.rs`
- `lib/kv-router/src/indexer/lower_tier_indexers.rs`
- `lib/llm/src/kv_router/route_lookup.rs`

Finally review backend integration plumbing:

- `components/src/dynamo/vllm/router_hints.py`
- `components/src/dynamo/vllm/handlers.py`

Tests cover:

- vLLM runtime metadata advertisement from backend router capabilities
- vLLM startup failure when router-hint capability lacks an advertisable endpoint
- pass-through allowlisting of `router_hint` into vLLM `SamplingParams`
- preservation of trusted/internal `kv_transfer_params` while dropping unrelated request-provided connector keys
- device-prefix retention, lower-tier extension collection, and source selection
- request attachment under `extra_args.kv_transfer_params.router_hint`
- malformed `kv_transfer_params` replacement instead of silent hint drop
- fail-closed cases for missing capability/candidates

## Related Issues

- Related DEP: https://github.com/ai-dynamo/dynamo/issues/11673 

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11695" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added experimental router hints for KV routing to improve worker selection based on cached KV prefixes.
  * Router hints can be enabled through configuration, CLI flags, or the `DYN_ROUTER_HINTS` environment variable.
  * Added support for forwarding router hints through vLLM requests and dynamically published model deployments.
  * Added source-control endpoint discovery for eligible KV transfer configurations.
* **Validation**
  * Router hints now require KV router mode and are disabled by load-aware routing.


components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,52 @@
...
+        host = _get(tier, "control_advertise_host") or _get(tier, "control_host")
+        if not isinstance(host, str) or not host or host in {"0.0.0.0", "::"}:
+            return None
+        return f"tcp://{host}:{control_port}"
+
+    return None
+
+
+def enable_router_hint_support(runtime_config: Any, engine_args: Any) -> None:
+    runtime_config.set_engine_specific(ROUTER_HINT_RUNTIME_CAPABILITY_KEY, "true")
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

components/src/dynamo/vllm/handlers.py
```diff
@@ -835,6 +836,13 @@ def build_sampling_params(
         configured_default = default_sampling_params.get("max_tokens", dynamic_default)
         sampling_params.max_tokens = min(configured_default, dynamic_default)
 
+    if isinstance(extra_args, dict):
+        kv_transfer_params = extra_args.get("kv_transfer_params")
+        if isinstance(kv_transfer_params, dict):
+            if sampling_params.extra_args is None:
+                sampling_params.extra_args = {}
+            sampling_params.extra_args["kv_transfer_params"] = kv_transfer_params
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

lib/kv-router/src/protocols.rs
```diff
@@ -181,6 +181,17 @@ pub trait WorkerConfigLike {
...
+    fn supports_router_hints(&self) -> bool {
+        false
+    }
+
+    /// Advertised peer-control endpoint used when this worker is selected as a router_hint source.
+    /// Backends that advertise router-hint support are responsible for publishing a usable endpoint.
+    fn router_hint_source_control_endpoint(&self) -> Option<&str> {
+        None
+    }
+
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

lib/llm/src/kv_router.rs
```diff
@@ -401,6 +453,56 @@ where
...
+                            .is_some_and(|config| config.supports_router_hints())
+                })?;
+            let source_config = configs.get(&source.worker_id)?;
+            let source_control_endpoint = source_config
+                .router_hint_source_control_endpoint()
+                .unwrap_or_else(|| {
+                    panic!(
+                        "router_hint-capable source worker {source:?} must advertise a control endpoint"
+                    )
+                })
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

lib/llm/src/kv_router.rs
```diff
@@ -588,8 +690,11 @@ where
             tokens,
             self.block_size,
             block_hashes,
-            cache_namespace.as_deref(),
-            retain_block_hashes,
+            TieredLookupOptions {
+                cache_namespace: cache_namespace.as_deref(),
+                retain_block_hashes,
+                retain_router_hint_chain: self.kv_router_config.router_hints,
+            },
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

lib/llm/src/kv_router/indexer/lookup.rs
```diff
@@ -172,6 +190,7 @@ impl<'a> LookupPipeline<'a> {
     async fn find_matches_by_tier(
         &self,
         sequence: HashInput<'_>,
+        lower_tier_options: LowerTierQueryOptions,
```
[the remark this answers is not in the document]
oandreeva-nv: Router hints are currently fail-closed for remote primary indexer mode: the remote protocol does not yet carry retain_router_hint_chain, so we log a warning, skip hint attachment, and leave protocol extension as follow-up work.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -304,6 +304,16 @@ impl KvPushRouter {
...
+        if let Some(router_hint) = selection.router_hint.as_ref()
+            && let Err(error) = backend_input.attach_router_hint(router_hint)
+        {
+            tracing::warn!(
+                request_id = %context_id,
+                worker_id = selection.instance_id,
+                error = %error,
+                "Failed to attach router_hint to backend request"
+            );
+        }
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

lib/llm/src/protocols/common/preprocessor.rs
```diff
@@ -365,6 +368,25 @@ impl PreprocessedRequest {
...
+                kv_transfer_params.insert(ROUTER_HINT_EXTRA_ARGS_KEY.to_string(), hint_value);
+            }
+        } else {
+            let mut kv_transfer_params = serde_json::Map::new();
+            kv_transfer_params.insert(ROUTER_HINT_EXTRA_ARGS_KEY.to_string(), hint_value);
+            map.insert(
+                KV_TRANSFER_PARAMS_EXTRA_ARGS_KEY.to_string(),
+                serde_json::Value::Object(kv_transfer_params),
+            );
+        }
```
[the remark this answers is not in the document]
oandreeva-nv: should be addressed, please double check

components/src/dynamo/vllm/router_hints.py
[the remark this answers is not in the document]
oandreeva-nv: Pushed follow-up fixes addressing the router-hint review comments:

  - Router-hint capability registration now requires a valid advertised source-control endpoint and enforces a single router-hint-capable secondary tier.
  - Router source selection also fails closed if endpoint metadata is missing, so missing/inconsistent metadata no longer panics the router.
  - vLLM `SamplingParams.extra_args` handling is now copy-on-write before inserting the allowlisted `router_hint`.
  - Router-hint retention is disabled for unsupported indexer configurations, and retained hash chains are trimmed to confirmed prefix depth.

lib/llm/src/kv_router.rs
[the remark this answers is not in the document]
oandreeva-nv: Double-checked. Remote indexer support is intentionally fail-closed for this PR rather than plumbed end-to-end. `supports_router_hint_chain_retention()` returns false for `Indexer::Remote`, and `query_tiered_matches` clears `retain_router_hint_chain` with a warning before constructing `LowerTierQueryOptions`, so remote primary lookup cannot silently attach incomplete router hints. The old `router_hints` config flag no longer exists; enablement is now capability-driven, with local-indexer support as an additional runtime gate. Extending the remote indexer protocol to carry retained router-hint chains remains follow-up work.

lib/kv-router/src/protocols.rs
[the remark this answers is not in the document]
oandreeva-nv: updated in [5fb5ff8](https://github.com/ai-dynamo/dynamo/pull/11695/commits/5fb5ff83b7506a22dfe5738e0c694aa29892d1c7)

lib/kv-router/src/indexer/concurrent_radix_tree_compressed/matches.rs
```diff
@@ -28,27 +42,35 @@ impl ConcurrentRadixTreeCompressed {
...
         sequence: S,
         early_exit: bool,
+        retain_router_hint_chain: bool,
     ) -> MatchDetails {
         let mut details = MatchDetails::new();
         if sequence.len() == 0 {
             return details;
         }
+        let mut router_hint_root_chain =
+            retain_router_hint_chain.then(|| Vec::with_capacity(sequence.len()));
```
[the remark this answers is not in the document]
oandreeva-nv: no observed overhead based on local experimentation

PeaBrane (commented):
Architectural note only: I have not done a broad correctness or merge-readiness review of this PR yet.

The initial indexer probe already produces the tiered matches used to derive WSPT's best eligible overlap, and this change derives router-hint candidates from that same snapshot. If a request waits long enough, the existing dequeue-time overlap refresher performs the conditional second probe and replaces only `OverlapSignals`; final worker selection can therefore use refreshed overlap while `router_hint_for_selection()` still uses the initial hint-candidate/target-prefix snapshot.

Could you evaluate having both the initial lookup and the existing refresh produce a common cache snapshot, for example `{ overlap: OverlapSignals, router_hint_candidates: Option<RouterHintRootCandidates> }`? When the existing refresh runs, it could replace both fields, so final worker selection and the better-source check use the same snapshot. This would preserve the current WSPT semantics—queue ordering remains based on enqueue-time overlap and refresh still happens after dequeue—while reusing the existing conditional refresh and avoiding a third post-selection indexer probe.

ziqifan617:
AI flagged a potential issue which I think worth follows up, though we haven't spent efforts on disagg mode yet:

If looking at the PrefillWorkerHandler:
```
sampling_params.extra_args[
            "kv_transfer_params"
        ] = kv_protocol.prefill_request_kv_transfer_params()
```
it will reset the kv_transfer_params and drop the router hint

Suggested fix — preserve the hint across the replacement at line 3454:
```
new_params = kv_protocol.prefill_request_kv_transfer_params()
existing = (sampling_params.extra_args or {}).get("kv_transfer_params")
if isinstance(existing, dict) and _ROUTER_HINT_EXTRA_ARGS_KEY in existing:
    new_params[_ROUTER_HINT_EXTRA_ARGS_KEY] = existing[_ROUTER_HINT_EXTRA_ARGS_KEY]
sampling_params.extra_args["kv_transfer_params"] = new_params
```

oandreeva-nv:
Thanks @PeaBrane, this is a fair architectural point. The current implementation intentionally keeps router-hint candidates tied to the initial lookup snapshot, while the existing dequeue-time refresh only updates OverlapSignals for final target selection.

I agree a shared cache snapshot shape would be cleaner long-term, especially if we want refreshed hint-source candidates without adding another post-selection lookup. The main reason I’m not doing that in this PR is scope/API boundary: OverlapSignals is the scheduler-facing data that already crosses from the dynamo-llm router path into the dynamo-kv-router scheduler path. Making refresh return { overlap, router_hint_candidates } would require extending that scheduler request/response shape and plumbing RouterHintRootCandidates through the same boundary.

For this PR, I think the current behavior is acceptable because the refresh path is primarily about improving target-worker selection under queueing delay, while the router hint is advisory source metadata. KVCC uses the hint to send a source-side pin/search request, and the source only returns descriptors for blocks it can actually find and pin. If the hinted blocks are unavailable or the transfer cannot complete, the backend falls back/recomputes rather than trusting the router hint as authoritative. For the initial vLLM/KVCC path, where secondary population is effectively write-through, a stale hint snapshot may reduce reuse quality or cause fallback, but should not affect target selection correctness or generation correctness.

I’ll keep this as a follow-up candidate: have both initial lookup and refresh produce a common { overlap, router_hint_candidates } snapshot so final target selection and hint-source selection are based on the same refreshed view.

PeaBrane:
@oandreeva-nv Thanks, that makes sense from a correctness perspective. The advisory fallback explains why the current behavior is safe, but I see that as a safety argument rather than a reason to keep two snapshot lifecycles.

Given the ongoing router modularization work, I’d prefer that we fold the unified snapshot into this PR unless there is a stronger architectural or ownership reason against it. This PR is establishing the router-hint lifecycle for the first time, so it seems worth getting the interface right now: have the initial lookup and existing dequeue-time refresh both produce a common `{ overlap, router_hint_candidates }` snapshot, while preserving the current refresh threshold, decay, queue ordering, and WSPT semantics.

That would reuse the existing tiered lookup rather than introducing a separate post-selection/third probe. I think the additional scheduler request/response plumbing is worthwhile here to avoid landing two snapshot semantics and revisiting the boundary later.

Is there a layering or ownership concern beyond the additional plumbing that I may be missing? I’d like to understand the tradeoff better on my end. Otherwise, my preference would be to fold this into the current PR if you’re open to it.

oandreeva-nv:
@PeaBrane sounds good, let me follow up with the required changes then

lib/llm/src/kv_router.rs
```diff
@@ -820,6 +904,8 @@ where
                 block_hashes: block_hashes_for_refresh,
                 isl_tokens,
                 overlap,
+                router_hint_candidates,
+                retain_router_hint_chain,
```
PeaBrane: The initial lookup masks this capability with `indexer.supports_router_hint_chain_retention()`, but this ungated value is carried into the scheduler and used directly by the aged dequeue refresh. As a result, `use_kv_events=false` and side-indexer configurations can start producing candidates only after queue aging, despite correctly failing closed initially. Please compute and carry the effective gated value once, or enforce the same gate inside the refresher, and add a queued-refresh regression.

components/src/dynamo/vllm/handlers.py
```diff
@@ -2088,6 +2129,12 @@ async def _register_lora_discovery(self, lora_name: str, lora_id: int) -> None:
             lora_needs_set = []
         if self.config.route_to_encoder:
             lora_needs_set.append(WorkerType.Encode)
+        enable_router_hint_support(
```
PeaBrane, line 2176: This fresh `ModelRuntimeConfig` still advertises the default DP range `(0, 1)`, while `enable_router_hint_support` builds endpoint keys from `self.dp_range`. For example, `(4, 2)` publishes endpoints for ranks 4/5 but the card advertises rank 0/size 1, breaking LoRA+DP routing/source lookup. Please set `data_parallel_start_rank` and `data_parallel_size` before registration, preferably through shared base-card setup.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -719,7 +743,13 @@ fn advance_state_to_breakpoint(
             }
         }
 
-        cur_hash = Some(edge.child_hash());
+        let child_hash = edge.child_hash();
+        if let Some(extensions) = router_hint_extensions.as_deref_mut() {
+            for worker in &active {
```
PeaBrane: This materializes the same matched hash once per active worker for every HostPinned block, then `merge_router_hint_tier_candidates` scans those per-worker vectors to emit one shared chain plus owner depths. The intermediate is `O(W × L)` allocations/storage on the lookup hot path and is repeated by dequeue refresh. Please build the canonical chain once and track only per-owner depths, or otherwise avoid the per-worker hash copies.

lib/kv-router/src/indexer/types.rs
```diff
@@ -358,12 +360,43 @@ pub struct MatchDetails {
...
 impl MatchDetails {
     pub fn new() -> Self {
         Self::default()
     }
+
+    pub fn retain_router_hint_root_candidates(
+        &mut self,
+        mut block_hashes: Vec<ExternalSequenceBlockHash>,
+    ) {
+        let mut owner_prefix_blocks: Vec<_> = self
```
PeaBrane, line 376: `owner_prefix_blocks` is derived from all local-hash owners, but the radix insert path only warns when another worker reports a different external block-hash chain and still coalesces ownership. That can select worker B while sending worker A's hashes; fallback remains safe, but remote reuse silently fails. Please retain per-owner hash compatibility or exclude mismatched owners from candidates, and add a divergent-hash regression.

PeaBrane (approved):
AI-generated review comments, posted on my behalf. The overall architecture and unified refresh/no-third-probe direction look sound. Approving on the condition that the substantive inline comments below are resolved before merge.

components/src/dynamo/vllm/handlers.py
oandreeva-nv: addressed

lib/llm/src/kv_router.rs
oandreeva-nv: addressed

lib/kv-router/src/indexer/lower_tier.rs
oandreeva-nv: addressed

lib/kv-router/src/indexer/types.rs
oandreeva-nv: addressed

jthomson04 (commented):
P1/P2 findings from a source-only review of the current head. No tests or benchmarks were run locally.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -973,6 +979,8 @@ impl<
                 effective_overlap_blocks: selected.selection.effective_overlap_blocks,
                 cached_tokens: selected.selection.cached_tokens,
                 selected_worker_tiers: selected.selected_worker_tiers,
+                overlap: request.overlap.clone(),
```
jthomson04: [P1] This clones all overlap maps and the optional candidate chain into every scheduling response, including when router hints are disabled. That adds O(workers) allocation/copy work to the request path. Please move ownership or return only the selected target prefix needed for hint construction.

lib/llm/src/kv_router.rs
```diff
@@ -850,6 +929,14 @@ where
 
         let supports_overlap_refresh = self.scheduler.supports_overlap_refresh();
         let retain_block_hashes = supports_overlap_refresh || return_routing_hashes;
+        let has_router_hint_capable_workers = self.has_router_hint_capable_workers();
+        let retain_router_hint_chain =
+            has_router_hint_capable_workers && self.indexer.supports_router_hint_chain_retention();
+        if has_router_hint_capable_workers && !retain_router_hint_chain {
```
jthomson04: [P1] This warning runs once per routed request whenever capable workers use an unsupported indexer configuration. Remote-primary or approximate-indexer deployments will flood logs and contend on the logging path. Please emit this during configuration/capability changes, or guard it so it logs once.

lib/llm/src/kv_router.rs
```diff
@@ -566,6 +593,58 @@ where
...
+            let prefix_blocks_to_beat =
+                usize::try_from(target_cached_prefix_blocks).unwrap_or(usize::MAX);
+            let (source, block_hashes) =
+                candidates.best_source(prefix_blocks_to_beat, |worker| {
+                    worker != target
+                        && configs.get(&worker.worker_id).is_some_and(|config| {
+                            config.supports_router_hints()
+                                && config.router_hint_worker_type() == Some(target_worker_type)
+                                && config
+                                    .router_hint_source_control_endpoint_for_dp_rank(worker.dp_rank)
```
jthomson04: [P2] The eligibility predicate constructs an owned endpoint String for every candidate, and the winner is looked up again below. ModelRuntimeConfig also allocates the DP-rank key and may parse JSON on each call. Please normalize endpoint metadata once and use a borrowed or otherwise non-allocating eligibility check.

lib/llm/src/kv_router.rs
```diff
@@ -947,6 +1042,14 @@ where
                 Err(error) => return Err(map_scheduler_error(error)),
             },
         };
+        let target_cached_prefix_blocks =
```
jthomson04: [P2] Router-hint construction also runs for WithoutAdmission, but FindBestMatchAdvisoryOutcome cannot carry the result. Conditional-bypass probes therefore scan candidates and clone the selected hash prefix only to discard it. Please skip hint construction for advisory selections.

components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,124 @@
...
+        tier for tier in _secondary_tiers(engine_args) if _supports_router_hint(tier)
+    ]
+
+
+def _router_hint_source_control_endpoint(tier: Any, port_offset: int = 0) -> str | None:
+    try:
+        control_port = int(_get(tier, "control_port")) + port_offset
+    except (TypeError, ValueError):
+        return None
+    if control_port <= 0:
```
jthomson04: [P2] This accepts final ports above 65535 and formats IPv6 hosts without brackets, so registration can succeed with unusable endpoints such as tcp://2001:db8::1:23280 or a DP-offset port of 65536. Please validate the final port and format IPv6 addresses correctly.

lib/kv-router/src/scheduling/queue.rs
oandreeva-nv: Addressed. SchedulingResponse no longer clones full overlap maps; the scheduler computes and returns only target_cached_prefix_blocks for hint construction. I also  moved ownership of SchedulingRequest through the advisory path where it was already  consumed, so router_hint_candidates can be moved instead of cloned.

lib/llm/src/kv_router.rs
oandreeva-nv: addressed, thanks for catching this

components/src/dynamo/vllm/router_hints.py
oandreeva-nv: addressed, thanks for catching this

lib/llm/src/kv_router.rs
oandreeva-nv: I addressed this with a small refactor. Router-hint worker metadata is now normalized once and read through borrowed metadata during source filtering, so we don’t allocate endpoint strings or parse endpoint JSON for every candidate. We only copy the endpoint after the final source is selected.

lib/llm/src/kv_router.rs
oandreeva-nv: no longer constructing hint in w/o admission case

components/src/dynamo/vllm/handlers.py
```diff
@@ -116,6 +117,9 @@
 _DELTA_REQUEST_OUTPUT_KIND = RequestOutputKind.DELTA
 _RL_INIT_WEIGHTS_TIMEOUT_ENV = "DYN_RL_INIT_WEIGHTS_TIMEOUT_S"
 _RL_INIT_WEIGHTS_TIMEOUT_DEFAULT_S = 30.0
+_LORA_LOCK_STRIPES = 64
```
karen-sy: I think we can remove `_LORA_LOCK_STRIPES`? Don't see used anywhere

jthomson04 (approved):

components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,141 @@
...
+    if address.is_unspecified:
+        return None
+    if address.version == 6:
+        return f"[{address.compressed}]"
+    return address.compressed
+
+
+def _router_hint_source_control_endpoint(tier: Any, port_offset: int = 0) -> str | None:
+    try:
+        control_port = int(_get(tier, "control_port")) + port_offset
```
karen-sy: Not a bug right now, but will there a guarantee from KVCC side that the control endpoint for local DP rank N is always `control_port + N`?

components/src/dynamo/vllm/handlers.py
```diff
@@ -817,6 +843,21 @@ def build_sampling_params(
     return sampling_params
 
 
+def _with_preserved_router_hint(
```
karen-sy: Looks like this helper is only used for prefill-side overwrite of the sampling params. But in decode, `build_sampling_params()` first preserves the router hint selected for the decode worker, but `DecodeWorkerHandler._generate_token_mode()` later replaces the entire dictionary at line 3283 with `prefill_result.disaggregated_params.kv_transfer_params`. That drops the decode worker’s router-generated hint before vLLM/KVCC receives the request.

Is this intended?

lib/llm/src/local_model/runtime_config.rs
```diff
@@ -787,6 +838,59 @@ mod tests {
...
+        assert_eq!(info.worker_type, "prefill");
+        assert!(info.source_control_endpoint.is_none());
+
+        config
+            .set_engine_specific(
+                ROUTER_HINT_SOURCE_CONTROL_ENDPOINTS_RUNTIME_KEY,
+                r#"{"0":"tcp://127.0.0.1:23280"}"#,
+            )
+            .unwrap();
+        let info = config.router_hint_metadata_for_dp_rank(0).unwrap();
```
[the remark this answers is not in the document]
oandreeva-nv: fixed

components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import Any
+
+from dynamo.common.constants import (
+    ROUTER_HINT_RUNTIME_CAPABILITY_KEY,
+    ROUTER_HINT_SOURCE_CONTROL_ENDPOINTS_RUNTIME_KEY,
+    ROUTER_HINT_WORKER_TYPE_RUNTIME_KEY,
+)
+
+
+def _set_engine_specific(runtime_config: Any, key: str, value: Any) -> None:
```
GuanLuo: Directly call `runtime_config.set_engine_specific(key, json.dumps(value))` instead of making a wrapper function

components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,141 @@
...
+
+    The PyO3 binding for ``set_engine_specific`` parses its argument with
+    ``serde_json::from_str``, so every value must be JSON text. Routing all
+    writes through here keeps that contract explicit — a bare ``"prefill"``
+    is not valid JSON and would raise at registration.
+    """
+    runtime_config.set_engine_specific(key, json.dumps(value))
+
+
+def _get(value: Any, key: str) -> Any:
```
GuanLuo: Seems defeating the purpose of typing

components/src/dynamo/vllm/router_hints.py
```diff
@@ -0,0 +1,141 @@
...
+    if isinstance(value, dict):
+        return value.get(key)
+    return getattr(value, key, None)
+
+
+def _secondary_tiers(engine_args: Any) -> list[Any]:
```
GuanLuo: give things proper type if possible

components/src/dynamo/vllm/tests/test_vllm_legacy_lora.py
```diff
@@ -107,6 +112,49 @@ async def test_prefill_load_records_and_publishes_without_eager_engine_add(
     assert kwargs["kv_cache_block_size"] == 1056
 
 
+@pytest.mark.asyncio
+async def test_prefill_lora_registration_preserves_worker_dp_range(monkeypatch):
```
GuanLuo, line 119: Why adding this test

components/src/dynamo/vllm/handlers.py
```diff
@@ -817,6 +843,21 @@ def build_sampling_params(
...
+) -> dict[str, Any]:
+    merged = dict(kv_transfer_params)
+    existing = (
+        extra_args.get(_KV_TRANSFER_PARAMS_EXTRA_ARGS_KEY)
+        if isinstance(extra_args, Mapping)
+        else None
+    )
+    if isinstance(existing, Mapping) and _ROUTER_HINT_EXTRA_ARGS_KEY in existing:
+        merged[_ROUTER_HINT_EXTRA_ARGS_KEY] = existing[_ROUTER_HINT_EXTRA_ARGS_KEY]
+    return merged
```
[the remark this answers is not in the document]
oandreeva-nv: addressed

lib/llm/src/kv_router.rs
```diff
@@ -566,6 +570,59 @@ where
         cache_hit_for_worker(cache_hit_estimates, worker)
     }
 
+    fn has_router_hint_capable_workers(&self) -> bool {
+        self.workers_with_configs.borrow().values().any(|config| {
+            let start = config.data_parallel_start_rank();
+            let end = start.saturating_add(config.data_parallel_size());
+            (start..end).any(|dp_rank| config.router_hint_metadata_for_dp_rank(dp_rank).is_some())
+        })
+    }
```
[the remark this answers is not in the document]
oandreeva-nv: Partially addressed: the request-path gate now checks one representative DP rank per worker instead of scanning every DP rank, so the default-path probe is O(workers), not O(workers * dp_size).

components/src/dynamo/vllm/handlers.py
oandreeva-nv: thanks!, deleted

components/src/dynamo/vllm/router_hints.py
oandreeva-nv: for now, yes, but if anything changes, will update

components/src/dynamo/vllm/handlers.py
oandreeva-nv: good catch, fixed

components/src/dynamo/vllm/handlers.py
```diff
@@ -817,6 +842,19 @@ def build_sampling_params(
...
+    sampling_params: SamplingParams,
+    kv_transfer_params: Mapping[str, Any],
+) -> None:
+    if sampling_params.extra_args is None:
+        sampling_params.extra_args = {}
+    merged = dict(kv_transfer_params)
+    existing = sampling_params.extra_args.get(_KV_TRANSFER_PARAMS_EXTRA_ARGS_KEY)
+    if isinstance(existing, Mapping) and _ROUTER_HINT_EXTRA_ARGS_KEY in existing:
+        merged[_ROUTER_HINT_EXTRA_ARGS_KEY] = existing[_ROUTER_HINT_EXTRA_ARGS_KEY]
+    sampling_params.extra_args[_KV_TRANSFER_PARAMS_EXTRA_ARGS_KEY] = merged
```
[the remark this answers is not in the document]
oandreeva-nv: Current code copies sampling_params.extra_args before writing:

 ```
 extra_args = (
      dict(sampling_params.extra_args)
      if isinstance(sampling_params.extra_args, dict)
      else {}
  )
  ...
  extra_args[_KV_TRANSFER_PARAMS_EXTRA_ARGS_KEY] = updated_params
  sampling_params.extra_args = extra_args
```
So it no longer does this unsafe in-place mutation: 
```
 sampling_params.extra_args["kv_transfer_params"] = merged
 ```
The build_sampling_params router-hint pass-through block also copies before mutation:

```
  passthrough_extra_args = (
      dict(sampling_params.extra_args)
      if isinstance(sampling_params.extra_args, dict)
      else {}
  )
  ...
  sampling_params.extra_args = passthrough_extra_args
```
So both paths now avoid mutating a shared/default extra_args dict.

lib/llm/src/kv_router/push_router.rs
```diff
@@ -350,6 +350,16 @@ impl KvPushRouter {
...
```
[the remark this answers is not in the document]
oandreeva-nv: addressed in two places.
 1. Rust router strips any incoming/stale hint before attaching its own current hint:

```
  let _ = backend_input
      .extra_args
      .as_mut()
      .and_then(serde_json::Value::as_object_mut)
      .and_then(|args| args.get_mut("kv_transfer_params"))
      .and_then(serde_json::Value::as_object_mut)
      .and_then(|params| params.remove("router_hint"));
```
So a request entering a router cannot carry an old router_hint through unchanged. If this router decides to attach a hint, it attaches a fresh one for this dispatch.

2. vLLM disagg handoff does not preserve prefill hints into decode.

Prefill path explicitly preserves the current request hint when replacing params with prefill protocol params:

```
  _update_kv_transfer_params(
      sampling_params,
      kv_protocol.prefill_request_kv_transfer_params(),
      preserve_router_hint=True,
  )

```
  Decode path applies prefill-produced kv_params without preserving router hint:

```
 _update_kv_transfer_params(sampling_params, kv_params)
```
and the helper only carries an existing hint when preserve_router_hint=True.

So a prefill-side hint can be used by prefill, but should not survive into the decode handoff. A decode-side hint would need to be freshly attached by the decode router for that decode dispatch.

components/src/dynamo/vllm/router_hints.py
oandreeva-nv: refactored

components/src/dynamo/vllm/router_hints.py
oandreeva-nv: refactored with actual types

components/src/dynamo/vllm/router_hints.py
oandreeva-nv: refactored with actual types

PR #11698, opened 2026-07-15
feat(router): add active sequence sparsity

jthomson04:
## Summary

- Add `--router-active-sequence-stride` / `DYN_ROUTER_ACTIVE_SEQUENCE_STRIDE` as a startup-only setting, defaulting to `1` and rejecting values below `1`.
- Retain rolling active-sequence hashes at stride endpoints while leaving KV-indexer hashes, routing-decision hashes, overlap signals, and token accounting unchanged.
- Correct externally visible active and potential decode-block estimates by scaling sparse prompt units while keeping generated-output blocks in physical units.
- Carry the stride on active-sequence add, free, and prefill-complete events; incompatible or malformed replica events are dropped before mutation and recorded through structured warnings and a Prometheus counter.

## Details

Active-sequence replica synchronization currently broadcasts the complete rolling-hash chain for every request. This makes payload size proportional to the full prompt length on every frontend.

This change computes the complete rolling chain first and then samples block positions `n, 2n, 3n, ...`. The sampled vector is used for local active-sequence tracking and is broadcast without receiver-side resampling. `BlockTracker` and `PromptMembershipTrie` remain sparsity-agnostic; outer load projection stores prompt and output components separately and converts sparse prompt units back into estimated physical blocks.

The corrected counts are estimates. Retained stride groups are fully represented, divergence inside a group can conservatively overcount overlap, and an incomplete trailing group is omitted.

`ActiveSequenceEvent.stride` is optional for wire compatibility: missing means stride `1`, while zero is malformed. A receiver drops only an incompatible event and continues processing subsequent compatible events.

Rollout must be coordinated. Deploy this binary everywhere with stride `1` first, then enable a common stride through a coordinated frontend restart. Older receivers may ignore the new field while still consuming the sparse vector.

- `lib/kv-router/src/scheduling/config.rs` for configuration and hash sampling.
- `lib/kv-router/src/sequences/prompt_registry.rs` and `multi_worker.rs` for load correction and event emission.
- `lib/kv-router/src/sequences/replica_sync.rs` and `lib/kv-router/src/protocols.rs` for compatibility enforcement.
- `lib/llm/src/kv_router/sequence.rs` and `metrics.rs` for frontend propagation and observability.

## Validation

- `cargo test -p dynamo-kv-router` — 752 passed, 2 ignored.
- `cargo test -p dynamo-llm`.
- `cargo check --manifest-path lib/bindings/python/Cargo.toml`.
- Shared KV-router configuration tests — 49 passed in an isolated CPU-only pytest environment.
- `cargo fmt --all -- --check`.
- Ruff lint and format checks for the changed Python files.
- `git diff --check`.

## Related Issues

- [x] Confirmed — no related issue.

## Summary by CodeRabbit

- **New Features**
  - Added configurable active-sequence sampling stride, available through CLI, environment variables, Python bindings, and router configuration.
  - Added stride-aware routing, load estimation, replay behavior, and replica synchronization.
  - Added validation to reject invalid stride values and incompatible replica events.
  - Added metrics for stride mismatches.

- **Bug Fixes**
  - Configuration errors now fail clearly instead of proceeding with invalid settings.

- **Documentation**
  - Documented stride behavior, synchronization requirements, upgrade considerations, and monitoring details.

### DIFF
--- a/components/src/dynamo/common/configuration/groups/kv_router_args.py
+++ b/components/src/dynamo/common/configuration/groups/kv_router_args.py
@@ -302,7 +302,9 @@ class KvRouterArgGroup(ArgGroup):
             help=(
                 "KV Router: Retain one active-sequence hash for every N complete prompt "
                 "blocks. A value of 1 disables sparsity. All router replicas must use "
-                "the same value."
+                "the same value. Active-block counts become estimates; prompts shorter "
+                "than N complete blocks and incomplete trailing groups contribute no "
+                "prompt hashes."
             ),
             arg_type=_positive_int,
         )
--- a/components/src/dynamo/common/configuration/groups/kv_router_args.py
+++ b/components/src/dynamo/common/configuration/groups/kv_router_args.py
@@ -35,6 +35,7 @@ _KV_ROUTER_FIELDS: tuple[str, ...] = (
     "use_kv_events",
     "router_replica_sync",
     "router_track_active_blocks",
+    "router_active_sequence_stride",
     "router_track_output_blocks",
     "router_assume_kv_reuse",
     "router_track_prefill_tokens",
@@ -101,6 +102,13 @@ def _default_prefill_load_scale() -> float:
     return 1.0
 
 
+def _positive_int(value: str) -> int:
+    parsed = int(value)
+    if parsed < 1:
+        raise argparse.ArgumentTypeError("value must be at least 1")
+    return parsed
+
+
 class KvRouterConfigBase(ConfigBase):
     """Mixin carrying the shared KvRouterConfig fields."""
 
@@ -114,6 +122,7 @@ class KvRouterConfigBase(ConfigBase):
     use_kv_events: bool
     router_replica_sync: bool
     router_track_active_blocks: bool
+    router_active_sequence_stride: int
     router_track_output_blocks: bool
     router_assume_kv_reuse: bool
     router_track_prefill_tokens: bool
@@ -283,6 +292,19 @@ class KvRouterArgGroup(ArgGroup):
                 "By default, active blocks are tracked for load balancing."
             ),
             obsolete_flag="--track-active-blocks",
+        )
+        add_argument(
+            g,
+            flag_name="--router-active-sequence-stride",
+            env_var="DYN_ROUTER_ACTIVE_SEQUENCE_STRIDE",
+            default=1,
+            dest="router_active_sequence_stride",
+            help=(
+                "KV Router: Retain one active-sequence hash for every N complete prompt "
+                "blocks. A value of 1 disables sparsity. All router replicas must use "
+                "the same value."
+            ),
+            arg_type=_positive_int,
         )
         add_negatable_bool_argument(
             g,
--- a/lib/kv-router/src/protocols.rs
+++ b/lib/kv-router/src/protocols.rs
@@ -3,7 +3,6 @@
 
 use std::collections::{HashMap, HashSet};
 use std::future::Future;
-use std::num::NonZeroUsize;
 use std::ops::Range;
 use std::sync::LazyLock;
 use std::time::Duration;
@@ -12,6 +11,9 @@ use dynamo_tokens::{SequenceHash, Token, compute_hash_v2, compute_next_sequence_
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use xxhash_rust::xxh3;
+
+use crate::active_sequence::ActiveSequenceStride;
+pub use crate::active_sequence::ActiveSequenceStrideError;
 
 const fn default_track_prefill_tokens() -> bool {
     true
@@ -633,18 +635,9 @@ pub struct ActiveSequenceEvent {
 
 impl ActiveSequenceEvent {
     /// Normalize the sender's stride, treating a missing value as the legacy stride of 1.
-    pub fn normalized_stride(&self) -> Result<NonZeroUsize, ActiveSequenceStrideError> {
-        match self.stride {
-            None => Ok(NonZeroUsize::MIN),
-            Some(stride) => NonZeroUsize::new(stride).ok_or(ActiveSequenceStrideError::Zero),
-        }
+    pub fn normalized_stride(&self) -> Result<ActiveSequenceStride, ActiveSequenceStrideError> {
+        ActiveSequenceStride::from_event_value(self.stride)
     }
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, thiserror::Error)]
-pub enum ActiveSequenceStrideError {
-    #[error("active-sequence stride must be greater than zero")]
-    Zero,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
@@ -2073,7 +2066,7 @@ mod tests {
         let serialized = serde_json::to_value(&event).unwrap();
         assert!(serialized.get("stride").is_none());
         let decoded: ActiveSequenceEvent = serde_json::from_value(serialized).unwrap();
-        assert_eq!(decoded.normalized_stride(), Ok(NonZeroUsize::MIN));
+        assert_eq!(decoded.normalized_stride(), Ok(ActiveSequenceStride::ONE));
     }
 
     #[test]
@@ -2092,7 +2085,10 @@ mod tests {
             serialized.get("stride").and_then(|value| value.as_u64()),
             Some(4)
         );
-        assert_eq!(event.normalized_stride(), Ok(NonZeroUsize::new(4).unwrap()));
+        assert_eq!(
+            event.normalized_stride(),
+            Ok(ActiveSequenceStride::new(4).unwrap())
+        );
 
         event.stride = Some(0);
         assert_eq!(
--- a/lib/kv-router/src/protocols.rs
+++ b/lib/kv-router/src/protocols.rs
@@ -622,8 +622,23 @@ pub struct ActiveSequenceEvent {
     pub worker: WorkerWithDpRank,
     pub data: ActiveSequenceEventData,
     pub router_id: u64,
+    /// Active-sequence sampling stride used by the sender. `None` is the
+    /// backwards-compatible representation of stride 1.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stride: Option<usize>,
     #[serde(default)]
     pub lora_name: Option<String>,
+}
+
+impl ActiveSequenceEvent {
+    /// Return the sender's effective stride, or `None` for a malformed zero.
+    pub fn effective_stride(&self) -> Option<usize> {
+        match self.stride {
+            None | Some(1) => Some(1),
+            Some(0) => None,
+            Some(stride) => Some(stride),
+        }
+    }
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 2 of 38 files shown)


lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -211,6 +248,7 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
             SequenceTrackerOptions {
                 replica_worker_policy: ReplicaWorkerPolicy::LazyRegister,
                 expiry_enabled: false,
+                active_sequence_stride: 1,
```
jthomson04: This hard-coded `1` is what breaks the mocker: `replay_slots` in `lib/mocker/src/replay/router_shared.rs` builds its tracker through `new_without_expiry`, while the replay routers (`replay/online/router.rs`, `replay/offline/components/router.rs`) sample hashes via `config.compute_seq_hashes_for_tracking(...)` with the user-supplied config. With stride > 1 the hashes get sampled but the tracker scales by 1, so replay under-counts decode blocks by the stride factor.

Suggest threading the stride through `new_without_expiry` (or asserting stride == 1 in `replay_router_config`). More broadly, the growing `new_with_*_and_*` constructor family (now 8 positional args behind a clippy allow) is how this kind of miss happens — making `SequenceTrackerOptions` public or adding a small builder would make the stride impossible to forget.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -162,6 +180,18 @@ fn kv_router_config_from_lookup(get_env: impl Fn(&str) -> Option<String>) -> KvR
...
+        let value = raw_value.parse::<usize>().unwrap_or_else(|_| {
+            panic!(
+                "DYN_ROUTER_ACTIVE_SEQUENCE_STRIDE must be an integer greater than or equal to 1"
+            )
+        });
+        assert!(
+            value >= 1,
+            "DYN_ROUTER_ACTIVE_SEQUENCE_STRIDE must be greater than or equal to 1"
+        );
+        config.router_active_sequence_stride = value;
```
jthomson04: This panics on a malformed value, but every sibling setting in `kv_router_config_from_lookup` silently ignores unparseable values (`parse_f64`/`parse_bool` return `Option`, and `dynamo_env_config_ignores_unparseable_values_and_validates_ranges` enshrines that convention). Fail-fast is defensible for a must-match-cluster-wide setting, but it's worth picking a lane: either the `Result`-returning style of `min_initial_workers_from_env`, or warn-and-default like the neighbors. As written, a typo'd env var crashes the frontend at startup.

lib/kv-router/src/sequences/replica_sync.rs
```diff
@@ -146,17 +146,36 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
     }
 
     fn apply_replica_event(&self, event: ActiveSequenceEvent, effects: &mut ReplicaBatchEffects) {
+        if event.router_id == self.router_id {
+            return;
+        }
+
+        let received_stride = event.effective_stride().unwrap_or(0);
+        if received_stride != self.active_sequence_stride {
+            tracing::warn!(
```
jthomson04: Two rollout-window consequences worth considering here:

1. **Log flooding.** This warns once per dropped event. During the coordinated stride change this PR prescribes, every not-yet-restarted peer emits add/free/prefill events for every request, so every restarted frontend logs a warning per event for the whole rollout window. Consider rate-limiting or logging once per `(source_router_id, received_stride)` — the counter already gives the per-event signal.
2. **Phantom load.** A request accepted before the divergence keeps its load when its `Free` arrives with the wrong stride (as `replica_sync_drops_only_mismatched_stride_events` shows), so replicas carry up to `router_ttl_secs` (120s default) of phantom decode load during rollout. By design, but worth a sentence in the operator docs alongside the rollout guidance.

lib/kv-router/src/protocols.rs
```diff
@@ -622,10 +622,25 @@ pub struct ActiveSequenceEvent {
...
+    /// backwards-compatible representation of stride 1.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub stride: Option<usize>,
     #[serde(default)]
     pub lora_name: Option<String>,
 }
 
+impl ActiveSequenceEvent {
+    /// Return the sender's effective stride, or `None` for a malformed zero.
+    pub fn effective_stride(&self) -> Option<usize> {
```
jthomson04: API-shape nit: `Option<usize>` where `None` means *malformed* — while `stride: None` on the wire means *1* — is an inversion that will trip up the next reader. An enum (`Stride::Valid(n) | Malformed`) or `Result` would be clearer. Relatedly, the caller funnels malformed into metrics via `unwrap_or(0)`, so dashboards see `received_stride="0"` as a magic value; a distinct label like `"malformed"` would read better.

lib/kv-router/src/sequences/block_tracker.rs
```diff
@@ -60,6 +60,18 @@ pub(super) struct ReleasedPromptPath {
...
 }
 
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub(super) struct BlockLoad {
+    pub(super) prompt_units: f64,
+    pub(super) output_blocks: f64,
+}
+
+impl BlockLoad {
+    pub(super) fn active_blocks(self, prompt_stride: usize) -> usize {
```
jthomson04: This is the same `prompt_units * stride + output_blocks → round` formula as `WorkerLoadSnapshot::active_decode_blocks` in `prompt_registry.rs`. Worth sharing one helper so the estimate definition lives in one place.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -305,6 +314,19 @@ def add_arguments(self, parser) -> None:
             ),
             obsolete_flag="--track-active-blocks",
         )
+        add_argument(
+            g,
+            flag_name="--router-active-sequence-stride",
```
jthomson04, line 298: Help-text suggestion: with stride N, prompts shorter than N complete blocks contribute zero prompt units (trailing group omitted), so at large `stride × block_size` a small-request-heavy workload systematically under-reports decode-block load. Worth a caveat here that counts become estimates and short prompts may be invisible to block tracking.

jthomson04 (commented):
Reviewed the full diff plus surrounding code at the PR head. Overall this is well-engineered: sampling at stride-group endpoints is sound for rolling hashes (each retained hash still uniquely identifies its full prefix), I traced every production path and found no double-sampling (SelectionCore samples once, including the cached-selection path; the lib/llm path samples inside `compute_seq_hashes_for_tracking`), replica sync uses name-based msgpack so the mid-struct optional field is wire-safe, and zero is rejected at every config boundary.

**One bug-shaped finding, in a file outside this diff:** the mocker/replay path is not wired for stride > 1. `replay_slots` (`lib/mocker/src/replay/router_shared.rs`) builds its tracker via `new_without_expiry`, which hard-codes `active_sequence_stride: 1`, but both replay routers sample hashes through `config.compute_seq_hashes_for_tracking(...)` with the user-supplied `KvRouterConfig`. With stride > 1, replay samples the hashes but never scales the load back up, so active/potential decode-block estimates under-count by the stride factor — silently skewing replay fidelity exactly when someone benchmarks this feature. Either thread the stride through, or reject stride > 1 in `replay_router_config`. (See inline comment on `new_without_expiry`.)

Other notes are inline: env-var panic vs. the file's ignore-unparseable convention, per-event warn flooding + phantom load during rollout, the `effective_stride` API shape, a duplicated formula, a help-text caveat, and two nits.

**Docs:** no user-facing docs are updated. Given the operational hazards (coordinated rollout required; old binaries deserialize sparse vectors as full chains and under-count peer load by ~stride×; dropped `Free` events leave phantom load until TTL), the flag deserves a section in the router docs rather than living only in the PR body.

**Tests:** coverage is strong (config validation across CLI/env/JSON/pyo3, sampling edge cases, serde round-trip + payload-size proof, mismatch dropping with metric assertions, projection scaling). Gaps: nothing exercises the mocker path with stride > 1 (would have caught the finding above), and no end-to-end test of two stride-N trackers syncing through the real publisher.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -492,6 +524,12 @@ pub struct KvRouterConfig {
     /// Whether to track active blocks in the router (default: true)
     pub router_track_active_blocks: bool,
 
+    /// Retain one active-sequence hash for every N complete prompt blocks.
+    /// A value of 1 disables sparsity.
+    #[serde(default = "default_active_sequence_stride")]
```
jthomson04: Nit: this `#[serde(default = ...)]` never fires — deserialization goes through `KvRouterConfigSerde` (`#[serde(try_from)]`), whose struct-level default supplies the 1. Harmless but misleading; the real default lives in the Serde mirror's `Default` impl.

lib/kv-router/src/scheduling/config.rs
```diff
@@ -823,19 +873,23 @@ impl KvRouterConfig {
...
+                    );
                 }
             };
-            Some(compute_seq_hash_for_block(block_hashes))
+            compute_seq_hash_for_block(block_hashes)
         } else {
             let mut rng = rand::rng();
-            Some((0..num_blocks).map(|_| rng.random::<u64>()).collect())
-        }
+            (0..num_blocks).map(|_| rng.random::<u64>()).collect()
```
jthomson04: Micro-nit: this generates `num_blocks` random hashes and then the sampling pass discards all but every Nth. Generating `num_blocks / stride` directly is equivalent for this branch.

jthomson04:
/ok to test [935e644](https://github.com/ai-dynamo/dynamo/commit/935e644c95fe1279d8866b8597706388ffbb090e)

lib/mocker/src/replay/offline/components/router.rs
```diff
@@ -214,7 +215,9 @@ impl OfflineReplayRouter {
         let config = replay_router_config(args, router_config);
         let worker_config_template = replay_worker_config(args);
         let workers_with_configs = replay_workers_with_configs(args, num_workers);
-        let slots = replay_slots(args, &workers_with_configs);
+        let active_sequence_stride = NonZeroUsize::new(config.router_active_sequence_stride)
+            .ok_or_else(|| anyhow!("router active-sequence stride must be greater than zero"))?;
+        let slots = replay_slots(args, &workers_with_configs, active_sequence_stride);
```
PeaBrane: With `slots` now stride-aware, the `Some(replay_hashes)` path in `build_pending_request` still passes the full `replay_hashes.sequence_hashes` chain when KV reuse is enabled. The tracker treats those dense hashes as sparse units and scales them again by N, overestimating active and potential decode load by roughly N×. Please sample that sequence before it enters `slots` and cover the `Some(ReplayRequestHashes)` path; the new test exercises only `None`.

lib/kv-router/src/services/selection/core/mod.rs
```diff
@@ -443,14 +445,21 @@ impl SelectionCore {
             &key.routing_group,
             block_size,
         );
-        let slots = Arc::new(ActiveSequencesMultiWorker::new_with_replica_worker_policy(
+        let active_sequence_stride =
+            NonZeroUsize::new(self.kv_router_config.router_active_sequence_stride)
+                .expect("KvRouterConfig validates active-sequence stride");
```
PeaBrane: `SelectionServiceBuilder::build` accepts `KvRouterConfig` without calling `validate_config`, so an embedded Rust caller can build successfully with stride 0 and then panic here on first worker registration. Please validate during build/startup or carry a validated `NonZeroUsize` into `SelectionCore`, returning a structured configuration error like the other entry points.

PeaBrane (commented):
Architectural suggestion: the two issues below share a missing invariant. `router_active_sequence_stride` remains a raw `usize`, while dense and sampled hashes are both represented as `Vec<SequenceHash>`, so every boundary must remember validation and exactly-once sampling.

Consider centralizing this behind an `ActiveSequenceStride(NonZeroUsize)` and an opaque tracked-hash constructor. This would make zero strides and dense/sparse representation mismatches harder to express. The local fixes below are still needed; the type refactor could follow separately if it is too broad for this PR.

jthomson04:
/ok to test [784f589](https://github.com/ai-dynamo/dynamo/commit/784f58950c920e8b3870442f0fc68f2514cbc321)

PR #11712, opened 2026-07-15
feat(snapshot): restore rootfs from SquashFS

galletas1712:
## Summary

- replace eager tar extraction of the captured container rootfs diff with a read-only SquashFS lower layer
- compose the restored root transparently with OverlayFS while preserving Kubernetes-managed child mounts and a pod-local writable upper layer
- keep `nsrestore` and CRIU on the stable placeholder control root, and pass the staged composed root to CRIU as the restored workload root
- preserve overlay whiteout/opaque-directory semantics and validate the mandatory SquashFS artifact with its manifest SHA-256
- pass detached rootfs, workspace, and checkpoint descriptors safely into namespace-scoped restore without exposing host loop devices beyond `/dev/loop-control`

This intentionally introduces the SquashFS artifact directly without a tar compatibility path or artifact schema framework. The checkpoint generation directory continues to provide atomic generation pinning, not format versioning.

## Validation

Validated at `3576bb21db`:

- `cd deploy/snapshot && go test -count=1 ./...`
- `cd deploy/snapshot && go test -race -count=1 ./internal/runtime ./internal/executor`
- `cd deploy/snapshot && go vet ./...`
- `git diff --check` and `gofmt -l` on all changed Go files
- focused unprivileged loop allocation and cleanup tests
- privileged `TestPrivilegedSnapshotOverPlaceholderRootComposition` (four passes)
- privileged `TestPrivilegedWorkspaceHandleCloneAcrossMountNamespaces`
- privileged `TestNSenterTargetSyntaxPreservesInheritedFD`

The privileged tests used Linux `6.17.0-35-generic`, real loop devices, SquashFS, and OverlayFS. They cover nine simultaneously live detached mounts, contention beyond eight allocation attempts, cleanup after mount and `open_tree` failures, snapshot-over-placeholder precedence, deletion and recursive opaque-directory semantics, owner/mode and ordinary-file xattr preservation, Kubernetes child-mount exclusion, and a same-workspace restore retry that removes stale upper-layer state.

Upstream Linux first provides the complete required syscall set in 5.8 (`open_tree`/`move_mount` in 5.2, `openat2` in 5.6, and `LOOP_CONFIGURE` in 5.8). Enterprise kernels may backport these independently, so rootfs preflight now performs a real scratch-file `LOOP_CONFIGURE` operation instead of relying on the reported kernel version.

The earlier [`DynamoCheckpoint` CI deploy flow](https://github.com/ai-dynamo/dynamo/actions/runs/29478775288) passed vLLM, SGLang, and TensorRT-LLM checkpoint/restore with coherent post-restore `RESTORE_OK` inference at `8678ad97837a2d6da3b9af6ee4f216c6879c052f`.

For current rebased head `3576bb21db`, [Pre Merge](https://github.com/ai-dynamo/dynamo/actions/runs/29582741286) passed, including the Snapshot job. The [PR workflow](https://github.com/ai-dynamo/dynamo/actions/runs/29582743422) could not start any DynamoCheckpoint deploy test because host Kyverno policy denied the synchronized operator ConfigMap in all three backend vClusters and the ordinary deploy vCluster. This reproduced the same shared-infrastructure failure as the pre-rebase attempt; no current-head post-restore inference ran. The separate ARM64 router synchronization failure is timing-sensitive and unrelated to the Snapshot paths changed here.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11712" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Snapshots now use validated, compressed root filesystem artifacts for improved restore handling.
  * Restore operations support dedicated root filesystem backing storage and safer filesystem composition.
  * Restore configuration now enforces isolated backing volumes for each target container.
  * Added support for loop-device-backed root filesystem mounts.

* **Bug Fixes**
  * Existing checkpoint artifacts are validated before resumed snapshots are marked successful.
  * Detects corrupted, incomplete, malformed, or unsafe checkpoint artifacts.
  * Restore failures now clean up loop devices reliably.

### DIFF
--- a/deploy/helm/charts/snapshot/templates/daemonset.yaml
+++ b/deploy/helm/charts/snapshot/templates/daemonset.yaml
@@ -149,6 +149,9 @@ spec:
             # Mount host cgroup for CRIU (write access needed for cgroup freezer)
             - name: host-cgroup
               mountPath: /sys/fs/cgroup
+            # The agent allocates loop numbers through the only host device it receives.
+            - name: host-loop-control
+              mountPath: /host/dev/loop-control
             # Kubelet PodResources API socket (for GPU UUID discovery)
             - name: kubelet-pod-resources
               mountPath: /var/lib/kubelet/pod-resources
@@ -214,6 +217,11 @@ spec:
           hostPath:
             path: /sys/fs/cgroup
             type: Directory
+        # Individual loop block nodes are created in the agent's ephemeral root.
+        - name: host-loop-control
+          hostPath:
+            path: /dev/loop-control
+            type: CharDevice
         # Kubelet PodResources API socket directory
         - name: kubelet-pod-resources
           hostPath:
--- a/deploy/snapshot/internal/runtime/overlay.go
+++ b/deploy/snapshot/internal/runtime/overlay.go
@@ -1,22 +1,41 @@
 package runtime
 
 import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
-	"github.com/go-logr/logr"
+	"golang.org/x/sys/unix"
 
 	"github.com/ai-dynamo/dynamo/deploy/snapshot/internal/types"
 )
 
 const (
-	rootfsDiffFilename   = "rootfs-diff.tar"
+	RootfsDiffFilename   = "rootfs-diff.squashfs"
 	deletedFilesFilename = "deleted-files.json"
+	// Bounds untrusted metadata while allowing at least ~4,000 PATH_MAX-sized entries and far more normal paths.
+	maxDeletedFilesSidecarSize = 16 << 20
 )
+
+type deletedFiles struct {
+	Whiteouts       []string `json:"whiteouts,omitempty"`
+	OpaqueDirectory []string `json:"opaqueDirectories,omitempty"`
+}
+
+func PreflightRootfsCapture() error {
+	if _, err := exec.LookPath("mksquashfs"); err != nil {
+		return fmt.Errorf("mksquashfs is required: %w", err)
+	}
+	return preflightRootfsMountCapability()
+}
 
 // GetRootFS returns the container's root filesystem path via /host/proc.
 func GetRootFS(pid int) (string, error) {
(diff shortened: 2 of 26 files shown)


deploy/snapshot/internal/types/manifest.go
```diff
@@ -179,6 +192,19 @@ func ReadManifest(checkpointDir string) (*CheckpointManifest, error) {
 	if strings.TrimSpace(data.CheckpointID) == "" {
 		return nil, fmt.Errorf("checkpoint manifest is missing checkpointId")
 	}
+	if err := validateRootFSSHA256(data.RootFSSHA256); err != nil {
+		return nil, err
+	}
 
 	return &data, nil
```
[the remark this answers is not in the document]
galletas1712: Confirmed; this incompatibility is intentional. The artifact changes from tar to mandatory SquashFS, and this PR deliberately does not add a tar fallback or schema compatibility layer. Legacy tar checkpoints must be recreated after upgrading. Nonterminal resume rejects them as `InvalidCheckpointArtifact` (covered by `TestReconcileSnapshotContent_ResumeRejectsTarOnlyArtifact`), while previously Ready artifacts are rejected when restore validates the old manifest.

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -33,184 +49,382 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
 	}
-	tarArgs = append(tarArgs, "-C", upperDir, "-cf", rootfsDiffPath, ".")
 
-	cmd := exec.Command("tar", tarArgs...)
-	output, err := cmd.CombinedOutput()
+	rootfsPath := filepath.Join(checkpointDir, RootfsDiffFilename)
+	if err := runMksquashfs(
+		upperDir,
+		rootfsPath,
+		buildExclusions(exclusions, bindMountDests, overlayEntries),
```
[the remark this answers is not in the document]
galletas1712: Fixed in `8678ad9783`. The existing lease-backed checkpoint context now flows through `CaptureRootfsDiff` and `runMksquashfs`, which uses `exec.CommandContext`. A focused fake-binary test verifies cancellation returns an error and removes the partial output. No process-group machinery was added because the current mksquashfs invocation uses in-process workers and no pseudo-file command helpers.

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -33,184 +49,382 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
+			result = append(result, exclusion, "... "+exclusion)
+		} else {
+			result = append(result, exclusion)
+		}
+	}
+	for _, path := range append(bindMountDests, overlayEntries...) {
+		path = strings.TrimPrefix(filepath.Clean(path), "/")
+		if path != "" && path != "." {
+			result = append(result, path)
+		}
```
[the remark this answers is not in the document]
galletas1712: Fixed in `8678ad9783`. Configured exclusions remain wildcard patterns, while bind-mount destinations and scanned overlay entries are treated as literal paths by escaping each non-separator byte before passing them to mksquashfs under `-wildcards`. The integration test verifies that literal `literal[ab]` is excluded while the wildcard decoy `literala` remains.

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -33,184 +49,382 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
 
-	whiteouts, err := findWhiteoutFiles(upperDir)
+		name := info.Name()
+		switch {
+		case name == ".wh..wh..opq":
+			parent := filepath.Dir(relative)
+			if parent != "." {
+				result.OpaqueDirectory = append(result.OpaqueDirectory, filepath.ToSlash(parent))
+			}
```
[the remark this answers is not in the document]
galletas1712: Not changing this branch. Capture reads a live kernel OverlayFS upperdir, where opacity is represented by the overlay opaque xattr rather than an OCI `.wh..wh..opq` marker. Root opacity is already captured as `.` through the root-directory xattr and replay supports that representation. A workload can create an ordinary root file with this literal name, so interpreting it as root opacity could incorrectly hide the entire lower root.

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -33,184 +49,382 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
+	if _, err := file.Seek(0, io.SeekStart); err != nil {
+		return nil, fmt.Errorf("seek deletion sidecar: %w", err)
+	}
+	data, err := io.ReadAll(file)
+	if err != nil {
+		return nil, fmt.Errorf("read deletion sidecar: %w", err)
+	}
+	var result deletedFiles
+	if err := json.Unmarshal(data, &result); err != nil {
+		return nil, fmt.Errorf("decode deletion sidecar: %w", err)
```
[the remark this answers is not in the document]
galletas1712: Fixed in `8678ad9783`. The sidecar is opened with `O_NONBLOCK|O_NOFOLLOW`, required to be a regular file at both the path-open and inherited-FD boundaries, and bounded to 16 MiB on write, stat, and a `limit+1` read to cover growth after stat. Tests cover valid round-trip behavior, nonblocking FIFO/inherited-FD rejection, and oversized sparse files. The bound allows at least roughly 4,000 PATH_MAX-sized entries and far more normal paths while bounding untrusted allocation.

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -33,184 +49,382 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
-	})
+	}
+	return &result, nil
+}
 
-	return whiteouts, err
+func validateRelativeArtifactPath(path string) error {
+	if path == "" || filepath.IsAbs(path) || filepath.Clean(path) != path ||
+		path == "." || strings.HasPrefix(path, ".."+string(os.PathSeparator)) {
+		return errors.New("path must be clean and relative")
```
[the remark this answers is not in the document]
galletas1712: Fixed in `8678ad9783`. `validateRelativeArtifactPath` now rejects exact `..` in addition to `../...`, with a focused regression assertion. The existing FD-relative/openat2 containment remains the defense-in-depth layer during replay.

deploy/snapshot/internal/runtime/root_compose.go
```diff
@@ -0,0 +1,569 @@
...
+	if err := createWhiteout(filepath.Join(restoreUpper, strings.TrimPrefix(RootfsBackingWorkspaceMountPath, "/"))); err != nil {
+		return nil, fmt.Errorf("hide rootfs backing workspace: %w", err)
+	}
+	options := strings.Join([]string{
+		"lowerdir=" + restorePlaceholder + ":" + restoreSnapshotLower,
+		"upperdir=" + restoreUpper,
+		"workdir=" + restoreWork,
+	}, ",")
+	if err := unix.Mount("overlay", restoreRoot, "overlay", 0, options); err != nil {
+		return nil, fmt.Errorf("mount placeholder-first composed root: %w", err)
```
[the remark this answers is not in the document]
galletas1712: The leftmost-precedence observation is correct, but placeholder-first is intentional compatibility with the previous `tar --skip-old-files` restore: existing placeholder/base paths win, snapshot-only entries fill gaps, and replayed whiteouts/opaque directories preserve deletion semantics in the writable upper. `TestPrivilegedPlaceholderFirstRootComposition` explicitly asserts the placeholder collision result. Reversing the order would change the restore contract rather than fix the SquashFS conversion, so this PR keeps the current ordering.

deploy/snapshot/internal/runtime/root_compose.go
```diff
@@ -0,0 +1,406 @@
...
+		return fmt.Errorf("inspect placeholder opaque path %q: %w", relative, err)
+	}
+	if info, err := os.Lstat(filepath.Join(snapshotRoot, relative)); err == nil {
+		if !info.IsDir() {
+			return fmt.Errorf("snapshot opaque path %q is not a directory", relative)
+		}
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("inspect snapshot opaque path %q: %w", relative, err)
+	}
+	upperFD, err := mkdirAllAt(upperRootFD, relative)
```
dfeigin-nv: Does this function copy mode too?

deploy/snapshot/internal/runtime/root_compose.go
```diff
@@ -0,0 +1,406 @@
...
+// supplied by the SquashFS layer below it.
+func replayOpaqueDirectory(
+	upperRootFD int,
+	relative, placeholderRoot, snapshotRoot string,
+) error {
+	if relative == "." {
+		relative = ""
+	}
+	if info, err := os.Lstat(filepath.Join(placeholderRoot, relative)); err == nil {
+		if !info.IsDir() {
```
dfeigin-nv: This return means we see the placeholder file instead of the checkpointed file, no?

deploy/snapshot/internal/runtime/root_compose.go
```diff
@@ -0,0 +1,406 @@
...
+
+	placeholderTarget, err := unix.Open(
+		restorePlaceholder,
+		unix.O_PATH|unix.O_DIRECTORY|unix.O_CLOEXEC|unix.O_NOFOLLOW,
+		0,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("open placeholder mountpoint: %w", err)
+	}
+	if err := moveMountToFD(placeholderFD, placeholderTarget); err != nil {
```
dfeigin-nv, line 132: If the restore fails, the mount will exist and no one will clean it up, right?

deploy/helm/charts/snapshot/templates/daemonset.yaml
```diff
@@ -149,6 +149,9 @@ spec:
             # Mount host cgroup for CRIU (write access needed for cgroup freezer)
             - name: host-cgroup
               mountPath: /sys/fs/cgroup
+            # The agent allocates loop numbers through the only host device it receives.
+            - name: host-loop-control
+              mountPath: /host/dev/loop-control
```
dfeigin-nv, line 154: Is this always available? openshift too?

deploy/snapshot/internal/runtime/squashfs_mount_test.go
```diff
@@ -0,0 +1,78 @@
+package runtime
+
+import (
+	"errors"
```
dfeigin-nv, line 4: Let's test a subset of these tests too?

| Operation before checkpoint | Expected after restore |
| :--- | :--- |
| Modify an existing image file | Modified contents |
| Add a new file | File exists |
| Delete an image file | File absent |
| Delete and recreate a file | Recreated contents/type |
| Rename a directory | Correct tree at new path |
| Make a directory opaque | Lower children remain hidden |
| Change mode/owner | Metadata preserved |
| Set file capabilities/xattrs | Capabilities/xattrs preserved |
| Create symlinks/hardlinks | Link behavior preserved |
| Modify a file beneath a Kubernetes volume mount | Volume contents win appropriately |

deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -1,23 +1,42 @@
...
+	// Bounds untrusted metadata while allowing at least ~4,000 PATH_MAX-sized entries and far more normal paths.
+	maxDeletedFilesSidecarSize = 16 << 20
 )
 
+type deletedFiles struct {
+	Whiteouts       []string `json:"whiteouts,omitempty"`
+	OpaqueDirectory []string `json:"opaqueDirectories,omitempty"`
+}
+
+func PreflightRootfsCapture() error {
```
dfeigin-nv, line 33: What is the earliest kernel you tested this on? We need to verify LOOP_CONFIGURE

deploy/snapshot/internal/runtime/squashfs_mount_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"testing"
```
dfeigin-nv, line 6: Please add a privileged failure/retry test that repeatedly runs concurrent rootfs mount operations and injects failures after loop configuration, after the SquashFS mount, after open_tree, and after root composition. After each case, verify that no loop device remains associated with the checkpoint image, no stale SquashFS/OverlayFS mounts remain, and a second restore in the same pod succeeds. Please also exercise contention beyond the eight allocation retries.

deploy/snapshot/internal/runtime/root_compose.go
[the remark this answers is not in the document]
galletas1712: Follow-up: after the human review clarified that a modified image-owned file must restore the checkpointed contents, I re-evaluated this. Preserving `tar --skip-old-files` here would preserve an old fidelity limitation rather than the intended snapshot contract. `aceb9908c9` now uses `lowerdir=snapshot:placeholder`, and the privileged test asserts that the snapshot collision wins.

deploy/snapshot/internal/runtime/root_compose.go
galletas1712: Yes. `copyDirectoryOwnerAndMode` now reads the winning lower directory with `Fstat`, applies UID/GID with `Fchown`, and then applies all permission and special bits with `Fchmod(mode & 07777)`. Replay selects snapshot metadata first and falls back to the placeholder for excluded/mount-backed paths. The privileged test checks UID, GID, and mode on both the composed root and a replay-created opaque directory.

deploy/snapshot/internal/runtime/root_compose.go
galletas1712: That was possible with the previous placeholder-first ordering. `aceb9908c9` changes the lower order so checkpoint/SquashFS content wins. For this early-return case, we now first verify that the snapshot path is a directory; only then can a placeholder file/symlink be ignored because there is no placeholder directory to traverse. Missing or malformed snapshot paths fail closed. The test covers an existing placeholder file modified by the snapshot and expects the snapshot contents.

deploy/snapshot/internal/runtime/root_compose.go
galletas1712: Correct; this was a lifecycle gap. `RootComposition` now owns the attached backing tree. `ComposeRoot` rolls it back if setup fails after attachment, and `executeRestore` defers `Close()` across CRIU and CUDA restore so success and downstream failure both lazy-unmount it. Cleanup errors are joined with the primary error. The dedicated upper/work workspace is also reset before every composition; the privileged test mutates the first composed root, closes it as a failed downstream restore would, retries in the same workspace, and verifies that the stale file is gone.

deploy/helm/charts/snapshot/templates/daemonset.yaml
galletas1712: It is not universally available, so the chart remains fail-closed: the `CharDevice` hostPath requires `/dev/loop-control` to exist. The node must have the loop driver loaded/built in. On OpenShift, the privileged SCC grants the agent permission to use the device and mount APIs, but it does not load the host driver or create `/dev/loop-control`. I documented those prerequisites and added a real `LOOP_CONFIGURE` capability preflight rather than treating device presence alone as sufficient.

deploy/snapshot/internal/runtime/squashfs_mount_test.go
galletas1712: Added a focused subset without turning this into a general filesystem conformance suite. The privileged test now covers: modified image file (snapshot wins), snapshot-only add, placeholder deletion, directory-to-file recreation, recursive opaque-directory hiding, UID/GID/mode preservation, a regular-file `user.*` xattr through SquashFS, and exclusion of both a directory child mount and a file bind mount so Kubernetes/CRIU can restore those mounts separately. Rename/redirect behavior remains explicitly unsupported by capture today, and native hardlink/symlink behavior was left out because it does not exercise Dynamo composition code.

deploy/snapshot/internal/runtime/overlay.go
galletas1712: The focused privileged validation for this revision ran on Linux `6.17.0-35-generic`; I do not want to claim an older personally tested kernel. Upstream introduced `open_tree`/`move_mount` in 5.2, `openat2` in 5.6, and `LOOP_CONFIGURE` in 5.8, making 5.8 the first upstream release with the full set. Enterprise kernels backport these independently, so the code now verifies `LOOP_CONFIGURE` by configuring and clearing a scratch-file loop device during preflight instead of relying only on `uname` or opening loop-control.

deploy/snapshot/internal/runtime/squashfs_mount_test.go
galletas1712: Addressed with focused lifecycle tests rather than a per-syscall production injection framework. Unprivileged deterministic tests cover: mount failure after loop configuration (exact loop cleared), `open_tree` failure after the SquashFS mount (staging unmounted and loop cleared), and joined primary/unmount/loop-clear errors. The privileged test holds nine real detached mounts live simultaneously, verifies no staging mount or UUID directory remains, then composes a root, mutates it, performs downstream-failure cleanup, and successfully composes again in the same pod workspace with the stale mutation absent. A separate retry test forces nine `EBUSY` results and succeeds on attempt ten. Allocation is now process-serialized, retries transient `EBUSY`/`ENOENT`/`ENXIO` up to the util-linux-sized 16-attempt budget, and delays retries to tolerate external-process contention.

PR #11719, opened 2026-07-15
feat(backends): handle conditional disagg bypass annotation

karen-sy:
> [!IMPORTANT]
> This PR is **[2/5]** in the conditional-disaggregation stack.
> Merge order is bottom-up: **#11718 merges to `main` first**, and **#11723 merges last**.
> Do not merge this PR until every earlier PR in the checklist has merged and this branch has been rebased/retargeted as needed.

Link to the DEP: [https://github.com/ai-dynamo/dynamo/issues/11514](https://github.com/ai-dynamo/dynamo/issues/11514)

Stack merge order:

- [x] #11718 — feat(router): add conditional disagg policy kernel
- [x] **#11719 — feat(backends): handle conditional disagg bypass annotation**
- [ ] #11720 — feat(router): add decode affinity scoring
- [ ] #11725 — feat(router): orchestrate conditional disagg bypass
- [ ] #11723 — feat(router): expose conditional disagg configuration

Reference to original (closed) combined PR, with addressed review comments: [ai-dynamo/dynamo#11357](https://github.com/ai-dynamo/dynamo/pull/11357)

## Related Issues

> ⚠️ **This section is required.** Choose one path below and delete the other.

**🔗 This PR is linked to an issue:**

- Closes #XXXX

**🚫 This PR is NOT linked to an issue**:
- [ ] Confirmed — no related issue

## Summary by CodeRabbit

- **New Features**
  - Added conditional remote-prefill bypass for eligible decode requests, allowing local prefill and generation when configured.
  - Added configurable bypass policies based on prompt size, prefill worker load, decode worker load, or a combination.
  - Added configuration options and validation for bypass thresholds and worker-busy conditions.
  - Improved decode-worker support for prefix caching and key-value cache event publishing.
- **Bug Fixes**
  - Unsupported bypass requests now fail early with a clear error message.
- **Tests**
  - Added coverage for bypass behavior, multimodal requests, configuration boundaries, and load-based decisions.

Issue #11514, opened 2026-07-10
DEP: Conditional Disaggregation


karen-sy:
## Summary

Motivation, design, and results so far on conditional disaggregation (PPD). 

Implementations are stacked here:
- [ ] #11718 — feat(router): add conditional disagg policy kernel
- [ ] #11719 — feat(backends): handle conditional disagg bypass annotation
- [ ] #11720 — feat(router): add decode affinity scoring
- [ ] #11725 — feat(router): orchestrate conditional disagg bypass
- [ ] #11723 — feat(router): expose conditional disagg configuration

## Motivation

Conditional disaggregation (C/D) is a feature that enables a hybrid of aggregated and disaggregated request routing. The router may serve a request `prefill worker -> decode worker`, or it may send the request directly to a decode worker and the backend runs local prefill plus decode there.

**For workloads with a high degree of KV reusage on long-ISL requests, e.g. multi-turn / agentic conversation scenarios, conditional disaggregation can help your deployment maintain predictable SLA.** Compared to unconditional disaggregation, it reduces memory pressure / TTFT on prefill workers by optimizing reuse of *decode-worker* KV cache; compared to unconditionally aggregated deployments, it avoids the heavy ITL penalty incurred by co-scheduling heavy prefill workload onto decode workers.

### Design
C/D hinges on a "bypass" decision -- whether an input should "bypass" a Prefill worker to perform both ctx+gen on a Decode worker. We propose that this bypass decision is primarily owned by the router, which has a view of per-worker load and kv block ownership state. The decision is parametrized as a `ConditionalDisaggPolicy`, each implementation of which defines a bypass threshold based on input sequence characteristics and/or worker state. 

Concretely, we can start with some static `ConditionalDisaggPolicy` variants:
- `ISL-based`: Bypass iff `ISL - kv cache hit < const threshold AND (effective ISL / ISL) < const threshold`
- `Load-based`: Bypass iff `prefill worker loadedness > const threshold AND decode worker loadedness < const threshold`
- `ISL-and-Load`: Bypass iff `ISL-based` OR `Load-based` is true

All of the above three options are implemented across the PR stack listed in the Summary.

We can extend these to dynamic policies based on live signals relative to SLA, for example:
- `ISL-based-dynamic`: Bypass iff `ISL - kv cache hit < f(threshold) AND ISL < g(threshold)` where `f` and `g` are additionally parameterized by current load signals and user-defined SLA
- `Load-based-dynamic`
[...]

PR #11357, opened 2026-07-07
feat(router): conditional disaggregation

karen-sy:
## Overview:

**For motivation/design/prelim results, see DEP here: https://github.com/ai-dynamo/dynamo/issues/11514**

Adds conditional disaggregation support to the KV router so selected requests can bypass the remote prefill worker and run prefill+decode locally on the chosen decode worker when the effective ISL is small enough or when the configured load policy says prefill is busy. Enables decode KV-affinity routing for conditional-disagg mode.
Supports vLLM and TRTLLM. 

### Summary

- Adds conditional-disagg router config, CLI flags, Python bindings, and env parsing.
- Adds conditional-disagg policies:
  - `isl_bounding`
  - `prefill_load`
  - `isl_or_load`
- Adds prefill-side and decode-side busy checks for gating bypass decisions.
- Adds decode-side circuit breaker support via `--router-conditional-disagg-decode-busy-threshold`.
- Adds decode-overlap affinity only when conditional-disagg is enabled.
- Annotates bypassed requests with `x-bypass-remote-prefill`.
- Wires vLLM decode workers to treat bypassed requests as AGG/local-prefill style requests.
- Wires TRT-LLM decode workers to treat bypassed requests as `context_and_generation`.
- Adds targeted unit coverage for router policy, decode gate behavior, vLLM bypass handling, and TRT-LLM bypass handling.

### Validation

- `cargo fmt --all`
- `cargo test -p dynamo-kv-router conditional_disagg`
- `cargo test -p dynamo-llm prefill_router`
- `cargo check` from `lib/bindings/python`
- `python -m py_compile` on modified Python handler/config/test files

## Where should the reviewer start?
- `docs/components/router`
  - docs update on relevant flags + added conditional disagg tuning guide

- `lib/kv-router/src/conditional_disagg.rs`
  - New conditional-disagg policy definitions and unit tests.

- `lib/llm/src/kv_router/prefill_router/conditional_bypass.rs`
  - Main bypass decision flow: decode worker selection, effective ISL calculation, prefill busy gate, and decode busy gate.

- `lib/llm/src/kv_router/prefill_router/mod.rs`
  - Integration point in the prefill router and decode router override behavior.

- `lib/kv-router/src/scheduling/selector.rs`
  - Decode-overlap affinity scoring path; normal disagg remains load-only.

- `components/src/dynamo/vllm/handlers.py`
  - vLLM bypass annotation handling.

- `components/src/dynamo/trtllm/request_handlers/handler_base.py`
[...]

PR #11723, opened 2026-07-15
feat(router): expose conditional disagg configuration

karen-sy:
> [!IMPORTANT]
> This PR is **[5/5]** in the conditional-disaggregation stack.
> Merge order is bottom-up: **#11718 merges to `main` first**, and **#11723 merges last**.
> Do not merge this PR until every earlier PR in the checklist has merged and this branch has been rebased/retargeted as needed.

Link to the DEP: [https://github.com/ai-dynamo/dynamo/issues/11514](https://github.com/ai-dynamo/dynamo/issues/11514)

Stack merge order:

- [x] #11719 — feat(backends): handle conditional disagg bypass annotation
- [x] #11720 — feat(router): add decode affinity scoring
- [x] #11718 — feat(router): add conditional disagg policy kernel
- [x] #11725 — feat(router): orchestrate conditional disagg bypass
- [x] **#11723 — feat(router): expose conditional disagg configuration**

Reference to original (closed) combined PR, with addressed review comments: [ai-dynamo/dynamo#11357](https://github.com/ai-dynamo/dynamo/pull/11357)

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

- **New Features**
  - Added experimental conditional disaggregation for KV-aware routing.
  - Supports configurable routing policies, sequence-length thresholds, and prefill/decode load thresholds.
  - Added Python configuration support with documented defaults.

- **Bug Fixes**
  - Added validation for unsupported settings, malformed configuration, invalid values, and required KV-event options.

- **Documentation**
  - Documented setup, supported policies, backend requirements, tuning guidance, and configuration options.
  - Marked TensorRT-LLM conditional disaggregation as supported.


MatejKosec (commented):
Two small logging consistency issues (details in inline comments).

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -105,6 +121,38 @@ def _default_prefill_load_scale() -> float:
...
+def warn_conditional_disagg_prefill_busy_threshold_resolution(
+    *,
+    policy: str,
+    busy_threshold: Optional[float],
+    queue_threshold: Optional[float],
+) -> None:
+    if policy not in LOAD_AWARE_CONDITIONAL_DISAGG_POLICIES:
+        return
+    if busy_threshold is not None:
+        logging.getLogger(__name__).info(
```
MatejKosec: Every other file in `components/src/dynamo/common/configuration/groups/` (and the broader codebase) declares a module-level logger and uses it throughout — for example `http_args.py` in this same directory does `logger = logging.getLogger(__name__)` right after imports, then calls `logger.warning(...)`. This file adds `import logging` but skips that step and instead calls `logging.getLogger(__name__)` inline on each log call site (here and at line 139). The fix is to add `logger = logging.getLogger(__name__)` at module level after the imports, then replace both `logging.getLogger(__name__).info(...)` calls with `logger.info(...)`.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -105,6 +121,38 @@ def _default_prefill_load_scale() -> float:
...
+        )
+    elif queue_threshold is not None:
+        logging.getLogger(__name__).info(
+            "conditional_disagg prefill-load condition using "
+            "--router-queue-threshold=%s because "
+            "--router-conditional-disagg-prefill-busy-threshold is unset",
+            queue_threshold,
+        )
+    else:
+        warnings.warn(
```
MatejKosec: Every other `warnings.warn` call in the repo is a deprecation notice — each one passes `DeprecationWarning` or `FutureWarning` to signal that a user is relying on something being removed. This call is different: it is a runtime misconfiguration warning, not a deprecation. The consistent choice here is `logger.warning(...)` (using the module-level logger mentioned in the comment above). That also unifies the three branches of this function under one notification mechanism instead of mixing `logging.info` for the two success paths and `warnings.warn` for the error path.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            ),
+            dest="conditional_disagg_enabled",
+        )
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-policy",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_POLICY",
+            default="isl_bounding",
+            help=(
+                "KV Router: Conditional-disagg bypass policy. "
```
MatejKosec: The gate flag `--router-conditional-disagg` at line 481 is correctly tagged `[EXPERIMENTAL]`, but all five sub-flags introduced here are missing it. The precedent in this same file is that each flag in an experimental group carries its own tag — `--use-remote-indexer`, `--shared-cache-multiplier`, and `--shared-cache-type` (lines 588–612) are all tagged individually even though they belong to the same experimental feature cluster. A user who finds one of these threshold flags from documentation or a colleague and runs `--help | grep router-conditional-disagg-policy` would see no experimental warning on the flag itself. The same applies to `--router-conditional-disagg-eff-isl-threshold`, `--router-conditional-disagg-eff-isl-ratio-threshold`, `--router-conditional-disagg-prefill-busy-threshold`, and `--router-conditional-disagg-decode-busy-threshold`.

MatejKosec (commented):
reviewed with comments

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+        add_negatable_bool_argument(
+            g,
+            flag_name="--router-conditional-disagg",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG",
+            default=False,
+            help=(
+                "[EXPERIMENTAL] KV Router: Enable conditional disaggregation. "
+                "When enabled, the frontend may skip remote prefill and route "
+                "a request directly to the cache-hot decode worker when the "
+                "configured policy decides local prefill+decode is preferable."
```
MatejKosec: Two prose issues in this help string.

First, the sentence has two "when" clauses: "When enabled, … when the configured policy decides". The second "when" arrives after a long subordinate clause, making it a garden-path sentence. Splitting works: "When enabled, the router may bypass the remote prefill step and route requests directly to a decode worker. The configured policy determines whether local prefill and decode is preferable for each request."

Second, `prefill+decode` uses `+` as a prose connector. No other help string in this file does this — they write ranges and relationships in words. "prefill and decode" is the plain form.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            g,
+            flag_name="--router-conditional-disagg-policy",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_POLICY",
+            default="isl_bounding",
+            help=(
+                "KV Router: Conditional-disagg bypass policy. "
+                "'isl_bounding': bypass when effective ISL is below both the "
+                "absolute and ratio thresholds. 'prefill_load': bypass when "
+                "the chosen prefill worker is busy. 'isl_or_load': bypass "
+                "when either policy would bypass."
```
MatejKosec: `'isl_or_load': bypass when either policy would bypass.` is self-referential — "either policy" only resolves if the reader already knows this is the union of `isl_bounding` and `prefill_load`. The docs policy table below this flag table writes it explicitly: "Either the `isl_bounding` condition OR the `prefill_load` condition is true." The help string should match that level of explicitness: `'isl_or_load': bypass when either the ISL condition or the prefill-load condition is met.`

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            choices=list(CONDITIONAL_DISAGG_POLICY_CHOICES),
+            dest="conditional_disagg_policy",
+        )
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-eff-isl-threshold",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_EFF_ISL_THRESHOLD",
+            default=2048,
+            help=(
+                "KV Router: For 'isl_bounding' and the ISL arm of "
```
MatejKosec: "the ISL arm of `isl_or_load`" — "arm" is undefined jargon. It does not appear in the docs prose, and the policy table (which describes `isl_or_load` as "either the `isl_bounding` condition OR the `prefill_load` condition") never uses the word. "the ISL condition within `isl_or_load`" says the same thing in terms the reader already has from the policy descriptions. The same phrasing recurs at line 522 for the ratio-threshold flag.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            arg_type=int,
+            dest="conditional_disagg_eff_isl_threshold",
+        )
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-eff-isl-ratio-threshold",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_EFF_ISL_RATIO_THRESHOLD",
+            default=0.7,
+            help=(
+                "KV Router: For 'isl_bounding' and the ISL arm of "
```
MatejKosec: `effective ISL / raw ISL` uses `/` as a division operator in prose. No other help string in this file uses mathematical notation this way — ranges are written as "Range: 0.0 to 1.0" and fractions in words. "the ratio of effective ISL to raw ISL" is the prose equivalent. (The docs flag table at line 146 of `router-disaggregated-serving.md` carries this verbatim from here, so both need updating together.)

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            arg_type=float,
+            dest="conditional_disagg_eff_isl_ratio_threshold",
+        )
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-prefill-busy-threshold",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_PREFILL_BUSY_THRESHOLD",
+            default=None,
+            help=(
+                "KV Router: Dedicated prefill-worker busy-line fraction for "
```
MatejKosec: "Dedicated prefill-worker busy-line fraction" — "busy-line" does not appear anywhere in the docs and does not match the flag's own name (`prefill_busy_threshold`). The docs flag table calls this "the prefill busy threshold". "Prefill-worker busy threshold for `prefill_load` and `isl_or_load`." aligns with that and with the flag name. The word "Dedicated" is also doing unclear work; what it signals — that this flag takes priority over the `--router-queue-threshold` fallback — is already explained by the next sentence.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -418,6 +472,88 @@ def add_arguments(self, parser) -> None:
...
+            arg_type=float,
+            dest="conditional_disagg_prefill_busy_threshold",
+        )
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-decode-busy-threshold",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_DECODE_BUSY_THRESHOLD",
+            default=None,
+            help=(
+                "KV Router: Decode-busy guard for conditional disagg. "
```
MatejKosec: "the selected decode worker's projected decode load" — "projected" implies estimation or prediction, but what is being projected (current KV occupancy? queued-request load? inferred future utilization?) is not defined here or in the docs flag table at line 148 of `router-disaggregated-serving.md`, which carries the same phrase verbatim. An operator tuning this threshold needs to know which metric is actually being compared against `threshold * total_kv_blocks(W)`.

docs/components/router/router-disaggregated-serving.md
```diff
@@ -96,3 +96,61 @@ graph TD
...
+
+| Flag                                                  | Default        | Use                                                                                                                                               |
+| ----------------------------------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `--router-conditional-disagg`                         | Disabled       | Enables conditional disaggregation. Requires `--router-mode kv` and separate prefill/decode worker pools.                                         |
+| `--router-conditional-disagg-policy`                  | `isl_bounding` | Selects the bypass policy: `isl_bounding`, `prefill_load`, or `isl_or_load`.                                                                      |
+| `--router-conditional-disagg-eff-isl-threshold`       | `2048`         | For `isl_bounding` and the ISL arm of `isl_or_load`, require effective ISL to be below this many tokens.                                          |
```
MatejKosec: Three issues in this flag table that need attention together.

**Lines 145–146, "the ISL arm of `isl_or_load`":** "arm" is not defined anywhere in this document. The policy table below uses "either the `isl_bounding` condition OR the `prefill_load` condition" without introducing the word. "the ISL condition within `isl_or_load`" is consistent with that phrasing and needs no prior knowledge of what an "arm" is.

**Line 146, `` `effective ISL / raw ISL` ``:** wrapping the slash expression in backticks makes it look like code syntax rather than a prose ratio. The tuning guidance at line 139 already writes "effective ISL / raw ISL ratio" in plain text, which is better; the table cell should match that or go further to "the ratio of effective ISL to raw ISL".

**Line 148, "projected decode load":** "projected" implies estimation or prediction but is never defined anywhere in this document. If this is current KV block occupancy, write that; if it is inferred from the pending-request queue, say so. An operator cannot tune this threshold without knowing which metric is being compared against the KV capacity fraction.

docs/components/router/router-disaggregated-serving.md
```diff
@@ -96,3 +96,61 @@ graph TD
...
+
+### Backend Requirements
+
+Conditional disaggregation requires decode-worker KV visibility. The router uses decode-side KV events to estimate effective ISL and decide whether local prefill+decode is cheaper than remote prefill.
+
+Configure workers as follows:
+
+| Backend | Requirement |
+| --- | --- |
+| vLLM | On decode workers, pass `--enable-conditional-disagg`. This opts decode workers into KV event publication and bypass handling. |
```
tedzhouhk: Passing `--enable-conditional-disagg` alone does not enable vLLM KV event publication. `create_kv_events_config()` returns the existing `engine_config.kv_events_config` or `None`; this flag only removes the decode-worker suppression. Following this table with the default `isl_bounding` policy therefore leaves the router with no decode overlap and the bypass never fires. Please also require an enabled `--kv-events-config` (with `enable_kv_cache_events: true`), or make the flag create one.

docs/components/router/router-disaggregated-serving.md
```diff
@@ -96,3 +96,61 @@ graph TD
...
+| `--router-conditional-disagg-eff-isl-threshold`       | `2048`         | For `isl_bounding` and the ISL condition within `isl_or_load`, require effective ISL to be below this many tokens.                                |
+| `--router-conditional-disagg-eff-isl-ratio-threshold` | `0.7`          | For `isl_bounding` and the ISL condition within `isl_or_load`, require the `effective ISL : raw ISL` ratio to be below this value. Must be in `[0.0, 1.0]`. |
+| `--router-conditional-disagg-prefill-busy-threshold`  | Unset          | Sets the prefill busy threshold for `prefill_load` and `isl_or_load`. When unset, those policies inherit `--router-queue-threshold` if it is set. |
+| `--router-conditional-disagg-decode-busy-threshold`   | Unset          | Decode-busy guard. When unset, the guard is disabled. When set, conditional-disagg bypass is disabled if the selected decode worker's projected active decode KV blocks exceed this fraction of KV capacity. This uses router-side active decode block accounting. |
```
tedzhouhk: This guard fails closed when the router cannot resolve `total_kv_blocks`, not only when projected load exceeds the threshold. TensorRT-LLM currently registers without `total_kv_blocks`, so setting this documented knob disables every TensorRT-LLM bypass. Please either publish the capacity or reject that unsupported combination; at minimum, document the capacity prerequisite and fail-closed behavior.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -379,6 +435,90 @@ def add_arguments(self, parser) -> None:
             ),
             arg_type=str,
         )
+        add_negatable_bool_argument(
+            g,
+            flag_name="--router-conditional-disagg",
```
tedzhouhk, line 556: These options live in the shared `KvRouterArgGroup`, which is also installed by `python -m dynamo.router`. The standalone router accepts and forwards `--router-conditional-disagg`, but conditional bypass is only executed by the frontend's `PrefillRouter`; the standalone service's plain `KvRouter` never consumes it. This is a silent no-op, even though native validation may log it as enabled. Please scope these options to the frontend or reject them in `DynamoRouterConfig.validate()`.

tedzhouhk (commented):
Four behavior/documentation mismatches found while reviewing the public conditional-disaggregation configuration.

docs/components/router/router-disaggregated-serving.md
```diff
@@ -96,3 +96,61 @@ graph TD
...
+
+Conditional disaggregation requires decode-worker KV visibility. The router uses decode-side KV events to estimate effective ISL and decide whether local prefill+decode is cheaper than remote prefill.
+
+Configure workers as follows:
+
+| Backend | Requirement |
+| --- | --- |
+| vLLM | On decode workers, pass `--enable-conditional-disagg`. This opts decode workers into KV event publication and bypass handling. |
+| TensorRT-LLM | Pass `--publish-kv-events` on prefill AND decode workers to opt them into KV-aware routing. |
+| SGLang | Not supported yet. Dynamo rejects SGLang conditional-disagg setups until the backend can run bypassed decode requests as local prefill+decode. |
```
tedzhouhk: This is not a setup-time rejection for the documented frontend flag. The frontend cannot validate the remote backend type: an SGLang deployment starts successfully, and a request marked with `x-bypass-remote-prefill` later raises `RuntimeError` in the decode handler. Please describe this as unsupported/request-time failure, or add actual fail-fast validation.

MatejKosec (dismissed):
All the logging, `[EXPERIMENTAL]` tag, and prose comments are addressed — good fixes throughout. Two small residual nits that are not blocking:

1. ```effective ISL : raw ISL``` ratio — the colon replaced the slash but is still inside backticks, making it look like a code literal. "the ratio of effective ISL to raw ISL" would be unambiguous prose, but this is a polish item.

2. "projected active decode KV blocks" — the docs addition ("This uses router-side active decode block accounting") names the mechanism, but "projected" still implies estimation without saying whether the count reflects current committed state or in-flight requests. Worth clarifying in a follow-up if anyone asks.

MatejKosec (changes requested):
I think we should revisit the overall number of flags needed to get the feature to be turned on and working correctly. It seems too easy to forget to activate one of the flags and then not even get warnings when something isn't configured correctly. 

To get a single bypass to fire with the default `isl_bounding` policy on vLLM, four things must all be true at once: `--router-mode kv` and `--router-conditional-disagg` on the frontend, `--enable-conditional-disagg` on every decode worker, and a KV events config with `enable_kv_cache_events: true` on every decode worker. That last point is tedzhouhk's comment — `--enable-conditional-disagg` only removes a suppression flag inside vLLM; the event publication pipeline is a separate config the worker flag does not touch. If any one of these is absent, the router evaluates the bypass condition on every eligible request, decides no, and routes normally. No log line fires. Nothing tells the operator the feature is running but doing nothing. This is the worst kind of misconfiguration: the deployment looks healthy and the feature silently does nothing.

**Some suggestions: **

1. When `--router-conditional-disagg` is set and a decode worker registers without KV event capability, the frontend should log a warning — something like: "conditional disagg is enabled but worker X is not publishing KV events; bypass will never fire for requests routed to this worker." The worker already advertises this on registration, so no new protocol is needed.

2. `DynamoRouterConfig.validate()` (or equivalent on the standalone router path) should reject `--router-conditional-disagg` with a clear error. Because these flags live in the shared `KvRouterArgGroup`, `python -m dynamo.router` currently parses and stores them, logs them as enabled, and then silently ignores them — the bypass logic only executes inside the frontend's `PrefillRouter`, which the standalone router never instantiates.

3. The validator in `FrontendArgs.validate()` should treat `prefill_load` or `isl_or_load` with no busy threshold as a hard error, not a warning. The `logger.warning` that was added is an improvement, but an operator who misses it ends up running the prefill-load policy with its condition permanently false for the lifetime of the process. A policy that can never fire should be rejected at startup, not warned about once.

Hope that helps.

alec-flowers:
I agree with @MatejKosec comment. Its very hard to configure properly and we need to make sure people/agents aren't turning things on willy nilly and its actually not working.

karen-sy:
thanks all for the reviews! will work on simplifying+clarifying the knobs

docs/fern/pages/developer-guide/knowledge-base/modular-components/router/disaggregated-serving.md
```diff
@@ -0,0 +1,162 @@
...
+
+Configure workers as follows:
+
+| Backend | Requirement |
+| --- | --- |
+| vLLM | Pass `--kv-events-config` on prefill and decode workers. |
+| TensorRT-LLM | Pass `--publish-kv-events` on prefill AND decode workers to opt them into KV-aware routing. |
+| SGLang | Not supported yet. |
+
+If decode workers do not publish KV events, the router cannot accurately assess bypass conditions.
```
MatejKosec: Suppose one decode worker starts without its KV-event configuration. The frontend still discovers that worker, and normal routing still sends requests to it. The worker serves those requests successfully.

The router, however, receives no cache events from that worker. It therefore sees no cached prefix for requests considered against that worker. With the default `isl_bounding` policy, the effective ISL then remains equal to the full input length, and the effective-to-raw ISL ratio remains `1.0`. The default ratio threshold is `0.7`, so the bypass check fails every time.

Nothing reports an error because the request still succeeds through normal remote prefill. Someone watching the deployment only sees that conditional disaggregation never uses the bypass path. The logs do not tell them whether their traffic failed the policy or whether the decode worker never supplied the cache information that the policy needs.

When conditional disaggregation is on and a decode worker registers without KV-event publication, could the frontend log one warning? For example: “Worker X does not publish KV events. Conditional disaggregation cannot bypass remote prefill for requests evaluated against this worker.” The frontend can continue using the worker for ordinary traffic; the warning only needs to explain why bypass will not work.

docs/fern/pages/developer-guide/knowledge-base/modular-components/router/disaggregated-serving.md
```diff
@@ -96,3 +96,67 @@ graph TD
...
+
+For workloads with a high degree of KV reusage on long-ISL requests, e.g. multi-turn / agentic conversation scenarios, conditional disaggregation can help your deployment maintain predictable SLA. Compared to unconditional disaggregation, it reduces memory pressure / TTFT on prefill workers by optimizing reuse of *decode-worker* KV cache; compared to unconditionally aggregated deployments, it avoids the heavy ITL penalty incurred by co-scheduling heavy prefill workload onto decode workers.
+
+Enable conditional disaggregation with `--router-conditional-disagg` on the frontend:
+
+```bash
+python -m dynamo.frontend \
+    --router-mode kv \
+    --router-conditional-disagg \
+    --router-conditional-disagg-config '{"policy":"isl_bounding"}'
```
MatejKosec: Could we show two examples here?

First, show the minimal setup:

```bash
python -m dynamo.frontend \
    --router-mode kv \
    --router-conditional-disagg
```

This tells the reader exactly what they need to turn the feature on with the default `isl_bounding` policy.

Then add a second example which introduces `--router-conditional-disagg-config` and changes something meaningful, such as the policy or one of its thresholds. That makes the distinction clear: the first example enables conditional disaggregation, while the second shows how to tune it.

The current example combines both steps and sets `isl_bounding`, which already serves as the default. A reader cannot tell that the JSON configuration is optional. Two examples would explain the interface more clearly than one example containing an argument that changes nothing.

docs/fern/pages/developer-guide/knowledge-base/modular-components/router/disaggregated-serving.md
karen-sy: warning should be covered by #12095

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -52,9 +56,32 @@
...
+)
+
+_CONDITIONAL_DISAGG_CONFIG_FIELDS: dict[str, str] = {
+    "policy": "conditional_disagg_policy",
+    "eff_isl_threshold": "conditional_disagg_eff_isl_threshold",
+    "eff_isl_ratio_threshold": "conditional_disagg_eff_isl_ratio_threshold",
+    "prefill_busy_threshold": "conditional_disagg_prefill_busy_threshold",
+    "decode_busy_threshold": "conditional_disagg_decode_busy_threshold",
+}
+
```
MatejKosec, line 86: human here: Do we want to maintain this mapping? If they all follow a prefix pattern, maybe there's a way to avoid this?

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -135,6 +203,13 @@ class KvRouterConfigBase(ConfigBase):
     serve_indexer: bool = False
     shared_cache_multiplier: float = 0.0
     shared_cache_type: str = "none"
+    conditional_disagg_enabled: bool = False
+    conditional_disagg_config: Optional[str] = None
+    conditional_disagg_policy: str = "isl_bounding"
+    conditional_disagg_eff_isl_threshold: int = 2048
+    conditional_disagg_eff_isl_ratio_threshold: float = 0.7
+    conditional_disagg_prefill_busy_threshold: Optional[float] = None
+    conditional_disagg_decode_busy_threshold: Optional[float] = None
```
MatejKosec, line 227: I see, so at the user level we expose a conditional dosage config json, but internally it's all serialized into the existing KvRouter config. That seems to be the pattern with other enhancements to kv-router as well.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -423,6 +528,34 @@ def add_arguments(self, parser) -> None:
...
+        add_argument(
+            g,
+            flag_name="--router-conditional-disagg-config",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_CONFIG",
+            default=None,
+            help=(
+                "[EXPERIMENTAL] KV Router: JSON object for conditional "
+                "disaggregation policy settings. Supported fields: policy, "
+                "eff_isl_threshold, eff_isl_ratio_threshold, "
+                "prefill_busy_threshold, decode_busy_threshold."
```
MatejKosec, line 577: After this consolidation, three messages in Rust `validate_kv_router_config` still tell users to set `--router-conditional-disagg-prefill-busy-threshold`, but that option no longer exists. Could we update those messages to point to `prefill_busy_threshold` inside `--router-conditional-disagg-config`, or to `--router-queue-threshold` where appropriate?

components/src/dynamo/frontend/frontend_args.py
```diff
@@ -158,6 +160,25 @@ def validate(self) -> None:
...
+            raise ValueError(
+                "--router-conditional-disagg-config eff_isl_threshold must be >= 0"
+            )
+        if not 0.0 <= self.conditional_disagg_eff_isl_ratio_threshold <= 1.0:
+            raise ValueError(
+                "--router-conditional-disagg-config eff_isl_ratio_threshold must be in [0.0, 1.0]"
+            )
+        if self.conditional_disagg_enabled and self.router_mode != "kv":
+            raise ValueError("--router-conditional-disagg requires --router-mode=kv")
+        if self.conditional_disagg_enabled and not self.use_kv_events:
```
MatejKosec, line 200: Could `KvRouterConfig::validate` enforce the same rules as the frontend? This check rejects conditional disaggregation with KV events off, and `apply_conditional_disagg_config()` rejects a load-based policy with no prefill or queue threshold. Direct `KvRouterConfig(...)`, `KvRouterConfig.from_json(...)`, and Rust serde construction currently accept both configurations. With no threshold, `prefill_load` never bypasses, while `isl_or_load` quietly loses its load-based condition. The entry point should not change which configurations the project accepts.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -145,9 +220,39 @@ def apply_load_aware_preset(self) -> None:
         for field, value in _LOAD_AWARE_KWARG_OVERRIDES.items():
             setattr(self, field, value)
 
+    def apply_conditional_disagg_config(self) -> None:
+        if self.conditional_disagg_config is not None:
+            parsed = _parse_conditional_disagg_config(self.conditional_disagg_config)
+            for key, value in parsed.items():
+                setattr(self, _CONDITIONAL_DISAGG_CONFIG_FIELDS[key], value)
+
+        if not self.conditional_disagg_enabled:
```
MatejKosec, line 251: A user can pass `--router-conditional-disagg-config` without `--router-conditional-disagg`. We parse and apply the JSON above, then return here, and the frontend accepts the configuration while the feature remains off. Could we reject the configuration argument unless the enable switch is present? The standalone router already rejects this combination. A direct error here would prevent someone from supplying tuning values which never take effect.

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -135,6 +203,13 @@ class KvRouterConfigBase(ConfigBase):
     serve_indexer: bool = False
     shared_cache_multiplier: float = 0.0
     shared_cache_type: str = "none"
+    conditional_disagg_enabled: bool = False
+    conditional_disagg_config: Optional[str] = None
+    conditional_disagg_policy: str = "isl_bounding"
```
MatejKosec, line 223: These values also appear in the Rust `Default` and serde path, the PyO3 constructor signature, the Python type stub, and the documentation. In the frontend path, `kv_router_kwargs()` passes every value into the binding, so these Python literals take precedence over the binding defaults. If Rust changes `2048` or `0.7` without the matching Python edit, CLI behavior will keep the old value while direct Rust and JSON construction will use the new one. Could we give these defaults one owner, or at least add a parity test which compares this class with `KvRouterConfig()`?

components/src/dynamo/common/configuration/groups/kv_router_args.py
```diff
@@ -423,6 +528,34 @@ def add_arguments(self, parser) -> None:
...
+            flag_name="--router-conditional-disagg-config",
+            env_var="DYN_ROUTER_CONDITIONAL_DISAGG_CONFIG",
+            default=None,
+            help=(
+                "[EXPERIMENTAL] KV Router: JSON object for conditional "
+                "disaggregation policy settings. Supported fields: policy, "
+                "eff_isl_threshold, eff_isl_ratio_threshold, "
+                "prefill_busy_threshold, decode_busy_threshold."
+            ),
+            arg_type=str,
```
MatejKosec: argparse already gives us a convenient way to convert an argument as it reads it through `type`. Could we use that here instead of accepting a plain string?

The code currently asks argparse for:

```python
arg_type=str
```

We could instead give it a small converter:

```python
def _conditional_disagg_config_arg(value: str) -> dict[str, Any]:
    try:
        return _parse_conditional_disagg_config(value)
    except ValueError as exc:
        raise argparse.ArgumentTypeError(str(exc)) from exc
```

and use:

```python
arg_type=_conditional_disagg_config_arg
```

I would keep all the current JSON and type checks inside `_parse_conditional_disagg_config`. Argparse would simply call that function while reading the command-line or environment value. `conditional_disagg_config` could then hold the resulting dictionary:

```python
conditional_disagg_config: Optional[dict[str, Any]] = None
```

At the moment, `FrontendConfig.validate()` calls:

```python
self.apply_conditional_disagg_config()
```

which reaches:

```python
parsed = _parse_conditional_disagg_config(self.conditional_disagg_config)
```

The frontend later calls `kv_router_kwargs()`, which calls `apply_conditional_disagg_config()` again. We therefore decode and check the same JSON twice.

If argparse gives us the checked dictionary at the start, `apply_conditional_disagg_config()` can use it directly:

```python
for key, value in self.conditional_disagg_config.items():
    setattr(self, _CONDITIONAL_DISAGG_CONFIG_FIELDS[key], value)
```

`FrontendConfig.validate()` should still handle the checks that need other arguments. For example, `_parse_conditional_disagg_config` cannot tell whether conditional disaggregation is enabled or whether `router_queue_threshold` supplies the prefill threshold. It should continue to decode the JSON and check its keys and value types; the frontend should continue to check how that configuration relates to the rest of the command line.

components/src/dynamo/common/configuration/groups/kv_router_args.py
karen-sy: I think thats a good point but the other KV router args also have duplicated defaults and don't have explicit checks about it. Think I'll keep the CD args consistent with that, though we can consider a followup PR which adds tests for all of the arguments for consistency

MatejKosec (approved):

alec-flowers (approved):

PeaBrane (approved):

PR #11724, opened 2026-07-15
feat(gms): surface scratch-KV engagement in shadow worker logs

mohammedabdulwahhab:
Present a clean summary of the scratch KV footprint when engaged during failover.

`[GMS] Scratch-KV engaged: 33 allocations, 69.65 GiB virtual reserved, 16.50 GiB physical backing on device 0`

### DIFF
--- a/lib/gpu_memory_service/integrations/vllm/worker.py
+++ b/lib/gpu_memory_service/integrations/vllm/worker.py
@@ -275,12 +275,9 @@ class GMSWorker(Worker):
             # The scratch pool is scoped to the KV tensors by
             # patch_kv_cache_pool_scope(), not the whole initialize_kv_cache.
             self.model_runner.initialize_kv_cache(kv_cache_config)
-            # The per-mapping "[GMS] Reserved ... scratch" line is logged at INFO
-            # on the gpu_memory_service logger, which the vLLM worker subprocess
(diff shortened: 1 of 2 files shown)


lib/gpu_memory_service/integrations/vllm/worker.py
```diff
@@ -270,9 +270,32 @@ def initialize_from_config(self, kv_cache_config) -> None:
...
+            # worker-logging fix is a follow-up).
+            n_scratch, virtual_bytes, physical_bytes = scratch_mgr.scratch_summary()
+            msg = (
+                f"[GMS] Scratch-KV engaged: {n_scratch} allocations, "
+                f"{virtual_bytes / (1 << 30):.2f} GiB virtual reserved, "
+                f"{physical_bytes / (1 << 30):.2f} GiB physical backing "
+                f"on device {device}"
+            )
+            logger.info(msg)
+            print(msg, flush=True)
```
galletas1712: why do we need a print here? also why the need for an extra msg variable?

mohammedabdulwahhab: The GMS logger is broken for our patches/class overrides that run in the vllm worker process context. I have a follow up to fix it.

galletas1712 (approved):

PR #11725, opened 2026-07-15
feat(router): orchestrate conditional disagg bypass

karen-sy:
> [!IMPORTANT]
> This PR is **[4/5]** in the conditional-disaggregation stack.
> Merge order is bottom-up: **#11718 merges to `main` first**, and **#11723 merges last**.
> Do not merge this PR until every earlier PR in the checklist has merged and this branch has been rebased/retargeted as needed.

Link to the DEP: [https://github.com/ai-dynamo/dynamo/issues/11514](https://github.com/ai-dynamo/dynamo/issues/11514)

Stack merge order:

- [ ] #11718 — feat(router): add conditional disagg policy kernel
- [ ] #11719 — feat(backends): handle conditional disagg bypass annotation
- [ ] #11720 — feat(router): add decode affinity scoring
- [ ] **#11725 — feat(router): orchestrate conditional disagg bypass**
- [ ] #11723 — feat(router): expose conditional disagg configuration

Reference to original (closed) combined PR, with addressed review comments: [ai-dynamo/dynamo#11357](https://github.com/ai-dynamo/dynamo/pull/11357)

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **New Features**
  * Added conditional routing that can bypass remote prefill and send requests directly to an appropriate decode worker.
  * Added session-affinity support to improve consistency when selecting workers across request phases.
  * Added advisory worker selection and load checks to make routing decisions without reserving capacity.
  * Added configurable safeguards based on prefill and decode worker workload.

* **Bug Fixes**
  * Improved propagation of routing affinity and configuration across prefill and decode paths.


PeaBrane (commented):
AI-assisted review of the conditional-disagg orchestration and stacked contracts. Posting these as neutral questions and edge cases for consideration.

lib/llm/src/kv_router/prefill_router/mod.rs
```diff
@@ -192,6 +214,62 @@ impl
             .get_optional::<SessionAffinityId>(SESSION_AFFINITY_CONTEXT_KEY)
             .map_err(|message| anyhow::anyhow!("invalid session affinity context: {message}"))?;
 
+        let decode_affinity_target =
+            self.decode_session_affinity_target(session_affinity.as_deref())?;
```
PeaBrane, line 236: AI-assisted review: The session-affinity lookup is not reserved, allowing concurrent first requests to choose conflicting workers.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,334 @@
...
+                routing_token_ids,
+                block_mm_infos,
+                req.router_config_override.as_ref(),
+                false,
+                lora_name,
+                cache_namespace,
+                priority_jump,
+                strict_priority,
+                expected_output_tokens,
+                None,
```
PeaBrane: AI-assisted review: The prefill-load probe ignores prefill session affinity and may inspect a different worker than normal P→D.

lib/llm/src/kv_router/prefill_router/mod.rs
```diff
@@ -192,6 +214,62 @@ impl
...
+                    if let Some(ref tracker) = req.tracker {
+                        let _decode_permit = tracker.set_phase(RequestPhase::Decode).await;
+                    }
+
+                    let routing = req.routing_mut();
+                    routing.decode_worker_id = Some(decision.worker.worker_id);
+                    routing.dp_rank = Some(decision.worker.dp_rank);
+
+                    req.annotations
+                        .push(BYPASS_REMOTE_PREFILL_ANNOTATION.to_string());
```
PeaBrane, line 271: AI-assisted review: No worker capability gate exists before adding the bypass marker, allowing SGLang or mixed-version requests to fail after dispatch.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,334 @@
...
+        };
+        input = input.with_decode_chosen_worker_busy(decode_busy);
+
+        let bypass = decode_gate_allows_bypass(policy_says_bypass, decode_busy);
+        let decode_gate_decision = if !policy_says_bypass {
+            "bypass_declined_by_policy"
+        } else if self.conditional_disagg_decode_busy_threshold.is_none() {
+            "bypass_allowed_decode_gate_disabled"
+        } else if decode_busy.is_none() {
+            "bypass_allowed_decode_busy_unknown"
```
PeaBrane: AI-assisted review: An explicitly configured decode-busy guard fails open when capacity is unknown; TensorRT-LLM currently leaves it unset.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,334 @@
...
+            crate::kv_router::FindBestMatchOutcome::QueueRejected { .. } => return Some(true),
+        };
+
+        router
+            .chooser
+            .worker_is_prefill_busy(worker, tokio::time::Instant::now(), threshold)
+    }
+}
+
+#[cfg(test)]
```
PeaBrane, line 380: AI-assisted review: Could we add an orchestration-level test covering affinity and explicit pins, unavailable capacity, annotation dispatch, and fallback?

PeaBrane (approved):
Approving on the condition that the AI-assisted review comments above are resolved.

lib/llm/src/kv_router/prefill_router/mod.rs
```diff
@@ -413,13 +526,21 @@ fn compute_bootstrap_room(dp_rank: Option<u32>, dp_size: Option<u32>, random_roo
...
+    // `overlap_score_credit` applies, unless the request already had an
+    // explicit override.
+    if !allow_decode_overlap_affinity {
+        override_config.overlap_score_credit = Some(0.0);
     }
+    override_config.assume_kv_reuse = Some(false);
+    override_config.track_prefill_tokens = Some(false);
+
+    override_config
 }
```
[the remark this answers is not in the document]
karen-sy: This is intentional. Otherwise requests that are routed P->D and not CD will have massive decode kv cache misses

lib/llm/src/kv_router/prefill_router/mod.rs
```diff
@@ -192,6 +214,62 @@ impl
...
+                        .push(BYPASS_REMOTE_PREFILL_ANNOTATION.to_string());
+
+                    let response_stream = next.generate(context.map(|_| req)).await?;
+                    let ctx = response_stream.context();
+                    let annotation = Annotated::<LLMEngineOutput>::from_annotation(
+                        BYPASS_REMOTE_PREFILL_ANNOTATION,
+                        &true,
+                    )?;
+                    let merged = stream::once(async move { annotation }).chain(response_stream);
+                    return Ok(ResponseStream::new(Box::pin(merged), ctx));
```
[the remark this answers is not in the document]
karen-sy: This is intentional. Dynamo response streams use `Annotated<T>` for both engine outputs and routing/control annotation, so this mirrors the existing stream-annotation pattern. The backend consumers for this annotation are covered in #11719. (#11725 only owns inserting the router-owned bypass marker when the policy approves bypass.)

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,334 @@
...
+                        worker_id = target.worker_id,
+                        "Skipping conditional disagg because decode affinity target has no resolved DP rank"
+                    );
+                    return Ok(None);
+                };
+                Some(WorkerWithDpRank::new(target.worker_id, dp_rank))
+            }
+            None => None,
+        };
+        let request_pinned_worker = req.routing.as_ref().and_then(|routing| {
```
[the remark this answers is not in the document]
karen-sy: Fixing by making the c/d probe distinguish “no explicit decode pin” from “explicit decode pin with unresolved dp_rank.” In the unresolved case, c/d now skips bypass instead of treating the request as unpinned, so the normal pinned-worker path preserves the existing validation/error behavior. (i.e. the `"Pinned worker {worker_id} requires an explicit dp_rank because it has multiple or unknown DP ranks"` error in P->D mode)

lib/llm/src/kv_router/prefill_router/mod.rs
karen-sy: True: c/d is a speculative probe, so it can query existing session affinity but can't really "reserve" a new affinity target before deciding bypass.
Adding that reservation feature would require an explicit reservation handoff into downstream routing to avoid double-booking or binding a worker for a request that later falls back to normal P→D.

For now I think this is ok, since the race condition happens in a pretty narrow setting (concurrent first requests for the SAME session before any affinity binding exists) that doesn't apply much to us yet. Once a binding exists, c/d respects it. Our target trace issues same-session turns sequentially, so this should not affect the intended workload. But worth tackling in a followup where parallel fanout workloads are more common.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
karen-sy: fixing by adding a `KvPushRouter::query_affinity_worker(request, phase)` that is called by c/d prefill-load peek

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
karen-sy: Fixing so that the behavior is:
If decode gate is disabled: unknown busy state does not block bypass. (`bypass_allowed_decode_busy_unknown`)
If decode gate is configured and busy state is unknown: deny bypass. (`bypass_denied_decode_busy_unknown`)

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
karen-sy: orchestration tests would require too much harness boilerplate, so added some more tests for the fixed behaviors

- explicit decode pin resolution now covers explicit rank, unique inferred rank, unresolved explicit pin, and no target
- decode-busy gate behavior now covers policy decline, disabled gate, busy worker, calm worker, and configured-but-unknown capacity fail-closed
- session affinity now has a read-only `query_affinity_worker(...)` test proving existing bindings can be inspected without reserving/rebinding

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -664,6 +690,31 @@ impl<
...
+        &self,
+        worker: WorkerWithDpRank,
+        projected_blocks: usize,
+        threshold: f64,
+    ) -> Option<bool> {
+        let workers = self.workers_with_configs.borrow();
+        let config = workers.get(&worker.worker_id)?;
+        let capacity = config.total_kv_blocks()? as f64;
+        Some(projected_blocks as f64 > threshold * capacity)
+    }
```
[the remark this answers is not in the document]
karen-sy: fixed, worker_is_prefill_busy() now mirrors queue behavior with DEFAULT_MAX_BATCHED_TOKENS fallback

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,457 @@
...
+            DecodePinResolution::Unresolved { worker_id } => {
+                tracing::debug!(
+                    request_id,
+                    worker_id,
+                    "Skipping conditional disagg because request has an explicit decode worker with no resolved DP rank"
+                );
+                return Ok(None);
+            }
+        };
+        let pinned_worker = request_pinned_worker.or(affinity_pinned_worker);
```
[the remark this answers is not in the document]
karen-sy: Explicit request decode pin now resolves before session-affinity fallback, so stale affinity cannot preempt a valid explicit pin.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,457 @@
...
+                routing_token_ids,
+                block_mm_infos,
+                req.router_config_override.as_ref(),
+                false,
+                lora_name,
+                cache_namespace,
+                priority_jump,
+                strict_priority,
+                None,
+                None,
```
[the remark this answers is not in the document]
karen-sy: Conditional-disagg decode and prefill probes now pass policy_class and agent_context.session_id into find_best_match_details_without_admission(...), matching real KvPushRouter selection

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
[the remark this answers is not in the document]
karen-sy: specifically, the bhavior now:
1. Resolve request’s explicit decode pin first.
2. If it resolves, use it.
3. Only if there is no request pin, resolve decode_affinity_target.
4. If affinity target has no resolvable rank, then skip conditional disagg.

MatejKosec:
One design question after tracing this path:

The scheduler already has the information needed here: which worker it selected, the projected load for that worker, and the worker’s capacity. At the moment, the advisory selection returns only part of that information. `PrefillRouter` then asks back down through `KvRouter` → `KvScheduler` → `LocalScheduler` → `SchedulerQueue` whether the selected worker exceeds a threshold. That is why methods such as `projected_decode_load_exceeds` and `worker_is_prefill_busy` appear at several layers even though the calculation itself happens only once.

Would it be cleaner for the advisory selection to return the relevant load measurements together with the selected worker? For example, a small result could contain the worker, cached tokens, projected decode blocks, and total KV capacity. The conditional-disaggregation code could then apply its configured threshold directly.

I think that would make the responsibilities easier to follow. The scheduler would report the measurements it already owns, while conditional disaggregation would decide what those measurements mean for the bypass policy. It would also let the worker selection and the associated load/capacity measurements come from the same observation, rather than selecting first and querying some of the state afterward.

I’m not suggesting a broad snapshot of everything in the scheduler—just a small, purpose-built advisory result. This would not remove the acknowledged race between advisory selection and real admission, but it would make this path more cohesive and avoid adding more one-off load queries through every layer as conditional disaggregation evolves.

I do not think this needs to block the PR, but I wanted to raise it as a possible cleanup direction.

MatejKosec:
2. `select_without_admission_inner` now has its own copy of most of the code in `select_worker_for_request`. Both functions calculate the current worker loads, remove workers that cannot accept the request, choose the best remaining worker, and collect information about that worker. The new function needs some extra information, which explains why the code was copied. The problem is that there are now two places that define how a worker is chosen. They behave the same today, but a future change to worker selection could be added to one function and missed in the other. Then the advisory path and the normal admission path could make different choices for reasons that are very difficult to see while reading the code. I think both paths should use one shared selection function, with the advisory path adding the chosen worker’s load and capacity to the result.

lib/llm/src/kv_router/prefill_router/conditional_bypass.rs
```diff
@@ -0,0 +1,464 @@
...
+        // decision; conditional-disagg is still deciding whether to enter that
+        // decode path, so booking here would double-count unless we added a
+        // reservation handoff to the downstream decode router. Use the selected
+        // worker's projected decode load, including this request, but allow for
+        // concurrent bypass decisions to race the same threshold.
+        let decode_gate_configured = self.conditional_disagg_decode_busy_threshold.is_some();
+        let decode_busy = if policy_says_bypass {
+            self.conditional_disagg_decode_busy_threshold
+                .and_then(|threshold| {
+                    decode_load.decode_load_exceeds(potential_decode_blocks, threshold)
```
MatejKosec, line 244: I think `potential_decode_blocks_after_admission` counts the incoming request twice when active-block tracking is enabled. The code that builds this number is spread across several files, so I want to trace the whole path.

The path begins in `find_best_match_details_with_policy_class_inner`. It creates sequence hashes for the incoming request and puts them in `ScheduleRequest::token_seq`:

```rust
let maybe_seq_hashes = self
    .kv_router_config
    .compute_seq_hashes_for_tracking_with_context(
        &self.tracking_hash,
        self.tracking_hash_scope(),
        tokens,
        router_config_override,
        hash_options,
        Some(&block_hashes),
    );

let schedule_request = ScheduleRequest {
    token_seq: maybe_seq_hashes,
    isl_tokens,
    // ...
};
```

When active-block tracking is enabled, `token_seq` contains one sequence hash for each complete prompt block represented by the tracker. A final partial block is not included. When active-block tracking is disabled, `token_seq` is `None`.

The scheduler then passes that request-specific sequence to `project_worker_loads`:

```rust
request.worker_loads = self
    .slots
    .project_worker_loads(request.token_seq.as_deref(), decay_now);
```

`project_worker_loads` uses the length of `token_seq` as the number of complete blocks in this request. It also finds how many of those blocks are already shared with an active sequence on each worker:

```rust
let query_len = token_sequence.map_or(0, |query| query.len());
let overlap_depth = matched_depth.get(&worker).copied().unwrap_or(0);

WorkerLoadProjection {
    active_prefill_tokens: load.active_tokens(decay_now),
    active_decode_blocks: load.active_blocks,
    active_requests: load.active_requests,
    additional_active_blocks: query_len.saturating_sub(overlap_depth),
}
```

The important field here is `additional_active_blocks`. It already describes the complete blocks from this incoming request that would increase the worker's active block footprint. If the request has four complete blocks and none of them overlap an active sequence, `additional_active_blocks` is four. If all four are already shared with an active sequence, it is zero.

`WorkerLoadProjection::potential_decode_blocks` then adds those request-specific blocks to the blocks that are already active on the worker:

```rust
pub fn potential_decode_blocks(self) -> usize {
    self.active_decode_blocks + self.additional_active_blocks
}
```

By the time `potential_decode_blocks()` returns, the result already includes the complete blocks that this request would add. `potential_decode_blocks_after_admission` then adds the full request size again:

```rust
self.worker_load_for(worker)
    .potential_decode_blocks()
    .saturating_add(self.request_blocks(block_size) as usize)
```

For example, suppose the block size is 16, the worker currently has 100 active blocks, and the prompt contains 32 tokens. The prompt contains two complete blocks. With no overlap against an active sequence, the calculation is:

```text
active_decode_blocks       = 100
additional_active_blocks   = 2
potential_decode_blocks()  = 102
request_blocks             = 2
returned value             = 104
```

The request should increase the worker's active block footprint from 100 to 102 blocks, but this function returns 104.

The same problem is visible when the incoming request fully overlaps an active sequence:

```text
active_decode_blocks       = 100
additional_active_blocks   = 0
potential_decode_blocks()  = 100
request_blocks             = 2
returned value             = 102
```

In that case, the two complete prompt blocks are already part of the worker's active block footprint. Admitting another request that shares those blocks should not add two new active blocks, but this function still adds them.

There is one reason that some addition is needed here. When active-block tracking is disabled, `token_seq` is `None`, so `query_len` and `additional_active_blocks` are both zero. A final partial prompt block is also absent from `token_seq`. Those blocks have not been included by `potential_decode_blocks()` and still need to be added.

It therefore looks as though this function should add only the part of `request_blocks` that is not already represented by `token_seq`, rather than adding the whole request:

```rust
let request_blocks = self.request_blocks(block_size) as usize;
let tracked_request_blocks = self.token_seq.as_ref().map_or(0, Vec::len);
let untracked_request_blocks =
    request_blocks.saturating_sub(tracked_request_blocks);

self.worker_load_for(worker)
    .potential_decode_blocks()
    .saturating_add(untracked_request_blocks)
```

With tracking disabled, `tracked_request_blocks` is zero, so the full request is still added. With tracking enabled, only a partial or otherwise untracked tail is added; the complete tracked blocks are not added a second time.

This value is then returned in the scheduling response and used by conditional disaggregation here:

```rust
decode_load.decode_load_exceeds(potential_decode_blocks, threshold)
```

The inflated value can make the decode-busy check say that the selected worker is over its configured threshold when its actual projected active footprint is still below that threshold. This is conservative: it prevents a bypass rather than allowing too much decode load. But it can cause requests to take the remote-prefill path when the configured policy should have allowed them to bypass it.

A focused test would make the intended calculation much easier to verify. I would cover active-block tracking disabled, tracking enabled with no overlap, tracking enabled with full active overlap, and a prompt with a final partial block.

karen-sy: another good catch, added some tests for all cases. Cases for behavior now should be:
- tracking disabled: add full request
- tracking enabled with complete blocks: do not double-add those complete blocks
- final partial block: still add it, because token_seq only represents complete blocks

RCA: I previously made a commit to make `potential_decode_blocks` post-admission, but it added the full request block count even though `project_worker_loads(...)` already includes the tracked complete-block portion through `additional_active_blocks`

lib/llm/src/kv_router/prefill_router/mod.rs
```diff
@@ -210,6 +232,64 @@ impl
...
+                    }
+
+                    let routing = req.routing_mut();
+                    routing.decode_worker_id = Some(decision.worker.worker_id);
+                    routing.dp_rank = Some(decision.worker.dp_rank);
+
+                    req.annotations
+                        .push(BYPASS_REMOTE_PREFILL_ANNOTATION.to_string());
+
+                    let response_stream = next.generate(context.map(|_| req)).await?;
```
MatejKosec, line 276: I think this path is missing a fallback when the advisory worker cannot be admitted.

`select_decode_worker_for_conditional_disagg` deliberately chooses a decode worker without admitting the request or holding that worker for it. If the policy approves the bypass, the chosen worker is then written into the request as an exact decode pin:

```rust
routing.decode_worker_id = Some(decision.worker.worker_id);
routing.dp_rank = Some(decision.worker.dp_rank);
```

The real admission happens afterward, inside:

```rust
let response_stream = next.generate(context.map(|_| req)).await?;
```

The scheduler state can change between those two operations. The selected worker might disappear, become ineligible, or no longer be able to accept the request. Several bypass decisions can also choose the same worker before any of those requests reaches real admission.

If the advisory selection itself returns an error, the code already falls back to normal remote prefill:

```rust
Err(error) => {
    tracing::warn!(
        request_id = %request_id,
        error = %error,
        "Conditional disagg decision failed; falling back to remote prefill"
    );
}
```

But if the advisory selection succeeds and the later exact admission fails, the `?` returns that error to the caller. The request does not fall back to remote prefill, even though remote prefill was still the ordinary route before the advisory bypass was chosen.

I think a failure to admit the advisory worker, provided it occurs before dispatch, should cancel the bypass and continue through the normal remote-prefill path. I would not catch every error from `next.generate`, because an error after dispatch may not be safe to retry. The missing case is specifically a rejected or invalid exact admission before the request has been sent to a backend.

A reservation passed from advisory selection into real admission would also close this case, but without such a handoff the bypass path needs a defined fallback when its advisory worker is no longer available.

lib/llm/src/kv_router.rs
```diff
@@ -807,34 +899,53 @@ where
...
+            {
+                Ok(response) => (response, None),
+                Err(KvSchedulerError::QueueRejected(rejection)) => {
+                    return Ok(FindBestMatchInnerOutcome::WithAdmission(
+                        FindBestMatchOutcome::QueueRejected { rejection },
+                    ));
+                }
+                Err(error) => return Err(map_scheduler_error(error)),
+            },
+            FindBestMatchAdmission::WithoutAdmission => match self
```
MatejKosec, line 935: I think the new without-admission path is also being counted as ordinary routing work in the existing per-request measurements.

Both branches of this match produce `response` and then continue through the same measurement code:

```rust
let (response, selected_worker_load) = match admission {
    FindBestMatchAdmission::WithAdmission { .. } => { ... }
    FindBestMatchAdmission::WithoutAdmission => { ... }
};

if let Some(m) = metrics::RoutingOverheadMetrics::get() {
    m.observe(
        hash_elapsed,
        seq_hash_elapsed,
        indexer_duration,
        shared_cache_duration,
        find_matches_elapsed,
        total_elapsed,
    );
}
```

The shared-cache measurements below are handled in the same way:

```rust
m.shared_cache_hit_rate
    .observe(hits.total_hits as f64 / num_blocks as f64);
m.shared_cache_beyond_blocks.observe(beyond as f64);
```

This means that an advisory inspection produces the same histogram observations as a routing call that actually admits and dispatches the request.

With conditional disaggregation enabled, one inference request can now run a decode advisory selection, sometimes a prefill advisory selection, and then the real routing selections. For example, a request using the prefill-load policy can contribute an advisory decode observation, an advisory prefill observation, and then one or more observations from the route it actually takes.

Those are recorded as separate histogram samples. They are not combined into one sample representing the request's total routing cost. The number of samples also depends on the configured policy and on whether the request bypasses remote prefill. Enabling conditional disaggregation can therefore change the meaning and weighting of these measurements, even when the underlying admitted routing work is unchanged.

I do not think this affects routing correctness, but it can make comparisons of routing latency and shared-cache behaviour misleading. I think the existing measurements should either be recorded only for the admission path, or the advisory probes should use separately named measurements.

lib/llm/src/kv_router/prefill_router/mod.rs
karen-sy: Leaving this potential race condition as a TODO item in mod.rs. Also created [DYN-3694](https://linear.app/nvidia-dynamo/issue/DYN-3694)

lib/llm/src/kv_router.rs
karen-sy: fixed so only requests admitted to scheduler is recorded for metrics

lib/llm/src/kv_router/prefill_router/mod.rs
MatejKosec: I traced this a little further to check whether the TODO would require a larger redesign. I do not think it would.

There is already a precise point inside `KvPushRouter` where the narrow fallback is safe. `KvPushRouter::generate` performs the real scheduler admission here:

```rust
self.select_with_affinity(&request, phase, is_query_only).await
```

This call only borrows `request`. If it returns an error, `track_selection` and `dispatch_selection` have not run, so no backend dispatch has begun. The scheduler has also either not booked the request or has rolled the booking back before returning the error. At that point, `KvPushRouter` still owns the original `SingleIn<PreprocessedRequest>`, including its controller, registry, cancellation state, stages, and metadata.

That ownership is important. `PrefillRouter` cannot safely reconstruct the same request context from a cloned `PreprocessedRequest`; it needs the original context back if it is going to continue through remote prefill.

One contained implementation would be for `KvPushRouter` to wrap only a non-cancellation failure from this exact admission call in a typed error or outcome which owns the request and preserves the original error as its source. Conceptually:

```rust
NotAdmitted {
    request,
    source,
}
```

This is similar to a channel send error returning the message that it could not send. `PrefillRouter` could recognize only this result, recover the same request, restore the complete `routing` value that existed before the advisory decode pin was installed, remove `x-bypass-remote-prefill`, and continue into the existing remote-prefill path. That path already changes the tracker phase back to `Prefill`.

Cancellation should remain cancellation. Errors from `track_selection`, `dispatch_selection`, the backend call, or response-stream setup should also remain terminal. In particular, I would not identify this case by catching `ResourceExhausted` or `CannotConnect` around the whole `next.generate(...)` call. Those error classes can also come from later stages, including stages where the request may already have reached the backend.

There is one slightly broader case worth making explicit in DYN-3694. After scheduler admission succeeds, runtime `dispatch_exact` performs a final discovery and overload check before it calls `addressed.generate`. A failure in that preflight is also known to be before the backend call, but the current routing path has already begun recording tracker state and transferring ownership by then. If the intended contract is to fall back for every failure that is definitely pre-send, exact dispatch should be split into preflight and send phases, with preflight occurring before those tracker updates and returning the request when it fails. That would be a somewhat larger local refactor across runtime `PushRouter`, `KvPushRouter`, and `PrefillRouter`, but it would still not require a scheduler redesign or a reservation protocol.

So the narrow scheduler-admission fallback is a small, local change. The broader guarantee covering every definitely-not-sent failure needs a clearer phase boundary, but it is still implementable without a major overhaul. The important part is to carry the request and the failure stage explicitly rather than trying to infer the stage from the ordinary error returned by `next.generate(...)`.

MatejKosec (approved):
Approved, once comments resolved.

lib/llm/src/kv_router/prefill_router/mod.rs
MatejKosec: I put the concrete implementation in #12467. It is a draft PR from my fork into this feature branch, so it does not alter this branch unless the proposal is accepted.

The proposal returns the original request when `select_with_affinity` fails with a non-cancellation error. That is the boundary before `track_selection`, `RequestGuard` creation, and backend dispatch. `PrefillRouter` restores the complete routing value from before the advisory pin, removes the bypass marker, and continues through normal remote prefill. Cancellation and every error after selection remain terminal.

The pipeline frontend uses a registration guard for the abandoned downstream call. The guard removes only the sender registered by that call, including when the call returns an immediate error or is cancelled.

The code comments and PR description now state these boundaries directly. I ran formatting and `git diff --check`, but I did not run the tests, so the proposal remains a draft.

PR #11729, opened 2026-07-15
fix(frontend): forward request chat_template_kwargs to vLLM template

### LINKED ISSUE #11704: [BUG]: vLLM chat-processor silently drops request body chat_template_kwargs
### Describe the Bug

With `--dyn-chat-processor=vllm`, `chat_template_kwargs` supplied in a chat-completion request payload (such as `{"enable_thinking": false}`) never reaches the chat template, so per-request template control is impossible (for example, thinking cannot be disabled on a per-request basis for reasoning models).
 
Cause is a serde field-name asymmetry at the Rust Python bindings:
 
1. Here the request field is named `chat_template_args`, with `alias = "chat_template_kwargs"`. serde's `alias` applies **only on deserialization**. The client's `chat_template_kwargs` is read into the Rust field `chat_template_args`.

https://github.com/ai-dynamo/dynamo/blob/b0b778678f37943f4f57726ec81783eb66a6e98f/lib/llm/src/protocols/openai/chat_completions.rs#L100-L107

2. The request is handed to the Python vLLM processor via `pythonize(py, &request)`. `pythonize` uses serde *serialization*, which emits the **Rust** field name. Since `alias` does not affect serialization, the Python dict is keyed `chat_template_args`, **not** `chat_template_kwargs`.

https://github.com/ai-dynamo/dynamo/blob/b0b778678f37943f4f57726ec81783eb66a6e98f/lib/bindings/python/rust/engine.rs#L278

3. The vLLM processor reconstructs `ChatCompletionRequest` and reads `request_for_sampling.chat_template_kwargs`. vLLM's model only knows the key `chat_template_kwargs`; it does not recognize **`chat_template_args`**. So the kwargs variable resolves to `{}`.

https://github.com/ai-dynamo/dynamo/blob/b0b778678f37943f4f57726ec81783eb66a6e98f/components/src/dynamo/frontend/prepost.py#L139

Net effect:  The template only ever renders with the model's built-in defaults.

### Steps to Reproduce

For instance, for Qwen models, the canonical way to disable reasoning is with the `enable_thinking` flag.

1. Start frontend:
```
   python -m dynamo.frontend \
     --dyn-chat-processor=vllm \
     --reasoning-parser qwen3
```
2. Start vLLM worker
```
   python -m dynamo.vllm \
     --model Qwen/Qwen3.6-27B-FP8 \
     ...
```
3. Send a chat request:
```
   curl http://localhost:8080/v1/chat/completions \
     -H 'Content-Type: application/json' \
     -d '{
              "model": "Qwen/Qwen3.6-27B-FP8",
              "messages": [{"role": "user", "content": "9.11 and 9.8, which is greater?"}],
              "chat_template_kwargs": {"enable_thinking": false}
       }'
```

### Expected Behavior

Response body with empty `reasoning_content`:
```json
{
    "choices": [
        "message": {
             "role": "assistant",
             "content": "To determine which number is greater beween **9.11** and **9.8** ... ",
        }
    ],
    ...
}
```

### Actual Behavior

Response body:
```json
{
    "choices": [
        "message": {
             "role": "assistant",
             "content": "**9.8 is greater**.\n\nWhen comparing decimals, ... ",
             "reasoning_content": "Here's a thinking process ...",
              ...
        }
    ],
    ...
}
```

### Environment

OS: ubuntu 24.04.3
ai-dynamo version: v1.3.0.dev1
CPU arch: x86_64
CUDA version: 13.2
GPU: 8xH100 80GB HBM3
Python: 3.12.3

### Additional Context

_No response_

### Screenshots

_No response_

KrishnanPrash:
#### Overview
With `--dyn-chat-processor=vllm`, per-request chat template controls sent in the request body (e.g. `chat_template_kwargs: {"enable_thinking": false}`) never reached the chat template, so things like disabling reasoning per-request were silently ignored. The vLLM processor read only vLLM's native `chat_template_kwargs` attribute, but requests cross the Rust->Python boundary serialized under the key `chat_template_args` (serde `alias` is deserialize-only, so `pythonize` emits the Rust field name). This forwards the client's kwargs into the template and hardens two adjacent edge cases the forwarding exposes.

Repro:
```
curl http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "Qwen/Qwen3.6-27B-FP8",
  "messages": [{"role":"user","content":"9.11 and 9.8, which is greater?"}],
  "chat_template_kwargs": {"enable_thinking": false}
}'
```

Before (`enable_thinking: false` ignored, model still reasons):
```json
"message": {
  "role": "assistant",
  "content": "**9.8 is greater** ...",
  "reasoning_content": "Here's a thinking process ..."
}
```

After (`enable_thinking: false` honored):
```json
"message": {
  "role": "assistant",
  "content": "To determine which is greater ...",
  "reasoning_content": null
}
```

#### Details
`_prepare_request` in `components/src/dynamo/frontend/prepost.py` now falls back to the raw dict's `chat_template_args` when vLLM's native `chat_template_kwargs` is absent, mirroring the dual-read already used in `sglang_prepost.py` and `common/utils/input_params.py`. Two edge bugs that the forwarding surfaces are fixed in the same function: (1) a nested `reasoning_effort` was unconditionally overwritten by the absent top-level field, now the top-level only wins when present; (2) a client nesting a renderer-managed key (`documents`, `tools`, `add_generation_prompt`, ...) triggered a duplicate-keyword `TypeError`, now the renderer-managed keys are applied last so the merge never raises. Adds 4 unit tests in `test_vllm_processor_unit.py::TestChatTemplateArgsPassthrough` covering the forwarding, the native key, the reasoning_effort guard, and the reserved-key merge.

#### Related Issues
- Resolves #11704


components/src/dynamo/frontend/prepost.py
```diff
@@ -136,8 +136,23 @@ def _prepare_request(
         )
         else None
     )
-    chat_template_kwargs = dict(request_for_sampling.chat_template_kwargs or {})
-    chat_template_kwargs["reasoning_effort"] = request_for_sampling.reasoning_effort
+    # The Rust bindings serialize this field as `chat_template_args` (serde
+    # `alias = "chat_template_kwargs"` applies on deserialize only), so a request
+    # arriving as a raw dict carries the client's kwargs there rather than under
+    # the vLLM-native `chat_template_kwargs`. Fall back to it so per-request
+    # template control is not silently dropped, mirroring sglang_prepost.py.
```
rmccorm4: > mirroring sglang_prepost.py

is this same support already added for sglang chat processor?

rmccorm4 (approved):

components/src/dynamo/frontend/prepost.py
KrishnanPrash: Yes 👍

components/src/dynamo/frontend/prepost.py
```diff
@@ -136,8 +136,19 @@ def _prepare_request(
...
     )
-    chat_template_kwargs = dict(request_for_sampling.chat_template_kwargs or {})
-    chat_template_kwargs["reasoning_effort"] = request_for_sampling.reasoning_effort
+    # serde's `alias` is deserialize-only, so pythonize emits the Rust field
+    # name `chat_template_args`; read it too or client kwargs are dropped.
+    raw_template_args = (
+        request.get("chat_template_args") if isinstance(request, dict) else None
+    )
+    chat_template_kwargs = dict(
+        request_for_sampling.chat_template_kwargs or raw_template_args or {}
```
indrajit96, line 145: AI flagged a review which seems reasonable

 [P1] Reject nested chat_template before forwarding. The new code copies chat_template_args wholesale into ChatParams. A client can therefore send:
```
    {"chat_template_kwargs": {"chat_template": "<attacker-controlled Jinja>"}}
```
Dynamo intentionally rejects top-level chat_template, but the nested form bypasses that validation. vLLM’s normal serving path explicitly blocks this unless trust_request_chat_template is enabled; Dynamo calls the renderer directly and skips that guard. ChatParams then promotes the nested value into the actual template. This recreates the resource-exhaustion path described in CVE-2025-61620 (https://github.com/vllm-project/vllm/security/advisories/GHSA-6fvq-23cw-5628). See the new
    forwarding code (https://github.com/ai-dynamo/dynamo/blob/0659872c3180047c0c504331117a7c2664343398/components/src/dynamo/frontend/prepost.py#L139-L176), Dynamo’s top-level rejection (https://github.com/ai-dynamo/dynamo/blob/0659872c3180047c0c504331117a7c2664343398/lib/llm/src/http/service/openai.rs#L4595-L4627), and vLLM’s trust check (https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/online_renderer.py#L261-L280).

Recommended fix: reject chat_template inside chat_template_args during Rust HTTP validation, or introduce an explicit trust opt-in matching vLLM. Add a regression test proving the nested key cannot reach the renderer.

I checked This is also present in Sgalng and Rust-HTTP path
Do we plan to cherrypick this change?
If NO
Worth fixing everywhere

indrajit96 (approved):
LGTM!
With a suggested good to have (Non Blocking)

components/src/dynamo/frontend/prepost.py
KrishnanPrash: Thanks for raising. Will address in follow-up PR since this would touch 3 separate paths and blow up scope for this PR.

PR #10999, opened 2026-06-26
feat(mm-routing): add Qwen3.5/3.6 + Kimi K2 family support

krishung5:
## Overview:

Add MM-aware KV routing support for the Qwen3.5/Qwen3.6 (hybrid GDN) and Kimi-K2.5/K2.6 model families. Bumps the mm-routing crates to `llm-multimodal=1.7.0` / `llm-tokenizer=1.4.1` resolved directly from crates.io (no patch override), promotes Qwen3.5-0.8B as the pre_merge gater on both vllm and sglang backends with a tightened (`min_avg_kv_hit_rate=0.7`) threshold, and fixes a launcher bug that was silently masking the sglang side of the gate.

Closes DIS-1968

## Details:

```mermaid
flowchart TB
    classDef router fill:#e6f3ff,stroke:#0366d6,stroke-width:2px,color:#000
    classDef payload fill:#fdf2e9,stroke:#9c5500,stroke-width:2px,color:#000
    classDef sgl fill:#e8f5e9,stroke:#2ea44f,stroke-width:2px,color:#000
    classDef vllm fill:#ffeaea,stroke:#d73a49,stroke-width:2px,color:#000
    classDef norm fill:#fff8c5,stroke:#b08800,stroke-width:2px,color:#000

    FE["<b>Rust Frontend</b>"]:::router

    %% Flow 1: what the router sees for scoring
    FE -->|"<b>(1) router input</b>"| ROUTER["<b>KV-Router</b><br/>scores workers by block-hash<br/>overlap over pad_value(mm_hash)-<br/>substituted tokens<br/><i>preprocessor.rs:1395-1414</i>"]:::router

    %% Flow 2: shared worker payload, fans out to both backends
    FE -->|"<b>(2) worker payload</b>"| PAYLOAD["<b>request to worker</b><br/>raw prompt tokens +<br/>extra_args.mm_hashes (hex list)<br/><i>SAME structure for both;</i><br/><i>each engine uses hashes differently</i><br/><i>preprocessor.rs:1280-1318</i>"]:::payload

    PAYLOAD --> SGL_W
    PAYLOAD --> VLLM_W

    SGL_W["<b>sglang worker</b><br/>passes mm_hashes to native engine<br/>(engine substitutes pad_value internally)<br/><i>decode_handler.py:449-462</i>"]:::sgl
[...]

PR #11737, opened 2026-07-15
feat(trtllm): native multimodal KV routing on rc21

KrishnanPrash:
#### Summary

TensorRT-LLM multimodal KV routing no longer needs a separate Python MM-router worker. This PR moves the aggregated TensorRT-LLM path onto Dynamo's existing Rust frontend + KV router and updates the worker integration for TensorRT-LLM 1.3.0rc21.

```text
BEFORE                         AFTER
Frontend (round-robin)         Frontend (Rust, --router-mode kv)
  |                              |
Python MM-router sidecar       TRT-LLM worker
  |
TRT-LLM worker (__internal)
```

The result deletes the duplicate tokenizer/image-processing/routing sidecar and its bespoke E2E harness while keeping one native request path.

#### How it works

- The Rust frontend computes the image routing hash and substitutes the corresponding `pad_value` into its routing-token view.
- Dynamo forwards those hashes to TensorRT-LLM as `multi_modal_uuids`; rc21 echoes them in KV-reuse events.
- rc21 preserves the real, in-vocabulary image marker in processed token IDs (the rc20 `vocab_size + 1` remap is gone).
- The worker resolves that real marker through the shared multimodal model registry and gives it to the KV-event publisher, which normalizes image-token runs to the same `pad_value` used by the frontend.
- The validated scope is aggregated Qwen2-VL/Qwen2.5-VL/Qwen3-VL and Kimi-K2.5/Kimi-K2.6. Other families fail closed to text-prefix routing.

The launcher now starts one real-named aggregated worker plus a `--router-mode kv` frontend. Its dedicated engine config enables block reuse and avoids the unused disaggregated cache transceiver. KV-event consolidation does not yet support multimodal events, so this uses the direct publisher path.

#### Validation

- **Full Dynamo E2E:** standard `tests/serve/test_trtllm.py` deployment passed in 73.67s. Two real Qwen3-VL image requests returned HTTP 200; request 2 reported `cached_tokens=25/25`, the router selected 8 overlapping blocks, and mean repeat `router_kv_hit_rate=0.889` (required >= 0.5).
- **Direct rc21 GPU protocol check:** Qwen3-VL resolved real image token `151655`; all 196 image markers in 7 KV blocks used that ID, none used `vocab_size + 1`, and an identical repeat reused 6/7 blocks (85.7%).
- **Focused tests:** 54/54 resolver, publisher, multimodal processor, and protocol tests passed against a freshly built `mm-routing` Python binding.
- **Protocol/config gates:** 2/2 rc21 wire-contract tests and 5/5 Qwen3-VL engine-config tests passed.
- **Static checks:** focused pre-commit suite, `mypy` for `llm_worker.py`, `cargo fmt --check`, shell syntax, and `git diff --check` passed.

#### Try it

```bash
bash examples/backends/trtllm/launch/agg_multimodal_router.sh
```

Send the same image twice; the repeat should report nonzero `usage.prompt_tokens_details.cached_tokens` and a high router KV-hit rate.

Resolves DIS-1932.


lib/bindings/python/rust/llm/kv.rs
```diff
@@ -968,6 +969,12 @@ impl KvEventPublisher {
             kv_block_size,
             dp_rank,
             warning_count: Arc::new(AtomicU32::new(0)),
+            // Unified-backend bridge only. TRT-LLM multimodal execution stays on
+            // the non-unified path (see components/src/dynamo/common/backend
+            // README + trtllm/llm_engine.py), whose direct `KvEventPublisher`
+            // constructor sets the marker; thread it here when MM moves to the
+            // unified path.
+            image_token_id: None,
```
rmccorm4: Why does unified backend require this change but existing implementations don't?

@connorcarpenter15 @tanmayv25 @KrishnanPrash

KrishnanPrash: It's the reverse, sorry, the comment was unclear. This PR wires the marker only on the non-unified path, which is where TRT-LLM multimodal actually runs (python -m dynamo.trtllm → init_llm_worker, whose KvEventPublisher gets image_token_id and normalizes the vocab_size+1 runs to the pad_value the frontend routes on). The unified backend is not wired for TRT-LLM MM in this PR, so from_arc correctly leaves it None; the comment is just a forward-looking note for when MM moves to the unified path. I've reworded it to say that. @connorcarpenter15 @tanmayv25 for the unified-backend MM plans.

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -968,6 +969,11 @@ impl KvEventPublisher {
             kv_block_size,
             dp_rank,
             warning_count: Arc::new(AtomicU32::new(0)),
+            // None: this bridge is unified-backend only, which this PR does not
+            // wire for TRT-LLM multimodal. MM runs on the non-unified path, where
+            // the `KvEventPublisher` constructor sets the marker. Thread it here
+            // if/when MM moves to the unified backend.
+            image_token_id: None,
```
rmccorm4: Let me rephrase my other question a bit --

Why does TRTLLM support require a generic KVEventPublisher struct change here, but SGLang/vLLM support does not need it? Is there something wrong/different with the way TRTLLM handles this compared to the other frameworks?

docs/features/multimodal/multimodal-kv-routing.md
```diff
@@ -7,7 +7,7 @@ subtitle: Route multimodal requests to workers with the best KV cache overlap
 
 ## Overview
 
-Multimodal KV routing extends Dynamo's KV-aware router to account for image content when computing cache overlap scores. An image hash (`mm_hash`) is computed per request — in the Rust frontend by default for vLLM backends, by vLLM's own processor when the chat-processor variant is enabled, or by a dedicated MM router worker for TRT-LLM backends — and included in per-block routing metadata. The KV router then selects the backend worker with the highest cache overlap, including overlap on image embedding blocks.
+Multimodal KV routing extends Dynamo's KV-aware router to account for image content when computing cache overlap scores. An image hash (`mm_hash`) is computed per request — in the Rust frontend by default (vLLM and TRT-LLM backends), or by vLLM's own processor when the chat-processor variant is enabled — and included in per-block routing metadata. The KV router then selects the backend worker with the highest cache overlap, including overlap on image embedding blocks.
```
krishung5: I think I missed adding sglang here in my previous PR, do you mind add it here along with TRTLLM 🙏

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -71,6 +71,33 @@
...
+_QWEN2_VL_FAMILY_MODEL_TYPES = frozenset({"qwen2_vl", "qwen2_5_vl", "qwen3_vl"})
+
+
+def _resolve_image_token_id(model_config: Any) -> Optional[int]:
+    """Resolve TRT-LLM's synthetic image-token marker (``vocab_size + 1``).
+
+    TRT-LLM writes this id at every image position in its KV-reuse events; the
+    KV-event publisher uses it to normalize image-token runs to their mm_hash
+    pad_value. Gated to the Qwen2-VL family, the only families where the
+    ``vocab_size + 1`` convention is confirmed; any other model_type returns
```
krishung5: Q: Re: ...the only families where the ``vocab_size + 1`` convention is confirmed;..., does this mean trtllm may do different convention for other model families?

lib/bindings/python/rust/llm/kv.rs
krishung5: Both vLLM and TRTLLM require reformatting the kv events before sending to the router. We use SGL's format so no struct change is needed for SGL. vLLM and TRTLLM emits the kv events in different paths: vllm emit the events through zmq so the normalization happens in zmq listener, while TRTLLM emit the events directly via the KVEventPublisher, so that's why we only see TRTLLM requires a struct change here.

I guess it'd be helpful for future readers if we could add some visualization to this PR as well. I had a graph [here](https://github.com/ai-dynamo/dynamo/pull/10999) but it only covers vllm and sgl.

KrishnanPrash: Nothing wrong or TRTLLM-specific, it's the event-ingestion path. vLLM/SGLang push their KV events to a ZMQ socket that dynamo subscribes to, and that ZMQ path already normalizes image-token runs via the `image_token_id` the constructor has long accepted (`KvEventSourceConfig::Zmq`). TRT-LLM's event API is pull-based (`get_kv_cache_events`), so dynamo re-publishes via the publish_stored push path instead of a ZMQ subscription. That push path was hardcoding `image_token_id=None`, so the same normalization the ZMQ path already does wasn't reachable from it. The struct field just carries the already-accepted constructor arg to the push path so it can reuse the identical normalizer; it's inert for the ZMQ (vLLM/SGLang) path.

tests/mm_router/test_trtllm_mm_router_e2e.py
```diff
@@ -4,10 +4,11 @@
 """End-to-end tests for multimodal KV routing with TRT-LLM.
```
krishung5: Can we rename this test to be `test_trtllm_mm_router_e2e.py` or similar? I think it's coming from historical changes but now it looks like this file only has the trtllm tests.

In addition to the e2e test, can we add some CPU-only gating test to make sure we know when upstream changes their protocol (vocab_size+1 in this case)? We have similar test for sglang here: https://github.com/ai-dynamo/dynamo/blob/7fc9e05a98a10c739391289db2da3f44a90c32ac/tests/serve/test_sglang_mm_hashes_protocol.py

tests/mm_router/test_trtllm_mm_router_e2e.py
```diff
@@ -370,20 +330,20 @@ def _check_text_only_overlap_repeated_prompt(
...
     overlap_2, total_2, _ = _send_request_get_overlap(
-        frontend_port, router_proc, payload, "same_3_images_req2"
+        frontend_port, frontend_proc, payload, "same_3_images_req2"
     )
     overlap_3, total_3, segment_3 = _send_request_get_overlap(
-        frontend_port, router_proc, payload, "same_3_images_req3"
+        frontend_port, frontend_proc, payload, "same_3_images_req3"
     )
 
     assert overlap_1 <= 1, (
```
krishung5: This test only asserts on the router's `[ROUTING] … overlap` debug logs. Could it reuse the shared [make_image_payload_cached_tokens](https://github.com/ai-dynamo/dynamo/blob/7fc9e05a98a10c739391289db2da3f44a90c32ac/tests/utils/multimodal.py#L83) helper (tests/utils/multimodal.py) so it validates the real cache hit, not just the routing decision? As written, a regression where routing is correct but the worker-side pad_value normalization silently mismatches would still pass.

krishung5 (commented):
Implementation lgtm overall! Left some comments mainly around the testing part.

components/src/dynamo/trtllm/workers/llm_worker.py
KrishnanPrash: Yes, and `a8baa0ad61` now handles exactly that. TRT-LLM writes two different image markers depending on the family: the Qwen2-VL family gets the synthetic `vocab_size + 1`, while other families (for example Kimi-K2.5) keep the real in-vocab id in their KV events. The resolver is now a two-bucket dispatch: a `vocab_size + 1` allowlist (Qwen) and a real-id allowlist (`{kimi_k25}`, resolved via the same frontend registry). Any unclassified family returns `None` (text-prefix routing) instead of guessing a marker, so the worker keys on whichever convention TRT-LLM actually emits.

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -71,6 +71,33 @@
 DEFAULT_KV_EVENT_BUFFER_MAX_SIZE = 100_000
 SPEC_DECODE_RUNTIME_KEY = "spec_decode"
 
+# Model families where TRT-LLM's synthetic image marker is confirmed to be
+# `vocab_size + 1`. Other multimodal families use a real in-vocab marker, so the
+# +1 convention is wrong for them and MM-aware KV routing must fall back.
+_QWEN2_VL_FAMILY_MODEL_TYPES = frozenset({"qwen2_vl", "qwen2_5_vl", "qwen3_vl"})
```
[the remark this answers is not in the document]
KrishnanPrash: Confirmed empirically: a KV-dump on Qwen2.5-VL-7B showed the `vocab_size + 1` marker (152065). `a8baa0ad61` keeps `qwen2_5_vl` in the explicit `_VOCAB_PLUS_ONE_MARKER_MODEL_TYPES` allowlist alongside `qwen2_vl` and `qwen3_vl`.

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -71,6 +71,37 @@
...
+        vocab_size = getattr(model_config, "vocab_size", None)
+        if vocab_size is None:
+            vocab_size = getattr(
+                getattr(model_config, "text_config", None), "vocab_size", None
+            )
+        if vocab_size is None:
+            return None
+        return int(vocab_size) + 1
+    except Exception:
+        return None
```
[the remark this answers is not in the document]
KrishnanPrash: The resolver has no broad handler. The only `except` is a narrow `ImportError` for the optional `mm-routing` binding (returns `None`). Config access is direct, so a genuinely malformed config surfaces its own error rather than being masked.

tests/mm_router/test_trtllm_mm_router_e2e.py
KrishnanPrash: Done. I removed the 715-line custom log-only E2E and moved this scenario into the standard `tests/serve/test_trtllm.py` deployment harness using `make_image_payload_cached_tokens`. The test now asserts both the real worker cache result and router behavior: the rc21 run reported `cached_tokens=25/25` on request 2 and mean repeat `router_kv_hit_rate=0.889` (threshold 0.5).

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -773,6 +884,7 @@ async def init_llm_worker(
                     zmq_topic="",
                     enable_local_indexer=config.enable_local_indexer,
                     kv_state_endpoint=config.kv_state_endpoint,
+                    image_token_id=image_token_id,
```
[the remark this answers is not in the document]
KrishnanPrash: Agreed that consolidation is outside the validated scope. Current `kvbm-consolidator` explicitly drops multimodal `BlockStored` events, so the supported path in this PR is aggregated + direct event publishing. I kept `image_token_id` on the ZMQ subscriber constructor so normalization remains correct when consolidation gains MM support, but the new launcher uses the direct path and the docs now call out consolidation as unsupported. The full rc21 E2E exercises the supported direct path.

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -67,10 +69,43 @@
...
+except ImportError:
+    resolve_routing_image_token_id = None  # type: ignore[assignment]
+
 # Default buffer size for kv cache events.
 DEFAULT_KV_EVENT_BUFFER_MAX_SIZE = 100_000
 SPEC_DECODE_RUNTIME_KEY = "spec_decode"
 
+# TRT-LLM 1.3.0rc21 keeps in-vocab image markers for these validated families.
+# Leave other families unresolved until their KV-event convention is verified.
+_MM_ROUTING_MODEL_TYPES = frozenset({"qwen2_vl", "qwen2_5_vl", "qwen3_vl", "kimi_k25"})
```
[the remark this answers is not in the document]
KrishnanPrash: Confirmed against the real HF configs, not inferred from names. Both `moonshotai/Kimi-K2.6` (snapshot `7eb5002f...`) and `nvidia/Kimi-K2.6-NVFP4` (snapshot `2fd3a800...`) report `model_type: "kimi_k25"`, `architectures: ["KimiK25ForConditionalGeneration"]`, and `media_placeholder_token_id: 163605`. K2.5 reports the same model type, so the single `kimi_k25` allowlist entry intentionally covers both versions.

krishung5 (approved):
LGTM, thanks!

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -968,6 +969,10 @@ impl KvEventPublisher {
             kv_block_size,
             dp_rank,
             warning_count: Arc::new(AtomicU32::new(0)),
+            // The unified backend does not run TRT-LLM multimodal, so this bridge
+            // carries no image marker; the non-unified path sets it via the
+            // constructor.
+            image_token_id: None,
```
rmccorm4, line 975: Why do we need this change for TRTLLM but didn't need it for vllm/sglang support?

(I might've already posted this somewhere but couldn't find it, so apologies if dupe)

indrajit96 (approved):
TRTLLM code owner approval

lib/bindings/python/rust/llm/kv.rs
KrishnanPrash: https://github.com/ai-dynamo/dynamo/pull/11737#discussion_r3606309792

PeaBrane (approved):
✅ AI reviewed.

PR #11747, opened 2026-07-16
fix(sglang): forward multimodal config and audio inputs

milesial:
## Summary

- forward Dynamo's `--enable-multimodal` setting to SGLang's parsed server arguments
- recognize `audio_url` alongside `image_url` and `video_url`
- extract and forward `audio_data` in aggregate, disaggregated prefill, and disaggregated decode paths
- reject unextracted audio content with the same guard used for other modalities

## Root cause

Dynamo consumes the shared `--enable-multimodal` option before SGLang parses its remaining arguments, so SGLang's `ServerArgs` did not receive the explicit setting. The request handlers also only extracted and forwarded image and video data, leaving audio URLs out of the engine call.

## Validation

- focused SGLang tests for audio extraction, multimodal guarding, and aggregate forwarding
- pre-commit hooks on all touched files

### DIFF
--- a/components/src/dynamo/sglang/args.py
+++ b/components/src/dynamo/sglang/args.py
@@ -405,6 +405,10 @@ async def parse_args(
     # otherwise fall back to default endpoints
     namespace = dynamo_config.namespace
 
+    # Dynamo's parser consumes --enable-multimodal; forward it to SGLang.
+    if dynamo_config.enable_multimodal:
+        parsed_args.enable_multimodal = True
+
     # If --embedding-worker is set, also set SGLang's --is-embedding flag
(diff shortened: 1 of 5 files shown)


Chokoyo (approved):
LGTM. On `--enable-multimodal`, SGLang already auto-enables multimodal for most VLMs, so forwarding is a no-op there, but this PR helps for the model SGLang explicitly disables multimodal by default (Gemma3, Llama4, and Step3VL), where the flag was previously swallowed by Dynamo's parser and never reached the engine. Thanks for the fix!

KrishnanPrash (approved):

PR #11748, opened 2026-07-16
fix(frontend): preserve SGLang reasoning delimiters

milesial:
## Summary

Preserve special delimiter tokens in the custom SGLang chat processor when either a tool-call or reasoning parser is active.

## Details

- Forward the active reasoning parser into backend output options.
- Keep parser delimiters visible during incremental detokenization so reasoning can populate `reasoning_content`.

## Validation

Not run; this PR intentionally contains no tests.

`components/src/dynamo/frontend/sglang_processor.py` and `components/src/dynamo/frontend/sglang_prepost.py`.

## Related Issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming output handling when reasoning or tool-call parsing is enabled.
  * Special tokens are now preserved consistently during incremental text generation.
  * Aligned behavior across worker-based and single-process request processing.

### DIFF
--- a/components/src/dynamo/frontend/sglang_prepost.py
+++ b/components/src/dynamo/frontend/sglang_prepost.py
@@ -950,9 +950,9 @@ class SglangStreamingPostProcessor:
             tool_call_parser_name
         )
         self._fast_plain_text = tool_call_parser is None and reasoning_parser is None
-        # Preserve special tokens when a tool call parser is active so
-        # delimiter tokens (e.g. <|tool_call|>) remain visible to the parser.
-        self._skip_special_tokens = tool_call_parser is None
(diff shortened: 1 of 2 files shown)


rmccorm4 (approved):
> Trailing EOS ids are handled by _strip_trailing_eos_token_ids (sglang_prepost.py:1096), but non-EOS special tokens are not filtered anywhere. This matches the behavior the tool-call path already had, so it is likely acceptable.

Agree with this analysis.

Additionally, both parsers would typically be set or neither at all. I don't think an edge case around having one parser set but not the other should be handled at this flag level anyways, it should be at a higher processing layer to handle that if it were to come up.

PR #11754, opened 2026-07-16
feat(vllm): add minimal grpc sidecar

connorcarpenter15:
## Summary

- add the vLLM sidecar crate at `lib/backend/vllm-sidecar`
- use the same sidecar terminology as the SGLang integration:
  - package/executable: `dynamo-vllm-sidecar`
  - Rust library: `dynamo_vllm_sidecar`
  - engine type: `VllmSidecarEngine`
- keep the sidecar as a standalone Rust executable composed with `dynamo_backend_common::run`
- connect to vLLM's released native gRPC `Generate` service
- support aggregated and NIXL prefill/decode generation
- leave the existing in-process vLLM backend unchanged

## Why

The sidecar is a process boundary: it implements Dynamo's `LLMEngine` contract and talks to vLLM's native gRPC server. vLLM runs independently while the sidecar owns Dynamo worker registration, request conversion, transport, cancellation, and disaggregation handoff.

The executable does not require a Python launcher or unreleased vLLM protocol extensions.

## Interfaces

Start vLLM with its released gRPC listener:

```bash
vllm-rs serve Qwen/Qwen3-0.6B --grpc-port 50051
```

Build and run the standalone sidecar from the Dynamo workspace:

```bash
cargo build --release -p dynamo-vllm-sidecar
./target/release/dynamo-vllm-sidecar \
    --vllm-endpoint http://127.0.0.1:50051 \
    --model-path Qwen/Qwen3-0.6B
```

## Protocol and feature scope

- vendor `rust/proto/vllm_grpc.proto` from vLLM v0.25.1 with its source revision and SHA-256
- support text and token generation, sampling, stopping, structured output, logprobs, cache options, priority, finish reasons, and usage
- forward `kv_transfer_params` unchanged between Protobuf `Struct` and JSON
- reject multimodal, LoRA, KV-aware routing, encode, beam-search, and `n > 1` requests before dispatch
- keep the plaintext, unauthenticated gRPC listener on loopback or a protected network

## Packaging scope

This PR does not add the executable to a wheel or runtime image. Distribution and process supervision will be designed and reviewed separately.

## Validation

- `cargo test -p dynamo-vllm-sidecar` — 11 unit tests, the executable-contract test, and doc tests pass
- `cargo test -p dynamo-backend-common args::tests` — 2 shared endpoint-normalization tests pass
- `cargo clippy -p dynamo-vllm-sidecar --all-targets -- -D warnings`
- `cargo metadata --locked --no-deps --format-version 1`
- `git diff --check`
- strict CODEOWNERS generation and 100% explicit-coverage validation
- vendored proto SHA-256 remains `7cccd0e1b2e54f189550e1090cc80321fc2bbd188c98a4e22701b28fdeb177b6`

## Related Issues


lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,573 @@
...
+        let count = prompt.num_prompt_tokens as usize;
+        if prompt.token_ids.len() != count
+            || prompt.logprobs.len() != count
+            || prompt.ranks.len() != count
+            || prompt.candidate_tokens.len() != count
+        {
+            return Err(client::protocol_error(
+                "prompt logprob arrays do not match num_prompt_tokens",
+            ));
+        }
```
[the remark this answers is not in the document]
connorcarpenter15: Verified against vLLM v0.25.1. rust/src/server/src/grpc/convert.rs::prompt_logprobs_to_proto explicitly inserts a zero logprob, zero rank, and empty candidate entry for prompt position zero. The parallel arrays therefore have num_prompt_tokens entries, and the fake server matches the released implementation. No change is needed.

lib/backend/vllm-remote/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+            .map_err(|error| invalid_argument(format!("invalid vLLM endpoint: {error}")))?
+            .connect_timeout(CONNECT_TIMEOUT);
+        let first = connect_until_ready(endpoint.clone()).await?;
+        let mut channels = vec![first];
+        let remaining = try_join_all((1..connections).map(|_| {
+            let endpoint = endpoint.clone();
+            async move { endpoint.connect().await }
+        }))
+        .await
+        .map_err(|error| cannot_connect(format!("failed to connect to vLLM: {error}")))?;
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in a83d31137e. Every pooled channel now uses connect_until_ready, so transient failures receive the same bounded retry behavior as the first channel.

.gitattributes
```diff
@@ -15,3 +15,6 @@
 # Exclude test data files from linguist language detection
 lib/llm/tests/data/** linguist-vendored
 lib/llm/tests/snapshots/** linguist-vendored
+
+# Preserve the released vLLM proto byte-for-byte, including its trailing blank line.
+lib/backend/vllm-remote/proto/vllm_grpc.proto whitespace=-blank-at-eof
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in a83d31137e. The vendored proto now uses -text to disable normalization, diff to keep it reviewable, and the existing whitespace override. Its SHA-256 remains identical to vLLM v0.25.1.

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,573 @@
...
+    validate_request(request, mode)?;
+
+    let sampling = &request.sampling_options;
+    let prompt_logprobs = request.output_options.prompt_logprobs;
+    let output_logprobs = request.output_options.logprobs;
+    let max_new_tokens = if mode.is_prefill() {
+        1
+    } else {
+        request.stop_conditions.max_tokens.unwrap_or(0)
+    };
```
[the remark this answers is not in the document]
connorcarpenter15: Verified against vLLM v0.25.1. A zero max_new_tokens leaves SamplingParams.max_tokens unset in rust/src/server/src/grpc/convert.rs. rust/src/text/src/lower.rs::resolve_max_tokens then uses the model generation default or max_model_len - prompt_len. The 20-token proto comment is stale, so no client fix is needed.

lib/backend/vllm-sidecar/README.md
```diff
@@ -0,0 +1,43 @@
...
+The initial protocol does not support multimodal input, LoRA, KV-aware data
+parallel routing, encode workers, beam search, or `n > 1`.
+
+## Run
+
+Start vLLM with its released gRPC listener:
+
+```bash
+vllm-rs serve Qwen/Qwen3-0.6B --grpc-port 50051
+```
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in a83d31137e. The README now states that the example listener is unauthenticated and plaintext, recommends loopback or a private interface, and requires network controls or a secure proxy for remote access.

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,207 @@
...
+        let client = self
+            .client
+            .get()
+            .ok_or_else(|| client::engine_shutdown("vLLM remote backend is not started"))?;
+        let request_id = ctx.id().to_string();
+        let proto_request =
+            build_generate_request(&request, &request_id, &self.model.served_name, self.mode)?;
+        let mut state = ResponseState::new(&request, self.mode);
+        let mut stream = client.generate_stream(proto_request).await?;
+        let cancel = self.cancel.clone();
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in a83d31137e. GenerateStream response-header setup now runs inside the cancellation select. The new cancellation_interrupts_pending_response_headers test covers cancellation before the server returns headers.

lib/backend/vllm-sidecar/proto/vllm_grpc.proto
tanmayv25, line 1: Why do we need to maintain a copy here? Why not pull the proto from the vllm commit?

connorcarpenter15: There are network issues with having to pull from a specific commit (e.g. when building with CI). We tried that with SGLang sidecar initially and had problems.

lib/backend/vllm-sidecar/src/args.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[command(flatten)]
+    pub common: CommonArgs,
+
+    /// vLLM gRPC endpoint as host:port or an http:// URL.
+    #[arg(long, env = "VLLM_GRPC_ENDPOINT")]
+    pub vllm_endpoint: String,
+
+    /// Hugging Face model ID or local path used for tokenization and templates.
+    #[arg(long)]
+    pub model_path: String,
```
alec-flowers, line 25: Why do we need this arg?

connorcarpenter15: vLLM doesn't expose any model or tokenizer metadata yet (we're in the process of adding this through `GetModelInfo` and `GetServerInfo`), so we need this flag for now so Dynamo can build it's local model card.

lib/backend/vllm-sidecar/src/args.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+pub(crate) fn normalize_endpoint(raw: &str) -> Result<String, String> {
```
alec-flowers: Do we do this the same as in SGLang? Could we use the same function from common or something like that?

lib/backend/vllm-sidecar/src/args.rs
```diff
@@ -0,0 +1,57 @@
...
+    let endpoint = raw.trim();
+    if endpoint.is_empty() {
+        return Err("vLLM gRPC endpoint must not be empty".to_string());
```
alec-flowers: Can we make this a better error message
"`--vllm-endpoint` must not be empty"

lib/backend/vllm-sidecar/src/args.rs
```diff
@@ -0,0 +1,57 @@
...
+use dynamo_backend_common::CommonArgs;
+
+#[derive(clap::Parser, Clone, Debug)]
+#[command(
+    name = "dynamo-vllm-sidecar",
+    about = "Run a Dynamo worker against vLLM's native gRPC service"
+)]
+pub(crate) struct Args {
+    #[command(flatten)]
+    pub common: CommonArgs,
```
alec-flowers, line 13: Just for me knowledge what is a commonArgs?

Also where is the namespace, component, and endpoint for sidecar specified? I assume component and endpoint are fixed but namespace

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::Duration;
+
+use dynamo_backend_common::{BackendError, DynamoError, ErrorType};
+use futures::future::try_join_all;
+use tokio::time::Instant;
+use tonic::transport::{Channel, Endpoint};
+
+use crate::proto as pb;
+
```
alec-flowers, line 14: The two timeouts should be configurable from env vars if someone wanted too.

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub(crate) fn connection_count(&self) -> usize {
+        self.channels.len()
+    }
+
+    pub(crate) async fn generate_stream(
+        &self,
+        request: pb::GenerateRequest,
+    ) -> Result<tonic::Streaming<pb::GenerateResponse>, DynamoError> {
+        let index = self.next.fetch_add(1, Ordering::Relaxed) % self.channels.len();
+        let mut client = pb::generate_client::GenerateClient::new(self.channels[index].clone());
```
alec-flowers: Tonic keeps a 4 MiB default decoding limit here. `PromptInfo` returns all prompt token IDs, logprobs, ranks, and candidates in one message, so a valid long-context prompt-logprob response can exceed the limit and fail locally. Can we set an explicit `max_decoding_message_size` (the SGLang sidecar uses 64 MiB) and add a streaming test with a response larger than 4 MiB?

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+        &self,
+        request: dynamo_backend_common::PreprocessedRequest,
+        ctx: GenerateContext,
+    ) -> Result<BoxStream<'static, Result<LLMEngineOutput, DynamoError>>, DynamoError> {
+        let client = self
+            .client
+            .get()
+            .ok_or_else(|| client::engine_shutdown("vLLM sidecar is not started"))?;
+        let request_id = ctx.id().to_string();
+        let proto_request = build_generate_request(&request, &request_id, self.mode)?;
```
alec-flowers: `generate` owns `PreprocessedRequest`, but the builder borrows it and then clones its largest fields. `ResponseState::new` only retains scalar metadata, so can we construct the state first and make `build_generate_request` consume both the request and request ID? That would let us move `token_ids`, stop/guided strings, cache salt, and the KV payload into protobuf instead of doing an O(prompt length) copy per request; it also removes the second request-ID allocation.

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+                state.reported_completion_tokens(),
+            ));
+            return Ok(Box::pin(futures::stream::once(async move { Ok(output) })));
+        };
+
+        Ok(Box::pin(async_stream::stream! {
+            loop {
+                tokio::select! {
+                    biased;
+                    _ = ctx.stopped() => {
```
alec-flowers: These cancellation futures are recreated after every streamed message. `ctx.stopped()` crosses an `async_trait` object boundary, so each call boxes a future and clones a watch receiver; `cancel.cancelled()` also repeatedly registers and drops its waiter. Can we create and pin both futures once before the loop, then select on `&mut` references?

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+                output.token_ids.len()
+            )));
+        }
+
+        self.completion_tokens = self.completion_tokens.saturating_add(output.num_tokens);
+        let mut mapped = LLMEngineOutput {
+            token_ids: if self.is_prefill {
+                Vec::new()
+            } else {
+                output.token_ids.clone()
```
alec-flowers: `convert` owns the decoded protobuf response, but this clones its token vector and the next line clones its text before the originals are dropped. The `then_some(output.text.clone())` expression is also eager, so prefill clones text that it immediately discards. Can we map/validate borrowed logprobs first and then move `output.token_ids` and `output.text` into `LLMEngineOutput`?

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+            }
+        }
+    }
+
+    let bypass_prefix_cache = bool_extra(extra, "bypass_prefix_cache")?
+        .or(bool_extra(extra, "skip_reading_prefix_cache")?)
+        .unwrap_or(false);
+    let caller_kv = extra
+        .and_then(|extra| extra.get("kv_transfer_params"))
+        .cloned();
```
alec-flowers: This deep-clones the KV JSON tree, after which `json_to_struct(&...)` recursively clones all keys and strings again into prost. Decode has the same issue with `prefill_result`, and the response direction borrows an owned prost `Struct` before dropping it. Since the request and response are owned, can we use consuming recursive JSON↔Struct converters that move keys/strings? Rebuilding containers is unavoidable; duplicating the source trees is not.

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+            for candidate in &prompt.candidate_tokens[index].tokens {
+                entries.insert(
+                    candidate.id.to_string(),
+                    serde_json::json!({
+                        "logprob": candidate.logprob,
+                        "rank": candidate.rank,
+                    }),
+                );
+            }
+            positions.push(serde_json::to_value(entries).map_err(|error| {
```
alec-flowers: This path repeatedly materializes the prompt-logprob payload: it first builds a `BTreeMap<String, Value>`, `to_value` reconstructs that as another JSON object, and the outer `json!({"prompt_logprobs": positions})` deep-copies the complete nested array again. Can we build `serde_json::Map`/`Value::{Object, Array}` directly? It would also be better to retain the compact owned `pb::PromptInfo` until the terminal chunk, so cancelled/failed generations never pay this conversion and in-flight requests do not retain the expanded JSON form.

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+    let count = output.token_ids.len();
+    if output.logprobs.len() != count
+        || output.ranks.len() != count
+        || output.candidate_tokens.len() != count
+    {
+        return Err(client::protocol_error(
+            "output logprob arrays do not match token_ids",
+        ));
+    }
+    let log_probs = output.logprobs.iter().copied().map(f64::from).collect();
```
alec-flowers: vLLM can legitimately emit `-inf` for impossible or masked tokens. Forwarding it unchanged is unsafe across the JSON transport: serde JSON turns non-finite values into `null`, which can fail typed logprob deserialization or drop the payload. The existing Python vLLM adapter clamps non-finite logprobs to a large finite-negative sentinel. Can we apply the same normalization to selected/candidate output and prompt logprobs, with `NEG_INFINITY` regression tests?

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+            token_ids: if self.is_prefill {
+                Vec::new()
+            } else {
+                output.token_ids.clone()
+            },
+            text: (!self.is_prefill && !output.text.is_empty()).then_some(output.text.clone()),
+            index: Some(0),
+            ..Default::default()
+        };
+        if self.expect_output_logprobs && !self.is_prefill {
```
alec-flowers: `expect_output_logprobs` loses the requested count, so `logprobs: 0` still populates `top_logprobs` with the selected-token entry. The existing vLLM adapter emits selected-token `log_probs` but suppresses `top_logprobs` when the requested count is zero. Can we retain the count and only assign `mapped.top_logprobs` when it is greater than zero?

alec-flowers (commented):
Focused review of the vLLM sidecar data path. I found several avoidable hot-path copies/allocations and a few production-scale logprob issues; details inline.

lib/backend/vllm-sidecar/src/client.rs
alec-flowers: Can we also name them something more clear. I coudn't figure out exactly how they were different 

Maybe
const CONNECT_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(30);
const CONNECT_RETRY_DEADLINE: Duration = Duration::from_secs(300);

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+use futures::stream::BoxStream;
+use tokio::sync::OnceCell;
+use tokio_util::sync::CancellationToken;
+
+use crate::args::{Args, normalize_endpoint};
+use crate::client::{self, VllmClient};
+use crate::convert::{ResponseState, build_generate_request};
+use crate::model::ConfiguredModel;
+
+const CONNECTIONS: usize = 8;
```
alec-flowers: How did we arrive at this? Can we also make this configurable from env var and maybe command line arg?

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+            endpoint,
+            connections,
+            model,
+            mode,
+            client: OnceCell::new(),
+            cancel: CancellationToken::new(),
+        }
+    }
+
+    pub fn from_env() -> Result<(Self, WorkerConfig), DynamoError> {
```
alec-flowers: `from_env()` is only used by `main`, while `from_args(Vec<String>)` currently has no callers; both just choose a Clap parsing entry point before delegating to `from_parsed`. Could we follow the existing SGLang/backend-common pattern and expose a single `from_args(Option<Vec<String>>)`? `None` would use `try_parse()` and `Some(argv)` would use `try_parse_from(argv)`. That removes a redundant public constructor, keeps parsing fallible as this `Result` API suggests, and avoids the slightly misleading `from_env` name since Clap is also reading argv. I think `new()` is still useful internally/tests, and `from_parsed()` can remain the shared validation/construction helper.

alec-flowers (commented):
One API-construction simplification follow-up.

lib/backend/vllm-sidecar/src/engine.rs
alec-flowers: Can we at minimum put a note at why its set to 8?

alec-flowers:
General comment of can we abstract certain helper features between sidecars maybe around the grpc side? We want to avoid things getting too bespoke. Ideally its just the minimal translation layer code. Even stuff like the args and parsing should be shared right?

lib/backend/vllm-sidecar/src/args.rs
alec-flowers: Can we add a comment that this is temporary with the reason we need it?

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+impl LLMEngine for VllmSidecarEngine {
+    async fn start(
+        &self,
+        _worker_id: u64,
+    ) -> Result<dynamo_backend_common::EngineConfig, DynamoError> {
+        if self.client.initialized() {
+            return Err(client::engine_shutdown("vLLM sidecar has already started"));
+        }
+        tracing::info!(
+            endpoint = %self.endpoint,
```
alec-flowers, line 135: Can we validate and sanitize this before logging it? `normalize_endpoint` currently preserves everything after `http://`, including userinfo, paths, and query strings, and this `info` event runs before `Endpoint::from_shared` validates the URI. A value such as `http://user:password@host:50051/?token=...` could therefore be persisted in logs. Ideally restrict the configured endpoint to plaintext scheme + authority and log only the sanitized scheme/host/port.

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,215 @@
...
+        let client = VllmClient::connect(&self.endpoint, self.connections).await?;
+        let connection_count = client.connection_count();
+        self.client
+            .set(client)
+            .map_err(|_| client::engine_shutdown("vLLM sidecar has already started"))?;
+        tracing::info!(
+            endpoint = %self.endpoint,
+            connections = connection_count,
+            model = %self.model.source,
+            "vLLM gRPC is ready"
```
alec-flowers: Could we avoid calling this “ready”? `VllmClient::connect` only establishes tonic transport channels; it does not invoke `Generate`, perform a health check, verify the peer implements vLLM’s service, or verify its loaded model. Also, `self.model.source` comes from `--model-path` and is the locally configured tokenization/template source, not server-discovered metadata. I suggest `"vLLM gRPC transport connected"`, renaming the field to `configured_model_source`, and adding `mode = %self.mode`. The common worker’s later `Serving ...` event can remain the true serving transition.

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+        client
+            .generate_stream(request)
+            .await
+            .map(tonic::Response::into_inner)
+            .map_err(|status| status_to_dynamo("GenerateStream", status))
+    }
+}
+
+async fn connect_until_ready(endpoint: Endpoint) -> Result<Channel, DynamoError> {
+    let deadline = Instant::now() + STARTUP_TIMEOUT;
```
alec-flowers: This is a per-connection deadline rather than an overall startup deadline. The first channel can consume five minutes, after which the remaining seven receive fresh five-minute deadlines; a 30-second connect attempt started near the deadline can also overrun it. Startup can therefore approach ten minutes while the returned error says “within 300s.” Can we create one deadline in `VllmClient::connect`, pass it to every pool connection, and cap each connect attempt and retry sleep at that shared deadline?

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,125 @@
...
+    loop {
+        match endpoint.clone().connect().await {
+            Ok(channel) => return Ok(channel),
+            Err(error) if Instant::now() < deadline => {
+                tracing::debug!(%error, "waiting for vLLM gRPC");
```
alec-flowers: Could we preserve the real transport cause and make retries identifiable? `tonic::transport::Error`'s `Display` is intentionally generic, so `%error` commonly emits only `transport error`; DNS, refusal, and timeout details are in its source chain/`Debug` representation. Please use `error = ?error` here and preserve the full source chain in the final returned error. It would also help to include the sanitized endpoint, pool slot, attempt, elapsed/remaining time, and retry interval. Once the first channel connects, seven loops can otherwise emit seven indistinguishable lines per second, so repeated identical failures should be throttled or summarized.

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+        };
+        if finish.num_output_tokens != self.completion_tokens {
+            return Err(client::protocol_error(format!(
+                "terminal num_output_tokens {} does not match streamed count {}",
+                finish.num_output_tokens, self.completion_tokens
+            )));
+        }
+
+        let reason = pb::finish_info::FinishReason::try_from(finish.finish_reason)
+            .map_err(|_| client::protocol_error("unknown finish reason"))?;
```
alec-flowers: Can this error include the observed numeric value, e.g. `unknown finish reason {}` with `finish.finish_reason`? This indicates a protobuf/server-version mismatch, and the raw enum value is the most useful evidence for diagnosing it.

lib/backend/vllm-sidecar/src/convert.rs
```diff
@@ -0,0 +1,576 @@
...
+        if !self.expect_prompt_logprobs {
+            return Ok(());
+        }
+        let count = prompt.num_prompt_tokens as usize;
+        if prompt.token_ids.len() != count
+            || prompt.logprobs.len() != count
+            || prompt.ranks.len() != count
+            || prompt.candidate_tokens.len() != count
+        {
+            return Err(client::protocol_error(
```
alec-flowers: When this protocol validation fails, can the error report the expected count and each observed array length (`token_ids`, `logprobs`, `ranks`, and `candidate_tokens`)? The corresponding output-logprob check below should do the same. These errors are already surfaced with the request ID by the common adapter, so retaining the exact wire dimensions would make version/shape mismatches immediately diagnosable.

alec-flowers (commented):
Focused logging/diagnostics review. The direct level choices make sense: lifecycle transitions belong at `info`, and expected connection retries belong at `debug`. The concerns inline are about making those events truthful, safe, and useful during an incident.

Two related cross-layer observations that do not attach cleanly to changed lines:

- Setup-time `engine.generate()` failures are mapped to `setup_failed` in `backend-common/src/adapter.rs`, but the full error is not logged there despite the comment saying it is available in the trace-correlated log stream. The mid-stream path already logs `request_id` and the typed error; setup failures should get the equivalent `debug` event in the common adapter.
- `backend-common` reports the elapsed `engine.start()` duration as `model_load_time_seconds`. For this remote sidecar that is gRPC connection-pool setup time, not vLLM model loading, so the log/metric name is misleading.

I would not add local startup `error!` or cleanup logs: returning one enriched startup error avoids log-and-return duplication, and the common worker already owns cleanup success/failure logging.

alec-flowers:
Non-blocking observability follow-up: I audited what Prometheus coverage is already supplied by the common worker before suggesting sidecar-specific metrics.

The runtime already provides request count/duration/inflight, bytes, cancellations, categorized handler errors, queue and permit saturation, capacity rejection, frontend-to-worker transit, worker time-to-first-response, and Tokio saturation metrics. vLLM should remain the source of truth for scheduler/GPU/KV state, TTFT/ITL, token throughput, and model execution latency. I don't think this sidecar should duplicate those.

The concrete gap is mid-stream engine failures. Setup failures are counted as `dynamo_component_errors_total{error_type="generate"}`, but gRPC read errors, malformed responses, and premature EOF are yielded as typed error frames; the common response pump notices `encoded.is_error` without incrementing the error counter.

I opened [DYN-3556](https://linear.app/nvidia/issue/DYN-3556/count-mid-stream-engine-failures-in-unified-backend-prometheus-metrics) for the minimal common-layer fix:

- reuse `dynamo_component_errors_total`
- add bounded `error_type="engine_stream"`
- increment exactly once for the first typed mid-stream error
- leave detailed gRPC status/protocol causes in logs and traces
- avoid request IDs, endpoints, channel indices, model paths, or free-form error labels

That gives us a failure-rate signal against the existing request counter and fixes vLLM, SGLang, and other unified backends consistently. I deliberately excluded sidecar TTFT/ITL, request latency, inflight, pool, connection-state, startup-retry, and per-chunk conversion metrics because they are duplicated, misleading about attribution, unavailable during failed startup, or add hot-path overhead.

One limitation: this remains traffic-driven. Detecting an idle-but-dead vLLM requires a real active health probe, not a potentially stale `connected=1` gauge.

lib/backend/vllm-sidecar/src/args.rs
connorcarpenter15: Fixed.

lib/backend/vllm-sidecar/src/args.rs
connorcarpenter15: Fixed.

lib/backend/vllm-sidecar/src/args.rs
connorcarpenter15: Added.

lib/backend/vllm-sidecar/src/args.rs
connorcarpenter15: `CommonArgs` contains the shared Dynamo worker options used by Rust backends. The sidecar flattens these into its CLI. It includes `namespace`, `component`, and `endpoint`, which default to `dynamo`, `backend`, and `generate`, but can be overridden through CLI flags or `DYN_*` environment variables. These identify the logical Dynamo endpoint; `--vllm-endpoint` separately identifies the vLLM gRPC server.

lib/backend/vllm-sidecar/src/engine.rs
connorcarpenter15: I arrived at this through my performance testing of the sidecars and gRPC connections. Setting it to 8 was enough to avoid throttling issues at higher concurrencies.

I'll put a note about why it's set to 8 and add a configurable env var/CLI arg.

lib/backend/vllm-sidecar/src/client.rs
connorcarpenter15: Done.

alec-flowers (approved):

lib/backend-common/src/args.rs
```diff
@@ -7,12 +7,63 @@
...
 
+use std::num::NonZeroUsize;
 use std::path::PathBuf;
+use std::time::Duration;
 
 use clap::Args;
 
 use crate::disagg::DisaggregationMode;
 
+const DEFAULT_GRPC_CONNECT_ATTEMPT_TIMEOUT_SECS: u64 = 30;
```
tanmayv25, line 18: Introducing gRPC here doesn't seem clean. As the plan is for experimental inference engine to use backend-common crate directly, most in-process engines should not be exposed to gRPC.

backend-common should stay gRPC free; gRPC helpers live in a new thin `dynamo-sidecar-grpc` crate.

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,228 @@
...
+
+pub(crate) fn engine_shutdown(message: impl Into<String>) -> DynamoError {
+    backend(BackendError::EngineShutdown, message)
+}
+
+fn cannot_connect(message: impl Into<String>) -> DynamoError {
+    backend(BackendError::CannotConnect, message)
+}
+
+pub(crate) fn status_to_dynamo(rpc: &str, status: tonic::Status) -> DynamoError {
```
tanmayv25, line 212: We can move this to  dynamo-sidecar-grpc as well.

lib/backend/vllm-sidecar/src/engine.rs
```diff
@@ -0,0 +1,229 @@
...
+                _ = shutdown.cancelled() => {}
+            }
+        });
+        let stream = tokio::select! {
+            biased;
+            _ = cancelled.as_mut() => None,
+            result = client.generate_stream(proto_request) => Some(result?),
+        };
+        let Some(mut stream) = stream else {
+            let output = LLMEngineOutput::cancelled().with_usage(usage(
```
tanmayv25: This LLMEngineOutput::cancelled().with_usage(...) is built identically here and in the loop below (line 186). Extract a small cancelled(&state) helper.

lib/backend/vllm-sidecar/src/client.rs
```diff
@@ -0,0 +1,228 @@
...
+    deadline: Instant,
+) -> Result<Channel, DynamoError> {
+    let started = Instant::now();
+    let mut attempt = 0_u64;
+    let mut last_error = None;
+    let mut last_logged_at = None;
+    let mut last_logged_error = None;
+    let mut suppressed_attempts = 0_u64;
+    loop {
+        let remaining = deadline.saturating_duration_since(Instant::now());
```
tanmayv25, line 90: The retry throttling logic can live in dynamo-sidecar-grpc. Other sidecars would inherit the same behavior.

tanmayv25 (commented):
I’d strongly prefer a dedicated `dynamo-sidecar-grpc` crate over a grpc feature flag on backend-common: it keeps transport stuff out of the common crate entirely, and we already have multiple consumers (sglang, trtllm, and the vllm sidecar ). There may be in-process consumers of unified backend interface. Additionally, imagine someone writing http based sidecar. 

The proposed layering:

- backend-common — engine contract and common error types.

- sidecar-grpc — transport + gRPC‑specific error helpers.

- {sglang,trtllm,vllm}-sidecar — protocol, proto, and engine‑specific message shapes.

tanmayv25 (approved):
Will investigate additional sidecar layer in follow-up PR once we have trtllm enabled as well.

nv-anants (approved):
.github and toml file changes look good!

jthomson04 (approved):

PR #11776, opened 2026-07-16
perf(kv-router): batch event-plane KV events

jthomson04:
## Overview

Native KV event batches can produce multiple ordered `RouterEvent`s after filtering, deduplication, and coalescing. The event plane currently serializes and publishes each resulting event as a separate message, multiplying envelope, serialization, transport, and receiver overhead.

This change publishes ordered `RouterEvent` vectors on both ZMQ and NATS Core, bounded by represented block count and event count rather than serialized byte size. The deprecated durable JetStream path retains its existing singleton payload.

## Summary

- collect the `RouterEvent`s emitted during an event-processor turn without changing coalescing, deduplication, event IDs, or local-indexer application
- partition ordered events before serialization at 128 events or 8,192 represented blocks, whichever limit is reached first, while always allowing at least one complete event per message
- count stored blocks and removed block hashes toward the block cap; `Cleared` events contribute zero blocks
- publish each resulting `Vec<RouterEvent>` through the existing production event-plane codec
- use the same batched payload and frontend decoder for ZMQ and NATS Core
- retain singleton `RouterEvent` publication and consumption for durable JetStream

## Details

The partitioner is allocation-free and never splits a `RouterEvent`. A normal message contains at most 128 events and 8,192 represented blocks. If one event represents more than 8,192 blocks, it is published intact as its own message. There is no serialized-byte-size check, speculative serialization, or subset reserialization.

This is a deliberate hard event-plane wire change: ZMQ and NATS Core publishers and subscribers now encode and decode `Vec<RouterEvent>` only. They must be upgraded together; rolling between singleton and vector payload versions is unsupported. The deprecated durable JetStream path, including the standalone-indexer flow that uses it, remains singleton-based.

NATS Core defaults to a 1 MiB `max_payload`. An unpartitioned adversarial batch containing 8,192 one-block events exceeds that limit because named MessagePack repeats event-level metadata. Regression tests reproduce that case and verify that every 128-event partition remains below 1 MiB, both for plain stored blocks and for one multimodal object with one offset per block. Dense 8,192-block production-shaped batches are also verified below the limit. Multimodal metadata is not intrinsically bounded, so an exceptional single event can still exceed the configured transport limit. That edge case is accepted: the failed publish is logged and dropped under the existing at-most-once semantics, with no retry, requeueing, backpressure, or byte-based splitting.

When one ordered output partitions into multiple event-plane messages, every partition is attempted even if an earlier publish fails. Each failed partition is logged with its transport, event count, and event-ID range, and the returned error reports the exact number of failed publishes and dropped events.

The existing queue sizes, overflow behavior, transport envelopes, publisher sequence numbers, worker event IDs, recovery behavior, shutdown cancellation between decoded events, and public APIs remain unchanged. There is no compatibility fallback or new runtime telemetry.

1. `lib/llm/src/kv_router/publisher/sinks.rs` for event/block partitioning, production-codec publication, and failure reporting
2. `lib/llm/src/kv_router/publisher/event_processor.rs` for ordered output collection and publication boundaries
3. `lib/llm/src/kv_router/indexer/recovery/subscriber.rs` for shared NATS Core/ZMQ batch decoding and ordered frontend delivery

## Related Issues

## Validation

- `cargo fmt --package dynamo-llm -- --check`
- `cargo test -p dynamo-llm kv_router::publisher::tests --lib` (78 passed)
- `cargo check -p dynamo-llm --bench kv_router_bench --features kv-router-stress`
- `cargo clippy -p dynamo-llm --lib --tests --no-deps -- -D warnings`
- `git diff --check`
- coverage includes event/block partitioning, both production caps, oversized singleton events, adversarial sparse-event NATS envelope sizes, ordering, deduplication, JetStream singleton behavior, aggregate partial-publish failures, and production-codec publisher/subscriber roundtripping

The existing mocker C512 topology was not used as a performance gate because `dynamo.mocker` feeds singleton native lists directly into the publisher and therefore cannot measure the intended native-batch message reduction.

## Summary by CodeRabbit

- **New Features**
  - Improved event delivery by publishing related updates in efficient batches.
  - Preserved event ordering while respecting delivery size limits.
  - Added support for processing multiple incoming events together.

- **Bug Fixes**
  - Improved resilience when receiving event updates, allowing processing to continue after recoverable failures.
  - Ensured pending updates are flushed during shutdown or cancellation.

- **Tests**
  - Added coverage for event batching, encoding, ordering, size limits, and recovery behavior.

### DIFF
--- a/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
@@ -10,9 +10,12 @@ use dynamo_kv_router::{
     protocols::{KV_EVENT_SUBJECT, RouterEvent},
 };
 use dynamo_runtime::{
-    component::Component, discovery::EventTransportKind, prelude::*,
-    transports::event_plane::EventSubscriber,
+    component::Component,
+    discovery::EventTransportKind,
+    prelude::*,
+    transports::event_plane::{EventEnvelope, EventSubscriber, TypedEventSubscriber},
 };
+use std::sync::Arc;
 use tokio_util::sync::CancellationToken;
 
 /// Start a simplified background task for event consumption using the event plane.
@@ -38,10 +41,9 @@ async fn start_kv_router_background_event_plane(
     // This ensures no events are lost between the initial dump fetch and the
     // subscription becoming active — the tree state at fetch time is guaranteed
     // to be a subset of what the subscription will deliver.
-    let mut subscriber =
+    let subscriber =
         EventSubscriber::for_component_with_transport(&component, KV_EVENT_SUBJECT, transport_kind)
-            .await?
-            .typed::<RouterEvent>();
+            .await?;
 
     // Brief delay to let the subscription fully establish with the NATS server
     // before recovery fetches the initial dump from workers.
@@ -81,46 +83,103 @@ async fn start_kv_router_background_event_plane(
     }
 
     tokio::spawn(async move {
-        loop {
-            tokio::select! {
-                biased;
-
-                _ = cancellation_token.cancelled() => {
-                    tracing::debug!("KV Router event plane background task received cancellation signal");
-                    break;
-                }
-
-                // Handle event consumption from event plane subscription
-                Some(result) = subscriber.next() => {
-                    let (envelope, event) = match result {
-                        Ok((envelope, event)) => (envelope, event),
-                        Err(e) => {
-                            tracing::warn!("Failed to receive RouterEvent from event plane: {e:?}");
-                            continue;
-                        }
-                    };
-
-                    tracing::trace!(
-                        "Received event from publisher {} (seq {})",
-                        envelope.publisher_id,
-                        envelope.sequence
-                    );
-
-                    tracing::trace!(
-                        "Forwarding live event to recovery coordinator for worker {} dp_rank {} event_id {}",
-                        event.worker_id,
-                        event.event.dp_rank,
-                        event.event.event_id
-                    );
-                    worker_query_client.handle_live_event(event).await;
-                }
+        let subscriber = match transport_kind {
+            EventTransportKind::Nats => {
+                RouterEventSubscriber::Singleton(subscriber.typed::<RouterEvent>())
             }
-        }
-
-        tracing::debug!("KV Router event plane background task exiting");
+            EventTransportKind::Zmq => {
+                RouterEventSubscriber::Batch(subscriber.typed::<Vec<RouterEvent>>())
+            }
+        };
+        consume_events(subscriber, worker_query_client, cancellation_token).await;
     });
 
     Ok(())
+}
+
+enum RouterEventSubscriber {
+    Singleton(TypedEventSubscriber<RouterEvent>),
+    Batch(TypedEventSubscriber<Vec<RouterEvent>>),
+}
+
+enum RouterEventPayload {
+    Singleton(RouterEvent),
+    Batch(Vec<RouterEvent>),
+}
+
+impl RouterEventSubscriber {
+    async fn next(&mut self) -> Option<Result<(EventEnvelope, RouterEventPayload)>> {
+        match self {
+            Self::Singleton(subscriber) => subscriber.next().await.map(|result| {
+                result.map(|(envelope, event)| (envelope, RouterEventPayload::Singleton(event)))
+            }),
+            Self::Batch(subscriber) => subscriber.next().await.map(|result| {
+                result.map(|(envelope, events)| (envelope, RouterEventPayload::Batch(events)))
+            }),
+        }
+    }
+}
+
+async fn consume_events(
+    mut subscriber: RouterEventSubscriber,
+    worker_query_client: Arc<WorkerQueryClient>,
+    cancellation_token: CancellationToken,
+) {
+    loop {
+        tokio::select! {
+            biased;
+
+            _ = cancellation_token.cancelled() => {
+                tracing::debug!("KV Router event plane background task received cancellation signal");
+                break;
+            }
+
+            Some(result) = subscriber.next() => {
+                let (envelope, payload) = match result {
+                    Ok((envelope, payload)) => (envelope, payload),
+                    Err(e) => {
+                        tracing::warn!("Failed to receive RouterEvent payload from event plane: {e:?}");
+                        continue;
+                    }
+                };
+
+                tracing::trace!(
+                    event_count = match &payload {
+                        RouterEventPayload::Singleton(_) => 1,
+                        RouterEventPayload::Batch(events) => events.len(),
+                    },
+                    "Received event payload from publisher {} (seq {})",
+                    envelope.publisher_id,
+                    envelope.sequence
+                );
+                match payload {
+                    RouterEventPayload::Singleton(event) => {
+                        forward_live_event(&worker_query_client, event).await;
+                    }
+                    RouterEventPayload::Batch(events) => {
+                        for event in events {
+                            if cancellation_token.is_cancelled() {
+                                break;
+                            }
+                            forward_live_event(&worker_query_client, event).await;
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    tracing::debug!("KV Router event plane background task exiting");
+}
+
+async fn forward_live_event(worker_query_client: &Arc<WorkerQueryClient>, event: RouterEvent) {
+    tracing::trace!(
+        "Forwarding live event to recovery coordinator for worker {} dp_rank {} event_id {}",
+        event.worker_id,
+        event.event.dp_rank,
+        event.event.event_id
+    );
+    worker_query_client.handle_live_event(event).await;
 }
 
 /// Helper to decide which subscriber (JetStream or Event Plane) to start based on config
--- a/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
@@ -104,11 +104,17 @@ async fn consume_events(
                 break;
             }
 
-            Some(result) = subscriber.next() => {
+            result = subscriber.next() => {
+                let Some(result) = result else {
+                    tracing::warn!("KV Router event-plane stream closed");
+                    break;
+                };
                 let (envelope, events) = match result {
                     Ok((envelope, events)) => (envelope, events),
                     Err(e) => {
-                        tracing::warn!("Failed to receive RouterEvent payload from event plane: {e:?}");
+                        tracing::warn!(
+                            "Failed to receive RouterEvent batch from event plane; publisher/subscriber versions or payload formats may not match: {e:?}"
+                        );
                         continue;
                     }
                 };
--- a/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
@@ -13,7 +13,7 @@ use dynamo_runtime::{
     component::Component,
     discovery::EventTransportKind,
     prelude::*,
-    transports::event_plane::{EventEnvelope, EventSubscriber, TypedEventSubscriber},
+    transports::event_plane::{EventSubscriber, TypedEventSubscriber},
 };
 use std::sync::Arc;
 use tokio_util::sync::CancellationToken;
@@ -43,7 +43,8 @@ async fn start_kv_router_background_event_plane(
     // to be a subset of what the subscription will deliver.
     let subscriber =
         EventSubscriber::for_component_with_transport(&component, KV_EVENT_SUBJECT, transport_kind)
-            .await?;
+            .await?
+            .typed::<Vec<RouterEvent>>();
 
     // Brief delay to let the subscription fully establish with the NATS server
     // before recovery fetches the initial dump from workers.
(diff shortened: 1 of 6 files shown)


jthomson04 (commented):
Reviewed the batching change end-to-end (codec, event-plane transport, bindings, bench, recovery paths). The core design is sound: ordering, coalescing, dedup, event IDs, and local-indexer application are genuinely preserved; the partitioner always yields at least one event per batch (no empty-slice loop) and oversized events go out intact; envelope sequences stay contiguous so gap detection is unaffected; Python/C bindings both route through `KvEventPublisher` so all real publishers emit the new format. Unit coverage of the partitioner is thorough.

**One issue that should be fixed in this PR:**

`lib/llm/benches/kv_router_bench.rs` is silently broken by the wire change. `build_tree_via_nats` (~line 747) and the replay path (~line 1164) hand-encode *singleton* `RouterEvent` payloads via `encode_event_with_envelope` and publish directly to the event-plane subject the frontend consumes. Post-PR the subscriber decodes `Vec<RouterEvent>`; a msgpack map cannot decode as an array, so every bench event will be warn-logged and dropped — the bench waits 2s, builds an empty tree, and produces meaningless numbers with no hard failure. Fix is one line at each call site: encode `std::slice::from_ref(&event)` instead of `&event`.

**Risk decisions worth explicit sign-off (details in inline comments):**
- block-count cap with no byte bound vs. NATS Core's 1 MB default `max_payload` (multimodal `mm_extra_info` is the unbounded term)
- same-version wire change with no rolling-upgrade story — at minimum make the decode-failure log point at version skew

**Smaller notes:**
- `consume_events` checks cancellation between events of a decoded batch and drops the remainder — fine for shutdown, but it makes a delivered message non-atomic, unlike pre-PR.
- The `first_error` accumulation loop is duplicated verbatim in both `publish_events` impls; a small helper would remove the repetition.
- Test gap: the msgpack roundtrip test re-implements decode with `rmp_serde::from_slice` rather than exercising `EventPublisher::publish_bytes` → `TypedEventSubscriber` through the actual codec — identical today, but it would not catch codec drift (moot if the `publish(&batch)` suggestion is taken).

lib/llm/src/kv_router/publisher/sinks.rs
```diff
@@ -8,20 +8,88 @@ use anyhow::Result;
...
 use dynamo_runtime::transports::nats::NatsQueue;
 
 use crate::kv_router::KV_EVENT_SUBJECT;
 
 pub(super) struct EventPlanePublisher(pub(super) EventPublisher);
 
-impl RouterEventSink for EventPlanePublisher {
-    fn publish_event(&self, event: &RouterEvent) -> impl Future<Output = Result<()>> + Send {
-        self.0.publish(event)
+pub(super) const MAX_EVENT_PLANE_KV_EVENT_BATCH_BLOCKS: usize = 8_192;
```
jthomson04, line 28: The cap is count-based with no byte bound. 8,192 plain stored blocks is ~400–500 KB of named msgpack — safely under NATS Core's 1 MB default `max_payload` — but `KvCacheStoredBlockData.mm_extra_info` carries a `Vec<BlockMmObjectInfo>` with unbounded offset lists per block, so a multimodal-heavy batch can exceed 1 MB. When that happens the entire multi-event message is rejected at publish and lost with a single error log. Pre-PR the worst-case message was one coalesced event (~`DEFAULT_MAX_BATCH_BLOCKS` = 128 blocks), roughly 64× smaller.

Suggestions (any one): a byte-aware fallback (halve the batch and retry on payload-too-large), a lower cap when `mm_extra_info` is present, or at minimum a comment documenting the max_payload assumption behind 8,192.

lib/llm/src/kv_router/publisher/sinks.rs
```diff
@@ -8,20 +8,88 @@ use anyhow::Result;
...
+        first_error.map_or(Ok(()), Err)
     }
 }
 
+impl RouterEventBatchSink for EventPlanePublisher {
+    async fn publish_events(&self, events: &[RouterEvent]) -> Result<()> {
+        let mut first_error = None;
+        for batch in event_plane_event_batches(events, MAX_EVENT_PLANE_KV_EVENT_BATCH_BLOCKS) {
+            let payload = encode_event_plane_batch(batch)?;
+            if let Err(error) = self.0.publish_bytes(payload).await
```
jthomson04: Observability regression: previously each failed publish logged its own error; now only the first error per turn survives, and the single log line in `publish_output` reports the whole `event_count`. Partial loss (batch 2 of 3 failed) is indistinguishable from total loss. Consider logging per-failed-batch here, or including a failed-batch count in the returned error.

lib/llm/src/kv_router/publisher/sinks.rs
```diff
@@ -8,20 +8,88 @@ use anyhow::Result;
...
+        }
+
+        let batch = &events[batch_start..batch_end];
+        batch_start = batch_end;
+        Some(batch)
+    })
+}
+
+/// Encode one complete ordered event-plane batch.
+pub(super) fn encode_event_plane_batch(events: &[RouterEvent]) -> Result<Vec<u8>> {
```
jthomson04: This hardcodes `rmp_serde::to_vec_named`, duplicating the codec's wire format outside the `Codec` abstraction. `self.0.publish(&batch)` serializes a `&[RouterEvent]` identically to `Vec<RouterEvent>` (still exactly one serialization per message, via `codec.encode_payload`) and would delete this helper plus the coupling. If a second codec variant is ever added, this call site silently diverges from what `typed::<Vec<RouterEvent>>()` decodes.

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
```diff
@@ -81,46 +84,62 @@ async fn start_kv_router_background_event_plane(
...
-                        event.worker_id,
-                        event.event.dp_rank,
-                        event.event.event_id
-                    );
-                    worker_query_client.handle_live_event(event).await;
+            Some(result) = subscriber.next() => {
+                let (envelope, events) = match result {
+                    Ok((envelope, events)) => (envelope, events),
+                    Err(e) => {
+                        tracing::warn!("Failed to receive RouterEvent payload from event plane: {e:?}");
```
jthomson04: During any mixed-version window (old worker + new frontend, or vice versa) every KV event lands here and routing quality degrades fleet-wide with only this generic warn. If the no-compat stance matches project policy, at least mention a possible publisher/subscriber version mismatch in the message so it's diagnosable from logs. A cheap alternative: on decode failure, try singleton `RouterEvent` as a fallback for one release — makes rollout order-independent at near-zero cost.

lib/llm/src/kv_router/publisher/sinks.rs
```diff
@@ -8,20 +8,88 @@ use anyhow::Result;
...
+            let payload = encode_event_plane_batch(batch)?;
+            if let Err(error) = self.0.publish_bytes(payload).await
+                && first_error.is_none()
+            {
+                first_error = Some(error);
+            }
+        }
+        first_error.map_or(Ok(()), Err)
+    }
+}
```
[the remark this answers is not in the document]
jthomson04: Addressed in ff7d395084. I kept the 8,192-block cap and added full-envelope regression tests through the production MsgpackCodec for both 8,192 plain stored blocks and 8,192 stored blocks with one multimodal object and one offset per block; both remain below the NATS Core default 1 MiB max_payload. The PR now documents that arbitrary multimodal metadata remains unbounded and that an oversize publish is logged and dropped under the existing best-effort semantics, with no byte splitting or retry.

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
```diff
@@ -38,10 +41,10 @@ async fn start_kv_router_background_event_plane(
     // This ensures no events are lost between the initial dump fetch and the
     // subscription becoming active — the tree state at fetch time is guaranteed
     // to be a subset of what the subscription will deliver.
-    let mut subscriber =
+    let subscriber =
         EventSubscriber::for_component_with_transport(&component, KV_EVENT_SUBJECT, transport_kind)
             .await?
-            .typed::<RouterEvent>();
+            .typed::<Vec<RouterEvent>>();
```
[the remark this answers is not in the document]
jthomson04: Confirmed as a deliberate hard wire break in ff7d395084 and documented explicitly in the PR. ZMQ and NATS Core publishers and subscribers encode and decode Vec<RouterEvent> only and must be upgraded together; there is no singleton compatibility fallback. The durable JetStream path, including its standalone-indexer flow, remains singleton-based.

lib/llm/src/kv_router/publisher/sinks.rs
jthomson04: Addressed in ff7d395084. The cap remains count-based, with a source comment and PR documentation explaining the NATS 1 MiB assumption and the accepted unbounded multimodal-metadata edge case. Production-codec envelope tests cover 8,192 plain blocks and one multimodal object plus one offset per stored block. Exceptional oversize messages retain the existing logged at-most-once drop behavior; no byte-aware split or retry was added.

lib/llm/src/kv_router/publisher/sinks.rs
jthomson04: Addressed in ff7d395084. Every ordered sub-batch is still attempted after a failure. Each failed event-plane publish now logs the transport, failed event count, and event-ID range, while the aggregate error reports the exact failed-publish and dropped-event counts. The singleton sink also logs each failed event. No retry, requeue, or new metric was added.

lib/llm/src/kv_router/publisher/sinks.rs
jthomson04: Addressed in ff7d395084. EventPlanePublisher now calls EventPublisher::publish(&batch), and the duplicate rmp-serde helper was removed. I also added a process-local EventPublisher-to-TypedEventSubscriber roundtrip that verifies the production codec transports the ordered vector.

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
jthomson04: Addressed in ff7d395084. The decode warning now explicitly calls out publisher/subscriber version or payload-format mismatch. The intentional Vec-only wire break remains; no singleton fallback was added.

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
```diff
@@ -81,46 +84,62 @@ async fn start_kv_router_background_event_plane(
...
+                    forward_live_event(&worker_query_client, event).await;
                 }
             }
         }
+    }
 
-        tracing::debug!("KV Router event plane background task exiting");
-    });
+    tracing::debug!("KV Router event plane background task exiting");
+}
```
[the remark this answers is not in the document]
jthomson04: Addressed in ff7d395084. The select branch now matches the Option explicitly, warns when the event-plane stream closes, and terminates the consumer. Existing decode-error handling and cancellation between events are unchanged.

jthomson04:
Addressed the review feedback in `ff7d395084`.

- Switched event-plane batches to `EventPublisher::publish(&batch)` and removed the duplicate MessagePack helper.
- Added production-codec coverage for the full 8,192-block envelope, including the ordinary stored-block case and one multimodal object/offset per block; both stay below the NATS Core default 1 MiB payload limit. The PR now documents the accepted unbounded-metadata edge case.
- Kept the deliberate `Vec<RouterEvent>`-only wire break and documented the required lockstep ZMQ/NATS Core upgrade. Durable JetStream and its standalone-indexer flow remain singleton-based.
- Restored per-failure visibility while retaining best-effort delivery: later ordered batches are attempted, each failure is logged with its transport/count/range, and the aggregate error reports exact failed-publish and dropped-event counts. There is still no retry, requeueing, backpressure, byte splitting, or new metric.
- Explicitly handle a closed subscriber stream and improve the decode warning for version/payload mismatch.
- Fixed both out-of-diff benchmark singleton encodes. The benchmark also had stale prompt-renderer imports on the current branch; those are corrected so the target compiles.
- Preserved cancellation checks between decoded events, including dropping the unprocessed remainder during shutdown.

No DEP was filed. This remains a scoped internal event-payload optimization: public APIs, the event envelope, transport selection/topology, queueing, and delivery semantics are unchanged. The non-rolling-compatible payload contract and operational upgrade requirement are now explicit in the PR.

Validation:

- `cargo test -p dynamo-llm kv_router::publisher::tests --lib` — 77 passed
- `cargo check -p dynamo-llm --bench kv_router_bench --features kv-router-stress`
- `cargo clippy -p dynamo-llm --lib --tests --no-deps -- -D warnings`
- `cargo fmt --package dynamo-llm -- --check`
- `git diff --check`

lib/llm/src/kv_router/publisher/sinks.rs
```diff
@@ -8,20 +8,130 @@ use anyhow::Result;
...
+
+        let mut batch_end = batch_start;
+        let mut batch_blocks = 0usize;
+        while let Some(event) = events.get(batch_end) {
+            let event_blocks = match &event.event.data {
+                KvCacheEventData::Stored(data) => data.blocks.len(),
+                KvCacheEventData::Removed(data) => data.block_hashes.len(),
+                KvCacheEventData::Cleared => 0,
+            };
+            if batch_end > batch_start && batch_blocks.saturating_add(event_blocks) > max_blocks {
```
PeaBrane: AI-reviewed note (non-blocking): the 8,192-block cap does not also bound event count or encoded bytes. A native batch can produce 8,192 independent one-block `Stored` events; named MessagePack repeats the per-event maps, putting that ordinary non-MM shape above NATS Core’s default 1 MiB. The current wire-size test covers 64×128-block events, not this maximum-overhead shape. Please add an event-count/byte bound and a singleton-store envelope regression.

PeaBrane (approved):
✅ These changes were AI-reviewed, including coalescing/deduplication, local-indexer ordering, event-plane batching, and recovery behavior. One non-blocking payload-bound note is attached.

lib/llm/src/kv_router/publisher/sinks.rs
jthomson04: Hmm. That's an extremely adversarial case, but I guess that's something we need to account for. I'll add an event-based cap in addition to block-based, which should take care of the byte limit as well.

PR #11779, opened 2026-07-16
refactor(aic): use core for runtime, full AIC for profiler

jasonqinzhou:
## Overview

runtime/modeling consumers use the standalone AIC core, while profiler workflow code continues to use the full upper AIC package.

## Summary

- move runtime, mocker, router estimation, planner performance modeling, and benchmarks to `aiconfigurator-core`
- retain full `aiconfigurator` for profiler CLI, generator, enumeration, picking, and task orchestration
- pin source and Rust dependencies to AIC `main` revision `2829e0983b8698558aea9d3ab05693dc61761839`
- build the upper and core AIC wheels from that same checkout for Dynamo runtime/planner images, then verify their package ownership and installed layering
- adapt Dynamo's speculative decoding boundary to AIC's current `nextn_accepted` scalar while preserving Dynamo's existing per-position acceptance-rate configuration

### Dependency boundary

| Consumer | AIC dependency |
| --- | --- |
| mocker/runtime/router/planner performance modeling | `aiconfigurator-core` |
| Rust forward-pass integration | `aiconfigurator-core` crate at the matching source revision |
| profiler search, picking, task execution, and deployment generation | full `aiconfigurator` |
| planner/runtime container wheelhouse | matching upper and core wheels built from one pinned checkout |

Published `ai-dynamo` metadata remains index-resolvable: the `mocker` extra declares `aiconfigurator-core==0.10.0` rather than a direct Git dependency. Source-based container and benchmark builds use the exact pinned revision.

## Packaging behavior

- the root `ai-dynamo` wheel continues to contain `dynamo/profiler/**`
- no `ai-dynamo-profiler` artifact or release step is created
- planner and Dynamo runtime images install the paired local `aiconfigurator` and `aiconfigurator-core` wheels
- AIC's release-wheel and installed-layer verifiers run during the image build
- Dynamo wheel smoke validation rejects direct URL/Git requirements in published metadata

## Validation

- `cargo check -p dynamo-mocker --features aic-forward-pass`
- focused mocker speculative-config tests: 2 passed
- built both AIC wheels from the pinned revision and passed AIC release-wheel verification
- installed the paired wheels in a clean Python 3.12 environment and passed `verify_installed_package_layers.py --expect full`
- verified all profiler upper/core imports against those wheels
- built and inspected `ai-dynamo`: profiler files are present and metadata contains only the core dependency under the `mocker` extra
- rendered planner/runtime Dockerfile targets successfully
- all changed-file pre-commit hooks passed

The PyO3 unit-test binary cannot link against the host Xcode Python in this local environment; both affected Rust crates compile successfully, and the pure Rust mocker tests pass.

1. `pyproject.toml`, `container/deps/requirements.planner.txt`, and `benchmarks/pyproject.toml` for the Python dependency split.
2. `container/templates/wheel_builder.Dockerfile`, `planner.Dockerfile`, and `dynamo_runtime.Dockerfile` for paired AIC wheel construction and installation.
3. `lib/bindings/python/src/dynamo/_internal/aic.py` and `lib/mocker/src/common/engine_perf.rs` for the core-only modeling and speculative-decoding migration.
4. The revert commit for removal of the standalone profiler-wheel direction.

## Related Issues

## Release prerequisite

- [ ] Publish `aiconfigurator-core==0.10.0` to the package index before the Dynamo release so the published `ai-dynamo[mocker]` extra is installable without the container wheelhouse.


jasonqinzhou:
@coderabbitai review

tedzhouhk (changes requested):
Requesting changes for the remaining P1 release, packaging, and DGDR issues:

1. **[P1] Do not create a hybrid AI Configurator installation in the planner image.**

   `components/profiler/pyproject.toml:16-17` installs both `aiconfigurator==0.10.0` from PyPI and the locally built split `aiconfigurator-core==0.10.0`. Both distributions claim `aiconfigurator_core/__init__.py` and `_aiconfigurator_core.abi3.so`. Reproducing the planner dependency installation produced a mixed package: the Python package came from the pinned split-core wheel, while the native extension was overwritten by the older monolithic PyPI wheel. Basic imports and a small task smoke test still pass, so this can silently pair Python and native code from different revisions.

   Build the upper `aiconfigurator` wheel from the same pinned revision, or use a published upper wheel that depends on the external split core without owning the same files. Please also add an image-level ownership/version assertion so this cannot regress unnoticed.

2. **[P1] Keep direct VCS references out of the published `ai-dynamo` metadata.**

   `pyproject.toml:83-86` places `aiconfigurator-core @ git+https://...` in the `mocker` extra. The built `ai_dynamo` wheel preserves that URL in `Requires-Dist`; public package indexes do not accept direct-reference metadata, even when guarded by an extra. `allow-direct-references = true` only permits the local Hatch build. Keep the temporary source pin in development/container requirements, or replace it with an index-resolvable version before producing the public wheel.

3. **[P1] Make the standalone profiler artifact installable from the release index.**

   `components/profiler/pyproject.toml:17` requires `aiconfigurator-core==0.10.0`, but that distribution is not currently published. The planner build succeeds only because `container/templates/wheel_builder.Dockerfile` builds the core wheel locally. `.github/workflows/release.yml:583-607` then publishes the Dynamo/runtime/KVBM/profiler wheels but does not stage or publish that core wheel, so an index-only installation of `ai-dynamo-profiler` cannot resolve its dependencies. Publish the compatible core first, or include it in the same release transaction and add a clean index-install validation.

4. **[P1] Include the Apache-2.0 license in the new profiler wheel.**

   `components/profiler/pyproject.toml:12` declares license text but does not declare `license-files`, and the component project has no copied license file. An isolated build of `ai_dynamo_profiler-1.3.0-py3-none-any.whl` contained no `LICENSE`, `LICENCE`, or `COPYING` member, while the root wheel correctly contained `ai_dynamo-1.3.0.dist-info/licenses/LICENSE`. Include the repository license in this distribution and add a wheel-content assertion.

5. **[P1] Complete the documentation migration for the package split.**

   - `docs/dynosim/sweeps.md:108` exposes only `components/src` and `lib/bindings/python/src`; the moved examples import `dynamo.profiler`, so the documented source-checkout workflow cannot run unless `components/profiler/src` is added or the profiler project is installed editable.
   - `docs/dynosim/sweeps.md:84-102` still installs only the old upper package and points troubleshooting at `site-packages/aiconfigurator/systems/data`, rather than documenting compatible upper/core/profiler packages and the new data location.
   - `components/profiler/README.md` names the new wheel but provides no installation, editable-development, or test commands. The main source-build documentation also stops after installing the root project, although that wheel now excludes `dynamo.profiler`.
   - `fern/components/profiler/profiler_guide.md:591` changes a source link to a repository-root relative path. The docs synchronization does not copy `components/profiler` into the published docs checkout, so this link will be broken; keep an absolute GitHub source URL with the new path.

The root/profiler wheel split itself is structurally reasonable: both wheels build, the shared `dynamo` namespace works, the root wheel excludes profiler code, and the rendered planner Dockerfile contains the expected profiler build/install stages. After the issues above are fixed, the full `planner-build` and `planner-test` lanes still need to run to validate the image-based DGDR path. Nightly profiler publication is intentionally not requested by this review.

jasonqinzhou:
@tedzhouhk Thanks for the detailed review. I addressed items 1, 2, 4, and 5 in e7b03e584:

1. The planner now builds both `aiconfigurator==0.10.0` and `aiconfigurator-core==0.10.0` from the same pinned AIC revision (`989dd6e9cbedcbd26d353da984af451b6b64fd3f`). The AIC wheel ownership/version verifier runs before installation, and the installed planner environment is checked again with `verify_installed_package_layers.py --expect full`.
2. Published `ai-dynamo` metadata now uses the exact release requirement `aiconfigurator-core==0.10.0`; the SHA pin remains in the container build context. Wheel archive checks reject any `git+` or other direct-reference requirement.
4. `ai-dynamo-profiler` now packages the repository Apache license, and its build verifies the exact license content, wheel pairing, dependency pins, and disjoint profiler payload.
5. Updated the DynoSim/source-build/profiler instructions for the split AIC install and current data path, added the profiler source install/test commands, fixed the source-checkout `PYTHONPATH`, and repaired the Fern links.

Validation completed: root and profiler wheels build and pass the archive verifier; planner/runtime templates render for amd64 and arm64; `fern check`, `fern docs broken-links`, and all touched-file pre-commit hooks pass.

Item 3 (making `aiconfigurator-core==0.10.0` available from the public release index before publishing `ai-dynamo-profiler`) remains open for now. I intentionally left the release workflow unchanged.

PR #11783, opened 2026-07-16
perf(trtllm): batch worker KV event drains

jthomson04:
## Summary

- add generic Rust and Python `KvEventPublisher.publish_batch` APIs so one ordered source batch reaches the existing KV event processor as one input
- use one list-based polling callback for legacy TRT events and stats, bounded to 256 items so a continuously ready TRT-LLM iterator cannot starve publication; persistent iterators resume on the next polling turn without reordering or loss
- group direct KV events once per non-empty attention-DP rank in each bounded chunk, while stats and the optional ZMQ consolidator continue to iterate their chunks as singletons
- keep the optional ZMQ path explicitly singleton-only and remove the unused direct singleton-publication fallback
- preserve within-rank order, event-gap tracking, global-attention filtering, existing global partial-block suppression, cache salt, multimodal metadata, LoRA, created events, and unknown-rank handling
- preserve legacy failure and metric semantics: partial fetched chunks publish before fetch errors propagate through `ManagedThread`, failed chunks are not recorded as successfully handled, and consolidator wire messages remain singleton-based
- leave `dynamo.trtllm.unified_main`, vLLM, and SGLang unchanged

## Validation

- `cargo test -p dynamo-llm kv_router::publisher::tests --lib` (80 passed)
- `cargo test --manifest-path lib/bindings/python/Cargo.toml --lib`
- `lib/bindings/python/.venv/bin/python -m pytest -q lib/bindings/python/tests/test_kv_event_publisher.py` (1 passed)
- `PYTHONPATH=components/src lib/bindings/python/.venv/bin/python -m pytest -q components/src/dynamo/trtllm/tests/test_trtllm_fpm_publisher.py` (49 passed)
- `MYPYPATH=components/src:lib/bindings/python/src mypy --explicit-package-bases components/src/dynamo/trtllm/publisher.py` (no issues)
- strict `cargo clippy` for `dynamo-llm` and the Python binding crate with `-D warnings`
- `cargo fmt --all -- --check` for the root and Python binding workspaces
- `pre-commit run --files <changed files>`
- no live TRT-LLM run or heavyweight image pull; the focused legacy adapter suite uses the existing local test environment

### DIFF
--- a/components/src/dynamo/trtllm/publisher.py
+++ b/components/src/dynamo/trtllm/publisher.py
@@ -29,7 +29,7 @@ import weakref
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from queue import Queue
-from typing import Any, Awaitable, Callable, Dict, Optional, Union
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, Optional, Union, cast
 
 import msgspec
 import zmq
@@ -38,6 +38,9 @@ from prometheus_client import CollectorRegistry
 from dynamo.common.utils.prometheus import LLMBackendMetrics
 from dynamo.llm import FpmDirectPublisher, KvEventPublisher, WorkerMetricsPublisher
 from dynamo.trtllm.utils.request_utils import stored_event_cache_salt
+
+if TYPE_CHECKING:
+    from dynamo._core import KvRemovedEventInput, KvStoredEventInput
 
 logger = logging.getLogger(__name__)
 
@@ -52,6 +55,8 @@ _KV_EVENTS_TIMEOUT_SEC = 0.0
 _PUBLISH_MIN_SLEEP_SEC = 0.01
 _PUBLISH_MAX_SLEEP_SEC = 0.1
 _PUBLISH_BACKOFF_FACTOR = 2.0
+# Keep a continuously ready TRT-LLM iterator from starving its batch handler.
+_POLLING_BATCH_MAX_ITEMS = 256
 _KV_EVENTS_MIN_SLEEP_SEC = 0.005
 _KV_EVENTS_MAX_SLEEP_SEC = 0.02
 _KV_EVENTS_BACKOFF_FACTOR = 1.5
@@ -564,7 +569,7 @@ class Publisher:
     async def _polling_loop(
         self,
         fetch_fn,
-        handler_fn,
+        batch_handler_fn,
         min_sleep: float,
         max_sleep: float,
         backoff_factor: float,
@@ -572,23 +577,43 @@ class Publisher:
     ):
         sleep_s = min_sleep
         while not self._stop_event.is_set():
-            had_data = False
-            batch_size = 0
+            batch = []
+            fetch_error = None
             try:
                 async for item in fetch_fn():
-                    had_data = True
-                    batch_size += 1
-                    handler_fn(item)
+                    batch.append(item)
+                    if len(batch) >= _POLLING_BATCH_MAX_ITEMS:
+                        break
             except (asyncio.TimeoutError, TimeoutError, asyncio.QueueEmpty):
                 pass
             except Exception as e:
-                logging.warning(f"Publisher polling loop error: {e}", exc_info=True)
-                raise
+                fetch_error = e
 
-            if batch_size and batch_size_handler_fn is not None:
-                batch_size_handler_fn(batch_size)
+            if batch:
+                try:
+                    batch_handler_fn(batch)
+                except Exception as e:
+                    logger.warning("Publisher polling loop error: %s", e, exc_info=True)
+                    if fetch_error is not None:
+                        raise e from fetch_error
+                    raise
 
-            if not had_data:
+            if fetch_error is not None:
+                logger.warning(
+                    "Publisher polling loop error: %s",
+                    fetch_error,
+                    exc_info=(
+                        type(fetch_error),
+                        fetch_error,
+                        fetch_error.__traceback__,
+                    ),
+                )
+                raise fetch_error
+
+            if batch and batch_size_handler_fn is not None:
+                batch_size_handler_fn(len(batch))
+
+            if not batch:
                 await asyncio.sleep(sleep_s)
                 sleep_s = min(max_sleep, sleep_s * backoff_factor)
             else:
@@ -758,9 +783,13 @@ class Publisher:
                     # stat shape.
                     logging.warning(f"FPM publish failed: {e}")
 
+        def handle_stats(stats):
+            for stat in stats:
+                handle_stat(stat)
+
         await self._polling_loop(
             lambda: self.engine.llm.get_stats_async(timeout=_STATS_TIMEOUT_SEC),
-            handle_stat,
+            handle_stats,
             _PUBLISH_MIN_SLEEP_SEC,
             _PUBLISH_MAX_SLEEP_SEC,
             _PUBLISH_BACKOFF_FACTOR,
@@ -785,19 +814,25 @@ class Publisher:
             lambda: self.engine.llm.get_kv_cache_events_async(
                 timeout=_KV_EVENTS_TIMEOUT_SEC
             ),
-            self._handle_kv_event,
+            self._handle_kv_event_drain,
             _KV_EVENTS_MIN_SLEEP_SEC,
             _KV_EVENTS_MAX_SLEEP_SEC,
             _KV_EVENTS_BACKOFF_FACTOR,
-            self._record_kv_event_drain_batch,
+            batch_size_handler_fn=self._record_kv_event_drain_batch,
         )
         return True
+
+    def _handle_kv_event_drain(self, events):
+        if self.zmq_kv_event_publisher:
+            self._handle_zmq_kv_event_batch(events)
+        else:
+            self._handle_kv_event_batch(events)
 
     def _record_kv_event_drain_batch(self, batch_size: int) -> None:
         if self.additional_metrics is not None:
             self.additional_metrics.record_kv_event_drain_batch(batch_size)
 
-    def _handle_kv_event(self, event):
+    def _normalize_kv_event(self, event):
         event_id = event["event_id"]
         attention_dp_rank = event.get("attention_dp_rank", 0)
 
@@ -906,39 +941,16 @@ class Publisher:
                 cache_salt is not None,
                 parent_hash is not None,
             )
-            # Publish to ZMQ if consolidator is enabled, otherwise publish to NATS
-            # Note: event_id is managed internally by the publisher (monotonic counter per dp_rank)
-            if self.zmq_kv_event_publisher:
-                # Consolidator enabled: publish to ZMQ only
-                self.zmq_kv_event_publisher.publish_stored(
-                    token_ids,
-                    num_block_tokens,
-                    block_hashes,
-                    parent_hash,
-                    block_mm_infos,
-                    attention_dp_rank,
-                    lora_name,
-                    cache_salt,
-                )
-            elif self.kv_event_publishers:
-                # No consolidator: publish to NATS (router subscribes directly)
-                # Route to correct publisher based on attention_dp_rank
-                publisher = self.kv_event_publishers.get(attention_dp_rank)
-                if publisher:
-                    publisher.publish_stored(
-                        token_ids,
-                        num_block_tokens,
-                        block_hashes,
-                        parent_hash,
-                        block_mm_infos,
-                        lora_name=lora_name,
-                        cache_salt=cache_salt,
-                    )
-                else:
-                    logging.warning(
-                        f"No publisher for attention_dp_rank={attention_dp_rank}, "
-                        f"available ranks: {list(self.kv_event_publishers.keys())}"
-                    )
+            return attention_dp_rank, {
+                "type": "stored",
+                "token_ids": token_ids,
+                "num_block_tokens": num_block_tokens,
+                "block_hashes": block_hashes,
+                "parent_hash": parent_hash,
+                "block_mm_infos": block_mm_infos,
+                "lora_name": lora_name,
+                "cache_salt": cache_salt,
+            }
         elif data["type"] == "removed":
             self.processing_initial_created_events = False
             removed_block_hashes: list[int] = []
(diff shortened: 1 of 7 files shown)


jthomson04:
@coderabbitai full review

jthomson04:
@coderabbitai full review

components/src/dynamo/trtllm/llm_engine.py
```diff
@@ -550,21 +551,44 @@ def _kv_events_poll_loop(self) -> None:
...
+                normalized = self._normalize_kv_event(event)
+            except (KeyError, TypeError, ValueError) as error:
+                self._warn_malformed_kv_event(error)
+                continue
+            if normalized is not None:
+                rank, normalized_event = normalized
+                events_by_rank.setdefault(rank, []).append(normalized_event)
+
+        for rank, normalized_events in events_by_rank.items():
+            self._kv_publishers[rank].publish_batch(normalized_events)
```
jthomson04: The old poll loop wrapped every dispatch in `except Exception` (warn once, continue), so a publish failure could never kill the thread. Now only `_normalize_kv_event` is guarded — any exception from this `publish_batch` call escapes `_kv_events_poll_loop`, which runs in a bare daemon thread with no supervisor. The thread dies via the default excepthook and KV event publishing silently stops for the worker's lifetime (the router quietly degrades, with only a one-time stderr traceback).

Two concrete triggers:
- a single token id outside `u32` range makes the Rust-side all-or-nothing `depythonize` reject the entire per-rank batch with `ValueError` — so one bad event now loses the whole drain *and* kills the publisher, where it previously lost just that one event
- a channel-closed/publish error during a shutdown race

`test_dispatch_kv_events_propagates_publish_failure` pins this propagation deliberately, so flagging as a design decision to revisit rather than an oversight. Note the asymmetry with the legacy path: `publisher.py`'s task runs under a `ManagedThread` with an `error_queue`, so propagation there is supervised; this path has nothing above it.

Suggestion: catch `ValueError` (or `Exception`) around the per-rank `publish_batch` with the existing one-shot warn — or wrap the `_dispatch_kv_events(events)` call in the poll loop — and consider a per-event fallback publish when a batch is rejected so one poisoned event doesn't discard its healthy neighbors.

components/src/dynamo/trtllm/llm_engine.py
```diff
@@ -550,21 +551,44 @@ def _kv_events_poll_loop(self) -> None:
...
+                "Dropping malformed KV event; suppressing further "
+                "tracebacks (last error: %s)",
+                error,
+                exc_info=True,
+            )
 
     def _dispatch_kv_event(self, event: dict[str, Any]) -> None:
-        """Forward stored / removed events to the right publisher. Other
-        event types are dropped — the Python publisher has no path for them."""
+        """Normalize and publish one event for compatibility with direct callers."""
```
jthomson04: Only tests call `_dispatch_kv_event` now — could migrate them to `_dispatch_kv_events` and delete this, though harmless to keep.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -954,26 +968,78 @@ def _handle_kv_event(self, event):
...
-                    )
+            return attention_dp_rank, {
+                "type": "removed",
+                "block_hashes": removed_block_hashes,
+            }
         elif data["type"] == "created" and self.processing_initial_created_events:
             self.update_max_window_size(event)
+        return None
+
+    def _handle_kv_event(self, event):
```
jthomson04: Malformed-event tolerance is now mode-dependent: `_handle_kv_event_batch` drops malformed events with a one-shot warning and continues, but this consolidator path still raises on the same input and kills its polling task. That matches the old behavior, but the two modes now respond differently to identical bad input — worth unifying while you're here.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -559,25 +560,56 @@ async def _polling_loop(
...
+                fetch_error = e
+
+            if batch and batch_handler_fn is not None:
+                try:
+                    batch_handler_fn(batch)
+                except Exception as e:
+                    logger.warning("Publisher polling loop error: %s", e, exc_info=True)
+                    if fetch_error is not None:
+                        raise e from fetch_error
+                    raise
```
jthomson04, line 599: When the batch handler raises, this re-raise skips `batch_size_handler_fn` below — but a fetch failure still records the drain metric. Slightly inconsistent drain-metric semantics.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -559,25 +560,56 @@ async def _polling_loop(
...
 
             if batch_size and batch_size_handler_fn is not None:
                 batch_size_handler_fn(batch_size)
 
+            if fetch_error is not None:
+                logger.warning(
+                    "Publisher polling loop error: %s",
+                    fetch_error,
+                    exc_info=(
```
jthomson04, line 605: Nit: `exc_info=fetch_error` works — logging accepts an exception instance directly.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -954,26 +968,78 @@ def _handle_kv_event(self, event):
...
+            if normalized is not None:
+                attention_dp_rank, normalized_event = normalized
+                events_by_rank.setdefault(attention_dp_rank, []).append(
+                    normalized_event
+                )
+
+        for attention_dp_rank, normalized_events in events_by_rank.items():
+            self._publish_direct_kv_batch(attention_dp_rank, normalized_events)
+
+    def _warn_malformed_kv_event(self, error: Exception) -> None:
```
jthomson04: `_warn_malformed_kv_event` and the group-by-rank loop are duplicated nearly verbatim between here and `llm_engine.py`, each with its own one-shot flag. Fine for now given the two paths, but a shared helper would prevent drift.

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -951,6 +952,64 @@ pub(crate) struct KvEventPublisher {
     warning_count: Arc<AtomicU32>,
 }
 
+#[derive(Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+enum KvEventInput {
```
jthomson04, line 1035: Consider `#[serde(deny_unknown_fields)]`: a typoed optional key (e.g. `cache_sal`) is currently silently ignored, which slightly undermines the "fully validated before enqueue" invariant. Callers are internal, so strictness is cheap here.

jthomson04 (commented):
Reviewed the full diff. The design is sound: ordered per-rank batching with all-or-nothing validation before enqueue is the right shape, GIL discipline in the binding is correct (depythonize under the GIL, conversion/hashing/publish under `allow_threads`), within-rank ordering is preserved end to end, and test coverage is thorough. One significant robustness concern on the unified path (inline), plus smaller consistency items.

Two non-inline notes:

- **Stale dependency note:** the PR body says "stacked on #11776, retarget after merge," but the PR is already a single commit based on `main`, and `main` already has the `Vec<PlacementEvent>` channel and a block-count flush bound in the event processor. Worth updating the body and confirming whether #11776's 8,192-block wire cap is actually a prerequisite — as far as I can tell the diff applies and functions against today's `main`.
- **Missing test:** nothing asserts the *poll loop's* terminal behavior on publish failure — the propagate tests stop at `_dispatch_kv_events`. Whichever way the inline thread-death question lands, the loop-level behavior deserves an explicit test.

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -951,6 +952,64 @@ pub(crate) struct KvEventPublisher {
...
+    },
+    Removed {
+        block_hashes: Vec<i64>,
+    },
+}
+
+fn depythonize_kv_event_inputs(events: &Bound<'_, PyAny>) -> PyResult<Vec<KvEventInput>> {
+    depythonize(events)
+        .map_err(|error| PyValueError::new_err(format!("invalid KV event batch: {error}")))
+}
```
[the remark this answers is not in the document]
jthomson04: Covered through the real binding path in test_kv_event_publisher.py: actual Python list and dict objects now exercise depythonize with mixed stored and removed inputs, every optional stored field, malformed-batch rejection, and unknown-field rejection. I ran the TCP integration test locally and it passed.

components/src/dynamo/trtllm/llm_engine.py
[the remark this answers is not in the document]
jthomson04: Restored the unified polling resilience from main. Each non-empty rank batch is attempted, _dispatch_kv_events raises the first failure afterward, and _kv_events_poll_loop logs it once and continues later drains. The new loop-level regression test verifies a second drain is published after the first fails.

jthomson04: Implemented at the poll-loop boundary while still attempting all rank groups before returning the first error. The loop logs once and survives, covered by a later-drain regression test. Batch rejection remains all-or-nothing and intentionally does not retry individual events.

components/src/dynamo/trtllm/llm_engine.py
jthomson04: Removed the test-only _dispatch_kv_event compatibility helper and migrated its remaining test to _dispatch_kv_events.

components/src/dynamo/trtllm/publisher.py
jthomson04: Keeping this mode-dependent behavior intentionally to preserve existing semantics: the optional ZMQ consolidator remains singleton-based and fatal on malformed input, while the new direct batch path can drop malformed items. Unifying these would change behavior outside this batching change.

components/src/dynamo/trtllm/publisher.py
jthomson04: Moved drain-size recording before batch-handler invocation, so an already-fetched drain is measured even when publication raises. Added a regression test that verifies the metric before the legacy error propagates.

components/src/dynamo/trtllm/publisher.py
jthomson04: Keeping the explicit type, value, traceback exc_info tuple. It preserves the exact fetch traceback and makes that intent clear alongside the chained batch-handler failure path.

components/src/dynamo/trtllm/publisher.py
jthomson04: Keeping the helpers separate for now because the lifecycles deliberately differ: unified polling logs and continues, while the legacy path reports through ManagedThread and stops. A shared helper would couple those error contracts and broaden this change.

lib/bindings/python/rust/llm/kv.rs
jthomson04: Added deny_unknown_fields to the internally tagged input enum. Both the Rust unit suite and the real Python-object binding integration test now cover a misspelled optional field rejecting the complete sequence.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -954,26 +968,78 @@ def _handle_kv_event(self, event):
...
+                self._warn_malformed_kv_event(error)
+                continue
+            if normalized is not None:
+                attention_dp_rank, normalized_event = normalized
+                events_by_rank.setdefault(attention_dp_rank, []).append(
+                    normalized_event
+                )
+
+        for attention_dp_rank, normalized_events in events_by_rank.items():
+            self._publish_direct_kv_batch(attention_dp_rank, normalized_events)
```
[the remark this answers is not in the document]
jthomson04: Replaced the global legacy partial-hash set with rank-local sets. The regression test uses the same hash on two ranks and verifies only the rank that recorded it as partial suppresses its removal.

lib/bindings/python/tests/test_kv_event_publisher.py
```diff
@@ -0,0 +1,56 @@
...
+    pytest.mark.pre_merge,
+    pytest.mark.integration,
+]
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("request_plane", ["tcp"], indirect=True)
+async def test_publish_batch_rejects_complete_malformed_python_list(
+    runtime: DistributedRuntime,
+) -> None:
```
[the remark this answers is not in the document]
jthomson04: Added a 30-second pytest timeout while retaining the asyncio and indirect TCP request-plane markers. The real TCP integration test passed locally in 0.23 seconds.

jthomson04:
Addressed the non-inline review feedback in `b0cd803978`:

- unified stored-event normalization now derives multimodal block metadata and both direct adapters forward optional Eagle mode without inference
- the unified poll loop attempts every rank batch, reports the first failure, logs once, and continues later drains; a loop-level regression test covers recovery
- the Rust publisher suite now covers closed-channel batch errors preserving event identity and order
- the PR body now describes the existing vector channel and processor cap on `main`; the stale #11776 dependency and 8,192-block claim are removed

Local validation passed for 80 Rust publisher tests, 3 binding unit tests, the real TCP binding integration test, and 52 legacy TRT publisher tests, plus strict clippy, formatting, and pre-commit. Unified TRT-LLM pytest remains CI-covered because no local TRT-LLM runtime or image was available.

components/src/dynamo/trtllm/llm_engine.py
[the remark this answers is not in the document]
jthomson04: Final disposition after de-scoping: all unified TRT changes were removed, and `components/src/dynamo/trtllm/llm_engine.py` now matches current `main` exactly. Unified publication failures therefore retain the existing log-and-continue behavior; this concern no longer applies to this PR.

jthomson04: Final disposition after de-scoping: the unified direct-batch path and its tests were removed. `dynamo.trtllm.unified_main` keeps the existing singleton dispatch and polling semantics from `main`, so this failure-mode change is no longer present.

components/src/dynamo/trtllm/llm_engine.py
jthomson04: Final disposition after de-scoping: the unified engine file and adapter tests were restored exactly to `main`. This PR no longer changes or adds unified dispatch helpers.

components/src/dynamo/trtllm/publisher.py
jthomson04: Updated in the scoped rewrite: the legacy direct batch handler no longer drops malformed entries. Normalization errors propagate through the existing `ManagedThread` error path, matching the singleton/consolidator fatal behavior from `main`.

components/src/dynamo/trtllm/publisher.py
jthomson04: Updated in the scoped rewrite: drain-size recording remains after successful batch handling. A batch-handler failure or fetch failure propagates without recording the failed drain, matching `main`; already-fetched events are still delivered as one partial batch before a fetch error is re-raised.

components/src/dynamo/trtllm/publisher.py
jthomson04: Final disposition after de-scoping: the unified normalization/batching helper was removed with the entire unified diff, and the legacy malformed-warning helper was also removed. There is no longer duplicated unified/legacy helper logic in this PR.

components/src/dynamo/trtllm/publisher.py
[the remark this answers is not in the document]
jthomson04: Final no-code disposition: this PR intentionally preserves `main`'s global partial-block set. Rank-local partial suppression would change existing legacy semantics independently of source batching, so that change and its regression test were removed from this scoped PR.

jthomson04:
Final scoped rewrite is now `70e292edc8`:

- replayed as one DCO- and GPG-signed commit on current `main`
- kept the generic batch publisher API and legacy `dynamo.trtllm` adoption only
- simplified legacy polling consumers around drained lists
- made the optional ZMQ adapter explicitly singleton-only and removed the unused direct singleton fallback
- retained the post-drain metric observer because moving it into the handler would change `main` semantics for partial fetch failures
- narrowed the optional ZMQ publisher and typed the normalized batch boundary so the components mypy job accepts the new API
- left unified TRT, vLLM, SGLang, and consolidator wire batching unchanged

Validation passed: targeted mypy, 80 Rust publisher tests, 3 binding unit tests, the bounded TCP binding integration test, 53 legacy TRT publisher tests, strict clippy, formatting checks, and pre-commit.

components/src/dynamo/trtllm/publisher.py
```diff
@@ -554,31 +557,49 @@ def _init_publish_kv_cache_events_thread(self):
...
-            had_data = False
-            batch_size = 0
+            batch = []
+            fetch_error = None
             try:
                 async for item in fetch_fn():
-                    had_data = True
-                    batch_size += 1
-                    handler_fn(item)
+                    batch.append(item)
```
PeaBrane, line 584: 🤖 AI-assisted finding: this waits for `fetch_fn()`’s async iterator to finish before calling the handler. In Dynamo’s pinned TRT-LLM 1.3.0rc20, `IterationResult.__anext__` stops only after `aqueue.get(timeout=...)` times out, so a continuously non-empty stats/KV queue can keep this loop appending indefinitely and prevent publication. The prior loop handled each yielded item immediately. Please bound a drain by count/time—or process bounded chunks—and cover a producer that stays non-empty until the handler runs.

PeaBrane (approved):
🤖 AI-assisted review. Approving as requested. The batch API and per-rank flow look coherent overall; one material forward-progress concern is noted inline.

components/src/dynamo/trtllm/publisher.py
jthomson04: Fixed in dcaf5862ca. The shared legacy polling loop now invokes its batch handler after at most 256 items, then immediately resumes the same persistent TRT-LLM iterator on the next polling turn, preserving order and successful-only chunk metrics. The new test_polling_loop_bounds_and_resumes_continuously_ready_producer regression keeps producing until the handler runs and verifies ordered 256 + 1 callbacks with no loss or duplication. The focused legacy suite passes 54 tests, and targeted mypy plus pre-commit are clean.

alec-flowers:
Its around 500 lines of test. Are these tests useful and doing something? Or just because AI put them?

jthomson04:
> Its around 500 lines of test. Are these tests useful and doing something? Or just because AI put them?
These can definitely be trimmed.

tanmayv25 (approved):

Issue #10774, opened 2026-06-17
ConcurrentRadixTreeCompressed: unbounded memory growth — internal nodes never reclaimed (split-but-never-merge) under prefix churn


Shang-Pin:
## Summary

`ConcurrentRadixTreeCompressed` grows memory without bound under sustained traffic with diverse prompt prefixes. The frontend/KV-router process RSS climbs continuously (we observed a ~20 GB/h climb on top of an already-large baseline) until it OOMs. A jemalloc heap profile attributes ~94% of the growth to `ConcurrentRadixTreeCompressed::apply_event → node::Node::from_blocks_for_worker`.

Root cause is structural: the tree **splits but never merges**, cleanup reclaims only **stale leaves**, and **sticky-internal** nodes are retained forever once they've had children. So the internal-node count grows monotonically with the number of distinct prefixes ever seen, which under real prompt churn is unbounded.

## Environment

- Reproduces on current `main` (the relevant CRTC code is unchanged).
- KV-aware routing: `router_mode = kv`, `use_kv_events = true`, `router_event_threads = 4` (the default — `> 1` selects `ConcurrentRadixTreeCompressed`).
- Sustained request load with high prompt diversity (many distinct prefixes over time).

## Evidence (jemalloc heap profile, `jeprof`)

Diff between the earliest and latest live-heap dumps (≈16.9 GB of growth):

```
Total: 16932050386 B
16883718187  99.7%  __rustc::__rdl_alloc
24646154    0.1%   ...concurrent_radix_tree_compressed::node::Node::from_blocks_for_worker   [cum 15.98 GB, 94.4%]
 9963296    0.1%   ...concurrent_radix_tree_compressed::ConcurrentRadixTreeCompressed::apply_event [cum 17.03 GB, 100.6%]
 3679398    0.0%   ...concurrent_radix_tree_compressed::node::Node::split_at_locked              [cum  0.98 GB,  5.8%]
```

Absolute (latest dump, ~19.3 GB live): same two frames dominate — `apply_event` cum 99.3%, `from_blocks_for_worker` cum 93.9%.

A telling secondary signal: KV-cache **remove events outpace store events** (~354/s vs ~139/s, zero `block_not_found`) yet memory still grows — i.e. the leak is in retained **node allocations**, not an event store/remove imbalance. Removes shorten per-worker cutoffs and drop stale leaves, but the internal `Node` structures (and their `edge` vectors allocated in `from_blocks_for_worker`) are never freed.

[...]

PR #11785, opened 2026-07-16
fix(kv-router): scrub lookup entry when a remove can't resolve its node

Shang-Pin:
## Overview:

`ConcurrentRadixTreeCompressed` permanently leaks per-worker lookup entries
when a remove event cannot resolve its node. In `remove.rs::apply_removed`,
the `resolve_lookup → None` branch skips the hash **without removing the
worker's lookup entry** — unlike the two sibling miss-branches in
`apply_removed_hash`, which both scrub. Since the tracked-blocks accounting is
the sum of `WorkerLookup` sizes, every such miss is a phantom tracked block,
and since all the miss paths log at `debug`, the leak is silent in production.

The state is reachable without exotic races — it is the designed interplay of
lazy lookup repair and children-clearing:

1. A cross-thread split moves hash `H` from node `N` to a suffix child `S`.
   The owning worker's lookup entry for `H` still points at `N` (stale by
   design; lazy repair).
2. Coverage-emptying removes drain `N.full_edge_workers`, and
   `clear_children_if_unreachable` drops `N`'s children — detaching `S`.
3. The owner's remove for `H` arrives: `N` no longer contains `H`, and the
   repair scan of `N`'s (now empty) subtree fails. The remove is skipped and
   the entry leaks forever — external sequence hashes are effectively unique,
   so nothing will ever reference that key again.

## Production data (gpt-oss-120b disaggregated, `router_event_threads=4`)

Prefill traffic supplies both ingredients at rate (prefix-fanout splits ×
aggressive unique-chain eviction); decode supplies neither:

- Prefill per-worker tracked blocks climbed monotonically ~19k/h to **1.79×
  the physical KV pool** (397k tracked vs 3 workers × 73,846 blocks) within
  9 h, with no worker churn and a verified-complete remove stream from the
  backend (see NVIDIA/TensorRT-LLM#16335 for the producer-side verification).
- The single-threaded `RadixTree` on the **identical event stream** stays
  bounded at exactly the pool sum.
- With this fix deployed, the same workload oscillates at the pool bound
  (`live prefill workers × 73,846`) in both directions for 24+ h — identical
  behavior to the plain tree.

This is likely a major contributor to the tracked-state growth reported in
#10774 (alongside the backend's missing-remove-events bugs, fixed in
NVIDIA/TensorRT-LLM#16335): mock workers emit remove patterns that rarely
exercise the split + children-clear interleaving, which is consistent with the
difficulty reproducing that issue end-to-end.

## Details:

- One scrub call in the resolve-miss branch, mirroring the scrubs in
  `apply_removed_hash`'s two miss branches. The invariant: a remove event is
  the backend asserting the worker no longer caches the block, so no lookup
  entry for that (worker, hash) may outlive the event — whether or not the
  node can still be found to update coverage.
- Off the hot path: executes only when resolve already missed (entry absent —
  a no-op delete — or entry stale with a failed subtree scan). No locking,
  shape-gate, split, cutoff, or repair semantics are touched, per the module
  guardrails; no benchmark delta is expected. Happy to run the preferred
  Mooncake benchmark setup if desired.
- Deterministic single-threaded reproduction included
  (`remove_after_split_and_children_clear_scrubs_lookup`): two lookup maps
  emulate two event threads through the split → coverage-drain →
  children-clear → remove sequence; it fails with two phantom entries without
  the fix.

@PeaBrane — one related observation for a possible follow-up, flagging rather
than patching per the guardrails: `repair_lookup_for_resolved_node` iterates
all workers and its helper inserts on vacant, so a cross-worker repair can in
principle resurrect entries a worker's removes already scrubbed (when removes
were applied against a detached node copy whose live twin still advertises
coverage). We have not observed a measurable residual from this in production
after this fix, but a repair-only update-without-insert variant may be worth
considering.

`lib/kv-router/src/indexer/concurrent_radix_tree_compressed/remove.rs` — the
`None` arm in `apply_removed`'s resolve match, compared against the two miss
branches in `apply_removed_hash`; then the new test at the bottom of
`concurrent_radix_tree_compressed/tests.rs`.

## Related Issues

- Relates to #10774

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale lookup data remaining after block removal misses.
  * Ensured removed entries are fully cleared during index updates and subsequent cleanup.

* **Tests**
  * Added regression coverage for lookup cleanup after index splits, child clearing, and lazy removal handling.

### DIFF
--- a/lib/kv-router/src/indexer/concurrent_radix_tree_compressed/remove.rs
+++ b/lib/kv-router/src/indexer/concurrent_radix_tree_compressed/remove.rs
@@ -90,6 +90,15 @@ impl ConcurrentRadixTreeCompressed {
                         block_hash = ?block_hash,
                         "Block not found during remove; skipping"
                     );
+                    // The remove event says this worker evicted the block, so
+                    // any lookup entry for it must not outlive the event. A
+                    // resolve miss with a live entry happens when the hash's
+                    // node was split off and the split child was later dropped
+                    // by clear_children_if_unreachable — without this scrub the
+                    // entry (and the per-worker tracked-block count) leaks
+                    // permanently. Mirrors the scrubs in apply_removed_hash's
+                    // miss branches.
+                    Self::remove_lookup_hashes(lookup, worker, [block_hash]);
                 }
             }
         }
--- a/lib/kv-router/src/indexer/concurrent_radix_tree_compressed/tests.rs
+++ b/lib/kv-router/src/indexer/concurrent_radix_tree_compressed/tests.rs
@@ -1309,3 +1309,73 @@ mod remove_cleanup_tests {
         assert_score(&index, &[1, 2, 3, 8, 9], WorkerWithDpRank::new(0, 0), 5).await;
     }
 }
+
+/// Deterministic reproduction of per-worker lookup-entry leakage under the
+/// split + children-clear + lazy-repair interplay (ai-dynamo/dynamo#10774,
+/// production signature: prefill tracked_blocks climbing ~2x the physical
+/// pool while RSS stays modest).
+///
+/// Two lookup maps emulate two event threads (sticky worker routing means a
+/// worker's events are serialized, but its lookup is never repaired by other
+/// threads' splits except lazily on access):
+///
+/// 1. w1 stores [1,2,3,4] -> node N; w1's lookup has 4 entries -> N.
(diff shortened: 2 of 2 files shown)


jthomson04 (commented):
Changes look good. Also adding @PeaBrane for viz

PeaBrane (approved):
✅

PR #11790, opened 2026-07-16
fix: Docs/nvbug 6389095 snapshots

kmkelle-nv:
## Overview:

These changes address NVBugs 6389095, which lists dozens of failing links in the Dynamo docs. This PR fixes  broken links in past-version docs by directly editing the docs-website branch. 

## Details:

**Changes**
Updates are made to 1.2.1, 1.1.1, and 0.9.1:
- fern/pages-v1.2.1/ — 34 links
- fern/pages-v1.1.1/ — 6 links
- fern/pages-v0.9.1/ — 25 links

**Approach:**
We chose to **not fix the source docs** because rebuilding doesn't output the same content as was previously published. Rebuilding regenerates a version's snapshot from its release tag. Since the original snapshots were built from main instead of the tag, the tag omits some pages that are live today. Rebuilding would drop these pages.

- Each broken link is repointed to an immutable target: the version tag (blob/vX.Y.Z/…), the file's introducing commit (full 40-char SHA) when it postdates the tag, or a (when no file target exists) GitHub release page. 
- Obsolete docs.nvidia.com/archive/… URLs are repointed to their published archive equivalents. 
- No mutable main references are introduced.

## Related Issues

**🔗 This PR is linked to an NVB issue:**
Relates to #6389095


nealvaidya:
> Each broken link is repointed to an immutable target: the version tag (blob/vX.Y.Z/…), the file's introducing commit (full 40-char SHA) when it postdates the tag, or a (when no file target exists) GitHub release page.

Most of these make sense, but I don't feel great about the last one (linking to the GitHub release page) since its going to an essentially unrelated piece of content. It looks like that only comes up on the version 0.9.1 docs, maybe we can split the difference and rebuild that one from the tag?

nealvaidya (approved):

PR #11798, opened 2026-07-16
ci(nightly): run GitLab trigger in parallel with tests

nv-nmailhot:
The nightly `release` job (which stages wheels/NGC and triggers the GitLab release-automation pipeline) depended on the full test matrix, so the GitLab trigger waited out the ~3.5h test suite before firing.

The tests never gated release on pass/fail — the `if:` uses `!cancelled()` and only strictly requires `dynamo-pipeline` — so they were purely a timing barrier. Point `release.needs` at the build jobs (container builds + dynamo-pipeline) instead, so the GitLab trigger and the test matrix both fan out from the builds and run concurrently.

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **Chores**
  * Updated nightly release automation to run after runtime build jobs and the pipeline complete.
  * Removed dependencies on the corresponding runtime test stages before triggering releases.

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -626,15 +626,10 @@ jobs:
       - compute-release-mode
       - notify-slack-start
       - manual-release-approval
-      - vllm-test
-      - vllm-multi-gpu-test
-      - sglang-test
-      - sglang-multi-gpu-test
-      - sglang-h100-test
-      - trtllm-test
-      - trtllm-multi-gpu-test
+      - vllm-build
+      - sglang-build
+      - trtllm-build
       - dynamo-pipeline
-      - rust-tests
     # !cancelled() lets framework/rust failures fall through; dynamo-pipeline
     # is gated strictly because stage-wheels-artifactory extracts wheels from
     # its image. manual-release-approval is `skipped` on cron, `success` on dispatch.
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -630,11 +630,17 @@ jobs:
       - sglang-build
       - trtllm-build
       - dynamo-pipeline
-    # !cancelled() lets framework/rust failures fall through; dynamo-pipeline
-    # is gated strictly because stage-wheels-artifactory extracts wheels from
-    # its image. manual-release-approval is `skipped` on cron, `success` on dispatch.
+    # !cancelled() lets the job evaluate the explicit gates below rather than
+    # implicitly requiring every need to succeed (notify-slack-start is
+    # best-effort). All three runtime builds and dynamo-pipeline are gated
+    # strictly: release.yml copies their -nightly ECR images to NGC and extracts
+    # wheels from the dynamo-pipeline image, so a failed build must not publish a
+    # partial release. manual-release-approval is `skipped` on cron, `success` on dispatch.
     if: ${{ !cancelled()
         && needs.compute-release-mode.outputs.release == 'true'
+        && needs.vllm-build.result == 'success'
+        && needs.sglang-build.result == 'success'
+        && needs.trtllm-build.result == 'success'
         && needs.dynamo-pipeline.result == 'success'
         && (needs.manual-release-approval.result == 'success'
             || needs.manual-release-approval.result == 'skipped') }}


.github/workflows/nightly-ci.yml
```diff
@@ -626,15 +626,10 @@ jobs:
...
-      - vllm-test
-      - vllm-multi-gpu-test
-      - sglang-test
-      - sglang-multi-gpu-test
-      - sglang-h100-test
-      - trtllm-test
-      - trtllm-multi-gpu-test
+      - vllm-build
+      - sglang-build
+      - trtllm-build
```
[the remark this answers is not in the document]
nv-anants: @nv-nmailhot this seems relevant, can you check?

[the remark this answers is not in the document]
nv-nmailhot: added now

.github/workflows/nightly-ci.yml
```diff
@@ -626,20 +626,21 @@ jobs:
...
       - dynamo-pipeline
-      - rust-tests
-    # !cancelled() lets framework/rust failures fall through; dynamo-pipeline
-    # is gated strictly because stage-wheels-artifactory extracts wheels from
-    # its image. manual-release-approval is `skipped` on cron, `success` on dispatch.
+    # !cancelled() lets the job evaluate the explicit gates below rather than
```
dmitry-tokarev-nv, line 633: **Tests can no longer gate the nightly release — confirming that's the intended policy.**

Removing the test jobs from `release.needs` (plus the `!cancelled()` + `needs.<build>.result == 'success'` gates) fully disconnects the framework and `rust-tests` results from the release decision — GitHub can't gate on a job that isn't a `need`. The description is right that they never *strictly* gated (only `dynamo-pipeline` did), but this makes it structural.

That contradicts the intent documented in `compute-release-mode` at `nightly-ci.yml:131`:

```yaml
schedule)
  # cron must keep tests on so a failing nightly blocks the release.
```

`run_tests=true` is still forced on cron *"so a failing nightly blocks the release,"* yet nothing consumes that result for gating now. Two ways to reconcile:

- **If publish-regardless is intended** (looks like it): update that comment so it stops asserting the opposite — e.g.
  ```yaml
  schedule)
    # cron runs the full matrix for signal/coverage; tests run concurrently with
    # the release and do NOT gate it (see the release job).
  ```
- **If a red nightly was meant to block release**, this PR is the wrong direction — it should *add* `needs.<test>.result == 'success'` gates rather than drop the test needs.

Stakes: on nightly, `release.yml` promotes the mutable floating tags `:nightly` / `:nightly-cuda13` (`release.yml:331-337`). Pre-PR that happened after the ~3.5h matrix; now it fires ~3.5h earlier while tests are still running, so a nightly whose tests later fail is promoted to the floating tag with no intervening gate. Same end-state as before (red images published regardless), but the ordering that let someone catch it is gone — worth an explicit sign-off from whoever owns line 131.

---

_Adjacent nit:_ the comment just above at `nightly-ci.yml:618-621` ("release.yml requires post-merge CI to have already pushed images to ECR for that SHA") doesn't hold on the nightly path — `release.yml` skips all post-merge-sourced images for nightly (`release.yml:341-398`) and uses only this run's `-nightly` outputs. The new comment right below already states the accurate behavior, so this one could be trimmed.

dmitry-tokarev-nv (commented):
One finding the existing threads don't cover: a test-gating policy/intent question (see inline comment). Not re-raising the runtime-build success gate — that's the CodeRabbit thread already handled by commit 6bb6feb.

.github/workflows/nightly-ci.yml
```diff
@@ -626,20 +626,21 @@ jobs:
...
       - notify-slack-start
       - manual-release-approval
-      - vllm-test
-      - vllm-multi-gpu-test
-      - sglang-test
-      - sglang-multi-gpu-test
-      - sglang-h100-test
-      - trtllm-test
-      - trtllm-multi-gpu-test
+      - vllm-build
```
pvijayakrish, line 629: Skipping all tests in nightly CI is not the long term behavior we want to have. Approving for now.

pvijayakrish (approved):

nv-anants (approved):

.github/workflows/nightly-ci.yml
nv-nmailhot: yes the goal is to bypass the tests as a gate

.github/workflows/nightly-ci.yml
nv-nmailhot: ack, this would be long term goal yes

Issue #11801, opened 2026-07-16
ai-dynamo frontend TCP accept-loop exhausts file descriptors and enters a runaway, unbounded retry loop under high concurrency, instead of backing off


mikekg:
# [Bug]: ai-dynamo frontend TCP accept-loop exhausts file descriptors and enters a runaway, unbounded retry loop under high concurrency, instead of backing off

## Affected component

**Not vLLM itself.** The crash is in NVIDIA's `ai-dynamo` project's Rust runtime crate —
specifically `dynamo_runtime::pipeline::network::tcp::server` — which implements the HTTP/TCP
frontend that sits in front of a vLLM (or sglang/trtllm) backend engine. vLLM's own
Python OpenAI-compatible server and request-handling code are never reached; the failure occurs
purely in the frontend's connection-accept loop, before any request reaches the model.

- **Repo**: `ai-dynamo/dynamo`
- **vLLM version**: pip-reported `0.1.dev17492+g454b47db8` (dev build inside the
  `vllm/vllm-openai:minimax-m3` container image; base vLLM release intended to track current
  `v0.25.1`)
- **ai-dynamo wheel version**: `1.2.0.dev20260426`
- **GPU / OS**: NVIDIA H100 80GB HBM3, Ubuntu-based container, Slurm-on-Kubernetes cluster
- **Model**: a 120B-parameter MoE model (NVFP4-quantized, hybrid Mamba2+attention+MoE
  architecture) served with `tensor-parallel-size=2`, `max-num-seqs=1024`, `kv-cache-dtype=fp8`
  — but model choice is very likely irrelevant; this is a frontend connection-handling bug that
  should reproduce with any model/backend behind the same `ai-dynamo` frontend, given enough
  concurrent connections.

## Summary

Under sustained high-concurrency load (concurrency=1024, non-keep-alive/per-request HTTP
connections), the dynamo frontend's TCP listener hits the process's open-file-descriptor limit
(`ulimit -n`) and its `accept()` retry loop enters an **unbounded, zero-backoff spin**: each
failed `accept()` call is retried again essentially immediately (observed retry interval:
**~8-9 microseconds** between consecutive log lines), rather than backing off, applying
backpressure, or gracefully degrading. This has two compounding effects:

1. **Log volume explosion**: because every failed accept is logged, the frontend's stdout log
   grew to **122 MB within ~78 seconds** of load starting, and to **15.8 GB / 86.9 GB** (two
[...]

PR #11802, opened 2026-07-16
fix(frontend): raise soft RLIMIT_NOFILE at startup to avoid accept() EMFILE spiral

mikekg:
Fixes the accept() failure spiral described in ai-dynamo/dynamo#11801.

The frontend's TCP accept loop (`lib/runtime/src/pipeline/network/tcp/server.rs`) retries `accept()` with no backoff on failure. Under high concurrency with a non-keep-alive HTTP client, established connections can reach roughly 2x the nominal benchmark concurrency, and the default soft `RLIMIT_NOFILE` (1024 on most distros) gets exhausted, causing `accept()` to fail with EMFILE in a tight zero-backoff loop that also floods stdout (observed 15.8GB/86.9GB logs within minutes).

This raises the process's soft `RLIMIT_NOFILE` at startup (`components/src/dynamo/frontend/main.py`), bounded by whatever hard limit the environment provides. No-op if the environment already has enough headroom. No changes to the accept-loop logic itself.

See ai-dynamo/dynamo#11801 for full repro details, environment, and logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved frontend startup reliability by automatically increasing the process file-descriptor limit when needed.
  * Respects the system’s maximum allowed limit and records the adjustment.

### DIFF
--- a/components/src/dynamo/frontend/main.py
+++ b/components/src/dynamo/frontend/main.py
@@ -20,6 +20,7 @@ import argparse
 import asyncio
 import logging
 import os
+import resource
 import signal
 import sys
 from argparse import Namespace
@@ -61,8 +62,6 @@ FRONTEND_FD_LIMIT_TARGET = 8192
 def _raise_fd_limit(target: int = FRONTEND_FD_LIMIT_TARGET) -> None:
     """Raise the process's soft RLIMIT_NOFILE toward `target`, bounded by the
     hard limit. No-op if the soft limit is already >= target."""
-    import resource
-
     soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
     new_soft = (
         target if hard == resource.RLIM_INFINITY else min(target, hard)
--- a/components/src/dynamo/frontend/main.py
+++ b/components/src/dynamo/frontend/main.py
@@ -51,6 +51,26 @@ logger = logging.getLogger(__name__)
 
 MIN_INITIAL_WORKERS_ENV = "DYN_ROUTER_MIN_INITIAL_WORKERS"
 
+# Default soft limit on most Linux distros is 1024, which can be exhausted by the
+# frontend's TCP listener under high concurrency (see the accept() loop in
+# lib/runtime/src/pipeline/network/tcp/server.rs), causing accept() to fail with
+# EMFILE. Raise it at startup, bounded by whatever hard limit the environment allows.
+FRONTEND_FD_LIMIT_TARGET = 8192
+
+
+def _raise_fd_limit(target: int = FRONTEND_FD_LIMIT_TARGET) -> None:
+    """Raise the process's soft RLIMIT_NOFILE toward `target`, bounded by the
+    hard limit. No-op if the soft limit is already >= target."""
+    import resource
+
+    soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    new_soft = min(target, hard)
+    if new_soft > soft:
+        resource.setrlimit(resource.RLIMIT_NOFILE, (new_soft, hard))
+        logger.info(
+            f"Raised RLIMIT_NOFILE soft limit {soft} -> {new_soft} (hard={hard})"
+        )
+
 
 def setup_engine_factory(
     config: FrontendConfig,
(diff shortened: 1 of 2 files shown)


components/src/dynamo/frontend/main.py
```diff
@@ -20,6 +20,7 @@
 import asyncio
 import logging
 import os
+import resource
```
RitwijParmar: Since this import is at module scope, `main.py` now fails to import on Windows because `resource` is Unix-only. Dynamo has Windows CI paths, so could we keep the import lazy and make `_raise_fd_limit` a no-op when `resource` is unavailable? A small platform test would make this startup hardening safe across the supported runners.

RitwijParmar (commented):
The EMFILE failure mode is real and the startup limit is a reasonable first layer. One portability issue needs attention before merge: `resource` is Unix-only, so importing it at module scope can make the frontend fail during import on Windows. I left an inline comment with the suggested lazy import and fallback. A small platform test would keep this safe across the supported runners.

components/src/dynamo/frontend/main.py
mikekg: OK, the style guide had escaalted global import :-)

We could put the import back into the _raise_fd_limit and `except:` return without raising the limit when we can't import resource. Wdyt?

biswapanda: I think this works fine. A good outcome is we always log the file descriptor limits after the fd limit set up.

components/src/dynamo/frontend/main.py
```diff
@@ -51,6 +52,26 @@
...
+def _raise_fd_limit(target: int = FRONTEND_FD_LIMIT_TARGET) -> None:
+    """Raise the process's soft RLIMIT_NOFILE toward `target`, bounded by the
+    hard limit. No-op if the soft limit is already >= target."""
+    soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    new_soft = (
+        target if hard == resource.RLIM_INFINITY else min(target, hard)
+    )
+    if new_soft > soft:
+        resource.setrlimit(resource.RLIMIT_NOFILE, (new_soft, hard))
+        logger.info(
```
biswapanda: I'm thinking that it might be useful to always info log the soft/hard limits? wdyt?

components/src/dynamo/frontend/main.py
```diff
@@ -51,6 +52,26 @@
 
 MIN_INITIAL_WORKERS_ENV = "DYN_ROUTER_MIN_INITIAL_WORKERS"
 
+# Default soft limit on most Linux distros is 1024, which can be exhausted by the
+# frontend's TCP listener under high concurrency (see the accept() loop in
+# lib/runtime/src/pipeline/network/tcp/server.rs), causing accept() to fail with
+# EMFILE. Raise it at startup, bounded by whatever hard limit the environment allows.
+FRONTEND_FD_LIMIT_TARGET = 8192
```
biswapanda, line 61: can we get this from an env variable so users can tweak it during deployment?

mikekg: Done.

rmccorm4:
/ok to test [f672fd8](https://github.com/ai-dynamo/dynamo/commit/f672fd8de623d12f4b77c55eeaf052934acae059)

components/src/dynamo/frontend/main.py
```diff
@@ -51,6 +51,46 @@
...
+        raw = os.getenv(FRONTEND_FD_LIMIT_ENV, str(FRONTEND_FD_LIMIT_TARGET))
+        try:
+            target = int(raw)
+        except ValueError:
+            target = 0
+    if target <= 0:
+        return
+    try:
+        import resource
+
```
biswapanda, line 80: should we short-circuit when soft limit is already set to RLIM_INFINITY (-1) this will prevent accidentally reducing unlimited soft limit (RLIM_INFINITY = -1) to default target (of 8192) via `min(target, hard)` path. 

the case is: 

soft = -1   # unlimited
hard = -1   # unlimited
target = 8192 # from default env var  FRONTEND_FD_LIMIT_TARGET = 8192

=> new_soft = 8192 which triggers `if new_soft (8192) > soft (-1)` path

biswapanda (approved):

components/src/dynamo/frontend/main.py
biswapanda: @mikekg ptal

Issue #11768, opened 2026-07-16
feat(request-trace): native S3 sink for request-trace records


nnshah1:
### Summary

Add a native **S3 `RequestTraceSink`** to the request-trace framework (consolidated in #11180) so deployments can write request/response payload and trace records straight to an S3 bucket **without running an OpenTelemetry collector**. A file/OTel/NATS sink already exists; S3 is a commonly-requested destination that currently requires a collector hop.

### Motivation

Request tracing captures input/output payloads and `request_end` records (hash ids, agentic context) for **perf tuning (ISL/OSL, prefix-cache behavior), offline replay, and payload capture/retention**. Today the only object-store path is `OTLP → collector → exporter`, which forces operators to run and maintain a collector purely to land records in a bucket. A **native S3 sink** removes that dependency: records are batched, compressed, and written directly to S3 from the frontend using the standard AWS credential chain. This is a destination gap, not a schema change — the same `dynamo.request.trace.v1` records, a new sink.

### Scope — 3 PRs
1. **Functional S3 `RequestTraceSink`** — writes existing `dynamo.request.trace.v1` records to a bucket, no schema change. Smallest reviewable unit; unblocks capture.
2. **Format/layout** — object keying/partitioning (`model/day/hour`), batching/rollup, compression (`.jsonl.gz`).
3. **Deploy docs (NOT operator)** — document how to provide S3 credentials to the frontend pod (exact mechanism TBD). No operator/CRD code.

### Implementation guidance (runtime)
- **Implement the trait** `RequestTraceSink` in `lib/llm/src/request_trace/sink.rs` — mirror `OtelRequestTraceSink` (`request_trace/otel_sink.rs`, ~453 lines): async `emit`, off-hot-path via the existing broadcast-channel worker.
- **Register the kind:** add `S3` to `RequestTraceSinkKind` (`request_trace/config.rs:30`), its `as_str` arm (`:40`), the parse arm (`:313`), and the build dispatch in `sink.rs` (~`:192`).
- **Config via env** (parallel to the existing `environment_names.rs::llm::request_trace` set): opt in via `DYN_REQUEST_TRACE_SINKS=...,s3`; add `DYN_REQUEST_TRACE_S3_{BUCKET,REGION,PREFIX,ROLL_BYTES,FLUSH_INTERVAL_MS}` (reuse `_FILE_ROLL_*` batching semantics).
[...]

PR #10903, opened 2026-06-23
feat(audit): S3-backed audit sink for request capture and replay

YiqiuLiu:
## Summary

Add an `s3` audit sink peer to the existing `stderr` / `nats` / `jsonl` / `jsonl_gz` sinks. Selected at runtime via `DYN_AUDIT_SINKS=s3`. Records are uploaded as rotated, gzip-compressed NDJSON segments to S3 (or any S3-compatible endpoint), partitioned by deployment and date.

Also includes:
- A `SegmentSink` trait extraction from `JsonlGzipWriter` so the rotation engine can drive both disk and cloud destinations
- A global head-based `DYN_AUDIT_SAMPLE_RATE` gate with `force_logging` bypass
- Auto-detection of `deployment` label from operator-injected `DYN_PARENT_DGD_K8S_NAME`
- Prometheus metrics for audit upload health
- Operator CRD addition (`spec.audit.aws_s3.irsaRoleArn`) for cloud-identity SA management

## Motivation

Three offline consumer patterns the existing sinks don't serve:

1. **Integration test fixtures** — capture production traffic slices, filter by model/token-count/date, replay as regression fixtures.
2. **Continuous tuning corpus** — feed real traffic shapes into autotune sweeps rather than synthetic prompts that miss prefix-cache patterns and tool-call diversity.
3. **Compliance/audit retention** — long-term queryable storage for regulatory review.

All three want records in cheap durable storage queryable by Athena/Glue/Spark/jq. S3 is the canonical answer.

### Rust (`lib/llm/`)

| File | Change |
|------|--------|
| `src/telemetry/jsonl_gz.rs` | Extract `SegmentSink` trait + `FileSegmentSink`. Add `roll_interval` for time-based rotation. Existing callers unchanged. |
| `src/telemetry/s3_segment_sink.rs` | **New.** `S3SegmentSink` impl using `aws-sdk-s3`. Buffers segments in memory, uploads on `close_segment`. Key format: `{prefix}/[{deployment}/]YYYY/MM/DD/HH/{pod}-{uuid8}-{seq}.jsonl.gz` |
| `src/audit/sink.rs` | `S3AuditSink` trait impl + `"s3"` match arm in `parse_sinks_from_env`. Bus-lag metrics wiring. |
| `src/audit/handle.rs` | `AuditRecord` gains `deployment: Option<String>` + `emitted_at_unix_ms: i64`. `create_handle` adds xxh3-based sampling gate. |
[...]

PR #11806, opened 2026-07-16
feat(request-trace): native S3 sink for request-trace records

YiqiuLiu:
## Summary

Adds a native S3 `RequestTraceSink` that writes existing `dynamo.request.trace.v1` records directly to a bucket -- parallel to the OTLP sink, no schema change, no partitioning. Smallest reviewable unit for the S3 destination work in ai-dynamo/dynamo#11768.

* New `RequestTraceSinkKind::S3` variant with `S3RequestTraceSink` (`lib/llm/src/request_trace/s3_sink.rs`), gated behind the new `request-trace-s3` cargo feature so the AWS SDK is opt-in.
* Records are batched in-process as gzipped JSONL and each finished batch is uploaded via a single `PutObject`. Object keys use a simple time-based layout for this PR (`{prefix}/{yyyy}/{mm}/{dd}/{host}-{HHMMSS}-{seq}.jsonl.gz`).
* Credentials come from the AWS SDK default provider chain -- env vars, IMDS, IRSA, Pod Identity, shared profiles. How the frontend pod is credentialed is a deployment concern and is not part of this PR.
* New env vars: `DYN_REQUEST_TRACE_S3_{BUCKET,REGION,PREFIX}`; opt in via `DYN_REQUEST_TRACE_SINKS=...,s3`.

## Scope split (per ai-dynamo/dynamo#11768)

This is PR 1 of 3:

1. **This PR** -- functional S3 sink, same schema, no layout changes.
2. Format/layout -- Hive-style `model=/date=/hour=` keys, tunable roll thresholds, Prometheus metrics.
3. Deploy docs -- how to provide AWS credentials to the frontend pod (IRSA / Pod Identity / static). No operator or CRD code, per ai-dynamo/dynamo#11768 explicit non-goals.

Supersedes [#10903](<https://github.com/ai-dynamo/dynamo/issues/10903>), which was pinned to the [pre-ai-dynamo/dynamo#11180](<https://github.com/pre-ai-dynamo/dynamo/issues/11180>) `audit` module.

## Test plan

- [X] `cargo fmt --check -p dynamo-llm`
- [X] `cargo-deny -L error --all-features check licenses bans` (bans ok, licenses ok)
- [X] `cargo metadata --locked --format-version 1`
- [X] `cargo doc --no-deps` -- 0 warnings from new files
- [X] `cargo test --locked --doc`
- [X] `cargo clippy --no-deps --all-targets -- -D warnings` (default features)
- [X] `cargo clippy --no-deps --all-targets --features request-trace-s3 -- -D warnings`
- [X] `cargo machete` -- no unused deps
- [X] `cargo test -p dynamo-llm --lib --features request-trace-s3 request_trace` -- 66/66 pass, including 4 new `s3_sink` tests
- [ ] MinIO integration test -- deferred; will add as a follow-up commit if reviewers want it in this PR

## Non-goals (deferred to PRs 2/3 per ai-dynamo/dynamo#11768)

* Rich object-key partitioning (model/day/hour).
* Sampling.
* New record schema fields.
* Any operator/CRD/IAM changes.

---

Open in Devin Review

## Summary by CodeRabbit

* **New Features**
  * Added optional S3-backed request tracing.
  * Request traces are uploaded as compressed, batched JSONL files with organized time-based paths.
  * Added configuration for S3 bucket, region, and object prefix.
  * Added safeguards for buffering, shutdown, and upload failures.
* **Documentation**
  * Documented the new S3 request-tracing configuration settings.


lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,406 @@
...
+            .finish()
+            .context("finalizing gzip batch for s3 upload")?;
+        self.uncompressed_bytes = 0;
+        self.lines = 0;
+        self.encoder = Some(GzEncoder::new(
+            Vec::with_capacity(DEFAULT_BUFFER_INITIAL_BYTES),
+            Compression::default(),
+        ));
+        Ok(bytes)
+    }
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. `GzipBatch` was replaced by `JsonlBatch`, which buffers raw JSONL and does gzip finalization inside `tokio::task::spawn_blocking` (same pattern as `telemetry::jsonl_gz`). The prior state-inconsistency window is gone: the batch resets its counters before the blocking task runs, and the encoder is constructed fresh inside that task per flush, so a `finish()` error can no longer leave the batch in a half-torn-down state.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    fn push(&mut self, record: &RequestTraceRecord) -> Result<()> {
+        let mut line = serde_json::to_vec(record).context("serializing request trace record")?;
+        line.push(b'\n');
+        let encoder = self
+            .encoder
+            .as_mut()
+            .expect("encoder present between rolls");
+        encoder
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. gzip encoding and finalization now run inside `tokio::task::spawn_blocking` in `JsonlBatch::take_finished`, matching the `telemetry::jsonl_gz` pattern. The async worker only serializes records to raw JSONL bytes (`Vec::extend_from_slice`) and hands them off for compression, so the compression CPU work no longer runs on the tokio async worker thread.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,406 @@
...
+            .unwrap_or_default();
+        let (yyyy, mm, dd, hh, mi, ss) = utc_date_parts(secs);
+        let mut key = String::new();
+        let prefix = self.options.prefix.trim_matches('/');
+        if !prefix.is_empty() {
+            key.push_str(prefix);
+            key.push('/');
+        }
+        key.push_str(&format!(
+            "{yyyy:04}/{mm:02}/{dd:02}/{host}-{hh:02}{mi:02}{ss:02}-{seq:06}.jsonl.gz",
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. Added a per-process `run_id` (v4 UUID, generated once at sink construction in `S3RequestTraceSink::from_policy`) and mixed it into every object key. New key format: `{prefix}/{yyyy}/{mm}/{dd}/{host}-{HHMMSS}-{run_id}-{seq:06}.jsonl.gz`. Pod restart within the same second or two frontends sharing a hostname now produce disjoint key namespaces. Updated the object-key tests to assert the new format.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,425 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! S3 destination for request trace records.
+//!
+//! Records are batched in-process as gzipped JSONL, and each finished batch is
+//! uploaded as one object via `PutObject`. Object keys use a simple time-based
+//! layout for PR 1 (`{prefix}/{yyyy}/{mm}/{dd}/{host}-{seq}.jsonl.gz`); richer
+//! partitioning ships in a follow-up.
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. Updated the module-level doc to match the actual format: `{prefix}/{yyyy}/{mm}/{dd}/{host}-{HHMMSS}-{run_id}-{seq}.jsonl.gz`.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,425 @@
...
+                            target: "dynamo_llm::request_trace",
+                            %error,
+                            "request trace s3: serialize failed during shutdown"
+                        );
+                    }
+                }
+                if !batch.is_empty() {
+                    upload_ready_batch(&uploader, &mut batch, &mut seq).await;
+                }
+                return;
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. Shutdown branch now calls `rx.close()` before draining, then reads to completion with `while let Some(record) = rx.recv().await`. This closes the send half so an in-flight `emit()` cannot enqueue after we start draining, and `recv().await` yields every already-queued record before returning `None`.

lib/runtime/src/config/environment_names.rs
```diff
@@ -545,6 +545,18 @@ pub mod llm {
...
+        /// `DYN_REQUEST_TRACE_SINKS` includes `s3`.
+        pub const DYN_REQUEST_TRACE_S3_BUCKET: &str = "DYN_REQUEST_TRACE_S3_BUCKET";
+
+        /// AWS region for the S3 request-trace sink. When unset the AWS SDK
+        /// default region resolution is used (env, profile, IMDS).
+        pub const DYN_REQUEST_TRACE_S3_REGION: &str = "DYN_REQUEST_TRACE_S3_REGION";
+
+        /// Optional object key prefix for the S3 request-trace sink. When unset
+        /// records land at the bucket root.
+        pub const DYN_REQUEST_TRACE_S3_PREFIX: &str = "DYN_REQUEST_TRACE_S3_PREFIX";
```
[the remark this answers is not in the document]
YiqiuLiu: Fixed. `DYN_REQUEST_TRACE_SINKS` docstring now lists `s3` and notes the `request-trace-s3` cargo feature and the `DYN_REQUEST_TRACE_S3_{BUCKET,REGION,PREFIX}` companions.

lib/llm/Cargo.toml
```diff
@@ -32,6 +32,8 @@ bench = ["dynamo-kv-router/bench"]
 kv-router-stress = ["dep:clap", "dep:indicatif", "bench"]
 mm-routing = ["dep:llm-multimodal", "dep:llm-tokenizer"]
 request-trace-bench = []
+# S3 destination for the request-trace sink. Pulls in aws-sdk-s3 + aws-config.
+request-trace-s3 = ["dep:aws-sdk-s3", "dep:aws-config"]
```
tmonty12, line 39: **[P1] Include the S3 sink in shipped Python wheels.** `request-trace-s3` is opt-in on `dynamo-llm`, but `dynamo-py3` neither forwards nor enables it, and the production `maturin` feature list omits it. Official wheels therefore compile the feature-disabled branch and cannot use `DYN_REQUEST_TRACE_SINKS=s3`. Please add a `dynamo-py3` forwarding feature and enable it in the production wheel build, or explicitly ship a separate S3-enabled artifact.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,430 @@
...
+    flush_tick.tick().await;
+
+    loop {
+        tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => {
+                // Close the receiver first so an in-flight `emit()` cannot
+                // land a record after we start draining. Then `recv()` yields
+                // every already-enqueued record and returns `None` once empty.
+                rx.close();
```
tmonty12, line 225: **[P1] Shutdown bypasses the 64 MiB roll limit.** Normal processing uploads after each push that crosses the threshold, but shutdown drains every queued record into one `JsonlBatch` and uploads only after the channel is empty. With up to 2,048 large records queued, the final batch can grow far beyond 64 MiB, causing excessive memory use or a very large single PUT and risking loss of the entire final batch. During shutdown draining, check `uncompressed_bytes()` after every successful push and upload when the roll threshold is reached, then upload the final remainder.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,430 @@
...
+use tokio_util::sync::CancellationToken;
+use uuid::Uuid;
+
+use dynamo_runtime::config::environment_names::llm::request_trace as env_request_trace;
+
+use super::RequestTraceRecord;
+use super::config::RequestTracePolicy;
+use super::sink::RequestTraceSink;
+
+const CHANNEL_CAPACITY: usize = 2048;
```
tmonty12, line 47: **[P2] Make the S3 roll threshold and flush interval configurable.** Fleet size, trace volume, object-count constraints, and delivery-latency requirements vary substantially, while these values are currently hard-coded. Please expose `DYN_REQUEST_TRACE_S3_ROLL_UNCOMPRESSED_BYTES` and `DYN_REQUEST_TRACE_S3_FLUSH_INTERVAL_MS`, parse positive values into `RequestTracePolicy`, and retain defaults when unset.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,430 @@
...
+const DEFAULT_ROLL_UNCOMPRESSED_BYTES: u64 = 64 * 1024 * 1024;
+const DEFAULT_FLUSH_INTERVAL: Duration = Duration::from_secs(60);
```
tmonty12: **[P2] Reduce the default S3 flush interval to 10 seconds.** The 60-second default leaves low-volume traces unavailable and only resident in process memory for up to a minute, increasing the crash and shutdown loss window. Use 10 seconds as the default; deployments prioritizing fewer objects can raise it through the new configuration knob.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,430 @@
...
+                "{} must be set when {} includes s3",
+                env_request_trace::DYN_REQUEST_TRACE_S3_BUCKET,
+                env_request_trace::DYN_REQUEST_TRACE_SINKS,
+            )
+        })?;
+        let prefix = policy.s3_prefix.clone().unwrap_or_default();
+        let host = hostname_or_fallback();
+        let run_id = Uuid::new_v4().simple().to_string();
+
+        let mut loader = aws_config::defaults(BehaviorVersion::latest());
```
tmonty12: **[P2] Bound S3 upload duration and define terminal failure behavior.** The AWS Rust SDK has no default operation or attempt deadline; stalled-stream protection does not bound a slow-progressing upload or delayed response. Configure approximately a 30-second attempt timeout and 90-second overall operation timeout. Standard SDK retries provide three total attempts for retryable timeouts, throttling, and transient failures, but after exhaustion this implementation discards the completed batch. Please retain or requeue the same object key and bytes after terminal failure, or explicitly meter and document the loss policy.

lib/runtime/src/config/environment_names.rs
```diff
@@ -545,6 +550,18 @@ pub mod llm {
         /// are recorded unredacted; avoid credential-bearing headers.
         pub const DYN_REQUEST_TRACE_HTTP_HEADER_CAPTURE_LIST: &str =
             "DYN_REQUEST_TRACE_HTTP_HEADER_CAPTURE_LIST";
+
+        /// S3 bucket for the S3 request-trace sink. Required when
```
tmonty12, line 558: **[P2] Document the S3 request-trace sink.** Please update `docs/observability/request-tracing.md` with an S3 configuration example and add the new S3 environment variables to the existing configuration table, matching the level of detail used for the other sinks.

lib/llm/Cargo.toml
YiqiuLiu: Fixed. Added a `request-trace-s3` feature to `dynamo-py3` (`lib/bindings/python/Cargo.toml`) that forwards to `dynamo-llm/request-trace-s3`, and enabled it in the production wheel build in `container/templates/wheel_builder.Dockerfile` (both the `ENABLE_MEDIA_FFMPEG=true` and `false` branches). Shipped wheels will now compile the S3 sink and accept `DYN_REQUEST_TRACE_SINKS=s3`.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Fixed. The shutdown-drain loop now enforces the roll threshold after every successful push, uploading a batch whenever `uncompressed_bytes >= roll_uncompressed_bytes` before continuing to drain. The trailing `is_empty()` check still uploads whatever remainder is left. A full channel of large records now produces a sequence of roll-sized PUTs during shutdown instead of one oversized final PUT.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Fixed. Added `DYN_REQUEST_TRACE_S3_ROLL_UNCOMPRESSED_BYTES` and `DYN_REQUEST_TRACE_S3_FLUSH_INTERVAL_MS` (see `environment_names.rs`). Both are parsed into `RequestTracePolicy` in `config.rs` with positive-value filtering, plumbed into the `run_worker` signature, and default when unset. Defaults: 64 MiB roll, 10 s flush.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Fixed. Default flush interval is now 10 s (`DEFAULT_S3_FLUSH_INTERVAL_MS = 10_000` in `config.rs`), and remains overridable via `DYN_REQUEST_TRACE_S3_FLUSH_INTERVAL_MS` for deployments that want fewer objects.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Partially addressed in this PR. The SDK client is now configured with `TimeoutConfig` — 30 s attempt timeout and 90 s operation timeout (`S3_ATTEMPT_TIMEOUT` / `S3_OPERATION_TIMEOUT` in `s3_sink.rs`). After SDK retries plus timeout expire, the batch is discarded and logged as `put_object failed after SDK retries; batch discarded` with `key`, `batch_bytes`, and the error — making the loss policy explicit and metered per-event. A persistent retry queue (disk spool for terminal failures) is deliberately deferred to the follow-up S3 layout PR to keep this PR focused on the functional sink per #11768.

lib/runtime/src/config/environment_names.rs
YiqiuLiu: Fixed. `docs/observability/request-tracing.md` now: (a) adds `s3` to the `DYN_REQUEST_TRACE_SINKS` values with a note about the `request-trace-s3` cargo feature, (b) adds table rows for `DYN_REQUEST_TRACE_S3_{BUCKET,REGION,PREFIX,ROLL_UNCOMPRESSED_BYTES,FLUSH_INTERVAL_MS}` with defaults/values/descriptions matching the neighbouring sinks, and (c) adds a prose subsection describing the sink pattern, object-key format, credential model, terminal-failure behavior, and a copy-paste config example.

docs/observability/request-tracing.md
```diff
@@ -40,7 +40,7 @@ export DYN_REQUEST_TRACE_FILE_PATH=/mnt/captures/run-42/request-trace
 | --- | --- | --- | --- |
 | `DYN_REQUEST_TRACE` | unset | Truthy value | Master switch. When enabled and `DYN_REQUEST_TRACE_RECORDS` is unset, emits `request_end,tool`. |
 | `DYN_REQUEST_TRACE_RECORDS` | `request_end,tool` when `DYN_REQUEST_TRACE=1`; unset otherwise | `request_end`, `request_payload`, `tool` | Comma-separated record types to emit. Setting this variable enables only the listed records. |
-| `DYN_REQUEST_TRACE_SINKS` | `file` | `file`, `stderr`, `nats`, `otel` | Comma-separated record sinks. |
+| `DYN_REQUEST_TRACE_SINKS` | `file` | `file`, `stderr`, `nats`, `otel`, `s3` | Comma-separated record sinks. The `s3` sink is compiled in only when `dynamo-llm` was built with the `request-trace-s3` cargo feature (enabled on shipped Python wheels). |
```
tmonty12: nit: can we remove the sentence "The `s3` sink is compiled in ..."

tmonty12 (approved):

docs/observability/request-tracing.md
YiqiuLiu: Removed. The cargo-feature note is dropped from the `DYN_REQUEST_TRACE_SINKS` table row; the compile-time detail still lives in the prose section further down for anyone doing a custom build.

container/templates/wheel_builder.Dockerfile
```diff
@@ -521,9 +521,9 @@ RUN --mount=type=secret,id=aws-web-identity-token,target=/run/secrets/aws-token
     uv build --wheel --out-dir /opt/dynamo/dist && \
     cd /opt/dynamo/lib/bindings/python && \
     if [ "$ENABLE_MEDIA_FFMPEG" = "true" ]; then \
-        maturin build --release --features "media-ffmpeg,kv-indexer,slot-tracker,select-service,mm-routing,aic-forward-pass" --out /opt/dynamo/dist; \
+        maturin build --release --features "media-ffmpeg,kv-indexer,slot-tracker,select-service,mm-routing,aic-forward-pass,request-trace-s3" --out /opt/dynamo/dist; \
```
nnshah1: @nv-anants , @saturley-hall for viz in case of any issue w.r.t dependencies

lib/bindings/python/Cargo.toml
```diff
@@ -40,6 +40,11 @@ nvtx = ["dynamo-runtime/nvtx"]
 # recipes pass this flag explicitly; the dev container's
 # `.devcontainer/post-create.sh` also enables it for local workflows.
 mm-routing = ["dynamo-llm/mm-routing"]
+# Enable the native S3 destination for the request-trace sink. Pulls in
+# aws-sdk-s3 + aws-config so the default wheel stays lean; production wheel
```
nnshah1, line 46: question: is this enabled in local container builds?

lib/runtime/src/config/environment_names.rs
```diff
@@ -476,11 +476,16 @@ pub mod llm {
...
 
-        /// Request trace sink selection. Comma-separated values: `file`, `stderr`, `nats`, `otel`.
+        /// Request trace sink selection. Comma-separated values: `file`,
+        /// `stderr`, `nats`, `otel`, `s3`.
         ///
         /// Legacy values map as follows: `jsonl` => `file` with `jsonl` format,
         /// `jsonl_gz` => `file` with `jsonl_gz` format, `stderr` => `stderr`,
         /// `nats` => `nats`, and `otel` => `otel`.
+        ///
+        /// The `s3` sink requires `dynamo-llm` to be built with the
```
nnshah1: this comment seems out of place -

lib/bindings/python/Cargo.toml
YiqiuLiu: Currently no — `maturin develop` in the local paths does not pass `--features request-trace-s3`, so `DYN_REQUEST_TRACE_SINKS=s3` fails on locally-built dev containers even though shipped wheels work. Four places would need the flag alongside the existing `mm-routing`:

1. `.devcontainer/post-create.sh:105` — VS Code Remote / Codespaces auto-bootstrap
2. `container/launch_message/dev.txt:11` — banner shown when users enter the dev container
3. `container/README.md:487` — dev build instructions
4. `container/README.md:547` — parallel dev build snippet lower in the same doc

Want it in all four, or just a subset? Happy to enable in whichever scope matches your intent.

lib/runtime/src/config/environment_names.rs
YiqiuLiu: Agreed. Dropped the paragraph in the next amend. The cargo-feature detail is a build-system concern, not an env-var-name concern; the S3 sink docs in `docs/observability/request-tracing.md` are the right place for it, and the individual `DYN_REQUEST_TRACE_S3_{BUCKET,REGION,PREFIX}` constants below already carry their own docstrings.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,463 @@
...
+                match message {
+                    Some(record) => {
+                        if let Err(error) = batch.push(&record) {
+                            tracing::warn!(
+                                target: "dynamo_llm::request_trace",
+                                %error,
+                                "request trace s3: serialize failed; dropping record"
+                            );
+                        } else if batch.uncompressed_bytes() >= roll_uncompressed_bytes {
+                            upload_ready_batch(&uploader, &mut batch, &mut seq).await;
```
nnshah1, line 260: **P2 — the S3 upload is `.await`-ed inline on the single drain task, so a slow `PutObject` blocks ingestion.**

Building on @tmonty12's upload-timeout bound (`:71`): even at the bounded 90s operation timeout, `run_worker` can't `rx.recv()` while a batch uploads, so the 2048-slot channel fills and `emit`'s `try_send` drops records — silently and correlated, exactly when S3 is degraded.

Same drain shape as `telemetry::jsonl_gz` (mpsc(2048), biased `select!`, flush/roll/gzip), but there the inline IO is a local disk flush (~ms) vs a 90s network call here.

Non-blocking / good follow-up: hand finished batches to a bounded set of concurrent upload tasks (semaphore-guarded spawn) so `recv` keeps draining. Cheap immediate win: a dropped-records counter/metric so the loss is at least observable. On the caller path this sink is already the best-isolated of the trace sinks (`try_send`/drop vs the file sink's blocking `.await`) — this just tightens the worker side.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,463 @@
...
+#[async_trait]
+impl RequestTraceSink for S3RequestTraceSink {
+    fn name(&self) -> &'static str {
+        "s3"
+    }
+
+    async fn emit(&self, record: &RequestTraceRecord) {
+        if let Err(error) = self.tx.try_send(record.clone()) {
+            match error {
+                mpsc::error::TrySendError::Full(_) => tracing::warn!(
```
nnshah1: **P3 (nit) — per-record `warn!` on backpressure floods logs.** One warn line per dropped record, under exactly the load that fills the channel, contending on the log-output lock. Prefer a dropped-records counter with a rate-limited/periodic warn (pairs with the metric suggested on the drain-loop finding).

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,463 @@
...
+            }
+        }
+    }
+
+    async fn shutdown(&self) {
+        self.shutdown.cancel();
+        let worker = self
+            .worker
+            .lock()
+            .expect("s3 sink shutdown mutex poisoned")
```
nnshah1: **Nit — `expect()` on the shutdown mutex in non-test code.** It only guards `Option<JoinHandle>` so poisoning is unlikely, but a poisoned lock turns a prior panic into a second panic during teardown. `lock().unwrap_or_else(|e| e.into_inner())` (or log-and-return) avoids that.

lib/llm/src/request_trace/s3_sink.rs
```diff
@@ -0,0 +1,463 @@
...
+    let doy = doe - (365 * yoe + yoe / 4 - yoe / 100);
+    let mp = (5 * doy + 2) / 153;
+    let d = doy - (153 * mp + 2) / 5 + 1;
+    let m = if mp < 10 { mp + 3 } else { mp - 9 };
+    let year = if m <= 2 { y + 1 } else { y };
+    (year, m as u32, d as u32, hh, mi, ss)
+}
+
+#[cfg(test)]
+mod tests {
```
nnshah1, line 446: **Test coverage — behavior gap.** The unit tests cover the pure `utc_date_parts` / `object_key` helpers well, but nothing exercises the parts that carry the risk: the `run_worker` drain loop, `emit` backpressure/drop, the shutdown drain, or upload failure/discard. A test that stalls the uploader and asserts `recv` keeps draining, and one asserting bounded log output under sustained `Full`, would cover behavior rather than the easy pure functions. (Static review — tests not run.)

nnshah1 (commented):
Post-approval pass (graham-style). Most prior findings are already fixed and @tmonty12 has approved — these are net-new: one substantive item on the drain/upload path (non-blocking; builds on the timeout Thomas already added) plus two nits and a test-coverage note.

— Neelay + 🤖

lib/bindings/python/Cargo.toml
nnshah1: I think its fine to keep as is - we can add a not in the dev build instructions if it possible to customize the dev build to add it. @keivenchang for viz.

nnshah1 (approved):
Open question on not blocking the trace ingestion thread for the s3 upload - we could move that to a seperate thread - have a few opens to ensure ops team sees the additional dependencies.

nealvaidya (approved):
approve docs

nv-anants (commented):
new deps are all permissive, looks good

lib/bindings/python/Cargo.toml
YiqiuLiu: Done. Kept the default local build lean and added a note next to both `maturin develop` snippets in `container/README.md`: add `--features request-trace-s3` to enable the S3 request-trace sink (`DYN_REQUEST_TRACE_SINKS=s3`).

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Addressed the observable-loss half now; deferred the concurrency rework.

- **Drop counter now:** `emit` no longer silently loses records — an `AtomicU64` counts every drop, one warning fires on the first drop, and a summary with the total fires at shutdown. Mirrors the OpenTelemetry Rust `BatchLogProcessor` drop-accounting pattern.
- **Concurrent uploads deferred:** hand-off to a bounded concurrent-upload pool (so `recv` keeps draining) is a follow-up, documented in a new module-doc note on `s3_sink.rs`. Worth noting the single-drain shape matches OTel `BatchLogProcessor` and the local `telemetry::jsonl_gz` writer, so this is a known trade-off rather than a regression; it belongs with the object-layout PR where the sink is restructured to carry more context.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Fixed. Replaced the per-record `warn!` with an `AtomicU64` drop counter: the first drop logs one warning, subsequent drops bump the counter silently, and a single summary with the running total logs at shutdown. No more one-line-per-dropped-record under backpressure. Same drop-accounting shape as the OpenTelemetry Rust `BatchLogProcessor`.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Fixed. Changed the shutdown lock to `self.worker.lock().unwrap_or_else(|poisoned| poisoned.into_inner())`, so a poisoned lock recovers the guard instead of turning a prior panic into a second panic during teardown.

lib/llm/src/request_trace/s3_sink.rs
YiqiuLiu: Added behavior tests for the drop path. Two new unit tests in `s3_sink.rs`:

- `emit_drops_and_counts_when_channel_is_full` — builds a sink whose bounded channel is never drained (the stalled-uploader condition), sends past capacity, and asserts the drop counter equals `total - capacity`.
- `note_dropped_warns_only_on_first_drop` — asserts the warning fires exactly once across 1001 drops while the counter still reflects all of them.

The "assert `recv` keeps draining under a stalled uploader" test is deferred to the follow-up that adds concurrent uploads — with the current single-drain design the drain genuinely blocks during an upload, so there is no drain-progress behavior to assert yet. The new tests cover `emit` backpressure and the drop accounting, which is the risk this PR introduces.

nnshah1:
/ok to test https://github.com/ai-dynamo/dynamo/commit/c3a50dc09620bdc93fe1c546223ea8a847fcf98f

nnshah1:
/ok to test https://github.com/ai-dynamo/dynamo/commit/c3a50dc09620bdc93fe1c546223ea8a847fcf98f

nv-anants (approved):

nv-anants:
/ok to test [ef0241e](https://github.com/ai-dynamo/dynamo/pull/11806/commits/ef0241e0bc7b759f3fd1df753dd9d9147388f828)

nv-anants:
/ok to test [9b417aa](https://github.com/ai-dynamo/dynamo/pull/11806/commits/9b417aa1e34b012cb48ceada51d0abfd97c57e28)

nv-anants:
/ok to test [34c5379](https://github.com/ai-dynamo/dynamo/pull/11806/commits/34c5379ac94b262f26fa90f067158cdf2b4b4f03)

nv-anants:
/ok to test [fedc2da](https://github.com/ai-dynamo/dynamo/pull/11806/commits/fedc2dabfb817f18499c296821306292dee12ed7)

PR #11809, opened 2026-07-16
fix(gms): make gpu_memory_service logs visible in the vLLM worker

mohammedabdulwahhab:
Fix gpu_memory_service logger when used inside a vllm worker process context. Previously there was no associated handler for these and INFO and DEBUG logs were getting dropped.

MX [handler](https://github.com/ai-dynamo/modelexpress/blob/4ab220619628bf5940ea4210cfe34d23c2a81298/modelexpress_client/python/modelexpress/__init__.py#L29) follows similar pattern.

## Summary by CodeRabbit

* **New Features**
  * Added configurable logging levels for GPU Memory Service worker processes through an environment variable.
  * Integrated GPU Memory Service logs with vLLM worker logging for consistent visibility and formatting.

* **Bug Fixes**
  * Improved projected memory availability reporting by routing messages through the configured logging system.


lib/gpu_memory_service/common/utils.py
```diff
@@ -17,6 +17,9 @@
 # constants in deploy/operator/internal/gms/gms.go.
 ENV_SCRATCH_KV_ENABLED = "DYN_GMS_SCRATCH_KV_ENABLED"
 ENV_VMM_GRANULARITY = "DYN_GMS_VMM_GRANULARITY"
+# Optional override for the gpu_memory_service log level (a logging level name,
+# e.g. "DEBUG"); when unset the vLLM worker integration falls back to vLLM's level.
+ENV_LOG_LEVEL = "DYN_GMS_LOG_LEVEL"
```
mohammedabdulwahhab: This env var is in common. How does it interact with the other contexts this library can be used in (GMS server, dynamo.vllm, etc). Should we make any changes?

galletas1712 (approved):

PR #11811, opened 2026-07-16
fix(container): source-build NIXL for the sglang EFA runtime image (DYN-3429)

dagil-nvidia:
## What

Fixes the SGLang CUDA **EFA** runtime hang (DYN-3429 / NVBug 6431579) by building a full `nixl-cu13` wheel from source and overlaying it onto the sglang EFA runtime image, replacing the inherited upstream `nixl-cu13==1.3.0`.

## Why

The upstream `lmsysorg/sglang` runtime ships `nixl-cu13==1.3.0`, whose libfabric rail manager issues unconditional `FI_MORE` batched writes on `efa`/`efa-direct` providers and hangs on AWS EFA. The fix is upstream-merged in `ai-dynamo/nixl` at commit **`322ba2a7`** (HEAD of `release/1.3.1`): it adds `is_efa_provider` and disables `FI_MORE` batching on EFA. This consumes that merged fix directly.

We pin the `release/1.3.1` **branch commit `322ba2a7`**, not the `v1.3.1` tag — the tag diverged from the branch and does **not** carry the fix.

This **supersedes** the interim binary-plugin-swap approach in #11549 (same base branch). Core and plugin both come from `322ba2a7`, so the overlay is ABI-coherent.

## How (scoped to the SGLang cuda EFA runtime target only)

- `container/context.yaml`: `sglang.cuda13.0.nixl_efa_ref = 322ba2a7`, consumed only by the EFA runtime build.
- `container/templates/args.Dockerfile`: resolve `NIXL_REF` to `nixl_efa_ref` only when `make_efa` and `target=runtime`.
- `container/templates/wheel_builder.Dockerfile`: build the `nixl-cu13` wheel for the sglang cuda EFA runtime (previously skipped for all sglang cuda runtime/dev/local-dev).
- `container/templates/aws.Dockerfile`: overlay the wheel_builder-built `nixl-cu13` wheel over the inherited one, mirroring the existing trtllm EFA NIXL override.

Rendering the Dockerfile matrix confirms only the sglang EFA runtime render pins `NIXL_REF=322ba2a7`, builds the wheel, and applies the overlay; non-EFA sglang runtime/dev images and every other framework are unchanged.

## Testing

- Verified: `container/render.py` renders the full framework/target/device/EFA matrix cleanly under StrictUndefined; scoping confirmed by grepping the rendered outputs (only sglang cuda EFA runtime changes).
- Still required (needs hardware, not runnable in the authoring env): multi-arch image build of the sglang EFA runtime, then the 9-case SGLang EFA regression on an AWS EFA cluster.

### DIFF
--- a/container/context.yaml
+++ b/container/context.yaml
@@ -104,6 +104,14 @@ sglang:
     # GPL ffmpeg/codec dpkgs; the GPL imageio-ffmpeg pip binary is replaced by an
     # in-tree LGPL ffmpeg in sglang_runtime.Dockerfile — no dpkg purge needed.)
     baseline_sbom: cuda@8b2705ea
+    # DYN-3429 / NVBug 6431579: EFA FI_MORE hang fix. Only the AWS EFA runtime
+    # overlay (templates/aws.Dockerfile) consumes this, installing a Dynamo-built
+    # nixl-cu13 wheel at this ref over the upstream lmsysorg/sglang nixl-cu13==1.3.0.
+    # Applied only when make_efa is true (see templates/args.Dockerfile), so non-EFA
+    # sglang runtime/dev images keep the framework default nixl_ref (v1.0.1) and the
+    # upstream runtime NIXL stack. Pin the ai-dynamo/nixl release/1.3.1 BRANCH commit;
+    # the v1.3.1 TAG diverged and does NOT carry the fix.
+    nixl_efa_ref: 322ba2a7
   xpu:
     base_image: intel/deep-learning-essentials
     base_image_tag: 2025.3.2-0-devel-ubuntu24.04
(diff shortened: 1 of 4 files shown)


dagil-nvidia:
Closing: the root cause is not FI_MORE. The image's libfabric 2.5.1 was built against AlmaLinux-8 rdma-core 48.0, which lacks the efadv symbols, so configure sets HAVE_EFA_DATA_PATH_DIRECT=0 and it falls back to the hanging rdma-core completion path. Rebuilding the same 2.5.1 against rdma-core 63.0 (in wheel_builder.Dockerfile) compiles data-path-direct in and passes 15/15 with stock NIXL 1.2.0 -- no NIXL/FI_MORE change needed. Pursuing that libfabric-vs-rdma-core build fix instead.

PR #11820, opened 2026-07-17
perf(logging): avoid dynamic filter span locks

jthomson04:
## Overview:

Use tracing-subscriber's metadata-only `Targets` filter when `DYN_LOG` and Dynamo's configured filters contain only static target/level directives. Preserve `EnvFilter` for dynamic span or field directives, whose matching requires its span-state machinery.

## Details:

- Removes the shared `EnvFilter` span-matcher `RwLock` from the normal static request-span lifecycle while retaining `request_span=trace` and existing span-event behavior.
- Preserves `EnvFilter::parse_lossy` behavior for `DYN_LOG`: zero-length comma-separated segments are ignored without trimming non-empty segments, and the configured default level applies only when no directives remain.
- A non-empty target-only `DYN_LOG` leaves unrelated targets disabled, as it does with `EnvFilter`.
- A dynamic `DYN_LOG` directive or a dynamic configured `log_filters` directive falls back to `EnvFilter`; directives invalid for both parsers retain the existing warning-and-ignore behavior.

Start with `lib/runtime/src/logging.rs`, particularly `targets_filter` and its regression tests. The function selects the static filter only when the complete effective directive set is compatible with `Targets`.

## Related Issues

## Performance impact

`request_span=trace` is a static directive, but the prior implementation always constructed `EnvFilter`. Tower/Hyper enters and exits the request span on every streaming response-body poll; `EnvFilter` takes its dynamic-span lock on these lifecycle callbacks even when no dynamic directives exist. On a 36+36 CPU split-NUMA frontend, that lock consumed 52.105% of frontend samples.

Matched Tyche AgentX c=2048 results (`Qwen/Qwen3-0.6B`, 32 mockers, 1e6 mocker speedup):

- exclusive `EnvFilter` samples: 52.105% -> 0.000%
- frontend CPU: 72.67 -> 35.81 ms/request
- throughput: 932.06 -> 1,133.44 req/s
- post-ramp TTFT p50/p95/p99: 1595.70/2111.06/2233.60 -> 180.93/1026.67/1311.83 ms

The whole-run p99 is ramp-dominated in the patched run: all requests at or above its whole-run p99 threshold began by 24.70 seconds. The post-30-second tail is the representative comparison.

## Validation

- `cargo test -p dynamo-runtime logging::tests:: -- --nocapture`
- `cargo clippy -p dynamo-runtime --all-targets -- -D warnings`
- Tyche two-node matched on-CPU profile: 226,267 samples, zero lost samples/chunks, zero AIPerf errors or cancellations, and load-node CPU below contamination gates (43.34% average, 75.17% p95).

Benchmark artifacts remain local to the Tyche harness and are not included in this PR.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved logging filter performance by using a faster static filtering approach whenever supported.
  * Preserved support for advanced dynamic logging directives through an automatic fallback.
  * Maintained existing request and span tracing behavior.

### DIFF
--- a/lib/runtime/src/logging.rs
+++ b/lib/runtime/src/logging.rs
@@ -1520,22 +1520,42 @@ fn filters(config: LoggingConfig) -> LoggingFilter {
 /// `RwLock`, and its span lifecycle callbacks acquire that lock even when the
 /// configured directives are all static.
 fn targets_filter(config: &LoggingConfig, dyn_log: Option<&str>) -> Option<Targets> {
-    let mut directives = vec![config.log_level.clone()];
+    // `EnvFilter::parse_lossy` ignores zero-length comma-separated segments,
+    // but leaves all other text untouched. In particular, do not trim here:
+    // whitespace may make the directive dynamic or invalid.
+    let dyn_directives = dyn_log
+        .into_iter()
+        .flat_map(|value| value.split(',').filter(|directive| !directive.is_empty()));
 
-    if let Some(value) = dyn_log {
+    let mut directives = Vec::new();
+    for directive in dyn_directives {
         // Parsing as `Targets` is also the feature test for whether the
-        // configured directives require EnvFilter's dynamic span matching.
-        value.parse::<Targets>().ok()?;
-        directives.push(value.to_string());
+        // configured directive requires EnvFilter's dynamic span matching.
+        directive.parse::<Targets>().ok()?;
+        directives.push(directive.to_string());
+    }
+
+    // EnvFilter uses the configured default only when DYN_LOG is absent or
+    // contains no directives. A target-only DYN_LOG must leave other targets
+    // disabled rather than inheriting the configured default level.
+    if directives.is_empty() {
+        directives.push(config.log_level.clone());
     }
 
     for (module, level) in &config.log_filters {
         let directive = format!("{module}={level}");
         match directive.parse::<Targets>() {
             Ok(_) => directives.push(directive),
-            Err(e) => {
-                eprintln!("Failed parsing filter '{level}' for module '{module}': {e}");
-            }
+            Err(_) => match directive.parse::<Directive>() {
+                // Valid span or field directives require EnvFilter's dynamic
+                // matching, so do not silently drop a configured filter.
+                Ok(_) => return None,
+                // Preserve the pre-fast-path warn-and-ignore behavior for
+                // directives that neither parser accepts.
+                Err(e) => {
+                    eprintln!("Failed parsing filter '{level}' for module '{module}': {e}");
+                }
+            },
         }
     }
 
@@ -1919,7 +1939,31 @@ pub mod tests {
     use tempfile::NamedTempFile;
 
     #[test]
-    fn target_level_directives_use_static_filter() {
+    fn unset_dyn_log_uses_configured_default() {
+        let filter = targets_filter(&LoggingConfig::default(), None)
+            .expect("an unset DYN_LOG should use Targets");
+
+        assert!(filter.would_enable("request_span", &tracing::Level::TRACE));
+        assert!(filter.would_enable("unlisted_target", &tracing::Level::INFO));
+        assert!(!filter.would_enable("unlisted_target", &tracing::Level::DEBUG));
+        assert!(!filter.would_enable("tower", &tracing::Level::WARN));
+    }
+
+    #[test]
+    fn empty_dyn_log_uses_configured_default() {
+        for dyn_log in ["", ",,"] {
+            let filter = targets_filter(&LoggingConfig::default(), Some(dyn_log))
+                .expect("an empty DYN_LOG should use Targets");
+
+            assert!(filter.would_enable("request_span", &tracing::Level::TRACE));
+            assert!(filter.would_enable("unlisted_target", &tracing::Level::INFO));
+            assert!(!filter.would_enable("unlisted_target", &tracing::Level::DEBUG));
+            assert!(!filter.would_enable("tower", &tracing::Level::WARN));
+        }
+    }
+
+    #[test]
+    fn target_only_dyn_log_does_not_use_configured_default() {
         let filter = targets_filter(
             &LoggingConfig::default(),
             Some("dynamo_runtime::logging=debug"),
@@ -1927,10 +1971,22 @@ pub mod tests {
         .expect("target/level directives should use Targets");
 
         assert!(filter.would_enable("request_span", &tracing::Level::TRACE));
-        assert!(filter.would_enable("unlisted_target", &tracing::Level::INFO));
+        assert!(!filter.would_enable("unlisted_target", &tracing::Level::INFO));
         assert!(!filter.would_enable("unlisted_target", &tracing::Level::DEBUG));
         assert!(filter.would_enable("dynamo_runtime::logging::child", &tracing::Level::DEBUG));
         assert!(!filter.would_enable("tower", &tracing::Level::WARN));
+    }
+
+    #[test]
+    fn trailing_empty_dyn_log_segments_do_not_enable_trace() {
+        let filter = targets_filter(
+            &LoggingConfig::default(),
+            Some("dynamo_runtime::logging=debug,"),
+        )
+        .expect("target/level directives should use Targets");
+
+        assert!(filter.would_enable("dynamo_runtime::logging", &tracing::Level::DEBUG));
+        assert!(!filter.would_enable("unlisted_target", &tracing::Level::TRACE));
     }
 
     #[test]
@@ -1942,6 +1998,19 @@ pub mod tests {
             )
             .is_none()
         );
+    }
+
+    #[test]
+    fn dynamic_log_filters_fall_back_to_env_filter() {
+        let config = LoggingConfig {
+            log_level: DEFAULT_FILTER_LEVEL.to_string(),
+            log_filters: HashMap::from([(
+                "dynamo_runtime[request{model=foo}]".to_string(),
+                "debug".to_string(),
+            )]),
+        };
+
+        assert!(targets_filter(&config, None).is_none());
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,137 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
...
+/// `RwLock`, and its span lifecycle callbacks acquire that lock even when the
+/// configured directives are all static.
+fn targets_filter(config: &LoggingConfig, dyn_log: Option<&str>) -> Option<Targets> {
+    let mut directives = vec![config.log_level.clone()];
+
+    if let Some(value) = dyn_log {
+        // Parsing as `Targets` is also the feature test for whether the
+        // configured directives require EnvFilter's dynamic span matching.
+        value.parse::<Targets>().ok()?;
+        directives.push(value.to_string());
```
jthomson04: **High: empty `DYN_LOG` segments enable TRACE for everything.**

`Targets::from_str` does **not** filter empty segments (unlike `EnvFilter::parse_lossy`, which does). An empty segment parses as `StaticDirective { target: Some(""), level: TRACE }`, and target matching uses `starts_with` — an empty target prefix-matches **every** target at TRACE. Since `Some("")` sorts as more specific than the bare default level, it wins first-match.

Concrete failure with `DYN_LOG=""` (set-but-empty, common in shell scripts and k8s manifests) or a trailing comma (`DYN_LOG="dynamo=debug,"`):
1. `value.parse::<Targets>()` succeeds, so the feature-test passes
2. the raw value is pushed and joined → `"info,,h2=error,..."`
3. the empty segment becomes `""=TRACE` → **the entire process logs at TRACE**

Previously both cases resolved to the plain `info` default via `parse_lossy`'s empty-segment filter. A TRACE flood on the frontend is worse than the lock contention this PR fixes.

Fix: split `DYN_LOG` on `,`, trim, and drop empty segments before both the feature-test parse and the join.

lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,137 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    LoggingFilter::Env(env_filter(&config))
+}
+
+/// Use the lock-free target/level filter when `DYN_LOG` contains no dynamic
+/// span or field directives. `EnvFilter` tracks dynamic span matches behind an
+/// `RwLock`, and its span lifecycle callbacks acquire that lock even when the
+/// configured directives are all static.
+fn targets_filter(config: &LoggingConfig, dyn_log: Option<&str>) -> Option<Targets> {
+    let mut directives = vec![config.log_level.clone()];
```
jthomson04: **Medium-high: default-level semantics change when `DYN_LOG` has only target-specific directives.**

`EnvFilter`'s builder only adds `with_default_directive(config.log_level)` when the env var yields **zero** directives (`from_directives` in tracing-subscriber's `env/builder.rs`). This line pushes `config.log_level` **unconditionally**.

- Old: `DYN_LOG=dynamo_llm=debug` → unmatched targets are **silent** (standard `RUST_LOG` semantics)
- New: `DYN_LOG=dynamo_llm=debug` → unmatched targets log at **INFO**

The new test asserts the new behavior (`would_enable("unlisted_target", INFO)`), so this may be intentional — but it contradicts the PR description's claim of preserved behavior, and the two code paths now **disagree with each other**: adding a dynamic directive (`foo[span]=debug`) flips you to the `EnvFilter` fallback where unmatched targets go silent again. Filter syntax shouldn't change the default level of unrelated targets.

Fix: only push `config.log_level` when `DYN_LOG` is unset or yields no directives, mirroring the builder. Combined with the empty-segment fix: build the directive list from filtered segments, and push the default only if that list is empty.

lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,137 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
...
+        value.parse::<Targets>().ok()?;
+        directives.push(value.to_string());
+    }
+
+    for (module, level) in &config.log_filters {
+        let directive = format!("{module}={level}");
+        match directive.parse::<Targets>() {
+            Ok(_) => directives.push(directive),
+            Err(e) => {
+                eprintln!("Failed parsing filter '{level}' for module '{module}': {e}");
```
jthomson04: **Medium-low: dynamic directives in `config.log_filters` are dropped instead of triggering the `EnvFilter` fallback.**

The fallback feature-test only inspects `DYN_LOG`. A `log_filters` config entry that parses as an `EnvFilter` `Directive` but not as `Targets` (e.g. key `"dynamo[request]"`) was honored by the old code but is now dropped with an `eprintln` while staying on the `Targets` path.

The built-in defaults are all static `module=level`, so this only bites user TOML configs, but the fix is small: if the directive fails `Targets` parsing but succeeds as a `Directive`, return `None` to fall back to `EnvFilter`; only `eprintln`-and-skip when it parses as neither (parity with the old behavior).

jthomson04 (commented):
Reviewed the `Targets` fast-path change. The optimization is sound and the root-cause analysis checks out — I verified in tracing-subscriber 0.3.23 source that `EnvFilter::on_enter`/`on_exit` unconditionally take the `by_id` `RwLock` even with zero dynamic directives, while `Targets` is lock-free. The `Filter<S>` delegation enum covers all 9 trait methods, and directive-precedence semantics (replace-on-equal-target, most-specific-first match) are shared between `Targets` and `EnvFilter` statics, so ordering parity holds.

That said, I found **one high-severity edge case** (empty `DYN_LOG` segments enable global TRACE logging) and **one user-visible semantic change** (default-level behavior when `DYN_LOG` contains only target-specific directives) — see inline comments. Both can be addressed with one small refactor of `targets_filter`.

**Test coverage gaps:**
- No test for `DYN_LOG` unset (`targets_filter(&config, None)`)
- No test for empty string / trailing comma (would have caught the TRACE-flood edge case)
- No test that the `Targets` path and the `EnvFilter` fallback agree on default-level semantics
- No test for a dynamic directive in `log_filters`

lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,136 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-fn filters(config: LoggingConfig) -> EnvFilter {
+enum LoggingFilter {
+    Targets(Targets),
+    Env(EnvFilter),
+}
+
+impl<S> Filter<S> for LoggingFilter {
```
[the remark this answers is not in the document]
jthomson04: Verified against the pinned `tracing-subscriber` 0.3.23 source: `filter/env/mod.rs:708` implements `Filter<S>` for `EnvFilter` without a `LookupSpan` bound. The current implementation also compiles under `cargo clippy -p dynamo-runtime --all-targets -- -D warnings`, so adding that bound would unnecessarily tighten the intended subscriber compatibility.

lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,136 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
...
+    }
+
+    for (module, level) in &config.log_filters {
+        let directive = format!("{module}={level}");
+        match directive.parse::<Targets>() {
+            Ok(_) => directives.push(directive),
+            Err(e) => {
+                eprintln!("Failed parsing filter '{level}' for module '{module}': {e}");
+            }
+        }
```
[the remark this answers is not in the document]
jthomson04: Fixed in [d5ae82ff05](https://github.com/ai-dynamo/dynamo/commit/d5ae82ff0583916e229af211a6f15512351a9da7): a configured `log_filters` directive that `Targets` rejects but `Directive` accepts now selects the existing `EnvFilter` path instead of being dropped. The regression uses `dynamo_runtime[request{model=foo}]=debug`.

lib/runtime/src/logging.rs
```diff
@@ -1421,13 +1423,136 @@ fn setup_logging() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    LoggingFilter::Env(env_filter(&config))
+}
+
+/// Use the lock-free target/level filter when `DYN_LOG` contains no dynamic
+/// span or field directives. `EnvFilter` tracks dynamic span matches behind an
+/// `RwLock`, and its span lifecycle callbacks acquire that lock even when the
+/// configured directives are all static.
+fn targets_filter(config: &LoggingConfig, dyn_log: Option<&str>) -> Option<Targets> {
+    let mut directives = vec![config.log_level.clone()];
```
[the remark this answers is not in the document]
jthomson04: Fixed in [d5ae82ff05](https://github.com/ai-dynamo/dynamo/commit/d5ae82ff0583916e229af211a6f15512351a9da7): the configured default is now added only when `DYN_LOG` has no non-empty segments. A target-only `DYN_LOG` therefore leaves unlisted targets disabled, matching `EnvFilter`; `target_only_dyn_log_does_not_use_configured_default` covers this.

lib/runtime/src/logging.rs
jthomson04: Fixed in [d5ae82ff05](https://github.com/ai-dynamo/dynamo/commit/d5ae82ff0583916e229af211a6f15512351a9da7): `DYN_LOG` now uses the same comma split as `EnvFilter::parse_lossy`, discarding only zero-length segments and not trimming non-empty text. Regressions cover literal empty input, all-empty `,,`, and a trailing comma without enabling unrelated targets at TRACE.

lib/runtime/src/logging.rs
jthomson04: Addressed by the shared default-level fix in [d5ae82ff05](https://github.com/ai-dynamo/dynamo/commit/d5ae82ff0583916e229af211a6f15512351a9da7): non-empty target-only `DYN_LOG` no longer seeds the configured default, so unrelated targets remain disabled. The added regression verifies that behavior.

lib/runtime/src/logging.rs
jthomson04: Addressed by the shared fallback fix in [d5ae82ff05](https://github.com/ai-dynamo/dynamo/commit/d5ae82ff0583916e229af211a6f15512351a9da7): any configured `log_filters` directive that requires dynamic span or field matching now retains the `EnvFilter` path rather than being silently ignored. The new field-matching TOML regression covers it.

ishandhanani (approved):

Issue #11661, opened 2026-07-14
DEP: EPP Embedded SelectionService Interface


tmonty12:
## Table of Contents

* [Summary](<#user-content-summary>)
* [Motivation](<#user-content-motivation>)
* [Proposal](<#user-content-proposal>)
  * [Design Inputs](<#user-content-design-inputs>)
  * [Component Boundary](<#user-content-component-boundary>)
  * [Motivation](<#user-content-motivation-2>)
  * [Goals](<#user-content-goals>)
  * [Worker Discovery](<#user-content-worker-discovery>)
  * [In-Process Interface](<#user-content-in-process-interface>)
  * [Aggregated Request Lifecycle](<#user-content-aggregated-request-lifecycle>)
  * [Tokenization](<#user-content-tokenization>)
    * [Problem with the Dynamo Preprocessor](<#user-content-problem-with-the-dynamo-preprocessor>)
    * [Initial Implementation: External Tokenizer Service](<#user-content-initial-implementation-external-tokenizer-service>)
      * [Backend Tokenization Services](<#user-content-backend-tokenization-services>)
    * [Long-Term: Tokens-In/Tokens-Out Execution](<#user-content-long-term-tokens-in-tokens-out-execution>)
  * [Disaggregated Routing](<#user-content-disaggregated-routing>)
    * [Overview](<#user-content-overview>)
    * [Plane 1: EPP Selection and Accounting](<#user-content-plane-1-epp-selection-and-accounting>)
    * [Plane 2: Decode Sidecar Execution](<#user-content-plane-2-decode-sidecar-execution>)
    * [Backend Execution Contracts](<#user-content-backend-execution-contracts>)
      * [vLLM](<#user-content-vllm-2>)
      * [SGLang](<#user-content-sglang-2>)
      * [TensorRT-LLM](<#user-content-tensorrt-llm-3>)
  * [Multi-Replica EPP State Sharing](<#user-content-multi-replica-epp-state-sharing-2>)
    * [Peer Discovery](<#user-content-peer-discovery>)
    * [State-Sharing Planes](<#user-content-state-sharing-planes>)
  * [EPP Embedded SelectionService vs Full Dynamo Runtime Router](<#user-content-epp-embedded-selectionservice-vs-full-dynamo-runtime-router>)
    * [Functional Gaps](<#user-content-functional-gaps>)
* [Implementation Milestones](<#user-content-implementation-milestones>)
[...]

PR #11827, opened 2026-07-17
feat(ext-proc): add vLLM render client for standalone EPP

tmonty12:
## Summary

- replace the offline tokenizer approach proposed in #11071 with a pooled HTTP client for vLLM's `/v1/chat/completions/render` endpoint
- add explicit standalone configuration through `DYN_EPP_VLLM_RENDER_URL` and the provider-neutral `DYN_EPP_TOKENIZATION_TIMEOUT_MS` (5-second default)
- classify renderer failures as typed `Unavailable`, `Timeout`, `UpstreamStatus`, or `InvalidResponse` errors
- forward the original OpenAI chat-completions JSON body unchanged and parse only the `token_ids` needed for routing

## Scope

This PR is intentionally client/config plumbing only. It does not wire tokenization into endpoint selection and does not choose fallback or request-failure behavior; that policy belongs in the later request-path integration PR.

This PR supports vLLM Render only. SGLang tokenizer support remains a separate follow-up with its own protocol client; both providers can consume the shared tokenization timeout.

## PR train

- stacked on #11070
- intended to replace #11071 without modifying that PR
- aligns with milestone 1 in #11661 (standalone EPP, aggregated vLLM)

## Validation

- `cargo fmt -p dynamo-ext-proc -- --check`
- `cargo test -p dynamo-ext-proc` (28 unit tests + 1 integration test)
- `cargo clippy -p dynamo-ext-proc --no-deps --all-targets -- -D warnings`

## Summary by CodeRabbit

* **New Features**
  * Added support for rendering chat requests through a vLLM tokenizer service.
  * Added configuration for the tokenizer service URL and request timeout.
  * Added validation to ensure the service URL and timeout settings are valid.
  * Added clear handling for service timeouts, unavailable services, upstream errors, and invalid responses.

### DIFF
--- a/deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
@@ -15,7 +15,7 @@ use validator::ValidationError;
 
 const DEFAULT_KV_EVENT_PORT: u16 = 5557;
 const DEFAULT_SELECTOR_THREADS: usize = 4;
-const DEFAULT_VLLM_RENDER_TIMEOUT_MS: u64 = 5_000;
+const DEFAULT_TOKENIZATION_TIMEOUT_MS: u64 = 5_000;
 
 /// Environment variable that selects the EPP operating mode.
 pub const DYN_EPP_MODE: &str = "DYN_EPP_MODE";
@@ -75,9 +75,9 @@ pub struct EppStandaloneConfig {
     #[validate(length(min = 1, message = "DYN_EPP_VLLM_RENDER_URL is required"))]
     #[validate(custom(function = "validate_vllm_render_url"))]
     pub vllm_render_url: String,
-    /// Deadline for calls to the vLLM renderer.
-    #[validate(range(min = 1, message = "DYN_EPP_VLLM_RENDER_TIMEOUT_MS must be >= 1"))]
-    pub vllm_render_timeout_ms: u64,
+    /// Deadline for calls to the configured tokenization provider.
+    #[validate(range(min = 1, message = "DYN_EPP_TOKENIZATION_TIMEOUT_MS must be >= 1"))]
+    pub tokenization_timeout_ms: u64,
     /// KV-cache block size; MUST equal the inference engine block size.
     #[validate(range(min = 1, message = "DYN_KV_CACHE_BLOCK_SIZE must be >= 1"))]
     pub block_size: u32,
@@ -118,8 +118,8 @@ impl EppStandaloneConfig {
             namespace: trimmed(get("POD_NAMESPACE")).unwrap_or_default(),
             model_name: trimmed(get("DYN_MODEL_NAME")).unwrap_or_default(),
             vllm_render_url: trimmed(get("DYN_EPP_VLLM_RENDER_URL")).unwrap_or_default(),
-            vllm_render_timeout_ms: opt_parse::<u64>(get, "DYN_EPP_VLLM_RENDER_TIMEOUT_MS")?
-                .unwrap_or(DEFAULT_VLLM_RENDER_TIMEOUT_MS),
+            tokenization_timeout_ms: opt_parse::<u64>(get, "DYN_EPP_TOKENIZATION_TIMEOUT_MS")?
+                .unwrap_or(DEFAULT_TOKENIZATION_TIMEOUT_MS),
             block_size: opt_parse::<u32>(get, "DYN_KV_CACHE_BLOCK_SIZE")?.unwrap_or(0),
             kv_event_port: opt_parse::<u16>(get, "DYN_EPP_KV_EVENT_PORT")?
                 .unwrap_or(DEFAULT_KV_EVENT_PORT),
@@ -241,7 +241,7 @@ mod tests {
         assert_eq!(cfg.namespace, "inference");
         assert_eq!(cfg.model_name, "Qwen/Qwen3-0.6B");
         assert_eq!(cfg.vllm_render_url, "http://vllm-render:8000");
-        assert_eq!(cfg.vllm_render_timeout_ms, DEFAULT_VLLM_RENDER_TIMEOUT_MS);
+        assert_eq!(cfg.tokenization_timeout_ms, DEFAULT_TOKENIZATION_TIMEOUT_MS);
         assert_eq!(cfg.block_size, 16);
         assert_eq!(cfg.kv_event_port, DEFAULT_KV_EVENT_PORT);
         assert!(cfg.replay_port.is_none());
@@ -378,14 +378,14 @@ mod tests {
     }
 
     #[test]
-    fn vllm_render_timeout_must_be_positive() {
+    fn tokenization_timeout_must_be_positive() {
         assert!(
             parse_cfg(&[
                 ("DYN_EPP_INFERENCE_POOL_NAME", "vllm-qwen-pool"),
                 ("POD_NAMESPACE", "inference"),
                 ("DYN_MODEL_NAME", "Qwen/Qwen3-0.6B"),
                 ("DYN_EPP_VLLM_RENDER_URL", "http://vllm-render:8000"),
-                ("DYN_EPP_VLLM_RENDER_TIMEOUT_MS", "0"),
+                ("DYN_EPP_TOKENIZATION_TIMEOUT_MS", "0"),
                 ("DYN_KV_CACHE_BLOCK_SIZE", "16"),
             ])
             .is_err()
--- a/deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
@@ -11,9 +11,11 @@
 //! [`EppStandaloneConfig::validate_config`] for field and cross-field checks.
 
 use validator::Validate;
+use validator::ValidationError;
 
 const DEFAULT_KV_EVENT_PORT: u16 = 5557;
 const DEFAULT_SELECTOR_THREADS: usize = 4;
+const DEFAULT_VLLM_RENDER_TIMEOUT_MS: u64 = 5_000;
 
 /// Environment variable that selects the EPP operating mode.
 pub const DYN_EPP_MODE: &str = "DYN_EPP_MODE";
@@ -69,6 +71,13 @@ pub struct EppStandaloneConfig {
     /// Served/catalog model identity used to group discovered workers.
     #[validate(length(min = 1, message = "DYN_MODEL_NAME is required"))]
     pub model_name: String,
+    /// Base URL of the vLLM renderer used to tokenize chat requests.
+    #[validate(length(min = 1, message = "DYN_EPP_VLLM_RENDER_URL is required"))]
+    #[validate(custom(function = "validate_vllm_render_url"))]
+    pub vllm_render_url: String,
+    /// Deadline for calls to the vLLM renderer.
+    #[validate(range(min = 1, message = "DYN_EPP_VLLM_RENDER_TIMEOUT_MS must be >= 1"))]
+    pub vllm_render_timeout_ms: u64,
     /// KV-cache block size; MUST equal the inference engine block size.
     #[validate(range(min = 1, message = "DYN_KV_CACHE_BLOCK_SIZE must be >= 1"))]
     pub block_size: u32,
@@ -108,6 +117,9 @@ impl EppStandaloneConfig {
             inference_pool_name: trimmed(get("DYN_EPP_INFERENCE_POOL_NAME")).unwrap_or_default(),
             namespace: trimmed(get("POD_NAMESPACE")).unwrap_or_default(),
             model_name: trimmed(get("DYN_MODEL_NAME")).unwrap_or_default(),
+            vllm_render_url: trimmed(get("DYN_EPP_VLLM_RENDER_URL")).unwrap_or_default(),
+            vllm_render_timeout_ms: opt_parse::<u64>(get, "DYN_EPP_VLLM_RENDER_TIMEOUT_MS")?
+                .unwrap_or(DEFAULT_VLLM_RENDER_TIMEOUT_MS),
             block_size: opt_parse::<u32>(get, "DYN_KV_CACHE_BLOCK_SIZE")?.unwrap_or(0),
             kv_event_port: opt_parse::<u16>(get, "DYN_EPP_KV_EVENT_PORT")?
                 .unwrap_or(DEFAULT_KV_EVENT_PORT),
@@ -121,6 +133,19 @@ impl EppStandaloneConfig {
     pub fn validate_config(&self) -> anyhow::Result<()> {
         self.validate()
             .map_err(|e| anyhow::anyhow!("invalid {STANDALONE_MODE} EPP config: {e}"))
+    }
+}
+
+fn validate_vllm_render_url(value: &str) -> Result<(), ValidationError> {
+    let valid_url = reqwest::Url::parse(value)
+        .map(|url| matches!(url.scheme(), "http" | "https") && url.host_str().is_some())
+        .unwrap_or(false);
+    if valid_url {
+        Ok(())
+    } else {
+        let mut error = ValidationError::new("vllm_render_url_invalid");
+        error.message = Some("DYN_EPP_VLLM_RENDER_URL must be an absolute HTTP(S) URL".into());
+        Err(error)
     }
 }
 
@@ -205,6 +230,7 @@ mod tests {
             ("DYN_EPP_INFERENCE_POOL_NAME", "vllm-qwen-pool"),
             ("POD_NAMESPACE", "inference"),
             ("DYN_MODEL_NAME", "Qwen/Qwen3-0.6B"),
+            ("DYN_EPP_VLLM_RENDER_URL", "http://vllm-render:8000"),
             ("DYN_KV_CACHE_BLOCK_SIZE", "16"),
         ])
         .expect("config should parse");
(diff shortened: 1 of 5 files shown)


deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
```diff
@@ -55,43 +55,45 @@ impl EppMode {
...
+    /// Served/catalog model identity used to group discovered workers.
     #[validate(length(min = 1, message = "DYN_MODEL_NAME is required"))]
     pub model_name: String,
+    /// Base URL of the vLLM renderer used to tokenize chat requests.
+    #[validate(length(min = 1, message = "DYN_EPP_VLLM_RENDER_URL is required"))]
+    #[validate(custom(function = "validate_vllm_render_url"))]
+    pub vllm_render_url: String,
+    /// Deadline for calls to the configured tokenization provider.
+    #[validate(range(min = 1, message = "DYN_EPP_TOKENIZATION_TIMEOUT_MS must be >= 1"))]
+    pub tokenization_timeout_ms: u64,
```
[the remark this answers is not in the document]
tmonty12: there are no existing deployments, standalone mode returns an error on startup when set

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,297 @@
...
+            "vLLM render timeout must be greater than zero"
+        );
+
+        let mut endpoint = Url::parse(base_url)
+            .with_context(|| format!("invalid vLLM renderer base URL {base_url:?}"))?;
+        anyhow::ensure!(
+            matches!(endpoint.scheme(), "http" | "https") && endpoint.host_str().is_some(),
+            "vLLM renderer base URL must be an absolute HTTP(S) URL"
+        );
+        endpoint.set_path(CHAT_RENDER_PATH);
```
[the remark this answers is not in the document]
tmonty12: [AI] Fixed in bcdfe3859a. The client now preserves the configured renderer base-path prefix when appending `/v1/chat/completions/render`, with end-to-end coverage for a prefixed URL with a trailing slash.

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
```diff
@@ -69,6 +71,13 @@ pub struct EppStandaloneConfig {
     /// Served/catalog model identity used to group discovered workers.
     #[validate(length(min = 1, message = "DYN_MODEL_NAME is required"))]
     pub model_name: String,
+    /// Base URL of the vLLM renderer used to tokenize chat requests.
+    #[validate(length(min = 1, message = "DYN_EPP_VLLM_RENDER_URL is required"))]
```
atchernych: I think we should rename this into something like DYN_EPP_TOKENIZER_URL
since we will support all engines in the future we would have to rename it anyway.
Also in the agg.yaml example we have an env var with an inference engine specific name.

deploy/inference-gateway/ext-proc/src/lib.rs
```diff
@@ -18,8 +18,10 @@ pub mod epp_standalone_config;
 pub mod picker;
 pub mod proto;
 pub mod server;
+pub mod vllm_render_client;
 
 pub use epp::Router;
 pub use epp_standalone_config::{EppMode, EppStandaloneConfig};
 pub use picker::{Endpoint, EndpointPicker, PickResult, RequestInfo};
 pub use server::ExtProcServer;
+pub use vllm_render_client::{VllmRenderClient, VllmRenderError};
```
atchernych, line 27: same here, let's remove the vllm from the `vllm_render_client`

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
```diff
@@ -124,6 +136,19 @@ impl EppStandaloneConfig {
     }
 }
 
+fn validate_vllm_render_url(value: &str) -> Result<(), ValidationError> {
```
atchernych: minor: since the new already validates then then we could keep only #[validate(length(min = 1))] on the config and let the client be the sole validator and avoid code duplication.

Alternatively, we could put the canonical check in vllm_render_client.rs and call it from the config's `[validate(custom)]`.

atchernych (changes requested):

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
tmonty12: Currently only vllm exposes this render API endpoint. The sglang tokenization url is a different API contract so it would be `DYN_EPP_SGLANG_TOKENIZATION_URL`. Need to distinguish between the APIs for request formation and response parsing.

In the long term, as/if the engine's converge to the same API we can drop the `VLLM` or `SGLANG` from the env var.

deploy/inference-gateway/ext-proc/src/lib.rs
tmonty12: same as above, vllm is the only engine that supports this render API so I think it's appropriate to keep for now

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
tmonty12: > Alternatively, we could put the canonical check in vllm_render_client.rs and call it from the config's [validate(custom)].

yep I agree with this approach

tmonty12: [AI] Fixed in a594bfc36b. URL parsing and HTTP(S)/host validation now live in one client-owned helper used by both the config validator and VllmRenderClient::new.

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
atchernych: You're right that vLLM's` /v1/chat/completions/render` and SGLang's tokenization API are different contract, but that's a client-code concern not an env-var one.

EPP has a homogeneous InferencePool with engine, one tokenizer URL set per deployment. Whoever configures would need a per engine name to look up.

You already named the timeout DYN_EPP_TOKENIZATION_TIMEOUT_MS being neutral, the URL should match I think.

tmonty12: how does EPP know what client to construct (either vllm render or sglang tokenizer) if it just receives `DYN_EPP_TOKENIZER_URL`? you would then need an additional env var to indicate the engine being used, which is why I've opted for just having `DYN_EPP_VLLM_RENDER_URL`.

the tokenization timeout is a neutral env var because it's a neutral thing - regardless of the client, we can configure a timeout for the tokenization request

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
```diff
@@ -69,6 +73,13 @@ pub struct EppStandaloneConfig {
     /// Served/catalog model identity used to group discovered workers.
     #[validate(length(min = 1, message = "DYN_MODEL_NAME is required"))]
     pub model_name: String,
+    /// Base URL of the vLLM renderer used to tokenize chat requests.
+    #[validate(length(min = 1, message = "DYN_EPP_VLLM_RENDER_URL is required"))]
+    #[validate(custom(function = "validate_vllm_render_url"))]
+    pub vllm_render_url: String,
```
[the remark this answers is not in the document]
tmonty12: [AI] Fixed in f278af2c4c. The zero replay-port fixture now supplies a valid renderer URL, so the test reaches replay-port validation.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    /// Forward an OpenAI chat-completions JSON body and return its prompt tokens.
+    ///
+    /// The body is sent unchanged so vLLM remains responsible for validating
+    /// engine-specific request fields and applying the chat template.
+    pub async fn render_chat(&self, request_body: &[u8]) -> Result<Vec<u32>, VllmRenderError> {
+        let response = self
+            .client
+            .post(self.endpoint.clone())
+            .header(CONTENT_TYPE, "application/json")
```
PeaBrane, line 133: AI review: vLLM requires Bearer auth for this endpoint when `--api-key`/`VLLM_API_KEY` is enabled, but neither the client nor standalone config can supply `Authorization`, so a protected renderer always returns 401. Could we add optional bearer-token plumbing, or explicitly document that this path supports unauthenticated internal renderers only?

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,327 @@
...
+            .body(request_body.to_vec())
```
PeaBrane: AI review (non-blocking): `render_chat(&[u8])` forces a full allocation here. The ext-proc path already retains the raw body and copies it into `RequestInfo`, so future integration can keep three full request bodies live. Consider accepting owned/shared `Bytes` (or transferring a `Vec<u8>`) before wiring this into the hot path.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,327 @@
...
+        let status = response.status();
+        if !status.is_success() {
+            return Err(VllmRenderError::UpstreamStatus {
+                status,
+                body: read_error_body(response).await,
+            });
+        }
+
+        let body = response
+            .bytes()
```
PeaBrane: AI review: error responses are bounded, but successful responses are collected without a byte limit before only `token_ids` are retained. vLLM multimodal render responses can contain serialized feature payloads, so concurrency can amplify RSS/latency, and the timeout does not bound bytes. Please consider bounded collection/streaming or an explicit text-only and response-size contract.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,327 @@
...
+        });
+        (format!("http://{address}"), task)
+    }
+
+    #[tokio::test]
+    async fn forwards_body_to_chat_render_endpoint() {
+        let (body_tx, mut body_rx) = mpsc::unbounded_channel();
+        let router = Router::new().route(
+            CHAT_RENDER_PATH,
+            post(move |body: Bytes| {
```
PeaBrane: AI review (non-blocking test hardening): this handler captures only the body. Removing or changing the production `Content-Type: application/json` header would leave this test green even though the real render endpoint may reject the request. Could the test capture/assert that header too?

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,327 @@
...
+            } if actual == timeout
+        ));
+        server.abort();
+    }
+
+    #[tokio::test]
+    async fn classifies_unavailable_renderer() {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let address = listener.local_addr().unwrap();
+        drop(listener);
```
PeaBrane: AI review (non-blocking test hardening): after dropping this listener, the address is unreserved until the client connects, so a parallel test/process can claim it and change the observed error. A deterministic server that accepts and immediately closes/resets the connection would avoid the port-reuse race.

PeaBrane (approved):
✅ AI-reviewed and approving this client/config-only slice. I left two compatibility/resource concerns and three non-blocking hardening notes inline.

Human review is still needed to decide whether authenticated renderer support and successful-response bounds belong in this PR or should be explicitly deferred. Please do not treat this automated approval as a substitute for that judgment.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
tmonty12: Addressed in e219cdb093: render_chat now takes Bytes by value, passes it directly to reqwest, and the forwarding test verifies the original body.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
tmonty12: Addressed in e219cdb093: added a 16 MiB default, zero validation, declared-length and streamed-body enforcement, plus overflow and exact-limit tests.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
tmonty12: Addressed in e219cdb093: the forwarding test now captures and asserts Content-Type: application/json.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
tmonty12: Addressed in e219cdb093: the test holds the listener, accepts one connection, and closes it before asserting Unavailable.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
tmonty12: let's defer authenticated render API support for a subsequent PR. initial scope is running the vllm render service as a sidecar or intra cluster Service

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,509 @@
...
+            .render_chat(Bytes::from_static(b"{}"))
+            .await
+            .unwrap_err();
+
+        assert!(matches!(
+            error,
+            VllmRenderError::ResponseTooLarge {
+                limit: actual_limit,
+                received,
+            } if actual_limit == limit && received == (FIRST_CHUNK.len() + SECOND_CHUNK.len()) as u64
```
dmitry-tokarev-nv: Non-blocking test-hardening nit: this pins `received` to the exact total (`20`). That only holds while hyper surfaces the server's `[11, 9]` chunk framing unchanged — if a chunk boundary ever lands inside `(11, 20)` (coalescing/re-chunking in a future hyper/reqwest), the guard still fires correctly but `received` differs and this assertion breaks. Asserting the invariant instead is equally strong and framing-independent:

```suggestion
            } if actual_limit == limit
                && received > limit as u64
                && received <= (FIRST_CHUNK.len() + SECOND_CHUNK.len()) as u64
```

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
dmitry-tokarev-nv: Weighing in since this thread is the last thing gating the PR.

I checked the code: there's currently **no engine/backend discriminator anywhere in `EppStandaloneConfig` or `main.rs`**. So for this vLLM-only milestone @tmonty12's point holds mechanically — a neutral `DYN_EPP_TOKENIZER_URL` can't by itself tell EPP whether to build a vLLM-render client (`/v1/chat/completions/render`) or an SGLang tokenizer client (different request/response contracts) without introducing a *second* discriminator env var. The engine-specific name is doing real work today as the implicit discriminator.

That said, @atchernych's consistency concern is fair and I don't think it needs to block: the timeout is genuinely engine-agnostic (a deadline), whereas the URL is bound to an engine-specific API path + response schema — so the neutral-timeout / specific-URL asymmetry is defensible rather than accidental.

A middle path that should satisfy both:
1. Keep `DYN_EPP_VLLM_RENDER_URL` for this slice.
2. Add a short doc note on the field spelling out the intended multi-engine model — one URL var per engine (`DYN_EPP_VLLM_RENDER_URL`, later `DYN_EPP_SGLANG_TOKENIZATION_URL`), engine inferred from which var is set, and a hard error if more than one is set.
3. Revisit collapsing to a neutral `DYN_EPP_TOKENIZER_URL` + explicit `DYN_EPP_TOKENIZER_KIND` when the second engine actually lands and forces the discriminator question.

That keeps the present config unambiguous while making the future direction explicit. Worth aligning the `agg.yaml` example env var name with whatever's chosen here too.

deploy/inference-gateway/ext-proc/src/vllm_render_client.rs
```diff
@@ -0,0 +1,509 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! HTTP client for vLLM's chat render endpoint.
+//!
+//! The client is intentionally limited to protocol handling. Callers decide how
+//! a render failure affects request routing.
```
dmitry-tokarev-nv, line 7: Following up on @PeaBrane's auth thread — since bearer-token plumbing is deferred, a one-line doc caveat here closes the "document unauthenticated-only" half of that ask and sets the scope explicitly:

```suggestion
//! a render failure affects request routing.
//!
//! The client sends no `Authorization` header, so it targets unauthenticated
//! in-cluster renderers (a sidecar or an internal Service). Authenticated
//! renderers (vLLM `--api-key` / `VLLM_API_KEY`) are a follow-up.
```

dmitry-tokarev-nv (approved):
lgtm

deploy/inference-gateway/ext-proc/src/epp_standalone_config.rs
atchernych: Long term the tokenizer can be an external service (plugin) not an instance of the engine worker. So long term we would need something like `DYN_EPP_TOKENIZER_SERVICE_URL` and discriminate in the service.  We can introduce an additional var for the protocol descriminator `DYN_EPP_TOKENIZER_PROTOCOL = vllm-render or sglang-tokenize`

atchernych (approved):

PR #11632, opened 2026-07-13
build(container): add a media-codec allowlist gate to the compliance stage (OPS-7665)

dmitry-tokarev-nv:
## Summary

Adds a compliance scan that **fails the image build** if a media-codec library or binary ships outside our in-tree ffmpeg allowlist.

We build an in-tree ffmpeg to a narrow media-codec set. This gate keeps the finished image aligned with that by catching media libraries introduced elsewhere — a distro `libav*`, a wheel-bundled `libavcodec`, `libx264`/`libx265`, or a stray/imageio-bundled `ffmpeg` — that a base-image or dependency bump could reintroduce.

- **Filesystem scan** (statically-bundled codec `.so` files don't show up as SBOM components) plus an optional SBOM component/version floor.
- **Config-driven**: `container/compliance/policy/codec_policy.yaml`. Matches under our in-tree `allow_paths` pass; reasoned per-entry `exceptions` (e.g. NVIDIA DALI's bundled `libav*`, owned by its own team) are logged but do not fail.
- Runs in the `licenses` stage of `compliance.Dockerfile`, so a violation fails the image build (same hook as the existing license policy gate).

Pairs with #11628, which builds the in-tree ffmpeg to that allowlist.

## Validation

- 10 unit tests (`container/compliance/tests/test_scan_codecs.py`) covering allow / exception / violation classification against the real bundled-library paths, the SBOM version floor, and exit codes — all pass.
- `container/render.py` renders the trtllm/sglang/vllm runtime Dockerfiles with the gate wired into the licenses stage.

Linear: OPS-7665

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11632" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a media-codec compliance gate for container builds.
  * Detects unapproved codec binaries and libraries outside the approved allowlist.
  * Checks SBOM entries for outdated codec versions.
  * Provides categorized findings, reports, and build-summary output.
  * Supports report-only scanning or failing builds when violations are found.
* **Bug Fixes**
  * Explicitly recognizes approved in-tree codecs and documented exceptions to reduce false positives.

PR #11836, opened 2026-07-17
build(container): media-codec cleanup + build gate + NVDEC video decode (OPS-7665)

dmitry-tokarev-nv:
## Summary

Consolidates the media-codec compliance work for all three runtime images and adds a
build gate that enforces it, then restores H.264/H.265 video input on the GPU so the
cleanup is not a capability regression for those codecs.

The removals and the gate have to land together: the gate only passes once every image
is clean, which made the original split PRs interdependent and impossible to green
individually.

### Scope, and how it grew

This began as a compliance cleanup: remove the software H.264/H.265/AAC codecs from the
three runtime images and gate the result so they cannot come back. It expanded twice,
both times because removal on its own would have shipped either a regression or an
incomplete result. Both expansions concern the GPU's dedicated video engines, and they
went in opposite directions.

**NVDEC — hardware decode, added.** Removing the software decoders would have dropped
H.264/H.265 video input outright. Those codecs now route to the GPU's NVDEC engine
through `PyNvVideoCodec` across all three backends, so the cleanup is not a capability
regression. This is the larger half of the diff and brought its own requirements: the
container must be granted the `video` driver capability
([ai-dynamo/velonix#550](https://github.com/ai-dynamo/velonix/pull/550) does this for the
Dynamo GPU testers), and three product defects surfaced only once the tests genuinely
executed against it.

**NVENC — hardware encode, removed.** The in-tree ffmpeg previously built `h264_nvenc`
and linked `nv-codec-headers`. A hardware H.264 encoder is still a distributable H.264
codec surface, so it is now omitted entirely: `--enable-nvenc` and the encoder are gone,
the headers are no longer cloned, and video generation encodes VP9 (libvpx) instead. A
post-build guard fails the build if any H.264/H.265/AAC/NVENC surface reappears.

**DALI — upgraded past its own cleanup.** The TensorRT-LLM base image ships a DALI whose
vendored ffmpeg carries the same decoders this PR removes everywhere else, and they
cannot be deleted independently of the package. Detailed below.

Together these mean the PR is no longer only a removal: it changes which engine serves
H.264/H.265 (GPU rather than CPU), and it drops one GPU path while adding another.

### Per image

- **In-tree ffmpeg** (`wheel_builder`): narrow decoder/demuxer/parser allowlist
  (VP8/VP9 + rawvideo); encoders reduced to `libvpx_vp9`. Bumped 8.1 → 8.1.2.
- **vLLM**: removes the codec-bearing video-decode wheels inherited from the base
  (`av`, `opencv`, `decord`, `torchcodec`). Keeps the in-tree ffmpeg + `imageio-ffmpeg`
  for the video-generation encode path, which now encodes VP9.
- **SGLang**: builds without the in-tree ffmpeg stack and purges the same wheels, with
  a build-failing guard.
- **TensorRT-LLM**: removes preinstalled `opencv-python-headless`.
- **Compliance gate** (`licenses` stage): filesystem + SBOM scan that fails the build if
  a media-codec library ships outside the allowlist or a reasoned exception.

### Hardware video decode (NVDEC)

Removing the software decoders would have dropped H.264/H.265 video input entirely, so
this PR routes those codecs to the GPU's dedicated NVDEC engine via `PyNvVideoCodec`,
across all three backends:

- Shared core `common/multimodal/nvdec_decoder.py`: codec probe → route → decode to
  `(T, H, W, 3)` uint8 RGB, matching the existing `VideoLoader` contract.
- Pinned `PyNvVideoCodec>=2.2.0`. 2.0.x bundles a full FFmpeg including `libavcodec`,
  which the gate rejects; 2.2.0 bundles only `libavutil` + `libavformat` (container
  demux, no codec) and links `libnvcuvid` at runtime.
- Sources: `http(s)`, `file://` and `data:` all hardware-decode. `file://` stays behind
  the existing `DYN_MM_LOCAL_PATH` policy.
- Requires a GPU with a decode engine and a container granted the `video` driver
  capability. NVDEC is present across the datacenter line (A100 5 engines, Hopper and
  Blackwell 7) — it is NVENC, the *encoder*, that those parts omit, which is also why
  the generation path encodes VP9 on CPU. Documented in
  `features/multimodal/video-decode-gpu-requirements.md`.

## Known limitation

**VP8/VP9/AV1 video input has no decoder in these images.** NVDEC covers H.264/H.265;
everything else relied on the removed software wheels. #12051 restores them behind an
opt-in installer. This is deliberate and documented on all three backend pages.

## Tests added

Video input previously had no serve-level cover outside vLLM. This adds it on the
other two backends, so all three now exercise video->text end to end:

| Test | Backend | What it covers |
|---|---|---|
| `mm_agg_video_nvdec` (h264, h265) | vLLM | NVDEC serve path |
| `video_e_pd_qwen_nvdec` (h264, h265) | SGLang | **new** — the encode-worker NVDEC path, with the embedding cache **disabled** so it runs the default configuration |
| `aggregated_multimodal_video_nvdec` (h264, h265) | TensorRT-LLM | **new** — this backend had *no* video serve test at all |
| `test_nvdec_decoder_gpu` (h264, hevc) | shared | real NVDEC decode against committed fixtures |
| `test_media_source` | shared | `file://` sandbox: prefix, traversal, symlink escape, data URIs |
| `test_video_encode_real` | all three | output really is VP9, via the ffmpeg in each image |

None of the new NVDEC cases install a decoder, so they exercise what a deployment
actually receives. The four pre-existing tests that *do* pip-install one are now marked
`installs_extra_dependencies`, so a green run cannot be mistaken for evidence that
VP8/VP9 works in an unmodified image (`-m "not installs_extra_dependencies"` selects
only unmodified-image behaviour).

## Validation

**CI, expanded suite** — [run 30513728452](https://github.com/ai-dynamo/dynamo/actions/runs/30513728452)
on `116728e016` ran the backend jobs with `(pre_merge or post_merge)` markers rather than
`pre_merge` alone, so the post_merge cases were exercised before merge: green across
vLLM, SGLang and TensorRT-LLM (Test, Multi-GPU Test and Compliance audit, amd64 + arm64).

**CI, merge-ready config** — [run 30562129491](https://github.com/ai-dynamo/dynamo/actions/runs/30562129491)
on [`6efdd3b2a6`](https://github.com/ai-dynamo/dynamo/commit/6efdd3b2a6550c6f2688cd0bbf660037e3f02e82)
green after the temporary markers and raised timeouts were reverted, confirming the
configuration that actually merges is clean at the original budgets.

**CI, latest** — [run 30598827010](https://github.com/ai-dynamo/dynamo/actions/runs/30598827010)
green on the final pre-rebase commit, again at the merge-ready budgets: **61 jobs
succeeded, 17 skipped, 0 failed**, covering every backend Test and Multi-GPU Test lane on
amd64 and arm64 plus all four Compliance audits (vLLM, SGLang, TensorRT-LLM, frontend).

The branch has since been rebased onto current `main`, which had moved the multimodal
documentation into `pages/use-cases/multimodal-serving/` and added the `request-trace-s3`
build feature. Both conflicts were merged rather than resolved in favour of one side, so
`request-trace-s3` is now built into every wheel while SGLang still builds without
`media-ffmpeg`.

**CI, rebased tree with post_merge markers** —
[run 30601314656](https://github.com/ai-dynamo/dynamo/actions/runs/30601314656) is
**green: 61 jobs succeeded, 17 skipped, 0 failed.** This is the widest run on this
branch — the backend lanes ran `(pre_merge or post_merge)` rather than `pre_merge` alone,
so the post_merge video cases were exercised before merge — with vLLM green at a raised
150m budget, TensorRT-LLM at 90m, and every Compliance audit passing.

The SGLang single-GPU lane needed re-runs to get there, for a reason unrelated to this
PR. On the first two attempts its parallel stage was killed at the job budget because one
router test, `test_router_e2e_with_sglang.py::test_sglang_indexers_sync[nats-tcp]`, wedged
for ~4245s against its own declared `timeout(320)` — while reporting **602 tests passed
and zero test-level failures**. That timeout cannot fire: the test is documented in-tree
(DYN-2784) as blocking forever in `KvRouter` on `min_initial_workers=2` when a worker dies
during launch, with the `pytest.mark.timeout` signal swallowed at the C-level syscall.
`GPU0: 5.4/22 GiB` against a 12.0 GiB profile shows only one of the two workers came up.

That flake reproduces on `main` independently of this branch —
[run 30606035706](https://github.com/ai-dynamo/dynamo/actions/runs/30606035706) wedged on
the same test, same worker, same `5.4/22 GiB`, for 6077s — so it is a pre-existing defect
being handled separately. The third attempt passed the lane cleanly.

The temporary markers and raised budgets have since been reverted, so the configuration
that merges is the stock one.

**CI now exercises hardware decode directly.** Until now the NVDEC cases could not run in
CI at all: the runners were not granted the `video` driver capability that `libnvcuvid`
needs, so every one of them skipped and the paths rested on a dev box.
[ai-dynamo/velonix#550](https://github.com/ai-dynamo/velonix/pull/550) added
`NVIDIA_DRIVER_CAPABILITIES=video,compute,utility` to the Dynamo GPU testers, and they now
execute on the runners' L4s — which have four NVDEC engines. In
[run 30685708717](https://github.com/ai-dynamo/dynamo/actions/runs/30685708717), with
**zero** capability-based skips anywhere:

| Backend | Test | Result |
|---|---|---|
| vLLM | `mm_agg_video_nvdec_qwen3-vl-2b_h264` serve e2e | passed, 138s |
| vLLM | `mm_agg_video_nvdec_qwen3-vl-2b_h265` serve e2e | passed, 104s |
| TensorRT-LLM | `aggregated_multimodal_video_nvdec` serve e2e | passed, 106s |
| SGLang | `video_e_pd_qwen_nvdec` serve e2e (embedding cache **disabled**) | passed |
| shared | `test_nvdec_decodes_real_clip[h264]` / `[hevc]` | passed — real `libnvcuvid` decode |

The NVDEC unit suites pass on all three backends in the same run. This closes the one
coverage gap this PR previously carried.

**GPU hardware** (RTX 6000 Ada dev box) — how the NVDEC paths were validated before the
runners could do it, and still where each fix is reproduced before pushing. Run against
the **built artifacts with no patching**, and every run asserts something actually
executed rather than trusting a green exit:

| Check | Result |
|---|---|
| vLLM NVDEC decode + H.264/H.265 serve e2e | `executed=5 skipped=0` |
| SGLang `video_e_pd_qwen_nvdec` serve e2e | passed, `executed=1 skipped=0`, no decoder fallback |
| TensorRT-LLM `aggregated_multimodal_video_nvdec` serve e2e | passed, `executed=1 skipped=0` |
| SGLang shipped `video_metadata` shim vs real `MMEncoder._encode` (sglang 0.5.16) | before fail -> after pass, embeddings `(648, 2048)` |
| TensorRT-LLM `VideoData` vs vendor `async_load_video` (1.3.0rc22) | structurally identical |
| `file://` and `data:` decode + sandbox refusals | decode ok; local-access-off, traversal and symlink escape all refused |
| Fixes present in the artifact, unpatched | verified on all three flavours |

Getting those tests to actually execute surfaced four defects CI could not have caught,
all fixed here: a `filterwarnings=error` interaction that cached the NVDEC capability
probe as unavailable for an entire session; a GPU test that built its fixture with an
encoder these images intentionally omit, so it could never run in the image it
validates; a source frame rate reported as `0`, which made every H.264/H.265 vLLM video
request fail downstream; and `file://`/`data:` video reaching no decoder at all.

**Review findings on the SGLang encode worker** (thanks @Chokoyo) — three, each confirmed
on GPU hardware before any code changed, and each with a regression test:

- *NVDEC never ran in the default configuration* (`9f0650b1ad`). The conversion was
  reachable only from `_encode_with_cache`, but the embedding cache defaults to disabled,
  so a stock deployment passed raw URLs to SGLang and every H.264/H.265 request failed.
  The serve test had masked it by setting a non-zero cache capacity; it now runs with the
  cache off.
- *SGLang's frame-selection policy was bypassed* (`9f0650b1ad`, completed by
  `10b56197e2`). Pre-sampled frames took `preprocess_video`'s `return vr, None` branch, so
  the model's `vision_config.video` policy never applied and metadata was synthesized. A
  `VideoDecoderWrapper` subclass over PyNvVideoCodec now lets SGLang choose frames from
  true source values: measured `nframes=4 -> 4 frames`, `fps=10.0` rather than a
  fabricated `24.0`, indices `[0, 3, 6, 9]`, with an ndarray control reproducing the old
  behaviour.
- *A URL-policy refusal degraded to an unvalidated fetch* (`c60d651be7`).
  `UrlValidationError` was caught alongside decode failures, so a rejected URL was handed
  to SGLang — which applies no policy of its own. Reproduced first: a loopback URL the
  policy refused was fetched anyway (38128 bytes from a blocked address), and a refused
  `file://` path resolved to a readable local file. After the fix the same probe records
  zero hits on that host. Refusals now propagate; genuine decode failures still fall back.

One related gap is deliberately **not** closed here: video whose codec is not
hardware-routed still reaches SGLang unvalidated, with no exception involved. That
predates this PR — `main`'s SGLang encode worker performs no URL validation at all — and
closing it would start rejecting input `main` accepts today, so it belongs in its own
change rather than folded into a compliance PR.

**Video content assertions rebuilt** (`bb5e1b230d`). The serve tests asserted words against
`ffmpeg testsrc2` colour bars — first `["red", "static", "still"]`, describing content those
clips never contained, then a wider list chosen from observed model output. Both were weak:
the first passed only when the model happened to say "still", and the second confirms the
model described something colourful, not that it saw the right video.

The fixtures are now a yellow triangle on a near-black background, generated once as raw
frames and encoded three ways (`lib/llm/tests/data/media/make_triangle_fixture.py`), so VP9,
H.264 and H.265 carry identical content and every codec path asserts the same word. Measured
round-trip against the source: **1.59 / 1.68 / 1.63** mean absolute per-pixel difference,
within 0.1 of each other — the previous fixtures differed by **17.77** between codecs, which
made cross-codec comparison impossible. Confirmed on hardware, where H.264 and H.265 each
returned the same sentence: "The video begins with a black screen. A yellow triangle appears
in the center of the screen." Forcing an impossible expected value fails the test, so a pass
is not vacuous.

One limit is recorded in the fixture comment rather than left implied: the triangle moves,
but that is **not** coverage of temporal sampling. Asked to describe the clip, Qwen2.5-VL-3B
reports it as "static and does not move" — it names the object reliably and the movement not
at all, so a regression collapsing frame selection to a single frame would not surface in the
response text. Cross-codec frame comparison is the check to build on for that, and the
identical-content fixtures now make it possible.

Every consumer routes through the shared `MULTIMODAL_VIDEO_EXPECTED` constant, so the
expectation cannot drift from the footage again — the vLLM and XPU suites had been asserting
against a clip they no longer used.

**VP9 rejection is now asserted** (`3d0a860927`). These images decode video through cv2,
which they no longer ship, so NVDEC covers H.264/H.265 and VP8/VP9/AV1 have no decode path at
all. Nothing exercised a codec the image cannot decode, so a base-image bump that
reintroduced cv2 — or a routing change that sent VP9 down the hardware path — would both have
gone unnoticed. Asserting the failure rather than leaving the gap untested is the approach
DALI took when it trimmed its own FFmpeg build (NVIDIA/DALI#6352). Verified against a real
image (1.3.0rc22) rather than assumed: VP9 probes as `vp9`, does not route to NVDEC, and
returns HTTP 400 carrying the upstream reason. The assertions are scoped to what this repo
owns — the status, the `Failed to load video (<url>)` prefix, and that the cause is preserved
rather than swallowed — so an upstream rewording cannot fail the test for no defect.

**CI status.** The most recent full run,
[30710098691](https://github.com/ai-dynamo/dynamo/actions/runs/30710098691) on `a887cfb186`,
predates the two commits above and so ran the earlier fixtures. Its video lanes passed —
`video_e_pd_qwen_nvdec`, `video_e_pd_qwen` and `multimodal_e_pd_qwen` — and the job failed on
the unrelated SGLang flake described earlier: `test_sglang_indexers_sync` wedged for 4235s
against its own declared `timeout=320s`, blocking three queued workers until the runner
reported `Executing the custom container implementation failed`. The triangle fixtures and the
VP9 test are validated on GPU hardware against the shipped images, but have not yet been
through CI.

**DALI upgraded past its own media-codec cleanup.** The TensorRT-LLM base image
ships DALI 2.1.0, whose vendored ffmpeg registers 446 decoders including h264,
hevc, aac, aac_fixed and aac_latm. Those cannot be removed on their own:
libavcodec, libavfilter, libavformat, libavutil and libswscale are each
DT_NEEDED by libdali.so, libdali_kernels.so, libdali_operators.so and every
python_function_plugin, so deleting the codec libraries breaks `import
nvidia.dali` outright. Upstream restricted that build
([DALI_deps#162](https://github.com/NVIDIA/DALI_deps/pull/162),
[DALI#6352](https://github.com/NVIDIA/DALI/pull/6352)) and released it in 2.1.1,
which registers 440 with all five absent and vp8/vp9/mjpeg/av1 retained.

The image now pins `nvidia-dali-cuda130>=2.1.1`. Removing the package outright
would also have worked — nothing declares a dependency on it, nothing imports it
across 43,455 files scanned, and the suite passes without it — but it belongs to
TensorRT-LLM, and a version bump clears the codecs without taking a package out
of someone else's image.

A whiteout entry comes with it, and that is the part that makes the upgrade real.
`pre_runtime` rebases on upstream and overlays `runtime_full`, and DALI's
libraries are hash-named, so an upgrade renames rather than overwrites. Built
both ways to check: **without** the whiteout the image keeps two libavcodec
copies — the upgraded one and the 2.1.0 one still registering h264/hevc/aac —
and reports version 2.1.0 while carrying both; **with** it, one copy and none of
them. Two build guards enumerate every bundled libavcodec, one after the upgrade
and one after the overlay COPY (the only stage where the fault can exist), and
`test_no_software_video_codecs.py` repeats the check on the shipped image. That
test was verified to fail on 2.1.0 before being trusted to pass afterwards.

The DALI waiver in `codec_policy.yaml` stays, with a corrected rationale: the deny
globs match on presence, and DALI still vendors a libavcodec on disk even with the
decoders gone. An earlier revision claimed the waiver could be narrowed to the
non-codec companions, which the link graph rules out.

**GPU validation** (RTX 6000 Ada) — unit suite green, and the TensorRT-LLM
integration lane green: 16 passed including `aggregated_multimodal_video_nvdec`,
`video_diffusion`, `image_diffusion` and the E/PD multimodal cases.

**One test is skipped, deliberately.**
`test_router_e2e_with_sglang.py::test_sglang_indexers_sync` hangs rather than
fails — a worker dies during launch, the router blocks forever waiting for two,
and the declared timeout cannot fire because the signal is swallowed at a C-level
syscall. The note previously scoped the skip to nightly on the basis that it
passed elsewhere; measured over five days that is 82% of nightly but also 63% of
post_merge, with observed hangs of 3659s, 4235s and 6077s, each holding a 12 GiB
reservation with three workers queued behind it. The comment records what would
let it come back: fixing the launch race, or giving the parallel orchestrator a
hard-kill for a child that outlives its timeout.

**CI, with the DALI work** —
[run 30778874769](https://github.com/ai-dynamo/dynamo/actions/runs/30778874769)
on `9c154b62cd` is **green: 61 jobs succeeded, 17 skipped, 0 failed**, with the
temporary post_merge widening in place so the video cases were exercised. This is
the first run to put the DALI change through a real build: both guards executed in
situ, the whiteout survived the full-root COPY, and license validation passed with
the three new SBOM components the upgrade pulls in — none of which any GPU run can
reach.

One job needed a re-run.
`test_mocker_disagg_startup_lifecycle[frontend-decode-prefill]` on arm64 asserted
the prefill worker id and received the decode worker's, a startup race in a router
mocker test this PR does not touch, which had passed on the three previous runs of
this branch. It passed on re-run.

The XPU lane is red for an unrelated pre-existing reason: the XPU image ships
without `gpu_memory_service`, so six `test_vllm_worker_factory` cases fail on
import. They fail identically on the untouched image. That lane runs only when the
`xpu` label is present, which was applied automatically because this PR touches two
`*xpu*` test files, so this is the first time it has executed here.

## Release note

**Behaviour change — video output codec.** `output_format: "mp4"` now returns a **VP9**
stream in an MP4 container, where it previously returned H.264. The file extension is
unchanged, so a client that hands the bytes straight to a player can fail with no error
from Dynamo. VP9-in-MP4 plays in Chrome, Firefox and ffmpeg; it does not play in Safari
or QuickTime.

Why it cannot stay H.264: NVIDIA omits the hardware encoder (NVENC) from the datacenter
line — A100, H100, H200, B200 and GB200 have none, against 5–7 NVDEC decode engines — so
a hardware H.264 path would work only on L4/L40S/RTX-class parts and fail on the GPUs
Dynamo is deployed on. The alternative is a software H.264 encoder, which is the codec
surface these images deliberately no longer carry. VP9 is the only option that behaves
identically on every supported GPU.

Consumers that need H.264 can transcode on their side:
`ffmpeg -i output.mp4 -c:v libx264 output-h264.mp4`.

Video *input* is unaffected: H.264 and H.265 are decoded on the GPU via NVDEC, which is
present across the datacenter lineup.

## Fixed during review

Reviewers found four defects that the tests here did not, and each is fixed with a
regression test. Worth listing because three of them share a shape: a control path that
looked healthy while doing nothing.

**The gate itself could pass every image** (`a1919637e2`, @dagil-nvidia). `CodecPolicy.load`
shape-checked `deny_globs` and `exceptions` but not `allow_paths` — the one key that fails
open. Written as a scalar it iterates per character, `"/"` is among them, and every
absolute path starts with `"/"`, so a single missing `- ` classified every violation as
`allowed` while still printing a full report. Reproduced on GPU:
`/opt/evil/libx264.so.164` → `('allowed', None)` as a scalar, `('violation', None)` as a
list. Three tests, verified with the check stripped back out.

**vLLM's frame-sampling contract was dropped for short clips** (`e1d8c454be`, @Chokoyo).
The NVDEC metadata omitted `do_sample_frames`; Qwen3-VL reads a missing flag as `False`,
so clips where every source frame is returned looked pre-sampled and the model's own fps
policy never ran. Their reproduction against vLLM 0.26.0 showed `video_grid_thw`
`[[5,16,20]]` instead of `[[2,16,20]]`. Asserted in a unit test because, as they noted, the
serve tests cannot see it — the answer does not change when the frame count does.

**SGLang downloaded every non-NVDEC video twice** (`1da12f978f`, Codex via @rmccorm4). The
encode worker fetched the payload to probe its codec, discarded it, and handed SGLang the
URL to fetch again. That doubles ingress, breaks one-use signed URLs, and — since SGLang
applies no URL policy of its own — meant the bytes it decoded were never the validated
ones. It now passes the fetched bytes; `load_video` takes `Union[str, bytes, VideoData]`,
so no shim is needed. Measured against a counting HTTP server: two origin fetches per
request before, one after.

**`DYN_DISABLE_NVDEC` meant opposite things per backend** (`0f4e71274d`, @dagil-nvidia).
One call site tested raw truthiness where the rest use `env_bool`, so `=0` and `=false`
disabled NVDEC on SGLang while leaving it on elsewhere.

Two further problems came from outside the PR review, both in the container/docs surface
where the test suite does not reach:

- **SGLang XPU was purged with no replacement** (`a4ba547211`). The purge sat outside the
  device conditional while the replacement ffmpeg is copied in under `device == "cuda"`,
  leaving that image with no ffmpeg at all. Gating the purge alone would have broken the
  build — the end-of-stage codec guard was not device-scoped either and only passed there
  *because* the purge had run; both are now gated together.
- **The SGLang support matrix contradicted its own page** (`e8ed9f2ba2`), listing video as
  supported in aggregated deployments where the paragraph nine lines below says it is not,
  and stating the disaggregated case unconditionally though `_NVDEC_UNSAFE_MODEL_TYPES`
  excludes the Qwen3-VL family.

**Coverage added along the way.** Video was only ever addressed over http; the local branch
reads through `read_local_media_bytes` behind a `DYN_MM_LOCAL_PATH` gate before joining the
same NVDEC routing, and had no serve coverage (`ea3743517b`). The note above those fixtures
also claimed `file://` is not hardware-decoded, which is wrong for vLLM and SGLang and
argued against writing the missing test; TensorRT-LLM is the real exception and is now named
as such.

**Supersedes and combines #11607, #11628 and #11632**, which stay open for their review
threads. #11872 (actionable decode errors) is stacked on this and rebases once it lands.

Linear: OPS-7665

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11836" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added automated media-codec compliance checks for container images, including SBOM-based version validation.
  * Added safeguards to prevent unsupported codec libraries and OpenCV from being included in selected runtime images.

* **Changes**
  * Updated bundled FFmpeg to version 8.1.2.
  * Reduced SGLang media support and removed its bundled FFmpeg stack.
  * Restricted supported video decoding to VP8/VP9 scenarios; H.264/H.265 software decoding is unavailable.

* **Documentation**
  * Clarified supported video codecs and known decoding limitations.


container/compliance/native_packages.yaml
```diff
@@ -78,17 +78,14 @@ packages:
...
   - name: ffmpeg
-    version: "8.1"
+    version: "8.1.2"
     license: LGPL-2.1-or-later
     source: https://ffmpeg.org/
     images:
       - vllm-runtime
       - vllm-runtime-efa
-      - sglang-runtime
-      - sglang-runtime-efa
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Fixed in 474b327a40. The TensorRT-LLM runtime does copy the in-tree ffmpeg + libvpx from the build stage, so I added `trtllm-runtime` and `trtllm-runtime-efa` to the `images:` lists for both the `ffmpeg` and `libvpx` entries — their generated notices now attribute both.

container/compliance/policy/codec_policy.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - "**/libavformat-*.so*"
+  - "**/libavfilter.so*"
+  - "**/libavfilter-*.so*"
+  - "**/libavutil.so*"
+  - "**/libavutil-*.so*"
+  - "**/libswscale.so*"
+  - "**/libswscale-*.so*"
+  - "**/libswresample.so*"
+  - "**/libswresample-*.so*"
+  - "**/libpostproc.so*"
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Fixed in 474b327a40. Added the hash-renamed variants `libx264-*.so*`, `libx265-*.so*`, `libfdk-aac-*.so*`, and `libpostproc-*.so*` (the `libav*` hashed variants were already present). Added a test asserting a vendored `libx264-<hash>.so` under a wheel `.libs/` is now denied.

container/compliance/scan_codecs.py
```diff
@@ -0,0 +1,260 @@
...
+class CodecPolicy:
+    def __init__(self, doc: dict):
+        self.deny_globs: list[str] = doc.get("deny_globs", []) or []
+        self.allow_paths: list[str] = doc.get("allow_paths", []) or []
+        self.deny_components: list[dict] = doc.get("deny_components", []) or []
+        self.exceptions: list[dict] = doc.get("exceptions", []) or []
+
+    @classmethod
+    def load(cls, path: Path) -> "CodecPolicy":
+        return cls(yaml.safe_load(path.read_text(encoding="utf-8")) or {})
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Fixed in 474b327a40. `CodecPolicy.load` now fails closed: it rejects a non-mapping document, a missing/empty `deny_globs`, and any exception missing `glob`/`reason`/`owner`, raising instead of defaulting to empty. Added tests for each case.

container/templates/vllm_runtime.Dockerfile
```diff
@@ -275,6 +275,30 @@ RUN --mount=type=bind,source=./container/deps/requirements.vllm.txt,target=/tmp/
...
+    rm -rf \
+        "${SITE_PACKAGES}"/av "${SITE_PACKAGES}"/av-*.dist-info "${SITE_PACKAGES}"/av.libs \
+        "${SITE_PACKAGES}"/cv2 "${SITE_PACKAGES}"/opencv_python*.dist-info "${SITE_PACKAGES}"/opencv_python*.libs \
+        "${SITE_PACKAGES}"/decord "${SITE_PACKAGES}"/decord-*.dist-info "${SITE_PACKAGES}"/decord.libs \
+        "${SITE_PACKAGES}"/decord2 "${SITE_PACKAGES}"/decord2-*.dist-info "${SITE_PACKAGES}"/decord2.libs \
+        "${SITE_PACKAGES}"/torchcodec "${SITE_PACKAGES}"/torchcodec-*.dist-info \
+        "${SITE_PACKAGES}"/PyNvVideoCodec "${SITE_PACKAGES}"/PyNvVideoCodec-*.dist-info "${SITE_PACKAGES}"/PyNvVideoCodec.libs \
+        /root/.cache/pip; \
+    ! python3 -c "import cv2" 2>/dev/null; \
+    ! python3 -c "import av" 2>/dev/null
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Fixed in 474b327a40. The post-purge guard now also checks `decord`, `torchcodec`, and `PyNvVideoCodec` in addition to `cv2` and `av` (decord2 imports as `decord`), so a package surviving the best-effort uninstall fails the build.

container/compliance/tests/test_scan_codecs.py
```diff
@@ -0,0 +1,139 @@
...
+    pytest.mark.post_merge,
+    pytest.mark.gpu_0,
+    pytest.mark.unit,
+]
+
+# The real shipped policy — the tests assert against it, not a fixture, so a
+# policy edit that would let a non-allowlisted media codec through fails a test here.
+_POLICY = CodecPolicy.load(
+    Path(__file__).resolve().parents[1] / "policy" / "codec_policy.yaml"
+)
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: The tests assert against the real shipped policy and do not mutate it, so there is no cross-test leakage today. The new fail-closed tests added in 474b327a40 load fresh policies via a local helper. Leaving the module-level instance as-is for now; happy to switch to a fixture if preferred.

container/templates/wheel_builder.Dockerfile
```diff
@@ -342,15 +363,37 @@ RUN --mount=type=secret,id=aws-web-identity-token,target=/run/secrets/aws-token
...
+        --enable-decoder=vp8,vp9,rawvideo \
         --disable-muxers \
         --enable-muxer=mov,mp4,matroska,webm \
-        --enable-protocol=file,pipe && \
+        --disable-demuxers \
+        --enable-demuxer=mov,matroska,rawvideo \
+        --disable-parsers \
+        --enable-parser=vp8,vp9 \
+        --disable-protocols \
+        --enable-protocol=file,pipe,fd && \
```
JanuszL, line 393: Perhaps you can log the output from the configuration step to verify how the build is being configured.

container/templates/sglang_runtime.Dockerfile
```diff
@@ -161,17 +142,97 @@ RUN --mount=type=bind,source=./container/deps/requirements.common.txt,target=/tm
...
+        /usr/local/include/libsw* \
+        /usr/local/lib/libav* \
+        /usr/local/lib/libpostproc* \
+        /usr/local/lib/libsw* \
+        /usr/local/lib/pkgconfig/libav*.pc \
+        /usr/local/lib/pkgconfig/libpostproc*.pc \
+        /usr/local/lib/pkgconfig/libsw*.pc \
+        /usr/local/src/ffmpeg \
+        /root/.cache/pip; \
+    ldconfig
```
saturley-hall, line 206: Is it possible to have all of this run in a final `runtime` stage that we ship so that end users can build up to that stage and also have multimodal support?

container/templates/wheel_builder.Dockerfile
dmitry-tokarev-nv: Thanks @JanuszL. We ended up going a step further than logging the configure step: a post-build guard in the same RUN inspects the *shipped* ffmpeg (`-encoders`/`-decoders`/`-parsers`) and fails the build if anything beyond the intended VP9 set appears (i.e. any H.264/H.265/AAC/NVENC surface). That verifies the actual built binary rather than just the requested `./configure` flags, so a config regression cannot pass silently — and the `./configure` output itself is still visible in the build log. Resolving; happy to add an explicit configure summary too if you would still find it useful.

container/templates/sglang_runtime.Dockerfile
dmitry-tokarev-nv: Good call — that is how it is set up now. The codec cleanup, the in-tree VP9 ffmpeg, and PyNvVideoCodec (NVDEC hardware decode) all run in the `pre_runtime` stage, and the shipped final stage is `FROM pre_runtime AS runtime`, so building `--target runtime` inherits all of it and end users get full multimodal support (images + H.264/H.265 video via NVDEC hardware decode). The layout shifted since this comment (rebase/squash), so the multimodal setup now lands in the shippable runtime lineage rather than an intermediate-only stage. Resolving — shout if you meant a separate, dedicated multimodal stage instead and we can revisit.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
```diff
@@ -319,6 +413,76 @@ def _aux_value_for_item(
...
+            elif is_local_media_url(normalized):
+                content = await read_local_media_bytes(normalized, self._url_policy)
+            else:
+                return None
+            if not should_use_nvdec(probe_video_codec(content)):
+                return None
+            frames, _meta = await asyncio.to_thread(
+                decode_video_nvdec, content, self.num_video_frames
+            )
+            return frames
```
Chokoyo: **Preserve SGLang's video sampling contract and source metadata**

Discarding _meta and returning a bare ndarray changes SGLang's video preprocessing semantics. In SGLang 0.5.16, the URL path returns a VideoDecoderWrapper; preprocess_video() reads the original frame count and avg_fps, applies vision_config.video (fps/nframes/min_frames/max_frames) to choose source indices, and returns the source fps, original frame count, and selected source indices. This path instead samples a backend-global frame count(32 by default) before SGLang sees the video, while the shim later synthesizes 24 fps and indices 0..31.

A 50-frame, 10-fps clip configured with fps=2,max_frames=8 produced 8 frames through the normal SGLang path, but 32 frames plus fps=24,duration=1.33 through this path. qwen2_5_vl not consuming the returned video_metadata downstream does not make this safe: SGLang uses the source frame count and fps earlier in preprocess_video() to determine the frame set. 

We should let SGLang own frame selection, ideally through an NVDEC-backed decoder implementing len/avg_fps/get_frames_as_tensor(indices), and propagate the real {fps,total_num_frames,frames_indices} instead of synthesizing it.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
```diff
@@ -352,8 +516,14 @@ async def _encode_with_cache(
         # SGLang's _encode outputs are already on CPU; use CPU as target for consistency
         target_device = torch.device("cpu")
         if uncached_urls:
+            # H.264/H.265 URLs are hardware-decoded to frames here; SGLang's
+            # _encode accepts pre-decoded ndarrays in place of URLs. Cache keys
+            # above remain keyed on the original URL, so caching is unaffected.
+            encode_inputs = await self._build_encode_inputs(
+                uncached_urls, modality_name
+            )
```
Chokoyo, line 677: **Make NVDEC conversion not depend on the embedding cache**

`_build_encode_inputs()` is only called from `_encode_with_cache()`. With the default --multimodal-embedding-cache-capacity-gb=0, generate() takes the no-cache branch and passes the original video URLs directly to encoder._encode(), so NVDEC is never reached. Since this PR also removes the software video decoders from the shipped SGLang image, supported H.264/H.265 requests fail instead of falling back successfully. The new E/PD test masks this by explicitly setting the cache capacity to 0.1. Please apply the NVDEC conversion before the cache/no-cache split, or support it in both branches, and add coverage with the cache disabled.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
```diff
@@ -319,6 +472,89 @@ def _aux_value_for_item(
...
+            elif is_local_media_url(normalized):
+                content = await read_local_media_bytes(normalized, self._url_policy)
+            else:
+                return None
+            if not should_use_nvdec(probe_video_codec(content)):
+                return None
+            # Constructing the decoder opens the container and reads its frame
+            # index, so keep it off the event loop.
+            return await asyncio.to_thread(NvdecVideoDecoder, content)
+        except Exception as exc:  # noqa: BLE001 - additive; fall back to URL
```
Chokoyo: Could this cause url rejected by dynamo fall through to sglang's unvalidated fetch path? The broad exception handler returns None, causing `_build_encode_inputs()` to restore the original URL and pass it to sglang.

Chokoyo (commented):
Hi @dmitry-tokarev-nv, thanks for address the previous comments. I found one additional concern and left an inline comment, not blocking for this PR.

I’m also curious whether the current changes could break XPU video support. I saw the runtime dockerfiles unconditionally remove codec from XPU builds, while the replacement NVDEC path is cuda only.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
dmitry-tokarev-nv: Confirmed and fixed in 9f0650b1ad — you were right on both counts, including that the test masked it.

`--multimodal-embedding-cache-capacity-gb` defaults to `0` (`runtime_args.py:284`, gated by `if capacity_gb > 0`), so a stock deployment took the no-cache branch, handed SGLang the raw URL, and every H.264/H.265 request failed for want of a decoder in this image. The serve case set `0.1`, which routed it through the one path that worked.

The conversion now runs in both branches. Deliberately not hoisted above the branch, since `_encode_with_cache` converts only its uncached subset and hoisting would decode on cache hits too.

Coverage for the case that was missing:

- `test_video_routes_through_nvdec_with_cache_disabled` drives `generate()` with `_embedding_cache = None` and asserts the decoder — not the URL — reaches `encoder._encode`. Asserting on the URL would have passed against the buggy version.
- `video_e_pd_qwen_nvdec` no longer sets the cache flag, so it exercises the default configuration. Verified end to end on an RTX 6000 Ada: passes for both H.264 and H.265 payloads.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
dmitry-tokarev-nv: Fixed in 9f0650b1ad by giving SGLang a real decoder instead of pre-sampled pixels. Your point that `qwen2_5_vl` ignoring `video_metadata` doesn't make it safe was the crux — frame count and fps are consumed by `smart_nframes` to *choose* the frame set, not merely passed downstream.

New `NvdecVideoDecoder(VideoDecoderWrapper)` exposes `__len__`, `avg_fps` and `get_frames_as_tensor(indices)` over PyNvVideoCodec, so `preprocess_video` takes its `is_video_obj` branch, applies the model's own `vision_config.video` policy against true source values, and NVDEC decodes only the frames SGLang asks for. Subclassing is required rather than duck typing, since the gate is `isinstance`.

Measured on GPU against the unpatched artifact, H.264 fixture, 10 frames @ 10 fps:

```
policy nframes=4 -> returned 4 frames
meta = {'fps': 10.0, 'duration': 1.0, 'total_num_frames': 10,
        'frames_indices': [0, 3, 6, 9]}
negative control: ndarray -> metadata None (policy skipped)
```

Real source fps and frame count, indices chosen by SGLang's own `linspace`, and the policy honoured — the old path always produced `min(32, total)` with a synthesized `fps=24.0`. The ndarray control reproduces the previous behaviour, so this is a measured difference rather than an assumption.

Two notes. `get_frames_as_tensor` returns NHWC uint8 CPU in pinned memory, matching the upstream decord branch, since `preprocess_video` immediately does `permute(0, 3, 1, 2)`. And the metadata shim is no longer on the NVDEC path — it now covers only genuine pre-decoded arrays from frontend decoding, and its docstring says so.

A follow-up (10b56197e2) was needed to make this actually work end to end: `load_video` had no case for an already-constructed `VideoDecoderWrapper` and rejected it, and `encode_server` binds its own `load_video`, so patching `base_processor` alone wasn't enough.

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
dmitry-tokarev-nv: Good catch — you're right, and I reproduced it on hardware before changing anything rather than reasoning about it.

`validate_media_url` raises `UrlValidationError` on a policy violation, the broad `except` treated that as a decode failure, and `_build_encode_inputs` restored the original URL. SGLang applies no policy of its own: `_normalize_video_input` fetches http(s) through `get_mm_http_session()` and resolves `file://` to a bare path.

Reproduction, serving a clip from `127.0.0.1` (blocked by the default policy) and watching whether that server is reached:

```
1. policy REJECTS it: http:// URLs are not allowed
2. _maybe_nvdec_decoder returned: None
3. _build_encode_inputs -> ['http://127.0.0.1:.../internal-secret.mp4']
4. SGLang _normalize_video_input fetched 38128 bytes
   loopback server hits: ['/internal-secret.mp4']
```

The refusal was being logged as `NVDEC decode failed ... using URL passthrough`, so a policy denial was indistinguishable from a codec problem. A refused `file://` path behaved the same way but resolved to a readable local file — a local read rather than SSRF.

Fixed in c60d651be7: `UrlValidationError` now propagates. It subclasses `ValueError`, which is how this handler already reports a bad request, so it surfaces as one. Genuine decode failures still fall back to the URL, with a test pinning each half. Re-running the same probe after the fix reports `loopback server hits: []` — enforcement measured as silence on the blocked host, not merely an exception being raised.

One scope caveat worth stating plainly: this covers *refusals*. Video whose codec isn't hardware-routed (VP9, AV1) still reaches SGLang unvalidated, with no exception involved, because `should_use_nvdec` simply returns false. That predates this PR — `main`'s SGLang encode worker has no URL validation at all — and closing it would start rejecting input `main` accepts today, so I've left it for a separate change rather than folding a behaviour break into this one. Happy to file that if you agree.

components/src/dynamo/common/multimodal/nvdec_decoder.py
```diff
@@ -0,0 +1,270 @@
...
+    stacked = np.ascontiguousarray(np.stack(frames)).astype(np.uint8, copy=False)
+    if stacked.ndim != 4 or stacked.shape[-1] != 3:
+        raise RuntimeError(
+            f"NVDEC frames have unexpected shape {stacked.shape}; expected (T,H,W,3)"
+        )
+    metadata = {
+        "fps": fps,
+        "frames_indices": indices.tolist(),
+        "total_num_frames": int(total),
+    }
```
Chokoyo, line 281: **Preserve vLLM's sampling contract for short NVDEC videos**

When `total <= num_frames`, this function returns every original source frame, but the metadata omits `do_sample_frames`. vLLM's native video loader sets this flag to `len(frames_indices) == total_num_frames`: `True` only when all source frames are returned, and `False` when the loader has already pre-sampled the video. Qwen3-VL defaults a missing flag to `False`, so the current short-video path bypasses model-specific FPS sampling.

Reproduced this with vLLM 0.26.0's real Qwen3-VL processor using identical pixels from the committed 10-frame, 10-fps H.264 fixture. With the metadata currently returned here, all 10 frames were consumed (`video_grid_thw=[[5,16,20]]`, placeholder length 440). Setting the computed flag to `True` for this all-frames case selected 4 frames (`video_grid_thw=[[2,16,20]]`, placeholder length 176). An answer-keyword E2E check on this static fixture does not detect the difference.

I think we can preserve vLLM's native contract at the NVDEC to vLLM handoff by setting `do_sample_frames = len(frames_indices) == total_num_frames`

Chokoyo (approved):
LGTM other than 
- https://github.com/ai-dynamo/dynamo/pull/11836/changes#r3707206003

container/templates/compliance.Dockerfile
```diff
@@ -123,6 +123,33 @@ RUN python3 -m compliance.policy.validate \
...
+# CUDA only, and deliberately so -- state the limit plainly rather than let the
+# gate read as covering every image. The purge this gate depends on (the
+# `device == "cuda"` block in vllm_runtime.Dockerfile) removes the base image's
+# media packages and installs the in-tree LGPL ffmpeg in their place. The XPU
+# image gets neither, so it still carries its base image's ffmpeg stack --
+# libavcodec, libx264 and libx265 among them -- and running the gate there would
+# fail the build on packages this repo does not choose or control. That image's
+# base is owned by Intel; bringing it into the allowlist is their call, not a
+# side effect of a Dynamo compliance change. Extending the purge to `xpu` is the
+# fix if that ownership question is ever settled the other way.
```
rmccorm4: Adjusting the wording, mostly duplicating what's said just above
```suggestion
# NOTE: CUDA only for now. XPU image is not currently published on NGC, so it does not 
# require the codecs to be removed yet.
```

components/src/dynamo/common/multimodal/nvdec_decoder.py
dmitry-tokarev-nv: Confirmed and fixed in e1d8c454be — you were right, and your reproduction was more precise than anything my tests could have told me.

`decode_video_nvdec` now reports `"do_sample_frames": len(indices) == total`, which is vLLM's rule verbatim. Put it in the shared decoder rather than the vLLM adapter because the decoder is the only thing that knows whether it already sampled; TensorRT-LLM selects its metadata keys explicitly so the extra key is inert there, and SGLang no longer goes through this function at all since it now gets a real `VideoDecoderWrapper`.

Your last sentence is the part I kept, because it generalises: *an answer-keyword E2E check on this static fixture does not detect the difference.* That is exactly right, and it is the second finding of yours to land in that blind spot. The serve tests assert on what the model says about the clip, and that sentence does not change when the frame count does — I had written the same caveat into the fixture comment and still shipped a defect underneath it. So the new test is a unit test, covering fewer frames than requested, exactly equal, more, and a single-frame clip; it asserts the flag and the returned count, then re-derives your rule from the returned metadata rather than restating the implementation.

Validated on an RTX 6000 Ada with the fix stripped back out as a control, since a test that passes with and without the thing it guards is worth nothing:

| | |
|---|---|
| new test, fix installed | 4 passed |
| new test, fix stripped (`1 → 0` verified) | fails with `KeyError: 'do_sample_frames'` |
| full nvdec unit suite | 32 passed |
| vLLM H.264/H.265 NVDEC serve cases | 2 passed |

That last row matters because this changes behaviour: vLLM now samples the 10-frame fixture down, so the model sees fewer frames than before. It still answers correctly.

Worth flagging that my first run of that validation was itself wrong — the container ran as non-root, the overlay copy failed silently, and all three stages tested the shipped code while the control cheerfully reported OK. The numbers above are from the corrected run, which fails outright if the fix is not actually installed.

container/templates/compliance.Dockerfile
dmitry-tokarev-nv: Adopted in e1d8c454be — your rationale is the better one. Mine leaned on who owns the base image, yours states the reason it is actually acceptable, and it avoids attributing anything to a partner in a public artifact.

I kept one sentence of the mechanical reason, so the next person knows enabling the gate is not a one-liner:

```
# CUDA only for now. The XPU image is not currently published on NGC, so it does
# not yet require the codecs to be removed. It would also fail this gate today:
# the purge the gate depends on sits in the `device == "cuda"` block of
# vllm_runtime.Dockerfile, so the XPU image still carries its base image's media
# stack. Enable both together if that changes.
```

Ten lines down to five, and the duplication with the block above is gone.

One thing I cannot check myself: **the NGC claim**. I have no way to verify what is or is not published there, and it is now the stated justification for the gate's scope — so if that is wrong, the comment is wrong in the direction that matters. Could you confirm? Happy to reword if the real reason is something else.

For context on why the XPU image would fail today: it carries `libavcodec`, `libx264` and `libx265` from its base, and the gate flags them on presence. That is unchanged from `main` — this PR is simply the first thing that ever looked, because that lane only runs when a PR carries the `xpu` label.

docs/fern/pages/developer-guide/knowledge-base/modular-components/backends/sglang/multimodal.md
```diff
@@ -12,9 +12,30 @@ This document provides a comprehensive guide for multimodal inference using SGLa
...
 | **Image** | HTTP/HTTPS URL | Yes | Yes | Vision encoder generates embeddings |
 | **Image** | Data URL (Base64) | No | No |  |
-| **Video** | HTTP/HTTPS/`file://` URL | Yes | Yes | Vision encoder generates embeddings |
+| **Video** | HTTP/HTTPS, `file://`, `data:` | Yes, H.264/H.265 | Yes, H.264/H.265 | Hardware-decoded on NVDEC, then encoded to embeddings |
 | **Audio** | HTTP/HTTPS URL | No | No | Not supported in SGLang backend |
 
+> [!IMPORTANT]
+> **Video input is limited to H.264 and H.265, and requires a separate encode
+> worker.** The runtime image ships no software video decoder, so these codecs are
+> decoded on the GPU by NVDEC and no other codec (VP8, VP9, AV1) has a decoder
```
sttts: nit: "codecs are decoded" sounds strange

docs/fern/pages/developer-guide/knowledge-base/modular-components/backends/sglang/multimodal.md
```diff
@@ -12,9 +12,30 @@ This document provides a comprehensive guide for multimodal inference using SGLa
...
+> available.
+>
+> Hardware decode is wired into the encode worker
```
sttts: there are two meanings of encode in this text which makes it sound confusing.

dagil-nvidia (commented):
**My thoughts**

The compliance work and the NVDEC routing both look right, and the validation record is better than most PRs this size get. Two things I would want settled before this merges.

`output_format: "mp4"` now returns VP9. That path produced H.264 before this PR, through `export_to_video` on vLLM and through `h264_nvenc` with a libx264 fallback on SGLang. The extension has not changed and the contents have. Is a client that asked for mp4 expected to cope with VP9 in that container? TRT-LLM is the exception, since that path had no fallback and was already failing on parts without NVENC.

The other is the gate itself, where `CodecPolicy.load` shape-checks `deny_globs` and `exceptions` but not `allow_paths`, and `allow_paths` is the one that fails open. Written as a scalar it iterates per character, the leading `/` matches every path, and every violation reports `allowed`. I ran it at head, where `/opt/evil/libx264.so.164` classifies `('allowed', None)` as a scalar against `('violation', None)` as a list. Two lines in `load` and a case in `test_scan_codecs.py` close it.

Three doc lines went stale with this change. `use-cases/diffusion/overview.md:168` still promises `h264_nvenc` in the TRT-LLM image and is not in the diff. `video-decode-gpu-requirements.md:131` sends people to `DYN_ENABLE_MEDIA_DECODERS`, which appears twice repo-wide and both are prose. The same page says VP8 and VP9 keep using the software decode path, which the backend pages in this PR contradict now that `sglang.enable_media_ffmpeg` is `false`.

**From Claude**

`DYN_DISABLE_NVDEC` goes through `env_bool` everywhere except `encode_worker_handler.py:514`, so `=0` disables NVDEC on SGLang alone. The ordering comment on `probe_video_codec` does not match the loop, which resolves by table order. `nvdec_video_decoder.py:48` reaches across a package boundary for three private names, and the `SimpleDecoder` setup is duplicated across the two decoders with different temp-file idioms. Known Limitations should name the `_NVDEC_UNSAFE_MODEL_TYPES` gate alongside VP8/VP9/AV1.

The `pre-commit` and XPU reds are not from this branch. The merge-base is July 31 and `main` fixed the ruff one on August 3 in e258f04f62, so a rebase clears it.

saturley-hall (approved):

dmitry-tokarev-nv:
Thanks @dagil-nvidia - both points were worth raising, and one of them was a real hole in the gate itself.

## `allow_paths` fails open

Confirmed and fixed in a1919637e2. You were right, and I reproduced it before changing anything:

```
scalar allow_paths -> ('allowed', None)      # /opt/evil/libx264.so.164
list   allow_paths -> ('violation', None)
```

`load()` now rejects a non-list `allow_paths`, with three tests: the scalar is refused at load, the behaviour it prevents is pinned so the reason stays visible, and a well-formed list still loads and still allows. Verified on GPU with the check stripped back out, where the suite fails as it must — a guard that passes with and without the thing it guards is worth nothing.

The docstring immediately above already said a gate that silently passes every image is worse than no gate. That was the intent; `allow_paths` was the one key it didn't cover, and it was the one that fails open.

## `output_format: "mp4"` now returns VP9

Worth separating what changed from what was already true, because "produced H.264 before this PR" holds on some hardware and not on the hardware this runs on.

H.264 output cannot come from NVENC on the parts Dynamo is deployed on. NVIDIA omits the encoder from the datacenter line entirely:

| GPU | NVDEC (decode) | NVENC (encode) |
|---|---|---|
| A100 | 5 engines | **none** |
| H100, H200 | 7 engines | **none** |
| B200, GB200 | 7 engines | **none** |
| L4 / L40S / RTX 6000 Ada | 3–4 engines | 2–3 |

That asymmetry is why decode and encode are not the same problem here: NVDEC is present everywhere, NVENC is present nowhere that matters.

So the pre-PR behaviour was not "H.264 everywhere":

| | before | on datacenter GPUs |
|---|---|---|
| vLLM | software H.264 via `export_to_video` | worked — and is the codec surface this PR removes |
| SGLang | `h264_nvenc` + **libx264 fallback** | the GPL fallback is what made it work |
| TRT-LLM | `h264_nvenc`, no fallback | **already failing**, as you noted |

Restoring H.264 for the datacenter case therefore means restoring a *software* encoder, which is exactly what this change removes. NVENC would cover only L4/RTX-class parts and would leave the output codec varying by GPU — a worse contract than a uniform change. VP9 is the only option that behaves identically on every supported GPU, needs no hardware encoder, and carries no software H.264. The container stays MP4; the stream inside is VP9.

The real cost is the one you identified: the extension is unchanged and the contents are not. VP9-in-MP4 plays in Chrome, Firefox and ffmpeg, but not in Safari or QuickTime, so a client that hands the bytes straight to a player can fail with no error from us. **That goes in the release notes** — it is the mitigation that makes this defensible rather than something to be discovered later.

## The rest

- `DYN_DISABLE_NVDEC` now reads through `env_bool` in the SGLang encode worker (0f4e71274d). Measured in the shipped image: `0` and `false` read `True` through the old raw truthiness and `False` through `env_bool`, so that one setting meant opposite things depending on the backend.
- The stale doc lines are corrected (e8ed9f2ba2): `diffusion/overview.md` no longer promises `h264_nvenc`, and the SGLang support matrix no longer claims aggregated video works — the paragraph nine lines below it said the opposite.
- `video-decode-gpu-requirements.md:131` pointing at `DYN_ENABLE_MEDIA_DECODERS` is left alone deliberately: that switch is implemented in #12051, which has not merged. Fixing it here would either document a flag that does not exist yet or pre-empt that PR.
- On the pre-commit and XPU reds: agreed they are not from this branch. The ruff one cleared with the merge. XPU has two failures, both traced and neither ours — `gpu_memory_service` is missing from that image and fails identically on an untouched one, and `google_crc32c` arrived with the ModelExpress 0.4.0 → 0.5.0 bump (#12455); it is imported by `modelexpress/metadata/artifact_manifest.py` and the 2-card lane passed on the commit immediately before that merge.

docs/fern/pages/use-cases/multimodal-serving/video-decode-gpu-requirements.md
```diff
@@ -0,0 +1,156 @@
...
+## Behavior when NVDEC is unavailable
+
+Hardware decode is additive and never blocks a request on its own: routing falls through
+to the software decode path.
+
+> [!IMPORTANT]
+> Dynamo's runtime images ship a VP8/VP9-only in-tree FFmpeg, so the software path cannot
+> decode H.264 or H.265. If NVDEC is unavailable in one of these images, those formats
+> have no decoder and the request fails with an unsupported-codec error. Install the
+> optional software decoders (`DYN_ENABLE_MEDIA_DECODERS`) or fix the `video` capability.
```
rmccorm4: Since this PR doesn't include the `DYN_ENABLE_MEDIA_DECODERS` change, should we remove the references to it here?

components/src/dynamo/common/multimodal/nvdec_decoder.py
```diff
@@ -0,0 +1,283 @@
...
+    else:
+        frames, meta = <existing decoder>
+
+The returned frames match the existing ``VideoLoader`` contract: a host numpy
+array of shape ``(T, H, W, 3)``, dtype ``uint8``, RGB, C-contiguous, plus a
+metadata dict ``{"fps", "frames_indices", "total_num_frames"}``.
+
+Gating: NVDEC is used when PyNvVideoCodec is importable and ``DYN_DISABLE_NVDEC``
+is not set. When it is unavailable (CPU image, unsupported profile) the caller
+falls back to the software carriers (``DYN_ENABLE_MEDIA_DECODERS``) or surfaces
```
rmccorm4: Since this PR doesn't include the `DYN_ENABLE_MEDIA_DECODERS` change, should we remove the references to it here?

rmccorm4 (approved):

components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py
```diff
@@ -319,6 +505,110 @@ def _aux_value_for_item(
...
+        # vLLM and TensorRT-LLM -- the same setting meaning opposite things
+        # depending on the backend.
+        if env_bool(DISABLE_ENV):
+            return False
+        if not nvdec_available():
+            return False
+        model_type = getattr(self.encoder, "model_type", "") or ""
+        return model_type not in _NVDEC_UNSAFE_MODEL_TYPES
+
+    async def _maybe_nvdec_decoder(self, url: str) -> Optional[Any]:
```
rmccorm4, line 527: Codex finding:

---

SGLang downloads fallback videos twice. [_maybe_nvdec_decoder](https://github.com/ai-dynamo/dynamo/blob/8d64ea9b30fa3851f8f8bffbdf6a7f76b6096c0b/components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py#L553-L566) downloads the complete payload to probe it, then [_build_encode_inputs](https://github.com/ai-dynamo/dynamo/blob/8d64ea9b30fa3851f8f8bffbdf6a7f76b6096c0b/components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py#L604-L610) restores the URL for non-HW codecs or NVDEC failures. SGLang fetches it again, doubling ingress/latency, breaking one-use signed URLs, and discarding redirect validation for the bytes ultimately decoded. Pass the already-fetched bytes into the fallback.

dmitrytokarev: Good catch. Addressing

dmitry-tokarev-nv: fixed

components/src/dynamo/common/multimodal/nvdec_decoder.py
dmitry-tokarev-nv: addressed

docs/fern/pages/use-cases/multimodal-serving/video-decode-gpu-requirements.md
dmitry-tokarev-nv: addressed

docs/fern/pages/use-cases/multimodal-serving/video-decode-gpu-requirements.md
```diff
@@ -0,0 +1,156 @@
+---
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+title: Video Decode GPU Requirements
+---
+
+Dynamo decodes H.264 and H.265 (HEVC) video input on the GPU using NVDEC, NVIDIA's
+dedicated hardware video decoder, through
+[PyNvVideoCodec](https://pypi.org/project/PyNvVideoCodec/). Other formats (VP8, VP9)
+continue to use the software decode path.
```
tanmayv25: **Docs contradiction: this page says VP8/VP9 video input uses a software decode path, but the shipped images have no VP8/VP9 input decoder.**

Lines 9–10 state that "Other formats (VP8, VP9) continue to use the software decode path," and the *Behavior when NVDEC is unavailable* section (124–131) reinforces it by framing only H.264/H.265 as lacking a software path — implying VP8/VP9 decode via the in-tree VP8/VP9-only FFmpeg.

That contradicts the backend pages added in this same PR, which say the opposite:

- `vllm/multimodal.md`: "The runtime images ship no software video decoder … no other codec (VP8, VP9, AV1) has a decoder available."
- `sglang/multimodal.md`: "Video in any other format (VP8, VP9, AV1) … has no decoder in this image."

…and the PR's own **Known limitation**: "VP8/VP9/AV1 video input has no decoder in these images."

These images are built with `enable_media_ffmpeg: false` in all four `context.yaml` blocks, so the Rust `media-ffmpeg` `VideoDecoder` is not compiled in — the in-tree VP8/VP9 FFmpeg serves only the video-*output* (VP9 encode) path. The Python input decoders (opencv/av/decord) are also purged. So no VP8/VP9/AV1 decoder — Rust or Python — is present in the shipped images; the VP9 serve tests install `decord2`/`opencv-python-headless` (`installs_extra_dependencies`) precisely because the image has none. An operator who follows this page and sends a VP9 clip gets an unsupported-codec failure, not the documented software decode.

Suggest rewording lines 9–10 (and the 124–131 section) to state that VP8/VP9/AV1 video input has no decoder in the shipped images — opt-in only via `DYN_ENABLE_MEDIA_DECODERS` — to match the backend pages.

dmitry-tokarev-nv: addressed locally, waiting for CI to finish and then will push the docs fix.

tanmayv25 (approved):

docs/fern/pages/use-cases/multimodal-serving/video-decode-gpu-requirements.md
dmitry-tokarev-nv: pushed docs fix. thank you

sttts (approved):

PR #11840, opened 2026-07-17
feat(trtllm): add TensorRT-LLM native gRPC sidecar backend

tanmayv25:
## Summary

Adds `dynamo-trtllm-sidecar`, a standalone Rust worker that drives an out-of-process TensorRT-LLM engine over its native `trtllm.TrtllmService` gRPC (`serve --grpc`, TRT-LLM `1.3.0rc21`+). It composes with `dynamo_backend_common::run` and owns Dynamo worker registration, request conversion, transport, cancellation, and abort; TensorRT-LLM runs as its own process (no Python launcher, no unreleased protocol extensions).

**Supported:** aggregated generation with sampling, stop conditions, structured output (JSON schema / regex / grammar / structural tag), logprobs, streaming deltas + terminal usage, and abort-on-cancel.

**Rejected before dispatch:** disaggregation, multimodal, LoRA, KV-aware routing, encode, beam search, `n > 1`.

**`max_tokens` note:** TensorRT-LLM's gRPC marks `max_tokens` REQUIRED, so an omitted `max_tokens` (which the frontend forwards as `None` for the backend to default) is filled to the model's remaining context via `--context-length`, mirroring the in-process backend's `_default_max_tokens`. The proper fix is upstream — tracked in NVIDIA/TensorRT-LLM#16549.

Container packaging is intentionally deferred to a follow-up.

## Validation

- `cargo test -p dynamo-trtllm-sidecar` — 14 unit tests + the executable-contract test pass
- `cargo clippy -p dynamo-trtllm-sidecar --all-targets` — clean
- End-to-end against a live `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc21` gRPC server (Qwen3-0.6B, single GPU) + Dynamo frontend: non-streaming, streaming (SSE deltas + terminal usage), logprobs (incl. top-k), stop conditions, abort-on-cancel, and the omitted-`max_tokens` fallback (`--context-length` → HTTP 200; without it → clean rejection) all verified.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11840" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a TensorRT-LLM sidecar executable that connects Dynamo workers to TensorRT-LLM through gRPC.
  * Supports streaming text generation, sampling controls, stop conditions, structured output, log probabilities, cancellation, and request aborts.
  * Added configurable endpoint, model path, and context-length options.

* **Documentation**
  * Added setup instructions, supported capabilities, limitations, and protocol documentation.

* **Tests**
  * Added coverage for request handling, streaming responses, cancellation, validation, connection pooling, and command-line configuration.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -84,6 +84,7 @@
 /components/src/dynamo/vllm/tests/                                           @ai-dynamo/dynamo-backend-vllm-codeowners
 
 # === backend-trtllm  (@ai-dynamo/dynamo-backend-trtllm-codeowners) ===
+/lib/sidecar/trtllm/                                                         @ai-dynamo/dynamo-backend-trtllm-codeowners
 /examples/backends/trtllm/                                                   @ai-dynamo/dynamo-backend-trtllm-codeowners
 /components/src/dynamo/trtllm/                                               @ai-dynamo/dynamo-backend-trtllm-codeowners
 /components/src/dynamo/trtllm/tests/                                         @ai-dynamo/dynamo-backend-trtllm-codeowners
--- a/lib/sidecar/trtllm/src/convert.rs
+++ b/lib/sidecar/trtllm/src/convert.rs
@@ -42,11 +42,11 @@ pub(crate) fn build_generate_request(
         sampling_config: Some(pb::SamplingConfig {
             beam_width: 1,
             num_return_sequences: 1,
-            top_k: normalize_top_k(sampling.top_k),
+            top_k: normalize_top_k(sampling.top_k)?,
             top_p: sampling.top_p,
             temperature: sampling.temperature,
             min_p: sampling.min_p,
-            seed: normalize_seed(sampling.seed),
+            seed: normalize_seed(sampling.seed)?,
             min_tokens: stop.min_tokens,
             repetition_penalty: sampling.repetition_penalty,
             presence_penalty: sampling.presence_penalty,
@@ -54,7 +54,7 @@ pub(crate) fn build_generate_request(
             ..Default::default()
         }),
         output_config: Some(pb::OutputConfig {
-            logprobs: output.logprobs.map(logprob_count).transpose()?,
+            logprobs: output.logprobs.map(request_logprob_count).transpose()?,
             // prompt_logprobs is intentionally not requested: it is rejected in
             // `validate_request` (no `LLMEngineOutput` field to surface it).
             // TRT-LLM streams delta tokens; input tokens must not be echoed.
@@ -67,7 +67,11 @@ pub(crate) fn build_generate_request(
         stop: stop.stop.clone().unwrap_or_default(),
         stop_token_ids: stop_token_ids(request),
         ignore_eos: stop.ignore_eos.unwrap_or(false),
-        include_stop_token_in_output: sampling.include_stop_str_in_output.unwrap_or(false),
+        // `include_stop_token_in_output` retains matched stop *token IDs*, a
+        // different axis from Dynamo's `include_stop_str_in_output` (stop
+        // *strings*). Mapping one to the other would leak hidden stop/EOS tokens,
+        // so leave it at the default (strip) and reject the request-level flag in
+        // `validate_request` instead.
         ..Default::default()
     })
 }
@@ -102,21 +106,35 @@ fn max_tokens(
     Ok(context_length.saturating_sub(prompt_len).max(1))
 }
 
-fn normalize_top_k(top_k: Option<i32>) -> Option<i32> {
-    // Dynamo uses -1 (or absence) for "consider all tokens"; TRT-LLM treats an
-    // unset/0 top_k the same way. Only forward a positive cap.
+fn normalize_top_k(top_k: Option<i32>) -> Result<Option<i32>, DynamoError> {
+    // Dynamo uses -1/0 (or absence) for "consider all tokens"; TRT-LLM treats an
+    // unset top_k the same way. Forward only a positive cap; reject other
+    // negatives rather than silently widening them to "all tokens".
     match top_k {
-        Some(value) if value > 0 => Some(value),
-        _ => None,
+        None | Some(-1) | Some(0) => Ok(None),
+        Some(value) if value > 0 => Ok(Some(value)),
+        Some(value) => Err(client::invalid_argument(format!(
+            "top_k must be -1, 0, or positive; got {value}"
+        ))),
     }
 }
 
-fn normalize_seed(seed: Option<i64>) -> Option<u64> {
-    seed.and_then(|seed| u64::try_from(seed).ok())
+fn normalize_seed(seed: Option<i64>) -> Result<Option<u64>, DynamoError> {
+    // TRT-LLM's seed is `uint64`; reject a negative seed rather than silently
+    // dropping it and losing reproducibility.
+    seed.map(|seed| {
+        u64::try_from(seed)
+            .map_err(|_| client::invalid_argument(format!("seed must be non-negative; got {seed}")))
+    })
+    .transpose()
 }
 
-fn logprob_count(count: u32) -> Result<i32, DynamoError> {
-    i32::try_from(count).map_err(|_| {
+fn request_logprob_count(count: u32) -> Result<i32, DynamoError> {
+    // TRT-LLM computes the selected-token logprob only when at least one logprob
+    // is requested, so floor the wire value at 1: `logprobs=0` (selected token,
+    // no alternatives) still yields the chosen-token logprob. The original count
+    // is preserved in `ResponseState` to decide whether to surface alternatives.
+    i32::try_from(count.max(1)).map_err(|_| {
         client::invalid_argument(format!("logprobs request must fit in i32; got {count}"))
     })
 }
@@ -251,7 +269,50 @@ fn validate_request(request: &PreprocessedRequest) -> Result<(), DynamoError> {
             "KV-aware data-parallel routing is not supported by the TensorRT-LLM sidecar",
         ));
     }
+    if request
+        .routing
+        .as_ref()
+        .and_then(|routing| routing.cache_namespace.as_deref())
+        .is_some_and(|namespace| !namespace.is_empty())
+    {
+        // `cache_namespace` is the request-scoped KV-cache isolation contract. The
+        // proto carries `cache_salt_id`, but TRT-LLM 1.3.0rc21 does not plumb it
+        // through the gRPC servicer, so honoring it would let requests from
+        // different namespaces share prefix-cache entries. Reject until supported.
+        return Err(client::invalid_argument(
+            "cache namespace isolation is not supported by the TensorRT-LLM sidecar",
+        ));
+    }
+    if request
+        .routing
+        .as_ref()
+        .and_then(|routing| routing.priority)
+        .is_some_and(|priority| priority != 0)
+    {
+        // The shared contract forwards a nonzero priority to the engine, but the
+        // TRT-LLM gRPC request has no priority field, so it would be silently
+        // dropped and change queue ordering. Reject until the wire protocol
+        // carries it.
+        return Err(client::invalid_argument(
+            "request priority is not supported by the TensorRT-LLM sidecar",
+        ));
+    }
+    if request.stop_conditions.max_thinking_tokens.is_some() {
+        // A reasoning-token budget the sidecar can neither forward nor enforce.
+        return Err(client::invalid_argument(
+            "max_thinking_tokens is not supported by the TensorRT-LLM sidecar",
+        ));
+    }
     let sampling = &request.sampling_options;
+    if sampling.include_stop_str_in_output == Some(true) {
+        // Retaining stop *strings* has no faithful mapping in the TRT-LLM gRPC
+        // request (its only control, `include_stop_token_in_output`, governs stop
+        // *token IDs* and would leak hidden stop/EOS tokens). Reject rather than
+        // mis-map — consistent with rejecting visible stop token IDs above.
+        return Err(client::invalid_argument(
+            "include_stop_str_in_output is not supported by the TensorRT-LLM sidecar",
+        ));
+    }
     if sampling.n.unwrap_or(1) != 1 {
         return Err(client::invalid_argument("n must be 1"));
     }
@@ -269,7 +330,8 @@ fn validate_request(request: &PreprocessedRequest) -> Result<(), DynamoError> {
 pub(crate) struct ResponseState {
     prompt_tokens: u32,
     completion_tokens: u32,
-    expect_output_logprobs: bool,
+    streamed_tokens: u32,
+    output_logprobs: Option<u32>,
 }
 
 impl ResponseState {
@@ -277,7 +339,8 @@ impl ResponseState {
         Self {
             prompt_tokens: request.token_ids.len() as u32,
             completion_tokens: 0,
-            expect_output_logprobs: request.output_options.logprobs.is_some(),
+            streamed_tokens: 0,
+            output_logprobs: request.output_options.logprobs,
         }
     }
 
@@ -298,7 +361,11 @@ impl ResponseState {
             Some(pb::generate_response::Response::Complete(complete)) => {
                 self.convert_complete(complete).map(Some)
             }
-            None => Ok(None),
+            // A response with neither payload is protocol drift, not an
+            // empty delta — fail rather than silently discard it.
+            None => Err(client::protocol_error(
+                "response carried neither a chunk nor a complete payload",
+            )),
         }
     }
 
@@ -334,6 +401,9 @@ impl ResponseState {
             chunk.prompt_tokens,
             chunk.completion_tokens,
         )?;
+        self.streamed_tokens = self
+            .streamed_tokens
+            .saturating_add(chunk.token_ids.len() as u32);
         if chunk.token_ids.is_empty() {
             return Ok(None);
         }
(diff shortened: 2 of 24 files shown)


tanmayv25:
Thanks for the reviews (@devin-ai-integration, @coderabbitai). Addressed in d8e7957:

**Fixed**
- **`stop_token_ids_visible` silently dropped** — now rejected in `validate_request`. TRT-LLM's single `include_stop_token_in_output` flag can't retain a stop token per-token, so forwarding it (with the global flag) would still mis-handle tool-parser EOS tokens; rejecting is the honest behavior, consistent with the vLLM sidecar.
- **`prompt_logprobs` requested but discarded** — now rejected in `validate_request` and no longer forwarded (no `LLMEngineOutput` field to surface them, so no wasted server compute).
- **`GetModelInfo` / `Abort` unbounded** — both now use a 30s RPC timeout; `start` still falls back to `--context-length` on timeout.
- **`AbortResponse.success` ignored** — `Client::abort` now errors when `success == false`.
- **`map_logprobs` empty slice silently bypassed** — bypasses only when logprobs weren't requested; a missing/short slice is now a protocol error.
- **`--context-length 0` accepted** — rejected in `from_parsed`.
- **Abort test didn't verify the RPC** — the fake service records abort IDs; new `abort_sends_the_abort_rpc_to_the_server` asserts the ID reaches the server. Added rejection cases (prompt_logprobs, visible stops) to the unsupported-features test.

**Verified, no change**
- **Usage cumulative assumption** — confirmed against live rc21 over multi-chunk streams; `max()` correctly tolerates chunks that omit the count. Added a clarifying comment.

**Validation:** `cargo test -p dynamo-trtllm-sidecar` (15 unit + integration + executable), `cargo clippy -p dynamo-trtllm-sidecar --all-targets` clean, and re-ran e2e against live `1.3.0rc21` — logprobs (incl. top-k over multi-token and streaming), no-`max_tokens` fallback, and abort.

lib/sidecar/trtllm/src/engine.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    pub fn from_env() -> Result<(Self, WorkerConfig), DynamoError> {
+        Self::from_parsed(<Args as clap::Parser>::parse())
+    }
+
+    pub fn from_args(argv: Vec<String>) -> Result<(Self, WorkerConfig), DynamoError> {
+        let args = <Args as clap::Parser>::try_parse_from(argv)
+            .map_err(|err| error::invalid_argument(err.to_string()))?;
+        Self::from_parsed(args)
+    }
```
connorcarpenter15, line 67: `from_env()` and `from_args()` do pretty much the same thing. In #11754, we folded `from_env()` into `from_args()` to just have one shared helper.

CODEOWNERS
```diff
@@ -85,6 +85,7 @@
 /components/src/dynamo/vllm/tests/                                           @ai-dynamo/dynamo-backend-vllm-codeowners
 
 # === backend-trtllm  (@ai-dynamo/dynamo-backend-trtllm-codeowners) ===
+/lib/trtllm-sidecar/                                                         @ai-dynamo/dynamo-backend-trtllm-codeowners
```
connorcarpenter15: Don't we want to put this under `/lib/backend/trtllm-sidecar/`?

connorcarpenter15 (commented):
Not sure if we want to hold off on this PR until we land the `sidecar-common` crate on main. We'd be adding a lot of logic that would immediately get removed.

tanmayv25:
### Rebased onto `refactor/sidecar-common` (stacked PR)

This PR now targets **`refactor/sidecar-common`** (#11844) instead of `main`, and the sidecar has been migrated onto the shared **`dynamo-sidecar-common`** crate that branch introduces:

- crate moved to **`lib/sidecar/trtllm`**, matching the relocated `vllm`/`sglang` sidecars
- dropped the crate-local `error.rs` and hand-rolled channel pool in favor of `dynamo_sidecar_common`'s `GrpcChannelPool` / `GrpcEndpoint` / `SidecarArgs` / `GrpcTransportConfig` and error module — mirroring the vLLM sidecar exactly
- the sidecar now honors the shared `--grpc-connections` / `--grpc-*-timeout` flags via `SidecarArgs`

All earlier review fixes are preserved (prompt_logprobs + visible stop-token rejections, `GetModelInfo`/`Abort` RPC timeouts, `AbortResponse.success` check, strict `map_logprobs`, `--context-length` `max_tokens` fallback + `== 0` rejection, and the `abort_sends_the_abort_rpc_*` test).

**Merge order:** this depends on #11844. Once #11844 lands in `main`, I'll retarget this back to `main`.

**Validation:** `cargo test -p dynamo-trtllm-sidecar` (14 unit + integration + executable), `cargo clippy --all-targets` clean, pre-commit + copyright + CODEOWNERS-strict pass, and re-ran e2e against live `1.3.0rc21` over the new `GrpcChannelPool` (incl. 8 concurrent requests round-robining a 4-connection pool, logprobs, streaming, no-`max_tokens` fallback).

.lycheeignore
```diff
@@ -28,3 +28,7 @@
 
 # NGC intermittently returns 503 for this version-filtered page from GitHub runners.
 ^https://catalog\.ngc\.nvidia\.com/orgs/nvidia/teams/ai-dynamo/containers/dynamo-frontend\?version=1\.2\.1$
+
+# NeurIPS proceedings PDFs refuse connections from GitHub runners (network/firewall).
+# The papers load fine from a normal browser.
+^https://proceedings\.neurips\.cc/
```
connorcarpenter15: I don't think this change should be folded into this PR if it's fixing a docs error.

lib/sidecar/trtllm/src/tests.rs
```diff
@@ -0,0 +1,608 @@
...
+        .collect()
+        .await
+}
+
+// ---------------------------------------------------------------------------
+// Request-building unit tests
+// ---------------------------------------------------------------------------
+
+#[test]
+fn request_maps_sampling_stop_and_output_fields() {
```
connorcarpenter15, line 300: `request_maps_sampling_stop_and_output_fields` pins incorrect behavior: the fixture combines a hidden stop-token ID with include_stop_str_in_output=true. Conversion enables TensorRT-LLM’s global token-inclusion flag, potentially exposing a token Dynamo marked hidden.

lib/sidecar/trtllm/src/tests.rs
```diff
@@ -0,0 +1,608 @@
...
+}
+
+#[test]
+fn omitted_max_tokens_default_is_floored_at_one() {
+    let mut req = request();
+    req.stop_conditions.max_tokens = None;
+    // Prompt already fills (or exceeds) the context: default must not underflow to 0.
+    let proto = build_generate_request(&req, "req", Some(2)).expect("build with fallback");
+    assert_eq!(proto.max_tokens, 1);
+}
```
connorcarpenter15, line 362: These tests could probably be merged

lib/sidecar/trtllm/src/tests.rs
```diff
@@ -0,0 +1,608 @@
...
+        build_generate_request(&req, "req", None).expect_err("oversized logprobs must fail");
+    assert!(error.to_string().contains("must fit in i32"));
+}
+
+// ---------------------------------------------------------------------------
+// Response-conversion unit tests
+// ---------------------------------------------------------------------------
+
+#[test]
+fn chunk_then_complete_produces_delta_then_terminal_usage() {
```
connorcarpenter15, line 458: Fairly overlapped with `aggregated_generation_streams_delta_then_terminal`, might want to remove.

lib/sidecar/trtllm/src/tests.rs
```diff
@@ -0,0 +1,608 @@
...
+    // opening the stream fails rather than yielding an error item.
+    let context = dynamo_backend_common::testing::mock_context();
+    let result = engine
+        .generate(request(), GenerateContext::new(context, None))
+        .await;
+    assert!(result.is_err());
+}
+
+#[tokio::test]
+async fn cancellation_yields_a_cancelled_terminal() {
```
connorcarpenter15, line 653: Should probably be replaced with `run_conformance` from `backend-common`.

connorcarpenter15 (commented):
Overall LGTM, left a few comments.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+    if request
+        .routing
+        .as_ref()
+        .is_some_and(|routing| routing.dp_rank.is_some() || routing.prefill_dp_rank.is_some())
+    {
+        return Err(client::invalid_argument(
+            "KV-aware data-parallel routing is not supported by the TensorRT-LLM sidecar",
+        ));
+    }
```
connorcarpenter15, line 271: [P1] Cache isolation is silently discarded. `routing.cache_namespace` is the request-scoped KV-cache isolation contract (`lib/llm/src/protocols/common/preprocessor.rs:57-63`) and the preprocessor populates it from `cache_salt`/`x-tenant-id` (`lib/llm/src/preprocessor.rs:1053-1086`), but this validator neither maps nor rejects it. Requests from different namespaces therefore reach TensorRT-LLM unsalted and can share backend prefix-cache entries. TRT-LLM v1.3.0rc21 also does not plumb the proto's `cache_salt_id` through its gRPC servicer, so please reject every nonempty cache namespace until end-to-end support is verified.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+            exclude_input_from_output: true,
+            ..Default::default()
+        }),
+        max_tokens: max_tokens(request, context_length)?,
+        streaming: true,
+        guided_decoding: guided_decoding(request)?,
+        stop: stop.stop.clone().unwrap_or_default(),
+        stop_token_ids: stop_token_ids(request),
+        ignore_eos: stop.ignore_eos.unwrap_or(false),
+        include_stop_token_in_output: sampling.include_stop_str_in_output.unwrap_or(false),
```
connorcarpenter15: [P1] This conflates stop-string retention with stop-token visibility. `stop_token_ids()` merges ordinary and hidden IDs (`convert.rs:124-136`), while preprocessing normally adds hidden EOS IDs (`lib/llm/src/preprocessor.rs:971-982`). Enabling TRT-LLM's global inclusion flag for `include_stop_str_in_output` can therefore retain hidden EOS tokens too, exposing them in token-ID output and letting the local decoder terminate before the server's usage terminal. It also conflicts with visible parser-end tokens, which are rejected at `convert.rs:222-233` even though `engine.rs:103-106` enables tool/reasoning parsers. Please model hidden and visible stops explicitly, or reject the unsupported combinations before dispatch.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+    if sampling.n.unwrap_or(1) != 1 {
+        return Err(client::invalid_argument("n must be 1"));
+    }
+    if sampling.best_of.unwrap_or(1) != 1 {
+        return Err(client::invalid_argument("best_of must be 1"));
+    }
+    if sampling.use_beam_search.unwrap_or(false) {
+        return Err(client::invalid_argument("beam search is not supported"));
+    }
+    Ok(())
```
connorcarpenter15, line 325: [P1] `max_thinking_tokens` is accepted but never enforced. The OpenAI preprocessing path deliberately forwards this backend control (`lib/llm/src/protocols/openai.rs:184-215`), but it is neither mapped nor rejected here. Reasoning-model requests can consequently exceed the requested thinking budget, increasing latency and token cost while appearing successful. Please reject `stop_conditions.max_thinking_tokens.is_some()` until the native gRPC contract supports it.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+    // unset/0 top_k the same way. Only forward a positive cap.
+    match top_k {
+        Some(value) if value > 0 => Some(value),
+        _ => None,
+    }
+}
+
+fn normalize_seed(seed: Option<i64>) -> Option<u64> {
+    seed.and_then(|seed| u64::try_from(seed).ok())
+}
```
connorcarpenter15: [P2] Several request controls fail open instead of preserving or rejecting their semantics. Invalid negative `top_k` values become “all tokens,” negative seeds become an absent seed, and nonzero backend `routing.priority` is ignored by `validate_request` even though the shared contract says it is forwarded to the engine (`lib/llm/src/protocols/common/preprocessor.rs:65-77`). These requests return success with different sampling, reproducibility, or queue ordering. Please make these conversions fallible and reject nondefault priority until the wire protocol supports it.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+    completion_tokens: u32,
+    expect_output_logprobs: bool,
+}
+
+impl ResponseState {
+    pub(crate) fn new(request: &PreprocessedRequest) -> Self {
+        Self {
+            prompt_tokens: request.token_ids.len() as u32,
+            completion_tokens: 0,
+            expect_output_logprobs: request.output_options.logprobs.is_some(),
```
connorcarpenter15: [P2] Reducing `output_options.logprobs` to a boolean loses the requested alternative count. `map_logprobs` consequently treats `logprobs=0` like a positive top-logprob request, always returns `top_logprobs=Some(...)`, and fabricates rank-0 candidate data when the server supplies no alternatives (`convert.rs:377-421`). The established sidecar contract keeps selected-token logprobs for zero while omitting top alternatives (`lib/sidecar/vllm/src/tests.rs:275-286`). Please retain `Option<u32>` in `ResponseState` and only construct alternatives for a positive count.

lib/sidecar/trtllm/src/convert.rs
```diff
@@ -0,0 +1,433 @@
...
+        &mut self,
+        response: pb::GenerateResponse,
+    ) -> Result<Option<LLMEngineOutput>, DynamoError> {
+        match response.response {
+            Some(pb::generate_response::Response::Chunk(chunk)) => self.convert_chunk(chunk),
+            Some(pb::generate_response::Response::Complete(complete)) => {
+                self.convert_complete(complete).map(Some)
+            }
+            None => Ok(None),
+        }
```
connorcarpenter15, line 369: [P2] Response conversion fails open on protocol drift. A response without its oneof payload is silently ignored here; terminal `output_token_ids` are never checked against the streamed deltas (`convert.rs:350-374`); and every unknown finish reason becomes a successful `Stop` (`convert.rs:425-431`). A malformed or version-skewed server can therefore produce partial or inconsistent output reported as success. Please track the expected request ID and streamed tokens, validate the terminal sequence/counts, and return protocol errors for missing payloads and unknown finish reasons.

connorcarpenter15 (changes requested):
Requesting changes for silent request-contract loss and fail-open response handling at the TensorRT-LLM conversion boundary. The inline comments cover cache isolation, stop visibility, thinking budgets, request controls, logprob semantics, and response protocol validation.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Fixed — `cache_namespace` is now rejected in `validate_request`. As you noted, rc21 does not plumb the proto’s `cache_salt_id` through the servicer, so failing closed is the honest behavior until it is supported end-to-end. Unit test added.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Fixed — stopped mapping `include_stop_str_in_output` onto TRT-LLM’s `include_stop_token_in_output` (different axis; would retain hidden stop/EOS token IDs). The flag now stays at its default (strip) and `include_stop_str_in_output=true` is rejected in `validate_request`, consistent with the existing visible-stop-token rejection. Covered by a unit test.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Fixed — `max_thinking_tokens` is now rejected in `validate_request` (no native gRPC field to enforce the budget), mirroring the vLLM sidecar.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Fixed — `normalize_top_k`/`normalize_seed` are now fallible: an invalid negative `top_k` (other than -1/0) and a negative seed are rejected rather than silently widened to “all tokens” / dropped. Nonzero `routing.priority` is also rejected in `validate_request`, since the gRPC request carries no priority field. Unit tests added.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Fixed — `ResponseState` now stores the requested count as `Option<u32>`. The wire request floors the count at 1 so the selected-token logprob is still computed, and `map_logprobs` surfaces top alternatives only for a positive count (selected-token logprob only for `logprobs=0`), matching the vLLM sidecar contract. Added a `logprobs=0` unit test.

lib/sidecar/trtllm/src/convert.rs
tanmayv25: Addressed — a response with neither payload now returns a protocol error (was silently dropped), an unknown `finish_reason` returns a protocol error (was fail-open to `Stop`), and the terminal’s cumulative `output_token_ids` are validated against the streamed delta count. I kept the terminal check tolerant of an omitted echo (empty `output_token_ids`) so a server that only carries finish state in the terminal is not false-rejected. Unit tests cover all three paths.

lib/sidecar/trtllm/src/tests.rs
tanmayv25: Fixed alongside the `include_stop_str_in_output` change — the fixture no longer sets that flag, and the test now asserts `include_stop_token_in_output` stays false, so no hidden stop-token exposure is pinned.

.lycheeignore
tanmayv25: Agreed — dropped the `.lycheeignore` commit from this PR entirely; it is unrelated to the sidecar and can land on its own.

lib/sidecar/trtllm/src/tests.rs
tanmayv25: I looked into this: no engine in the repo currently adopts `run_conformance`, and the sibling vLLM/sglang sidecars use the same bespoke fake-server style. I kept these integration tests consistent with the siblings rather than introduce a divergent harness in just this crate. Happy to migrate all three sidecars to `run_conformance` as a separate follow-up if you would prefer that direction.

lib/sidecar/trtllm/src/tests.rs
tanmayv25: Reviewed — I kept these as separate cases; after this revision they also anchor the new rejection/logprobs coverage and read more clearly as individual failure messages. Can consolidate if you feel strongly.

lib/sidecar/trtllm/src/tests.rs
tanmayv25: The unit test does overlap the integration path, but it is the cleanest place to assert multi-token logprob mapping + terminal usage directly on `ResponseState` without the transport, and it now sits alongside the new protocol-drift unit tests. Fine to drop if you would rather rely on the integration test.

connorcarpenter15 (approved):

rmccorm4 (approved):
Approving the CODEOWNERS change for @ai-dynamo/dynamo-process-codeowners

rmccorm4:
/ok to test [d39b2fb](https://github.com/ai-dynamo/dynamo/pull/11840/commits/d39b2fb327537749bfefdf9f249d615f1ba3c8e9)

nv-tusharma (approved):

PR #11849, opened 2026-07-17
refactor(vllm): generalize custom encoder artifacts

furionw:
_Based on #12416._

## Why

Custom encoder drivers currently hard-code tensor results even though downstream adapters own prompt construction. A generic artifact type lets model-specific adapters consume structured results without teaching the batching and async layers model semantics.

## What Change

- Generalize the backend, async driver, and adapter over `ArtifactT`.
- Keep tensor and placeholder validation inside the linear adapter.
- Preserve opaque artifact ordering through cross-request batching.

## Test Plan

- Focused custom-encoder and handler unit tests: 69 passed.
- Pre-commit hooks passed.

## Summary by CodeRabbit

* **New Features**
  * Custom vision encoders can now produce backend-specific artifacts, enabling more flexible multimodal processing.
  * Arbitrary encoder artifacts are preserved through asynchronous encoding and prompt preparation.
  * Removed the requirement for a hardcoded image placeholder token.

* **Bug Fixes**
  * Improved validation when required image token configuration is missing.

* **Tests**
  * Added coverage for custom artifact handling, ordering, and configuration validation.

PR #11850, opened 2026-07-17
feat(vllm): add Qwen3-VL custom encoder adapter

furionw:
_Based on #11849._

## Why

Qwen3-VL external image embeddings must preserve the pre-merge grid and packed DeepStack features so vLLM can reconstruct native multimodal inputs without rerunning the vision tower. Malformed encoder artifacts should fail at the adapter boundary, before they reach EngineCore. A model-aware adapter supports both dense and MoE Qwen3-VL decoders while keeping feature construction with the producer.

## What Change

- Add packed Qwen3-VL artifacts and native prompt adaptation.
- Validate tensor and grid consistency before engine handoff.
- Support dense and MoE architectures with shared model-config helpers.

## Test Plan

- 26 focused custom-encoder adapter and handler tests passed.
- `pre-commit` passed for all changed files.


components/src/dynamo/vllm/multimodal_utils/custom_encoder/adapter/factory.py
```diff
@@ -23,4 +31,20 @@ def create_custom_encoder_adapter(
...
 
+    if model_config is None:
+        raise ValueError("CustomEncoder requires the resolved vLLM ModelConfig")
+    architectures = _model_architectures(model_config)
+    qwen3_vl_architectures = [
+        architecture
+        for architecture in architectures
+        if architecture in _QWEN3_VL_ARCHITECTURES
+    ]
+    if qwen3_vl_architectures:
```
GuanLuo: Move this to `qwen3_vl.py`?

components/src/dynamo/vllm/multimodal_utils/model_config.py
```diff
@@ -20,6 +20,14 @@ def _hidden_size(model_config: Any) -> int:
     return value
 
 
+def _model_architectures(model_config: Any) -> tuple[str, ...]:
```
GuanLuo, line 23: This `model_config.py` seems universal for vLLM and can be promoted to under `multimodal_utils`?

components/src/dynamo/vllm/multimodal_utils/custom_encoder/adapter/factory.py
furionw: moved

components/src/dynamo/vllm/multimodal_utils/model_config.py
furionw: moved.

components/src/dynamo/vllm/multimodal_utils/custom_encoder/adapter/factory.py
```diff
@@ -23,4 +31,15 @@ def create_custom_encoder_adapter(
 ) -> CustomEncoderAdapter[Any]:
     """Create the adapter selected by the resolved downstream decoder."""
 
+    if model_config is None:
+        raise ValueError("CustomEncoder requires the resolved vLLM ModelConfig")
+    architectures = _model_architectures(model_config)
```
GuanLuo: Can move to exception if only needed there

GuanLuo (approved):

components/src/dynamo/vllm/multimodal_utils/custom_encoder/adapter/factory.py
furionw: moved. thank you

PR #11860, opened 2026-07-17
fix(mocker): publish native ZMQ KV batches

jthomson04:
## Summary

- Publish every nonempty deferred mocker KV visibility boundary as one native ZMQ event batch.
- Keep normal event sinks per-event while preserving singleton fallback behavior for raw sinks without native batching.

## Details

- Deferred events are converted once, forwarded individually to the normal sink, and emitted as one raw ZMQ batch.
- When both sinks are configured, a normal-sink failure is warned but does not suppress the complete raw batch; the two transports are intentionally independent.
- Default raw batching attempts every event and returns the first failure only after attempting the remainder.

- `lib/mocker/src/scheduler/kv_event_sink.rs`: deferred visibility-boundary forwarding and its coverage.
- `lib/mocker/src/services/zmq_events.rs`: one queued vector maps to one native ZMQ payload.

## Related Issues

- [x] Confirmed — no related issue.

## Validation

- `cargo test -p dynamo-mocker --lib`
- `cargo clippy -p dynamo-mocker --lib --tests -- -D warnings`

### DIFF
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -58,6 +58,20 @@ pub struct RawKvEvent {
 /// Trait for publishing transport-specific raw KV event payloads.
 pub trait RawKvEventSink: Send + Sync {
     fn publish(&self, event: RawKvEvent) -> anyhow::Result<()>;
+
+    /// Publishes raw events that share one source visibility boundary.
+    ///
+    /// Implementations that do not have a native batch representation retain
+    /// singleton delivery semantics by default.
+    fn publish_batch(&self, events: Vec<RawKvEvent>) -> anyhow::Result<()> {
+        let mut first_error = None;
+        for event in events {
+            if let Err(error) = self.publish(event) {
+                first_error.get_or_insert(error);
+            }
+        }
+        first_error.map_or(Ok(()), Err)
+    }
 }
 
 /// Shared KV event publisher bundle used by schedulers and KV managers.
@@ -112,6 +126,30 @@ impl KvEventPublishers {
             })?;
         }
 
+        Ok(())
+    }
+
+    /// Publishes a normal KV event without also forwarding it to a raw sink.
+    ///
+    /// Deferred live-scheduler forwarding uses this to preserve existing
+    /// per-event error handling for normal sinks while sending the matching raw
+    /// events as one native batch.
+    pub(crate) fn publish_event_sink_only(
+        &self,
+        event: KvCacheEvent,
+        storage_tier: StorageTier,
+    ) -> anyhow::Result<()> {
+        if let Some(sink) = self.event_sink.as_ref() {
+            sink.publish_with_storage_tier(event, storage_tier)?;
+        }
+        Ok(())
+    }
+
+    /// Publishes raw events as one source visibility boundary.
+    pub(crate) fn publish_raw_batch(&self, events: Vec<RawKvEvent>) -> anyhow::Result<()> {
+        if let Some(sink) = self.raw_sink.as_ref() {
+            sink.publish_batch(events)?;
+        }
         Ok(())
     }
 }
@@ -1396,8 +1434,48 @@ impl MockEngineArgs {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Mutex;
+
     use super::*;
     use serde_json::json;
+
+    #[derive(Default)]
+    struct FailingRawSink {
+        attempts: Mutex<Vec<u64>>,
+    }
+
+    impl RawKvEventSink for FailingRawSink {
+        fn publish(&self, event: RawKvEvent) -> anyhow::Result<()> {
+            self.attempts.lock().unwrap().push(event.event.event_id);
+            if event.event.event_id == 2 {
+                anyhow::bail!("injected raw sink failure");
+            }
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn raw_sink_batch_fallback_attempts_later_events_after_failure() {
+        let sink = FailingRawSink::default();
+        let error = sink
+            .publish_batch(
+                (1..=3)
+                    .map(|event_id| RawKvEvent {
+                        event: KvCacheEvent {
+                            event_id,
+                            data: dynamo_kv_router::protocols::KvCacheEventData::Cleared,
+                            dp_rank: 0,
+                        },
+                        block_token_ids: None,
+                        storage_tier: StorageTier::Device,
+                    })
+                    .collect(),
+            )
+            .unwrap_err();
+
+        assert_eq!(error.to_string(), "injected raw sink failure");
+        assert_eq!(*sink.attempts.lock().unwrap(), vec![1, 2, 3]);
+    }
 
     #[test]
     fn direct_request_priorities_are_backward_compatible() {
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -44,6 +44,23 @@ pub trait KvCacheEventSink: Send + Sync {
     ) -> anyhow::Result<()> {
         self.publish(event)
     }
+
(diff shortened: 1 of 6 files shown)


lib/mocker/src/common/protocols.rs
```diff
@@ -58,6 +58,17 @@ pub struct RawKvEvent {
 /// Trait for publishing transport-specific raw KV event payloads.
 pub trait RawKvEventSink: Send + Sync {
     fn publish(&self, event: RawKvEvent) -> anyhow::Result<()>;
+
+    /// Publishes raw events that share one source visibility boundary.
+    ///
+    /// Implementations that do not have a native batch representation retain
+    /// singleton semantics by default.
+    fn publish_batch(&self, events: Vec<RawKvEvent>) -> anyhow::Result<()> {
```
jthomson04, line 83: The `?` here means a failure on event N silently drops events N+1.. of the batch, whereas the old flush loop warned and continued to the next event. All in-tree non-ZMQ raw sinks are test capture sinks, so impact is low — but consider warn-and-continue here, or a doc note that a mid-batch failure drops the remainder.

jthomson04 (commented):
## Review: fix(mocker): publish native ZMQ KV batches

**Overview:** Previously each deferred KV event flushed as its own single-event ZMQ payload, while real vLLM emits one multi-event batch per engine step. This PR batches all events from one visibility boundary into a single native wire-format batch: `RawKvEventSink::publish_batch` (per-event default, overridden by `ZmqKvEventSink`), a reusable `encode_event_batch`, and a split deferred-flush path that keeps per-event publishes for the normal sink.

**Strengths:**
- Well-scoped; matches the PassStart/PassEnd flush points in `vllm/live.rs` and `sglang/live.rs`.
- The test round-trips through the router's real `decode_event_batch` — a genuine cross-crate wire-compatibility check.
- One-timestamp-per-batch matches vLLM native semantics better than the old per-event timestamps.
- Performance is strictly better: one channel send, one msgpack serialization, and one ZMQ multipart per boundary; the raw path now moves `block_token_ids` instead of the old `to_vec()` copy.

**Notes (details inline):**
- Behavior change: raw batch is now sent even when normal-sink publishes fail (old code short-circuited via `?`). Arguably an improvement, but worth calling out in the description.
- Default `publish_batch` drops the remainder of a batch on first error, where the old loop warned and continued per event.
- Minor: two separate `impl ZmqKvEventSink` blocks; `publish_event_with_storage_tier` naming is easy to confuse with `publish_with_storage_tier`.
- Pre-existing, not introduced here: the `assert_eq!` in `convert_to_zmq_events` panics the spawned publisher task; with batching, one malformed event now takes the whole batch with it.

**Risk:** downstream ZMQ consumers now receive multi-event payloads where the mocker previously only sent single-event ones — but real vLLM already emits multi-event batches, and the decode test proves the router handles them.

**Verdict:** LGTM with minor suggestions.

lib/mocker/src/common/protocols.rs
```diff
@@ -114,6 +125,30 @@ impl KvEventPublishers {
 
         Ok(())
     }
+
+    /// Publishes a normal KV event without also forwarding it to a raw sink.
+    ///
+    /// Deferred live-scheduler forwarding uses this to preserve existing
+    /// per-event error handling for normal sinks while sending the matching raw
+    /// events as one native batch.
+    pub(crate) fn publish_event_with_storage_tier(
```
jthomson04: Naming nit: `publish_event_with_storage_tier` is one word away from the existing `publish_with_storage_tier` on the same struct, which is easy to misread at call sites. Something like `publish_event_sink_only` would be more self-describing. The doc comment does mitigate this.

lib/mocker/src/services/zmq_events.rs
```diff
@@ -237,12 +221,53 @@ impl ZmqKvEventSink {
...
     fn publish(&self, event: RawKvEvent) -> anyhow::Result<()> {
+        self.enqueue(vec![event])
+    }
+
+    fn publish_batch(&self, events: Vec<RawKvEvent>) -> anyhow::Result<()> {
+        self.enqueue(events)
+    }
+}
+
+impl ZmqKvEventSink {
```
jthomson04: Style: this is a second `impl ZmqKvEventSink` block (the constructor lives in the first one above the trait impl). Consider merging `enqueue` into the existing impl block — the split reads like an accident.

lib/mocker/src/scheduler/kv_event_sink.rs
```diff
@@ -206,15 +206,30 @@ pub(crate) fn publish_deferred_kv_events(
...
+
+    for event in &raw_events {
+        if let Err(error) =
+            sinks.publish_event_with_storage_tier(event.event.clone(), event.storage_tier)
+        {
             tracing::warn!("Failed to forward buffered KV event: {error}");
         }
     }
+
+    if let Err(error) = sinks.publish_raw_batch(raw_events) {
```
jthomson04, line 231: Behavior change worth confirming: previously a normal-sink failure short-circuited via `?` inside `KvEventPublishers::publish_with_storage_tier`, which also skipped that event's raw publish. Now the raw batch is always sent even if some normal-sink publishes failed. I'd call this an improvement (the two transports shouldn't be coupled), but it's a semantic change the PR description doesn't call out.

lib/mocker/src/scheduler/kv_event_sink.rs
```diff
@@ -226,3 +241,84 @@ pub(crate) fn publish_deferred_fpm(sink: &FpmPublisher, snapshots: Vec<ForwardPa
...
+        }
+
+        fn publish_batch(&self, events: Vec<RawKvEvent>) -> Result<()> {
+            self.batches.lock().unwrap().push(events);
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn deferred_visibility_boundary_emits_one_native_zmq_batch() {
```
jthomson04: Coverage gaps (all minor): the default `publish_batch` fallback is never exercised since `CapturingRawSink` overrides it; `publish_deferred_kv_events` is only tested with `event_sink: None`, so the "normal sink still gets per-event publishes" half of the split is untested; and there's no empty-input early-return test.

lib/mocker/src/common/protocols.rs
jthomson04: Addressed in ec6d4b8928: the fallback now attempts every singleton event, retains the first error after the loop, and has coverage proving events 1, 2, and 3 are all attempted when event 2 fails.

lib/mocker/src/common/protocols.rs
jthomson04: Renamed to `publish_event_sink_only` in ec6d4b8928.

lib/mocker/src/services/zmq_events.rs
jthomson04: Merged `enqueue` into the original `impl ZmqKvEventSink` block in ec6d4b8928.

lib/mocker/src/scheduler/kv_event_sink.rs
jthomson04: Confirmed. The sinks are intentionally independent: a normal-sink failure is warned but does not suppress the complete raw batch. This is now stated in the PR Details section.

lib/mocker/src/scheduler/kv_event_sink.rs
jthomson04: Covered in ec6d4b8928: added fallback failure/continuation coverage and extended the deferred-boundary test to verify normal per-event delivery, one raw native batch, and empty-drain suppression.

PeaBrane (approved):
✅

jthomson04:
/ok to test [093792e](https://github.com/ai-dynamo/dynamo/commit/093792e47f06b5591d4e954b7701d87d66ac74e1)

PR #11861, opened 2026-07-17
refactor(vllm): derive prompt usage from request output

furionw:
## Why

Dynamo threads an embedding sequence length beside the prompt solely to report usage, duplicating information returned by vLLM. Supported vLLM builds already preserve mixed-prompt token IDs and synthesize one placeholder ID per pure-embedding row, so `RequestOutput.prompt_token_ids` is the authoritative sequence length. Using that output directly removes handler coupling and prevents the prompt builder from deciding response accounting.

## What Change

- Remove embedding sequence-length plumbing from prompt construction and generation.
- Derive prompt usage from returned token IDs, preserving empty-prompt semantics.
- Align unit and E2E coverage with the vLLM output contract.

## Test Plan

Focused unit tests (92 passed, 5 skipped):

```bash
python -m pytest \
  components/src/dynamo/vllm/tests/test_vllm_prompt_embeds.py \
  components/src/dynamo/vllm/tests/test_vllm_worker_handler.py \
  -q
```

Prompt-embeddings E2E on an H100 with Python 3.12.3, vLLM 0.25.1, and `Qwen/Qwen3-0.6B`:

```bash
python -m pytest \
  tests/frontend/test_prompt_embeds.py::TestPromptEmbedsE2E::test_usage_prompt_tokens_not_zero \
  -v -s --timeout=900
```

The fixture launched a real Dynamo frontend and vLLM worker with `--enable-prompt-embeds`. It submitted base64-encoded `20 × 1024` embeddings, then asserted `prompt_tokens == 20` and `total_tokens == prompt_tokens + completion_tokens`. Result: 1 passed in 84.98 seconds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token usage reporting for requests that use prompt embeddings.
  * Ensured prompt token counts are derived consistently across streaming, prefill, and decode flows.
  * Simplified prompt handling to provide more reliable completion usage information.

* **Tests**
  * Updated coverage for embedding-based prompts, cached tokens, and decode-only scenarios.
  * Clarified documentation around placeholder tokens used in embedding prompts.

### DIFF
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -2507,7 +2507,7 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
 
     def _create_prompt_from_embeddings(
         self, prompt_embeds_base64: str
-    ) -> tuple[EmbedsPrompt, int, torch.Tensor]:
+    ) -> tuple[EmbedsPrompt, torch.Tensor]:
         """
         Decode prompt embeddings and create EmbedsPrompt for vLLM.
 
@@ -2515,9 +2515,8 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
             prompt_embeds_base64: Base64-encoded PyTorch tensor
 
         Returns:
-            Tuple of (EmbedsPrompt, sequence_length, tensor) where:
+            Tuple of (EmbedsPrompt, tensor) where:
             - EmbedsPrompt: The vLLM prompt input
-            - sequence_length: Extracted from tensor shape for usage statistics
             - tensor: The decoded tensor (for logging shape/dtype)
 
         Raises:
@@ -2529,13 +2528,10 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
                 f"prompt embeds should have dim 2 after vllm processing, but found dim {embeddings_tensor.dim()}"
             )
 
-        # Extract sequence length from tensor shape for usage reporting
-        sequence_length = embeddings_tensor.shape[0]
-
         # EmbedsInputs TypedDict has: {type: 'embeds', prompt_embeds: Tensor, cache_salt?: str}
         prompt = EmbedsPrompt(prompt_embeds=embeddings_tensor)
 
-        return prompt, sequence_length, embeddings_tensor
+        return prompt, embeddings_tensor
 
     def _build_prompt_from_request(
         self,
@@ -2545,7 +2541,7 @@ class BaseWorkerHandler(ABC, Generic[RequestT, ResponseT]):
         log_prefix: str = "",
         mm_processor_kwargs: Dict[str, Any] | None = None,
         mixed_embeds: tuple[torch.Tensor, list[int], list[bool]] | None = None,
-    ) -> tuple[TokensPrompt | EmbedsPrompt | None, int | None, Dict[str, Any] | None]:
+    ) -> tuple[TokensPrompt | EmbedsPrompt | None, Dict[str, Any] | None]:
         """
         Build a prompt from request, handling both prompt_embeds and token_ids.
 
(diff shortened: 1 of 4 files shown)


components/src/dynamo/vllm/handlers.py
```diff
@@ -2640,37 +2628,30 @@ def _build_prompt_from_request(
...
-            prompt_tokens = embedding_sequence_length
-        elif request_output.prompt_token_ids:
-            prompt_tokens = len(request_output.prompt_token_ids)
-        else:
-            prompt_tokens = None
+        prompt_tokens = (
+            len(request_output.prompt_token_ids)
+            if request_output.prompt_token_ids is not None
+            else None
+        )
```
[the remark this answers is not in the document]
furionw: Good callout — confirmed against upstream history. vLLM commit [`9a4600e4`](https://github.com/vllm-project/vllm/commit/9a4600e4dcbbd13988c31d5198d3ab8b4172ecca) introduced V1 prompt-embedding support and the `RequestOutput` fallback that synthesizes `[0] * len(prompt_embeds)` when no prompt token IDs were supplied. The first stable tag containing that behavior is v0.11.0; v0.10.2 does not contain it.

Dynamo currently pins vLLM 0.25.1 for its Python/CUDA/CPU builds, and the XPU image contains vLLM 0.24.0, so every supported runtime is above that floor. The added E2E exercises this contract with the pinned v0.25.1 runtime. Given that supported-version coverage, we do not need to retain a separate embedding-length side channel or add a runtime guard.

karen-sy (approved):

tests/frontend/test_prompt_embeds.py
```diff
@@ -266,9 +266,9 @@ def test_usage_prompt_tokens_not_zero(self, dynamo_client):
         """
         CRITICAL REGRESSION TEST: Ensure prompt_tokens is correctly reported.
```
furionw, line 267: i am removing the below comments because they are no longer accurate.

dynamo doesn't need to specify what is happening inside engine anyway.

arguably, they can be 

```
        vLLM represents a pure embedding prompt with one placeholder token ID
        per embedding row. The worker derives usage from those returned IDs,
        so this assertion verifies that contract end to end.
```

but i opt-in for simplicity

components/src/dynamo/vllm/handlers.py
keivenchang, line 2654: Hey Qi, I think this changes the empty-list case from fail-visible to fail-silent — the old truthy check (`elif request_output.prompt_token_ids:`) mapped `[]` to `prompt_tokens=None` (usage visibly absent), while `is not None` maps `[]` to a hard `0`, which is exactly the silent-zero symptom the E2E regression test guards against. Given the vLLM floor you confirmed on the other thread this edge may never fire, so I think you can either keep the truthy check as cheap defense-in-depth (an engine that doesn't synthesize placeholder IDs then fails loudly with missing usage instead of reporting 0), or leave it as-is if you want `[]` to mean a genuine zero-length prompt. One related note — the sibling copy in `omni/output_formatter.py` (`_build_completion_usage`, ~line 433) still maps empty to `None` via `is_empty_payload`, so the two copies now disagree on this exact edge; might be worth aligning them whichever way you pick.

keivenchang (approved):

components/src/dynamo/vllm/handlers.py
furionw: Good catch. vLLM rejects a zero-length decoder prompt before generation, so this should be unreachable on the supported path. I still adopted the truthy check as cheap defense-in-depth: `[]` now maps to unavailable usage (`None`), while every nonempty ID list—including `[0] * N`—reports its length. The empty-list unit case now expects both `prompt_tokens` and `total_tokens` to be `None` in commit [`516b14124c`](https://github.com/ai-dynamo/dynamo/commit/516b14124c405bdc0089e593edeb27d707e33426), which also aligns this handler with the Omni formatter. Thanks!

PR #11870, opened 2026-07-18
feat(runtime): support explicit transport host overrides

jthomson04:
## Summary

- Add `DYN_EVENT_PLANE_HOST` to override the address advertised by direct ZMQ event publishers.
- Give `DYN_EVENT_PLANE_HOST` and `DYN_TCP_RESPONSE_STREAM_HOST` one contract: each accepts an IPv4 or IPv6 literal, or an exact network interface name.
- Trim configured host values, reject wildcard advertisement addresses and non-Unicode values, and preserve the existing IPv4-first automatic selection with IPv6 fallback.
- Build advertised endpoints with `SocketAddr`, preserve the transport host settings and `DYN_TCP_RPC_HOST` across snapshot restore, and document the supported configuration.

Direct ZMQ continues to bind `0.0.0.0`; this PR changes its advertised discovery address only. Full IPv6 binding and address-selection policy remain in #13126. NATS and ZMQ broker mode are unchanged.

## Validation

- `cargo test -p dynamo-runtime utils::ip_resolver -- --test-threads=1` — 10 passed
- `cargo test -p dynamo-runtime event_plane -- --test-threads=1` — 43 passed
- `cargo test -p dynamo-runtime --lib pipeline::network::tcp::server::tests -- --test-threads=1` — 39 passed
- `cargo check -p dynamo-runtime --all-targets`
- `cargo clippy -p dynamo-runtime --lib --tests -- -D warnings`
- `PYTHONPATH=components/src uv run --no-project --with pytest --with pytest-asyncio python -m pytest -c /dev/null -p no:cacheprovider --asyncio-mode=auto components/src/dynamo/common/tests/test_snapshot_restore_context.py` — 5 passed
- `fern check`
- `fern docs broken-links`

### DIFF
--- a/components/src/dynamo/common/snapshot/constants.py
+++ b/components/src/dynamo/common/snapshot/constants.py
@@ -32,6 +32,8 @@ RESTORE_RUNTIME_ENV_NAMES = {
     "DYN_EVENT_PLANE",
     "DYN_EVENT_PLANE_HOST",
     # DistributedRuntime infrastructure env read after restore.
+    "DYN_TCP_RESPONSE_STREAM_HOST",
+    "DYN_TCP_RESPONSE_STREAM_PORT",
     "NATS_SERVER",
     "ETCD_ENDPOINTS",
     # Runtime system server/readiness env read after restore.
--- a/components/src/dynamo/common/snapshot/constants.py
+++ b/components/src/dynamo/common/snapshot/constants.py
@@ -30,6 +30,7 @@ RESTORE_RUNTIME_ENV_NAMES = {
     "DYN_DISCOVERY_BACKEND",
     "DYN_REQUEST_PLANE",
     "DYN_EVENT_PLANE",
+    "DYN_EVENT_PLANE_HOST",
     # DistributedRuntime infrastructure env read after restore.
     "NATS_SERVER",
     "ETCD_ENDPOINTS",
--- a/components/src/dynamo/common/snapshot/constants.py
+++ b/components/src/dynamo/common/snapshot/constants.py
@@ -33,6 +33,7 @@ RESTORE_RUNTIME_ENV_NAMES = {
     # DistributedRuntime infrastructure env read after restore.
     "DYN_EVENT_PLANE_HOST",
     "DYN_TCP_RPC_HOST",
+    "DYN_TCP_RPC_PORT",
     "DYN_TCP_RESPONSE_STREAM_HOST",
     "DYN_TCP_RESPONSE_STREAM_PORT",
     "NATS_SERVER",
--- a/components/src/dynamo/common/snapshot/constants.py
+++ b/components/src/dynamo/common/snapshot/constants.py
@@ -30,8 +30,9 @@ RESTORE_RUNTIME_ENV_NAMES = {
     "DYN_DISCOVERY_BACKEND",
     "DYN_REQUEST_PLANE",
     "DYN_EVENT_PLANE",
-    "DYN_EVENT_PLANE_HOST",
     # DistributedRuntime infrastructure env read after restore.
+    "DYN_EVENT_PLANE_HOST",
+    "DYN_TCP_RPC_HOST",
     "DYN_TCP_RESPONSE_STREAM_HOST",
     "DYN_TCP_RESPONSE_STREAM_PORT",
     "NATS_SERVER",
--- a/docs/design-docs/event-plane.md
+++ b/docs/design-docs/event-plane.md
@@ -54,7 +54,7 @@ python3 -m dynamo.vllm --event-plane zmq --model Qwen/Qwen3-0.6B
 | Variable | Description | Default |
 |----------|-------------|---------|
 | `DYN_EVENT_PLANE` | Transport: `nats` or `zmq` | `zmq` (all backends) |
-| `DYN_EVENT_PLANE_HOST` | IP advertised by direct ZMQ publishers; ignored by NATS and ZMQ broker mode | Auto-detected local IP |
+| `DYN_EVENT_PLANE_HOST` | IPv4 address or interface advertised by direct ZMQ publishers; ignored by NATS and ZMQ broker mode | Auto-detected local IPv4 address |
 | `NATS_SERVER` | NATS server URL (NATS transport only) | `nats://localhost:4222` |
 
 When `DYN_EVENT_PLANE` is not set, the default is **zmq**. Set `DYN_EVENT_PLANE=nats`
@@ -91,9 +91,11 @@ When using ZMQ (`DYN_EVENT_PLANE=zmq`):
 - When publishers come and go (e.g., workers scaling up/down), subscribers dynamically adjust their connections.
 
 To advertise an address on a specific network interface, such as InfiniBand, set
-`DYN_EVENT_PLANE_HOST` to that interface's IPv4 or IPv6 address. When unset, Dynamo
-advertises an auto-detected local address. This setting only changes direct ZMQ publisher
-discovery endpoints; it does not apply to NATS or ZMQ broker mode.
+`DYN_EVENT_PLANE_HOST` to an IPv4 address or interface name. Interface names resolve to
+the lowest usable IPv4 address. IPv6 literals and `0.0.0.0` are rejected because direct
+ZMQ publishers bind IPv4 listeners and discovery requires a dialable address. When unset,
+Dynamo advertises an auto-detected local IPv4 address. This setting only changes direct
+ZMQ publisher discovery endpoints; it does not apply to NATS or ZMQ broker mode.
 
 Example setup:
 
--- a/docs/design-docs/event-plane.md
+++ b/docs/design-docs/event-plane.md
@@ -54,6 +54,7 @@ python3 -m dynamo.vllm --event-plane zmq --model Qwen/Qwen3-0.6B
 | Variable | Description | Default |
 |----------|-------------|---------|
 | `DYN_EVENT_PLANE` | Transport: `nats` or `zmq` | `zmq` (all backends) |
+| `DYN_EVENT_PLANE_HOST` | IP advertised by direct ZMQ publishers; ignored by NATS and ZMQ broker mode | Auto-detected local IP |
 | `NATS_SERVER` | NATS server URL (NATS transport only) | `nats://localhost:4222` |
 
 When `DYN_EVENT_PLANE` is not set, the default is **zmq**. Set `DYN_EVENT_PLANE=nats`
@@ -89,10 +90,16 @@ When using ZMQ (`DYN_EVENT_PLANE=zmq`):
 - Subscribers automatically discover and connect to all active publishers.
 - When publishers come and go (e.g., workers scaling up/down), subscribers dynamically adjust their connections.
 
+To advertise an address on a specific network interface, such as InfiniBand, set
+`DYN_EVENT_PLANE_HOST` to that interface's IPv4 or IPv6 address. When unset, Dynamo
+advertises an auto-detected local address. This setting only changes direct ZMQ publisher
+discovery endpoints; it does not apply to NATS or ZMQ broker mode.
+
 Example setup:
 
 ```bash
 export DYN_EVENT_PLANE=zmq
+export DYN_EVENT_PLANE_HOST=192.0.2.10
 
 # Start workers -- each binds a ZMQ socket, registers with discovery
 python3 -m dynamo.vllm --model Qwen/Qwen3-0.6B \
--- a/docs/fern/pages/reference/components/runtime-configuration.mdx
+++ b/docs/fern/pages/reference/components/runtime-configuration.mdx
@@ -132,9 +132,10 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
 </ParamField>
 
 <ParamField path="DYN_TCP_RESPONSE_STREAM_HOST" type="string" default="auto-detected local address">
-  IPv4 address or network interface name used to bind and advertise the TCP response stream server.
-  An interface name resolves to its lowest usable IPv4 address. Dynamo rejects explicit IPv6 and
-  unspecified addresses such as `0.0.0.0`.
+  IPv4 or IPv6 address, or exact network interface name, used to bind and advertise the TCP response
+  stream server. If an interface name occurs more than once in the operating system's interface
+  list, Dynamo uses the last address. Dynamo trims whitespace and treats an empty value as unset.
+  Unspecified addresses such as `0.0.0.0` and `::` are invalid.
 </ParamField>
 
 <ParamField path="DYN_TCP_RESPONSE_STREAM_PORT" type="integer" default="0">
@@ -151,10 +152,15 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
   Environment variable: `DYN_EVENT_PLANE`
 </ParamField>
 
-<ParamField path="DYN_EVENT_PLANE_HOST" type="string" default="auto-detected local IPv4 address">
-  IPv4 address or network interface name advertised by direct ZMQ event publishers. An interface
-  name resolves to its lowest usable IPv4 address. Dynamo rejects IPv6 and unspecified addresses
-  such as `0.0.0.0`. This variable does not affect NATS or ZMQ broker mode.
+<ParamField path="DYN_EVENT_PLANE_HOST" type="string" default="auto-detected local address">
+  IPv4 or IPv6 address, or exact network interface name, advertised by direct ZMQ event publishers.
+  If an interface name occurs more than once in the operating system's interface list, Dynamo uses
+  the last address. Dynamo trims whitespace and treats an empty value as unset. Unspecified addresses
+  such as `0.0.0.0` and `::` are invalid.
+
+  This variable changes the discovery address, not the listener address. Direct ZMQ continues to bind
+  an IPv4 wildcard, so an IPv6 value does not enable IPv6 listening. This variable does not affect
+  NATS or ZMQ broker mode.
 </ParamField>
 
 <ParamField path="--connector" type="string" default="null">
--- a/docs/fern/pages/reference/components/runtime-configuration.mdx
+++ b/docs/fern/pages/reference/components/runtime-configuration.mdx
@@ -140,6 +140,12 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
   Environment variable: `DYN_EVENT_PLANE`
 </ParamField>
 
+<ParamField path="DYN_EVENT_PLANE_HOST" type="string" default="auto-detected local IP">
+  IPv4 or IPv6 address advertised by direct ZMQ event publishers. Set this variable to an address
+  on the network interface that subscribers use to reach the publisher. Dynamo rejects hostnames.
+  This variable does not affect NATS or ZMQ broker mode.
+</ParamField>
+
 <ParamField path="--connector" type="string" default="null">
   KV-cache transfer connector. Accepts zero or more values. Deprecated for vLLM — use `--kv-transfer-config` instead. For TRT-LLM, valid options are `nixl`, `lmcache`, `kvbm`, `null`, and `none`.
 
--- a/docs/fern/pages/reference/components/runtime-configuration.mdx
+++ b/docs/fern/pages/reference/components/runtime-configuration.mdx
@@ -131,6 +131,17 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
   <span className="enum-values"><span className="enum-label">Allowed values:</span> <Badge intent="note" minimal>json</Badge> <Badge intent="note" minimal>msgpack</Badge></span>
 </ParamField>
 
+<ParamField path="DYN_TCP_RESPONSE_STREAM_HOST" type="string" default="auto-detected local address">
+  IPv4 address or network interface name used to bind and advertise the TCP response stream server.
+  An interface name resolves to its lowest usable IPv4 address. Dynamo rejects explicit IPv6 and
+  unspecified addresses such as `0.0.0.0`.
+</ParamField>
+
+<ParamField path="DYN_TCP_RESPONSE_STREAM_PORT" type="integer" default="0">
+  Port used by the TCP response stream server. When unset or set to `0`, the operating system assigns
+  a free port. Values must be between `0` and `65535`.
+</ParamField>
+
 <ParamField path="--event-plane" type="string" default="zmq">
   Event publishing transport. ZMQ is the default for every discovery backend; select `nats`
   explicitly to use NATS Core. See [Event Plane](../../developer-guide/knowledge-base/concepts/system-architecture/architecture.md#event-plane) for design details.
@@ -140,10 +151,10 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
   Environment variable: `DYN_EVENT_PLANE`
 </ParamField>
 
-<ParamField path="DYN_EVENT_PLANE_HOST" type="string" default="auto-detected local IP">
-  IPv4 or IPv6 address advertised by direct ZMQ event publishers. Set this variable to an address
-  on the network interface that subscribers use to reach the publisher. Dynamo rejects hostnames.
-  This variable does not affect NATS or ZMQ broker mode.
+<ParamField path="DYN_EVENT_PLANE_HOST" type="string" default="auto-detected local IPv4 address">
+  IPv4 address or network interface name advertised by direct ZMQ event publishers. An interface
+  name resolves to its lowest usable IPv4 address. Dynamo rejects IPv6 and unspecified addresses
+  such as `0.0.0.0`. This variable does not affect NATS or ZMQ broker mode.
 </ParamField>
 
 <ParamField path="--connector" type="string" default="null">
--- a/docs/fern/pages/reference/components/runtime-configuration.mdx
+++ b/docs/fern/pages/reference/components/runtime-configuration.mdx
@@ -139,8 +139,9 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
 </ParamField>
 
 <ParamField path="DYN_TCP_RESPONSE_STREAM_PORT" type="integer" default="0">
-  Port used by the TCP response stream server. When unset or set to `0`, the operating system assigns
-  a free port. Values must be between `0` and `65535`.
+  Port used by the TCP response stream server. Dynamo trims surrounding whitespace. When the value
+  is unset, empty, or `0`, the operating system assigns a free port. Values must be between `0` and
+  `65535`.
 </ParamField>
 
 <ParamField path="--event-plane" type="string" default="zmq">
--- a/lib/runtime/src/config/environment_names.rs
+++ b/lib/runtime/src/config/environment_names.rs
@@ -622,7 +622,8 @@ pub mod tcp_response_stream {
     /// If unset or 0, the OS assigns a free ephemeral port.
     pub const DYN_TCP_RESPONSE_STREAM_PORT: &str = "DYN_TCP_RESPONSE_STREAM_PORT";
 
-    /// Host/interface for the TCP response stream server.
+    /// Advertised IPv4 address or interface for the TCP response stream server.
+    /// IPv6 and unspecified addresses are rejected.
     /// If unset, the server auto-detects a routable local IP.
     pub const DYN_TCP_RESPONSE_STREAM_HOST: &str = "DYN_TCP_RESPONSE_STREAM_HOST";
 }
@@ -641,8 +642,9 @@ pub mod event_plane {
     /// Event plane codec selection: "json" or "msgpack".
     pub const DYN_EVENT_PLANE_CODEC: &str = "DYN_EVENT_PLANE_CODEC";
 
-    /// Host IP advertised by direct ZMQ event publishers.
-    /// If unset, the runtime auto-detects a local IP address.
+    /// IPv4 address or interface advertised by direct ZMQ event publishers.
+    /// IPv6 and unspecified addresses are rejected.
+    /// If unset, the runtime auto-detects a local IPv4 address.
     pub const DYN_EVENT_PLANE_HOST: &str = "DYN_EVENT_PLANE_HOST";
 
     /// Bounded capacity of the direct ZMQ event-subscriber's merged event channel.
--- a/lib/runtime/src/config/environment_names.rs
+++ b/lib/runtime/src/config/environment_names.rs
@@ -641,6 +641,10 @@ pub mod event_plane {
     /// Event plane codec selection: "json" or "msgpack".
     pub const DYN_EVENT_PLANE_CODEC: &str = "DYN_EVENT_PLANE_CODEC";
 
+    /// Host IP advertised by direct ZMQ event publishers.
+    /// If unset, the runtime auto-detects a local IP address.
+    pub const DYN_EVENT_PLANE_HOST: &str = "DYN_EVENT_PLANE_HOST";
+
     /// Bounded capacity of the direct ZMQ event-subscriber's merged event channel.
     ///
     /// Many peer publishers (e.g. every other frontend under replica-sync) feed
@@ -875,6 +879,7 @@ mod tests {
             // Event Plane
             event_plane::DYN_EVENT_PLANE,
             event_plane::DYN_EVENT_PLANE_CODEC,
+            event_plane::DYN_EVENT_PLANE_HOST,
             event_plane::DYN_ZMQ_EVENT_SUBSCRIBER_CHANNEL_CAPACITY,
             // ZMQ Broker
             zmq_broker::DYN_ZMQ_BROKER_URL,
--- a/lib/runtime/src/config/environment_names.rs
+++ b/lib/runtime/src/config/environment_names.rs
@@ -806,6 +806,10 @@ pub mod event_plane {
     /// Event plane codec selection: "json" or "msgpack".
     pub const DYN_EVENT_PLANE_CODEC: &str = "DYN_EVENT_PLANE_CODEC";
 
+    /// Host IP advertised by direct ZMQ event publishers.
+    /// If unset, the runtime auto-detects a local IP address.
+    pub const DYN_EVENT_PLANE_HOST: &str = "DYN_EVENT_PLANE_HOST";
+
     /// Bounded capacity of the direct ZMQ event-subscriber's merged event channel.
     ///
     /// Many peer publishers (e.g. every other frontend under replica-sync) feed
@@ -1066,6 +1070,7 @@ mod tests {
             // Event Plane
             event_plane::DYN_EVENT_PLANE,
             event_plane::DYN_EVENT_PLANE_CODEC,
+            event_plane::DYN_EVENT_PLANE_HOST,
             event_plane::DYN_ZMQ_EVENT_SUBSCRIBER_CHANNEL_CAPACITY,
             // ZMQ Broker
             zmq_broker::DYN_ZMQ_BROKER_URL,
--- a/lib/runtime/src/config/environment_names.rs
+++ b/lib/runtime/src/config/environment_names.rs
@@ -745,8 +745,8 @@ pub mod tcp_response_stream {
     /// If unset or 0, the OS assigns a free ephemeral port.
     pub const DYN_TCP_RESPONSE_STREAM_PORT: &str = "DYN_TCP_RESPONSE_STREAM_PORT";
 
-    /// Advertised IPv4 address or interface for the TCP response stream server.
-    /// IPv6 and unspecified addresses are rejected.
+    /// IP address or exact interface used to bind and advertise the TCP response stream server.
+    /// Unspecified addresses are rejected.
     /// If unset, the server auto-detects a routable local IP.
     pub const DYN_TCP_RESPONSE_STREAM_HOST: &str = "DYN_TCP_RESPONSE_STREAM_HOST";
 
@@ -807,9 +807,9 @@ pub mod event_plane {
     /// Event plane codec selection: "json" or "msgpack".
     pub const DYN_EVENT_PLANE_CODEC: &str = "DYN_EVENT_PLANE_CODEC";
 
-    /// IPv4 address or interface advertised by direct ZMQ event publishers.
-    /// IPv6 and unspecified addresses are rejected.
-    /// If unset, the runtime auto-detects a local IPv4 address.
+    /// IP address or exact interface advertised by direct ZMQ event publishers.
+    /// Unspecified addresses are rejected.
+    /// If unset, the runtime auto-detects a local IP address.
     pub const DYN_EVENT_PLANE_HOST: &str = "DYN_EVENT_PLANE_HOST";
 
     /// Bounded capacity of the direct ZMQ event-subscriber's merged event channel.
--- a/lib/runtime/src/config/environment_names.rs
+++ b/lib/runtime/src/config/environment_names.rs
@@ -745,7 +745,8 @@ pub mod tcp_response_stream {
     /// If unset or 0, the OS assigns a free ephemeral port.
     pub const DYN_TCP_RESPONSE_STREAM_PORT: &str = "DYN_TCP_RESPONSE_STREAM_PORT";
 
-    /// Host/interface for the TCP response stream server.
+    /// Advertised IPv4 address or interface for the TCP response stream server.
+    /// IPv6 and unspecified addresses are rejected.
     /// If unset, the server auto-detects a routable local IP.
     pub const DYN_TCP_RESPONSE_STREAM_HOST: &str = "DYN_TCP_RESPONSE_STREAM_HOST";
 
@@ -806,8 +807,9 @@ pub mod event_plane {
     /// Event plane codec selection: "json" or "msgpack".
     pub const DYN_EVENT_PLANE_CODEC: &str = "DYN_EVENT_PLANE_CODEC";
 
-    /// Host IP advertised by direct ZMQ event publishers.
-    /// If unset, the runtime auto-detects a local IP address.
+    /// IPv4 address or interface advertised by direct ZMQ event publishers.
+    /// IPv6 and unspecified addresses are rejected.
+    /// If unset, the runtime auto-detects a local IPv4 address.
     pub const DYN_EVENT_PLANE_HOST: &str = "DYN_EVENT_PLANE_HOST";
 
     /// Bounded capacity of the direct ZMQ event-subscriber's merged event channel.
--- a/lib/runtime/src/distributed.rs
+++ b/lib/runtime/src/distributed.rs
@@ -401,9 +401,10 @@ impl DistributedRuntime {
                     })?,
                     Err(_) => 0,
                 };
-                let interface = std::env::var(tcp_response_stream::DYN_TCP_RESPONSE_STREAM_HOST)
-                    .ok()
-                    .filter(|h| !h.is_empty());
+                let interface = crate::utils::ip_resolver::host_override_from_env(
+                    tcp_response_stream::DYN_TCP_RESPONSE_STREAM_HOST,
+                )
+                .map_err(|error| PipelineError::Generic(error.to_string()))?;
 
                 let host_suffix = interface
                     .as_ref()
--- a/lib/runtime/src/distributed.rs
+++ b/lib/runtime/src/distributed.rs
@@ -40,6 +40,48 @@ use tokio_util::sync::CancellationToken;
 type EndpointDiscoverySourceMap = HashMap<Endpoint, Weak<EndpointDiscoverySource>>;
 type RoutingOccupancyMap = HashMap<Endpoint, Weak<RoutingOccupancyState>>;
 
+fn parse_tcp_response_stream_port(value: Option<&str>) -> Result<u16, PipelineError> {
+    let Some(port) = value.map(str::trim).filter(|value| !value.is_empty()) else {
+        return Ok(0);
+    };
+
+    port.parse::<u16>().map_err(|_| {
+        PipelineError::Generic(format!(
+            "invalid {}: '{}' is not a valid port number",
+            tcp_response_stream::DYN_TCP_RESPONSE_STREAM_PORT,
+            port
+        ))
+    })
+}
+
+#[cfg(test)]
+mod unit_tests {
+    use super::parse_tcp_response_stream_port;
+    use crate::pipeline::PipelineError;
+
+    #[test]
+    fn response_stream_port_trims_and_treats_empty_as_unset() {
+        for value in [None, Some(""), Some(" \t ")] {
+            assert_eq!(parse_tcp_response_stream_port(value).unwrap(), 0);
+        }
+        assert_eq!(
+            parse_tcp_response_stream_port(Some(" 8080 ")).unwrap(),
+            8080
+        );
+    }
+
+    #[test]
+    fn response_stream_port_rejects_invalid_values() {
+        let error = parse_tcp_response_stream_port(Some(" 65536 ")).unwrap_err();
+        assert!(matches!(
+            error,
+            PipelineError::Generic(message)
+                if message
+                    == "invalid DYN_TCP_RESPONSE_STREAM_PORT: '65536' is not a valid port number"
+        ));
+    }
+}
+
 /// Distributed [Runtime] providing cluster-wide communication, transport, and discovery resources.
 ///
 /// `DistributedRuntime` is not a process singleton. Calling [`DistributedRuntime::new`] more than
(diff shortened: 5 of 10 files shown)


jthomson04 (commented):
Reviewed the transport host overrides. The change is clean and well-tested, and the `String` → `IpAddr`/`SocketAddr` refactor is a solid improvement (IPv6 bracketing stays consistent with the auto-detect path, and validation fails fast before any bind).

One thing I'd like to resolve before merge, plus a couple of optional notes — left inline:

1. **(please address)** `DYN_TCP_RESPONSE_STREAM_HOST` isn't preserved across snapshot restore, unlike its new sibling `DYN_EVENT_PLANE_HOST`. If a deployment pins the TCP host to a node-specific IP literal, that value freezes at snapshot time and the restored process will try to bind the old node's IP. See comment on `constants.py`.
2. **(optional)** API asymmetry: the event-plane host accepts IP literals only, while the TCP host also accepts interface names (IPv4-only resolution). See comment on `event_plane/mod.rs`.
3. **(optional)** The `DYN_TCP_RESPONSE_STREAM_HOST` interface-name behavior change isn't documented. See comment on `event-plane.md`.

Minor cosmetic (no inline): after the `IpAddr` change, `tracing::debug!("tcp transport service on {local_ip}:{local_port}")` in `tcp/server.rs` prints an unbracketed IPv6 (`…::1:8080`) — ambiguous in logs; consider logging the `SocketAddr`.

components/src/dynamo/common/snapshot/constants.py
```diff
@@ -30,6 +30,7 @@
     "DYN_DISCOVERY_BACKEND",
     "DYN_REQUEST_PLANE",
     "DYN_EVENT_PLANE",
+    "DYN_EVENT_PLANE_HOST",
```
jthomson04: `DYN_EVENT_PLANE_HOST` is added here, but its sibling `DYN_TCP_RESPONSE_STREAM_HOST` — extended by this same PR for the same IPoIB use case — isn't in `RESTORE_RUNTIME_ENV_NAMES`.

Both host vars are read by the Rust runtime during `DistributedRuntime` init *after* restore (the event publisher, and `TcpStreamServer` in `distributed.rs`), which is exactly why the event-plane host needed adding here. `write_snapshot_restore_context` only captures names in `_SUPPORTED_RESTORE_ENV_NAMES`, and `_apply_restore_env` drops anything outside it — so a node-specific IP literal in `DYN_TCP_RESPONSE_STREAM_HOST` is frozen at snapshot time and won't refresh on restore to a new node, causing a bind failure.

Suggest adding `DYN_TCP_RESPONSE_STREAM_HOST` (and arguably `DYN_TCP_RESPONSE_STREAM_PORT`) to the set, or explicitly documenting that TCP host pinning must use an interface name (node-portable) to survive restore.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -38,13 +39,35 @@ use std::task::{Context, Poll};
...
+fn event_plane_host_from_env() -> Result<String> {
+    let Ok(host) = std::env::var(DYN_EVENT_PLANE_HOST) else {
+        return Ok(local_ip_for_advertise());
+    };
+
+    if host.is_empty() {
+        return Ok(local_ip_for_advertise());
+    }
+
+    let ip = host.parse::<IpAddr>().map_err(|_| {
```
jthomson04: Optional / usability: this only accepts an IP literal, whereas `DYN_TCP_RESPONSE_STREAM_HOST` now also accepts an interface name (via `resolve_configured_host`). The PR summary frames both as "explicitly selected interfaces," so an operator configuring both for the same IPoIB NIC has to give an interface name to one and an IP to the other. Consider reusing `resolve_configured_host` here so the event-plane host can also take an interface name — with the bonus that interface names are node-portable across snapshot restore. Fine to defer, but worth noting the asymmetry.

docs/design-docs/event-plane.md
```diff
@@ -54,6 +54,7 @@ python3 -m dynamo.vllm --event-plane zmq --model Qwen/Qwen3-0.6B
 | Variable | Description | Default |
 |----------|-------------|---------|
 | `DYN_EVENT_PLANE` | Transport: `nats` or `zmq` | `zmq` (all backends) |
+| `DYN_EVENT_PLANE_HOST` | IP advertised by direct ZMQ publishers; ignored by NATS and ZMQ broker mode | Auto-detected local IP |
```
jthomson04: The docs cover `DYN_EVENT_PLANE_HOST` well, but the parallel change to `DYN_TCP_RESPONSE_STREAM_HOST` (now accepts an interface name in addition to an IP literal, IPv4-only resolution) isn't documented anywhere. Worth a short note so operators know the two knobs differ in what they accept.

jthomson04:
@coderabbitai full review

jthomson04 (commented):
## Code review

Clean, well-tested change. Comments below are mostly minor.

### Strengths
- **Good refactor** — moving `IpResolver`/`DefaultIpResolver` into `ip_resolver.rs` (server re-exports them) removes the prior `ip_resolver → server` import cycle and centralizes resolution. The default trait method for `list_afinet_netifas` keeps the existing `server.rs` test mock compiling untouched.
- **`SocketAddr` is a real correctness fix, not just cosmetic** — the old `format!("{}:{}", local_ip, port)` produced malformed `::1:port` for the IPv6 auto-detect fallback; `SocketAddr::to_string()` correctly brackets it (`[::1]:port`).
- **Deterministic interface selection** — the old path collapsed interfaces into a `HashMap<String, IpAddr>`, silently discarding all but one arbitrary address per name. Sorting + lowest-IPv4 is a genuine improvement.
- **Strong test coverage** — literal / interface / IPv6-reject / unspecified-reject / no-usable-IPv4 / not-found, plus valid+invalid publisher integration tests and the snapshot apply-and-clear update. Validation runs *before* the ZMQ bind, so an invalid host fails fast.

### Issues & suggestions

**1. Default event-plane path changed for IPv6-only hosts (low, likely intended).** The direct-ZMQ default moved from `local_ip_for_advertise()` (returns bracketed IPv6 when no IPv4 exists) to `local_ipv4_for_advertise()`, which falls back to `127.0.0.1`. Since direct ZMQ always binds the IPv4 wildcard (`tcp://0.0.0.0:0`), the old IPv6 advertisement was already unreachable — so this is arguably a fix, but worth a line in the PR description: an IPv6-only host now advertises loopback instead of a scoped-but-broken IPv6.

**2. "Lowest *usable* IPv4" only excludes `0.0.0.0` (low).** `resolve_advertised_ipv4` filters unspecified addresses but not loopback (`127.0.0.0/8`) or link-local (`169.254.0.0/16`). If an interface carries such an address alongside a routable one and it sorts lower (e.g. `169.254.x` < `172.16.x`), the non-routable address wins. Won't bite the documented IB/IPoIB case, but "lowest" ≠ "lowest routable" — consider filtering loopback/link-local, or noting the assumption in the docs.

**3. IP literals aren't verified against local interfaces (low).** Interface *names* are validated against `list_afinet_netifas()`; an IP *literal* is only checked for "IPv4 and not unspecified." A typo'd/foreign IP silently becomes the advertised address while the socket binds elsewhere. Matches the "explicit override = operator's responsibility" model, so defensible — just flagging the asymmetry.

**4. Snapshot constant grouping is slightly off (nit).** In `constants.py`, `DYN_EVENT_PLANE_HOST` sits under *"Parsed Python runtime config that must also refresh the in-memory config passed to create_runtime()"*, but it's read only on the Rust side at publisher creation. It fits better next to `DYN_TCP_RESPONSE_STREAM_HOST/PORT` under *"infrastructure env read after restore."* `RESTORE_RUNTIME_ENV_NAMES` is a set, so this is comment accuracy only, not behavior.

**5. `ServerOptions.interface` is now a misnomer (nit).** The field accepts an IP literal *or* an interface name; the name `interface` (and the `Interface not found` error text) under-describes it. Rename is out of scope, but a doc comment noting it accepts either would help future readers.

**6. Observation (no action):** the two transports intentionally differ — the TCP response-stream server *binds* to the resolved IP, while direct ZMQ binds the wildcard and only changes the *advertised* address. Both preserve prior per-transport behavior, so correct; just worth being explicit that `DYN_TCP_RESPONSE_STREAM_HOST` restricts the listen address whereas `DYN_EVENT_PLANE_HOST` does not.

### Verdict
Approve with minor nits. The only ones I'd act on before merge are **#1** (one line in the PR description) and **#4** (move the constant under the right comment); the rest are optional polish.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,6 +45,52 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+        if interface_found {
+            anyhow::anyhow!("Interface has no usable IPv4 address: {host_or_interface}")
+        } else {
+            anyhow::anyhow!("Interface not found: {host_or_interface}")
+        }
+    })
+}
+
+/// Resolve the IPv4 address used by IPv4-only advertised transports.
+pub(crate) fn local_ipv4_for_advertise<R: IpResolver>(resolver: &R) -> Ipv4Addr {
```
jthomson04: **Correctness — IPv6 auto-detect silently collapses to `127.0.0.1`.**

This drops the IPv6 fallback that `local_ip_for_advertise()` provided, and logs nothing.

On an IPv6-only or v6-preferred host, `local_ip()` returns `IpAddr::V6(..)`. The old event-plane path fell back to `local_ipv6()` and produced a bracketed `[2001:db8::1]`, giving a working `tcp://[2001:db8::1]:port` public endpoint. This code matches only `Ok(IpAddr::V4(..))`, so every direct-ZMQ publisher registers `tcp://127.0.0.1:<port>` in discovery. Remote subscribers dial their own loopback, connect to nothing (or to an unrelated local publisher), and because the event plane is best-effort nothing errors — KV/routing events are silently lost.

At minimum this needs a `tracing::warn!` on the fallback, matching the equivalent `TcpStreamServer` path at `server.rs:186`.

lib/runtime/src/pipeline/network/tcp/server.rs
```diff
@@ -185,19 +169,11 @@ impl TcpStreamServer {
...
-                    .get(&interface)
-                    .ok_or(PipelineError::Generic(format!(
-                        "Interface not found: {}",
-                        interface
-                    )))?
-                    .to_string()
-            }
+        let local_ip = match options.interface.as_deref() {
+            Some(host) => IpAddr::V4(
+                resolve_advertised_ipv4(host, &resolver)
```
jthomson04: **Correctness — an explicit IPv6 host is now a hard `DistributedRuntime` startup failure, while auto-detect four lines below still permits IPv6.**

Set `DYN_TCP_RESPONSE_STREAM_HOST=eth0` on an IPv6-only host (or to an IPv6 literal). The old code did `list_afinet_netifas()?.into_iter().collect::<HashMap<_,_>>().get(&interface)`, yielding the interface's IPv6, bound via `TcpListener::bind("2001:db8::1:8080")` (std's `ToSocketAddrs` splits at the last `:` and `getaddrinfo` accepts the bare literal), and `TcpStream::connect` on the client resolved the same string — so it worked end to end.

This code returns `Interface has no usable IPv4 address: eth0` / `IPv6 addresses are not supported for advertised hosts`, which becomes `PipelineError::Generic` and fails `DistributedRuntime::tcp_server()`, taking the whole worker down. The inconsistency is the sharp part: leaving the variable *unset* on the same host still yields an IPv6 bind via the `None` branch below.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,6 +45,52 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+    let interfaces = resolver.list_afinet_netifas()?;
+    let interface_found = interfaces.iter().any(|(name, _)| name == host_or_interface);
+    let mut ipv4_addresses: Vec<_> = interfaces
+        .into_iter()
+        .filter_map(|(name, ip)| match ip {
+            IpAddr::V4(ip) if name == host_or_interface && !ip.is_unspecified() => Some(ip),
+            _ => None,
+        })
+        .collect();
+    ipv4_addresses.sort_unstable();
```
jthomson04: **Correctness — the lowest-IPv4 rule prefers a `169.254.0.0/16` link-local alias over the real routable address.**

An IPoIB NIC that failed DHCP once carries both `169.254.7.3` and the real `172.16.0.87`. `Ipv4Addr: Ord` compares octets, so `169 < 172` and `sort_unstable()` + `.next()` returns the zeroconf address. With `DYN_EVENT_PLANE_HOST=ib0` the publisher advertises an unroutable link-local address to every peer; with `DYN_TCP_RESPONSE_STREAM_HOST=ib0` the response-stream listener *binds* to it, so workers on other nodes can never call home.

This is squarely in the path this PR exists to serve (pinning to a NUMA-local IPoIB NIC). The filter above only excludes `is_unspecified()` — it should also exclude `is_link_local()` and `is_loopback()`, or prefer a globally routable address rather than the numerically smallest one.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -926,6 +944,194 @@ mod tests {
...
+                assert_ne!(port, 0);
+
+                drop(publisher);
+            },
+        )
+        .await;
+    }
+
+    #[tokio::test]
+    async fn direct_zmq_publisher_rejects_invalid_configured_host() {
```
jthomson04: **Test isolation — this test mutates process-global env while sibling tests run in parallel, and two of those siblings were not given the guard.**

`temp_env::async_with_vars` sets `DYN_EVENT_PLANE_HOST=::1` process-wide, but libtest runs the other tests in this module on parallel threads. Meanwhile `direct_zmq_endpoint_scopes_are_isolated` (mod.rs:1220) and `direct_zmq_publishers_in_one_endpoint_fan_into_one_subscriber` (mod.rs:1313) create direct-ZMQ publishers *without* clearing the variable — even though this PR added exactly that guard to two sibling tests (e.g. `same_topic_publishers_are_independent_across_recreation`, mod.rs:1395).

Reproduced deterministically:

```
DYN_EVENT_PLANE_HOST=::1 cargo test -p dynamo-runtime --lib \
  transports::event_plane::tests::direct_zmq_endpoint_scopes_are_isolated
```
```
panicked at: create endpoint A publisher: Invalid DYN_EVENT_PLANE_HOST value '::1':
IPv6 addresses are not supported for advertised hosts
```

The same command against the guarded test passes. Both tests at mod.rs:1220 and mod.rs:1313 need `(DYN_EVENT_PLANE_HOST, None::<&str>)` added, otherwise this is a flake on a clean machine, not just under an inherited env.

components/src/dynamo/common/snapshot/constants.py
```diff
@@ -30,7 +30,10 @@
...
     # DistributedRuntime infrastructure env read after restore.
+    "DYN_TCP_RESPONSE_STREAM_HOST",
+    "DYN_TCP_RESPONSE_STREAM_PORT",
```
jthomson04: **Correctness — this list gains the response-stream and event-plane vars but omits `DYN_TCP_RPC_HOST`, the request-plane advertised host, so it stays stale across CRIU restore.**

A deployment that sets `DYN_TCP_RPC_HOST` from the Downward API (`status.podIP`) or to a pinned NIC address hits the exact bug this PR fixes for the other two vars. `write_snapshot_restore_context()` only captures names in `_SUPPORTED_RESTORE_ENV_NAMES` and `_apply_restore_env()` only rewrites those names, so the restored process keeps the snapshot-time value. `lib/runtime/src/pipeline/network/manager.rs:95` and `lib/runtime/src/component/endpoint.rs:328` then advertise the *old* pod's IP for the request plane while the two newly-listed vars correctly advertise the new pod's — routers dial a dead address and every request to that worker fails.

`DYN_SYSTEM_HOST` is already in this list for the same reason; `DYN_TCP_RPC_HOST` belongs here too, and it carries the actual traffic.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -41,13 +41,31 @@ use std::task::{Context, Poll};
...
-use crate::utils::local_ip_for_advertise;
+use crate::utils::ip_resolver::{
+    DefaultIpResolver, IpResolver, local_ipv4_for_advertise, resolve_advertised_ipv4,
+};
+
+fn event_plane_host_from_env() -> Result<String> {
+    event_plane_host_from_env_with_resolver(&DefaultIpResolver)
+}
+
+fn event_plane_host_from_env_with_resolver<R: IpResolver>(resolver: &R) -> Result<String> {
```
jthomson04: **Observability — a wrong-but-parseable value produces a healthy process advertising an unreachable endpoint, with no log line at all.**

Set `DYN_EVENT_PLANE_HOST=172.16.0.78` when the NUMA-local IPoIB NIC is actually `172.16.0.87`. `resolve_advertised_ipv4` returns the literal unchanged (the parse branch at `ip_resolver.rs:56` never consults the resolver), the PUB socket still binds `tcp://0.0.0.0:0`, discovery registers `tcp://172.16.0.78:<port>`, and every subscriber's ZMQ connect silently retries forever. Because the event plane is best-effort, nothing surfaces — routing quality just degrades.

The TCP path at least fails fast at `bind`. Here there is neither validation nor an `info`/`warn` of the resolved advertised host. Separately, a typo'd *interface* name is only caught when the first direct-ZMQ publisher is constructed, potentially after readiness has already been reported — resolving once during `DistributedRuntime` construction would make it fail fast (and would also fix the repeated `getifaddrs`, see the comment on line 55).

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,6 +45,52 @@ pub fn tcp_rpc_host_from_env() -> String {
...
 
+/// Resolve an explicit advertised host as a dialable IPv4 address.
+///
+/// The input can be an IPv4 literal or a network interface name. Interfaces
+/// with multiple IPv4 addresses select the lowest address deterministically.
+pub(crate) fn resolve_advertised_ipv4<R: IpResolver>(
+    host_or_interface: &str,
+    resolver: &R,
+) -> Result<Ipv4Addr> {
+    if let Ok(ip) = host_or_interface.parse::<IpAddr>() {
```
jthomson04: **Error handling — values are neither trimmed nor distinguished from non-interface strings, so whitespace and hostnames both surface as "Interface not found".**

A Kubernetes manifest yields `DYN_EVENT_PLANE_HOST=" 10.0.0.5"` (YAML block scalar or trailing space). `parse::<IpAddr>()` fails on the leading space, the value falls through to the interface branch, and the operator sees:

```
Invalid DYN_EVENT_PLANE_HOST value ' 10.0.0.5': Interface not found:  10.0.0.5
```

A variable named `_HOST` also invites a hostname — `DYN_TCP_RESPONSE_STREAM_HOST=worker-0.dynamo.svc.cluster.local` fails with `Interface not found: worker-0.dynamo.svc.cluster.local`, which never tells the user that hostnames are unsupported. `distributed.rs:404` only applies `.filter(|h| !h.is_empty())`, no `trim`.

Related: `event_plane/mod.rs:60` matches `Ok(host) if !host.is_empty()` with a catch-all `_ =>`, so `VarError::NotUnicode` silently auto-detects instead of erroring.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -3,12 +3,35 @@
...
 use std::net::{IpAddr, Ipv4Addr};
 
 const FALLBACK: IpAddr = IpAddr::V4(Ipv4Addr::LOCALHOST);
 
+/// IP address resolution interface used by networking components and tests.
+pub trait IpResolver {
+    fn local_ip(&self) -> Result<IpAddr, Error>;
+    fn local_ipv6(&self) -> Result<IpAddr, Error>;
+
+    fn list_afinet_netifas(&self) -> Result<Vec<(String, IpAddr)>, Error> {
```
jthomson04: **Testability — this default body calls the real global `list_afinet_netifas()`, defeating the trait's dependency-injection purpose for any implementor that doesn't override it.**

`FailingIpResolver` (`server.rs:1409`) implements only `local_ip`/`local_ipv6`. Any future test that exercises the interface path — `TcpStreamServer::new_with_resolver(ServerOptions { interface: Some("eth0"), .. }, FailingIpResolver)` — would silently read the CI host's real NIC table instead of a controlled fixture: passes on a laptop with an `eth0`, fails in a container without one.

The trait exists precisely so IP resolution can be injected (the doc comment says "used by networking components and tests"). The default body should live on `DefaultIpResolver` only, forcing every mock to declare its interface list, as `MockIpResolver` and `EventPlaneHostResolver` already do.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -459,14 +478,13 @@ impl EventPublisher {
...
-                    // Get local IP for public endpoint
+                    // Use the configured or auto-detected local IP for the public endpoint.
                     let actual_port: u16 = actual_bind_endpoint
                         .rsplit(':')
                         .next()
                         .and_then(|s| s.parse().ok())
                         .expect("Failed to parse port from bind endpoint");
-                    let local_ip = local_ip_for_advertise();
-                    let public_endpoint = format!("tcp://{}:{}", local_ip, actual_port);
+                    let public_endpoint = format!("tcp://{}:{}", advertised_host, actual_port);
```
jthomson04: **Altitude — the IPv6 ban exists to work around this `format!`, and this PR already made the correct fix elsewhere.**

`server.rs:456` was changed from `format!("{}:{}", ip, port)` to `SocketAddr::new(ip, port).to_string()` precisely because a bare `{host}:{port}` cannot express IPv6. This line keeps the string form and compensates by rejecting IPv6 upstream in `resolve_advertised_ipv4`. Building `tcp://{SocketAddr}` here would support both families with no ban at all — and would remove the need for `local_ipv4_for_advertise` and its loopback collapse.

Stepping back, the PR leaves three parallel mechanisms with three different contracts:

| var | semantics | IPv6 | interface names | registered in `environment_names.rs` |
|---|---|---|---|---|
| `DYN_TCP_RPC_HOST` | advertise | bracketed, allowed | no | no (undocumented) |
| `DYN_TCP_RESPONSE_STREAM_HOST` | bind **and** advertise | rejected when explicit | yes | yes |
| `DYN_EVENT_PLANE_HOST` | advertise only | rejected | yes | yes |

A single `advertised_host_from_env(var, resolver) -> IpAddr` plus a shared `SocketAddr` endpoint formatter would cover all three — and would make `DYN_TCP_RPC_HOST` accept interface names too, which this PR's own goal (pin every transport to a NUMA-local IPoIB NIC) requires, since the request plane carries the actual traffic.

lib/runtime/src/pipeline/network/tcp/server.rs
```diff
@@ -477,7 +453,7 @@ impl ResponseService for TcpStreamServer {
     async fn register(&self, options: StreamOptions) -> PendingConnections {
         // oneshot channels to pass back the sender and receiver objects
 
-        let address = format!("{}:{}", self.local_ip, self.local_port);
+        let address = SocketAddr::new(self.local_ip, self.local_port).to_string();
```
jthomson04: **Efficiency — `register()` is on the per-request egress path, and this rebuilds an immutable value on every call.**

`egress/addressed_router.rs:773` calls `self.resp_transport.register(options)` once per outbound request. Each call formats `SocketAddr::new(self.local_ip, self.local_port)` into a fresh `String` and then clones it once or twice into `TcpStreamConnectionInfo` (server.rs:474 and server.rs:520) — three allocations per request for a value fixed at construction.

Since the diff already touches this line: store the rendered `address: Arc<str>` alongside `local_ip`/`local_port` in `new_with_resolver` and clone the `Arc` here.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -41,13 +41,31 @@ use std::task::{Context, Poll};
...
     MAX_JSON_SAFE_PUBLISHER_ID,
 };
 use crate::protocols::EndpointId;
 use crate::traits::DistributedRuntimeProvider;
-use crate::utils::local_ip_for_advertise;
+use crate::utils::ip_resolver::{
+    DefaultIpResolver, IpResolver, local_ipv4_for_advertise, resolve_advertised_ipv4,
+};
+
+fn event_plane_host_from_env() -> Result<String> {
```
jthomson04: **Efficiency — this re-reads the environment and re-enumerates every network interface on each `EventPublisher` construction.**

With `DYN_EVENT_PLANE_HOST=ib0`, every direct-mode publisher in the process calls `std::env::var` plus `resolver.list_afinet_netifas()`, which walks `getifaddrs` and allocates a `Vec<(String, IpAddr)>` of the whole NIC table. Under replica-sync a frontend creates a publisher per endpoint/topic, so startup pays that syscall and allocation N times for a value that cannot change.

Memoize in a `static OnceLock<Result<Ipv4Addr, String>>`, or resolve once during `DistributedRuntime` construction — the latter also makes a bad value fail fast (see the comment on line 59).

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,6 +45,52 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+            IpAddr::V4(ip) if ip.is_unspecified() => {
+                bail!("unspecified IPv4 addresses cannot be advertised")
+            }
+            IpAddr::V4(ip) => Ok(ip),
+            IpAddr::V6(_) => bail!("IPv6 addresses are not supported for advertised hosts"),
+        };
+    }
+
+    let interfaces = resolver.list_afinet_netifas()?;
+    let interface_found = interfaces.iter().any(|(name, _)| name == host_or_interface);
```
jthomson04: **Simplification — two full passes over the interface list plus a `Vec` sort, only to take the first element.**

`interfaces.iter().any(..)` on the line above scans the whole list to compute `interface_found` even on the happy path, then this scans it again and `sort_unstable()`s a `Vec` whose only consumer is `.into_iter().next()`.

`ipv4_addresses.into_iter().min()` gives the same result in one pass with no sort, and `interface_found` can be folded into the single filter pass, removing the intermediate `Vec` entirely. (If the link-local finding on line 75 is addressed, the selection predicate changes anyway — worth doing both at once.)

lib/runtime/src/config/environment_names.rs
```diff
@@ -745,7 +745,8 @@ pub mod tcp_response_stream {
     /// If unset or 0, the OS assigns a free ephemeral port.
     pub const DYN_TCP_RESPONSE_STREAM_PORT: &str = "DYN_TCP_RESPONSE_STREAM_PORT";
 
-    /// Host/interface for the TCP response stream server.
+    /// Advertised IPv4 address or interface for the TCP response stream server.
```
jthomson04: **Docs — this says "Advertised", but the value is also the listener's *bind* address, and it now reads almost identically to `DYN_EVENT_PLANE_HOST`, which really is advertise-only.**

The two constants sit in the same file with near-identical wording ("Advertised IPv4 address or interface ... IPv6 and unspecified addresses are rejected"), yet `DYN_EVENT_PLANE_HOST` leaves the PUB socket on the `tcp://0.0.0.0:0` wildcard while this one is passed straight to `TcpListener::bind` (`server.rs:371`).

A user who learns the advertise-only semantics from the event plane and sets this to a NAT/LB/VIP address gets `Failed to start TcpStreamServer: ... Cannot assign requested address` at DRT init. The docs page added by this PR says "used to bind and advertise" — this doc comment should say the same, and the two constants should be worded so the difference is obvious.

components/src/dynamo/common/snapshot/constants.py
```diff
@@ -30,7 +30,10 @@
     "DYN_DISCOVERY_BACKEND",
     "DYN_REQUEST_PLANE",
     "DYN_EVENT_PLANE",
+    "DYN_EVENT_PLANE_HOST",
```
jthomson04: **Docs — this is filed under the "Parsed Python runtime config" comment, but it has no parsed config field; it belongs in the "DistributedRuntime infrastructure env" group two lines down.**

That first comment block documents a contract: those names are copied into the reparsed config by `_copy_restore_runtime_config` via `_RESTORE_RUNTIME_CONFIG_FIELDS = (namespace, discovery_backend, request_plane, event_plane)`. `DYN_EVENT_PLANE_HOST` has no entry there and no argument in `DynamoRuntimeArgGroup` — it is read only by Rust through `std::env::var`, exactly like `NATS_SERVER` and the two `DYN_TCP_RESPONSE_STREAM_*` names this same diff placed under the second comment.

As written, a maintainer adding an env var next to this one will believe they also need a parsed field and a `_RESTORE_RUNTIME_CONFIG_FIELDS` entry, or will assume one already exists for this one.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -41,13 +42,44 @@ use std::task::{Context, Poll};
...
+fn event_plane_host_from_env() -> Result<IpAddr> {
+    event_plane_host_from_env_with_resolver(&DefaultIpResolver)
+}
+
+fn event_plane_host_from_env_with_resolver<R: IpResolver>(resolver: &R) -> Result<IpAddr> {
+    let Some(host) = host_override_from_env(DYN_EVENT_PLANE_HOST)? else {
+        return Ok(resolve_local_host(resolver));
+    };
+
+    resolve_host_or_interface(&host, resolver)
```
jthomson04: **[1] Correctness — an IPv6 `DYN_EVENT_PLANE_HOST` advertises an endpoint no subscriber can reach.**

`resolve_host_or_interface` accepts both families here, and `direct_zmq_public_endpoint` faithfully emits `tcp://[2001:db8::10]:PORT` into discovery. But the publisher binds `tcp://0.0.0.0:0` (line 486) and no ZMQ socket in this crate sets `ZMQ_IPV6`, which libzmq defaults to `0`.

The new doc text says "Direct ZMQ continues to bind an IPv4 wildcard, so an IPv6 value does not enable IPv6 listening" — that describes the mechanism but understates the consequence. It isn't that IPv6 listening stays off; it's that the advertised endpoint is *dead*. `zmq_connect` to a bracketed IPv6 endpoint fails `EINVAL` with `ZMQ_IPV6=0`, and `DynamicSubscriber::consume_endpoint_stream` handles that with a single `tracing::warn!` and a drop from `active_endpoints` — no retry. KV events silently stop flowing from that worker and routing quietly degrades.

The same is reachable without an explicit override: `resolve_local_host` falls back to `local_ipv6()` when no IPv4 is found.

Either reject IPv6 here with a message naming the wildcard-bind reason, or bind `tcp://[::]:0` with `ZMQ_IPV6=1` when the advertised host is v6. `direct_zmq_public_endpoint_formats_ipv4_and_ipv6` and `direct_zmq_advertise_host_preserves_ipv6_fallback_and_rejects_wildcards` currently assert the broken configuration is correct.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,14 +45,65 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+) -> Result<IpAddr> {
+    let host_or_interface = host_or_interface.trim();
+    if let Ok(ip) = host_or_interface.parse::<IpAddr>() {
+        if ip.is_unspecified() {
+            bail!("unspecified IP addresses cannot be advertised");
+        }
+        return Ok(ip);
+    }
+
+    let interfaces: HashMap<String, IpAddr> = resolver.list_afinet_netifas()?.into_iter().collect();
```
jthomson04: **[2] Correctness — `HashMap` last-wins silently prefers IPv6, and this value now *binds*.**

The doc comment and `interface_lookup_preserves_existing_last_address_behavior` both frame "last enumerated address wins" as preserved behavior, which is accurate as far as it goes. What neither addresses: on a stock dual-stack Linux host `list_afinet_netifas()` returns `("eth0", 10.0.0.5)` *then* `("eth0", fe80::...)` / `("eth0", 2001:db8::5)`, so last-wins means IPv6-wins for essentially every real NIC. And `IpAddr` has no scope id, so a link-local `fe80::` arrives here stripped of its `%eth0`.

Two consequences. `DYN_EVENT_PLANE_HOST=eth0` advertises the v6 address, which is dead per [1]. `DYN_TCP_RESPONSE_STREAM_HOST=eth0` is worse, because `TcpStreamServer::new_with_resolver` passes the result straight to `TcpListener::bind` — a scopeless `fe80::` either fails `EINVAL` or binds an address IPv4 clients cannot dial.

This path was previously near-dead: the old `format!("{}:{}", ip, port)` produced a malformed string for any v6 address and failed loudly. The `SocketAddr` fix in this PR is correct in isolation, but it makes the nondeterministic interface selection load-bearing for the first time. Worth an explicit rule — prefer IPv4, or prefer the first global-scope address — rather than inheriting `getifaddrs` ordering.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,14 +45,65 @@ pub fn tcp_rpc_host_from_env() -> String {
     std::env::var("DYN_TCP_RPC_HOST").unwrap_or_else(|_| local_ip_for_advertise())
```
jthomson04: **[3] Correctness — `DYN_TCP_RPC_HOST` is promoted to a restore knob but skips the entire new host contract.**

This is the one variable in the family still on raw `std::env::var`: no trim, empty is not treated as unset, no IPv6 bracketing, and — most importantly — no unspecified-address guard, on a value that `endpoint.rs:337` publishes into discovery.

Three concrete breaks:

1. A ConfigMap or YAML block scalar yields `DYN_TCP_RPC_HOST=" 10.0.0.5 "`. `manager.rs:270` does `format!("{}:{}", host, port).parse::<SocketAddr>()` → `Invalid TCP bind address`, process refuses to start — while the sibling `DYN_TCP_RESPONSE_STREAM_HOST` tolerates the identical value.
2. `_apply_restore_env` writes empty strings verbatim, so a restore context carrying `DYN_TCP_RPC_HOST=""` yields `":8080"` → parse failure at restore.
3. `DYN_TCP_RPC_HOST=0.0.0.0` is accepted, binds the wildcard as intended, and then advertises `0.0.0.0:8080/<id>/<name>` into discovery — the exact unspecified-advertisement this PR's new guard was written to prevent, on the variable that most needs it.

It also has no constant in `config/environment_names.rs` and no entry in the new docs page, unlike every other host in the family. Routing it through `host_override_from_env` + `resolve_host_or_interface` closes all three.

components/src/dynamo/common/snapshot/constants.py
```diff
@@ -31,6 +31,10 @@
     "DYN_REQUEST_PLANE",
     "DYN_EVENT_PLANE",
     # DistributedRuntime infrastructure env read after restore.
+    "DYN_EVENT_PLANE_HOST",
+    "DYN_TCP_RPC_HOST",
```
jthomson04: **[4] Correctness — `DYN_TCP_RPC_PORT` is missing while its HOST sibling is added; a restored worker binds the snapshotted pod's port.**

The response-stream family got both HOST and PORT (lines 36–37) and `DYN_SYSTEM_PORT` is already here, so this reads as an oversight rather than a decision.

Snapshot in pod A with `DYN_TCP_RPC_PORT=25000`, restore into pod B whose manifest sets `25001`. `write_snapshot_restore_context` only captures names in `_SUPPORTED_RESTORE_ENV_NAMES`, so the variable is never written *or cleared*, and the CRIU'd process keeps `25000` in its own environ. Both `NetworkConfig::from_env` (`manager.rs:96`) and `build_transport_type_inner` (`endpoint.rs:331`) read it after `refresh_snapshot_restore_config` runs (`vllm/main.py:211` → `create_runtime` at `:226`), so the restored worker binds and advertises `25000` while pod B's Service and NetworkPolicy expect `25001`. Routers get connection refused against a worker that looks healthy.

lib/runtime/src/distributed.rs
```diff
@@ -401,9 +401,10 @@ impl DistributedRuntime {
                     })?,
                     Err(_) => 0,
                 };
-                let interface = std::env::var(tcp_response_stream::DYN_TCP_RESPONSE_STREAM_HOST)
-                    .ok()
-                    .filter(|h| !h.is_empty());
+                let interface = crate::utils::ip_resolver::host_override_from_env(
```
jthomson04: **[5] Correctness — the port two lines above the host gets none of the host's new leniency, and fails lazily.**

`DYN_TCP_RESPONSE_STREAM_PORT` was added to `RESTORE_RUNTIME_ENV_NAMES` in this PR, which means it now flows through `_apply_restore_env` — which writes empty strings verbatim. A DGD that renders `env: [{name: DYN_TCP_RESPONSE_STREAM_PORT, value: ""}]` (a routine Helm outcome for an optional value) produces `Ok("")` here, `"".parse::<u16>()` fails, and `tcp_server()` returns `invalid DYN_TCP_RESPONSE_STREAM_PORT: '' is not a valid port number`. Same for `" 8080 "`. The host directly below treats both as unset.

Worse than the asymmetry: `tcp_server` is a `OnceCell`, so this doesn't surface at startup. The process boots, registers, serves — and dies on the first request that needs a response stream. Empty-as-unset (→ `0`, OS-assigned) plus a trim would match the contract the next statement already implements.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -441,6 +473,7 @@ impl EventPublisher {
                     )
                 } else {
                     // DIRECT MODE: Bind PUB socket
+                    let advertised_host = event_plane_host_from_env()?;
```
jthomson04: **[6] Robustness — validation is per-publisher and lazy, so a typo fails mid-traffic and re-syscalls forever.**

`DYN_EVENT_PLANE_HOST=eth99` starts cleanly, registers the model, and serves requests. The first component that later constructs an `EventPublisher` in direct mode — a KV-event publisher during model registration, a replica-sync publisher on first peer discovery — hits `event_plane_host_from_env()?` and fails with `Invalid DYN_EVENT_PLANE_HOST value 'eth99'`, taking down whatever operation happened to trigger it rather than failing the deployment at boot.

The same call also re-reads the environment, runs a blocking `getifaddrs` syscall, and builds a full `HashMap` of every address on every NIC — on an async runtime thread, once per publisher. A `OnceLock<IpAddr>` resolved during `DistributedRuntime::new` gives you fail-fast at boot and one syscall for the process lifetime. The TCP host has the same shape inside its `OnceCell` (see [5]).

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,14 +45,65 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+///
+/// Interface lookup preserves the existing TCP response-stream behavior: when
+/// an interface name occurs more than once, the last enumerated address wins.
+pub(crate) fn resolve_host_or_interface<R: IpResolver>(
+    host_or_interface: &str,
+    resolver: &R,
+) -> Result<IpAddr> {
+    let host_or_interface = host_or_interface.trim();
+    if let Ok(ip) = host_or_interface.parse::<IpAddr>() {
+        if ip.is_unspecified() {
```
jthomson04: **[7] Correctness — the wildcard guard misses the IPv4-mapped wildcard `::ffff:0.0.0.0`.**

`Ipv6Addr::is_unspecified()` is `true` only for `::`, so `::ffff:0.0.0.0` parses cleanly, passes this check, and reaches `direct_zmq_public_endpoint`, which emits `tcp://[::ffff:0.0.0.0]:PORT` into discovery — precisely the wildcard advertisement this guard, the docs ("Unspecified addresses such as `0.0.0.0` and `::` are invalid"), and `explicit_host_rejects_unspecified_addresses` all claim is impossible.

`ip.to_canonical().is_unspecified()` closes it; the test deserves the mapped form as a third case.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -3,11 +3,34 @@
...
 
-const FALLBACK: IpAddr = IpAddr::V4(Ipv4Addr::LOCALHOST);
+const FALLBACK: IpAddr = IpAddr::V4(std::net::Ipv4Addr::LOCALHOST);
+
+/// IP address resolution interface used by networking components and tests.
+pub trait IpResolver {
+    fn local_ip(&self) -> Result<IpAddr, Error>;
+    fn local_ipv6(&self) -> Result<IpAddr, Error>;
+
+    fn list_afinet_netifas(&self) -> Result<Vec<(String, IpAddr)>, Error> {
```
jthomson04: **[8] Testability — this default body calls the real OS enumerator, defeating the seam the trait exists for.**

The trait's own doc says it is the resolution interface "used by networking components and tests." A default that calls the global `list_afinet_netifas()` means any implementor that forgets to override it silently reads the host's real NICs. That is already live: `FailingIpResolver` (`tcp/server.rs:1408`), which backs `test_server()` and every test in that module, does not override it — so any test reaching the interface-name branch depends on CI machine networking rather than a fixture.

`EventPlaneHostResolver` and `MockIpResolver` both implement it already. Making it required costs two lines in `DefaultIpResolver` and removes the trap for every future implementor.

lib/runtime/src/pipeline/network/tcp/server.rs
```diff
@@ -185,19 +169,12 @@ impl TcpStreamServer {
...
-                    .to_string()
-            }
+        let local_ip = match options.interface.as_deref() {
+            Some(host) => resolve_host_or_interface(host, &resolver).map_err(|error| {
+                PipelineError::Generic(format!(
+                    "Failed to resolve configured TCP host '{host}': {error}"
+                ))
+            })?,
             None => {
                 let resolved_ip = resolver.local_ip().or_else(|err| match err {
```
jthomson04: **[9] Correctness — the extracted helper and this surviving copy implement opposite error policies.**

This PR extracted `resolve_local_host` (`ip_resolver.rs:95`) but left this chain inlined, and the two now disagree on the question they both answer. The comment right below is explicit: fall back to loopback *only* on `LocalIpAddressNotFound`, propagate I/O and platform errors "so misconfigured hosts fail fast instead of silently binding to 127.0.0.1."

`resolve_local_host` ends in `.unwrap_or(FALLBACK)` — it swallows every error, including the ones this code deliberately propagates, and logs nothing at all (this path at least emits a `tracing::warn!`). The event plane now routes through it, so a transient `getifaddrs` I/O error makes a worker advertise `tcp://127.0.0.1:PORT` to the entire cluster, silently. That's the failure this code was hardened against, reintroduced next door.

Pick one policy and have both callers use it. Given the event-plane value goes into discovery, the fail-fast one seems right.

docs/fern/pages/reference/components/runtime-configuration.mdx
```diff
@@ -131,6 +131,18 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
   <span className="enum-values"><span className="enum-label">Allowed values:</span> <Badge intent="note" minimal>json</Badge> <Badge intent="note" minimal>msgpack</Badge></span>
 </ParamField>
 
+<ParamField path="DYN_TCP_RESPONSE_STREAM_HOST" type="string" default="auto-detected local address">
+  IPv4 or IPv6 address, or exact network interface name, used to bind and advertise the TCP response
```
jthomson04: **[10] Design — this knob both binds and advertises, and the new validation rejects the value a bind address most needs.**

"used to bind and advertise" is an honest description of the implementation, and that's the problem: a multi-NIC or NAT'd deployment wants to listen on all interfaces and advertise one routable address. Neither half is expressible here. `0.0.0.0` is rejected outright by `resolve_host_or_interface`, and any accepted value goes verbatim to `TcpListener::bind`, so advertising a NAT or load-balancer address that isn't local to the pod fails `EADDRNOTAVAIL`.

`DYN_EVENT_PLANE_HOST` got exactly the right split in this same PR — wildcard bind, override advertise only, documented at lines 161–163. The response-stream host kept one overloaded knob, and inherited a validation rule written for advertise semantics that now contradicts its own documented bind purpose. Worth either splitting it the same way or documenting the restriction as deliberate.

lib/runtime/src/pipeline/network/tcp/server.rs
```diff
@@ -1596,6 +1573,44 @@ mod tests {
         assert!(socket_addr.port() > 0, "Server should have a valid port");
     }
 
+    #[tokio::test]
+    async fn configured_ip_literals_bind_and_format_addresses() {
+        for (host, expected_ip) in [
+            ("127.0.0.1", "127.0.0.1".parse::<IpAddr>().unwrap()),
+            ("::1", "::1".parse::<IpAddr>().unwrap()),
```
jthomson04: **[11] Test reliability — this leg binds `::1` unconditionally and unwraps.**

The second iteration reaches `TcpListener::bind([::1]:0)` and `.unwrap()`s at line 1590. On a runner with `net.ipv6.conf.all.disable_ipv6=1`, a Docker network without IPv6, or a kernel without `CONFIG_IPV6`, that returns `EAFNOSUPPORT`/`EADDRNOTAVAIL`, `tcp_listener` sends `Err`, and the test panics — a red build with nothing to do with the change.

A probe bind on `::1` that `return`s early (or `eprintln!` + skip) keeps the bracketing assertion on hosts that have IPv6 and costs nothing on hosts that don't.

lib/runtime/src/utils/ip_resolver.rs
```diff
@@ -22,14 +45,65 @@ pub fn tcp_rpc_host_from_env() -> String {
...
+        return Ok(ip);
+    }
+
+    let interfaces: HashMap<String, IpAddr> = resolver.list_afinet_netifas()?.into_iter().collect();
+    let ip = interfaces.get(host_or_interface).copied().ok_or_else(|| {
+        anyhow::anyhow!(
+            "'{host_or_interface}' is not a valid IP address and no network interface with that name was found"
+        )
+    })?;
+    if ip.is_unspecified() {
```
jthomson04: **[12] Simplification — second copy of the unspecified check, plus a whole-interface-table `HashMap` to read one key.**

This block is identical to lines 77–79. It means the `::ffff:0.0.0.0` hole in [7] has to be fixed twice, and a third resolution branch will want a third copy. Checking once on the resolved `ip` immediately before the final `Ok` covers every branch — including future ones — for free.

Two smaller things in the same function: line 75 trims a value that `host_override_from_env` already trimmed at line 63 (harmless, but it implies the caller can't be trusted, which invites the next reader to add a third trim). And line 83 allocates a `HashMap` over every address on every NIC to look up one name — on every direct-ZMQ publisher construction, per [6]. `.into_iter().filter(|(n, _)| n == host_or_interface).map(|(_, ip)| ip).last()` is allocation-free and states the documented last-wins rule outright instead of leaving it to `HashMap` insert-overwrite semantics.

lib/runtime/src/distributed.rs
```diff
@@ -401,9 +401,10 @@ impl DistributedRuntime {
...
+                    tcp_response_stream::DYN_TCP_RESPONSE_STREAM_HOST,
+                )
+                .map_err(|error| PipelineError::Generic(error.to_string()))?;
 
                 let host_suffix = interface
```
jthomson04: **[13] Naming — three names for one value, and they now disagree about what it is.**

`ServerOptions.interface` widened in this PR from interface-name-only to IP-literal-or-interface-name, but nothing downstream was renamed. So in these six lines the value is bound as `interface` (line 404), formatted into a log that calls it a *host* (line 411), and passed to a field called `interface` whose resolver error says "Failed to resolve configured TCP host" (`server.rs:175`).

A reader tracing `DYN_TCP_RESPONSE_STREAM_HOST=10.0.0.5` through those three names can't tell from the types whether an IP literal is even legal. `ServerOptions` is `pub` with a derived builder and `Serialize`/`Deserialize`, so renaming the field is breaking and probably not worth it here — but the field needs a doc comment stating the accepted forms, and this local should be `host`.

lib/runtime/src/pipeline/network/tcp/server.rs
```diff
@@ -47,26 +47,10 @@ use crate::pipeline::{
...
-impl IpResolver for DefaultIpResolver {
-    fn local_ip(&self) -> Result<std::net::IpAddr, Error> {
-        local_ip()
-    }
-
-    fn local_ipv6(&self) -> Result<std::net::IpAddr, Error> {
-        local_ipv6()
-    }
-}
+pub use crate::utils::ip_resolver::{DefaultIpResolver, IpResolver};
```
jthomson04: **[14] Simplification — re-export with no remaining consumer.**

After the move, every in-repo reference imports from `utils::ip_resolver` (`event_plane/mod.rs:52-55`, and this file's line 50 for `resolve_host_or_interface`); only this module's own body and its test module use the bare names. So this leaves `dynamo_runtime::pipeline::network::tcp::server::IpResolver` as a permanently supported alias for an audience that doesn't exist — and it's the path a future reader hits first, pointing away from the canonical definition where the DI-breaking default from [8] actually lives. Importing the two names alongside line 50 and dropping the `pub use` is the same line count.

PeaBrane (approved):

galletas1712 (approved):

PR #11873, opened 2026-07-18
feat(vllm): add mock gRPC server for sidecar tests

connorcarpenter15:
## Overview:

Adds a CPU-only Mocker-backed implementation of vLLM's native Generate gRPC API so the real Dynamo vLLM sidecar can be tested without vLLM, a model, or GPUs.

## Details:

- Adds `dynamo-vllm-mocker-server` under `lib/mocker/servers/vllm`.
- Uses `dynamo_mocker::LiveEngine` for request submission, deterministic streaming, capacity errors, metrics, and cancellation cleanup.
- Supports aggregated, prefill, and decode wire flows. Disaggregated mode validates and round-trips handoff metadata without moving KV data.
- Emits deterministic token IDs, synthetic logprobs, usage, and length termination.
- Temporarily imports the sidecar-owned vLLM 0.25.1 generated contract through `dynamo_vllm_sidecar::proto`, avoiding a duplicate proto or standalone shared crate until vLLM publishes its upstream package.
- Documents local usage and validates the real sidecar against the CPU-only server.

Validation:

- `cargo fmt --all --check`
- `cargo metadata --no-deps --format-version 1`
- `cargo test -p dynamo-mocker -p dynamo-vllm-mocker -p dynamo-vllm-sidecar`
- `cargo clippy -p dynamo-mocker -p dynamo-vllm-mocker -p dynamo-vllm-sidecar --all-targets -- -D warnings`
- `cargo run -q -p dynamo-vllm-mocker --bin dynamo-vllm-mocker-server -- --help`

Start with `lib/mocker/servers/vllm/src/server.rs` for the RPC lifecycle and `lib/mocker/servers/vllm/src/server_request.rs` for request validation and wire conversion. The real sidecar boundary is covered by `lib/mocker/servers/vllm/tests/sidecar.rs`.

## Related Issues

- Closes [DIS-2474](https://linear.app/nvidia/issue/DIS-2474/add-a-vllm-compatible-mocker-grpc-server)
- Relates to #11917


jthomson04 (commented):
Four non-blocking review comments.

lib/mocker/servers/vllm/tests/sidecar.rs
```diff
@@ -0,0 +1,229 @@
...
+        .expect("prefill response should carry an opaque KV handoff");
+    assert_eq!(handoff["do_remote_prefill"], true);
+    assert!(handoff["remote_engine_id"].is_string());
+
+    let mut decode_request = request(3);
+    decode_request.prefill_result = Some(PrefillResult {
+        disaggregated_params: handoff,
+        prompt_tokens_details: None,
+    });
+    let decode_outputs = collect(&decode, decode_request).await;
```
jthomson04, line 189: [P3] Could this test pass a sentinel field that the decode mock explicitly checks? The payload is intentionally opaque, but the current decode validation only reads known rendezvous fields, so this test would still pass if the sidecar dropped unknown fields.

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+            .min(MAX_CANDIDATES))
+            .map(|offset| selected.wrapping_add(offset as u32))
+            .collect(),
+        Some(pb::candidate_tokens::Select::TokenIds(ids)) => {
+            ids.ids.iter().copied().take(MAX_CANDIDATES).collect()
+        }
+        Some(pb::candidate_tokens::Select::All(true)) => (1..=MAX_CANDIDATES)
+            .map(|offset| selected.wrapping_add(offset as u32))
+            .collect(),
+        Some(pb::candidate_tokens::Select::All(false)) => Vec::new(),
```
jthomson04, line 431: [P3] TopN, TokenIds, and All(true) are silently capped at 20. Please either reject requests beyond the mock's supported limit or clearly expose the limitation; otherwise tests can mistake a truncated response for vLLM behavior.

lib/mocker/servers/vllm/src/server.rs
```diff
@@ -0,0 +1,200 @@
...
+        let (prepared, mut live, _permit) = self.start_generation(request.into_inner()).await?;
+        let mut output_ids = Vec::with_capacity(prepared.max_output_tokens);
+        while let Some(signal) = live.recv().await {
+            let token_id = checked_token(&signal).map_err(|status| *status)?;
+            output_ids.push(token_id);
+            if signal.completed {
+                return Ok(Response::new(pb::GenerateResponse {
+                    prompt_info: Some(prepared.prompt_info()),
+                    outputs: Some(prepared.sequence_output(&output_ids, true)),
+                }));
```
jthomson04, line 145: [P3] The integration tests exercise GenerateStream but not a successful Generate call. Since the mock advertises both RPCs, please add one minimal unary test covering accumulated output and terminal metadata.

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
+    Uuid::from_bytes(bytes)
+}
+
+fn synthetic_token_seed(seed: u64, request_id: &str) -> u64 {
+    let mut hasher = blake3::Hasher::new();
+    hasher.update(&seed.to_le_bytes());
+    hasher.update(request_id.as_bytes());
+    let bytes = hasher.finalize();
+    u64::from_le_bytes(bytes.as_bytes()[..8].try_into().unwrap())
```
jthomson04: [P3] The digest is always long enough, but this remains an unwrap in non-test code. Copy the first eight bytes into a fixed-size array so the invariant is explicit without a fallible conversion.

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+
+        match (remote_decode, remote_prefill) {
+            (Some(true), Some(true)) => Err(Status::invalid_argument(
+                "KV transfer payload cannot request both remote prefill and remote decode",
+            )
+            .into()),
+            (Some(true), _) => {
+                if DECODE_RENDEZVOUS_FIELDS
+                    .iter()
+                    .any(|key| params.fields.contains_key(*key))
```
PeaBrane, line 349: AI-assisted review — please evaluate carefully: this appears to accept only the mock's synthetic NIXL-like payload shape. Dynamo's NIXL prefill payload includes the `remote_*` keys with null values, so the key-presence check rejects it; Mooncake decode payloads use `transfer_id` / `remote_bootstrap_addr` and omit several fields required below. That means the advertised disaggregated flow may reject valid connector payloads rather than treating the handoff as opaque. Would validating only the role flags, or selecting an explicit connector schema, be safer?

lib/mocker/servers/vllm/src/server.rs
```diff
@@ -0,0 +1,200 @@
...
+                outputs: None,
+            };
+
+            let mut generated = 0usize;
+            while let Some(signal) = live.recv().await {
+                let token_id = checked_token(&signal).map_err(|status| *status)?;
+                generated += 1;
+                yield pb::GenerateResponse {
+                    prompt_info: None,
+                    outputs: Some(prepared.sequence_output(&[token_id], signal.completed)
```
PeaBrane, line 197: AI-assisted review — please evaluate carefully: this one-token-per-poll response stream sits directly on `LiveEngine`'s fixed eight-signal request queue, whose dispatcher cancels when that queue is full. With the documented `speedup_ratio=1000`, a producer getting briefly more than eight tokens ahead of Tonic can close the channel before a terminal signal, and this handler then returns `INTERNAL`. A pump/batching boundary plus an actual Tonic test longer than eight tokens may be needed to distinguish a disconnected client from normal transport skew.

lib/mocker/servers/vllm/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    tracing::info!(
+        listen = %args.listen,
+        model = %service.config().model,
+        mode = %service.config().mode,
+        "starting Mocker-backed vLLM gRPC server"
+    );
+    tonic::transport::Server::builder()
+        .add_service(GenerateServer::new(service))
+        .serve_with_shutdown(args.listen, async {
```
PeaBrane, line 86: AI-assisted review — please evaluate carefully: Tonic's graceful shutdown waits for active HTTP/2 connections, but this Ctrl-C signal is not propagated into `VllmMockerService` or its active streams. A client that stays connected while no longer consuming `GenerateStream` can therefore keep the response body and `LiveRequest` alive indefinitely and prevent process exit. Consider a shared shutdown token for active streams and/or a bounded graceful-drain timeout.

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+                    "text tokenization is not available in the CPU-only Mocker server; send token_ids",
+                )
+                .into());
+            }
+            None => return Err(Status::invalid_argument("prompt is required").into()),
+        };
+        if prompt_tokens.is_empty() {
+            return Err(Status::invalid_argument("token_ids must not be empty").into());
+        }
+        if request.truncate_prompt_tokens > 0 {
```
PeaBrane, line 90: AI-assisted review — please evaluate carefully: vLLM 0.25.1 rejects any nonzero `truncate_prompt_tokens` as unsupported, whereas the mock applies truncation. Because this server advertises the native Generate API, direct compatibility tests could pass here with a request the real endpoint rejects. Should this return `INVALID_ARGUMENT`, or should the fidelity difference be an explicit documented limitation?

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+        }
+    }
+
+    pub(super) fn output_token(&self, position: usize) -> u32 {
+        deterministic_token_id(self.output_token_seed, position)
+    }
+
+    pub(super) fn prompt_info(&self) -> pb::PromptInfo {
+        let wants_logprobs = self.response.prompt_logprobs;
+        let wants_tokens = self.response.prompt_token_ids || wants_logprobs;
```
PeaBrane, line 213: AI-assisted review — please evaluate carefully: vLLM 0.25.1 treats `prompt_token_ids` and `prompt_logprobs` as independent response flags, but this returns token IDs whenever logprobs are requested. The current sidecar sets both together, so the immediate impact is limited, but direct tests can consume data the real endpoint suppresses. Keeping these flags independent would preserve protocol fidelity.

lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -0,0 +1,529 @@
...
+            max_output_tokens,
+            priority: request.priority,
+            response: request.response.unwrap_or_default(),
+            mode: config.mode,
+        })
+    }
+
+    pub(super) fn direct_request(&self) -> DirectRequest {
+        DirectRequest {
+            tokens: self.prompt_tokens.clone(),
```
PeaBrane, line 193: AI-assisted review — please evaluate carefully: `PreparedRequest` retains the original prompt vector for the whole generation while this clones it into the scheduler's `ActiveSequence`. At the default 256 admitted RPCs, long-context requests can turn this into hundreds of MiB of avoidable duplicate storage. When prompt details were not requested, could the vector move into `DirectRequest` while the response side retains only the prompt length and handoff block count?

lib/mocker/servers/vllm/src/server_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+        VllmMockerService::new(disabled, MockEngineArgs::default())
+            .err()
+            .unwrap()
+            .to_string()
+            .contains("max_concurrent_requests")
+    );
+}
+
+#[tokio::test]
+async fn unary_generate_maps_capacity_rejection_to_resource_exhausted() {
```
PeaBrane, line 207: AI-assisted review — please evaluate carefully: this verifies capacity rejection only through unary `Generate`, while the sidecar exclusively uses `GenerateStream`. In the streaming path the RPC first succeeds and `RESOURCE_EXHAUSTED` arrives later as a stream item, so error propagation and cleanup are materially different. A streaming capacity-rejection test would protect the production sidecar path.

PeaBrane (commented):
AI-assisted static review; no tests were run. These comments are intentionally advisory, and I would prefer careful human evaluation of each before deciding whether a change is warranted. Existing open review-thread concerns were omitted.

lib/mocker/servers/vllm/tests/sidecar.rs
connorcarpenter15: Good catch — the round trip had no teeth for opacity. Fixed: the prefill handoff's non-rendezvous `mocker_request_id` is now an explicit sentinel (`HANDOFF_SENTINEL_FIELD`) that the decode role requires in `KvTransferRole::classify`, so a decode payload that dropped it is rejected with `INVALID_ARGUMENT`. The integration test now asserts the sentinel survives the prefill→decode round trip, and a new unit test (`decode_rejects_a_handoff_missing_the_opacity_sentinel`) fails loudly if it's dropped. So the disagg flow now catches a sidecar that stops forwarding opaque fields verbatim.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Documented the limit — the 20-candidate cap is now called out in the README's *Deliberate limitations*. Note vLLM's default `max_logprobs` is also 20, but it rejects over-limit requests rather than truncating. Since the candidate count here is driven by the sidecar's logprob count (itself bounded), I opted to document rather than hard-reject, to avoid diverging from the inputs the sidecar actually sends. Happy to switch to a hard `INVALID_ARGUMENT` if you'd prefer strict parity.

lib/mocker/servers/vllm/src/server.rs
connorcarpenter15: Added — `unary_generate_accumulates_output_and_terminal_metadata` now covers a successful unary `Generate`: it asserts the accumulated `token_ids`/`num_tokens` and the terminal `FinishInfo` (Length + `num_output_tokens`). The prior unary coverage only exercised the rejection path.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Fixed — `synthetic_token_seed` now copies the first eight digest bytes into a `[0u8; 8]` with `copy_from_slice`, matching `stable_uuid` right above it. The 8-byte invariant is now explicit and there's no fallible conversion.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Thanks — I checked this against what the Dynamo sidecar actually puts on the wire. In `build_kv_parameters` (`lib/sidecar/vllm/src/convert.rs`) the prefill path inserts only `do_remote_decode=true` (no null `remote_*` keys), and the decode path forwards the prefill handoff verbatim. So for the current sidecar + NIXL flow the key-presence checks are exactly compatible — the passing `prefill_handoff_round_trips_through_a_decode_server` integration test exercises this against the real `VllmSidecarEngine`. The null-valued `remote_*` keys you mention are vLLM's *internal* connector dataclass, not the sidecar wire payload; and Mooncake isn't a target here (the mock's prefill role emits the NIXL-shaped payload by design). I've kept the validation strict on purpose so the disagg round trip has teeth (I actually strengthened it — see jthomson04's opacity-sentinel thread), and clarified in the README that decode expects a NIXL-shaped payload. If we later want connector-agnostic decode, relaxing to role-flags-only would be the move — happy to do that as a follow-up if you think it's worth it.

lib/mocker/servers/vllm/src/server.rs
connorcarpenter15: Confirmed and addressed — nice find. You're right: `LiveEngine`'s fixed 8-signal per-request buffer sheds a slow consumer (dispatcher `try_send` → `Full` → cancel), and this handler would then surface that as `INTERNAL`. The shedding itself lives in `lib/mocker/src/live.rs` (a separate PR, #11917), so I fixed it at this server's boundary: `generate_stream` now runs a pump that drains the engine promptly into a per-request buffer bounded by the request's own token budget, so transport skew no longer trips the shed. The pump also selects on `Sender::closed()`, so a dropped client still cancels scheduler work promptly (verified by the existing `dropping_sidecar_stream_cancels_mocker_work`).

Added the Tonic-level test you asked for: `streaming_survives_a_producer_that_outruns_a_stalled_consumer` — instant producer, 50 tokens, a stalled consumer — which fails with the old code (INTERNAL after ~8 tokens) and passes now (all 50 tokens + a clean `LENGTH` terminal).

lib/mocker/servers/vllm/src/main.rs
connorcarpenter15: Fair in principle — a stuck-but-connected client could delay graceful shutdown. I've left it as-is for this CPU-only test binary: the integration harness uses its own `serve_with_incoming_shutdown` + oneshot (`tests/sidecar.rs`), the binary is for manual/example use, and the pump added for the streaming-shed fix means a dropped client now cancels promptly. I don't think a shared shutdown token / drain deadline earns its complexity here, but I'm happy to add a bounded graceful-drain timeout if you feel strongly.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Verified this is moot for the target path: the sidecar hardcodes `truncate_prompt_tokens: 0` (`convert.rs`), so it never sends a nonzero value — the mock's truncation branch is unreachable through the sidecar. It would only matter for hand-crafted direct-proto tests, which aren't the mock's purpose. I've left the lenient behavior rather than adding an `INVALID_ARGUMENT` the sidecar can never trigger; easy to revisit if we add direct-proto conformance tests that need strict vLLM parity.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Confirmed there's no observable difference on the target path: the sidecar sets `prompt_token_ids` and `prompt_logprobs` together (`prompt_token_ids: prompt_logprobs.is_some()`, `convert.rs`), which is exactly the coupling here. Decoupling them would be a one-line change but has zero effect for any request the sidecar issues, so I've left it as-is. Will decouple if we ever exercise the flags independently.

lib/mocker/servers/vllm/src/server_request.rs
connorcarpenter15: Real observation, but I've left it: this is a CPU-only test mock, the tests use tiny prompts, and nobody drives 256 × long-context requests through it, so the duplicate prompt storage isn't a concern in this context. The move-when-details-not-requested optimization would complicate `PreparedRequest` (it still needs prompt length + block count for `prompt_info`/`handoff`) for no practical benefit here. Noting it in case the mock ever grows a load-testing use case.

lib/mocker/servers/vllm/src/server_tests.rs
connorcarpenter15: Good call — added `streaming_generate_maps_capacity_rejection_to_resource_exhausted`. It drives the same undersized KV cache through `generate_stream` and asserts the production shape you described: the RPC opens successfully, `prompt_info` is yielded, then `RESOURCE_EXHAUSTED` arrives as a later stream item. That covers the sidecar's actual path, which the unary rejection test didn't.

PeaBrane (approved):
The review comments have been addressed. Approving on the condition that the promised soak test is run and its results are delivered offline.

jthomson04 (approved):

lib/mocker/servers/vllm/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+description = "Mocker-backed implementation of vLLM's native Generate gRPC API"
+readme = "README.md"
+
+[[bin]]
+name = "dynamo-vllm-mocker-server"
+path = "src/main.rs"
+
+[dependencies]
+dynamo-mocker = { workspace = true }
+dynamo-vllm-sidecar = { path = "../../../sidecar/vllm", version = "1.3.0" }
```
nv-anants: can you remove version from here?

will be hard to keep in sync, its already out of sync i think - main moved to 1.4.0

connorcarpenter15: Fixed.

nv-anants (approved):

PR #11874, opened 2026-07-18
feat(sglang): add mock gRPC server for sidecar tests

connorcarpenter15:
## Overview:

Add a CPU-only `dynamo-sglang-mocker-server` that exercises the real SGLang sidecar gRPC path without SGLang, model weights, or GPUs.

## Details:

- Place the server under `lib/mocker/servers/sglang` and temporarily reuse `dynamo_sglang_sidecar::proto`, matching the vLLM mocker layout from #11873 without a standalone proto crate.
- Support aggregated, prefill, and decode roles with deterministic tokens/logprobs, targeted aborts, capacity behavior from the real Mocker scheduler, and metadata-only rendezvous handoff.
- Validate and round-trip bootstrap host, port, and room metadata without opening a KV transport or moving KV data.
- Cover discovery, generation, logprobs, prefill/decode handoff, and abort cleanup through the real SGLang sidecar.

Validation:

- `cargo metadata --locked --format-version 1` for the root, Python bindings, KVBM bindings, and runtime examples workspaces
- `cargo clippy -p dynamo-sglang-mocker --all-targets --all-features -- -D warnings`
- `cargo test -p dynamo-mocker` (597 passed, 1 ignored)
- `cargo test -p dynamo-sglang-mocker` (10 passed)

Start with `lib/mocker/servers/sglang/src/server.rs` for the RPC lifecycle and `lib/sidecar/sglang/src/proto.rs` for the temporary sidecar-owned protocol export. The real sidecar boundary is covered by `lib/mocker/servers/sglang/tests/sidecar.rs`.

## Related Issues

- [DIS-2475: Add an SGLang-compatible Mocker gRPC server](https://linear.app/nvidia/issue/DIS-2475/add-an-sglang-compatible-mocker-grpc-server)

## Summary by CodeRabbit

* **New Features**
  * Added a CPU-only SGLang-compatible mock server for local development and testing.
  * Supports aggregated and disaggregated prefill/decode serving modes, streaming generation, health and model discovery, request cancellation, and deterministic outputs.
  * Added optional log probabilities and top-logprob metadata to generated responses.

* **Bug Fixes**
  * Improved scheduler cancellation so completed work exits promptly.

* **Compatibility**
  * Added safeguards to preserve SGLang gRPC request and wire-format compatibility.


ishandhanani (approved):

lib/mocker/servers/sglang/src/server.rs
```diff
@@ -0,0 +1,497 @@
...
+            .clone()
+            .try_acquire_owned()
+            .map_err(|_| Status::resource_exhausted("Mocker concurrent request limit reached"))?;
+        let prepared = PreparedRequest::new(request, &self.config).map_err(|status| *status)?;
+        let live = self
+            .engine
+            .submit(prepared.direct_request())
+            .await
+            .map_err(|error| {
+                if error.to_string().contains("already active") {
```
jthomson04: **Brittle cross-crate error-string match.**

This depends on the exact wording of `bail!("request {client_id} is already active")` in `lib/mocker/src/live.rs`. Rewording that message silently downgrades duplicate-`rid` from `ALREADY_EXISTS` to `INTERNAL`, and there is no test covering the duplicate-`rid` path to catch it. The vLLM sibling (`lib/mocker/servers/vllm/src/server.rs`) doesn't do this — it maps every submission failure to `INTERNAL`.

Suggest having `LiveEngine::submit` return a distinguishable error (typed variant, or at minimum an exported `const` for the message) and adding a test that submits the same `rid` twice.

lib/sidecar/sglang/src/protocol.rs
```diff
@@ -427,9 +427,24 @@ fn prompt_logprobs_from_meta(meta: &HashMap<String, String>) -> Result<Option<Va
         _ => Vec::new(),
     };
 
-    let mut payload = Vec::with_capacity(input_logprobs.len() + 1);
-    payload.push(Value::Null);
+    // Current SGLang encodes the first requested prompt position as
+    // `[null, token_id, null]`; older releases omitted it entirely.
+    let has_native_sentinel = input_logprobs.first().is_some_and(|entry| {
```
jthomson04, line 439: **The sentinel is inferred from one array but applied to two.**

`has_native_sentinel` is derived only from `input_token_logprobs[0]`, but it then shifts the index used against `input_top_logprobs` further down (`input_top_logprobs.get(index)`). If SGLang ever emits the leading sentinel in one array and not the other, every top-logprob entry shifts by one position and gets silently attributed to the wrong prompt token — no error, just wrong data.

Consider deriving the flag per-array, or validating that the two arrays agree in length/shape before indexing them with a shared offset.

lib/sidecar/sglang/src/protocol.rs
```diff
@@ -705,18 +720,18 @@ mod tests {
         let meta = HashMap::from([
             (
                 "input_token_logprobs".to_string(),
-                json!([[-0.1, 10, "a"], [-0.2, 11, "b"]]).to_string(),
+                json!([[null, 10, null], [-0.2, 11, "b"]]).to_string(),
```
jthomson04, line 771: **Legacy-SGLang coverage was replaced rather than added.**

The new comment above `has_native_sentinel` justifies the branch by saying "older releases omitted it entirely" — but this test was rewritten from the non-sentinel form to the sentinel form, so the `!has_native_sentinel` path (the one that comment exists for) now has zero coverage.

Please keep both: this test for the sentinel encoding, plus the original `json!([[-0.1, 10, "a"], [-0.2, 11, "b"]])` / `json!([[[-0.3, 12, "c"]], []])` case asserting `prompt[1]["10"]["logprob"] == -0.1`.

lib/mocker/servers/sglang/src/server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    output_ids: vec![output_id],
+                    meta_info: prepared.meta_info(&output_tokens, signal.completed),
+                    finished: signal.completed,
+                };
+                if signal.completed {
+                    return;
+                }
+            }
+            Err(Status::internal(
+                "Mocker output channel closed before a terminal response",
```
jthomson04, line 286: **Abort ends the stream with `INTERNAL` instead of an SGLang-shaped terminal.**

After the `Abort` RPC, `live.recv()` returns `None`, the pump exits, `signal_rx` closes, and this trailing `Err` fires. So a *successful* cancellation and a dead engine are reported identically as `INTERNAL`.

Real SGLang ends an aborted stream with a terminal chunk carrying `finish_reason: {"type": "abort"}`, which `protocol.rs` explicitly handles (`abort_terminal_preserves_failure_metadata_as_error`) — so that sidecar path can never be exercised end-to-end against this mock. `sidecar_abort_releases_mocker_work` doesn't surface it because the sidecar's own `ctx.stopped()` select wins first and it yields `LLMEngineOutput::cancelled()`.

Either track that an abort was requested and end the stream cleanly (or synthesize the abort terminal), or call this out explicitly under "Deliberate limitations" in the README.

lib/mocker/servers/sglang/src/server_request.rs
```diff
@@ -0,0 +1,269 @@
...
+            (
+                "prompt_tokens".to_string(),
+                Value::from(self.prompt_tokens.len()).to_string(),
+            ),
+            (
+                "mocker_request_id".to_string(),
+                Value::String(self.request_id.clone()).to_string(),
+            ),
+        ]);
+        if self.return_logprob {
```
jthomson04, line 145: **`output_top_logprobs` is emitted even when `top_logprobs_num == 0`.**

This gates only on `return_logprob`, so with `top_logprobs_num == 0` the mock emits `output_top_logprobs: [[], [], ...]`. Real SGLang omits the key entirely in that case. Downstream, `extract_logprobs` turns that into `top_logprobs: Some(vec![vec![]])` where production would yield `None` — a divergence that could mask a frontend bug rather than expose it.

Same applies to `input_top_logprobs` in the terminal branch below. Suggest gating both on `self.top_logprobs_num > 0`.

(Related, on the validation side: `top_logprobs_num` above 20 is silently clamped to `MAX_TOP_LOGPROBS` while negative values are rejected with `INVALID_ARGUMENT`. The asymmetry is undocumented, and real SGLang rejects over-limit values.)

jthomson04 (commented):
Five correctness/fidelity findings from reading the new mock server against the real sidecar and scheduler code. Nothing here looks blocking-critical, but 1 and 2 are the ones I'd most want addressed before merge.

lib/mocker/servers/sglang/src/server.rs
connorcarpenter15: Fixed in dcefe7d96c. I removed the error-string match and now map every LiveEngine submission failure to INTERNAL, matching the vLLM adapter and native SGLang behavior. No duplicate-rid test is needed because there is no longer a distinct status contract to preserve.

lib/sidecar/sglang/src/protocol.rs
connorcarpenter15: Fixed in dcefe7d96c. When top logprobs are present, the sidecar now requires equal array lengths and matching sentinel encoding before sharing indices. Inconsistent metadata returns a protocol error instead of silently shifting entries.

lib/sidecar/sglang/src/protocol.rs
connorcarpenter15: Fixed in dcefe7d96c. The existing test now covers both native sentinel and legacy non-sentinel encodings, and also verifies that inconsistent encodings are rejected.

lib/mocker/servers/sglang/src/server.rs
connorcarpenter15: Documented in dcefe7d96c under Deliberate limitations. Abort releases Mocker scheduler state but does not synthesize an SGLang abort terminal; a caller that continues polling receives INTERNAL when the output channel closes. The real sidecar cancellation path remains covered end to end.

lib/mocker/servers/sglang/src/server_request.rs
connorcarpenter15: Fixed in dcefe7d96c. The mock now omits input and output top-logprob metadata when the requested count is zero, and rejects values outside 0 through 20 instead of silently clamping. Coverage checks both the wire shape and upper bound.

jthomson04 (approved):

nv-anants (approved):

rmccorm4 (approved):
approving CODEOWNERS / lock files for @ai-dynamo/dynamo-process-codeowners

PR #11883, opened 2026-07-19
fix(snapshot): write rootfs-diff.tar atomically, handle empty archive

dfeigin-nv:
## Summary

Port correctness fixes from [Radostin Stoyanov (rst0git)](https://github.com/rst0git) at [doublewordai/dynamo@e7c78a2](https://github.com/doublewordai/dynamo/commit/e7c78a25987e2c8441b5cc5f8101573f0115838b).

- **Atomic capture**: `CaptureRootfsDiff` previously wrote directly to `rootfs-diff.tar`. A failed or interrupted capture left a corrupt partial archive that restore would silently attempt to extract. Now writes to a `.tmp` file and renames atomically on success; `defer` removes the temp on any error path.
- **Empty archive guard**: reject a zero-byte tar (empty upperdir edge case) before publishing.
- **`ApplyRootfsDiff` stat errors**: non-ENOENT stat errors (permissions, I/O failure) were silently swallowed and fell through to `tar -xf`. Now propagated. Zero-byte archive is treated as no-op instead of producing a confusing `tar: archive is empty` error.
- **Tests**: add `TestCaptureRootfsDiff` and `TestApplyRootfsDiff` covering round-trip, no-temp-leak-on-failure, and missing/empty/invalid archive cases. We had zero test coverage of these functions.

## Test plan

- [ ] `go test ./internal/runtime/... -run TestCaptureRootfsDiff` — 2 subtests pass
- [ ] `go test ./internal/runtime/... -run TestApplyRootfsDiff` — 3 subtests pass
- [ ] Full CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot root filesystem archive creation to prevent incomplete or empty archives from being published.
  * Added safer handling for archive creation failures, including cleanup of temporary files.
  * Applying snapshots now safely skips missing or empty root filesystem archives while reporting other file access errors.
  * Invalid non-empty archives are correctly rejected instead of being silently ignored.

* **Tests**
  * Added coverage for archive creation, application, cleanup, and error scenarios.

### DIFF
--- a/deploy/snapshot/internal/runtime/overlay.go
+++ b/deploy/snapshot/internal/runtime/overlay.go
@@ -50,12 +50,31 @@ func GetOverlayUpperDir(pid int) (string, error) {
 }
 
 // CaptureRootfsDiff captures the overlay upperdir to a tar file.
+//
+// The archive is written to a temporary file first and renamed atomically on
+// success, so a partial or failed capture never leaves a corrupt archive at the
+// final path.
 func CaptureRootfsDiff(upperDir, checkpointDir string, exclusions types.OverlaySettings, bindMountDests []string) (string, error) {
 	if upperDir == "" {
 		return "", fmt.Errorf("upperdir is empty")
 	}
 
 	rootfsDiffPath := filepath.Join(checkpointDir, rootfsDiffFilename)
+
+	tmpFile, err := os.CreateTemp(checkpointDir, rootfsDiffFilename+".*.tmp")
+	if err != nil {
+		return "", fmt.Errorf("failed to create temporary rootfs diff: %w", err)
+	}
+	tmpPath := tmpFile.Name()
+	if err := tmpFile.Close(); err != nil {
+		_ = os.Remove(tmpPath)
+		return "", fmt.Errorf("failed to close temporary rootfs diff: %w", err)
+	}
+	defer func() {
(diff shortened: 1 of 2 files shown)


deploy/snapshot/internal/runtime/overlay.go
```diff
@@ -50,28 +50,60 @@ func GetOverlayUpperDir(pid int) (string, error) {
...
+		return "", fmt.Errorf("failed to stat temporary rootfs diff: %w", err)
+	}
+	if info.Size() == 0 {
+		return "", fmt.Errorf("tar produced empty rootfs diff")
+	}
+
+	if err := os.Rename(tmpPath, rootfsDiffPath); err != nil {
+		return "", fmt.Errorf("failed to publish rootfs diff: %w", err)
+	}
+	tmpPath = ""
```
galletas1712, line 105: do we need this line?

galletas1712 (approved):

deploy/snapshot/internal/runtime/overlay.go
galletas1712: I mean tmpPath = ""

dfeigin-nv: It's actually important for the defer to not try to remove the renamed file

Issue #11875, opened 2026-07-18
DEP: Native composable worker scoring and picking


ishandhanani:
## Summary

Add native, composable worker scoring and picking inside Dynamo's existing `WorkerSelector` boundary. A custom router or EPP binary can link scorer and picker implementations from a normal Rust crate and provide them when it constructs `SelectionService`; no Dynamo source fork or runtime-loaded library is required.

Dynamo continues to own policy-class queues, scheduling, worker discovery, `RoutingEligibility`, final target validation, accounting, reservation, rollback, and metrics. The built-in path remains concrete and statically dispatched. Custom policy dispatch is opt-in.

```text
EPP / frontend
  -> Selector
  -> SelectionService
  -> policy-class queues / scheduling
  -> existing RoutingEligibility
  -> WorkerSelector
       -> additive lower-is-better Scorer(s)
       -> Picker
  -> host validation / accounting / reservation
```

## Motivation

Dynamo already has a public `WorkerSelector<C>` trait, and `LocalScheduler` is generic over it. The missing pieces are narrower scorer and picker contracts plus a way for `SelectionService`, which currently hard-codes `DefaultWorkerSelector`, to construct an opted-in custom implementation.

The current default selector combines KV/load cost calculation with zero-temperature tie selection and nonzero-temperature sampling. Customers who want to change ranking or picking should not need to replace Dynamo's queueing, eligibility, reservation, rollback, or metrics.

A runtime-loaded C-ABI selector was explored in #11876 and #11884. That remains a possible separate solution for prebuilt binaries, but it is unnecessary for Rust crates linked into a custom image and should not shape the native API.

### Preserve the existing scheduling boundary

This work builds on the existing `WorkerSelector<C>` trait and `LocalScheduler` generic. It does not introduce another scheduling abstraction.

`RoutingEligibility` continues to enforce allowed workers, overload state, routing constraints, pinning, and worker availability exactly as it does today. Eligibility is not customizable through this API, and there is no policy filter stage.

The scorer/picker extension sees only the dense set of workers allowed by existing eligibility:

- Each scorer contributes a finite, lower-is-better cost for each eligible worker row.
- Dynamo adds contributions from scorers in order.
- One picker observes the eligible rows and accumulated costs and returns at most one row index.
- Dynamo validates that the returned index names an eligible row before mapping it to a worker target.
- Policy code never books capacity directly.

[...]

PR #11886, opened 2026-07-19
feat(router): load runtime worker selector plugins

ishandhanani:
This loads trusted worker-selector `cdylib` plugins at KV-router construction and runs them inside the existing scheduler actor. The stock path remains statically dispatched and unchanged, while configured plugins can compose existing cache, load, session, taint, and capacity signals without forking Dynamo.

### How This Was Implemented

- Erases selector type parameters only after each scheduler actor starts, preserving public generic types without adding a stock-path branch or vtable.
- Loads and validates the v1 ABI from the shared `ModelManager::kv_chooser_for` boundary used by decode, deferred prefill, Python, C, and ext-proc KV routers.
- Reuses selector-owned candidate storage and exposes only already-computed request/worker signals.
- Keeps eligibility, pin resolution, selected-target validation, accounting, and optional fallback to the exact configured default selector host-owned.
- Replaces mirrored ABI fixtures with three focused tests: descriptor rejection, eager symbol binding, and one real `cdylib` selection/fallback smoke test.

<details>

#### Mental model

The environment is checked once per KV-router construction. With no plugin configured, the existing concrete `DefaultWorkerSelector` actor is started; with a plugin configured, Dynamo loads native state for that router role and starts the same actor shape with `RuntimePluginSelector`.

```mermaid
flowchart LR
    Build["ModelManager::kv_chooser_for"] --> Config{"Plugin configured?"}
    Config -->|"No"| Stock["Queue actor<DefaultWorkerSelector>"]
    Config -->|"Yes"| Load["Validate + create cdylib state"]
    Load --> Plugin["Queue actor<RuntimePluginSelector>"]
    Plugin --> Decision{"Candidate or UseDefault"}
    Decision --> Validate["Host validation + accounting"]
    Stock --> Validate
```

#### State and ordering

Each decode or prefill router owns independent plugin state and its library handle. Calls to that state are serialized by the queue actor; eligible candidates arrive in unspecified order, so plugins can tie-break on stable worker ID, DP rank, or stable routing ID. `UseDefault` invokes a `DefaultWorkerSelector` constructed from the same router configuration.

#### Request lifecycle

Dynamo resolves pins and filters eligibility before plugin selection. The loader uses an absolute path with eager local symbol binding, validates the descriptor version, size, and callbacks, rejects failed or null state creation, and then maps existing scheduling signals into callback-scoped borrowed records. A successful candidate index is validated again by the target-only selector adapter before Dynamo books its normal accounting snapshot; callback errors and invalid indices fail the request. Destruction runs before the selector-owned library handle is released.

Plugins are trusted native code in the Dynamo process; there is no isolation, hot reload, or separate service. The callback runs synchronously on the scheduler actor and must remain bounded and non-blocking. Non-KV router modes do not construct a KV chooser and therefore ignore these KV-specific environment variables.

</details>

### Validation

- `cargo fmt --manifest-path lib/kv-router/examples/custom-worker-selector/Cargo.toml -- --check`
- `cargo test -p dynamo-kv-router scheduling::runtime_plugin --lib`
- `cargo check -p dynamo-llm`
- `cargo check --manifest-path lib/bindings/python/Cargo.toml --lib`
- `cargo clippy -p dynamo-kv-router -p dynamo-llm --all-targets -- -D warnings`

### Related Issues

- Stacked on #11884
- Relates to #11875


ishandhanani:
Moved to https://github.com/ishandhanani/dynamo/pull/5 for continued development outside ai-dynamo/dynamo. The DEP remains ai-dynamo/dynamo#11875.

PR #11903, opened 2026-07-20
fix(runtime): file discovery loses concurrent registrations - make bucket lookup-or-insert atomic

h-avsha:
## Overview:

With `--discovery-backend file`, a worker can silently lose some of its endpoint registrations
~10s after a clean startup: the frontend drops the model (`/v1/models` empty, every request
503s) while the worker stays fully healthy, and nothing ever re-registers - only a restart
recovers. We hit this on two consecutive restarts of a single-pod deployment (frontend + vllm
worker sharing `/tmp/dynamo_store_kv`).

Root cause: `get_or_create_bucket` does the `active_dirs` lookup and insert as two separate
locked steps. Concurrent registrations into a brand-new bucket (a worker registering its 3
endpoints within ~300µs is the normal startup sequence) each construct their own `Directory`
with their own empty `owned_files` set; only the last one stays in the map. Files recorded on
the discarded ones are never mtime-refreshed by the keep-alive thread, so the TTL cleanup scan
deletes them 10s later. The only log trace is a single `Expired path=...` INFO line.

## Details:

All changes in `lib/runtime/src/storage/kv/file.rs`:

- `get_or_create_bucket` / `get_bucket`: lookup-or-insert under a single lock via
  `entry(...).or_insert_with_key(...)`, so concurrent callers always share one `Directory`
  (and one `owned_files`). On-disk `create_dir_all` stays outside the lock (idempotent).
- Watcher: match `Modify(Data(_))` instead of only `Modify(Data(DataChange::Content))` - Linux
  inotify reports in-place writes as `DataChange::Any`, so non-rename writers were invisible.
  Keep-alive touches are `Modify(Metadata)` and remain ignored.
- New test `test_concurrent_get_or_create_bucket_returns_one_directory`: 50 rounds × 4
  concurrent registrations into a fresh bucket on a multi-threaded runtime. Fails on the first
  round against the unfixed code - the race is near-deterministic, matching the 2-for-2
  production hit rate.
- New test `test_insert_existing_revision_zero_does_not_take_ownership`: pins the deliberately
  unchanged create-if-absent semantics (a revision-0 `insert` returning `Exists` must not take
  ownership of the winner's file), since ownership is now exactly what decides whether a
  registration survives.

All file-store tests pass; clippy and rustfmt clean.

The two `entry(...).or_insert_with_key(...)` blocks in `get_or_create_bucket` and `get_bucket`
— they are the fix. Then `test_concurrent_get_or_create_bucket_returns_one_directory`, which
demonstrates the race against the old code.

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11903" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent file storage access so simultaneous requests for the same bucket consistently share the same storage instance.
  * Improved file watching to detect a broader range of file modification events, ensuring updates are reflected more reliably.
  * Added safeguards and coverage for concurrent bucket creation and file ownership handling.

### DIFF
--- a/lib/runtime/src/storage/kv/file.rs
+++ b/lib/runtime/src/storage/kv/file.rs
@@ -140,11 +140,22 @@ impl Store for FileStore {
                 ));
             }
         } else {
-            // Create
+            // Create. Idempotent, so concurrent callers racing past the fast path
+            // above are fine.
             fs::create_dir_all(&p).map_err(to_fs_err)?;
         }
-        let dir = Directory::new(self.root.clone(), p.clone(), ttl.unwrap_or(DEFAULT_TTL));
-        self.active_dirs.lock().insert(p, dir.clone());
+        // Lookup-or-insert under a single lock: concurrent callers must all get the
+        // SAME Directory. A duplicate Directory has its own owned_files set that the
+        // keep-alive thread never walks, so every file registered through it would
+        // silently expire after TTL.
+        let dir = self
+            .active_dirs
+            .lock()
+            .entry(p)
+            .or_insert_with_key(|p| {
+                Directory::new(self.root.clone(), p.clone(), ttl.unwrap_or(DEFAULT_TTL))
+            })
+            .clone();
         Ok(dir)
     }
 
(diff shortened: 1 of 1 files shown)


PeaBrane (commented):
Thanks for tracking this down. The concurrent bucket-registration race is already fixed on `main` by #11841. That change makes the `active_dirs` entry insertion atomic in both `get_or_create_bucket` and `get_bucket`, and includes `concurrent_bucket_open_shares_ownership_registry`, a barrier-synchronized 32-caller regression test that directly verifies all bucket handles share one ownership registry.

Given that overlap—and the resulting conflict with `main`—I think we can close this PR. The broader `Modify(Data(_))` watcher change is separate from the bucket race and could be proposed independently with watcher-focused coverage if still needed.

rmccorm4:
> Given that overlap—and the resulting conflict with main—I think we can close this PR. The broader Modify(Data(_)) watcher change is separate from the bucket race and could be proposed independently with watcher-focused coverage if still needed.

Closing this PR due to previous comment and 👍 from @h-avsha on it -- feel free to re-open or open a new one as needed and we can help review @h-avsha !

Issue #11642, opened 2026-07-14
DEP: Rewrite Dynamo custom resources to v1beta1 storage


carlory:
## Summary

Dynamo now automatically migrates existing custom resources to each CRD's current storage version when the cluster-wide Operator starts. The implementation landed in [#11904](https://github.com/ai-dynamo/dynamo/pull/11904).

The change:

- promotes `v1beta1` to the storage version for `DynamoComponentDeployment`, `DynamoGraphDeployment`, and `DynamoGraphDeploymentScalingAdapter`;
- includes `DynamoGraphDeploymentRequest` in migration because it already used `v1beta1` storage but upgraded clusters can still report or contain older stored versions;
- keeps `v1alpha1` served and relies on the existing conversion webhook for compatibility;
- registers an asynchronous CRD migration controller only in the cluster-wide Operator;
- rewrites existing objects through the status subresource using Server-Side Apply;
- narrows each migrated CRD's `status.storedVersions` to its current storage version after object migration succeeds; and
- removes managedFields entries only when they reference API versions that the live CRD no longer serves.

The controller is a Dynamo-owned, controller-runtime-only port derived from the Cluster API CRD migrator at v1.13.4. Dynamo does not add a runtime dependency on `sigs.k8s.io/cluster-api`.

## Motivation

Changing `spec.versions[*].storage` affects new writes, but it does not rewrite objects already persisted in etcd. Until existing objects are rewritten, Kubernetes can retain older versions in `CustomResourceDefinition.status.storedVersions`, which prevents safe retirement of the old storage representation.

Users should not need to run an administrative script during an Operator upgrade. Migration should run inside the Operator, retry transient failures, preserve the served `v1alpha1` API, and expose completion through Kubernetes metadata.

## Goals

- Automatically rewrite existing instances of the affected Dynamo CRDs into the current storage version.
- Keep `v1alpha1` served and usable during migration.
- Run migration only from the cluster-wide Operator.
- Make migration asynchronous, idempotent, leader-elected, and retryable.
- Update `status.storedVersions` only after object migration succeeds.
- Preserve object spec, generation, status values, and conversion-preservation annotations.
- Prepare for a future release to stop serving `v1alpha1` by including managedFields cleanup for versions that are no longer served.
- Provide an observable completion marker for each configured CRD.

## Non-Goals

- Removing `v1alpha1` from the CRDs, Go types, scheme registration, or conversion code.
[...]

PR #11904, opened 2026-07-20
feat(operator): migrate CRD storage to v1beta1

carlory:
## Summary

- port the Cluster API CRD migrator as a controller-runtime-only internal package with upstream provenance
- promote DynamoComponentDeployment, DynamoGraphDeployment, DynamoGraphDeploymentRequest, and DynamoGraphDeploymentScalingAdapter storage to v1beta1 while continuing to serve v1alpha1
- register asynchronous migration in the cluster-wide operator, add RBAC, generated manifests, tests, and operator documentation

Part of #11642

## Validation

- GOCACHE=/private/tmp/dynamo-go-cache go test ./internal/crdmigrator ./api ./api/v1alpha1 ./api/v1beta1 ./cmd -count=1
- GOCACHE=/private/tmp/dynamo-go-cache go vet ./internal/crdmigrator ./cmd ./api/...
- verified the migrator dependency graph contains no sigs.k8s.io/cluster-api dependency
- make manifests
- rendered the operator Helm chart in cluster-wide and namespace-restricted modes
- Kind Kubernetes 1.36 upgrade E2E: created v1alpha1-stored objects for all four CRDs, installed the modified operator, verified storedVersions and observed-generation converged to v1beta1, and read every object through both v1beta1 and v1alpha1 APIs

## Summary by CodeRabbit

* **New Features**
  * Added automatic migration of Dynamo custom resources to the v1beta1 storage version.
  * Maintained transparent compatibility with served v1alpha1 APIs.
  * Added migration progress tracking, retries, and cleanup handling.

* **Documentation**
  * Documented migration behavior, completion checks, retries, and restricted-namespace considerations.
  * Updated Kubernetes API references to identify v1beta1 as the storage version.

* **Bug Fixes**
  * Improved operator permissions for monitoring and updating custom resource definitions and their statuses.

### DIFF
--- a/deploy/operator/cmd/main.go
+++ b/deploy/operator/cmd/main.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -68,6 +69,7 @@ import (
 	internalcert "github.com/ai-dynamo/dynamo/deploy/operator/internal/cert"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/controller"
 	commonController "github.com/ai-dynamo/dynamo/deploy/operator/internal/controller_common"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/crdmigrator"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/namespace_scope"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/observability"
@@ -586,6 +588,36 @@ func registerControllers(
 	operatorImage string,
 	operatorPullPolicy corev1.PullPolicy,
 ) error {
+	if operatorCfg.Namespace.Restricted == "" {
+		// The cluster-wide manager cache covers all namespaces. ExcludedNamespaces only filters
+		// business-controller events and does not narrow the cache used by the migrator.
+		migrator := &crdmigrator.CRDMigrator{
+			Client:    mgr.GetClient(),
+			APIReader: mgr.GetAPIReader(),
+			Config: map[client.Object]crdmigrator.ByObjectConfig{
+				&nvidiacomv1beta1.DynamoComponentDeployment{}: {
+					UseCache:                            true,
+					UseStatusForStorageVersionMigration: true,
+				},
+				&nvidiacomv1beta1.DynamoGraphDeployment{}: {
+					UseCache:                            true,
+					UseStatusForStorageVersionMigration: true,
+				},
+				&nvidiacomv1beta1.DynamoGraphDeploymentRequest{}: {
+					UseCache:                            true,
+					UseStatusForStorageVersionMigration: true,
+				},
+				&nvidiacomv1beta1.DynamoGraphDeploymentScalingAdapter{}: {
+					UseCache:                            true,
+					UseStatusForStorageVersionMigration: true,
+				},
+			},
+		}
+		if err := migrator.SetupWithManager(mgr, ctrlcontroller.Options{MaxConcurrentReconciles: 1}); err != nil {
+			return fmt.Errorf("unable to create CRD migrator controller: %w", err)
+		}
+	}
+
 	setupOptions := controller.SetupOptions{
 		Config:        operatorCfg,
 		RuntimeConfig: runtimeConfig,
--- a/deploy/operator/internal/crdmigrator/crd_migrator.go
+++ b/deploy/operator/internal/crdmigrator/crd_migrator.go
@@ -1,0 +1,400 @@
+/*
+SPDX-FileCopyrightText: Copyright 2025 The Kubernetes Authors.
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+
+Derived from kubernetes-sigs/cluster-api/controllers/crdmigrator at v1.13.4,
+commit 27f464418c195d96ae2ef4b96f3b6a047ea89310.
+*/
+
+// Package crdmigrator contains a controller-runtime-only CRD migrator.
+package crdmigrator
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"time"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	// ObservedGenerationAnnotation records the CRD generation reconciled by the migrator.
+	// Dynamo uses an NVIDIA-owned domain instead of the upstream Cluster API annotation domain.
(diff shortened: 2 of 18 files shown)


deploy/operator/cmd/main.go
```diff
@@ -584,6 +586,34 @@ func registerControllers(
...
+				&nvidiacomv1beta1.DynamoComponentDeployment{}: {
+					UseCache:                            false,
+					UseStatusForStorageVersionMigration: true,
+				},
+				&nvidiacomv1beta1.DynamoGraphDeployment{}: {
+					UseCache:                            false,
+					UseStatusForStorageVersionMigration: true,
+				},
+				&nvidiacomv1beta1.DynamoGraphDeploymentRequest{}: {
+					UseCache:                            false,
```
carlory: runtimeConfig.ExcludedNamespaces is a business-controller event filter. It is not a controller-runtime cache namespace selector. 
The current cluster-wide manager cache therefore watches custom resources from all namespaces, including excluded namespaces.

However, this is only the current implementation behavior. I'm not unsure that it is a stable contract. If yes, it is better to use `cache`. cc @sttts

sttts: I think we should always migrate all namespaces, don't exclude any. Storage consistency is always owned by the cluster-wide operator globally.

deploy/operator/cmd/main.go
```diff
@@ -584,6 +586,34 @@ func registerControllers(
 	operatorImage string,
 	operatorPullPolicy corev1.PullPolicy,
 ) error {
+	if operatorCfg.Namespace.Restricted == "" {
+		migrator := &crdmigrator.CRDMigrator{
+			Client:    mgr.GetClient(),
+			APIReader: mgr.GetAPIReader(),
+			Config: map[client.Object]crdmigrator.ByObjectConfig{
+				&nvidiacomv1beta1.DynamoComponentDeployment{}: {
+					UseCache:                            false,
```
sttts: why the UseCache false?

docs/fern/pages/developer-guide/knowledge-base/kubernetes/kubernetes-operator/dynamo-operator.md
```diff
@@ -145,6 +145,37 @@ For user-focused workflows, see:
...
+leaves these completion markers unchanged. The controller retries after the dependency recovers.
+
+> [!IMPORTANT]
+> Configure GitOps tools to preserve or ignore
+> `crd-migration.cluster.x-k8s.io/observed-generation`. Removing the annotation repeatedly causes
+> the Operator to scan the configured resources again.
+
+A cluster-wide Operator with `dynamo-operator.upgradeCRD=false` still runs migration, but it does
+not change the CRD spec. It waits until the external CRD owner declares the new storage version.
+Namespace-restricted Operators do not run storage migration.
```
sttts: Maybe one liner about: downgrade of the operator is supported, nothing breaks.

deploy/operator/internal/crdmigrator/crd_migrator.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+func newTTLSet(ttl time.Duration) *ttlSet {
+	return &ttlSet{ttl: ttl, entries: map[string]time.Time{}}
+}
+
+func (s *ttlSet) Add(key string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.entries[key] = time.Now().Add(s.ttl)
```
sttts: **[P3] Reclaim expired migration-cache entries**

`Add` records an expiry, but expired entries are removed only when `Has` is called for that exact key. Because the key also includes `crd.Generation`, completion marks that generation observed, and later generations use different keys, successful entries remain in `entries` until process restart. This retains roughly one entry per migrated object per CRD generation; the upstream cache cited by this port runs a periodic sweep specifically to avoid indefinite growth. Could we add lifecycle-safe pruning (or clear completed-generation entries) and cover expiry without re-reading the same key?

carlory: Fixed in 7b0079ed8f by porting the periodically swept TTL cache from Cluster API v1.13.4 (util/cache/cache.go, commit 27f464418c195d96ae2ef4b96f3b6a047ea89310). The source provenance comments and maintenance notes now cover both the migrator and cache. I also added a regression test proving an expired entry is reclaimed by the sweep without re-reading that key; unit tests, race tests, and vet pass.

deploy/operator/cmd/main.go
carlory: Switched all four configured resources to UseCache: true in 7b0079ed8f. In cluster-wide mode the manager cache covers all namespaces; ExcludedNamespaces only filters business-controller events and does not narrow the cache. I added a comment at the registration site to make that dependency explicit.

deploy/operator/cmd/main.go
carlory: Agreed. The migrator remains registered only on the cluster-wide operator and migrates resources from all namespaces, including namespaces excluded from business-controller reconciliation. Addressed in 7b0079ed8f.

docs/fern/pages/developer-guide/knowledge-base/kubernetes/kubernetes-operator/dynamo-operator.md
carlory: Added the downgrade compatibility note in 7b0079ed8f: older Operator versions continue to access migrated resources through the served v1alpha1 API.

deploy/operator/internal/crdmigrator/crd_migrator.go
```diff
@@ -0,0 +1,388 @@
...
+		if err := r.Client.Status().Patch(ctx, crd, patch); err != nil {
+			return ctrl.Result{}, fmt.Errorf("patch CRD %s storedVersions: %w", crd.Name, err)
+		}
+	}
+
+	if r.phases.Has(CleanupManagedFieldsPhase) {
+		if err := r.cleanupManagedFields(ctx, crd, objects, cfg, storageVersion); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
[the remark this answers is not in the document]
carlory: Confirmed against Cluster API v1.13.4: upstream uses the same ordering and passes the pre-migration object list into managed-fields cleanup. A conflict is transient. If the cached re-read does not catch up within RetryOnConflict, the reconcile returns an error without writing the completion marker and controller-runtime requeues it. Because storedVersions was already pruned, the next reconcile skips the storage rewrite, lists the objects again, and retries cleanup from a fresh snapshot. I propose keeping the upstream ordering unchanged.

deploy/operator/internal/crdmigrator/crd_migrator.go
```diff
@@ -0,0 +1,388 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	// ObservedGenerationAnnotation records the CRD generation reconciled by the migrator.
+	ObservedGenerationAnnotation = "crd-migration.cluster.x-k8s.io/observed-generation"
+	storageMigrationCacheTTL     = time.Hour
```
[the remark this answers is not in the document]
carlory: I agree that we should decide this before release, but I would like maintainer input on whether we need a persistent annotation at all. One option is to use an NVIDIA-owned key such as nvidia.com/dynamo-crd-migration-observed-generation. Another option is to keep the observed CRD generation only in process memory, which avoids mutating the CRD and avoids creating a public GitOps preservation contract. The tradeoff is that an Operator restart loses that state and triggers one more scan. Storage migration itself remains guarded durably by status.storedVersions, so after a completed migration the restart path would skip rewrites and repeat only the idempotent managed-fields inspection and cleanup. The in-memory option also means there is no durable marker that managed-fields cleanup completed. @sttts, which ownership and persistence model do you prefer here?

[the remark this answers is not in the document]
carlory: Note: Dynamo’s built-in CRD apply path preserves the migration annotation. The crd-apply init container uses Server-Side Apply with the dynamo-crd-apply field manager. Because the migration annotation is written by the controller and is not part of the CRD apply configuration, subsequent CRD upgrades do not remove it.

docs/fern/pages/developer-guide/knowledge-base/kubernetes/kubernetes-operator/dynamo-operator.md
```diff
@@ -146,6 +146,39 @@ For user-focused workflows, see:
 - **[Managing Models with DynamoModel Guide](managing-models-dynamomodel.md)**
 - **[Snapshotting GPU Workers](snapshot.md)** for `DynamoCheckpoint`
 
+### CRD Storage Migration
```
sttts, line 149: I am undecided whether we actually want this section in the docs. CRD Storage Migration is built in a way that it is invisible. That's by design. It only become important if it does not succeed continuously, and eventually we are dropping v1alpha1. Then this would explode.

Maybe shrinking the section and expressing that narrative above would be good. Under normal circumstances the admin has nothing to do, but watching progress and success would be good.

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/pull/11904/commits/a28f3dd30414a01866c4bff3cf73d822134ade70

carlory:
/ok to test https://github.com/ai-dynamo/dynamo/pull/11904/commits/ba07d9423e68ed832a9f058664770240b8564ab0

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/commit/ba07d9423e68ed832a9f058664770240b8564ab0

deploy/operator/internal/crdmigrator/crd_migrator.go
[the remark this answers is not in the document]
carlory: Renamed the durable completion marker to crd-migration.nvidia.com/observed-generation in 40ecc14336 and updated the documented JSONPath. This preserves restart-safe migration state while using an NVIDIA-owned annotation domain.

sttts (approved):

Issue #8058, opened 2026-04-10
TCP response server always advertises bound IP, breaking multi-cluster and NAT'd deployments


fergusfinn:
When the frontend and workers run on different networks (separate clusters, across a VPN, behind a NAT), the TCP response path breaks silently. The frontend's `TcpStreamServer` auto-detects its pod IP and embeds it in `connection_info`. Workers try to connect back to that IP to deliver responses, but it's not reachable from their network.

The request plane already has `DYN_TCP_RPC_HOST` for controlling the advertised address, but the response server has no equivalent. There's also no way to separate bind and advertise addresses on either server — if the frontend is reachable through a load balancer or proxy at a different address than it binds to, there's no way to express that.

The result is that inference responses are generated successfully by workers but never delivered to the frontend. Requests hang until they time out.

It would be useful to have env vars for the response server (`DYN_TCP_RESP_HOST`, `DYN_TCP_RESP_ADVERTISE_HOST`, etc.) and advertise overrides for the request plane, so the server can bind locally but tell workers to connect to a routable address.

PR #11905, opened 2026-07-20
fix(runtime): skip link-local when binding TCP stream to interface

smb2love:
## Overview

Resolve TCP response-stream startup failures when a configured network interface has multiple addresses and a link-local address is selected.

## Summary

- preserve all addresses returned for the configured interface instead of collapsing them into a `HashMap`
- select the first routable unicast address, skipping loopback, link-local, multicast, unspecified, and broadcast addresses
- preserve explicit IP literals, including wildcard addresses
- use `IpAddr` and `SocketAddr` internally so IPv6 bind and advertised endpoints are formatted correctly

## Details

Interface-name resolution previously collected `(interface, address)` pairs into a `HashMap<String, IpAddr>`. Interfaces commonly have several IPv4 and IPv6 addresses, so this discarded all but one address and could leave an IPv6 link-local address. Binding that address without a scope ID fails on Linux.

The new resolution path scans all addresses for the requested interface and chooses the first routable unicast address. Explicit IP literals continue to bypass interface filtering.

`lib/runtime/src/pipeline/network/tcp/server.rs`, starting with `is_routable_unicast` and `resolve_interface_bind_ip`.

## Validation

- `cargo test -p dynamo-runtime --lib` — 489 passed, 2 ignored
- `cargo clippy -p dynamo-runtime --lib --tests -- -D warnings`
- generated format-patch applies cleanly to `origin/main`

## Related Issues
 #8058/#8059 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP server binding when configured with a network interface name.
  * Prevented binding to invalid or non-routable addresses, including loopback, multicast, unspecified, and link-local addresses.
  * Added support for IP address literals with more reliable listener address handling.
  * Improved error reporting when no suitable address is available for the selected interface.

### DIFF
--- a/docs/fern/pages/reference/components/runtime-configuration.mdx
+++ b/docs/fern/pages/reference/components/runtime-configuration.mdx
@@ -134,7 +134,8 @@ Unless a field is marked environment-only, it has both a CLI flag and an environ
 <ParamField path="DYN_TCP_RESPONSE_STREAM_HOST" type="string" default="auto-detected local address">
   IPv4 or IPv6 address, or exact network interface name, used to bind and advertise the TCP response
   stream server. If an interface name occurs more than once in the operating system's interface
-  list, Dynamo uses the last address. Dynamo trims whitespace and treats an empty value as unset.
+  list, Dynamo uses the last usable address, skipping link-local, multicast, broadcast, and
+  unspecified addresses. Dynamo trims whitespace and treats an empty value as unset.
   Unspecified addresses such as `0.0.0.0` and `::` are invalid.
 </ParamField>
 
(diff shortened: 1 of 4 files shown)


smb2love:
@GuanLuo @biswapanda @jthomson04 @ishandhanani — could you take a look?
Small fix in `lib/runtime/.../tcp/server.rs`: when `DYN_TCP_RESPONSE_STREAM_HOST` is an interface name, don't collapse addresses into a HashMap (often picks `fe80::/...` and bind fails with EINVAL). Prefer first routable unicast; also keep bind/advertise as `IpAddr`/`SocketAddr` so IPv6 gets brackets.
Happy to address feedback.

smb2love:
@GuanLuo @biswapanda @jthomson04 @ishandhanani  Would appreciate a quick review!

PR #11909, opened 2026-07-20
feat(operator): support Recreate strategy for managed rolling updates

julienmancuso:
## Summary

Add component-scoped `Recreate` strategy support to operator-managed DynamoGraphDeployment worker updates.

When a worker component specifies:

```yaml
podTemplate:
  metadata:
    annotations:
      nvidia.com/deployment-strategy: Recreate
```

the operator now:

1. Scales every old DCD generation for that component to zero.
2. Waits until each DCD has observed the scale-down and reports zero non-terminated replicas.
3. Scales the replacement DCD to the requested replica count.

Strategy resolution is shared between the DGD rollout coordinator and the generated Kubernetes Deployment, ensuring both layers interpret the annotation consistently.

`Recreate` is evaluated independently per component, so Recreate and RollingUpdate workers can progress concurrently within the same DGD. The `maxSurge` and `maxUnavailable` annotations are ignored for Recreate components.

Existing RollingUpdate behavior and the fallback for missing or unrecognized strategy values remain unchanged.

The rolling-update documentation now describes Recreate configuration, scope, and availability implications.

## Motivation

Operator-managed updates create separate DCD generations. Setting the Kubernetes Deployment strategy to Recreate only controls pods within one DCD; it does not prevent old and new DCD generations from running concurrently.

This change adds cross-generation Recreate semantics for deployments that cannot temporarily surge—particularly GPU-constrained clusters—or where old and new worker generations must not overlap.

## Safety and correctness

The replacement generation remains scaled to zero until every old DCD reports:

- `spec.replicas: 0`
- `status.observedGeneration >= metadata.generation`
- `status.component.replicas: 0`

This prevents stale cache observations, pending controller writes, absent status, or a remaining older DCD generation from triggering an early cutover.

Recreate intentionally introduces an availability gap while old workers stop and new workers start.

## Validation

- `go test ./internal/controller -count=1`
- `go test ./api/... ./internal/... -count=1`
- `go vet ./...`
- `fern check`
- `fern docs broken-links`
- `git diff --check`

Test coverage includes:

- Default, explicit RollingUpdate, Recreate, and unrecognized strategy values
- Pending and stale scale-down observations
- Missing component status
- Remaining non-terminated replicas
- Default replica handling
- Multiple old DCD generations
- Holding the replacement generation at zero during drain
- Starting the replacement after drain completion
- Independent Recreate and RollingUpdate component progression
- Ignoring surge configuration for Recreate

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11909" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added component-level `Recreate` deployment strategy support for worker updates.
  - Existing workers now scale down fully before new workers start, helping prevent overlap during replacement.
  - Unrecognized or missing strategy settings continue to use `RollingUpdate` by default.

- **Documentation**
  - Documented configuration, behavior, availability considerations, and rolling-update parameter support for the new strategy.


docs/kubernetes/rolling-update.md
```diff
@@ -326,6 +326,44 @@ VllmDecodeWorker:
...
+      type: worker
+      replicas: 1
+      podTemplate:
+        metadata:
+          annotations:
+            nvidia.com/deployment-strategy: Recreate
+        spec:
+          containers:
+            - name: main
+            image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.2.1
```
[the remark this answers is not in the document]
julienmancuso: fixed

deploy/operator/internal/controller/dynamographdeployment_rollingupdate.go
```diff
@@ -1231,18 +1249,35 @@ func (r *DynamoGraphDeploymentReconciler) buildRollingUpdateContext(
...
-		maxScaledDown := max(int32(0), (oldState.Spec+newState.Spec)-minAvailable-newUnavailable)
-		oldUnhealthy := max(int32(0), oldState.Spec-oldState.Available)
-		// availableSurplus is how many extra available replicas we have above minAvailable (min 0)
-		availableSurplus := max(int32(0), (oldState.Available+newState.Available)-minAvailable)
-		oldTarget := max(int32(0), oldState.Spec-min(maxScaledDown, oldUnhealthy+availableSurplus))
-
-		// Surge budget uses Spec (declared intent) like K8s Deployment controller; scheduler enforces actual resource constraints.
-		scaleUpBudget := max(int32(0), desired+maxSurge-oldState.Spec-newState.Spec)
-		newTarget := min(desired, newState.Spec+scaleUpBudget)
+		annotations := dynamo.GetPodTemplateAnnotations(spec)
```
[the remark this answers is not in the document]
julienmancuso: fixed

deploy/operator/internal/controller/deployment_strategy.go
```diff
@@ -0,0 +1,25 @@
...
+
+package controller
+
+import "github.com/ai-dynamo/dynamo/deploy/operator/internal/common"
+
+const (
+	KubeAnnotationDeploymentStrategy                    = "nvidia.com/deployment-strategy"
+	KubeAnnotationDeploymentRollingUpdateMaxSurge       = "nvidia.com/deployment-rolling-update-max-surge"
+	KubeAnnotationDeploymentRollingUpdateMaxUnavailable = "nvidia.com/deployment-rolling-update-max-unavailable"
+)
```
sttts, line 14: devil's advocate: why are these not real fields?

sttts: nvmd, not for this RP anyway. But maybe something to think about.

deploy/operator/internal/controller/dynamographdeployment_rollingupdate.go
```diff
@@ -812,6 +813,26 @@ func (r *DynamoGraphDeploymentReconciler) getOldWorkerDCDsByComponent(
...
+// unprocessed write or absent status as a completed drain.
+func oldWorkerDCDsAtZero(dcds []*nvidiacomv1beta1.DynamoComponentDeployment) bool {
+	for _, dcd := range dcds {
+		if dcd == nil || dcd.Spec.Replicas == nil || *dcd.Spec.Replicas != 0 {
+			return false
+		}
+		if dcd.Status.ObservedGeneration < dcd.Generation {
+			return false
+		}
+		if dcd.Status.Component == nil || dcd.Status.Component.Replicas != 0 {
```
sttts, line 857: we cannot trust `dcd.Status.Component.Replicas`. It only says that there is no non-terminating pod.

Compare https://github.com/kubernetes/kubernetes/blob/v1.36.2/pkg/controller/deployment/recreate.go#L74-L103 for how deployment reconciler checks the pods of the replicaset manually to implement recreate semantics.

julienmancuso: fixed

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -2763,6 +2763,11 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 				GenericFunc: func(ge event.GenericEvent) bool { return true },
 			}),
 		).
+		Watches(
```
tmonty12, line 2777: instead of having this watch for terminal phase DGD pods that seems brittle, could we instead just requeue when there are more than one pods in terminating phase?

julienmancuso: The Pod event is only used to wake the level-driven reconcile. the barrier always re-lists pods from the cache. We watch deletion and non-terminal-to-terminal transitions, which are exactly the events that can clear it, while creation/membership changes close races.
Requeueing would require polling while blocked and would be less immediate. Is there a specific Pod transition you think the predicate misses?

sttts: Number of pods can be higher, e.g. hundreds or more. Pods are huge objects. So listing them is expensive, and we cannot filter by deletion. So I think polling is not a good idea. In contrast, an informer means lots of memory potentially. 

One can prune the pods, controller-runtime has some feature for that (transformation?), deleting spec and status. Actually, there is also a mode to watch only for metadata via some MIME type I believe. That's likely the best approach.

About brittleness, watches have guarantees to always get event. The guarantees are strong. No risk there.

julienmancuso: The operator already keeps full Pod information in memory for other features (topology_label_controller.go)
Our new watch simply reuses that existing information, so it adds almost no extra memory. A metadata-only watch would create another separate copy containing only Pod names and labels and that copy would not include the Pod phase we need to know whether it has actually stopped.
So:
- Current approach: reuse existing Pod data and check whether each Pod stopped.
- Metadata-only approach: create another cache, then still need another way to obtain the missing Pod phase.

sttts: Reuse is good. But does the other usecase need more than metadata either?

sttts: I see https://github.com/ai-dynamo/dynamo/issues/11983 :+1:

julienmancuso: implemented as part of [perf(operator): reduce memory retained by the shared Pod cache](https://github.com/ai-dynamo/dynamo/pull/12014#top)

deploy/operator/internal/controller/dynamographdeployment_rollingupdate.go
```diff
@@ -812,6 +813,69 @@ func (r *DynamoGraphDeploymentReconciler) getOldWorkerDCDsByComponent(
...
+	podList := &corev1.PodList{}
+	if err := r.List(
+		ctx,
+		podList,
+		client.InNamespace(dgd.Namespace),
+		client.MatchingLabels{
+			consts.KubeLabelDynamoGraphDeploymentName: dgd.Name,
+			consts.KubeLabelDynamoComponent:           componentName,
+		},
+	); err != nil {
```
sttts, line 962: Are we sure here this is supported by an index?

Example: two DGDs, each 1000 pods. Will the terminating one keep listing 1000+N pods here?

sttts: Codex:

This is cache-backed, but not label-indexed.

The reconciler uses `mgr.GetClient()`, and the Pod watch ensures that the shared Pod informer is registered and synchronized before reconciliation starts. Therefore, this `List` is served from the controller-runtime Pod cache rather than directly from the API server.

However, `client.MatchingLabels` does not automatically create or use label indexes. controller-runtime first uses its built-in namespace index for `client.InNamespace(...)`, retrieves all cached Pods in that namespace, and then evaluates both label requirements in memory.

The complexity is therefore `O(number of Pods in the namespace)` per call. If that scan is too broad, we would need to register an explicit composite field index for the DGD and component and query it using `MatchingFields`.

julienmancuso: yes good catch, fixed in 5b49d4695483b3c8f8d8bbd5bb70829b5903e3c7 by adding and using a new index

sttts:
### P2 — Keep zero-replica Recreate rollouts open through Pod termination

This path requires a **hash-relevant worker template change while the desired replica count is zero**. A replicas-only change does not create a new worker hash.

For example, assume the existing generation uses `worker:v1` and a template update changes the desired generation to `worker:v2` while the component is scaled to zero:

```yaml
# Desired state on the DynamoGraphDeployment
kind: DynamoGraphDeployment
metadata:
  name: example
spec:
  components:
    - name: worker
      type: worker
      replicas: 0
      podTemplate:
        metadata:
          annotations:
            nvidia.com/deployment-strategy: Recreate
        spec:
          containers:
            - name: main
              image: worker:v2
```

The managed rollout represents the old and new worker generations as separate, hash-named DCDs. They temporarily coexist during the rollout; they do **not** have identical hash-relevant specs:

```yaml
# Existing generation: oldhash is derived from the worker:v1 template
kind: DynamoComponentDeployment
metadata:
  name: example-worker-oldhash
  labels:
    nvidia.com/dynamo-worker-hash: oldhash
spec:
  name: worker
  type: worker
  replicas: 0
  podTemplate:
    spec:
      containers:
        - name: main
          image: worker:v1
status:
  component:
    replicas: 0
    readyReplicas: 0
```

```yaml
# Replacement generation: newhash is derived from the worker:v2 template
kind: DynamoComponentDeployment
metadata:
  name: example-worker-newhash
  labels:
    nvidia.com/dynamo-worker-hash: newhash
spec:
  name: worker
  type: worker
  replicas: 0
  podTemplate:
    spec:
      containers:
        - name: main
          image: worker:v2
status:
  component:
    replicas: 0
    readyReplicas: 0
```

Although the old DCD reports zero replicas, one of its Pods can still be non-terminal during its deletion grace period:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: example-worker-oldhash-abcde
  deletionTimestamp: "2026-07-24T12:00:00Z"
  labels:
    nvidia.com/dynamo-graph-deployment-name: example
    nvidia.com/dynamo-worker-hash: oldhash
    nvidia.com/selector: example-worker-oldhash
status:
  phase: Running
```

`continueRollingUpdate` calls `completedWorkerComponents`, which decides completion from the DGD desired count and the two DCD status objects:

```yaml
completionEvaluation:
  desiredReplicasFromDGD: 0
  newGenerationFromDCDStatus:
    readyReplicas: 0
    newReady: "readyReplicas >= desiredReplicas => 0 >= 0 => true"
  oldGenerationFromDCDStatus:
    readyReplicas: 0
    oldGone: "readyReplicas == 0 => true"
  result:
    componentUpdated: true
```

This completion path does not check the Pod shown above. Because `continueRollingUpdate` runs before `reconcileResources`/`buildRollingUpdateContext`, the Pod-aware `oldWorkerPodsTerminated` barrier in the Recreate replica planner cannot prevent this completion. `completeRollingUpdate` can therefore delete the old DCD and record `newhash` as the current worker hash while the old Pod still has `status.phase: Running`.

A later replicas-only scale-up keeps the `worker:v2` template unchanged:

```yaml
kind: DynamoGraphDeployment
metadata:
  name: example
spec:
  components:
    - name: worker
      type: worker
      replicas: 1
      podTemplate:
        spec:
          containers:
            - name: main
              image: worker:v2
```

`replicas` is intentionally excluded from the worker hash, so this remains `newhash`. Since `newhash` was already recorded as current, `buildRollingUpdateContext` takes its current-hash fast path and does not evaluate `oldWorkerPodsTerminated` again. The replacement Pod may consequently start before the old Pod exits, violating the Recreate no-overlap contract.

Before completing a rollout, require `oldWorkerDCDsAtZero` and `oldWorkerPodsTerminated` for every Recreate component, including components whose desired replica count is zero. Keep the rollout active and retain the old hash until that condition succeeds.

Please add a state-machine test covering:

```yaml
testSequence:
  - desired: {image: worker:v2, replicas: 0}
    oldDCD: {hash: oldhash, readyReplicas: 0}
    newDCD: {hash: newhash, readyReplicas: 0}
    oldPod: {deletionTimestamp: set, phase: Running}
    expectedRolloutPhase: InProgress
  - oldPod: {phase: Succeeded}
    expectedRolloutPhase: Completed
  - desired: {image: worker:v2, replicas: 1}
    expectedWorkerHash: newhash
    expected: replacement starts only after the old Pod is terminal
```

sttts (approved):

PR #11915, opened 2026-07-20
fix(bindings): keep discovery alive while HTTP service drains in-flight requests

jacksonrnewhouse:
## Summary

When a process serving the OpenAI-compatible frontend (`HttpService` from the Python bindings) receives SIGTERM, `Runtime::shutdown` Phase 2 waits on the `GracefulShutdownTracker` — but only component push endpoints ever register with it. A frontend-only process therefore has an empty tracker, and Phase 3 cancels the primary token milliseconds after SIGTERM, while axum is still draining in-flight connections (Phase 1 intentionally keeps accepting for `DYN_HTTP_GRACEFUL_SHUTDOWN_TIMEOUT_SECS`, then drains without bound).

The primary-token cancel tears down the KV-store discovery watch mid-drain. The watch exit path publishes an empty instance list while the router's `routing_instances` snapshot stays frozen, so the draining frontend attempts to route to a router it can still name but can no longer resolve, and fails in-flight requests with `Instance <id> not found and no other instances available` → HTTP 529, even though the router is healthy. The router-side request guard's `mark_free` fails on the same path, leaking queued-token admission until the 300s stale-sequence expiry.

The window is proportional to request duration, so short-request deployments only see a rare race — but with long-context traffic (multi-minute prefills queued at the router), every frontend scale-down or rollout fails live requests.

Fix: register the HTTP service with the graceful-shutdown tracker for the lifetime of its serve+drain future. Phase 2 then holds — bounded by `DYN_RUNTIME_GRACEFUL_SHUTDOWN_TIMEOUT_SECS` — until in-flight requests finish, and discovery/backend connections outlive the drain. One guard, no new API.

Note for operators: `terminationGracePeriodSeconds` and `DYN_RUNTIME_GRACEFUL_SHUTDOWN_TIMEOUT_SECS` must be sized above the longest expected stream for the drain this enables to complete.

## Validation

Reproduced and validated on production-shaped Kubernetes deployments (GLM-class model, TensorRT-LLM backend, split frontend/router/worker, 200K–1M-token requests):

- Without the fix: killing a frontend pod with requests in flight produced the discovery-teardown log signature (`KeyValueStoreManager.watch failed`, `No KV router workers available`, `router request guard failed to free request`, `Instance not found`) and HTTP 529s; the SIGTERM→Phase 3 gap was under 1ms.
- With the fix: deleted **both** frontend pods with four in-flight ~600K-token requests — all four completed HTTP 200 with correct outputs; each pod held Phase 2 for exactly as long as its remaining streams needed (83s and 161s), then completed Phase 3 teardown normally.
- `cargo check` / `cargo fmt` clean on `lib/bindings/python`; existing shutdown-phase unit test semantics unchanged (the tracker path this uses is already covered by `shutdown_cancels_main_token_after_graceful_timeout`).

Known follow-ups (intentionally out of scope): the discovery `endpoint_watcher` publishes an empty instance list on local cancellation (turns "my watch died" into "cluster empty"), and the router request-guard leak could use a liveness lease rather than the 300s expiry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown handling for the HTTP server.
  * In-flight requests now receive additional time to complete during server shutdown.


rmccorm4:
Thanks for the contribution @jacksonrnewhouse - looks like this was covered in https://github.com/ai-dynamo/dynamo/pull/12031 which was already merged

PR #11917, opened 2026-07-20
feat(mocker): add live request engine

connorcarpenter15:
## Overview:

Add a reusable, protocol-neutral `LiveEngine` that lets network-facing adapters drive Mocker through request-scoped submission, streaming, cancellation, metrics, and bounded response buffering.

This shared boundary will support the vLLM and SGLang mock gRPC servers. It is also intended to become the scheduler-facing implementation behind the Mocker backend in `lib/llm`.

## Details:

- Expose request-scoped Mocker submission, output streams, targeted cancellation, scheduler metrics, and stream-drop cleanup.
- Run the vLLM and SGLang schedulers through the shared live-loop boundary.
- Isolate caller-provided request IDs from internal scheduler IDs so cancellation and ID reuse cannot misroute stale output.
- Bound retained response signals globally while preventing a slow reader from blocking unrelated requests.
- Preserve focused coverage for streaming, shutdown, cancellation races, duplicate IDs, backpressure, and capacity cleanup.
- Validated with:
  - `cargo metadata --no-deps --format-version 1`
  - `cargo test -p dynamo-mocker` (539 passed, 1 ignored)

- `lib/mocker/src/live.rs` for the public `LiveEngine` API and request lifecycle.
- `lib/mocker/src/scheduler/live_boundary.rs` for the shared scheduler integration.
- `lib/mocker/src/live/tests.rs` for the focused lifecycle and backpressure coverage.

## Related Issues

- Closes [DIS-2473: Expose Mocker as a live request engine for sidecar adapters](https://linear.app/nvidia/issue/DIS-2473/expose-mocker-as-a-live-request-engine-for-sidecar-adapters)
- Follow-up: [DIS-2478: Adapt the `lib/llm` Mocker backend to `LiveEngine`](https://linear.app/nvidia/issue/DIS-2478/adapt-the-libllm-mocker-backend-to-liveengine)
- Prerequisite for #11873.


lib/mocker/src/live.rs
```diff
@@ -0,0 +1,695 @@
...
+            .context("declared output length exceeds the live response budget representation")?;
+        anyhow::ensure!(
+            output_capacity <= self.inner.output_budget_capacity,
+            "declared output length {output_capacity} exceeds the live response budget"
+        );
+        let output_budget = self
+            .inner
+            .output_budget
+            .clone()
+            .acquire_many_owned(output_permits)
```
PeaBrane: Could we add a short `NOTE` here explaining that this intentionally reserves the request’s entire declared response until `LiveRequest` is dropped? That reservation lets every admitted response be buffered without a slow reader blocking unrelated admitted requests while keeping worst-case memory bounded. Releasing permits per `recv()` would break that guarantee unless slow clients were cancelled or outputs dropped. Renaming `MAX_BUFFERED_OUTPUT_SIGNALS` to `MAX_RESERVED_OUTPUT_SIGNALS` would make the policy especially clear.

lib/mocker/src/live/tests.rs
```diff
@@ -0,0 +1,281 @@
...
+                output_token_ids: Some(vec![11; 100]),
+                uuid: Some(uuid),
+                ..Default::default()
+            })
+            .await
+            .unwrap();
+        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+
+        assert!(
+            tokio::time::timeout(std::time::Duration::from_secs(1), engine.cancel(uuid))
```
PeaBrane: Could this test deterministically queue an old output batch before reusing the client ID? Because cancellation uses `discard_pending_output=true`, the boundary suppresses pending old signals, so this can pass even if `scheduler_id` regresses to `client_id`. Pausing the dispatcher, establishing that the old terminal batch is queued, removing the old route and reusing the client ID, then reopening dispatch and asserting that only token `22` reaches the replacement would exercise the intended isolation.

lib/mocker/src/live/tests.rs
```diff
@@ -0,0 +1,281 @@
...
+        });
+        tokio::pin!(pending);
+        assert!(
+            tokio::time::timeout(std::time::Duration::from_millis(20), &mut pending)
+                .await
+                .is_err(),
+            "submission should wait rather than exceed the global output budget"
+        );
+
+        drop(slow);
```
PeaBrane: Could we add separate coverage that dropping the stream actually removes scheduler state and the route, rather than only releasing the semaphore permit? `_output_budget` is released independently, so this test could pass even if `spawn_cancellation` stopped running. A long active request followed by stream drop, observed route/occupancy cleanup, and successful reuse of the same client ID would cover that contract. A companion case aborting `submit()` while its command is deferred would also exercise the delicate `Submitting → cancellation after admission` path.

PeaBrane (approved):
✅ The LiveEngine boundary reuses the existing scheduler infrastructure cleanly and keeps replay isolated. I left three non-blocking comments to document the reservation invariant and strengthen lifecycle regression coverage.

lib/mocker/src/live.rs
jthomson04: I think the invariant itself needs reconsideration, not only documentation. This makes LiveEngine admission depend on the sum of callers’ declared maximum output lengths. With 32,768 permits, one request declaring 32,768 output tokens excludes every other request until its `LiveRequest` is dropped; 8,192-token requests are limited to four. Tokio’s FIFO-fair `acquire_many` can also head-of-line block smaller requests that would otherwise fit. These requests never reach the Mocker scheduler, so its capacity controls and metrics cannot observe them. Could we bound actual queued output instead—for example, small per-route buffers with `try_send` and cancellation of slow consumers? A NOTE and rename would clarify the policy, but wouldn’t remove this second admission-control layer.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -889,5 +607,271 @@ impl LiveEffectsPublisher {
...
+        let _ = publisher
+            .apply_cancellation(
+                core,
+                cancellation,
+                true,
+                scheduler_elapsed_ms(scheduler_start),
+            )
+            .await;
+    }
+    while let Ok(command) = command_rx.try_recv() {
```
jthomson04: These drain-until-empty loops can starve scheduler execution under sustained submissions. `apply_command` wakes a submitter, which can refill `command_rx` before `try_recv()` observes it empty, so the bounded channel does not bound the work performed per pass. Could we process a fixed snapshot or command budget per cycle so the scheduler is guaranteed a turn? The same pattern appears in `drain_live_control`.

lib/mocker/src/live.rs
```diff
@@ -0,0 +1,695 @@
...
+        let client_id = request.uuid.unwrap_or_else(Uuid::new_v4);
+        let scheduler_id = Uuid::new_v4();
+        request.uuid = Some(scheduler_id);
+        // A request emits at most one signal per effective output token (or one
+        // rejection signal). The engine-wide semaphore bounds the sum of these
+        // per-request channels while allowing the shared dispatcher to keep
+        // unrelated routes moving when one client reads slowly.
+        let (tx, rx) = mpsc::channel(output_capacity);
+        let route = Arc::new(RequestRoute::new(client_id, scheduler_id, tx));
+        match self.inner.routes.by_client.entry(client_id) {
```
jthomson04, line 319: The duplicate-ID check happens only after waiting for the output permits. If the existing request holds the available budget, submitting the same ID blocks instead of immediately returning “already active.” Could we atomically reserve/check the client ID before waiting, and roll that reservation back if admission fails?

lib/mocker/src/live.rs
```diff
@@ -0,0 +1,695 @@
...
+            Some(SchedulerOutputSender::Bounded(output_tx)),
+            KvEventPublishers::default(),
+            Some(cancel.clone()),
+            FpmPublisher::default(),
+        );
+        let command_tx = scheduler.command_sender();
+        let cancellation_tx = scheduler.cancellation_sender();
+        let metrics_rx = scheduler.metrics_receiver();
+        let routes = Arc::new(RequestRoutes::default());
+        runtime.spawn(run_output_dispatcher(
```
jthomson04: This dispatcher’s `JoinHandle` is discarded. If the task panics or exits unexpectedly while the engine remains alive, the scheduler will eventually block on the bounded output channel while new submissions can still be accepted without a response consumer. Could we supervise the task and cancel or mark the engine unhealthy on unexpected termination?

lib/mocker/src/scheduler/live_boundary.rs
PeaBrane: I think we may be able to defer this unless the patch is easy. We generally don't really have hygenic starvation handling. But can be patched if a soak exposes this @connorcarpenter15 to confirm this

lib/mocker/src/live.rs
connorcarpenter15: Removed the global reservation. Routes now use fixed buffers with `try_send`, and full streams are cancelled, leaving admission to the scheduler.

lib/mocker/src/live/tests.rs
connorcarpenter15: Updated the test to queue old output, reuse the client ID, reopen dispatch, and assert only token `22` arrives.

lib/mocker/src/live/tests.rs
connorcarpenter15: Added both cases: active-stream drop verifies cleanup and ID reuse; deferred-submit abort verifies cancellation after admission.

lib/mocker/src/scheduler/live_boundary.rs
connorcarpenter15: Kept changes small so external lanes process an entry snapshot per cycle, and scheduler-owned deferred commands drain fully.

lib/mocker/src/live.rs
connorcarpenter15: The permit wait is gone. Client IDs are reserved before the first await, so duplicates fail immediately.

lib/mocker/src/live.rs
connorcarpenter15: The dispatcher is now supervised. Unexpected exit cancels the engine, closes routes, and rejects new submissions.

jthomson04 (approved):

PeaBrane (approved):

PR #11920, opened 2026-07-20
perf(router): batch active sequence replica events

jthomson04:
## Summary

Batch frontend active-sequence replica-sync lifecycle events on ZMQ while preserving singleton NATS publication and the existing receive-side coalescing behavior.

## Overview:

Reduce ZMQ message fanout by collecting lifecycle events with the shared 256-event and 1 ms policy. NATS keeps the existing ActiveSequenceEvent wire format, so no batch payload-size cap or partitioning is required.

## Details:

- Select singleton or batched publication from the runtime event transport.
- Publish each NATS lifecycle event immediately with the existing singleton wire format.
- Send ZMQ lifecycle events through the existing 100,000-event bounded queue and publish one ActiveSequenceEventBatch per collection window.
- Decode NATS singletons and ZMQ batches through one shared transport-aware subscriber used by replica apply and LoRA load tracking.
- Preserve the public by-reference SequencePublisher method and add a default owned helper; only the runtime overrides the owned path to avoid cloning large AddRequest token vectors.
- Keep receive-side ReplicaBatchEffects coalescing, best-effort failures, and standalone slot-tracker wire behavior unchanged.

- lib/llm/src/kv_router/sequence.rs for transport selection, ZMQ batching, and shared subscriber flattening.
- lib/llm/src/lora/load_estimator.rs for reuse of the shared subscriber and the LoRA regression test.
- lib/kv-router/src/sequences/multi_worker.rs for the source-compatible owned-event helper.

## Related Issues

## Validation

- cargo fmt --all -- --check
- cargo test -p dynamo-llm kv_router::sequence::tests --lib
- cargo test -p dynamo-llm mixed_replica_batch_updates_lora_load_in_order --lib
- cargo test -p dynamo-kv-router replica_sync --lib
- cargo test -p dynamo-kv-router --features standalone-slot-tracker scoped_publisher_drops_when_outbound_channel_is_full --lib
- cargo clippy -p dynamo-kv-router -p dynamo-llm -p dynamo-mocker --lib --tests -- -D warnings
- cargo clippy -p dynamo-kv-router --features standalone-slot-tracker --lib --tests -- -D warnings

### DIFF
--- a/lib/llm/src/lora/load_estimator.rs
+++ b/lib/llm/src/lora/load_estimator.rs
@@ -743,6 +743,12 @@ impl Default for LoadEstimator {
 mod tests {
     use super::*;
     use dynamo_kv_router::protocols::{ActiveSequenceEventBatch, WorkerWithDpRank};
+    use dynamo_runtime::config::environment_names::zmq_broker as broker_env;
+    use dynamo_runtime::distributed::DistributedConfig;
+    use dynamo_runtime::transports::event_plane::EventPublisher;
+    use dynamo_runtime::{DistributedRuntime, Runtime};
+
+    use crate::kv_router::ACTIVE_SEQUENCES_SUBJECT;
 
     fn lora_event(
         request_id: &str,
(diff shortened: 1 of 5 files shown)


jthomson04:
## Code review

Reviewed the batched replica-sync path end to end (transport wiring, partitioning, paused-time tests, receiver apply loop). Overall: correct, well-tested, clear throughput win (up to 256:1 fewer NATS publishes). No blocking bugs. Notes below.

### What's solid
- **Envelope handling is correct.** `publish(&T)` == `encode_payload` + `publish_bytes_ref`, so the manual `encode_replica_batch` + `publish_bytes_ref` path is equivalent — the manual encode is justified because the encoded size is needed to partition. The typed `ActiveSequenceEventBatch` decodes bytes serialized from `ActiveSequenceEventBatchRef` identically; the roundtrip test confirms it.
- **Partitioning is correct.** MessagePack prefix length is monotonic in event count, so the binary search over `[2, len-1]` (with `best_len = 1` seed and the full-slice / single-event fast paths) always finds the largest fitting prefix. `len ∈ {1,2,3,4}` edge cases check out.
- **Clean lifecycle** — on drop the channel closes and the publisher task exits; `child_token()` covers cancellation. `.first()/.last().expect()` are safe since `collect_replica_batch` always seeds `first_event`.
- **Strong unit coverage** — time/count caps under `start_paused`, payload split + envelope-under-1 MB assertion, single-oversized-event preservation.

### Issues & risks
- **The "ordered" framing overstates the guarantee (pre-existing, not a regression).** `spawn_publish_event` does `tokio::spawn(publish_event(...))` *per event*, so the order events reach the mpsc channel is decided by task scheduling, not by the logical `add_request` → `free` order. Intra-batch order is faithfully preserved, but that order is only "channel arrival order," which was already nondeterministic before this PR. Batching neither fixes nor worsens it — worth softening the "ordered" language, and confirming `apply_replica_event` is order-tolerant (it appears to be).
- **NATS `max_payload` is a hardcoded assumption.** `MAX_REPLICA_BATCH_PAYLOAD_BYTES = 1_000_000` targets the NATS default 1 MB; the test only guards that default. A deployment that lowers `max_payload` would have well-formed batches rejected at publish time. Consider reading `client.server_info().max_payload`, or documenting the assumption. (Also dead overhead on the ZMQ transport, which has no such limit — harmless.)
- **Oversized single events and mid-batch failures are silently lossy.** An oversized event is published intact (warn only) and will be rejected by NATS → logged as error, no retry → that request's state diverges on the peer. `publish_replica_batch` likewise continues past a publish error, leaving gaps. Both are pre-existing best-effort semantics, but batching makes them explicit fallbacks — a one-line comment stating replica sync is lossy-by-design would help the next reader.
- **Extra per-event allocation.** `publish_event` now does `event.clone()` to move into the channel (trait takes `&ActiveSequenceEvent`), where the old path serialized from the reference. For `AddRequest` with a large `token_sequence` this is a real heap copy on the hot path, on top of the retained per-event `tokio::spawn`. Likely a net win given the publish reduction, but a genuine added cost.

### Nits
- `REPLICA_EVENT_CHANNEL_CAPACITY = 100_000` and `MAX_REPLICA_BATCH_PAYLOAD_BYTES` have no rationale comment; 100k buffered events (each potentially carrying a large `token_sequence`) is a nontrivial memory ceiling under backpressure.
- Constants are split across crates (`MAX_REPLICA_BATCH_{EVENTS,DURATION}` in `protocols.rs`, payload/channel caps in `sequence.rs`) — the crate boundary justifies it, but a cross-reference comment would aid discoverability.
- `encode_replica_batch_partitions` re-encodes the full remaining slice each outer iteration → O(n log n) encodes in the pathological all-large-events case. Fine given the fast path dominates.

_Reviewed with Claude Code._

jthomson04:
Addressed the review feedback in 77d4e327be.

- Batching is now ZMQ-only. NATS keeps publishing one ActiveSequenceEvent per message, preserving its existing wire format and eliminating the batch max-payload and partitioning concerns.
- ZMQ retains the 256-event / 1 ms collection policy and the existing 100,000-event bounded queue. The pre-existing spawned publisher tasks await queue capacity rather than deliberately dropping events.
- Replica apply and LoRA tracking now share one transport-aware subscriber: singleton decoding for NATS and batch decoding plus event flattening for ZMQ. This fixes the LoRA wire mismatch without duplicating transport-selection logic.
- The public SequencePublisher::publish_event(&ActiveSequenceEvent) API remains source-compatible. A default owned helper was added, and only RuntimeSequencePublisher overrides it to move large AddRequest payloads into the ZMQ queue without cloning.
- Batch wording now states the actual guarantee: publisher-queue arrival order. Logical ordering across the existing spawned producer tasks remains unchanged.
- Manual MessagePack size checks, binary-search partitioning, oversized-event handling, and their tests were removed.

Validation is green: formatting; focused NATS-wire/ZMQ-batching, LoRA, and 13 replica-apply tests; the standalone publisher regression with its feature enabled; and clippy with warnings denied for dynamo-kv-router, dynamo-llm, dynamo-mocker, and standalone slot-tracker.

lib/llm/src/lora/load_estimator.rs
```diff
@@ -745,6 +742,60 @@ impl Default for LoadEstimator {
...
+        ActiveSequenceEventData::AddRequest {
+            token_sequence: None,
+            track_prefill_tokens: false,
+            expected_output_tokens: None,
+            prefill_load_hint: None,
+        }
+    }
+
+    #[test]
+    fn mixed_replica_batch_updates_lora_load_in_order() {
```
PeaBrane: Non-blocking: this test bypasses the subscription path it is meant to protect—it unwraps `batch.events` and calls the pre-existing `handle_event` directly. It would still pass if `RuntimeSequenceSubscriber` failed to decode or flatten a ZMQ batch, which was the regression fixed earlier in this PR. Could we drive a batch through the shared subscriber and assert the final LoRA counts instead?

PeaBrane (approved):
✅

lib/llm/src/lora/load_estimator.rs
jthomson04: Addressed in 9f3cb73fca. `mixed_replica_batch_updates_lora_load_in_order` now uses a process-local ZMQ endpoint, publishes an `ActiveSequenceEventBatch` through the production event publisher, receives the flattened lifecycle events through `RuntimeSequenceSubscriber`, and then verifies the LoRA counts. The focused LoRA and sequence tests pass, as does clippy.

ishandhanani (approved):

Issue #11899, opened 2026-07-20
DEP: Scale direct-ZMQ KV-event ingestion across publishers


PeaBrane:
## Summary

Direct-ZMQ KV-event ingestion stops scaling well before the recovery-aware consumer funnel or concurrent radix-tree indexer. A replay campaign on current `main` localized the saturated ceiling to the combined low-level ZMQ publish, socket receive, merged-stream fan-in, and envelope/payload decode segment before `WorkerQueryClient::handle_live_batch`.

This DEP proposes landing the stage-cut benchmark and ingress observability, then making direct-ZMQ ingestion concurrency-preserving across publishers while retaining FIFO order within each `(worker_id, dp_rank)` source. Follow-up stage isolation found that receive, decode, and concurrent `WorkerQueryClient` admission are each fast in isolation; the remaining ceiling appears when they share an oversized runtime/executor budget. The preferred production direction is therefore direct per-source delivery on a modest, explicit event-plane executor, not another aggregate stream or a large sharded queue layer.

## Motivation

`DynamicSubscriber` creates one receive task per discovered publisher, but all tasks `try_send` raw envelopes into one bounded MPSC stream. Envelope decode, payload decode, and the recovery-aware consumer then run after that merge. The downstream indexer already preserves source ordering while distributing unrelated workers across flume/CRTC lanes, so serializing unrelated publishers before that point can leave available indexer concurrency unused.

### Benchmark setup

- Bases: initial baseline/proof `f48f8182462f79dceefc5ac156c174f631b704e7`; stage/cardinality/topology/shard follow-up `74b79d04bfde03c83198b706785a767bacb659d7` (then-current `origin/main`).
- Host: exclusive 144-core Neoverse-V2 CPU node, two NUMA nodes; Rust 1.96.1 optimized bench profile.
- Trace: pinned Mooncake JSONL, SHA-256 `b434f1816a707f4bac697235588184ebc374c9907cb981bb65fb0643471fe711`.
- Corpus: 256 logical publishers, 102,788 store events, 1,012,008 blocks, approximately 70 MB of pre-encoded payload, 512-token blocks.
- Normal headline configuration: 64 events/envelope, 32 indexer workers, 64 Tokio workers, 3 repetitions.
- Recovery traffic was disabled, but sources used the production `ActiveLiveOnly` path, including publisher lookup, sequence tracking, and the live source mutex.
[...]

PR #11925, opened 2026-07-20
perf(kv-router): isolate direct ZMQ ingress runtime

PeaBrane:
## Summary

- Stack on #11938 and move its direct-ZMQ source tasks from the application Tokio runtime onto one service-owned ingress runtime.
- Default the isolated runtime to two Tokio workers through `DYN_ROUTER_ZMQ_INGRESS_THREADS`; invalid or unset values use `2`, while `0` restores the legacy aggregated subscriber.
- Keep the runtime KV-local and direct-ZMQ-only. NATS, brokered ZMQ, generic subscribers, request-plane TCP, CRTC threads, and queue-accepted flume semantics remain unchanged.
- Own runtime startup, cancellation, completion, and bounded shutdown with the KV event subscription handle.

This is intentionally a small, workload-sensitive optimization on top of the transport/lifecycle change. The dedicated runtime isolates ingress work from request-plane Tokio tasks; its size is not derived from host CPU count.

Reviewer guide: the entire stack-specific behavior is in `subscriber.rs`; `environment_names.rs` only registers the environment variable.

### Incremental performance

Using the same Mooncake corpus:

| Path | Median events/s |
|---|---:|
| Legacy aggregation | 555k |
| Direct delivery on application Tokio | 963k |
| Direct delivery, dedicated ingress 2 | 1.229M |

The two-thread pool was **28% faster than direct-on-main** and **2.21x aggregation** in the isolated sweep. Longer interaction runs measured 1.214M/s versus 1.138M/s at CRTC-32 and 1.250M/s versus 1.022M/s at CRTC-64.

Two is the smallest robust measured isolation point, not a universal optimum. Wider pools increased CRTC/ArcSwap contention; pool 16 fell to 1.072M/s at CRTC-64 while ArcSwap debt/CAS overhead rose from 3.5% to 7.5% of total cycles. Accordingly, this PR keeps the setting tunable and does not auto-scale it with CPU count.

### Coexistence and correctness

- At constrained 4 / 8 / 16 CPU budgets, additive ingress-2 changed throughput by -2.4% / -2.2% / +6.1% versus direct-on-main; a constant total CPU budget was consistently 1-3% better than additive allocation.
- Two-node live-mocker coexistence with constant-budget main-62 + ingress-2 changed request p99 by -0.03% versus aggregation, with exact event accounting.
- The broader RCA completed 310 valid runs with exact issued/decoded/handled/applied counts and zero ordering errors.
- Environment/topology selection and deterministic subscription shutdown tests pass; the combined stack previously passed `cargo test -p dynamo-llm --no-default-features`.
- `DYN_ROUTER_ZMQ_INGRESS_THREADS=0` is the immediate rollback.

### DIFF
--- a/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
@@ -166,16 +166,13 @@ async fn run_subscription_supervisor<T: RecoveryTarget>(
     }
 
     client.shutdown().await;
-    if cancellation_token.is_cancelled() {
-        metrics.set_kv_event_source_mismatch_workers(
-            &model,
-            worker_type,
-            &serving_endpoint.namespace,
-            &serving_endpoint.component,
-            &serving_endpoint.name,
-            0,
-        );
-    }
+    clear_mismatch_metric_on_cancellation(
+        &metrics,
+        &cancellation_token,
+        &model,
+        worker_type,
+        &serving_endpoint,
+    );
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -277,6 +274,26 @@ pub(super) fn update_subscription_failure_metric(
         &serving_endpoint.component,
         &serving_endpoint.name,
         view.sources.len(),
+    );
+}
+
+pub(super) fn clear_mismatch_metric_on_cancellation(
+    metrics: &RouterWorkerStatusMetrics,
+    cancellation_token: &CancellationToken,
+    model: &str,
+    worker_type: &str,
+    serving_endpoint: &EndpointId,
+) {
+    if !cancellation_token.is_cancelled() {
+        return;
+    }
+    metrics.set_kv_event_source_mismatch_workers(
+        model,
+        worker_type,
+        &serving_endpoint.namespace,
+        &serving_endpoint.component,
+        &serving_endpoint.name,
+        0,
     );
 }
 
--- a/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/subscriber.rs
@@ -7,16 +7,20 @@ use anyhow::Result;
 use dynamo_kv_router::protocols::{KV_EVENT_SUBJECT, RouterEvent};
 use dynamo_runtime::{
     component::{Component, Endpoint},
+    config::environment_names::router::DYN_ROUTER_ZMQ_INGRESS_THREADS,
     discovery::EventTransportKind,
     protocols::EndpointId,
     traits::DistributedRuntimeProvider,
-    transports::event_plane::{EventSubscriber, TypedEventSubscriber},
+    transports::event_plane::{EventSubscriber, TypedEventSubscriber, uses_direct_zmq},
 };
 use tokio::sync::{Semaphore, oneshot};
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 
-use super::{IndexerRecoveryTarget, RecoveryTarget, worker_query::WorkerQueryClient};
+use super::{
+    IndexerRecoveryTarget, RecoveryTarget, direct_zmq::run_direct_zmq_supervisor,
+    worker_query::WorkerQueryClient,
+};
 use crate::{
     discovery::{KvSourceMembershipView, KvSourceMembershipWatch, KvSourceStatus},
     kv_router::{Indexer, metrics::RouterWorkerStatusMetrics},
@@ -24,6 +28,8 @@ use crate::{
 
 const SUBSCRIPTION_INITIAL_BACKOFF: Duration = Duration::from_millis(100);
 const SUBSCRIPTION_MAX_BACKOFF: Duration = Duration::from_secs(5);
+const DEFAULT_ZMQ_INGRESS_THREADS: usize = 2;
+const INGRESS_RUNTIME_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
 
 enum ScopeExit {
     Rebind,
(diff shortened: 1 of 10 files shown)


lib/llm/src/kv_router/indexer/recovery/subscriber.rs
```diff
@@ -274,38 +280,163 @@ fn update_subscription_failure_metric(
...
-    startup_rx.await.map_err(|_| {
-        anyhow::anyhow!("KV event subscription supervisor exited before reporting readiness")
-    })?;
-    Ok(())
+    let (completion_tx, completion_rx) = oneshot::channel();
+    let owner_cancel = cancel.clone();
+    std::thread::Builder::new()
+        .name("dynamo-kv-zmq-owner".to_string())
+        .spawn(move || {
+            let runtime = match tokio::runtime::Builder::new_multi_thread()
```
jthomson04, line 416: What's the motivation behind a dedicated tokio runtime for this? Hardcoding to a two-thread default is concerning. If somebody's running on a massive cpu, they shouldn't have to manually change this value to get good perf.

jthomson04 (commented):
Primary concern is the static default of two tokio workers in the dedicated runtime. I also find it very odd that a bigger runtime would lead to lower ingest perf. Was there cpu contention between the benchmarker and ingress or something like that?

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
PeaBrane: Thanks — agreed that the fixed default and the inverse scaling need more context.

The dedicated runtime is intended as isolation, not as a way to scale ingress threads with host CPU count. In the controlled full-delivery pool sweep, the application runtime stayed at 64 workers and the results were:

| Path | Events/s |
|---|---:|
| Aggregated ZMQ, main runtime 64 | 555k |
| Direct delivery, main runtime 64 | 963k |
| Dedicated pool 1 | 641k |
| **Dedicated pool 2** | **1.229M** |
| Dedicated pool 4 | 1.088M |
| Dedicated pool 8 | 930k |
| Dedicated pool 16 | 876k |
| Recovery-aware funnel control | 1.773M |

This was an exclusive 144-core Neoverse-V2 host with randomized repeats. A paired supplement reconfirmed pool 2 at 1.275M/s (1.252M/s combined six-run median), and the null issuer sustained 2.67M/s, so issue capacity was not the candidate bottleneck. A separate two-node live-mocker coexistence run also kept the load generator off the SUT; constant-budget main-62 + ingress-2 changed request p99 by -0.03% versus aggregation with exact event accounting. That said, the isolated burst harness still colocated publishers and the SUT, so I do not want to claim that every form of scheduler contention has been ruled out.

The reason more workers hurt is not yet proven. Tokio task migration/work stealing and cache locality are candidates, as are ZMQ/reactor scaling and increased contention where decoded batches enter `WorkerQueryClient`/flume/CRTC.

A focused RCA campaign is active WIP:

1. Cross pools 1/2/4/8/16 with cumulative cuts: raw receive, envelope decode, payload decode, predecoded handler admission, and the full path.
2. Put publishers/issuer on a separate host and pin ingress workers within one NUMA domain.
3. Compare one N-worker Tokio runtime against N statically sharded single-thread runtimes on the same CPUs. This should separate multi-thread scheduling/migration from actual useful parallelism.
4. Record cycles, IPC, cache misses, context switches, CPU migrations, run queue, and per-thread utilization; repeat at fixed-total-work publisher cardinalities 4/32/256.

Until that is resolved, `2` is the smallest measured winner rather than a CPU-count-derived optimum; it is tunable with `DYN_ROUTER_ZMQ_INGRESS_THREADS`, and `0` retains the aggregate rollback. The exact production patch still needs its constrained-CPU and soak validation before default-on rollout.

PeaBrane: RCA complete. The dedicated runtime is useful for isolation, but it is not a CPU-scaling mechanism; `2` is a conservative measured default, not something we should derive from host core count.

Across 310 valid runs, the receive, envelope-decode, and payload-decode cuts stayed at roughly 1.7–1.9M events/s for pool sizes 1/2/4/8/16. The regression appeared only when the real `WorkerQueryClient`/flume/CRTC handler was enabled. A static-shard control (N independent one-worker Tokio runtimes) was no better than one N-worker runtime: static/pool medians were 1.03x at 2, 0.98x at 8, and 0.93x at 16. That rules out Tokio work stealing/task migration as the main cause.

The longer interaction sweep locates the knee downstream:

| CRTC workers | direct main | ingress 2 | ingress 16 | funnel |
|---:|---:|---:|---:|---:|
| 32 | 1.138M/s | 1.214M/s | 1.173M/s | 1.350M/s |
| 64 | 1.022M/s | 1.250M/s | 1.072M/s | 1.302M/s |

Perf corroborates this: widening ingress 2→16 increased CRTC-worker cycle share from 16.5%→20.0% at CRTC-32 and 14.2%→22.8% at CRTC-64. At CRTC-64, ArcSwap debt/CAS overhead increased from 3.5%→7.5% of total cycles while throughput fell. The wider ingress burst is amplifying contention in the shared radix/ArcSwap update path and oversubscribing downstream work; ZMQ receive/decode itself is not the bottleneck.

The in-process issuer was kept as a diagnostic rather than assumed free: the null issuer met its 124 ms target (127.7–128.9 ms completion, 3–4 ms p99 issue lag), we used 4x-longer traces, and the prior two-node live-mocker run kept generation off the SUT. On constrained 4/8/16-CPU runs, direct-on-main was already 1.54–2.12x aggregate; additive ingress-2 was within -2.4%/-2.2% on 4/8 CPUs and +6.1% on 16 CPUs. Keeping the total budget constant was consistently 1–3% better than adding threads.

Implementation conclusion: keep direct concurrent delivery and the small isolated pool with `0` as rollback; do not scale the default upward with CPU count, and do not add static shards. The optimum moves with publisher cardinality, so a larger automatic pool would be less robust. I have preserved the full command/result bundle and am adding the same headline to DEP #11899.

PR #11930, opened 2026-07-21
feat(kv-router): normalize key-only chunk removals to block-wise in the publisher

Change72:
## Summary

Handle key-only lower-tier removals by expanding them to block-wise in the
worker-side KV publisher, and keep the lower-tier indexer purely block-wise.

vLLM's lower-tier (offload) removals are moving to **key-only**: a removal carries
only the chunk key (tail block hash), not the chunk's constituent block hashes.
The router must reconstruct the members. This PR does that in the publisher
(before batching), so the lower-tier indexer never has to be chunk-aware.

## Why in the publisher, not the indexer

vLLM emits one `BlockStored` per chunk, but the publisher's `event_processor`
coalesces adjacent parent-chaining stores into one multi-block store, which
erases chunk boundaries **before** any indexer sees them. A chunk-aware indexer
that treats `blocks.last()` as the tail therefore misreads a real chunk tail as an
interior block once its store was merged — leaking or mis-deleting on removal. The
normalizer sits before batching, where each store is still one chunk, so it can
build an accurate `tail -> members` map and expand removals there. After expansion
everything downstream is block-wise and the existing per-tier dedup filter handles
shared-block reference counting.

This supersedes this PR's earlier indexer-side (tail-form) approach, which the
batching above defeats.

## What changes

- **`lib/llm/src/kv_router/publisher/normalizer.rs` (new):** per-worker map keyed
  by `(dp_rank, storage_tier, tail_hash) -> [member block hashes]`, built from
  store events. Each removal is expanded to block-wise (union-dedup, so a fat
  removal collapses to the same set — the PR is safe while vLLM still emits fat
  removals, no flag day). Duplicate/retry stores are suppressed so the
  always-incrementing dedup filter counts each chunk once. The map is cleared per
  rank on `Cleared`, in lockstep with `dedup.clear_rank`.
- **`lib/kv-router/src/indexer/lower_tier.rs`:** lower-tier removal is now
  infallible — a removal for an unknown worker or block is an idempotent no-op
  instead of `Err(BlockNotFound)`, because after the normalizer + dedup filter the
  indexer can legitimately receive a removal for a block it does not hold (an
  untracked-hash removal the dedup filter passes through, a `Cleared` race, or a
  store that indexed only a prefix under an edge conflict). The indexer stays
  block-wise; no chunk awareness.

## Reviewer notes

- The two correctness-critical points: (1) the normalizer map is cleared in
  lockstep with `dedup.clear_rank` on reset, and (2) it is partitioned by
  `(dp_rank, storage_tier)` exactly like the dedup filter. Both leak / over-delete
  if wrong.
- Known limit (documented in the module): a shared tier where the storing worker
  is gone by removal time cannot be expanded by the per-worker map. This is moot
  for worker-local tiers (map gone ⇒ cache gone) and is left to a future
  lease/TTL/scrub mechanism.
- Metrics for the new path are in the stacked follow-up (#12308).

## Test plan

- `cargo test -p dynamo-llm --lib kv_router::publisher::normalizer` — 7 new unit
  tests pass (key-only expansion, duplicate-store suppression, shared-prefix
  survival, fat-removal union-dedup, unknown-tail pass-through, per-rank/tier
  independence, clear_rank).
- `cargo test -p dynamo-kv-router --lib` — 860 passed, 0 failed (no regression
  from the infallible-removal change).
- `cargo clippy -p dynamo-llm -p dynamo-kv-router --all-targets -- -D warnings` —
  clean.
- `cargo fmt --all -- --check` — clean.

AI assistance was used to design and draft this change; every line was reviewed by
the submitter.

### DIFF
--- a/lib/kv-router/src/indexer/lower_tier.rs
+++ b/lib/kv-router/src/indexer/lower_tier.rs
@@ -288,14 +288,21 @@ impl LowerTierIndexer {
         worker: WorkerWithDpRank,
         block_hashes: &[ExternalSequenceBlockHash],
     ) -> Result<(), KvCacheEventError> {
+        // Lower-tier removals are idempotent no-ops for state we do not hold.
+        // With the publisher-side normalizer + dedup filter, the indexer can
+        // legitimately receive a removal for a block (or worker) it never indexed
+        // — an untracked-hash removal the dedup filter passes through, a race
+        // with a Cleared reset, or a store that only indexed a prefix under an
+        // edge conflict. Skipping (rather than aborting the batch) avoids leaving
+        // a partial removal behind.
         let remove_worker_entry = {
             let Some(worker_map) = worker_blocks.get_mut(&worker) else {
-                return Err(KvCacheEventError::BlockNotFound);
+                return Ok(());
             };
 
             for block_hash in block_hashes {
                 let Some(key) = worker_map.remove(block_hash) else {
-                    return Err(KvCacheEventError::BlockNotFound);
+                    continue;
                 };
 
                 self.remove_worker_from_edge(key, worker);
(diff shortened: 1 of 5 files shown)


Change72:
Split into two stacked PRs. This PR (#11930) is now the **mechanism-only** half; the lower-tier removal **metrics/observability** half is #12308, stacked on top of this one.

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `effe4b2` | unsigned |

PR #11938, opened 2026-07-21
perf(kv-router): ingest direct ZMQ events concurrently [DYN-3849]

PeaBrane:
## Summary

- Deliver direct-ZMQ KV events through one FIFO task per publisher on the application Tokio runtime, allowing different publishers to enter `WorkerQueryClient` concurrently.
- Add a true single-consumer ZMQ stream that retains native subject filtering, frame validation, `Codec` decoding, and the 100k receive HWM while bypassing the lossy 1,024-entry broadcast hop.
- Preserve membership fencing, recovery mutex and sequence validation, flume/CRTC admission, per-source reconnect/replacement, deterministic shutdown, and lifecycle/error metrics.
- Leave NATS, brokered ZMQ, generic subscribers, and request-plane TCP unchanged.

This is the transport and lifecycle base for #11899. The dedicated ingress runtime and `DYN_ROUTER_ZMQ_INGRESS_THREADS` tuning/rollback are intentionally separated into stacked PR #11925.

Reviewer guide: `direct_zmq.rs` and `worker_query.rs` contain the source lifecycle and live-ingestion semantics; `zmq_transport.rs` contains the single-consumer transport primitive; `subscriber.rs` selects direct ZMQ and runs it on the application runtime.

## Validation

- Same Mooncake corpus: current aggregation **555k events/s**; direct delivery on the application runtime **963k events/s** (**1.73x**).
- Constrained runs with four CRTC workers retained **1.54x / 2.02x / 2.12x** aggregate throughput at 4 / 8 / 16 CPUs respectively.
- The broader RCA completed 310 valid runs with exact issued/decoded/handled/applied accounting and zero ordering errors; receive/decode sustained roughly 1.7-1.9M events/s, locating the remaining saturation downstream in CRTC rather than this transport path.
- Two-node live-mocker coexistence measured 560.83 ms request p99 for direct-on-main versus 560.94 ms for aggregation, with exact event accounting.
- `cargo test -p dynamo-llm --no-default-features direct_zmq_multi_node_replacement_isolated_by_global_rank` (1 passed)
- `cargo test -p dynamo-llm --no-default-features kv_router::indexer::recovery::subscriber::tests --lib` (1 passed)
- `cargo test -p dynamo-runtime direct_zmq_topology_selection_is_narrow --lib` (1 passed)
- The combined stack previously passed `cargo test -p dynamo-llm --no-default-features`; formatting, DCO hooks, and `git diff --check` pass on this split.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/11938" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added direct ZMQ event ingestion for faster, single-consumer processing of live updates.
  * Added lifecycle monitoring for ingestion sources, batches, and transitions.
* **Bug Fixes**
  * Improved handling of publisher changes, disconnects, stale connections, and malformed or mismatched messages.
  * Ensured event-processing metrics reset correctly after shutdown.
* **Reliability**
  * Added graceful cancellation and cleanup for background event processing.
  * Improved reconnection behavior with controlled retry backoff.

### DIFF
--- a/lib/llm/src/kv_router.rs
+++ b/lib/llm/src/kv_router.rs
@@ -210,6 +210,7 @@ where
     cancellation_token: CancellationToken,
     client: Client,
     is_eagle: bool,
+    kv_event_subscription: Option<indexer::KvEventSubscriptionHandle>,
     _served_indexer_handle: Option<ServedIndexerHandle>,
     /// Optional external shared KV cache pool. When present, `find_best_match`
     /// queries it in parallel with the indexer and factors shared hits into scoring.
@@ -298,30 +299,34 @@ where
         .await?;
 
         // Start KV event subscription if needed — skip when using a remote indexer.
-        if kv_router_config.use_remote_indexer {
+        let kv_event_subscription = if kv_router_config.use_remote_indexer {
             tracing::info!("Skipping KV event subscription (using remote indexer)");
+            None
         } else if kv_router_config.should_subscribe_to_kv_events() {
             let membership_watch = kv_source_membership.ok_or_else(|| {
                 anyhow::anyhow!(
                     "KV source membership watch is required when local KV event subscription is enabled"
                 )
             })?;
-            indexer::start_subscriber(
-                endpoint.clone(),
-                indexer.clone(),
-                membership_watch,
-                model_name.clone().unwrap_or_else(|| "unknown".to_string()),
-                worker_type,
-                cancellation_token.child_token(),
+            Some(
+                indexer::start_subscriber(
+                    endpoint.clone(),
+                    indexer.clone(),
+                    membership_watch,
+                    model_name.clone().unwrap_or_else(|| "unknown".to_string()),
+                    worker_type,
+                    cancellation_token.child_token(),
+                )
+                .await?,
             )
-            .await?;
         } else {
             tracing::info!(
                 "Skipping KV event subscription (use_kv_events={}, overlap_score_credit={})",
                 kv_router_config.use_kv_events,
                 kv_router_config.overlap_score_credit,
             );
-        }
+            None
+        };
 
         let served_indexer_handle = if kv_router_config.serve_indexer {
             let model_name = model_name.clone().ok_or_else(|| {
@@ -352,6 +357,7 @@ where
             cancellation_token,
             client,
             is_eagle,
+            kv_event_subscription,
             _served_indexer_handle: served_indexer_handle,
             shared_cache,
             lora_filter,
@@ -369,6 +375,14 @@ where
 
     pub fn kv_router_config(&self) -> &KvRouterConfig {
         &self.kv_router_config
+    }
+
+    /// Cancel background work and wait for KV event ingestion to stop.
+    pub async fn shutdown(mut self) {
+        self.cancellation_token.cancel();
+        if let Some(subscription) = self.kv_event_subscription.take() {
+            subscription.shutdown().await;
+        }
     }
 
     pub fn is_eagle(&self) -> bool {
--- a/lib/llm/src/kv_router/indexer/recovery/direct_zmq.rs
+++ b/lib/llm/src/kv_router/indexer/recovery/direct_zmq.rs
@@ -1,0 +1,836 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+    time::Duration,
+};
+
+use anyhow::Result;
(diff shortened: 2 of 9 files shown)


lib/llm/src/kv_router/indexer/recovery/subscriber.rs
```diff
@@ -274,38 +274,126 @@ fn update_subscription_failure_metric(
...
+        model,
+        worker_type,
+        &serving_endpoint.namespace,
+        &serving_endpoint.component,
+        &serving_endpoint.name,
+        0,
+    );
+}
+
+pub(crate) struct KvEventSubscriptionHandle {
```
jthomson04, line 299: > [!NOTE]
> 🤖 **This comment is AI-generated** (Claude Code review).

`start_subscriber` used to be fire-and-forget; with this change, the returned handle's `Drop` cancels the subscription's child token. That means any future caller that writes `start_subscriber(...).await?;` without binding the handle gets a subscription that is silently cancelled at the end of the statement. The one production caller (`KvRouter`) stores the handle, but the `Result` alone satisfies must-use, so this is an easy regression to reintroduce later.

Suggestion: mark the handle itself as must-use and document the drop semantics, so the discarded-handle pattern produces a compiler warning:

```rust
/// Dropping this handle cancels the KV event subscription.
#[must_use = "dropping the handle cancels the KV event subscription"]
pub(crate) struct KvEventSubscriptionHandle {
```

lib/llm/src/kv_router/indexer/recovery/direct_zmq.rs
```diff
@@ -0,0 +1,836 @@
...
+            continue;
+        }
+        let Some(source) = sources.get_mut(&publisher_id) else {
+            continue;
+        };
+        if source.state.active_bindings() == Some(&bindings) {
+            continue;
+        }
+        let SourceState::Preconnected { activate } =
+            std::mem::replace(&mut source.state, SourceState::Fenced)
```
jthomson04, line 524: > [!NOTE]
> 🤖 **This comment is AI-generated** (Claude Code review).

`std::mem::replace` executes unconditionally, so when the `let ... else` pattern does not match, the source's state is still replaced with `SourceState::Fenced` and the old state is dropped in the `else` arm — without going through `leave_source_state`.

This path is reachable: `current_ready` is computed *before* the `stale_after_sync` restart loop, so a just-restarted publisher (now `Connecting`) still appears in `ready_publishers` and gets silently flipped to `Fenced` here. Today that happens to be benign — `Connecting` and `Fenced` behave identically (`is_ready()` false, no gauge), and the next `Ready` signal self-heals via `transition_source_state` — but if either state ever gains a gauge or distinct behavior, this becomes a real bug and a gauge leak.

Suggestion: guard before mutating, e.g.

```rust
if !matches!(source.state, SourceState::Preconnected { .. }) {
    continue;
}
let SourceState::Preconnected { activate } =
    std::mem::replace(&mut source.state, SourceState::Fenced)
else {
    unreachable!()
};
```

or route the transition through `transition_source_state` so state changes and gauge accounting stay in one place.

jthomson04 (approved):

lib/llm/src/kv_router/indexer/recovery/subscriber.rs
PeaBrane: Fixed in 1e6f0786c0. The handle now documents its drop-cancels behavior and is marked `#[must_use]`, so discarding the successful result produces a compiler warning.

lib/llm/src/kv_router/indexer/recovery/direct_zmq.rs
PeaBrane: Fixed in 1e6f0786c0. Reconciliation now verifies that the source is `Preconnected` before replacing its state, so a stale ready snapshot cannot silently turn a restarted `Connecting` source into `Fenced`.

lib/llm/src/kv_router.rs
```diff
@@ -371,6 +377,14 @@ where
         &self.kv_router_config
     }
 
+    /// Cancel background work and wait for KV event ingestion to stop.
+    pub async fn shutdown(mut self) {
+        self.cancellation_token.cancel();
+        if let Some(subscription) = self.kv_event_subscription.take() {
+            subscription.shutdown().await;
+        }
+    }
```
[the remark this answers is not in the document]
PeaBrane: Agreed that deterministic async shutdown is not currently reachable through production `Arc<KvRouter>` ownership. The last-`Arc` drop still cancels ingestion safely, and production does not currently require awaiting task completion. I am deferring an Arc-compatible deterministic shutdown API rather than expanding the lifecycle design in this PR.

PR #11942, opened 2026-07-21
refactor(mocker): isolate offline replay policies [DYN-3850]

PeaBrane:
## Summary

- introduce router-neutral admission, placement, engine-observation, and worker-lifecycle contracts for offline replay
- keep round-robin inside the runnable offline core as the zero-sized, statically dispatched default
- isolate concrete KV routing and KV-event protocol code under `offline/extensions/kv_router` and `offline/extensions/kv_events`
- preserve aggregated/disaggregated capacity, topology, handoff, planner, and KVBM liveness semantics
- add reusable deterministic report and production performance support for future offline replay parity campaigns

This PR does not create a new crate or feature-gate the existing unconditional `dynamo-kv-router` Cargo dependency. The source firewall is complete: `offline/core` has no KV-router/adapter symbols, and literal `dynamo_kv_router` references occur only inside the two KV extension directories. The feature-free replay benchmark remains runnable without `dynamo-runtime`.

No generated parity reports or campaign scripts are committed. The non-default `replay-bench` feature only makes request IDs and KV selector sampling deterministic; it does not gate replay implementation code or change feature-off production routing.

### Reviewer guide

- Core contracts and built-in round-robin: `lib/mocker/src/replay/offline/core/`
- Concrete composition: `lib/mocker/src/replay/offline/entrypoints.rs`
- KV routing/event extensions: `lib/mocker/src/replay/offline/extensions/`
- Reusable parity/performance output: `lib/bench/offline_replay_bench.rs`

## Validation

Direct runtime comparison: parity baseline `b9849f051ba253b9767bf3ae12c6f0404eeb622b` to runtime candidate `048a52ad100f3f41ccb1863b6ff6d3b3736154e5`, same x86_64 Linux host, Rust 1.96.1, release profile, locked dependencies, and Mooncake-1000 trace SHA-256 `e2bec9bccd21978f69a2ace7886ee2e5192a64839f7aa2c4dd5d8766fd392510`. The follow-up HEAD commit only removes campaign artifacts/tests and wires semantic KVBM replay tests into pre-merge CI.

- canonical parity passed with 20 byte-identical baseline runs, 20 byte-identical candidate runs, and exact cross-revision bytes for all eight normal cells: vLLM PassStart and SGLang PassEnd × aggregated/disaggregated × round-robin/KV
- the same byte gate passed for all four 3,072-block KVBM-stress cells; a second bounded edge pass verified all 1,000 requests completed in each cell
- the candidate passed both forced G1→G2 eviction / G2→G1 restoration tests, each covering aggregated/disaggregated × round-robin/KV; these tests now run in the existing Linux KVBM pre-merge step
- production-selector performance passed after 5 warmup and 30 randomized A/B pairs; one-sided 95% upper bounds were 1.0086 (aggregated RR), 1.0380 (disaggregated RR), 1.0056 (aggregated KV), and 0.9986 (disaggregated KV)
- release size passed: binary ratio 1.0407 and `.text` ratio 1.0470
- after fixture cleanup, 161 scoped offline tests passed with default features and the same 161 passed with `replay-bench`; the seeded KV selector test passed and the replay benchmark target built with `replay-bench`
- Clippy with warnings denied and rustfmt passed for `lib/mocker`, `lib/kv-router`, and `lib/bench`; `git diff --check` and repository hooks passed
- the final Linux KVBM suite reported 622 passed / one unrelated SGLang scheduler timing failure / 1 ignored under parallel load; that sole timing test passed immediately in isolation, and the replay-specific forced lifecycle tests passed

Parity used two checksum-pinned baseline-only harness patches: report-envelope metadata and the existing seeded selector under `replay-bench`. Feature-off production routing remained unchanged. Full reports, traces, copied binaries, worktrees, bundles, and cluster scratch were deleted after retaining compact checksums and statistics outside the repository.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-iteration JSONL exports for benchmark timings and canonical replay reports.
  * Added support for deterministic replay outputs, including stable request identifiers and routing selections.
  * Expanded offline replay to support round-robin and KV-router modes across aggregated and disaggregated scenarios.
  * Added engine observation and KV-event capture for replay artifacts, including offload scenarios.

* **Bug Fixes**
  * Improved replay consistency and comparability across repeated benchmark runs.

* **Tests**
  * Added coverage for deterministic routing, replay parity, lifecycle handling, and offload behavior.


lib/mocker/src/replay/offline/core/round_robin.rs
```diff
@@ -0,0 +1,341 @@
...
+            .ok_or_else(|| anyhow!("round-robin placement requires a request UUID"))?;
+        let active_count = self.workers.values().map(Vec::len).sum::<usize>();
+        let index = self.next % active_count;
+        let scheduler_id = self
+            .workers
+            .values()
+            .flat_map(|ranks| ranks.iter().copied())
+            .nth(index)
+            .expect("active round-robin pool must contain a scheduler");
+        self.next = index + 1;
```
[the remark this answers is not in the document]
PeaBrane: Confirmed this is unreachable at the runtime boundary. `execute_action` checks the corresponding engine’s `has_active_workers()` before either route calls placement; actions wait when the pool is empty and are woken only after worker readiness updates both engine and placement topology. This preserves the pre-existing invariant, and the scale-to-zero test covers the wait/wake path, so I’m keeping the hot path branch-free.

PeaBrane:
> Explicitly flush output writers before drop to avoid silently truncated parity JSONL.

Addressed in `88eb2012b9`. Both JSONL `BufWriter`s now flush explicitly after the iteration loop, and final flush failures propagate with output-specific context. Validated with `cargo clippy --locked -p dynamo-bench --no-deps --all-targets -- -D warnings`.

lib/bench/offline_replay_bench.rs
```diff
@@ -210,6 +249,14 @@ struct Args {
     #[arg(long, default_value_t = 1)]
     iterations: usize,
 
+    /// Emit one JSON object per measured replay iteration to this path.
+    #[arg(long)]
+    timings_jsonl: Option<PathBuf>,
+
+    /// Emit one canonical full replay report per iteration for parity checks.
+    #[arg(long)]
+    canonical_reports_jsonl: Option<PathBuf>,
```
dreamtalen, line 259: Non-blocking: I tried `--canonical-reports-jsonl` and found that it only produces identical reports when `replay-bench` is enabled. Without that, identical inputs produced different reports because request UUIDs and KV-selector placement remain random.

Since this report is intended for parity checks, could we document somewhere or fail fast when it is used without `replay-bench`?

dreamtalen (approved):

tedzhouhk (approved):

tedzhouhk (approved):

lib/bench/offline_replay_bench.rs
PeaBrane: Addressed in `1d20bf4886`. The CLI now documents that `--canonical-reports-jsonl` requires `replay-bench`, and fails immediately when the flag is used without that feature, before reading the trace or opening output. Validated with Clippy under both `offline-replay` alone and `replay-bench`, with warnings denied.

saturley-hall (approved):
Reasonable change to CI to improve testing coverage

PR #11235, opened 2026-07-06
ci(operator): run GPU-free deploy e2e tests

sttts:
## Overview

Run the GPU-free Dynamo operator deploy E2E suite in PR and post-merge CI by reusing the deploy vCluster. The lane exercises the installed Helm chart, Kubernetes RBAC, DGDR reconciliation, profiling jobs, and cleanup without allocating GPUs.

## Summary

- add a reusable operator E2E workflow that reconnects to the existing deploy vCluster and self-bootstraps it for reruns after cleanup
- build and copy the commit-scoped planner image needed by mocker-mode DGDR profiling
- run only validation and rapid `gpu_0` specs, excluding the separately labeled H100 support matrix
- collect JUnit results, logs, Kubernetes events, and operator diagnostics on failure
- pass only the secrets required by the new reusable workflow calls and disable persisted checkout credentials
- remove stale cluster-scoped node access from the profiling-job role; GPU discovery runs in the operator, not the profiling job

## Details

The first live CI run caught a real chart/operator integration defect: the operator dynamically binds the profiling-job ClusterRole, but that role included `nodes get/list/watch` while the operator service account held only `nodes get` with GPU discovery disabled. Kubernetes rejected the RoleBinding as privilege escalation. The profiling job does not perform node discovery, so this PR removes that stale permission instead of broadening the operator's access.

## Where should reviewer start?

1. `.github/workflows/shared-operator-e2e.yml` for the test selection, vCluster reuse, and diagnostics.
2. The `planner-copy-to-acr` and `deploy-test-operator-gpu0` jobs in `.github/workflows/pr.yaml` and `.github/workflows/post-merge-ci.yml` for gating and least-privilege secrets.
3. `deploy/helm/charts/platform/components/operator/templates/profiling-job-rbac.yaml` and its Helm regression test for the RBAC correction.

## Related Issues

- [DYN-3452](https://linear.app/nvidia/issue/DYN-3452/expand-operator-live-cluster-e2e-coverage-in-ci-using-the-shared)

## Validation

- pre-commit hooks on all changed files
- Helm unit tests: 16 passed
- actionlint on all three modified workflows, ignoring only repository custom-runner-label warnings
- YAML parsing with `yq`
- `git diff --check`
- dry-run of the exact Make/Ginkgo selection: 27 of 39 specs selected, 0 failed, H100 support matrix excluded, and JUnit XML generated
- live Kubernetes rerun pending the next pushed commit

PR #11946, opened 2026-07-21
test(deploy): port operator DGDR e2e coverage to pytest

sttts:
## Summary

Depends on #11949.

- Supersedes #11235 by moving all 39 DGDR live-cluster tests from Ginkgo to `tests/deploy`.
- This PR enables **CPU-only CI coverage**: 27 `gpu_0` cases run against mocked hardware in isolated, always-cleaned namespaces.
- The 12 real-GPU cases are ported but not scheduled in CI. Enabling the two `gpu_1` and ten H100 `gpu_8` cases will follow after a GPU-capacity discussion.
- Leaves the legacy Kind smoke test deprecated and retains the profiling RBAC fix.

The CPU job is independent of the vLLM, SGLang, and TensorRT-LLM deploy-test jobs.

## Validation

- 39 tests collected; 27 CPU cases selected in CI
- pre-commit, marker report, and workflow lint
- legacy Go package compile


tests/deploy/README.md
```diff
@@ -0,0 +1,42 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Deployment tests
+
+This directory contains live-cluster pytest suites. `test_deploy.py` validates
```
sttts: Should we rename to test_dgd.py then? Anything else in there?

sttts: Yes. It contains the regular DGD serving test and the GAIE DGD test, nothing outside DGD coverage. Renamed it to `test_dgd.py` and updated the deploy-test action default and README.

tests/deploy/dgdr_utils.py
```diff
@@ -0,0 +1,696 @@
...
+        }
+        for service in expected_services:
+            assert service in services or service in components
+
+    if manifest["spec"].get("autoApply", True) is False:
+        return result, output
+
+    result = await manager.wait_for_phase(
+        name, "Deployed", manager.config.deploy_timeout
+    )
```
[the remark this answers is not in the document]
sttts: Fixed in 855d9a77d5: the Deployed wait now uses `profiling_timeout + deploy_timeout`, matching the Go test.

sttts:
Fixed the CodeRabbit nit in 855d9a77d5: external YAML/Kubernetes validation now uses explicit raises instead of runtime asserts.

julienmancuso (commented):
Five findings from review of the current PR head.

tests/deploy/test_dgdr_h100.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+H100_TTFT_MS = _env_float("DGDR_H100_TTFT_MS", 500.0)
+H100_ITL_MS = _env_float("DGDR_H100_ITL_MS", 30.0)
+H100_ISL = _env_int("DGDR_H100_ISL", 3000)
+H100_OSL = _env_int("DGDR_H100_OSL", 300)
+H100_GPU_COUNT = 8
+
+BACKENDS = [
+    pytest.param("vllm", marks=pytest.mark.vllm),
```
julienmancuso: **[P1] Remove local backend markers from remote-cluster tests**

`tests/conftest.py` adds a skip to every `vllm`/`sglang`/`trtllm`-marked item when that Python module is not installed. These are remote-cluster deploy tests, and the deploy runner installs only `container/deps/requirements.test.txt`, so the three parameterized groups plus the separately marked GPT-OSS case become 10 skips instead of coverage. Please remove or rework all four framework markers, or exempt remote deploy tests from the local-backend dependency check.

tests/deploy/dgdr_utils.py
```diff
@@ -0,0 +1,775 @@
...
+
+
+def unique_name(config_: DGDRTestConfig, suffix: str) -> str:
+    """Return a short, collision-resistant Kubernetes resource name."""
+
+    digest = hashlib.sha1(suffix.encode(), usedforsecurity=False).hexdigest()[:6]
+    if config_.name_prefix:
+        return f"{config_.name_prefix[:37]}-{digest}"
+    millis = time.time_ns() // 1_000_000 % 100000
+    return f"dgdr-test-{suffix[:25]}-{millis}-{digest}"
```
julienmancuso: **[P1] Bound the generated name for profiler resources**

The empty-prefix path can produce names too long for the existing profiler guard. For `lifecycle-ready`, this is typically 38 characters; appending `-dgd` and the mocker's `Frontend` service yields 50, above `_MAX_COMBINED_RESOURCE_NAME_LENGTH = 45`, so the README command—which supplies no name prefix—fails during profiling. The `name_prefix[:37]` branch can exceed the same limit. Please bound the final DGDR name against the service-name constraint.

.github/filters.yaml
```diff
@@ -172,6 +173,16 @@ deploy:
   - '!**/*.md'
   - '!**/*.rst'
 
+dgdr:
+  - '.github/workflows/shared-dgdr-deploy-test.yml'
+  - 'tests/conftest.py'
```
julienmancuso, line 209: **[P2] Route DGDR-only changes through the prerequisite jobs**

This filter includes `tests/conftest.py`, but that path sets `dgdr=true` while `deploy=false` and `operator=false`. Both the `operator` and `deploy-operator` job conditions in `pr.yaml` omit the `dgdr` output, so the new `deploy-test-dgdr` job is skipped for exactly this declared dependency. Please include `dgdr` in both prerequisite conditions, or remove this path from the filter if it should not trigger the lane.

tests/deploy/test_dgdr.py
```diff
@@ -0,0 +1,466 @@
...
+    "case",
+    [
+        "missing-model",
+        "thorough-auto",
+        "invalid-backend",
+        "invalid-search-strategy",
+        "invalid-optimization-type",
+    ],
+)
+async def test_dgdr_validation_webhook_rejects_invalid_specs(
```
julienmancuso, line 121: **[P2] Add bounded timeouts to the live-cluster validation tests**

These validation cases all perform Kubernetes API or `kubectl` I/O, but none of the seven validation tests has a `pytest.mark.timeout`. A stalled API call can consume the entire 120-minute workflow timeout; the repo's pytest rules require explicit timeouts for network, polling, and subprocess tests. Please add a bounded timeout to the validation group.

tests/deploy/dgdr_utils.py
```diff
@@ -0,0 +1,775 @@
...
+        self._created_names: list[str] = []
+
+    async def init(self) -> None:
+        kubeconfig = os.environ.get("KUBECONFIG")
+        if kubeconfig and os.path.exists(kubeconfig):
+            await config.load_kube_config(config_file=kubeconfig)
+        else:
+            try:
+                config.load_incluster_config()
+            except Exception:
```
julienmancuso: **[P3] Catch only the missing in-cluster configuration error**

Catching `Exception` here turns programming or configuration failures into a silent fallback to the default kubeconfig, which can hide the real failure and target a different cluster. Please catch the specific `kubernetes_asyncio` configuration exception raised when in-cluster configuration is unavailable.

tests/deploy/test_dgdr_h100.py
sttts: Fixed in e7241eee08: removed the local backend markers; all 10 H100 cases are collected without local backend modules.

tests/deploy/dgdr_utils.py
sttts: Fixed in e7241eee08: DGDR names now fit both the profiler combined-name limit and the Grove label limit, with unit coverage.

.github/filters.yaml
sttts: Fixed in 6d0a802e9a: dgdr changes now trigger both operator and deploy-operator prerequisites.

tests/deploy/test_dgdr.py
sttts: Fixed in e7241eee08: all seven live-cluster validation tests now have a 120s timeout.

tests/deploy/dgdr_utils.py
sttts: Fixed in e7241eee08: the fallback catches only config.ConfigException; unexpected errors propagate, with unit coverage.

hhzhang16 (approved):

.github/workflows/shared-dgdr-deploy-test.yml
```diff
@@ -0,0 +1,135 @@
...
+        required: false
+
+jobs:
+  dgdr-deploy-test:
+    name: CPU / ${{ matrix.suite }}
+    runs-on: prod-deploy-tester-v1
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      max-parallel: 2
```
nv-tusharma: any particular reason to limit max-parallel to 2 here, I believe we should be fine from the CPU side for resources

.github/workflows/shared-dgdr-deploy-test.yml
```diff
@@ -0,0 +1,135 @@
...
+          vcluster_namespace: ${{ inputs.namespace }}
+
+      - name: Set up vCluster kubeconfig
+        id: kubeconfig
+        shell: bash
+        env:
+          KUBECONFIG_BASE64: ${{ steps.connect-vcluster.outputs.kubeconfig_base64 }}
+        run: |
+          set -euo pipefail
+          echo "${KUBECONFIG_BASE64}" | base64 -d > "${RUNNER_TEMP}/dgdr-test-kubeconfig"
```
nv-tusharma: how much of the vcluster setup can we extend from here: https://github.com/ai-dynamo/dynamo/blob/main/.github/actions/setup-dynamo-operator/action.yml
Maybe it's worth refactoring the vcluster setup into it's own action and extending it across both workflows?

nv-tusharma:
A few minor/nit observations (non-blocking):

- Several per-test timeouts in test_dgdr.py (3660–4260s, and 10920s on test_dgdr_profiling_backends_sequentially) exceed the CPU lane's 120-min job cap, so the job is killed before those timeouts ever apply — worth aligning them to the real job budget.
- dgdr_utils.py verify_inference_endpoints uses a substring model match (assert model in models.stdout), which is brittle if the served model id is normalized/path-stripped. Real-GPU only, so low impact.
- conftest.py pytest_runtest_makereport only records call.when == 'call', so failures during fixture setup skip the on-failure diagnostics — consider also capturing 'setup'.

tests/deploy/conftest.py
```diff
@@ -66,6 +67,65 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default="/models",
         help="Mount path for the model cache PVC (used with --model-cache-pvc).",
     )
+    parser.addoption(
+        "--dgdr-model",
+        default="Qwen/Qwen3-0.6B",
+        help="Default model for DynamoGraphDeploymentRequest tests.",
+    )
+    parser.addoption(
```
nv-tusharma: Can we use framework alternatively for this?

https://github.com/ai-dynamo/dynamo/blob/main/tests/deploy/conftest.py#L29-L34

tests/deploy/conftest.py
```diff
@@ -66,6 +67,65 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        default=600,
+        help="Seconds to wait for an auto-applied DGD.",
+    )
+    parser.addoption(
+        "--dgdr-name-prefix",
+        default="",
+        help="Short prefix for resources created by DGDR tests.",
+    )
+    parser.addoption(
+        "--dgdr-pvc-name",
```
nv-tusharma: additionally, for the dgdr paths that overlap with the existing args, does it make sense to just use the default args we already leverage? https://github.com/ai-dynamo/dynamo/blob/main/tests/deploy/conftest.py#L57-64

kthui (approved):
`tests/fault_tolerance/deploy/` looks good to me

.github/workflows/shared-dgdr-deploy-test.yml
sttts: Removed `max-parallel`; the three CPU suites can now run concurrently.

.github/workflows/shared-dgdr-deploy-test.yml
sttts: The workflow already reuses the shared check, setup, and connect actions. I also removed the duplicate kubeconfig decoding and now use the kubeconfig produced by `connect-vcluster` directly. A further split would not simplify this path because its missing-vCluster fallback needs the full operator setup.

tests/deploy/conftest.py
sttts: Yes. Removed `--dgdr-backend`; DGDR now uses the existing `--framework` option and defaults to vLLM when it is unset.

tests/deploy/conftest.py
sttts: Yes. Removed the overlapping DGDR PVC name and mount options; DGDR now uses `--model-cache-pvc` and `--model-cache-mount`. Only the DGDR-specific snapshot path remains separate.

sttts:
Addressed the three follow-up nits in e701d9cfd2: the only timeout above the 120-minute job cap is now bounded to 110 minutes, model-list validation parses exact JSON model IDs, and setup-phase failures now trigger failed-test diagnostics. The existing 61–71 minute per-test timeouts remain below the job cap.

rmccorm4 (approved):
Approving for the `tests/README.md` change

nv-tusharma (approved):
LGTM, for .github/ and tests/deploy changes

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `943f405` | unsigned |
| `8bb1931` | unsigned |
| `d0430ee` | unsigned |
| `3e47cdf` | unsigned |
| `0597802` | unsigned |
| `8525a4f` | unsigned |
| `db297b4` | unsigned |
| `07d9e4d` | unsigned |
| `f543273` | unsigned |
| `34ebf51` | unsigned |
| `7d42dc5` | unsigned |
| `ec0bed9` | unsigned |
| `2d656e2` | unsigned |
| `2a6cf41` | unsigned |
| `43285ce` | unsigned |

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/pull/11946/commits/1bfbf54cf4a5ce86dbbb5c5568975d4ee003fc3a

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/commit/43285ce939f05d772c51f1614f5b9c8ee61740e4

PR #11950, opened 2026-07-21
feat: add Get Dynamo install-command selector to docs 

nv-anants:
Adds an interactive install-command selector to the docs. In the Quickstart's **Get Dynamo** section, readers pick a backend (vLLM / SGLang / TensorRT-LLM), a stable or nightly Dynamo build, a backend version, and an install form (container / wheel), and get the exact, copy-pastable command.

- **Pure CSS, no client JavaScript.** A server component  (`fern/components/InstallSelector.tsx`) expands generated data into hidden  radios + `[data-*]` command blocks + a `:has()` stylesheet that reveals exactly  one block per selection — mirroring the existing `RecipeStyles.tsx` pattern.
-  **Dev-only; release snapshots stay frozen.** The interactive selector renders only in  `dev` docs. When a versioned snapshot (`vX.Y.Z`) is cut, the release job runs  `freeze_install_selector.py`, which replaces the selector with static  `docker run …:<version>` tabs — so release docs show correct, pinned commands with no  live or nightly content.
- **Complete commands from live sources.**  `.github/scripts/build_install_selector.py` builds the data from  `container/context.yaml` git history + the public NGC tag list + the  `pypi.nvidia.com` index (no credentials). Nightly pins resolve to the exact  published `YYYYMMDD-<sha>` container tag and `X.Y.Z.devYYYYMMDD` wheel; GC'd or  skipped nights never appear, so a dead command is never emitted.
- **Scope:** the 3 most recent stable releases (each shown as its own option) and  the last 30 days of nightlies.
- **Generated at docs-build time, not committed** (gitignored). `fern-docs.yml`  runs the generator before `fern generate`; a daily `schedule` keeps the rolling  window current with no source commits.

Also adds a **Nightly Release Info** reference page (nightly containers, wheels, and install patterns), with the interactive backend-version → build lookup handled by the selector. The Support Matrix `main (ToT)` row is refreshed at docs-build time by the same generator (`--refresh-support-matrix`).

closes: OPS-7770

## Summary by CodeRabbit

- **New Features**
  - Added an interactive install selector for choosing backend, hardware, and release version.
  - Added copyable Docker installation commands for stable and nightly builds.
  - Added a Nightly Release Info reference page with artifact details and installation guidance.
  - Added daily updates for nightly release information.

- **Documentation**
  - Updated the Quickstart to use the install selector.
  - Expanded release artifact documentation and cross-links.
  - Clarified nightly image tags, wheel installation, version pinning, and shipped artifacts.


nealvaidya:
For some of the longer container names, the copy button overlaps, makes it hard to see the tag name 
image

nv-anants:
> For some of the longer container names, the copy button overlaps, makes it hard to see the tag name

thanks @nealvaidya , add some word warp. Looks better now - https://nvidia-preview-019f86a2-9e42-7094-bbc2-c7229ab6197c.docs.buildwithfern.com/dynamo/dev/getting-started/quickstart#get-dynamo

nealvaidya (approved):

.github/scripts/freeze_install_selector.py
```diff
@@ -0,0 +1,66 @@
...
+        "Containers have all dependencies pre-installed. Pick your backend:\n\n<Tabs>\n"
+        + body
+        + "\n</Tabs>"
+    )
+
+
+def freeze(text: str, version: str) -> str:
+    if "BEGIN:install-selector" not in text:
+        return text
+    text = re.sub(
```
dagil-nvidia: **Silent-failure risk (would ship broken release docs):** `freeze()` checks for the `BEGIN` marker but never verifies the substitution happened. If the `END:install-selector` marker is missing or malformed, this `re.sub` matches nothing while the import-strip below still runs — the released Quickstart then renders a literal `<InstallSelector />` as "Unsupported JSX tag" (the failure mode the component's own docstring warns about), and `fern check` won't catch it.

Suggest `re.subn` + `raise SystemExit` if `BEGIN` is present but the replacement count is 0 (same for the import strip). Also consider printing a warning in the missing-file branch of `main()` so a path typo in the workflow doesn't no-op silently forever — today it's indistinguishable from the intended older-snapshot case.

.github/workflows/fern-docs.yml
```diff
@@ -116,6 +125,16 @@ jobs:
           path: docs-checkout
           fetch-depth: 1
           token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate install-selector data
+        # Writes the generated TS module into fern/components/ (picked up by the
+        # component sync below). Needs full history and tags to read
+        # container/context.yaml history, release tags, and nightly commit versions.
+        run: |
+          git -C source-checkout fetch --unshallow --tags 2>/dev/null || git -C source-checkout fetch --tags
```
dagil-nvidia, line 134: **Silent shallow-history degradation:** if `--unshallow` fails for a transient reason (not just "already complete"), the `|| git fetch --tags` fallback leaves a depth-1 clone. `read_history()` then sees one commit, the nightly window list silently collapses to just the current version, and the docs publish anyway.

Suggest checking `git rev-parse --is-shallow-repository` after the fetch and failing the step, rather than sending stderr to /dev/null.

.github/scripts/build_install_selector.py
```diff
@@ -0,0 +1,497 @@
...
+
+def nightly_latest_commands(
+    label: str, wheel_version: str | None = None
+) -> dict[str, str]:
+    img, extra, wheel = META[label]
+    cmds = {"container": run(f"{REGISTRY}/{img}-runtime-nightly:latest")}
+    if wheel:
+        # Pin the wheel: --extra-index-url pools with public PyPI, so an unpinned
+        # --pre install can resolve a stable release instead of the nightly.
+        pin = f"=={wheel_version}" if wheel_version else ""
```
dagil-nvidia: **Unpinned fallback contradicts the script's guarantee:** when the NGC lookup fails (`hit is None` upstream), `wheel_version` is `None` and this emits `uv pip install --pre --extra-index-url … "ai-dynamo[extra]"` with no pin — and the comment right above explains that exact command can resolve a *stable* release instead of a nightly. Since the module docstring promises "a dead command is never emitted", suggest omitting the wheel command when there's no pin, like `nightly_pinned_commands` already does.

Related: wheel existence is never verified against `pypi.nvidia.com` — a night where the container published but the wheel upload failed still yields a `==X.Y.Z.devYYYYMMDD` command that 404s. A HEAD check (or a docstring caveat) would close that.

.github/workflows/fern-docs.yml
```diff
@@ -478,6 +498,13 @@ jobs:
             done
           fi
 
+      - name: Freeze the install selector in the versioned Quickstart
+        # Dev keeps the interactive selector; a release snapshot is static and pinned.
+        run: |
+          python3 source-checkout/.github/scripts/freeze_install_selector.py \
```
dagil-nvidia, line 514: **Implicit cross-branch coupling worth a comment:** on tag builds the generate step doesn't run, so the dev version's `import { INSTALL_DATA } from "./install-selector-data"` only resolves because a *prior dev sync* committed the generated module to `docs-website`. That works (data is stale-but-present during the tag publish), but it breaks if `docs-website`'s ignore rules ever pick up the new root-`.gitignore` entry. A one-line comment here noting the dependency would save the next person a debugging session.

fern/components/InstallSelector.tsx
```diff
@@ -0,0 +1,228 @@
...
+    ? "Latest stable release that supports this version"
+    : e.latest
+      ? "Latest nightly build"
+      : "Latest nightly in this version's range";
+  const ships = `Supports ${label} ${e.backend_version}`;
+
+  const cmd = (e.commands as Record<string, string>)[form];
+  const hint = !isStable && e.note ? `<p class="is-hint">${esc(e.note)}</p>` : "";
+  const body = cmd
+    ? `<div class="is-cmd"><button type="button" class="is-copy" onclick="navigator.clipboard&&navigator.clipboard.writeText(this.closest('.is-cmd').querySelector('pre').innerText);this.textContent='Copied!';setTimeout(()=>{this.textContent='Copy'},1200)">Copy</button><pre>${esc(cmd)}</pre>${hint}</div>`
```
dagil-nvidia, line 80: **"No client JavaScript" isn't quite true:** the selection logic is pure CSS, but this Copy button is an inline `onclick` handler injected via `dangerouslySetInnerHTML`. Two consequences: (1) the file/PR claim is inaccurate — suggest softening to "no bundled/hydrated JS; copy uses an inline handler"; (2) inline handlers are the first thing a CSP without `unsafe-inline` blocks — worth verifying the Fern-hosted site's CSP allows it, since the button would silently do nothing otherwise.

.github/scripts/build_install_selector.py
```diff
@@ -0,0 +1,497 @@
...
+    return subprocess.check_output(["git", *args], cwd=repo_root, text=True)
+
+
+def read_current(repo_root: Path) -> dict[str, tuple[str, str]]:
+    """{label: (version, raw_runtime_image_tag)} from context.yaml at HEAD."""
+    doc = yaml.safe_load((repo_root / CONTEXT).read_text())
+    out: dict[str, tuple[str, str]] = {}
+    for fw in FRAMEWORKS:
+        tag = tag_from_context(doc, fw)
+        if tag is None:
```
dagil-nvidia, line 110: **`read_current` skips the `version_re` filter that `read_history` applies:** the history walk explicitly drops pre-layout base-image tags (the `FRAMEWORKS` comment cites vLLM `13.0.2`), but this function — which feeds the Support Matrix ToT row via `--refresh-support-matrix` — takes `runtime_image_tag` at HEAD unvalidated. If `context.yaml` ever again holds a non-backend tag, the published ToT row gets a bogus version. Suggest applying `fw.version_re` here and failing loudly.

.github/scripts/build_install_selector.py
```diff
@@ -0,0 +1,497 @@
...
+        t
+        for t in ngc_tag_list(f"nvidia/ai-dynamo/{image}-runtime")
+        if re.fullmatch(r"\d+\.\d+\.\d+", t)
+    }
+
+
+# --------------------------------------------------------------------------- #
+# stable map (support-matrix.md), limited to the newest releases
+# --------------------------------------------------------------------------- #
+_ROW = re.compile(
```
dagil-nvidia, line 238: **Silent mis-mapping if the table is ever reordered:** `_ROW` hard-codes the Backend Dependencies column order (SGLang, TensorRT-LLM, vLLM). `TOT_RE` fails loudly on drift, but this one mis-maps *silently* if a column is reordered or inserted — every stable command would then reference the wrong backend version. Parsing (or at least asserting) the header row would close this.

dagil-nvidia (commented):
Reviewed the full diff — the architecture here is right (generate at docs-build time, freeze at release), and the resilience details (NGC fallback warnings, `TOT_RE.subn` count check, escaped HTML interpolation, the `always()`/`needs.result` gating for scheduled runs) are well done.

I left 7 inline comments. The one I'd ask for before merge is the `freeze_install_selector.py` substitution-count check — as written it can silently ship broken release docs. The rest are cheap hardening or follow-ups.

**Minor (no inline comment):**
- `reltags_by_img: dict[str, set] = {}` is assigned `None` in the except branch — annotation should be `dict[str, set[str] | None]` (mypy will flag it).
- `ngc_tag_list` doesn't handle Docker Registry v2 pagination (`Link` header / `n` param). If NGC caps the tag list, older nightlies vanish silently — worth verifying against the live repo.
- If a framework's `stable` list ends up empty after NGC filtering, the selector shows a version row with no chips and no output block — no "nothing available" message.
- `quickstart.mdx`: `../getting-started/local-installation.md` from a file already in `getting-started/` works but `local-installation.md` is cleaner.
- In versioned snapshots, `nightly-release-info.md` still says "use the backend selector in the Quickstart", but the frozen Quickstart has no selector.

**Test coverage:** none for either script, and both sit on release-critical paths. `freeze()`, `update_tot()`, `stable_by_framework()`, and `latest_in_window()` are pure functions — cheap to unit test. Also, the `--check` flag exists but nothing in CI invokes it; either wire it into a docs CI job or drop it.

.github/scripts/freeze_install_selector.py
nv-anants: Fixed — freeze() uses re.subn now and raises SystemExit if the BEGIN marker is present but the block substitution or import strip isn't exactly 1; main() also warns on a missing file. So a missing/duplicate END marker fails the release build instead of shipping a literal

.github/workflows/fern-docs.yml
nv-anants: Added — after the fetch it checks git rev-parse --is-shallow-repository and fails the step if still shallow, so the nightly window can't silently collapse.

.github/scripts/build_install_selector.py
nv-anants: Both addressed — the wheel command is omitted when there's no pin (matching nightly_pinned_commands), and pins are verified against the pypi.nvidia.com index (pypi_dev_versions() → wheel_version_for()), so a container-only night no longer emits a 404ing wheel.

.github/scripts/build_install_selector.py
nv-anants: Fixed — read_current now applies fw.version_re and raises if a tag isn't a recognized backend version, so a stray tag can't publish a bogus ToT row.

.github/scripts/build_install_selector.py
nv-anants: Added a _HEADER_RE assertion — stable_by_framework fails loudly if the Backend Dependencies header order drifts, same as TOT_RE.

.github/workflows/fern-docs.yml
nv-anants: Added a comment on the freeze step noting the tag build doesn't regenerate the module — the live dev pages resolve their import from the module a prior dev sync committed to docs-website.

fern/components/InstallSelector.tsx
nv-anants: Softened the docstring — selection is pure CSS with no bundled/hydrated JS, and the copy button is called out as an inline onclick with the CSP unsafe-inline caveat. I couldn't verify the hosted site's CSP locally; if the preview's Copy button works we're covered — let me know if you know the policy.

.github/scripts/build_install_selector.py
```diff
@@ -0,0 +1,527 @@
...
+    device: str  # sub-key holding runtime_image_tag
+    version_re: "re.Pattern[str]"  # accepts only real backend tags
+
+
+# Device keys track the CUDA targets nightly builds. version_re drops pre-layout
+# base-image tags (e.g. vLLM 13.0.2) so the history walk keeps only real versions.
+FRAMEWORKS = [
+    Framework("vLLM", "vllm", "cuda13.0", re.compile(r"^v\d+\.\d+\.\d+")),
+    Framework("SGLang", "sglang", "cuda13.0", re.compile(r"^v\d+\.\d+")),
+    Framework("TensorRT-LLM", "trtllm", "cuda13.1", re.compile(r"^\d+\.\d+\.\d+rc\d+")),
```
pvijayakrish: The TRT-LLM version regex `r"^\d+\.\d+\.\d+rc\d+"` only matches RC tags. For GA releases it might silently drop from the stable selector.

nv-anants: thanks! fixed in https://github.com/ai-dynamo/dynamo/pull/11950/changes/8c1b0a0786eb7a38f611ba7cb0b02855a25cd806

pvijayakrish (approved):

PR #11951, opened 2026-07-21
fix(helm): disable bundled NATS by default

### LINKED ISSUE #11921: fix(helm): stop installing NATS by default
## Problem

The platform Helm chart still sets `global.nats.install: true`. That installs bundled NATS, configures the operator with its address, and causes the operator to inject `NATS_SERVER` into generated workloads.

This default was introduced in #9232/#9281 because the runtime event plane then defaulted to NATS for distributed discovery backends. Since #10902, the runtime defaults are TCP for the request plane and ZMQ for the event plane across all discovery backends. The chart comment still states the old NATS-default rationale.

The injected environment variable is not inert: the runtime treats the presence of `NATS_SERVER` as an explicit request to enable NATS and eagerly opens a connection, even when the default TCP/ZMQ transports do not use it.

The older use of NATS object storage for model metadata was already removed in #3659.

## Expected behavior

- Set `global.nats.install` to `false` by default.
- Continue supporting bundled NATS through explicit `global.nats.install=true`.
- Continue supporting external NATS through `dynamo-operator.natsAddr`.
- Continue injecting `NATS_SERVER` only when one of those NATS configurations is explicitly supplied.
- Update chart documentation and tests to reflect the opt-in default.

NATS remains available for explicit features such as `DYN_EVENT_PLANE=nats`, legacy `DYN_REQUEST_PLANE=nats`, and NATS request-trace sinks.

julienmancuso:
## Overview

Dynamo now defaults to TCP for the request plane and ZMQ for the event plane, so standard deployments no longer require NATS. However, the platform Helm chart still installed bundled NATS and injected `NATS_SERVER` into workloads by default.

## Summary

Disable bundled NATS by default while preserving explicit NATS configurations for legacy transports, NATS event planes, request-trace sinks, and external NATS servers.

## Details

- Change `global.nats.install` from `true` to `false`.
- Preserve opt-in bundled NATS through `global.nats.install=true`.
- Preserve external NATS configuration through `dynamo-operator.natsAddr`.
- Continue injecting `NATS_SERVER` whenever bundled or external NATS is explicitly configured.
- Change the operator Tilt workflow’s `enable_nats` default to `false`.
- Update Helm tests for the new default and explicit NATS configurations.
- Update and regenerate the platform chart documentation.

### Upgrade impact

This is a breaking default change for releases that implicitly relied on bundled NATS.

Accepting the new default during an upgrade removes the bundled NATS resources and removes `NATS_SERVER` from reconciled workload specifications, which can trigger rolling updates.

Deployments that need to preserve the previous behavior can avoid the NATS-related rollout by explicitly setting:

```bash
helm upgrade <release> <chart> \
  --set global.nats.install=true
```

Deployments using an external NATS server should continue setting `dynamo-operator.natsAddr`.

- `deploy/helm/charts/platform/values.yaml`
- `deploy/helm/charts/platform/tests/nats_config_test.yaml`
- `deploy/helm/charts/platform/README.md.gotmpl`

## Validation

- `make -C deploy/helm/charts/platform generate-helm-docs`
- `make -C deploy/helm/charts/platform lint`
- `make -C deploy/helm/charts/platform test`
  - 7 test suites passed.
  - 30 tests passed.
- Verified that a default Helm render contains no bundled NATS resources, `natsAddress`, or `NATS_SERVER`.
- `git diff --check`

## Related Issues

- Closes #11921

## Summary by CodeRabbit

- **New Features**
  - NATS is now optional and disabled by default for standard TCP/ZMQ request and event transports.
  - Added support and guidance for enabling bundled NATS or connecting to an external NATS service.
- **Documentation**
  - Updated configuration guidance, transport behavior, and upgrade instructions.
  - Added notes about potential workload updates when transitioning from bundled NATS.
- **Bug Fixes**
  - Improved configuration handling so external NATS addresses are applied when bundled NATS is disabled.


deploy/helm/charts/platform/README.md
```diff
@@ -93,7 +115,7 @@ Kubernetes: `>=1.30.0-0`
...
 | global.etcd.install | bool | `false` | Whether this chart should install the bundled etcd subchart. When true, deploys etcd and auto-configures the operator with its address. When false, etcd is not deployed. Use dynamo-operator.etcdAddr to point at an external instance if you are bringing your own etcd. |
-| global.nats.install | bool | `true` | Whether this chart should install the bundled NATS subchart. When true, deploys NATS and auto-configures the operator with its address. When false, NATS is not deployed. Use dynamo-operator.natsAddr to point at an external instance if you are bringing your own NATS. Defaults to true (since release 1.1.0) because the Dynamo runtime's event plane (DYN_EVENT_PLANE) defaults to NATS for distributed backends (etcd/kubernetes). |
+| global.nats.install | bool | `false` | Whether this chart should install the bundled NATS subchart. When true, deploys NATS and auto-configures the operator with its address. When false, NATS is not deployed. Use dynamo-operator.natsAddr to point at an external instance if you are bringing your own NATS. NATS is optional because the default request and event planes use TCP and ZMQ, respectively. |
```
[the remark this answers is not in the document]
julienmancuso: fixed

nnshah1 (approved):

MR_DESCRIPTION.md
nnshah1, line 1: I think this would be deleted before merge?

nnshah1: Also seems unrelated

deploy/operator/internal/dynamo/graph_test.go
```diff
@@ -1714,6 +1714,48 @@ func Test_mergeEnvs(t *testing.T) {
 	}
 }
 
+func TestAddStandardEnvVars_NATS(t *testing.T) {
```
tmonty12, line 1717: nit: test feels quite specific to the PR. i think we should make it more generic with a table test containing operatorConfiguration and envs (stresses the other configured URLs) or just remove this test

deploy/helm/charts/platform/README.md
```diff
@@ -41,6 +41,28 @@ The Dynamo Platform Helm chart deploys the complete Dynamo Kubernetes Platform i
 
 ## 🔄 Upgrading Notes
 
+### Bundled NATS is now disabled by default
```
tmonty12, line 44: nice

tmonty12 (approved):

deploy/operator/internal/dynamo/graph_test.go
julienmancuso: removed

MR_DESCRIPTION.md
julienmancuso: fixed

PR #11957, opened 2026-07-21
test: guard TRT-LLM engine arg preservation

### LINKED ISSUE #9288: [CONTRIBUTION]: [FEATURE]: Enforce user-config preservation in TRT-LLM worker arg_map
### Type of Change

Refactoring

### Problem Statement

  ## Feature request

  Add programmatic guardrails so user-supplied engine arguments (YAML, `--extra-engine-args`, `--override-engine-args`) cannot be silently clobbered by Dynamo defaults when building the TRT-LLM `arg_map` in `components/src/dynamo/trtllm/workers/llm_worker.py`.                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ## Describe the problem you're encountering
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  The Dynamo→TRT-LLM config bridge in `init_llm_worker` is leaky. User YAML is parsed into `arg_map`, then `arg_map` is mutated in several places (`if config.publish_events_and_metrics:` block, dict↔`KvCacheConfig` conversion, `override_engine_args` deep-update) before being handed to `LLM(...)`. Nothing verifies that user-supplied values actually survive.                                                                                                                                
   
  This same six-line block has regressed three times in the past few releases:                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                  
  1. `84c7d1e234` (Sep 2025) — dict→`KvCacheConfig` refactor dropped the `if not event_buffer_max_size` guard, causing user-supplied buffer sizes to be silently overwritten with the 1024 default.                                                                                                                                                                                                                                                                                                   
  2. PR [#5198](https://github.com/ai-dynamo/dynamo/pull/5198) (Jan 2026) — partially fixed: preserved most `KvCacheConfig` settings via `model_dump`, but did not restore the conditional on `event_buffer_max_size` itself.
  3. PR [#9284](https://github.com/ai-dynamo/dynamo/pull/9284) — fixes `event_buffer_max_size` properly, but in the same change drops `arg_map["return_perf_metrics"] = config.publish_events_and_metrics`. After that PR, vanilla `--publish-events-and-metrics` deployments (without OTEL launch scripts that explicitly inject `return_perf_metrics: true`) silently lose TRT-LLM's `PerfMetricsManager` instrumentation — GPU forward/sample timing, `step_metrics`, `ctx_chunk_metrics`,         
  `time_breakdown` analysis, OTEL forward/sample spans. None of this is caught by existing tests.                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  The recurring pattern: Dynamo's internal massaging of `arg_map` clobbers or fails to forward user intent, and the failure mode is silent (no warning, no exception, no log), so it ships and is only discovered when someone notices missing data in production.                                                                                                                                                                                                                                    
                                                                  
  ## Describe alternatives you've tried                                                                                                                                                                                                                                                                                                                                                                                                                                                               
                                                                  
  - Adding individual conditional guards per field (the approach used in #5198 and #9284) — fixes one field at a time; doesn't prevent the next regression on a different field.                                                                                                                                                                                                                                                                                                                      
  - Documentation (`docs/backends/trtllm/trtllm-observability.md:212` documents the `return_perf_metrics` contract) — does not enforce.
  - Code review — has not caught the recurring class of bug.                                                                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  A systemic guardrail in the worker init pipeline would prevent the next instance of this bug class.                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                            

### Proposed Solution

Concretely, two complementary mechanisms:                                                                                                                                                                                                                                                                                                                                                                                                                                                           
                                                                  
  1. **Audit log of user-vs-default provenance.** Either snapshot the post-YAML `arg_map` and diff against the final `arg_map` before `LLM(**arg_map)`, or track per-key provenance (`{"event_buffer_max_size": "user_yaml", "backend": "dynamo_default", ...}`). Emit a warning whenever a Dynamo-default code path overwrites a user-supplied key. The existing `_warn_override_collisions` helper at `llm_worker.py:98` already implements this pattern in one direction (warning when             
  `override_engine_args` clobbers a value) — apply it across the whole pipeline.
  2. **Regression test pinning.** Feed YAML configs with non-default values for the historically-fragile fields (`kv_cache_config.event_buffer_max_size`, `kv_cache_config.free_gpu_memory_fraction`, `kv_cache_config.cache_transceiver_config`, `return_perf_metrics`, `backend`, `enable_iter_perf_stats`) and assert they survive end-to-end into the final `arg_map` produced by the worker init.                      

### Estimated PR Size

M (51-200 lines)

### Files/Components Affected

TRTLLM

cpakkamisaac-sae:
## Overview
Adds guardrails for TRT-LLM engine argument preservation when KV event publishing is enabled.

## Details
- Reuses the shared `warn_override_collisions` helper in `llm_worker` instead of the worker-local duplicate.
- Teaches collision warnings to recurse through model-like config objects that expose `model_dump()`, so nested overrides report precise dotted paths.
- Adds regression coverage for `--publish-kv-events` defaults and user-provided final engine args, including `backend`, `return_perf_metrics`, `enable_iter_perf_stats`, `kv_cache_config.event_buffer_max_size`, `kv_cache_config.free_gpu_memory_fraction`, and nested `cache_transceiver_config`.
- Refactors the existing engine-arg capture mocks into a helper used by the related worker tests.

## Where should reviewers start?
Start with `components/src/dynamo/trtllm/tests/test_trtllm_unit.py`, then check the small utility reuse in `components/src/dynamo/trtllm/workers/llm_worker.py` and `components/src/dynamo/trtllm/utils/trtllm_utils.py`.

## Related Issues
Fixes #9288

## Testing
- `python3 -m py_compile components/src/dynamo/trtllm/workers/llm_worker.py components/src/dynamo/trtllm/utils/trtllm_utils.py components/src/dynamo/trtllm/tests/test_trtllm_unit.py`
- `git diff --check`
- `PYTHONPATH=components/src python3 -c "... warn_override_collisions ..."`

Focused pytest was not run locally because the available environment does not have `pytest`/TensorRT-LLM installed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved configuration override warnings to detect collisions within nested model-based configuration values.
  - Ensured resolved TRT-LLM engine settings remain consistent with configured overrides.
  - Improved handling of KV event publishing, including default event and metrics settings while preserving user-provided engine options.

- **Tests**
  - Expanded coverage for nested configuration warnings, engine argument resolution, and KV event publishing scenarios.

### DIFF
--- a/components/src/dynamo/trtllm/tests/test_trtllm_unit.py
+++ b/components/src/dynamo/trtllm/tests/test_trtllm_unit.py
@@ -7,6 +7,7 @@ import asyncio
 import os
 import re
 import warnings
+from dataclasses import dataclass
 from pathlib import Path
 from unittest import mock
 
@@ -22,9 +23,13 @@ if not torch.cuda.is_available():
 
 from dynamo.trtllm.args import Config, parse_args
 from dynamo.trtllm.constants import DisaggregationMode, Modality
+from dynamo.trtllm.engine import Backend
 from dynamo.trtllm.tests.conftest import make_cli_args_fixture
 from dynamo.trtllm.utils.trtllm_utils import deep_update, warn_override_collisions
-from dynamo.trtllm.workers.llm_worker import init_llm_worker
+from dynamo.trtllm.workers.llm_worker import (
+    DEFAULT_KV_EVENT_BUFFER_MAX_SIZE,
+    init_llm_worker,
+)
 
 # Get path relative to this test file
 REPO_ROOT = Path(__file__).resolve().parents[5]
@@ -317,6 +322,46 @@ def test_deep_update_adds_new_keys():
     assert target == {"a": 1, "b": 2, "c": {"nested": 3}}
 
 
+def test_deep_update_merges_model_dump_configs():
+    """deep_update preserves sibling settings from model-like configs."""
+
+    class ModelDumpConfig:
+        def model_dump(self):
+            return {
+                "event_buffer_max_size": 0,
+                "free_gpu_memory_fraction": 0.85,
+            }
+
+    target = {"kv_cache_config": ModelDumpConfig()}
+    source = {"kv_cache_config": {"event_buffer_max_size": 4096}}
+    deep_update(target, source)
+    assert target == {
+        "kv_cache_config": {
+            "event_buffer_max_size": 4096,
+            "free_gpu_memory_fraction": 0.85,
+        }
+    }
+
+
+def test_deep_update_merges_dataclass_configs():
+    """deep_update preserves sibling settings from dataclass configs."""
+
+    @dataclass
+    class DataclassConfig:
+        event_buffer_max_size: int = 0
+        free_gpu_memory_fraction: float = 0.85
+
+    target = {"kv_cache_config": DataclassConfig()}
+    source = {"kv_cache_config": {"event_buffer_max_size": 4096}}
+    deep_update(target, source)
+    assert target == {
+        "kv_cache_config": {
+            "event_buffer_max_size": 4096,
+            "free_gpu_memory_fraction": 0.85,
+        }
+    }
+
+
 # ---- Tests for trtllm_utils.warn_override_collisions ----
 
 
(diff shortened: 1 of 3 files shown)


cpakkamisaac-sae:
Closing this older PR because it has diverged from current TRT-LLM work and overlaps newer active fixes. Any remaining #9288 scope can be revisited as a focused change against current main.

PR #11960, opened 2026-07-21
refactor(mocker): share live scheduler loop

connorcarpenter15:
## Overview:

Extract the duplicated Mocker vLLM and SGLang live scheduler actor loops into one shared implementation. This is the first prerequisite split from #11917; request cancellation and the public live request engine remain outside this PR.

## Details:

- Move request intake, pass timing, command handling, lifecycle publication, metrics, and progress waits into `scheduler/live_boundary.rs`.
- Keep the existing role-derived `controls_enabled` behavior and unbounded output channel semantics.
- Reduce the vLLM and SGLang live modules to scheduler wrappers and engine-specific hooks.
- Relocate the three existing boundary-publication tests without expanding their coverage.
- Fix shutdown observation between continuously productive zero-duration passes in a separate commit with one focused regression test.
- Do not add `CancelRequest`, a cancellation lane, bounded output plumbing, or `LiveEngine`.

- `lib/mocker/src/scheduler/live_boundary.rs` for the shared actor loop.
- `lib/mocker/src/scheduler/vllm/live.rs` and `lib/mocker/src/scheduler/sglang/live.rs` for the engine-specific adapters.
- `lib/mocker/src/scheduler/live_boundary/tests.rs` for the relocated ordering tests and shutdown regression.

## Related Issues

- [DIS-2473: Expose Mocker as a live request engine for sidecar adapters](https://linear.app/nvidia/issue/DIS-2473/expose-mocker-as-a-live-request-engine-for-sidecar-adapters)
- Relates to #11873

## Testing

- `cargo test -p dynamo-mocker --lib -- --skip services::bootstrap::tests` (512 passed, 1 pre-existing ignored)
- Focused live-boundary, vLLM live-scheduler, and SGLang lifecycle tests
- KVBM-offload test attempted locally but blocked during dependency build because the host has CUDA 11; KVBM requires CUDA 12+

## Summary by CodeRabbit

- **Improvements**
  - Improved live scheduling coordination across supported runtime backends.
  - Ensured outputs, lifecycle events, metrics, acknowledgements, and related updates are published in a consistent order.
  - Improved handling of deferred work and internal scheduling deadlines for smoother progress.
  - Added more reliable cancellation and shutdown behavior, including protection against endless zero-duration processing loops.

- **Tests**
  - Added coverage for publication ordering, command-driven effects, mid-pass acknowledgements, and clean shutdown behavior.

### DIFF
--- a/lib/mocker/src/scheduler/live_boundary.rs
+++ b/lib/mocker/src/scheduler/live_boundary.rs
@@ -199,6 +199,11 @@ async fn run_live_scheduler<C: LiveBoundaryCore>(
     let mut deferred_commands = VecDeque::new();
 
     loop {
+        // Productive zero-duration passes may never enter one of the
+        // cancellation-aware waits below.
+        if cancel_token.is_cancelled() {
+            break;
+        }
         if !receive_until_live_schedulable(
             core,
             &mut request_rx,
--- a/lib/mocker/src/scheduler/live_boundary.rs
+++ b/lib/mocker/src/scheduler/live_boundary.rs
@@ -1,16 +1,28 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::common::protocols::{FpmPublisher, KvEventPublishers, OutputSignal};
+use std::collections::VecDeque;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use crate::common::protocols::{
+    DirectRequest, FpmPublisher, KvEventPublishers, MockEngineArgs, OutputSignal,
+};
+use crate::common::utils::sleep_until_precise;
 use crate::scheduler::kv_event_sink::{
-    DeferredKvPublish, DeferredKvPublishBuffer, publish_deferred_fpm, publish_deferred_kv_events,
+    DeferredKvPublish, DeferredKvPublishBuffer, capture_deferred_kv_publish_sink,
+    publish_deferred_fpm, publish_deferred_kv_events,
 };
 use crate::scheduler::vllm::MockerMetrics;
 use crate::scheduler::{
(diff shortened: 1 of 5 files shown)


lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -53,6 +79,214 @@ pub(crate) trait LiveBoundaryCore {
...
+    cancel_token: CancellationToken,
+    controls_enabled: bool,
+) {
+    let scheduler_start = Instant::now();
+    let mut deferred_commands = VecDeque::new();
+
+    loop {
+        // Productive zero-duration passes may never enter one of the
+        // cancellation-aware waits below.
+        if cancel_token.is_cancelled() {
```
PeaBrane, line 204: Non-blocking: this observes cancellation once the token is set, but a productive zero-duration loop can still starve an external canceller on a current-thread runtime. Zero modeled duration does not necessarily mean zero progress, so fixing this correctly requires care rather than treating every zero-duration pass as idle. Since the behavior is pre-existing, I’d prefer preserving it in this refactor and fixing it separately. Could you leave a TODO here for cooperative scheduling and an external-canceller regression test?

PeaBrane (approved):
Nice cleanup—the shared boundary centralizes live-loop invariants while preserving the vLLM/SGLang-specific timing and KVBM hooks. Offline replay bypasses this actor and drives `EngineCore` directly, so I don’t think the offline parity/performance campaign is a meaningful gate here. One non-blocking TODO below.

lib/mocker/src/scheduler/live_boundary.rs
connorcarpenter15: Good catch. I agree this is pre-existing behavior and that this refactor should preserve the productive zero-duration semantics rather than fold in a cooperative-scheduling change. I’ll handle the yield behavior and add a current-thread external-canceller regression in the immediate cancellation follow-up, #11964, where external cancellation becomes part of the scheduler contract. That keeps the behavioral change and its test together, so I’m going to leave this PR unchanged rather than add a TODO that the next PR would immediately remove.

PR #11964, opened 2026-07-21
feat(mocker): add targeted request cancellation

connorcarpenter15:
## Overview:

Add targeted request cancellation to the shared live Mocker scheduler as a follow-up to #11960.

This is the cancellation-only prerequisite for sidecar-facing Mocker adapters. It intentionally excludes the `LiveEngine`, bounded output channels, and protocol-specific adapters.

Scope note: this PR stops at the `lib/mocker` live-scheduler boundary. The existing `lib/llm::MockEngine` context-stop and response-stream-drop paths do not yet call the new cancellation lane. That end-to-end integration is tracked by [DIS-2478](https://linear.app/nvidia/issue/DIS-2478/wire-mockengine-lifecycle-cancellation-into-the-mocker-scheduler), including disaggregated cancellation coverage.

## Details:

- Add an idempotent `CancelRequest` scheduler command for vLLM and SGLang requests.
- Add a dedicated cancellation lane that remains responsive during modeled passes and in aggregated workers without enabling general lifecycle controls there.
- Cooperatively yield during productive zero-duration passes so external cancellation and shutdown tasks can run on current-thread Tokio runtimes.
- Remove cancelled scheduler state and suppress pending output immediately. During a modeled pass, published running/waiting metrics refresh at the next pass boundary; exact mid-pass metrics are deferred rather than adding an O(all resident requests) snapshot to every pass.
- Update offline replay accounting for applied request cancellations.
- Cover waiting/running cancellation for both engines, mid-pass cancellation at the shared live boundary, explicit pending-output discard after an idempotent `Noop`, and offline in-flight cleanup.
- Replace the inherited self-triggered shutdown test with current-thread external-canceller regression coverage.

- `lib/mocker/src/scheduler/live_boundary.rs`
- `lib/mocker/src/scheduler/vllm/core.rs`
- `lib/mocker/src/scheduler/sglang/core.rs`

## Related Issues

- [DIS-2478: Wire MockEngine lifecycle cancellation into the Mocker scheduler](https://linear.app/nvidia/issue/DIS-2478/wire-mockengine-lifecycle-cancellation-into-the-mocker-scheduler)
- [DIS-2473: Expose Mocker as a live request engine for sidecar adapters](https://linear.app/nvidia/issue/DIS-2473/expose-mocker-as-a-live-request-engine-for-sidecar-adapters)
- Depends on #11960
- Relates to #11873

## Testing

- `cargo test -p dynamo-mocker` (525 passed, 1 ignored)

## Summary by CodeRabbit

* **New Features**
  * Added request cancellation support across live scheduling workflows.
  * Cancellations now work for requests that are waiting or actively running.
  * Pending output can be discarded when a request is cancelled.

* **Bug Fixes**
  * Corrected in-flight request accounting after successful cancellation.
  * Prevented cancelled requests from delivering pending output.

### DIFF
--- a/lib/mocker/src/scheduler/live_boundary.rs
+++ b/lib/mocker/src/scheduler/live_boundary.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::VecDeque;
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
@@ -24,12 +24,6 @@ use crate::scheduler::{
 use tokio::sync::{mpsc, watch};
 use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
-
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub(crate) enum RequestResidency {
-    Running,
-    Waiting,
-}
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum PublishedEffect {
@@ -64,8 +58,6 @@ pub(crate) trait LiveBoundaryCore {
     fn live_metrics(&self) -> MockerMetrics;
 
     fn pass_boundary_metrics(&self, pass_metrics: MockerMetrics) -> MockerMetrics;
-
-    fn visit_live_request_residencies(&self, visitor: &mut dyn FnMut(Uuid, RequestResidency));
 
     fn live_internal_deadline_ms(&self) -> Option<f64> {
         None
@@ -307,7 +299,6 @@ pub(crate) struct LiveEffectsPublisher {
     kv_event_publishers: KvEventPublishers,
     fpm_publisher: FpmPublisher,
     captured_kv_events: DeferredKvPublishBuffer,
-    visible_residencies: Mutex<HashMap<Uuid, RequestResidency>>,
     #[cfg(test)]
     publication_log: Option<std::sync::Arc<std::sync::Mutex<Vec<PublishedEffect>>>>,
 }
@@ -641,7 +632,6 @@ impl LiveEffectsPublisher {
             kv_event_publishers,
             fpm_publisher,
             captured_kv_events,
-            visible_residencies: Mutex::new(HashMap::new()),
             #[cfg(test)]
             publication_log: None,
         }
@@ -654,11 +644,6 @@ impl LiveEffectsPublisher {
     ) -> Self {
         self.publication_log = Some(log);
         self
-    }
-
-    #[cfg(test)]
-    fn published_metrics(&self) -> MockerMetrics {
-        self.metrics_tx.borrow().clone()
     }
 
     pub(crate) fn capture_pass(&self, pass: EnginePassResult) -> PendingLivePass {
@@ -731,7 +716,6 @@ impl LiveEffectsPublisher {
         // disappeared. Those cleanup events belong to this same boundary.
         self.publish_router_effects(self.captured_kv_events.drain(), None);
         self.publish_lifecycle(pending.pass.lifecycle_events).await;
-        self.refresh_visible_residencies(core);
         let metrics = core.pass_boundary_metrics(pending.pass.mocker_metrics);
         self.record(PublishedEffect::Metrics);
         let _ = self.metrics_tx.send(metrics);
@@ -772,10 +756,6 @@ impl LiveEffectsPublisher {
         now_ms: f64,
     ) -> Option<SchedulerCommandResult> {
         let SchedulerCommandEnvelope { command, reply } = envelope;
-        let cancellation_id = match &command {
-            SchedulerCommand::CancelRequest { request_id } => Some(*request_id),
-            _ => None,
-        };
         let result = core.apply_live_command(command, allow_destination_admission, now_ms);
         match result {
             Ok(mut effects) => {
@@ -791,33 +771,10 @@ impl LiveEffectsPublisher {
                 }
                 self.record(PublishedEffect::Ack);
                 let _ = reply.send(Ok(effects));
-                let cancelled_residency = if command_result == SchedulerCommandResult::Applied {
-                    cancellation_id.and_then(|request_id| {
-                        self.visible_residencies.lock().unwrap().remove(&request_id)
-                    })
-                } else {
-                    None
-                };
                 if allow_destination_admission {
                     self.publish_lifecycle(lifecycle_events).await;
-                    self.refresh_visible_residencies(core);
                     self.record(PublishedEffect::Metrics);
                     let _ = self.metrics_tx.send(core.live_metrics());
-                } else if let Some(residency) = cancelled_residency {
-                    // The core has already executed the pending pass. Update
-                    // only the queue where this request lived at the last
-                    // published boundary; aggregate post-pass metrics cannot
-                    // identify which visible request was cancelled.
-                    let mut published = self.metrics_tx.borrow().clone();
-                    let visible_count = match residency {
-                        RequestResidency::Running => &mut published.running_requests,
-                        RequestResidency::Waiting => &mut published.waiting_requests,
-                    };
-                    if *visible_count > 0 {
-                        *visible_count -= 1;
-                        self.record(PublishedEffect::Metrics);
-                        self.metrics_tx.send_replace(published);
-                    }
                 }
                 Some(command_result)
             }
@@ -827,14 +784,6 @@ impl LiveEffectsPublisher {
                 None
             }
         }
-    }
-
-    fn refresh_visible_residencies<C: LiveBoundaryCore>(&self, core: &C) {
-        let mut visible_residencies = self.visible_residencies.lock().unwrap();
-        visible_residencies.clear();
-        core.visit_live_request_residencies(&mut |request_id, residency| {
-            visible_residencies.insert(request_id, residency);
-        });
     }
 
     pub(crate) async fn retry_destinations<C: LiveBoundaryCore>(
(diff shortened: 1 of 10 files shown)


lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -319,48 +329,56 @@ impl PendingLivePass {
...
         let internal_deadline = wait_for_internal_deadline(scheduler_start, internal_deadline_ms);
         tokio::pin!(internal_deadline);
         tokio::select! {
             biased;
             _ = cancel_token.cancelled() => return false,
-            command = command_rx.recv() => {
+            cancellation = cancellation_rx.recv() => {
+                let Some(cancellation) = cancellation else {
+                    return false;
+                };
```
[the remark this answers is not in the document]
connorcarpenter15: Confirmed that this does not require a code change. `LiveSchedulerState` already owns an `Arc<LiveCancelGuard>`; dropping the final scheduler handle cancels the actor token even if an independently cloned `request_sender()` remains. That pre-existing ownership model means a request-sender clone never kept the scheduler actor alive. The cancellation-token branch is also biased ahead of the closed cancellation-channel branch, so the new channel does not change the handle-drop lifetime semantics.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -684,19 +799,64 @@ impl LiveEffectsPublisher {
...
+        self.visible_residencies
+            .lock()
+            .unwrap()
+            .retain(|request_id, residency| {
+                let Some(current) = core.live_request_residency(*request_id) else {
+                    return false;
+                };
+                *residency = current;
+                true
+            });
```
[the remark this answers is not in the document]
connorcarpenter15: Good catch. Fixed in `1c00009c28`. The live boundary now rebuilds `visible_residencies` from a complete core snapshot at pass and command metrics publication boundaries for both vLLM and SGLang, so ordinary `request_tx` submissions are included. I also removed the test-only residency setter; the existing mid-pass cancellation regression now obtains residency through the snapshot and fails against the previous retain-only implementation. `cargo fmt`, Clippy, and the full `dynamo-mocker` suite pass (523 passed, 1 ignored).

PeaBrane (commented):
High-level scope note: could we make it explicit in the PR description and leave a NOTE/TODO in code that this only adds cancellation at the `lib/mocker` live-scheduler boundary and does not yet hook the existing `lib/llm::MockEngine` context stop/drop path into it? I’m okay deferring that integration to follow-up work; making the scope explicit would let that follow-up faithfully test end-to-end cancellation behavior, potentially including disaggregated cases, rather than implying live Mocker cancellation is already end to end here.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -657,6 +731,7 @@ impl LiveEffectsPublisher {
         // disappeared. Those cleanup events belong to this same boundary.
         self.publish_router_effects(self.captured_kv_events.drain(), None);
         self.publish_lifecycle(pending.pass.lifecycle_events).await;
+        self.refresh_visible_residencies(core);
```
PeaBrane: [P2] Refresh residencies on the offload metrics path

Could we apply the same residency refresh before `advance_offload` publishes `core.live_metrics()`? vLLM offload progress can complete a swap-in and move the request into waiting during a modeled pass. That increments the published `waiting_requests`, but the snapshot still lacks the UUID; a same-pass cancellation is `Applied` yet cannot decrement the visible metric, leaving it stale until pass end. Please cover swap-in promotion followed by mid-pass cancellation.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -684,19 +791,52 @@ impl LiveEffectsPublisher {
...
                 self.record(PublishedEffect::Ack);
                 let _ = reply.send(Err(error));
+                None
             }
         }
     }
 
+    fn refresh_visible_residencies<C: LiveBoundaryCore>(&self, core: &C) {
+        let mut visible_residencies = self.visible_residencies.lock().unwrap();
+        visible_residencies.clear();
```
PeaBrane: [P2] Avoid rebuilding the full residency map every live pass

This refresh clears and rehashes every running and waiting UUID after each live pass and boundary-applied command, even when cancellation is unused. Because the waiting backlog is not capped by `max_num_seqs`, accelerated Mocker runs can pay O(passes × resident requests). Could this use actor-owned residency deltas, or at least avoid rebuilding on zero-duration/no-cancellation-observable passes?

PeaBrane (commented):
There are two reasonable contracts here:

- **Option 1 — simpler and preferred for this PR:** Cancellation immediately removes scheduler state and suppresses pending output, while published running/waiting metrics are only guaranteed to refresh at the next pass boundary. Add a concise NOTE/TODO explaining that exact mid-pass metrics would require incremental per-request residency accounting. Under this contract, remove `visible_residencies`, its full per-pass rebuild, the residency visitor plumbing, and the manual mid-pass counter adjustment. The next two inline comments then become irrelevant.

- **Option 2 — exact mid-pass metrics, which I do not prefer for this PR and would defer:** Introduce carefully designed incremental per-request residency accounting. It must cover ordinary admissions, running/waiting transitions, and KVBM swap-in promotion without imposing a complete O(all resident requests) rebuild on every pass. Under this contract, the next two inline comments are important and should be addressed.

Either way, core cancellation, pending-output suppression, and KV/offload cleanup should be tested independently. Add focused coverage for `Noop + discard_pending_output=true`, real vLLM/SGLang scheduler cancellation, and offline `in_flight` accounting. Those provide more useful signal than broad offline replay parity, which cannot currently construct `CancelRequest`.

lib/mocker/src/scheduler/live_boundary.rs
connorcarpenter15: Addressed by taking Option 1 in `859d79e0cb`. The residency map and manual mid-pass metric correction were removed, so offload promotion no longer has a snapshot that must be refreshed. Cancellation still removes scheduler state and suppresses pending output immediately; published running/waiting metrics are explicitly guaranteed to refresh at the next pass boundary. Under that contract, the swap-in-promotion metrics test is no longer needed.

lib/mocker/src/scheduler/live_boundary.rs
connorcarpenter15: Fixed in `859d79e0cb`. I removed `visible_residencies`, the full pass/command rebuilds, the vLLM/SGLang residency visitor plumbing, and the manual mid-pass counter adjustment. The narrower contract is documented on `SchedulerHandle::cancellation_sender` and in the PR description: scheduler state and pending output update immediately, while modeled-pass running/waiting metrics refresh at the next pass boundary.

connorcarpenter15:
Addressed the high-level review in `859d79e0cb`:

- Made the `lib/mocker`-only scope explicit in the PR description and added a `TODO(DIS-2478)` at `lib/llm::MockEngine` for wiring context stop/stream drop into the cancellation lane and adding end-to-end/disaggregated coverage. The follow-up is tracked in [DIS-2478](https://linear.app/nvidia/issue/DIS-2478/wire-mockengine-lifecycle-cancellation-into-the-mocker-scheduler).
- Took Option 1: cancellation removes scheduler state and suppresses pending output immediately; modeled-pass running/waiting metrics refresh at the next pass boundary. Removed the per-pass residency map, visitor plumbing, and manual metric adjustment.
- Kept the real vLLM/SGLang waiting/running cancellation coverage and added focused checks for `Noop + discard_pending_output=true` and offline `in_flight` cleanup.

Validation is clean: formatting and Clippy pass; `cargo test -p dynamo-mocker` reports 525 passed, 1 ignored.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -319,48 +320,56 @@ impl PendingLivePass {
...
+            .output_signals
+            .retain(|signal| signal.uuid != request_id);
+        self.pass.completed_requests = self
+            .pass
+            .output_signals
+            .iter()
+            .filter(|signal| signal.completed)
+            .count();
+        let (output_tokens, decode_forwards) =
+            crate::scheduler::accept_length_sample(&self.pass.output_signals);
```
PeaBrane, line 335: Non-blocking quick win: this currently rescans `output_signals` and `accept_length_sample` adds another scan plus a `HashSet` allocation. Since every removed signal belongs to one UUID, the `retain` could count removed completed/visible signals and checked-subtract those counters, including one decode forward when appropriate. Fine as a follow-up.

lib/mocker/src/scheduler/mod.rs
```diff
@@ -565,6 +596,56 @@ mod tests {
...
+            core.execute_hidden_pass(0.0);
+            assert_eq!(request_metrics(&core).running_requests, 1);
+            assert_eq!(
+                core.apply_command(SchedulerCommand::CancelRequest {
+                    request_id: running_id,
+                })
+                .unwrap(),
+                SchedulerCommandResult::Applied
+            );
+            assert_eq!(core.num_requests(), 0);
```
PeaBrane, line 650: Non-blocking coverage quick win: asserting `active_decode_blocks > 0` before cancellation and `== 0` afterward for both engines would independently demonstrate ordinary KV cleanup; `num_requests() == 0` currently proves scheduler-state removal. A pending-swap-in/offload case can remain follow-up work.

PeaBrane (approved):
✅ The residency-map overhead and metric-contract concerns are addressed by the simpler pass-boundary semantics, and the live-mocker scope is now explicit. The focused Noop/discard and offline-accounting coverage look good. I left two non-blocking quick wins below; neither blocks this PR.

lib/mocker/src/scheduler/live_boundary.rs
connorcarpenter15: Agreed. I’m leaving this as follow-up work because suppression only runs when cancellation lands mid-pass, and the current recomputation keeps the output/accounting invariants straightforward. No change in this PR.

lib/mocker/src/scheduler/mod.rs
connorcarpenter15: Added in c0cf874565. The existing cross-engine cancellation test now verifies nonzero KV occupancy before cancellation and reduced occupancy afterward, with vLLM reaching zero. SGLang intentionally retains evictable radix-prefix cache blocks in `active_decode_blocks`, so exact zero is not its correct post-cancel contract; the assertion instead proves request-owned KV occupancy is released. The full `dynamo-mocker` suite passes (525 passed, 1 ignored).

PR #11982, opened 2026-07-21
refactor(trtllm): report max_num_seqs per attention DP rank

kthui:
## Overview:

Standardize Python TensorRT-LLM’s existing `max_num_seqs` registration field as a per-attention-DP-rank capacity.

TensorRT-LLM exposes `max_batch_size` as an engine-wide concurrent-request limit, while Dynamo separately reports `data_parallel_size`. The runtime can therefore derive the local engine capacity consistently as:

`max_num_seqs × data_parallel_size`

## Details:

- Read the finalized post-initialization capacity from `engine.llm.args.max_batch_size`.
- Obtain the represented attention-DP size from `engine.get_attention_dp_size()`.
- Publish:

  `max_num_seqs = floor(max_batch_size / data_parallel_size)`

- Preserve exact capacity when `max_batch_size` is divisible by DP size.
- Conservatively round down when it is not divisible, ensuring the shared runtime never overstates TensorRT-LLM capacity.
  - Example: engine-wide capacity `37` with DP size `4` publishes `max_num_seqs=9`, producing a derived capacity of `36`, not an unsafe `40`.
- Leave `max_num_seqs` unavailable when:
  - Capacity or DP metadata is invalid.
  - Engine capacity is smaller than the DP size.
  - An encode registration has no local scheduler capacity.
- Continue publishing the existing `data_parallel_size` field.
- Add backend-local unit tests for DP-one, exact division, conservative division, invalid metadata, capacity smaller than DP size, and encode-without-scheduler behavior.

1. `components/src/dynamo/trtllm/utils/trtllm_utils.py`
   - `per_dp_rank_max_num_seqs()` and its conservative normalization rules.

2. `components/src/dynamo/trtllm/workers/llm_worker.py`
   - Publication of the finalized per-rank `max_num_seqs` and attention-DP size.

3. `components/src/dynamo/trtllm/tests/test_runtime_metadata.py`
   - Unit coverage at the backend registration contract boundary.

## Related Issues

- Prerequisites of #11928.

**🚫 This PR is NOT linked to a GitHub issue**:

- Closes DIS-2467

## Summary by CodeRabbit

* **New Features**
  * Improved runtime capacity calculations for data-parallel inference.
  * Added support for encode-only and disaggregated configurations.

* **Bug Fixes**
  * Corrected maximum sequence capacity assignment across attention data-parallel ranks.
  * Added validation for invalid or incompatible batch-size and parallelism settings.
  * Added warnings when configured capacity cannot be evenly distributed.

### DIFF
--- a/components/src/dynamo/trtllm/tests/test_runtime_metadata.py
+++ b/components/src/dynamo/trtllm/tests/test_runtime_metadata.py
@@ -5,7 +5,11 @@ from types import SimpleNamespace
 
 import pytest
 
-from dynamo.trtllm.utils.trtllm_utils import get_spec_decode_runtime_data
+from dynamo.trtllm.constants import DisaggregationMode
+from dynamo.trtllm.utils.trtllm_utils import (
+    get_spec_decode_runtime_data,
+    per_dp_rank_max_num_seqs,
+)
 
 pytestmark = [
     pytest.mark.unit,
@@ -13,6 +17,53 @@ pytestmark = [
     pytest.mark.gpu_0,
     pytest.mark.pre_merge,
 ]
+
+
+@pytest.mark.parametrize(
+    ("max_batch_size", "data_parallel_size", "expected"),
+    [
+        (37, 1, 37),
+        (36, 4, 9),
+        (37, 4, 9),
+        (None, 1, None),
+        (0, 1, None),
+        (-1, 1, None),
+        (True, 1, None),
+        ("37", 1, None),
+        (37, None, None),
+        (37, 0, None),
+        (37, -1, None),
+        (37, True, None),
+        (37, "4", None),
+        (3, 4, None),
+    ],
+)
+def test_max_num_seqs_is_reported_per_dp_rank(
+    max_batch_size, data_parallel_size, expected
+):
+    engine = SimpleNamespace(
+        llm=SimpleNamespace(args=SimpleNamespace(max_batch_size=max_batch_size))
+    )
+
+    result = per_dp_rank_max_num_seqs(engine, data_parallel_size)
+
+    assert result == expected
+    if result is not None:
+        assert result * data_parallel_size <= max_batch_size
+
+
+def test_encode_registration_without_local_scheduler_reports_no_max_num_seqs():
+    engine = SimpleNamespace(
+        disaggregation_mode=DisaggregationMode.ENCODE,
+        encoder_available=False,
+    )
+
+    assert per_dp_rank_max_num_seqs(engine, 1) is None
+
+
+def test_per_dp_rank_max_num_seqs_fails_if_engine_contract_changes():
+    with pytest.raises(AttributeError):
+        per_dp_rank_max_num_seqs(SimpleNamespace(), 1)
 
 
 def test_spec_decode_runtime_data_uses_max_draft_len():
--- a/components/src/dynamo/trtllm/utils/trtllm_utils.py
+++ b/components/src/dynamo/trtllm/utils/trtllm_utils.py
@@ -7,6 +7,8 @@ import logging
 from collections.abc import Mapping
 from typing import Any
 
+from dynamo.trtllm.constants import DisaggregationMode
+
 
 def deep_update(target: dict[str, Any], source: Mapping[str, Any]) -> None:
     """Recursively update nested dictionaries.
@@ -77,3 +79,35 @@ def get_spec_decode_runtime_data(engine_args: Any) -> dict[str, Any] | None:
     if method:
         data["method"] = str(method)
     return data
+
+
+def per_dp_rank_max_num_seqs(engine: Any, data_parallel_size: int) -> int | None:
+    """Return TRT-LLM's post-initialization request capacity per attention-DP rank."""
+    if (
+        getattr(engine, "disaggregation_mode", None) == DisaggregationMode.ENCODE
+        and not engine.encoder_available
+    ):
+        return None
+
+    value = engine.llm.args.max_batch_size
+    if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
+        return None
+    if (
+        not isinstance(data_parallel_size, int)
+        or isinstance(data_parallel_size, bool)
+        or data_parallel_size <= 0
+    ):
+        return None
+
+    per_rank = value // data_parallel_size
+    if per_rank <= 0:
+        return None
+    if value % data_parallel_size:
+        logging.warning(
+            "TRT-LLM max_batch_size=%d is not divisible by attention DP size=%d; "
+            "advertising conservative per-rank max_num_seqs=%d",
+            value,
+            data_parallel_size,
+            per_rank,
+        )
+    return per_rank
--- a/components/src/dynamo/trtllm/workers/llm_worker.py
+++ b/components/src/dynamo/trtllm/workers/llm_worker.py
@@ -65,7 +65,11 @@ from dynamo.trtllm.request_handlers.handlers import (
     RequestHandlerConfig,
     RequestHandlerFactory,
 )
-from dynamo.trtllm.utils.trtllm_utils import deep_update, get_spec_decode_runtime_data
+from dynamo.trtllm.utils.trtllm_utils import (
+    deep_update,
+    get_spec_decode_runtime_data,
+    per_dp_rank_max_num_seqs,
+)
 
 # Default buffer size for kv cache events.
 DEFAULT_KV_EVENT_BUFFER_MAX_SIZE = 100_000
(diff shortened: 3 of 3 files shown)


components/src/dynamo/trtllm/utils/trtllm_utils.py
```diff
@@ -77,3 +79,35 @@ def get_spec_decode_runtime_data(engine_args: Any) -> dict[str, Any] | None:
...
+
+def per_dp_rank_max_num_seqs(engine: Any, data_parallel_size: int) -> int | None:
+    """Return TRT-LLM's post-initialization request capacity per attention-DP rank."""
+    if (
+        getattr(engine, "disaggregation_mode", None) == DisaggregationMode.ENCODE
+        and not engine.encoder_available
+    ):
+        return None
+
+    value = engine.llm.args.max_batch_size
```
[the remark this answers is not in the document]
kthui: Thanks—confirmed against our pinned TensorRT-LLM `1.3.0rc21`. `MultimodalEncoder` inherits `_TorchLLM`, whose `args` is `TorchLlmArgs`; `BaseLlmArgs` defines `max_batch_size`, and Dynamo passes `max_batch_size` explicitly when constructing the encoder. Supported ENCODE workers therefore expose `engine.llm.args.max_batch_size`. Unsupported encoder models leave `encoder_available` false, so this helper returns `None` before accessing `.llm`. The direct access is intentional fail-fast behavior if that pinned engine contract changes, so no code change is needed.

nnshah1 (commented):
@tanmayv25 - is this accurate to make this assumption? Does trtllm internally enforce each DP rank batch size as (max_batch_size / dp ranks)?

nnshah1:
@kthui is this currently optional / not always available?

> leave max_num_seqs unavailable when:

components/src/dynamo/trtllm/utils/trtllm_utils.py
```diff
@@ -77,3 +79,35 @@ def get_spec_decode_runtime_data(engine_args: Any) -> dict[str, Any] | None:
...
+    if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
+        return None
+    if (
+        not isinstance(data_parallel_size, int)
+        or isinstance(data_parallel_size, bool)
+        or data_parallel_size <= 0
+    ):
+        return None
+
+    per_rank = value // data_parallel_size
```
tanmayv25, line 102: **Major — this division may be backwards.** It's correct only if TRT-LLM's `max_batch_size` is an *engine-wide* limit summed across attention-DP ranks (the claim added at `llm_worker.py:550`). Two counter-signals:

- The sibling vLLM worker — the "parity" this PR cites — sets `max_num_seqs` **without** dividing by `data_parallel_size` (`vllm/main.py:697`), advertising its per-rank scheduler bound directly.
- TRT-LLM's PyTorch attention-DP path appears to treat `max_batch_size` as a **per-rank** scheduler limit.

If it's already per-rank, dividing by `get_attention_dp_size()` (= TP degree) understates capacity by a factor of TP (e.g. TP=4 → advertises 9 instead of 37) — the opposite of parity, and it makes the router under-schedule. Please confirm `max_batch_size` semantics against TRT-LLM source/docs (the PR's own comment is circular); if per-rank, drop the `// data_parallel_size` division and advertise `max_batch_size` directly, mirroring vLLM. The arithmetic itself (floor / round-down / warn) is fine — the concern is the premise.

components/src/dynamo/trtllm/utils/trtllm_utils.py
```diff
@@ -77,3 +79,35 @@ def get_spec_decode_runtime_data(engine_args: Any) -> dict[str, Any] | None:
...
+    ):
+        return None
+
+    value = engine.llm.args.max_batch_size
+    if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
+        return None
+    if (
+        not isinstance(data_parallel_size, int)
+        or isinstance(data_parallel_size, bool)
+        or data_parallel_size <= 0
```
tanmayv25, line 98: Minor — this `data_parallel_size` validation (None / 0 / negative / bool / str) is unreachable from the only caller. `get_attention_dp_size()` (`engine.py:171-178`) can only ever return `1` or a positive `tensor_parallel_size`. The `max_batch_size` check above is justified (user-overridable via `--extra-engine-args`), but this branch guards inputs the call path can't produce — consider trimming it (and the corresponding test rows).

components/src/dynamo/trtllm/workers/llm_worker.py
```diff
@@ -537,18 +541,27 @@ async def init_llm_worker(
...
+        # Set data_parallel_size for attention DP mode
+        # This enables the router's scheduler to correctly iterate over all dp_ranks
+        # Need to name ADP as `data_parallel_size` for parity with other frameworks
+        attention_dp_size = engine.get_attention_dp_size()
+        max_num_seqs = per_dp_rank_max_num_seqs(engine, attention_dp_size)
+        if max_num_seqs is not None:
+            runtime_config.max_num_seqs = max_num_seqs
+        runtime_config.data_parallel_size = attention_dp_size
+
         runtime_config.max_num_batched_tokens = engine_args["max_num_tokens"]
```
tanmayv25, line 565: Minor — the reworded comment above says values are read from the initialized engine *because they can be overridden by `--extra-engine-args`*, but only `max_num_seqs` was switched to `engine.llm.args`. This line still reads the pre-override `engine_args["max_num_tokens"]`, so it has the same override problem the PR just fixed for `max_num_seqs`. Either apply the same treatment here or scope the comment to `max_num_seqs` only.

components/src/dynamo/trtllm/tests/test_runtime_metadata.py
```diff
@@ -15,6 +19,53 @@
...
+):
+    engine = SimpleNamespace(
+        llm=SimpleNamespace(args=SimpleNamespace(max_batch_size=max_batch_size))
+    )
+
+    result = per_dp_rank_max_num_seqs(engine, data_parallel_size)
+
+    assert result == expected
+    if result is not None:
+        assert result * data_parallel_size <= max_batch_size
```
tanmayv25, line 52: Minor — this invariant only checks the floor arithmetic; it can't catch a wrong *direction* of the relationship (see the major comment on the division). And nothing here exercises the `llm_worker.py` wiring, so if the division is backwards CI still passes green. Consider a test that pins the intended engine-wide ↔ per-rank relationship.

tanmayv25 (commented):
## Review summary

Mechanically clean and well-guarded — the crash-safety (ENCODE guard), `None`-handling, and test hygiene are all sound. But the change rests on one assumption that in-repo evidence contradicts: that TRT-LLM's `max_batch_size` is engine-wide across attention-DP ranks. See the inline comments — one major (the per-rank division may be backwards; please confirm `max_batch_size` semantics) and three minor.

**Checked, not issues:** no crash risk (the `ENCODE and not encoder_available` guard short-circuits the only post-init path where `engine.llm` would raise); consumer-safe (`max_num_seqs` is `Option<u64>` defaulting to `None`, no consumer unwraps it); `logging` is already imported so the new warning is safe.

_Filed as a non-blocking comment review._

tanmayv25:
I dug into the TRT-LLM source to settle this — short answer: **no, TRT-LLM doesn't divide `max_batch_size` across DP ranks. `max_batch_size` is already the per-attention-DP-rank cap**, so aggregate concurrency is `max_batch_size × dp_size`, not `max_batch_size` total. That means `max_num_seqs = max_batch_size // dp_size` under-reports each rank's real capacity by a factor of `dp_size`.

Tracing the PyTorch executor path (pinned `v1.1.0rc5`; `main` is identical):

- Capacity scheduler budget is `max_batch_size * pp_size` — only a pipeline-parallel multiplier, no DP divisor: [`_util.py:630`](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/_torch/pyexecutor/_util.py#L630) → `BindCapacityScheduler(scheduler_capacity, …)` ([L649](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/_torch/pyexecutor/_util.py#L649)).
- Micro-batch scheduler gets the raw `max_batch_size`: [`_util.py:655`](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/_torch/pyexecutor/_util.py#L655).
- Each attention-DP rank runs its own executor and sizes resources as `tp_size=1` ([`_util.py:97`](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/_torch/pyexecutor/_util.py#L97)); requests are routed to distinct DP ranks. Nothing on the `LlmArgs → scheduler` path divides by DP — there's even a special case that *adds* 1 to capacity under attention DP ([L646-647](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/_torch/pyexecutor/_util.py#L646)), never subtracts.

Your other assumption is correct, though: the attention-DP degree does equal `tensor_parallel_size` — [`mapping.py:403-404`](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.1.0rc5/tensorrt_llm/mapping.py#L403), `dp_size = tp_size if enable_attention_dp else 1`.

So for real vLLM parity (vLLM advertises its per-rank scheduler bound directly, without dividing), I'd drop the `// data_parallel_size` and publish `max_num_seqs = max_batch_size` as-is, keeping `data_parallel_size = attention_dp_size` as its own field. Happy to be double-checked on the links above.

components/src/dynamo/trtllm/utils/trtllm_utils.py
kthui: Thanks — you were right to challenge the premise. We re-audited the TensorRT-LLM version pinned by Dynamo (`v1.3.0rc21`) and confirmed that `max_batch_size` is already a per-attention-DP-rank cap. The executor derives aggregate attention-DP admission by multiplying the per-rank cap by the rank count; it does not divide `max_batch_size` across those ranks.

Our earlier engine-wide interpretation was an unsupported inference from Dynamo's wrapper shape. The `37 / 4 -> 9` tests then encoded that assumption rather than validating the upstream contract, exactly as you noted.

Because the division in this PR is backwards and would under-report capacity by approximately the DP factor, I have closed the PR without merge. Existing Dynamo publication of `max_num_seqs = max_batch_size` plus the separate attention-DP size already satisfies the runtime PR's PP-one contract. The three minor findings are moot with this change withdrawn; any future effective-cap or pipeline-parallel accounting should be handled separately. Thanks for tracing this to the scheduler semantics.

PR #11991, opened 2026-07-22
fix(inference-gateway): forward full request body to EPP KV-router scorer

avinash-rafay:
## Overview

The EPP KV-router prefill/decode scorer builds its routing request via `BuildOpenAIRequest`, which reconstructs each message from the framework's **typed** `ChatCompletions` view as only `{role, content}`. The upstream GAIE `Message` type models just `Role` and `Content`, so `tool_calls`, `tool_call_id`, `name`, `tools`, and reasoning/thinking content are dropped at parse time and never reach the body handed to the Rust router.

The Rust router (`route_decode_request` → `preprocess_request`) parses that JSON into the strict `NvCreateChatCompletionRequest` and re-renders the model's chat template to tokenize for KV-aware routing. Feeding it the stripped body breaks two things:

1. **Multi-turn tool calls are unroutable (503 "no healthy upstream").** A `role:"tool"` message missing the required `tool_call_id` fails deserialization (`missing field tool_call_id`); `route_decode_request` returns an error, no worker is selected, and the request is dropped by the gateway. Single-turn requests work; any conversation with a tool result is dead.
2. **Non-streaming reasoning/tool parsing is lost.** With `tool_calls` / `tools` / reasoning content missing, the chat template renders an incomplete prompt, so tokenization diverges from what the worker processes and the reasoning/tool parsers do not engage.

## Summary

Forward the full original request body to the EPP scorer's Rust router instead of a lossy `{role, content}` reconstruction, so tool-calling and reasoning fields survive for the router's strict parse and chat-template rendering. The same fix resolves both issues, since both stem from the router being handed an incomplete request.

## Details

- `BuildOpenAIRequest` now forwards the raw request body from `req.Body.Payload` (a `PayloadMap` that retains every field), overriding only `model` with the resolved target.
- The previous typed reconstruction is preserved as `buildOpenAIRequestFromTyped`, used only when the raw payload is unavailable; `nvext` and `cache_salt` handling are unchanged there.
- The tool fields cannot be recovered from the typed `Message` (it has no `ToolCalls` / `ToolCallID` / `Name`), so the raw payload is the only source that still carries them — a typed-path tweak isn't possible.
- Adds `TestBuildOpenAIRequest_PreservesToolCallFields`.

- `deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go` — `BuildOpenAIRequest` and the new `buildOpenAIRequestFromTyped` fallback.
- `deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin_test.go` — new `TestBuildOpenAIRequest_PreservesToolCallFields`.

## Validation

- `go test ./deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/...` — new test plus existing `BuildOpenAIRequest` tests pass (payload-bearing cases go through the passthrough, preserving `nvext.agent_hints.priority` and `cache_salt`; completion cases with no payload use the typed fallback).
- Validated end-to-end on a KV-routed disaggregated deployment: multi-turn tool calls now route instead of returning 503, and non-streaming requests parse reasoning/tool output.

## Related Issues

### DIFF
--- a/deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
+++ b/deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
@@ -283,16 +283,10 @@ func SerializeEndpointsToJSON(endpoints []schedtypes.Endpoint) (string, error) {
 	return string(data), nil
 }
 
-// BuildOpenAIRequestJSON serializes the request as an OpenAI-compatible JSON
-// body for the Rust router's FFI.
-//
-// A successfully parsed OpenAI request always yields a populated Payload holding
-// the complete body, so it is forwarded verbatim (overriding only the resolved
-// target model). This preserves tool_calls, tool_call_id, tools, and reasoning
-// content that the router's strict parse and chat-template rendering require —
-// fields the typed ChatCompletions view does not model. When the payload is
-// unavailable the request cannot be KV-routed, so an error is returned and the
-// scorer falls back to non-KV routing rather than a lossy reconstruction.
+// BuildOpenAIRequestJSON forwards the full request body (req.Body.Payload) to
+// the Rust router's FFI, overriding only the model, so tool-calling and
+// reasoning fields survive the router's parse and chat-template render. Errors
+// when no payload is available so the scorer falls back to non-KV routing.
 func BuildOpenAIRequestJSON(req *schedtypes.InferenceRequest) (string, error) {
 	if req == nil || req.Body == nil {
 		return "", fmt.Errorf("missing request body")
@@ -305,7 +299,7 @@ func BuildOpenAIRequestJSON(req *schedtypes.InferenceRequest) (string, error) {
 
 	requestBody := make(map[string]any, len(pm))
 	maps.Copy(requestBody, pm)
-	// Route on the resolved target model, not whatever alias the caller sent.
+	// Route on the resolved target model.
 	if strings.TrimSpace(req.TargetModel) != "" {
 		requestBody["model"] = req.TargetModel
 	}
--- a/deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
+++ b/deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
@@ -283,127 +283,38 @@ func SerializeEndpointsToJSON(endpoints []schedtypes.Endpoint) (string, error) {
 	return string(data), nil
 }
 
-func BuildOpenAIRequest(req *schedtypes.InferenceRequest) (map[string]any, error) {
+// BuildOpenAIRequestJSON serializes the request as an OpenAI-compatible JSON
+// body for the Rust router's FFI.
+//
+// A successfully parsed OpenAI request always yields a populated Payload holding
+// the complete body, so it is forwarded verbatim (overriding only the resolved
+// target model). This preserves tool_calls, tool_call_id, tools, and reasoning
+// content that the router's strict parse and chat-template rendering require —
+// fields the typed ChatCompletions view does not model. When the payload is
+// unavailable the request cannot be KV-routed, so an error is returned and the
+// scorer falls back to non-KV routing rather than a lossy reconstruction.
+func BuildOpenAIRequestJSON(req *schedtypes.InferenceRequest) (string, error) {
 	if req == nil || req.Body == nil {
-		return nil, fmt.Errorf("missing request body")
+		return "", fmt.Errorf("missing request body")
 	}
 
-	// Prefer the full, original request body. The Rust router parses this JSON
-	// and re-renders the model's chat template to tokenize for KV-aware routing,
-	// so it needs the complete message structure — tool_calls, tool_call_id,
-	// tools, reasoning/thinking content, multimodal parts, etc. The typed
-	// ChatCompletions view only models role+content; reconstructing from it
-	// drops those fields, which breaks routing in two ways:
-	//  1. On tool-calling turns the router's strict parse fails (a role:"tool"
-	//     message is missing the required tool_call_id), route_decode_request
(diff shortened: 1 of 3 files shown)


deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
```diff
@@ -283,12 +284,47 @@ func SerializeEndpointsToJSON(endpoints []schedtypes.Endpoint) (string, error) {
...
+	//  1. On tool-calling turns the router's strict parse fails (a role:"tool"
+	//     message is missing the required tool_call_id), route_decode_request
+	//     returns an error, no worker is selected, and the request is dropped
+	//     as 503 "no healthy upstream" — multi-turn tool calling is unroutable.
+	//  2. Even when parsing succeeds, the chat template renders an incomplete
+	//     prompt, so reasoning/tool parsing is lost (e.g. on non-streaming
+	//     requests) and tokenization no longer matches what the worker sees.
+	if pm, ok := req.Body.Payload.(fwkrh.PayloadMap); ok && len(pm) > 0 {
+		requestBody := make(map[string]any, len(pm))
+		maps.Copy(requestBody, pm)
```
tmonty12: This helper’s output is only consumed as FFI JSON. Could it return a serialized JSON string—perhaps as BuildOpenAIRequestJSON—instead of exposing map[string]any? Keep the shallow clone internally to override the top-level model, then marshal before returning.

deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
```diff
@@ -283,12 +284,47 @@ func SerializeEndpointsToJSON(endpoints []schedtypes.Endpoint) (string, error) {
...
+		if strings.TrimSpace(req.TargetModel) != "" {
+			requestBody["model"] = req.TargetModel
+		}
+		return requestBody, nil
+	}
+
+	// Fallback: the raw payload is unavailable (not unmarshaled). Reconstruct a
+	// minimal request from the framework's typed view. Tool-calling turns cannot
+	// be represented here, but plain chat/completion routing still works.
+	return buildOpenAIRequestFromTyped(req)
```
tmonty12: Successful OpenAI HTTP parsing guarantees a populated PayloadMap containing the complete request body needed for Rust preprocessing. Falling back to the typed role/content view reintroduces the field loss this PR fixes. Please remove the fallback and return ExactRoutingUnavailable when that invariant is not met, allowing the scorer to use non-KV routing.

deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
```diff
@@ -283,12 +284,47 @@ func SerializeEndpointsToJSON(endpoints []schedtypes.Endpoint) (string, error) {
 }
 
 func BuildOpenAIRequest(req *schedtypes.InferenceRequest) (map[string]any, error) {
-	requestBody := make(map[string]any)
-
 	if req == nil || req.Body == nil {
 		return nil, fmt.Errorf("missing request body")
 	}
 
+	// Prefer the full, original request body. The Rust router parses this JSON
```
tmonty12: nit: remove long comment

tmonty12 (commented):
Several very long comments that I think should be removed within `BuildOpenAIRequest`, method docstring for `buildOpenAIRequestFromTyped` and method docstring above `TestBuildOpenAIRequest_PreservesToolCallFields`

deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
avinash-rafay: Done in e0363b27de — `BuildOpenAIRequestJSON` now returns the serialized JSON string (shallow-clones the payload internally to override `model`, then marshals). `buildRequestJSON` in shared.go is now a thin wrapper, and its `json`/`fmt` imports were dropped.

deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
avinash-rafay: Agreed — removed the typed fallback entirely in e0363b27de. When the Payload isn't a populated `PayloadMap`, `BuildOpenAIRequestJSON` returns an error; the decode/prefill scorers already treat a build error by returning `uniformScores` (non-KV routing), which matches the intended behavior. This also let me delete the now-unused `buildOpenAIRequestFromTyped`, `addCompletionPrompt`, `extractNvext`, and `extractTopLevelCacheSalt`. Completions now flow through the passthrough too (the Rust preprocessor handles them via the `prompt` field), covered by a new test; added a missing-payload error test as well.

deploy/inference-gateway/epp/pkg/plugins/dynamo_kv_scorer/plugin.go
avinash-rafay: Done — trimmed to a concise docstring in e0363b27de.

tmonty12 (approved):

athreesh (approved):
Reviewed the final diff and the follow-up changes from the prior review. The scorer now forwards the original PayloadMap to the Rust router as FFI JSON, overrides only the resolved model, and returns an error when the raw payload invariant is missing so the disagg scorers fall back to non-KV routing instead of rebuilding a lossy typed request. The call sites in prefill/decode already handle that error path with uniform scoring, and the added tests cover priority/cache_salt forwarding, tool-call field preservation, completions payload forwarding, and missing-payload behavior.

I also checked the head CI rollup; the relevant checks are passing. I attempted the targeted Go test locally, but the sparse checkout cannot link the CGO package because lib/libdynamo_llm_capi.a is not present in the local checkout. No blocking findings from my review.

PR #11992, opened 2026-07-22
fix(router): serialize active sequence publishing

jthomson04:
## Overview:

Serialize active-sequence lifecycle publication without the duplicate tracker-owned dispatcher. Lifecycle mutations now synchronously enqueue into the transport publisher, so accepted events retain FIFO order without spawning one task per event.

## Details:

- replace the async `SequencePublisher` publication API with an owned synchronous enqueue boundary
- use one bounded 100,000-event queue per runtime publisher
- feed ZMQ directly into the existing batch publisher task and use one equivalent singleton publisher task for NATS
- drop the newest event on queue overflow while leaving the local lifecycle mutation successful
- aggregate repeated full-queue and unexpectedly-closed diagnostics at 30-second intervals
- classify queue closure with the publisher cancellation token and log expected shutdown closure once at debug level
- preserve request, worker, capacity, and standalone model/routing-group context in queue diagnostics
- preserve the existing NATS singleton and ZMQ batch wire formats, batch limits, and concurrent replica ingestion
- keep standalone replica sync on its existing process-level outbound queue

- `lib/kv-router/src/sequences/multi_worker.rs` for the synchronous tracker-to-publisher boundary and failure-log aggregation
- `lib/llm/src/kv_router/sequence.rs` for the shared sender, cancellation classification, and transport-specific publisher tasks

## Related Issues

## Summary

- eliminate the tracker queue-to-task-to-runtime-queue chain
- guarantee FIFO ordering among successfully enqueued lifecycle events
- retain bounded best-effort behavior without blocking local state updates
- avoid overflow log floods and shutdown-time error noise

## Validation

- `cargo test -p dynamo-kv-router sequences::multi_worker::tests --lib`
- `cargo test -p dynamo-kv-router --features standalone-slot-tracker services::common::replica_sync::tests --lib`
- `cargo test -p dynamo-llm kv_router::sequence::tests --lib`
- `cargo test -p dynamo-llm mixed_replica_batch_updates_lora_load_in_order --lib`
- `cargo check -p dynamo-mocker --lib`
- `cargo clippy -p dynamo-kv-router -p dynamo-llm --lib --tests -- -D warnings`

### DIFF
--- a/lib/kv-router/src/sequences/multi_worker.rs
+++ b/lib/kv-router/src/sequences/multi_worker.rs
@@ -10,7 +10,7 @@
 //! this crate while the runtime glue stays in `lib/llm`.
 
 use dynamo_tokens::SequenceHash;
-use parking_lot::RwLock;
+use parking_lot::{Mutex, RwLock};
 use rustc_hash::FxHashMap;
 use std::collections::HashMap;
 use std::env;
@@ -40,6 +40,9 @@ use crate::protocols::{
 // in ActiveSequencesMultiWorker::force_expire_requests_across_all_workers for
 // more details.
 const FORCE_EXPIRE_REQUESTS_ACROSS_ALL_WORKERS_INTERVAL: Duration = Duration::from_secs(60);
+
+/// Minimum interval between repeated logs for a saturated or unexpectedly closed publish queue.
+const SEQUENCE_PUBLISH_FAILURE_LOG_INTERVAL: Duration = Duration::from_secs(30);
 
 /// Environment override for the stale active-request cleanup guard.
 const DYN_ROUTER_ACTIVE_REQUEST_EXPIRY_SECS: &str = "DYN_ROUTER_ACTIVE_REQUEST_EXPIRY_SECS";
@@ -117,6 +120,104 @@ pub trait SequencePublisher: Send + Sync {
 
     /// Observe that a worker/dp_rank was removed from the router.
     fn observe_worker_removed(&self, _worker: &WorkerWithDpRank, _worker_type: &str) {}
+}
+
+/// Admission failures reported by bounded active-sequence publisher queues.
+#[derive(Debug, thiserror::Error)]
+pub enum SequencePublishQueueError {
+    #[error(
+        "active-sequence publish queue full; dropping newest event \
+         (request_id={request_id}, worker={worker:?}, capacity={capacity})"
+    )]
+    Full {
+        request_id: String,
+        worker: WorkerWithDpRank,
+        capacity: usize,
+    },
+    #[error(
+        "active-sequence publish queue closed; dropping event \
+         (request_id={request_id}, worker={worker:?}, capacity={capacity})"
+    )]
+    Closed {
+        request_id: String,
+        worker: WorkerWithDpRank,
+        capacity: usize,
+        during_shutdown: bool,
+    },
+}
+
+impl SequencePublishQueueError {
+    pub fn full(event: ActiveSequenceEvent, capacity: usize) -> Self {
+        Self::Full {
+            request_id: event.request_id,
+            worker: event.worker,
+            capacity,
+        }
+    }
+
+    pub fn closed(event: ActiveSequenceEvent, capacity: usize, during_shutdown: bool) -> Self {
+        Self::Closed {
+            request_id: event.request_id,
+            worker: event.worker,
+            capacity,
+            during_shutdown,
+        }
+    }
+}
+
+#[derive(Default)]
+struct RateLimitedPublishFailure {
+    last_logged_at: Option<Instant>,
+    suppressed_since_last_log: u64,
+}
+
+impl RateLimitedPublishFailure {
+    fn record(&mut self, now: Instant) -> Option<u64> {
+        let should_log = self.last_logged_at.is_none_or(|last_logged_at| {
+            now.saturating_duration_since(last_logged_at) >= SEQUENCE_PUBLISH_FAILURE_LOG_INTERVAL
+        });
+        if should_log {
+            let failures_since_last_log = self.suppressed_since_last_log.saturating_add(1);
+            self.last_logged_at = Some(now);
+            self.suppressed_since_last_log = 0;
+            Some(failures_since_last_log)
+        } else {
+            self.suppressed_since_last_log = self.suppressed_since_last_log.saturating_add(1);
+            None
+        }
+    }
+}
+
+#[derive(Default)]
+struct SequencePublishFailureLogState {
+    full: RateLimitedPublishFailure,
+    unexpected_closed: RateLimitedPublishFailure,
+    shutdown_closed_logged: bool,
+}
+
+#[derive(Default)]
+struct SequencePublishFailureLogLimiter {
+    state: Mutex<SequencePublishFailureLogState>,
+}
+
+impl SequencePublishFailureLogLimiter {
+    fn record_full(&self, now: Instant) -> Option<u64> {
+        self.state.lock().full.record(now)
+    }
+
+    fn record_unexpected_closed(&self, now: Instant) -> Option<u64> {
+        self.state.lock().unexpected_closed.record(now)
+    }
+
+    fn record_shutdown_closed(&self) -> bool {
+        let mut state = self.state.lock();
+        if state.shutdown_closed_logged {
+            false
+        } else {
+            state.shutdown_closed_logged = true;
+            true
+        }
+    }
 }
 
 /// No-op publisher for callers that do not need active-sequence event transport.
@@ -213,6 +314,7 @@ pub struct ActiveSequencesMultiWorker<P: SequencePublisher> {
     block_size: usize,
     pub(super) router_id: u64,
     pub(super) publisher: Arc<P>,
+    publish_failure_logs: SequencePublishFailureLogLimiter,
     remote_state_updates: watch::Sender<()>,
     #[cfg(test)]
     remote_state_update_count: AtomicUsize,
@@ -355,6 +457,7 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
             block_size,
             router_id,
             publisher,
+            publish_failure_logs: SequencePublishFailureLogLimiter::default(),
             remote_state_updates,
             #[cfg(test)]
             remote_state_update_count: AtomicUsize::new(0),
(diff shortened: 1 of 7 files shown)


jthomson04:
## Code Review

**Verdict: solid fix.** The change is correct, well-tested, and improves resource bounds over the old per-event spawn. Remaining points are minor observations and polish suggestions, not blockers.

### Overview

Replaces the per-event `tokio::spawn` in `spawn_publish_event` with a single lazily-initialized dispatcher task per tracker (`ActiveSequencePublishQueue`), fed by a bounded `mpsc` channel (capacity 100,000). Events are now published strictly in enqueue order; on overflow the newest event is dropped with an error log. Since NATS preserves per-connection publish order, this gives peers in-order lifecycle events (`AddRequest` → `MarkPrefillCompleted` → `Free`), fixing the race where a `Free` could be published before its `AddRequest` and leave phantom load on peer routers.

### Correctness

- **Lazy init via `OnceLock` is the right call.** `get_or_init` only runs when `replica_sync` is enabled and the caller is inside a runtime — the same `tokio::spawn` requirement the old code had, so no regression. The plain (non-tokio) `#[test]` for the disabled path doubles as a regression guard: if anyone moves the spawn ahead of the `replica_sync` check, that test panics.
- **Shutdown/lifecycle is clean.** The dispatcher exits when the channel closes (when the tracker — the only sender holder — is dropped), draining remaining events first. No cancellation token needed, no publisher `Arc` leak.
- **Overflow test is deterministic.** Waiting for the `entered_rx.recv()` of `AddRequest` guarantees the consumer has pulled it off the queue before the capacity-1 channel is refilled, so the drop of `Free` isn't racy. `Notify` permit semantics also make `notify_one` safe even if it fires before the consumer reaches `notified().await`.
- **Residual ordering race (pre-existing):** enqueue happens *after* the per-worker lock is released in `mutate_request_worker_*_state`, so two concurrent lifecycle calls for the same request could theoretically enqueue in the opposite order from their local mutations. In practice a request's lifecycle is driven by one task, and this PR strictly narrows the old race — just noting the guarantee is "publish order = enqueue order," not "publish order = mutation order."
- **Drop-newest tradeoff:** dropping a `Free` whose `AddRequest` was delivered leaves phantom load on peers until the stale-request expiry guard clears it (self-healing, ~60s). Drop-oldest would instead under-count peer load. Given overflow at 100k implies the publisher is stalled anyway, drop-newest + expiry safety net is reasonable — a one-line comment on the constant noting the reliance on the expiry guard would help future readers.

### Design & performance

- **Net improvement on both axes.** The old code could spawn an unbounded number of tasks under a stalled NATS connection (unbounded memory); the new code is bounded and replaces a per-event task allocation with a cheap non-blocking `try_send` on the caller's hot path.
- **Throughput ceiling:** publishes are now serialized, so sustained throughput is bounded by one-at-a-time `publish_event_owned` latency. `RuntimeSequencePublisher::publish_event` is a NATS client-buffer write, so this should be microseconds and the 100k buffer absorbs bursts — worth keeping in mind if the transport ever gains per-publish flushing.
- **Log flood on sustained overflow:** the `Full` branch emits one `tracing::error!` per dropped event. If the queue is pinned at 100k, that's thousands of error lines per second. Consider rate-limiting the log or a periodic "dropped N events" summary, and/or a drop counter metric — a drop here directly means peer-state divergence.
- Minor: the stored `capacity` field duplicates `event_tx.max_capacity()`; the field could be dropped. Cosmetic only.

### Tests

Good coverage of the three behaviors that matter: ordering under a blocked publisher, deterministic drop-newest on overflow, and no dispatcher spawn when replica sync is disabled. The `BlockingEventPublisher` harness is deterministic (no sleeps). The untested `TrySendError::Closed` branch is only reachable if the dispatcher task dies — fine to leave uncovered.

### Suggested follow-ups (non-blocking)

1. Rate-limit or aggregate the queue-full error log; consider a dropped-event counter metric.
2. Add a brief comment on `ACTIVE_SEQUENCE_PUBLISH_QUEUE_CAPACITY` noting that dropped events are reconciled by the stale-request expiry guard, and one on the `OnceLock` field explaining why init is lazy (spawn needs a runtime; only needed when replica sync is on).

jthomson04:
## Code Review (re-review of the reworked design)

**Verdict: the redesign is a clear improvement over the previous revision — LGTM.** Instead of bolting a lazily-spawned dispatcher onto the tracker (`OnceLock` + `get_or_init` + spawn), the trait boundary itself is now synchronous: `publish_event`/`publish_event_owned` (async, borrowed + owned variants) collapse into one sync `enqueue_event(ActiveSequenceEvent) -> anyhow::Result<()>`. That eliminates the runtime-context subtlety, the duplicate queue-behind-a-queue on the ZMQ path, and the `event.clone()` in `ScopedSequencePublisher`. Remaining points are observations and small suggestions, not blockers.

### Overview

- Trackers now enqueue lifecycle events synchronously into the publisher; the transport side owns one bounded 100k-event channel per `RuntimeSequencePublisher`, created in `create_multi_worker_sequences`.
- NATS gets a new `run_replica_singleton_publisher` task (sequential, preserves the singleton wire format); ZMQ feeds the existing `run_replica_batch_publisher` directly, which preserves order (sequential batch publishes, in-order drain — verified against the existing implementation).
- Overflow drops the newest event and surfaces an `Err` that the tracker logs; the local mutation always succeeds.

### Correctness

- **FIFO guarantee holds end-to-end** on both transports for successfully enqueued events. One residual note: enqueue happens after the per-worker lock is released in `mutate_request_worker_*_state`, so the guarantee is "publish order = enqueue order," not "= mutation order." In practice a request's lifecycle is single-tasked, and this is strictly better than the old racing spawns.
- **ZMQ path semantic change worth being deliberate about:** previously `event_tx.send(event).await` inside a spawned task meant no event was ever dropped (at unbounded task cost); now a full queue drops the newest event. The PR body and the updated comment on `REPLICA_EVENT_CHANNEL_CAPACITY` document this, and drop-newest is reconciled by the stale-request expiry guard — fine, just confirming it's intentional.
- **`ScopedSequencePublisher` severity flip:** a full standalone replica queue used to `trace!` and return `Ok`; it now returns `Err`, which the tracker logs at **error** level per event. Consistent with the runtime path, but it converts a previously silent expected-overload signal into a per-event error flood.
- **Shutdown noise:** when the cancellation token fires, the publisher task exits and drops `event_rx`; in-flight requests still calling `free`/`mark_prefill_completed` will hit `TrySendError::Closed` and emit error-level logs. Consider treating `Closed` as expected-during-shutdown (warn/debug), since it's not actionable.

### Design & code quality

- The sync `enqueue_event` boundary is the right shape — lifecycle callers never await transport I/O, and the trait no longer needs the owned/borrowed async pair. Test doubles across `queue.rs`, `push_router.rs`, and the mocker all got simpler as a result.
- `run_replica_singleton_publisher` mirrors the batch publisher's structure (select on cancellation, exit on channel close). One small inconsistency: on cancellation the batch publisher flushes its in-progress batch (`stop_after_flush`) while the singleton drops everything queued immediately. Harmless at shutdown, just asymmetric.
- Error messages embed `request_id`/`worker`/`capacity` via string interpolation while the tracker's `tracing::error!` carries the same info as structured fields — acceptable, but the info appears twice in the log line.

### Performance

- Hot path improves vs. both main and the previous revision: no per-event task spawn, no clone, one `try_send`. Bounded memory under a stalled transport (the old ZMQ path was bounded, but the NATS path could accumulate unbounded spawned tasks).
- NATS publishes are now serialized through one task; since `EventPublisher::publish` is a client-buffer write this is a non-issue, with the 100k buffer absorbing bursts.

### Tests

- Coverage is appropriately layered: tracker-level ordering and disabled-path tests in `multi_worker.rs` (now trivially synchronous via the recording publisher), sender-level FIFO/overflow tests in `sequence.rs`, and the `ScopedSequencePublisher` full-queue error test.
- Gap: `run_replica_singleton_publisher` itself has no test (order of actual publishes, exit on cancellation), while the batch publisher does. A small paused-clock test mirroring the existing batch tests would close it.
- Validation list is thorough (fmt, targeted tests including the standalone feature, clippy `-D warnings`, mocker check).

### Suggested follow-ups (non-blocking)

1. Rate-limit or aggregate the queue-full error logging (carried over from the previous review — a pinned queue now produces per-event error logs on *both* the runtime and standalone paths), and consider a dropped-event counter metric.
2. Downgrade the `Closed` case to warn/debug or gate it on cancellation, so shutdown doesn't emit error-level noise.
3. Add a small test for `run_replica_singleton_publisher` (publish order + cancellation exit).

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -634,7 +774,7 @@ impl<P: SequencePublisher + 'static> ActiveSequencesMultiWorker<P> {
         });
         self.add_request_local(req, decay_now, lazily_register_worker)?;
         if let Some(event) = event {
-            self.spawn_publish_event(event);
+            self.enqueue_publish_event(event);
```
PeaBrane, line 783: **Advisory / non-blocking:** FIFO begins at the enqueue boundary. Because the local mutation completes before this call, overlapping lifecycle operations for the same request can still enqueue in a different order. Given replica sync is best-effort and peers self-heal through expiry, accepting that seems reasonable; a short comment documenting this boundary would keep the guarantee from being read as mutation-order FIFO.

lib/llm/src/kv_router/sequence.rs
```diff
@@ -150,6 +179,43 @@ impl SequencePublisher for RuntimeSequencePublisher {
...
+) {
+    loop {
+        let event = tokio::select! {
+            _ = cancellation_token.cancelled() => break,
+            event = event_rx.recv() => match event {
+                Some(event) => event,
+                None => break,
+            },
+        };
+        if let Err(error) = publisher.publish_event(&event).await {
```
PeaBrane: **Small patch:** Could we make the in-flight publish cancellation-aware, not just the receive? Prolonged NATS reconnect/backpressure can leave `publish_event` pending and prevent this detached consumer from observing cancellation. Wrapping this await in `tokio::select!` with `cancellation_token.cancelled()` would make shutdown prompt; the existing blocking-publisher test can cancel before releasing the publish and assert the task exits. Dropping the in-flight event is consistent with the best-effort contract.

lib/kv-router/src/sequences/multi_worker.rs
```diff
@@ -87,21 +90,11 @@ fn active_request_expiry_duration_from_lookup(
...
-
-    /// Publish an owned replica-sync event, avoiding a clone when supported by the transport.
-    fn publish_event_owned(
-        &self,
-        event: ActiveSequenceEvent,
-    ) -> impl Future<Output = anyhow::Result<()>> + Send {
-        async move { self.publish_event(&event).await }
-    }
+    /// Enqueue a replica-sync event for publication to peer routers.
+    fn enqueue_event(&self, event: ActiveSequenceEvent) -> anyhow::Result<()>;
```
PeaBrane, line 100: **Advisory / non-blocking:** Failure classification is currently an implicit downcast contract: queue publishers must return `SequencePublishQueueError` for the tracker's rate limiting and shutdown-aware logging. Current implementations comply, so this is fine today; consider a typed admission result if the trait gains more publishers or wrappers.

PeaBrane (commented):
One small cancellation-hardening request plus two non-blocking advisories.

PeaBrane (approved):
✅

lib/kv-router/src/sequences/multi_worker.rs
jthomson04: Addressed in 1cec3a31b6. The enqueue helper now documents that FIFO begins after local mutation, so overlapping lifecycle calls can enqueue out of mutation order; replica sync remains best-effort and peer expiry reconciles stale state.

lib/llm/src/kv_router/sequence.rs
jthomson04: Addressed in 1cec3a31b6. The singleton publisher now selects cancellation against every in-flight publish, so transport backpressure cannot delay shutdown. The existing test now starts a blocked publish, cancels without releasing it, and verifies prompt task exit. Focused and module tests plus clippy passed.

lib/kv-router/src/sequences/multi_worker.rs
jthomson04: Addressed in 1cec3a31b6 by making the current contract explicit in SequencePublisher::enqueue_event: queue-backed implementations preserve SequencePublishQueueError as the source so tracker downcasting can classify saturation and closure. I kept the anyhow signature to avoid broad trait churn; a typed admission result remains appropriate if the publisher set grows.

PR #11993, opened 2026-07-22
feat(runtime): configure worker inhibition with direct-routing bypass

### LINKED ISSUE #11583: [CONTRIBUTION]: Disable push router quarantine / fault rejection
### Type of Change

- Bug fix

---

### Summary

Dynamo currently appears to automatically mark remote worker instances as down in the request path when certain transport/backend errors occur. This behavior is useful for production safety in many cases, but it can also make debugging, recovery flows, and controlled failure testing difficult because a transient request failure can remove a worker from the routable pool.

I'd like Dynamo to expose an environment variable that disables this automatic worker quarantine / fault-detection behavior, while keeping the current behavior **enabled by default**.

**Proposed environment variable:**

```text
DYN_DISABLE_FAULT_INJECTION=true
```

Alternative (if a more implementation-specific name is preferred):

```text
DYN_DISABLE_REMOTE_QUARANTINE=true
```

---

## Current behavior

In `PushRouter::from_client_with_monitor`, fault detection is enabled by default.

When `fault_detection_enabled` is `true`, Dynamo reports an instance as down whenever an inhibited transport/backend error is observed.

Examples include:

- Stream item errors matching `is_inhibited(...)`
- Request failures matching `is_inhibited(...)`
- Backend response inactivity timeout paths

These paths ultimately invoke:

```rust
client.report_instance_down(instance_id)
```

This effectively quarantines the worker from future routing decisions.

There is already a separate constructor:

```rust
from_client_no_fault_detection(...)
```

which disables this behavior for certain internal recovery/query paths. However, there is currently no runtime or environment-level mechanism to disable fault detection for the normal router construction path.

---

## Why this is a problem

In some deployments, a transient or expected backend/transport failure should fail only the individual request—not remove the worker from the routable pool.

### Transient backend failures

A worker may temporarily fail a single request.

Automatically reporting the instance as down can unnecessarily reduce available routing capacity.

### Debugging and fault reproduction

During debugging it is often desirable to repeatedly hit the same worker.

Automatic quarantine hides the original failure mode by removing the worker after the first failure.

### Controlled failure testing

Some test environments intentionally inject failures.

In these cases the desired behavior is:

- Request fails
- Worker remains routable

rather than:

- Request fails
- Worker is quarantined

### External orchestration

Many deployments already rely on external health checking, Kubernetes lifecycle management, or orchestration-level replacement.

In those environments, Dynamo's request-path quarantine can be overly aggressive or duplicate existing health management.

---

## Desired behavior

The default behavior should remain unchanged.

By default:

- Fault detection remains enabled.
- Workers continue to be reported down on inhibited errors.

When the following environment variable is set:

```text
DYN_DISABLE_FAULT_INJECTION=true
```

Dynamo should:

- Still return request errors to the caller.
- Still log and trace failures.
- **Not** call `report_instance_down(instance_id)` from the request path.
- Route using the non-quarantined discovery instance list where appropriate, consistent with the existing `from_client_no_fault_detection` behavior.

When the environment variable is **unset**, behavior should remain exactly as it is today.

---

## Suggested implementation

The existing implementation already contains:

- `fault_detection_enabled`
- `from_client_no_fault_detection(...)`

The environment variable could simply control the default value used by `from_client_with_monitor`.

Conceptually:

```rust
fault_detection_enabled = !env::var_truthy("DYN_DISABLE_FAULT_INJECTION");
```

or equivalently:

```text
default:
    fault_detection_enabled = true

if DYN_DISABLE_FAULT_INJECTION is truthy:
    fault_detection_enabled = false
```

This preserves full backward compatibility while allowing deployments to opt out when appropriate.

---

## Acceptance criteria

- [ ] Add an environment variable to disable automatic request-path worker quarantine/fault detection.
- [ ] Default behavior remains unchanged when the environment variable is unset.
- [ ] Prevent request-path calls to `report_instance_down(instance_id)` for inhibited transport/backend errors.
- [ ] Prevent quarantine caused by backend response inactivity timeout.
- [ ] Update documentation or the environment variable registry if applicable.
- [ ] Add tests covering both the default-enabled and env-disabled behavior where practical.

---

## Proposed environment variable

**Preferred**

```text
DYN_DISABLE_FAULT_INJECTION
```

**Alternative**

```text
DYN_DISABLE_REMOTE_QUARANTINE
```

The important aspect is that this is a **disable-style flag**, since the existing behavior is already enabled by default.

---

### Proposed Solution

Implement an environment-variable-controlled switch that disables request-path worker quarantine by defaulting `fault_detection_enabled` to `false` when `DYN_DISABLE_FAULT_INJECTION` is truthy, while otherwise preserving the existing behavior.

---

### Estimated PR Size

- S (11–50 lines)

---

### Files / Components Affected

- `PushRouter`
- Fault detection / request-path routing
- Environment variable configuration
- Documentation (if applicable)
- Tests (if applicable)

kthui:
## Overview:

Adds `DYN_RUNTIME_INHIBITED_DURATION_SECS` to make local worker inhibition configurable and updates Direct dispatch to honor an upstream-selected worker while it remains in service discovery.

The existing five-second duration remains the default. A positive value customizes the maximum local inhibition window, while `0` disables local worker inhibition.

## Details:

- Read `DYN_RUNTIME_INHIBITED_DURATION_SECS` once during runtime client initialization.
- Preserve the existing five-second default when the variable is unset.
- Fall back to the default and log a warning when the value is invalid.
- Treat `0` as disabling local inhibition while continuing to process discovery updates.
- Keep random and round-robin selection filtered by local inhibition.
- Make Direct dispatch bypass local inhibition and validate its selected worker against service discovery instead.
- Keep overload tracking independent and enforced.
- Add unit tests for duration configuration, disabled inhibition, and Direct-dispatch behavior.
- Document the configuration and routing behavior in the fault-tolerance guide.

- `lib/runtime/src/component/client.rs` — duration resolution and zero-duration behavior.
- `lib/runtime/src/pipeline/network/egress/push_router.rs` — Direct-dispatch discovery validation and unit coverage.
- `lib/runtime/src/config/environment_names.rs` — environment-variable definition.
- `docs/fault-tolerance/README.md` — user-facing configuration and routing behavior.

## Related Issues

- Closes #11583
- Closes DIS-2476


lib/runtime/src/component/client.rs
```diff
@@ -575,6 +601,14 @@ impl Client {
 
     /// Mark an instance as down/unavailable
     pub fn report_instance_down(&self, instance_id: u64) {
+        if self.reconcile_interval.is_zero() {
+            tracing::debug!(
+                instance_id,
+                "local worker inhibition is disabled; leaving instance routable"
+            );
+            return;
+        }
```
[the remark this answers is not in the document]
kthui: Confirmed—this is intentional for the explicit `0` setting. It allows a worker that transiently failed a single request to remain eligible, so a retry or migration may immediately reselect that worker; discovery remains authoritative if it later removes the worker. The default remains five seconds, so existing behavior is unchanged.

docs/fault-tolerance/README.md
```diff
@@ -71,6 +71,12 @@ Dynamo provides multiple health check mechanisms:
 
 See [Health Checks](../observability/health-checks.md) for details.
 
+### Local Worker Inhibition
+
+After a request-path failure, a client temporarily removes that worker from its local routing set while service discovery propagates the worker's state. `DYN_INHIBITED_DURATION_SECS` sets the maximum local inhibition window and is read once during client initialization. The default is `5` seconds; set it to `0` to disable local inhibition. Discovery updates remain authoritative and can restore or remove workers sooner.
+
+Changes to the environment variable take effect the next time the process starts.
+
```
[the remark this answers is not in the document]
kthui: Thanks for the suggestion. The runtime warning already identifies the invalid value and states that the five-second default is used, so we’ll keep the documentation focused on supported configuration values.

docs/fault-tolerance/README.md
```diff
@@ -71,6 +71,12 @@ Dynamo provides multiple health check mechanisms:
 
 See [Health Checks](../observability/health-checks.md) for details.
 
+### Local Worker Inhibition
+
+After a request-path failure, a client temporarily removes that worker from its local routing set while service discovery propagates the worker's state. `DYN_INHIBITED_DURATION_SECS` sets the maximum local inhibition window and is read once during client initialization. The default is `5` seconds; set it to `0` to disable local inhibition. Discovery updates remain authoritative and can restore or remove workers sooner.
```
nnshah1: client here is a little confusing - maybe a 'runtime client'

nnshah1:
```suggestion
After a request-path failure, that worker is removed from the local routing set while service discovery propagates the worker's state. `DYN_INHIBITED_DURATION_SECS` sets the maximum local inhibition window and is read once during client initialization. The default is `5` seconds; set it to `0` to disable local inhibition. Discovery updates remain authoritative and can restore or remove workers sooner.
```

docs/fault-tolerance/README.md
```diff
@@ -86,6 +92,7 @@ For Kubernetes deployments, [Shadow Engine Failover](../kubernetes/shadow-engine
 | Decode blocks threshold | `DYN_ACTIVE_DECODE_BLOCKS_THRESHOLD` | unset |
 | Prefill tokens threshold | `DYN_ACTIVE_PREFILL_TOKENS_THRESHOLD` | unset |
 | Prefill tokens fraction threshold | `DYN_ACTIVE_PREFILL_TOKENS_THRESHOLD_FRAC` | unset |
+| Local worker inhibition | `DYN_INHIBITED_DURATION_SECS` | `5` seconds (`0` disables) |
```
nnshah1: Question: does this effect direct, and random and round robin?

I think semantically it should only effect random and round robin - direct should go through without inhibition - is that feasible?

lib/runtime/src/component/client.rs
```diff
@@ -648,7 +683,7 @@ impl Client {
                             cancel_token.cancel();
                         }
                     }
-                    _ = tokio::time::sleep(reconcile_interval) => {
+                    _ = tokio::time::sleep(reconcile_interval), if !reconcile_interval.is_zero() => {
```
nnshah1, line 686: Question: how does this effect overloaded instances?

lib/runtime/src/component/client.rs
```diff
@@ -15,6 +15,7 @@ use futures::StreamExt;
 use rand::Rng;
 
 use crate::component::{Endpoint, Instance};
+use crate::config::environment_names::runtime as env_runtime;
```
michaelfeil, line 18: Typically uses OnceLock.

michaelfeil (commented):
Seems good to me. If duration is zero, you can perform additional performance savings, e.g. cloning client per connection etc. most of them were rather cheap clones, like cancellation token and client arc IIRC.

docs/fault-tolerance/README.md
kthui: Thanks—updated this to “a runtime client” for clarity.

kthui: Thanks. I kept the actor explicit to clarify that the worker is removed only from that runtime client’s local routing set, while adopting “runtime client” to resolve the ambiguity.

docs/fault-tolerance/README.md
kthui: Yes—implemented. Random and round-robin remain inhibition-filtered; direct dispatch honors the selected worker while it remains in discovery.

lib/runtime/src/component/client.rs
kthui: Thanks, Michael. I used `LazyLock` because it handles the self-contained first-access initialization in one step with less boilerplate, and both types are in `std::sync`. I’m happy to switch to `OnceLock` if that’s the preferred pattern.

lib/runtime/src/component/client.rs
kthui: Good question. If an overloaded instance is temporarily excluded through the penalty-box (`report_instance_down`) path, setting the duration to `0` disables that layer of protection. Explicit `ResourceExhausted` responses are handled separately by `self.client.mark_overloaded_immediate(instance_id)`, so that overload protection remains active.

lib/runtime/src/component/client.rs
michaelfeil: its just what much of the codebase uses.

michaelfeil (approved):

lib/runtime/src/config/environment_names.rs
```diff
@@ -756,6 +759,7 @@ mod tests {
             runtime::DYN_RUNTIME_NUM_WORKER_THREADS,
             runtime::DYN_RUNTIME_MAX_BLOCKING_THREADS,
             runtime::DYN_RUNTIME_GRACEFUL_SHUTDOWN_TIMEOUT_SECS,
+            runtime::DYN_INHIBITED_DURATION_SECS,
```
nnshah1: for consitency 
```suggestion
            runtime::DYN_RUNTIME_INHIBITED_DURATION_SECS,
```

lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -812,19 +812,9 @@ where
     where
         F: FnOnce(&mut T, u64) -> anyhow::Result<M>,
     {
-        // When fault detection is disabled, check the raw discovery list
-        // (not filtered by report_instance_down) so transient failures
-        // don't poison the instance for subsequent retries.
-        let found = {
-            if self.fault_detection_enabled {
```
nnshah1, line 819: question: if we already had a "fault_detection_enabled" flag that can disable fault_detection (including inhibition it seems) - why did we need this change? Is it that we didn't have this exposed by a env variable? If we also expose this flag - what additional does it control besides local worker inhibition?

nnshah1 (commented):
one small naming change and a question about the  other boolean.

nnshah1 (approved):
naming change and a question on the purpose of the other variable gating inhibition. I think having a flag to disable detection and having a flag to set the inhibition time is reasonable.

sttts (approved):

lib/runtime/src/config/environment_names.rs
kthui: Thanks, Neelay—updated the constant and environment variable to `DYN_RUNTIME_INHIBITED_DURATION_SECS` for consistency.

lib/runtime/src/pipeline/network/egress/push_router.rs
kthui: Good question. The distinction is both scope and behavior. `fault_detection_enabled` is an internal, per-router construction setting; setting it to `false` bypasses the broader fault-detection path, including response-inactivity timeouts and overload checks and marking.

`DYN_RUNTIME_INHIBITED_DURATION_SECS` is an external operator setting limited to the penalty box. A positive value configures the inhibition duration, while `0` makes `report_instance_down` a no-op. The remaining fault-detection protections stay enabled.

For direct dispatch, local inhibition is bypassed regardless, but the timeout and overload behavior still differs.

Let's take the discussion offline on if we are going to expose the `fault_detection_enabled` as an environment variable.

PR #11880, opened 2026-07-19
feat(sglang): support frontend image decoding in multimodal EPD

Chokoyo:
## Overview

Enable SGLang multimodal E/PD deployments to consume images decoded by the Dynamo frontend.

The frontend sends decoded RGB pixels and a canonical content hash to the encode worker. The encode worker loads pixels through NIXL, runs the existing SGLang `MMEncoder`, and forwards the resulting embeddings to the PD worker.

This change is image-only. Existing URL-based image and video paths remain supported; frontend-decoded video is out of scope, add support after https://github.com/ai-dynamo/dynamo/pull/11364 is merged.

## Details

- Allow `--frontend-decoding` on the frontend-facing E/PD encode worker while continuing to reject it on internal multimodal workers.
- Accept both `Url` and `Decoded` image wire variants in the encode worker.
- Compute and serialize the canonical `xxh3-64(shape + dtype + decoded RGB bytes)` key on the Rust decode thread.
- Reuse the canonical key for MM-aware routing and the SGLang embedding cache.
- Probe the embedding cache before reading decoded pixels through NIXL, allowing cache hits to skip both pixel transfer and vision encoding.
- Preserve mixed cache hit/miss ordering and existing URL-based behavior.
- Release decoded PIL buffers before streaming downstream responses.
- Add launcher support and post-merge E/PD coverage using Qwen3-VL.

## Where should the reviewer start?

- `components/src/dynamo/sglang/request_handlers/multimodal/encode_worker_handler.py`
- `lib/llm/src/preprocessor/media/rdma.rs`
- `components/src/dynamo/sglang/backend_args.py`
- `tests/serve/test_sglang.py`

## Validation

- Targeted SGLang unit tests: 86 passed
- `cargo test -p dynamo-llm content_hash`: 2 passed
- `pre-commit run --from-ref origin/main --to-ref HEAD`
- SGLang E/PD frontend-decoding E2E: 1 passed in 55.69s
- Verified repeated image requests reuse the embedding cache

## Related Issues
- DIS-2451 (Support Matrix)

## Summary by CodeRabbit

* **New Features**
  * Added frontend decoding support for multimodal image processing in SGLang launch workflows.
  * Added embedding-cache support for frontend-decoded media, improving reuse across repeated requests.
  * Added stable content-based caching for decoded images.
  * Added stricter validation for multimodal media payloads and clearer configuration errors.

* **Bug Fixes**
  * Prevented ambiguous or malformed media variants from being accepted.
[...]

PR #12004, opened 2026-07-22
feat(vllm): support frontend image decoding with EPD

Chokoyo:
## Summary

Enable opt-in frontend image decoding for vLLM multimodal E/PD and E/P/D deployments. The existing backend-decoding path remains the default.

Prefill/PD forwards only the decoded-image descriptor. On a cache miss, the Encode worker reads RGB pixels directly from the frontend-owned buffer with one NIXL transfer; pixels are not relayed through Prefill/PD. 

NIXL for decoded pixels transfers only **ONCE** from frontend to Encoder worker on EC miss.

```mermaid
sequenceDiagram
    participant F as Frontend
    participant P as Prefill / PD
    participant E as Encode

    F->>F: Decode and register RGB
    F->>P: Request + NIXL descriptor

    alt Prefill cache hit
        P->>P: Reuse embeddings
    else Prefill cache miss
        P->>E: Forward descriptor only
        alt Encode cache miss
            F-->>E: RGB via one direct NIXL transfer
            E->>E: Vision encode
        end
        E-->>P: Embeddings
    end

    P->>P: Prefill and continue to Decode
```

## Details

- Forward frontend-decoded image descriptors from Prefill/PD to the Encode worker.
- Read decoded RGB pixels directly from frontend memory through NIXL.
- Reuse the canonical `content_hash` for Prefill and Encode embedding-cache keys.
- Validate URL and decoded-image variants at dispatch and encode boundaries.
- Bypass caching when a canonical content hash is unavailable.
- Add opt-in `--frontend-decoding` support to the E/PD and E/P/D launchers.
- Preserve existing URL-image behavior and process videos on Prefill/PD.
- Add unit coverage for E/PD and E/P/D serving cases.

## Where Should the Reviewer Start?

- `components/src/dynamo/vllm/multimodal_utils/request_processor.py`
- `components/src/dynamo/vllm/multimodal_utils/prefill_worker_utils.py`
- `components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py`
- `examples/backends/vllm/launch/disagg_multimodal_e_pd.sh`
- `examples/backends/vllm/launch/disagg_multimodal_epd.sh`

## Validation

- Targeted unit tests: `12 passed`
- Pre-commit on affected files: passed
- Local E/PD E2E: two base64 frontend-decoded requests returned `Green`; the second request hit the embedding cache.
- Local E/P/D E2E: `mm_epd_qwen3-vl-2b_b64_frontend_decoding` passed in 91.71 seconds and returned `green`.
- Added E/PD and E/P/D cases to post-merge CI.

## Related Issues
- Based on #11880
- DIS-2451

## Summary by CodeRabbit

* **New Features**
  * Added optional frontend image decoding for multimodal inference.
  * Supports both image URLs and frontend-decoded image data in requests.
  * Added `--frontend-decoding` options to multimodal launch scripts.
  * Enables content-based caching for decoded images when valid identifiers are available.
* **Bug Fixes**
  * Added validation for malformed, unsupported, conflicting, or empty image inputs.
  * Prevents invalid images from being dispatched or cached.
* **Tests**
  * Added coverage for decoded-image processing, caching, validation, and launch topologies.


components/src/dynamo/vllm/multimodal_utils/request_processor.py
```diff
@@ -488,21 +493,26 @@ async def extract_multimodal_data(
...
-                        image_urls.append(item[URL_VARIANT_KEY])
-                    else:
+                    if isinstance(item, dict) and UUID_ONLY_VARIANT_KEY in item:
+                        supported = False
+                        break
+                    _url, decoded = parse_image_item(item)
+                    if decoded is not None and not self.enable_frontend_decoding:
                         supported = False
+                        break
+                    image_items_for_encoder.append(item)
```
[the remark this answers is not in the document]
Chokoyo: This is expected. uuid-only request expected a embedding cache hit and skipped. If there is a EC miss, the request should failed.

- https://docs.vllm.ai/en/stable/features/multimodal_inputs/#embedding-inputs

tests/serve/multimodal_profiles/vllm.py
```diff
@@ -236,14 +236,32 @@
...
+                tests=[
+                    MmCase(payload=make_image_payload(["green"])),
+                    # Rust frontend decode -> NIXL RGB transfer -> Encode ->
+                    # Prefill embedding handoff -> Decode generation.
+                    MmCase(
+                        suffix="b64_frontend_decoding",
+                        payload=make_image_payload_b64(["green"]),
+                        extra_script_args=["--frontend-decoding"],
+                    ),
+                ],
```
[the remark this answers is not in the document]
Chokoyo: Profiled both E/P/D cases with tests/utils/profile_pytest.py. Reduced the P/D KV cache from 1.714 GB to 512 MB without changing the model. Both the regular and frontend-decoding cases passed and peaked at 18.7 GB, so the topology can remain in the 24 GB post-merge tier

lib/llm/src/kv_router/encoder_router.rs
```diff
@@ -318,20 +342,29 @@ impl
...
-                .context("Encoder router is active but not initialized")?;
+        let binding = self.binding.load_full();
+        let encode_future = async move {
+            let binding = binding.context("Encoder router is active but not initialized")?;
             let response = binding.router.generate(encode_context).await?;
             Self::consume_encode_stream(response).await
-        }
-        .await;
+        };
+        let encode_result = await_encode_with_source_guards(encode_future, source_guards).await;
```
KrishnanPrash: **AI Generated Comment**

**Encoder-hop guard may never run for vLLM**

Does anything advertise `encoder_result_handoff` for vLLM? I ask because activation depends on it: `model_manager.rs:2146` reads that key from the card and defaults to `false`, which makes `set_target(None)` fire, which leaves the router `Unavailable`, which returns early at the top of `generate()`:

```rust
// :340, first line of generate()
if self.lifecycle_state() != Active
    || !Self::should_encode(&request)
{
    return next.generate(...).await;
}
```

The only setter I can find in the repo is `model_manager.rs:3251`, inside `#[cfg(test)]`. If that is right, the hop these launch scripts use is the Python one at `prefill_worker_utils.py:222`, and it has no equivalent guard. That path is exposed because the frontend media cache is opt-in and off by default (`loader.rs:384`), so the in-flight request is the only owner of the pinned buffer.

If a follow-up will advertise the capability, then this is groundwork and the code is fine, but the lifetime thread should stay open rather than resolved.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
```diff
@@ -190,36 +243,20 @@ async def generate(
...
+                        wire_items.append(
+                            {DECODED_VARIANT_KEY: group_mm_input.image_decoded}
                         )
-                        continue
-                    loaded_images.append(result)
-                if collective_exceptions:
-                    raise ValueError(
-                        f"Errors occurred during image loading:\n{collective_exceptions}"
-                    )
+                loaded_images = await self.image_loader.load_image_batch(wire_items)
```
KrishnanPrash, line 259: **AI Generated Comment**

**Grayscale images look like they break on the decoded path**

This call now sends decoded descriptors through `_read_and_convert_nixl_image`, which I think only handles RGB. The Rust decoder keeps the source channel count, so a grayscale PNG arrives as `(H, W, 1)` (`decoders/image.rs:89`), and the one normalization step is `array[..., :3]` (`media_nixl.py:108`), which does nothing to a 1-channel array. `Image.fromarray` then raises on it. The URL path avoids this only because it ends in `.convert("RGB")`.

```python
# image_loader.py:184
# before
return Image.fromarray(arr)
# after
return Image.fromarray(arr).convert("RGB")
```

Pre-existing on the aggregated path, but this PR makes it reachable in E/PD and E/P/D, and the new CI cases are RGB only.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
```diff
@@ -119,6 +129,52 @@ def cleanup(self):
...
+                "encode worker."
+            )
+        if has_url:
+            return EmbeddingCache.generate_hash_key(group_input.image_url)
+        if not self._enable_frontend_decoding:
+            raise ValueError(
+                "Received a frontend-decoded image but --frontend-decoding is "
+                "not enabled on the encode worker. Enable it on both the "
+                "frontend-facing worker and the encode worker."
+            )
```
KrishnanPrash, line 162: **AI Generated Comment**

**Flag mismatch is only caught per request**

The message is clear, but this fires per request instead of at startup. If `--frontend-decoding` reaches the PD worker and not the encode worker, every image fails after the frontend has already fetched, decoded, and registered it.

Could this move to a startup check? SGLang does exactly that in `backend_args.py:229` with `validate_multimodal_topology()`. The encode worker also already registers a card (`worker_factory.py:734`), so advertising the flag there would remove the need to keep two launch flags in sync by hand.

KrishnanPrash (commented):
Left a series of AI generated comments. Let me know what you think 👍

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
```diff
@@ -119,6 +129,52 @@ def cleanup(self):
...
+            and self.embedding_cache is not None
+            and not self._decoded_content_hash_warning_emitted
+        ):
+            logger.warning(
+                "Frontend-decoded image descriptor has a missing or invalid "
+                "canonical content_hash; this item will bypass the encode-worker "
+                "embedding cache. Ensure the frontend and encode worker use "
+                "compatible Dynamo versions and the descriptor is not corrupted."
+            )
+            self._decoded_content_hash_warning_emitted = True
```
KrishnanPrash, line 175: **AI Generated Comment**

**A missing content hash turns both caches off quietly**

The skip-the-cache behaviour is right. My concern is only how visible it is. `content_hash` is `skip_serializing_if` (`rdma.rs:52`), so a version-skewed frontend just omits it and every decoded item goes unkeyed. This warns once per process, the prefill side warns not at all, and there is no counter.

A deployment could run for weeks doing a full vision encode per request with both caches off. The once-only flag also means one bad early request silences every later warning. A counter would make this visible.

lib/llm/src/kv_router/encoder_router.rs
```diff
@@ -25,11 +26,34 @@ use dynamo_runtime::{
...
+    }
+
+    match tokio::spawn(async move {
+        let _source_guards = source_guards;
+        future.await
+    })
+    .await
+    {
+        Ok(result) => result,
+        Err(error) if error.is_panic() => std::panic::resume_unwind(error.into_panic()),
```
KrishnanPrash: **AI Generated Comment**

**Detached hop has no deadline or cap**

Scoping this to the non-empty case is the right call. The part worth bounding is what the spawned branch commits to: it drains the whole encode stream with no deadline, and nothing caps how many can be in flight. A wedged encode worker plus a burst of disconnects leaves N tasks each pinning a decoded buffer until the stream ends.

Worth noting in a comment either way that this intentionally drops the last cancellation path, since `Context::with_id_and_metadata` already gives the encode side its own controller. Fine as a follow-up if the `encoder_result_handoff` question above holds.

tests/serve/multimodal_profiles/vllm.py
```diff
@@ -236,14 +236,33 @@
...
             ),
             "epd": TopologyConfig(
                 marks=[pytest.mark.post_merge],
                 timeout_s=300,
                 single_gpu=True,
-                requested_vllm_kv_cache_bytes=1_714_881_000,
-                tests=[MmCase(payload=make_image_payload(["green"]))],
+                profiled_vram_gib=18.7,
+                requested_vllm_kv_cache_bytes=536_870_912,
```
KrishnanPrash, line 255: **AI Generated Comment**

**KV cache value contradicts the script's own comment**

512 MiB is the exact value the launch script calls unusable: `disagg_multimodal_epd.sh:166-169` says vLLM 0.20+ rejects it and defaults to 2 GiB. Your profiling run passed, so I think the comment is the stale half rather than this value, and five topologies in this file now use 512 MiB.

Could you scope that comment to LLaVA-1.5-7b instead of stating it as a general floor? Otherwise the next person sizing a topology will believe it.

components/src/dynamo/vllm/multimodal_utils/prefill_worker_utils.py
```diff
@@ -228,47 +271,49 @@ async def _fetch_embeddings(
...
-    for idx, url in enumerate(image_urls):
+    for idx, item in enumerate(image_items):
         if cache is not None:
-            key = get_embedding_hash(url)
-            cached = cache.get(key)
+            key = _image_item_cache_key(item)
+            cached = cache.get(key) if key is not None else None
             if cached is not None:
-                logger.debug(f"[{request_id}] Cache hit for URL index {idx}")
+                logger.debug(f"[{request_id}] Cache hit for image index {idx}")
```
KrishnanPrash: **AI Generated Comment**

**f-string in a logging call**

`.ai/python-guidelines.md:274` asks for lazy formatting in logging calls. Same point CodeRabbit raised earlier.

```python
# before
logger.debug(f"[{request_id}] Cache hit for image index {idx}")
# after
logger.debug("[%s] Cache hit for image index %d", request_id, idx)
```

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
Chokoyo: Fixed in 63ab5e20ea. Added to RGB conversion in frontend

components/src/dynamo/vllm/multimodal_utils/prefill_worker_utils.py
Chokoyo: Fixed in 63ab5e20ea.

lib/llm/src/kv_router/encoder_router.rs
Chokoyo: Removed the Rust detached/source-guard path in 63ab5e20ea. Tested NIXL Read for pixel transfer from Frontend to E, didn't obeserved issues in https://github.com/ai-dynamo/dynamo/pull/12004#discussion_r3731100728 for now.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
Chokoyo: Follow up this in future PRs

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
Chokoyo: Follow up this in future PR

tests/serve/multimodal_profiles/vllm.py
Chokoyo: Added the comment in bbdc8e7584 for future reference.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
```diff
@@ -190,36 +243,20 @@ async def generate(
...
+                        wire_items.append(
+                            {DECODED_VARIANT_KEY: group_mm_input.image_decoded}
                         )
-                        continue
-                    loaded_images.append(result)
-                if collective_exceptions:
-                    raise ValueError(
-                        f"Errors occurred during image loading:\n{collective_exceptions}"
-                    )
+                loaded_images = await self.image_loader.load_image_batch(wire_items)
```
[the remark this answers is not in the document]
rmccorm4: **[P1] Retain the frontend NIXL source until this read completes**

Confirmed this is an active correctness bug at the current head. `RdmaMediaDataDescriptor::source_storage` is the only in-flight owner of the registered allocation and is skipped during serialization (`lib/llm/src/preprocessor/media/rdma.rs:55-58`); the frontend media cache that could provide another owner is disabled by default (`loader.rs:388-416`). `AddressedPushRouter::generate` borrows the request only until the worker response prologue arrives and returns a stream that does not capture it (`addressed_router.rs:579-622,764-775`). The worker sends that prologue before polling the demand-driven Python generator (`push_handler.rs:640-686`; `bindings/python/rust/engine.rs:72-73,115-138`). The frontend can therefore deregister and free the region before the Python prefill hop dispatches the descriptor and this `load_image_batch` call performs the NIXL read.

Please retain the decoded source `Arc`s until the first real prefill response or terminal close; that item cannot be produced until the encode worker has completed this read. The lifetime guard must also survive client cancellation before the first item, since a wrapper owned only by the caller would be dropped too early.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
rmccorm4: Deferred: this is an observability improvement, not a correctness requirement for frontend decoding. A missing or invalid `content_hash` already bypasses the cache safely and emits a warning. Adding a cache-bypass metric is out of scope for this PR and can be considered in a follow-up.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
rmccorm4: Rejecting per request with clear error for now is fine for this PR

lib/llm/src/kv_router/encoder_router.rs
rmccorm4: Superseded by 63ab5e20ea and completed in 1a48ec95d0. The encoder_router detached path was removed, so this no longer depends on encoder_result_handoff being advertised. Source lifetime retention now runs in the generic AddressedPushRouter path used by the vLLM Python encode hop.

lib/llm/src/kv_router/encoder_router.rs
rmccorm4: Fixed in 1a48ec95d0. The old whole-stream detached drain is gone. Only dispatch through the first response is detached, retained dispatches are process-wide capped at 1024 with ResourceExhausted on saturation, and the stream tail is handed back to normal caller-owned polling and cancellation.

components/src/dynamo/vllm/multimodal_handlers/encode_worker_handler.py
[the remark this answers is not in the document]
rmccorm4: Fixed in 1a48ec95d0. Migration now retains the decoded source_storage Arcs in a take-once guard, propagates it to the actual prefill/PD request context, and AddressedPushRouter keeps it alive through the first real worker response even if the caller drops its waiter. Focused cancellation and lifetime tests were added in addressed_router.rs.

rmccorm4:
/ok to test [1a48ec9](https://github.com/ai-dynamo/dynamo/pull/12004/commits/1a48ec95d007c09b3cb5ab56f80e08449f7ba857)

rmccorm4 (approved):

components/src/dynamo/common/multimodal/image_loader.py
```diff
@@ -178,10 +178,10 @@ async def _read_and_convert_nixl_image(
         """Read decoded image via NIXL and convert numpy array to PIL Image."""
         assert self._nixl_connector is not None
         arr = await read_decoded_media_via_nixl(self._nixl_connector, metadata)
-        # TRT-LLM's input processor requires PIL Images (accesses .height/.width
-        # for token count calculation). fromarray() is near-zero-cost: it wraps
-        # the existing numpy buffer without copying pixel data.
-        return Image.fromarray(arr)
+        if arr.ndim == 3 and arr.shape[-1] == 1:
```
GuanLuo, line 181: Why decoded media shape may be like this?

components/src/dynamo/common/multimodal/image_loader.py
```diff
@@ -178,10 +178,10 @@ async def _read_and_convert_nixl_image(
...
+            arr = arr.squeeze(axis=-1)
+        image = Image.fromarray(arr)
+        return image if image.mode == "RGB" else image.convert("RGB")
```
GuanLuo, line 184: Will `image.convert("RGB")` be no-op if `image.mode == "RGB"`? If so, can always `return image.convert("RGB")`

lib/llm/src/migration.rs
```diff
@@ -411,6 +412,21 @@ where
...
+                .as_ref()
+                .into_iter()
+                .flat_map(|media| media.values())
+                .flatten()
+                .filter_map(|item| match item {
+                    MultimodalData::Decoded(descriptor) => descriptor.source_storage.clone(),
+                    _ => None,
+                })
+                .collect::<Vec<_>>();
+            if !source_guards.is_empty() {
```
GuanLuo, line 427: Are you using RAII to make sure the source storage is not released until read (use first response as indicator)?

GuanLuo: This seems to be missing piece of frontend decoding, any ideas on why this issue wasn't exposed previously?

GuanLuo (approved):

components/src/dynamo/common/multimodal/image_loader.py
rmccorm4: No it won't be a no-op - Pillow returns self.copy() when the requested mode already matches. I verified it produces a distinct image object and pixel core. Unconditionally converting would add a full image allocation/copy on the common RGB path.

components/src/dynamo/common/multimodal/image_loader.py
rmccorm4: The Rust decoder preserves the source channel count while always producing HWC tensors. Grayscale becomes (H, W, 1)

Pillow rejects (H, W, 1) arrays; it accepts (H, W) as mode L. The squeeze in image_loader.py (line 181) is therefore necessary. This is covered at test_image_loader.py (line 301)

lib/llm/src/migration.rs
rmccorm4: We retain the NIXL source storage with an RAII guard until the worker returns its first real response, which confirms the image read has completed. The guard also survives client cancellation.

This was previously hidden because normal requests kept the storage alive long enough. E/P/D adds another hop before the Encode worker reads the image, making early cancellation expose the lifetime race. The new cancellation test covers that case.

rmccorm4:
/ok to test [1635ddb](https://github.com/ai-dynamo/dynamo/pull/12004/commits/1635ddb30c5cb97e8e3134fe0d76e54533cccf3c)

ishandhanani (approved):

PR #12011, opened 2026-07-22
feat(gms): add GMS V1 lifecycle

galletas1712:
## Summary

Adds an experimental, isolated GMS V1 process-restore lifecycle while leaving the existing V0 implementation and behavior unchanged.

V1 preserves committed weight backing across suspend/resume, releases ephemeral KV backing, and remaps both domains into the preserved engine process. V0 remains the default; V1 is selected explicitly with `--use-v1`.

This is the bottom PR in stack #12691. #12392 adds exact-ID weight artifact save/load on top.

## Scope

This PR adds:

- typed Unix-socket RPC with transient CUDA FD transfer;
- RW/RO admission, publication, and disconnect cleanup;
- V1-local allocation, mapping, session, and server policy;
- persistent weight and ephemeral KV domains;
- a concise vLLM sleep/wake integration; and
- additive rank-local and supervisor `--use-v1` launch paths.

V0 and V1 are mutually exclusive launch profiles. V1 reuses the unchanged V0 allocator extension; it does not add a second allocator shim or extract V0 into a shared core.

This PR intentionally excludes weight artifact save/load (#12392), SGLang/TensorRT-LLM V1 integrations, operator/recipe changes, and KV failover/layout adoption.

## Validation

Current head: `f8895f5a50e655e9f3ec31a84f80fcbc008656bb`

- complete available GMS suite on the final stack: `66 passed, 8 skipped`;
- focused V1 protocol, memory-manager, parameter-storage, and server lifecycle coverage;
- clean wheel and sdist build, including packaged V1 modules;
- Ruff, compileall, pre-commit, and `git diff --check` passed; and
- every changed path resolves only to `@ai-dynamo/dynamo-gms-codeowners`.

- exact-head V0 shadow-engine failover on an nscale B200: vLLM `1 passed in 179.78s`, SGLang `1 passed in 118.85s`; TensorRT-LLM remains source-skipped for OPS-4450. This is V0 regression evidence and does not validate V1.

### DIFF
--- a/lib/gpu_memory_service/v1/README.md
+++ b/lib/gpu_memory_service/v1/README.md
@@ -1,0 +1,169 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# GPU Memory Service V1
+
+GMS V1 is an experimental, rank-local memory owner for engines restored by
+Dynamo Snapshot. It is fail-stop: errors terminate the worker rather than
+retrying, rolling back, falling back to native allocation, or continuing after
+a partial wake.
+
+## Snapshot assumptions
+
+Dynamo Snapshot captures only after whole-engine GMS sleep. Restore preserves:
+
+- the same Python and Torch process state;
+- the same TensorImpls and post-partition StorageImpl graph;
+- tensor layouts;
+- GMS allocation IDs; and
+- CUDA virtual-address reservations.
+
+Model construction, model loading, and non-Parameter storage copy-out do not
+run again after restore. The rank-local GMS sidecar survives separately and
+retains committed weight allocations. Default-allocator copies are ordinary
+process-owned Snapshot state.
+
+KV TensorImpls, mapping records, allocation IDs, sizes, and VA reservations also
+survive. KV physical backing and contents do not. Wake creates fresh backing
+under the saved IDs and maps it at the preserved VAs before vLLM prepares the
+cache for use.
+
+## Ownership
+
+`GMSV1Worker._maybe_get_memory_pool_context(tag)` is the vLLM routing seam.
+`weights` and `kv_cache` delegate to one `GMSV1SleepModeBackend`; every other
+tag follows vLLM's normal implementation.
+
+```text
+GMSV1Worker
+  _maybe_get_memory_pool_context(tag)
+    -> GMSV1SleepModeBackend
+         owns one CUDAPluggableAllocator
+         owns temporary weights MemPool + long-lived KV MemPool
+         owns two GMSClientMemoryManager instances
+              weights -> weights.sock
+              kv      -> kv_cache.sock
+
+sidecar per rank:
+  weights.sock  -> GMSServerMemoryManager + independent allocations and sessions
+  kv_cache.sock -> GMSServerMemoryManager + independent allocations and sessions
+```
+
+The pluggable allocator outlives both MemPools. Allocation callbacks route by
+the active domain. Free callbacks route by VA ownership because Torch can free
+outside the context that allocated the storage. C free callbacks cannot
+reliably propagate Python exceptions, so the backend latches and surfaces the
+first callback failure.
+
+V1 reuses the existing
+`gpu_memory_service.client.torch.extensions._allocator_ext` native shim
+unchanged and permits one V1 allocator/backend owner per process. The default
+V0 and `--use-v1` launch profiles are mutually exclusive. Mixed initialization
+is unsupported by contract and prevented by launch/process topology; no runtime
+cross-profile arbitration is provided or needed.
+
+Both client domains use the same `GMSClientMemoryManager` class and the same
+V0-style operations:
+
+| Operation | Purpose |
+|---|---|
+| `connect` / `disconnect` | Acquire or release the socket lease |
+| `create_mapping` / `destroy_mapping` | Own one server allocation and local VA |
+| `commit` | Change the same weights socket and mappings from RW to RO |
+| `unmap_all_vas` | Drop imported handles but preserve IDs, sizes, and VAs |
+| `reallocate_all_handles` | Recreate ephemeral server backing under saved IDs |
+| `remap_all_vas` | Install server backing at saved VAs |
+| `close` | Release local mappings and the socket |
+
+Each Unix-domain socket is its lease. A writer excludes all other sessions.
+Committed weights can have shared readers. Waiting writers have priority over
+new readers. A same-socket commit changes RW to RO atomically. Disconnecting an
+uncommitted writer clears its complete allocation epoch before another writer
+is admitted. Reader disconnect only releases that reader.
+
+The server retains CUDA handles rather than persistent export FDs. Each export
+creates a transient FD, transfers it with `SCM_RIGHTS`, and closes the server
+copy. Client import consumes and closes the received copy. The wire protocol is
+a small typed `msgspec.Struct` MessagePack protocol; there is no JSON
+compatibility path.
+
+## Weight construction
+
+V1 uses vLLM's normal model loader inside its broad weights allocation scope:
+
+1. Connect `weights` RW and enter the temporary GMS weights MemPool.
+2. Run model construction, loading, quantization, and post-load transforms.
+3. Leave the GMS MemPool.
+4. Copy live non-Parameter tensors from GMS storage to Torch's default
+   allocator while preserving TensorImpl identity and non-Parameter aliases.
+5. Synchronize CUDA.
+6. Destroy the weights MemPool and collect dead mappings.
+7. Surface any allocator callback failure.
+8. Protect surviving Parameter mappings RO and commit the same socket RW to RO.
+
+The storage copy operates on each overlapping connected component of bounding
+storage byte ranges, not on an entire StorageImpl and not once per tensor.
+Relative aliases and offsets inside a copied component are preserved. Disjoint
+components get separate storage. Absolute non-empty storage offsets may be
+rebased. Mixed Parameter/non-Parameter aliasing is deliberately severed.
+
+See `parameter_storage.py` for the detailed before/after diagram and
+implementation.
+
+The commit log reports Parameter span bytes, retained aligned GMS bytes,
+uncovered retained bytes and ratio, copied-out bytes, and retained allocation
+count.
+
+## Sleep and wake
+
+```mermaid
+sequenceDiagram
+    participant W as vLLM worker
+    participant WM as weights manager
+    participant KM as KV manager
+    participant WS as weights server
+    participant KS as KV server
+
+    W->>WM: unmap_all_vas, disconnect RO
+    Note over WS: committed weight backing remains
+    W->>KM: unmap_all_vas, disconnect RW
+    Note over KS: clear KV epoch
+    Note over W: Dynamo Snapshot captures sleeping engine
+
+    W->>KM: connect RW
+    W->>KM: reallocate_all_handles, remap_all_vas
+    Note over KS: fresh KV backing at preserved VAs
+    W->>WM: connect RO against saved server identity/GPU
+    W->>WM: remap_all_vas
+```
+
+Suspend order is weights then KV so the exclusive KV lease remains held until
+local weight memory is asleep. Resume order is KV then weights.
+
+Server nonce and physical GPU identity are saved on first connection. Every
+wake must match them. V1 does not reconstruct models, retain KV contents, scan
+raw mappings, validate model-specific layouts, or implement SGLang integration.
+
+## Running V1
+
+Start one V1 child per visible device:
+
+```text
+python3 -m gpu_memory_service.cli.server --use-v1
+```
+
+The supervisor discovers the visible devices and monitors the children. To
+start one rank-local child directly:
+
+```text
+gpu-memory-service --use-v1 --device 0
+```
+
+Select the worker while retaining vLLM's normal load format:
+
+```text
+python -m dynamo.vllm ... \
+  --worker-cls gpu_memory_service.v1.integrations.vllm.worker.GMSV1Worker
+```
--- a/lib/gpu_memory_service/v1/README.md
+++ b/lib/gpu_memory_service/v1/README.md
@@ -1,0 +1,176 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# GPU Memory Service V1
+
+GMS V1 is an experimental, rank-local memory owner for engines restored by
+Dynamo Snapshot. It is fail-stop: errors terminate the worker rather than
+retrying, rolling back, falling back to native allocation, or continuing after
+a partial wake.
+
+## Snapshot assumptions
+
+Dynamo Snapshot captures only after whole-engine GMS sleep. Restore preserves:
+
+- the same Python and Torch process state;
+- the same TensorImpls and post-partition StorageImpl graph;
+- tensor layouts;
+- GMS allocation IDs; and
+- CUDA virtual-address reservations.
+
+Model construction, model loading, and non-Parameter storage copy-out do not
+run again after restore. The rank-local GMS sidecar survives separately and
+retains committed weight allocations. Default-allocator copies are ordinary
+process-owned Snapshot state.
+
+KV TensorImpls, mapping records, allocation IDs, sizes, and VA reservations also
+survive. KV physical backing and contents do not. Wake creates fresh backing
+under the saved IDs and maps it at the preserved VAs before vLLM prepares the
+cache for use.
+
+## Ownership
+
+`GMSV1Worker._maybe_get_memory_pool_context(tag)` is the vLLM routing seam.
+vLLM calls its allocation-scope labels tags. V1 maps the recognized `weights`
+and `kv_cache` tags to isolated GMS domains owned by one
+`GMSV1SleepModeBackend`; every other tag follows vLLM's normal implementation.
+
+```text
+GMSV1Worker
+  _maybe_get_memory_pool_context(tag)
+    -> GMSV1SleepModeBackend
+         owns one CUDAPluggableAllocator
+         owns temporary weights MemPool + long-lived KV MemPool
+         owns two GMSClientMemoryManager instances
+              weights -> weights.sock
+              kv      -> kv_cache.sock
+
+sidecar per rank:
+  weights.sock  -> GMSServerMemoryManager + independent allocations and sessions
+  kv_cache.sock -> GMSServerMemoryManager + independent allocations and sessions
+```
+
+The pluggable allocator outlives both MemPools. Allocation callbacks route by
+the active domain. Free callbacks route by VA ownership because Torch can free
+outside the context that allocated the storage. C free callbacks cannot
+reliably propagate Python exceptions, so the backend latches and surfaces the
+first callback failure.
+
+V1 reuses the existing
+`gpu_memory_service.client.torch.extensions._allocator_ext` native shim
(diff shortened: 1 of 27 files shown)


lib/gpu_memory_service/v1/client/memory_manager.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def _check(self) -> None:
+        if self._failure is not None:
+            raise self._failure
+
+    def _latch(self, message: str, cause: Exception) -> RuntimeError:
+        self.disconnect()
+        if self._failure is None:
+            self._failure = RuntimeError(f"{message}: {cause}")
+        return self._failure
```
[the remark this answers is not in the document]
galletas1712: Fixed in `2ae3bb1b`. `_latch` now records the first operational failure and detaches the session before fallible disconnect cleanup. A secondary close failure is logged without replacing or clearing the original latch. The focused regression test verifies subsequent operations raise the same latched exception object.

lib/gpu_memory_service/v1/client/memory_manager.py
```diff
@@ -0,0 +1,264 @@
...
+                    self._identity,
+                )
+                if session.identity[1] != self._gpu_identity():
+                    session.close()
+                    raise RuntimeError("GMS sidecar is on another physical GPU")
+                if self._identity is None:
+                    self._identity = session.identity
+                self._session = session
+            except Exception as exc:
+                raise self._latch("GMS connect failed", exc) from exc
```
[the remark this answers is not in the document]
galletas1712: This is intentional in the base PR. #12011 is the same-sidecar Snapshot profile, so preserving the server nonce protects against reconnecting to the wrong server incarnation. Fresh-server cold hydration belongs to stacked PR #12392, which owns the different identity policy and exact-ID restore flow. We are not weakening the base identity check.

lib/gpu_memory_service/v1/integrations/vllm/backend.py
```diff
@@ -0,0 +1,261 @@
...
+            if device != self._device:
+                raise RuntimeError(
+                    f"allocator callback device {device} != {self._device}"
+                )
+            domain = self._active_domain.get()
+            if domain == _WEIGHTS:
+                return self._weights.create_mapping(size)
+            if domain == _KV_CACHE:
+                return self._kv_cache.create_mapping(size)
+            raise RuntimeError("GMS allocator callback has no active domain")
```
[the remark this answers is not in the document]
galletas1712: No change. PyTorch invokes the selected pluggable allocator callback synchronously on the allocating thread, and `torch.cuda.use_mem_pool()` itself is thread-local. An allocation performed on another host thread would not inherit the selected pool; it would bypass this callback rather than invoke it with a missing `ContextVar`. The supported vLLM construction path and real-CUDA lifecycle test use the synchronous pool scope.

lib/gpu_memory_service/v1/client/session.py
```diff
@@ -0,0 +1,162 @@
...
+                response,
+                received_fd,
+                HandshakeResponse,
+            )
+            self._granted_lock_type = handshake.lock_type
+            self._identity = (handshake.server_nonce, handshake.gpu_uuid)
+        except Exception:
+            self._socket.close()
+            self._socket = None
+            raise
```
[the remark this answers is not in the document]
galletas1712: Deferred as out of scope for #12011. Startup availability is now handled separately: the client retries only `FileNotFoundError` / `ConnectionRefusedError` until its rank-local server starts. A blanket timeout on handshake and every established RPC is not a safe quick fix because lock admission can be workload-shaped and mutating request timeouts create unknown outcomes. That needs an operation-specific liveness policy rather than one socket timeout.

lib/gpu_memory_service/core/client/torch/extensions/allocator.cpp
```diff
@@ -0,0 +1,147 @@
...
+  const Callbacks& cb = callbacks();
+  if (!cb.malloc_cb) {
+    return nullptr;
+  }
+
+  PyGILState_STATE gstate = PyGILState_Ensure();
+
+  PyObject* args = Py_BuildValue("(niK)", size, device, (unsigned long long)stream);
+  PyObject* result = PyObject_CallObject(cb.malloc_cb, args);
+  Py_DECREF(args);
```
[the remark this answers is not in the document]
galletas1712: Fixed in `2ae3bb1b`. Both callback paths now check `Py_BuildValue` before call/decref and release the GIL on failure. The malloc path also checks `PyLong_AsUnsignedLongLong` for a pending Python error before using the returned address. The native extension rebuild and callback/error smoke checks passed.

lib/gpu_memory_service/core/client/torch/storage_rebinding.py
```diff
@@ -0,0 +1,107 @@
...
+"""Alias-preserving storage copies and TensorImpl rebinding."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Iterable
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torch
```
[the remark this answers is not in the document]
galletas1712: Fixed in the final rebased stack (`2ae3bb1b`). Torch imports in `storage_rebinding.py` and `parameter_storage.py`, plus the V1 CLI selector import, are now absolute module-scope imports. An AST scan across the base and both stacked PR ranges found no newly introduced function-local production imports.

lib/gpu_memory_service/core/server/rpc.py
```diff
@@ -0,0 +1,127 @@
...
+                    except Exception:
+                        return
+                finally:
+                    if export_fd >= 0:
+                        os.close(export_fd)
+        except Exception:
+            return
+        finally:
+            if session is not None:
+                self.server.manager.close(session)
```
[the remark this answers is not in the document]
galletas1712: Addressed in `2ae3bb1b` with a deliberately small logging-only change. Expected request failures retain the existing `ErrorResponse` behavior, including manager-side `OSError`; peer transport failures close at debug level; unexpected failures retain tracebacks. The original request/session/FD cleanup semantics are preserved rather than introducing a new RPC policy framework.

lib/gpu_memory_service/core/server/rpc.py
```diff
@@ -0,0 +1,127 @@
...
+
+class GMSRPCServer(socketserver.ThreadingUnixStreamServer):
+    daemon_threads = True
+
+    def __init__(self, path: str, manager: GMSServerMemoryManager):
+        self.path = path
+        self.manager = manager
+        self._prepare_socket_path()
+        super().__init__(path, _GMSRequestHandler)
+        os.chmod(path, 0o600)
```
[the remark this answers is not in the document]
galletas1712: Not addressed; this hardening is out of scope for #12011. We retained the existing post-bind `chmod(0600)`. On the target Linux environment, connecting to a pathname Unix socket requires write permission, so the typical intermediate `0755` mode from umask `022` is not group/other-connectable; same-UID processes are unaffected by changing `0755` to `0600`. Avoiding a process-global umask change keeps this PR focused.

lib/gpu_memory_service/tests/v1/test_torch_integration.py
```diff
@@ -0,0 +1,295 @@
...
+    assert model.weight.tolist()[12:20] == list(map(float, range(12, 20)))
+    assert accounting == (51 * 4, (16 + 4 + 4) * 4)
+
+
+@pytest.mark.post_merge
+@pytest.mark.integration
+@pytest.mark.gpu_1
+@pytest.mark.skipif(not HAS_CUDA, reason="CUDA is required")
+def test_real_cuda_allocator_and_dual_domain_lifecycle() -> None:
+    """Use a subprocess because Torch allocator callbacks are process-global."""
```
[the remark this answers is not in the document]
galletas1712: Fixed in `2ae3bb1b`. The CUDA subprocess lifecycle test now has `@pytest.mark.timeout(180)`, safely above its existing 120-second subprocess timeout. The final real-CUDA test passed against the built wheel.

lib/gpu_memory_service/v1/integrations/vllm/backend.py
```diff
@@ -0,0 +1,261 @@
...
+            "copied_out_bytes=%d retained_gms_allocation_count=%d",
+            self._device,
+            parameter_span_bytes,
+            retained_aligned_gms_bytes,
+            parameter_span_bytes / retained_aligned_gms_bytes,
+            uncovered_retained_gms_bytes,
+            uncovered_retained_gms_bytes / retained_aligned_gms_bytes * 100,
+            copied_out_bytes,
+            len(self._weights.mappings),
+        )
```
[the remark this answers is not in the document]
galletas1712: No change because this path is not reachable after a successful weights commit. Empty allocation sets are rejected by `GMSClientMemoryManager.commit()`, and every accepted mapping has a positive aligned size, so `retained_aligned_gms_bytes` is positive whenever telemetry runs. Adding a denominator fallback would hide a violated lifecycle invariant rather than handle a supported case.

galletas1712:
Review-body-only suggestions that did not have individual inline threads:

- **`zip(..., strict=True)`** — fixed in `2ae3bb1b`.
- **Move the subprocess payload to a helper module** — valid maintainability idea, deferred; extracting a one-off process-isolation probe would expand this already-large base PR without changing behavior.
- **Log every free-callback failure immediately** — deferred. The first allocator failure is already latched and surfaced at lifecycle boundaries; adding additional callback logging is diagnostic follow-up work, not required for correctness here.
- **Redesign the callback singleton to tolerate calls before registration** — deferred. The supported flow registers callbacks before constructing the allocator; a thread-safe publication redesign for unsupported direct symbol invocation is out of scope.
- **Docstring coverage warning** — no repository 80% docstring gate exists. We are not adding broad boilerplate docstrings solely for the advisory bot percentage.
- **PR template** — the description now includes reviewer-start guidance and related-work/issue disposition.

The final review-fix commit is intentionally small (`+124/-22`) and was validated with two full pre-commit passes, repeated core/snapshot tests, real-CUDA lifecycle coverage, retained V0 tests, and wheel/install smoke checks.

galletas1712:
/ok to test [e5a624c](https://github.com/ai-dynamo/dynamo/commit/e5a624c6c68785133c2274c22fe93d4100ac23eb)

lib/gpu_memory_service/v1/client/session.py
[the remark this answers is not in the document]
galletas1712: Agreed with the follow-up conclusion. We are intentionally not adding a blanket socket timeout: lock admission may be workload-shaped, while timing out a mutating RPC leaves its outcome ambiguous. Startup connection availability is handled separately. An operation-specific liveness and recovery policy belongs in a dedicated follow-up, not this lifecycle PR.

lib/gpu_memory_service/v1/cli.py
```diff
@@ -0,0 +1,79 @@
...
+from concurrent.futures import FIRST_COMPLETED, ThreadPoolExecutor, wait
+from contextlib import ExitStack
+from threading import Event
+
+from gpu_memory_service.common.vmm import get_vmm
+from gpu_memory_service.v1 import device as device_identity
+from gpu_memory_service.v1.device import get_socket_path
+from gpu_memory_service.v1.server import GMSRPCServer, GMSServerMemoryManager
+
+_DOMAINS = ("weights", "kv_cache")
```
mohammedabdulwahhab, line 18: I think we were previously referring to these as tags. I like domains better though.

lib/gpu_memory_service/v1/cli.py
```diff
@@ -0,0 +1,79 @@
...
+    gpu_uuid = device_identity.get_device_uuid(args.device)
+    with ExitStack() as stack:
+        servers = [
+            stack.enter_context(
+                GMSRPCServer(
+                    get_socket_path(args.device, domain),
+                    GMSServerMemoryManager(gpu_uuid, vmm, args.device),
+                )
+            )
+            for domain in _DOMAINS
```
mohammedabdulwahhab, line 66: Might there ever be a reason to start only a weights server (or kv cache server)? Do we want to preserve that optionality?

lib/gpu_memory_service/v1/client/mapping.py
```diff
@@ -0,0 +1,89 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""V1-local CUDA VMM mapping records and operations."""
```
mohammedabdulwahhab, line 4: I noticed that these files kind of cleanly separate into server and client concerns. Was it intentional to remove the nesting from v0?

Might be easier to reason about the context given the subpath you are in (but at the cost of nesting)

lib/gpu_memory_service/v1/client/memory_manager.py
```diff
@@ -0,0 +1,272 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mohammedabdulwahhab, line 1: Is there an argument to just call this client.py so a reader can quickly locate/reason the client side APIs?

lib/gpu_memory_service/v1/client/memory_manager.py
```diff
@@ -0,0 +1,272 @@
...
+    def owns(self, va: int) -> bool:
+        with self._lock:
+            return va in self._mappings
+
+    def connect(self, lock_type: RequestedLockType) -> None:
+        with self._lock:
+            self._check()
+            if self._session is not None:
+                raise RuntimeError("GMS memory manager is already connected")
+            try:
```
mohammedabdulwahhab, line 79: Don't these identity checks restrict out of band snapshot?

lib/gpu_memory_service/v1/mapping.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+from dataclasses import dataclass
+
+from gpu_memory_service.common.locks import GrantedLockType
+from gpu_memory_service.common.vmm import VMMDevice
+
+
+@dataclass(frozen=True)
+class LocalMapping:
+    """One local VA reservation associated with opaque server backing."""
```
mohammedabdulwahhab: This might be the most important class in GMS. An more detailed comment here documenting the fields would be useful for a reader.

I think most important is that allocation_id is the join key across the client/server.

lib/gpu_memory_service/v1/README.md
```diff
@@ -0,0 +1,169 @@
...
+process-owned Snapshot state.
+
+KV TensorImpls, mapping records, allocation IDs, sizes, and VA reservations also
+survive. KV physical backing and contents do not. Wake creates fresh backing
+under the saved IDs and maps it at the preserved VAs before vLLM prepares the
+cache for use.
+
+## Ownership
+
+`GMSV1Worker._maybe_get_memory_pool_context(tag)` is the vLLM routing seam.
```
mohammedabdulwahhab, line 35: I think we're now referring this to domain right? We should be consistent across domain/tag.

lib/gpu_memory_service/v1/README.md
```diff
@@ -0,0 +1,169 @@
...
+`weights` and `kv_cache` delegate to one `GMSV1SleepModeBackend`; every other
+tag follows vLLM's normal implementation.
+
+```text
+GMSV1Worker
+  _maybe_get_memory_pool_context(tag)
+    -> GMSV1SleepModeBackend
+         owns one CUDAPluggableAllocator
+         owns temporary weights MemPool + long-lived KV MemPool
```
mohammedabdulwahhab, line 45: Isn't this the other way around?

mohammedabdulwahhab: Aha I see, looks like we destroy the weights mempool as our new pruning operation. We might need some more explanation around this as its a bit confusing.

lib/gpu_memory_service/v1/integrations/vllm/backend.py
```diff
@@ -0,0 +1,291 @@
...
+        try:
+            with (
+                torch.cuda.device(self._device),
+                torch.cuda.use_mem_pool(pool, device=self._device),
+            ):
+                yield
+        finally:
+            self._active_domain.reset(token)
+
+    def _destroy_weights_pool(self) -> None:
```
mohammedabdulwahhab, line 222: This is functionally equivalent to pruning but needs to be better documented. Kind of confusing why we're wholesale destroying the pool.

mohammedabdulwahhab (approved):

lib/gpu_memory_service/v1/README.md
galletas1712: nice catch! yes I'll rename this to domain.

lib/gpu_memory_service/v1/cli.py
galletas1712: Kept `tag` at the vLLM hook boundary because that is vLLM’s term, and clarified in the README that V1 maps the recognized `weights` and `kv_cache` tags to isolated GMS domains. The sidecar and backend internals consistently use domain.

lib/gpu_memory_service/v1/cli.py
galletas1712: I kept coordinated startup for now. The only current V1 caller always needs both domains and treats them as one rank-local sidecar lifecycle. A selector would add unused configuration and allow an unsupported partial topology; we can add it when a real weights-only or KV-only caller defines the desired supervision semantics.

lib/gpu_memory_service/v1/client/mapping.py
galletas1712: Agreed. I reorganized V1 into `v1/client/` (`mapping`, `memory_manager`, `parameter_storage`, `session`) and `v1/server/` (`allocations`, `rpc`). I left the shared wire contract and device identity at the V1 root instead of creating a generic `common` bucket.

lib/gpu_memory_service/v1/client/memory_manager.py
galletas1712: Addressed with a `v1/client/` package rather than one broad `client.py`: the public manager remains easy to find at `v1/client/memory_manager.py`, while its mapping/session/storage collaborators now sit beside it without collapsing distinct responsibilities into one file.

lib/gpu_memory_service/v1/client/memory_manager.py
galletas1712: Yes, and that restriction is intentional for this bottom PR’s process-local sleep/wake path: it must reconnect to the same sidecar identity. The stacked out-of-band artifact PR (#12392) removes the persisted server-nonce requirement so a fresh server can load the artifact, while retaining the physical-GPU check.

lib/gpu_memory_service/v1/mapping.py
galletas1712: Expanded `LocalMapping`’s docstring to document every field and explicitly call out `allocation_id` as the client/server join key, with the distinction between requested bytes, aligned physical bytes, and the preserved VA reservation.

lib/gpu_memory_service/v1/README.md
galletas1712: Updated. The README now distinguishes vLLM allocation-scope tags from the GMS domains to which the two recognized tags are routed.

lib/gpu_memory_service/v1/README.md
galletas1712: Expanded the weight-construction section: destroying the temporary MemPool is the pruning step. It releases cached/unreferenced blocks through free callbacks, but live Parameter-backed allocations remain and form the committed set.

lib/gpu_memory_service/v1/integrations/vllm/backend.py
galletas1712: Added the same explanation directly on `_destroy_weights_pool`: deleting the temporary pool drains cached/unreferenced mappings, while blocks retained by live Parameter storage survive and are committed.

PR #9683, opened 2026-05-18
feat(planner): power infrastructure — pod annotation, RBAC, config, Prometheus

kaim-eng:
Part of the \[PR ai-dynamo/dynamo#9369 split plan\]. **PR 1b — Planner Power Infrastructure (Phase 1).**

**Predecessor:** ai-dynamo/dynamo#9682 — Power Agent DaemonSet (landed on `main`).  <br>**Successor:** [#9684](<https://github.com/ai-dynamo/dynamo/issues/9684>) (Draft) — Power budget enforcement.  <br>**Follow-up:** ai-dynamo/dynamo#11491 — aggregate-planner startup/scaling resolution (pre-existing gap; see below).

## Summary

Wires the planner from config to per-pod power annotation. **Phase 1 only** — it emits and observes power infrastructure; nothing here changes scaling policy to consume power (budget-gated scaling and the AIC closed loop are deferred to later PRs). Power-awareness is **off by default**.

**17-file diff. Contents:**

* **Config (6 fields):** `enable_power_awareness` (default `False`), `total_gpu_power_limit`, `prefill_engine_gpu_power_limit`, `decode_engine_gpu_power_limit`, `power_agent_safe_default_watts`, `power_annotation_interval_seconds` (default 60s sweep throttle). Budget/safe-default are required only when `enable_power_awareness=True`; all watt/budget fields are `ge=1` validated.
* **Annotation sweep (**`core/base.py`**):** `_apply_power_annotations()` emits `dynamo.nvidia.com/gpu-power-limit` per worker pod; `_publish_power_budget_metrics()` exposes advisory Phase-1 gauges. Guarded behind `enable_power_awareness` + non-advisory mode + Kubernetes connector; advisory mode is a hard no-op for pod mutation; steady-state sweeps are throttled with a scale-up force window; blocking K8s calls run via `asyncio.to_thread`. The sweep resolves aggregate (`mode=agg`) workers via `agg_worker_k8s_name`.
* **Kubernetes helpers:** `patch_pod_annotation`, `list_pods_by_label`, `get_component_pods`, plus frontend admission helpers (`list_frontend_pods`, `resolve_frontend_http_port`, `post_busy_threshold`).
* **Prometheus/DCGM:** advisory `get_total_dgd_power()` query (Phase-3 support, not yet wired) + a NaN-filter fix in the increase query.
[...]

PR #10785, opened 2026-06-17
feat: experimental squeeze_evolve (multi-model evolutionary test-time scaling)

Nietzsche2000:
## Summary

Adds `components/src/dynamo/squeeze_evolve/`, a standalone Dynamo service that serves [Squeeze-Evolve](https://arxiv.org/abs/2604.07725) (verifier-free evolutionary test-time scaling) as a chat model for **multi-model orchestration**. A single `/v1/chat/completions` request becomes one Squeeze-Evolve problem: an evolutionary loop generates a population of candidates and, each round, scores candidate groups by answer diversity and routes them across N model tiers, so easy groups run on the cheap models and only the hard ones reach the expensive models, for single-model or better accuracy at lower cost.

- **N single-model routers:** each tier is its own `dynamo.vllm` deployment and the service owns one `KvRouter` per tier. Dynamo does the within-tier worker selection (KV-aware); Squeeze-Evolve does the across-tier selection (by fitness). One model per router.
- **Dynamo-native:** registers as `ModelInput.Text` + `ModelType.Chat` behind `dynamo.frontend`, tokenizes per tier itself, and dispatches via each tier's `KvRouter.generate_from_request` (token-in / token-out). Plain functions + asyncio; no external package, Pydantic, operator registry, or storage. Styled after `thunderagent_router`.
- **Routing direction:** a group whose candidates agree (low answer diversity, "easy") routes to the cheap tier; one that disagrees (high diversity, "hard") routes to the expensive tier. Implemented as negative-diversity fitness so it matches the paper's confidence signal (high confidence = easy = cheap).
- **Config:** tiers via a single `--tiers` JSON flag (cheapest first); per-tier `KvRouter` knobs reuse `dynamo.router`'s `--router-*` flags.
- Status is **experimental:** opt-in service, no default frontend wiring, `--tiers` schema still scoped for review.

## Status

Experimental. Not wired into any default frontend path. Opt-in: start one `dynamo.vllm` worker per tier, then `python -m dynamo.squeeze_evolve --tiers '[...]' --model-name <name>`, then point `dynamo.frontend` at it. Only the registered `--model-name` is affected; other models on the frontend are untouched.

## Validation

[...]

PR #12012, opened 2026-07-22
feat(planner): DGD-owned GPU power caps and budget-aware scaling

kaim-eng:
## Summary

Add Kubernetes-only, power-aware scaling to the Planner. Per-GPU caps are owned by DynamoGraphDeployment (DGD) worker `podTemplate` annotations, propagated to Pods by the operator, and enforced by the Power Agent. The Planner is read-only with respect to those caps: it resolves each role's per-replica power cost at startup and clamps scaling proposals against `total_gpu_power_limit`.

This replaces the Planner-owned Pod PATCH design from #9683, #9684, and #9685. The power path has no Pod write surface and requires no `pods/patch` RBAC.

### Power ownership and admission contract

- Worker components declare `dynamo.nvidia.com/gpu-power-limit` on their DGD `podTemplate`.
- The operator propagates the annotation to worker Pods; the Power Agent applies the requested per-GPU cap.
- After the DGD and worker rollout settle, the Planner verifies the annotation on every relevant labelled, non-terminal Pod, resolves the effective scalar `nvidia.com/gpu` count and `multinode.nodeCount`, and caches watts per logical replica.
- The operator's validating webhook rejects adding, removing, or changing the power annotation on an existing component. For a power-annotated component, it also makes the effective scalar GPU count and node count immutable and rejects DRA-backed GPU allocation.
- The shipped Helm configuration uses `failurePolicy: Fail` by default. Phase-1 cached accounting assumes this admission contract is enforced.
- Changing a cached power input requires deleting and recreating the DGD and restarting the Planner against the replacement deployment.

Power awareness is supported for Kubernetes disaggregated Planner deployments. Configuration validation requires `environment=kubernetes`, `mode=disagg`, and a positive `total_gpu_power_limit`.

### Budget and rollout behavior

- Deployment state carries the startup-cached per-GPU cap and per-replica watts for each required role.
- Initialization is strict and fail-closed, including a minimum-footprint check for `min_endpoint`.
- Runtime refresh does not re-read the power annotation. It uses one DGD GET and one DGD-scoped Pod LIST, dispatched off the async planning loop, to derive ready counts and deployment-wide rollout stability.
- The final projection applies the GPU budget first and the power budget second.
- The power budget is a ceiling: it may reduce a proposal below the GPU floor when necessary, never upscales decode, preserves partial-proposal provenance, and stages opposing rebalances so scale-up does not rely on unfinished scale-down capacity.
- While either power-relevant role is rolling, scale-ups are held at ready count. Baseline-filled peer targets are removed using the original proposal mask so an unrelated in-flight desired count is not overwritten.

The total budget is a projection over requested caps, not measured hardware draw. Effective-cap feedback and dynamic cap retargeting are outside this phase.

### Additional cleanup

This PR also removes unused diagnostic-projection helpers from `engine_adapter.py` and its module-local `__all__`. Repository-wide search found no call sites.

## Where should reviewers start?

1. Admission invariant:
   - `deploy/operator/internal/webhook/validation/dynamographdeployment.go`
   - `deploy/operator/internal/webhook/validation/dynamographdeployment_helpers.go`
   - `deploy/operator/internal/webhook/validation/dynamographdeployment_validation_envtest_test.go`
2. DGD parsing and startup lifecycle:
   - `components/src/dynamo/planner/monitoring/dgd_services.py`
   - `components/src/dynamo/planner/environment/base.py`
   - `components/src/dynamo/planner/connectors/kubernetes.py`
3. Final budget boundary and proposal provenance:
   - `components/src/dynamo/planner/core/budget.py`
   - `components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py`
   - `components/src/dynamo/planner/plugins/orchestrator/pipeline.py`
4. Behavioral coverage:
   - `components/src/dynamo/planner/tests/unit/test_power_budget.py`
   - `components/src/dynamo/planner/tests/unit/test_power_environment.py`
   - `components/src/dynamo/planner/tests/unit/test_kubernetes_connector.py`
   - `deploy/operator/internal/webhook/validation/dynamographdeployment_validation_envtest_test.go`

## Related issues

Supersedes the unmerged Planner-owned Pod-PATCH stack: #9683, #9684, and #9685.

## Validation

Current tip: `354a0240c9491d7953b97f59a11b20c1eb3377fb`.

Passing on the current tip:

- Pre-merge checks, pre-commit, CODEOWNERS, DCO, and CodeQL
- Operator Integration and the Operator image job
- Helm Chart Tests
- Power Agent
- Planner compliance on amd64 and arm64
- Fern configuration, links, and docs preview

The full CI run is still in progress. The Planner CPU-test jobs on amd64 and arm64 are currently failing in the parallelized CPU-test step; detailed logs are pending workflow completion and must be investigated or retried before merge.


kaim-eng:
Review nudge for CODEOWNER stamps on tip `57d2efdc9a` — https://github.com/ai-dynamo/dynamo/pull/12012

**What this PR does:** planner stops PATCHing Pod power caps; it reads `dynamo.nvidia.com/gpu-power-limit` from DGD worker podTemplates, projects watts, and clamps scale-up against `total_gpu_power_limit`. Enforcement stays with operator + Power Agent.

**Current state:** short CI green / near-green; full CI re-triggered via `/ok to test 57d2efdc9a`.

Please stamp your area when you have a chance:

- @ai-dynamo/dynamo-planner-codeowners — planner config/core/env/budget clamp, power tests, `examples/power-aware-budget/`
- @ai-dynamo/dynamo-operator-codeowners — power-agent + helm (`deploy/power-agent/`, planner/power-agent helm templates), k8s connector co-own
- @ai-dynamo/dynamo-observability-codeowners — `planner_metrics.py` co-own (power diagnostic metric surface)
- @ai-dynamo/dynamo-docs-codeowners — `docs/components/planner/planner-examples.md`
- @ai-dynamo/dynamo-ops-codeowners — `.github/codeowners/areas.yaml` ownership change for the new example path
- @ai-dynamo/dynamo-process-codeowners — generated `CODEOWNERS` regeneration

Happy to walk any of the above live if useful.

components/src/dynamo/planner/core/budget.py
```diff
@@ -215,78 +210,197 @@ def proportional_clamp_single(
...
+
+    eff_p = proposed_p if proposed_p is not None else current_p
+    eff_d = proposed_d if proposed_d is not None else current_d
+    if project_watts(eff_p, eff_d, p_watts, d_watts) <= total_budget:
+        return proposed_p, proposed_d, None
+
+    p_adjustable = proposed_p is not None and p_watts is not None and p_watts > 0
+    d_adjustable = proposed_d is not None and d_watts is not None and d_watts > 0
+
+    if p_adjustable and d_adjustable:
```
tedzhouhk, line 363: [P1] Do not credit an uncompleted scale-down when admitting a scale-up.

This branch validates only the final `(prefill, decode)` tuple. For example, current `(1, 4)` at 1000 W/replica with a 5000 W budget accepts proposed `(4, 1)`. `DisaggPlanner` then applies the component targets through separate Kubernetes/DGDSA updates (prefill first), so the deployment can pass through `(4, 4) = 8000 W` before decode reaches 1. Even issuing the down update first would not prove the old pods have terminated before the up begins, and the rollout guard only runs on the next tick.

Please stage power-constrained rebalances: emit/await the scale-downs first, then admit scale-up against the post-down stable counts (or provide equivalent atomic/serialized actuation). Add this `(1,4) -> (4,1)` case as a regression.

tedzhouhk (commented):
One blocking correctness issue in the power-budget actuation path.

components/src/dynamo/planner/config/planner_config.py
```diff
@@ -671,6 +727,24 @@ def _validate_ca_bundle_path(cls, v: Optional[str]) -> Optional[str]:
         ),
     )
 
+    @model_validator(mode="before")
```
tedzhouhk: [P1] Remove migration handling for a configuration API that never shipped.

None of `_REMOVED_POWER_WRITE_FIELDS` exists in this PR's base or in `main`; these names only came from the still-open, unmerged #9683–#9685 stack. There is therefore no released Planner config to migrate. Keeping a permanent validator, tests, and migration documentation for an unpublished implementation adds dead compatibility surface and incorrectly presents these fields as a former public contract. Please delete the field blacklist, this validator, and the corresponding tests/documentation.

components/src/dynamo/planner/environment/base.py
```diff
@@ -157,10 +165,29 @@ async def apply_scaling(
...
 
-    async def _refresh_deployment_state(self) -> None:
+    async def _refresh_deployment_state(self, *, is_initialization: bool) -> None:
         self._refresh_worker_info()
         self._refresh_gpu_counts()
-        await self._refresh_replica_counts()
+        self._refresh_power_configs(is_initialization=is_initialization)
+        try:
+            await self._refresh_replica_counts()
+        except Exception as exc:
```
tedzhouhk: [P1] Preserve fail-stop refresh behavior when power awareness is disabled.

This catch now affects every Planner. With power awareness off (the default), a replica-count failure retains stale counts but does not latch `power_scale_up_blocked`, so the tick continues and may issue scaling from stale inventory; previously the refresh failed and the tick could not act. Re-raise unless the power-specific conservative mode is enabled, and even there catch only expected API/transport failures rather than every `Exception`. The broadening in `_refresh_gpu_counts` has the same feature-disabled regression and should be gated/narrowed as well.

components/src/dynamo/planner/environment/base.py
```diff
@@ -249,6 +291,166 @@ def _refresh_gpu_counts(self) -> None:
         if self.require_decode:
             state.decode.num_gpus = decode_gpus
 
+    def _refresh_power_configs(self, *, is_initialization: bool) -> None:
```
tedzhouhk: [P1] Treat the DGD power annotation as static for the Planner lifetime.

There is no current requirement to support live annotation changes. This runtime state machine re-reads the cap every tick and adds last-good/max(old,new), sticky scale-up blocking, restart-required recovery, and blocked-state propagation across environment/state/capabilities/diagnostics. A transient GET failure or edit can permanently disable scale-up until restart. Read and validate the annotation once during initialization after `wait_for_deployment_ready`, cache `watts_per_replica`, and fail startup if it is unavailable or malformed. Annotation changes can take effect through the normal worker rollout plus Planner restart; please remove the runtime re-read/change-detection/sticky-block machinery.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
```diff
@@ -1049,11 +938,31 @@ def _project_scale_to(self, outcome, worker_counts: WorkerCounts):
         if p_unchanged and d_unchanged:
             return None
 
+        # Restore the proposal mask before the final budget boundary (see
+        # docstring). Ready-equal echoes are charged via ``current_*`` inside
+        # the clamps, never returned as adjustable targets.
+        if num_p is not None and num_p == current_p:
```
tedzhouhk: [P2] Do not change scale-projection semantics globally for a power-only feature.

This ready-equal masking runs even when `enable_power_awareness` is false, changing the emitted `ScalingDecision`/DGD write shape for every disaggregated Planner. If this behavior is required specifically by the power constraint, gate it behind power awareness. If it is a general Planner correctness fix, split it into a separate change with independent compatibility coverage.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
```diff
@@ -1112,16 +1173,91 @@ def clamp_single(component: str, replicas: Optional[int]) -> Optional[int]:
...
+                max(base_d, min_endpoint),
+                p_gpu,
+                d_gpu,
+                min_gpus,
+                max_gpus,
+                min_endpoint,
+            )
+            return clamped_p, clamped_d
+
+        if proposed_p != proposed_d:
```
tedzhouhk: [P2] Keep the existing GPU-budget policy out of this power PR.

This fixed-peer/residual-budget path rewrites partial-proposal handling for all disaggregated deployments using GPU budgets, regardless of whether power awareness is enabled. That substantially expands the regression surface beyond reading a DGD power constraint and clamping scale-up. Preserve the existing GPU-budget behavior here, or split this general GPU-budget correction into a focused PR with its own behavioral review.

components/src/dynamo/planner/monitoring/dgd_services.py
```diff
@@ -215,8 +306,99 @@ def get_component_from_type_or_name(
         if not _can_use_explicit_component_name(component, component_type):
             raise SubComponentNotFoundError(component_type.value)
         matching_components.append((component_name, component))
+    elif not matching_components and component_type == SubComponentType.DECODE:
```
tedzhouhk: [P2] Keep generic `type: worker` resolution local to the power parser.

`get_component_from_type_or_name` is shared by model-name lookup, GPU-count discovery, replica inspection, and scaling. Adding this fallback here therefore changes component selection for callers unrelated to power. Resolve the aggregate worker inside `resolve_component_power_configs` (or a dedicated helper) so the new power feature does not alter the semantics of the shared resolver.

components/src/dynamo/planner/connectors/base.py
```diff
@@ -55,6 +56,22 @@ def get_gpu_counts(
     ) -> tuple[Optional[int], Optional[int]]:
         pass
 
+    def get_component_power_configs(
```
tedzhouhk, line 88: [P2] Do not put a Kubernetes-only power capability on every Planner connector.

`PlannerConfig` already restricts power awareness to `environment="kubernetes"`, so Virtual and Global connectors have no authoritative DGD and should not need this method or rejecting stubs. Use a narrow `PowerConfigProvider` capability (or a Kubernetes-specific dependency in the environment) instead of expanding the common `PlannerConnector` protocol for a feature that cannot apply to most implementations.

tedzhouhk (commented):
Additional planner findings: three blocking scope/correctness issues and four non-blocking cross-cutting behavior changes.

components/src/dynamo/planner/core/budget.py
```diff
@@ -215,78 +210,245 @@ def proportional_clamp_single(
...
+    (unproposed) component alone leaves no room to even seat ``min_endpoint``,
+    so the proposed scale-up is refused (held at ``min(proposed, current)``)
+    rather than the unproposed component being silently mutated.
+    """
+    if avail < min_endpoint * watts:
+        held, capped = _hold_at_current(proposed, current)
+        # ``suppressed`` tracks whether the proposal was actually held back.
+        # When ``current`` is None, ``_hold_at_current`` leaves the proposal
+        # unchanged and ``capped`` is False — do not claim a suppression.
+        return (held if held is not None else min_endpoint), capped
```
dmitry-tokarev-nv: **Nit (dead branch):** the `else min_endpoint` fallback here is unreachable.

`proposed` is typed `int` (non-optional), and the sole caller — `apply_power_budget`'s `p_adjustable != d_adjustable` branch — only reaches `_shrink_single` when the proposed count is not `None`. `_hold_at_current` returns a `None` first element *only* when `proposed is None`, so `held` is never `None` here and the `else min_endpoint` can't be taken.

```python
return (held if held is not None else min_endpoint), capped
```

can just be `return held, capped`. Harmless as written, but the guard implies a `None` case the signature already rules out — worth dropping so the contract reads clearly.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
```diff
@@ -1051,9 +1052,21 @@ def _project_scale_to(self, outcome, worker_counts: WorkerCounts):
...
+        # rolling role's transient ready count would overwrite its larger
+        # in-flight desired and cancel the rollout — so a scale-down of one role
+        # must not drag the other role's echoed ready count along, and a
+        # scale-up held at ready (see ``_hold_scale_up_during_rollout``)
+        # collapses to "no change" instead of re-emitting ready.
+        if num_p is not None and num_p == current_p:
+            num_p = None
+        if num_d is not None and num_d == current_d:
+            num_d = None
+        if num_p is None and num_d is None:
```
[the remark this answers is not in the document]
kaim-eng: Addressed. Ready-equal masking is now gated behind `enable_power_awareness` (see `engine_adapter.py` `_project_scale_to`); the non-power path keeps the original both-unchanged check. Landed in `85841f0e8e`.

components/src/dynamo/planner/monitoring/dgd_services.py
```diff
@@ -215,8 +306,99 @@ def get_component_from_type_or_name(
...
+            if get_component_type(curr_component)
+            == V1BETA1_GENERIC_WORKER_COMPONENT_TYPE
+        ]
+        if len(generic_workers) == 1:
+            matching_components.append(generic_workers[0])
+        elif len(generic_workers) > 1:
+            component_names = [name for name, _ in generic_workers]
+            raise DuplicateSubComponentError(component_type.value, component_names)
+        else:
+            raise SubComponentNotFoundError(component_type.value)
```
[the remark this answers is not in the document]
kaim-eng: Addressed. The shared `get_component_from_type_or_name` is back to its prior semantics. The generic `type: worker` fallback now lives only inside `_resolve_one_power_config`. Landed in `91770a3fe9`.

components/src/dynamo/planner/monitoring/dgd_services.py
```diff
@@ -174,12 +193,84 @@ def get_gpu_count(self) -> int:
...
+            self.service.get("podTemplate", {}).get("metadata", {}).get("annotations")
+            or {}
+        )
+        raw = annotations.get(POWER_ANNOTATION_KEY)
+        if raw is None:
+            raise PowerAnnotationMissingError(self.name)
+        try:
+            watts = int(str(raw).strip())
+        except (ValueError, TypeError):
+            raise PowerAnnotationInvalidError(self.name, str(raw))
```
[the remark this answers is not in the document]
kaim-eng: Addressed. Watts parse now chains with `except (ValueError, TypeError) as err: raise PowerAnnotationInvalidError(...) from err`. Landed in `61c12e2a36`.

components/src/dynamo/planner/tests/integration/test_power_no_mutation.py
```diff
@@ -0,0 +1,149 @@
...
+        ),
+        decode=EngineCapabilities(
+            num_gpu=4, power_watts_per_replica=decode_cfg.watts_per_replica
+        ),
+        power_scale_up_blocked=False,
+    )
+    adapter = _bare_adapter(caps)
+    wc = WorkerCounts(ready_num_prefill=2, ready_num_decode=2)
+    new_p, new_d = adapter._apply_final_budget(4, 4, wc)  # 4*700 + 4*1200 = 7600 > 5000
+    assert new_p * 700 + new_d * 1200 <= 5000  # clamped to fit
```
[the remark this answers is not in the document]
kaim-eng: Addressed. Step 3 now sets `expected_num_prefill/decode` equal to ready so the deployment is stable and the proportional power clamp (not the rollout hold) is what constrains `(4,4)` under 5000 W.

examples/power-aware-budget/README.md
```diff
@@ -0,0 +1,82 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# DGD-owned power caps and budget-aware scaling
```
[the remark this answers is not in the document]
kaim-eng: Rejecting this one. Per `docs/documentation-style-guide.md` §Examples and recipes and `docs/AGENTS.md`, example READMEs use the HTML-comment SPDX form with a normal `#` H1 (no Fern frontmatter). Fern frontmatter / no-body-H1 rules apply to pages under `docs/`, not `examples/`. Sibling examples (`examples/global_planner/README.md`, etc.) follow the same pattern. Pre-merge Fern checks are already green on this tip.

examples/power-aware-budget/README.md
```diff
@@ -0,0 +1,82 @@
...
+budget so the Planner keeps its **projected** power draw within a configured
+rack/DGD budget when it scales replicas.
+
+> [!NOTE]
+> These manifests are a **contract fragment**, not a fully deployable DGD: they
+> show only the power-relevant fields (annotations, GPU limits, Planner config).
+> A runnable deployment also needs container images, model args, env/secrets,
+> and a Frontend — see [examples/backends/vllm/deploy/](../backends/vllm/deploy/)
+> for a complete DGD to merge these fields into.
+
```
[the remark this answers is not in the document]
kaim-eng: Rejecting this one. The blank line between `[!NOTE]` and `[!IMPORTANT]` is intentional so GitHub/Fern keep them as two separate callouts. Replacing it with a lone `>` separator merges them into one blockquote and breaks distinct alert rendering. This is the standard GFM pattern used elsewhere in the repo.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
```diff
@@ -1063,6 +1076,147 @@ def _apply_final_budget(
...
+        d_watts = d_caps.power_watts_per_replica if d_caps else None
+
+        ready_p = worker_counts.ready_num_prefill
+        ready_d = worker_counts.ready_num_decode
+
+        num_p, num_d = self._hold_scale_up_during_rollout(
+            num_p, num_d, ready_p, ready_d, p_watts, d_watts, worker_counts
+        )
+
+        new_p, new_d, reason = apply_power_budget(
```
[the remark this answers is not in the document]
kaim-eng: Addressed. Proposal-mask preservation before the power clamp is in place: roles whose merged target equals ready are passed as `None` into `apply_power_budget`, so a one-role proposal cannot drag the peer through `_shrink_pair`. Coverage restored/extended via `test_project_scale_to_budget_preserves_single_component_target_mask` on tip `0e7d6eb663`.

components/src/dynamo/planner/core/budget.py
kaim-eng: Addressed. `apply_power_budget` now detects opposing rebalances (`_is_opposing_rebalance`) and stages via `peak_parallel_watts`: if the transient peak (e.g. `(1,4)->(4,1)` peaking at `(4,4)`) would exceed budget, scale-up legs are held at current and the reason is `power_rebalance_staged`. Scale-down emits this tick; scale-up is admitted after settle. Regression coverage for the `(1,4)->(4,1)` case is in `test_power_budget.py`. Landed in `8508f84df8`.

components/src/dynamo/planner/config/planner_config.py
kaim-eng: Addressed. `_REMOVED_POWER_WRITE_FIELDS`, the validator, and the corresponding tests/docs are gone. The PR adds only `enable_power_awareness` + `total_gpu_power_limit`. Landed in `85841f0e8e`.

components/src/dynamo/planner/environment/base.py
kaim-eng: Addressed. Broad `except Exception` on replica/GPU refresh is gone; fail-stop behavior when power awareness is off matches `origin/main`. Power-gated startup load is the only place that retains a power-specific failure path. Landed in `91770a3fe9`.

components/src/dynamo/planner/environment/base.py
kaim-eng: Addressed. Caps are read once in `_load_static_power_caps_at_startup()` after `wait_for_deployment_ready`; no per-tick re-read, no sticky `power_scale_up_blocked`, no change-detection machinery. Cap changes require worker rollout + Planner restart.

Follow-up on tip `0e7d6eb663`: because `num_gpus` still refreshes every tick while `power_watts_per_replica` is startup-cached, we now fail closed if per-pod GPUs diverge from the startup snapshot (`_assert_power_topology_static`), requiring a restart rather than projecting against stale wattage. That keeps the static-cap contract intact without reintroducing runtime cap re-reads.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
kaim-eng: Addressed. Ready-equal masking is gated behind `enable_power_awareness`; the else branch preserves the original both-unchanged check for non-power disagg. Landed in `85841f0e8e`.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
kaim-eng: Addressed. Partial residual / fixed-peer GPU-budget rewrite was reverted; disagg GPU clamping is back to the `origin/main` joint-clamp behavior. Proposal-mask invariant coverage restored via `test_project_scale_to_budget_preserves_single_component_target_mask`. Landed in `85841f0e8e` / `61c12e2a36`.

components/src/dynamo/planner/monitoring/dgd_services.py
kaim-eng: Addressed. Generic `type: worker` fallback removed from the shared resolver and scoped inside `_resolve_one_power_config` only. Landed in `91770a3fe9`.

components/src/dynamo/planner/connectors/base.py
kaim-eng: Addressed. `get_component_power_configs` is no longer on `PlannerConnector`. The environment uses `getattr(..., None)` and fails closed if the connector lacks it; only the Kubernetes connector implements it. Landed in `91770a3fe9`.

components/src/dynamo/planner/core/budget.py
kaim-eng: Addressed. Dropped the unreachable `else min_endpoint` fallback and asserted `held is not None` so the non-optional `proposed` contract is explicit for mypy. Landed in `d82fc0f2a6`.

kaim-eng:
All 16 review threads on tip `0e7d6eb663` now have replies and are resolved:

- Ted's P1/P2 items — fixed and pushed (staged rebalance, init-only static caps, fail-stop refresh, scope trims, plus a fail-closed runtime topology guard so stale `power_watts_per_replica` cannot under-count after a GPUs-per-replica change)
- Dmitry's dead-branch nit — fixed
- CodeRabbit B904 / clamp-test — fixed; README frontmatter + MD028 intentionally rejected per example-README style guide (Fern frontmatter rules are for `docs/`, not `examples/`)
- Devin / review-agent notes — fixed (mask gating + localized generic-worker fallback)

Short CI is green; merge is still blocked only on `REVIEW_REQUIRED`.

@tedzhouhk — when you have a chance, could you re-review the tip and stamp if the earlier findings look closed?
@ai-dynamo/dynamo-operator-codeowners @ai-dynamo/dynamo-observability-codeowners — still showing as requested CODEOWNER teams; a stamp on your areas would unblock merge.

components/src/dynamo/planner/core/budget.py
tedzhouhk: Reopening: the fix only stages a proposal that is already an opposing rebalance before clamping. The clamp can synthesize a new opposing rebalance afterward. On current tip `5fe990dd57`, with stable current `(4, 1)`, proposal `(5, 5)`, 1000 W/replica, and a 5000 W budget, `apply_power_budget` returns `(2, 3, "power_budget_clamped")`; the settled target fits, but separate actuation can peak at `(4, 3) = 7000 W`. The final `OrchestratorEngineAdapter` path emits `ScalingDecision(2, 3)`, so this also turns a pure scale-up proposal into an unrequested scale-down/scale-up rebalance. Please run peak/staging against the final clamped target (or only admit scale-up increments against current residual headroom) and add this case as a regression.

components/src/dynamo/planner/environment/base.py
tedzhouhk: Reopening: later commits reintroduced the runtime annotation check. On current tip `5fe990dd57`, every `refresh()` calls `_assert_power_config_static()`, which calls `_resolve_power_configs()` and compares a startup fingerprint containing both `(gpus_per_replica, gpu_power_limit_watts)`. That is a per-tick DGD annotation re-read and change-detection path, despite the static-configuration contract and the earlier reply saying there is no per-tick re-read. Please remove `_power_startup_fingerprint`, `_assert_power_config_static()`, and the runtime drift tests; read and validate the annotation once during initialization, with changes taking effect only after worker rollout plus Planner restart.

components/src/dynamo/planner/core/budget.py
kaim-eng: Addressed on current tip `84a831686d`. `apply_power_budget` now checks `_is_opposing_rebalance` and `peak_parallel_watts` before accepting a settled-fit tuple. For the exact current `(1,4)` → proposed `(4,1)` case at 1000 W/replica with a 5000 W budget, it holds the prefill scale-up and emits `(1,1,power_rebalance_staged)`, allowing decode to scale down first; prefill can scale up only after the deployment settles. The exact regression is `test_apply_power_budget_rebalance_stages_scale_down_first`. We also added post-clamp staging for rebalances synthesized by `_shrink_pair`. @tedzhouhk, could you please resolve this thread if the implementation now satisfies the P1?

components/src/dynamo/planner/environment/base.py
kaim-eng: Addressed on current tip `84a831686d`. The runtime power fingerprint/re-read machinery (`_power_startup_fingerprint`, `_assert_power_config_static`, and runtime drift tests) has been removed. DGD-owned caps are resolved and validated once in `_load_static_power_caps_at_startup()` after `wait_for_deployment_ready`; `refresh()` never re-resolves or drift-checks the power annotation. Changes take effect through worker rollout plus Planner restart. The replacement regression `test_refresh_does_not_reread_power_annotation` asserts that refresh does not call `get_component_power_configs` and retains the startup-cached watts. @tedzhouhk, could you please resolve this now-outdated thread if the static-cap behavior matches your request?

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
tedzhouhk: Reopening: commit `fffa5265c2` reintroduced the fixed-peer/residual GPU-budget policy this thread asked to keep out of the power PR. `_apply_gpu_final_budget` now takes the residual path whenever exactly one disaggregated role is proposed, with no `enable_power_awareness` guard. This changes power-off behavior too: with current `(7,1)`, a decode-only proposal of `7`, one GPU/replica, and `max_gpu_budget=8`, the new code holds decode at `1` and emits no decision, whereas the prior joint-clamp path emitted decode `4`. The residual policy may be a valid general GPU-budget correctness fix, but it is unrelated scope and needs its own behavioral review. Please remove it from this PR (or limit any special handling here strictly to the power-induced proposal-mask path) and land the general correction separately.

kaim-eng: Accepted — residual sizing is now gated behind `enable_power_awareness` on tip `d0efb223f5`. Power-off disagg keeps the historical joint-then-discard path: current `(7,1)`, decode-only proposal `7`, ceiling `8` still emits `(None, 4)`. The residual path only runs on the power-induced proposal-mask case (ready-equal peer collapsed to `None` before the GPU clamp), which is what made the over-ceiling applied state reachable in this PR. Added `test_power_off_partial_proposal_keeps_joint_gpu_clamp` as the scope guard; a general residual GPU-budget correction stays out of this PR. @tedzhouhk, please resolve if this matches the requested power-path-only limit.

kaim-eng:
**codeowners red on tip `d0efb223f5`:** not from the power-budget diff. Our legitimate `examples/power-aware-budget/` ownership edit flips the gate to full-tree strict, which then fails on unowned `components/src/dynamo/squeeze_evolve/**` already on `main` (landed in #10785).

Deferring a local workaround — asked @Nietzsche2000 on #10785 to cover that tree on `main` so every policy-touching PR stops failing for the same reason: https://github.com/ai-dynamo/dynamo/pull/10785#issuecomment-5070524518

kaim-eng:
@tedzhouhk — all three of your re-opened findings are addressed on current tip \`84a831686d\`. Quick summary of what landed:

- **budget.py P1 (re-open):** Post-clamp opposing rebalance is now staged too. After `_shrink_pair`, the result is checked via `_is_opposing_rebalance` + `peak_parallel_watts`; if the clamped target would still peak over budget during parallel rollout, the scale-up leg is held at current. Your exact case (`current (4,1)`, proposal `(5,5)`, 1000 W/replica, 5000 W budget) now returns `(2, 1, "power_rebalance_staged")` — covered in `test_apply_power_budget_clamp_synthesized_rebalance_is_staged`.

- **environment/base.py P1 (re-open):** Runtime fingerprint machinery (`_power_startup_fingerprint`, `_assert_power_config_static`) is gone. DGD annotation is read exactly once in `_load_static_power_caps_at_startup` after `wait_for_deployment_ready`; `refresh()` never re-resolves or drift-checks. Covered in `test_refresh_does_not_reread_power_annotation`.

- **engine_adapter.py P2 (re-open):** Residual GPU-budget path is now gated behind `enable_power_awareness`; power-off disagg falls through to the historical joint-clamp. `test_power_off_partial_proposal_keeps_joint_gpu_clamp` is the scope guard.

Could you re-review the tip and stamp if these match what you asked for? Happy to walk any of them live if useful.

components/src/dynamo/planner/environment/base.py
kaim-eng: Addressed in `483b3fb25b`.

Removed the startup power fingerprint state and all runtime drift checks. `refresh()` no longer re-resolves the DGD power annotation or performs a per-tick DGD GET.

Power caps are now resolved and validated once from the settled startup snapshot, then remain static for the Planner lifetime. Annotation or GPU-topology changes take effect through the worker rollout followed by a Planner restart.

I also replaced the runtime-drift tests with `test_refresh_does_not_reread_power_annotation`, which explicitly verifies that `refresh()` does not call `get_component_power_configs` and retains the startup-cached watts.

Ted, could you please take another look and resolve this thread if it now matches the intended contract?

kaim-eng:
C:/Program Files/Git/ok to test 483b3fb25bedd6faf7a361c25d233900fbad59ef

components/src/dynamo/planner/core/budget.py
kaim-eng: Reverified on exact head `483b3fb25b`.

Both unsafe actuation shapes are now covered at the final target boundary:

- A proposed opposing rebalance `(1,4) -> (4,1)` is checked with `peak_parallel_watts` before settled-fit acceptance and stages to `(1,1,power_rebalance_staged)`, so only the decode scale-down is emitted.
- A clamp-synthesized rebalance from current `(4,1)`, proposal `(5,5)` is checked again after `_shrink_pair`; it stages the clamped `(2,3)` target to `(2,1,power_rebalance_staged)`, keeping the parallel peak at 5000 W rather than 7000 W.

The exact regressions are `test_apply_power_budget_rebalance_stages_scale_down_first` and `test_apply_power_budget_clamp_synthesized_rebalance_is_staged`. Planner CPU and compliance CI are green on amd64 and arm64. Resolving as addressed.

components/src/dynamo/planner/environment/base.py
kaim-eng: Reverified on exact head `483b3fb25b`.

Initialization waits for a settled DGD snapshot, refreshes deployment state from that snapshot, and calls `_load_static_power_caps_at_startup(deployment=deployment)` once. `_resolve_power_configs()` is documented and used only by that startup path. The `_power_startup_fingerprint` and `_assert_power_config_static()` runtime machinery are absent, and `refresh()` does not resolve power configs.

`test_refresh_does_not_reread_power_annotation` loads the startup caps, resets the resolver mock, executes `refresh()`, asserts `get_component_power_configs` was not called, and verifies the cached watts remain unchanged. Planner CPU and compliance CI are green on amd64 and arm64. Resolving as addressed.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
kaim-eng: Reverified on exact head `483b3fb25b`.

The fixed-peer/residual GPU-budget branch is now entered only under `if self._config.enable_power_awareness`. With power awareness disabled, partial disaggregated proposals continue through the historical joint proportional clamp and then discard the unproposed peer; the general GPU-budget policy is unchanged by this PR.

`test_power_off_partial_proposal_keeps_joint_gpu_clamp` uses the exact scope-guard case from this thread: current `(7,1)`, decode-only proposal `7`, one GPU per replica, and `max_gpu_budget=8` still emits `(None,4)`. Planner CPU and compliance CI are green on amd64 and arm64. Resolving as addressed.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
```diff
@@ -288,14 +328,279 @@ def get_service_replica_status(
...
+        component_name: str,
+        component_spec: dict,
+        component_names: list[str],
+    ) -> tuple[bool, str]:
+        # Prefer status.components[*].componentNames (same source Grove uses).
+        # During a rolling update that list includes both old and new DCDs, so
+        # accepting "first ready hash from current-worker-hash" would wrongly
+        # settle on the still-annotated old revision.
+        if component_names:
+            for dcd_name in component_names:
```
tedzhouhk: [P1] `status.components[*].componentNames` cannot be used as DCD names here. The operator API defines this field as the names of the underlying workloads (`Deployment`, `LeaderWorkerSet`, `PodClique`, etc.); for a normal DCD-backed worker the entry is typically `demo-vllmdecodeworker-deployment` while the DCD is `demo-vllmdecodeworker`. This loop therefore GETs a nonexistent DCD and the power-aware Planner waits 30 minutes before startup times out. The new tests mask this by putting DCD names into `componentNames` while declaring `componentKind: Deployment`. Please use `componentNames` only for the Grove kinds; for Deployment/LWS derive the DCD name from the DGD/component/current worker hash (or drop the backing-CR settlement), and add a regression test with the real workload name shape.

tedzhouhk (approved):
Approving as Planner code owner; please address the inline P1 before merge.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
kaim-eng: Addressed in two follow-ups.

**`55e7faa516` — stop treating `componentNames` as DCD names for Deployment/LWS.** `_is_dcd_backing_ready` no longer accepts `component_names`; for Deployment/LWS it always derives the DCD name from the DGD + component + worker-hash candidates (same shape the operator uses). `componentNames` is still used only for Grove kinds (`PodClique` / `PodCliqueScalingGroup`), where those entries really are the backing CR names. The rolling-update phase guard is unchanged and still runs before DCD settlement.

Tests now use the real workload name shape (`…-deployment` in `componentNames`) and assert that only the hash-derived DCD is queried — including an explicit regression that records every DCD GET and requires the workload name not appear.

**`52dd3edd55` — dual-hash probe order matches the operator bridge.** With both annotations present, `_worker_hash_candidates` now probes v1 before v2 (mirroring `workerHashForDCDGeneration` / `activeWorkerHashCandidates`), so a stale-but-ready v2 DCD cannot short-circuit settlement while the active v1 DCD still lags. Regression: both DCDs present, v2 ready / v1 lagging → settlement rejects and names the v1 DCD.

deploy/helm/charts/platform/components/operator/templates/planner.yaml
```diff
@@ -39,6 +39,9 @@ rules:
 - apiGroups: ["nvidia.com"]
   resources: ["dynamocomponentdeployments", "dynamographdeployments"]
   verbs: ["get", "list", "create", "update", "patch"]
+- apiGroups: ["grove.io"]
+  resources: ["podcliques", "podcliquescalinggroups"]
+  verbs: ["get", "list"]
```
sttts: this seems wrong. Planner becoming dependent on the scheduling primitives breaks the layering.

We have two levels of permissible data sourcing:
- DGDs – for intent level information
- Pods – for low-level status.

Our pods are labelled, and those labels are considered a stable interface. If planner want to reason about the actual state of deployments, it should look at pods.

sttts: Compare @julienmancuso's https://github.com/ai-dynamo/dynamo/pull/11909/changes#diff-126954d7a81ef5503ce05f5ab44c2d78bf45aef6c28973cb0f70692cf30e585bR868 where pods are checked for being terminated. In your case, you can check power settings on pods.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
```diff
@@ -288,14 +328,277 @@ def get_service_replica_status(
...
+        component_name: str,
+        worker_suffix: str,
+        component_spec: dict,
+    ) -> str:
+        base = f"{dgd_name}-{component_name.lower()}"
+        if (
+            worker_suffix
+            and get_component_type(component_spec) in WORKER_COMPONENT_TYPES
+        ):
+            return f"{base}-{worker_suffix}"
```
sttts: [P1] Planner must work with every workload backend.

Supported Planner interfaces exist for exactly this reason:

- DGD: desired deployment intent.
- DGDSA `/scale`: replica actuation.
- Labelled Pods: actual applied state.

The intermediate workload objects are operator-internal:

- Deployment and LeaderWorkerSet reach `_is_dcd_backing_ready()` and depend on the reconstructed DCD name `<dgd>-<component>-<worker-hash>`.
- Grove reaches its intermediate CRs directly.
- Adding, replacing, or changing a workload backend would therefore require Planner changes.

**This freezes the operator implementation: workload backends, DCD naming, and intermediate APIs could no longer evolve without coordinated Planner changes.**

See the [previous layering comment](https://github.com/ai-dynamo/dynamo/pull/12012#discussion_r3663954921). Please remove DCD/Grove discovery and verify the desired power annotation on relevant non-terminal Pods.

sttts (commented):
Concrete layering violation at the DCD name reconstruction site.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
```diff
@@ -288,14 +328,277 @@ def get_service_replica_status(
...
+
+        During a v1/v2 compatibility generation both annotations are present and
+        the operator keeps the v1 DCD active, probing it first
+        (``activeWorkerHashCandidates`` -> ``workerHashForDCDGeneration``). Probe
+        v1 first too: a stale-but-ready v2 DCD must not short-circuit
+        settlement while the active v1 DCD still lags.
+        """
+        annotations = deployment.get("metadata", {}).get("annotations") or {}
+        candidates: list[str] = []
+        for key in (WORKER_HASH_V1_ANNOTATION, WORKER_HASH_V2_ANNOTATION):
```
sttts: [P1] Worker hashes are operator-internal and must not be exported to Planner.

- Reads `current-worker-hash` and `current-worker-hash-v2`.
- Mirrors the operator's v1-before-v2 migration precedence.
- Makes an internal rollout representation a Planner API contract.
- Prevents the operator from changing or removing the hash scheme independently.

Planner does not need hashes:

- Read desired power intent from the DGD.
- Read applied power state from labelled, non-terminal Pods.

See the [layering comment](https://github.com/ai-dynamo/dynamo/pull/12012#discussion_r3663954921).

sttts (commented):
The worker-hash abstraction is operator-internal and cannot become a Planner contract.

deploy/helm/charts/platform/components/operator/templates/planner.yaml
kaim-eng: Addressed in `815ea9222c`.

The Planner no longer reads DCDs, Grove CRs, or worker-hash annotations. The only startup settlement check is a `pods/list` against the two stable, published pod labels (`nvidia.com/dynamo-graph-deployment-name` and `nvidia.com/dynamo-component`), followed by verifying that every non-terminal pod carries the raw `dynamo.nvidia.com/gpu-power-limit` annotation verbatim from the DGD snapshot. The power cap itself is still read from the DGD `podTemplate.metadata.annotations` (intent layer). Pods are checked only to confirm the operator propagated that intent before the Planner caches the cap at startup.

RBAC changes: `grove.io` rule removed; `dynamocomponentdeployments` removed from the nvidia.com rule; `pods` list-only added. No pod write surface is introduced. Could you resolve this thread if this matches what you had in mind?

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
kaim-eng: Addressed in `815ea9222c`. Settlement now selects worker pods by the stable `nvidia.com/dynamo-graph-deployment-name` + `nvidia.com/dynamo-component` labels and checks the `dynamo.nvidia.com/gpu-power-limit` annotation on every non-terminal pod — no DCD, no PodClique, no PodCliqueScalingGroup. This path works for Deployment, LeaderWorkerSet, or any future Grove-based backend identically, because all of them produce pods with these labels. Could you resolve this thread?

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
kaim-eng: Addressed in `815ea9222c`. No worker-hash annotation (`current-worker-hash`, `current-worker-hash-v2`) is read anywhere in the power path. DCD name reconstruction (`<dgd>-<component>-<hash>`) is gone entirely. The pod-annotation convergence check uses only the two stable published pod labels and the DGD annotation key. Could you resolve this thread?

kaim-eng:
C:/Program Files/Git/ok to test 815ea9222c

kaim-eng:
C:/Program Files/Git/ok to test 2d18408226

kaim-eng:
Rebased tip is `13cd948f0c`; the pod-settlement fix is now in `aa09376da7` with cleanup in `34f0f98750`. All three threads remain addressed: power intent comes from the DGD, applied state comes from labeled non-terminal Pods, and the power path has no DCD, Grove, or worker-hash dependency. Focused WSL Helm RBAC tests pass 4/4. Resolving the addressed threads now.

components/src/dynamo/planner/monitoring/dgd_services.py
```diff
@@ -220,3 +331,146 @@ def get_component_from_type_or_name(
...
+        if len(generic_workers) == 1:
+            name, component = generic_workers[0]
+            return Service(name=name, service=component)
+        if len(generic_workers) > 1:
+            component_names = [name for name, _ in generic_workers]
+            raise DuplicateSubComponentError(sub_component_type.value, component_names)
+        raise
+
+
+def _resolve_one_power_config(
```
sttts, line 402: am worried this is the wrong semantics. Why is planner looking into the deployment to get power settings, why is that not on the actual pods?

sttts: After tracing how these values are used, I think the current semantics are wrong. Planner does not apply the GPU power cap; it only uses it to account projected replica cost against the total power budget. The accounting should therefore distinguish actual state from desired future state:

- Existing committed budget: sum the power annotation and allocated GPU count of every labelled, non-terminal Pod.
- Additional scale-up budget: use the current DGD's power limit and GPU topology for replicas which do not exist yet.
- Scale-down: do not credit the released budget until the corresponding Pods are actually terminal/gone.

Conceptually:

```
projected power = power of existing Pods + power of additional replicas from DGD intent
```

The current implementation instead caches `gpu_power_limit_watts` and `gpus_per_replica` from the current DGD and applies that homogeneous cost to both current and projected replica counts. That retroactively treats every existing Pod as if it already matched the current DGD. The Pod settlement check only proves that the power annotation matches; it does not prove that an old Pod has the current GPU count or `nodeCount` topology. An old revision with the same annotation but a different GPU topology can therefore pass settlement while the calculated budget is wrong.

Summing existing Pods also needs no Grove/LWS/Deployment knowledge or replica grouping: the stable DGD/component labels, Pod annotation, and Pod GPU resources are sufficient. The DGD remains the correct source only for the cost of future replicas. Please change the accounting semantics accordingly.

julienmancuso (commented):
Review findings: one P1 correctness issue and three P2 design/robustness issues. The P1 is related to the open actual-vs-desired accounting discussion, but covers the remaining startup-cache failure after a rollout has fully settled.

components/src/dynamo/planner/environment/base.py
```diff
@@ -249,6 +334,131 @@ def _refresh_gpu_counts(self) -> None:
         if self.require_decode:
             state.decode.num_gpus = decode_gpus
 
+    def _load_static_power_caps_at_startup(
+        self, deployment: Optional[dict] = None
+    ) -> None:
+        """Read DGD-owned per-GPU caps once at startup; fail closed if bad.
```
julienmancuso, line 341: [P1] Refresh or invalidate cached power costs after a DGD change

This cache is permanent for the Planner lifetime. For example, suppose Planner starts with a 250 W cap and 4 GPUs per replica (1000 W/replica) under a 2500 W budget. If the DGD changes to 400 W and the workers roll without restarting Planner, then after the rollout is fully settled Planner still admits two replicas as 2000 W even though their requested draw is 3200 W. The startup settlement check only proves consistency of the initial snapshot; nothing detects later DGD generation/config drift or prevents scale-up with stale costs. This is related to the actual-vs-desired accounting thread, but remains after all old Pods are gone unless future-replica cost is read from the current DGD. Re-read/adopt the cost after convergence, or retain a generation/config fingerprint and fail closed on drift until it is reconciled.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
```diff
@@ -288,14 +325,142 @@ def get_service_replica_status(
...
+        )
+
+    def _list_pods_for_component(self, dgd_name: str, component_name: str) -> list:
+        """List all pods (any phase) for the given DGD + component label pair."""
+        label_selector = (
+            f"{DYNAMO_DGD_NAME_LABEL}={dgd_name},"
+            f"{DYNAMO_COMPONENT_LABEL}={component_name}"
+        )
+        return (
+            self.core_api.list_namespaced_pod(
```
julienmancuso, line 361: [P2] Avoid synchronous Pod LISTs on every planning tick

Every power-aware refresh calls `get_actual_worker_counts(check_terminating_pods=True)`. For a stable disaggregated deployment that reaches this synchronous Kubernetes client call once for prefill and once for decode, on the async Planner loop. A slow API server can therefore stall metric collection and planning, while every Planner adds two LIST requests per tick. At minimum, list once by the DGD label and partition the result by component; preferably use an async client/watch-backed cache or offload the blocking I/O.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
```diff
@@ -1049,12 +936,42 @@ def _project_scale_to(self, outcome, worker_counts: WorkerCounts):
         if p_unchanged and d_unchanged:
             return None
 
+        if self._config.enable_power_awareness:
+            # Restore the proposal mask before the final budget boundary (see
+            # docstring). Ready-equal echoes are charged via ``current_*`` inside
+            # the clamps, never returned as adjustable targets.
+            if num_p is not None and num_p == current_p:
```
julienmancuso: [P2] Preserve proposal provenance instead of inferring it from equality

`target == ready` does not prove that the target is a merge-generated baseline echo. During a rollout, an explicit request to set a role to its transient ready count is indistinguishable here from an echo and is silently converted to `None`; if the peer is also changing, the peer update is applied while this explicit cancel/scale-down intent is lost. A target can also coincidentally equal ready in stable operation. Carry the originally proposed component mask through the merge/projection pipeline and use that mask at the budget boundary instead of reconstructing provenance from values.

components/src/dynamo/planner/connectors/base.py
```diff
@@ -68,4 +72,75 @@ async def set_component_replicas(
...
+
+
+_POWER_AWARE_REQUIRED: tuple[str, ...] = (
+    "get_graph_deployment",
+    "get_component_power_configs",
+    "wait_for_settled_graph_deployment",
+)
+
+
+def is_power_aware_connector(obj: object) -> TypeGuard[PowerAwareConnector]:
```
julienmancuso, line 125: [P2] Make this TypeGuard establish the full protocol

`PowerAwareConnector` has four required members, including `get_actual_worker_counts(..., check_terminating_pods=...)`, but `_POWER_AWARE_REQUIRED` checks only the other three. A connector can therefore pass this guard while exposing only the older `PlannerConnector.get_actual_worker_counts` signature; the power-aware refresh then passes `check_terminating_pods=True` and fails with `TypeError`. A `TypeGuard` must establish the complete predicate it claims. Include the runtime-count capability/signature in the check, move the keyword into the base connector contract, or replace this structural guard with an explicit capability adapter.

components/src/dynamo/planner/monitoring/dgd_services.py
kaim-eng: The DGD is the source of power intent for each component: the per-GPU cap and GPU topology used to project the cost of replicas that do not exist yet. The operator propagates that intent to Pods, the Power Agent enforces the cap, and the Planner uses it for scaling-budget calculations.

At startup, the Planner also lists Pods to verify that the DGD annotation has propagated and the worker rollout has settled. It does not derive the per-replica power cost from individual Pods in this phase.

kaim-eng: You're right that homogeneous DGD-sourced accounting would be incorrect if an admitted DGD update could leave existing Pods with a different cap or GPU topology.

This update closes that failure mode for the supported phase-1 configuration through CRD transition validation. For every power-annotated DGD component, CEL rules reject addition, removal, or changes to:
- `dynamo.nvidia.com/gpu-power-limit`
- the effective main-container `nvidia.com/gpu` count, using limits-first/request-fallback semantics
- `multinode.nodeCount`, defaulting to 1

Therefore, admitted DGD updates cannot create mixed-cost worker generations within that DGD. Replica counts and unrelated Pod-template fields remain mutable.

Actual-Pod committed-budget accounting would be needed to support mixed generations, live cap retargeting, or DRA-backed GPU allocation. Those configurations remain explicitly unsupported and out of scope for this phase.

components/src/dynamo/planner/environment/base.py
kaim-eng: Fixed through CRD transition validation. CEL rules now reject any DGD update that changes the power annotation, effective `nvidia.com/gpu` count, or `multinode.nodeCount` on a power-annotated component.

The stale-cache scenario you described — such as changing the DGD cap to 400 W while Planner retains a 250 W startup value — is therefore rejected before it can trigger a worker rollout. Changing those inputs requires deleting and recreating the DGD and starting Planner against the replacement deployment.

The startup-cache docstrings, PlannerConfig descriptions, and power-aware documentation now state this contract explicitly.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
kaim-eng: Fixed. `get_actual_worker_counts` is Pod-list free again. The power-aware path now uses a dedicated `get_power_aware_worker_counts` method that:
1. Reads the DGD once.
2. Issues one DGD-scoped Pod LIST using `nvidia.com/dynamo-graph-deployment-name=<name>`.
3. Partitions the result locally by `nvidia.com/dynamo-component`.
4. Runs the terminating-Pod checks against those in-memory partitions.

The synchronous DGD GET and Pod LIST are executed together in one `asyncio.to_thread` dispatch, so they do not block the Planner event loop. Watch-backed caching remains a possible follow-up.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
kaim-eng: Fixed as suggested. `PipelineOutcome` now carries `proposed_components: frozenset[ComponentKey]`, derived from the raw PROPOSE-stage `OverrideResult.targets` whose `replicas` value is not `None`.

That mask is preserved through RECONCILE and CONSTRAIN. `_project_scale_to` uses it to remove baseline-filled peers before applying the final budget, instead of inferring provenance from `target == ready`.

A target is now suppressed as a no-op only when the settled expected count is known and equal. When expected state is unknown during a rollout, an explicit target equal to transient ready is preserved because it may intentionally cancel the in-flight target. Rollout-held scale-ups emit `None` directly.

components/src/dynamo/planner/connectors/base.py
kaim-eng: Fixed using the explicit-capability approach. `PowerAwareConnector` now defines `get_power_aware_worker_counts`, and `_POWER_AWARE_REQUIRED` includes it as the fourth required method.

The power path calls this dedicated method rather than overloading `get_actual_worker_counts` with a Kubernetes-specific keyword argument. A connector implementing only the previous three power capabilities now fails the TypeGuard before invocation.

The ordinary `PlannerConnector.get_actual_worker_counts` signature remains unchanged, so Virtual and other non-Kubernetes connectors are unaffected.

kaim-eng:
C:/Program Files/Git/ok to test 3983b3d983

components/src/dynamo/planner/monitoring/dgd_services.py
kaim-eng: At startup, Planner waits for the DGD generation and rollout to settle, then verifies that every relevant labelled, non-terminal Pod carries the exact DGD power annotation before caching the per-replica cost. Stefan's reference implementation (incorporated into this branch) replaces the generated CEL rules with a Go validating webhook that rejects adding, removing, or changing that annotation and rejects changes to the effective `nvidia.com/gpu` count or `multinode.nodeCount`. It also rejects DRA-backed allocation for a power-annotated component on create and update. Therefore, under the supported operator-managed contract, a DGD cannot produce mixed per-logical-replica power costs, and homogeneous DGD-sourced accounting is equivalent to summing those Pod costs. The shipped Helm configuration defaults the webhook to `failurePolicy: Fail`. Please resolve if this closes the concern.

components/src/dynamo/planner/environment/base.py
kaim-eng: The stale-cache scenario — DGD cap changed without a Planner restart — is now blocked at the k8s API level. Stefan's reference implementation (incorporated into this branch) provides a Go admission webhook that rejects any DGD update changing the power annotation, effective GPU count, or node count on a power-annotated component, and rejects DRA-backed allocation on create and update. The shipped Helm configuration defaults the webhook to `failurePolicy: Fail` (this is configurable). The `_load_static_power_caps_at_startup` docstring has been updated to name the webhook and its DRA constraint explicitly. Please resolve if this closes the concern.

components/src/dynamo/planner/connectors/base.py
kaim-eng: Fixed in the current tip: `get_power_aware_worker_counts` is both defined on `PowerAwareConnector` and listed as the fourth entry in `_POWER_AWARE_REQUIRED`. `is_power_aware_connector` uses `inspect.getattr_static` to avoid MagicMock false-positives. Please resolve.

components/src/dynamo/planner/plugins/orchestrator/engine_adapter.py
kaim-eng: Fixed in the current tip: `PipelineOutcome` carries `proposed_components: frozenset[ComponentKey]` derived from raw PROPOSE-stage targets. `_project_scale_to` uses that mask to drop baseline-filled peers before the budget boundary instead of inferring provenance from `target == ready`. Post-budget no-op suppression compares against `expected_*` (not `ready_*`) to preserve explicit targets during rollouts. Please resolve.

components/src/dynamo/planner/connectors/clients/kubernetes_api.py
kaim-eng: Fixed in the current tip: the power-aware path calls `get_power_aware_worker_counts`, which dispatches `_get_power_aware_worker_counts_sync` via `asyncio.to_thread`. That function performs exactly one DGD GET and one DGD-scoped Pod LIST (`DYNAMO_DGD_NAME_LABEL` selector), partitions the result in-memory by component, and runs terminating-Pod checks against those partitions — no per-component LISTs and no event-loop blocking. Please resolve.

kaim-eng:
C:/Program Files/Git/ok to test 354a0240c9

components/src/dynamo/planner/monitoring/dgd_services.py
kaim-eng: This sourcing from the Deployment/DGD was @tedzhouhk's explicit design direction — see his P1 comment [#discussion_r3634837362](https://github.com/ai-dynamo/dynamo/pull/12012#discussion_r3634837362):

> "Read and validate the annotation once during initialization after `wait_for_deployment_ready`, cache `watts_per_replica`, and fail startup if it is unavailable or malformed. Annotation changes can take effect through the normal worker rollout plus Planner restart."

The rationale: the DGD is the authoritative source of *intent* — it owns the per-GPU cap and GPU topology for replicas that don't exist yet. Pods carry that intent only after the operator propagates it; summing pod-level values would mix Planner-layer accounting with actuation-layer state. Ted wanted the Planner to stay at the intent layer, which also aligns with how the profiler sources per-component configuration.

The invariant that makes homogeneous DGD-sourced accounting correct is enforced by the Go admission webhook: it rejects any change to the power annotation or effective GPU count on an existing DGD component, so no admitted update can produce mixed-cost worker generations within a deployment.

sttts (approved):

kaim-eng:
C:/Program Files/Git/ok to test 6dc517e616

kaim-eng:
C:/Program Files/Git/ok to test deb67c35b6

jthomson04 (approved):
approving planner_metrics.py observability changes.

dmitry-tokarev-nv (approved):
Only approving process changes from DevOps side.

Issue #11983, opened 2026-07-21
perf(operator): reduce Pod cache footprint on large clusters


julienmancuso:
## Problem

The operator's controller-runtime cache stores full `corev1.Pod` objects because several controllers watch or read Pods. Event predicates reduce reconciliations, but they do not reduce what the shared informer retains.

On clusters with many Pods, this may consume unnecessary operator memory. Optimizing a single watch as metadata-only may not help if another controller still requires the full-Pod informer, and could accidentally create a second cache.

## Goal

Determine whether the shared Pod cache has a meaningful memory cost at large-cluster scale and, if it does, reduce that cost without changing controller behavior.

## Proposed work

1. Measure operator memory with one representative high-Pod-count workload after the cache has synchronized.
2. Identify the Pod fields required by the current controllers that watch or read Pods.
3. If the measured cost is material, implement the simplest controller-runtime-supported cache configuration that reduces it without introducing duplicate cluster-wide Pod caches.
4. Report the before/after memory results and run the existing controller tests.

Possible mechanisms include a shared cache transform, object selectors, metadata-only caching, or uncached reads for the few paths that need complete Pods. The implementation should choose the simplest option supported by the measurements and field requirements.

## Done when

- [ ] The baseline measurement and test setup are documented.
- [ ] Required Pod data is identified.
- [ ] If optimization is warranted, before/after results demonstrate the improvement and existing controller tests pass.
- [ ] The resulting design does not accidentally maintain both full-object and metadata-only caches for all Pods.

If the baseline shows that Pod caching is not a material contributor to operator memory, documenting that result is sufficient to close the issue without adding complexity.

## Related

- Recreate rollout and metadata-only watch discussion: #11909
- API-server list/watch pressure: #7416

sttts:
Big (not giant) cluster: 10k nodes, 100000 pods. Each pod is up to 2MB => 200GB of memory. Easy math.

sttts:
But it is not only memory, listing 100.000 pods will take minutes. And it is load on the apiserver to send out full pod events on every change (and pods change frequently).

sttts:
> several controllers currently need real Pod fields:
> - Failover needs status.phase and phase transitions: [failover_cascade_controller.go (line 86)](/Users/sschimanski/src/ai-dynamo/dynamo/deploy/operator/internal/controller/failover_cascade_controller.go:86).
[...]

PR #12014, opened 2026-07-22
perf(operator): reduce memory retained by the shared Pod cache

julienmancuso:
## Summary

Reduce the operator’s steady-state memory usage by projecting Pods before they enter the shared controller-runtime informer cache.

The operator previously retained complete `corev1.Pod` objects for every Pod in scope. Since the topology, checkpoint, failover, snapshot, model, and DGDR controllers already share that informer, optimizing an individual controller watch would not reduce the underlying cache footprint.

This change:

- Configures one Pod-specific transform on the existing shared typed Pod informer.
- Retains the Pod fields currently consumed by operator controllers:
  - identity, labels, annotations, owner references, finalizers, and deletion state
  - node assignment
  - container names, commands, and arguments
  - relevant DownwardAPI volume field paths
  - Pod phase and Ready condition
  - waiting and terminated container diagnostic state
- Drops heavyweight unused fields before Pods are committed to the cache.
- Preserves namespace-restricted cache configuration.
- Uses the same cache configuration in controller envtests.
- Keeps full-object access available through the manager’s uncached `APIReader`.
- Adds contract coverage for topology, checkpoint, snapshot, failover, model, DGDR, and the Recreate strategy introduced by #11909.
- Verifies that topology metadata patches do not accidentally overwrite fields removed by the projection.

This does not introduce a metadata-only informer or a second Pod watch. Controllers continue using the existing typed Pod informer and cached client.

The transform reduces retained heap usage. Kubernetes still sends and the operator still decodes complete Pod objects before applying the projection, so LIST/WATCH network traffic is unchanged.

Related to #11983. PR #11909 will be rebased onto this change and should use the shared typed Pod watch.

## Validation

- `go test ./internal/controller -count=1`
- `go test ./internal/podcache ./internal/testing/operatorenv ./cmd/... -count=1`
- `go test ./internal/... -run '^$' -count=1`
- `go vet ./internal/podcache ./internal/controller`
- `git diff --check`

## Summary by CodeRabbit

* **Improvements**
  * Reduced cached Pod data to the information required for controller operations, improving cache efficiency.
  * Added safeguards to ensure the Pod cache is configured correctly before the operator starts.
  * Preserved essential Pod metadata, status, topology, checkpoint, and failover information.

* **Bug Fixes**
  * Prevented the operator from starting when Pod cache configuration fails.

* **Tests**
  * Added coverage for Pod projection, cache configuration, controller compatibility, and metadata updates.


deploy/operator/internal/podcache/transform.go
```diff
@@ -0,0 +1,202 @@
...
+// whenever a field outside this projection is required.
+func Project(pod *corev1.Pod) *corev1.Pod {
+	if pod == nil {
+		return nil
+	}
+
+	// All labels, annotations, owner references, identity, finalizers, and
+	// deletion state are retained. ManagedFields is the large metadata field no
+	// controller consumes.
+	pod.ManagedFields = nil
```
sttts, line 73: I think this is right, just double check that we are allowed mutate the pod. I think we can.

sttts (approved):

deploy/operator/internal/podcache/transform.go
julienmancuso: yes :
[client-go v0.36.2 — TransformFunc documentation](https://pkg.go.dev/k8s.io/client-go@v0.36.2/tools/cache#TransformFunc)
“TransformFunc sees the object before any other actor, and it is now safe to mutate the object in place instead of making a copy.”

PR #12019, opened 2026-07-22
feat(recipes): add GPQA-Diamond accuracy check for GLM-5.2

### LINKED ISSUE #12018: feat(recipes): general accuracy check example for recipes
## Type of Change
New feature

## Problem Statement
Accuracy is usually evaluated alongside throughput and latency when validating a deployment, to confirm it performs in line with the checkpoint's published results. Recipes ship perf benchmarks (`perf.yaml`) but there is no equivalent starting point for an accuracy run, so anyone wanting one builds it from scratch.

## Proposed Solution
Add a single model-agnostic accuracy check at `recipes/accuracy/` that runs a public benchmark against any deployed recipe with AIPerf and grades the answers, so the score reflects the whole serving path. The result is read as an A/B test against the checkpoint's published model-card number rather than as an absolute capability claim.

Everything model-specific — endpoint, model ID, benchmark, dataset size, sampling parameters — is lifted into a `CONFIGURE` block of env vars, so the check is pointed at whichever recipe is deployed.

## Estimated PR Size
M (101-300 lines) — ~335 lines across 2 new files.

## Files/Components Affected
- `recipes/accuracy/accuracy.yaml` (new) — the Job
- `recipes/accuracy/README.md` (new) — methodology and how to read the result

marckarp:
## What
Adds a GPQA-Diamond accuracy check to the GLM-5.2 recipe: an aiperf-based Kubernetes Job that A/B tests a deployed endpoint against the checkpoint's model-card baseline, plus measured results for all four targets and docs.

## Why
The checkpoints the recipe serves are already accuracy-validated ([nvidia/GLM-5.2-NVFP4](https://huggingface.co/nvidia/GLM-5.2-NVFP4) publishes GPQA-Diamond results for the quantized model alongside its FP8 baseline). However, it can still be reassuring to perform a final accuracy check to ensure no bugs have crept in along the path to deployment — a stale env var or misconfigured flag can silently corrupt outputs while throughput looks healthy.

## How
`recipes/glm-5.2/accuracy/accuracy.yaml` runs `aiperf profile --accuracy-benchmark gpqa_diamond`: it sends the 198 GPQA-Diamond questions as chat requests through a deployed DGD and grades each answer (simple-evals prompt, lighteval letter extraction) against ground truth, comparing the score (A) to the checkpoint's published model-card number (B). One Job covers all four targets via the `ENDPOINT` env var. Sampling matches the model-card methodology (temperature 1.0, top_p 0.95, max_new_tokens 100k).

## Results
Measured against the NVIDIA model-card baseline (B200 serves NVFP4, H200 serves FP8), 0 errors and 0 unparsed answers on every run. Each score is a proportion over the same 198 questions, so it carries a binomial 95% confidence interval of ±(1.96·√(p(1−p)/198)) ≈ ±4.0–4.6 points:

| Config | Score | 95% CI | Model card | Card inside CI? |
|---|---:|---|---:|---|
| disagg-b200 | 90.91 (180/198) | [86.9, 94.9] | 89.39 | ✅ |
| agg-b200 | 88.89 (176/198) | [84.5, 93.3] | 89.39 | ✅ |
| disagg-h200 | 89.39 (177/198) | [85.1, 93.7] | 89.52 | ✅ |
| agg-h200 | 87.88 (174/198) | [83.3, 92.4] | 89.52 | ✅ |

Every model-card baseline falls within its config's confidence interval (largest deviation: agg-h200 at 0.75σ below baseline, two-sided p≈0.45). The scores are statistically indistinguishable from the model card. Note the resolution limit: a single n=198 run can detect regressions of roughly ≥4 points — sufficient to catch deployment bugs, which typically cause large drops.

## Testing
- Ran the accuracy Job against all four deployed GLM-5.2 DGDs (B200, H200) — results above.
- `fern check` passes with 0 errors on the docs page.

Fixes #12018

### DIFF
--- a/docs/recipes/glm-5-2.mdx
+++ b/docs/recipes/glm-5-2.mdx
@@ -231,6 +231,31 @@ All four targets serve GLM-5.2 on SGLang with KV-aware routing and EAGLE-style M
 | **KV offload** | HiCache CPU | HiCache CPU | None | None |
 | **Max context** | 500K | 500K | 250K | 250K |
 
+## Accuracy
+
+The checkpoints these targets serve are already accuracy-validated — [nvidia/GLM-5.2-NVFP4](https://huggingface.co/nvidia/GLM-5.2-NVFP4) publishes GPQA-Diamond results for the quantized model alongside its FP8 baseline. However, it can still be reassuring to perform a final accuracy check to ensure no bugs have crept in along the path to deployment. The `accuracy/accuracy.yaml` Job does this as an A/B test against the model card: it runs AIPerf's GPQA-Diamond benchmark, sending the 198 real GPQA-Diamond questions as chat requests through your deployed DGD (A) and comparing the graded score to the checkpoint's published number (B). Each answer is graded by extracting the chosen A/B/C/D letter and exact-matching against ground truth. Compare the score to the model-card number, not to a fixed threshold.
+
+Edit `ENDPOINT` in `accuracy/accuracy.yaml` to your deployed frontend service, then run the Job. `glm-5-2-gpqa` is a fixed-name Job, so clear any prior run before re-applying (for example when switching `ENDPOINT` to a different target):
+
+```bash
+kubectl delete job glm-5-2-gpqa -n ${NAMESPACE} --ignore-not-found
+kubectl apply -f recipes/glm-5.2/accuracy/accuracy.yaml -n ${NAMESPACE}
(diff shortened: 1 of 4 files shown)


marckarp:
Superseded by #12495, which reworks this into a general, model-agnostic accuracy example at `recipes/accuracy/` per review feedback. The Job is the same; everything model-specific is now configurable, and the GLM-5.2 results are dropped so there is nothing per-recipe to maintain.

PR #12020, opened 2026-07-22
fix(vllm): eliminate measurement-noise sources in FPM self-benchmark …

liyuanzhe1991:
## Overview:

Single-sample self-benchmark `wall_time` measurements are intermittently poisoned by 3–200× outliers, forcing expensive workarounds (measure-N-times-take-median). We root-caused **two independent mechanisms** and fix both, making single-sample collection reach median-of-5 database quality at ~1/3.5 the cost:

1. **CPython gen2 (full) GC inside the engine processes.** A gen2 pass walks every tracked object, so its pause grows with the heap — measured **0.4 s → 9 s** over a 30 k-point sweep — and under TP/DP lockstep a single worker's pause stalls the whole forward pass. Triggers follow CPython's 25 % long-lived-growth rule, producing outliers at geometrically spaced, run-independent iteration indices (ratio ≈ 1.23, identical across nodes/days/topologies) at a fleet-constant ~0.08 %/sample rate.
2. **Eager-shape first-execution cost.** Shapes beyond the cudagraph capture range pay a ~1 s one-off on their first execution — five consecutive repeats of decode batch 513 read `[1126, 147, 147, 144, 144]` ms. Captured shapes are immune because capture itself executes them at startup; a single-sample sweep books the one-off as the eager point's latency.

## Details:

**GC policy (env-gated, off by default — zero behavior change unless enabled):**
- New `dynamo/vllm/gc_policy.py`: on activation, disable automatic gen2 (`gc.set_threshold(t0, t1, 1<<30)`), `gc.freeze()` the engine-init heap once, then run a daemon thread executing a **bare `gc.freeze()`** per tick (O(1) generation-list splice: no traversal, no pause). `gc_maintain()` (collect+freeze) is provided for explicit reclamation from untimed windows.
- Worker processes are covered via vLLM's `worker_extension_cls` (`FpmGcWorkerExtension`, auto-injected in benchmark mode when `DYN_FPM_GC_POLICY` is set and the user configured no extension — collision-safe); the engine-core process starts the policy from `InstrumentedScheduler._bench_init`. No vLLM changes.
- Two rejected variants are documented in the module docstring (both reproduce measurable stalls): periodic `gc.collect()` re-introduces growing full-heap walks (outlier rate 0.082 % → 0.712 %), and per-tick `gc.get_freeze_count()` walks the whole permanent generation (0.4–4 s once millions of objects are frozen).

**Eager-shape warmup (benchmark mode only):**
- `_bench_build_grid` prepends one **discarded warmup replica** per eager shape, deduped by the shape driver (token count for prefill, batch size for decode; typically a handful of replicas ≈ +10 s bring-up).
- Replicas run the normal injection/lockstep/validation path on every rank — DP grids stay identical by construction — and are dropped at save time. Published benchmark IDs stay contiguous (native-artifact contract) because real points are numbered first; `counter_id` stamping uses `point.benchmark_id` directly, so execution order and ID order are independent.

### Validation (GLM-5.2-NVFP4, 4×B200, vLLM 0.25.1)

**Stage 1+2 — A/B on identical 30 480-measurement sweeps** (same node, same GPUs, same point list; only the GC policy toggled):

| run | outliers >3× coordinate median | max pause |
|---|---|---|
| baseline | 25 (0.082 %) — geometric gen2 ladder | **8 900 ms** |
| GC policy on | 2 (0.007 %) | 1 131 ms |

The 2 residual outliers are exactly the first samples of the two eager shapes (decode batch 513 / 1024): their 5 consecutive repeats read `[1126, 147, 147, 144, 144]` ms — first execution pays the one-off, the rest are warm. That is the fingerprint fix 2 targets.

**Stage 3 — both fixes, fresh standard single-sample collection** (the production protocol):

| metric | before fixes | after both fixes |
|---|---|---|
| curve-neighbor poisoned rows (whole DB) | 12, worst 34× | **0** |
| eager decode points (B=513 / B=1024, min-KV) | 1 216 / 1 184 ms (first-touch) | **190 / 185 ms (warm)** |
| LOO prefill (4 097 pts) | 8.02 % / max 2 913 % | **4.74 % / max 125 %** (median-of-5 reference: 4.17 % / 129 %) |
| holdout vs frozen truths (509 pts) | 2.77 % / max 37.8 % | **2.24 % / max 12.2 %** (reference: 1.96 % / 13.5 %) |

Single-sample collection reaches median-of-5 database quality at ~1/3.5 the cost.

**End-to-end, fresh standard single-sample collections with both fixes:**

- curve-neighbor poisoned rows: **12 → 0**
- the two eager decode points read **190 / 185 ms (warm)** vs 1216 / 1184 ms first-touch before
- LOO prefill **4.74 % / max 125 %** and holdout-vs-frozen-truths **2.24 % / max 12.2 %** match the median-of-5 reference (4.17 % / 129 %; 1.96 % / 13.5 %) — at ~1/3.5 the collection cost

No sweep-duration regression, no change to the non-outlier latency distribution, no OOM across four full sweeps with gen2 disabled (worker RSS ≈ +1–2 GB/h, decelerating).

- `components/src/dynamo/vllm/gc_policy.py` — the GC policy incl. design rationale in the module docstring
- `components/src/dynamo/vllm/instrumented_scheduler.py` — `_bench_eager_warmup_points`, the save-time discard, ID numbering; 2-line GC hook in `_bench_init`
- `components/src/dynamo/vllm/args.py` — benchmark-mode `worker_extension_cls` injection
- `components/src/dynamo/vllm/tests/test_gc_policy.py`

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12020" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional garbage-collection policy to improve benchmark stability and reduce pauses.
  * Added configurable background memory maintenance, with manual maintenance controls available during execution.
  * Added eager warmup for benchmark shapes without CUDA-graph capture.

* **Bug Fixes**
  * Warmup runs are now excluded from reported benchmark results, preventing them from affecting performance metrics.

* **Tests**
  * Added coverage for garbage-collection configuration, maintenance behavior, interval handling, and benchmark warmup processing.

### DIFF
--- a/components/src/dynamo/vllm/args.py
+++ b/components/src/dynamo/vllm/args.py
@@ -351,17 +351,6 @@ def update_engine_config_with_dynamo(
                 "scheduler_cls"
             ] = "dynamo.vllm.instrumented_scheduler.InstrumentedScheduler"
             logger.info("Benchmark mode: auto-enabling InstrumentedScheduler")
-        if os.environ.get("DYN_FPM_GC_POLICY"):
-            existing_ext = getattr(engine_config, "worker_extension_cls", None)
-            if not existing_ext:
-                from dynamo.vllm.gc_policy import WORKER_EXTENSION_CLS
-
-                defaults["worker_extension_cls"] = WORKER_EXTENSION_CLS
-                logger.info(
-                    "Benchmark mode: DYN_FPM_GC_POLICY set, injecting "
-                    "worker_extension_cls=%s",
-                    WORKER_EXTENSION_CLS,
-                )
         elif existing_cls is not None and "InstrumentedScheduler" not in str(
             existing_cls
         ):
@@ -370,6 +359,27 @@ def update_engine_config_with_dynamo(
                 f"--scheduler-cls is set to '{existing_cls}'. Either remove "
                 f"--scheduler-cls or use a subclass of InstrumentedScheduler."
             )
+        if os.environ.get("DYN_FPM_GC_POLICY"):
+            # Class path as a literal, not an import: importing
+            # dynamo.vllm.gc_policy auto-starts the policy in the importing
+            # process, and this launcher process must stay untouched.
+            worker_extension_cls = "dynamo.vllm.gc_policy.FpmGcWorkerExtension"
+            existing_ext = getattr(engine_config, "worker_extension_cls", None)
+            if not existing_ext:
+                defaults["worker_extension_cls"] = worker_extension_cls
+                logger.info(
+                    "Benchmark mode: DYN_FPM_GC_POLICY set, injecting "
+                    "worker_extension_cls=%s",
+                    worker_extension_cls,
+                )
+            elif str(existing_ext) != worker_extension_cls:
+                raise ValueError(
+                    f"DYN_FPM_GC_POLICY requires "
+                    f"worker_extension_cls='{worker_extension_cls}' so model "
+                    f"workers apply the GC policy, but --worker-extension-cls "
+                    f"is set to '{existing_ext}'. Remove it or unset "
+                    f"DYN_FPM_GC_POLICY."
+                )
         benchmark_config: Dict[str, Any] = {
             "mode": dynamo_config.benchmark_mode,
             "warmup_iterations": dynamo_config.benchmark_warmup_iterations,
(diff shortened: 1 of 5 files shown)


tedzhouhk:
Turning off the GC for all processes seems a little too aggressive. We want the engine to serve regular traffics normally after self-benchmarking.

According to the description, the root cause of the GC is in the GPU runner process. This is a different process than the `InstrumentedScheduler`. I guess the best solution is exposing an API from model executor to query the last GC snapshot (i.e., something like `model_executor.collective_rpc("fpm_gc_snapshot")`). But this requires vllm side code change and takes a long time. Consider the 0.08% low GC chance, can we do some anomaly detection and rerun the affected points?

PR #12021, opened 2026-07-22
fix: synchronize self-benchmark capacity across DP ranks

liyuanzhe1991:
## Overview:

Self-benchmarking with attention-DP > 1 intermittently deadlocks at startup: each DP rank probes its local KV-cache capacity independently, and any run-to-run variation (block-pool rounding, allocator state) makes ranks generate *different* benchmark grids. The lockstep synchronization protocol then waits forever on points that only a subset of ranks generated. In our GLM-5.2-NVFP4 / 4×B200 campaigns, decode benchmarks with dp=4 failed 3 out of 4 attempts before this fix.

## Details:

- Each rank builds a local capacity envelope (`_bench_make_local_capacity`: max_model_len, max_num_running_reqs, max_num_scheduled_tokens, usable KV blocks) and exchanges it through the existing benchmark synchronizer before grid generation.
- Grid construction reads capacity exclusively through `_bench_capacity_limit()` / `_bench_grid_usable_blocks()`, which return the **minimum across ranks** — every rank derives byte-identical grids by construction.
- `_bench_decode_feasible_max_batch_size()` now computes from the synchronized envelope, so both call sites (the native decode grid and explicit benchmark points) agree across ranks.
- Single-rank (dp=1) behavior is unchanged: the synchronized envelope degenerates to the local one.

Validated on 4-rank and 8-rank attention-DP topologies (GLM-5.2-NVFP4, B200, vLLM 0.25.1): previously flaky decode cells pass deterministically; full prefill+decode campaigns (5 466 and 6 508 grid points) completed with zero lockstep timeouts.

- `components/src/dynamo/vllm/instrumented_scheduler.py`:
  - `_BenchmarkCapacityEnvelope`, `_bench_make_local_capacity`, `_bench_capacity_limit`, `_bench_grid_usable_blocks`
  - the capacity exchange in the `_BenchmarkSynchronizer` handshake
  - `_bench_decode_feasible_max_batch_size` (now envelope-based)
- `components/src/dynamo/vllm/tests/test_vllm_instrumented_scheduler.py` — capacity-divergence regression tests

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Improved distributed benchmark setup by negotiating shared capacity across ranks.
  * Benchmark grids now adapt to the common capacity available across participating ranks.
  * Benchmark results include the negotiated common capacity when available.

* **Bug Fixes**
  * Added validation to detect capacity and benchmark-grid mismatches before warmup.
  * Improved KV-cache capacity accounting for prefill and decode benchmark scenarios.

* **Tests**
  * Added regression coverage for capacity negotiation, grid synchronization, and mismatch handling.

### DIFF
--- a/components/src/dynamo/vllm/instrumented_scheduler.py
+++ b/components/src/dynamo/vllm/instrumented_scheduler.py
@@ -2926,8 +2926,7 @@ class InstrumentedScheduler(AsyncScheduler):
         # their nominal coordinate).
         max_model_len = self._bench_capacity_limit("max_model_len")
         if any(
-            max(context_len, 2) + 2 > max_model_len
-            for context_len in context_lengths
+            max(context_len, 2) + 2 > max_model_len for context_len in context_lengths
         ):
             return False
         required_blocks = sum(
@@ -4023,8 +4022,13 @@ class InstrumentedScheduler(AsyncScheduler):
             },
             "capacity": {
                 "common": (
-                    asdict(self._bench_negotiated_capacity)
-                    if getattr(self, "_bench_negotiated_capacity", None) is not None
+                    asdict(negotiated_capacity)
+                    if (
+                        negotiated_capacity := getattr(
+                            self, "_bench_negotiated_capacity", None
+                        )
+                    )
+                    is not None
                     else None
                 ),
             },
--- a/components/src/dynamo/vllm/instrumented_scheduler.py
+++ b/components/src/dynamo/vllm/instrumented_scheduler.py
@@ -2182,7 +2182,15 @@ class InstrumentedScheduler(AsyncScheduler):
             "prefill_cudagraph_mode": self._bench_prefill_cudagraph_mode,
             "decode_cudagraph_mode": self._bench_decode_cudagraph_mode,
             "prefill_capture_sizes": self._bench_prefill_capture_sizes,
-            "decode_capture_sizes": self._bench_decode_capture_sizes,
+            # ``_bench_decode_capture_sizes`` is filtered by
+            # ``max_num_running_reqs`` — a negotiable capacity value that may
+            # legitimately differ across ranks — so hash the unfiltered
+            # configuration and re-filter after negotiation.
+            "decode_capture_sizes": (
+                list(self._bench_cudagraph_capture_sizes)
+                if self._bench_decode_cudagraph_mode != "NONE"
+                else []
+            ),
         }
         encoded = json.dumps(payload, sort_keys=True, separators=(",", ":")).encode()
         return hashlib.sha256(encoded).hexdigest()
@@ -2241,6 +2249,14 @@ class InstrumentedScheduler(AsyncScheduler):
             asdict(local_capacity),
             asdict(common_capacity),
         )
+        # The activation-time filter used the local request limit; re-filter
+        # with the negotiated one so every rank builds the decode grid from
+        # the same capture list.
+        self._bench_decode_capture_sizes = [
+            size
+            for size in self._bench_decode_capture_sizes
+            if size <= common_capacity.max_num_running_reqs
+        ]
 
         self._bench_grid_built = True
         mode = self._bench_config.mode
@@ -2313,7 +2329,7 @@ class InstrumentedScheduler(AsyncScheduler):
         capture_size, padding_tokens, reasons = self._bench_cudagraph_metadata(
             candidate.total_prefill_tokens,
             self._bench_prefill_capture_sizes,
-            self.max_num_scheduled_tokens,
+            self._bench_capacity_limit("max_num_scheduled_tokens"),
         )
         return BenchmarkPoint(
             point_type="prefill",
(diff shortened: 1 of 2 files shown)


components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -1828,10 +2080,118 @@ def _bench_init(self, vllm_config: "VllmConfig") -> None:
...
+            ),
+            "use_eagle": getattr(
+                getattr(self, "kv_cache_manager", None), "use_eagle", False
+            ),
+            "uses_per_group_cache_lookup": self._bench_uses_per_group_cache_lookup(),
+            "manager_layout": manager_layout,
+            "prefill_cudagraph_mode": self._bench_prefill_cudagraph_mode,
+            "decode_cudagraph_mode": self._bench_decode_cudagraph_mode,
+            "prefill_capture_sizes": self._bench_prefill_capture_sizes,
+            "decode_capture_sizes": self._bench_decode_capture_sizes,
```
tedzhouhk: **[P1] Normalize decode capture sizes after capacity negotiation**

`_bench_decode_capture_sizes` is filtered with each rank's local `max_num_running_reqs` before this digest is computed. If ranks report different request limits—which this envelope explicitly tries to reconcile—they hash different capture lists, and `_BenchmarkCapacityEnvelope.common()` raises `grid invariants differ` before it can choose the shared minimum. Please hash the unfiltered capture configuration and derive the filtered decode list from `common_capacity.max_num_running_reqs` after negotiation.

components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -1961,7 +2327,7 @@ def _bench_raise_explicit_infeasible(
         )
 
     def _bench_generate_prefill_grid(self) -> None:
-        max_tokens = self.max_num_scheduled_tokens
+        max_tokens = self._bench_capacity_limit("max_num_scheduled_tokens")
```
tedzhouhk, line 2396: **[P1] Use the common limit for explicit prefill metadata**

This generated-grid path correctly uses `_bench_capacity_limit`, but `_bench_materialize_prefill_candidate()` still passes the local `self.max_num_scheduled_tokens` as `axis_limit` to `_bench_cudagraph_metadata()`. With per-rank limits of 8192 and 10000, an explicit 8192-token point gets different `engine_limit`/`geometric_axis` reasons, so the final grid digests differ even though the point is feasible under the negotiated capacity. Please use the negotiated limit in the explicit path as well.

components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -2360,7 +2731,12 @@ def _bench_cached_kv_read_tokens(self, req: Request) -> int:
         return cached_tokens
 
     def _bench_generate_decode_grid(self) -> None:
-        if self.max_model_len < 3:
+        max_model_len = self._bench_capacity_limit("max_model_len")
+        max_num_running_reqs = self._bench_capacity_limit("max_num_running_reqs")
```
tedzhouhk: **[P1] Remove the unused capacity locals**

Both `max_num_running_reqs` and `max_num_scheduled_tokens` are assigned but never read in this function. Ruff reports F841 for these lines, and the required pre-commit check is currently failing. Please remove both assignments.

tedzhouhk (commented):
Three blocking findings from the agentic review are noted inline.

components/src/dynamo/vllm/instrumented_scheduler.py
liyuanzhe1991: Fixed in 64efd51f0 (carried through the rebase): the grid-invariants digest now hashes the unfiltered cudagraph capture configuration, and `_bench_build_grid` re-filters `_bench_decode_capture_sizes` with the negotiated `max_num_running_reqs` after negotiation, so ranks that differ only in request limits negotiate the minimum instead of failing the invariant check. Regression test: `test_capacity_digest_ignores_request_limit_filtered_capture_sizes`.

components/src/dynamo/vllm/instrumented_scheduler.py
liyuanzhe1991: Fixed in 64efd51f0: `_bench_materialize_prefill_candidate` now passes `self._bench_capacity_limit("max_num_scheduled_tokens")` so explicit prefill points get identical cudagraph metadata (engine_limit vs geometric_tail) on every rank. Regression test: `test_explicit_prefill_point_uses_negotiated_scheduled_token_limit`.

components/src/dynamo/vllm/instrumented_scheduler.py
liyuanzhe1991: Fixed in 64efd51f0: both unused locals removed; ruff is clean.

tedzhouhk (approved):
Verified the three previously raised P1 concerns against the current head: decode capture invariants are negotiated correctly, explicit prefill metadata uses the common limit, and the unused locals are removed. The added regression coverage matches the fixes.

PR #12033, opened 2026-07-22
feat(mocker): add selectable native G1 manager

dreamtalen:
## Overview

This PR adds a selectable, self-contained Native G1 KV manager for the shared vLLM/TensorRT-LLM mock scheduler. It removes KVBM dependencies from the Native implementation while keeping the existing KVBM path available during migration.

KVBM remains the default. Users must explicitly select `g1_backend=native` (or `--g1-backend native`). The Native path is G1-only and cannot be combined with the legacy `kvbm-offload` G2/G3/G4 implementation.

## Summary

- Add a minimal physical block-pool model that tracks real capacity, distinct physical copies for duplicate hashes, cached-prefix pins, request references, and inactive LRU eviction.
- Add request-owned vLLM block tables and computed-boundary cache visibility, including Store/Remove events normalized to router-visible hash-presence transitions.
- Add a temporary `G1Manager` composition boundary so the shared scheduler can select KVBM or Native without leaking KVBM symbols into the Native implementation.
- Add source firewalls that keep the physical pool independent of KVBM, scheduler, and Dynamo router/event adapters, and keep the Native backend independent of all KVBM symbols.
- Support vLLM and TensorRT-LLM scheduling, including prefix caching, chunked prefill, MTP/NextN, preemption, and handoff lifecycle behavior.

### Native G1 semantics

The Native implementation models the physical behavior that affects scheduling and replay results without reproducing vLLM's concrete numeric block IDs:

- each request owns an append-only physical block table;
- duplicate full-block hashes consume distinct physical capacity;
- reusable prefix blocks are pinned atomically with fresh-capacity reservation;
- newly allocated blocks remain request-private until their computed-token boundary;
- cached copies transition between active references and inactive LRU state;
- eviction publishes a Removed event only when the final physical copy of a hash disappears; and
- prefix-cost accounting distinguishes active and inactive cached overlap.

Destination reservations and ordinary decode reservations use the same physical pool. Preemption, MTP recomputation, handoff activation/cancellation, and terminal cleanup are covered through the shared scheduler tests.

### Dependency boundary

`lib/mocker/src/kv_manager/g1_manager.rs` is a temporary static-composition layer. It adapts scheduler-facing `MoveBlock`/lineage data into the Native backend's local types and dispatches to either implementation. Once the legacy KVBM G1 path is removed, this wrapper can be deleted.

The source-firewall tests enforce the intended dependency direction:

- `cache/vllm_block_pool.rs` cannot depend on KVBM, scheduler code, or Dynamo router/event adapters.
- `kv_manager/vllm_backend.rs` cannot reference any KVBM symbol or legacy G1/offload type.

### Intentional shared vLLM scheduler corrections

KVBM remains the default and its manager implementation is retained, but this PR intentionally corrects three backend-independent vLLM scheduler semantics. The default KVBM replay path is therefore not byte-identical to `main` at these boundaries:

- preemption prefix cost uses the complete known request context (prompt plus retained generated tokens), rather than only the original prompt;
- a fully cached, exactly block-aligned context recomputes its final block to produce logits; and
- a sampled decode token acquires KV only when the next scheduler step computes it, so a terminal dangling token never allocates a slot.

These corrections apply to both KVBM and Native and align the shared scheduler with vLLM v1 behavior.

### Performance and replay parity

Release offline replay of the full 23,608-request Mooncake tool-agent trace at commit `e917a453e9aaa288887d65daf39994d98275bc63`. Both arms use the same release binary and differ only in `g1_backend`. Times are medians from 10 randomized adjacent pairs after 3 warmups per arm.

| Scenario | KVBM replay wall time | Native replay wall time |
| --- | ---: | ---: |
| 1 worker, round-robin, 16,384 blocks | 4.005 s | **1.484 s** |
| 4 workers, KV router, 4,096 blocks/worker | 7.256 s | **4.318 s** |

The deterministic one-worker case is the G1 semantic-parity gate. The following fields matched exactly between KVBM and Native:

| Replay result | KVBM | Native |
| --- | ---: | ---: |
| Completed requests | 23,608 / 23,608 | 23,608 / 23,608 |
| Input tokens | 202,940,084 | 202,940,084 |
| Output tokens | 4,299,817 | 4,299,817 |
| Processed tokens | 207,239,901 | 207,239,901 |
| First-admission prefix-cache reused ratio | 35.931094% | 35.931094% |
| Mean TTFT | 1.397206 ms | 1.397206 ms |
| Mean TPOT | 0.058566 ms | 0.058566 ms |

The four-worker KV-router runs also completed all 23,608 requests and 207,239,901 tokens. Equal-score routing ties produce the expected small cache-reuse variation across measured runs: 35.5218%–35.6718% for KVBM and 35.5611%–35.6875% for Native, so this row is an end-to-end performance diagnostic rather than the exact G1 parity gate.

Configuration: vLLM aggregated offline replay, block size 64, Mooncake trace block size 512, `max_num_seqs=256`, `max_num_batched_tokens=8192`, prefix caching and chunked prefill enabled, `speedup_ratio=1000`, `arrival_speedup_ratio=4`, and G1 only.

### Real vLLM parity validation

Compared real vLLM, AIC + Native G1, and AIC + KVBM G1 using the same 1,000-request Mooncake trace with `MiniMaxAI/MiniMax-M2.5` on 4× B200 at TP=4. The real runtime used vLLM 0.25.1; the mocker used AIC 0.9.0 with its vLLM 0.19.0 B200 profile. All arms used BF16 KV cache, block size 32, `max_num_seqs=256`, `max_num_batched_tokens=8192`, prefix caching, chunked prefill, and `arrival_speedup_ratio=4`.

Real vLLM reported 54,454 total KV blocks and reserves one null block, leaving 54,453 usable blocks. Both mocker arms were configured with the same 54,453 usable blocks to isolate scheduler and block-management parity. No timing calibration was applied (`speedup_ratio=1.0`).

| Metric | Real vLLM | AIC + Native G1 | AIC + KVBM G1 | Native error vs real |
| First-admission cached tokens | 2,916,448 | 2,884,928 | 2,884,928 | -1.08% |
| First-admission cache reuse | 29.7839% | 29.4620% | 29.4620% | -0.3219 pp |
| Mean TTFT | 75,473 ms | 73,620 ms | 73,620 ms | -2.46% |
| Mean TPOT | 129.60 ms | 129.67 ms | 129.67 ms | +0.06% |
| Mean E2E | 98,391 ms | 94,935 ms | 94,935 ms | -3.51% |

All three runs completed exactly 1,000 requests with 9,792,030 input tokens and 199,084 output tokens. Native and KVBM produced identical first-admission cache reuse and effectively identical aggregate latency.

### Full legacy KVBM regression

To validate the shared scheduler changes against the legacy path, I ran a full multi-tier regression against the PR base and candidate revision, with three runs per revision.

Configuration:

- Mooncake 1,000-request trace;
- MiniMax-M2.5/B200 AIC with uncalibrated `speedup_ratio=1.0`;
- legacy KVBM G1, two prefill and two decode workers, and KV routing;
- G1/G2/G3 capacities of 32,768 / 65,536 / 98,304 blocks, with G4 enabled; and
- prefix caching and chunked prefill enabled.

All six runs completed 1,000/1,000 requests with identical workload totals:

| Semantic result | Base | This PR |
| --- | ---: | ---: |
| Completed requests | 1,000 / 1,000 | 1,000 / 1,000 |
| Input tokens | 9,792,030 | 9,792,030 |
| Output tokens | 199,084 | 199,084 |
| Processed tokens | 9,991,114 | 9,991,114 |

All tier directions were exercised across the campaign. Transfer values below are three-run medians in `reservations / blocks`:

| Direction | Base | This PR |
| --- | ---: | ---: |
| G1→G2 store | 10,958 / 293,118 | 11,097 / 289,985 |
| G2→G1 load | 68 / 20,838 | 68 / 23,137 |
| G2→G3 store | 6,618 / 148,843 | 6,145 / 145,735 |
| G3→G2 load | 22 / 4,353 | 26 / 6,155 |
| G2→G4 store | 9,694 / 251,936 | 9,068 / 245,856 |
| G4→G2 load | 2 / 176 | 2 / 144 |

The dominant store block volumes differed by -1.07% for G1→G2, -2.09% for G2→G3, and -2.41% for G2→G4. Restore volume is more sensitive to routing and cache-hit order: reservation medians remained close, while this PR reused 89,824 additional cached tokens (+2.45%), directionally consistent with restoring longer prefixes.

| Three-run median | Base | This PR | Delta |
| --- | ---: | ---: | ---: |
| First-admission cached tokens | 3,663,328 | 3,753,152 | +2.45% |
| Mean TTFT | 16,517.5 ms | 16,149.7 ms | -2.23% |
| Mean TPOT | 13.098 ms | 13.184 ms | +0.66% |
| Mean E2E | 19,400.8 ms | 19,123.6 ms | -1.43% |

Exact cache/timing byte parity is not claimed because production KV-router tie selection and asynchronous offload ordering are not seeded. This PR also intentionally aligns shared scheduler behavior with vLLM v1, which affects the legacy KVBM path. No completion, token-accounting, or material performance regression was observed.

The regression compared base `bdff18f11c` with candidate `a251873436`. The amended review follow-up changes Native G1 plus zero-output-only scheduler gating and test integration; the Mooncake campaign has nonzero outputs and does not exercise that gate.

1. `lib/mocker/src/cache/vllm_block_pool.rs` — leaf physical-capacity, pin/reference, duplicate-copy, and LRU semantics.
2. `lib/mocker/src/kv_manager/vllm_backend.rs` — request ownership, cache finalization, reservations, events, and prefill-cost accounting.
3. `lib/mocker/src/kv_manager/g1_manager.rs` — temporary KVBM/Native composition boundary.
4. `lib/mocker/src/scheduler/vllm/core.rs` and policy tests — scheduler integration, computed-boundary visibility, MTP, and preemption behavior.
5. `lib/mocker/src/kv_manager/vllm_firewall_tests.rs` — enforced dependency direction.

## Validation

- `cargo test -p dynamo-mocker --lib` — 620 passed, 1 ignored.
- `cargo test -p dynamo-mocker --lib --features kvbm-offload` — 717 passed, 1 ignored.
- `cargo clippy -p dynamo-mocker --all-targets --no-default-features -- -D warnings`.
- `PYTHONPATH=components/src dynamo/bin/python -m pytest components/src/dynamo/mocker/tests/unit/test_config.py -q -k native_g1` — 4 passed.
- Release replay campaigns described above; semantic invariants matched in every warmup and measured sample.
- Full legacy KVBM regression described above; all six base/candidate runs completed with exact request and token totals, and all G1–G4 transfer directions were exercised.

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Added configurable G1 backend selection, with KVBM as the default and native mode available for vLLM and TRT-LLM.
  * Added native prefix-cache reuse, capacity tracking, eviction, and KV event support.
  * Added command-line configuration and Python accessors for selecting and viewing the G1 backend.

* **Bug Fixes**
  * Improved prefix-cache accounting, block promotion, reservation handling, and scheduling behavior across supported backends.
  * Added safeguards for invalid backend combinations and incomplete cache blocks.


PeaBrane (commented):
@dreamtalen Thanks for running the Native/KVBM parity and performance comparisons—the main G1 path looks well covered. Since this PR also touches shared scheduler behavior, would it be possible to add one full KVBM regression run with KV routing, disaggregated P/D, and G2/G3/G4 offload enabled together? Byte parity would be ideal, with a brief note for any intentional differences.

(human edit: I guess kvbm offloading is not possible because the new g1 does not support it yet, so perhaps instead with the old kvbm-logical as G1 in place)

lib/mocker/src/kv_manager/vllm_backend.rs
```diff
@@ -0,0 +1,1007 @@
...
+                    _ => panic!("Deref block kind disagrees with request table"),
+                }
+            }
+            owned.split_off(start)
+        };
+        if self.request_blocks[&request_id].is_empty() {
+            self.request_blocks.remove(&request_id);
+        }
+
+        for block in released.into_iter().rev() {
```
PeaBrane, line 685: Could we document this ordering contract here and cover it with a focused eviction-order regression? Because the pool is lineage-agnostic, tail-before-head eviction after a request release depends on this reverse traversal assigning older LRU timestamps to suffix/leaf blocks. The guarantee is procedural rather than intrinsic to the pool, so a future change to `Deref` batching or release order could otherwise alter it.

PeaBrane (commented):
One documentation/coverage note on the native LRU release-order contract.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -0,0 +1,482 @@
...
+pub(crate) struct ReserveOutcome {
+    pub(crate) reservation: BlockReservation,
+    /// Hashes whose final cache-visible physical copy was evicted.
+    pub(crate) removed: Vec<SequenceHash>,
+}
+
+pub(crate) struct VllmBlockPool {
+    capacity: usize,
+    copies: SlotMap<BlockCopyId, BlockCopy>,
+    by_hash: FxHashMap<SequenceHash, Vec<BlockCopyId>>,
```
PeaBrane: Since the common case appears to be one physical copy per hash, this representation allocates a separate `Vec` for every distinct hash. Would an inline-one representation (for example, `SmallVec<[BlockCopyId; 1]>` or an equivalent enum) be worth considering? Non-blocking, but it could matter at replay scale.

lib/mocker/src/kv_manager/vllm_backend.rs
```diff
@@ -0,0 +1,1007 @@
...
+            kv_cache_trace::log_vllm_trace(
+                if is_store { "allocation" } else { "eviction" },
+                self.dp_rank,
+                self.block_size,
+                self.num_active_blocks(),
+                self.num_inactive_blocks(),
+                self.max_capacity(),
+            );
+        }
+        if self.kv_event_publishers.is_empty() {
```
PeaBrane, line 806: When KV-event capture is disabled, the publisher set is fixed and empty, but the Native path has already constructed `PendingStore`, cloned token metadata, and built store-group vectors before reaching this check. Could event materialization be gated earlier while retaining cache finalization and tracing? This is a performance suggestion, not a correctness concern.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -0,0 +1,482 @@
...
+            *refs == 0 && *pins == 0
+        };
+        if should_deactivate {
+            self.insert_inactive(id);
+        }
+    }
+
+    /// Release all unconsumed capacity and prefix pins.
+    pub(crate) fn cancel(&mut self, reservation: BlockReservation) {
+        for (hash, id) in reservation.prefix {
```
PeaBrane: `reservation.prefix` is recorded head-to-tail, and cancellation unpins it in that same order. Because `unpin()` assigns increasing inactive timestamps and eviction removes the oldest entry, an inactive prefix `A → B` can become eligible as parent `A` before leaf `B`. The default KVBM lineage backend intrinsically selects leaves, and ordinary Native dereference explicitly releases in reverse order, so this looks like a Native semantic regression on cancellation. Could cancellation unpin tail-to-head and add a reserve → cancel → capacity-pressure regression through the router indexer? A full disaggregated E2E parity test would also likely catch this if it exercises destination cancellation followed by pressure.

PeaBrane (commented):
Three follow-ups from reviewing Native G1 cancellation semantics and hot-path allocation shape.

dreamtalen:
@PeaBrane Done — I added the full setup, semantic results, tier-transfer counts, and latency comparison under **Full legacy KVBM regression** in the PR description.

TL;DR: base and this PR both completed all three 1,000-request runs with exact request/token totals, and every G1–G4 store/load direction was exercised. Mean TPOT changed by +0.66%, while mean TTFT and E2E improved by 2.23% and 1.43%. Exact cache/timing byte parity is not claimed because KV-router tie selection and asynchronous offload ordering are not seeded.

lib/mocker/src/kv_manager/vllm_backend.rs
dreamtalen: Done in `09f0957`. I documented the tail-first release-order contract in `process_deref()` and added `request_release_evicts_leaf_before_parent` to guard it under capacity pressure. The test verifies that releasing an `A → B` request makes leaf `B` the older eviction candidate while parent `A` remains resident.

lib/mocker/src/cache/vllm_block_pool.rs
dreamtalen: Done in `09f0957`, thanks for the suggestion. `by_hash` now uses `SmallVec<[BlockCopyId; 1]>`, so the common one-copy-per-hash case avoids a separate heap allocation while preserving support for duplicate physical copies.

lib/mocker/src/kv_manager/vllm_backend.rs
dreamtalen: Done in `09f0957`. Store-event materialization is now gated before constructing `PendingStore`, cloning token metadata, or building store-group vectors. Cache finalization still runs unconditionally, and tracing remains enabled through the same gate. I also added `event_enabled_finalization_preserves_store_payload` to cover the event-enabled path.

lib/mocker/src/cache/vllm_block_pool.rs
dreamtalen: Good catch — fixed in `09f0957`. Cancellation now unpins the reserved prefix tail-to-head. I added both a focused pool regression and a router-indexer regression covering reserve → cancel → capacity pressure; they verify that the leaf is evicted before its parent and that the resulting Removed event is applied cleanly.

dreamtalen: Good catch — fixed in `09f0957`. Cancellation now unpins the reserved prefix tail-to-head. I added both a focused pool regression and a router-indexer regression covering reserve → cancel → capacity pressure; they verify that the leaf is evicted before its parent and that the resulting Removed event is applied cleanly.

PeaBrane (approved):
✅

nv-anants (approved):

PR #12034, opened 2026-07-22
perf(mocker): optimize SGLang radix cache lifecycle [DYN-3851]

PeaBrane:
## Summary

Stacked on #12025.

- Compress SGLang mocker radix-cache growth and avoid per-token leaf insertion, repeated prefix materialization, and redundant KV-index copies on scheduler hot paths.
- Add a move-only `ActiveKvLease` that keeps KV indices, the cached-token boundary, and the protected radix tail in one ownership object. `SglangKvManager` now owns extend, finish, abort, and retract transitions so scheduler callers cannot independently free the wrong suffix or unlock an inconsistent path.
- Preserve the optimized layout: the lease has exactly the footprint of the previous three fields and adds no refcount, mutex, heap allocation, or per-token outer-`Option` branch.
- Add regressions for duplicate canonicalization/missing-parent behavior, retained-tail ownership, cache-owned prefix indices, lease layout, and suffix-only retraction.

Reviewer guide: the compressed-radix behavior is in `radix_cache.rs`; ownership and state transitions are concentrated in `sglang_backend.rs`; the remaining scheduler changes replace direct field/free/unlock manipulation with those manager APIs. Upstream SGLang data-structure parity is intentionally out of scope.

## Validation

- Canonical 1,000-request Mooncake replay, 10 alternating paired trials: median wall time fell from `32.205051s` to `19.394082s` (`39.779%` ratio-of-medians improvement, `40.205%` median paired improvement, `1.661x` speedup). Request and token totals were unchanged: 1,000 completions, 9,792,030 input tokens, and 199,084 output tokens.
- Lease-only overhead campaign, 12 alternating paired 50-iteration trials: six wins and six losses, `-0.140%` median paired delta, with all deterministic replay fields matching. Samply showed no lease-only transition among the top 40 self-time entries.
- Samply profiling reduced radix insertion sampled CPU from `2.190s` to `0.227s` and total sampled CPU from `9.889s` to `5.656s` in the local profiling workload.
- `cargo test -p dynamo-mocker`: 553 passed, 1 ignored.
- The repository Python marker hook could not run against the stale local `_core.abi3.so` because it lacks `FrontendExtensionContext`; all other commit hooks passed.

### DIFF
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -5,8 +5,8 @@
 //!
 //! Reference: sglang/python/sglang/srt/mem_cache/radix_cache.py
 
+use rustc_hash::{FxHashMap, FxHashSet};
 use slotmap::{SlotMap, new_key_type};
-use std::collections::{HashMap, HashSet};
 use std::time::Instant;
 
 new_key_type! {
@@ -16,23 +16,32 @@ new_key_type! {
 
 /// Manages free / allocated token slot indices for the KV cache pool.
 pub struct TokenPool {
+    next_fresh: usize,
     free: Vec<usize>,
     total: usize,
 }
 
 impl TokenPool {
     pub fn new(total: usize) -> Self {
-        let free: Vec<usize> = (0..total).rev().collect();
-        Self { free, total }
+        Self {
+            next_fresh: 0,
+            free: Vec::new(),
+            total,
+        }
     }
 
     /// Allocate `n` token slots. Returns `None` if not enough free slots.
     pub fn allocate(&mut self, n: usize) -> Option<Vec<usize>> {
-        if self.free.len() < n {
+        if self.available() < n {
             return None;
         }
-        let start = self.free.len() - n;
-        let indices: Vec<usize> = self.free.drain(start..).collect();
+
+        let recycled = n.min(self.free.len());
+        let fresh = n - recycled;
+        let mut indices = Vec::with_capacity(n);
+        indices.extend(self.free.drain(self.free.len() - recycled..));
+        indices.extend(self.next_fresh..self.next_fresh + fresh);
+        self.next_fresh += fresh;
         Some(indices)
     }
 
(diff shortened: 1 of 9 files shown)


tedzhouhk:
One semantic concern worth calling out is that this builds on a pre-existing limitation in the mock radix cache: it does not maintain one node per KV block/page. A node may contain multiple page-aligned blocks, while `evict(n)` removes an entire leaf node. As a result, eviction can already overshoot the requested token count depending on the incidental compressed-tree shape.

This PR makes that issue more pronounced for incrementally materialized requests. Previously, re-walking from the root would encounter the locked retained prefix and append newly materialized pages as a child. With `insert_from_node()`, a uniquely locked retained tail is extended in place, removing that node boundary.

For example, with `page_size = 4`, let request A materialize `[1,2,3,4]` and then grow to `[1,2,3,4,5,6,7,8]`:

```text
before: root -> [1,2,3,4] -> [5,6,7,8]
after:  root -> [1,2,3,4,5,6,7,8]
```

After A finishes and releases its lease, `evict(4)` removes only `[5,6,7,8]` in the former shape, but removes all eight tokens in the latter because eviction operates on whole leaves. The first four tokens therefore remain reusable before this change but disappear after it. This can change subsequent prefix reuse, scheduling, TTFT, and total replay duration even when request and token totals still match.

Compressed radix nodes are reasonable; the concern is coupling eviction granularity to incidental node boundaries. Could we either evict/trim a leaf at page granularity, or preserve the page boundary when extending a retained tail? We should also add a parity regression that covers post-eviction prefix reuse, rather than validating only request/token totals.

PeaBrane:
> Could we either evict/trim a leaf at page granularity, or preserve the page boundary when extending a retained tail?

Addressed in [`fe022b8dfa`](https://github.com/ai-dynamo/dynamo/commit/fe022b8dfa2ce7efb2fc10bf9bac43f9dc070049). Eviction now trims only the page-aligned suffix of a compressed leaf, keeps the retained prefix indexed, and returns/publishes removal for only the evicted suffix indices. The partial path reuses the existing leaf/value storage, so retained-tail insertion keeps its compressed shape without adding another eviction allocation.

The new lifecycle regression materializes 4 tokens, extends the retained leaf to 8, releases the lease, evicts 4, and verifies local prefix reuse falls from 8 to 4 while router overlap falls from 2 blocks to 1 with exactly one Removed block. Full `dynamo-mocker` suite: 556 passed, 1 ignored.

PR #12036, opened 2026-07-22
fix(frontend): return 4xx for Backend(InvalidArgument) on streaming c…

ritazh:
…hat/responses

## Overview:

A /v1/chat/completions or /v1/responses request with `stream: true` that hits a synchronous backend InvalidArgument at request-parse time (before any token is generated) currently emits:
  - HTTP 200 (stream opens)
  - SSE data frame: {"error":{"message":"Internal server error","type":"internal_server_error","code":500}}
  - data: [DONE]

The identical request with `stream: false` correctly returns HTTP 400 with the typed message, because the non-streaming branch already calls `check_for_backend_error()` before folding the stream for chat_completions and responses. The streaming branches skip this check and rely on `monitor_for_disconnects_with_timeout()`, which hardcodes `SanitizedError::Internal` for any Err and loses the `Backend(InvalidArgument)` classification.

Reproducible with zai-org/GLM-5.2 (text-only) + image_url content + `stream: true`: the model correctly refuses (ValueError: Received multimodal data but multimodal processing is not enabled), then returns Backend(InvalidArgument), but the streaming SSE frame still surfaces `code: 500`.

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming chat completion and response requests now return the correct HTTP error status when a backend error occurs before streaming begins.
  * Prevented initial backend errors from being incorrectly delivered as server-sent events after a successful response has already started.


rmccorm4:
Hi @ritazh, thanks for the contribution! 

1. Can you take a look into the failing test case: https://github.com/ai-dynamo/dynamo/actions/runs/29952953360/job/89035017580?pr=12036?
2. Can you check out the open review comments from Devin, dynamo-review-agent, and Coderabbit?

lib/llm/src/http/service/openai.rs
```diff
@@ -1962,11 +1962,24 @@ async fn chat_completions(
...
+        // `Backend(InvalidArgument)` from a text-only model receiving image
+        // content, or from an unsupported JSON-schema keyword), we can still
+        // return the correct 4xx status + typed body — same code path as the
+        // non-streaming path. Once HTTP 200 is committed the
+        // downstream `monitor_for_disconnects` fallback can only report errors
+        // as a hardcoded generic 500 SSE frame, so catching here is the only
+        // way to preserve `Backend(InvalidArgument)` -> HTTP 400 on the
+        // streaming path.
         stream_handle.arm(); // allows the system to detect client disconnects and cancel the LLM generation
+        let stream = check_for_backend_error(stream)
```
jh-nv: Please see the AI comment regarding the possible scenario that backend never returns and address it.

lib/llm/src/http/service/openai.rs
```diff
@@ -2441,10 +2454,21 @@ async fn responses(
...
-        // must be delivered as SSE events in the stream. This is standard SSE behavior.
+        // Peek at the first non-annotation event before committing HTTP 200: if
+        // the engine surfaces a synchronous backend error at t=0 (e.g.
+        // `Backend(InvalidArgument)` from a text-only model receiving image
+        // content, or from an unsupported JSON-schema keyword), return the
+        // correct 4xx status + typed body — same as the non-streaming path.
+        // Once HTTP 200 is committed the downstream error frame is a hardcoded generic 500.
         stream_handle.arm(); // allows the system to detect client disconnects and cancel the LLM generation
+        let engine_stream =
+            check_for_backend_error(engine_stream)
```
jh-nv: This, and also the one  at line 1976.

the new fallible preflight check introduces an early-return path after the handle has been armed.

When check_for_backend_error(...) returns a typed 4xx, ? exits the handler and drops the armed handle. ConnectionHandle::drop() then reports ClosedUnexpectedly, which can incorrectly count the intentional validation response as a client disconnect/cancellation and kill the context.

Please either move stream_handle.arm() until after the preflight check succeeds, or disarm it before propagating the detected HTTP error.

ritazh:
Addressed review comments:
- replaced pre-peek `stream_handle.arm()` + `check_for_backend_error(...).await` with new s`treaming_preflight_peek` helper: short 50ms `tokio::select!` window, `Pin<Box<dyn Stream>>` return so no borrowed lifetimes leak into `Sse::new;` falls through to SSE on timeout and lets `monitor_for_disconnects` own the long safety net. Extracted `backend_error_response()` shared with `check_for_backend_error`.

rmccorm4:
/ok to test [59adeaf](https://github.com/ai-dynamo/dynamo/pull/12036/commits/59adeaf82f8e2cffca3a7f85de485dabf5c20150)

rmccorm4:
/ok to test [b29377d](https://github.com/ai-dynamo/dynamo/pull/12036/commits/b29377d01b1bd198ac5078536d5107a3bef5f37d)

lib/llm/src/http/service/openai.rs
```diff
@@ -1739,6 +1727,84 @@ pub(super) async fn check_for_backend_error(
...
+                code: status_code.as_u16(),
+                details: None,
+            }),
+        ),
+    }
+}
+
+/// Short pre-commit peek window. synchronous validation/backend errors should surface quickly,
+/// and the short window avoids delaying SSE headers for normal high-TTFT requests.
+const PRE_COMMIT_ERROR_PEEK_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(50);
```
jh-nv: I am concerned about this windowed check here, as it will make the response time dependent, rather than deterministic, and error that comes at 49ms vs 51 ms, changes what client sees. 

I think there is a deeper architecture design change needed to fix this. we need to separate synchronous request validation/admission from asynchronous token generation. 

If this fix is urgently needed, let's make the value configurable and pick the default value from your observed p99 validation/admission latency. and clearly document the behavior. 

@rmccorm4 please also chime in.

GuanLuo: Seconding @jh-nv 's comment. For streaming, is there a particular reason to surface the inference-time error to the HTTP status instead of as a SSE error event?

The example error is related to "business logic", so I think the proper resolution is to keep such error as SSE event, but fixing the error message to reflect the real cause instead of just "internal server error". In such way, we keep the HTTP status to reflect the connection status.

ritazh: Thanks for the review!

Re: the peek window concern: you're right that a 50ms boundary trades determinism for pragmatism. The pre-fix state was deterministic in the wrong direction: every sync backend error surfaced as HTTP 200 + `{"code":500,"message":"Internal server error"}` with the real cause sanitized away. Happy to take your fallback ask: I'll make the window env-tunable (`DYN_HTTP_PRE_COMMIT_ERROR_PEEK_MS`, default 50).

Full agree that separating synchronous validation/admission from async generation is the right long-term fix.

Re: HTTP status vs SSE event: The error class we're catching is request validation (text-only model + image_url, unsupported JSON-schema keyword, etc.), not runtime; same class as malformed JSON, which returns HTTP 400. Three reasons to prefer HTTP status:
  1. The non-streaming path already returns HTTP 400 for this exact case; flipping stream: true/false on the same body shouldn't change the error contract.
  2. Load-balancer / retry / circuit-breaker layers classify by HTTP status. Wrapping validation errors as HTTP 200 forces every intermediary to deep-inspect SSE payloads.
  3. OpenAI's own streaming API returns 4xx pre-stream and 200 + SSE for successful streams. Dynamo diverging breaks clients built against that contract.

I'd like to keep the pre-commit peek so validation errors that we CAN catch before headers go out surface with the correct status.

lib/llm/src/http/service/openai.rs
```diff
@@ -1739,6 +1727,117 @@ pub(super) async fn check_for_backend_error(
...
+            )),
+        }
+    });
+    *CACHED
+}
+
+/// Streaming-path preflight peek. Polls the stream for a bounded window
+/// (`DYN_HTTP_PRE_COMMIT_ERROR_PEEK_MS`, default 10ms) buffering leading
+/// annotation frames.
+async fn streaming_preflight_peek(
```
GuanLuo: This function seems to be an enhanced version of `check_for_backend_error`, would it be possible to merge them into a single function to be used in both non-streaming and streaming case?

lib/llm/src/http/service/openai.rs
```diff
@@ -1739,6 +1727,117 @@ pub(super) async fn check_for_backend_error(
...
+fn pre_commit_error_peek_timeout() -> Option<std::time::Duration> {
+    use std::sync::LazyLock;
+    static CACHED: LazyLock<Option<std::time::Duration>> = LazyLock::new(|| {
+        match std::env::var(env_llm::DYN_HTTP_PRE_COMMIT_ERROR_PEEK_MS)
+            .ok()
+            .and_then(|s| s.parse::<u64>().ok())
+        {
+            Some(0) => None,
+            Some(ms) => Some(std::time::Duration::from_millis(ms)),
+            None => Some(std::time::Duration::from_millis(
```
GuanLuo: Can `None => None`? So the default behavior is the same as previously (open SSE stream without delay)

lib/llm/src/http/service/openai.rs
```diff
@@ -1739,6 +1727,117 @@ pub(super) async fn check_for_backend_error(
...
+
+/// Read the pre-commit peek window from the environment.
+///
+/// `Some(dur)` — poll for that duration before committing SSE.
+/// `None` — the peek is disabled entirely (all errors surface as SSE frames
+/// post-HTTP-200).
+///
+/// Cached in a `LazyLock` because it's on the streaming hot path and env
+/// values are fixed at process start.
+fn pre_commit_error_peek_timeout() -> Option<std::time::Duration> {
```
GuanLuo, line 1791: Okay to put here for now, but can you add FIXME for unifying env var initialization along with other env vars (which hasn't been standerized)

GuanLuo (approved):

ritazh:
@GuanLuo addressed your nits. PTAL. Good to merge?

GuanLuo (approved):

jh-nv (approved):

PR #12043, opened 2026-07-22
perf(mocker): streamline SGLang radix scheduling [DYN-3851]

PeaBrane:
## Summary

- Keep each SGLang request in one append-only `u32` token sequence so decode caching borrows the retained prefix instead of rebuilding prompt plus output on every completed page.
- Use native `u32` radix keys and router hash inputs, append token-pool allocations directly into active leases, consume decode reservations with a vector cursor, release suffixes through borrowed slices, and sample decode bursts inline.
- Skip redundant release/reacquire walks when a retained radix node is unchanged while preserving path reference counting for interleaved requests that share prefixes.
- Validate pool capacity before reserving allocation memory and cover failed append atomicity plus exact, unique release of partially consumed decode reservations.

Reviewer guide: `request.rs` and `radix_cache.rs` contain the representation changes; `sglang_backend.rs` contains the slot and lease lifecycle; `decode.rs` contains the hot-path allocation changes. The remaining changes adapt scheduler call sites and regression fixtures.

Non-goals: this does not remove radix path reference counts, change canonical page ownership, alter eviction recency semantics, or convert `ActiveKvLease` to a tail-only representation.

## Validation

- Canonical 1,000-request Mooncake replay, four SGLang workers with KV-aware routing, eight alternating 50-iteration pairs on a pinned physical CPU core: median wall time moved from `10.676775s` to `7.380573s` (**30.873% lower**, **1.4466x**), with the optimized candidate winning 8/8 pairs. Baseline CV was 0.617%; optimized CV was 0.504%.
- Deterministic correctness reports matched for request counts, token counts, prefix reuse, simulated duration, worker time, and GPU-hour accounting.
- `cargo test -p dynamo-mocker`: 561 passed, 1 ignored, 0 failed.
- `cargo fmt --all -- --check` and `git diff --check` passed.

## Summary by CodeRabbit

- **Improvements**
  - Improved token and KV-cache handling for more efficient request processing.
  - Reduced unnecessary memory allocations during token allocation, caching, and request completion.
  - Improved handling of partially consumed reservations and cache cleanup.

- **Bug Fixes**
  - Corrected release behavior when only part of a decode reservation is used.
  - Prevented redundant cache reference updates during request finalization.

- **Tests**
  - Added coverage for allocation failures, atomic behavior, and partial reservation cleanup.

### DIFF
--- a/lib/mocker/src/scheduler/sglang/decode.rs
+++ b/lib/mocker/src/scheduler/sglang/decode.rs
@@ -166,11 +166,7 @@ pub(super) fn cleanup_completed_request(
         return;
     }
     let lease = std::mem::take(&mut request.kv_lease);
-    if tokens_to_cache <= request.prompt_len() {
-        kv_manager.finish(&request.prompt_tokens()[..tokens_to_cache], lease);
-    } else {
-        kv_manager.finish(request.sequence_prefix(tokens_to_cache), lease);
-    }
+    kv_manager.finish(request.sequence_prefix(tokens_to_cache), lease);
 }
 
 pub(super) fn simulate_decode_step_with_sampler(
--- a/lib/mocker/src/scheduler/sglang/decode.rs
+++ b/lib/mocker/src/scheduler/sglang/decode.rs
@@ -63,12 +63,8 @@ pub(super) fn cache_materialized_prefix(
         );
     }
 
-    if aligned_tokens <= req.prompt_tokens.len() {
-        kv_manager.extend_cached_prefix(&req.prompt_tokens[..aligned_tokens], &mut req.kv_lease);
-    } else {
-        let sequence = req.sequence_prefix(aligned_tokens).into_owned();
-        kv_manager.extend_cached_prefix(&sequence, &mut req.kv_lease);
-    }
+    let sequence = &req.sequence_tokens[..aligned_tokens];
+    kv_manager.extend_cached_prefix(sequence, &mut req.kv_lease);
     req.debug_assert_invariants(config.block_size);
 }
 
@@ -171,10 +167,9 @@ pub(super) fn cleanup_completed_request(
     }
     let lease = std::mem::take(&mut request.kv_lease);
     if tokens_to_cache <= request.prompt_len() {
-        kv_manager.finish(&request.prompt_tokens[..tokens_to_cache], lease);
+        kv_manager.finish(&request.prompt_tokens()[..tokens_to_cache], lease);
     } else {
-        let sequence = request.sequence_tokens().into_owned();
-        kv_manager.finish(&sequence[..tokens_to_cache], lease);
+        kv_manager.finish(request.sequence_prefix(tokens_to_cache), lease);
     }
 }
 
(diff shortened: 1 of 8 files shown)


lib/mocker/src/scheduler/sglang/decode.rs
```diff
@@ -290,23 +285,18 @@ pub(super) fn simulate_decode_step_with_sampler(
...
-                remaining.min(1)
-            } else if let Some(sampler) = sampler.as_deref_mut() {
-                sampler.sample_output_tokens(remaining)
-            } else {
-                remaining.min(1)
-            }
-        })
-        .collect::<Vec<_>>();
-    output_signals.reserve(sampled_bursts.iter().copied().sum::<usize>());
+    output_signals.reserve(running.len());
```
dreamtalen, line 284: This only reserves one signal per request, but MTP can emit up to `max_burst` signals per request. Shall we use `output_signals.reserve(reserved_tokens) `here instead?

dreamtalen (approved):
one comment on output_signals capacity, otherwise LGTM

lib/mocker/src/scheduler/sglang/decode.rs
PeaBrane: Good point. I am keeping `running.len()` intentionally here. `reserved_tokens` is the worst-case KV safety budget, not the expected number of emitted signals: with `aic_nextn=5` it can reserve six `OutputSignal`s per running request even when acceptance emits only one. Since `OutputSignal` is substantially larger than the removed `usize` burst entry, reserving that upper bound would systematically increase per-pass memory. The current reservation covers the one-token common case and lets speculative acceptance grow the vector amortized; growth is bounded because the burst is capped at six. If profiling later shows this reallocation is material, we can revisit an expected-size or reusable-buffer strategy without reserving the worst case on every pass.

lib/mocker/src/scheduler/sglang/decode.rs
```diff
@@ -171,10 +167,9 @@ pub(super) fn cleanup_completed_request(
     }
     let lease = std::mem::take(&mut request.kv_lease);
     if tokens_to_cache <= request.prompt_len() {
-        kv_manager.finish(&request.prompt_tokens[..tokens_to_cache], lease);
+        kv_manager.finish(&request.prompt_tokens()[..tokens_to_cache], lease);
```
jthomson04: Now that `SglangRequest` stores the entire sequence contiguously, both branches pass the same prefix. Drop the branch and call `kv_manager.finish(request.sequence_prefix(tokens_to_cache), lease)` directly.

jthomson04 (approved):

lib/mocker/src/scheduler/sglang/decode.rs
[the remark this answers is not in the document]
PeaBrane: Keeping `running.len()` intentionally. `reserved_tokens` is the worst-case KV safety budget, not the expected signal count; with speculative decoding it can reserve up to six comparatively large `OutputSignal`s per request even when acceptance emits one. The current reservation covers the common case and lets the bounded accepted burst grow amortized, avoiding systematic worst-case memory overhead.

lib/mocker/src/scheduler/sglang/decode.rs
PeaBrane: Fixed in 6cd1405ab6: completion now passes `request.sequence_prefix(tokens_to_cache)` directly.

PR #12049, opened 2026-07-23
fix(mocker): bound offline replay memory growth

tedzhouhk:
## Summary

- Replace the unbounded global ITL and per-user token-rate sample vectors with bounded DDSketch distributions at 0.1% relative accuracy.
- Fold token gaps into the sketches when a request terminates, then retain only the first, second, and last token timestamps plus output length.
- Keep full-prompt trace requests compact while they wait in aggregated and disaggregated prefill queues, materializing tokens only at worker submission for source-first replay.
- Preserve exact request counts, routing inputs, per-request JSONL fields, TTFT, TTST, TPOT, E2E, mean, min, max, and standard deviation behavior; only the two global quantile families become approximate.

## Details

The DDSketch stores are capped at 32,768 bins, covering roughly 10^28 same-sign dynamic range while bounding the two global sketches to about 1 MiB total. Mean and variance use Welford accumulation.

Compact prompts still materialize transiently during workload setup to preserve the existing engine-block hash semantics. That temporary vector is dropped immediately; the queued payload retains only input length, request metadata, and trace hash IDs.

Aggregated replay and source-first vLLM disaggregated replay now keep that compact payload through router and worker-capacity waits. Disaggregated prefill expands the prompt at the prefill worker submission boundary. Destination-first SGLang still expands at decode reservation before prefill; an inline TODO records the follow-up needed to keep decode routing and reservation compact as well.

- `lib/mocker/src/replay/collector.rs` for bounded distributions and terminal compaction.
- `lib/mocker/src/loadgen/types.rs` and `lib/mocker/src/loadgen/driver.rs` for deferred prompt payloads.
- `lib/mocker/src/replay/offline/agg.rs` for aggregated replay integration.
- `lib/mocker/src/replay/offline/disagg.rs` for source-first prefill deferral and the decode-side follow-up boundary.

## Validation

- `cargo test -p dynamo-mocker --lib` — 586 passed, 1 ignored.
- `cargo clippy -p dynamo-mocker --all-targets -- -D warnings`.
- `cargo test -p dynamo-mocker --lib --features aic-forward-pass --no-run`.
- Release Rust offline replay parity for the original memory changes against base `1bf8080969` and candidate `4345a8f900`:
  - 5,000-request aggregated vLLM and SGLang traces, repeated per revision.
  - Aggregated/disaggregated vLLM and SGLang canonical reports matched after excluding wall-clock rates, floating reduction-order noise, and the intentionally approximate quantiles.
  - Maximum observed DDSketch quantile relative error was 0.0954%, below the configured 0.1%.
  - Pre-existing cross-process aggregated vLLM routing nondeterminism was reproduced on both base and candidate; matching routing outcomes had identical canonical digests.
- Incremental release Rust disaggregated replay comparison between `4345a8f900` and `afdc754a57`:
  - 100-request vLLM and SGLang traces, two runs per revision.
  - Request counts, token totals, simulated duration, routing reuse, throughput, and latency distributions matched; differences were limited to wall-clock rates and last-bit floating reduction noise.
- Multi-pass agentic/Graham review completed with no remaining P0-P3 findings.

Performance timing was not treated as authoritative because another replay campaign was active on the host.

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12049" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


lib/mocker/src/replay/collector.rs
```diff
@@ -352,6 +359,104 @@ struct TraceRequestStats {
...
+
+        self.sketch.add(value);
+        self.count += 1;
+        let delta = value - self.mean;
+        self.mean += delta / self.count as f64;
+        let delta_after_mean_update = value - self.mean;
+        self.sum_squared_deviations += delta * delta_after_mean_update;
+        self.min = self.min.min(value);
+        self.max = self.max.max(value);
+    }
```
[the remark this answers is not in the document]
tedzhouhk: Addressed in 6838704462. Added `streaming_distribution_preserves_all_zero_samples`, which inserts 128 zero-valued samples, verifies DDSketch records them in the zero bucket, and exercises `finish()` to confirm mean/min/max/std and every reported percentile remain exactly 0.0 without an error or panic. Full `dynamo-mocker` lib suite and strict clippy pass.

lib/mocker/src/replay/collector.rs
```diff
@@ -352,6 +359,104 @@ struct TraceRequestStats {
...
+        let quantile = if span == 0 || rank >= span {
+            1.0
+        } else {
+            (rank as f64 + 0.5) / span as f64
+        };
+        match self.sketch.quantile(quantile) {
+            Ok(value) => value,
+            Err(error) => panic!("invalid built-in DDSketch quantile {quantile}: {error}"),
+        }
+    }
```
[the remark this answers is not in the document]
tedzhouhk: Addressed in 6838704462. Added `streaming_percentiles_select_the_historical_rounded_rank`, covering 10 sample sizes (including even/odd and boundary-adjacent sizes) and 12 percentiles from 0 through 100. Each result is checked against the historical `percentile_rank` selection within the configured DDSketch 0.1% relative-error bound, confirming ddsketchy's floored-rank convention matches this mapping. Full `dynamo-mocker` lib suite and strict clippy pass.

lib/mocker/src/replay/collector.rs
```diff
@@ -542,18 +660,51 @@ impl TraceRequestStats {
...
+        &mut self,
+        include_in_distributions: bool,
+        itl_distribution: &mut StreamingDistribution,
+        output_token_throughput_per_user: &mut StreamingDistribution,
+    ) {
+        let TokenTimeline::Recording(times) = &self.token_timeline else {
+            return;
+        };
+
+        if include_in_distributions {
```
PeaBrane, line 689: Non-blocking: this moves O(output length) DDSketch folding onto live replay’s single output demux. For long completions, worker outputs continue accumulating on the unbounded channel while this loop runs, and later batches receive timestamps only when dequeued, so collector work can inflate unrelated token/terminal timings. Could we move or chunk this aggregation off the demux critical path, or timestamp outputs before queueing?

lib/mocker/src/loadgen/driver.rs
```diff
@@ -715,15 +725,37 @@ impl WorkloadDriver {
...
+                        input_length,
+                        hash_ids,
+                        self.trace_block_size,
+                    );
+                    // The router needs engine-block hashes at arrival, but it
+                    // does not need to retain the expanded prompt. Materialize
+                    // once transiently for hashing, then keep only the compact
+                    // payload until a worker admission.
                     let replay_hashes = self.include_replay_hashes.then(|| {
+                        let request_tokens = request.prompt_tokens();
```
PeaBrane, line 760: Non-blocking: this expands the entire prompt to compute replay hashes and then drops the vector; aggregated immediate dispatch expands the same deferred payload again in `into_direct_request`. Compared with the base, that adds one O(ISL) allocation/fill per KV-routed request, including the default no-queue case. Could the replay hashes be derived incrementally from the compact trace blocks instead?

PeaBrane (approved):
These are AI comments.

lib/mocker/src/loadgen/driver.rs
tedzhouhk: added a TODO for this perf optimization, not sure if router can just use mooncake hash (or hash again from mooncake hash)

lib/mocker/src/replay/collector.rs
tedzhouhk: it would need to fallback to original behavior (continuously growing memory in live mocker) to address this. Another way is to do metrics in a separate thread (too big for this PR). Since live mocker is usually not perf critical, I just keep the current behavior and added a todo

nv-anants (approved):
toml change look good!

PR #12051, opened 2026-07-23
feat(media): explicit installer for additional media decoders (OPS-7795)

dmitry-tokarev-nv:
## Summary

Explicit, operator-run installer for the media-decoder packages the runtime images deliberately omit — reworked from the earlier env-var design per review feedback (@rmccorm4, Neelay): **nothing installs implicitly**.

```bash
python -m dynamo.common.utils.install_media_decoders <vllm|sglang|trtllm>
```

Linear OPS-7795

## Details

- **Env-var pathway removed completely.** `DYN_ENABLE_MEDIA_DECODERS`, `DYN_MEDIA_DECODER_PACKAGES`, `DYN_MEDIA_DECODER_PIP_ARGS`, `DYN_MEDIA_DECODER_TIMEOUT_S` and the startup hooks in the vLLM / SGLang / TRT-LLM / omni entrypoints are gone. The former overrides are CLI flags: `--packages`, `--pip-args`, `--timeout-s`, `--dry-run`. A sweep test fails if any `__main__.py` ever calls the installer again or the module reads the environment.
- **Versions no longer float** (raised by @saturley-hall). Each package installs at a validated lower bound with the next major capped:
  | Package | Spec | Validated by |
  |---|---|---|
  | opencv-python-headless | `>=4.13.0.92,<5` | GPU serve tests (vLLM/TRT-LLM video) |
  | decord2 | `>=3.4.0,<4` | GPU serve tests (SGLang video) |
  | av | `>=18.0.0,<19` | last full-suite image generation (vLLM audio) |

  The cap is not hypothetical: `opencv-python-headless` 5.x is already on PyPI, so a floating install today pulls a major nothing has validated. The serve tests now install from the same `VALIDATED_SPECS` constant, so tests exercise exactly what the docs prescribe.
- **Failure semantics flip with ownership**: the old startup hook swallowed errors to protect the worker; an explicit admin command exits non-zero when pip fails or the module still does not import afterwards.
- Kept from the previous review round: idempotency (skip when the module already imports), cross-process file lock, `--no-deps` default so the image's pinned stack (numpy under torch/vLLM) cannot shift, credential redaction in logged pip commands, bounded pip timeout.
- **Docs**: the Additional Media Decoders page documents the explicit workflow — installer, plain-pip equivalents at the same bounds, a baked image layer as the recommended Kubernetes shape, pod-command form for quick evaluation, and air-gapped usage (`--pip-args '--no-index --find-links /wheels'`). Per @rmccorm4, the validated pip install commands are also added to the Video Decode GPU Requirements page next to the NVDEC fallback discussion.

`components/src/dynamo/common/utils/media_decoders.py` (the whole design in one module), then `docs/fern/pages/use-cases/multimodal-serving/additional-media-decoders.md` for the operator-facing story.

## Validation

- **Real-image validation, 17/17 on all three runtime images** (`.codec-audit/validate_media_decoders_real.sh`, x86_64 GPU host, real pip, no mocks): dry-run lists the bounded specs; real install succeeds (`cv2 4.14.0.94`, `av 18.0.0`, `decord2 3.4.0` — all inside bounds); each carrier **decodes the baked H.264 fixture** (10/10 frames via cv2, av, and decord — import alone proves too little); trtllm install does not pull in `av` (scope check); second run is a no-op; and two must-fail controls (`--pip-args=--no-index` with no wheelhouse, impossible pin) exit non-zero for the right reason.
- The real-image pass caught two bugs the mocked unit suite structurally could not, both fixed in `4088cd6fa7`: post-install verification now runs in a **fresh interpreter** (as the images' non-root user, pip defaults to user-site installs, and a user-site dir created after the parent interpreter started is never on its `sys.path`), and `--pip-args` is documented in the `=` form (argparse rejects a spaceless leading-dash value as a separate token).
- 29/29 unit tests (`components/src/dynamo/common/tests/test_media_decoders.py`): version-bound invariants, per-backend install sets, only-missing/idempotent/racing-process behavior, `--no-deps` default vs override-with-deps, failure raises, CLI exit codes, an unmocked fresh-probe test, and the no-implicit-install sweep.
- `ruff format`/`ruff check` (v0.5.2) clean; nav YAML parses; branch rebased onto current main (post codec-removal merge).

## Related Issues

Relates to #11836 (codec removal, merged)

### DIFF
--- a/components/src/dynamo/common/tests/test_media_decoders.py
+++ b/components/src/dynamo/common/tests/test_media_decoders.py
@@ -22,11 +22,19 @@ pytestmark = [pytest.mark.pre_merge, pytest.mark.unit, pytest.mark.gpu_0]
 
 @pytest.fixture
 def sandboxed(monkeypatch):
-    """Stub out the lock and import-cache rewrite for unit runs."""
+    """Stub out the lock, cache rewrite, and fresh-interpreter probe."""
     monkeypatch.setattr(
         media_decoders, "_cross_process_lock", lambda: contextlib.nullcontext()
     )
     monkeypatch.setattr(media_decoders.importlib, "invalidate_caches", lambda: None)
+    # Route the fresh-interpreter probe through _module_available so one stub
+    # controls both, and the probe never spawns a real python (which would
+    # also collide with the subprocess.run recorders below).
+    monkeypatch.setattr(
+        media_decoders,
+        "_modules_missing_fresh",
+        lambda mods: [m for m in mods if not media_decoders._module_available(m)],
+    )
     return monkeypatch
 
 
@@ -273,6 +281,21 @@ def test_pending_subset_installs_only_still_missing(sandboxed):
     assert media_decoders.VALIDATED_SPECS["opencv-python-headless"] not in cmd
 
 
+def test_modules_missing_fresh_real_probe():
+    """Unmocked: the fresh-interpreter probe distinguishes real modules.
+
+    The probe exists because pip may install into a user-site directory
+    created after the parent interpreter started; only a fresh interpreter
+    (like the worker launched later) is guaranteed to see it. Exercise the
+    real subprocess path here since every other test stubs it out.
+    """
+    missing = media_decoders._modules_missing_fresh(
+        ["json", "definitely_not_a_module_xyz"]
+    )
+    assert missing == ["definitely_not_a_module_xyz"]
+    assert media_decoders._modules_missing_fresh([]) == []
+
+
 def test_redact_masks_url_credentials():
     line = "pip install --index-url https://user:secret@pypi.corp/simple pkg"
     masked = media_decoders._redact(line)
@@ -328,6 +351,18 @@ def test_cli_packages_override_reaches_pip(sandboxed):
     (cmd,) = calls
     assert "av==18.0.0" in cmd
     assert "--no-deps" not in cmd
+
+
+def test_cli_pip_args_equals_form_single_flag(sandboxed):
+    """--pip-args=--no-index must parse: a spaceless leading-dash value as a
+    separate token is an argparse usage error, so the docs teach the = form.
+    Found on a real image: the separate-token form exited 2, not 1."""
+    present: set[str] = set()
+    _set_available(sandboxed, present)
+    calls = _record_pip_and_mark(sandboxed, present, "decord")
+    assert media_decoders.main(["sglang", "--pip-args=--no-index"]) == 0
+    (cmd,) = calls
+    assert "--no-index" in cmd
 
 
 def test_cli_pip_args_reach_pip(sandboxed):
--- a/components/src/dynamo/common/tests/test_media_decoders.py
+++ b/components/src/dynamo/common/tests/test_media_decoders.py
@@ -1,0 +1,387 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the explicit media-decoder installer.
+
+The pip install is fully mocked -- these tests never touch the network or the
+real site-packages.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from dynamo.common.utils import media_decoders
+
+pytestmark = [pytest.mark.pre_merge, pytest.mark.unit, pytest.mark.gpu_0]
+
+
+@pytest.fixture
+def sandboxed(monkeypatch):
+    """Stub out the lock and import-cache rewrite for unit runs."""
+    monkeypatch.setattr(
+        media_decoders, "_cross_process_lock", lambda: contextlib.nullcontext()
+    )
+    monkeypatch.setattr(media_decoders.importlib, "invalidate_caches", lambda: None)
+    return monkeypatch
+
+
+def _record_pip(monkeypatch, *, fail: bool = False) -> list[list[str]]:
+    """Replace subprocess.run with a recorder; return the list of commands."""
+    calls: list[list[str]] = []
+
+    def fake_run(cmd, check=False, **kwargs):
+        calls.append(list(cmd))
+        if fail:
+            raise subprocess.CalledProcessError(1, cmd)
+        return subprocess.CompletedProcess(cmd, 0)
+
+    monkeypatch.setattr(media_decoders.subprocess, "run", fake_run)
+    return calls
+
+
+def _record_pip_kwargs(monkeypatch) -> list[dict]:
+    """Replace subprocess.run with a kwargs recorder (for timeout checks)."""
+    calls: list[dict] = []
+
+    def fake_run(cmd, check=False, **kwargs):
+        calls.append({"cmd": list(cmd), **kwargs})
+        return subprocess.CompletedProcess(cmd, 0)
+
+    monkeypatch.setattr(media_decoders.subprocess, "run", fake_run)
+    return calls
+
+
+def _set_available(monkeypatch, present) -> None:
+    """Stub _module_available; `present` is a set of importable module names."""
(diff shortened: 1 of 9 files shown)


dagil-nvidia (commented):
Where is this GPU requirement doc? And if it does exist, please make sure to add a link to it

saturley-hall (approved):

components/src/dynamo/common/tests/test_media_decoders.py
```diff
@@ -0,0 +1,269 @@
...
+    ):
+        monkeypatch.delenv(key, raising=False)
+    media_decoders._completed.clear()
+    # Never take a real file lock or rewrite import caches in unit tests.
+    monkeypatch.setattr(
+        media_decoders, "_cross_process_lock", lambda: contextlib.nullcontext()
+    )
+    monkeypatch.setattr(media_decoders.importlib, "invalidate_caches", lambda: None)
+    yield monkeypatch
+    media_decoders._completed.clear()
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Overtaken by the redesign: all four `DYN_MEDIA_DECODER_*` env vars (including `TIMEOUT_ENV`) were removed entirely — timeout is now the `--timeout-s` CLI flag and the fixture no longer touches the environment. Resolving as moot.

components/src/dynamo/common/utils/install_media_decoders.py
```diff
@@ -0,0 +1,320 @@
...
+
+logger = logging.getLogger(__name__)
+
+ENABLE_ENV = "DYN_ENABLE_MEDIA_DECODERS"
+PACKAGES_ENV = "DYN_MEDIA_DECODER_PACKAGES"
+PIP_ARGS_ENV = "DYN_MEDIA_DECODER_PIP_ARGS"
+TIMEOUT_ENV = "DYN_MEDIA_DECODER_TIMEOUT_S"
+
+_DEFAULT_TIMEOUT_S = 600
+_LOCK_PATH = Path(tempfile.gettempdir()) / "dynamo_media_decoders.lock"
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Confirmed and fixed in 25091c2ae7 — and thank you, this was real: reproduced on a runtime image by planting a symlink at the fixed lock path and driving the installer through the lock; `open(_LOCK_PATH, "w")` truncated the symlink target to empty. The lock now opens with `os.open(O_CREAT | O_RDWR | O_NOFOLLOW, 0o600)`: symlinks are refused, nothing is ever truncated (no `O_TRUNC`), and the existing best-effort fallback proceeds lock-less rather than failing. Verified on the same image post-fix: victim file intact, and the normal lock + real-install path still works end to end. Unit tests added for both the symlink-refusal and the 0600 normal path.

components/src/dynamo/common/utils/media_decoders.py
```diff
@@ -0,0 +1,320 @@
...
+            _pip_install(packages, extra_args, with_deps=with_deps)
+    except Exception as exc:  # noqa: BLE001 - never break worker startup
+        logger.error(
+            "Failed to install media decoder backend(s) %s: %s. Requests needing "
+            "these input formats will fail with an actionable error at request "
+            "time. For offline/air-gapped clusters, pre-install the package(s) "
+            "into the image or point %s at a local wheelhouse "
+            "(e.g. '--find-links /wheels').",
+            _redact(" ".join(packages)),
+            exc,
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Confirmed and fixed in 25091c2ae7. Reproduced on a runtime image: with `--pip-args="--index-url https://user:token@…"`, `str(CalledProcessError)` embedded the full pip command and the raw token reached the error log — `_redact()` only covered the pre-exec command line, exactly as reported. The CLI boundary now logs `_redact(str(exc))`; verified post-fix on the same image (zero raw-token occurrences, `***@` marker present) plus a caplog unit test.

components/src/dynamo/common/utils/media_decoders.py
```diff
@@ -0,0 +1,320 @@
...
+_NO_DECODE_BACKENDS = frozenset({"trtllm"})
+
+# Process-local guard so repeated calls in one interpreter are cheap no-ops.
+_completed: set[str] = set()
+
+
+def _module_available(module: str) -> bool:
+    """Return True if `module` can be imported in this interpreter."""
+    try:
+        return importlib.util.find_spec(module) is not None
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Confirmed and fixed in 25091c2ae7. Reproduced on a runtime image: installed decord2, deleted its native libs — `find_spec` still resolved while `import decord` failed — and the installer exited 0 with 'already present', silently skipping the broken package. The fresh-interpreter probe now performs real imports (in a subprocess, so the parent's module state stays clean) and serves all three checks: pre-install skip, post-lock re-check, post-install verify. A broken package now exits non-zero with a `--force-reinstall` hint, since pip treats the requirement as already satisfied. Verified post-fix on the same image; unit test added with a planted package that raises on import.

components/src/dynamo/common/utils/install_media_decoders.py
rmccorm4, line 1: nit - can we call this `install_media_decoders.py` so it's more clearly an install script?

docs/fern/pages/use-cases/multimodal-serving/additional-media-decoders.md
```diff
@@ -0,0 +1,107 @@
...
+It installs the table's validated specs with `--no-deps`, skips packages that already import, serializes concurrent runs with a file lock, and exits non-zero if the install fails or the module still does not import afterwards. To see what it would do first:
+
+```bash
+python -m dynamo.common.utils.media_decoders vllm --dry-run
+```
+
+To pin different versions or install a curated subset, pass explicit specs (this install resolves dependencies, so pin transitive versions if that matters):
+
+```bash
+python -m dynamo.common.utils.media_decoders vllm --packages "opencv-python-headless==4.13.0.92"
```
rmccorm4: To be honest, I'm not sure we need to expose customizability and such like arbitrary packages, they can just run `pip install <packages>` directly for whatever their need is if they need to customize, right? This would also let them choose to do `pip install` vs `uv pip install` vs ... of their choice

I do think the mapping of vllm/sglang/trtllm to necessary libs and mapping of pyav, opencv, etc. that this script does is a useful reference to capture in one place for users.

I think it's fine to start with this just to provide something for the release with the out-of-box tested combinations of packages/versions, but at the same time for future looking thoughts, for any custom/extra scenarios, we can probably just direct users to running `pip install` commands directly rather than trying to capture all the customization cases for them in this script

rmccorm4: I'm not sure if this is a different story for `ffmpeg` on rust frontend or other possibly more complicated things to install

rmccorm4: Let me know what you think, I appreciate all the work here.

rmccorm4 (approved):
LGTM, left a comment on long term need for an installer script vs. just referring to a handful of `pip install` commands in the documentation -- but I think the installer script is fine to give a solution for the v1.4.0 release.

docs/fern/pages/use-cases/multimodal-serving/additional-media-decoders.md
dmitry-tokarev-nv: Yeah, --packages we can drop for simplicity. Doing it now.
ffmpeg - much much more complex issue and I think we should leave it out for 1.4.0.

components/src/dynamo/common/utils/install_media_decoders.py
dmitry-tokarev-nv: will do

dmitry-tokarev-nv: done

docs/fern/pages/use-cases/multimodal-serving/additional-media-decoders.md
dmitry-tokarev-nv: done

Issue #12053, opened 2026-07-23
vLLM POC: GMS-owned persistent KV warm failover


hutm:
## Summary

Prove the smallest reviewable **vLLM-only** loop in which GMS-owned GPU KV survives one inference-engine process failure and a local shadow vLLM resumes from the preserved cache with exactly correct output.

Linear project: [vLLM POC — GMS KV Failover](https://linear.app/nvidia/project/vllm-poc-gms-kv-failover-722cab6cb4ed)

This issue covers only the first merge slice of the larger GMS train: runtime prerequisites followed by the eight-part vLLM Minimal POC. Everything after that slice is explicitly out of scope.

## Problem

Today a vLLM process normally owns both the physical GPU KV-cache allocation and the engine-local metadata used to find reusable KV blocks. If the process crashes, the allocation lifetime and usable content-to-block association are lost together, forcing a replacement process to start cold even when the KV bytes could have survived.

The POC separates those responsibilities:

* **GMS owns** daemon-lifetime CUDA VMM allocations, crash-safe leases and fencing, and a flat survivable record for committed KV content.
* **vLLM owns** scheduling, block tables, prefix indexing, cache policy, block hashes, and all engine-specific execution state.

GMS does not mirror vLLM's prefix tree. It stores only enough information to validate and recover a committed physical block.

## Local POC topology

```text
frontend
   |
   +--> primary vLLM
   `--> shadow vLLM
             |
             +--> GMS allocation service
             `--> GMS content-directory service
```

The primary publishes a recovery record only after KV bytes are complete and the corresponding lease is sealed. After primary failure, the shadow obtains a newer fenced writer epoch, reattaches the same daemon-owned VMM mappings, adopts only compatible committed records, installs the recovered physical blocks into its native vLLM state, and resumes serving.

## Published official PR train

1. [#12056 — persistent allocation wire contracts](https://github.com/ai-dynamo/dynamo/pull/12056)
2. [#12057 — crash-safe shared KV lease ring](https://github.com/ai-dynamo/dynamo/pull/12057)
3. [#12032 — bounded parent-first process teardown](https://github.com/ai-dynamo/dynamo/pull/12032)
4. [#12055 — bounded fast-failover unregister](https://github.com/ai-dynamo/dynamo/pull/12055)

[...]

PR #12057, opened 2026-07-23
feat(gms): arbitrate persistent KV blocks with crash-safe leases

hutm:
## Overview:

This PR prevents the primary and shadow from using the same KV slot at the same time. Each slot has a small crash-safe lease, so completed blocks survive while abandoned work can be reclaimed.

## Details:

- Add a native shared-memory ring with FREE, LEASED, SEALED, and TRANSITION states.
- Fence ownership by PID and generation and recover interrupted mutations after crashes.
- Seal batches atomically and reject corrupt or truncated maps unless reset is explicitly enabled.

### Why it is needed

Persistent bytes alone are unsafe; primary and shadow need block-granular ownership without copying vLLM's prefix tree into GMS.

`lib/gpu_memory_service/integrations/common/kv_lease_client.py`, then the native ring crash tests.

## Validation

- Native lease-ring checkpoint: 27 passed, including multi-process crash cases.
- Rust formatting and pinned project lint pass.

## Related Issues

- Relates to #12053

## Stack

Checkpoint 4/12 for #12053.

- Previous: #14577
- Next: #14578
- Linear project: [vLLM POC — GMS KV Failover](https://linear.app/nvidia/project/vllm-poc-gms-kv-failover-722cab6cb4ed)


lib/gpu_memory_service/integrations/common/kv_lease_client.py
```diff
@@ -0,0 +1,838 @@
...
+                    f"path={shm_path} existing={existing_blocks} requested={total_blocks}"
+                )
+            map_size = (
+                _KV_LEASE_SHM_HEADER_SIZE
+                + int(existing_blocks) * _KV_LEASE_SHM_RECORD_SIZE
+            )
+            if stat.st_size < map_size:
+                os.ftruncate(fd, map_size)
+            buf = mmap.mmap(fd, map_size)
+            return fd, buf
```
hhzhang16, line 329: When the header is valid but the file is shorter than the declared geometry, this function extends it with zeros. Those zero records mean `FREE`, so blocks previously owned by the old process can be allocated again and silently recreates lost lease records as free capacity. I reproduced this: after truncating an allocation containing two live leases, the replacement reacquired the same block IDs and generations. This is an issue because the content directory may still advertise those blocks as valid cached KV while the lease ring says they are newly free/reusable. The file must fail closed on a size mismatch, not extend itself.

lib/gpu_memory_service/integrations/common/kv_lease_client.py
```diff
@@ -0,0 +1,838 @@
...
+        target_logger.warning("%s %s", message, rendered)
+    else:
+        target_logger.warning("%s", message)
+
+
+def _env_enabled(name: str, default: bool) -> bool:
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    return value.strip().lower() not in ("", "0", "false", "no", "off")
```
hhzhang16, line 78: This enables any value except its denylist; would it be better to instead use an allowlist for explicit truthy-ness and disable otherwise?

lib/gpu_memory_service/integrations/common/kv_lease_client.py
hutm: Addressed. Existing maps are never extended after validation: the implementation computes the expected geometry and requires os.fstat(fd).st_size == expected_size before mmap. A valid header with a truncated body now raises RuntimeError with file size mismatch. test_shared_memory_lease_client_rejects_truncated_valid_map covers both direct construction and the environment factory.

lib/gpu_memory_service/integrations/common/kv_lease_client.py
hutm: Addressed. Lease feature flags now use the shared is_truthy_env allowlist, so only recognized true values enable the feature. test_kv_lease_opt_in_flags_accept_only_recognized_truthy_values covers unknown and explicit false values.

PR #12061, opened 2026-07-23
refactor(planner): decouple global planner decision logic from K8s and drt

sachalmalick:
## Overview:

Create a boundary between Dynamo Runtime based API (scale_handler.py), core Global Planner decision logic (now in orchestrator.py) and kubernetes scaling (now in CapacityManager)

## Details:

Create a boundary between Dynamo Runtime based API (scale_handler.py), core Global Planner decision logic (now in orchestrator.py) and kubernetes scaling (now in CapacityManager).
This is the first step towards creating a Global Planner 

orchestrator.py
capacity_manager.py

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes-backed capacity discovery, monitoring, and replica scaling.
  * Added global planning for GPU budgets, authorization, participant registration, and coordinated scale requests.
  * Added support for pairing scale operations across pools while respecting minimum and maximum GPU limits.
  * Added current replica reporting and handling for generic worker roles.

* **Bug Fixes**
  * Improved resilience when Kubernetes discovery or capacity observations fail.

* **Tests**
  * Added comprehensive coverage for capacity management, orchestration, budgeting, authorization, and scaling behavior.

### DIFF
--- a/components/src/dynamo/global_planner/capacity_manager.py
+++ b/components/src/dynamo/global_planner/capacity_manager.py
@@ -1,0 +1,285 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Capacity backend for the GlobalPlanner — reads and writes scaling state.
+
+:class:`CapacityManager` is the infrastructure-facing base class the
+:class:`~dynamo.global_planner.orchestrator.Orchestrator` drives to **observe**
+current pool state and **actuate** replica changes. It defines a neutral,
+infrastructure-agnostic surface (``observe`` / ``actuate`` / ``discover`` /
+``ensure_participant`` / ``knows`` / ``current_replicas``) keyed by opaque
+``participant_id``; the orchestrator never sees Kubernetes concepts.
+
+:class:`KubernetesCapacityManager` is the concrete Kubernetes backend — the only
+place in the control loop that touches Kubernetes. A ``participant_id`` there is
+``"{cluster_name}/{deployment_name}"``, and the read/write logic is lifted
+verbatim from the pre-refactor ``ScaleRequestHandler`` so behavior is unchanged.
+
+Offline replay can add another backend (in-memory) by subclassing
+:class:`CapacityManager` and implementing the same surface — no orchestrator
+change required.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Optional
+
+from dynamo.planner import KubernetesConnector, TargetReplica
+from dynamo.planner.connectors.clients.kubernetes_api import KubernetesAPI
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class PoolSpec:
+    """Snapshot of one pool's state read from a backend.
+
+    Infrastructure-agnostic: the budget math consumes only these three fields.
+    Pools with ``gpu_per_replica == 0`` are included for completeness but
+    contribute 0 to budget totals.
+    """
+
+    sub_type: str
+    current_replicas: int
+    gpu_per_replica: int
+
+
+# participant_id -> (sub_type -> PoolSpec)
+PoolSnapshot = dict[str, dict[str, PoolSpec]]
+
+
+class CapacityManager:
+    """Base capacity backend: observe and actuate scaling state.
+
+    Neutral contract keyed by opaque ``participant_id``. Concrete backends
+    override the observe/actuate surface; ``discover`` defaults to a no-op for
+    backends (such as an in-memory replay backend) that have nothing to
+    pre-populate.
+    """
+
+    def discover(self, managed_callers: Optional[set[str]]) -> list[str]:
+        """Pre-populate the participant set at startup (no-op by default).
+
+        ``managed_callers`` scopes discovery in explicit mode, or ``None`` for
+        implicit mode. Returns the discovered participant identities.
+        """
+        return []
+
+    def ensure_participant(
+        self,
+        participant_id: str,
+        *,
+        caller_name: str,
+        cluster_name: str,
+        deployment_name: str,
+    ) -> None:
+        """Register a participant (idempotent) so it counts toward the budget and
+        can be actuated."""
+        raise NotImplementedError
+
+    def knows(self, participant_id: str) -> bool:
+        """Whether ``participant_id`` has been registered/discovered."""
+        raise NotImplementedError
+
+    def observe(self) -> PoolSnapshot:
+        """Read current pool state for every known participant."""
+        raise NotImplementedError
+
+    async def actuate(
+        self,
+        participant_id: str,
+        targets: list[TargetReplica],
+        *,
+        blocking: bool,
+    ) -> None:
+        """Apply desired replica targets to one participant."""
+        raise NotImplementedError
+
+    def current_replicas(self, participant_id: str) -> dict[str, int]:
+        """Read back current replica counts as ``{sub_type: replicas}``."""
+        raise NotImplementedError
+
+
+class KubernetesCapacityManager(CapacityManager):
+    """Observe / actuate capacity via ``KubernetesConnector`` per deployment.
+
+    A ``participant_id`` is ``"{cluster_name}/{deployment_name}"``. One
+    ``KubernetesConnector`` is cached per participant.
+    """
+
+    def __init__(self, cluster_name: str):
+        self.cluster_name = cluster_name
+        # participant_id -> KubernetesConnector
+        self.connectors: dict[str, KubernetesConnector] = {}
+
+    # ------------------------------------------------------------------ #
+    # Discovery / registration                                           #
+    # ------------------------------------------------------------------ #
+
+    def _managed_deployment_names(
+        self, managed_callers: Optional[set[str]]
+    ) -> Optional[set[str]]:
+        """Derive the deployment names this GlobalPlanner manages.
+
+        Returns a set of deployment names in explicit mode, or ``None`` in
+        implicit mode. The operator convention is
+        ``DYN_NAMESPACE = "{cluster_name}-{deployment_name}"``, so the deployment
+        name is the caller identity with the cluster prefix stripped.
+        """
+        if managed_callers is None:
+            return None
+
+        prefix = f"{self.cluster_name}-"
+        names = set()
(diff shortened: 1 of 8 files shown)


components/src/dynamo/global_planner/capacity_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                k8s_namespace=cluster_name,
+                parent_dgd_name=deployment_name,
+                raise_not_ready=True,
+            )
+            self.connectors[participant_id] = connector
+            logger.debug(f"Created new connector for {participant_id}")
+        else:
+            logger.debug(f"Reusing cached connector for {participant_id}")
+
+    def knows(self, participant_id: str) -> bool:
```
sachalmalick: rename to participant_exists or exists

components/src/dynamo/global_planner/capacity_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                current_replicas=replicas,
+                gpu_per_replica=gpu_per_replica,
+            )
+        return pools
+
+    # ------------------------------------------------------------------ #
+    # Actuate                                                            #
+    # ------------------------------------------------------------------ #
+
+    async def actuate(
```
sachalmalick: change this to scale() instead of actuate()

components/src/dynamo/global_planner/capacity_manager.py
```diff
@@ -0,0 +1,285 @@
...
+    ) -> None:
+        """Apply desired replica targets to one participant."""
+        raise NotImplementedError
+
+    def current_replicas(self, participant_id: str) -> dict[str, int]:
+        """Read back current replica counts as ``{sub_type: replicas}``."""
+        raise NotImplementedError
+
+
+class KubernetesCapacityManager(CapacityManager):
```
sachalmalick: move this to its own file.

components/src/dynamo/global_planner/capacity_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                    f"expected prefix '{prefix}'; cannot derive deployment name"
+                )
+        return names
+
+    def discover(self, managed_callers: Optional[set[str]]) -> list[str]:
+        """Pre-populate connectors for deployments managed by this GlobalPlanner.
+
+        Ensures the GPU budget accounts for deployments that already exist at
+        startup, even if they haven't sent a scale request yet. In explicit mode
+        (``managed_callers`` set) only matching deployments are discovered; in
```
sachalmalick: oh you know what lets change this back to managed_deployments that made way more sense

components/src/dynamo/global_planner/capacity_manager.py
```diff
@@ -0,0 +1,285 @@
...
+infrastructure-agnostic surface (``observe`` / ``actuate`` / ``discover`` /
+``ensure_participant`` / ``knows`` / ``current_replicas``) keyed by opaque
+``participant_id``; the orchestrator never sees Kubernetes concepts.
+
+:class:`KubernetesCapacityManager` is the concrete Kubernetes backend — the only
+place in the control loop that touches Kubernetes. A ``participant_id`` there is
+``"{cluster_name}/{deployment_name}"``, and the read/write logic is lifted
+verbatim from the pre-refactor ``ScaleRequestHandler`` so behavior is unchanged.
+
+Offline replay can add another backend (in-memory) by subclassing
```
sachalmalick: delete this we're not there yet

components/src/dynamo/global_planner/orchestrator.py
```diff
@@ -0,0 +1,917 @@
...
+
+    ``approved`` requests apply the requesting pool's targets plus any
+    ``selected_partners`` (empty for a standalone apply). A rejected result
+    carries the soft-denial ``reject_message`` the caller returns unchanged.
+    """
+
+    approved: bool
+    # (participant_id, sub_type, applied_desired, spec) — partners to apply
+    # alongside the request. Empty means "standalone / no-budget path".
+    selected_partners: list[tuple[str, str, int, PoolSpec]]
```
sachalmalick: can you not do these extremely ugly nested types please
`list[tuple[str, str, int, PoolSpec]]`
this is a sign that ur missing a definition of a concrete type

components/src/dynamo/global_planner/orchestrator.py
```diff
@@ -0,0 +1,917 @@
...
+
+    # ------------------------------------------------------------------ #
+    # observe -> decide -> actuate                                       #
+    # ------------------------------------------------------------------ #
+
+    async def submit(
+        self,
+        participant_id: str,
+        targets: list[TargetReplica],
+        *,
```
sachalmalick: tbh, i dont like the wild card here.  do we need it?

components/src/dynamo/global_planner/orchestrator.py
```diff
@@ -0,0 +1,917 @@
...
+        blocking: bool,
+        deployment_name: str,
+        caller_name: str,
+    ) -> ScaleOutcome:
+        """Arbitrate and execute one scale request.
+
+        The observe -> decide -> actuate span is serialized under the scale lock
```
sachalmalick: its not called actuate anymore, can we agree to stop saying actuate please?

tedzhouhk:
## Global Planner follow-up findings (mostly pre-existing)

**Important context:** the runtime correctness issues below already exist in the base implementation and were moved into the new `Orchestrator`; they were **not introduced by this PR**. I am OK with not fixing those historical issues as part of this refactor. However, because several affect hard GPU bounds and partial execution, it's better adding some`TODO`s.

The refactor-specific items are explicitly marked below.

### P1

1. **Authorization is not bound to the scaling target.** Only `caller_namespace` is authorized, while `k8s_namespace` and `graph_deployment_name` from the same request are trusted. An allowed caller can potentially target another namespace/DGD. (`scale_handler.py:141-184`)

2. **The planner cannot guarantee a complete, serialized cluster snapshot.** Startup discovery fails open; participant registration happens outside the budget lock; and multiple Global Planner replicas have process-local locks, connector registries, and intent caches. Any of these can under-count GPUs and allow the hard ceiling to be exceeded. (`kubernetes_capacity_manager.py:85-186`, `scale_handler.py:175-192`, `orchestrator.py:131-152`)

3. **Boundary and partner-selection logic does not reliably preserve the intended budget semantics.** This includes: rejecting incremental recovery when already outside a boundary, computing tolerance from unselected candidates, stopping greedy search after an unpartializable candidate, preferring a cached pair over an already-valid standalone request, and losing the same-participant preference during global sorting. (`orchestrator.py:448-530`, `627-671`, `747-842`, `884-890`)

4. **The budget-approved plan is not equivalent to the Kubernetes execution sequence.** Duplicate pool targets are modeled differently from how they are applied; same-DGD multi-target scaling is a sequence of independent PATCHes rather than an atomic operation; and down-before-up ordering does not mean GPUs have actually been released when `blocking=False`. A plan whose final total is in bounds can transiently—or after a partial failure, permanently—exceed the ceiling. (`orchestrator.py:316-347`, `434-440`, `900-915`; `planner/connectors/kubernetes.py:699-723`)

5. **Partial failures do not have transaction semantics.** A later participant PATCH failure is swallowed, subsequent scale-ups may continue, and the requester can still receive `SUCCESS`; cancellation can similarly leave a half-applied transfer with no reconciliation record. (`orchestrator.py:249-269`, `349-395`)

### P2

1. **Request/config validation is incomplete.** Unknown, duplicate, empty, or negative targets can make mediation differ from backend execution; configuration also does not reject `min > max`, invalid TTL values, or distinguish an empty allowlist from `None`. (`orchestrator.py:119-152`, `429-458`, `569-587`, `900-915`)

2. **Participant and intent state have no lifecycle reconciliation.** Connector discovery is one-shot and add-only, deleted DGDs remain cached, and expired/satisfied intents are skipped but never pruned. (`kubernetes_capacity_manager.py:85-186`, `orchestrator.py:569-587`, `649-656`)

3. **The request path has scaling and event-loop costs.** Every request observes every participant—even with budget enforcement disabled—creating roughly O(N²) Kubernetes reads across N local planners; synchronous Kubernetes readback/PATCH work also runs on the async request path, and `blocking=True` can hold the global lock through readiness waits. (`orchestrator.py:249-303`, `442-443`; `kubernetes_capacity_manager.py:288-320`)

4. **[Refactor-specific] The new core boundary is not actually transport/backend-neutral.** `Orchestrator`/`CapacityManager` still use `TargetReplica`, protocol `ScaleStatus`, planner exceptions, and namespace/deployment registration fields despite claiming a neutral boundary. A domain DTO/status/participant descriptor should live inside the core, with conversion at the handler/backend adapters. (`capacity_manager.py:23-103`, `orchestrator.py:4-14`, `32-43`, `86-95`, `170-187`)

5. **The same configuration is used for two different concepts.** `managed_namespaces` becomes `managed_deployments`, is used as the caller authorization allowlist, and is also passed to Kubernetes discovery where names are derived through a namespace-prefix convention. Please separate `authorized_callers` from backend discovery scope. (`scale_handler.py:73-82`, `orchestrator.py:143-166`, `193-200`, `kubernetes_capacity_manager.py:60-83`)

6. **[Refactor-specific] The compatibility accessors are not compatible with the previous writable API.** `max_total_gpus`, `min_total_gpus`, and `managed_namespaces` are now read-only properties; existing callers that mutate them receive `AttributeError`, and `managed_namespaces` also changes shape. Either provide forwarding setters or document the API break. (`scale_handler.py:115-129`)

7. **The tests do not currently protect these concurrency/execution invariants.** The concurrent-insert test is a false positive because its monkeypatch signature causes a swallowed `TypeError`; several handler unit tests can perform real Kubernetes discovery; and coverage is missing for HA, cancellation, intermediate PATCH failure, out-of-bound recovery, tolerance inflation, and duplicate-target execution. (`tests/unit/test_scale_request_handler.py:497-503`, `1241-1280`)

sachalmalick:
@tedzhouhk - in response to your list of notes, I added TODOs for all the existing issues in global planner that this re-factor re-surfaced.

Points 4 and Point 6 were refactor specific

For point 4 Orchestrator/CapacityManager can use these types as they are generic enough in my opinion. 
For point 6, I dont think these properties should be mutable anyway.  Do you think they should be subject to change after initialization?

tedzhouhk:
> For point 6, I dont think these properties should be mutable anyway. Do you think they should be subject to change after initialization?

make sense, immutable sounds good.

tedzhouhk (approved):

PR #12070, opened 2026-07-23
fix(recipes): make gpt-oss-120b disagg + perf run as shipped

RomaA2000:
## Summary

Two "recipe fails on apply as shipped in `main`" bugs in `recipes/gpt-oss-120b`, both reported from a Kubernetes-native operator dev cluster (B200):

**1. disagg deploys force etcd discovery at a hardcoded endpoint.**
`vllm/disagg-{b200,h200}-agentic/deploy.yaml` set `ETCD_ENDPOINTS` + `DYN_DISCOVERY_BACKEND: etcd` pointing at `dynamo-platform-etcd.dynamo-system`. On operator / Kubernetes-native clusters (no local etcd) the disagg pod can't reach it and exits ~20s. The **agg** variants set no discovery env and deploy fine (operator injects discovery); the disagg pods carry the same `nvidia.com/dynamo-component-type` labels, so removing the forcing lets discovery use the cluster's backend. `NATS_SERVER` is kept (message bus; `dynamo-system` is the normal location and the reporting cluster kept it). Same fix pattern as bug 6435308 (DeepSeek-V4-Pro).

**2. perf.yaml passes a synthetic-only flag in trace mode.**
Both aiperf calls combined `--input-file` (mooncake trace replay) with `--prompt-input-tokens-block-size 512`. That flag (`isl-block-size`) is synthetic-dataset-only, and the pinned `aiperf==0.10.0` errors on the combo (`isl-block-size is only supported with synthetic datasets, not --input-file`), so every trace-mode perf run fails on launch. Removed from both the warmup and the scored call.

## Validation

- `yaml.safe_load_all` parses all three edited files.
- disagg env after the change: `NATS_SERVER` + `DYN_NAMESPACE` + `NCCL_*` retained; only the two etcd vars removed.
- perf.yaml: `--prompt-input-tokens-block-size` no longer present in either aiperf invocation.
- Cross-checked that the working agg variants already set none of the removed discovery env.

## Summary by CodeRabbit

* **Performance**
  * Updated GPT-OSS 120B benchmarking profiles by removing a prompt token block-size setting from warmup and scored runs.
  * Preserved existing benchmark configuration, including concurrency, streaming, timeouts, worker sizing, and artifact handling.

* **Deployment**
  * Simplified agentic deployment environment configuration for B200 and H200 setups.
  * Retained NATS connectivity while removing obsolete service-discovery settings.


karen-sy (approved):

karen-sy:
/ok to run [a791c7e](https://github.com/ai-dynamo/dynamo/pull/12070/commits/a791c7eb365437bf655eb8989b8170c68f7ba582)

PR #10494, opened 2026-06-09
feat(operator): add runtimeVersionOverride to DGD components

tmonty12:
## Summary

Adds `runtimeVersionOverride` as an explicit declaration of Dynamo runtime compatibility when an image tag does not provide it.

The override controls version-dependent operator behavior only. It never rewrites or pins the image reference.

### API and CRDs

- Adds `runtimeVersionOverride` to:
  - DGD component specs
  - DCD specs
  - DGDR specs
- Supports both `v1alpha1` and `v1beta1`, including conversion.
- Accepts canonical core semver only, for example `1.4.0`.
- Updates generated CRDs, Helm references, and API documentation.
- Excludes the field from the legacy worker hash, so changing compatibility metadata alone does not roll Pods.

### Admission

- **DGD:** requires a `main` image; a non-semver image requires `runtimeVersionOverride`.
- **DCD:** requires a `main` image but does not require the override because DCDs are controller-generated.
- **DGD/DCD updates:** ratchet legacy objects. An unchanged legacy image/version tuple remains valid, while changing an image to a non-semver value requires an override.
- **DGDR:** requires the override when `spec.image` has no parseable semantic-version tag.
- Keeps DGDR specs immutable after profiling starts, except for the deferred `autoApply: false` flow:
  - select the runtime override during Profiling or Ready
  - enable auto-apply once the request is Ready

### DGDR controller and profiler

- Propagates the DGDR override into newly generated DGD components.
- Preserves explicit component overrides when the controller fills missing values.
- Applies the override defensively immediately before creating a new DGD.
- Keeps the stored Ready snapshot immutable; it is not repaired or rewritten.
- Does not mutate an existing deployed DGD through a DGDR update.

### Documentation and test support

- Documents custom-image, digest, and non-semver-tag usage.
- Updates DGDR/profiler examples and generated references.
- Adds focused unit, webhook, conversion, controller, and envtest coverage.
- Updates deployment fixtures to provide deterministic runtime compatibility versions.

## Validation

- `go test ./internal/runtimeversion ./api/v1alpha1 -count=1`
- `KUBEBUILDER_ASSETS=<envtest assets> go test ./internal/webhook/validation ./internal/controller -count=1`
- `make SHELL=/bin/sh manifests`
- `make SHELL=/bin/sh generate-api-docs`
- `git diff --check`

## Tracking

[...]

Issue #9020, opened 2026-05-01
DEP: Operator–Runtime Version Compatibility Contract


tmonty12:
# Operator–Runtime Version Compatibility

**Authors:** Thomas Montfort

**Area:** Operator / Runtime Version Compatibility

### Version History

| Date | Version | Changes |
|---|---|---|
| May 1, 2025 | First draft | Initial proposal with declarative `runtime-versions.yaml` manifest and workload-aware admission validation |
| May 8, 2025 | Second draft | Replaced `runtime-versions.yaml` with version-gated code for frozen injection. Dropped workload-aware validation in favor of blanket N-2 guarantee backed by CI/QA test matrix. Added admission warnings with opt-in enforcement. Added EPP and Planner component test matrices. Added forward compatibility (N+1) policy and testing. Added migration path, implementation plan, and alternatives considered. |
| Jun 4, 2025 | Third draft | Added in success criteria for Planner test matrix. |

## Summary

This document proposes a framework for ensuring version compatibility between the Dynamo platform/operator and the runtime container images it manages. Today, there is no guarantee whether a given runtime image version is compatible with a given operator version, or whether a specific DynamoGraphDeployment (DGD) workload will function correctly with a particular operator–runtime combination. This leads to silent failures, CrashLoopBackoffs, and degraded functionality that is only discovered at deploy time.

We propose adopting a version skew policy — inspired by the [Kubernetes Version Skew Policy](https://kubernetes.io/releases/version-skew-policy/) — where the operator guarantees compatibility with runtime versions N, N-1, and N-2 as well as N+1. The compatibility will be proven through a CI/QA test matrix for single-node and multi-node configurations, non-inclusive of experimental features. To prevent spurious rolling updates of a DGD workload during operator upgrades, all injection changes are version-gated in operator code, ensuring the operator produces a deterministic pod spec for a given (DGD spec, runtime version) pair.

## Motivation

### Background

The Dynamo operator is responsible for translating a user's `DynamoGraphDeployment` (DGD) specification into running Kubernetes workloads. As part of this translation, the operator **injects** a substantial set of configuration into the runtime pods it creates: environment variables, CLI flags, health probes, volumes, init containers, labels, and ports. The runtime image must understand and correctly consume this injected configuration for the deployment to function.

[...]

PR #12072, opened 2026-07-23
fix(operator): propagate DGDR runtime version overrides

tmonty12:
#### Overview

Closes the DGDR auto-apply compatibility gap introduced by #10494.

#10494 requires each DGD component whose main image has a custom or otherwise non-semver tag to declare a `runtimeVersionOverride`. A DGDR could still accept such an image, generate a DGD without per-component overrides, and then fail when auto-apply submitted that DGD to admission.

This PR carries the runtime compatibility intent from the DGDR through profiler-based DGD generation, so a DGDR accepted by admission produces an admission-valid DGD.

#### Semantics

- `spec.runtimeVersionOverride` declares the Dynamo runtime compatibility version for every component generated from the DGDR. It must be canonical core semver (`major.minor.patch`).
- When an explicit `spec.image` has no parseable semantic-version tag, the override is required on create and every non-deletion update. Existing invalid objects must be repaired on their next update instead of being ratcheted indefinitely.
- When the image has a parseable semantic-version tag, the override remains optional and generated components can derive compatibility from that tag.
- When `spec.image` is omitted, the defaulter uses a canonicalized, parseable `operatorVersion` for the planner image tag (for example, `v1.2.3` becomes `1.2.3` and `1.2` becomes `1.2.0`). Unknown or invalid operator versions are not used as image tags.
- The profiler applies the override to every generated legacy v1alpha DGD service after optional Planner injection, covering frontend, workers, and Planner consistently.

#### Scope

- Adds the DGDR field to v1alpha1 and v1beta1 APIs, conversion coverage, generated CRDs, API documentation, and profiler Pydantic models.
- Adds DGDR admission validation for custom/non-semver images and strict update-time repair.
- Propagates the override through all DGD generation paths, including mocker and Planner-enabled configurations.
- Adds controller envtests for defaulting, validation, conversion, custom-tag auto-apply, and generated DGD contents.
- Documents compatibility requirements for older profiler images and raw DGD overrides.
- Updates generated documentation and admission smoke fixtures so repository and deployment checks exercise valid DGDs.

#### Validation

- `KUBEBUILDER_ASSETS=... go test ./api/... ./internal/webhook/defaulting ./internal/webhook/validation -count=1`
- `KUBEBUILDER_ASSETS=... go test ./internal/controller -count=1`
- `go test ./api -run 'TestFuzzRoundTripMutability/DGDR|TestFuzzRoundTrip_DGDR' -roundtrip-fuzz-iters=3000 -count=1 -v`
- `PYTHONPATH=components/src python -m pytest components/src/dynamo/profiler/tests/unit/test_planner_image_selection.py` (10 passed)
- Go/Python DGD override boundary integration (1 passed)
- `make check` with the CI-pinned Go and Python toolchain
- `ruff check`, `ruff format --check`, and `git diff --check`

The all-resource 3,000-iteration fuzz command still reproduces the same pre-existing DGD `VolumeMount` ordering mismatch at iteration 1405 on the parent PR head; all DGDR fuzz paths pass.

#### Related

- Parent: #10494
- Linear: https://linear.app/nvidia/issue/DEP-940/add-runtimeversion-field-to-crd-with-semver-parsing-from-image-tags
- DEP: https://github.com/ai-dynamo/dynamo/issues/9020


docs/kubernetes/dgdr.md
```diff
@@ -78,6 +79,42 @@ spec:
...
+Profilers from Dynamo 1.4.0 or later propagate the DGDR-level override to every
+generated component, so `spec.runtimeVersionOverride` also covers a worker image
+replaced through `overrides.dgd`.
+
+With a profiler based on Dynamo 1.3.0 or earlier, the DGDR-level field is
+ignored. However, `overrides.dgd` is a raw embedded resource and preserves a
+per-component `runtimeVersionOverride`. Use the legacy `v1alpha1`
+service-shaped DGD override understood by those profilers, and set the field on
+the affected service rather than relying on the DGDR-level
+`spec.runtimeVersionOverride`.
```
sttts: I have problems to understand what this is saying. Wondering how to simplify it. Can we protect the user to shot himself into the foot via some validation?

docs/kubernetes/dgdr.md
```diff
@@ -52,6 +52,7 @@ spec:
 |---|---|---|---|
 | `model` | Yes | — | HuggingFace model ID (e.g. `Qwen/Qwen3-0.6B`) |
 | `image` | No | — | Container image for the profiling job. Dynamo >= 1.1.0: use `dynamo-planner`; earlier versions: use `dynamo-frontend`. |
+| `runtimeVersionOverride` | Conditional | — | Dynamo runtime version propagated to generated DGD components when `image` has no parseable semantic-version tag or its tag does not identify the intended runtime. |
```
sttts, line 55: is the term runtimeVersion correct here? Is it that a 1.4.3 planner will always create DGDs with 1.4.3 runtimes?

sttts (commented):
Codex review, focused on the added Kubernetes DGDR image/runtime-version documentation.

The new `docs/kubernetes/dgdr.md` section is much longer than the user-facing semantics require. I would shorten it to the behavior a DGDR author needs to know: omitted `spec.image` defaults to the operator-versioned planner image; explicit `spec.image` tags are treated as the Dynamo runtime version only by convention; and `spec.runtimeVersionOverride` is the escape hatch when that convention is not true.

Please update the field table row so the default is visible:

```md
| `image` | No | `nvcr.io/nvidia/ai-dynamo/dynamo-planner:<operatorVersion>` | Container image for the profiling job. Dynamo >= 1.1.0: use `dynamo-planner`; earlier versions: use `dynamo-frontend`. |
```

Please replace the long `Profiler Image Version Compatibility` section with something closer to this:

```md
### Profiler Image Version Compatibility

When `spec.image` is omitted, the operator defaults it on create to
`nvcr.io/nvidia/ai-dynamo/dynamo-planner:<operatorVersion>`.

When you set `spec.image`, tag it with the Dynamo runtime version in semantic
version form, for example `dynamo-planner:1.4.0`. Admission can only check that
the tag parses as a semantic version; it cannot verify what runtime the image
contains. Set `spec.runtimeVersionOverride` for tagless, digest-only,
non-semver, or semver-tagged images where the tag is not the Dynamo runtime
version.

Dynamo 1.3.0 and older profiler images do not understand DGDR-level
`spec.runtimeVersionOverride`. If you use one with Dynamo Operator 1.4.0 or
later, use a runtime-version image tag such as `dynamo-planner:1.3.0`. For
images changed through `spec.overrides.dgd`, apply the DGD component
runtime-version rules to the overridden component.
```

docs/kubernetes/dgdr.md
```diff
@@ -52,6 +52,7 @@ spec:
 |---|---|---|---|
 | `model` | Yes | — | HuggingFace model ID (e.g. `Qwen/Qwen3-0.6B`) |
 | `image` | No | — | Container image for the profiling job. Dynamo >= 1.1.0: use `dynamo-planner`; earlier versions: use `dynamo-frontend`. |
```
sttts, line 54: Codex review: This row still says `image` has no default, but the defaulting webhook sets omitted `spec.image` on create to `nvcr.io/nvidia/ai-dynamo/dynamo-planner:<operatorVersion>`. Please make the default visible here so users know leaving it empty is intentional.

Suggested replacement:

```md
| `image` | No | `nvcr.io/nvidia/ai-dynamo/dynamo-planner:<operatorVersion>` | Container image for the profiling job. Dynamo >= 1.1.0: use `dynamo-planner`; earlier versions: use `dynamo-frontend`. |
```

docs/kubernetes/dgdr.md
```diff
@@ -78,6 +79,43 @@ spec:
...
+generated component, so `spec.runtimeVersionOverride` also covers a worker image
+replaced through `overrides.dgd`.
+
+With a profiler based on Dynamo 1.3.0 or earlier, the DGDR-level field is
+ignored. Those profilers also discard the embedded override's `apiVersion` and
+`kind`, then directly merge the remaining dictionary into a generated
+`v1alpha1` DGD. Raw storage preserves the override fields but does not convert
+between DGD schemas. Use a `v1alpha1`-shaped override under `spec.services` and
+set `runtimeVersionOverride` on the affected service. A `v1beta1`
+`spec.components` override is not translated into the generated service.
```
sttts, line 117: Codex review: This section is doing too much for the Kubernetes reference page. The user-facing rule can be much shorter: empty `spec.image` gets the operator-versioned planner image, explicit image tags are treated as the Dynamo runtime version only by convention, and `spec.runtimeVersionOverride` is the escape hatch when that convention is not true.

Please replace this section with something closer to this:

```md
### Profiler Image Version Compatibility

When `spec.image` is omitted, the operator defaults it on create to
`nvcr.io/nvidia/ai-dynamo/dynamo-planner:<operatorVersion>`.

When you set `spec.image`, tag it with the Dynamo runtime version in semantic
version form, for example `dynamo-planner:1.4.0`. Admission can only check that
the tag parses as a semantic version; it cannot verify what runtime the image
contains. Set `spec.runtimeVersionOverride` for tagless, digest-only,
non-semver, or semver-tagged images where the tag is not the Dynamo runtime
version.

Dynamo 1.3.0 and older profiler images do not understand DGDR-level
`spec.runtimeVersionOverride`. If you use one with Dynamo Operator 1.4.0 or
later, use a runtime-version image tag such as `dynamo-planner:1.3.0`. For
images changed through `spec.overrides.dgd`, apply the DGD component
runtime-version rules to the overridden component.
```

docs/kubernetes/dgdr.md
sttts: Yes. `spec.image` is the profiler/planner image, and `derive_backend_image` replaces only the image name while preserving the tag. With the default path, an operator version `1.4.3` selects `dynamo-planner:1.4.3`, and the generated backend images keep `:1.4.3`. `runtimeVersionOverride` is therefore the explicit Dynamo runtime compatibility version for cases where that tag convention does not hold. The term is correct.

components/src/dynamo/profiler/utils/dgd_generation.py
```diff
@@ -156,11 +157,24 @@ def assemble_final_config(
...
 
+    apply_runtime_version_override(dgdr, base)
     if config_maps:
         return config_maps + [base]
     return base
 
 
+def apply_runtime_version_override(dgdr, config_dict: dict) -> None:
+    """Apply the DGDR runtime version to every generated DGD service."""
+    override = dgdr.runtimeVersionOverride
```
sttts, line 168: [P2] Propagation is currently owned by the selected profiler image. Admission accepts a custom-tag DGDR when `spec.runtimeVersionOverride` is set, but a Dynamo 1.3-or-older profiler ignores that new field and emits custom-tag DGD components without overrides. The controller does not reapply the DGDR value, so #10494 admission rejects the generated DGD. Please copy the override into every generated component in the operator when consuming profiler output, and add a controller test whose profiler YAML intentionally omits component overrides.

sttts (commented):
Static re-review of `5d01fd4` after the force-push; no tests run.

This is not merge-ready yet:

- The child head no longer descends from the current parent head `59646f7`; GitHub reports `CONFLICTING`/`DIRTY` and the PR diff repeats the parent stack (44 files). Please replay the four DGDR-specific commits after `842ecf5` onto the current parent branch.
- The existing update-validation thread remains valid: a legacy custom-tag DGDR in an immutable phase cannot accept a metadata-only update and cannot add the override either.
- The requested docs changes are not present at this head: the field table still shows no image default, and the compatibility section remains the longer implementation-heavy version.

I answered and resolved only the terminology question; the concrete docs threads remain open.

PR #12082, opened 2026-07-23
feat(replay): Add global planner + multi-model sim support

sachalmalick:
## Overview:

In this PR we add support for supporting multi-model / multi-deployments in Replay where scaling decisions are mediated by Global Planner.  This is a step towards being able to do comprehensive Inference Provider simulation with offline replay.  Inference Providers like Baseten, DeepInfra, Together, Firework, etc and even frontier labs like Anthropic and OpenAI are responsible for serving multiple models using the capacity they have available.  With this PR we can simulate how Global Planner would resolve competing scaling requests from different models / deployments with shared gpu budgets.

## Details:

TLDR, we introduce a world controller that is playing out all of the deployments "simultaneously" - still through a single thread.  It does this by finding the next valid timestamp barrier which is the min of every deployments next "data-plane" event and ever deployments next local planner deadline. Then advances all the deployments up until that timestamp.  Then finds the subsets that have planner ticks due.  Process them in a batch, takes their scaling proposals, runs them all through GlobalPlanner for mediation.

Definitely read `planner_world.rs` starting from `run` also read `world.py` and `global_planner.py`

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Added multi-deployment offline replay with shared global planning and coordinated scaling.
  * Added support for aggregated and disaggregated deployments, trace workloads, and configurable synthetic workloads.
  * Added structured replay reports with per-deployment results, scaling events, mediation outcomes, and planner events.
  * Added public replay-world APIs and configuration options, including workload controls and planner resource limits.
  * Added asynchronous planner operation for responsive replay execution.

* **Bug Fixes**
  * Improved scaling-state tracking, including partial targets and draining workers.
  * Strengthened validation and cleanup behavior for invalid inputs and replay failures.

### DIFF
--- a/components/src/dynamo/replay/global_planner.py
+++ b/components/src/dynamo/replay/global_planner.py
@@ -1,0 +1,321 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""In-memory Global Planner capacity backend for offline Replay.
+
+The production Global Planner observes and patches Kubernetes through a
+``CapacityManager``. Offline Replay uses the same orchestrator with this backend:
+configured participants are pre-registered, desired replica counts are mirrored
+in memory, and approved scale calls are staged as ordered actions for the Rust
+simulation world to apply after the Python callback returns.
+
+One orchestrator submission may scale several participants. ``begin_batch`` /
+``finish_batch`` make those calls visible to the orchestrator as one overlay while
+keeping the committed mirror unchanged until the controller accepts the complete
+batch. ``abort_batch`` discards both the overlay and its actions.
+"""
+
+from __future__ import annotations
+
+import operator
+from collections.abc import Iterable
+from dataclasses import dataclass
+from typing import Optional
+
+from dynamo.global_planner.capacity_manager import (
+    CapacityManager,
+    PoolSnapshot,
+    PoolSpec,
+)
+from dynamo.planner import SubComponentType, TargetReplica
+
+__all__ = [
+    "ReplayCapacityManager",
+    "ReplayParticipantSpec",
+    "ReplayScaleAction",
+    "ReplayScaleTarget",
+]
+
+
+@dataclass(frozen=True)
+class ReplayParticipantSpec:
+    """Pre-registered Global Planner participant and its capacity pools."""
+
+    participant_id: str
+    pools: tuple[PoolSpec, ...]
+
+
+@dataclass(frozen=True)
+class ReplayScaleTarget:
+    """One immutable pool target staged for the simulation world."""
+
+    sub_type: str
+    desired_replicas: int
+    component_name: Optional[str] = None
+
+
+@dataclass(frozen=True)
+class ReplayScaleAction:
+    """One ordered backend ``scale`` call staged during an active batch."""
+
+    participant_id: str
+    targets: tuple[ReplayScaleTarget, ...]
+    blocking: bool
+
+
+_PoolKey = tuple[str, str]
+_VALID_ROLES = frozenset(role.value for role in SubComponentType)
+
+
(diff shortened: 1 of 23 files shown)


components/src/dynamo/replay/world.py
```diff
@@ -0,0 +1,759 @@
...
+)
+from dynamo.planner.plugins.clock import VirtualClock
+from dynamo.replay.global_planner import (
+    ReplayCapacityManager,
+    ReplayParticipantSpec,
+    ReplayScaleAction,
+)
+
+
+def _require_integer(value: object, *, name: str, minimum: int) -> int:
```
sachalmalick, line 39: this is ugly / an overkill

components/src/dynamo/replay/world.py
```diff
@@ -0,0 +1,759 @@
...
+            not math.isfinite(self.arrival_speedup_ratio)
+            or self.arrival_speedup_ratio <= 0
+        ):
+            raise ValueError("arrival_speedup_ratio must be finite and positive")
+        _require_integer(self.trace_block_size, name="trace_block_size", minimum=1)
+
+
+@dataclass(frozen=True)
+class ReplaySyntheticWorkload:
+    """One deployment-local deterministic synthetic workload distribution."""
```
sachalmalick: how are we mapping this to arrival rate distributions?

components/src/dynamo/replay/world.py
```diff
@@ -0,0 +1,759 @@
...
+            next_ticks[participant_id] = result.next_tick_ms
+
+        self._capacity_manager.begin_batch()
+        actions: tuple[ReplayScaleAction, ...]
+        try:
+            for participant_id in sorted(proposals):
+                proposal = proposals[participant_id]
+                if proposal is None:
+                    continue
+                if not self._orchestrator.is_authorized(participant_id):
```
sachalmalick, line 445: i dont think we need this check its sim itll aalways be authorized.

components/src/dynamo/replay/global_planner.py
```diff
@@ -0,0 +1,321 @@
...
+    participant_id: str
+    targets: tuple[ReplayScaleTarget, ...]
+    blocking: bool
+
+
+_PoolKey = tuple[str, str]
+_VALID_ROLES = frozenset(role.value for role in SubComponentType)
+
+
+def _non_negative_int(value: object, *, field: str) -> int:
```
sachalmalick, line 70: delete, overkill

components/src/dynamo/replay/global_planner.py
```diff
@@ -0,0 +1,321 @@
...
+                    sub_type=role,
+                    desired_replicas=desired,
+                    component_name=target.component_name,
+                )
+            )
+            updates.append((key, desired))
+
+        for key, desired in updates:
+            overlay[key] = desired
+        actions.append(
```
sachalmalick, line 288: not really good to mutate function args like this.

sachalmalick: yea we could just be mutating self.\_batch_actions right? just curious if im missing a reason for this pattern

components/src/dynamo/replay/world.py
```diff
@@ -0,0 +1,759 @@
...
+            result = await self._sessions[participant_id].propose_tick_async(
+                by_id[participant_id]
+            )
+            proposals[participant_id] = result.proposal
+            next_ticks[participant_id] = result.next_tick_ms
+
+        self._capacity_manager.begin_batch()
+        actions: tuple[ReplayScaleAction, ...]
+        try:
+            for participant_id in sorted(proposals):
```
sachalmalick, line 441: i think we should be more thoughtful about the order in which we send to global planner.  right now i think theres an assumption that a) the scale_lock that forces serialization in global planner is gonna continue that way, i dont think that will scale, so we will need to be able to model partial parallel arrivals at global planner door.  also yea lemme look into what the default ordering after sorted is, maybe literally just a random shuffle here would be more realistic.

PR #12086, opened 2026-07-23
fix: count only generate workers in kv router bench

oandreeva-nv:
## Overview:

Fixes KV router benchmark worker discovery so the benchmark only uses backend workers that can serve generation requests

## Details:

`/health` reports all discovered component endpoints, not only generation workers. `kv_router_bench` was collecting every `instance_id`, which could make the benchmark count non-generation endpoints as workers and build synthetic KV state for IDs that are not valid generation targets.

This change filters discovered instances to `endpoint == "generate"` before deduplicating worker IDs.

 Validation:

 ```bash
  cargo bench --package dynamo-llm --bench kv_router_bench --features kv-router-stress --no-run
```

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **New Features**
  * Added an `--endpoint` option for selecting benchmark workers.
  * Worker discovery now filters by namespace, component, and endpoint.
  * Configuration output now displays the selected endpoint.

* **Bug Fixes**
  * Prevented duplicate workers from appearing in discovery results.
  * Improved messages and errors when matching workers are found or unavailable.

### DIFF
--- a/lib/llm/benches/kv_router_bench.rs
+++ b/lib/llm/benches/kv_router_bench.rs
@@ -625,9 +625,7 @@ async fn discover_worker_ids(
     let worker_ids: Vec<WorkerId> = health
         .instances
         .iter()
-        .filter(|i| {
-            i.namespace == namespace && i.component == component && i.endpoint == endpoint
-        })
+        .filter(|i| i.namespace == namespace && i.component == component && i.endpoint == endpoint)
         .map(|i| i.instance_id)
         .collect();
 
--- a/lib/llm/benches/kv_router_bench.rs
+++ b/lib/llm/benches/kv_router_bench.rs
@@ -612,9 +612,14 @@ async fn discover_worker_ids(frontend_url: &str) -> Result<Vec<WorkerId>> {
         .await
         .context("Failed to parse health response")?;
 
-    let worker_ids: Vec<WorkerId> = health.instances.iter().map(|i| i.instance_id).collect();
+    let worker_ids: Vec<WorkerId> = health
+        .instances
+        .iter()
+        .filter(|i| i.endpoint == "generate")
+        .map(|i| i.instance_id)
+        .collect();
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/llm/benches/kv_router_bench.rs [hunk 1]
```diff
@@ -612,9 +612,14 @@ async fn discover_worker_ids(frontend_url: &str) -> Result<Vec<WorkerId>> {
         .await
         .context("Failed to parse health response")?;
 
-    let worker_ids: Vec<WorkerId> = health.instances.iter().map(|i| i.instance_id).collect();
+    let worker_ids: Vec<WorkerId> = health
+        .instances
+        .iter()
+        .filter(|i| i.endpoint == "generate")
```
PeaBrane: `/health` is backed by `AllEndpoints`, so matching only the leaf name can also include endpoints such as `prefill.generate` or `encode.generate` from other components or namespaces. Since this benchmark already knows `--namespace` and `--component` and publishes KV events to that exact component subject, could we deserialize those fields and match the full `(namespace, component, endpoint)` identity? If custom endpoint names are intended to work here, a `--endpoint` option defaulting to `generate` would keep this aligned with `EndpointId` semantics.

PeaBrane (approved):
Nice focused fix. Filtering non-generation endpoint records out of benchmark worker discovery is the right correction and prevents synthetic KV state from being assigned to obviously invalid targets. I left an endpoint-identity suggestion below. ✅

lib/llm/benches/kv_router_bench.rs
oandreeva-nv: addressed

jthomson04 (approved):

rmccorm4:
/ok to test [4ed654f](https://github.com/ai-dynamo/dynamo/pull/12086/commits/4ed654fb7c7c46272b2ae6112d1d048e0964b039)

PR #12092, opened 2026-07-24
fix(frontend): reject over-context requests before streaming

Aphoh:
## Overview

Dynamo currently commits HTTP 200 for streaming requests before a backend can reject an
over-context token budget. That turns a client error into an inline server error, triggers
meaningless retries, and distorts monitoring.

This change makes each engine publish the request overflows that the frontend may reject
early. The shared OpenAI preprocessor applies only those advertised rejections, before
backend dispatch and before streaming headers are sent. Backend-owned clamping and
truncation remain in the backend.

## Summary

- add a minimal engine/frontend capability:

  ```text
  TokenBudget {
      combined_limit: N,
      reject_prompt_overflow: bool,
      reject_total_overflow: bool,
  }
  ```

  A `false` flag delegates that overflow dimension to the backend. Missing or malformed
  metadata also delegates both dimensions, so unofficial backends and version-skewed
  deployments stay online.

- publish the resolved backend behavior:

  | Backend mode | Reject prompt overflow early | Reject total overflow early |
  |---|---:|---:|
  | vLLM generation | Yes | Yes |
  | SGLang default | Yes | Yes |
  | SGLang `--allow-auto-truncate` | No | No |
  | SGLang total-token validation disabled | Yes | No |
  | TensorRT-LLM | Yes when `max_seq_len` is explicit; otherwise delegated | No |

- publish matching metadata for base and dynamic LoRA model cards; embedding-only and
  engine-less encode cards remain excluded where generation policy does not apply
- reject exact over-budget requests in the canonical preprocessor for Chat Completions,
  Responses, and legacy Completions
- defer prompt embeddings and multimodal lengths the frontend cannot prove
- preserve explicit requests on delegated paths; the frontend does not reproduce backend
  clamping or truncation
- preserve immediate streaming headers, SSE keep-alives, backend inactivity timeouts, and
  client cancellation for admitted requests
- propagate typed `InvalidArgument` as JSON HTTP 400 and classify it as a validation failure
  in frontend metrics without changing the backend's wire message
- document the capability and fail-open behavior for custom Python workers

### Why frontend admission is needed

The engine remains the source of policy. Standalone SGLang rejects an over-budget request in
its tokenizer manager before starting a stream. Dynamo's tokenized path bypasses that layer
and sends directly to the scheduler, where the same condition can surface only after HTTP 200
has been committed. Engine-published `TokenBudget` restores the backend's public admission
semantics without hard-coding one policy for every engine.

### TensorRT-LLM inferred-limit caveat

When `max_seq_len` is explicit, Dynamo publishes that limit and can reject prompt overflow
before backend dispatch. When it is omitted, TensorRT-LLM 1.3.0rc23 infers and may clamp
the realized limit inside its initialized executor worker, but does not expose that value
through the parent `LLM` API in the normal IPC/RPC execution path. Dynamo therefore omits
the token-budget capability and delegates overflow validation to TensorRT-LLM for this case.
The streaming preflight still propagates the backend rejection as HTTP 400 before SSE headers,
but default inferred-limit deployments do not receive frontend token-count admission or a
populated `context_length`. Dynamo intentionally does not duplicate TensorRT-LLM model-specific
inference; a future TRT-LLM realized-limit getter would let this path publish the same metadata.

### Test matrix

- publisher mappings for vLLM, the three distinct SGLang modes, and TensorRT-LLM
- exact combined-limit acceptance and one-token overflow rejection
- independent prompt-overflow and total-overflow delegation
- prompt-at-limit rejection and omitted-output handling
- missing and malformed metadata fail open without mutating the request
- prompt embeddings and inexact multimodal requests defer to the backend
- Chat, Responses, and legacy Completions wiring
- streaming typed `InvalidArgument` returns JSON HTTP 400 before SSE headers

### vLLM comparison

Using standalone vLLM and vLLM behind Dynamo with Qwen3-0.6B and `max_model_len=64`:

| Case | Standalone vLLM | Released Dynamo + vLLM | Patched Dynamo + vLLM |
|---|---:|---:|---:|
| Chat: 48 input + 17 output | 400 | 200 SSE, output clamped | 400 JSON |
| Chat: 48 input + 16 output | 200 SSE | 200 SSE | 200 SSE |
| Chat: 64 input, output omitted | 400 | 400 | 400 |
| Completions: 48 input + 17 output | 400 | 200 SSE, output clamped | 400 JSON |
| Responses disabled: explicit combined overflow | 400 | accepted when prompt itself fits | 400 JSON |

Standalone vLLM's Responses `truncation=auto` behavior is a separate prompt-transformation
capability. This PR does not emulate it in the frontend.

### Related work

This reconciles #11570. It includes that PR's Responses `max_output_tokens` handoff and
legacy Completions coverage, while adding:

- backend-published admission capability instead of a frontend-global context rule
- independent prompt and combined-total rejection/delegation
- vLLM parity for explicit combined overflow
- TensorRT-LLM's prompt-reject/total-defer contract
- reserved-token-aware SGLang limits and `--allow-auto-truncate`
- base and dynamic LoRA registration coverage
- inexact multimodal and prompt-embedding deferral
- the HTTP 400/metrics contract without delaying valid streaming headers

Co-authored coverage from Tianhe Zhang is credited in commit `8628dadd2b`.

- `components/src/dynamo/common/token_budget.py` and
  `lib/llm/src/local_model/runtime_config.rs` define the cross-language contract
- `lib/llm/src/preprocessor.rs` applies frontend rejection policy
- backend registration code publishes vLLM, SGLang, and TensorRT-LLM behavior
- `lib/llm/tests/preprocessor.rs` and backend runtime-metadata tests capture the matrix

## Related Issues

Related pull request: #11570.

## Validation

- `cargo clippy -p dynamo-llm --lib --tests -- -D warnings`
- focused `dynamo-llm` TokenBudget unit tests
- `cargo test -p dynamo-llm --test preprocessor context_length_validation` (9 passed)
- `cargo test -p dynamo-llm --test http-service test_http_service` (1 passed)
- repository-pinned Black and Ruff on changed Python files
- SGLang 1.3.0 container policy and LoRA matrix
- TensorRT-LLM runtime-metadata tests
- focused vLLM publisher validation
- original SGLang/Qwen3-8B streaming overflow reproduction
- standalone vLLM vs released/patched Dynamo comparison
- thermonuclear maintainability review, with findings incorporated

### Live backend parity validation

A pre-merge GPU differential test starts the native OpenAI server and Dynamo with the same
backend, model, and context limit. It pins upstream behavior and Dynamo behavior for a
within-budget request, streaming output overflow, and streaming prompt overflow. Accepted
clamp/truncate cases also compare token usage, so HTTP 200 alone cannot mask policy drift.

| Backend mode | Native behavior | Dynamo behavior | Focused CI |
|---|---|---|---|
| vLLM default | Reject / Reject at request time | Same HTTP 400 behavior | [passed](https://github.com/ai-dynamo/dynamo/actions/runs/30309543372) |
| SGLang default | Reject / Reject at request time | Same HTTP 400 behavior | [passed](https://github.com/ai-dynamo/dynamo/actions/runs/30309545930) |
| SGLang auto-truncate | Clamp output / truncate prompt | Same accepted result and token usage | [passed](https://github.com/ai-dynamo/dynamo/actions/runs/30309545930) |
| TensorRT-LLM default | Clamp output; prompt rejection arrives after HTTP 200 as an SSE error | Same rejection semantics, moved to HTTP 400 preflight | [passed](https://github.com/ai-dynamo/dynamo/actions/runs/30309514795) |

The TensorRT-LLM transport difference is intentional: native TensorRT-LLM currently commits
streaming headers before prompt-length validation. The prompt-rejection capability lets
Dynamo preserve the engine decision while returning the actionable client status before
headers.


rmccorm4:
Related: https://github.com/ai-dynamo/dynamo/pull/11570

lib/llm/src/preprocessor.rs
```diff
@@ -930,21 +986,25 @@ impl OpenAIPreprocessor {
...
+        let exact_prompt_len = Self::exact_prompt_len(
             preprocessed
                 .mm_routing_info
                 .as_ref()
                 .map(|mm| mm.expanded_prompt_len),
-            has_images,
+            preprocessed.multi_modal_data.as_ref(),
             preprocessed.token_ids.len(),
         );
+        if let (Some(prompt_len), Some(max_tokens)) =
```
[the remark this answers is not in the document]
Aphoh: Addressed in the current head. The legacy Completions operator calls `validate_preprocessed_token_budget` immediately after `builder.build()` and before `next.generate`; prompt embeddings and prompt lengths the frontend cannot prove remain delegated to the backend. Coverage includes the operator-level no-dispatch rejection case.

lib/llm/src/http/service/openai.rs
```diff
@@ -1958,14 +1958,28 @@ async fn chat_completions(
...
+            // A backend 400 is an engine-owned request validation failure.
+            // It may not use the frontend's "Validation:" message prefix.
+            let error_type = if error_response.0 == StatusCode::BAD_REQUEST {
+                ErrorType::Validation
+            } else {
+                extract_error_type_from_response(&error_response)
+            };
+            inflight_guard.mark_error(error_type);
+            error_response
+        })?;
```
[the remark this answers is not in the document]
Aphoh: Addressed in the current head. Streaming requests no longer await `check_for_backend_error` before constructing the SSE response; they return headers immediately and retain SSE keep-alives. Exact context-budget rejection now happens in the shared preprocessor from engine-published policy, before backend dispatch, so valid long-prefill requests keep their original first-token behavior while over-budget requests still return HTTP 400.

lib/llm/src/http/service/openai.rs
```diff
@@ -1958,14 +1958,28 @@ async fn chat_completions(
     // apply any annotations to the front of the stream
     let stream = stream::iter(annotations).chain(stream);
 
+    // Backends can report request validation failures as the first stream
+    // event. Inspect it before committing an HTTP status.
+    let stream = check_for_backend_error(stream)
```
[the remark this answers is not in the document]
Aphoh: Addressed in the current head by removing first-event preflight from the streaming path. `monitor_for_disconnects` again wraps the stream from the start, so backend inactivity timeout, client disconnect detection, and context cancellation remain effective. Strict token-budget failures are rejected during preprocessing before generation.

lib/llm/src/preprocessor.rs
```diff
@@ -930,21 +1017,12 @@ impl OpenAIPreprocessor {
...
-            has_images,
-            preprocessed.token_ids.len(),
-        );
+        let exact_prompt_len = Self::validate_preprocessed_token_budget(
+            &preprocessed,
+            self.strict_request_token_limit,
+        )?;
         if preprocessed.stop_conditions.max_tokens.is_none()
-            && let Some(prompt_len) = effective_prompt_len
+            && let Some(prompt_len) = exact_prompt_len
```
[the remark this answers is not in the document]
Aphoh: Addressed in the current head. Omitted output defaults use `strict_request_token_limit` when published (falling back to `context_length` only without a strict policy), and the derived budget is revalidated before dispatch. `test_omitted_output_budget_uses_strict_engine_limit` covers the reserved-token case.

biswapanda:
should we encode the behavior for prompt / output overflow (instead of single strict_request_token_limit)
```
TokenBudget {
    combined_limit: N,
    output_overflow: Reject | Clamp | Backend,
    prompt_overflow: Reject | Truncate | Backend,
}
```

Aphoh:
Implemented this as a typed `TokenBudget` contract in e649241f3b. The frontend now rejects only advertised `Reject` cases and leaves `Clamp`, `Truncate`, and `Backend` requests unchanged for the engine. Published mappings are vLLM Reject/Reject, SGLang default Reject/Reject, SGLang auto-truncate Clamp/Truncate, and TRT-LLM Clamp/Reject. SGLang reserved tokens, disabled total-token validation, and both vLLM/SGLang dynamic LoRA cards are covered. The strict review caught the SGLang LoRA registration gap before push; it now reuses the canonical runtime metadata builder and has focused coverage.

Aphoh:
CI update for `bdd13b2e24`:

- Focused live parity CI is green for [vLLM](https://github.com/ai-dynamo/dynamo/actions/runs/30309543372), [SGLang default + auto-truncate](https://github.com/ai-dynamo/dynamo/actions/runs/30309545930), and [TensorRT-LLM](https://github.com/ai-dynamo/dynamo/actions/runs/30309514795). Each run compares native-backend behavior with the Dynamo frontend for within-budget, streaming output-overflow, and streaming prompt-overflow requests.
- The same TensorRT-LLM parity test also passed in the full PR matrix. The unrelated TRT serve lane then lost several existing configurations during process startup; a targeted single-GPU rerun is in progress.
- All six failed deploy lanes hit the same cluster conversion-webhook `connection refused` error across vLLM, SGLang, and TensorRT-LLM.
- All three failed dev-image builds hit the same external NVIDIA ARM package 404 for `nsight-systems-2025.5.1`.

No current failure points to the token-budget implementation or parity suite.

lib/llm/src/preprocessor.rs
```diff
@@ -665,6 +771,9 @@ impl OpenAIPreprocessor {
 
         // // Initialize runtime config from the ModelDeploymentCard
         let runtime_config = mdc.runtime_config.clone();
+        let token_budget = runtime_config
+            .get_engine_specific::<TokenBudget>(TOKEN_BUDGET_RUNTIME_KEY)
```
nnshah1, line 978: This is the only `get_engine_specific` caller that propagates the error with `?`. The two siblings deliberately tolerate a bad value — `shared_cache.rs:242` matches `Err(error) => warn!(...) → None`, and `push_router.rs:408` uses `.ok().flatten()`.

With `?` here, a newer backend that publishes an `output_overflow`/`prompt_overflow` variant this frontend's enum does not know yet (version skew during a rolling upgrade) fails to deserialize and takes the whole model offline, instead of degrading to “defer to backend.” Malformed JSON does the same.

Suggest matching the sibling pattern: on `Err`, `tracing::warn!(%error, key = TOKEN_BUDGET_RUNTIME_KEY, ...)` and fall back to `None` (defer) rather than propagating. That also keeps the backend authoritative on the uncertain path.

— Neelay + 🤖

lib/llm/src/http/service/openai.rs
```diff
@@ -403,7 +403,7 @@ impl ErrorMessage {
             return (
                 StatusCode::BAD_REQUEST,
                 Json(ErrorMessage {
-                    message: dynamo_err.message().to_string(),
+                    message: format!("{VALIDATION_PREFIX}{}", dynamo_err.message()),
```
nnshah1: `VALIDATION_PREFIX` is now prepended to **every** `InvalidArgument` returned by `find_invalid_argument_in_chain`, not just the new token-budget rejections — this branch also carries backend-originated 400s (guided-decoding bounds, sampling params, etc.), whose wire message previously had no prefix.

It's consistent with the rest of the file, so likely intentional, but it's a user-visible message change riding in a token-budget PR and any client/test matching the raw backend message will shift. Worth a line in the PR description, or scope the prefix to the frontend-generated token-budget error if it was not meant to be blanket.

— Neelay + 🤖

lib/llm/src/preprocessor.rs
```diff
@@ -4775,42 +4898,281 @@ mod tests {
...
+    ) -> TokenBudget {
+        TokenBudget {
+            combined_limit,
+            output_overflow,
+            prompt_overflow,
+        }
+    }
+
+    #[test]
+    fn test_requested_token_budget_policy_matrix() {
```
nnshah1, line 5608: Minor: this matrix enumerates 8 cases that mostly re-exercise the same two branches (prompt-overflow reject, output-overflow reject). Behavior coverage > input coverage — consider trimming to exact-limit-accepted, one-token-over-rejected, one defer (Clamp/Backend), and the `usize::MAX` overflow case, and dropping the rest. Same applies to the 5-case `test_runtime_metadata.py` parametrize.

— Neelay + 🤖

components/src/dynamo/sglang/request_handlers/handler_base.py
```diff
@@ -411,6 +411,20 @@ async def load_lora(self, request: Optional[Dict[str, Any]] = None):
                                 else:
                                     lora_worker_type = WorkerType.Aggregated
                                     lora_needs = []
+
+                            # Reuse the base-model metadata builder so LoRA
+                            # cards advertise the same token-overflow policy,
+                            # parser configuration, and routing capabilities.
+                            # Keep this import lazy because the repository's
+                            # static test-collection environment intentionally
+                            # lacks parts of SGLang imported by register.py.
```
nnshah1: Nit: this 6-line comment over-narrates — the reuse intent is evident from `get_runtime_config`. The only non-obvious part worth keeping is *why* the import is lazy. Suggest trimming to one line, e.g. `# Lazy import: static test collection intentionally lacks parts of SGLang imported by register.py`.

— Neelay + 🤖

components/src/dynamo/vllm/main.py
```diff
@@ -634,6 +635,15 @@ def setup_vllm_engine(
     )
 
 
+def _set_vllm_token_budget(
```
nnshah1: Nit: `_set_vllm_token_budget` sits one underscore away from `set_vllm_token_budget` and exists only to skip `ModelType.Embedding`. The generation-only invariant now lives in two shapes — the LoRA path (`handlers.py`) calls the unguarded `set_vllm_token_budget` directly. Consider inlining `if model_type != ModelType.Embedding:` at the single call site, or renaming to something intent-revealing like `set_generation_token_budget`.

— Neelay + 🤖

components/src/dynamo/sglang/register.py
```diff
@@ -38,7 +44,9 @@
 SPEC_DECODE_RUNTIME_KEY = "spec_decode"
 
 
-def _register_model_source_path(engine: sgl.Engine, server_args: ServerArgs) -> str:
+def _register_model_source_path(
+    engine: Optional[sgl.Engine], server_args: ServerArgs
```
GuanLuo, line 42: Why making `engine` optional in every function?

components/src/dynamo/vllm/capacity.py
```diff
@@ -7,9 +7,30 @@
...
+    OutputOverflow,
+    PromptOverflow,
+    TokenBudget,
+    publish_token_budget,
+)
+
 logger = logging.getLogger(__name__)
 
 
+def set_vllm_token_budget(runtime_config: Any, max_model_len: int | None) -> None:
```
GuanLuo: Unify the helper name, TRTLLM worker uses different name

components/src/dynamo/vllm/main.py
```diff
@@ -634,6 +635,15 @@ def setup_vllm_engine(
...
+    runtime_config: ModelRuntimeConfig,
+    model_type: ModelType,
+    max_model_len: int,
+) -> None:
+    if model_type != ModelType.Embedding:
```
GuanLuo: What happen if we always set the budget?

GuanLuo (approved):

lib/llm/src/local_model/runtime_config.rs
```diff
@@ -20,6 +20,37 @@ pub use dynamo_parsers::tool_calling::StructuralTagSchemaMode;
 // Reserve a topology namespace so generated taints can be rebuilt without touching caller taints.
 pub const TOPOLOGY_TAINT_PREFIX: &str = "dynamo.topology/";
 
+/// Runtime-data key for an engine-published token-overflow contract.
+pub const TOKEN_BUDGET_RUNTIME_KEY: &str = "token_budget";
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum OutputOverflow {
```
nnshah1: Design question on the enum shape (tracing the consumers, the preprocessor only ever branches on `== Reject` — `runtime_config.rs` is the sole non-test reader, at `preprocessor.rs:426/435/1115`). `Clamp`, `Truncate`, and `Backend` are all pass-through today, and each publisher already picks Reject-vs-not from its own engine config. So for the *current* frontend behavior a per-axis `reject | defer` (bool) carries everything the code path uses, and it also removes the forward-compat liability I flagged at `preprocessor.rs:775` (an unknown future variant can't fail to deserialize a bool).

Is the three-way distinction load-bearing for a near-term `--output-overflow-policy` (frontend-side clamp/truncate, per the design thread), i.e. is it deliberately provisioning that? If so, makes sense to keep — but then could the deserialize default to *defer* on an unknown variant rather than failing model load? If Option B isn't imminent, `reject | defer` would be simpler and closes the version-skew hole. Either way the behavior is identical today; this is about the published contract's shape.

— Neelay + 🤖

Aphoh: Addressed in f6b35893bc. The contract is now combined_limit plus reject_prompt_overflow and reject_total_overflow booleans. False means defer to the backend. Both flags deserialize with a false default, and malformed metadata warns and falls back to no capability, so version skew remains fail-open.

lib/llm/src/preprocessor.rs
Aphoh: Addressed in f6b35893bc. The Rust matrix is reduced from eight cases to the five behavioral boundaries: exact limit, one over, full prompt, prompt defer, and total defer. The SGLang publisher matrix is reduced from five cases to the three distinct modes. I dropped the old usize::MAX case because prompt_len already entered the prompt-overflow branch, so it did not actually exercise saturating addition.

components/src/dynamo/vllm/capacity.py
Aphoh: Addressed in f6b35893bc. Backend helpers now consistently use publish_vllm_token_budget and publish_trtllm_token_budget.

components/src/dynamo/sglang/request_handlers/handler_base.py
Aphoh: Addressed in f6b35893bc. The lazy-import explanation is now one line.

lib/llm/src/http/service/openai.rs
Aphoh: Addressed in f6b35893bc. InvalidArgument messages are no longer prefixed on the wire. ErrorMessage now carries a serde-skipped metrics classification so the clean backend message is preserved while the request is still counted as a validation error; the streaming HTTP test asserts both.

components/src/dynamo/vllm/main.py
Aphoh: Addressed in f6b35893bc. The generation-only guard is inlined at the base-model registration call, the wrapper is removed, and the shared helper is renamed publish_vllm_token_budget.

components/src/dynamo/vllm/main.py
Aphoh: The budget only affects generation admission, so publishing it for an embedding-only card would currently be unused. I kept the generation-only guard to avoid advertising an unvalidated capability, but inlined it at the one base registration call in f6b35893bc; LoRA registrations are generation workers and publish directly.

components/src/dynamo/sglang/register.py
Aphoh: Narrowed in f6b35893bc. The new token-budget and bootstrap helpers now require an engine. The registration entry points and source-path helper remain optional because the SGLang multimodal encode path genuinely calls them with engine=None; that path has no tokenizer manager and intentionally omits this generation capability.

lib/llm/src/preprocessor.rs
Aphoh: Addressed in f6b35893bc. Invalid token_budget metadata now emits a structured warning and falls back to None, so overflow handling is delegated instead of taking the model offline. The focused preprocessor test also covers malformed metadata and startup fallback.

biswapanda (approved):

jh-nv (approved):

nnshah1 (approved):

dmitry-tokarev-nv (approved):
LGTM from DevOps side

connorcarpenter15 (approved):

galletas1712 (approved):

components/src/dynamo/common/token_budget.py
```diff
@@ -0,0 +1,43 @@
...
+    remains responsible for any clamping, truncation, or rejection after that.
+    """
+
+    combined_limit: int
+    reject_prompt_overflow: bool
+    reject_total_overflow: bool
+
+    def __post_init__(self) -> None:
+        if self.combined_limit < 0:
+            raise ValueError("combined_limit must be non-negative")
```
rmccorm4, line 29: TRTLLM test fails here: https://github.com/ai-dynamo/dynamo/actions/runs/31063332480/job/92503193545?pr=12092
```
          File "/opt/dynamo/venv/lib/python3.12/site-packages/dynamo/common/token_budget.py", line 28, in __post_init__
            if self.combined_limit < 0:
               ^^^^^^^^^^^^^^^^^^^^^^^
        TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

Aphoh: Fixed the startup failure in d9414b32e9. Clarification after tracing the exact TensorRT-LLM 1.3.0rc23 path: `max_seq_len=None` is resolved inside the initialized executor worker, but rc23 does not expose that realized value through the parent `LLM` API in the normal IPC/RPC path. The guard is therefore a compatibility fallback, not equivalent to publishing the inferred limit: Dynamo omits this capability and delegates validation to TensorRT-LLM. This PR still propagates the backend rejection as HTTP 400 during streaming preflight, but inferred-limit deployments do not get frontend token-count admission or populated `context_length`. The PR description now documents this caveat and the intended follow-up is a TRT-LLM realized-limit getter.

PR #12093, opened 2026-07-24
refactor(operator): organize APIs by group

carlory:
## Summary

- move Dynamo resource APIs under `deploy/operator/api/dynamo/{v1alpha1,v1beta1}` to organize API packages by group
- update Go consumers, Kubebuilder metadata, generators, CODEOWNERS, and source-path documentation
- preserve the public Kubernetes API identities `nvidia.com/v1alpha1` and `nvidia.com/v1beta1`

## Validation

- `GOCACHE=/tmp/dynamo-go-cache go test ./api/... ./internal/...`
- `GOCACHE=/tmp/dynamo-go-cache go test ./api -run TestFuzzRoundTrip -roundtrip-fuzz-iters=3000 -count=1 -v`
- generated Pydantic model validation
- `GOOS=linux GOCACHE=/tmp/dynamo-go-cache go test -run "^$" -exec /usr/bin/true ./cmd/snapshotctl ./internal/controller` in `deploy/snapshot`
- regenerated deepcopy code, CRDs, RBAC, API reference, Pydantic models, and CODEOWNERS
- `git diff --check`

Cluster E2E was not run locally because it requires an active Kubernetes test environment and installs external operators.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12093" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added Dynamo Kubernetes API resources for graph deployments, component deployments, models, profiling requests, checkpoints, scaling adapters, and pod snapshots.
  - Added support for topology placement, GPU memory services, failover, checkpoint policies, multinode deployments, and KV-transfer routing.
  - Added API version conversion support between v1alpha1 and v1beta1.

- **Bug Fixes**
  - Improved resource serialization by removing empty configuration artifacts while preserving meaningful and unknown fields.

- **Documentation**
  - Updated API references and development guidance to reflect the current Dynamo resource locations.


.github/codeowners/areas.yaml
```diff
@@ -223,6 +223,7 @@ areas:
   - components/src/dynamo/common/memory/
   - components/src/dynamo/common/tests/
   - components/src/dynamo/common/utils/
+  - components/src/dynamo/squeeze_evolve/
```
carlory, line 226: It is not related to this PR. but the ci is broken without this patch.

sttts:
Devil's advocate: why not `api/nvidia.com/{v1beta1,v1alpha1}` ? Of without the `.com` if that makes problems.

carlory:
@sttts  To be honest, the API package naming in the Dynamo project seems a little unusual. For comparison, consider these open source projects:

Kueue

- `config.kueue.x-k8s.io` for the Kueue operator configuration.
- `kueue.x-k8s.io` for Kueue features. 

LWS

- `config.lws.x-k8s.io` for the LWS operator configuration. the lws word is the project's name.
- `leaderworkerset.x-k8s.io` for LWS features
- `disaggregatedset.x-k8s.io` for PD features

In these examples, x-k8s.io is the shared subdomain, while the prefix indicates the API's purpose or scope.

Dynamo

- `operator.config.dynamo.nvidia.com` for the Dynamo operator configuration, even though its directory is simply named config. And it seems that the `operator` prefix is unnecessary.
- `nvidia.com` for Dynamo features.

In my opinion, `dynamo.nvidia.com` would be a better choice for the dynamo API group. It clearly identifies the API as belonging to Dynamo, whereas nvidia.com represents NVIDIA as a whole.
However, changing it now would introduce a breaking API change.

hhzhang16:
This MR seems to break Go source compatibility for consumers importing the old API packages, even though the Kubernetes GVK remains unchanged

carlory:
@hhzhang16  It seems that no other open source project import our API package.

- https://grep.app/search?q=%22github.com%2Fai-dynamo%2Fdynamo%2Fdeploy%2Foperator%2Fapi%2Fv1beta1%22
- https://grep.app/search?q=%22github.com%2Fai-dynamo%2Fdynamo%2Fdeploy%2Foperator%2Fapi%2Fv1alpha1%22

dagil-nvidia (commented):
Content-clean but integration-stale: at this PR's own base, every gate is green and the 195-file diff is a proven pure-mechanical `api/{v1alpha1,v1beta1}` to `api/dynamo/{v1alpha1,v1beta1}` rewrite with zero behavior change. But main has moved 657 commits since the base, and a textual merge now breaks the build. Nothing is wrong with the approach itself; it needs a rebase and a mechanical re-apply.

## Verified at the PR base (head 3bfa663, reviewed in a local worktree)

- CODEOWNERS regenerates byte-identical (615 rules, zero drift); strict path coverage passes; matcher tests 86/86; ownership routing for old paths, new paths, and `api/` root all resolve to the operator team, same as before the move.
- `go build ./...` and `go vet ./...` clean in `deploy/operator` and (GOOS=linux) `deploy/snapshot`; `go test ./api/...` green including the roundtrip fuzz package.
- deepcopy (controller-gen v0.17.3), CRD manifests, RBAC plus the helm-chart role copy, and the generated pydantic model all regenerate byte-identical to the committed files.
- Whole-diff multiset check: all 223 removed lines map 1:1 onto added lines after the single path substitution. The only non-move hunks are the squeeze_evolve ownership addition, its generated CODEOWNERS counterpart, a stale-path fix in `PROJECT`, and a default-path update in `generate_pydantic_from_go.py`.
- API identity is untouched (`Group: "nvidia.com"` in both `groupversion_info.go` files), so live CRs are unaffected.
- Tree-wide grep at the base: zero stale references to the old layout.

## Needed before merge (rebase requirements, not defects in the diff)

1. Rebase and re-apply the rewrite over current main. 53 Go files on main import the old `deploy/operator/api/{v1alpha1,v1beta1}` path and do not exist at this base (for example `internal/controller/dgd_*_reconciler.go`, `internal/checkpoint/compatibility.go`, `api/conversion_webhook_test.go`); a textual merge deletes the package they import and the operator module stops compiling. The two `docs/kubernetes/*.md` files this PR edits have since moved under `docs/fern/pages/**` (modify/delete conflict); the same stale strings now live in `dynamo-operator.md`, `api-reference-k8s.md`, and `full-api-reference.mdx`. After rebasing, rerun the same substitution over the new consumers, then `make generate manifests` and the CODEOWNERS generator.
2. Drop the `.github/codeowners/areas.yaml` squeeze_evolve hunk on rebase. It fixed a real coverage gap at this base, but main has since covered that path and routes it to the router team; keeping this line would silently re-route it. One more rebase trap the strict gate will not catch: main added a shared-ownership override for `deploy/operator/api/v1alpha1/dynamocheckpoint*` (operator plus gms); that glob must be rewritten to the new path or the checkpoint types silently lose co-ownership, because the parent directory still covers them.

## Minor

- `CONVERSION.md`'s updated command was verified live (`go test ./api/dynamo/v1alpha1 -count=1` passes); just confirm the file has not drifted on main during rebase.

Pre-existing, not this PR: the broken relative link in `recipes/kimi-k2.5/tokenspeed/agg/nvidia/README.md:221`, and `deploy/snapshot`'s darwin build failure (Linux-only syscalls without build tags).

Comment-only review; no approval implied and severity calls stay with the human reviewers.

dagil-nvidia (changes requested):
Requesting changes on the two rebase requirements detailed in the review comment above:

1. Rebase onto current main and re-apply the path rewrite over the 53 newer Go files that import the old `api/{v1alpha1,v1beta1}` layout, then rerun `make generate manifests` and the CODEOWNERS generator. As it stands, a merge breaks the operator build.
2. Drop the squeeze_evolve ownership hunk (main now routes that path to the router team), and rewrite main's shared `dynamocheckpoint*` ownership override to the new path so the checkpoint types keep their co-ownership.

The rewrite itself verified clean end to end at this PR's base - happy to re-review quickly after the rebase.

PR #12100, opened 2026-07-24
feat(audio): stream speech responses from vLLM-Omni

ptarasiewiczNV:
## Overview

Add incremental audio delivery to `POST /v1/audio/speech` for vLLM-Omni TTS workers. Voice clients such as Pipecat can begin playback while speech generation is still running instead of waiting for the complete utterance.

Binary PCM and WAV responses at the default playback speed stream incrementally. Formats that require a complete file, non-default speed adjustment, and URL uploads accumulate raw model output and encode or upload it exactly once before the frontend responds.

## Details

- Request incremental audio output from vLLM-Omni for every speech request so cumulative and per-step engine output can be normalized consistently.
- Track request-local audio state and emit every sample exactly once.
- Emit one open-ended WAV header followed by PCM chunks, with validated mono or stereo channel layout.
- Aggregate raw audio before encoding MP3 and other whole-file formats, applying speed adjustment, or uploading a URL response.
- Let the worker own codec and aggregation policy; the frontend streams binary worker responses and preserves URL responses as JSON.
- Validate the first non-empty worker response before sending HTTP headers, then preserve cancellation, backend errors, and request metrics for the body stream.
- Share streaming sampling-parameter construction between realtime and audio serving.

## Validation

- Model-backed candidate image built from implementation revision `223f843b5e6ced5f6123b0f79e2e456840fa34a8` with vLLM `0.26.0` and vLLM-Omni `0.26.0rc1`.
- Current head `da91878ed2c34609012c6a29b7795557a587459d` only merges three non-overlapping `main` commits; all ten PR files are unchanged from the tested revision.
- Focused vLLM-Omni formatter, audio-handler, and Omni-handler suite: 118 passed.
- Full Rust HTTP service integration suite: 17 passed, including incremental delivery, complete-file buffering, cancellation, validation errors, and metrics.
- Pre-commit hooks, Ruff, Rust formatting, and `git diff --check` passed.
- Model-backed E2E with `Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice`:
  - PCM: first body at 0.438 s, completion at 2.413 s, 18 client reads.
  - WAV: first body at 0.055 s, completion at 2.378 s, 20 client reads and one RIFF header.
  - MP3: headers/body at finalization (2.010 s), decoded as 24 kHz mono Layer III.
  - Speed-request WAV and URL modes withheld headers until finalization; URL output referenced a valid generated WAV.
  - WAV, MP3, and URL artifacts decoded as non-silent 24 kHz mono audio.
- Earlier Pipecat validation received 35 incremental audio frames through `OpenAITTSService`.

1. `lib/llm/src/http/service/openai.rs`: binary body streaming, first-response validation, and cancellation.
2. `components/src/dynamo/vllm/omni/output_formatter.py`: request-local streaming and aggregation state.
3. `components/src/dynamo/vllm/omni/omni_handler.py`: engine delta output and finalization.
4. `components/src/dynamo/vllm/omni/audio_handler.py`: request normalization and response-mode selection.

## Related Issues

- [x] [DYN-3482: Stream output chunks from `/v1/audio/speech`](https://linear.app/nvidia/issue/DYN-3482/voice-stream-output-chunks-from-v1audiospeech)

## Summary by CodeRabbit

* **New Features**
  * Added streaming audio responses for speech requests using non-URL audio sources.
  * Supports incremental PCM/WAV delivery, stereo audio, format-specific content types, and final audio aggregation.
  * Preserves complete-file JSON responses for URL-based audio requests.
* **Bug Fixes**
  * Improved handling of audio chunking, duplicate or empty chunks, cancellation, decoding errors, and backend failures.
  * Added validation for audio channel layouts and sample rates.
* **Tests**
  * Expanded coverage for streaming, buffering, cancellation, response formats, and audio shape handling.


components/src/dynamo/vllm/omni/audio_handler.py
```diff
@@ -187,8 +187,19 @@ async def build_engine_inputs(self, req: NvCreateAudioSpeechRequest):
...
+        output_format = (req.response_format or "wav").lower()
+
+        # URL delivery, whole-file encoders, and speed adjustment require the
+        # complete waveform before the worker can emit a response.
+        returns_audio_bytes = req.data_source != "url"
+        supports_chunk_encoding = output_format in {"pcm", "wav"}
+        uses_default_speed = req.speed is None or req.speed == 1.0
+        stream_audio = (
+            returns_audio_bytes and supports_chunk_encoding and uses_default_speed
+        )
```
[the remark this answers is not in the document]
ptarasiewiczNV: Solved with addition of `frontend_accepts_audio_chunks` that is only set by update frontend

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -295,21 +315,42 @@ async def format(
...
+                mm_output,
+                chunk_state=(
+                    stream_state if stream_state is not None else aggregate_state
+                ),
             )
+            if audio_np.size == 0:
+                return None
+
+            if aggregate_state is not None:
+                self._append_audio_chunk(aggregate_state, audio_np, sample_rate)
```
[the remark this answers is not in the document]
ptarasiewiczNV: Addressed. Empty audio payloads now return None for both streaming and aggregate generation and are skipped by the handler. The previous failed response remains only for non-incremental calls. Covered by test_empty_incremental_frame_is_skipped.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -354,6 +436,12 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
             )
 
         if isinstance(audio_val, list):
+            if chunk_state is not None:
+                new_audio = audio_val[chunk_state.emitted_chunks :]
+                chunk_state.emitted_chunks = len(audio_val)
+                audio_val = new_audio
+                if not audio_val:
+                    return np.empty(0, dtype=np.float32), self._sample_rate(mm_output)
```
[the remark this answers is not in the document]
ptarasiewiczNV: This follows vLLM-Omni’s native speech streaming contract: list payloads are cumulative, while tensor payloads are per-step deltas. In DELTA mode vLLM-Omni drains the audio modality after each snapshot, so the current Qwen3-TTS path emits fresh tensors, not repeated single-element lists. Both supported shapes are covered by tests; treating single-element lists as deltas would make cumulative one-chunk snapshots ambiguous and could duplicate audio.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -354,6 +436,12 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
...
             audio_val = torch.cat(audio_val, dim=-1)
```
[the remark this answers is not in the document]
ptarasiewiczNV: This follows vLLM-Omni’s native speech streaming contract: list payloads are cumulative, while tensor payloads are per-step deltas. In DELTA mode vLLM-Omni drains the audio modality after each snapshot, so the current Qwen3-TTS path emits fresh tensors, not repeated single-element lists. Both supported shapes are covered by tests; treating single-element lists as deltas would make cumulative one-chunk snapshots ambiguous and could duplicate audio.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -363,19 +451,85 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
...
+            16,
+            1,
+            num_channels,
+            sample_rate,
+            byte_rate,
+            block_align,
+            bits_per_sample,
+            b"data",
+            placeholder_size,
+        )
```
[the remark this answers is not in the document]
ptarasiewiczNV: Confirmed and intentional. Streaming starts before the final payload size is known, so the WAV header uses open-ended sizes matching vLLM-Omni’s native implementation. Producing final RIFF sizes would require buffering the complete waveform and defeat streaming. Pipecat/ffmpeg playback accepts this; clients saving a strict WAV file should finalize the RIFF/data sizes after the stream completes.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -363,19 +454,85 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
...
+            if audio_np.shape[0] != 1:
+                raise ValueError(
+                    f"Expected one audio batch, got shape {audio_np.shape}"
+                )
+            audio_np = audio_np[0]
+
+        if audio_np.ndim == 1:
+            return audio_np, 1
+
+        if audio_np.ndim == 2 and audio_np.shape[0] in (1, 2):
```
rmccorm4: Could we keep accepting frame-major audio here as well? vLLM-Omni's encoder handles both `(channels, frames)` and `(frames, channels)`, and the previous formatter also accepted `(frames, 1)` and `(frames, 2)`. With the current check, a generic model that returns the standard soundfile layout—for example `(2048, 2)`—gets a failed response. I think this should normalize either layout, with regression tests for frame-major mono and stereo.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -363,19 +454,85 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
...
 
-        return audio_np, sample_rate
+    def _encode_audio_chunk(
+        self,
+        audio_np: np.ndarray,
+        sample_rate: int,
+        fmt: str,
+        stream_state: AudioStreamState,
+    ) -> tuple[bytes, str]:
+        audio_np, num_channels = self._normalize_audio_layout(audio_np)
```
rmccorm4: Can we carry the first chunk's sample rate and channel count in `AudioStreamState` and validate the remaining chunks against them? The WAV header is emitted from the first chunk, but later chunks are encoded without checking either value. If one changes, the response keeps the original header and silently produces malformed audio. The aggregate path already rejects both mismatches, so the streaming path should do the same.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -11,8 +11,11 @@
 import asyncio
 import base64
 import logging
+import struct
+import tempfile
```
rmccorm4: `tempfile` looks left over from an earlier version and is unused now. `ruff check` reports F401 on this line, so it needs to be removed for lint to pass.

rmccorm4 (commented):
A few findings from a local review:

components/src/dynamo/vllm/omni/output_formatter.py
ptarasiewiczNV: fixed

components/src/dynamo/vllm/omni/output_formatter.py
ptarasiewiczNV: Good find, fixed

components/src/dynamo/vllm/omni/output_formatter.py
ptarasiewiczNV: Good point, addressed in 63a46f5. `AudioStreamState` now retains the first chunk’s sample rate and channel count, and subsequent chunks are validated through the same helper used by the aggregate path. Added regression coverage for both mismatch cases.

components/src/dynamo/vllm/omni/utils.py
```diff
@@ -20,6 +21,20 @@
...
+def streaming_sampling_params(
+    engine_client: Any, sampling_params_list: list[Any] | None = None
+) -> list[Any] | None:
+    """Return request parameters or engine defaults configured for streaming."""
+    source = (
+        sampling_params_list
+        if sampling_params_list is not None
+        else engine_client.default_sampling_params_list
+    )
+    if not source:
```
GuanLuo: `streaming_sampling_params` returns `None` when the engine exposes no `default_sampling_params_list`. Audio requests never supply their own list, so for them this fallback is the only path.

Two consequences, both silent:

- DELTA is never requested, so the engine keeps sending cumulative output. `_extract_audio_tensor` then re-emits it every step in stream mode, and re-appends it in aggregate mode (an O(n²)-length file).
- `inputs.sampling_params_list` stays `None`, so `is_multi_stage` in `_generate_with_lora_admission_lock` is false and a multi-stage pipeline releases the LoRA admission lock after the first output — the exact failure the assignment at `omni_handler.py:360-364` was changed to prevent.

Worth noting the only test for this helper (`test_omni_handler.py:106`) passes an explicit params list, so it never reaches this branch. Is an empty `default_sampling_params_list` reachable in a real TTS deployment?

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -363,19 +461,93 @@ def _extract_audio_tensor(self, mm_output: Dict[str, Any]) -> tuple:
...
+            audio_np = audio_np[0]
+
+        if audio_np.ndim == 1:
+            return audio_np, 1
+
+        if audio_np.ndim != 2:
+            raise ValueError(f"Unexpected audio shape {audio_np.shape}")
+
+        # Prefer vLLM-Omni's channel-first contract for ambiguous small shapes,
+        # while retaining compatibility with soundfile-style frame-major output.
```
GuanLuo: This tests `shape[0] in (1, 2)` before `shape[1]`, so a 2-frame frame-major mono chunk — shape `(2, 1)`, which is the `(N, 1)` layout this helper is meant to accept — resolves as 2-channel stereo.

Mid-stream that flips `num_channels`, `_validate_audio_metadata` raises "Audio channel count changed while generating", and since the 200 and the wav header have already gone out, the client receives a **truncated body under a success status** rather than an error.

The tie-break reads as deliberate given the comment directly above it, so the real question is whether a 2-frame chunk can reach here at all. If it can, this is a silent truncation; if it cannot, it is worth a line saying so.

GuanLuo (commented):
Four findings from a pass deliberately scoped to what earlier review here has **not** already covered. I read the existing threads first, so this leaves out the cumulative-vs-delta list question (settled above) and the `sampling_params_list` / LoRA thread (CodeRabbit's suggestion is applied verbatim at head).

Two of the four are in the `realtime_utils.py` → `utils.py` move. Its halves render far apart in the diff, so they are easy to miss together: 22 lines left the old file and 14 arrived in the new one.

Reviewed at `789de81`.

components/src/dynamo/vllm/omni/utils.py
```diff
@@ -11,6 +11,7 @@
 from vllm.sampling_params import SamplingParams
 from vllm_omni.distributed.omni_connectors.utils.serialization import OmniSerializer
 from vllm_omni.entrypoints.stage_utils import shm_read_bytes
+from vllm_omni.entrypoints.utils import coerce_param_message_types
```
GuanLuo, line 14: Moving this helper out of `realtime_utils.py` dropped its `try/except Exception → logger.warning → return None` wrapper and hoisted `coerce_param_message_types` from inside that `try` to module scope.

A vllm_omni build missing that symbol now fails the import of the entire `dynamo.vllm.omni` package. Before, it degraded this one feature and logged a warning.

The description frames this as sharing an existing helper between realtime and audio serving, which is why the dropped guard is easy to miss — the removal is in the other file.

lib/llm/src/http/service/openai.rs
```diff
@@ -4563,6 +4652,128 @@ async fn audio_speech(
...
+                        stream_handle.disarm();
+                        break;
+                    }
+                }
+            }
+        };
+
+        return Response::builder()
+            .header("content-type", content_type)
+            .body(Body::from_stream(body_stream))
```
GuanLuo, line 4896: Complete-file formats (mp3/flac/aac/opus, or `speed != 1.0`) are fully decoded into `first_chunks` before this point, so the entire body is already in memory — but it is still returned through `Body::from_stream`, so the response loses the `Content-Length` that `Body::from(audio_bytes)` set before this PR.

Clients that size the download or drive a progress bar from the header lose that for the buffered formats, which are exactly the ones where the length is known up front.

GuanLuo (approved):
Approved to unblock, the structure and workflow looks good to me. The comments are identified by agent which is reasonable but not blocking.

components/src/dynamo/vllm/omni/utils.py
ptarasiewiczNV: Thanks for raising this. Normal audio requests pass None, so we select engine_client.default_sampling_params_list, which vLLM-Omni populates once per initialized stage. We fixed the empty case by preserving [] instead of converting it to None. The empty list is passed to AsyncOmniEngine, which raises its canonical missing-stage-parameters error before producing output. This avoids silently falling back to uncoerced defaults and ensures successful requests retain the effective per-stage list for LoRA lock handling.

components/src/dynamo/vllm/omni/utils.py
ptarasiewiczNV: This is intentional, we need this function to be importable to prepare sampling params. If it's not available this means the there is broken vllm omni dependency / we need to update our workflow to a new omni version.

components/src/dynamo/vllm/omni/output_formatter.py
ptarasiewiczNV: True, fixed. We now use the state derived from the first chunk to set channel axis. In case of ambiguity (still possible for the first chunk) we log a warning and follow vllm omni channel first assumption.

lib/llm/src/http/service/openai.rs
ptarasiewiczNV: True, now I split `returns_audio_bytes` from `streams_audio_chunks`, which is now true only for PCM/WAV at speed 1.0. Only that path advertises chunk support and uses Body::from_stream. MP3/FLAC/AAC/Opus and speed-adjusted responses are fully collected and returned with Body::from(audio_bytes), restoring Content-Length.

components/src/dynamo/vllm/omni/omni_handler.py
```diff
@@ -351,8 +357,13 @@ async def _generate_openai_mode(
...
         }
-        if inputs.sampling_params_list is not None:
-            generate_kwargs["sampling_params_list"] = inputs.sampling_params_list
+        sampling_params_list = inputs.sampling_params_list
+        if inputs.request_type == RequestType.AUDIO_GENERATION:
+            sampling_params_list = streaming_sampling_params(
+                self.engine_client, sampling_params_list
+            )
+        if sampling_params_list is not None:
+            generate_kwargs["sampling_params_list"] = sampling_params_list
```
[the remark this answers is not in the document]
ptarasiewiczNV: Normal audio requests pass None, so we select engine_client.default_sampling_params_list, which vLLM-Omni populates once per initialized stage. We fixed the empty case by preserving [] instead of converting it to None. The empty list is passed to AsyncOmniEngine, which raises its canonical missing-stage-parameters error before producing output. This avoids silently falling back to uncoerced defaults and ensures successful requests retain the effective per-stage list for LoRA lock handling.

PR #12108, opened 2026-07-24
fix(mocker): model SGLang KV allocation by page

tedzhouhk:
## Overview

The SGLang mocker currently allocates and retains individual token-slot IDs even though SGLang's paged allocator owns and frees whole pages. This undercounts active partial-page occupancy, retains per-token metadata in the completed radix cache, and can produce capacity and eviction behavior that the real allocator cannot.

## Motivation

Offline replay maintains an independent SGLang radix tree for every simulated worker. On large production traces, retaining a token-slot ID for every token in every completed cached page makes the radix-tree memory footprint grow with both cached tokens and worker count; this became the dominant memory cost in high-worker-count Astra replay runs. Completed pages only need page/block identity, so this PR replaces that unnecessary per-token state with compact per-page metadata while preserving SGLang allocator semantics.

## Summary

- Replace the token-slot pool with a whole-page allocator while preserving flattened token indices only for bounded active requests.
- Store one router-compatible block hash and one physical page ID per completed radix page instead of per-token IDs.
- Make prefill, decode, destination reservations, eviction, retraction, and worker KV metrics account for physical page ownership.
- Protect a matched prefix before eviction and allocation, avoiding a false-OOM race when the matching prefix is the LRU victim.
- Hash and validate only the newly completed suffix during retained-prefix insertion, avoiding quadratic decode growth.

## Details

Completed radix edges now retain `LocalBlockHash + KvPageId` per page. Active partial pages consume their full physical capacity and extend in place until the page boundary; decode reserves new pages only when a request crosses that boundary. Finished partial tails release the entire page, while complete pages remain compactly indexed in the radix tree.

The 64-bit local block hash is intentionally the page identity, matching router-side indexing. Hash collisions are therefore treated as identical pages, rather than retaining an exact per-token collision guard.

At constrained capacity, replay results can intentionally differ from the previous mocker because unused positions in an active partial page are no longer exposed as allocatable capacity.

- `lib/mocker/src/cache/radix_cache.rs`: `PagePool`, compact radix representation, suffix-only insertion, and page-granular eviction.
- `lib/mocker/src/kv_manager/sglang_backend.rs`: request ownership, atomic match/protect/evict/allocate, decode and destination reservations, canonicalization, and router events.
- `lib/mocker/src/scheduler/sglang/{prefill,decode,core}.rs`: page-aware admission, reservation, and KV usage accounting.

## Validation

- `cargo clippy -p dynamo-mocker --lib --tests -- -D warnings`
- `cargo test -p dynamo-mocker --lib -- --test-threads=1` — 600 passed, 1 ignored
- `cargo bench -p dynamo-bench --bench offline_replay_bench --no-default-features --features replay-bench --no-run`
- Independent ownership and SGLang allocator semantic audits passed.
- Deterministic SGLang KV-router replay, two independent processes per revision/topology:
  - aggregated, 4 workers, 16,384 blocks/worker: baseline and candidate canonical reports match exactly;
  - disaggregated, 2P+2D, 131,072 blocks/worker: canonical reports match exactly;
  - disaggregated, 2P+2D, 16,384 blocks/worker: deterministic expected differences from whole-page pressure semantics; increasing capacity removes all differences.
- Paired release replay performance, 5 warmups + 30 pairs:
  - aggregated candidate/baseline median ratio: 0.762;
  - disaggregated candidate/baseline median ratio: 0.806.
- 5,000-request aggregated replay max RSS: 591,052 KiB to 339,860 KiB (-42.5%).

## Related Issues

## Summary by CodeRabbit

- **New Features**
  - KV-cache management now uses page-based allocation, improving handling of partial pages and cache reuse.
  - Decode and destination reservations more accurately track available cache capacity.

- **Bug Fixes**
  - Improved cache eviction and cleanup to prevent stale blocks and preserve active generated output.
  - Refined prefill, decode, activation, and retraction capacity calculations for more reliable scheduling.

- **Tests**
  - Updated coverage for page allocation, eviction, cancellation, re-prefill, and mixed decode workflows.

### DIFF
--- a/lib/mocker/src/scheduler/sglang/decode.rs
+++ b/lib/mocker/src/scheduler/sglang/decode.rs
@@ -20,6 +20,22 @@ pub(super) struct DecodeResult {
     pub(super) end_ms: f64,
 }
 
+fn decode_page_growth_needed(
+    running: &[SglangRequest],
+    block_size: usize,
+    max_burst: usize,
+) -> usize {
+    running
+        .iter()
+        .map(|req| {
+            let burst = max_burst.min(req.remaining_output_tokens());
+            let target =
+                super::config::ceil_to_block(req.current_sequence_len() + burst, block_size);
+            target.saturating_sub(req.allocated_tokens)
+        })
+        .sum()
+}
+
 fn decode_capacity_state(
     running: &[SglangRequest],
     kv_manager: &SglangKvManager,
@@ -31,15 +47,7 @@ fn decode_capacity_state(
     // Full partial pages are already owned by PagePool and excluded from
     // `actual_available`; subtracting their slack again would double-charge it.
     let logical_available = actual_available;
-    let page_growth_needed = running
-        .iter()
-        .map(|req| {
-            let burst = max_burst.min(req.remaining_output_tokens());
-            let target =
-                super::config::ceil_to_block(req.current_sequence_len() + burst, config.block_size);
-            target.saturating_sub(req.allocated_tokens)
-        })
-        .sum();
+    let page_growth_needed = decode_page_growth_needed(running, config.block_size, max_burst);
 
     (actual_available, logical_available, page_growth_needed)
 }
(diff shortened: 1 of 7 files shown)


lib/mocker/src/scheduler/sglang/decode.rs
```diff
@@ -263,14 +261,21 @@ pub(super) fn simulate_decode_step_with_sampler(
         unscaled_time
     };
 
-    let reserved_tokens = running
+    let reserved_page_tokens = running
```
dreamtalen: Non-blocking: can we extract this page-growth calculation logic into a helper? It is repeated in `page_growth_needed` at L34 and L112

dreamtalen (approved):

lib/mocker/src/scheduler/sglang/decode.rs
tedzhouhk: Done in 7657397c08. I extracted `decode_page_growth_needed` and now use it for the capacity preflight, post-retraction eviction, and decode-page reservation calculations. `cargo test -p dynamo-mocker scheduler::sglang` passes (54 tests).

PR #12109, opened 2026-07-24
fix(glm52): chat template parity fixes — arguments normalization and streaming tool_call recovery

peilii:
## Summary

- **Fix: tool_calls[*].function.arguments JSON string → dict before MiniJinja render** — GLM-5.2 Jinja template iterates arguments with \`{% for k, v in _args.items() %}\` which requires a dict; the wire schema stores arguments as a JSON-object string. Added \`normalize_tool_call_arguments()\` shared helper in \`preprocessor/prompt.rs\`, called from both \`NvCreateChatCompletionRequest::messages()\` (the \`/v1/chat/completions\` path RWLT uses) and \`UnifiedRequest::messages()\`.

- **Fix: streaming glm47 truncated tool_call → preserve as content** — When \`max_tokens\` is hit mid-\`<tool_call>\` block, \`glm47_parser\` drops the block silently (\`allow_eof_recovery=false\`). Added content-recovery to the streaming jail path in \`preprocessor.rs::apply_tool_calling_jail()\`: buffers input text, detects \`finish_reason=length\` with no tool_calls output, and emits the dropped text as a synthetic content chunk before the finish. Also covers the non-streaming aggregation path in \`aggregator.rs\`.

## Context

Discovered during GLM-5.2 NVfp4 GB300 RWLT benchmarking. Dynamo showed ~3K fewer ISL tokens and shorter OSL vs TRT-LLM on identical trajectories. OSL gap included 1315 truncated tool_call drops per SLO50 run (confirmed from frontend logs) and Dynamo prose-before-tool fraction of 24% vs TRT-LLM 88%. The streaming fix matches TRT-LLM behavior where partial XML is returned as content rather than silently dropped.

Companion fix in AA repo: \`peili/dynamo_conv_routing@767073e\` (AgentPerf \`setdefault\` for \`reasoning_content\`).

## Test plan
- [ ] Rebuild image from this branch
- [ ] Parity test: compare \`server_input_tokens\` on shared (conversation_id, conversation_idx, turn) pairs between Dynamo and TRT-LLM
- [ ] Confirm truncated tool_call drop warnings go to 0 in frontend logs
- [ ] Confirm ISL gap narrows with AA reasoning_content fix active

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12109" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/lib/llm/src/preprocessor/prompt.rs
+++ b/lib/llm/src/preprocessor/prompt.rs
@@ -35,13 +35,335 @@ pub trait MediaRequestExt {
     fn media_io_kwargs(&self) -> Option<&MediaDecoder>;
 }
 
+/// How a chat template expects tool_calls[*].function.arguments to be passed.
+/// Inferred once from the Jinja source at formatter construction; never from the
+/// served model name, which is arbitrary and not reliable for template detection.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Default)]
+pub enum ToolArgumentsMode {
+    /// Template receives arguments as a JSON-object string (OpenAI wire default).
+    #[default]
+    JsonString,
+    /// Template calls `.items()` on arguments (e.g. GLM-5.2's
+    /// `{% for k, v in _args.items() %}`), so arguments must be a parsed object.
+    ParsedObject,
+}
+
+pub fn detect_tool_arguments_mode(template: &str) -> ToolArgumentsMode {
+    // GLM-5.2: `{% set _args = tc.arguments %}{% for k, v in _args.items() %}`
+    if template.contains("_args.items()") || template.contains("arguments.items()") {
+        ToolArgumentsMode::ParsedObject
+    } else {
+        ToolArgumentsMode::JsonString
+    }
+}
+
+thread_local! {
+    /// Argument mode for the current formatter.render() call. Set by the preprocessor
+    /// immediately before the synchronous `apply_template` invocation; never across an
+    /// `.await` point. Using a thread-local avoids adding a field to
+    /// `NvCreateChatCompletionRequest` (which would require updating every struct literal).
+    static RENDER_TOOL_ARGUMENTS_MODE: std::cell::Cell<ToolArgumentsMode> =
+        const { std::cell::Cell::new(ToolArgumentsMode::JsonString) };
+}
+
+/// RAII guard that sets the thread-local tool-argument mode for the duration of a
+/// synchronous rendering call and restores the previous mode on drop.
+///
+/// SAFETY: Only use in a *synchronous* (non-`async`) scope with no `.await`
+/// between guard creation and drop. Thread-locals are not preserved across
+/// async executor boundaries — the task may resume on a different OS thread.
+pub(crate) struct ToolArgumentsModeGuard {
+    previous: ToolArgumentsMode,
+}
+
+impl ToolArgumentsModeGuard {
+    pub(crate) fn new(mode: ToolArgumentsMode) -> Self {
+        let previous = RENDER_TOOL_ARGUMENTS_MODE.with(|m| m.replace(mode));
+        Self { previous }
+    }
+}
+
+impl Drop for ToolArgumentsModeGuard {
+    fn drop(&mut self) {
+        RENDER_TOOL_ARGUMENTS_MODE.with(|m| m.set(self.previous));
+    }
+}
+
+pub(crate) fn get_tool_arguments_mode_for_render() -> ToolArgumentsMode {
+    RENDER_TOOL_ARGUMENTS_MODE.with(|m| m.get())
+}
+
+/// Extract the Jinja template source from a ModelDeploymentCard for analysis.
+///
+/// Priority order:
+/// 1. `mdc.chat_template_file` — standalone `.jinja` or `chat_template.json` file.
+/// 2. `mdc.prompt_formatter` — `tokenizer_config.json` with an embedded
+///    `"chat_template"` string (the normal HF layout for most models).
+///
+/// This covers both layouts so models that ship only a tokenizer config are not
+/// silently left in [`ToolArgumentsMode::JsonString`] when their template calls
+/// `.items()` on tool-call arguments.
+pub fn mdc_jinja_template_text(mdc: &ModelDeploymentCard) -> Option<String> {
+    fn read_embedded(checked_file: &crate::common::checked_file::CheckedFile) -> Option<String> {
+        let path = checked_file.path()?;
+        let contents = std::fs::read_to_string(path).ok()?;
+        let config: serde_json::Value = serde_json::from_str(&contents).ok()?;
+        let value = config.get("chat_template")?;
+        if let Some(s) = value.as_str() {
+            return Some(s.to_owned());
+        }
+        // Some HF configs store templates as [{name, template}, ...]. Concatenate
+        // so .items() in any variant is visible to detect_tool_arguments_mode.
+        if let Some(arr) = value.as_array() {
+            let combined: String = arr
+                .iter()
+                .filter_map(|v| v.get("template").and_then(|t| t.as_str()))
+                .collect::<Vec<_>>()
+                .join("\n");
+            if !combined.is_empty() {
+                return Some(combined);
+            }
+        }
+        None
+    }
+
+    if let Some(artifact) = mdc.chat_template_file.as_ref() {
+        match artifact {
+            PromptFormatterArtifact::HfChatTemplateJinja { file, .. } => {
+                if let Some(path) = file.path() {
+                    if let Ok(s) = std::fs::read_to_string(path) {
+                        return Some(s);
+                    }
+                }
+            }
+            // HfChatTemplateJson and HfTokenizerConfigJson both embed the template
+            // under the "chat_template" JSON key; read_embedded handles both.
+            PromptFormatterArtifact::HfChatTemplateJson { file, .. }
+            | PromptFormatterArtifact::HfTokenizerConfigJson(file) => {
+                if let Some(s) = read_embedded(file) {
+                    return Some(s);
+                }
+            }
+        }
+    }
+
+    // Fallback: normal HF layout stores tokenizer_config.json in mdc.prompt_formatter;
+    // chat_template_file is None unless a separate template file was present.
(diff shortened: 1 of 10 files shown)


rmccorm4:
@peilii can these types of fixes go directly to `main` and cherry pick into necessary branch after?

lib/llm/src/preprocessor/prompt.rs
```diff
@@ -35,13 +35,50 @@ pub trait MediaRequestExt {
     fn media_io_kwargs(&self) -> Option<&MediaDecoder>;
 }
 
+/// Whether a model's Jinja chat template expects tool_calls[*].function.arguments
+/// as a parsed object (dict) rather than a JSON-object string.
+/// GLM-5.2 does (`{% for k, v in _args.items() %}`); other models may not.
+pub(crate) fn template_wants_arguments_as_dict(model: &str) -> bool {
+    let m = model.to_ascii_lowercase();
+    m.contains("glm")
```
rmccorm4: Is there a smarter way to check this from template itself rather than based on model name?

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -419,6 +419,20 @@ impl DeltaAggregator {
                     choice.text = content.unwrap_or_default();
                 } else if is_harmony_parser(parser) && contains_harmony_protocol(&choice.text) {
                     choice.text = content.unwrap_or_default();
+                } else if matches!(
+                    choice.finish_reason,
+                    Some(dynamo_protocols::types::FinishReason::Length)
+                ) && choice.text.contains("<tool_call>")
```
rmccorm4: This "<tool_call>" check probably isn't correct -- this is looking for a leaking tool call tag (ex: "<tool_call>"), but the tool call tag may be different across models. Not all models will emit a token that translates to the literal text "<tool_call>". This is probably specific to GLM in the context of this PR, but the code doesn't represent that.

lib/llm/src/preprocessor/prompt.rs
```diff
@@ -35,13 +35,50 @@ pub trait MediaRequestExt {
...
+                                    obj.insert("arguments".to_string(), parsed);
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
rmccorm4, line 207: This is really deeply nested with { { { { } } } } -- can you flatten it? Try inverting some of the conditions to exit early for example or assign the arrays after the if conditions.

One example from codex:
```rust
pub(crate) fn normalize_tool_call_arguments(messages_json: &mut serde_json::Value) {
    let Some(messages) = messages_json.as_array_mut() else {
        return;
    };

    for message in messages {
        let Some(tool_calls) = message
            .get_mut("tool_calls")
            .and_then(serde_json::Value::as_array_mut)
        else {
            continue;
        };

        for tool_call in tool_calls {
            let Some(arguments) = tool_call.pointer_mut("/function/arguments") else {
                continue;
            };

            let Some(arguments_str) = arguments.as_str() else {
                continue;
            };

            let Ok(parsed) = serde_json::from_str(arguments_str) else {
                continue;
            };

            *arguments = parsed;
        }
    }
}
```

rmccorm4 (changes requested):
Please target fix to `main` first and cherry pick into your necessary feat branch after so we don't lose track of getting this to fix main for upcoming v1.4.0

rmccorm4:
> _Review generated with AI assistance_

Reviewed locally at `715f5ef` against merge-base `4dcdeca02`.

## Blocking

**1. Branch does not compile.** `cargo check -p dynamo-llm --lib`:

```
lib/llm/src/preprocessor.rs:2778:85  E0308: expected `&str`, found `&ChatCompletionMessageContent`
lib/llm/src/preprocessor.rs:2838:66  E0599: no method named `len` on `&ChatCompletionMessageContent`
lib/llm/src/preprocessor.rs:2865:65  E0308: expected `ChatCompletionMessageContent`, found `String`
```

`delta.content` is `Option<ChatCompletionMessageContent>`, an enum of `Text(String) | Parts(Vec<...>)` (`dynamo-protocols/src/types/chat.rs:488`). This is a design question rather than three casts: the recovery scheme assumes flat `String` content it can byte-count and slice, so the `Parts` variant needs an explicit decision. Matches CI — `rust-tests (.)` and `rust-clippy (.)` are red.

Also `prompt.rs:62`: `warning: unused doc comment` — a `///` comment cannot attach to a `thread_local!` macro invocation. Use `//` or move it inside. This alone fails clippy under `-D warnings`.

**2. `cargo fmt --check` fails in four places**: `prompt.rs:129`, `preprocessor.rs:77`, `preprocessor.rs:2775`, and a stray blank line added at `chat_completions.rs:128` that looks like an unrelated edit artifact.

**3. The renderer already performs this normalization on this exact path.** GLM-5.2's standalone `chat_template.jinja` is read into `config.chat_template` at `prompt.rs:335` and handed to `PromptFormatter::from_parts` — the same `HfTokenizerConfigJsonFormatter` whose `render()` already calls `normalize_tool_calls_arguments_in_messages()` (`dynamo-renderer/src/template/oai.rs:500`), gated on whether the template contains `arguments is string`.

That leaves two cases, and it would help to know which one you are hitting:
- The template has no `arguments is string`, the renderer normalizes already, and `detect_tool_arguments_mode` adds a second no-op pass.
- The template does contain it, the renderer is skipping deliberately, and this change silently overrides that opt-out. Per `dynamo-renderer/src/template.rs:109-120` the reason for the opt-out is that pre-parsing forces the `tojson` branch and re-emits with minijinja's compact separators, breaking append-only prefix matching across multi-step tool-use turns — the same ISL/prefix-reuse parity this PR is chasing.

Either way the detection belongs in dynamo-renderer next to the existing `template_handles_args_string`, not as a parallel mechanism in `messages()`. A rendered-prompt dump for a two-turn tool-use conversation would settle it.

**4. Speculative prefill renders with a different mode, producing a prefix mismatch.** `speculative_prefill.rs:155` calls `formatter.render(&prefill_request)` with no `ToolArgumentsModeGuard`, from a spawned background task, and `SpeculativePrefillRequest::messages()` (`speculative_prefill.rs:54`) never normalizes. So for GLM-5.2 the speculative path renders `arguments` as a JSON string while the main path renders it as an object. The warmed KV prefix will not match the real next-turn request, defeating speculative prefill precisely on tool-use turns — the same class of failure this PR is trying to fix.

**5. The vLLM/SGLang parity decision is overridden unconditionally.** `dynamo-parsers/src/tool_calling/parsers.rs:186-189`: "GLM-4.7 intentionally omitted: match upstream vLLM/SGLang behavior (drop the call when `</tool_call>` is missing)." Every glm47 deployment flips to TRT-LLM behavior, hardcoded on parser name, with no flag and no update to that comment. The new code comments acknowledge the markup leak, which is an improvement, but acknowledging it is not the same as making it opt-in.

## Design

**6. The thread-local is a hidden global for what is a per-request parameter.** `RENDER_TOOL_ARGUMENTS_MODE` (`prompt.rs:65`) couples `OpenAIPreprocessor` to `OAIChatLikeRequest::messages()` with no compile-time link. The guard at `preprocessor.rs:876` is correct today — `apply_template` is sync and nothing awaits inside the block — but that correctness rests on a comment, and finding 4 is the failure mode already present in-tree. The stated reason for avoiding a struct field (updating every struct literal) does not apply if the detection lives in the renderer, where the formatter already owns per-template flags.

**7. `detect_tool_arguments_mode` is a narrow substring match with a silent-miss gap.** `prompt.rs:53` matches only `_args.items()` and `arguments.items()`; the minijinja filter form (`arguments | items`, used in dynamo-renderer's own test at `oai.rs:1591`) does not match. And `mdc_jinja_template_text` (`prompt.rs:101`) returns `None` when the template is embedded in `tokenizer_config.json` rather than a standalone `.jinja`, so the fix silently does not apply to those deployments.

**8. Streaming and non-streaming now do opposite things on the same event.** Streaming (`preprocessor.rs:2831`) emits the raw `<tool_call>` markup as content for TRT-LLM parity; non-streaming (`aggregator.rs:422`) sets `choice.text = String::new()`. The same truncated turn yields markup with `stream=true` and an empty turn with `stream=false`.

The aggregator arm is also a behavior change relative to base, in the opposite direction from the PR's stated goal: before this PR the arm did not exist and `choice.text` retained the raw text; now it is wiped. If dropping is the intended non-streaming semantic, that deserves its own callout in the description rather than arriving inside a parity fix.

## Correctness

**9. Recovery chunk is emitted before its own chunk.** `preprocessor.rs:2878-2881` pushes `recovery` ahead of `nv_chunk`. The jail's `finalize()` emission can carry both the pre-marker prose and `finish_reason` in one chunk (jail entered mid-chunk, prose in `accumulated_content`), so the client receives `<tool_call>...` before the prose that preceded it. Appending `dropped` to the triggering choice's own `delta.content` on that chunk is order-safe and removes the clone/retain/usage-clearing entirely.

**10. The byte-offset heuristic is fragile.** `input_text[emitted_content_len..]` assumes emitted content is a verbatim prefix of jail input. The finalize path `trim_end()`s the prefix, and any path moving text into `reasoning_content` desyncs the offset. `.get(..).unwrap_or("")` prevents the panic but yields wrong text rather than nothing, so a garbled mid-word slice reaches the client. Locating the first `<tool_call>` in `input_text` and emitting from there is more robust.

## Testing

**11. No tests**, across 238 added lines. `lib/llm/tests/test_streaming_tool_parsers.rs` already drives `apply_tool_calling_jail` directly. Suggested minimum: truncated block recovered; complete block not recovered; n=2 with one truncated choice; non-glm47 parser is a no-op; `detect_tool_arguments_mode` over both the GLM-5.2 template and the `| items` filter form; and a two-turn render asserting the prompt prefix is stable.

## Minor

- Both new log messages contain a long run of embedded spaces from a wrapped string literal (`preprocessor.rs:2849`, `aggregator.rs:435`) and will render that way in logs.
- `ToolArgumentsMode`, `detect_tool_arguments_mode`, and `mdc_jinja_template_text` are `pub`, adding public API surface for what are internal helpers. `pub(crate)` matches the rest of the module.
- `normalize_tool_call_arguments` does `tc.pointer("/function/arguments")` and then re-fetches `tc.get_mut("function")` — two lookups where the renderer's equivalent does one.

nv-yna (approved):

PR #12120, opened 2026-07-24
refactor(operator): move component reconciliation into a workload program

julienmancuso:
## Summary

This MR introduces the first concrete composition-first boundary in the `DynamoGraphDeployment` reconciler.

The common DGD reconciliation flow now:

1. Resolves shared reconciliation inputs.
2. Builds ephemeral `graphReconcileState`.
3. Selects one complete workload program.
4. Invokes that program once.
5. Applies the resulting common status.

The component pathway is implemented as `componentProgram`, which now owns the complete DCD graph reconciliation flow:

- DCD generation and synchronization
- checkpoint startup policy application
- preservation of the immutable DCD backend framework
- restart annotation propagation
- managed rolling-update orchestration
- readiness calculation and status aggregation

The Grove pathway remains behind a temporary adapter around the existing Grove reconciliation method. This preserves current behavior while establishing the seam for a later dedicated `groveProgram` extraction.

`graphReconcileState` remains intentionally ephemeral. Results are published back to the common reconciliation flow only after the selected program completes successfully.

This MR is stacked on #12115.

## Motivation

The DGD controller currently selects between workload implementations inside a shared reconciliation method. As more pathways and workload APIs are added, that structure makes the common controller increasingly aware of provider-specific behavior.

This change establishes a composition-first model where each pathway is represented by a complete reconciler rather than a collection of lifecycle callbacks.

The common controller owns shared DGD concerns, while the selected workload program owns its workload graph and reconciliation ordering.

## Design

The new boundary consists of:

- `graphReconcileState`: ephemeral inputs and outputs for one reconciliation
- `workloadProgram`: the contract implemented by a complete workload pathway
- `componentProgram`: the concrete DCD-based implementation
- `groveProgram`: a temporary adapter around the existing Grove implementation
- `selectWorkloadProgram`: centralized pathway selection

The program contract deliberately contains a single `Reconcile` method. This keeps ordering and policy decisions inside each complete pathway instead of exposing provider-specific hooks to the common controller.

The existing managed rolling-update helpers remain on `DynamoGraphDeploymentReconciler` for now, but `componentProgram` owns when and how they participate in the component reconciliation flow. Moving those helpers behind a narrower dependency boundary is deferred to a follow-up MR.

## Behavior and compatibility

This is intended to be a behavior-preserving refactor:

- Pathway selection semantics are unchanged.
- Existing DCD generation and synchronization behavior is unchanged.
- Checkpoint startup behavior is unchanged.
[...]

PR #12121, opened 2026-07-24
refactor(operator): move rollout control into workload programs

julienmancuso:
## Summary

This MR moves workload rollout control out of the outer `DynamoGraphDeployment` reconciliation flow and into the selected workload program.

After finalization, the outer DGD controller now:

1. Builds the ephemeral reconciliation state.
2. Selects exactly one workload program.
3. Invokes that program once.
4. Projects the successful result into DGD status.

Each concrete program composes its complete ordered reconciliation flow.

The component pathway now performs:

1. Worker-hash migration.
2. Managed or unsupported-path rollout handling.
3. Shared resource reconciliation.
4. Component-specific multinode validation.
5. Restart-state resolution.
6. DCD workload reconciliation.
7. Shared checkpoint-readiness and scaling-adapter processing.

The Grove pathway performs the equivalent composition while retaining its existing unsupported managed-rollout behavior and temporary Grove workload adapter.

This MR is stacked on the preceding workload-program MR.

## Motivation

The outer DGD controller previously selected the workload pathway in one place while independently deciding rollout behavior in another.

That left the common controller coupled to component-specific decisions such as:

- whether managed rolling updates are supported;
- when to initialize worker hashes;
- when to start or advance a rollout;
- how unsupported pathways acknowledge worker changes.

It also meant the selected workload program did not yet own its complete graph-level state machine.

This change makes rollout behavior part of each concrete program’s composition while keeping shared resource operations reusable and explicit.

### Complete workload programs

The `workloadProgram` contract remains a single `Reconcile` method. No provider lifecycle callbacks such as `Prepare`, `Render`, `Scale`, or `Cleanup` are added.

`componentProgram.Reconcile` and `groveProgram.Reconcile` explicitly compose their operations in order. Earlier operations enrich `graphReconcileState`; later operations consume those resolved values.

### Shared reconciliation steps

The former `reconcileResources` flow is split into focused shared operations:

- `reconcileProgramInputs` reconciles common resources and records resolved checkpoint and multinode information.
- `resolveProgramRestartState` computes restart state after pathway-specific validation.
- `reconcileProgramResult` applies checkpoint-readiness policy and reconciles scaling adapters.

These are concrete shared operations composed by each program, not hooks driven by a generic framework.

### Rollout ownership

`componentProgram` now decides whether a component deployment uses:

- the existing managed rolling-update state machine for single-node component workloads; or
- the existing unsupported-path worker-hash behavior for multinode component workloads.

`groveProgram` explicitly composes the unsupported-path worker-hash behavior.

The underlying rolling-update helper implementations remain on `DynamoGraphDeploymentReconciler` temporarily. This MR moves control-flow ownership without combining it with a large mechanical dependency/receiver migration.

### Failure semantics

Program-specific rollout failures retain their existing DGD failure reasons:

- `failed_to_migrate_worker_hash`
- `failed_to_initialize_worker_hash`
- `rolling_update_failed`

Other program failures retain `failed_to_reconcile_the_resources`.

A program publishes `state.Result` only after its entire reconciliation sequence succeeds. A rollout, shared-resource, workload, or post-processing error therefore cannot replace the last-known-good component or restart status with a partial result.

## Behavior and compatibility

This is intended to be a behavior-preserving refactor:

- Finalization remains owned by the outer DGD controller.
- The outer DGD controller remains the only DGD status writer.
- Workload-program selection semantics are unchanged.
- Managed rolling updates remain limited to single-node component workloads.
- Grove and multinode workloads retain their existing unsupported-path behavior.
- Worker-hash migration and compatibility behavior are unchanged.
- Rollout reconciliation still occurs before shared resource reconciliation.
- Component multinode validation still occurs before restart-state resolution.
- Checkpoint readiness and scaling-adapter ordering are unchanged.
- Rolling-update status continues to override the overall DGD state where applicable.
- Results are still projected only after successful resource reconciliation.

There are no CRD, rendered-resource, ownership, watch-registration, or persisted-state changes.

## Tests

Added or updated coverage for:

- component-owned managed-rollout selection;
- managed rollout startup for single-node workloads;
- unsupported-path behavior for multinode component workloads;
- Grove unsupported-path failure handling;
- preservation of the previous program result on rollout errors;
- preservation of rollout-specific failure reasons;
- workload adapters independently from the complete program composition;
- shared resource validation before workload reconciliation.

The existing rolling-update characterization suite remains unchanged apart from transferring the managed-rollout support predicate to `componentProgram`.

## Summary by CodeRabbit

* **Improvements**
  * Improved deployment reconciliation reliability and status reporting.
  * Added clearer handling for worker rollout, migration, and resource reconciliation failures.
  * Improved rolling-update tracking and completion behavior.
  * Added support for pathway-specific reconciliation and more consistent readiness reporting.

* **Tests**
  * Expanded coverage for reconciliation failures, status preservation, workload selection, and rolling updates.

### DIFF
--- a/deploy/operator/internal/controller/dynamographdeployment_program.go
+++ b/deploy/operator/internal/controller/dynamographdeployment_program.go
@@ -19,6 +19,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	nvidiacomv1alpha1 "github.com/ai-dynamo/dynamo/deploy/operator/api/v1alpha1"
@@ -42,6 +43,7 @@ type graphReconcileState struct {
 	DGD             *nvidiacomv1beta1.DynamoGraphDeployment
 	HasMultinode    bool
 	RestartState    *dynamo.RestartState
+	RestartStatus   *nvidiacomv1beta1.RestartStatus
 	CheckpointInfos map[string]*checkpoint.CheckpointInfo
 	Result          ReconcileResult
 }
@@ -52,6 +54,31 @@ type graphReconcileState struct {
 // lifecycle callbacks.
 type workloadProgram interface {
 	Reconcile(context.Context, *graphReconcileState) error
+}
+
+type workloadProgramFailure struct {
+	reason Reason
+	err    error
+}
+
+func (e *workloadProgramFailure) Error() string {
+	return e.err.Error()
+}
+
+func (e *workloadProgramFailure) Unwrap() error {
+	return e.err
+}
+
+func failWorkloadProgram(reason Reason, err error) error {
+	return &workloadProgramFailure{reason: reason, err: err}
+}
+
+func workloadProgramFailureReason(err error) (Reason, bool) {
+	var failure *workloadProgramFailure
+	if !errors.As(err, &failure) {
+		return "", false
+	}
+	return failure.reason, true
 }
 
 // groveReconcileFunc is the temporary strangler seam around the existing Grove
@@ -71,25 +98,52 @@ type componentProgram struct {
 	lwsEnabled bool
 }
 
+// Reconcile composes the complete component pathway. Each earlier operation
+// enriches the ephemeral state consumed by the later workload and result
+// operations; only the final successful result is published back to state.
 func (p *componentProgram) Reconcile(ctx context.Context, state *graphReconcileState) error {
 	log.FromContext(ctx).Info(
 		"Reconciling Dynamo components deployments",
-		"hasMultinode", state.HasMultinode,
+		"hasMultinode", state.DGD.HasAnyMultinodeComponent(),
 		"lwsEnabled", p.lwsEnabled,
 	)
 
-	result, err := p.reconcile(ctx, state)
+	if err := p.reconcileWorkerRollout(ctx, state.DGD); err != nil {
+		return err
+	}
+	if err := p.reconciler.reconcileProgramInputs(ctx, state); err != nil {
+		return err
+	}
+	if state.HasMultinode && !p.lwsEnabled {
+		err := fmt.Errorf("no multinode orchestrator available")
+		log.FromContext(ctx).Error(
+			err,
+			err.Error(),
+			"hasMultinode", state.HasMultinode,
+			"lwsEnabled", p.lwsEnabled,
+		)
+		return fmt.Errorf("failed to reconcile Dynamo components deployments: %w", err)
+	}
+	p.reconciler.resolveProgramRestartState(ctx, state)
+
+	result, err := p.reconcileWorkloads(ctx, state)
+	if err != nil {
+		return fmt.Errorf("failed to reconcile Dynamo components deployments: %w", err)
+	}
+	result, err = p.reconciler.reconcileProgramResult(ctx, state, result)
 	if err != nil {
 		return err
 	}
+
 	state.Result = result
 	return nil
 }
 
-// reconcile owns the component pathway's complete DCD graph reconciliation.
+// reconcileWorkloads owns the component pathway's complete DCD graph
+// reconciliation.
 // Managed rolling-update helpers remain on the DGD reconciler until their
 // dedicated extraction; this program owns when they participate in the flow.
-func (p *componentProgram) reconcile(
+func (p *componentProgram) reconcileWorkloads(
 	ctx context.Context,
 	state *graphReconcileState,
 ) (ReconcileResult, error) {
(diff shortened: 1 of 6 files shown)


sttts:
> **Disclosure:** This comment was prepared and posted by OpenAI Codex at @sttts's request after reviewing this PR on top of #12115 and #12120.

I think this PR makes the important architectural move in the right direction: the selected program now owns the order and timing of rollout preparation, shared reconciliation steps, restart resolution, workload reconciliation, and result processing. The outer DGD reconciler selects one program and invokes it once rather than driving provider hooks through a common lifecycle.

So this is not an objection to the composition-first direction. My concern is narrower: before this boundary becomes the contract between separately packaged programs, I would give the program invocation a clearer input/output shape.

The current contract is:

```go
type workloadProgram interface {
    Reconcile(context.Context, *graphReconcileState) error
}
```

`graphReconcileState` currently contains several different categories of data:

| Current field | Actual role | Natural home |
|---|---|---|
| `DGD` | Mutable primary object and user intent | `Request.DGD` |
| `HasMultinode` | Resolved fact | A typed intermediate result, or immutable request facts if already known |
| `CheckpointInfos` | Result of shared checkpoint reconciliation | A typed intermediate result consumed by later steps |
| `RestartState` | State derived after earlier reconciliation and validation | A local typed value passed to the workload step |
| `RestartStatus` | Desired DGD status output | `Result.Status.Restart` |
| `Result` | Program output | Returned `Result.Status` |

This makes `graphReconcileState` simultaneously the request, scratch space, temporal protocol, and result. The ordering is therefore encoded indirectly through field mutation: one operation must populate a field before another operation may consume it. Those temporal preconditions are not visible in the function signatures.

That is especially relevant to the design goal here. The program correctly owns the flow, but the shared mutable state makes part of that flow implicit again.

## Suggested program contract

I would follow controller-runtime's `Request` / `Result` terminology:

```go
type workloadProgram interface {
    Reconcile(
        context.Context,
        Request,
    ) (Result, error)
}

type Request struct {
    // DGD is the mutable primary object for this reconciliation.
    //
    // The program may mutate and directly persist non-status fields. Status
    // persistence remains owned by the outer DGD reconciler.
    DGD *nvidiacomv1beta1.DynamoGraphDeployment

    // Facts contains immutable observations that were genuinely resolved
    // independently of the selected program. This can initially be empty.
    Facts ResolvedFacts
}

type Result struct {
    // Standard controller-runtime requeue semantics.
    ctrl.Result

    // Status is the complete desired DGD status produced by the program.
    // The outer reconciler persists it once. Nil means that the program did
    // not produce a successful status result.
    Status *nvidiacomv1beta1.DynamoGraphDeploymentStatus
}
```

I do not think we need a generic `PrimaryMutation`. The DGD is explicitly the mutable primary object. Program-specific non-status mutations, such as worker-hash annotations, are uncommon and can be written directly by the program that owns them. Finalizer ownership remains with the outer controller.

Status is different because there should be exactly one status writer. The selected program returns the desired `DynamoGraphDeploymentStatus`; the outer DGD reconciler adds controller-owned fields and conditions such as `Ready`, `ObservedGeneration`, and any common condition projection, then performs one `Status().Update()`.

The current code already has this useful property: DGD conditions are accumulated in memory and there is one central DGD status update. The proposed result type makes that ownership part of the program contract rather than an implicit property of the enclosing reconciler.

Conceptually:

```go
programResult, err := program.Reconcile(ctx, Request{
    DGD: dgd,
})
if err != nil {
    return ctrl.Result{}, err
}

if programResult.Status != nil {
    dgd.Status = *programResult.Status
}

// Add controller-owned Ready/ObservedGeneration/common conditions in memory.
// The existing outer status path persists the complete status exactly once.
return programResult.Result, nil
```

## Keep derived state local and typed

I would not move the current `reconcileProgramInputs` flow back into the outer controller merely to populate `Request.Facts`. That would give part of the ordering back to the framework shell.

Instead, the concrete program should continue to decide when shared operations run, but those operations should return structured values rather than mutate a universal state blob:

```go
func (p *componentProgram) Reconcile(
    ctx context.Context,
    req Request,
) (Result, error) {
    inputs, err := p.reconcileInputs(ctx, req.DGD, req.Facts)
    if err != nil {
        return Result{}, err
    }

    if inputs.HasMultinode && !p.lwsEnabled {
        return Result{}, fmt.Errorf("no multinode orchestrator available")
    }

    restart := p.resolveRestart(req.DGD, inputs)

    workloads, err := p.reconcileWorkloads(ctx, componentWorkloadRequest{
        DGD:         req.DGD,
        Inputs:      inputs,
        Restart:     restart.State,
        Checkpoints: inputs.Checkpoints,
    })
    if err != nil {
        return Result{}, err
    }

    status := req.DGD.DeepCopy().Status
    applyComponentResult(&status, restart, workloads)

    return Result{
        Status: &status,
    }, nil
}
```

The exact intermediate types are less important than the direction: each value should state what has been established and what the next operation may rely on. `Request` is the primary object plus immutable observations; `Result` is the requeue decision plus the complete DGD status. Derived rollout, restart, checkpoint, and workload state stays local to the concrete program and appears in the signatures that consume it.

## Comparison

| Current PR | Suggested refinement |
|---|---|
| Program owns complete control flow | Unchanged |
| Outer controller selects and invokes one program | Unchanged |
| Outer controller owns finalization and the single DGD status write | Unchanged |
| Programs directly reconcile child/provider resources | Unchanged |
| Non-status DGD mutations happen in their owning program | Unchanged |
| Mutable `graphReconcileState` carries input, scratch state, and output | Separate `Request` and returned `Result` |
| Later steps rely on fields populated by earlier steps | Typed intermediate values make dependencies explicit |
| `ctrl.Result` cannot be returned by a program | Embedded in the program `Result` |

This should be a focused contract refactor, not an architectural rewrite. The implementation already lands on the composition-first side with respect to flow ownership. I would make this change before extracting Grove and the other pathways into separate packages, while the boundary is still package-private and inexpensive to change. Otherwise `graphReconcileState` is likely to harden into the long-term provider protocol.

julienmancuso:
@coderabbitai full review

julienmancuso:
thanks @sttts , that makes sense,
I updated the MR.
there is one intentional difference from the example contract: Result.Status can be non-nil together with an error.
This preserves status progress produced by an earlier successful step.
for example, starting a rolling update or resolving checkpoint status, when a later step encounters a transient error. The request DGD’s status is never mutated directly, the partial status still travels through the result and is persisted once by the high-level reconciler.
On a completely successful reconciliation, Result.Status is the complete desired status. On error, it represents the accumulated status up to the failing step, after which the high-level reconciler adds the failure/Ready condition.
Does that error-result semantic match what you had in mind, or would you prefer the stricter rule that Status must always be nil when returning an error?

deploy/operator/internal/controller/dynamographdeployment_program.go
```diff
@@ -31,27 +32,93 @@ import (
...
 
-// graphReconcileState is ephemeral state shared by the common DGD flow and the
-// selected workload program during one reconciliation. It is intentionally
-// small: dependencies belong to concrete programs, and provider-native API
-// objects must not be added here.
-type graphReconcileState struct {
+// resolvedFacts contains immutable observations resolved independently of the
+// selected workload program. It is intentionally empty until such a fact is
+// demonstrated; program-derived values remain typed locals.
+type resolvedFacts struct{}
```
sttts: **Disclosure: This comment was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

I would remove both `resolvedFacts` and the `Facts` field from `workloadProgramRequest` for now.

The outer reconciler currently does not resolve any program-independent observations before selecting and invoking the program. All ephemeral computed state is intentionally derived inside the selected program and in sequence: `programInputs`, `programRestart`, the component rolling-update context, and Grove render inputs. There is therefore no concrete value that belongs in `Facts` today.

Keeping an empty placeholder anticipates an abstraction without a demonstrated contract and creates an obvious place for a generic state bag to grow back over time. The honest current boundary seems to be:

```go
type workloadProgramRequest struct {
    // DGD is the mutable primary object. Programs may persist non-status
    // mutations directly; proposed status is returned through Result.
    DGD *nvidiacomv1beta1.DynamoGraphDeployment
}
```

If a genuinely program-independent observation is later computed outside the selected program, we can add it then with a domain-specific type and name. Until that exists, the minimal request makes the ownership boundary clearer.

sttts (commented):
One request-contract simplification.

PR #12137, opened 2026-07-24
feat(examples): restructure hello_world as a two-stage ladder with a custom-engine stage

### LINKED ISSUE #12136: [CONTRIBUTION]: Hello-world example for writing custom unified backends (frontend + KV routing primitives, no GPU)
### Type of Change

New feature

### Problem Statement

Dynamo's core primitives — the frontend/engine split (who tokenizes, who generates), model registration and discovery, the `LLMEngine` lifecycle, KV events, and the router's prefix matching — are individually documented but hard to understand *concretely* until you watch them interact in a system you can read end to end. Real backends (vLLM/SGLang/TRT-LLM) are too large to learn from, and every moving part normally requires GPUs and model weights just to observe.

This gap matters most for the exact audience the unified backend targets: **teams bringing their own inference engine to Dynamo.** Today their path is the guide (`docs/development/unified-backends.md`) plus reverse-engineering `sample_engine.py` — there is no runnable, minimal, out-of-tree-shaped example of a custom engine going through the full journey: implement `LLMEngine` → serve through the standard OpenAI frontend → publish KV events that drive KV-aware routing → containerize → deploy as a `DynamoGraphDeployment`.

The existing `examples/custom_backend/hello_world` demonstrates the raw runtime endpoint with a custom client — its README notes it "does not demonstrate the standard Dynamo FrontEnd-Backend deployment." The in-tree `sample_engine.py` is a good API reference but isn't packaged as an example, doesn't cover the container/Kubernetes path, and its counter-based KV block hashes mean repeat prompts don't actually demonstrate routing stickiness.

### Proposed Solution

Add `examples/custom_backend/hello_engine/`: a hello-world **custom unified backend** — a ~235-line `LLMEngine` that streams a hardcoded sentence. No GPU, no model weights, no downloads, so every Dynamo primitive is observable on a laptop:

- **the contract, readable**: `from_args → start → generate → cleanup` with cancellation, `max_tokens`, finish reasons, and usage handled per the docs
- **KV routing you can watch work**: events with content-derived block hashes, so identical prompts genuinely pin to the worker holding the prefix (verified: `0/13 → 12/13` block overlap in router scoring; a 2-worker DGD pins repeat prompts 3–0)
- a bundled ~6 KB byte-level mock tokenizer (256 tokens + EOS, generator included) so registration downloads nothing
- Dockerfile on the published `dynamo-frontend` base + a v1beta1 `DynamoGraphDeployment` (1 frontend + 2 CPU-only workers)
- README with a plain-English walkthrough of the `LLMEngine` contract and how KV events reach the router, plus Docker and Kubernetes runbooks

**Who this helps:** anyone integrating a custom engine via the unified backend path copies this as their starting skeleton; anyone learning Dynamo gets a complete, inspectable system in ~250 lines.

Validated on bare-metal, plain Docker (file discovery + ZMQ event plane), and a kind cluster DGD (kubernetes discovery + NATS event plane) — together exercising both event planes.

### Estimated PR Size

XL (501-1000 lines)

### Files/Components Affected

- `examples/custom_backend/hello_engine/` (new)
- `examples/README.md` (index entry)
- `docs/development/unified-backends.md` (cross-link)
- No runtime/library code modified.

marckarp:
## Overview:

Restructures `examples/custom_backend/hello_world/` into a two-stage ladder and adds the new second stage: a hello-world custom **unified backend** — a ~235-line `LLMEngine` that streams a hardcoded sentence and serves through the standard OpenAI frontend with KV-aware routing. No GPU, no model weights, no downloads.

```
examples/custom_backend/hello_world/
├── README.md   # the ladder: which stage to start with
├── basic/      # the original minimal runtime example (unchanged, paths updated)
└── engine/     # NEW: complete custom engine on the unified backend contract
```

## Details:

Per review discussion, rather than a separate sibling example, the engine lives as stage 2 of hello_world: `basic/` stays the minimal runtime intro (endpoint + worker, no model), `engine/` is the copyable skeleton for custom-engine integrators and a complete, inspectable Dynamo system for learners.

- `LLMEngine` contract: `from_args → start → generate → cleanup`, with cancellation, `max_tokens`, finish reasons, and usage
- KV events with content-derived block hashes → repeat prompts genuinely pin to the worker holding the prefix (flow diagram in the engine README)
- bundled ~6 KB byte-level mock tokenizer (generator included) → model registration downloads nothing
- Docker build + CPU-only `DynamoGraphDeployment` for both stages
- Docs updated: examples index, unified-backends guide link, examples reference page. The runtime development guide references the Rust runtime `hello_world` under `lib/`, unaffected.

## Validation
- Pure reorganization + docs for the restructure commit: files moved as 100% renames, no code changes
- Pre-commit hooks pass on all touched files; yaml parses; doc link targets verified
- Engine behavior validated in the original commit (end-to-end serving + KV routing logs shown in the README)

## Related Issues

- Closes #12136


marckarp:
Restructured per the ladder discussion: hello_engine is now the second stage of hello_world rather than a separate example.

- `examples/custom_backend/hello_world/basic/` — the original minimal runtime example, unchanged (paths updated)
- `examples/custom_backend/hello_world/engine/` — the custom engine on the unified backend contract
- A top-level README explains the progression and when to start where
- Docs links updated (writing-unified-backends, examples reference); the runtime development guide references the Rust runtime hello_world under lib/, so it's unaffected

No code changes in the restructure, pure reorganization. Also rebased onto main and all commits are signed.

On the earlier CodeRabbit review: all findings are addressed in the current code — explicit exceptions instead of asserts (engine.py raises RuntimeError before start; the tokenizer generator raises on round-trip/EOS mismatch), `max_tokens=0` handled via an `is None` check with an empty terminal `length` chunk, cancellation re-checked immediately after the await, README describes tokens as bytes, and the layout fence carries a language tag.

@coderabbitai resolve

tanmayv25 (commented):
### Review summary

> [!NOTE]
> **This review was AI-assisted.** It was produced with Claude Code running a deep-review skill: three scoped review passes (framework-contract conformance, reader/structure advocacy, ops & repo conventions) plus an independent unprimed cross-check. Every claim below was verified against the framework source before posting, and two candidate findings were dropped because they did not survive that check. Please still apply your own judgement — treat these as findings to confirm, not verdicts.

The two-stage ladder is the right call, and the teaching content is above the bar for examples in this repo. The routing table in `hello_world/README.md` is genuinely the best-written part of the PR — "Start here if" gives a real decision rule rather than a difficulty label. Two things need fixing before merge, though, and one of them undercuts the example's main claim.

#### Blockers

**1. The move breaks a pre-merge CI test that this PR does not update.**
`lib/bindings/python/tests/test_example_hello_world.py:28` still resolves to `examples/custom_backend/hello_world` and then runs `python3 hello_world.py` / `client.py` with that as cwd (`:36`, `:56`). Both moved into `basic/`. The test is marked `pre_merge` + `integration` + `gpu_0` (`:14-18`), and the diff does not touch `lib/bindings/python/tests/`. Fix: append `"basic"` to the join at `:28`. (Not inline-commentable — the file isn't in the diff.)

**2. The KV publishing is incorrect beyond the exact-repeat demo, and the README documents the wrong mechanism.** See the inline comments on `engine.py:52` and `engine.py:224`. In short: `parent_hash` is never passed, so new tail blocks are anchored at the radix root instead of under the shared prefix; and the "skip already-published blocks" dedup forges adjacency between non-adjacent blocks. The demo only works because an identical repeat prompt publishes nothing at all. Since `engine/` is the one artifact in the tree where KV routing visibly works, this is the highest-value thing to get right.

#### Structural

- **There is no way to run `engine/` without Docker.** Stage 1 is two `python` commands; stage 2's first executable instruction is `docker build`. Inline comment on `README.md:63`.
- **The two stages don't share a shape.** `basic/` is *learn → run → read → deploy*; `engine/` is *deploy → read*, with the walkthrough 135 lines down. Inline comment on `README.md:199`.
- **Three artifacts now claim "how to write a custom engine"** — the unified-backends guide, `sample_engine.py` (+ its README and AGENTS.md), and this. Nothing says which is canonical, and the new contract table is a fourth, drifted copy. Suggest stating the split explicitly: guide = the contract, `sample_engine.py` = the reference implementation, `engine/` = the deployable end-to-end walkthrough. The README already has the right instinct at `:288-290` and `:311-315` — just finish it.
- **No test coverage.** Nothing outside the example directory references `hello_engine`. An example shipping an image build, a DGD manifest and a KV-routing claim, with nothing asserting any of it, will drift.

#### Smaller items not worth an inline comment

- `examples/README.md:85` says "`hello_world` and `hello_engine`" — `hello_engine` is a package name, not a directory a reader can find. Suggest `hello_world/basic` and `hello_world/engine`. The same line still says "Python 3.9+", now contradicted by the new subproject's `>=3.11`.
- `hello_world/README.md:29` links out of `docs/` with a relative `../` escape. It resolves, so the linter stays quiet, but `docs/fern/AGENTS.md:24` asks for absolute `github.com/ai-dynamo/dynamo` URLs for targets outside `docs/`.
- `engine.py` banner comments: the bracketed timing hints (`[called once per request]`) are genuinely useful and not derivable from the file — keep those. The box-drawing frames and the banners that restate the method name below them read as narration. Worth noting the inversion: `cleanup()` gets a full banner box, while `model_name=args.tokenizer_repo` (`:116`) — the frontend's tokenizer-resolution path, and the single line a copier most needs to understand — gets five words.
- Dead `[project.scripts] hello-engine` (nothing invokes it; Dockerfile, DGD and README all use `python3 -m`). `uv pip install tokenizers` is redundant — the base image pulls it transitively via `transformers`. `COPY README.md` is unused by hatchling and invalidates the layer cache on every README edit.
- `basic/deploy/hello_world.yaml:9` still says `backendFramework: vllm` for an example with no vLLM — pre-existing, but the PR is already editing that file.
- Consider a dual-owner entry in `.github/codeowners/areas.yaml` for the new Dockerfile, matching the tritonserver/diffusers/mocker pattern. CI's coverage gate passes either way.

#### Verified correct — flagging so nobody re-reviews these

The `LLMEngine` ABC is fully and correctly implemented (all four abstract methods, correct signatures, `from_args` correctly a classmethod); `run()` wiring and sync `main()`; every `WorkerConfig` / `EngineConfig` / `LlmRegistration` field name, default and required-ness; `served_model_name=None` deferral; the `finish_reason` values and `completion_usage` shape; `is_stopped()` as the cancellation API; `python3` resolving to the venv and `--no-deps` being right; the DGD schema (`components` really is an array, `name: main` is required, and omitting `backendFramework`/probes/resources is correct because the operator injects them); and `tokenizer.json` matching `generate.py` by inspection.

Also worth stating: the `DYN_*` env fallbacks in `from_args` initially looked like redundant duplication, but they are **necessary** — `RuntimeConfig::apply_to_env` (`lib/backend-common/src/worker.rs:98-121`) writes set fields *into* the environment, overwriting operator-injected values, so without them `DYN_DISCOVERY_BACKEND=kubernetes` would be silently clobbered by `"etcd"`. Good catch by the author; leaving it documented here so a future reviewer doesn't "simplify" it away.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,243 @@
...
+        # [called once after start()] "Build me a publisher and hand it
+        # to the callback below."
+        return [PushSource(on_ready=self._on_publisher_ready, dp_rank=0)]
+
+    def _on_publisher_ready(self, publisher) -> None:
+        # The framework hands us a live KvEventPublisher (wired to ZMQ
+        # or NATS — we never know which). We just keep it.
+        self._publisher = publisher
+
+    def _publish_prompt_blocks(self, prompt_tokens: list[int]) -> None:
```
tanmayv25, line 219: **Blocker — this breaks KV routing for anything but the exact-repeat demo.**

Two distinct defects here:

**`parent_hash` is never passed.** With `parent_hash=None` and no `start_position`, the indexer anchors the run at position 0 / the root — see `lib/kv-router/src/indexer/README.md:250-266`. Since this method publishes only *new* blocks, the ordinary case (shared system prompt + new suffix) grafts the tail blocks at the root. Two consequences: the worker is never credited for the tail on a later request, and any future prompt *beginning* with those tokens scores a false prefix hit.

**Deduping intermediate blocks forges adjacency.** If blocks 1 and 3 are new but block 2 was published earlier, this emits `token_ids = B1+B3` as one contiguous run — asserting to the router that B3 follows B1. The same defect collapses a prompt containing two identical 16-token blocks.

The demo works only because a second identical prompt publishes nothing at all.

Suggested fix: derive the external hash as a real chain (`compute_next_sequence_hash`, `lib/tokens/src/lib.rs:77-83`), publish the prompt's full contiguous block run anchored with `parent_hash` at the last already-known block, and drop the `_published_blocks` dedup entirely. Re-stores are idempotent in the tree, so republishing the full prefix is the right thing for a hello-world to do — and it's simpler code.

That also fixes a side issue: `_published_blocks` (`:67`) grows without bound for the process lifetime, and there's no `publish_removed` counterpart, so this worker's overlap score is permanently inflated. `sample_engine.py:280-284` models the release side.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,243 @@
...
+
+HARDCODED_REPLY = (
+    "Hello! I am a hand-written Dynamo engine. I run no model - every "
+    "token you are reading was hardcoded by a human. I exist to show "
+    "how a custom engine plugs into Dynamo's frontend, router, and "
+    "KV-aware routing."
+)
+
+
+def _block_hash(token_block: list[int]) -> int:
```
tanmayv25: This docstring teaches the wrong mental model, which matters a lot in a file whose stated purpose is showing newcomers how KV-aware routing works.

Prefix matching does **not** key on this hash. It keys on `tokens_hash`, which the framework recomputes itself from the `token_ids` you pass — `compute_block_hash_for_seq(...)[0]` at `lib/kv-router/src/zmq_wire/convert.rs:380-402`. The `block_hashes` argument is an `ExternalSequenceBlockHash`: node *identity* used for parent resolution, `publish_removed`, and the `lookup` map — not the match key.

`sample_engine.py:265` demonstrates this by using a plain monotonic counter for the same field.

So the blake2b content hash buys nothing for matching, and actively hurts the identity contract: identical content at two different sequence positions collapses to one node key. Content-only hashing is exactly what sequence-hash chaining exists to prevent.

Worth rewriting this docstring and `README.md:306-309` together.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,243 @@
...
+            if len(reply) == len(self._reply_token_ids)
+            else "length",
+            "completion_usage": _usage(len(reply)),
+        }
+
+    # ------------------------------------------------------------------
+    # cleanup   [called at shutdown; must be idempotent + null-safe]
+    # ------------------------------------------------------------------
+    async def cleanup(self) -> None:
+        self._publisher = None
```
tanmayv25, line 203: Minor shutdown race: `cleanup()` nulls `_publisher` while an in-flight `generate()` may already have passed the `if self._publisher is None` guard at `:226`, giving an `AttributeError` on `publish_stored`. Narrow, but easy to avoid — take a local reference in `_publish_prompt_blocks` and use that.

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+What the framework calls, and when:
+
+| Method | Required? | When the framework calls it | Ours does |
+|---|---|---|---|
+| `from_args()` | yes | first — parse CLI, return `(engine, WorkerConfig)` | flags for model name, namespace, discovery |
+| `start()` | yes | once at boot — load your model, return `EngineConfig` | load bundled tokenizer, encode the hardcoded sentence |
+| `generate()` | yes | once **per request** — yield token chunks | stream the sentence one token at a time |
+| `cleanup()` | yes | at shutdown — must be idempotent + null-safe | drop references |
+| `abort()` / `drain()` | no (no-op defaults) | on client cancel / before shutdown | not overridden |
```
tanmayv25: **There is no `drain()` method on `BaseEngine`.** The hooks are `abort()` (`components/src/dynamo/common/backend/engine.py:208`) and `is_quiescent()` (`:220`); draining is the Worker's job, not the engine's. A reader taking this table as the contract will go looking for something that doesn't exist.

Also, the heading at `:201` says "five methods" over a six-row table naming seven.

Broader suggestion: **delete this table and link to the unified-backends guide instead.** It's a fourth copy of the lifecycle — `components/src/dynamo/common/backend/AGENTS.md:11-70` and that directory's `README.md:19-37` already have it, correctly. Deleting removes both the bug and the drift surface in one move, and the surrounding prose (the "framework calls YOU" framing, the `generate()` rules, the code sample) is the genuinely valuable part and stands on its own.

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+for p in $(kubectl get pods -n <namespace> -o name | grep hello-engine-worker); do
+  echo "$p: $(kubectl logs -n <namespace> $p | grep -c 'request received') requests, \
+$(kubectl logs -n <namespace> $p | grep -c 'published') KV publishes"
+done
+```
+
+Expected: **all requests on one worker, zero on the other** — the router
+learned which worker holds the prefix from the engine's KV events and
+pinned every repeat prompt to it. One `published` line despite several
+requests is the engine's dedup (identical blocks are published once).
```
tanmayv25, line 337: This advertises the dedup as a feature, but it's the source of the KV-correctness bug flagged on `engine.py:224` — skipping already-published blocks is what produces false root-anchored blocks and forged adjacency. If the dedup goes away, this line should go with it.

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+    W1 -. "KV events<br/>(block hashes)" .-> F
+```
+
+The engine publishes a KV event for each prompt's blocks; the router's
+radix tree learns which worker "holds" which prefix, so a repeat prompt
+routes back to the same worker (`0/13 → 12/13 blocks overlap`) while
+the other worker stays idle. With no cache knowledge, requests would
+round-robin instead.
+
+## Container
```
tanmayv25, line 236: **The ladder's second rung is missing: there's no way to run this without Docker.**

`basic/` runs in two `python` commands. Here, the first executable instruction is `docker build` — I grepped, and this README has no `pip install`, no `uv pip`, and no bare `python3 -m hello_engine.main` outside a `docker run`. A reader who just ran stage 1 locally is now asked to build a container image, which also undercuts the "no GPU, no downloads, nothing to install" pitch in the ladder README.

Everything needed already exists — the engine defaults to `--discovery-backend etcd` but accepts `file`, and the frontend does too. So this is roughly:

```bash
uv pip install -e .
python3 -m hello_engine.main --discovery-backend file --event-plane zmq &
python3 -m dynamo.frontend --http-port 8000 --discovery-backend file --router-mode kv
```

Suggest making that the **first** runnable section, before `## Container`.

Relatedly: prerequisites are never stated as such anywhere. `examples/README.md:73` tells arrivals "Most examples require etcd and NATS" and hands them a docker-compose command. This README should say plainly, once, near the top: no etcd, no NATS, no GPU, no network.

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+  ≈ the full ASCII reply).
+- **Swap it**: `--tokenizer-repo Qwen/Qwen3-0.6B` (or any HF repo) uses a
+  real tokenizer instead.
+
+The two sidecar files are the minimum paperwork Dynamo's model card
+requires: `tokenizer_config.json` carries the chat template that renders
+`messages` into a prompt; `config.json` provides `model_type` /
+`eos_token_id` / context-length fields.
+
+## Walkthrough: how this engine works
```
tanmayv25: **Ordering: the walkthrough is the reason this example exists, and it's 135 lines down.**

Current heading order is *motivation → Container → Docker → Kubernetes → Verify → tokenizer → Walkthrough*. Compare `basic/README.md`: *Architecture → Components → Prerequisites → Running → Expected Output → Code Structure → Deploy → Next step*.

Stage 1 is "learn → run → read → deploy". Stage 2 is "deploy → read". For a two-stage ladder the rungs should feel like the same object twice; right now they read as two documents by different authors.

Suggest moving `## Walkthrough` and `## The bundled mock tokenizer` above `## Container`. Cheap change, high payoff.

(The mermaid diagram at `:42-55` earns its place, by the way — showing the KV-event back-edge as a dotted line is the one thing prose handles badly.)

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+prompt always produces the same block identities — identical prompts
+genuinely match in the router's tree instead of being random noise.
+
+Not implemented here (deliberately): the sibling **KV metrics** channel
+(`ComponentSnapshot` gauges — "how full is my cache"). See
+`component_metrics_dp_ranks()` / `attach_snapshot_publisher()` in
+[`sample_engine.py`](https://github.com/ai-dynamo/dynamo/blob/main/components/src/dynamo/common/backend/sample_engine.py)
+for that pattern.
+
+## Layout
```
tanmayv25, line 372: The Layout section describes what each file *is*, but nothing says what a reader must **change**. Since the PR sells `engine/` as "the copyable skeleton for custom-engine integrators", that's the whole stated value proposition and it's currently undocumented.

A short "what to change when you copy this" table would cover it — roughly ten lines:

- `pyproject.toml`: name, `[tool.hatch...] packages`, `[project.scripts]`
- `Dockerfile`: base image tag
- `from_args`: `--component`, `--served-model-name`, `--endpoint` defaults
- `model_name=args.tokenizer_repo` — repoint at your real model, or the frontend tokenizes with the mock
- `LlmRegistration` values, especially **`kv_cache_block_size`, which must match your engine's real block size or KV events are silently dropped** (`lib/kv-router/src/zmq_wire/convert.rs:427-441` breaks out of the loop with a warning on mismatch)
- `deploy/dgd.yaml`: both image references and `metadata.name`

That `kv_cache_block_size` one in particular is a silent failure a copier will not diagnose on their own.

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,325 @@
...
+    F -- "route by prefix overlap" --> W1
+    F -.-> W2
+    W1 -- "hardcoded token stream" --> F
+    F -- "text" --> C
+    W1 -. "KV events<br/>(block hashes)" .-> F
+```
+
+The engine publishes a KV event for each prompt's blocks; the router's
+radix tree learns which worker "holds" which prefix, so a repeat prompt
+routes back to the same worker (`0/13 → 12/13 blocks overlap`) while
```
tanmayv25: These numbers contradict the rest of the document: `0/13 → 12/13` here, but `0/4 → 3/4` at `:35-37`, `:166-167` and `:301-303`. Pick one prompt and be consistent — a reader who notices will assume they're misreading.

Two more in the same vein: `:28` says "the engine is ~200 lines" (it's 243) and `:321` calls `main.py` a "3-line entry point" (it's 14). Small, but this is a document whose whole value is precision.

examples/custom_backend/hello_world/engine/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Hello-world Dynamo custom engine on the published Dynamo base image.
+# The base carries the Dynamo runtime + ai-dynamo Python package; we
+# layer the engine wheel on top. No GPU, no model weights.
+ARG BASE_IMAGE=nvcr.io/nvidia/ai-dynamo/dynamo-frontend
+ARG BASE_IMAGE_TAG=1.3.0
```
tanmayv25: This pins two minor versions behind the tree — the worktree is at `1.5.0` (root `pyproject.toml:6`, `Cargo.toml:45`). I checked, and `v1.3.0` does carry everything this example uses (`dynamo.common.backend.run`, `kv_event_sources`, the `WorkerConfig` fields), so the build works today — but an example shipped in the 1.5.0 tree pinned to a 1.3.0 base will drift on every release.

There's precedent both ways (`examples/diffusers/Dockerfile:27` pins `1.3.0.post1`; `recipes/qwen3-0.6b/vllm/agg/gaie/deploy.yaml:18` tracks `1.5.0`), but the dominant example convention is a `my-tag` placeholder plus a note. Same applies to `ai-dynamo>=1.3.0` in `pyproject.toml:14`.

examples/custom_backend/hello_world/engine/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "hello-dynamo-engine"
+version = "0.1.0"
+description = "Hello-world Dynamo unified backend: hardcoded token stream + synthetic KV events, no GPU"
+requires-python = ">=3.11"
```
tanmayv25: This is stricter than the repo's own floor (root `pyproject.toml:14` is `>=3.10`, with an explicit `typing_extensions` back-port at `:24-26` specifically to keep 3.10 working) with no stated reason — and it's unenforced in the image anyway because of the `--no-deps` install. It also contradicts `examples/README.md:85` ("Python 3.9+"), a line this PR edits.

Suggest `>=3.10` unless something here genuinely needs 3.11.

Separately, `[project.scripts] hello-engine` on `:19` is dead — the Dockerfile ENTRYPOINT, the DGD, and the README all use `python3 -m hello_engine.main`. Either use it in the ENTRYPOINT or drop it.

examples/custom_backend/hello_world/basic/README.md
```diff
@@ -0,0 +1,120 @@
...
+### Client (`client.py`)
+
+- **`worker`**: A dynamo worker that connects to the backend service and processes the streaming response
+
+## Deployment to Kubernetes
+
+Note that this a very simple degenerate example which does not demonstrate the standard Dynamo FrontEnd-Backend deployment. The hello-world client is not a web server, it is a one-off function which sends the predefined text "world,sun,moon,star" to the backend. The example is meant to show the HelloWorldWorker. As such you will only see the HelloWorldWorker pod in deployment. The client will run and exit and the pod will not be operational.
+
+
+Follow the [Quickstart Guide](../../../docs/fern/pages/kubernetes/getting-started/quickstart.mdx) to install Dynamo Kubernetes Platform.
```
tanmayv25: **Broken link** — the file moved down a level but the link wasn't re-based. `../../../` from `basic/` lands at `examples/`, so this resolves to `examples/docs/fern/...`. Confirmed by the repo's own linter:

```
$ python3 docs/fern/scripts/docs_lint.py --scan examples
ERROR LINK  examples/custom_backend/hello_world/basic/README.md:101  broken relative link
```

Note the fix is **not** adding a fourth `../`: `docs/fern/AGENTS.md:24` requires absolute `github.com/ai-dynamo/dynamo` URLs for targets outside `docs/`.

marckarp:
Thanks for the deep review @tanmayv25 — all of it landed. Blockers:

1. Test path fixed (637b8abc).
2. KV publishing rewritten (29ba8315): block hashes are now chained on their parent, and publishes carry `parent_hash` so tails anchor under the shared prefix — the forged-adjacency dedup is structurally impossible now. Validated end to end with two workers and the KV router: fresh prompt 0/11 overlap, exact repeat 10/11 pinned with nothing republished, and the case that was broken before — shared prefix + new tail — now scores 9/12 and publishes only the 2 new blocks under the correct parent. Added 5 unit tests covering the chaining contract (including the cross-position collision you caught), passing inside the engine image.

Structural + smaller items (fe3f5444): Run-without-Docker section (tested verbatim in a clean python:3.12-slim container; noted Linux-only wheels), walkthrough and sample logs updated to the chained mechanism, canonical-split note added (guide = contract, sample_engine.py = reference, engine/ = walkthrough), model_name/tokenizer-resolution comment expanded, dead script entry + redundant tokenizers install + README COPY dropped (image rebuilt and worker boot verified), backendFramework removed from basic's DGD (server dry-run validated against the operator CRD), examples/README naming + Python 3.11, absolute URL in the ladder README.

Two judgment calls, open to changing: kept the section banners since they carry the timing hints you flagged as useful (the model_name inversion is fixed by the expanded comment), and skipped the codeowners entry since the coverage gate passes — happy to add it if you'd like the dual-owner pattern.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
marckarp: Fixed across 29ba8315 + b1ddba33, landing on exactly your suggested shape: no dedup, no `_published_blocks` — the engine publishes the prompt's full block run from position 0 on every request with monotonic counter IDs (sample_engine's pattern), and re-stores are idempotent in the tree. Both defects are structurally gone: nothing is ever root-anchored mid-sequence, and there's no skip logic to forge adjacency. The unbounded-set side issue is gone with the set itself. Validated e2e with two workers: 0/11 cold → 10/11 exact repeat → 9/12 shared prefix + new tail, all pinned to the publishing worker; plus unit tests on the publish contract (full run, position 0, fresh IDs per publish, partial final block excluded).

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
marckarp: You're right, and this reshaped the fix: since matching keys on the recomputed `tokens_hash` and `block_hashes` are only node identity, the blake2b content hashing bought nothing — b1ddba33 deletes it entirely in favor of a monotonic counter, per sample_engine. The new `_publish_prompt_blocks` docstring now leads with exactly this distinction ("matching is on token content / block_hashes are node identities"), and the README's design-note paragraph was rewritten the same way — it now calls out "matching keys on the tokens, not on our hashes" as the thing newcomers most often misread.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
marckarp: Fixed in b1ddba33 — `_publish_prompt_blocks` now takes a local reference to the publisher and uses that throughout, so a concurrent `cleanup()` can't null it mid-publish.

examples/custom_backend/hello_world/engine/README.md
marckarp: Fixed in d12b40f0: `drain()` removed from the table — the optional hooks are now listed as `abort()` and `is_quiescent()` (verified against the ABC), with a note that draining itself is the Worker's job and the engine only answers `is_quiescent()`. Heading now says "four required methods," which matches the four `yes` rows.

examples/custom_backend/hello_world/engine/README.md
marckarp: The dedup is gone entirely as of b1ddba33 (full-run republish with counter IDs, per your suggestion on the publish path), and this line went with it — the section now describes the republish-every-request behavior and points at delta-publishing + `publish_removed` as what a real engine does instead.

examples/custom_backend/hello_world/engine/README.md
marckarp: Added in fe3f5444: a "Run without Docker" section right after the intro, same shape as basic/ (install, then two terminals — worker and frontend — then curl). Executed verbatim in a clean python:3.12-slim container to make sure it works as written. One honest constraint documented: the ai-dynamo wheels are Linux-only, so macOS readers are pointed at the Container path.

examples/custom_backend/hello_world/engine/README.md
marckarp: Reordered in ae7c74c1 to match basic/'s shape: learn (intro + flow diagram) → run (the new no-Docker section) → read (the walkthrough, now at line ~101 instead of ~237) → deploy (container, DGD, verify), with the tokenizer and layout as appendices. Stage 1 and stage 2 now read the same way.

examples/custom_backend/hello_world/engine/README.md
marckarp: Added in 5097da4d: a "What to change when you copy this" table closing the README — package name, base image tag, the from_args defaults, `model_name` (with the mock-vocab consequence), `LlmRegistration` with your `kv_cache_block_size` warning called out in bold as the silent trap, the generate() chunk contract, and the dgd.yaml image/name fields. (Your `[project.scripts]` row became moot — that entry was dead and is deleted per the other thread.)

examples/custom_backend/hello_world/engine/README.md
marckarp: Cleaned up in 15ea37a4 plus the earlier rewrites: all overlap examples now come from one real run (the 0/11 → 10/11 → 9/12 sequence, captured from actual logs during e2e validation), the stray 0/4 in the verify section is now generic, "~200 lines" is "~250" (current count 255 after the review changes), and the "3-line entry point" claim is gone (main.py is 14 lines with its comments).

examples/custom_backend/hello_world/engine/Dockerfile
marckarp: Bumped to 1.4.0 in 3bc96c12 (rebuilt and boot-verified) — that's the newest published dynamo-frontend tag; the tree's 1.5.0 doesn't exist on NGC until the 1.5.0 release ships, so an example has to trail by one at release-cut time by construction. The copy-checklist table now also tells copiers to pin to the release they build against. If there's a repo convention for auto-bumping example pins at release time, happy to wire into it — otherwise this stays a manual bump each release, as with the precedent you found.

examples/custom_backend/hello_world/engine/pyproject.toml
marckarp: Lowered to >=3.10 in 2d7689a0, matching the repo floor — the engine uses nothing 3.11-specific (compile-verified under python:3.10). The examples/README prerequisite line and the Run-without-Docker section now say 3.10+ consistently.

examples/custom_backend/hello_world/basic/README.md
marckarp: Fixed in 9f79b909 with the absolute GitHub URL (per the docs convention for out-of-docs targets, and immune to future moves). Verified with `docs_lint.py --scan examples` — no findings remain under hello_world. Good catch on the scan flag; the default CI lint only covers docs/, which is how this slipped through.

tanmayv25 (approved):

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,253 @@
...
+            request_plane=args.request_plane,
+            event_plane=args.event_plane,
+        )
+        return engine, worker_config
+
+    # ------------------------------------------------------------------
+    # start: load "the model"   [called once at boot]
+    # ------------------------------------------------------------------
+    async def start(self, worker_id: int) -> EngineConfig:
+        del worker_id
```
alec-flowers, line 123: Why delete this?

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,253 @@
...
+        from tokenizers import Tokenizer
```
alec-flowers, line 124: Can you move import to top of file?

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,253 @@
...
+                Tokenizer.from_pretrained, self.tokenizer_repo
+            )
+        self._reply_token_ids = tokenizer.encode(
+            HARDCODED_REPLY, add_special_tokens=False
+        ).ids
+        logger.info("HelloEngine ready: reply is %d tokens", len(self._reply_token_ids))
+
+        return EngineConfig(
+            model=self.tokenizer_repo,
+            served_model_name=None,  # Worker uses WorkerConfig.served_model_name
```
alec-flowers, line 140: hmmm if this is taken care of why do we have served_model_name avaialble? Will it override it if selected @tanmayv25 ?

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
```diff
@@ -0,0 +1,253 @@
...
+    # ------------------------------------------------------------------
+    # KV events: opt in, receive the publisher, publish per prompt
+    # ------------------------------------------------------------------
+    async def kv_event_sources(self) -> list[KvEventSource]:
+        # [called once after start()] "Build me a publisher and hand it
+        # to the callback below."
+        return [PushSource(on_ready=self._on_publisher_ready, dp_rank=0)]
+
+    def _on_publisher_ready(self, publisher) -> None:
+        # The framework hands us a live KvEventPublisher (wired to ZMQ
```
alec-flowers, line 215: Do we need this extra commentary about zmq/nats?

examples/custom_backend/hello_world/engine/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
alec-flowers, line 1: Does this require a dockerfile?

alec-flowers: I see we have k8s and DGD support

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
tanmayv25: It is to show that this example does not consume `worker_id`.  Certain features required a unique id to operate.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
tanmayv25: @marckarp  let's clean this verbosity.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
tanmayv25: Both fields exist because they're set at different points in the lifecycle. `WorkerConfig` is built in `from_args()`, so it carries operator/CLI intent known before the engine starts. `EngineConfig` comes back from `start()`, so an engine that only learns its name after loading can still advertise one without a flag. Every real backend sets it on `WorkerConfig` via `from_runtime_config`; only the sample engines set it on `EngineConfig`.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
marckarp: Done in 4698bd35 — moved to module level.

examples/custom_backend/hello_world/engine/src/hello_engine/engine.py
marckarp: Trimmed in 4698bd35 — the callback comment is one line now.

tanmayv25:
/ok to test [3e94a60](https://github.com/ai-dynamo/dynamo/pull/12137/commits/3e94a60f06694a89ec7569eacc97ccf0795f8813)

examples/custom_backend/hello_world/engine/README.md
```diff
@@ -0,0 +1,392 @@
...
+for p in $(kubectl get pods -n <namespace> -o name | grep hello-engine-worker); do
+  echo "$p: $(kubectl logs -n <namespace> $p | grep -c 'request received') requests, \
+$(kubectl logs -n <namespace> $p | grep -c 'published') KV publishes"
+done
+```
+
+Expected: **all requests on one worker, zero on the other** — the router
+learned which worker holds the prefix from the engine's KV events and
+pinned every repeat prompt to it. One `published` line despite several
+requests is the engine's dedup (identical blocks are published once).
```
dmitry-tokarev-nv, line 337: **[P2] The README promises a dedup the engine does not have.** `engine/README.md:337` tells the reader that one `published` line for several identical requests is "the engine's dedup". `_publish_prompt_blocks` publishes on every call, and line 227 of the same file says so. Please drop the dedup sentence and state the real count.

<details>
<summary>Measured: three identical prompts, three publish lines</summary>

Three copies of the long prompt from the verify section, single worker, image built from this branch:

```
requests: 5
published-lines: 4
... published 1 KV block(s) for prompt      <- the earlier short "hi" request
... published 12 KV block(s) for prompt
... published 12 KV block(s) for prompt
... published 12 KV block(s) for prompt
```

Router scoring over the same three requests confirms the pinning claim is correct:

```
with 0/13 blocks overlap
with 12/13 blocks overlap
with 12/13 blocks overlap
```

So the "all requests on one worker" half of the Expected block holds. Only the publish count is wrong. Commit b1ddba33 removed the content hashing that would have made a dedup possible, and this sentence survived it.
</details>

dmitry-tokarev-nv (approved):
## Verified review

Both stages run as written. I built the engine image from the example `Dockerfile` on the real `nvcr.io/nvidia/ai-dynamo/dynamo-frontend:1.4.0` base, ran the README Docker recipe verbatim, and got a served OpenAI response plus real KV-aware routing. One P2 on a README line that describes output the code does not produce.

<details>
<summary>What I ran, and what it returned</summary>

Merge base `165d854ba6` (2026-09-02), head `3e94a60f06`. History is linear and additive after one rebase, so the diff is the author's own change.

Stage 2, `engine/`, AMD64 Linux, Docker, no GPU:

- `docker build -t hello-dynamo-engine:0.1.0 .` succeeded on the 1.4.0 base.
- The two `docker run` commands and the `curl` from the README, copied verbatim, returned `{"id":"chatcmpl-...","finish_reason":"stop","usage":{"prompt_tokens":24,"completion_tokens":208,"total_tokens":232}}` with the hardcoded sentence as content. `/v1/models` listed `hello-engine`.
- KV routing is real. Three copies of the long prompt from the verify section gave `0/13 blocks overlap`, then `12/13`, then `12/13`.

Stage 1, `basic/`: started `hello_world.py` and `client.py` in the same image with `DYN_DISCOVERY_BACKEND=file DYN_EVENT_PLANE=zmq`. The backend logged `[backend] Received request: world,sun,moon,star` and the request completed.

Interface check: `from_args`, `start(worker_id)`, `generate(request, context)`, `cleanup`, `kv_event_sources` all match the current `BaseEngine` and `LLMEngine` in `components/src/dynamo/common/backend/engine.py`. `PushSource(on_ready=..., dp_rank=...)` and `publish_stored(token_ids, num_block_tokens, block_hashes, parent_hash)` match `publisher.py` and `_core.pyi`. No swallowed errors: `generate` raises on a missing `start`, and the null publisher guard is the only early return.

Tests: the six cases in `lib/bindings/python/tests/test_example_hello_engine.py` pass, and they pin what they claim. I mutated the engine three ways and each mutation broke the expected cases, with the unmutated control green.

| mutation | result |
|---|---|
| publish the partial final block too | 2 failed, 4 passed |
| stop advancing the block id counter | 2 failed, 4 passed |
| remove the null publisher guard | 1 failed, 5 passed |
| control, no mutation | 6 passed |

Orphans and references: no path in the repository still points at the old flat layout. The other `hello_world` hits are the Rust runtime example under `lib/`, which this change does not touch. Both new paths have a CODEOWNERS owner (`who_owns.py` returns `dynamo-runtime-codeowners`). No link-checker ignore entry was added.

Style and gates: `pre-commit run --files <21 changed files>` passed every hook that selected a file, including `codespell`, `check yaml`, `ruff`, `black`, and the Fern asset path check. The marker report hook failed only on local port allocation in `tests/serve/conftest.py`, and its own summary printed `Missing sets: 0`.

What I did not verify: the Kubernetes path. I have no cluster, so `deploy/dgd.yaml`, the operator environment injection, and the "workers go Ready in 30 to 60s" claim are unverified.
</details>

PR #12138, opened 2026-07-24
fix(router): clear a worker's overloaded mark on every healthy load sample

yifjiang:
> **Draft** — opening for discussion/validation. The fix is a small, targeted change; CI has not yet been run and I'd welcome maintainer input on the approach before finalizing.

## Overview

A worker that is ever marked overloaded via the request path (`Client::mark_overloaded_immediate`) can be **excluded from routing permanently**, causing the router to return `ResourceExhausted` / "All workers are busy, please retry later" **forever — even when the worker is idle and healthy.** Only a frontend restart recovers.

## Root cause

Two independent code paths add a worker to the router's overloaded set:

1. **Monitor path** — `KvWorkerMonitor` computes an overloaded set from load thresholds and calls `publish_overloaded_instances(...)` to push it to the routing `Client`.
2. **Request fast-path** — on a `ResourceExhausted` / "worker at capacity" response, `push_router` calls `Client::mark_overloaded_immediate(instance_id)`, which inserts the worker into `RoutingInstances.overloaded_ids` **directly, outside** the monitor's `overloaded_tracker`.

The fast-path mark (2) is only cleared when the monitor republishes the full set (1) via `set_overloaded_instances`. But in `worker_monitor.rs` that publish is gated on `overloaded_changed`:

```rust
if overloaded_changed {
    let overloaded_instances = overloaded_tracker.ids();
    publish_overloaded_instances(&client, &prefill_client_holder, &overloaded_instances);
}
```

When load thresholds are **unconfigured (the default)**, `overloaded_tracker` never changes, so `overloaded_changed` is never `true`, so the monitor **never republishes**. The immediate mark set in (2) is therefore never overwritten — the worker stays in `overloaded_ids` indefinitely, `free_ids = routable − overloaded` stays empty, and every subsequent request hits the "All workers are busy" branch. The worker itself is idle/healthy the whole time (its per-request permits are RAII and released).

## Fix

Publish the overloaded set on **every fresh load sample** rather than only when the threshold-derived set changes, so the healthy set overwrites the stale immediate mark once the transient overload passes. `overloaded_changed` is retained only to gate a debug log. One file, `lib/llm/src/discovery/worker_monitor.rs`.

## Reproduction

Single worker, `DYN_ENGINE_REQUEST_LIMIT` small (e.g. 8) + a small overflow queue, model slow enough that the 8 engine slots stay occupied under load. Drive it closed-loop at concurrency well above `limit + queue` for ~25 s so the engine + queue co-saturate and the worker emits "worker at capacity" at least once. After the burst drains, every request returns a permanent `500` / `ResourceExhausted "All workers are busy"` while the worker sits idle (`sglang:num_running_reqs=0`). With this change, the endpoint recovers on the next load sample.

(Observed live on a GB200 deployment; the failure and the fix were both confirmed against the exact code in `worker_monitor.rs`.)

## Testing

- [ ] `cargo test -p dynamo-llm` (not yet run in this draft)
- [ ] add a regression test asserting the overloaded set is republished on an unchanged healthy sample after an immediate mark

Happy to add the regression test and adjust the approach per review.

### DIFF
--- a/lib/llm/src/discovery/worker_monitor.rs
+++ b/lib/llm/src/discovery/worker_monitor.rs
@@ -992,9 +992,7 @@ impl WorkerLoadMonitor for KvWorkerMonitor {
                         }
 
                         // Recompute the full overloaded set only when thresholds change;
-                        // otherwise incrementally update just this worker. When the set
-                        // changes, publish to both the decode Client and (in disaggregated
-                        // serving) the prefill Client — see `publish_overloaded_instances`.
+                        // otherwise incrementally update just this worker.
                         let overloaded_changed = if thresholds_changed {
                             last_thresholds = cfg.clone();
                             let overloaded_workers =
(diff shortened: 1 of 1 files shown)


PeaBrane:
/ok to test [9a34dc3](https://github.com/ai-dynamo/dynamo/pull/12138/commits/9a34dc31061caa15fbafd84676250e4e4570fe77)

TaekyungHeo:
Additional exact reproduction from Aria ComputeLab SC qualification:

- Dynamo SGLang 1.1 same-node disaggregated: jobs `4021214`, `4021255`, `4021306`
- Dynamo SGLang 1.2 same-node disaggregated: jobs `4021546`, `4021608`, `4021927`, `4021954`
- Two B200 GPUs, prefill/decode/frontend all become ready, NIXL UCX initializes, and the frontend discovers both model roles.
- The frontend then logs `KvWorkerMonitor: prefill endpoint watcher activated, tracking 0 workers`; the first chat request returns 503 `service overloaded or rate limited` while both workers are healthy.
- Starting the frontend before workers, using direct routing, varying the readiness-to-request delay, and forcing a decode-block threshold did not recover the free set.
- Control rows on the same internal-only provider fabric pass: Dynamo vLLM disaggregated `4021028`; Dynamo TensorRT-LLM 1.2 disaggregated `4021982`.

This is consistent with the PR's root cause: the initial/unchanged healthy empty overloaded set is not republished, so the routing client's free set remains empty. These are qualification-only canaries, not performance claims.

rmccorm4:
I believe https://github.com/ai-dynamo/dynamo/pull/14210 will supersede this PR

rmccorm4:
Closing this as superseded by https://github.com/ai-dynamo/dynamo/pull/14210. @yifjiang feel free to open a new PR if needed.

PR #12135, opened 2026-07-24
test(vllm): cover classify and pooling endpoints

jh-nv:
## Summary

- Add shared classify/pooling payload builders and aggregate vLLM end-to-end coverage.
- Cover text and pre-tokenized inputs, batched tokens, truncation side, all encoding modes, classify parity, and token_classify.
- Stack layer extracted from #12058.

## Validation

- All 10 final live endpoint payload cases passed against cross-encoder/nli-MiniLM2-L6-H768.
- 104 targeted Python tests passed across the stack.
- 40 targeted Rust classify/pooling protocol tests passed.
- Pre-commit hooks passed for every changed file.

## Stack

- **14 of 14**
- Previous: [#12134](https://github.com/ai-dynamo/dynamo/pull/12134)
- Next: final layer
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `jihao/vllm-pooling-example`
- Head: `jihao/vllm-pooling-e2e`

PR #12127, opened 2026-07-24
feat(http): serve classify and pooling requests

jh-nv:
## Summary

- Add the classify and pooling HTTP handlers and service-v2 route registration.
- Support vLLM-compatible float, base64, bytes, and bytes_only pooling responses, including binary metadata.
- Backend worker activation remains out of scope for this layer.
- Stack layer extracted from #12058.

## Validation

- Targeted dynamo-llm classify and pooling tests passed.
- The final stack's HTTP behavior was verified against a live local vLLM Dynamo service.
- Pre-commit hooks passed for the changed files.

## Stack

- **6 of 14**
- Previous: [#12126](https://github.com/ai-dynamo/dynamo/pull/12126)
- Next: [#12128](https://github.com/ai-dynamo/dynamo/pull/12128)
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `jihao/vllm-pooling-metrics`
- Head: `jihao/vllm-pooling-http`

PR #12125, opened 2026-07-24
feat(http): define classify and pooling endpoints

jh-nv:
## Summary

- Define classify and pooling endpoint kinds and their service-state flags.
- Handle the new variants exhaustively while leaving route registration for the next HTTP layer.
- Stack layer extracted from #12058.

## Validation

- Targeted dynamo-llm tests passed.
- Pre-commit hooks passed for the changed files.

## Stack

- **4 of 14**
- Previous: [#12124](https://github.com/ai-dynamo/dynamo/pull/12124)
- Next: [#12126](https://github.com/ai-dynamo/dynamo/pull/12126)
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `jihao/vllm-pooling-discovery`
- Head: `jihao/vllm-pooling-endpoint-types`

PR #12122, opened 2026-07-24
feat(llm): add classify protocol types

jh-nv:
## Summary

- Add OpenAI-compatible classify request, response, streaming, and aggregation protocol types.
- Keep the types dormant: this layer does not register or expose a new HTTP endpoint.
- First PR in the stack extracted from #12058.

## Validation

- 20 targeted Rust classify protocol tests passed.
- Pre-commit hooks passed for the changed files.
- The complete stacked tree was exercised against a local vLLM Dynamo service.

## Stack

- **1 of 14**
- Previous: `main`
- Next: [#12123](https://github.com/ai-dynamo/dynamo/pull/12123)
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `main`
- Head: `jihao/vllm-classify-protocol`

PR #12139, opened 2026-07-24
feat(http): serve classify and pooling requests

jh-nv:
## Summary

- Add classify and pooling endpoint state, request handling, and metrics.
- Support vLLM-compatible `float`, `base64`, `bytes`, and `bytes_only` pooling responses, including binary metadata.
- Leave built-in vLLM backend activation to the next PR, keeping this layer scoped to frontend/HTTP integration.
- This is PR 2 of 4 and consolidates drafts #12125–#12127.

## Validation

- This commit's pre-commit hooks passed.
- On the complete four-PR stack: 20 classify and 20 pooling Rust tests passed.
- Live endpoint validation passed for JSON and binary response formats.
- The final tree exactly matches a clean merge of the former 14-PR stack onto current `main`.

## Stack

Merge in this order:

1. [#12140 — core protocols and model discovery](https://github.com/ai-dynamo/dynamo/pull/12140)
2. [#12139 — frontend HTTP integration](https://github.com/ai-dynamo/dynamo/pull/12139) **(this PR)**
3. [#12142 — vLLM worker integration and activation](https://github.com/ai-dynamo/dynamo/pull/12142)
4. [#12141 — examples and end-to-end validation](https://github.com/ai-dynamo/dynamo/pull/12141)

The original monolithic #12058 remains open; this four-PR stack replaces only the earlier 14-draft split (#12122–#12135).

## Related Issues

- Relates to DIS-2528

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible classification and pooling endpoints.
  * Classification requests now support cache-salt validation and return classification results.
  * Pooling requests support JSON or binary responses, optional metadata, multiple data types, and tensor validation.
  * Added endpoint configuration, routing, metrics, and model support for classification and pooling.
* **Tests**
  * Added coverage for binary pooling responses and invalid empty cache salts.

### DIFF
--- a/lib/llm/src/http/service/openai.rs
+++ b/lib/llm/src/http/service/openai.rs
@@ -58,10 +58,15 @@ use crate::protocols::openai::{
         NvCreateChatCompletionRequest, NvCreateChatCompletionResponse,
         NvCreateChatCompletionStreamResponse,
     },
+    classify::{NvCreateClassifyRequest, NvCreateClassifyResponse},
     completions::{NvCreateCompletionRequest, NvCreateCompletionResponse},
     delta_common,
     embeddings::{NvCreateEmbeddingRequest, NvCreateEmbeddingResponse},
     images::{NvCreateImageRequest, NvImagesResponse},
+    pooling::{
+        NvCreatePoolingRequest, NvCreatePoolingResponse, PoolingEmbedDType, PoolingEncodingFormat,
+        PoolingEndianness, PoolingOutput,
+    },
     responses::{NvCreateResponse, NvResponse, ResponseParams, chat_completion_to_response},
     videos::{NvCreateVideoRequest, NvVideosResponse},
 };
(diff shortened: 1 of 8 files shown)


lib/llm/src/http/service/openai.rs
```diff
@@ -1337,6 +1342,385 @@ fn decode_base64_embedding_to_floats(s: &str) -> Result<Vec<f32>, anyhow::Error>
...
+            let err_response =
+                ErrorMessage::from_anyhow(anyhow::Error::new(e), "Failed to fold pooling stream");
+            inflight.mark_error(extract_error_type_from_response(&err_response));
+            err_response
+        })?;
+
+    let response = match response_encoding {
+        PoolingEncodingFormat::Float | PoolingEncodingFormat::Base64 => {
+            Json(response).into_response()
+        }
```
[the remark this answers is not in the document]
jh-nv: Verified against the complete stack. The vLLM worker sets `shape` only for `encoding_format == "bytes"`; its parametrized output-format test asserts that `shape` is absent for `float`, `base64`, and `bytes_only`. Therefore the supported worker cannot leak this field in Float/Base64 responses. Frontend sanitization could be separate defensive hardening for an out-of-contract custom worker, but there is no current behavior to fix here.

lib/llm/src/http/service/openai.rs
```diff
@@ -1337,6 +1342,385 @@ fn decode_base64_embedding_to_floats(s: &str) -> Result<Vec<f32>, anyhow::Error>
...
+
+    // vLLM currently rejects dimensionality reduction on `/pooling`.
+    if request.dimensions.is_some() {
+        return Err(pooling_or_classify_bad_request(
+            "dimensions is currently not supported".to_string(),
+        ));
+    }
+    let response_encoding = request.encoding_format;
+    let response_dtype = request.embed_dtype.unwrap_or_default();
+    let response_endianness = request.endianness.unwrap_or_default();
```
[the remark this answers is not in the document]
jh-nv: Verified against both sides of the internal wire contract. Rust defaults omitted values to `float32` / `native`, and the worker independently defaults to the same two values. Encoded-length validation depends on dtype width only; endianness does not change the expected byte count. The binary-format tests also validate returned dtype/endianness metadata, so there is no default mismatch in the supported stack.

rmccorm4 (commented):
Two inline findings from local review.

lib/llm/src/model_type.rs
```diff
@@ -227,6 +227,12 @@ impl ModelType {
         if self.contains(Self::Realtime) {
             endpoint_types.push(crate::endpoint_type::EndpointType::Realtime);
         }
+        if self.contains(Self::Classify) {
```
rmccorm4, line 230: *From local review:*

[P1] Removing one model disables this endpoint for every surviving model of the same type. These new mappings feed process-wide `AtomicBool` flags in `ServiceV2`, while `ModelUpdate::Removed` unconditionally clears every endpoint mapped from the removed card. With pooling models A and B registered, removing A therefore makes B's `/pooling` route return 404 until another pooling add event; classify is affected identically. Please derive availability from the remaining model cards or maintain per-endpoint membership/reference counts, and cover add-A/add-B/remove-A.

lib/llm/src/http/service/openai.rs
```diff
@@ -1337,6 +1342,385 @@ fn decode_base64_embedding_to_floats(s: &str) -> Result<Vec<f32>, anyhow::Error>
...
+) -> Result<Response, ErrorResponse> {
+    // return a 503 if the service or model is not ready
+    check_ready(&state)?;
+    check_model_serving_ready(&state, &request.model)?;
+
+    if !state.nvext_enabled() {
+        warn_nvext_disabled("classify", request.nvext.is_some(), &headers);
+        request.nvext = None;
+    }
+    validate_pooling_cache_salt(request.cache_salt.as_deref())?;
```
rmccorm4: *From local review:*

[P2] This validation returns before `create_inflight_guard`, so an invalid `cache_salt` 400 is absent from the request and validation-error metrics. Pooling has the same gap for `cache_salt` and `dimensions` at lines 1617-1623. Please create and mark the guard before local validation, or route these checks through a common metered validation path, and add route-level metric assertions.

lib/llm/src/model_type.rs
jh-nv: I don't think this is reachable. The watcher only emits a `Removed` card for a model type once no model of that type remains, and narrows the card to that one type — `removed_model_cards` (`watcher.rs:312`) gates on `is_model_type_list_empty`, which has `Classify`/`Pooling` arms and runs after `remove_model()`. That's been on main since #12140.

Checked it end-to-end with two real pooling workers: killed pool-a, and `/v1/pooling` + `/v1/classify` kept serving pool-b (200 across 5 probes, no `disabled` in the frontend log). Then killed pool-b as well and both endpoints went 404 — so the check can detect a gated-off endpoint; it just never gets gated off while another model of that type is still registered.

Added the add-A/add-B/remove-A  tests

lib/llm/src/http/service/openai.rs
jh-nv: Fixed

rmccorm4 (approved):

PR #12141, opened 2026-07-24
test(vllm): validate classify and pooling endpoints

jh-nv:
## Summary

- Add the aggregated vLLM classify launch example.
- Add end-to-end classify/pooling payload coverage, including pre-tokenized truncation and binary formats.
- Complete the four-PR stack without changing runtime behavior beyond the preceding implementation layers.
- This is PR 4 of 4 and consolidates drafts #12134–#12135.

## Validation

- This commit's pre-commit hooks passed.
- `bash -n examples/backends/vllm/launch/agg_classify.sh` passed.
- 228 focused Python tests passed, including pooling payload tests.
- A live local vLLM/Dynamo stack passed classify, pooling, flat/nested token truncation, `base64`, `bytes`, `bytes_only`, and `token_classify` smoke tests.
- The final tree exactly matches a clean merge of the former 14-PR stack onto current `main`.

## Stack

Merge in this order:

1. [#12140 — core protocols and model discovery](https://github.com/ai-dynamo/dynamo/pull/12140)
2. [#12139 — frontend HTTP integration](https://github.com/ai-dynamo/dynamo/pull/12139)
3. [#12142 — vLLM worker integration and activation](https://github.com/ai-dynamo/dynamo/pull/12142)
4. [#12141 — examples and end-to-end validation](https://github.com/ai-dynamo/dynamo/pull/12141) **(this PR)**

The original monolithic #12058 remains open; this four-PR stack replaces only the earlier 14-draft split (#12122–#12135).

## Related Issues

**🔗 This PR is linked to an issue:**

- Closes DIS-2528

## Summary by CodeRabbit

- **New Features**
  - Added an example launch configuration for aggregated vLLM classification and pooling services.
  - Added support for classification and pooling requests using text, token IDs, batches, truncation, and optional request parameters.

- **Bug Fixes**
  - Improved validation of classification and pooling responses, including probabilities, token usage, tensor shapes, binary data, and base64 encoding.

- **Tests**
  - Added comprehensive coverage for JSON, binary, token-wise, and encoded pooling responses, along with classification workflows.

PR #12133, opened 2026-07-24
feat(vllm): advertise the pooling endpoint

jh-nv:
## Summary

- Advertise /v1/pooling from the already-enabled pooling-family worker.
- Keep classify and pooling endpoint publication explicit and covered at the factory boundary.
- Stack layer extracted from #12058.

## Validation

- A live local vLLM Dynamo stack served both /v1/classify and /v1/pooling.
- Live checks passed for float, base64, bytes, bytes_only, pre-tokenized truncation on both sides, batched token IDs, and token_classify.
- Pre-commit hooks passed for the changed files.

## Stack

- **12 of 14**
- Previous: [#12132](https://github.com/ai-dynamo/dynamo/pull/12132)
- Next: [#12134](https://github.com/ai-dynamo/dynamo/pull/12134)
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `jihao/vllm-classify-enable`
- Head: `jihao/vllm-pooling-enable`

PR #12128, opened 2026-07-24
feat(vllm): add pooling-family worker handler

jh-nv:
## Summary

- Add a standalone vLLM worker handler for classify, pooling, score, and token-classification tasks.
- Apply pre-tokenized truncation through vLLM tokenization semantics and support all four pooling encodings.
- Keep the module dormant; worker-factory wiring is introduced later in the stack.
- Stack layer extracted from #12058.

## Validation

- 26 focused pooling-handler unit tests passed once the following test layer is applied.
- Live tests covered string, list[int], list[list[int]], truncation-side, binary encoding, and token_classify requests.
- Pre-commit hooks passed for the changed file.

## Stack

- **7 of 14**
- Previous: [#12127](https://github.com/ai-dynamo/dynamo/pull/12127)
- Next: [#12129](https://github.com/ai-dynamo/dynamo/pull/12129)
- Original unsplit PR: [#12058](https://github.com/ai-dynamo/dynamo/pull/12058)
- Base: `jihao/vllm-pooling-http`
- Head: `jihao/vllm-pooling-worker`

PR #12142, opened 2026-07-24
feat(vllm): enable classify and pooling workers

jh-nv:
## Summary

- Add the standalone vLLM pooling-family worker and its unit coverage.
- Wire worker creation, configuration, registration safeguards, and classify/pooling activation.
- Apply `truncate_prompt_tokens` to pre-tokenized inputs using vLLM-compatible truncation semantics.
- This is PR 3 of 4 and consolidates drafts #12128–#12133.

## Validation

- This commit's pre-commit hooks passed.
- On the complete four-PR stack: 228 focused Python tests passed.
- Live validation passed for `/v1/classify`, `/v1/pooling`, flat/nested token truncation, and `token_classify`.
- The final tree exactly matches a clean merge of the former 14-PR stack onto current `main`.

## Stack

Merge in this order:

1. [#12140 — core protocols and model discovery](https://github.com/ai-dynamo/dynamo/pull/12140)
2. [#12139 — frontend HTTP integration](https://github.com/ai-dynamo/dynamo/pull/12139)
3. [#12142 — vLLM worker integration and activation](https://github.com/ai-dynamo/dynamo/pull/12142) **(this PR)**
4. [#12141 — examples and end-to-end validation](https://github.com/ai-dynamo/dynamo/pull/12141)

The original monolithic #12058 remains open; this four-PR stack replaces only the earlier 14-draft split (#12122–#12135).

## Related Issues

- Relates to DIS-2528

## Summary by CodeRabbit

* **New Features**
  * Added support for classification workers with `/classify` request handling.
  * Added pooling and classification responses with validated inputs, configurable output formats, usage metadata, and batching.
  * Classification workers now support lifecycle management, health checks, cancellation, and scheduler-aware processing.
* **Bug Fixes**
  * Improved worker configuration validation and incompatibility checks.
  * Pooling and classification workers now default prefix caching to disabled when unspecified, while other worker types retain existing behavior.
  * Improved classification and pooling worker registration and routing.


components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,502 @@
...
+                    "prompt": encode_arg,
+                    "pooling_params": pooling_params,
+                    "request_id": request_id,
+                }
+                if priority != 0:
+                    encode_kwargs["priority"] = priority
+                # Token-ID prompts have already been passed through vLLM's
+                # TokenizeParams. Text prompts are tokenized inside AsyncLLM.
+                if tokenization_kwargs is not None and isinstance(prompt, str):
+                    encode_kwargs["tokenization_kwargs"] = tokenization_kwargs
```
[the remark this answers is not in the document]
jh-nv: Verified against vLLM 0.25.1. `TokenizeParams.with_kwargs` explicitly consumes `truncate_prompt_tokens`, `truncation_side`, and `add_special_tokens`; `get_encode_kwargs` forwards the special-token setting and `_token_truncation` applies the requested/default side. These values are not passed blindly to the Hugging Face tokenizer. Source: https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/renderers/params.py#L245-L331. The current implementation matches that API, so no code change is needed.

components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,502 @@
...
+        return await self._run_pooling_batch(prompts, _encode_one)
+
+    async def generate(
+        self, request: dict, context: Context
+    ) -> AsyncIterator[dict[str, Any]]:
+        """Handle one internal classify or pooling wire request."""
+        if "encoding_format" in request:
+            async for response in self._generate_pooling(request, context):
+                yield response
+            return
```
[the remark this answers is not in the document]
jh-nv: This is an intentional internal-wire discriminator, not an optional caller hint. `NvCreatePoolingRequest.encoding_format` is non-optional with a serde default and is always serialized, including `"float"` when the HTTP client omits it; the round-trip test asserts that field is present on the internal wire. The classify request type has no `encoding_format` field, and worker tests cover both key-present (pooling) and key-absent (classify) dispatch. An explicit request-kind field would be broader protocol churn without changing the enforced invariant.

components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,503 @@
...
+        for key, value in raw.items():
+            try:
+                normalized[int(key)] = str(value)
+            except (TypeError, ValueError):
+                continue
+        return normalized
+
+    async def _resolve_pooling_task(self, requested_task: str | None) -> str:
+        if requested_task is not None:
+            return requested_task
```
furionw, line 80: Upstream vLLM `/pooling` does not support switching tasks per request: it pins requests to the configured `pooling_task`, even when the model supports another task. Here, Dynamo appears to forward the requested task and therefore allows per-request task switching when the underlying pooling model supports it. Is this difference intentional? If so, could we document it and add coverage for this behavior?

components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,503 @@
...
+    raise TypeError(
+        f"Unsupported 'input' element type {type(first).__name__}; "
+        "expected str, int, or list[int]"
+    )
+
+
+def _pooling_output_to_list(data: Any) -> list[float]:
+    """Convert sequence-level pooling output to a flat float list."""
+    if isinstance(data, torch.Tensor):
+        return data.detach().cpu().flatten().tolist()
```
furionw: Should we reject non-1-D output here instead of flattening it? `/classify` expects one probability vector with shape `[num_classes]`. For example, if a malformed token-level result for four tokens and three classes reaches this path:

```text
[[0.1, 0.2, 0.7],
 [0.1, 0.8, 0.1],
 [0.2, 0.7, 0.1],
 [0.05, 0.9, 0.05]]
```

Flattening produces 12 values, so the response reports `num_classes=12`; its argmax can be index 10 even though valid class indices are only 0–2, leaving no matching label. Native vLLM rejects classification output when `ndim != 1`: https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/outputs.py#L288-L306.

A correctly matched `task="classify"` model should normally return a 1-D vector, so this is defensive validation to avoid hiding a model or pooler mismatch.

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -584,6 +585,14 @@ async def create(
             )
             return
 
+        # This getattr keeps the factory wiring dormant until the
```
karen-sy: > until the --classify-worker configuration flag is introduced

nit on the comment block, but the `classify-worker` flag **is** introduced by this PR, right? in `backend_args.py`

components/src/dynamo/vllm/backend_args.py
```diff
@@ -706,3 +720,31 @@ def _validate_realtime_worker_exclusivity(self) -> None:
             raise ValueError("--realtime cannot be combined with --benchmark-mode.")
         if getattr(getattr(self, "engine_args", None), "enable_lora", False):
             raise ValueError("--realtime cannot be combined with --enable-lora.")
+
+    def _validate_classify_worker_exclusivity(self) -> None:
```
karen-sy, line 724: `--classify-worker --headless` currently passes validation, but `main.worker()` returns through `run_dynamo_headless(config)` before `WorkerFactory.create(...)`, so `_create_classify_worker()` never runs and the process never registers the classify/pooling endpoint.

So i think we should also reject that combination of flags

components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,503 @@
...
+                "total_tokens": prompt_tokens,
+                "completion_tokens": 0,
+            },
+        }
+
+
+def _build_pooling_tokenize_params(
+    engine_client: Any, tokenization_kwargs: dict[str, Any] | None
+) -> Any:
+    if tokenization_kwargs is None:
```
karen-sy, line 332: Should we keep Dynamo on vLLM’s default tokenization/post-tokenization path even when no explicit tokenization kwargs are set? Upstream vLLM still runs pre-tokenized pooling/classify inputs through `TokenizeParams.apply_post_tokenization()`, which preserves default validation/truncation behavior. Here we only build `TokenizeParams` if the request includes overrides, so token-ID inputs may take a different path from native vLLM.

components/src/dynamo/vllm/backend_args.py
karen-sy: Also, should `--classify-worker` reject `--enable-lora`, or should this handler pass `lora_request` into `engine_client.encode()` like upstream vLLM pooling serving does? Right now the worker accepts the flag but appears to run classify/pooling requests without adapter selection, which could silently use the base model for a request targeting a LoRA adapter.

components/src/dynamo/vllm/pooling_handlers.py
jh-nv: fixed.

components/src/dynamo/vllm/worker_factory.py
jh-nv: Removed — along with the `getattr` default, which isn't needed now that the flag exists.

components/src/dynamo/vllm/pooling_handlers.py
jh-nv: Thanks! I researched on this, and I think vllm does not pin the task. it has default task (which Dynamo resolves to the same default), and it supports request to set its task.


```Verified at three levels on v0.26.0 (our pin), plus the docs:

  ┌──────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │    Level     │                                                   Evidence                                                    │
  ├──────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ API surface  │ PoolingCompletionRequest.task: PoolingTask | None = None → to_pooling_params() forwards it                    │
  ├──────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Execution    │ DispatchPooler.forward does groupby(pooling_metadata.tasks) — per-request, and a single batch can mix tasks   │
  ├──────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Model wiring │ for_seq_cls builds both classify and token_classify poolers; for_embedding builds both embed and token_embed  │
  ├──────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Docs         │ "set the task directly when using the Pooling API", example {"model": ..., "input": prompts, "task": "embed"} │
  └──────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

File: components/src/dynamo/vllm/backend_args.py (hunk 6 above)
Comment (jh-nv) [reply components/src/dynamo/vllm/backend_args.py:724]: > --classify-worker --headless
This is now fixed by rejecting the combination. 


> --classify-worker --enable-lora
This is good catch. Currently, we are rejecting the combination. Want to land this PR first this way. We need a dedicated fix for LoRA support for anything extends EmbeddingWorkerHandler, so it is bigger than the scope of this PR. Will build a follow up for this.

File: components/src/dynamo/vllm/pooling_handlers.py (hunk 7 above)
Comment (jh-nv) [reply components/src/dynamo/vllm/pooling_handlers.py:332]: Good catch! now `_build_pooling_tokenize_params` returns the renderer defaults insted of none, with overrides layered on top,  so token-ID inputs always go through `apply_post_tokenization` like upstream,  including its default truncation to `max_model_len`, which an over-long pre-tokenized prompt was previously skipping entirely.

File: components/src/dynamo/vllm/backend_args.py [hunk 8]
```diff
@@ -706,3 +720,31 @@ def _validate_realtime_worker_exclusivity(self) -> None:
...
+        if self.enable_multimodal:
+            raise ValueError(
+                "--classify-worker cannot be combined with multimodal flags."
+            )
+        if self.benchmark_mode is not None:
+            raise ValueError(
+                "--classify-worker cannot be combined with --benchmark-mode. "
+                "Benchmark mode injects InstrumentedScheduler, which is a "
+                "generation scheduler and not compatible with pooling engines."
+            )
```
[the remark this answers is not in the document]
jh-nv: Fixed in `b15bb8ba0e3` — `--headless` is now rejected, mirroring `_validate_realtime_worker_exclusivity`. Confirmed the mechanism: `main.worker()` returns at `run_dynamo_headless()` (main.py:160-162) before `factory.create()` at 183, so nothing registered. Test: `test_headless_combination_rejected`.

On LoRA: also rejected for now, but only as a stopgap. `EmbeddingWorkerHandler` is a plain class rather than a `BaseWorkerHandler` subclass, so it inherits none of the LoRA machinery — meaning the same silent-base-model behaviour is already live on `/v1/embeddings` in main. Proper plumbing is a handler-hierarchy change spanning both pooling roles, tracked in DIS-2639.

components/src/dynamo/vllm/pooling_handlers.py
```diff
@@ -0,0 +1,503 @@
...
+    if isinstance(data, torch.Tensor):
+        return data.detach().cpu().flatten().tolist()
+    if isinstance(data, (list, tuple)):
+        if data and isinstance(data[0], (list, tuple)):
+            return [float(value) for row in data for value in row]
+        return [float(value) for value in data]
+    raise TypeError(
+        f"Unsupported PoolingOutput.data type {type(data).__name__}; "
+        "expected torch.Tensor or list"
+    )
```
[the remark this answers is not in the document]
jh-nv: Fixed in `5acf9c6e56d`. `_classification_output_to_list` now raises on multi-row output instead of flattening, matching vLLM's `ndim != 1` rejection, while still unwrapping the legitimate `[1, num_classes]` singleton batch dim. Tests cover both accepted shapes, both rejected forms (tensor and list), and an end-to-end case asserting the handler surfaces the error rather than a mislabelled 200.

components/src/dynamo/vllm/tests/test_vllm_worker_factory.py
```diff
@@ -596,6 +609,81 @@ async def test_passes_snapshot_engine(self, factory: WorkerFactory) -> None:
...
+
+    engine_client = Mock()
+    vllm_config = Mock(model_config=Mock())
+    engine_tuple: EngineSetupResult = (
+        engine_client,
+        vllm_config,
+        Mock(),
+        "/tmp/prom",
+        None,
+    )
```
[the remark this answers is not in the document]
jh-nv: Leaving this one as-is. Our ruff config has no `select`, so the bandit `S` rules aren't enabled and S108 doesn't fire here. The literal is a placeholder string inside a mock `engine_tuple` and never reaches the filesystem, so there's no temp-file race behind the CWE-377 reference.

This PR adds one such literal, matching three pre-existing ones already in the same file (lines 591, 734, 819), so changing only the new one would make it the odd one out. Happy to switch all four to `tmp_path` in a follow-up if you'd prefer consistency the other way.

furionw (approved):
Thank you

karen-sy (approved):

PR #12155, opened 2026-07-25
fix(http): map request validation failures to HTTP 400 instead of 500

yifjiang:
> **Draft** — extracted from an internally-validated fix; opening for review. CI has not been run here.

## The bug

When an incoming request fails validation, `ValidateEngine::generate` returns a generic `anyhow::anyhow!("Validation failed: ...")` error. That error carries no type information, so the HTTP layer cannot tell it apart from an internal/server failure and returns **HTTP 500** for what is actually a **client input error**.

The OpenAI HTTP service already contains a `find_invalid_argument_in_chain` helper that maps a typed `InvalidArgument` error to a `400 invalid_request_error`, but:

1. `ValidateEngine` never produced a typed error for that helper to match on, and
2. the Anthropic HTTP service had no equivalent helper at all,

so validation failures on both surfaces still fell through to 500.

## The fix

Two halves that pair up:

- **`lib/llm/src/engines.rs`** — `ValidateEngine::generate` now returns a typed `DynamoError` with `ErrorType::InvalidArgument` (message preserved as `"Validation failed: ..."`) instead of an opaque `anyhow!`.

- **`lib/llm/src/http/service/anthropic.rs`** — adds `find_invalid_argument_in_chain`, which walks the error chain for a `DynamoError` whose type is `InvalidArgument` (or `Backend(InvalidArgument)`) and returns a `400 invalid_request_error` (marking the request as a `Validation` error for metrics). This mirrors the mapping the OpenAI service already performs.

Because the OpenAI service already had the helper and wiring, the typed error from `ValidateEngine` means it now also returns 400 for validation failures without any further change there.

Unit tests are included for both halves:
- `validate_engine_classifies_request_errors_as_invalid_argument` (engines.rs) — asserts the validation path yields `InvalidArgument` and never reaches the inner engine.
- `anthropic_invalid_argument_is_found_through_error_context` (anthropic.rs) — asserts the helper finds the typed error through an `anyhow` context wrapper.

## Testing checklist

- [ ] `cargo build -p dynamo-llm`
- [ ] `cargo test -p dynamo-llm engines::tests::validate_engine_classifies_request_errors_as_invalid_argument`
- [ ] `cargo test -p dynamo-llm http::service::anthropic::tests::anthropic_invalid_argument_is_found_through_error_context`
- [ ] `cargo clippy -p dynamo-llm`
- [ ] Manual: send a request that fails validation to the Anthropic `/v1/messages` endpoint and confirm the response is HTTP 400 with `type: "invalid_request_error"` (previously 500).
- [ ] Manual: confirm the OpenAI endpoint returns HTTP 400 for the same validation failure.

## Notes for reviewers

The edit maps cleanly onto current `main`: the validation block in `engines.rs` and the error-handling ladder in `anthropic_messages` are unchanged from the internally-validated base, and the `find_invalid_argument_in_chain` helper is byte-for-byte equivalent to the one already present in the OpenAI service. No unrelated changes are bundled in.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid request validation failures are now consistently classified as client errors.
  * The Anthropic Messages API returns HTTP 400 with an `invalid_request_error` response for invalid input.
  * Error messages now provide clearer details about the validation issue, even when errors are wrapped by other processing layers.
  * Invalid requests are rejected before backend generation begins, avoiding unnecessary processing.


rmccorm4 (approved):

rmccorm4:
Thanks @yifjiang for the contribution!

PR #12157, opened 2026-07-25
feat(embeddings): tokenizer alignment, multi-process pool, and worker byte-transport

anmolgupt:
## Overview
Embedding-worker improvements for the vLLM backend:
- Tokenizer alignment with vLLM pooling (`add_special_tokens` defaults to true)
- Optional Rust frontend tokenization (`DYN_EMBEDDING_FRONTEND_TOKENIZATION`)
- Multi-process Dynamo endpoint pool sharing one `EngineCore` per GPU
- Faster worker→frontend embedding bytes (`numpy.tobytes` + base64)
### 1. Tokenizer alignment + optional frontend tokenization
- Request field `add_special_tokens` (optional). Precedence: request > `DYN_EMBEDDING_TOKENIZATION_ADD_SPECIAL_TOKENS` > **default `true`**.
- Frontend tokenization for embeddings is opt-in via --embedding-frontend-tokenization or DYN_VLLM_EMBEDDING_FRONTEND_TOKENIZATION; embedding workers use vLLM tokenization by default.
- Embedding tokenizers are constructed only for the embedding pipeline and initialized lazily (with/without specials as separate variants).
- `add_special_tokens=true` uses the HuggingFace backend and bypasses the L1 tokenizer cache. `false` keeps the normal backend/cache path.
- Frontend applies `truncate_prompt_tokens` on the tokens path.
### 2. Multi-process endpoint pool
- `--embedding-worker-processes N` (parent counts as one). Shares one `EngineCore`; adds request-handling/tokenization capacity, not GPU throughput.
- Children bind `DYN_SYSTEM_PORT + index` when the parent port is enabled.
- Rejects fixed `DYN_TCP_RPC_PORT`, intra-pod failover, and overlapping system/FPM/NIXL port reservations when `N > 1`.
- Mutually exclusive with `--data-parallel-size > 1` and `--enable-lora`.
### 3. Worker→frontend transport
- Tokens path wire type: `EmbeddingsEngineOutput.embeddings: Vec<String>` (base64 float32).
- Worker always encodes with little-endian `numpy.tobytes` + base64 (no transport env var).
- **Rolling upgrade:** frontend and worker must be upgraded together when using the tokens path (`DYN_EMBEDDING_FRONTEND_TOKENIZATION`). Default text path is unaffected.
- Empty worker embedding outputs now surface as stream errors instead of returning an empty `data` list.
## Reviewer starting points
1. `components/src/dynamo/vllm/handlers.py` — tokens vs text input, encode path
2. `lib/llm/src/preprocessor.rs` — specials default, lazy tokenizers, truncation
3. `lib/llm/src/protocols/common/llm_backend.rs` — wire type
4. `components/src/dynamo/vllm/embedding_worker_processes.py` — process pool + ports
5. Tests under `components/src/dynamo/vllm/tests/` and `lib/llm/tests/preprocessor.rs`

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **New Features**
  * Added configurable embedding worker processes for improved embedding throughput.
  * Embedding requests now support pre-tokenized inputs, truncation controls, and optional special-token handling.
  * Added support for float and base64 embedding response formats, including dimension validation.
* **Bug Fixes**
  * Improved tokenizer consistency and right-truncation behavior across embedding requests.
  * Added clearer validation for incompatible worker, parallelism, LoRA, failover, and port configurations.
* **Tests**
  * Expanded coverage for tokenization, serialization, configuration validation, worker lifecycle, and cleanup.


lib/llm/src/protocols/common/llm_backend.rs
```diff
@@ -390,11 +390,11 @@ impl MaybeError for LLMEngineOutput {
...
 }
 
-/// Raw output from embedding engines containing embedding vectors
+/// Raw output from embedding engines containing base64-encoded float32 vectors.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct EmbeddingsEngineOutput {
-    /// Generated embedding vectors (one per input text)
-    pub embeddings: Vec<Vec<f64>>,
+    /// One standard-base64 string per input embedding.
+    pub embeddings: Vec<String>,
```
tzulingk, line 397: This changes the worker↔frontend wire contract (was Vec<Vec<f64>>). Please (1) confirm nothing else deserializes EmbeddingsEngineOutput, and (2) note in the PR description that the tokens path now needs frontend + worker deployed together — a rolling upgrade with mismatched versions will break it.

components/src/dynamo/vllm/handlers.py
```diff
@@ -3796,27 +3861,44 @@ async def _encode_one(idx: int, prompt: Any):
             if pending:
                 await asyncio.gather(*pending, return_exceptions=True)
 
+        use_numpy = (
+            _embedding_response_transport() == _EMBEDDING_RESPONSE_TRANSPORT_NUMPY
```
tzulingk: _embedding_response_transport() re-reads os.environ on every request. Resolve it once at worker init and pass the result in — it can't change at runtime.

lib/llm/src/protocols/common/preprocessor.rs
```diff
@@ -400,6 +390,7 @@ pub struct PreprocessedEmbeddingRequest {
     pub model: String,
 
     /// Encoding format preference
+    #[serde(skip_serializing_if = "Option::is_none")]
```
tzulingk, line 509: **Not a regression — flagging a gap this PR makes reachable.**

`truncate_prompt_tokens` has never been carried by this struct, and `OpenAIPreprocessor::preprocess_embedding_request` has never applied it. Truncation support lives only in the Python handler (`handlers.py`, in `EmbeddingWorkerHandler.generate`) and only for raw-text inputs.

That's been harmless so far because nothing could reach the tokens embedding pipeline — `worker_factory.py::_create_embedding_worker` registers `ModelInput.Text` unconditionally on main, so the frontend always builds the `Text + Embedding` branch in `discovery/watcher.rs`, which constructs no `OpenAIPreprocessor` at all. This PR changes that registration to `_embedding_model_input()`, so with `DYN_EMBEDDING_FRONTEND_TOKENIZATION=1` the preprocessor is built for the first time and the gap becomes visible.

**What a user would see.** They send `truncate_prompt_tokens` and it's ignored. Their long input either exceeds `max_model_len` and errors, or returns a vector for text they explicitly asked to truncate. No warning either way. The same request behaves differently depending on an env var they may not know is set.

**Why I'm raising it here rather than as a separate issue.** This PR adds exactly the plumbing that would carry it — `add_special_tokens` now rides through this struct and gets applied during Rust tokenization. Adding truncation alongside it is a small delta now and a rediscovery later.

**What I'd like, in preference order.**

**Option A — close it here.** Add the field:

```rust
/// Maximum prompt tokens to keep; -1 means the model's maximum length.
#[builder(default)]
pub truncate_prompt_tokens: Option<i64>,
```

Then apply it in `preprocess_embedding_request` after tokenizing, since Rust owns tokenizing on this path:

```rust
if let Some(limit) = request.truncate_prompt_tokens {
    let max = if limit == -1 { model_context_length } else { limit as usize };
    for ids in all_token_ids.iter_mut() {
        ids.truncate(max);
    }
}
builder.truncate_prompt_tokens(request.truncate_prompt_tokens);
```

Plus a test asserting the Text and Tokens paths produce the same token count for the same input and limit.

**Option B — defer it explicitly.** File a tracked issue, note the limitation in the `DYN_EMBEDDING_FRONTEND_TOKENIZATION` documentation, and log a warning when a request carries `truncate_prompt_tokens` on the tokens path so it isn't silent. In that case I'd ask that we not recommend the tokens path in any recipe or model card until it's closed.

What I'd like to avoid is shipping the env var with the gap unrecorded.

lib/llm/src/preprocessor.rs
```diff
@@ -2216,15 +2090,35 @@ impl OpenAIPreprocessor {
         let mut annotations = HashMap::new();
         let mut builder = PreprocessedEmbeddingRequest::builder();
 
+        let effective_add_special_tokens = request
```
tzulingk, line 3716: **Pre-existing divergence that this PR exposes — and here the PR adds the fix but doesn't take it.**

To be clear on what's yours: the tokens embedding path has always used `self.tokenizer`, which doesn't add marker tokens, while vLLM's pooling default does. That divergence predates this PR and was harmless because nothing could reach the path (`worker_factory.py::_create_embedding_worker` registered `ModelInput.Text` unconditionally). This PR makes it reachable and, to your credit, adds the mechanism to correct it.

My concern is only the default. When neither the request field nor the env var is set, this still falls back to the legacy tokenizer. Your own test pins that: `request_add_special_tokens_is_tri_state` asserts omitting the field gives the same IDs as passing `false`, and `true` prepends BOS.

**Why it matters — the configuration matrix.** Getting vLLM-equivalent vectors requires two env vars, read by two different processes, with no cross-validation:

- `DYN_EMBEDDING_FRONTEND_TOKENIZATION` — read in `worker_factory.py::_embedding_model_input`, so it must be set on the **worker**.
- `DYN_EMBEDDING_TOKENIZATION_ADD_SPECIAL_TOKENS` — read in `preprocessor.rs`, reached via `ModelWatcher` → `OpenAIPreprocessor::new_for_embeddings`. `ModelWatcher` is constructed in the frontend entrypoints (`entrypoint/input/http.rs`, `entrypoint/input/grpc.rs`), so it must be set on the **frontend**.

| worker: `..._FRONTEND_TOKENIZATION` | frontend: `..._ADD_SPECIAL_TOKENS` | tokenized by | marker tokens | matches vLLM |
|---|---|---|---|---|
| unset / `0` / `false` / `no` | (not consulted) | vLLM | yes | yes |
| `1` / `true` | `1` / `true` / `yes` | Rust | yes | yes |
| `1` / `true` | **unset** | Rust | **no** | **no — silent** |
| `1` / `true` | `0` / `false` / `no` | Rust | no | no (deliberate) |
| `1` / `true` | anything else, e.g. `on` | Rust | **no** (warns) | **no** |

So the only safe way to enable frontend tokenization is to set the worker variable **and** set the frontend variable to a recognized true value. Miss the frontend one — or typo it — and every request still succeeds while returning different vectors.

In Kubernetes those are two containers in two pod specs. Nothing validates that they agree, and the failure is invisible.

There's also a truthiness mismatch between the two parsers. `_embedding_model_input` treats anything outside `("", "0", "false", "no")` as ON, so `on`, `yes`, and `TRUE` all enable the path. `parse_embedding_add_special_tokens` accepts only `1/true/yes` and `0/false/no`, and warns-then-ignores everything else. So `..._FRONTEND_TOKENIZATION=on` works while `..._ADD_SPECIAL_TOKENS=on` quietly doesn't. Two variables that must agree, with different notions of true.

**Why the consequence is severe.** Every request returns HTTP 200 with a plausible vector. But vectors written during ingestion stop matching vectors computed at query time, so retrieval quality degrades quietly. By the time anyone notices, the remedy is re-embedding the corpus.

**What to change — preferred.** Default to `true` so one env var is sufficient and the frontend one becomes an override rather than a requirement:

```rust
let effective_add_special_tokens = request
    .add_special_tokens
    .or(self.embedding_add_special_tokens_default)
    .or(Some(true));
```

Then update `request_add_special_tokens_is_tri_state` to assert `omitted == explicit_true`, and `invalid_env_add_special_tokens_preserves_legacy_behavior` to expect the vLLM default.

This is easy to accept because there's no existing user to break: no in-tree worker registered `ModelInput::Tokens` for embeddings before this PR, so nobody depends on the current default. Changing it now costs nothing. Changing it after someone builds an index costs them a re-embed.

**What to change — acceptable alternative.** Keep the default, but make the unsafe row of that table impossible. `new_for_embeddings` already knows the model registered for token input, and it runs in the frontend where the second variable lives, so it can validate:

```rust
if embedding_add_special_tokens_env().is_none() {
    anyhow::bail!(
        "token-input embedding models require \
         DYN_EMBEDDING_TOKENIZATION_ADD_SPECIAL_TOKENS to be set to true or false. \
         Leaving it unset produces token IDs that differ from vLLM's pooling default."
    );
}
```

Failing at startup is fine. Silent divergence is the row I want removed.

**Third thing, independent of which option you pick.** Please align the two parsers on one truthiness rule, and have `_embedding_model_input` reject unrecognized values instead of treating them as ON. Right now a typo in one variable enables the feature and a typo in the other silently disables the correctness fix.

**One documentation request.** The original write-up said "text inputs now default to vLLM pooling behavior," and the earlier prototype did default to `true`. If dropping that was deliberate, please note it in the PR description — the description and the code currently disagree, and readers will trust the description.

components/src/dynamo/vllm/embedding_worker_processes.py
```diff
@@ -0,0 +1,481 @@
...
+    parent_pid: int,
+) -> dict[str, str]:
+    env = os.environ.copy()
+    env[_ROLE_ENV] = _CHILD_ROLE
+    env[_INDEX_ENV] = str(process_index)
+    env[_PARENT_PID_ENV] = str(parent_pid)
+    env[_ENGINE_ADDRESSES_ENV] = addresses_json
+    # Each Dynamo runtime process needs a distinct system-status listener.
+    # Port zero asks the OS to select a collision-free ephemeral port.
+    env["DYN_SYSTEM_PORT"] = "0"
```
tzulingk: **Problem.** `DYN_SYSTEM_PORT` is the health and metrics listener (documented in `deploy/operator/docs/footer.md`), and the Kubernetes operator sets it explicitly on worker containers (`deploy/operator/internal/dynamo/component_worker.go`, and again in `component_planner.go` and `failover.go`). Setting it to `0` tells each child to bind whatever port the OS hands out.

**Why it matters.** With `--embedding-worker-processes 8`, the operator's configured port reaches the parent and nothing else. The other seven bind ports nobody can discover. Two concrete consequences:

- Prometheus scrapes 1 of 8 processes, so throughput and latency dashboards under-report by roughly 8x with no indication that data is missing.
- Per-child health checks are impossible. A child that's running but wedged looks healthy from outside the pod.

**What to change.** Give each child a deterministic port derived from the parent's, and leave children disabled if the parent's is disabled:

```python
parent_port = os.environ.get("DYN_SYSTEM_PORT")
if parent_port and int(parent_port) > 0:
    env["DYN_SYSTEM_PORT"] = str(int(parent_port) + process_index)
# else: inherit the parent's disabled setting rather than forcing a port
```

Then two follow-ups:

1. Document in the `--embedding-worker-processes` help text that the pool occupies `DYN_SYSTEM_PORT` through `DYN_SYSTEM_PORT + N - 1`, so operators can reserve the range.
2. Add a validation error if that range would collide with anything else the worker binds.

If metrics from the children are already aggregated somewhere I've missed, point me at it and I'll withdraw this.

components/src/dynamo/vllm/handlers.py
```diff
@@ -3846,6 +3939,29 @@ async def _encode_one(idx: int, prompt: Any):
         }
 
 
+def _embedding_response_transport() -> str:
```
tzulingk: **Problem.** The faster encoder is opt-in behind an undocumented env var, so the default deployment keeps using the slow one.

**Why it matters.** Nobody sets undocumented env vars. As written, the main serialization optimization in this PR merges and then does nothing in practice. We also carry two encoders, an env parser with fallback semantics, and three tests whose only purpose is proving the two encoders agree.

**Why a flag isn't needed here.** Flags are for changes that are risky or that change behavior. Neither applies:

- The two encoders produce identical bytes — your own `test_pooling_tensor_base64_matches_struct_pack` proves it.
- numpy isn't a new dependency. torch already requires it, and `Tensor.numpy()` is core torch.
- The only real difference is byte order: `tobytes()` uses native order, `struct.pack("<...")` forces little-endian. They diverge only on big-endian CPUs, and no big-endian CUDA platform exists.

**What to change.**

1. Delete `_embedding_response_transport()`, the three `_EMBEDDING_RESPONSE_TRANSPORT_*` constants, the `use_numpy` variable, and `_encode_row_base64`.

2. Call the numpy encoder directly in the response loop:

```python
encoded = _pooling_output_to_base64(embedding_row)
```

3. Make byte order explicit in `_flatten_pooling_tensor` so the little-endian caveat disappears entirely rather than being argued away:

```python
return data.detach().cpu().flatten().to(torch.float32).contiguous().numpy().astype("<f4", copy=False)
```

(Return the numpy array and let `_pooling_output_to_base64` call `.tobytes()` on it; `_pooling_output_to_list` can call `.tolist()`.)

4. Delete `test_invalid_response_transport_warns_and_uses_default`, `test_text_numpy_response_transport_uses_base64_bytes`, and `test_tokens_numpy_response_matches_default_wire_shape`. Keep `test_pooling_tensor_base64_matches_struct_pack` as the regression guard that the bytes are still spec-correct.

lib/llm/src/preprocessor.rs
```diff
@@ -643,13 +628,48 @@ impl OpenAIPreprocessor {
...
+        let PromptFormatter::OAI(formatter) = embedding_prompt_formatter(&mdc)?;
+        Self::new_with_parts(mdc, formatter, tokenizer)
+    }
+
     pub fn new_with_parts(
         mdc: ModelDeploymentCard,
         formatter: Arc<dyn OAIPromptFormatter>,
         tokenizer: crate::tokenizers::Tokenizer,
     ) -> Result<Arc<Self>> {
+        let supports_embedding = mdc.model_type.supports_embedding();
```
tzulingk: **Problem.** This gate is looser than it looks. `ModelType` is bitflags (see the `bitflags!` block in `lib/llm/src/model_type.rs` — `Chat` and `Embedding` are separate bits), so a card can be `Chat | Embedding`. And `new_with_parts` is the general chat and completions constructor — it's called from the two `ModelInput::Tokens` chat/completions branches in `discovery/watcher.rs` and from `entrypoint/input/common.rs`.

**Why it matters.** For a model registered as both chat and embedding, the chat and completions pipelines each build two HuggingFace tokenizers they will never use — paying model load time and holding the memory for the process lifetime.

**What to change — minimal.** Build them on first use instead of at construction. Change the two fields to lazily-initialized cells:

```rust
embedding_tokenizer_with_special_tokens: std::sync::OnceLock<Arc<dyn Tokenizer>>,
embedding_tokenizer_without_special_tokens: std::sync::OnceLock<Arc<dyn Tokenizer>>,
```

Store the `ModelDeploymentCard` (or just the pieces needed) so `preprocess_embedding_request` can call `get_or_try_init` on the branch it actually takes. Chat pipelines then never construct either one, and an embedding pipeline only builds the variant it uses.

**What to change — better, if you're open to it.** Pass `add_special_tokens` as an argument to `encode()` rather than baking it into the tokenizer at construction. That's what your earlier prototype did (`encode_with_special_tokens(s, add_special_tokens)`), and it solves three problems at once: one tokenizer instead of three, no interaction with the tokenizer cache (see my comment on `model_card.rs:1258`), and no unreachable branch (see my comment on `model_card.rs:1232`).

What made you move away from that? If there's a reason the per-call form doesn't work, please capture it in a comment — the construction-time form costs us in all three of those places, so a future reader will wonder.

lib/llm/src/model_card.rs
```diff
@@ -1286,7 +1255,7 @@ impl ModelDeploymentCard {
                     Arc::new(wrap_hf(hf))
                 };
 
-                if cache_enabled {
+                if cache_enabled && !options.add_special_tokens {
```
tzulingk, line 1433: **Problem.** Two conditions combine here in a way a reader won't notice. `force_hugging_face` (via the `use_fast` computation in `tokenizer_with_options`) turns off the fast tokenizer backend, and this line turns off the tokenizer cache. Both trigger on the embedding tokenizer with `add_special_tokens = true`.

Since `true` is the vLLM-correct setting for pooling models, **the correct configuration is the one that gets neither the fast backend nor the cache.**

No objection to the behavior — the reasoning in the comment below is sound, since an incremental cache could re-run the post-processor per segment and add markers repeatedly. My concern is that this is a surprising property of a performance PR and nothing states it.

**What to change.**

1. Add a sentence to the PR description under the tokenizer section: enabling `add_special_tokens` selects the HuggingFace backend and bypasses the tokenizer cache, and here's why that's acceptable for embeddings (embedding inputs aren't extended incrementally the way chat history is, so the prefix cache buys little).

2. Add that same "acceptable for embeddings because…" reasoning to the code comment below, so the next reader doesn't have to rediscover it.

3. One question worth checking before we accept the cache loss: the doc comment on the tokenizer cache in this file distinguishes a plain cache from the `cache_extend` incremental behavior. Is the correctness problem specific to `cache_extend`? If a whole-input cache is safe with special tokens, we could keep caching and disable only the extend path.

components/src/dynamo/vllm/backend_args.py
```diff
@@ -178,6 +182,17 @@ def add_arguments(self, parser) -> None:
             "and InstrumentedScheduler injection (none apply to pooling models).",
         )
 
+        add_argument(
+            g,
+            flag_name="--embedding-worker-processes",
+            env_var="DYN_VLLM_EMBEDDING_WORKER_PROCESSES",
+            default=1,
+            arg_type=int,
+            help="Number of Dynamo embedding endpoint processes sharing one "
```
tzulingk, line 248: **A rule in the help text**, something like: "Choose roughly `min(CPU cores available to this pod, expected peak concurrency)` (**just an example, need your judgement here**). Values above the core count are unlikely to help. Only useful when concurrency > 1, since a single in-flight request occupies one process."

tzulingk (changes requested):

lib/llm/src/protocols/common/llm_backend.rs
anmolgupt: Only the vLLM tokens embedding path uses EmbeddingsEngineOutput

components/src/dynamo/vllm/handlers.py
anmolgupt: dropped the env var

lib/llm/src/protocols/common/preprocessor.rs
anmolgupt: added Option A support

lib/llm/src/preprocessor.rs
anmolgupt: default is not true for frontend tokenization. env parser for frontend/backend are aligned

components/src/dynamo/vllm/embedding_worker_processes.py
anmolgupt: Added more checks and improved the help text for the arg.

components/src/dynamo/vllm/handlers.py
anmolgupt: Now it always uses numpy

lib/llm/src/preprocessor.rs
anmolgupt: added lazy init support.
dynamo-tokenizers's Encoder only has encode(input) function.

lib/llm/src/model_card.rs
anmolgupt: updated the comment

components/src/dynamo/vllm/backend_args.py
anmolgupt: Added more info in the arg.

components/src/dynamo/vllm/backend_args.py
```diff
@@ -785,3 +732,107 @@ def _validate_embedding_worker_exclusivity(self) -> None:
...
+        tcp_rpc_port = _configured_fixed_port("DYN_TCP_RPC_PORT")
+        if request_plane == "tcp" and tcp_rpc_port is not None:
+            raise ValueError(
+                "DYN_TCP_RPC_PORT cannot be fixed when "
+                "--embedding-worker-processes is greater than 1 because every "
+                "endpoint process needs a unique TCP RPC listener. Unset "
+                "DYN_TCP_RPC_PORT to use OS-assigned ports."
+            )
+
+        system_range = self._validate_system_port_range()
```
tzulingk, line 858: **Child processes validate a port range they don't own.**

`args.py:109` calls `dynamo_config.validate()` unconditionally, and
`embedding_worker_processes.py::create_shared_embedding_engine_client` launches
children with `[sys.executable, "-m", "dynamo.vllm", *sys.argv[1:]]` — the same
argv, so `embedding_worker_processes` is still N in the child. But by then
`_child_environment` (line 180) has rewritten `DYN_SYSTEM_PORT` to `base + index`.

So a child at index `i` reaches this line and validates the range
`[base+i, base+i+N-1]`, which it never claims. Two consequences:

1. `_validate_system_port_range` can raise for a child near the top of the port
   space over a range that doesn't exist. The parent already validated the real
   range correctly, so this is a pure false negative at startup.
2. `_validate_port_reservation_collisions` then compares `DYN_FORWARDPASS_METRIC_PORT`
   and `NIXL_TELEMETRY_PROMETHEUS_PORT` against that phantom range, so it can
   report a collision that isn't real or miss one that is.

Only the parent owns the range, so only the parent should check it:

```python
from .embedding_worker_processes import is_embedding_process_child

if is_embedding_process_child():
    return

system_range = self._validate_system_port_range()
self._validate_port_reservation_collisions(system_range)

File: lib/llm/src/preprocessor.rs [hunk 11]
```diff
@@ -98,25 +102,45 @@ fn routing_priorities(hints: Option<&AgentHints>) -> (Option<f64>, Option<u32>,
     (priority_jump, strict_priority, priority)
 }
 
-fn invalid_argument_error(message: impl Into<String>) -> anyhow::Error {
-    DynamoError::builder()
-        .error_type(ErrorType::InvalidArgument)
-        .message(message.into())
-        .build()
-        .into()
+fn decode_base64_to_floats(encoded: &str) -> Result<Vec<f32>, String> {
```
tzulingk: **Two identical base64 decoders — please consolidate.**
This is the same function as `decode_base64_embedding_to_floats` at
`lib/llm/src/http/service/openai.rs:1231`: same alphabet, same multiple-of-4
check, same `f32::from_le_bytes` loop. Only the error type differs (`String` vs
`anyhow::Error`). Each also has its own tests, and both tests here are strict
subsets of the three `decode_base64_embedding_to_floats_*` tests at
`openai.rs:6236`.

Could we consolidate?

lib/llm/src/preprocessor.rs
```diff
@@ -2805,38 +2783,37 @@ impl OpenAIPreprocessor {
...
+        // Honor the client's OpenAI `encoding_format` at this boundary:
+        // pass base64 through, or decode to floats (the public default).
         let encode_base64 = matches!(
             original_request.inner.encoding_format,
             Some(dynamo_protocols::types::EncodingFormat::Base64)
         );
         stream.map(move |output| {
             output.map_data(|engine_output| {
-                // Convert engine output to OpenAI response format
+                if engine_output.embeddings.is_empty() {
```
tzulingk, line 4613: **New failure mode, unmentioned and untested.**

An empty `embeddings` vector now becomes a stream error rather than an empty
response list. I think that's the right call — an embedding response with no
vectors is never useful, and failing loudly beats handing the client an empty
`data` array.

But it isn't in the PR description and nothing covers it. Please add a test that
an empty `EmbeddingsEngineOutput` surfaces as an error, and note the change in the
description under the transport section, since it is technically a behavior change
on the response path.

tzulingk (approved):
Overall looks good

lib/llm/src/preprocessor.rs
anmolgupt: Added the fix

lib/llm/src/preprocessor.rs
anmolgupt: Added the fix

components/src/dynamo/vllm/backend_args.py
anmolgupt: Added the fix

rmccorm4:
/ok to test [7b2e869](https://github.com/ai-dynamo/dynamo/pull/12157/commits/7b2e8698594f5e7c8ef9cb1a37a380efde85661c)

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -935,7 +956,7 @@ async def _create_embedding_worker(
                     health_check_payload=embedding_health_check_payload,
                 ),
                 self.register_vllm_model(
-                    ModelInput.Text,
+                    _embedding_model_input(),
```
rmccorm4: Can we reuse the existing `--use-vllm-tokenizer` flag for now instead of introducing new env var?

ref: https://github.com/ai-dynamo/dynamo/blob/8ccbe9636a329f2bde931b470832ddeabc56f4c0/components/src/dynamo/vllm/worker_factory.py#L1294

tzulingk: How about keeping it opt-in, but make it a first-class knob instead of a bare env var. Register
it in backend_args.py via add_negatable_bool_argument, default off. That gives us the
flag plus the paired DYN_* env var in one shot, so it shows up in --help and in the
vLLM config reference, and it lands on DynamoVllmConfig where we can reject combining
it with --use-vllm-tokenizer at parse time.

And let's write the sunset into the help text and release notes: next release, pooling
workers start reading --use-vllm-tokenizer like generative workers do (so false =
frontend tokenizes), and we delete this flag. That makes it explicitly a one-release
compatibility gate to give users notice, not a permanent second tokenizer switch.

lib/llm/src/protocols/common/llm_backend.rs
jh-nv: [note] this endpoint is not released in 1.4.0, so it is safe for this breaking change.

lib/llm/src/model_card.rs
```diff
@@ -1231,7 +1231,26 @@ impl ModelDeploymentCard {
...
     pub fn tokenizer(&self) -> anyhow::Result<crate::tokenizers::Tokenizer> {
-        let tokenizer_backend = self.runtime_config.effective_tokenizer_backend();
+        self.tokenizer_with_options(Default::default(), false)
+    }
+
+    pub(crate) fn embedding_tokenizer_with_options(
+        &self,
+        options: crate::tokenizers::TokenizerOptions,
+    ) -> anyhow::Result<crate::tokenizers::Tokenizer> {
+        self.tokenizer_with_options(options, true)
```
jh-nv: Here are hardcode to force_hugging_face, rather  than depending on options.add_special_tokens, so both add_special_tokens=true and false force TokenizerBackend::Default.

anmolgupt: Thanks! Added the change to use FastTokenizer when add_special_tokens is False.

jh-nv (approved):

rmccorm4 (approved):

anmolgupt:
@rmccorm4 , @jh-nv: Could you please launch the CI pipeline tests to complete the pending checks?

Issue #12016, opened 2026-07-22
[CONTRIBUTION]: ThunderAgent Improvements


ashnamehrotra:
### Type of Change

Bug fix

### Problem Statement

# ThunderAgent: Kubernetes Integration Bugs & Observability Gaps

## Summary

During end-to-end testing of ThunderAgent on Kubernetes via the Dynamo Operator (DGD deployment), several bugs prevent core functionality from working and observability gaps make it impossible to verify behavior. This tracking issue covers the fixes needed to make ThunderAgent demo-ready and production-viable on K8s.

## Context

- Deployment: DGD with Frontend → ThunderAgentRouter → VllmDecodeWorker (1x A100, Qwen3-0.6B)
- Image: built from main branch (post-PR #9448)
- Operator: Dynamo Operator in `dynamo-system` namespace

---

### 1. WorkerCapacityProvider namespace mismatch prevents pause/resume

**Severity: High — core scheduling feature is non-functional on K8s**

The ThunderAgent scheduler's pause/resume logic never fires in K8s because `WorkerCapacityProvider` can't discover the worker's MDC.

- Worker publishes MDC at: `dynamo-system-thunderagent-demo-<hash>.backend`
- Router subscribes at: `dynamo.backend` (from `--endpoint dynamo.backend.generate`)
- Result: `capacity.snapshot()` always returns empty, scheduler tick no-ops

The CLI reference script (`run_minimax_8xh100.sh`) works because all components share a flat namespace. The K8s operator uses namespaced discovery that doesn't match.

**Fix:** Either the operator needs to wire up the namespaced endpoint correctly, or ThunderAgentRouter needs to discover the worker's actual published namespace from the runtime context.

---

### 2. `session-final` still forwards to GPU engine

When `x-dynamo-session-final: true` is sent, ThunderAgent still calls `KvRouter.generate_from_request()` and runs full inference before releasing the program. This wastes a GPU forward pass on a throwaway "done" message.

**Expected:** Short-circuit after releasing the program — return immediately without an engine call.

**File:** `components/src/dynamo/thunderagent_router/__main__.py`

---

### 3. DGDR profiler doubles model path

The profiler generates `--model-path /opt/models/opt/models` by concatenating `modelCache.pvcMountPath` with `HF_HOME`. Requires manual fixup of the generated DGD.

**Steps to reproduce:**
1. Submit a DGDR with `modelCache.pvcMountPath: /opt/models` and env `HF_HOME=/opt/models`
[...]

PR #12160, opened 2026-07-25
fix(router): honor Kubernetes worker namespace suffix

cpakkamisaac-sae:
## Overview:

Fix standalone router worker-endpoint resolution for Kubernetes deployments that publish worker-class components in a suffixed Dynamo namespace during rolling-update-aware DGD rendering.

## Details:

- Resolve the router's target worker endpoint with `get_worker_namespace()` so `DYN_NAMESPACE_WORKER_SUFFIX` is included when the operator injects it into worker-class router pods.
- Preserve `config.namespace` as the base `DYN_NAMESPACE`, so the standalone router continues to serve its own endpoints in the DGD namespace while discovering backend workers in the suffixed worker namespace.
- Add isolated router config tests covering default endpoint parsing, `DYN_NAMESPACE`, and `DYN_NAMESPACE_WORKER_SUFFIX`.
- Add the `router` pytest marker to the new router test module.

Validation:

- `PYTHONPATH=components/src UV_CACHE_DIR=/tmp/codex-uv-cache uv run --no-project --with pytest --with pytest-asyncio python -m pytest -c /dev/null components/src/dynamo/router/tests/test_args.py components/src/dynamo/router/tests/test_standalone_router.py`
- Validated in AKS with `aks-dynamo-thunderagent` in resource group `cpakkamisaac-sae-msft-east` (`eastus`) using a `Standard_NC4as_T4_v3` GPU node pool and model `Qwen/Qwen3-0.6B`.
- Kubernetes runtime validation used Ashna's demo runtime image with an in-container startup patch equivalent to this PR change; no PR-specific container image was built.
- Reproduced the namespace mismatch on the earlier `dynamo-platform-1.2.1` chart: the router watched `dynamo-system-thunderagent-demo.backend.generate` while the vLLM worker published under `dynamo-system-thunderagent-demo-588c6644.backend.generate`; the second same-session ThunderAgent request returned HTTP 500 with no routable endpoints.
- Validated the fix behavior on `dynamo-platform-1.2.1`: router discovery watched `dynamo-system-thunderagent-demo-03302af8.backend.generate`, the DGD became Ready, and first plus second same-session `/v1/chat/completions` calls returned HTTP 200.
- Upgraded the same AKS cluster to latest Dynamo release `v1.3.0` / chart `dynamo-platform-1.3.0`, applying the large Grove/KAI CRDs with server-side apply where needed. The DGD became Ready, router logs showed discovery against `dynamo-system-thunderagent-demo-03302af8.backend.generate`, `/v1/models` returned HTTP 200, and both first plus second same-session `/v1/chat/completions` calls returned HTTP 200.
- `v1.3.0` still does not list `router` as a valid `DynamoComponentDeployment.spec.type` enum value (`frontend`, `worker`, `prefill`, `decode`, `planner`, `epp` are valid), so the demo manifest still needs `ThunderAgentRouter` represented as `componentType: worker`. That is separate from the router worker-namespace bug fixed here.

Start with `components/src/dynamo/router/args.py`, specifically `DynamoRouterConfig.validate()`, then review `components/src/dynamo/router/tests/test_args.py` for the namespace-resolution cases.

## Related Issues

- Relates to #12016


components/src/dynamo/router/tests/test_args.py
```diff
@@ -0,0 +1,193 @@
...
+
+import importlib.util
+import os
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+pytestmark = [pytest.mark.pre_merge, pytest.mark.unit, pytest.mark.gpu_0]
```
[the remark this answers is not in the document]
cpakkamisaac-sae: Fixed in `2cbd3115` by adding `pytest.mark.router` to the new router test module. Re-ran the focused router tests successfully.

cpakkamisaac-sae:
Hi @ishandhanani, could you please review this PR when you have a chance? It fixes the namespace mismatch described in #12016 by honoring the Kubernetes worker-namespace suffix for router endpoints used by the ThunderAgent demo. I added unit coverage and validated the fix on AKS against Dynamo Platform v1.3.0.

components/src/dynamo/router/args.py
```diff
@@ -16,8 +18,11 @@
     KvRouterConfigBase,
 )
 from dynamo.common.configuration.utils import add_argument, add_negatable_bool_argument
+from dynamo.common.utils.namespace import get_worker_namespace
 from dynamo.llm import AicPerfConfig, KvRouterConfig
 
+logger = logging.getLogger(__name__)
```
ishandhanani: Should we use the dynamo logging utility here?

components/src/dynamo/router/tests/test_args.py
```diff
@@ -0,0 +1,198 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ishandhanani: We don't need this test file

components/src/dynamo/router/args.py
```diff
@@ -42,7 +47,19 @@ def validate(self) -> None:
...
             )
-        self.namespace = parts[0]
+        endpoint_namespace, component, endpoint_name = parts
+        self.namespace = os.environ.get("DYN_NAMESPACE") or endpoint_namespace
+
+        worker_namespace = get_worker_namespace(self.namespace)
+        if worker_namespace != endpoint_namespace:
+            old_endpoint = self.endpoint
+            self.endpoint = f"{worker_namespace}.{component}.{endpoint_name}"
+            logger.info(
```
ishandhanani: TBH not even sure we need this log at all

ishandhanani (commented):
Some nits. Pls ping me on slack when this is ready

cpakkamisaac-sae:
@ishandhanani updated based on your recommendations.

Changes made:
- Removed the extra namespace-resolution log and `logging` import.
- Removed the added `router/tests/test_args.py` file.
- Kept the runtime namespace fix minimal in `components/src/dynamo/router/args.py`.

I also revalidated the behavior end-to-end on AKS with the ThunderAgent deployment from #12016: the router resolved the suffixed worker namespace, discovered the vLLM worker, and both `/v1/models` and `/v1/chat/completions` returned successfully.

ishandhanani (approved):

PR #12161, opened 2026-07-25
perf(tokens): remove redundant inner DashMap sharding

jthomson04:
## Summary

- replace each positional radix bucket's inner `DashMap` with `FxHashMap`
- retain the outer `DashMap` as the concurrency boundary
- add focused same-position concurrent insertion coverage
- reduce peak mocker RSS by 5.51 GiB (35.3%) in a 32-worker AgentX A/B

Every inner-map access already holds the outer `DashMap`'s exclusive `RefMut`, so inner sharding provided no additional concurrency while replicating shard scaffolding for every position.

## Validation

- `cargo test -p dynamo-tokens positional_radix_tree`
- `cargo test -p kvbm-logical registry::tests`
- `cargo test -p kvbm-logical concurrent`
- unprofiled AgentX A/B: 32 workers, 128,000 blocks per worker, block size 16, concurrency 512
  - baseline: 15,996.0 MiB peak RSS, 243.90 req/s
  - patched: 10,356.2 MiB peak RSS, 251.97 req/s
  - reduction: 5,639.9 MiB / 5.51 GiB / 35.3%
  - both arms: 32/32 workers, zero errors, zero cancellations, zero affinity violations

The A/B contains one accepted run per arm. Throughput was healthy in both arms, but the observed difference is not presented as a speedup claim.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved registration cleanup to prevent newer registrations from being removed when older handles are dropped.
  - Improved reliability when multiple registration operations occur concurrently.

- **Performance**
  - Optimized internal token indexing to reduce nested map overhead.

- **Tests**
  - Added coverage for concurrent token insertion and registration replacement scenarios.
  - Updated test handling for safer value retrieval.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2904,6 +2904,7 @@ dependencies = [
  "derive-getters",
  "itertools 0.14.0",
  "rmp-serde",
+ "rustc-hash 2.1.2",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
--- a/lib/bindings/kvbm/Cargo.lock
+++ b/lib/bindings/kvbm/Cargo.lock
@@ -1906,6 +1906,7 @@ dependencies = [
  "bytemuck",
  "dashmap",
  "derive-getters",
+ "rustc-hash 2.1.2",
  "serde",
  "thiserror 2.0.18",
  "uuid",
--- a/lib/bindings/python/Cargo.lock
+++ b/lib/bindings/python/Cargo.lock
@@ -2582,6 +2582,7 @@ dependencies = [
  "bytemuck",
  "dashmap",
  "derive-getters",
+ "rustc-hash 2.1.2",
  "serde",
  "thiserror 2.0.18",
  "uuid",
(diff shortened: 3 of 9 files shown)


jthomson04:
Three new `clippy::map_clone` errors under `-D warnings`.

CI runs `cargo clippy --no-deps --all-targets -- -D warnings` (`.github/workflows/pre-merge.yml:232`). Because the inner map is now an `FxHashMap`, `get()` returns `Option<&V>` instead of `Option<dashmap::mapref::one::Ref<'_, K, V>>`, which makes clippy fire on pre-existing test code this PR does not touch:

- `lib/tokens/src/lib.rs:2893` — `.map(|v| v.clone())` → `.cloned()`
- `lib/tokens/src/lib.rs:2913` — `.map(|v| *v)` → `.copied()`
- `lib/tokens/src/lib.rs:2914` — `.map(|v| *v)` → `.copied()`

Reproduced on the PR head vs. the merge-base (`61d0289`):

```
$ cargo clippy --no-deps -p dynamo-tokens --all-targets -- -D warnings
# 61d0289 (base): exit 0
# 36c0431 (PR head): exit 101 — "could not compile `dynamo-tokens` (lib test) due to 3 previous errors"
```

The validation list in the description covers `cargo fmt --all -- --check` but not clippy, which is likely how this slipped through.

PeaBrane (approved):

nv-anants (approved):

PR #12163, opened 2026-07-25
fix(vllm): seed KV gauges for every DP rank, not just the last

### LINKED ISSUE #12052: vLLM StatLoggerFactory seeds KV gauges for only the last DP rank (prefill + decode)
### Summary

`StatLoggerFactory` in `components/src/dynamo/vllm/publisher.py` seeds KV cache gauges for only one data-parallel rank. `create_stat_logger()` stores each new publisher in a single field, `self.created_logger`, overwriting the previous one on every per-rank call. The seed helpers `set_num_gpu_blocks_all()` and `init_publish()` then act on that single field, so only the last DP rank created gets its `dynamo_component_total_blocks` and `dynamo_component_gpu_cache_usage_percent` labels seeded before the first scheduler record.

This is pre-existing behavior on the decode/aggregated path (`worker_factory.py` lines ~282–304). PR #12041, which fixes #11919 by adding the same seed step to the prefill worker, mirrors the existing decode contract and so inherits the same limitation. The `dynamo-review-agent` bot flagged it on that PR.

### Impact

Low severity, startup-only. Each per-rank `DynamoStatLoggerPublisher` is a distinct object that vLLM retains and drives; its `record()` writes labels keyed by its own `dp_rank` on every scheduler step. So once real scheduler stats flow, every DP rank populates its own labels regardless of the overwrite. The gap is confined to the pre-first-record window: for `DP > 1`, non-last ranks lack their zero-valued seeded labels until the first record arrives, then populate normally. There is no impact for `DP = 1` (the common case), which PR #12041 fully fixes.

### Fix

Have `StatLoggerFactory` retain every created publisher and broadcast the seed to all of them. This corrects both the prefill and decode paths, since they share the factory.

```diff
-        self.created_logger: Optional[DynamoStatLoggerPublisher] = None
+        self.created_loggers: list[DynamoStatLoggerPublisher] = []

     def create_stat_logger(self, dp_rank: int) -> StatLoggerBase:
         ...
         logger = DynamoStatLoggerPublisher(
             endpoint=self.endpoint,
             dp_rank=dp_rank,
             component_gauges=self.component_gauges,
         )
-        self.created_logger = logger
-
+        self.created_loggers.append(logger)
         return logger

     def set_num_gpu_blocks_all(self, num_blocks: int) -> None:
-        if self.created_logger:
-            self.created_logger.set_num_gpu_block(num_blocks)
+        for logger in self.created_loggers:
+            logger.set_num_gpu_block(num_blocks)

     def init_publish(self) -> None:
-        if self.created_logger:
-            self.created_logger.init_publish()
+        for logger in self.created_loggers:
+            logger.init_publish()
```

`created_logger` and `create_stat_logger` are referenced only within `publisher.py`; no other module reads `factory.created_logger`, so the change is self-contained. Add a unit test that drives `create_stat_logger()` for two DP ranks and asserts both publishers receive the seeded block count and `init_publish()`.

### Related

- Surfaced by PR #12041 (review comment on `worker_factory.py`)
- Fixes the general case behind #11919

1fanwang:
## Summary

`StatLoggerFactory` kept only the last publisher vLLM asked it to create, so the KV
capacity seed reached a single data-parallel rank.

vLLM calls the factory once per DP rank and retains each returned
`DynamoStatLoggerPublisher`. The factory stored each one in a single field,
overwriting the previous, so `set_num_gpu_blocks_all()` and `init_publish()`, called
once at engine start (`worker_factory.py:920-921`), only ever reached the last rank.

That gap is not startup-only. `num_gpu_block` defaults to `1` (`publisher.py:35`) and
`record()` re-reads it on every scheduler iteration (`publisher.py:64-68`), so an
unseeded rank publishes `kv_used_blocks = int(1 * kv_cache_usage)`, `0` for any usage
below 100%, and pins `total_blocks` at `1` for the worker's lifetime. The KV router
reads that rank as permanently empty and keeps sending it work, so the cost is routing
correctness rather than a missing metric label.

This hits the internal and hybrid DP load-balancing shapes, where one worker owns
several ranks (`get_dp_range_for_worker`). External LB gives each worker exactly one
rank, so it and `DP = 1` are unaffected.

The factory now retains every per-rank publisher and broadcasts the seed to all of them.

Fixes #12052

## Validation

Real `vllm==0.23.0` and the `dynamo.vllm` from this branch, DP 2 over 16384 KV blocks
(8192 per rank) at `kv_cache_usage=0.5`, driving the real `DynamoStatLoggerPublisher`
for three scheduler iterations. Only `WorkerMetricsPublisher`, the Rust/NATS boundary,
which carries the values but does not compute them, is substituted, to capture what
the KV router would receive.

| dp_rank | `num_gpu_block` | `kv_used_blocks` published | `total_blocks` gauge |
| --- | --- | --- | --- |
| 0 (before) | 1 | `[0, 0, 0]` | 1.0 |
| 1 (before) | 8192 | `[4096, 4096, 4096]` | 8192.0 |
| 0 (after) | 8192 | `[4096, 4096, 4096]` | 8192.0 |
| 1 (after) | 8192 | `[4096, 4096, 4096]` | 8192.0 |

Rank 0 stays empty on every iteration before the fix, not just the first, which is the
router-visible symptom. `test_unseeded_rank_reports_empty_to_the_kv_router` pins those
values so the regression cannot come back as a passing test.

<details>
<summary>Raw logs</summary>

Unit tests, before the fix (`publisher.py` reverted to `main`, tests kept):

```console
$ PYTHONPATH=components/src python -m pytest \
    components/src/dynamo/vllm/tests/test_vllm_publisher.py -v
platform linux -- Python 3.12.13, pytest-9.1.1
rootdir: /repo
configfile: pyproject.toml

published  = [(0, 0), (1, 4096), (0, 0), (1, 4096), (0, 0), (1, 4096)]
E       AssertionError: assert [(0, 0), (1, ...0), (1, 4096)] == [(0, 4096), (...6), (1, 4096)]
E         At index 0 diff: (0, 0) != (0, 4096)

FAILED ...::test_factory_returns_noop_logger_for_embedding_worker
FAILED ...::test_seed_broadcasts_to_all_dp_ranks
FAILED ...::test_unseeded_rank_reports_empty_to_the_kv_router
3 failed, 3 passed in 4.97s
```

Unit tests, with the fix:

```console
$ PYTHONPATH=components/src python -m pytest \
    components/src/dynamo/vllm/tests/test_vllm_publisher.py -v
collected 6 items

components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_factory_returns_noop_logger_for_embedding_worker PASSED [ 16%]
components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_noop_stat_logger_record_is_safe_with_none_stats PASSED [ 33%]
components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_factory_embedding_flag_skips_component_gauges_assert PASSED [ 50%]
components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_factory_default_is_chat_path PASSED [ 66%]
components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_seed_broadcasts_to_all_dp_ranks PASSED [ 83%]
components/src/dynamo/vllm/tests/test_vllm_publisher.py::test_unseeded_rank_reports_empty_to_the_kv_router PASSED [100%]

6 passed in 3.60s
```

Standalone driver producing the table above, before the fix:

```console
$ PYTHONPATH=components/src python dp_rank_repro.py
engine start: 16384 KV blocks / DP 2 -> per_rank_kv_blocks=8192
publisher.py under test: /repo/components/src/dynamo/vllm/publisher.py
factory retains: NoneType (main: single 'created_logger' field)

steady state: 3 scheduler iterations at kv_cache_usage=0.5
dp_rank  num_gpu_block  kv_used_blocks published    total_blocks gauge
0        1              [0, 0, 0]                   1.0
1        8192           [4096, 4096, 4096]          8192.0

expected per rank: kv_used_blocks=[4096, 4096, 4096], total_blocks=8192.0
RESULT: FAIL - a rank reports empty to the KV router
```

and with the fix:

```console
$ PYTHONPATH=components/src python dp_rank_repro.py
engine start: 16384 KV blocks / DP 2 -> per_rank_kv_blocks=8192
publisher.py under test: /repo/components/src/dynamo/vllm/publisher.py
factory retains: list (main: single 'created_logger' field)

steady state: 3 scheduler iterations at kv_cache_usage=0.5
dp_rank  num_gpu_block  kv_used_blocks published    total_blocks gauge
0        8192           [4096, 4096, 4096]          8192.0
1        8192           [4096, 4096, 4096]          8192.0

expected per rank: kv_used_blocks=[4096, 4096, 4096], total_blocks=8192.0
RESULT: PASS - every rank reports its real occupancy
```

<details>
<summary>dp_rank_repro.py</summary>

```python
import asyncio
import sys
from types import SimpleNamespace

from prometheus_client import CollectorRegistry
from vllm.v1.metrics.stats import SchedulerStats

import dynamo.vllm.publisher as pub
from dynamo.common.utils.prometheus import LLMBackendMetrics
from dynamo.vllm.capacity import per_rank_kv_blocks

TOTAL_KV_BLOCKS, DP_SIZE, USAGE, ITERATIONS = 16384, 2, 0.5, 3
published = []

class RecordingMetricsPublisher:
    async def create_endpoint(self, _endpoint):
        return None

    def publish(self, dp_rank, **kwargs):
        published.append((dp_rank, kwargs["kv_used_blocks"]))

def gauge(registry, suffix, dp_rank):
    for family in registry.collect():
        if family.name.endswith(suffix):
            for sample in family.samples:
                if sample.labels.get("dp_rank") == str(dp_rank):
                    return sample.value
    return None

async def main():
    pub.WorkerMetricsPublisher = RecordingMetricsPublisher
    registry = CollectorRegistry()
    gauges = LLMBackendMetrics(
        registry, model_name="Qwen/Qwen3-0.6B", component_name="backend"
    )
    factory = pub.StatLoggerFactory(endpoint=SimpleNamespace(), component_gauges=gauges)
    loggers = [factory(SimpleNamespace(), r) for r in range(DP_SIZE)]
    await asyncio.sleep(0)

    per_rank = per_rank_kv_blocks(TOTAL_KV_BLOCKS, DP_SIZE)
    print(f"engine start: {TOTAL_KV_BLOCKS} KV blocks / DP {DP_SIZE} "
          f"-> per_rank_kv_blocks={per_rank}")
    print(f"publisher.py under test: {pub.__file__}")
    print(f"factory retains: {type(getattr(factory, 'created_loggers', None)).__name__}"
          f" (main: single 'created_logger' field)")

    factory.set_num_gpu_blocks_all(per_rank or 0)
    factory.init_publish()
    published.clear()

    for _ in range(ITERATIONS):
        for logger in loggers:
            logger.record(SchedulerStats(kv_cache_usage=USAGE), None)

    print(f"\nsteady state: {ITERATIONS} scheduler iterations at kv_cache_usage={USAGE}")
    print(f"{'dp_rank':<9}{'num_gpu_block':<15}{'kv_used_blocks published':<28}"
          f"{'total_blocks gauge':<20}")
    ok = True
    for rank, logger in enumerate(loggers):
        seen = [used for r, used in published if r == rank]
        total = gauge(registry, "total_blocks", rank)
        print(f"{rank:<9}{logger.num_gpu_block:<15}{str(seen):<28}{str(total):<20}")
        if seen != [int(per_rank * USAGE)] * ITERATIONS or total != per_rank:
            ok = False

    print(f"\nexpected per rank: kv_used_blocks={[int(per_rank * USAGE)] * ITERATIONS}, "
          f"total_blocks={per_rank}.0")
    print("RESULT:", "PASS - every rank reports its real occupancy" if ok
          else "FAIL - a rank reports empty to the KV router")
    return 0 if ok else 1

sys.exit(asyncio.run(main()))
```

</details>
</details>

### DIFF
--- a/components/src/dynamo/vllm/publisher.py
+++ b/components/src/dynamo/vllm/publisher.py
@@ -141,7 +141,7 @@ class StatLoggerFactory:
         self.endpoint = endpoint
         self.component_gauges = component_gauges
         self.embedding_worker = embedding_worker
-        self.created_logger: Optional[DynamoStatLoggerPublisher] = None
+        self.created_loggers: list[DynamoStatLoggerPublisher] = []
 
     def create_stat_logger(self, dp_rank: int) -> StatLoggerBase:
         # Embedding workers have no KV cache and no scheduler stats worth
@@ -159,7 +159,7 @@ class StatLoggerFactory:
             dp_rank=dp_rank,
             component_gauges=self.component_gauges,
         )
-        self.created_logger = logger
+        self.created_loggers.append(logger)
 
         return logger
 
@@ -168,9 +168,9 @@ class StatLoggerFactory:
 
     # TODO Remove once we publish metadata to shared storage
     def set_num_gpu_blocks_all(self, num_blocks: int) -> None:
-        if self.created_logger:
-            self.created_logger.set_num_gpu_block(num_blocks)
+        for logger in self.created_loggers:
+            logger.set_num_gpu_block(num_blocks)
 
     def init_publish(self) -> None:
-        if self.created_logger:
-            self.created_logger.init_publish()
+        for logger in self.created_loggers:
+            logger.init_publish()
--- a/components/src/dynamo/vllm/tests/test_vllm_publisher.py
+++ b/components/src/dynamo/vllm/tests/test_vllm_publisher.py
@@ -63,7 +63,7 @@ def test_factory_returns_noop_logger_for_embedding_worker(monkeypatch):
     # downstream ``init_publish`` / ``set_num_gpu_blocks_all`` calls in
     # the chat path are safe no-ops if anyone ever wires them on the
     # embedding branch by mistake.
-    assert factory.created_logger is None
+    assert factory.created_loggers == []
 
 
 def test_noop_stat_logger_record_is_safe_with_none_stats():
@@ -125,3 +125,39 @@ def test_factory_default_is_chat_path(monkeypatch):
     assert constructed[0]["endpoint"] is endpoint
     assert constructed[0]["dp_rank"] == 3
     assert constructed[0]["component_gauges"] is component_gauges
+
+
+def test_seed_broadcasts_to_all_dp_ranks(monkeypatch):
+    """Regression for #12052: the pre-first-record seed must reach every
+    per-rank publisher, not just the last one created.
+
+    vLLM calls the factory once per data-parallel rank, and each returned
+    ``DynamoStatLoggerPublisher`` is a distinct object vLLM retains. If the
+    factory only remembers the last one, ``set_num_gpu_blocks_all`` /
+    ``init_publish`` seed the ``total_blocks`` and ``gpu_cache_usage_percent``
+    gauges for a single rank, so non-last ranks lack their zero-valued labels
+    until the first scheduler record arrives.
+    """
+    created = []
+
+    def _fake_publisher(*_args, **kwargs):
+        m = Mock(spec=DynamoStatLoggerPublisher)
+        m.dp_rank = kwargs.get("dp_rank")
+        created.append(m)
+        return m
+
+    monkeypatch.setattr(publisher_mod, "DynamoStatLoggerPublisher", _fake_publisher)
+
+    factory = StatLoggerFactory(
+        endpoint=SimpleNamespace(), component_gauges=SimpleNamespace()
+    )
+    factory.create_stat_logger(dp_rank=0)
+    factory.create_stat_logger(dp_rank=1)
+
+    factory.set_num_gpu_blocks_all(123)
+    factory.init_publish()
+
+    assert len(created) == 2
+    for publisher in created:
+        publisher.set_num_gpu_block.assert_called_once_with(123)
+        publisher.init_publish.assert_called_once_with()
--- a/components/src/dynamo/vllm/tests/test_vllm_publisher.py
+++ b/components/src/dynamo/vllm/tests/test_vllm_publisher.py
@@ -3,20 +3,23 @@
 
 """Unit tests for the vLLM stat-logger factory.
 
-These tests focus on the embedding-worker gating path: vLLM workers run
-either a chat/decode engine or a pooling (embedding) engine, and the
-chat-shaped Prometheus collectors are only meaningful on the former.
-The factory is the single seam where vLLM calls into dynamo per dp_rank,
-so it is also the seam where the embedding worker must short-circuit
-the chat-shaped pipeline.
+The factory is the single seam where vLLM calls into dynamo per dp_rank, which
+makes it the seam for two things: the embedding worker short-circuiting the
+chat-shaped pipeline (vLLM workers run either a chat/decode engine or a pooling
+engine, and the chat-shaped Prometheus collectors only mean something on the
+former), and the per-rank KV capacity seed reaching every rank the worker owns.
 """
 
+import asyncio
 from types import SimpleNamespace
 from unittest.mock import Mock
 
 import pytest
+from prometheus_client import CollectorRegistry
+from vllm.v1.metrics.stats import SchedulerStats
 
 import dynamo.vllm.publisher as publisher_mod
+from dynamo.common.utils.prometheus import LLMBackendMetrics
 from dynamo.vllm.publisher import (
     DynamoStatLoggerPublisher,
     NoopStatLogger,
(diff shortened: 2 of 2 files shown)


components/src/dynamo/vllm/publisher.py
```diff
@@ -141,7 +141,7 @@ def __init__(
         self.endpoint = endpoint
         self.component_gauges = component_gauges
         self.embedding_worker = embedding_worker
-        self.created_logger: Optional[DynamoStatLoggerPublisher] = None
+        self.created_loggers: list[DynamoStatLoggerPublisher] = []
```
jthomson04, line 144: Non-blocking: a dict keyed by `dp_rank` would be more defensive if the factory were ever called twice for one rank. Nothing does today, so the list is fine.

jthomson04 (approved):
Fix looks right, but the framing undersells it — this isn't startup-only.

`worker_factory.py:437` is the only caller of `set_num_gpu_block`, and `num_gpu_block` defaults to `1` (`publisher.py:35`). `record()` reads it every iteration (`publisher.py:64-68`), so on `main` a non-last DP rank stays at `1` for the worker's lifetime and publishes `int(1 * kv_cache_usage)` → `kv_used_blocks=0` permanently, with `total_blocks` pinned at 1. A rank that always looks empty is one the KV router keeps loading up.

Only the `init_publish()` half is a real pre-first-record seed. Hits hybrid/internal DP LB (`handlers.py:569-587`); external LB is unaffected.

Worth updating the PR body and #12052 — it's a routing-correctness fix.

components/src/dynamo/vllm/publisher.py
1fanwang: thanks for the pointer! I still kept the list after some testing: The factory cannot know which object vLLM retained, so keying by `dp_rank` would silently drop one: if vLLM kept the publisher from the first call and the dict kept the second, the retained one never gets seeded — this bug again. Seeding every created publisher is safe because both seed writes are idempotent (`set_num_gpu_block` sets a field, `init_publish` writes zeros for its own `dp_rank`).

1fanwang:
Thanks @jthomson04 — you were right, and the body now leads with the routing consequence rather than calling it startup-only.

PR #12165, opened 2026-07-25
feat(snapshot): agentInject agent-only checkpoint storage (POC, do not merge)

Ronkahn21:
## POC — do not merge

Proof-of-concept for an **agent-only** checkpoint storage mode (`accessMode: agentInject`) where the
workload pod never mounts the checkpoint PVC. The node agent owns the PVC and grafts the checkpoint
into the restore container's mount namespace via `open_tree(OPEN_TREE_CLONE)` + `move_mount`, so CRIU
can read it. Opened to exercise CI on the change; not intended to merge as-is.

### What it does
- New opt-in `accessMode: agentInject` (agent config + operator config + Helm).
- Operator skips injecting the checkpoint PVC into capture/restore pods (leaves `PVCName` empty) and
  stamps the resolved access mode; the agent treats the stamped value as authoritative.
- `nsrestore` gains `--checkpoint-fd`: the agent clones its own checkpoint dir into a detached mount
  (`open_tree`), passes the fd via `exec.Cmd.ExtraFiles`, and `nsrestore` grafts it with `move_mount`
  inside the container before CRIU restore.
- Capture is unchanged from `agentMount` (agent writes via its own mount).

### Validation done outside CI
- Cross-namespace mechanism proven on a live kernel: plain `/proc/self/fd` bind is `EINVAL`
  cross-ns; `open_tree` + `move_mount` works.
- Full real-CRIU run: checkpoint a live process, graft via `open_tree`+`move_mount`, `criu restore`
  from the grafted mount, process resumes.
- Both Go modules build; operator unit tests pass.

### Not covered here
- Full GPU/k8s e2e (agent ↔ PodSnapshotContent, CUDA-checkpoint) runs on the cluster, not in this PR.

See `plans/snapshot-migration/agent-inject-cross-ns-explainer.md` for the kernel-level writeup.

## Summary by CodeRabbit

* **New Features**
  * Added `agentInject` checkpoint storage mode for PVC-backed snapshots.
  * Restore operations can now access checkpoint data without mounting the PVC into workload containers.
  * Added support for restoring checkpoints through inherited file descriptors.

* **Bug Fixes**
  * PVC names are no longer required when using `agentInject`.
  * Updated storage validation and workload preparation to handle the new mode correctly.

* **Documentation**
  * Added configuration guidance for `agentInject` behavior and setup.

### DIFF
--- a/deploy/snapshot/internal/executor/nsrestore.go
+++ b/deploy/snapshot/internal/executor/nsrestore.go
@@ -21,6 +21,11 @@ type RestoreOptions struct {
 	CUDADeviceMap  string
 	CgroupRoot     string
 	TargetPodIP    string
+	// CheckpointFD, when >= 0, is an inherited fd of a detached checkpoint mount
+	// (agentInject mode). nsrestore grafts it at CheckpointPath inside this
+	// container's mount namespace before reading the checkpoint. -1 means the
+	// checkpoint is already visible (agentMount/podMount).
+	CheckpointFD int
 }
 
 type RestoreInNamespaceResult struct {
@@ -38,7 +43,23 @@ func RestoreInNamespace(ctx context.Context, opts RestoreOptions, log logr.Logge
 		"has_cuda_map", opts.CUDADeviceMap != "",
 		"cgroup_root", opts.CgroupRoot,
 		"target_pod_ip_present", opts.TargetPodIP != "",
+		"checkpoint_fd", opts.CheckpointFD,
 	)
+
+	// agentInject: graft the agent-supplied detached checkpoint mount into this
+	// container's mount namespace at CheckpointPath before anything reads it. This
+	// must precede ReadManifest/ApplyRootfsDiff, which both read the checkpoint dir.
+	if opts.CheckpointFD >= 0 {
+		detach, err := snapshotruntime.AttachCheckpointTree(opts.CheckpointFD, opts.CheckpointPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to graft checkpoint mount: %w", err)
+		}
+		defer func() {
+			if err := detach(); err != nil {
+				log.Error(err, "Failed to detach grafted checkpoint mount")
+			}
+		}()
+	}
 
 	manifestReadStart := time.Now()
 	m, err := types.ReadManifest(opts.CheckpointPath)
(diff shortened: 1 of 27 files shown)


deploy/snapshot/internal/executor/nsrestore.go
```diff
@@ -38,8 +43,24 @@ func RestoreInNamespace(ctx context.Context, opts RestoreOptions, log logr.Logge
...
+	// container's mount namespace at CheckpointPath before anything reads it. This
+	// must precede ReadManifest/ApplyRootfsDiff, which both read the checkpoint dir.
+	if opts.CheckpointFD >= 0 {
+		detach, err := snapshotruntime.AttachCheckpointTree(opts.CheckpointFD, opts.CheckpointPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to graft checkpoint mount: %w", err)
+		}
+		defer func() {
+			if err := detach(); err != nil {
+				log.Error(err, "Failed to detach grafted checkpoint mount")
```
julienmancuso, line 59: we should fail the restore if cleanup fails

deploy/snapshot/internal/runtime/nsmount.go
```diff
@@ -0,0 +1,210 @@
...
+		"",
+		unix.MOVE_MOUNT_F_EMPTY_PATH|unix.MOVE_MOUNT_T_EMPTY_PATH,
+	); err != nil {
+		return nil, fmt.Errorf("move_mount -> %s: %w", target, err)
+	}
+	return func() error {
+		// UMOUNT_NOFOLLOW prevents final-component symlink substitution during
+		// cleanup. MNT_DETACH removes the graft from pathname lookup
+		// immediately even if CRIU or the workload still holds open files.
+		if err := unix.Unmount(target, unix.MNT_DETACH|unix.UMOUNT_NOFOLLOW); err != nil {
```
julienmancuso, line 205: This pins the target for attachment, but cleanup later unmounts by pathname. The restored workload can rename an ancestor before cleanup, causing detachment to miss the injected mount. UMOUNT_NOFOLLOW only prevents dereferencing a final symlink, it does not pin the path...
The cleanuo should be tied to a stable mount identity or suspend the workload through detach

PR #12166, opened 2026-07-25
fix(operator): support vLLM multinode with GPU DRA

### LINKED ISSUE #12164: [BUG]: vLLM multi-node silently skips distributed launch setup for pure GPU DRA ResourceClaims
## Describe the Bug

cc @negz @julienmancuso @tmonty12

A vLLM `DynamoGraphDeployment` can pass Kubernetes Dynamic Resource Allocation (DRA) `ResourceClaims` through to its pods, and pure GPU DRA works for a single-node vLLM component. However, when the same GPU allocation mechanism is used for a standard multi-node vLLM TP/PP or DP component, the operator cannot determine the number of GPUs per node.

The multi-node launch-selection code reads GPU counts exclusively from scalar `resources.requests` / `resources.limits` entries such as `nvidia.com/gpu`, custom `*/gpu` resources, or MIG resources. It does not resolve or count devices requested by `resources.claims` / `spec.resourceClaims` and their referenced `ResourceClaimTemplate`.

Consequently, a pure-DRA container produces `containerGPUs == 0`. Both normal multi-node detection functions return `false`, and the operator silently skips the role-specific MP/Ray or hybrid-DP launch configuration. The generated leader and worker pods retain the same original vLLM command instead of receiving the required `--nnodes`, `--node-rank`, `--master-addr`, `--headless`, or DP coordination flags.

This appears to be a missing integration between generic ResourceClaim passthrough and the independently developed vLLM multi-node auto-configuration path. Even if pure-DRA vLLM multi-node is considered unsupported today, accepting the manifest and rendering commands that cannot form the requested distributed engine is a bug; the operator should at minimum reject the combination with a clear validation error.

There are two distinct layers here:

1. **DRA allocation:** Kubernetes allocates the requested devices and exposes them in each pod.
2. **vLLM distributed launch configuration:** DGD must render different commands for the leader and workers.

Manual vLLM flags cannot repair an unallocated/Pending ResourceClaim. This issue concerns the second layer when the claims are valid and allocated.

## Steps to Reproduce

The examples use two GPUs total and TP=2 because that makes the decision easy to inspect. Replace the image as appropriate for the environment.

### Case A: single-node pure DRA control case

One pod receives two GPUs from DRA. No cross-node command rewriting is required.

```yaml
apiVersion: resource.k8s.io/v1
kind: ResourceClaimTemplate
metadata:
  name: two-gpus
spec:
  spec:
    devices:
      requests:
      - name: gpus
        exactly:
          deviceClassName: gpu.nvidia.com
          count: 2
---
kind: DynamoGraphDeployment
metadata:
  name: vllm-single-dra
spec:
  backendFramework: vllm
  components:
  - name: frontend
    type: frontend
    replicas: 1
    podTemplate:
      spec:
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest

  - name: agg
    type: worker
    replicas: 1
    podTemplate:
      spec:
        resourceClaims:
        - name: gpu
          resourceClaimTemplateName: two-gpus
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest
          command: [python3, -m, dynamo.vllm]
          args:
          - --model
          - Qwen/Qwen3-0.6B
          - --tensor-parallel-size
          - "2"
          resources:
            claims:
            - name: gpu
```

The command remains local, as expected:

```console
python3 -m dynamo.vllm \
  --model Qwen/Qwen3-0.6B \
  --tensor-parallel-size 2
```

The DRA driver exposes both devices to the one pod, so vLLM can create its local TP=2 workers.

### Case B: multi-node pure DRA

Each pod receives one GPU from DRA, while TP=2 is intended to span the two nodes.

```yaml
apiVersion: resource.k8s.io/v1
kind: ResourceClaimTemplate
metadata:
  name: one-gpu
spec:
  spec:
    devices:
      requests:
      - name: gpu
        exactly:
          deviceClassName: gpu.nvidia.com
          count: 1
---
kind: DynamoGraphDeployment
metadata:
  name: vllm-multinode-dra
  annotations:
    # This selects MP only after distributed launch has been detected.
    # It does not bypass the zero-GPU detection described below.
    nvidia.com/vllm-distributed-executor-backend: mp
spec:
  backendFramework: vllm
  components:
  - name: frontend
    type: frontend
    replicas: 1
    podTemplate:
      spec:
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest

  - name: agg
    type: worker
    replicas: 1
    multinode:
      nodeCount: 2
    podTemplate:
      spec:
        resourceClaims:
        - name: gpu
          resourceClaimTemplateName: one-gpu
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest
          command: [python3, -m, dynamo.vllm]
          args:
          - --model
          - Qwen/Qwen3-0.6B
          - --tensor-parallel-size
          - "2"
          resources:
            claims:
            - name: gpu
```

For the vLLM backend, the effective input is equivalent to:

```text
resources.requests[GPU] = absent
resources.limits[GPU]   = absent
resources.claims        = [{name: gpu}]

getContainerGPUs() = 0
```

Therefore:

```text
needsTensorParallelMultinodeLaunch = false
needsDataParallelMultinodeLaunch   = false
```

The MP annotation does not help because `shouldUseMpBackend()` is only consulted after `needsTensorParallelMultinodeLaunch()` returns true.

The leader and worker consequently retain equivalent commands:

```console
# Leader: no multi-node flags injected
python3 -m dynamo.vllm \
  --model Qwen/Qwen3-0.6B \
  --tensor-parallel-size 2

# Worker: also no role-specific multi-node flags injected
python3 -m dynamo.vllm \
  --model Qwen/Qwen3-0.6B \
  --tensor-parallel-size 2
```

Each pod has one allocated GPU but independently tries to start TP=2. The worker is not rendered as a headless rank, and neither pod receives rendezvous information.

### Case C: scalar GPU resource comparison

The equivalent scalar-resource manifest exposes the count to the backend:

```yaml
kind: DynamoGraphDeployment
metadata:
  name: vllm-multinode-scalar
  annotations:
    nvidia.com/vllm-distributed-executor-backend: mp
spec:
  backendFramework: vllm
  components:
  - name: frontend
    type: frontend
    replicas: 1
    podTemplate:
      spec:
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest

  - name: agg
    type: worker
    replicas: 1
    multinode:
      nodeCount: 2
    podTemplate:
      spec:
        containers:
        - name: main
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:latest
          command: [python3, -m, dynamo.vllm]
          args:
          - --model
          - Qwen/Qwen3-0.6B
          - --tensor-parallel-size
          - "2"
          resources:
            limits:
              nvidia.com/gpu: "1"
```

Now the operator computes:

```text
containerGPUs = 1
worldSize     = TP * PP = 2 * 1 = 2
2 > 1         = distributed launch required
```

The rendered leader command is conceptually:

```console
python3 -m dynamo.vllm \
  --model Qwen/Qwen3-0.6B \
  --tensor-parallel-size 2 \
  --distributed-executor-backend mp \
  --nnodes 2 \
  --master-addr <leader-hostname> \
  --master-port 29500 \
  --node-rank 0
```

The Grove worker command is conceptually:

```console
/bin/sh -c 'exec python3 -m dynamo.vllm \
  --model Qwen/Qwen3-0.6B \
  --tensor-parallel-size 2 \
  --distributed-executor-backend mp \
  --nnodes 2 \
  --master-addr <leader-hostname> \
  --master-port 29500 \
  --node-rank $((GROVE_PCLQ_POD_INDEX + 1)) \
  --headless'
```

For the only worker in a two-node component, the resulting rank is 1.

### Inspect the rendered pods

```bash
kubectl get pods \
  -l nvidia.com/dynamo-graph-deployment-name=vllm-multinode-dra
```

```bash
kubectl get pods \
  -l nvidia.com/dynamo-graph-deployment-name=vllm-multinode-dra \
  -o json |
jq -r '
  .items[] |
  .metadata.name as $pod |
  .spec.containers[] |
  select(.name == "main") |
  [
    $pod,
    "  command: " + ((.command // []) | join(" ")),
    "  args:    " + ((.args // []) | join(" ")),
    "  limits:  " + ((.resources.limits // {}) | tostring),
    "  claims:  " + ((.resources.claims // []) | tostring)
  ] | join("\n")
'
```

The DRA side can be inspected independently:

```bash
kubectl get resourceclaimtemplates
kubectl get resourceclaims
kubectl describe resourceclaimtemplate one-gpu
kubectl describe resourceclaim <generated-claim-name>
```

## Expected Behavior

For an allocated pure GPU DRA claim, one of the following should happen:

1. **Preferred:** DGD determines the per-node accelerator count from the referenced `ResourceClaim` / `ResourceClaimTemplate` and renders the same correct role-specific multi-node commands that it renders for scalar GPU resources.
2. **Explicit topology input:** DGD accepts a backend-independent explicit field such as `gpusPerNode` / `acceleratorsPerNode` and uses that value for launch planning while Kubernetes DRA remains responsible for allocation.
3. **Minimum safe behavior:** admission or reconciliation rejects standard vLLM multi-node TP/PP/DP with pure ResourceClaims when the GPU count cannot be established, explaining that distributed launch auto-configuration requires an explicit per-node device count.

An explicit `nvidia.com/vllm-distributed-executor-backend: mp|ray` annotation should not result in silently identical leader and worker commands.

## Actual Behavior

The operator accepts the DGD and passes the DRA claims into the generated pod specs, but standard TP/PP and DP multi-node detection sees zero GPUs and falls through to:

```text
No need to inject tensor or data parallel flags for multinode deployments
```

No distributed launch flags are injected. Leader and worker commands remain effectively identical.

The special elastic-EP path is an exception because it is selected directly by `--enable-elastic-ep` without consulting this GPU-count check. This issue concerns the normal multi-node TP/PP and hybrid-DP paths.

## Code Analysis

At current `main` (`daf44c6bf194ae7c56961b14f6cfa7edfe25af28`):

1. [`updateVLLMMultinodeArgs()`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/backend_vllm.go#L267-L295) dispatches to MP, Ray, elastic EP, hybrid DP, or no-op.
2. [`needsTensorParallelMultinodeLaunch()`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/backend_vllm.go#L534-L542) returns `false` when `containerGPUs == 0`.
3. [`needsDataParallelMultinodeLaunch()`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/backend_vllm.go#L550-L558) has the same zero-GPU short-circuit.
4. [`getGPUQuantity()` / `resourceListGPUValue()`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/backend_vllm.go#L574-L610) inspect only scalar `Requests` and `Limits`. They do not inspect or resolve `ResourceRequirements.Claims`.
5. [`resourceRequirementsWithFallback()`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/backend_vllm.go#L613-L623) even preserves fallback `Claims`, but no subsequent GPU-count function consumes them.
6. The [`MultinodeSpec`](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/api/v1beta1/common.go#L71-L78) documentation says: `Total GPUs used is nodeCount * container GPU request`, reflecting the original scalar-resource assumption.

A simple `len(resources.Claims)` fix would not be correct. A single ResourceClaim may request zero, one, or many devices; may contain several device requests; and may be unrelated to GPUs. The count and DeviceClass live in the referenced `ResourceClaim` or `ResourceClaimTemplate`, not in `corev1.ResourceRequirements.Claims`.

### Why operator-managed GMS DRA can work

GMS DRA does not disprove the issue. In `GenerateBasePodSpec`, backend-specific command mutation happens before the GMS path replaces the scalar GPU resource with a DRA claim:

1. [`backend.UpdateContainer()` runs first](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/graph.go#L1521-L1534), while the original scalar GPU count is still visible.
2. [`dra.ApplyClaim()` runs later](https://github.com/ai-dynamo/dynamo/blob/daf44c6bf194ae7c56961b14f6cfa7edfe25af28/deploy/operator/internal/dynamo/graph.go#L1600-L1614) and replaces that scalar resource with the generated DRA claim.

Pure user-supplied DRA has no temporary scalar GPU count for the backend to observe.

## Relevant Git History

The history suggests two independently developed paths that were never joined:

- [#2269](https://github.com/ai-dynamo/dynamo/pull/2269), **Julien Mancuso (`@julienmancuso`)**, added Grove multi-node support in August 2025. The vLLM launch path was based on scalar container GPU resources.
- [#3595](https://github.com/ai-dynamo/dynamo/pull/3595), **Thomas Montfort (`@tmonty12`)**, added vLLM data-parallel multi-node support in October 2025, again deriving local DP layout from the scalar GPU count.
- [#3446](https://github.com/ai-dynamo/dynamo/pull/3446), **Rohan Varma (`@nvrohanv`)**, added generic ResourceClaim passthrough as part of an MPI-flow fix in October 2025. Its ResourceClaim test also retained a scalar `GPU: "4"`; it did not establish pure-DRA vLLM multi-node behavior.
- [#4408](https://github.com/ai-dynamo/dynamo/pull/4408), **Thomas Montfort (`@tmonty12`)**, added custom GPU types in November 2025. This introduced/retained the explicit `containerGPUs == 0 => false` behavior while supporting more scalar GPU resource names, but did not resolve claims.
- [#5006](https://github.com/ai-dynamo/dynamo/pull/5006), **Matej Kosec (`@MatejKosec`)**, fixed vLLM multi-node TP and DP behavior in December 2025, still using the scalar GPU-count model.
- [#7777](https://github.com/ai-dynamo/dynamo/pull/7777), **Julien Mancuso (`@julienmancuso`)**, added inter-pod GMS in April 2026. Its DRA path retains the scalar count until after backend command generation, as described above.
- [#9253](https://github.com/ai-dynamo/dynamo/pull/9253), **Yao Qing (`@yao531441`)**, added vLLM XPU DRA examples in July 2026. The pure-DRA examples are single-node, so they do not exercise the multi-node launch decision.

A search for issues and PRs containing combinations of `vLLM`, `DRA`, `multinode`, `ResourceClaim`, and `getContainerGPUs` did not find an existing issue specifically tracking this interaction.

## Manual Workaround and Why Flags Alone Are Insufficient

Setting only:

```yaml
metadata:
  annotations:
    nvidia.com/vllm-distributed-executor-backend: mp
```

or adding only:

```text
--distributed-executor-backend mp
--nnodes 2
```

is insufficient. The DGD provides one user pod template, while leader and worker need different `--node-rank` values, and workers additionally need `--headless`. The operator normally generates those differences.

For Grove, a diagnostic workaround can use its injected role and rank environment variables to implement the missing role-specific entrypoint manually:

```yaml
command:
- /bin/sh
- -ec
args:
- |
  MASTER_ADDR="${GROVE_PCSG_NAME}-${GROVE_PCSG_INDEX}-agg-ldr-0.${GROVE_HEADLESS_SERVICE}"

  case "${GROVE_PCLQ_NAME}" in
    *-ldr)
      echo "Starting vLLM leader, master=${MASTER_ADDR}:29500"
      exec python3 -m dynamo.vllm \
        --model Qwen/Qwen3-0.6B \
        --tensor-parallel-size 2 \
        --distributed-executor-backend mp \
        --nnodes 2 \
        --master-addr "${MASTER_ADDR}" \
        --master-port 29500 \
        --node-rank 0
      ;;

    *-wkr)
      NODE_RANK=$((GROVE_PCLQ_POD_INDEX + 1))
      echo "Starting vLLM worker, rank=${NODE_RANK}, master=${MASTER_ADDR}:29500"
      exec python3 -m dynamo.vllm \
        --model Qwen/Qwen3-0.6B \
        --tensor-parallel-size 2 \
        --distributed-executor-backend mp \
        --nnodes 2 \
        --master-addr "${MASTER_ADDR}" \
        --master-port 29500 \
        --node-rank "${NODE_RANK}" \
        --headless
      ;;

    *)
      echo "Unknown Grove clique role: ${GROVE_PCLQ_NAME}" >&2
      exit 1
      ;;
  esac
```

Because the rendered command contains the `--distributed-executor-backend mp` pair, the existing vLLM `UpdatePodSpec` detection should still inject the `wait-for-leader-mp` init container for workers. This is useful for validating the diagnosis but is not a desirable long-term API contract; it duplicates backend-specific operator behavior and relies on Grove implementation details.

Adding a fake `nvidia.com/gpu` request alongside the DRA claim is not a safe workaround because it can cause two independent GPU allocations.

## Suggested Fix Direction

Potential approaches, ordered from most explicit to most automatic:

1. **Add an explicit per-node accelerator count to multi-node configuration.** This cleanly separates launch topology from Kubernetes allocation and works for DRA, scalar extended resources, and future device mechanisms.
2. **Resolve referenced ResourceClaims/ResourceClaimTemplates during reconciliation.** Count only compatible GPU/XPU device requests, account for `ExactCount` and named requests, and return a clear error for ambiguous/mixed claims.
3. **Use operator-created metadata when the operator created the claim.** GMS already knows the original GPU count; this could be generalized, but it would not solve arbitrary user-provided claims.
4. **At minimum, validate and reject.** If `multinode.nodeCount > 1`, TP/PP/DP implies cross-node execution, scalar GPU count is zero, and only claims are present, reject the DGD with an actionable message rather than silently rendering identical commands.

Suggested tests:

- Multi-node TP with one GPU per pod supplied exclusively by a GPU `ResourceClaimTemplate` renders MP/Ray leader and worker commands.
- Multi-node DP with pure DRA computes the correct `--data-parallel-size-local` and start ranks.
- A non-GPU ResourceClaim does not count as a GPU.
- Multiple device requests and non-`ExactCount` requests are handled or rejected explicitly.
- An unresolved/missing ResourceClaimTemplate produces an actionable reconciliation/admission error.
- Single-node pure DRA remains unchanged.
- Scalar GPU resources and GMS-generated DRA remain unchanged.
- Explicit MP/Ray annotations do not silently no-op when the distributed topology cannot be determined.

## Environment

- Repository: `ai-dynamo/dynamo`
- Analyzed revision: `daf44c6bf194ae7c56961b14f6cfa7edfe25af28` (current `main` at the time of analysis)
- Backend: vLLM
- Deployment path: Kubernetes DGD, Grove multi-node
- DRA API shape: `resource.k8s.io/v1` `ResourceClaimTemplate`
- DeviceClass examples: `gpu.nvidia.com`; the same structural issue applies to other accelerator DeviceClasses such as `gpu.intel.com`

This report is based on the deterministic operator rendering path plus the reported observation that pure DRA works for single-node but not standard vLLM multi-node. Runtime-specific logs can be added from a concrete cluster reproduction, but the zero-GPU fallthrough is directly visible in the current code.

## Backend scope update

A follow-up report clarified the backend scope:

- **vLLM:** remains the confirmed affected backend and the scope of this issue and PR #12166.
- **SGLang:** multi-node with pure GPU DRA was tested successfully and is not affected by the reported behavior.
- **TensorRT-LLM:** may have a similar integration gap, but this has not been verified yet. Treat it as a follow-up investigation rather than confirmed scope.

sttts:
## Summary

* Resolve a component's per-node GPU count from user-supplied `ResourceClaim` and `ResourceClaimTemplate` objects when no scalar GPU resource is present.
* Pass the lazy, memoized count through both Grove and LeaderWorkerSet rendering. vLLM consumes it for multinode MP/Ray and hybrid-DP launch flags; TensorRT-LLM consumes it for the multinode leader's `mpirun` world size.
* Preserve pure DRA allocation: the operator reads claim metadata for launch planning but never adds a synthetic `nvidia.com/gpu` request or limit.
* Watch referenced `ResourceClaim` and `ResourceClaimTemplate` objects and relevant cluster-scoped `DeviceClass` objects so dependency changes re-render existing workloads.
* Add the required ResourceClaim and DeviceClass read/watch RBAC and regenerate the operator and Helm RBAC manifests.
* Reject ambiguous or non-deterministic device requests instead of silently rendering broken distributed launch flags.

Fixes ai-dynamo/dynamo#12164

## Root cause

The multinode backend logic derived the per-container GPU count only from scalar entries in `resources.requests` and `resources.limits`. A pure-DRA container has `resources.claims` instead, so vLLM saw a count of zero and skipped the normal multinode launch setup. Leader and worker could consequently receive effectively identical commands.

Operator-managed GMS DRA did not expose this gap because its scalar GPU count remains present while backend arguments are rendered and is converted to a claim afterward. User-supplied DRA has no temporary scalar count.

## Implementation

`dra.ResolveGPUCount` keeps scalar resources authoritative for compatibility. When no scalar GPU is present, it maps the main container's claim references to the pod's `resourceClaims` and reads the referenced `resource.k8s.io/v1` `ResourceClaim` or `ResourceClaimTemplate`.

The resolver:

* supports deterministic `ExactCount` GPU requests, including Kubernetes' default count of one;
* supports `firstAvailable` when every alternative is a GPU class and all alternatives request the same count;
* respects the optional container request selector and deduplicates repeated claim/request references;
* ignores requests that are deterministically non-GPU;
* rejects `allocationMode: All`, mixed GPU/non-GPU alternatives, differing alternative counts, missing request names, missing claim objects, and unclassifiable DeviceClasses with actionable reconciliation errors;
* classifies DeviceClasses by `extendedResourceName`, unambiguous conjunctive `device.driver == ...` selectors, or a conventional GPU class name only when no selectors are present.

Rendering passes a `ContainerGPUCount` closure rather than eagerly resolving every component. `sync.OnceValues` shares the result and error across all roles of one component:

* multinode vLLM resolves it when producing distributed launch flags;
* multinode TensorRT-LLM resolves it only for the leader;
* SGLang, Noop, and single-node vLLM do not resolve it because they do not consume the count.

DCD and DGD controllers watch DRA dependencies. Claim/template events enqueue only multinode components that reference the changed object; DeviceClass events enqueue relevant DRA-backed multinode workloads. DGD dependency watches are limited to the Grove pathway because LWS rendering is owned by the DCD controller.

For two nodes with one DRA GPU each and `--tensor-parallel-size 2`, vLLM now enters the same MP/Ray launch path as two scalar-GPU pods. With MP selected, the leader receives `--nnodes 2`, `--node-rank 0`, and rendezvous flags; the worker receives its computed nonzero rank plus `--headless`. Hybrid-DP similarly derives `--data-parallel-size-local` and the per-node start rank from the resolved count.

## RBAC

The generated operator and Helm roles include:

* `get/list/watch` for `resourceclaims`;
* `get/list/watch` for cluster-scoped `deviceclasses`;
* `get/list/watch` for `resourceclaimtemplates` in addition to the existing lifecycle permissions required by operator-managed templates.

## Validation

After rebasing onto `e523f7ceb9`:

* `go test ./internal/dra ./internal/dynamo`
* `go test ./internal/controller -run '^(TestRenderMultinodePodTemplateSpecs_VLLMMultinodeDRA|TestGenerateWorkerPodTemplateSpecDoesNotRequireGPUResource|TestDynamoComponentDeploymentReconciler_.*WorkloadComponentType)$'`
* `git diff --check`

Coverage includes direct ResourceClaim and ResourceClaimTemplate resolution, request filtering and deduplication, default counts, DeviceClass classification, ambiguous request rejection, scalar precedence, lazy backend consumption, dependency event mapping, Grove TP/MP rendering, LeaderWorkerSet TP/MP rendering, TensorRT-LLM world-size calculation, and hybrid-DP flag injection.

## Summary by CodeRabbit

* **New Features**
  * Added support for GPU-aware multinode vLLM workloads using Device Resource Allocation (DRA).
  * Preserved resource claims while generating leader and worker pod configurations.
  * Added support for resolving GPU counts from resource claims when scalar GPU resources are unavailable.

* **Bug Fixes**
  * Improved multinode launch and parallelism configuration when GPUs are provisioned through DRA.

### DIFF
--- a/deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
@@ -30,6 +30,7 @@ import (
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	resourcev1 "k8s.io/api/resource/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"emperror.dev/errors"
@@ -53,6 +54,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
@@ -814,8 +816,10 @@ func (r *DynamoComponentDeploymentReconciler) generateDeployment(ctx context.Con
 		},
 	}
 
+	renderer := r.workloadRenderer()
+	containerGPUs := renderer.containerGPUCount(ctx, opt.dynamoComponentDeployment)
 	// nolint: gosimple
-	podTemplateSpec, err := r.workloadRenderer().generatePodTemplateSpec(ctx, opt.dynamoComponentDeployment, dynamo.RoleMain)
+	podTemplateSpec, err := renderer.generatePodTemplateSpec(ctx, opt.dynamoComponentDeployment, dynamo.RoleMain, containerGPUs)
 	if err != nil {
 		return
 	}
@@ -908,7 +912,23 @@ func (r *DynamoComponentDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
 		})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&networkingv1.Ingress{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		WithEventFilter(commonController.EphemeralDeploymentEventFilter(r.Config, r.RuntimeConfig))
+		WithEventFilter(deploymentEventFilter(r.Config, r.RuntimeConfig))
+
+	if r.RuntimeConfig.Gate.Enabled(features.DRA) {
+		m = m.Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimToDCDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).Watches(
+			&resourcev1.ResourceClaimTemplate{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimTemplateToDCDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).Watches(
+			&resourcev1.DeviceClass{},
+			handler.EnqueueRequestsFromMapFunc(r.mapDeviceClassToDCDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		)
+	}
 
 	if r.RuntimeConfig.Gate.Enabled(features.LWS) {
 		m.Owns(&leaderworkersetv1.LeaderWorkerSet{}, builder.WithPredicates(predicate.Funcs{
--- a/deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
@@ -38,6 +38,7 @@ import (
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/common"
 	commonconsts "github.com/ai-dynamo/dynamo/deploy/operator/internal/consts"
 	commonController "github.com/ai-dynamo/dynamo/deploy/operator/internal/controller_common"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/dra"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/dynamo"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/gms"
@@ -99,6 +100,7 @@ type DynamoComponentDeploymentReconciler struct {
 
 // +kubebuilder:rbac:groups=scheduling.volcano.sh,resources=podgroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=resource.k8s.io,resources=resourceclaims;resourceclaimtemplates,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -928,14 +930,7 @@ func (r *DynamoComponentDeploymentReconciler) generatePodTemplateSpec(ctx contex
 	podAnnotations := dynamo.GetDCDKubeAnnotations(dcd)
 	kubeName := dcd.Name
 
-	// Convert user-provided metrics annotation into controller-managed label
-	// By default (no annotation), metrics are enabled
-	if podAnnotations[commonconsts.KubeAnnotationEnableMetrics] == commonconsts.KubeLabelValueFalse {
-		// Explicitly disabled, don't add the label
-	} else {
-		// Any other value (including empty) enables metrics
-		podLabels[commonconsts.KubeLabelMetricsEnabled] = commonconsts.KubeLabelValueTrue
-	}
+	applyMetricsLabel(podLabels, podAnnotations)
 
 	if parentName := dcd.GetLabels()[commonconsts.KubeLabelDynamoGraphDeploymentName]; parentName != "" {
 		podLabels[commonconsts.KubeLabelDynamoGraphDeploymentName] = parentName
@@ -974,6 +969,11 @@ func (r *DynamoComponentDeploymentReconciler) generatePodTemplateSpec(ctx contex
 		}
 	}
 
+	gpusPerNode, err := r.resolveGPUsPerNode(ctx, dcd, component)
+	if err != nil {
+		return nil, err
+	}
+
 	podSpec, err := dynamo.GenerateBasePodSpecForController(
 		dcd,
 		r.DockerSecretRetriever,
@@ -983,6 +983,7 @@ func (r *DynamoComponentDeploymentReconciler) generatePodTemplateSpec(ctx contex
 		checkpointInfo,
 		dynamo.GenerateBasePodSpecForControllerOptions{
 			WorkloadComponentType: nvidiacomv1beta1.ComponentType(componentType),
+			GPUsPerNode:           gpusPerNode,
 		},
 	)
 	if err != nil {
@@ -1056,6 +1057,43 @@ func (r *DynamoComponentDeploymentReconciler) generatePodTemplateSpec(ctx contex
 		},
 		Spec: *podSpec,
 	}, nil
+}
+
+func applyMetricsLabel(podLabels, podAnnotations map[string]string) {
+	// Metrics are enabled unless the pod annotation explicitly disables them.
+	if podAnnotations[commonconsts.KubeAnnotationEnableMetrics] != commonconsts.KubeLabelValueFalse {
+		podLabels[commonconsts.KubeLabelMetricsEnabled] = commonconsts.KubeLabelValueTrue
+	}
+}
+
+func (r *DynamoComponentDeploymentReconciler) resolveGPUsPerNode(
+	ctx context.Context,
+	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
+	component *nvidiacomv1beta1.DynamoComponentDeploymentSharedSpec,
+) (int64, error) {
+	backendFramework, err := dynamo.GetBackendFrameworkFromDynamoComponent(dcd)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to determine backend framework")
+	}
+	if backendFramework != dynamo.BackendFrameworkVLLM || component.GetNumberOfNodes() <= 1 {
+		return 0, nil
+	}
+
+	claimPodSpec := &corev1.PodSpec{}
+	if component.PodTemplate != nil {
+		claimPodSpec = &component.PodTemplate.Spec
+	}
+	gpuCount, err := dra.ResolveGPUCount(
+		ctx,
+		r.Client,
+		dcd.Namespace,
+		claimPodSpec,
+		dynamo.GetMainContainerResources(component),
+	)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to resolve GPUs per node")
+	}
+	return int64(gpuCount), nil
 }
 
 func (r *DynamoComponentDeploymentReconciler) generateService(ctx context.Context, opt generateResourceOption) (*corev1.Service, bool, error) {
--- a/deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
+++ b/deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"sync"
 
 	"emperror.dev/errors"
 	configv1alpha1 "github.com/ai-dynamo/dynamo/deploy/operator/api/config/v1alpha1"
@@ -86,17 +87,18 @@ func (r *dcdWorkloadRenderer) renderMultinodePodTemplateSpecs(
 	if err != nil {
 		return nil, nil, err
 	}
+	containerGPUs := r.containerGPUCount(ctx, dcd)
 
 	leaderLabels := make(map[string]string, len(podLabels))
 	maps.Copy(leaderLabels, podLabels)
-	leaderPodTemplateSpec, err := r.generateLeaderPodTemplateSpec(ctx, dcd, leaderLabels)
+	leaderPodTemplateSpec, err := r.generateLeaderPodTemplateSpec(ctx, dcd, leaderLabels, containerGPUs)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	workerLabels := make(map[string]string, len(podLabels))
 	maps.Copy(workerLabels, podLabels)
-	workerPodTemplateSpec, err := r.generateWorkerPodTemplateSpec(ctx, dcd, workerLabels)
+	workerPodTemplateSpec, err := r.generateWorkerPodTemplateSpec(ctx, dcd, workerLabels, containerGPUs)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -104,12 +106,22 @@ func (r *dcdWorkloadRenderer) renderMultinodePodTemplateSpecs(
 	return leaderPodTemplateSpec, workerPodTemplateSpec, nil
 }
 
+func (r *dcdWorkloadRenderer) containerGPUCount(
+	ctx context.Context,
+	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
+) dynamo.ContainerGPUCount {
+	return sync.OnceValues(func() (int64, error) {
+		return dynamo.ResolveContainerGPUs(ctx, r.reader, dcd.Namespace, &dcd.Spec.DynamoComponentDeploymentSharedSpec)
+	})
+}
+
 func (r *dcdWorkloadRenderer) generateLeaderPodTemplateSpec(
 	ctx context.Context,
 	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
 	labels map[string]string,
+	containerGPUs dynamo.ContainerGPUCount,
 ) (*corev1.PodTemplateSpec, error) {
-	leaderPodTemplateSpec, err := r.generatePodTemplateSpec(ctx, dcd, dynamo.RoleLeader)
+	leaderPodTemplateSpec, err := r.generatePodTemplateSpec(ctx, dcd, dynamo.RoleLeader, containerGPUs)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate leader pod template")
 	}
@@ -129,8 +141,9 @@ func (r *dcdWorkloadRenderer) generateWorkerPodTemplateSpec(
 	ctx context.Context,
 	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
 	labels map[string]string,
+	containerGPUs dynamo.ContainerGPUCount,
 ) (*corev1.PodTemplateSpec, error) {
-	workerPodTemplateSpec, err := r.generatePodTemplateSpec(ctx, dcd, dynamo.RoleWorker)
+	workerPodTemplateSpec, err := r.generatePodTemplateSpec(ctx, dcd, dynamo.RoleWorker, containerGPUs)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate worker pod template")
 	}
@@ -150,6 +163,7 @@ func (r *dcdWorkloadRenderer) generatePodTemplateSpec(
 	ctx context.Context,
 	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
 	role dynamo.Role,
+	containerGPUs dynamo.ContainerGPUCount,
 ) (*corev1.PodTemplateSpec, error) {
 	component := &dcd.Spec.DynamoComponentDeploymentSharedSpec
 	componentType, err := r.getDCDWorkloadComponentType(ctx, dcd)
@@ -216,6 +230,7 @@ func (r *dcdWorkloadRenderer) generatePodTemplateSpec(
 		role,
 		commonconsts.MultinodeDeploymentTypeLWS,
 		checkpointInfo,
+		containerGPUs,
 		dynamo.GenerateBasePodSpecForControllerOptions{
 			WorkloadComponentType: nvidiacomv1beta1.ComponentType(componentType),
 		},
--- a/deploy/operator/internal/controller/dynamographdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeployment_controller.go
@@ -27,6 +27,7 @@ import (
 
 	networkingv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/scale"
@@ -262,7 +263,22 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			UpdateFunc:  func(de event.UpdateEvent) bool { return true },
 			GenericFunc: func(ge event.GenericEvent) bool { return true },
 		})).
-		WithEventFilter(commoncontroller.EphemeralDeploymentEventFilter(r.Config, r.RuntimeConfig))
+		WithEventFilter(deploymentEventFilter(r.Config, r.RuntimeConfig))
+	if r.RuntimeConfig.Gate.Enabled(features.DRA) {
+		ctrlBuilder = ctrlBuilder.Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimToDGDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).Watches(
+			&resourcev1.ResourceClaimTemplate{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimTemplateToDGDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).Watches(
+			&resourcev1.DeviceClass{},
+			handler.EnqueueRequestsFromMapFunc(r.mapDeviceClassToDGDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		)
+	}
 	if r.RuntimeConfig.Gate.Enabled(features.Istio) {
 		ctrlBuilder = ctrlBuilder.Owns(&networkingv1beta1.DestinationRule{}, builder.WithPredicates(predicate.Funcs{
 			CreateFunc:  func(ce event.CreateEvent) bool { return false },
--- a/deploy/operator/internal/dra/dra.go
+++ b/deploy/operator/internal/dra/dra.go
@@ -8,6 +8,7 @@ package dra
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -31,6 +32,11 @@ const (
 	// source of truth for this string across the operator.
 	DefaultDeviceClassName = "gpu.nvidia.com"
 )
+
+// deviceDriverEqualityPattern extracts positive device.driver equality constraints
+// from CEL conjunctions. classifyDeviceClass deliberately ignores disjunctions,
+// where one matching branch does not classify every device selected by the class.
+var deviceDriverEqualityPattern = regexp.MustCompile(`(?:^|&&)\s*\(*\s*device\.driver\s*==\s*["']([^"']+)["']`)
 
 // ApplyClaim replaces the first container's scalar GPU resources with a shared
 // DRA ResourceClaim. Every container that references this claim name will share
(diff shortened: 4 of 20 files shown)


sttts:
One concern with `isGPUDeviceClassName`: DRA `DeviceClass` names are cluster-defined and do not have to follow a `gpu.*` or `*/gpu` naming convention. A custom GPU class such as `shared-h100.example.com` would therefore be treated as non-GPU, yielding a count of zero and potentially recreating the broken multinode rendering this PR fixes.

Could we support an explicit/configurable set of GPU DeviceClass names (or another authoritative signal), and fail with an actionable error when a container-selected claim cannot be classified? Silently ignoring an unrecognized class seems risky here.

deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
```diff
@@ -910,6 +912,19 @@ func (r *DynamoComponentDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
 		Owns(&networkingv1.Ingress{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		WithEventFilter(commonController.EphemeralDeploymentEventFilter(r.Config, r.RuntimeConfig))
 
+	if r.RuntimeConfig.Gate.Enabled(features.DRA) {
+		m.Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimToDCDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
```
sttts: ).Watches

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
```diff
@@ -290,6 +298,37 @@ func (r *dcdWorkloadRenderer) generatePodTemplateSpec(
 	}, nil
 }
 
+func (r *dcdWorkloadRenderer) resolveGPUsPerNode(
+	ctx context.Context,
+	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
+) (int64, error) {
+	if !dcd.IsMultinode() {
+		return 0, nil
```
sttts: this is strange. Either return the same value, or restrict the func by name to only mean multi-node GPUs or so. Zero is really not the value I would expect for single-node here.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -263,6 +264,18 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			GenericFunc: func(ge event.GenericEvent) bool { return true },
 		})).
 		WithEventFilter(commoncontroller.EphemeralDeploymentEventFilter(r.Config, r.RuntimeConfig))
+	if r.RuntimeConfig.Gate.Enabled(features.DRA) {
+		ctrlBuilder.Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimToDGDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
```
sttts: ).Watches

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
```diff
@@ -290,6 +298,37 @@ func (r *dcdWorkloadRenderer) generatePodTemplateSpec(
 	}, nil
 }
 
+func (r *dcdWorkloadRenderer) resolveGPUsPerNode(
```
sttts: do we really have the func twice???? Don't the share a spec?

deploy/operator/internal/dra/dra.go
```diff
@@ -32,6 +33,8 @@ const (
 	DefaultDeviceClassName = "gpu.nvidia.com"
 )
 
+var deviceDriverEqualityPattern = regexp.MustCompile(`(?:^|&&)\s*\(*\s*device\.driver\s*==\s*["']([^"']+)["']`)
```
sttts, line 39: go doc, why we need this regex. What is it for?

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1558,6 +1562,42 @@ func GenerateBasePodSpec(
...
+		deployerOverride,
+		backendRenderContext{},
+	)
+}
+
+// generateBasePodSpec accepts controller-resolved resource information while
+// preserving GenerateBasePodSpec as the client-independent rendering entrypoint.
+//
+//nolint:gocyclo
+func generateBasePodSpec(
```
sttts: do not inject useless helper indirections.

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1428,6 +1428,10 @@ type MultinodeDeployer interface {
 	NeedsDNSWait() bool          // returns true if DNS wait is needed to launch multinode components
 }
 
+type backendRenderContext struct {
+	GPUsPerNode int64
```
sttts: why is this in the context?

deploy/operator/internal/dra/dra.go
```diff
@@ -104,6 +104,166 @@ func ExtractGPUCountFromResourceRequirements(resources corev1.ResourceRequiremen
...
+		name := *podClaim.ResourceClaimName
+		claim := &resourcev1.ResourceClaim{}
+		if err := cl.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, claim); err != nil {
+			return nil, fmt.Errorf("failed to get ResourceClaim %s/%s: %w", namespace, name, err)
+		}
+		return &claim.Spec, nil
+	default:
+		return nil, fmt.Errorf("pod resourceClaim %q does not reference a ResourceClaim or ResourceClaimTemplate", podClaim.Name)
+	}
+}
```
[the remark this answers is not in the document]
sttts: Fixed. Both DCD and DGD now watch ResourceClaim and ResourceClaimTemplate changes and enqueue deployments that reference them. DeviceClass changes are watched too, because they also affect the resolved GPU count.

deploy/operator/internal/dra/dra.go
```diff
@@ -104,6 +104,166 @@ func ExtractGPUCountFromResourceRequirements(resources corev1.ResourceRequiremen
...
+
+func gpuCountFromDeviceRequest(request resourcev1.DeviceRequest) (int, bool, error) {
+	if request.Exactly != nil {
+		if !isGPUDeviceClassName(request.Exactly.DeviceClassName) {
+			return 0, false, nil
+		}
+		count, err := exactDeviceCount(request.Exactly.AllocationMode, request.Exactly.Count)
+		return count, true, err
+	}
```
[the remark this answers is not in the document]
sttts: Fixed. A request with neither exactly nor firstAvailable now returns a clear error, which ResolveGPUCount propagates.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -474,14 +480,13 @@ func hasFlag(expandedArgs []string, flag string) bool {
...
-func injectDataParallelLaunchFlags(container *corev1.Container, role Role, serviceName string, multinodeDeployer MultinodeDeployer, resources *corev1.ResourceRequirements, numberOfNodes int32) {
+func injectDataParallelLaunchFlags(container *corev1.Container, role Role, serviceName string, multinodeDeployer MultinodeDeployer, gpusPerNode int64, numberOfNodes int32) {
 	expandedArgs := getExpandedArgs(container)
 	leaderHostname := multinodeDeployer.GetLeaderHostname(serviceName)
 
 	// Calculate engines per node
-	containerGPUs := getContainerGPUs(resources)
 	worldSize := getWorldSize(expandedArgs) // TP * PP per engine
-	dataParallelSizeLocal := containerGPUs / worldSize
+	dataParallelSizeLocal := gpusPerNode / worldSize
```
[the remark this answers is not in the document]
sttts: No code change needed: injectDataParallelLaunchFlags is called only when needsDataParallelMultinodeLaunch is true. With worldSize <= 0 that predicate is false, so the division is unreachable.

deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
sttts: Fixed: the returned builder is now assigned back to m.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
sttts: Renamed the shared function to ResolveMultinodeVLLMGPUsPerNode and documented that non-vLLM and single-node components intentionally return zero.

deploy/operator/internal/controller/dynamographdeployment_controller.go
sttts: Fixed: the returned builder is now assigned back to ctrlBuilder.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
sttts: Unified. DCD and DGD now use the same SharedSpec-based ResolveMultinodeVLLMGPUsPerNode implementation.

deploy/operator/internal/dra/dra.go
sttts: Added GoDoc: the regex extracts positive device.driver equality constraints from CEL conjunctions; disjunctions are deliberately ignored because one matching branch cannot classify the whole DeviceClass.

deploy/operator/internal/dynamo/graph.go
sttts: Removed the helper. GenerateBasePodSpec now accepts the resolved GPUs-per-node value directly.

deploy/operator/internal/dynamo/graph.go
sttts: Removed the one-field context and pass gpusPerNode directly through the rendering path.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
```diff
@@ -290,6 +297,23 @@ func (r *dcdWorkloadRenderer) generatePodTemplateSpec(
...
 
+func (r *dcdWorkloadRenderer) resolveMultinodeVLLMGPUsPerNode(
+	ctx context.Context,
+	dcd *nvidiacomv1beta1.DynamoComponentDeployment,
+) (int64, error) {
+	backendFramework, err := dynamo.GetBackendFrameworkFromDynamoComponent(dcd)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to determine backend framework")
+	}
+	return dynamo.ResolveMultinodeVLLMGPUsPerNode(
```
sttts: why is this vllm specific? Is it?

deploy/operator/internal/dynamo/graph.go
```diff
@@ -2004,13 +2009,43 @@ func generatePodSpecForComponent(
...
+// ResolveMultinodeVLLMGPUsPerNode returns the GPU count needed to derive vLLM's
+// distributed launch flags. Other backends and single-node components return zero.
+func ResolveMultinodeVLLMGPUsPerNode(
+	ctx context.Context,
+	reader ctrlclient.Reader,
+	namespace string,
+	component *v1beta1.DynamoComponentDeploymentSharedSpec,
+	backendFramework BackendFramework,
+) (int64, error) {
+	if component == nil || backendFramework != BackendFrameworkVLLM || !component.IsMultinode() {
```
sttts: no, please do not invent return values for unsupported input. This is a bad pattern. Don't call the function in those circumstances. At best, return an error here.

sttts: and if I am not mistaken, nothing below has anything to do with vllm?

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
sttts: Correct: the GPU-count resolution itself is not vLLM-specific. I renamed it to ResolveGPUsPerNode and removed backend/node-count knowledge from it. The DCD call site now checks multinode + vLLM before calling it.

deploy/operator/internal/dynamo/graph.go
sttts: Agreed. ResolveGPUsPerNode now only resolves a valid component and errors on nil; it no longer invents zero for non-vLLM or single-node input. Both call sites perform those checks before calling it.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -262,7 +263,24 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
...
 		})).
-		WithEventFilter(commoncontroller.EphemeralDeploymentEventFilter(r.Config, r.RuntimeConfig))
+		WithEventFilter(deploymentEventFilter(r.Config, r.RuntimeConfig))
+	if r.RuntimeConfig.Gate.Enabled(features.DRA) {
+		ctrlBuilder = ctrlBuilder.Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.mapResourceClaimToDGDRequests),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
+			Watches(
```
sttts: no, the point was also to put the Watches on the line with the dot

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1991,6 +2028,7 @@ func generatePodSpecForComponent(
 	checkpointInfo *checkpoint.CheckpointInfo,
 	deployerOverride MultinodeDeployer,
 	groveClusterTopologyDomains []v1beta1.TopologyDomain,
+	gpusPerNode int64,
```
sttts: can't this be derived from component?

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1558,6 +1556,42 @@ func GenerateBasePodSpec(
...
+		parentGraphDeploymentName,
+		namespace,
+		role,
+		numberOfNodes,
+		operatorConfig,
+		multinodeDeploymentType,
+		serviceName,
+		checkpointInfo,
+		deployerOverride,
+		0,
```
sttts: can't this be derived from component?

deploy/operator/internal/dynamo/graph.go
```diff
@@ -2545,6 +2612,18 @@ func GenerateGrovePodCliqueSet(
 
 		numberOfNodes := component.GetNumberOfNodes()
 		isMultinode := numberOfNodes > 1
+		var gpusPerNode int64
+		if backendFramework == BackendFrameworkVLLM && isMultinode {
+			gpusPerNode, err = ResolveGPUsPerNode(
```
sttts: am curious: why does this only matter for vLLM?

deploy/operator/internal/controller/dynamographdeployment_controller.go
sttts: Fixed in c4a2636a7b. Both DGD and DCD builder chains now use ).Watches(, so Watches stays on the same line as the dot as requested.

deploy/operator/internal/dynamo/graph.go
sttts: Only the scalar GPU limit/request can be derived from component. For DRA the component contains references; the count lives in ResourceClaim/ResourceClaimTemplate and DeviceClass objects and needs a Kubernetes reader. In c4a2636a7b I made that distinction explicit: nil means derive scalar resources from component, while a non-nil resolvedGPUsPerNode carries the externally resolved DRA count. This removes the 0 sentinel.

deploy/operator/internal/dynamo/graph.go
sttts: Same distinction here: the public reader-independent wrapper now passes nil, meaning derive the scalar count from component, instead of passing 0. Only the controller/Grove paths provide an externally resolved DRA count.

deploy/operator/internal/dynamo/graph.go
sttts: This PR needs it specifically for vLLM because its injected TP/DP flags compare the configured world size with GPUs per node. SGLang multinode injection uses only node count and rank. TensorRT-LLM does derive GPUs per node for mpirun -n, so it may have the same pure-DRA gap; since that path has not been validated, I am not claiming or adding TensorRT-LLM support in this vLLM PR.

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1403,18 +1404,34 @@ func ParseBackendFramework(framework string) (BackendFramework, error) {
...
 }
 
+// ContainerGPUCount lazily resolves the main container's scalar or DRA-backed
+// GPU count. The same resolver can be shared across all roles of a component.
+type ContainerGPUCount func() (int64, error)
+
+// MemoizeContainerGPUCount returns a resolver that evaluates resolve at most
+// once and reuses both its count and error.
+func MemoizeContainerGPUCount(resolve ContainerGPUCount) ContainerGPUCount {
+	return sync.OnceValues(resolve)
```
sttts: arg, inline this. Why a func

deploy/operator/internal/dynamo/backend_vllm_test.go
```diff
@@ -187,11 +174,65 @@ func TestVLLMBackend_UpdateContainer(t *testing.T) {
...
+			expectProbesRemoved: true,
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			g := gomega.NewGomegaWithT(t)
+			containerGPUs := tt.gpuCount
+			if tt.resolvedContainerGPUCount > 0 {
+				containerGPUs = tt.resolvedContainerGPUCount
```
sttts: warum brauchen wir gpuCount und resolvedContainerGPUCount

sttts: Good catch. These fields were leftover transition plumbing. I collapsed them into a single `containerGPUs` field and removed the unrelated `component.Resources` mutation; this backend test now gets the GPU count only through the resolver, matching production behavior. Re-reviewed at `8128a45039` with no further findings.

julienmancuso:
k8s determines the actual device using both DeviceClass rules and request-level selectors
so there would be some unusual cases where we would still not count correctly the number of gpus 
```
apiVersion: resource.k8s.io/v1
kind: DeviceClass
metadata:
  name: gpu.nvidia.com
spec: {}  # The class itself does not restrict the device driver
---
apiVersion: resource.k8s.io/v1
kind: ResourceClaimTemplate
metadata:
  name: selected-device
spec:
  spec:
    devices:
      requests:
      - name: device
        exactly:
          deviceClassName: gpu.nvidia.com
          selectors:
          - cel:
              expression: device.driver == "dra.net.example.com"
          count: 1
```

or 

```
deviceClassName: accelerators.example.com
selectors:
- cel:
    expression: device.driver == "gpu.nvidia.com"
```

but i guess these are non-common cases and can be ignored

julienmancuso (approved):

sttts:
@julienmancuso Good catch — fixed in `589a095662`.

The resolver now evaluates the `DeviceClass` selectors together with the request-level selectors for both `exactly` and `firstAvailable`, matching Kubernetes' conjunctive selection semantics. It caches the `DeviceClass` object rather than the final classification so the same class can be classified independently for different requests.

This covers both examples:
- a GPU-looking empty class narrowed to a non-GPU driver is not counted;
- an otherwise unclassifiable class narrowed to a GPU driver is counted.

Conflicting GPU/non-GPU driver constraints now fail with an actionable error. Tests cover both examples, conflicting selectors, per-request classification of the same class, and `firstAvailable` request selectors.

PR #12170, opened 2026-07-25
fix(runtime): restrict event publisher IDs to 63 bits

jthomson04:
## Summary

Restrict event publisher IDs to the signed 63-bit range so Kubernetes discovery preserves their exact value.

## Overview:

Event publishers previously generated full-range random `u64` IDs. Kubernetes stores `DiscoveryMetadata` in untyped JSON, whose Go conversion path uses signed 64-bit integers and falls back to floating point for larger values.

That creates two failure modes:

- if the rounded value still fits in `u64`, discovery returns a different publisher ID than the MessagePack event envelope, so the subscriber cannot bind the event batch to its publisher;
- if the rounded value exceeds `u64::MAX`, deserializing `DiscoveryMetadata` fails and the entire pod CR is excluded from discovery, hiding its endpoints, model cards, and event registrations.

## Details:

- constrain generated publisher IDs to 63 bits before the same ID is reused by event envelopes, channel discovery, and source discovery;
- reject out-of-range `EventChannel` and `EventSource` IDs at the Kubernetes registration boundary rather than silently remapping them;
- test the signed-integer invariant and the serialized Kubernetes metadata representation.

- `lib/runtime/src/transports/event_plane/mod.rs` for publisher-ID generation;
- `lib/runtime/src/discovery/kube.rs` for the Kubernetes registration guard;
- `lib/runtime/src/discovery/kube/crd.rs` for the metadata serialization regression coverage.

## Related Issues

## Validation

- `cargo test -p dynamo-runtime discovery::kube --lib --quiet`
- `cargo test -p dynamo-runtime event_plane --lib --quiet`
- `git diff --check`

### DIFF
--- a/lib/runtime/src/transports/event_plane/mod.rs
+++ b/lib/runtime/src/transports/event_plane/mod.rs
@@ -259,6 +259,11 @@ impl Stream for DeduplicatingStream {
     }
 }
 
+/// Keep publisher IDs representable as signed integers in Kubernetes discovery metadata.
+fn discovery_safe_publisher_id(random_id: u64) -> u64 {
+    random_id & (i64::MAX as u64)
+}
+
 /// Event publisher for a specific topic.
 pub struct EventPublisher {
     transport_kind: EventTransportKind,
@@ -387,9 +392,11 @@ impl EventPublisher {
         // can host multiple publishers for the same scope/topic, each with its
         // own ZMQ endpoint and sequence space, so the process ID is not unique
         // enough here.
-        let publisher_id = rand::rngs::OsRng
-            .try_next_u64()
-            .map_err(|error| anyhow::anyhow!("failed to generate publisher ID: {error}"))?;
+        let publisher_id = discovery_safe_publisher_id(
+            rand::rngs::OsRng
+                .try_next_u64()
+                .map_err(|error| anyhow::anyhow!("failed to generate publisher ID: {error}"))?,
+        );
         let discovery = Some(drt.discovery());
         let runtime_handle = drt.runtime().secondary();
         let subject = scope.subject(&topic);
@@ -914,6 +921,12 @@ fn current_timestamp_ms() -> u64 {
 mod tests {
     use super::*;
     use crate::config::environment_names::zmq_broker as broker_env;
+
+    #[test]
+    fn publisher_ids_fit_kubernetes_discovery_integer_range() {
+        assert_eq!(discovery_safe_publisher_id(42), 42);
+        assert_eq!(discovery_safe_publisher_id(u64::MAX), i64::MAX as u64);
+    }
 
     #[test]
     fn direct_zmq_topology_selection_is_narrow() {
--- a/lib/runtime/src/transports/event_plane/mod.rs
+++ b/lib/runtime/src/transports/event_plane/mod.rs
@@ -259,7 +259,7 @@ impl Stream for DeduplicatingStream {
     }
 }
 
-/// Keep publisher IDs representable as signed integers in Kubernetes discovery metadata.
+/// Keep the shared wire, channel, and source publisher ID representable in Kubernetes metadata.
 fn discovery_safe_publisher_id(random_id: u64) -> u64 {
     random_id & (i64::MAX as u64)
 }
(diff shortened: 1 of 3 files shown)


lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -915,6 +922,12 @@ mod tests {
     use super::*;
     use crate::config::environment_names::zmq_broker as broker_env;
 
+    #[test]
+    fn publisher_ids_fit_kubernetes_discovery_integer_range() {
+        assert_eq!(discovery_safe_publisher_id(42), 42);
+        assert_eq!(discovery_safe_publisher_id(u64::MAX), i64::MAX as u64);
```
jthomson04: This asserts the implementation rather than the invariant — `discovery_safe_publisher_id(u64::MAX) == i64::MAX as u64` is a restatement of `& (i64::MAX as u64)` and would pass for any masking bug that happens to be written the same way twice.

The property that actually matters is int64 representability:

```suggestion
        assert_eq!(discovery_safe_publisher_id(42), 42);
        assert!(i64::try_from(discovery_safe_publisher_id(u64::MAX)).is_ok());
```

Stronger still would be a round-trip that exercises the real constraint — serialize a `DiscoverySpec::EventChannel` to `serde_json::Value` and assert the `publisher_id` number reports `is_i64()`. That fails on `main` and passes here, which makes it a regression test for the actual Kubernetes round-trip rather than a unit test of a bitmask.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -259,6 +259,11 @@ impl Stream for DeduplicatingStream {
     }
 }
 
+/// Keep publisher IDs representable as signed integers in Kubernetes discovery metadata.
+fn discovery_safe_publisher_id(random_id: u64) -> u64 {
```
jthomson04, line 263: Placement note: this helper lives in the event-plane transport module, but the constraint it encodes is imposed entirely by the Kubernetes discovery backend. Nothing at this layer prevents the next `DiscoverySpec` variant from skipping it.

Concretely, `DiscoverySpec::EventSource` (`lib/runtime/src/discovery/mod.rs:585`) carries its own `publisher_id: u64` into the same `DiscoveryMetadata` blob. It has no construction site in the repo today, so nothing is broken — but it inherits this bug the moment someone wires it up.

Applying the mask at the discovery registration boundary instead would make it unskippable for every current and future variant. Fine to leave as-is for this fix if you'd rather keep the diff minimal; flagging so it's a deliberate choice.

jthomson04 (commented):
Review notes on the fix. The bound itself is right: the operative constraint is that `k8s.io/apimachinery/pkg/util/json`'s `convertNumber` tries `Int64()` first and only falls back to `Float64()` when that fails, so masking to 63 bits keeps every publisher ID on the int64 path and out of the float fallback entirely. Same boundary through structured-merge-diff. Three follow-ups, one of which is just the PR description.

### The PR description understates the failure mode

The body says the mismatch "prevents subscribers from attributing those event batches to their discovered publisher." The blast radius is larger than that. Once the apiserver hands back `1.8446744073709552e+19`, `serde_json::from_value::<DiscoveryMetadata>` at `lib/runtime/src/discovery/kube/daemon.rs:294` fails on the `u64` field, and the `Err` arm drops the **entire CR** into `invalid_crs` — endpoints, model cards, everything for that pod, not just the event channel.

That is the same catastrophic mode already documented in the `deserialize_null_default` comment at `lib/runtime/src/discovery/metadata.rs:13-31`: the worker becomes invisible to service discovery and inference requests 404. Worth stating in the description, since it changes how urgently this wants backporting.

lib/runtime/src/transports/event_plane/mod.rs
jthomson04: Addressed in 445475b66b. The unit test now checks the actual invariant with `i64::try_from(...).is_ok()`, and the existing Kubernetes metadata round-trip test now uses `i64::MAX` and asserts that the serialized event-source `publisher_id` is an `i64` JSON number before deserializing the CR metadata.

lib/runtime/src/transports/event_plane/mod.rs
jthomson04: Addressed in 445475b66b. I kept the 63-bit normalization at publisher creation because the MessagePack envelope, `EventChannel`, and `EventSource` must all retain the same exact ID; masking only at Kubernetes registration would recreate the discovery/wire mismatch. The Kubernetes registration boundary now rejects out-of-range IDs for both discovery variants, so future or independent call sites cannot silently publish an unsafe value. The current `EventSource` path in `lib/llm/src/kv_router/publisher/mod.rs` also reuses the already-constrained event-publisher ID.

PeaBrane (approved):

PR #11014, opened 2026-06-27
fix(runtime): give event publishers unique discovery identities

zhongdaor-nv:
## Summary

- Give every `EventPublisher` a fork-safe, per-incarnation publisher ID and use the same ID in both `EventEnvelope` and event-channel discovery.
- Keep the existing four-part discovery key and serialized `DiscoveryInstance::EventChannel` shape while allowing multiple publishers for the same process/component/topic.
- Track ZMQ publisher connections by full `DiscoveryInstanceId` and ignore unrelated topics.
- Keep asynchronous discovery unregister work alive through graceful-shutdown Phase 2.
- Make discovery registration terminology unambiguous: conversion uses a documented `default_instance_id`, registration code logs only the final `instance_id`, and EventPublisher logs use `publisher_id`. The old public `with_instance_id` method remains as a compatibility alias.
- Add direct ZMQ regression coverage for two same-topic publishers and drop-to-immediate-recreate, including a fresh identity/sequence space, eventual discovery state, and continued delivery from both surviving and recreated publishers.
- Remove the mocker E2E autouse fixture that forced NATS, so nondurable mocker coverage exercises the default ZMQ event plane.

The root cause was that each publisher bound its own ZMQ endpoint and owned its own sequence counter, but all publishers in one process registered with the process-level discovery ID. Their discovery keys therefore collided, subscribers only connected to one endpoint, and dropping any colliding publisher could remove the shared registration.

This also makes `(publisher_id, sequence)` valid for broker deduplication across multiple publishers and prevents a late unregister from an old publisher incarnation from removing its replacement. The wire schema and discovery path structure are unchanged. The public Rust `DiscoverySpec::EventChannel` variant now requires `publisher_id`.

## Validation

- Rebased onto `origin/main` at `2469510c7d6` and resolved the event-plane conflict while retaining both main's cached NATS subject path and this change's publisher lifecycle handling.
- `cargo fmt --all`
- `git diff --check`
- `cargo test -p dynamo-runtime transports::event_plane:: --lib` — 18 passed
- Drop-to-recreate direct-ZMQ regression repeated 20 times without failure
- `cargo test -p dynamo-runtime --lib` — 433 passed, 2 ignored
- `cargo clippy -p dynamo-runtime --lib --tests -- -D warnings`
- `cargo check -p dynamo-llm -p dynamo-backend-common`
- `maturin develop --uv`
[...]

PR #12185, opened 2026-07-26
fix(kv-router): restore multinode SGLang KV events

rmccorm4:
Credit to @furionw for identifying the bug and the fix

## Summary

Fix KV-aware routing for multinode SGLang workers on Kubernetes.

- Publish KV events only from the leader node when multinode workers share a single DP-rank slice.
- Preserve per-node publishing when DP attention assigns distinct rank slices.
- Restrict event publisher IDs to 53 bits so they survive float64-backed JSON discovery round trips without precision loss.
- Add regression coverage for multinode publisher selection and JSON-safe publisher IDs.

Related: [DIS-2519](https://linear.app/nvidia/issue/DIS-2519/fix-multinode-sglang-kv-event-ingestion-on-kubernetes)

## Validation

- `cargo test -p dynamo-runtime publisher -- --nocapture`
  - 7 passed
- `components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py`
  - 5 passed locally
  - The SGLang dependency-presence skip was disabled because SGLang is not installed locally; these tests exercise dependency-free capacity logic.
- `pre-commit run --files <changed files>`
  - All hooks passed
- `cargo fmt --all --check`
  - Passed
- Python compilation and `git diff --check`
  - Passed

## Root cause analysis

These were two recently activated failures.

The TP-only multinode collision became possible when non-leader SGLang nodes began publishing under the leader worker ID while every node continued advertising the shared DP-rank slice `[0, 1)`. Endpoint-scoped, fail-closed source membership then made duplicate `(worker ID, DP rank)` advertisements ambiguous, leaving direct-ZMQ with no ready source. DP-attention deployments assign distinct rank slices and do not have this collision.

The publisher-ID failure originated in #11014, which replaced Kubernetes' bounded pod instance ID with a full-range random `u64` publisher ID. Values above `i64::MAX` can fall back to floating-point representation in Kubernetes' unstructured JSON path and return rounded. Once direct-ZMQ began strictly comparing the discovered publisher ID with the frame and envelope IDs, every mismatched event was dropped.

#12170 bounded publisher IDs to 63 bits, matching Kubernetes' signed-integer representation. This change uses 53 bits as a conservative bound that is also safe through generic float64-backed JSON processing. Proving that 53 bits, rather than 63 bits, is strictly required would need an exact Kubernetes round-trip failure for an ID between `2^53` and `i64::MAX`.

Non-Kubernetes discovery backends avoid this JSON serialization issue. NATS does not perform the same discovered-ID-to-frame comparison, although malformed Kubernetes discovery metadata can still affect event registrations or exclude the containing worker metadata resource.

## Suggested follow-up testing

Add a pre-merge Kubernetes deployment test for multinode SGLang KV-event ingestion:

1. Deploy one KV-routing frontend and two logical SGLang workers.
2. Configure each worker as a two-node TP-only gang with DP attention disabled and KV events enabled.
3. Wait for discovery and assert:
   - exactly one active KV-event source exists per logical worker;
   - no sources remain `preconnected`;
   - no source is reported as ambiguous;
   - `identity_mismatch` remains zero.
4. Send two requests with distinct, sufficiently long prefixes, wait for their KV events to propagate, and replay each prefix.
5. Assert:
   - direct-ZMQ batch count is greater than zero;
   - successfully applied KV-cache event count is greater than zero;
   - each warm replay routes to the same logical worker as its corresponding cold request;
   - `identity_mismatch` remains zero after the replays.
6. Capture the publisher IDs advertised in `DynamoWorkerMetadata` and compare them with the IDs observed by the frontend subscriber.

The Kubernetes test should cover topology, discovery, and end-to-end attribution. Keep the deterministic Rust publisher-ID bound test as the non-probabilistic guard against accidentally restoring full-range `u64` generation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved KV-event publishing across multi-node deployments to prevent duplicate event sources and ensure correct routing.
  - Added support for publishing KV events from the appropriate nodes when data-parallel attention is enabled.
  - Improved publisher ID handling so event identifiers remain reliable during transport and JSON-based discovery.
- **Tests**
  - Expanded coverage for single-node, multi-node, and data-parallel configurations, including publisher ID integrity.

### DIFF
--- a/components/src/dynamo/sglang/publisher.py
+++ b/components/src/dynamo/sglang/publisher.py
@@ -26,7 +26,11 @@ from dynamo.llm import KvEventPublisher, WorkerMetricsPublisher
 from dynamo.runtime import Endpoint
 from dynamo.sglang._disagg import SGLANG_WORKER_GROUP_ID_KEY, get_sglang_worker_group_id
 from dynamo.sglang.args import Config
-from dynamo.sglang.capacity import kv_metrics_block_values, local_dp_rank_bounds
+from dynamo.sglang.capacity import (
+    kv_metrics_block_values,
+    local_dp_rank_bounds,
+    publishes_kv_events,
+)
 
 
 def get_local_dp_rank_range(server_args) -> range:
@@ -296,7 +300,14 @@ class DynamoSglangPublisher:
             List of KvEventPublisher instances if KV event publishing is enabled,
             empty list otherwise.
         """
-        if self.dynamo_args.use_kv_events:
+        if self.dynamo_args.use_kv_events and not publishes_kv_events(self.server_args):
+            logging.info(
+                "Non-leader node (node_rank=%s) shares the leader's single KV "
+                "rank slice; skipping KV event publishing so the router sees "
+                "exactly one source per (worker_id, dp_rank).",
+                getattr(self.server_args, "node_rank", 0) or 0,
+            )
+        elif self.dynamo_args.use_kv_events:
             kv_events = json.loads(self.server_args.kv_events_config)
             base_ep = kv_events.get("endpoint")
             if not base_ep:
--- a/components/src/dynamo/sglang/publisher.py
+++ b/components/src/dynamo/sglang/publisher.py
@@ -576,7 +576,9 @@ async def handle_non_leader_node(
     )
 
     try:
-        if publisher.dynamo_args.use_kv_events:
+        if publisher.dynamo_args.use_kv_events and publishes_kv_events(
+            publisher.server_args
+        ):
             kv_worker_id = await _resolve_multinode_leader_worker_id(
                 publisher.generate_endpoint,
                 publisher.server_args,
--- a/components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py
+++ b/components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py
@@ -8,6 +8,12 @@ from types import SimpleNamespace
 
 import pytest
 
+from dynamo.sglang.capacity import (
+    local_dp_rank_bounds,
+    model_card_dp_rank_bounds,
+    publishes_kv_events,
+)
+
 pytestmark = [
     pytest.mark.unit,
     pytest.mark.sglang,
@@ -22,8 +28,6 @@ pytestmark = [
 
 
 def test_model_card_registration_keeps_global_dp_range():
-    from dynamo.sglang.capacity import model_card_dp_rank_bounds
-
     server_args = SimpleNamespace(
         dp_size=16,
         enable_dp_attention=True,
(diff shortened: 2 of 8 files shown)


furionw (approved):

furionw:
Thank you

components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py
```diff
@@ -32,3 +36,43 @@ def test_model_card_registration_keeps_global_dp_range():
...
+    base = dict(dp_size=1, enable_dp_attention=False, nnodes=1, node_rank=0)
+    base.update(kwargs)
+    return SimpleNamespace(**base)
+
+
+def test_single_node_publishes_kv_events():
+    assert publishes_kv_events(_args()) is True
+
+
+def test_multinode_without_dp_attention_publishes_only_from_leader():
```
PeaBrane, line 51: The helper-level assertion is useful, but it still passes while `handle_non_leader_node()` resolves the leader ID before this predicate is consulted in `init_kv_event_publish()`. Could we add or update the publisher-level test to assert that a TP-only follower neither calls `_resolve_multinode_leader_worker_id()` nor constructs a `KvEventPublisher`? That would protect the initialization ordering as well as the predicate.

lib/runtime/src/transports/event_plane/mod.rs
```diff
@@ -259,9 +259,11 @@ impl Stream for DeduplicatingStream {
     }
 }
 
-/// Keep the shared wire, channel, and source publisher ID representable in Kubernetes metadata.
+const PUBLISHER_ID_BITS: u32 = 53;
+
+/// Keep publisher IDs exactly representable in float64-backed JSON metadata.
 fn discovery_safe_publisher_id(random_id: u64) -> u64 {
-    random_id & (i64::MAX as u64)
+    random_id >> (u64::BITS - PUBLISHER_ID_BITS)
```
PeaBrane: The 53-bit generator fixes the current in-tree `EventPublisher` callers. As a follow-up hardening point, `KubeDiscoveryClient::validate_kubernetes_publisher_id()` still accepts values through `i64::MAX`, so a future direct `DiscoverySpec::{EventChannel, EventSource}` caller could bypass this invariant. Would it be safer to enforce the 53-bit bound at the Kubernetes discovery boundary—or via a validated publisher-ID type—so the serialization constraint has one owner?

PeaBrane (commented):
Two additional follow-ups after filtering the existing discussion.

PeaBrane (approved):
✅

PeaBrane:
Tracking the remaining SGLang KV-event source-ownership cases in #12186.

The DEP keeps one leader-owned logical serving endpoint and worker ID. Nonleader nodes remain non-routable, but may publish a DP-rank KV source under that leader ID when they host the event-emitting scheduler. It covers normal internal DP with DP-attention disabled and DP-attention groups that span nodes; the TP-only collision fixed here remains in scope of this PR.

components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py
rmccorm4: Addressed in [05e6515](https://github.com/ai-dynamo/dynamo/commit/05e651508a2625a91c73da330cf30316c1ec37c4). `handle_non_leader_node()` now checks `publishes_kv_events()` before resolving the leader worker ID. I also added a publisher-level regression test asserting that a TP-only follower neither awaits `_resolve_multinode_leader_worker_id()` nor constructs a `KvEventPublisher`, and updated the positive non-leader case to use DP attention.

lib/runtime/src/transports/event_plane/mod.rs
rmccorm4: Addressed in [4f2e106](https://github.com/ai-dynamo/dynamo/commit/4f2e106b0431ac9a96ad86d4ee59537000995334). The event-plane generator and Kubernetes validator now share a single `MAX_JSON_SAFE_PUBLISHER_ID` limit. Kubernetes discovery rejects IDs above `2^53 - 1`, including direct `DiscoverySpec::{EventChannel, EventSource}` callers, while the generic discovery API remains `u64`. I also added exact-boundary coverage and updated the CRD round-trip test to use the maximum JSON-safe ID.

components/src/dynamo/sglang/publisher.py
```diff
@@ -296,7 +300,14 @@ def init_kv_event_publish(self) -> List[KvEventPublisher]:
             List of KvEventPublisher instances if KV event publishing is enabled,
             empty list otherwise.
         """
-        if self.dynamo_args.use_kv_events:
+        if self.dynamo_args.use_kv_events and not publishes_kv_events(self.server_args):
```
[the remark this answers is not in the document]
rmccorm4: Addressed in [05e6515](https://github.com/ai-dynamo/dynamo/commit/05e651508a2625a91c73da330cf30316c1ec37c4). `handle_non_leader_node()` now checks `publishes_kv_events()` before calling `_resolve_multinode_leader_worker_id()`, so TP-only followers proceed directly to the idle wait. The commit also adds a publisher-level regression test asserting that the leader resolver and `KvEventPublisher` constructor are not called for that topology.

rmccorm4:
@ishandhanani for review as well

KrishnanPrash (approved):

PR #12189, opened 2026-07-27
refactor(mocker): introduce KV cache groups in the native vLLM backend

nealvaidya:
## Summary

The mocker's native vLLM KV manager assumes one homogeneous block pool. vLLM assumes the opposite: a model's layers are split into **KV cache groups** — one per attention type — each keeping its own per-request block table while drawing from a single physical pool with a uniform page size, with cache visibility namespaced per group (`BlockHashWithGroupId`). This PR builds that seam and stops there: `VllmKvManager` becomes a coordinator over N groups.

Exactly one group — the main attention group — is registered when this lands, so **there is no behavioral change**: no config knob, no scheduler decision change, no new event, no capacity-accounting change. `VllmKvManager`'s `pub(crate)` / `pub(super)` surface is unchanged from `main`, and the diff is confined to `lib/mocker/src/cache/vllm_block_pool.rs` and `lib/mocker/src/kv_manager/`.

Three invariants make that claim checkable in review:

1. Every non-test `GroupedHash` takes its group from the group that owns the block — `self.group` inside a group, `attention_id()` in the coordinator — so no call site can name a group it isn't.
2. `VllmKvManager::new` pushes exactly one `GroupManager`, so every per-group loop runs exactly once, `other_groups()` is empty, and each summed per-group quantity is the attention group's own.
3. Pool capacity, reservations, and the LRU stay global. That *is* vLLM's shared pool, and it is what makes one group indistinguishable from today.

Group implementations move into a new `kv_manager/vllm_groups/` module — `FullAttentionGroup` in its own file, the `GroupManager` contract in `mod.rs` — so most of the diff reads as a file move plus a namespacing change, and the follow-up adds a sibling file instead of reopening the coordinator's file. The coordinator resolves the attention group by kind rather than by index, the way vLLM resolves `full_attention_group_id`, so nothing depends on where that group sits among the others.

One rule has to land here, because this is where the event path gets its filter: only the main attention group's blocks may be reported to the router. `KvCacheEvent` carries no group identity, so a second group's `Removed(H)` would tell the router an attention block is gone while it is still cached. A firewall test keeps group implementations away from the publishers entirely.

Stacked follow-up: `nealv/mocker-mamba-hybrid-kv-cache` adds the Mamba group (`none` / `align` cache modes) that makes this coordination non-trivial. That is where behavior changes.

## Validation

- `cargo test -p dynamo-mocker` — 637 pass (`main`'s suite plus the four added here). No existing test's expected values changed; edits to existing tests are call-site updates for the group-keyed pool.
- Added: pool key namespacing (the same `SequenceHash` in two groups is two distinct copies, and evicting one reports the group-scoped key while leaving the other matchable); the pool release semantics the group layer depends on (a released *cached* copy stays matchable, a released *private* copy is recycled); an attention eviction reaching the event sink as a removal, which the new group filter made worth pinning; and rejection of a partial release before any group mutates.
- `cargo clippy -p dynamo-mocker --all-targets` and `cargo fmt -p dynamo-mocker -- --check` clean.
- `cargo check --workspace` is not usable as a gate here: it fails on macOS inside `dynamo-llm` for pre-existing Linux-only reasons (`dynamo_memory::numa`, `nix::fcntl::fallocate`, `OFlag::O_DIRECT`).

<details>
<summary>Design plan for this branch — rationale, module layout, and what is deliberately deferred to PR 2 (file links are repo-relative paths)</summary>

# PR 1 — KV cache groups in the mocker's native vLLM backend

Stacked PR 1 of 2, split out of [mocker_hybrid_kv_cache_a3cb85d6.plan.md](mocker_hybrid_kv_cache_a3cb85d6.plan.md).

| | |
|---|---|
| Branch | `nealv/mocker-kv-cache-groups` |
| Worktree | `/Users/nealv/Documents/dynamo/dynamo-mocker-kv-groups` |
| Base | `origin/main` @ `a9a7db43a3` |
| Followed by | PR 2 — `nealv/mocker-mamba-hybrid-kv-cache` ([plan](pr2_mocker_mamba_hybrid_cache.md)) |
| Title | `refactor(mocker): introduce KV cache groups in the native vLLM backend` |

## The contract this PR must hold

Generalize the mocker's native vLLM KV manager from *one homogeneous block pool* into a *coordinator over N groups sharing one physical pool*, and stop there. Exactly one group — the main attention group — exists when this PR lands, so **there is no behavioral change**: no new config knob, no scheduler decision change, no new event, no capacity-accounting change.

Three invariants make that claim checkable in review:

1. Every `GroupedHash` built in non-test code takes its group from the group that owns the block — `self.group` inside a group, `attention_id()` in the coordinator — so no call site can name a group it isn't. `rg "GroupedHash::new\(" lib/mocker/src` should show nothing else outside `#[cfg(test)]`.
2. `VllmKvManager::new` pushes exactly one `GroupManager`, so every per-group loop runs exactly once, `other_groups()` is empty, and every summed per-group quantity is the attention group's own.
3. The diff touches only [lib/mocker/src/cache/vllm_block_pool.rs](lib/mocker/src/cache/vllm_block_pool.rs) and [lib/mocker/src/kv_manager/](lib/mocker/src/kv_manager/). No scheduler, `protocols.rs`, bindings, or KV-event-protocol file changes — those all belong to PR 2.

Every existing `#[cfg(test)]` assertion keeps its expected values. Test edits are limited to call sites (the pool's helpers now take a key) plus the two new pool tests below.

## Why groups at all

vLLM splits a hybrid model's layers into **KV cache groups** — one per attention/SSM type — each with its own per-request block table, all drawing from **one shared physical pool** with a uniform page size. Cache keys are namespaced per group (`BlockHashWithGroupId`), and `HybridKVCacheCoordinator` reconciles per-group prefix hits down to a single agreed token boundary. PR 2 adds the Mamba group that makes this non-trivial; this PR builds the seam.

One piece of PR 2's rationale has to land here, because this PR is where the event path gets its group filter: **only the attention group may publish KV events.** Real vLLM tags `BlockStored` / `BlockRemoved` with `group_idx` and `kv_cache_spec_kind`, and Dynamo's `ZmqEventNormalizer` drops the non-main-attention ones ([lib/kv-router/src/zmq_wire/filter.rs](lib/kv-router/src/zmq_wire/filter.rs)) — but that filter exists only on the ZMQ wire path. The mocker's primary path is the typed `KvCacheEventSink`, and `KvCacheEvent` ([lib/kv-router/src/protocols.rs](lib/kv-router/src/protocols.rs)) carries no group identity, so a second group's `Removed(H)` would tell the router that attention block `H` is gone while it is still cached. The coordinator therefore filters removals to the attention group's own id and a firewall test keeps group implementations away from the publishers entirely.

## Module layout

The first pass at this work kept everything in `vllm_backend.rs` and produced a 1,500-line in-place rewrite. Splitting the group implementations into their own module makes this PR read as *a file move plus a namespacing change*, and lets PR 2 add a sibling file instead of reopening the coordinator's file.

| File | Contents |
|---|---|
| `kv_manager/vllm_groups/mod.rs` | The group contract: the `GroupManager` enum, its dispatch, and its by-kind accessors, plus the vocabulary crossing the coordinator↔group boundary — `CacheHit`, `GroupAllocation`, `PendingStore`, `StoredBlock` |
| `kv_manager/vllm_groups/full_attention.rs` | `FullAttentionGroup` plus its private block-table types `OwnedBlock` and `FullBlock` |
| `kv_manager/vllm_backend.rs` | `VllmKvManager` (the coordinator), the reservation types `NativeDecodeBlockReservation` / `NativeDestinationReservation` / `VllmBlockLayout`, event shaping (`StoreGroup`, `publish_store_sequence`, `publish_removed`, `publish_kv_event`), and `get_prefill_cost` |
| `cache/vllm_block_pool.rs` | `KvCacheGroupId`, `GroupedHash`, and the group-keyed pool |

`GroupManager` is an enum, not a `dyn` trait: static dispatch keeps the hot path free of vtables, and the set of group kinds is closed and small.

```mermaid
flowchart TB
    SEQ[ActiveSequence blocks] --> COORD

    subgraph COORD [VllmKvManager coordinator]
      REC[collect prefix keys, sum fresh needs, one pool reserve, commit per group]
      G0[GroupManager::FullAttention  group 0]
      G1[group 1 — PR 2]
    end

    REC --> G0
    REC -.-> G1
    G0 -->|"GroupedHash(0, seq_hash)"| POOL[VllmBlockPool shared capacity + LRU]
    G1 -.->|"GroupedHash(1, seq_hash)"| POOL
    G0 -->|store/remove events| EV[KvEventPublishers]
```

### 1. Group-scope the pool keys

[vllm_block_pool.rs](lib/mocker/src/cache/vllm_block_pool.rs) keys visibility on a bare `SequenceHash`, so two groups caching the same token prefix would collide. Introduce the pool's own leaf-level key type (the file must stay free of `crate::` per [vllm_firewall_tests.rs](lib/mocker/src/kv_manager/vllm_firewall_tests.rs)):

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub(crate) struct KvCacheGroupId(pub(crate) u8);

/// vLLM's `BlockHashWithGroupId`: cache visibility is per KV cache group.
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
pub(crate) struct GroupedHash {
    pub(crate) group: KvCacheGroupId,
    pub(crate) hash: SequenceHash,
}
```

`reserve(&[GroupedHash], fresh)`, `ReserveOutcome::removed`, `cache_private`, `allocate_cached`, `prefix_hit`, `activate_pin`, and `cancel` all move to `GroupedHash`, and the internals follow the same vocabulary: the cache index becomes `by_key` and `CopyState::Cached` stores a `key`. Capacity, `reserved`, and the LRU stay **global** — that is exactly vLLM's shared pool, and keeping them global is what makes "one group" indistinguishable from today.

### 2. Move the attention block table into `vllm_groups/full_attention.rs`

`FullAttentionGroup` is today's per-request block-table logic, unchanged except that it carries a `KvCacheGroupId` and routes every pool call through `fn key(&self, hash) -> GroupedHash`. Its surface, all `pub(crate)` inside the private `vllm_groups` module, or private:

`new`, `id`, `key`, `holds_request`, `block_count`, `prefix_keys`, `fresh_blocks`, `resident_prefix`, `longest_cache_hit`, `commit`, `finalize_computed_prefix`, `deref`, `promote`, and `validate_fresh_partials`.

It owns its `request_blocks: FxHashMap<Uuid, Vec<OwnedBlock>>` and `partial_uuids` set. It does **not** publish: `commit` and `promote` return `StoredBlock`s and the coordinator decides what reaches the event sink.

The coordinator↔group vocabulary in `mod.rs`:

```rust
/// One group's prefix-cache lookup result, in tokens.
#[derive(Clone, Copy, Debug, Default)]
pub(crate) struct CacheHit {
    pub(crate) tokens: usize,
    /// Portion of `tokens` whose copies are already active, and so are not
    /// re-consumed from physical capacity when the request claims them.
    pub(crate) active_tokens: usize,
}

/// Everything a group needs to size and commit one allocation step.
pub(crate) struct GroupAllocation<'a> {
    pub(crate) request_id: Uuid,
    pub(crate) blocks: &'a [UniqueBlock],
    pub(crate) local_hashes: &'a [BlockHash],
    pub(crate) token_ids: Option<&'a [Vec<u32>]>,
    pub(crate) parent: Option<&'a UniqueBlock>,
    /// Leading entries of `blocks` that are already resident and reusable.
    pub(crate) reusable_prefix_blocks: usize,
    /// Destination transfers arrive already computed, so their fresh blocks
    /// enter the prefix cache immediately instead of waiting for a compute
    /// watermark.
    pub(crate) cache_fresh: bool,
}
```

`GroupAllocation` replaces the long argument lists the entrypoints pass down, and owns the check that a step's metadata describes its own blocks (`validate`) — that check is about the struct's internal consistency, not about any group's state. Add no field without a live consumer in this PR — PR 2 adds `target_tokens` when the Mamba group needs the watermark.

### 3. Turn `VllmKvManager` into a coordinator

`VllmKvManager` stays the scheduler-facing type — its `pub(crate)` / `pub(super)` API must be byte-identical to `main` so [g1_manager.rs](lib/mocker/src/kv_manager/g1_manager.rs) and the scheduler are untouched. Internally it gains `groups: Vec<GroupManager>` and every mutating entrypoint follows one shape:

1. Ask every group for its cached-prefix keys (`prefix_keys`) and fresh-block need (`fresh_blocks`).
2. Issue **one** `pool.reserve(&all_prefix_keys, sum_fresh)`.
3. Commit each group against that shared reservation (`commit_groups`).
4. Publish only the attention group's stores, and filter removals: `publish_removed` keeps `key.group == self.attention_id()`.

The entrypoints that run this sequence are `use_for_request`, `use_decode_reservation_for_request`, `reserve_destination_at` / `destination_alloc` / `activate_destination`, `deref_for_request`, `promote_for_request`, `finalize_computed_prefix`, and `reserve_decode_blocks`.

What the coordinator may ask a group to do is the whole reviewable seam, so keep it on `GroupManager` rather than in a match arm in the coordinator: `prefix_keys` and `fresh_blocks` (sizing), `commit` and `finalize_computed_prefix` (making blocks cache-visible), `release`, and `longest_cache_hit`. Two consequences shape the signatures: `commit` drains a shared iterator of activated prefix pins, so a group takes exactly the pins it asked for and leaves the rest to the groups behind it; and `finalize_computed_prefix` takes *token* watermarks rather than block indices, so a group quantizes with its own block size instead of the coordinator's.

Three mechanics worth knowing before starting:

- **Borrow discipline.** Any method that hands the pool to a group while iterating groups must destructure first — `let Self { pool, groups, .. } = self;` — otherwise `&mut self.pool` and `&mut self.groups` collide. This applies to `deref_for_request`, `promote_for_request`, `finalize_computed_prefix`, and `commit_groups`.
- **Attention-group access by kind, never by index.** vLLM sorts groups and resolves `full_attention_group_id` by kind; do the same with `GroupManager::as_full_attention{,_mut}` behind `attention()` / `attention_mut()` / `attention_id()`, and reach the rest through `other_groups()`. Nothing then depends on where the attention group sits among the others, and PR 2 adds a group without revisiting these call sites.
- **A release is total.** Every `Deref` signal derives its block list from the sequence's whole allocated footprint, and a group whose table isn't parallel to the attention table drops a request's state wholesale, so `deref_for_request` checks the release covers the attention table *before* any group mutates — releases run group by group and cannot be rolled back. Trimming a live request is a real vLLM operation (`remove_skipped_blocks`) that the mocker does not model yet, so the check is an assertion rather than a branch.

There is no separate `UnitaryKVCacheCoordinator`-style fast path: with one group, every per-group loop runs once and `other_groups()` contributes no keys, no fresh blocks, and no commits.

### 4. Cross-group reconciliation in `get_prefill_cost`

The reconciliation structure lands here; the group that makes it bite lands in PR 2. Attention reports a contiguous hit; every other group may only shorten it:

```rust
let mut reconciled = attention.tokens;
for group in self.other_groups() {
    let Some(hit) = group.longest_cache_hit(&self.pool, blocks) else {
        continue;
    };
    reconciled = reconciled.min(hit.tokens);
}
```

Groups share a block size, so a single pass converges (vLLM's `is_simple_hybrid` shortcut) — worth a comment, since vLLM's general coordinator loops. `active_cached_tokens` is the attention active-prefix capped by the reconciled length. `PrefillCost` itself is **unchanged** in this PR: no new fields, so [protocols.rs](lib/mocker/src/common/protocols.rs) and [policy.rs](lib/mocker/src/scheduler/vllm/policy.rs) stay out of the diff.

`longest_cache_hit` returns `Option<CacheHit>` so a group can say "I constrain no boundary" without inventing a token count.

### 5. Firewall tests

[vllm_firewall_tests.rs](lib/mocker/src/kv_manager/vllm_firewall_tests.rs) `include_str!`s each source file it guards, so add `vllm_groups/mod.rs` and `vllm_groups/full_attention.rs` to the no-KVBM/no-legacy-G1 list, and generalize the event check to the whole group layer:

```rust
/// A group implementation must never reach the typed event sink: `KvCacheEvent`
/// has no group field, so the router would read another group's hash as an
/// attention block. Only the coordinator publishes.
#[test]
fn groups_never_publish_kv_events() { /* over every vllm_groups/* source */ }
```

This is the enforcement point PR 2 inherits for free when it adds `mamba.rs`.

## Tests

New, and specific to what this PR adds:

- **Pool group-namespacing**: the same `SequenceHash` in group 0 and group 1 occupies two distinct copies; evicting the group-1 copy reports the group-scoped key and leaves the group-0 block matchable by `prefix_hit`.
- **Release semantics the group layer depends on**: a released *cached* copy stays matchable as an inactive `prefix_hit` while a released *private* copy is recycled outright — vLLM's `free_blocks` sending hashed blocks to the free-queue tail and unhashed blocks to the head.

Everything else is a regression check: `cargo test -p dynamo-mocker` with no expected-value edits, `cargo clippy -p dynamo-mocker --all-targets`, `cargo fmt -p dynamo-mocker -- --check`. (`cargo check --workspace` fails on macOS inside `dynamo-llm` for pre-existing Linux-only reasons — `dynamo_memory::numa`, `nix::fcntl::fallocate`, `OFlag::O_DIRECT`.)

## Deliberately deferred to PR 2

Each of these is either a Mamba fact or a field with no live consumer here; a reviewer should not have to accept always-zero values or a no-op hook:

| Deferred | Why it can't justify itself in PR 1 |
|---|---|
| `vllm_groups/mamba.rs`, `MAMBA_GROUP` | the whole point of PR 2 |
| `CacheHit::deferred` | only a state cached in the current pass can defer a request |
| `GroupAllocation::target_tokens`, `VllmKvManager::target_tokens` | nothing consumes the watermark until a group pages state by tokens |
| `GroupManager::{overhead_blocks, decode_blocks_per_new_block, new_scheduling_pass}`, `VllmKvManager::begin_scheduling_pass` (+ its `G1Manager` / [core.rs](lib/mocker/src/scheduler/vllm/core.rs) plumbing) | constant 0 / 1 / no-op with one group |
| `PrefillCost::{group_overhead_blocks, deferred_by_group}`, its `Default` derive, and the `decide_waiting_admission` changes | admission has nothing extra to charge or defer yet |
| `cache_group_boundary_states`, `is_hybrid`, `FullAttentionGroup::full_block_hash` | boundary-state caching exists only for a state-holding group |
| `MambaCacheMode`, `has_mamba_layers`, `mamba_block_size` and their validation/accessors | configuration for a group that doesn't exist yet |
| `mamba_block_aligned_split` and its `schedule_request` call | the only scheduler-visible behavior change in the stack |
| `VllmKvManager::new_with_event_sink` → `new` rename | the rename exists to add the `mamba` argument |

## As implemented — divergences from the sketch above

All of these are consequences of PR 1 having exactly one group; each is listed in PR 2's contract extensions.

| Divergence | Why |
|---|---|
| `vllm_groups` items are `pub(crate)`, not `pub(super)` | `vllm_backend` is a sibling of `vllm_groups`, not a descendant, so `pub(super)` inside `full_attention.rs` would hide `FullAttentionGroup` from the coordinator (and block the `pub use` re-export). `vllm_groups` is a private module, so reachability is still the `kv_manager` subtree — the same idiom `cache/vllm_block_pool.rs` uses |
| `GroupManager::as_full_attention` can't fail, and `attention()` still `expect`s | `GroupManager` has one variant here, so the accessor's `match` is irrefutable and the `expect` is unreachable — but it states the invariant the coordinator actually relies on ("a model has a main attention group"), which survives PR 2 adding a variant that returns `None` |
| `GroupManager::prefix_keys(alloc)` takes no `&VllmBlockPool`, and `longest_cache_hit(pool, blocks)` no `max_tokens` | the attention group needs neither, and an unused parameter would be dead weight a reviewer has to accept. PR 2 widens both when the Mamba group needs the pool and the running boundary |
| `FullAttentionGroup::deref` returns `()`, matching `main`'s `process_deref` | nothing acts on "the request's table just emptied" until another group has per-request state to free |
| `commit_groups`'s shared prefix-pin iterator is drained entirely by the attention group | the drain and the assert that nothing is left over are the contract PR 2 needs, but with one group there is no second consumer to leave pins for |
| `NativeDestinationReservation` gains `attention_fresh` / `attention_prefix` | both have live consumers here — `transferable_prompt_tokens` must report attention blocks rather than the whole reservation, and `activate_destination` must rebuild the alloc with the same reusable prefix the reservation pinned |
| Two tests beyond the two pool tests: `attention_eviction_is_published_as_a_removal` and `deref_rejects_a_partial_release_before_mutating` | `publish_removed` gained a group filter and no existing test asserts a removal reaches the sink; and the release-is-total precondition is new, so it needs a test that the rejection leaves the block table and the pool untouched. PR 2 extends the first with the shared-hash Mamba case |
| The pool's second-group test constant is `SECOND`, not `MAMBA` | no Mamba concept exists in this PR |

Result: 637 tests pass (`main`'s suite plus the four added here), clippy clean, `cargo fmt --check` clean, and `VllmKvManager`'s `pub(crate)` / `pub(super)` signature list is byte-identical to `main`.

## PR mechanics

- Conventional-commit title, `Summary` and `Validation` sections in the description, `git commit -s` (DCO), per [AGENTS.md](AGENTS.md).
- Say explicitly in `Summary` that this is refactor-only, list the three invariants above, and point at PR 2 as the behavioral follow-up.
- Full CI runs only after a maintainer comments `/ok to test <sha>`.

</details>

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12189" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Improvements**
  * Improved KV-cache management across different cache groups.
  * Prevented identical cached sequences from being incorrectly shared between separate groups.
  * Improved prefix-cache reuse and block allocation for requests using multiple cache groups.
  * Enhanced cache lifecycle handling, including promotion, release, eviction, and restoration of cached data.
  * Improved cache event reporting and handling of partial request data for more consistent behavior.

### DIFF
--- a/lib/mocker/src/kv_manager/vllm_backend.rs
+++ b/lib/mocker/src/kv_manager/vllm_backend.rs
@@ -224,9 +224,23 @@ impl VllmKvManager {
         self.process_use(&alloc, None)
     }
 
+    /// Release a request's blocks from every group.
+    ///
+    /// A release always covers the request's whole footprint: each `Deref` signal
+    /// derives its block list from the sequence's allocated tokens. Groups whose
+    /// table is not parallel to the attention table lean on that, dropping all of
+    /// a request's state rather than trimming it. Trimming a live request is a
+    /// real vLLM operation — `remove_skipped_blocks` — that this mocker does not
+    /// model yet, so a partial release stops here rather than guessing.
     pub(super) fn deref_for_request(&mut self, request_id: Uuid, blocks: &[UniqueBlock]) {
-        let (pool, attention) = self.attention_mut();
-        attention.deref(pool, request_id, blocks);
+        let Self { pool, groups, .. } = self;
+        for group in groups.iter_mut() {
+            group.release(pool, request_id, blocks);
+        }
+        assert!(
+            !self.attention().holds_request(request_id),
+            "releasing part of a request's table is not modeled yet"
+        );
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -781,6 +795,20 @@ mod tests {
             .pool
             .prefix_hit(GroupedHash::new(manager.attention_id(), hash))
             .is_some()
+    }
+
+    /// Groups other than attention drop a request's whole state on release, so a
+    /// release that leaves the request holding blocks has to stop rather than
+    /// half-apply. Trimming a live request is vLLM's `remove_skipped_blocks`,
+    /// which the mocker does not model.
+    #[test]
+    #[should_panic(expected = "not modeled")]
+    fn releasing_part_of_a_request_table_is_rejected() {
+        let mut manager = attention_manager(4);
+        let owner = Uuid::from_u128(1);
+        ready(use_full(&mut manager, owner, &[7, 8], 0));
+
+        manager.deref_for_request(owner, &[UniqueBlock::FullBlock(8)]);
     }
 
     #[test]
--- a/lib/mocker/src/kv_manager/vllm_backend.rs
+++ b/lib/mocker/src/kv_manager/vllm_backend.rs
@@ -3,8 +3,11 @@
 
 //! vLLM G1 manager over a minimal physical block-pool model.
 //!
-//! The manager owns request block tables and KV-event metadata. The pool owns
-//! physical occupancy, duplicate copies, prefix pins, and LRU eviction.
+//! [`VllmKvManager`] is the coordinator. It owns one manager per KV cache group,
+//! sizes every allocation as a single reservation against the shared pool, and
+//! reconciles per-group prefix hits the way vLLM's `HybridKVCacheCoordinator`
+//! does. Per-group request block tables live in the group managers; the pool
+//! owns physical occupancy, duplicate copies, prefix pins, and LRU eviction.
 
 use dynamo_kv_router::protocols::{
     ExternalSequenceBlockHash, KvCacheEvent, KvCacheEventData, KvCacheRemoveData, KvCacheStoreData,
@@ -12,38 +15,17 @@ use dynamo_kv_router::protocols::{
 };
 use dynamo_tokens::blocks::UniqueBlock;
 use dynamo_tokens::{BlockHash, SequenceHash};
-use rustc_hash::{FxHashMap, FxHashSet};
 use uuid::Uuid;
 
-use crate::cache::vllm_block_pool::{BlockCopyId, BlockReservation, ReserveOutcome, VllmBlockPool};
+use crate::cache::vllm_block_pool::{BlockReservation, GroupedHash, ReserveOutcome, VllmBlockPool};
 use crate::common::kv_cache_trace;
 use crate::common::protocols::{KvEventPublishers, PrefillCost};
 use crate::common::sequence::ActiveSequence;
 
-struct PendingStore {
-    parent_hash: Option<SequenceHash>,
-    local_hash: Option<BlockHash>,
-    token_ids: Option<Vec<u32>>,
-}
-
-struct FullBlock {
-    copy: BlockCopyId,
-    hash: SequenceHash,
-    /// Whether a freshly allocated full block still needs to become cache-visible.
-    pending_cache: bool,
-    /// Event metadata retained until `pending_cache` is finalized.
-    pending_store: Option<PendingStore>,
-}
-
-enum OwnedBlock {
-    Partial { copy: BlockCopyId, uuid: Uuid },
-    Full(FullBlock),
-}
-
-struct StoredBlock {
-    hash: SequenceHash,
-    metadata: PendingStore,
-}
+use super::vllm_groups::{
+    ATTENTION_GROUP, CacheHit, FullAttentionGroup, GroupAllocation, GroupManager, PendingStore,
+    StoredBlock,
+};
 
 struct StoreGroup {
     parent_hash: Option<SequenceHash>,
@@ -82,7 +64,6 @@ impl StoreGroup {
         }
     }
 }
-
 pub(crate) struct NativeDecodeBlockReservation {
     pool: BlockReservation,
 }
@@ -97,11 +78,16 @@ pub(crate) struct NativeDestinationReservation {
     request_id: Uuid,
     pool: BlockReservation,
     layout: Option<VllmBlockLayout>,
+    /// Attention blocks the reservation covers, excluding other groups' share.
+    attention_fresh: usize,
+    /// Leading attention prefix pins, which precede other groups' pins in the
+    /// reservation.
+    attention_prefix: usize,
 }
 
 impl NativeDestinationReservation {
     pub(crate) fn transferable_prompt_tokens(&self, block_size: usize) -> usize {
-        self.pool.fresh_len().saturating_mul(block_size)
+        self.attention_fresh.saturating_mul(block_size)
     }
 
     #[cfg(test)]
@@ -140,8 +126,9 @@ impl VllmBlockLayout {
 
 pub(crate) struct VllmKvManager {
     pool: VllmBlockPool,
-    request_blocks: FxHashMap<Uuid, Vec<OwnedBlock>>,
-    partial_uuids: FxHashSet<Uuid>,
+    /// Group 0 is always the main attention group; a hybrid model's
+    /// non-attention state groups follow it.
+    groups: Vec<GroupManager>,
     block_size: usize,
     enable_prefix_caching: bool,
     kv_event_publishers: KvEventPublishers,
@@ -163,14 +150,22 @@ impl VllmKvManager {
         }
         Self {
             pool: VllmBlockPool::new(max_capacity),
-            request_blocks: FxHashMap::default(),
-            partial_uuids: FxHashSet::default(),
+            groups: vec![GroupManager::FullAttention(FullAttentionGroup::new(
+                ATTENTION_GROUP,
+                block_size,
+                enable_prefix_caching,
+            ))],
             block_size,
             enable_prefix_caching,
             kv_event_publishers,
             dp_rank,
             next_event_id: 0,
         }
+    }
+
+    fn attention(&self) -> &FullAttentionGroup {
+        let GroupManager::FullAttention(group) = &self.groups[0];
+        group
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -183,19 +178,22 @@ impl VllmKvManager {
         parent: Option<&UniqueBlock>,
         reusable_prefix_blocks: usize,
     ) -> VllmAcquire<usize> {
-        self.process_use(
+        let alloc = GroupAllocation {
             request_id,
             blocks,
             local_hashes,
             token_ids,
             parent,
             reusable_prefix_blocks,
-            None,
-        )
+            cache_fresh: false,
+        };
+        self.process_use(&alloc, None)
     }
 
     pub(super) fn deref_for_request(&mut self, request_id: Uuid, blocks: &[UniqueBlock]) {
-        self.process_deref(request_id, blocks);
+        let Self { pool, groups, .. } = self;
+        let GroupManager::FullAttention(attention) = &mut groups[0];
+        attention.deref(pool, request_id, blocks);
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -208,7 +206,24 @@ impl VllmKvManager {
         local_hash: Option<BlockHash>,
         token_ids: Option<Vec<u32>>,
     ) {
-        self.process_promote(request_id, uuid, hash, parent_hash, local_hash, token_ids);
+        let materialize_store_events = self.materialize_store_events();
+        let Self { pool, groups, .. } = self;
+        let GroupManager::FullAttention(attention) = &mut groups[0];
+        let stored = attention.promote(
+            pool,
+            request_id,
+            uuid,
+            hash,
+            PendingStore {
+                parent_hash,
+                local_hash,
+                token_ids,
+            },
+            materialize_store_events,
+        );
+        if let Some(stored) = stored {
+            self.publish_store_sequence(vec![Some(stored)]);
+        }
     }
 
     /// Publish full blocks completed by this scheduling decision.
@@ -238,43 +253,17 @@ impl VllmKvManager {
         }
 
         let materialize_store_events = self.materialize_store_events();
-        let Some(blocks) = self.request_blocks.get_mut(&request_id) else {
-            panic!("request {request_id} owns no block table")
+        let stores = {
+            let Self { pool, groups, .. } = self;
+            let GroupManager::FullAttention(attention) = &mut groups[0];
+            attention.finalize_computed_prefix(
+                pool,
+                request_id,
+                first_new_block,
+                completed_blocks,
+                materialize_store_events,
+            )
         };
-        let completed_blocks = completed_blocks.min(blocks.len());
-        if first_new_block >= completed_blocks {
-            return;
-        }
-        let mut stores = materialize_store_events
-            .then(|| Vec::with_capacity(completed_blocks - first_new_block));
-        for block in &mut blocks[first_new_block..completed_blocks] {
-            match block {
-                OwnedBlock::Full(full) => {
-                    if !full.pending_cache {
-                        if let Some(stores) = &mut stores {
-                            stores.push(None);
-                        }
-                        continue;
-                    }
-                    full.pending_cache = false;
-                    let metadata = full.pending_store.take();
-                    let became_visible = self.pool.cache_private(full.copy, full.hash);
-                    if let Some(stores) = &mut stores {
-                        stores.push(became_visible.then(|| {
-                            StoredBlock {
-                                hash: full.hash,
-                                metadata: metadata.expect(
-                                    "materialized pending store must retain event metadata",
-                                ),
-                            }
-                        }));
-                    } else {
-                        debug_assert!(metadata.is_none());
-                    }
-                }
-                OwnedBlock::Partial { .. } => break,
-            }
-        }
         if let Some(stores) = stores {
             self.publish_store_sequence(stores);
         }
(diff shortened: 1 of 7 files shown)


lib/mocker/src/kv_manager/vllm_backend.rs
```diff
@@ -471,270 +481,62 @@ impl VllmKvManager {
...
+        let mut fresh = 0;
+        for group in &self.groups {
+            prefix.extend(group.prefix_keys(alloc));
+            fresh += group.fresh_blocks(alloc);
+        }
+        (prefix, fresh)
+    }
+
+    /// Commit every group against the shared reservation, publishing only the
+    /// main attention group's store events.
```
dreamtalen, line 507: Wondering whether publishing only the main attention group’s store events is aligned with upstream vLLM. Do we also need to emit events for the other cache groups?
cc @PeaBrane, would events of other cache groups be useful to the router?

lib/mocker/src/kv_manager/vllm_backend.rs
```diff
@@ -183,19 +205,34 @@ impl VllmKvManager {
...
-            None,
-        )
+            cache_fresh: false,
+        };
+        self.process_use(&alloc, None)
     }
 
+    /// Release a request's blocks from every group.
     pub(super) fn deref_for_request(&mut self, request_id: Uuid, blocks: &[UniqueBlock]) {
-        self.process_deref(request_id, blocks);
```
dreamtalen, line 198: This appears to be a behavior change even for the attention-only path: the previous manager supported suffix releases, while this now requires whole-request release. Could we run a replay-parity test against main—perhaps using dynamo-kv-replay-parity—to confirm that current scheduler flows do not regress?

lib/mocker/src/kv_manager/vllm_backend.rs
nealvaidya: This is to align with our current router behavior -- right now, vllm publishes events for all groups, and on the router side we filter out all of the non-attention groups when ingesting the zmq events. Since we're bypassing zmq in the mocker and pushing directly to the router, we're just don't publish events for the non main group. So while the behavior right now matches what the real system does, there is potential for divergence in the future if we do start using the events from non-main groups.

lib/mocker/src/kv_manager/vllm_backend.rs
nealvaidya: Sure, I can run the parity test. From offline code analysis, the suffix release path was never used -- all of the calling sites strictly used whole-request release, so I figured it was an ok simplification to make since the semantics of suffix release are very different for other kinds of groups. But lmk if you think it needs to be re-added

PR #12195, opened 2026-07-27
fix(operator): preserve runtime versions across updates

sttts:
## Summary

- ratchet `runtimeVersionOverride` admission across DGDR, DGD, and DCD updates so unchanged legacy images remain updateable
- require an override when an update introduces or changes a non-semver runtime image
- allow an override-only DGDR repair while the remaining spec stays immutable
- make the DGDR controller apply the request-level override to profiler output before persistence and again before DGD creation
- cover v1alpha1 and v1beta1 source paths, current profiler output, and persisted legacy profiler output

This is a stacked fix for #10494 and targets `tmonty12/dyn-940-runtime-version`.

Linear: [DYN-940](https://linear.app/nvidia/issue/DYN-940/autoscaling-with-planner-grove)

## Root cause

The new create-time runtime-version requirement was applied inconsistently on updates: full revalidation trapped unchanged legacy resources, while the selected profiler image remained responsible for propagating a field that older profiler versions do not understand.

## User impact

Existing resources with unchanged custom image tags remain editable, but changing to a non-semver image still requires an explicit runtime version. DGDRs using older profiler images no longer produce DGD manifests that fail admission after the request itself was accepted.

## Validation

- `gofmt` on all changed Go files
- repository pre-commit hooks passed for both commits
- `git diff --check`
- GPG signatures verified for both commits
- tests not run, per review constraint

### DIFF
--- a/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
@@ -462,22 +462,20 @@ func (r *DynamoGraphDeploymentRequestReconciler) Reconcile(ctx context.Context, 
 		return ctrl.Result{}, nil
 	}
 
-	// Check for spec changes (immutability enforcement)
+	// Observe spec changes admitted by the validation webhook before resuming the phase machine.
 	if dgdr.Status.ObservedGeneration > 0 && dgdr.Status.ObservedGeneration != dgdr.Generation {
-		// Spec changed after initial processing
 		if dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseProfiling || dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseDeploying ||
 			dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseReady || dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseDeployed {
-			logger.Info("Spec change detected in immutable phase",
+			logger.Info("Observing admitted spec repair in immutable phase",
 				"phase", dgdr.Status.Phase,
 				"observedGeneration", dgdr.Status.ObservedGeneration,
 				"currentGeneration", dgdr.Generation)
 
-			r.Recorder.Event(dgdr, corev1.EventTypeWarning, nvidiacomv1beta1.EventReasonSpecChangeRejected,
-				fmt.Sprintf(MessageSpecChangeRejected, dgdr.Status.Phase))
-
-			// Keep the old observedGeneration to continue rejecting changes
-			// No phase transition - stay in current phase with old spec
-			return ctrl.Result{}, nil
+			dgdr.Status.ObservedGeneration = dgdr.Generation
+			if err := r.Status().Update(ctx, dgdr); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to observe admitted DGDR spec repair: %w", err)
+			}
+			return ctrl.Result{Requeue: true}, nil
 		}
 	}
 	// Phase machine: handle different phases
--- a/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
@@ -955,6 +955,7 @@ func (r *DynamoGraphDeploymentRequestReconciler) createDGD(ctx context.Context, 
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to unmarshal generated deployment from annotation: %w", err)
 	}
+	applyDGDRRuntimeVersionOverride(dgdr, generatedDGD)
 
 	// Determine DGD name and namespace from generated deployment
 	dgdName := generatedDGD.Name
@@ -2052,6 +2053,7 @@ func (r *DynamoGraphDeploymentRequestReconciler) generateDGDSpec(ctx context.Con
 	if err != nil {
 		return nil, "", fmt.Errorf("failed to extract DGD from %s: %w", outputFile, err)
 	}
+	applyDGDRRuntimeVersionOverride(dgdr, dgd)
 
 	// Override the profiler-generated name with a DGDR-scoped unique name.
 	// The profiler emits a static topology-derived name (e.g. "vllm-agg") which
@@ -2103,6 +2105,19 @@ func (r *DynamoGraphDeploymentRequestReconciler) generateDGDSpec(ctx context.Con
 	}
 	dgdr.ResourceVersion = apply.GetResourceVersion()
 	return profilingResults, dgd.Name, nil
+}
+
+// applyDGDRRuntimeVersionOverride makes the DGDR authoritative across profiler versions.
+func applyDGDRRuntimeVersionOverride(
+	dgdr *nvidiacomv1beta1.DynamoGraphDeploymentRequest,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) {
+	if dgdr.Spec.RuntimeVersionOverride == "" {
+		return
+	}
+	for i := range dgd.Spec.Components {
+		dgd.Spec.Components[i].RuntimeVersionOverride = dgdr.Spec.RuntimeVersionOverride
+	}
 }
 
 // encodeBetaDGDManifest returns JSON/YAML manifest bytes for a beta DGD.
--- a/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
@@ -54,6 +54,7 @@ import (
 	nvidiacomv1beta1 "github.com/ai-dynamo/dynamo/deploy/operator/api/v1beta1"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/consts"
 	commonController "github.com/ai-dynamo/dynamo/deploy/operator/internal/controller_common"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/dgdrutil"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/dynamo"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/gpu"
@@ -462,20 +463,51 @@ func (r *DynamoGraphDeploymentRequestReconciler) Reconcile(ctx context.Context, 
 		return ctrl.Result{}, nil
 	}
 
-	// Observe spec changes admitted by the validation webhook before resuming the phase machine.
+	currentSpecFingerprint, err := dgdrutil.SpecFingerprint(&dgdr.Spec)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Record the fingerprint for the spec associated with the observed generation.
+	if dgdr.Status.ObservedGeneration > 0 &&
+		dgdr.Status.ObservedGeneration == dgdr.Generation &&
+		dgdr.Status.ObservedSpecFingerprint != currentSpecFingerprint {
+		dgdr.Status.ObservedSpecFingerprint = currentSpecFingerprint
+		if err := r.Status().Update(ctx, dgdr); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to record observed DGDR spec fingerprint: %w", err)
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// Verify immutable-phase repairs against the last observed spec before resuming the phase machine.
 	if dgdr.Status.ObservedGeneration > 0 && dgdr.Status.ObservedGeneration != dgdr.Generation {
 		if dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseProfiling || dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseDeploying ||
 			dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseReady || dgdr.Status.Phase == nvidiacomv1beta1.DGDRPhaseDeployed {
-			logger.Info("Observing admitted spec repair in immutable phase",
+			repair, err := dgdrutil.IsRuntimeVersionOverrideRepair(&dgdr.Spec, dgdr.Status.ObservedSpecFingerprint)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if repair {
+				logger.Info("Observing verified runtime version override repair in immutable phase",
+					"phase", dgdr.Status.Phase,
+					"observedGeneration", dgdr.Status.ObservedGeneration,
+					"currentGeneration", dgdr.Generation)
+
+				dgdr.Status.ObservedGeneration = dgdr.Generation
+				dgdr.Status.ObservedSpecFingerprint = currentSpecFingerprint
+				if err := r.Status().Update(ctx, dgdr); err != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to observe DGDR runtime version override repair: %w", err)
+				}
+				return ctrl.Result{Requeue: true}, nil
+			}
+
+			logger.Info("Spec change detected in immutable phase",
 				"phase", dgdr.Status.Phase,
 				"observedGeneration", dgdr.Status.ObservedGeneration,
 				"currentGeneration", dgdr.Generation)
-
-			dgdr.Status.ObservedGeneration = dgdr.Generation
-			if err := r.Status().Update(ctx, dgdr); err != nil {
-				return ctrl.Result{}, fmt.Errorf("failed to observe admitted DGDR spec repair: %w", err)
-			}
-			return ctrl.Result{Requeue: true}, nil
+			r.Recorder.Event(dgdr, corev1.EventTypeWarning, nvidiacomv1beta1.EventReasonSpecChangeRejected,
+				fmt.Sprintf(MessageSpecChangeRejected, dgdr.Status.Phase))
+			return ctrl.Result{}, nil
 		}
 	}
 	// Phase machine: handle different phases
--- a/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
@@ -551,9 +551,6 @@ func (r *DynamoGraphDeploymentRequestReconciler) handlePendingPhase(ctx context.
 			return r.updatePhaseWithCondition(ctx, dgdr, nvidiacomv1beta1.DGDRPhaseFailed, nvidiacomv1beta1.ConditionTypeValidation, metav1.ConditionFalse, nvidiacomv1beta1.EventReasonValidationFailed, err.Error())
 		}
 
-		// Set observedGeneration to track the spec we're processing
-		dgdr.Status.ObservedGeneration = dgdr.Generation
-
 		dgdr.AddStatusCondition(metav1.Condition{
 			Type:               nvidiacomv1beta1.ConditionTypeValidation,
 			Status:             metav1.ConditionTrue,
@@ -2433,8 +2430,10 @@ func setSucceededCondition(dgdr *nvidiacomv1beta1.DynamoGraphDeploymentRequest, 
 func (r *DynamoGraphDeploymentRequestReconciler) updatePhase(ctx context.Context, dgdr *nvidiacomv1beta1.DynamoGraphDeploymentRequest, phase nvidiacomv1beta1.DGDRPhase, message string) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 	logger.Info("Updating DGDR phase", "name", dgdr.Name, "phase", phase, "message", message)
+	if err := observeCurrentDGDRSpec(dgdr); err != nil {
+		return ctrl.Result{}, err
+	}
 	dgdr.Status.Phase = phase
-	dgdr.Status.ObservedGeneration = dgdr.Generation
 	setSucceededCondition(dgdr, phase)
 	if err := r.Status().Update(ctx, dgdr); err != nil {
 		return ctrl.Result{}, err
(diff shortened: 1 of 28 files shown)


deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
```diff
@@ -2105,6 +2107,19 @@ func (r *DynamoGraphDeploymentRequestReconciler) generateDGDSpec(ctx context.Con
...
+// applyDGDRRuntimeVersionOverride makes the DGDR authoritative across profiler versions.
+func applyDGDRRuntimeVersionOverride(
+	dgdr *nvidiacomv1beta1.DynamoGraphDeploymentRequest,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) {
+	if dgdr.Spec.RuntimeVersionOverride == "" {
+		return
+	}
+	for i := range dgd.Spec.Components {
+		dgd.Spec.Components[i].RuntimeVersionOverride = dgdr.Spec.RuntimeVersionOverride
```
sttts, line 2151: don't we overrule the profiler if it has an opinion?

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
```diff
@@ -86,15 +86,30 @@ func (v *DynamoComponentDeploymentValidator) validateUpdate(
...
 	allErrs := validation.validateDynamoComponentDeployment(newDCD)
-	alpha, err := alphaDynamoComponentDeploymentForValidation(newDCD)
+	newAlpha, err := alphaDynamoComponentDeploymentForValidation(newDCD)
 	if err != nil {
 		return nil, fmt.Errorf("cannot validate preserved v1alpha1 DynamoComponentDeployment fields: %w", err)
 	}
-	allErrs = append(allErrs, validation.validateDynamoComponentDeploymentV1alpha1(alpha)...)
+	allErrs = append(allErrs, validation.validateDynamoComponentDeploymentV1alpha1(newAlpha)...)
+
+	validation.runtimeVersionSource = runtimeVersionSource
```
sttts, line 92:
```suggestion
    // one line comment saying what is done here in this block
	validation.runtimeVersionSource = runtimeVersionSource
```

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
```diff
@@ -86,15 +86,30 @@ func (v *DynamoComponentDeploymentValidator) validateUpdate(
...
+		}
+		if err := runtimeVersionOverrideUpdateErrorV1Alpha1(
+			&newAlpha.Spec.DynamoComponentDeploymentSharedSpec,
+			&oldAlpha.Spec.DynamoComponentDeploymentSharedSpec,
+			field.NewPath("spec"),
+		); err != nil {
+			allErrs = append(allErrs, err)
+		}
+	}
 	allErrs = append(allErrs, validation.validateDynamoComponentDeploymentUpdate(newDCD, oldDCD)...)
```
sttts, line 105: empty line above because the block is a semantical unit.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -107,14 +107,40 @@ func (v *DynamoGraphDeploymentValidator) ValidateUpdate(
 	newDGD *nvidiacomv1beta1.DynamoGraphDeployment,
 	userInfo *authenticationv1.UserInfo,
 	operatorPrincipal string,
+) (admission.Warnings, error) {
+	return v.validateUpdate(ctx, oldDGD, newDGD, userInfo, operatorPrincipal, runtimeVersionSourceV1Beta1)
+}
+
+func (v *DynamoGraphDeploymentValidator) validateUpdate(
```
sttts: why this helper? Can't we add the runtimeVersionSource directly?

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest.go
```diff
@@ -173,3 +176,18 @@ func dgdrRuntimeVersionOverrideRequired(spec *nvidiacomv1beta1.DynamoGraphDeploy
 	_, err := runtimeversion.ParseImageVersion(spec.Image)
 	return err != nil
 }
+
+func isDGDRRuntimeVersionOverrideRepair(
```
sttts, line 181: go doc

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
```diff
@@ -145,19 +144,35 @@ func runtimeVersionOverrideErrorV1Alpha1(
...
-	if spec.ExtraPodSpec != nil && spec.ExtraPodSpec.MainContainer != nil {
-		image = spec.ExtraPodSpec.MainContainer.Image
-	}
-	if image == "" {
-		return field.Required(fldPath.Child("extraPodSpec", "mainContainer", "image"), "is required")
+	image, imagePath := runtimeVersionImageAndPathV1Alpha1(spec, fldPath)
+	return runtimeVersionError(image, spec.RuntimeVersionOverride, imagePath, fldPath.Child("runtimeVersionOverride"))
+}
+
+func runtimeVersionOverrideUpdateErrorV1Alpha1(
```
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
```diff
@@ -145,19 +144,35 @@ func runtimeVersionOverrideErrorV1Alpha1(
...
+	oldImage, _ := runtimeVersionImageAndPathV1Alpha1(oldSpec, fldPath)
+	if newImage == oldImage && newSpec.RuntimeVersionOverride == oldSpec.RuntimeVersionOverride {
 		return nil
 	}
-	if _, err := runtimeversion.ParseImageVersion(image); err != nil {
-		return field.Required(overridePath, "is required when the specified main container image has no parseable semantic-version tag")
+	return err
+}
+
+func runtimeVersionImageAndPathV1Alpha1(
```
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
```diff
@@ -462,19 +467,52 @@ func runtimeVersionOverrideError(
...
 		if index := containerIndexByName(spec.PodTemplate.Spec.Containers, consts.MainContainerName); index >= 0 {
-			image = spec.PodTemplate.Spec.Containers[index].Image
 			imagePath = imagePath.Index(index).Child("image")
+			return spec.PodTemplate.Spec.Containers[index].Image, imagePath
 		}
 	}
+	return "", imagePath
+}
+
+func runtimeVersionError(
```
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
```diff
@@ -462,19 +467,52 @@ func runtimeVersionOverrideError(
...
+	}
+
+	oldImage, _ := runtimeVersionImageAndPath(oldSpec, fldPath)
+	if newImage == oldImage && newSpec.RuntimeVersionOverride == oldSpec.RuntimeVersionOverride {
+		return nil
+	}
+	return err
+}
+
+func runtimeVersionImageAndPath(
```
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
```diff
@@ -462,19 +467,52 @@ func runtimeVersionOverrideError(
 	spec *nvidiacomv1beta1.DynamoComponentDeploymentSharedSpec,
 	fldPath *field.Path,
 ) *field.Error {
-	overridePath := fldPath.Child("runtimeVersionOverride")
-	image := ""
+	image, imagePath := runtimeVersionImageAndPath(spec, fldPath)
+	return runtimeVersionError(image, spec.RuntimeVersionOverride, imagePath, fldPath.Child("runtimeVersionOverride"))
+}
+
+func runtimeVersionOverrideUpdateError(
```
sttts: these non-method funcs make me nervous. We are breaking the structural validation pattern.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
sttts: Done in 3da3d88b88: added a one-line comment explaining that source-version runtime validation is re-enabled for the old/new ratchet.

deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
sttts: Yes, intentionally. `DGDR.spec.runtimeVersionOverride` is defined as authoritative for every generated component, and the current Python profiler already overwrites component values with it. This Go overlay preserves the same contract for older profilers and persisted output; a conflicting profiler value is subordinate to the explicit request value.

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest.go
sttts: Done in 3da3d88b88: added Go doc describing the override-only repair predicate.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
sttts: Done in 3da3d88b88: removed the extra helper and added `runtimeVersionSource` directly to `ValidateUpdate`. I aligned the DCD validator the same way.

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
sttts: Agreed. Refactored in 3da3d88b88: the runtime checks now follow the structural validation pattern as methods on `sharedValidation` and return `field.ErrorList`.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
sttts: Done in 3da3d88b88: added the blank line before the shared update-validation block.

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
sttts: Agreed. Refactored in 3da3d88b88: the update ratchet is now a `sharedValidation` method returning `field.ErrorList`, with the source-version wrapper remaining structural.

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
sttts: Agreed. Refactored in 3da3d88b88: image/path extraction is now part of the `sharedValidation` method set rather than a free function.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
sttts: Agreed. Refactored in 3da3d88b88: the shared runtime validation core is method-based and returns `field.ErrorList`.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
sttts: Agreed. Refactored in 3da3d88b88: the beta image/path lookup is now a `sharedValidation` method.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
sttts: Agreed. Refactored in 3da3d88b88: create and update runtime validation now use methods on `sharedValidation`, preserving the existing structural pattern.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
sttts: This is a general rule actually. We should have oneliner comment above every non-oneliner code block telling the story. And every semantical blocks with multiple lines needs to be separated by other code with empty lines (not tailing, not pending blocks of course). But this belongs into the operator AGENTS.md.

deploy/operator/internal/webhook/validation/shared_v1beta1.go
```diff
@@ -458,27 +459,87 @@ func (v *sharedValidation) validateExperimentalSpecUpdate(
...
+func (v *sharedValidation) validateRuntimeVersionUpdate(
+	newImage string,
+	newOverride string,
+	oldImage string,
+	oldOverride string,
+	imagePath *field.Path,
+	overridePath *field.Path,
+) field.ErrorList {
+	allErrs := v.validateRuntimeVersion(newImage, newOverride, imagePath, overridePath)
+	if len(allErrs) == 0 || newImage == "" {
```
sttts: not sure what this case means. Need short end of line comment on the return?

sttts: Reworked in ddac79a158 rather than adding an end-of-line comment: the opaque combined return is gone. The exact API-type update validators now state the cases directly under a story comment—missing images are always rejected, while only an unchanged legacy invalid image/override tuple is ratcheted. Pure lookup/comparison helpers were also moved to `shared_helpers.go` per `STRUCTURAL_VALIDATION.md`.

deploy/operator/internal/webhook/validation/dynamocomponentdeployment.go
sttts: Done in ddac79a158: added the general story-comment and semantic-block spacing rules to `deploy/operator/AGENTS.md`.

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest.go
```diff
@@ -149,14 +149,17 @@ func (v *dynamoGraphDeploymentRequestValidation) validateDynamoGraphDeploymentRe
...
 		allErrs = append(allErrs, field.Required(
 			fldPath.Child("runtimeVersionOverride"),
 			"is required when spec.image has no parseable semantic-version tag",
 		))
 	}
 
-	if isImmutableDGDRPhase(oldPhase) && !apiequality.Semantic.DeepEqual(newSpec, oldSpec) {
+	if isImmutableDGDRPhase(oldPhase) &&
+		!apiequality.Semantic.DeepEqual(newSpec, oldSpec) &&
+		!isDGDRRuntimeVersionOverrideRepair(newSpec, oldSpec) {
```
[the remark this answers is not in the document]
sttts: Fixed in 7bdbcfb322. The reconciler now advances status.observedGeneration for an admitted spec repair in an immutable phase and requeues before resuming the phase machine, so profiling/deployment monitoring no longer stops after the repair. The controller envtest now covers repairing a legacy custom-image DGDR and verifies that the new generation is observed.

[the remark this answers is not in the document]
sttts: Corrected in 576b3b3236; this supersedes the broader handling in 7bdbcfb322. The controller now persists the fingerprint of the spec associated with `status.observedGeneration`. Admission permits the immutable-phase repair only when the old spec still matches that controller-owned fingerprint, and the reconciler independently verifies that removing the new `runtimeVersionOverride` reproduces the same fingerprint before advancing `observedGeneration`. Any unrelated or admission-bypassed spec change remains blocked. Legacy objects must first be reconciled once to backfill the fingerprint.

julienmancuso (approved):

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest.go
```diff
@@ -173,3 +179,25 @@ func dgdrRuntimeVersionOverrideRequired(spec *nvidiacomv1beta1.DynamoGraphDeploy
...
+
+// isDGDRRuntimeVersionOverrideRepair reports whether an update only repairs a missing required override.
+func isDGDRRuntimeVersionOverrideRepair(
+	newSpec *nvidiacomv1beta1.DynamoGraphDeploymentRequestSpec,
+	oldSpec *nvidiacomv1beta1.DynamoGraphDeploymentRequestSpec,
+	observedSpecFingerprint string,
+) bool {
+	if !dgdrRuntimeVersionOverrideRequired(oldSpec) ||
+		newSpec.RuntimeVersionOverride == oldSpec.RuntimeVersionOverride ||
+		dgdrRuntimeVersionOverrideRequired(newSpec) {
```
tmonty12, line 187: **[P1] This predicate makes the intended repair unreachable.** `dgdrRuntimeVersionOverrideRequired(newSpec)` tests whether the image tag is non-semver, not whether the override is missing. For the repair case—same `:custom` image, adding `runtimeVersionOverride: 1.2.3`—it returns true, so this helper returns false and immutable-phase admission rejects the patch. The controller's new fingerprint/acknowledgement path can therefore never run. Remove this condition or negate it; the equality check below already proves that no field other than the override changed. Please add an immutable-phase admission test for a non-semver legacy request that adds only the override.

deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
```diff
@@ -462,21 +463,50 @@ func (r *DynamoGraphDeploymentRequestReconciler) Reconcile(ctx context.Context,
...
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if repair {
+				logger.Info("Observing verified runtime version override repair in immutable phase",
+					"phase", dgdr.Status.Phase,
+					"observedGeneration", dgdr.Status.ObservedGeneration,
+					"currentGeneration", dgdr.Generation)
+
+				dgdr.Status.ObservedGeneration = dgdr.Generation
```
tmonty12, line 501: **[P1] A verified repair acknowledges the generation without repairing generated artifacts.** In `Ready` with `autoApply: false`, `status.profilingResults.selectedConfig` was generated before the override existed and still contains non-semver component images without `runtimeVersionOverride`. The following Ready reconcile is a no-op, so exporting and applying that status manifest still fails DGD admission. If immutable-phase repair is supported, re-decode and re-encode both persisted forms—the selected-config status JSON and `AnnotationGeneratedDGDSpec`—before acknowledging the generation. Please add a lifecycle envtest that reaches Ready from legacy profiler output, adds the override, then applies the emitted selected config successfully.

docs/fern/components/profiler/profiler-guide.md
```diff
@@ -208,12 +208,13 @@ spec:
...
-> With Dynamo Operator 1.4.0 or later, an explicitly selected profiler based on
-> Dynamo 1.3.0 or earlier must use a `spec.image` tag that accurately identifies
-> its runtime with semantic versioning, such as `:1.3.0`. Older profilers
-> silently discard the DGDR-level `spec.runtimeVersionOverride` field and
-> therefore cannot propagate it into the generated DGD.
+> [!NOTE]
+> The DGDR-level `spec.runtimeVersionOverride` is authoritative for every
+> component in the generated DGD. The operator applies it after processing
+> profiler output and DGD overrides, including when a profiler based on Dynamo
+> 1.3.0 or earlier discards the field while parsing the DGDR. Set the override
```
tmonty12, line 215: **[P2] The linked compatibility documentation still says the opposite.** `docs/kubernetes/dgdr.md#profiler-image-version-compatibility` says pre-1.4 profilers cannot propagate `runtimeVersionOverride` and must use semver profiler-image tags. With controller-side post-processing, profiler version no longer determines whether the DGDR-level override reaches the generated DGD. Please remove or rewrite that section in the same PR so the link does not contradict this note.

deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
```diff
@@ -2105,6 +2137,19 @@ func (r *DynamoGraphDeploymentRequestReconciler) generateDGDSpec(ctx context.Con
 	return profilingResults, dgd.Name, nil
 }
 
+// applyDGDRRuntimeVersionOverride makes the DGDR authoritative across profiler versions.
```
tmonty12, line 2142: **Ownership question:** now that the controller applies this after parsing profiler output and DGD overrides, should the profiler's `apply_runtime_version_override()` logic from #12072 remain? For the operator-managed DGDR path it duplicates ownership of the final field and is no longer needed for legacy-profiler compatibility. I would make the controller the sole authoritative writer, unless directly running the profiler and consuming its DGD output without the operator is a supported interface; if it is, that separate contract should be documented.

tmonty12 (commented):
Inline review comments on the immutable repair path and runtimeVersionOverride ownership.

deploy/operator/internal/webhook/validation/dynamographdeploymentrequest.go
sttts: The repair is reachable as written: `dgdrRuntimeVersionOverrideRequired` returns false immediately when `RuntimeVersionOverride != ""`, so it is false for the new repaired spec. The immutable non-semver repair is already covered by `adding runtime version override repairs legacy custom image in immutable phase`; it seeds `profiler:custom` without an override and admits the override-only update. No predicate change needed.

deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
sttts: Fixed in 1691afce48. Before acknowledging the repaired generation, the reconciler now decodes, updates, and re-encodes both persisted forms: `status.profilingResults.selectedConfig` and `nvidia.com/generated-dgd-spec`. The lifecycle envtest now seeds legacy Ready artifacts, adds the override, verifies both repaired manifests, and creates the selected DGD through admission.

docs/fern/components/profiler/profiler-guide.md
sttts: This section is already rewritten at the current head (34aa14ca97): it says the DGDR override is authoritative, the operator reapplies it after profiler output and DGD overrides, and that this behavior applies across profiler versions, including profilers through 1.3.0 that discard the field while parsing. The linked text no longer requires semver tags as the legacy-profiler workaround.

deploy/operator/internal/controller/dynamographdeploymentrequest_controller.go
sttts: Keeping both is intentional because standalone profiler output is supported (`profile_sla.py` explicitly has a non-operator path). The profiler owns correctness when its generated DGD is consumed directly; the controller is authoritative for operator-managed DGDRs and reapplies the value after all overrides. Commit 1691afce48 documents that split explicitly in the profiler guide.

Issue #11225, opened 2026-07-05
DEP: Multi-DC KV-Aware Request Routing


Kaonael:
### Area

router

### Summary

Introduce a new KV-aware routing subsystem that operates across data centers. It consists of two components:
a Relay, which publishes each data center's available KV cache blocks and worker metrics, and a Global Router, which proxies incoming requests and decides which data center should serve each one based on that information. To keep the per-block cache index compact enough to share across every data center without shipping the KV cache blocks themselves, the subsystem introduces a new Cuckoo filter-based indexer.

### Motivation

Dynamo can already make KV-aware routing decisions within a single DGD (Dynamo Graph Deployment), preferring the worker whose cache already holds the longest matching prefix for an incoming prompt. That reasoning stops at the boundary of one DGD: there is no mechanism to tell whether a different DGD — potentially running in a different data center — is already holding a better-matching cache for the same prompt, or is simply less loaded and would serve the request faster.

As deployments grow beyond a single data center, for capacity, latency, or resilience reasons, traffic needs to be distributed not just within a DGD but across many DGDs running in different data centers. Without cross-data-center awareness, a request can land on a DGD whose cache is cold for that prompt while another DGD already has it hot, forcing an unnecessary full re-prefill; or it can land on a data center that is already GPU-saturated, adding queueing delay a less-loaded data center would not have incurred.

This DEP extends Dynamo's handling of KV cache events so that routing decisions can be made across a fleet of DGDs spread across multiple data centers, not only within one. The routing decision should weigh:

- **KV cache utilization/overlap** — how much of the incoming request's prompt is already cached in each candidate data center.
- **GPU load** — how loaded the workers in each candidate data center currently are.
- **Latency** — the network cost of reaching each candidate data center.

Combining these signals lets the system prefer a data center that is both cache-warm and has spare serving capacity, instead of optimizing for cache locality alone (which can overload a single "hot" data center) or for load alone (which discards useful cache locality and causes unnecessary re-prefill).

[...]

PR #12199, opened 2026-07-27
feat(llm): add pool-scoped KV DC Relay runtime

Kaonael:
## Summary

  This PR formalizes the Dynamo-local pool model and per-pool runtime ownership for the KV DC Relay.

  The central invariant is:

  > Each materializable serving endpoint maps to one atomic Dynamo `PoolId`. Each active pool owns one Relay actor, one CKF state, and one producer generation.

  Independent endpoints or deployments remain separate pools even when they serve the same canonical model. A canonical model may therefore appear in multiple pool descriptors.
  Comparing those pools and selecting between them remains consumer or global-router policy.

  ## Details

  ### Pool discovery and model projection

  - Projects Dynamo `ModelDeploymentCard`s into endpoint-centric pool memberships.
  - Requires each materializable endpoint to resolve to:
    - one `IndexerDomainId`;
    - one canonical base-model identity;
    - one unambiguous KV-state endpoint.
  - Derives `PoolId` from the resolved indexer domain and DC identity.
  - Normalizes canonical model and alias registrations.
  - Represents LoRA adapters explicitly while keeping them attached to the pool of their backing base model.
  - Treats multiple domains, multiple base models, conflicting request-facing names, and duplicate live `PoolId` ownership as fail-closed materialization conflicts.
  - Keeps invalid aliases and orphan adapter cards as soft discovery errors when the remaining endpoint projection is valid.
  - Reconciles discovery periodically, rebinds all scoped streams after a stream terminates, and opens configured namespace queries concurrently.
  - Suppresses duplicate discovery upserts before rebuilding the complete membership projection.

  ### Pool runtime and producer contract

  - Adds `PoolRegistry` as the owner of active pool actors, layout generations, reservations, and catalog publication.
  - Exposes the producer-side contract through `DcPoolCatalog` and `DcPoolDescriptor`.
  - Publishes the serving endpoint, `ProducerIdentity`, and normalized base/LoRA registrations for each active pool.
  - Keeps derived cross-pool model and alias lookup indexes out of the Relay contract.
  - Separates the backend-specific DRT instance identity from a random per-`KvDcRelay::start` Relay incarnation.
  - Uses the Relay incarnation and monotonic layout generation to prevent producer identity reuse across Relay and pool lifetimes.
  - Updates catalog entries incrementally and materializes their ordered representation lazily.
  - Allocates and initializes CKF state through `spawn_blocking`, outside the registry mutex.
  - Uses bounded concurrent CKF allocation, two-phase reservations, and RAII rollback so cancellation, shutdown, or allocation failure cannot leave stale reservations or publish
  incomplete pools.

  ### Lifecycle, recovery, and fault handling

  - Reuses Dynamo’s worker-query recovery model, with `(worker_id, dp_rank)` identifying a source and `SourceEpoch` fencing source replacement.
  - Refreshes registration-only changes in place when the pool binding remains unchanged.
  - Rebuilds the producer when the indexer domain, KV-state endpoint, or worker/rank topology changes.
  - Withdraws a pool from `DcPoolCatalog` before waiting for recovery and actor teardown.
  - Routes terminal actor and recovery failures through the same fenced-retirement path.
  - Makes actor fault publication cancellation-aware and drains faults concurrently during teardown.
  - Coalesces pending source faults by `(worker_id, dp_rank)`, retains the newest epoch and strongest action, and keeps the pending state bounded.
  - Gives producer-fencing faults priority over weaker recovery work; overflow also fails safe by fencing the producer.

  ## Out of scope

  This PR intentionally does not add:
  - WAN transport or a global consumer;
  - consumer-side model/alias lookup policy;

  1. `components/src/dynamo/kv_dc_relay/README.md`
     - The intended endpoint-scoped pool model.
  2. `lib/llm/src/kv_dc_relay/identity.rs`
     - Relay identity, producer catalog, canonical registrations, and LoRA targets.
  3. `lib/llm/src/kv_dc_relay/discovery.rs`
     - Projection from model cards into materializable endpoint memberships.
  4. `lib/llm/src/kv_dc_relay/pool_registry.rs`
     - Actor ownership, allocation reservations, catalog publication, and retirement.
  5. `lib/llm/src/kv_dc_relay/host.rs`
     - Discovery reconciliation, pool transitions, recovery, fencing, and fault coalescing.
  6. `lib/llm/src/kv_dc_relay/actor.rs` and `lib/llm/src/discovery/kv_source_membership.rs`
     - Producer identity construction, cancellation-aware fault delivery, and binding-relevant source membership.
     
  ## Validation
  - `cargo test -p dynamo-llm kv_dc_relay:: --lib`
    - 65 tests passed.
  - `cargo test -p dynamo-llm --features ckf-diagnostics kv_dc_relay:: --lib`
    - 66 tests passed.
  - `cargo test -p dynamo-llm kv_source_membership::tests --lib`
    - 6 tests passed.
  - `cargo clippy -p dynamo-llm --all-targets --features ckf-diagnostics -- -D warnings`
  - `git diff --check origin/main...HEAD`

  ## Related Issues

  - Relates to #11225

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12199" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added pool-based KV cache relay management with contributor-aware ownership and lifecycle controls.
  * Added model identity, alias, routing, and pool-binding support.
  * Added configurable discovery across namespaces and endpoint prefixes.
  * Added publication subscriptions with incremental updates and snapshot recovery.
  * Added options to track publication activity without materializing payloads.

* **Bug Fixes**
  * Improved handling of conflicting memberships, stale sources, publication gaps, and departed endpoints.
  * Added validation for model identities, aliases, and format metadata.

* **Documentation**
  * Added an overview of KV cache relay pools and their runtime responsibilities.


lib/llm/src/kv_dc_relay/resolution.rs
```diff
@@ -47,10 +47,12 @@ pub(crate) struct EndpointLocator {
 }
 
 impl EndpointLocator {
+    #[cfg(test)]
     pub(crate) fn new(dc_id: DcId, endpoint_id: EndpointId) -> Self {
         Self { dc_id, endpoint_id }
     }
 
+    #[cfg(test)]
```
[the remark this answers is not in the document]
Kaonael: Done

lib/llm/src/kv_dc_relay/identity.rs
```diff
@@ -0,0 +1,322 @@
...
+    #[test]
+    fn pool_identity_sources_report_derived_components() {
+        let explicit = PoolIdentitySources::from_pool(pool(IdentitySource::Explicit));
+        assert_eq!(explicit.routing_scope(), IdentitySource::Explicit);
+        assert!(explicit.is_derived());
+
+        let derived = PoolIdentitySources::from_pool(pool(IdentitySource::DefaultDerived));
+        assert_eq!(derived.routing_scope(), IdentitySource::DefaultDerived);
+        assert!(derived.is_derived());
+    }
```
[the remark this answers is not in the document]
Kaonael: Done

lib/llm/src/kv_dc_relay/discovery.rs
```diff
@@ -157,92 +350,416 @@ impl MembershipState {
...
-            .retain(|endpoint, _| endpoints.contains_key(endpoint));
-        self.previous = endpoints.clone();
-        DcMembershipView { endpoints }
+        self.domain_generations
+            .retain(|slot_id, _| domains.contains_key(slot_id));
+        self.warned_invalid_models = invalid_models;
+        self.warned_invalid_aliases = invalid_aliases;
+        let domains = Arc::new(domains);
+        self.previous = domains.clone();
+        DcMembershipView { domains }
```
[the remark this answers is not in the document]
Kaonael: Done

lib/llm/src/kv_dc_relay/discovery.rs
```diff
@@ -5,66 +5,252 @@ use std::collections::{HashMap, HashSet};
...
+                self.watch_all
+                    || self.namespaces.iter().any(|namespace| {
+                        prefix == namespace
+                            || prefix
+                                .strip_prefix(namespace)
+                                .is_some_and(|suffix| suffix.starts_with('.'))
+                    }),
+                "KV DC Relay endpoint prefix {prefix} is outside the configured namespaces"
+            );
+        }
```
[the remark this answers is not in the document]
Kaonael: Done

lib/llm/src/kv_dc_relay/pool_registry.rs
```diff
@@ -0,0 +1,1393 @@
...
+                tracing::error!(%pool_id, "KV DC Relay pool actor fault channel closed unexpectedly");
+                cancel.cancel();
+            }
+            return;
+        };
+        let contributor_id = fault.contributor_id;
+        let route = routes.read().get(&contributor_id).cloned();
+        if let Some(route) = route {
+            match route.try_send(fault) {
+                Ok(()) => {}
```
hhzhang16: This queue contains source-scoped faults, not interchangeable duplicate notifications. If it is full, the dropped fault may carry `FenceAndRebuildProducer`, while the queued fault may not. That can leave a suspect producer generation active. Please make overflow fail safe: either fence/cancel the pool generation, or coalesce pending faults while preserving the strongest required action. Please also add a test showing that a queued non-fencing fault cannot suppress a later fencing fault.

lib/llm/src/kv_dc_relay/publication_hub.rs
```diff
@@ -0,0 +1,540 @@
...
+    }
+    let retirement = result
+        .as_deref()
+        .unwrap_or("publication hub stopped")
+        .to_string();
+    let mut current = state.lock();
+    current.ready = false;
+    current.last_error = result;
+    drop(current);
+    let _ = updates.send(PublicationUpdate::Retired(retirement));
```
hhzhang16: A terminal hub failure only changes hub-local state and sends `Retired`; it does not fence or withdraw the owning PoolEntry. `catalog_from_state` will therefore continue advertising this pool and its model bindings even though publication is permanently unavailable.
If the hub remains in this PR, terminal failure needs to notify the registry so the generation is fenced/withdrawn and contributors can rebind. Please add a test asserting that a terminal hub failure removes the generation from `watch_catalog`.

lib/llm/src/kv_dc_relay/pool_registry.rs
```diff
@@ -0,0 +1,1393 @@
...
+            drop(state);
+            let hub = match PublicationHub::start(
+                handle.clone(),
+                LaneLease::new(
+                    ConsumerInstanceId::new(self.process_incarnation),
+                    0,
+                    layout_generation,
+                ),
+                self.actor_config.publication_fanout_capacity,
+            )
```
hhzhang16: This eagerly creates a publication hub and full CKF mirror for every pool, although the subscription/catalog consumers are not wired yet. At the default sizing this is roughly 4 MiB of mirror state per pool, plus snapshot copying and continuous delta processing.
Since WAN consumption is explicitly deferred, please either start the hub lazily when a real consumer subscribes or move it into the consumer/WAN PR. If it remains eager, this PR needs a concrete current consumer and validation of the per-pool resource cost.

lib/llm/src/kv_dc_relay/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+This module owns the internal model used to aggregate KV-cache state within a data center. It does
+not decide whether a pool can accept a request.
+
+## Domain model
+
+A `PoolId` identifies one physical KV-cache compatibility domain in one data center. Compatible
+serving endpoints attach to that pool as independent contributors. The pool owns one serialized
+CKF actor, so worker and rank identities from different endpoints cannot overwrite each other's
+state.
```
hhzhang16: Not sure if I see the production staging path that makes distinct endpoints “compatible” here. The default routing scope is derived from endpoint identity, and within this repo I found no production code populating a shared `ModelDeploymentCard.indexer_identity`; the registry test demonstrates sharing by directly injecting the same PoolId.

lib/llm/src/kv_dc_relay/discovery.rs
```diff
@@ -157,92 +358,414 @@ impl MembershipState {
...
             let changed = self
                 .previous
-                .get(&endpoint)
+                .get(&slot_id)
                 .is_none_or(|previous| !same_membership(previous, &candidate));
-            let generation = self
-                .endpoint_generations
-                .entry(endpoint.clone())
-                .or_default();
+            let generation = self.domain_generations.entry(slot_id.clone()).or_default();
```
hhzhang16: This now preserves correctness across slot reappearance, but domain_generations and next_layout_generation retain every historical identity for the process lifetime. For a long-lived relay with endpoint churn, could a process-wide monotonic generation allocator provide non-reuse without permanent per-identity tombstones?

hhzhang16 (changes requested):

lib/llm/src/kv_dc_relay/pool_registry.rs
Kaonael: I guess now it's irrelevant - intermediate queue removed

lib/llm/src/kv_dc_relay/publication_hub.rs
Kaonael: PublicationHub now not exists after removing ContributorId model

lib/llm/src/kv_dc_relay/pool_registry.rs
Kaonael: Also irrelevant after removing ContributorId model

lib/llm/src/kv_dc_relay/README.md
Kaonael: Adopted to discussed model with independent pools

lib/llm/src/kv_dc_relay/discovery.rs
Kaonael: Yes, great idea, thank you!

PeaBrane:
/ok to test https://github.com/ai-dynamo/dynamo/pull/12199/commits/6e06e64ee5c90d18de70bf0d9de9233c61217fa6

lib/llm/src/kv_dc_relay/discovery.rs
```diff
@@ -157,92 +350,416 @@ impl MembershipState {
...
     let mut retry_delay = Duration::from_millis(100);
+    let mut watch_failures = 0u64;
+    let mut reconcile_failures = 0u64;
     loop {
         let stream_cancel = cancel.child_token();
-        let stream = discovery
-            .list_and_watch(DiscoveryQuery::AllModels, Some(stream_cancel.clone()))
-            .await;
-        let mut stream = match stream {
+        let streams = open_query_streams(&discovery, &queries, &stream_cancel).await;
```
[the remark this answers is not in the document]
PeaBrane: Addressed in the current head: each namespace watch is wrapped to emit a terminal `Ok(None)` before exhaustion. The outer membership loop treats that sentinel as a closure, breaks, cancels the remaining streams, and reopens all configured query watches.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -402,21 +415,42 @@ impl KvDcRelay {
             membership: Mutex::new(Some(membership)),
             supervisor: Mutex::new(Some(supervisor)),
             statuses,
+            pools,
         })
     }
 
+    pub async fn model_pool_bindings(&self) -> Vec<ModelPoolBinding> {
```
PeaBrane: Could we keep this PR’s producer-side contract to `DcPoolCatalog`/`DcPoolDescriptor` and defer the derived `model_pool_bindings*` and `model_alias_bindings` lookup APIs until the consumer exists? The Relay is the right place to normalize local `ModelDeploymentCard`s into pool advertisements tied to `ProducerIdentity`, but flattening those advertisements into model/alias indexes feels like consumer/global-router responsibility. Deferring that part would keep the producer contract minimal and avoid cementing a lookup shape before the actual consumer is wired.

PeaBrane (commented):
Suggestion on keeping the producer/consumer catalog boundary minimal.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -369,27 +375,34 @@ impl KvDcRelay {
...
-        .await?;
+        let membership =
+            DcMembershipWatch::start(component.drt().discovery(), discovery, cancel.clone())
+                .await?;
         let membership_rx = membership.subscribe();
         let statuses = Arc::new(RwLock::new(HashMap::new()));
         let dc_id: Arc<str> = Arc::from(dc_id);
         let process_incarnation = component.drt().connection_id();
+        let ckf_dc_id = stable_dc_id(dc_id.as_ref());
+        let pools = Arc::new(PoolRegistry::new(
```
PeaBrane, line 516: Could we make the Relay incarnation unique per `KvDcRelay::start`? Recreating the Relay on the same `Component` reuses this DRT connection ID, while the new `PoolRegistry` restarts `layout_generation` at 1 and catalog revision at 0. For the same `PoolId` and deterministic CKF format, the recreated CKF can therefore have the same `ProducerIdentity` as the retired one, so delayed old snapshots/deltas are indistinguishable. A per-start incarnation or process-global generation allocator would preserve the intended fencing boundary.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -676,9 +781,41 @@ async fn run_endpoint_slot(
             }
         }
 
+        if let (Some(active), Some(membership)) = (runtime.as_mut(), membership.as_ref())
+            && membership.is_materializable()
+            && active.registrations != membership.registrations
+        {
+            match pools
+                .replace_registrations(&mut active.attachment, membership.registrations.clone())
```
PeaBrane: Could we compute and compare `desired_binding` before refreshing registrations, and only update them in place when it still equals `active.binding`? A single MDC update can change both the canonical registration and indexer domain; this currently publishes the new model against the old `PoolId`/CKF before the later `binding_changed` path retires it. Recovery shutdown can extend the lifetime of that inconsistent catalog revision.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -964,37 +1129,52 @@ async fn start_endpoint_actor(
...
+async fn stop_endpoint_pool(active: EndpointPoolRuntime, pools: &PoolRegistry) {
     active.recovery.shutdown().await;
-    if let Err(error) = active.handle.shutdown().await {
-        tracing::warn!(%error, endpoint = %active.handle.scope.pool_binding.serving_endpoint().endpoint_id(), "Failed to drain KV DC Relay endpoint actor");
+    let pool_id = active.attachment.pool_id;
+    let endpoint = active.binding.kv_state_endpoint.clone();
+    // Keep this await inline: `detach` owns the actor fault receiver and must complete its
+    // actor round-trip before the endpoint-slot task exits. The host joins every slot during
+    // retirement and shutdown rather than aborting it.
+    if let Err(error) = pools.detach(active.attachment).await {
```
PeaBrane: Could the catalog generation be withdrawn before awaiting recovery shutdown? `detach()` performs the withdrawal, but it runs only after this await, so an endpoint departure, source loss, or binding replacement can leave the retiring pool advertised for the duration of the drain. A generation-checked withdraw/mark-unavailable step here would let graceful cleanup remain private without presenting the pool as active.

lib/llm/src/kv_dc_relay/pool_registry.rs
```diff
@@ -0,0 +1,733 @@
...
+            if entry.fenced {
+                return Ok(());
+            }
+            entry.fenced = true;
+            entry.cancel.cancel();
+            let handle = entry.handle.clone();
+            publish_catalog(&mut state, self.process_incarnation, &self.catalog_tx);
+            handle
+        };
+        handle.fence().await
```
PeaBrane: Could fencing be made cancellation-aware with respect to the bounded fault channel? The actor can block in `fault_tx.send(...).await` outside its outer fence select, while the endpoint task owns the sole receiver and does not begin detach/fault draining until this wait returns. Once the 16-entry channel refills, fencing and shutdown can hang indefinitely. Selecting the fault send against the fence token, or draining faults concurrently while fencing, would close that cycle.

lib/llm/src/kv_dc_relay/pool_registry.rs
```diff
@@ -0,0 +1,733 @@
...
+                existing.endpoint,
+                request.endpoint
+            );
+        }
+        validate_registrations(&request.registrations)?;
+
+        let layout_generation = allocate_layout_generation(&mut state)?;
+        let mut config = CkfConfig::new(self.actor_config.expected_unique_blocks);
+        config.publish_every_n_events = self.actor_config.publication_threshold;
+        let (handle, faults) = KvDcRelayHandle::spawn_with_publication_delay(
```
PeaBrane: This CKF construction occurs while the single registry mutex is held from the ownership check through catalog publication. At the default capacity it synchronously allocates roughly 4 MiB per pool, and each successive attach also rebuilds/clones the full catalog, serializing startup and producing approximately quadratic catalog copying across a burst of pools. Could we reserve the pool/generation under the lock, allocate outside it, then revalidate and commit—or otherwise avoid holding the global lock across CKF allocation?

lib/llm/src/kv_dc_relay/discovery.rs
```diff
@@ -157,115 +366,359 @@ impl MembershipState {
...
             tokio::select! {
                 _ = cancel.cancelled() => return,
                 event = stream.next() => match event {
-                    Some(Ok(event)) => {
+                    Some(Ok(Some(event))) => {
+                        watch_failures = 0;
+                        retry_delay = Duration::from_millis(100);
                         state.apply(event, &filter);
-                        sender.send_replace(state.view(&filter));
+                        publish_membership_if_changed(&sender, state.view(&filter));
```
PeaBrane: Could we avoid rebuilding the complete membership projection for no-op discovery upserts? `DcMembershipWatch::start` first seeds state with `list_queries`, while `list_and_watch` then replays the same initial inventory as individual `Added` events. Each replayed event reaches `state.view()`, which now re-hashes, clones, and sorts all retained cards, making initial synchronization effectively O(N²). Having `apply` report whether the stored card actually changed would remove the common duplicate work; batching resync bursts would help further.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -1167,48 +1354,113 @@ mod tests {
...
+            conflicts: Vec::new(),
         }
     }
 
-    #[test]
-    fn simultaneous_endpoints_cannot_own_one_pool() {
-        let domain_id = IndexerDomainId::new(
-            CacheSemanticsId::new([1; 16], IdentitySource::Explicit),
-            RoutingScopeId::new([2; 16], IdentitySource::Explicit),
+    #[tokio::test]
```
PeaBrane, line 1855: Could we add deterministic transition-level coverage around the new lifecycle state machine, beyond the current helper tests? The highest-value cases are: same-process Relay recreation yields a distinct producer identity; a simultaneous registration/binding change never publishes the new registration on the old pool; catalog withdrawal precedes recovery drain; and fencing remains live with a full fault queue. Those assertions would directly protect the producer-generation contract introduced here.

PeaBrane (commented):
Neutral architectural follow-up—not a blocking verdict. The common pattern in the remaining findings is that `PoolRegistry` state currently serves both as local actor ownership and as externally visible “active catalog” truth. A useful invariant would be: every advertised descriptor names one unique, currently active producer generation. Concretely, give each Relay start a unique incarnation, publish or refresh catalog state only after validating the binding transition, and withdraw a generation before draining it. Keep consumer routing indexes derived from the pool-scoped catalog as discussed separately. This adds explicit activation/withdrawal steps, but centralizes lifecycle consistency now, before consumers depend on the contract.

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Yes, make sense

lib/llm/src/kv_dc_relay/host.rs
Kaonael: wow, now I know a little bit more about `component.drt().connection_id()`, thank you.
I guess we can use for this `EventPublisher::publisher_id()`-inspired mechanics and generate random u64, wdyt?

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done with u64 random

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/pool_registry.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/pool_registry.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/discovery.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/host.rs
PeaBrane: Thanks — the original old-pool publication ordering is fixed. I think there is one small follow-up hole in the new deferral condition: `runtime_config_changed` suppresses registration refresh for any runtime-config change, while the source watcher only republishes when the derived `KvSourceMembershipView` changes. For example, if one MDC update changes an alias and `context_length`, the registration refresh is skipped, but `context_length` does not affect that source view, so no follow-up event necessarily wakes the active slot and the catalog can retain the old alias. Could we defer only for a binding-relevant change, or explicitly schedule a reconciliation after the skip? A regression test for alias plus an unrelated runtime-field change should cover it.

lib/llm/src/kv_dc_relay/host.rs
```diff
@@ -964,37 +1149,148 @@ async fn start_endpoint_actor(
...
+    receiver: &mut tokio::sync::mpsc::Receiver<U>,
+    pending: &mut VecDeque<U>,
+    future: impl Future<Output = T>,
+) -> T {
+    tokio::pin!(future);
+    loop {
+        tokio::select! {
+            result = &mut future => return result,
+            item = receiver.recv() => match item {
+                Some(item) => pending.push_back(item),
```
hhzhang16: Thanks for addressing the blocking/full fault-channel issue! I think moving the faults into this `VecDeque` leaves two related concerns:

1. This queue is unbounded. Nothing is dropped anymore, but while recovery is running, every incoming fault can be appended here. If recovery stalls or faults continue arriving faster than they are handled, memory can grow without bound.

2. The faults do not all have equal priority. For example, `FenceAndRebuildProducer` should supersede pending `RetryWorker`-style actions for the same producer/source; there is little value in completing or retaining recovery work for a producer that must be fenced and rebuilt so we don't want to defer the handling for fence faults behind the in-progress recovery and pending work.

Could we coalesce pending actions per producer/source while preserving the strongest action? For example:

- `Retry + Retry` → one `Retry`
- `Retry + Fence` → `Fence`
- `Fence + Retry` → `Fence`

Alternatively, could a fencing action immediately withdraw/cancel the current producer recovery?

Could we also add tests with a larger and more varied sequence of faults? In particular, it would be useful to verify that the pending state remains bounded, duplicate retries are coalesced, and a fencing fault supersedes weaker actions rather than merely being retained for eventual processing.

PeaBrane (approved):

PeaBrane:
/ok to test [6ab1c26](https://github.com/ai-dynamo/dynamo/pull/12199/commits/6ab1c261b7e75281bc27d23fc1d11b35d725fc07)

lib/llm/src/kv_dc_relay/host.rs
PeaBrane: I think it’s reasonable to defer the general coalescing work. These faults come from fairly exceptional allocation, protocol, or invariant failures. Normal CKF capacity pressure doesn’t use this path, and the actor/mailbox are already bounded and serialized. The current change at least ensures a later fence isn’t lost. Making it interrupt an in-flight recovery would also require thinking carefully about cancellation safety, so I’d prefer to leave that as follow-up hardening unless we have a concrete fault-burst scenario we need to support now.

Leaving the call up to you guys. A TODO or NOTE for deferral here is also fine

lib/llm/src/kv_dc_relay/host.rs
PeaBrane: Sorry, I should have double-checked the lifecycle assumption behind my earlier comment. On reflection, the DRT connection/instance ID seems like the right producer-replica identity if our contract is one Relay lifetime per DRT. Separate Relay replicas, even when watching the same pools, should normally have separate DRTs and instance IDs, similar to how the router tests model separate replicas.

The collision here comes specifically from the E2E test stopping and recreating a Relay inside the same DRT. If it is only test convenience, I think the test should use a fresh DRT and we should retain the DRT identity. If in-process Relay restart is an intended contract, then a per-start Relay incarnation works, although it should be named and documented as such. 

(In Rust, the existing test pattern is essentially let first_drt = shared_drt(store.path()).await; let second_drt = shared_drt(store.path()).await;, where both DRTs use the same discovery backend but receive distinct connection_id() values. Noting that we don't enforce DRT to be a process singleton, mainly for this testing sake ...)

Also ok with keeping the current u64 random behavior, though it is a bit semantically ambiguous

PeaBrane: Discussed offline, and probably safer to just use random u64 for now (the current shape)

Kaonael: Done in worker-inspired way

lib/llm/src/kv_dc_relay/host.rs
Kaonael: Done

lib/llm/src/kv_dc_relay/host.rs
Kaonael: @hhzhang16 nice catch! Please check latest [commit](https://github.com/ai-dynamo/dynamo/pull/12199/commits/e6ec1225126c91e60b5c6b5d20cc830ffa60a98a)

hhzhang16 (approved):

PR #12201, opened 2026-07-27
fix(tool-calling): Enforce `parallel_tool_calls: false` in chat completions

pskiran1:
## Overview:

`parallel_tool_calls: false` was accepted in the request schema but not
universally enforced. The existing enforcement relied on the generation-time
structural-tag path, which only activates under specific `tool_choice` and
guided-decoding configurations. Outside those configurations, the model could
emit multiple tool calls that were passed through unchanged, violating the
client's request.

This PR adds a post-parse fallback that caps the response to the first tool
call when `parallel_tool_calls` is `false`, covering both streaming and
non-streaming chat completion responses across all backends (vLLM, SGLang,
TRT-LLM). The approach mirrors [vLLM's `maybe_filter_parallel_tool_calls`](https://github.com/vllm-project/vllm/blob/91945b6e4ade361125837228179cee01e6573023/vllm/entrypoints/openai/utils.py#L22-L25).

Fixes DIS-2456.

## Details:

hree files changed:
**`lib/llm/src/protocols/openai.rs`**
- Added `parallel_tool_calls: Option<bool>` field to `ParsingOptions`.
- Added `with_parallel_tool_calls()` builder method to thread the value from
  the request into the aggregator.

**`lib/llm/src/protocols/openai/chat_completions/aggregator.rs`**
- After all tool-call parsing and finalization in `DeltaAggregator::apply`,
  if `parallel_tool_calls == Some(false)`, each choice's `tool_calls` is
  truncated to the first entry.
- Added unit test `test_parallel_tool_calls_false_caps_to_first_call`.

**`lib/llm/src/http/service/openai.rs`**
- In the chat completions handler, the request's `parallel_tool_calls` value
  is threaded into `ParsingOptions` (for non-streaming) and captured as a
  local boolean (for streaming).
- In the streaming loop, when enforcement is active, tool-call deltas
  with index > 0 are removed. If a delta becomes empty after filtering,
  `tool_calls` is set to `None` so the existing empty-chunk filter drops it.

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

- **New Features**
  - Added support for respecting the `parallel_tool_calls` setting when processing tool calls.
  - Requests that disable parallel tool calls now retain only the first tool call, including during streaming responses.
  - Requests that allow parallel tool calls continue to support multiple tool calls.

- **Bug Fixes**
  - Prevented additional tool-call entries from appearing when parallel tool calls are disabled.

- **Tests**
  - Added coverage for limiting streamed tool calls to a single entry.


rmccorm4:
/ok to test [9f5006f](https://github.com/ai-dynamo/dynamo/pull/12201/commits/9f5006f630d83ed41934360fea7e81683852d681)

rmccorm4 (approved):

PR #12204, opened 2026-07-27
fix(llm): GLM-5.2 chat template compatibility — argument normalization and tool_call recovery

peilii:
## Summary

GLM-5.2's Jinja chat template calls `.items()` on tool-call arguments, expecting a parsed dict rather than the OpenAI wire-schema JSON string. It also emits `<tool_call>` XML blocks that can be truncated at `max_tokens`. This PR fixes both, plus a speculative-prefill rendering inconsistency.

- **Argument mode detection** (`preprocessor/prompt.rs`): scan the Jinja template source at formatter construction for `.items()` patterns. Introduce `ToolArgumentsMode` enum and `detect_tool_arguments_mode`. Use a thread-local `ToolArgumentsModeGuard` (RAII, sync-only scope) so the mode is applied only during the synchronous `apply_template` call with no async boundary risk.

- **Argument normalization** (`preprocessor/prompt.rs`): `normalize_tool_call_arguments` parses `tool_calls[*].function.arguments` from JSON string to object in `NvCreateChatCompletionRequest::messages()` when `ParsedObject` mode is active. Non-object and malformed-JSON values are replaced with `{}` and warned (by length only — no raw argument content in logs).

- **Array template variants** (`preprocessor/prompt.rs`): `mdc_jinja_template_text` now handles HF configs that store `chat_template` as `[{name, template}, ...]` by scanning all variant strings, so `.items()` in any variant is detectable.

- **Non-streaming recovery** (`aggregator.rs`): extract the last incomplete `<tool_call>` block before parsing. If a second block is truncated after a complete first one, it is appended to `choice.text` after parsing (the parser drops it silently otherwise).

- **Streaming recovery** (`preprocessor.rs`): on `finish_reason=length`, use `rfind("<tool_call>")` to recover the last block without a matching `</tool_call>`. Avoids duplicating already-emitted complete blocks. Per-choice `ChoiceRecovery` buffers input text; recovered content is emitted as a synthetic chunk before the finish chunk.

- **Speculative prefill** (`preprocessor/speculative_prefill.rs`): propagate `tool_arguments_mode` into `SpeculativePrefillRequest::messages()` so the prefill render applies the same argument normalization as the main request, keeping KV prefix hashes consistent.

## Validation

- `cargo test -p dynamo-llm`: all tests pass, including 8 new unit tests covering `detect_tool_arguments_mode`, `normalize_tool_call_arguments` (object, non-object, malformed JSON, non-assistant messages), and `mdc_jinja_template_text` (tokenizer_config.json and chat_template.json paths)

### DIFF
--- a/lib/llm/src/preprocessor.rs
+++ b/lib/llm/src/preprocessor.rs
@@ -2973,18 +2973,15 @@ impl OpenAIPreprocessor {
                 p.template = Some(metrics);
             }
             // Buffer input content only for glm47 (truncation recovery).
-            if is_glm47 {
-                if let Some(data) = &a.data {
-                    let mut cr = choice_recovery_in.lock().expect("choice recovery poisoned");
-                    for choice in &data.inner.choices {
-                        if let Some(ChatCompletionMessageContent::Text(content)) =
-                            &choice.delta.content
-                        {
-                            cr.entry(choice.index)
-                                .or_default()
-                                .input_text
-                                .push_str(content);
-                        }
+            if is_glm47 && let Some(data) = &a.data {
+                let mut cr = choice_recovery_in.lock().expect("choice recovery poisoned");
+                for choice in &data.inner.choices {
+                    if let Some(ChatCompletionMessageContent::Text(content)) = &choice.delta.content
+                    {
+                        cr.entry(choice.index)
+                            .or_default()
+                            .input_text
+                            .push_str(content);
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -416,9 +435,40 @@ impl DeltaAggregator {
...
+                    && matches!(
+                        choice.finish_reason,
+                        Some(dynamo_protocols::types::FinishReason::Length)
+                    )
+                    && choice.text.contains("<tool_call>")
+                {
+                    tracing::warn!(
+                        parser,
+                        "glm47: partial <tool_call> returned as content on length finish \
+                         (TRT-LLM parity; raw markup in content)"
```
rmccorm4, line 452: let's remove these "trtllm-parity" from the log messages since this the frontend/parser is generic to all backends - it would confuse users
```suggestion
                        "glm47: partial <tool_call> returned as content on length finish"
```

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -416,9 +435,40 @@ impl DeltaAggregator {
...
+                // Recover any tail saved before parsing — the parser drops a truncated
+                // second block silently when an earlier complete block was already parsed.
+                if let Some(tail) = glm47_truncated_tail
+                    && !choice.text.contains(&tail)
+                {
+                    tracing::warn!(
+                        parser,
+                        tail_bytes = tail.len(),
+                        "glm47: truncated later <tool_call> appended as content \
+                         (TRT-LLM parity; raw markup in content)"
```
rmccorm4, line 465:
```suggestion
                        "glm47: truncated later <tool_call> appended as content"
```

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3024,59 @@ impl OpenAIPreprocessor {
...
+                            } else {
+                                None
+                            }
+                        });
+                        if let Some(recovered) = recovered {
+                            tracing::warn!(
+                                choice_index = choice.index,
+                                recovered_bytes = recovered.len(),
+                                "glm47 streaming: partial <tool_call> emitted as content \
+                                 on length finish (TRT-LLM parity; raw markup in content)"
```
rmccorm4, line 3056:
```suggestion
                                 on length finish"
```

rmccorm4 (commented):
Same as @KrishnanPrash feedback - please split up the argument normalization and the tool call recovery PRs. Argument normalization can probably be taken in easily. Tool call recovery could use some thorough review. Also most tool call recovery logic is in frontend-crates/ repo along with the specific parser logic, and not in this repo - so that would dictate splitting it up anyways

PR #12206, opened 2026-07-27
feat(sidecar): add trtllm sidecar deploy example and minimal image

tanmayv25:
> [!WARNING]
> **Experimental.** This TensorRT-LLM sidecar and its deployment example are experimental and not yet packaged for distribution. The manifest, flags, and behavior may change without notice.

## Summary

Adds a container image and a Kubernetes example for the TensorRT-LLM Rust sidecar (`dynamo-trtllm-sidecar`).

- `lib/sidecar/trtllm/Dockerfile` — builds the sidecar into a slim image. No GPU, no TensorRT-LLM.
- `lib/sidecar/trtllm/deploy/agg.yaml` — runs the sidecar next to a stock TensorRT-LLM engine in one pod. Serves `Qwen/Qwen3-0.6B` on one GPU. The engine runs as a native sidecar (an initContainer with `restartPolicy: Always`), so it starts before the worker and stops after it — the worker no longer races the engine on startup.
- `lib/sidecar/trtllm/README.md` — a quick start: build/push the image, create the HF secret, deploy, send a request.

The engine is pinned to `1.3.0rc21` (rc22 crashes on startup). There is no published sidecar image yet, so the manifest uses a `<your-registry>` placeholder.

## Validation

Tested end to end on GB200 (arm64) clusters:

- Builds pass: `cargo build`, `docker build` (amd64), `docker buildx` (arm64).
- Deployed `agg.yaml`: the engine renders as a native-sidecar initContainer, the worker starts only after it is ready and comes up with **0 restarts**, pod reaches `2/2 Running`, and the worker registers the model.
- Engine binds `127.0.0.1` only — verified via `/proc/net/tcp` inside the engine and a cross-pod connection refused to the pod IP.
- `GET /v1/models` lists the model. `POST /v1/chat/completions` returns a reply.


lib/sidecar/trtllm/deploy/agg.yaml
```diff
@@ -0,0 +1,81 @@
...
+              ephemeral-storage: 2Gi
+          command:
+          - python3
+          - -m
+          - tensorrt_llm.commands.serve
+          args:
+          - Qwen/Qwen3-0.6B
+          - --grpc
+          - --host
+          - 0.0.0.0
```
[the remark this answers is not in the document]
tanmayv25: Fixed in 41c40b7 — the engine now binds `--host 127.0.0.1`, so the unauthenticated gRPC is reachable only over pod loopback, matching the manifest's stated security boundary.

One knock-on change: because the engine is now loopback-only, the `tcpSocket` startupProbe couldn't stay (the kubelet dials the pod IP, which can't reach a `127.0.0.1`-only listener). It's replaced with an `exec` probe that checks `127.0.0.1:50051` from inside the engine container, preserving the native-sidecar startup ordering. Thanks for the catch.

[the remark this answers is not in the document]
tanmayv25: Fixed in 41c40b7 — the engine now binds `--host 127.0.0.1`, so the unauthenticated gRPC is reachable only over pod loopback, matching the manifest's stated security boundary.

One knock-on change: because the engine is now loopback-only, the `tcpSocket` startupProbe couldn't stay (the kubelet dials the pod IP, which can't reach a `127.0.0.1`-only listener). It's replaced with an `exec` probe that checks `127.0.0.1:50051` from inside the engine container, preserving the native-sidecar startup ordering. Thanks for the catch.

lib/sidecar/trtllm/README.md
```diff
@@ -48,5 +48,76 @@ dynamo-trtllm-sidecar \
...
+
+In `deploy/agg.yaml`, set the `main` worker image to the one you just pushed.
+If your registry is private, add `imagePullSecrets` to the worker pod spec.
+
+### 3. Create the Hugging Face token secret
+
+```bash
+kubectl create secret generic hf-token-secret \
+  --from-literal=HF_TOKEN=<your-hf-token> -n <namespace>
+```
```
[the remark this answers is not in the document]
tanmayv25: Fixed in 41c40b7 — the secret command now reads the token from `$HF_TOKEN` (`--from-literal=HF_TOKEN="$HF_TOKEN"`), matching the repo's recipe convention, so the literal token no longer lands in the command or shell history. Added a note pointing to `--from-file` / an external secret manager for stricter (CI) environments. Thanks.

connorcarpenter15 (approved):

lib/sidecar/trtllm/deploy/agg.yaml
```diff
@@ -0,0 +1,120 @@
...
+          # Gates `main` on the gRPC port opening. An exec probe (not tcpSocket)
+          # because the engine binds loopback only, which the kubelet cannot reach
+          # from the pod IP. Checks the port is listening, not that the model has
+          # loaded; the sidecar retries if the engine is not serving yet.
+          startupProbe:
+            exec:
+              command:
+              - python3
+              - -c
+              - import socket; socket.create_connection(('127.0.0.1', 50051), 2).close()
```
atchernych: is this a temporary startup probe?
startupProbe only proves 127.0.0.1:50051 accepts TCP. After that, if trtllm-engine crashes/restarts, main (dynamo-trtllm-sidecar) can stay Ready and keep getting traffic.
TCP accept ≠ gRPC serving ≠ model loaded so this is not a true StartupProbe
At least do something like below
```
startupProbe:
  exec:
    command: ["python3", "-c", "import grpc; grpc.insecure_channel('127.0.0.1:50051').unary_unary('/trtllm.TrtllmService/HealthCheck')(b'', timeout=3)"]
  periodSeconds: 5
  timeoutSeconds: 10        # must exceed the 3s RPC timeout + interpreter spawn
  failureThreshold: 60      # ~5 min for model load; raise for big models


```

tanmayv25: Fixed it!

lib/sidecar/trtllm/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# TensorRT-LLM runtime, so it ships in a slim Debian image separate from the
+# heavy TRT-LLM engine container. The two run side by side in one pod and talk
+# over loopback (see deploy/agg.yaml).
+#
+# Build from the repository root (the workspace is needed to compile the crate):
+#   docker build -f lib/sidecar/trtllm/Dockerfile -t dynamo-trtllm-sidecar:1.3.0 .
+
+# ---- builder ----------------------------------------------------------------
+# Pinned to the workspace toolchain (rust-toolchain.toml: 1.96.1).
+FROM rust:1.96.1-bookworm AS builder
```
nv-tusharma, line 10: need to use NVIDIA base images if we're looking to release this container

nv-tusharma:
do we want to add this example to CI, we have deploy-* tests that should be able to deploy these, likely would also need a build job to build the sidecar container. We should look to add to nightly.

atchernych (approved):

nv-anants (approved):
dockerfiles and examples are experimental for now, looks good. Will need closer review and changes like base-image, CI, compliance etc changes when this goes official

lib/sidecar/trtllm/Dockerfile
nv-tusharma: since this is experimental not required today, resolving my comment

nv-tusharma (approved):

PR #12207, opened 2026-07-27
fix(thunderagent): account for hybrid KV blocks

jthomson04:
## Summary

- publish vLLM's initialized, hybrid-aware `kv_cache_size_tokens` in optional runtime metadata while preserving the existing physical block metadata
- make ThunderAgent prefer the authoritative token capacity, fall back to `kv_cache_block_size * total_kv_blocks`, and add native offloading capacity to either source
- carry each worker's block size into scheduling and use block-rounded footprints for admission, placement, demotion, pause, and resume decisions
- document the capacity source and lifecycle-specific block accounting

## Root cause

ThunderAgent previously derived vLLM capacity only from the physical block count and event block size. That product is not the authoritative usable token capacity for hybrid KV cache layouts. The scheduler also counted raw in-flight tokens plus a flat buffer, so concurrent partial blocks were invisible even though each request occupied a full allocation block. The error becomes large for hybrid models with block sizes around 4,160 tokens.

REASONING programs and new or resumed placements now reserve `ceil((token_total + buffer) / block_size)` blocks. ACTING programs count only complete reusable blocks before the existing acting weight or idle-time decay is applied.

SGLang Mamba-state capacity, TensorRT-LLM multi-pool capacity, and DP-rank-aware accounting remain follow-up work.

## Validation

- `pytest` ThunderAgent capacity, router, and program-state tests: 42 passed
- targeted vLLM runtime metadata tests: 2 passed
- `ruff check` on changed Python files
- `fern check`
- `fern docs broken-links`

## Summary by CodeRabbit

- **New Features**
  - Added hybrid-aware KV-cache capacity reporting from vLLM.
  - Improved worker capacity tracking with per-worker block sizes and retention limits.
  - Added block-aligned scheduling for admissions, pauses, acting workloads, and resumptions.
  - Added native offloading capacity to available worker capacity.

- **Bug Fixes**
  - Added validated fallback handling when authoritative cache capacity is unavailable or invalid.
  - Improved capacity metadata validation and resilience to malformed runtime data.

- **Documentation**
  - Documented KV-cache sources, fallback behavior, block accounting, and scheduling rules.


components/src/dynamo/thunderagent_router/router.py
```diff
@@ -322,17 +367,20 @@ async def _pause_until_safe(self, capacities: dict[int, int]) -> None:
...
-                target_limit = capacity * pause_target
+                target_limit = capacity.retention_tokens * pause_target
                 paused_this_tick = 0
                 marked_this_tick = 0
                 # Bound the inner loop by total program count so a candidate
                 # transitioning out from under us can't spin the tick.
                 for _ in range(len(self._table.programs) + 1):
-                    if self._worker_used(worker_id) <= target_limit:
+                    if (
+                        self._worker_used(worker_id, capacity.block_size)
```
jthomson04: **Blocking: this loop can pause programs that release nothing.**

With `floor()` accounting, any ACTING program whose `token_total < block_size` contributes **0** to `_worker_used`. `_smallest_candidates` selects ACTING programs by *ascending* `token_total`, so it hands back exactly those zero-contribution programs first, and the loop `continue`s without checking whether progress was made.

Concretely: hybrid worker, `block_size=4160`, `retention_tokens=1M`, 200 ACTING programs holding ~3k tokens each (0 accounted) plus a few large REASONING programs pushing used past `0.95 * 1M`. The tick pauses **all 200** short programs - each releasing nothing - before `acting is None` finally lets it reach the REASONING branch.

Previously every pause released at least `buffer_per_program`, so the loop had guaranteed monotone progress. That guarantee is gone.

Secondary effect: `_worker_used` is O(programs) and is called once per iteration, so this is O(n^2) **while holding `self._lock`** - a direct latency spike on the `before_request` admission path.

Suggested fix - skip candidates that release nothing:

```python
acting, reasoning = self._smallest_candidates(worker_id)
if acting is not None and self._program_tokens(acting, capacity.block_size) == 0:
    acting = None  # pausing frees no blocks; fall through to reasoning
```

or track `used` before/after and `break` when a pause doesn't reduce it.

components/src/dynamo/vllm/main.py
```diff
@@ -688,6 +692,7 @@ async def register_vllm_model(
         )
         num_gpu_blocks = 0
     runtime_config.total_kv_blocks = per_rank_kv_blocks(num_gpu_blocks, dp_range[1])
+    _publish_kv_cache_capacity(runtime_config, runtime_values["kv_cache_size_tokens"])
```
jthomson04: **Blocking: DP-rank scale mismatch between the two capacity sources.**

The line above normalizes `total_kv_blocks` to **per-rank** via `per_rank_kv_blocks`, but `kv_cache_size_tokens` is published raw. With internal or hybrid DP load balancing, `get_dp_range_for_worker` returns `dp_size > 1` (`handlers.py:951`), so ThunderAgent's new authoritative budget is **`dp_size x` larger** than the fallback it replaces - it will over-admit by that factor.

The PR body lists "DP-rank-aware accounting" as follow-up work, but this isn't deferred polish: the change swaps a per-rank number for an aggregate one on the same code path, so the regression lands now.

Either divide consistently, or gate publishing:

```python
if dp_range[1] == 1:
    _publish_kv_cache_capacity(runtime_config, runtime_values["kv_cache_size_tokens"])
```

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -460,22 +508,32 @@ def group_key(program: Program) -> int:
...
 
             backend_caps.sort(key=lambda x: -x[1])
 
-            total_capacity = sum(r for _, r in backend_caps)
+            total_capacity = sum(remaining for _, remaining, _ in backend_caps)
             resumable_programs: list[Program] = []
             cumulative = 0
             for program in paused_programs:
-                required = program.token_total + self._cfg.buffer_per_program
+                required = min(
```
jthomson04: This pre-check isn't meaningful across heterogeneous block sizes. `required` is the cost on the *smallest-block* worker, while `total_capacity` sums remaining across *all* workers. A program that only fits on the large-block worker is admitted to `resumable_programs` at the small-block price.

Nothing breaks - the placement loop below just `continue`s past it - but the pre-check no longer bounds anything, which was its purpose. Consider computing per-worker `required`, or dropping the pre-check entirely and letting placement be the only gate.

Also: `min(...)` is recomputed per program even though the `backend_caps` block sizes are invariant inside this loop. Hoist it.

components/src/dynamo/thunderagent_router/capacity.py
```diff
@@ -3,33 +3,54 @@
...
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
+        return None
+    if isinstance(value, float) and not math.isfinite(value):
+        return None
+    value = int(value)
+    return value if value > 0 else None
+
+
+@dataclass(frozen=True)
+class WorkerCapacity:
```
jthomson04, line 33: `_round_up_tokens` / `_complete_block_tokens` in `router.py` divide by `block_size` unguarded. `_parse_capacity` guarantees positivity today, but this frozen dataclass doesn't.

A `block_size=0` instance raises `ZeroDivisionError` inside `_scheduler_tick`, which the loop's `except Exception` swallows - repeating every tick until the 10-failure counter **permanently halts the scheduler loop**. Cheap insurance: a `__post_init__` guard here, or `max(1, block_size)` in the two helpers.

Separately, `_positive_int` above is a third copy of the bool/finite/positive validation that already exists in `common/native_offloading.py` and now `common/kv_cache_capacity.py`.

components/src/dynamo/common/kv_cache_capacity.py
```diff
@@ -0,0 +1,36 @@
...
+KV_CACHE_CAPACITY_RUNTIME_KEY = "kv_cache_capacity"
+
+
+def kv_cache_capacity(total_tokens: object) -> dict[str, int] | None:
+    """Build runtime metadata from an authoritative KV cache token capacity."""
+    if (
+        isinstance(total_tokens, bool)
+        or not isinstance(total_tokens, (int, float))
+        or total_tokens <= 0
+    ):
```
jthomson04: This module is a near-verbatim clone of `common/native_offloading.py` - 25 of 36 lines are identical modulo identifiers. Worth extracting a shared `_token_capacity_payload(value)` and `get_capacity_tokens(runtime_data, key)`, leaving each module with just its key constant and thin wrappers.

A three-way copy (counting `_positive_int` in `thunderagent_router/capacity.py`) is worth collapsing now, before a fourth backend needs one.

components/src/dynamo/vllm/main.py
```diff
@@ -775,6 +794,11 @@ def get_engine_cache_info(engine: AsyncLLM) -> dict[str, Any]:
             "num_gpu_blocks": engine.vllm_config.cache_config.num_gpu_blocks,
             "block_size": engine.vllm_config.cache_config.block_size,
             "kv_event_block_size": kv_event_block_size,
+            "kv_cache_size_tokens": getattr(
+                engine.vllm_config.cache_config,
+                "kv_cache_size_tokens",
```
jthomson04: Silent no-op risk: this `getattr` fails closed with **no log**, and `_publish_kv_cache_capacity` returning `False` is also silent. If the attribute doesn't exist on the pinned vLLM or is renamed upstream, the whole feature quietly reverts to the legacy path with nothing in the logs. Add a `logging.warning` on the miss path.

I couldn't verify `kv_cache_size_tokens` exists on vLLM 0.24.0's `CacheConfig` (per `container/compliance/base_sboms/`) - vllm isn't installed in my checkout. Worth confirming against a real engine, since `test_engine_cache_info_includes_authoritative_capacity_and_physical_blocks` uses `SimpleNamespace` and would pass either way.

jthomson04 (commented):
## Review summary

Direction is right and the docs updates are unusually clear about the accounting model. The validation in the new metadata module is appropriately paranoid (rejects `bool`, non-finite floats, non-positives), and the `[1, 64, 4160]` block-size parametrization across the new router tests is the right shape.

Two issues I'd want fixed before merge (inline), plus smaller items.

### Blocking

1. **`_pause_until_safe` can pause programs that free nothing** (`router.py:379`). With `floor()` accounting, any ACTING program with `token_total < block_size` contributes 0 to `_worker_used`, and `_smallest_candidates` returns exactly those first. The old code released at least `buffer_per_program` per pause, guaranteeing monotone progress; that guarantee is gone.
2. **DP-rank scale mismatch** (`vllm/main.py:694-695`). `total_kv_blocks` is normalized per-rank by `per_rank_kv_blocks`, but `kv_cache_size_tokens` is published raw — so with internal/hybrid DP LB the new "authoritative" budget is `dp_size x` larger than the fallback it replaces.

### Non-blocking

- `_greedy_resume`'s feasibility pre-check mixes a min-over-workers cost with a sum-over-workers capacity, so it no longer bounds anything (inline).
- `WorkerCapacity` has no invariant enforcement; a `block_size=0` instance raises `ZeroDivisionError` inside `_scheduler_tick`, which the loop swallows until the 10-failure counter **permanently halts the scheduler** (inline).
- `common/kv_cache_capacity.py` is a near-verbatim clone of `common/native_offloading.py`, and `_positive_int` in `thunderagent_router/capacity.py:28` is a third copy of the same validation (inline).
- Silent no-op risk on the vLLM publish path (inline).
- `_publish_kv_cache_capacity` returns a `bool` the single call site discards.
- `_greedy_resume`'s 5-tuple `target` pack/unpack (`router.py:551-568`) is hard to follow; storing just `index` and re-reading `backend_caps[index]` would read better.
- Docstring at `router.py:5` says "BFD restore", but sorting descending by remaining and taking the first fit is *worst*-fit. Pre-existing, just noting it while the function is being touched.

### Semantics worth confirming

The `floor()` model for ACTING is defensible - vLLM frees the trailing partial block on request completion, so only complete blocks survive as reusable prefix cache. But it creates a discontinuity of up to `2 x block_size` at the REASONING->ACTING boundary (`ceil(t + buffer)` -> `floor(t)`), which `test_acting_transition_counts_only_complete_reusable_blocks` encodes directly. Combined with item 1, that discontinuity is what makes the pause loop misbehave. Please confirm the "partial tail block is released between turns" assumption holds for hybrid Mamba groups too, since that's the exact configuration this targets.

Separately: `kv_cache_block_size` in the MDC is `kv_event_block_size` (`main.py:749`), from `select_main_attention_block_size` - the *main attention group's* block size, overridable via `additional_config["dynamo_kv_event_block_size"]`. This PR repurposes it as "the worker's physical allocation granularity." They coincide in practice for hybrid models, but an operator setting the override silently makes all the rounding arithmetic meaningless. Worth a comment noting the coupling.

### Test gaps

- No test for the `_pause_until_safe` sub-block regression above - the item most likely to bite in production.
- No DP > 1 test for the vLLM publish path.
- `test_publish_kv_cache_capacity_preserves_physical_block_metadata` uses a hand-rolled `FakeRuntimeConfig` and `SimpleNamespace`, so it passes regardless of whether `kv_cache_size_tokens` exists on the real `CacheConfig`. It validates the plumbing, not the contract.
- Missing: `kv_cache_capacity` present with `total_kv_blocks` absent or 0. This is a genuine behavior improvement (worker now included where previously dropped) but untested - note it also means Ray-DP workers carrying the `total_kv_blocks=0` sentinel could now enter the snapshot.
- No test for `_publish_kv_cache_capacity` returning `False` when `total_tokens is None` (the common legacy path).

### Performance / security

`_greedy_resume` lost the `min_required > remaining: break` early exit, making placement O(programs x workers) per tick under the lock - acceptable at realistic worker counts, but it compounds with the O(n^2) issue in item 1. Relaxing the `backend_caps` filter from `r > buffer_per_program` to `r > 0` is correct now that `required` includes the buffer. The `self._parsed` JSON cache is preserved correctly across the `int -> WorkerCapacity` change. Nothing concerning security-wise.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 7ae575bc6d. `_smallest_candidates` now receives the worker block size and skips ACTING candidates whose existing accounted footprint is zero. The pause loop, candidate ordering, and REASONING fallback are otherwise unchanged. Added a 4,160-token regression proving the sub-block program stays active while the next positive-footprint ACTING program is paused.

components/src/dynamo/vllm/main.py
jthomson04: Verified against the pinned vLLM 0.25.1 contract and kept the raw value intentionally. `CacheConfig.kv_cache_size_tokens` is explicitly per-DP-engine, and `core_client` deliberately leaves it unsummed while aggregating `num_gpu_blocks`; only the latter needs `per_rank_kv_blocks`. Added an inline explanation plus a DP=2 regression that expects 246 blocks -> 123 while preserving 456,789 capacity tokens. Sources: https://docs.vllm.ai/en/v0.25.1/api/vllm/config/cache/ and https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/v1/engine/core_client.py#L667-L685

components/src/dynamo/thunderagent_router/router.py
jthomson04: Addressed in 7ae575bc6d under the deployment invariant that workers serving one model use the same KV block size. The common block size is now hoisted once and used for both aggregate feasibility and placement. The existing aggregate precheck, group ordering, smallest-first admission, largest-first packing, and backend ordering are unchanged. Added a two-worker shared-block-size regression that locks the existing selection and placement behavior.

components/src/dynamo/thunderagent_router/capacity.py
jthomson04: Fixed in 7ae575bc6d. `WorkerCapacity.__post_init__` now rejects non-integer, boolean, zero, and negative retention/block values before scheduler arithmetic can see them. The duplicate parser was also replaced with the shared `positive_int` helper. Focused invariant tests cover both fields.

components/src/dynamo/common/kv_cache_capacity.py
jthomson04: Fixed in 7ae575bc6d. Added `dynamo.common.token_capacity` with the shared positive-value, payload, and runtime-data parsing helpers. The KV-cache and native-offloading modules now retain only their key constants and thin public wrappers; ThunderAgent capacity parsing uses the same validator. Wrapper parity is covered by a focused parameterized test.

components/src/dynamo/vllm/main.py
jthomson04: Fixed in 7ae575bc6d. Dynamo pins vLLM 0.25.1, so `get_engine_cache_info` now accesses `cache_config.kv_cache_size_tokens` directly. `_publish_kv_cache_capacity` no longer returns an ignored boolean and logs a warning when no usable capacity is available. The contract test now uses a real vLLM `CacheConfig`, and a separate test covers the warning/fallback path.

jthomson04:
@coderabbitai full review

jthomson04 (commented):
### Code review: 15 findings

Reviewed for recall at extra-high effort. Ordered most severe first; each is an inline comment below.

**Findings 1-4 share one root cause.** This PR introduces two different footprint formulas -- block-rounded-up (`ceil((t + buffer)/bs)*bs`) for REASONING, admission, and resume, and block-floored (`floor(t/bs)*bs`) for ACTING -- but the transitions between them are ungated:

- resume reserves the rounded-up figure and the next tick accounts the floored one (finding 1)
- the granularity used for that reservation is taken from a single arbitrary worker (finding 2)
- the floored formula collapses to 0 below one block, disabling both backpressure bands (finding 3)
- a returning program crosses from floored back to rounded-up with no admission check (finding 4)

The gap between the two formulas is up to `2 * block_size`, i.e. roughly 8,300 tokens for the 4,160-token hybrid blocks this PR targets. They are commented separately but are best fixed together.

**Findings 2 and 7 are regressions introduced by commit 2 against commit 1** of this PR (per-worker resume granularity, and the `getattr` guard on `cache_config.kv_cache_size_tokens`).

Finding 6 is anchored to the nearest line inside a diff hunk; its body states the line it actually refers to.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -484,29 +553,42 @@ def group_key(program: Program) -> int:
...
             for program in resumable_programs:
                 if not backend_caps:
                     break
-                worker_id, remaining = backend_caps[0]
-                if min_required > remaining:
-                    break
-                required = program.token_total + self._cfg.buffer_per_program
-                if required > remaining:
+                target = None
+                required = self._resume_tokens(program, block_size)
```
jthomson04: **1. Resume reserves a block-rounded footprint that the next tick immediately discounts**

`_resume_tokens` reserves `ceil((token_total + buffer)/block_size)*block_size`, but a program paused by `_pause_acting_locked` keeps `status == ACTING`, and `_resume_program` (router.py:600-618) only changes `lifecycle`. So the next tick accounts that same program at `floor(token_total/block_size)*block_size` (router.py:272).

block_size=4160, buffer_per_program=100, program P paused with token_total=4160:

- tick N: `required` = ceil(4260/4160)*4160 = 8320, deducted from that worker's `remaining`
- tick N+1: `_worker_used` counts P as 4160, so `remaining` is 4160 higher than tick N believed, and another program is resumed onto the same worker

This repeats every tick until every paused program is resumed. They then take their next turn, flip to REASONING, jump back to 8320 each, and the worker overshoots `pause_threshold` and mass-pauses them again.

Pre-PR `required` was `t + buffer` and post-resume accounting was `t*weight + buffer`, so the two matched exactly and this loop did not exist.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -460,22 +519,32 @@ def group_key(program: Program) -> int:
...
+            backend_caps = [(w, r, b) for w, r, b in backend_caps if r > 0]
             if not backend_caps:
                 return
 
             backend_caps.sort(key=lambda x: -x[1])
 
-            total_capacity = sum(r for _, r in backend_caps)
+            # Workers serving one model deployment use the same KV allocation
+            # granularity, so the aggregate feasibility pass has one cost.
+            block_size = backend_caps[0][2]
```
jthomson04: **2. One worker's block size is applied to every worker's resume reservation**

`block_size = backend_caps[0][2]` takes the granularity of whichever worker currently has the most free capacity and uses it for `_resume_tokens` on all workers (router.py:562). `worker_block_size` is unpacked at 567 and written back at 588 but never used in any arithmetic.

This is a regression against commit 1 of this PR, which computed `required = self._resume_tokens(program, block_size)` per candidate worker inside the placement loop. Commit 2 hoisted it out.

Heterogeneous block sizes are reachable inside a single deployment: `configure_kv_event_block_size` falls back to `cache_config.block_size` whenever the `get_kv_cache_group_metadata` RPC throws (components/src/dynamo/vllm/cache_info.py:50-60), so one worker can publish `kv_cache_block_size` 16 while its peers publish 4160.

- 16-block worker sorts first: a 5000-token program gets `required`=5100 but lands on a 4160-block worker that will actually consume 8320 -> 2x over-subscription
- 4160-block worker sorts first: programs that would fit on the 16-block worker are rejected

`backend_caps` is sorted by live remaining capacity (538), so which granularity wins changes tick to tick.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -245,18 +245,46 @@ async def _scheduler_tick(self) -> None:
...
         if program.status != ProgramStatus.ACTING:
-            return program.token_total
+            return self._round_up_tokens(
+                program.token_total + self._cfg.buffer_per_program,
+                block_size,
+            )
+
+        # At a completed tool turn, only full vLLM cache blocks remain reusable;
+        # the trailing partial attention/Mamba-aligned block is released.
+        reusable_tokens = self._complete_block_tokens(program.token_total, block_size)
```
jthomson04: **3. Sub-block ACTING programs count as zero, so utilization can read 0% on a loaded worker**

`_complete_block_tokens(t, block_size)` returns 0 for any `token_total < block_size`, and the old `_worker_used` term `len(programs) * buffer_per_program` was removed, so nothing else contributes.

Hybrid model, block_size=4160 (the case this PR targets), 500 programs parked at a tool boundary with token_total=3000 each -> `_worker_used` = 0:

- `_apply_soft_demotes` (345-361): util = 0, below `soft_demote_threshold`, skipped
- `_pause_until_safe` (372-374): base_used=0 <= retention*0.95 -> `continue`
- `_select_worker_for_new_program_locked` (334): used=0 -> admits every new program

Both backpressure bands are permanently disabled and admission is unbounded.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -245,18 +245,46 @@ async def _scheduler_tick(self) -> None:
...
+    def _program_tokens(
+        self,
+        program: Program,
+        block_size: int,
+        *,
+        decayed: bool = False,
+    ) -> int:
         if program.status != ProgramStatus.ACTING:
-            return program.token_total
+            return self._round_up_tokens(
```
jthomson04: **4. ACTING -> REASONING footprint jumps by up to 2x block_size with no admission gate**

REASONING reserves `ceil((t + buffer)/block_size)*block_size` while ACTING accounts `floor(t/block_size)*block_size` (router.py:272); the delta is between 1 and 2 blocks. `_admit_locked` only runs a capacity check when `was_new and program.assigned_worker_id is None` (router.py:156), so a returning program re-enters with no check at all.

block_size=4160, buffer=100, ten ACTING programs at token_total=4100 pinned to worker 1 -> `_worker_used` = 0. All ten finish their tool calls and call `before_request`: each returns `None, False` at line 156, `begin_request` sets REASONING, and each becomes ceil(4200/4160)*4160 = 8320. Accounted usage goes 0 -> 83,200 between two scheduler ticks with nothing able to refuse it.

Pre-PR the same transition went from `t*weight + buffer` to `t + buffer`, i.e. no jump at all at the default `acting_token_weight=1.0`.

components/src/dynamo/thunderagent_router/capacity.py
```diff
@@ -56,40 +76,42 @@ def snapshot(self) -> dict[int, int]:
...
-            ):
-                result = int(block_size) * int(total_blocks)
-                runtime_data = (card.get("runtime_config") or {}).get(
-                    "runtime_data", {}
-                )
+            block_size = positive_int(card.get("kv_cache_block_size"))
+            runtime_config = card.get("runtime_config") or {}
+            if block_size is not None and isinstance(runtime_config, dict):
+                runtime_data = runtime_config.get("runtime_data", {})
+                retention_tokens = get_kv_cache_capacity_tokens(runtime_data)
```
jthomson04: **5. `kv_cache_size_tokens` is a logical-token figure being compared against physical block footprints**

vLLM computes it as `int(max_concurrency * max_model_len)` (`get_kv_cache_capacity`, `vllm/v1/core/kv_cache_utils.py`), not a physical pool size. The router then compares it against footprints denominated in physical block tokens (`ceil(t/bs)*bs`, router.py:249-251).

Hybrid model, block_size=4160, user passes `--max-model-len 2048`: `num_block_per_request` = 1, so `max_concurrency = num_blocks` and `kv_cache_size_tokens = num_blocks * 2048`, while the physical pool is `num_blocks * 4160`. Every program is then counted at >= 4160 against a budget of `num_blocks * 2048`, so each consumes two budget-blocks: the worker is declared full at half its real program capacity and `pause_threshold` fires at roughly 47% real utilization.

That is worse than the `block_size * total_kv_blocks` fallback this branch replaces, and it shifts with `--max-model-len` even though the physical pool is identical.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -350,20 +403,22 @@ async def _pause_until_safe(self, capacities: dict[int, int]) -> None:
                         continue
```
jthomson04, line 455: **6. The pause loop marks every REASONING program on the worker in a single tick**

This refers to the reasoning branch at router.py:395-403, specifically `reasoning.marked_for_pause = True` at line 401 -- anchored here because 395-401 are unchanged context outside the diff hunk.

Setting `marked_for_pause` changes neither `lifecycle` nor `status`, so `_worker_used` is identical on the next iteration, and `_smallest_candidates` skips already-marked programs (430-431). The loop therefore marks the next one, and the next, bounded only by `range(len(self._table.programs) + 1)` (381).

The new zero-footprint skip at 433 makes this branch far more reachable: on a hybrid worker where every ACTING program sits under one block, `_smallest_candidates` returns `acting=None` on the first iteration and the loop falls straight through to the reasoning branch. Every in-flight REASONING program gets marked, and `after_request` (205-210) then pauses all of them at their next tool boundary, dropping the worker to near-zero instead of stopping at `pause_target`.

components/src/dynamo/vllm/main.py
```diff
@@ -775,6 +801,7 @@ def get_engine_cache_info(engine: AsyncLLM) -> dict[str, Any]:
             "num_gpu_blocks": engine.vllm_config.cache_config.num_gpu_blocks,
             "block_size": engine.vllm_config.cache_config.block_size,
             "kv_event_block_size": kv_event_block_size,
+            "kv_cache_size_tokens": engine.vllm_config.cache_config.kv_cache_size_tokens,
```
jthomson04: **7. Dropped `getattr` guard turns a missing optional field into a registration failure**

Commit 2 of this PR replaced `getattr(engine.vllm_config.cache_config, "kv_cache_size_tokens", None)` with a bare attribute access. `get_engine_cache_info` catches and re-raises (main.py:822-824), so `register_vllm_model` propagates and the worker never registers.

`kv_cache_size_tokens` is recent: absent from `vllm/config/cache.py` at tag v0.22.0, present at v0.25.1 (the current pin in pyproject.toml). Any older or forked vLLM -- vllm-omni, a user-supplied base image, a pin rollback -- now takes a full registration outage instead of losing one optional metric.

The accompanying test cannot catch this. `test_engine_cache_info_includes_authoritative_capacity_and_physical_blocks` builds a real `CacheConfig` but then does `cache_config.kv_cache_size_tokens = 456_789` (components/src/dynamo/vllm/tests/test_vllm_unit.py:547). `@config` produces a non-slotted pydantic dataclass, so that assignment succeeds whether or not the field is declared, and the test passes green against a vLLM that would crash in production. A `hasattr` or `dataclasses.fields` assertion would be needed to make it meaningful.

components/src/dynamo/thunderagent_router/capacity.py
```diff
@@ -3,33 +3,53 @@
...
 
 
+@dataclass(frozen=True)
+class WorkerCapacity:
+    """A worker's retention budget and physical allocation granularity."""
+
+    retention_tokens: int
+    block_size: int
+
+    def __post_init__(self) -> None:
```
jthomson04, line 39: **8. `__post_init__` can raise onto the request hot path, and its float check is unreachable**

`snapshot()` wraps only `self._subscriber.get_model_cards()` in try/except (capacity.py:73-77); `_parse_capacity` and the `WorkerCapacity(...)` construction at 112 sit outside it. `before_request` -> `_admit_locked` -> `self._capacity.snapshot()` (router.py:159) has no try/except either, so any `ValueError` raised here surfaces as a 500 on a live chat-completions request rather than a dropped worker.

The two validation layers also disagree. `test_worker_capacity_requires_positive_integers` asserts `block_size=1.5` raises, but a card publishing `"kv_cache_block_size": 1.5` never reaches this guard: `positive_int` (components/src/dynamo/common/token_capacity.py:12-19) returns `int(1.5) == 1` first, and the worker is silently accounted with a 1-token block. The reachable layer is the permissive one.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -460,22 +519,32 @@ def group_key(program: Program) -> int:
...
+                    - self._worker_used(
+                        worker_id,
+                        capacity.block_size,
+                        decayed=False,
+                    ),
+                    capacity.block_size,
+                )
+                for worker_id, capacity in capacities.items()
             ]
+            backend_caps = [(w, r, b) for w, r, b in backend_caps if r > 0]
```
jthomson04: **9. Relaxed filter makes the aggregate resume feasibility pass over-optimistic**

The filter went from `r > self._cfg.buffer_per_program` to `r > 0` while `required` became block-rounded, so remaining slivers smaller than one block still count toward `total_capacity` (543).

Eight workers each with `remaining` = 4000 and block_size = 4160: `total_capacity` = 32,000, so the feasibility loop (546-550) marks up to seven 4160-token programs resumable. The placement loop then finds `required`=4160 > remaining=4000 on every worker and `continue`s past all of them, resuming nothing.

The old code excluded workers at or below `buffer_per_program` and short-circuited via `min_required`; this version instead does a full programs x workers scan every tick under `self._lock` and returns `resumed_this_tick == 0`.

components/src/dynamo/thunderagent_router/tests/test_capacity.py
```diff
@@ -48,17 +59,54 @@ def _card(
...
+def test_snapshot_uses_authoritative_capacity_without_physical_blocks(total_blocks):
+    provider, _ = _make_provider(
+        {"1": _card(4160, total_blocks, capacity_total_tokens=1_234_567)}
+    )
+    assert provider.snapshot() == {
+        1: WorkerCapacity(retention_tokens=1_234_567, block_size=4160)
+    }
 
 
 def test_snapshot_adds_native_offloading_tokens_to_retention_budget():
```
jthomson04: **10. The only coverage of "offloading + physical block-count fallback" was deleted**

This test now also sets `capacity_total_tokens=15_000`, so it short-circuits on the authoritative branch and no longer exercises `retention_tokens = block_size * total_blocks` followed by `+= offloaded_tokens` (capacity.py:104-111).

That is the path every non-vLLM worker takes: SGLang publishes `native_offloading_capacity` (components/src/dynamo/sglang/register.py:496) but not `kv_cache_capacity`. After this edit nothing covers the combination -- `test_snapshot_falls_back_to_physical_kv_pool_tokens` has no host tokens and `test_snapshot_ignores_invalid_native_offloading_capacity` supplies invalid ones.

A refactor that moved `+= offloaded_tokens` inside the `if retention_tokens is None` block would silently drop the entire HiCache retention budget for SGLang with a fully green suite.

components/src/dynamo/thunderagent_router/capacity.py
```diff
@@ -56,40 +76,42 @@ def snapshot(self) -> dict[int, int]:
...
-                if offloaded_tokens is not None:
-                    result += offloaded_tokens
+                if retention_tokens is not None:
+                    if offloaded_tokens is not None:
+                        retention_tokens += offloaded_tokens
+                    result = WorkerCapacity(
+                        retention_tokens=retention_tokens,
+                        block_size=block_size,
+                    )
         self._parsed[card_json] = result
```
jthomson04: **11. `_parsed` grows without bound, keyed on the full card JSON**

Every MDC republish that changes any field -- `total_kv_blocks` after a restart, a LoRA adapter count, a topology taint, a tokenizer-backend flip -- adds a new key holding a multi-KB card string that is never evicted, for the router process's entire lifetime.

The comment at 50-51 says the cache exists so "a subsequent snapshot() call avoids re-parsing on the request hot path", which only needs the current card per worker. Keying on `worker_id` with a stored card hash, or an LRU bounded by worker count, gives the same hit rate at fixed memory.

components/src/dynamo/common/kv_cache_capacity.py
```diff
@@ -0,0 +1,20 @@
...
+"""Runtime metadata for authoritative KV cache token capacity."""
+
+from __future__ import annotations
+
+from dynamo.common.token_capacity import get_capacity_tokens, token_capacity_payload
+
+KV_CACHE_CAPACITY_RUNTIME_KEY = "kv_cache_capacity"
+
+
+def kv_cache_capacity(total_tokens: object) -> dict[str, int] | None:
```
jthomson04: **12. Two modules that are now pure pass-throughs to the same helpers**

`kv_cache_capacity(x)` is `return token_capacity_payload(x)`, and `native_offloading_capacity(x)` (components/src/dynamo/common/native_offloading.py:13-15) is the identical line. `get_kv_cache_capacity_tokens` and `get_native_offloading_capacity_tokens` differ only in which constant they forward. That is two modules and four indirection hops to express two string constants, plus two import lines in capacity.py:19-21.

Moving both constants into `token_capacity.py` and calling `get_capacity_tokens(runtime_data, KV_CACHE_CAPACITY_RUNTIME_KEY)` directly deletes both files, and `test_capacity_metadata_wrappers_share_validation` -- a parametrized test whose only job is asserting the two wrappers behave identically -- goes with them.

components/src/dynamo/vllm/main.py
```diff
@@ -765,6 +774,23 @@ def _base_model_lora_capacity(config: Config, model_type: ModelType) -> int | No
     return config.engine_args.max_loras
 
 
+def _publish_kv_cache_capacity(
+    runtime_config: ModelRuntimeConfig, total_tokens: object
+) -> None:
+    """Publish vLLM's initialized hybrid-aware KV capacity when available."""
+    capacity = kv_cache_capacity(total_tokens)
+    if capacity is None:
+        logging.warning(
```
jthomson04: **13. WARNING fires on workers that will never run ThunderAgent**

`cache_config.kv_cache_size_tokens` is assigned only inside `EngineCore._initialize_kv_caches` when `kv_cache_groups` is non-empty, so it stays `None` for embedding and pooling models and for `--data-parallel-backend ray` (which already logs its own warning at main.py:689).

Every such worker now emits "vLLM did not report an authoritative KV cache token capacity; ThunderAgent will use the legacy block-count fallback" at startup, naming an experimental component the operator has never heard of. `register_vllm_model` is also called per model surface, so multimodal deployments log it more than once.

`logging.debug`, or a gate on the value actually being expected for this worker type, would fit better.

components/src/dynamo/thunderagent_router/router.py
```diff
@@ -484,29 +553,42 @@ def group_key(program: Program) -> int:
...
                 updated_remaining = remaining - required
-                if updated_remaining > self._cfg.buffer_per_program:
-                    backend_caps[0] = (worker_id, updated_remaining)
+                if updated_remaining > 0:
+                    backend_caps[index] = (
+                        worker_id,
+                        updated_remaining,
+                        worker_block_size,
+                    )
                     backend_caps.sort(key=lambda x: -x[1])
```
jthomson04: **14. Full re-sort of the worker list after every placement, under the request lock**

The list is already sorted descending and exactly one element's value decreased, so a single in-place reinsertion (or `heapq`) restores the invariant in O(W) / O(log W) instead of O(W log W) per program.

With W workers and P resumable programs this adds O(P * W log W) on top of the O(P * W) first-fit scan at 563-576, all while holding `self._lock`, which `before_request` also needs (router.py:95). At 64 workers and a few hundred paused programs that is tens of thousands of comparisons per tick blocking every request on the router.

components/src/dynamo/vllm/tests/test_vllm_unit.py
```diff
@@ -537,6 +538,82 @@ def set_model_load_time(self, _load_time):
     assert default_sampling_params == {"temperature": 0.7}
 
 
+def test_engine_cache_info_includes_authoritative_capacity_and_physical_blocks():
+    from vllm.config import CacheConfig
```
jthomson04: **15. Function-body import without the justifying comment the guidelines require**

`.ai/python-guidelines.md` states under "Keep imports at the top of the file": "**Always flag** any `import` statement that appears inside a function body, method, or class", and "Exceptions exist, but they should be rare and justified in a code comment." This import carries no comment.

The file's own precedent shows the expected form: `_load_vllm_main` (test_vllm_unit.py:64-72) carries a four-line docstring explaining exactly why that load is deferred. Without one here a reader cannot tell whether this is guarding the lightweight pre-commit collection environment or is simply an oversight.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. ACTING usage now applies the acting weight or decay, adds the buffer, and rounds up with the worker block size. Resume uses the same block-rounded footprint. A two-tick regression test verifies that another paused program does not fit on the next tick.

components/src/dynamo/thunderagent_router/capacity.py
jthomson04: Fixed in 158f7e10c1. The parse cache is now keyed by worker ID and stores only that worker's current raw card and parsed value. Each snapshot replaces the cache with the current worker set. Tests cover cache hits, card replacement, and removal of departed workers.

components/src/dynamo/thunderagent_router/capacity.py
jthomson04: Fixed in 158f7e10c1. MDC parsing now accepts only positive integers and rejects booleans, floats, and coerced strings. Each card parse is isolated behind `TypeError`, `ValueError`, and `OverflowError` handling, so one bad card is skipped instead of failing a request. The tests cover both strict input validation and per-worker failure isolation.

components/src/dynamo/thunderagent_router/capacity.py
jthomson04: Fixed in 158f7e10c1. The `kv_cache_size_tokens` publication and consumption path was removed. ThunderAgent now uses the physical device pool, `kv_cache_block_size * total_kv_blocks`, and adds native offloading capacity when present.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. The pause loop now tracks projected usage. It subtracts each new REASONING pause mark and accounts for already-pending releases once, so it stops at `pause_target` instead of marking every program. A two-cycle regression test verifies that only the required programs remain marked.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. Returning programs now pass through an admission check on their sticky worker after `begin_request` changes them to REASONING. If the block-rounded worker usage is over capacity, the program queues with reason `returning_full`. The regression test covers the ACTING-to-REASONING expansion.

components/src/dynamo/vllm/main.py
jthomson04: Fixed in 158f7e10c1. The vLLM authoritative-capacity lookup and publication were removed, so registration no longer reads `kv_cache_size_tokens` at all. The associated synthetic test was removed with that path.

components/src/dynamo/vllm/main.py
jthomson04: Fixed in 158f7e10c1. The vLLM capacity publication path and its ThunderAgent warning were removed. Workers that do not run ThunderAgent no longer emit this warning.

components/src/dynamo/vllm/tests/test_vllm_unit.py
jthomson04: Fixed in 158f7e10c1. The function-body import was part of the retired vLLM capacity-publication test. That test and import were removed together, so no undocumented local import remains.

components/src/dynamo/common/kv_cache_capacity.py
jthomson04: Fixed in 158f7e10c1. The new `kv_cache_capacity.py` and shared `token_capacity.py` pass-through layers were deleted with the invalid vLLM capacity path. The existing native-offloading helper remains self-contained, so the added wrapper duplication is gone.

components/src/dynamo/thunderagent_router/tests/test_capacity.py
jthomson04: Fixed in 158f7e10c1. The physical-pool plus native-offloading test is restored as a direct case: block size 16, 1000 blocks, and 300 offloaded tokens produce a 16,300-token retention budget.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. `_greedy_resume` now calculates `_resume_tokens` with each candidate worker's block size. The heterogeneous 4160/16 regression test verifies that a 5000-token program is placed only on the worker where its worker-specific footprint fits.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. The aggregate feasibility prepass was removed. Resume now checks each program directly against each worker-specific block-rounded remainder, and workers with unusable sub-block capacity are filtered out. The eight-workers-at-4000 regression case now resumes nothing without an aggregate false positive.

components/src/dynamo/thunderagent_router/router.py
jthomson04: Fixed in 158f7e10c1. ACTING programs no longer use floor accounting. The scheduler applies the acting weight or decay, adds the per-program buffer, and rounds up to the worker block size. The sub-block regression test verifies that these programs contribute utilization and can be paused.

jthomson04:
/ok to test [158f7e1](https://github.com/ai-dynamo/dynamo/commit/158f7e10c126b118484eca0e86a807137fc3d144)

Issue #10572, opened 2026-06-11
[CONTRIBUTION]: priority_load_shed_percent / load shed last contribution


michaelfeil:
### Type of Change

Feature

### Problem Statement

- Queue depth limits currently apply uniformly to normal and priority traffic.
- During short overload spikes, priority requests may be rejected at the same pending-queue cap as normal requests.
- Operators may want priority traffic to tolerate a controlled amount of additional queued work without globally raising queue limits.
- Raising the global queue cap helps priority traffic but also allows normal traffic to build a deeper queue, increasing tail latency and memory pressure.
- Completely bypassing queue limits for priority traffic is also risky because it can create unbounded overload.
- The desired behavior is a bounded per-request cap adjustment: priority requests can exceed the base queue-depth limit by a configured percentage, but still hit a hard backpressure limit.
- This should work with existing queue-depth tiers, especially caps based on pending ISL tokens or missing-prefill work.

### Proposed Solution

- Add a request-level `priority_load_shed_percent` admission parameter.
- Expose the parameter through `nvext` and all relevant external request interfaces, not only internal scheduler structs.
- Apply it only when the request has a positive priority boost.
- Increase the applicable queue-depth cap by the configured percentage for that request.
- Continue returning backpressure once the boosted cap is exceeded, preserving a bounded overload policy.
- Keep default behavior unchanged when `priority_load_shed_percent` is absent or zero.
- Treat the parameter as a request admission hint, not a global queue configuration change.
- Compute the effective cap as `base_cap + (base_cap * priority_load_shed_percent / 100)`.
- Apply the boosted cap only to the current request's admission decision; do not mutate the configured base cap.
- Require a positive priority signal, such as `priority_jump > 0`, before applying the boosted cap so normal traffic cannot opt into a higher cap accidentally.
- Preserve existing backpressure behavior when no base cap applies, for example when queue depth is unbounded.
- Add validation or clamping if the request/API surface can receive out-of-range values.
- Example external request shape:

```json
{
  "model": "example-model",
  "messages": [{"role": "user", "content": "hello"}],
  "nvext": {
    "priority_jump": 5.0,
    "priority_load_shed_percent": 20
  }
}
```

- Example parsed extension type:

```rust
#[derive(Default, Deserialize)]
struct NvExt {
    #[serde(default)]
    priority_jump: f64,
    #[serde(default)]
    priority_load_shed_percent: u8,
}

struct SchedulingRequest {
    priority_jump: f64,
[...]

PR #12209, opened 2026-07-27
feat(router): bounded queue-cap boost for priority requests

ZhuohaoNi:
## Overview

Adds `priority_load_shed_percent`, a per-request admission hint that lets a priority
request be admitted against a stretched router queue-depth cap instead of the base cap.

The queue-depth caps (`request_queue_limit_per_worker`,
`raw_isl_token_queue_limit_per_worker`, `cached_token_queue_limit_per_worker`) apply
identically to normal and priority traffic today, so a priority request is shed at the
same cap as everything else during a short overload spike. Raising the configured cap
deepens the queue for normal traffic too, and exempting priority traffic removes
backpressure entirely.

```
effective_cap = base_cap + base_cap * priority_load_shed_percent / 100
```

## Details

- Applied only when `priority_jump > 0`, so normal traffic cannot opt into a deeper queue.
- Applied only to that request's admission decision; the configured cap is never mutated.
- Backpressure still fires at the stretched cap, so this is a bounded relaxation, not a
  bypass.
- Absent or `0` reproduces current behavior exactly. `u8` bounds growth to 2.55x, and
  saturating arithmetic handles large caps.
- Applied where all three tiers are compared, so requests, raw ISL tokens, and cached
  tokens stay consistent.
- `QueueRejection` already reports the limit it rejected against, and that limit is now
  the effective cap, so a boosted rejection is distinguishable from a base-cap rejection
  without a new field.
- `RouterRequest::New` carries the value with a serde default, skipped on serialize when
  zero, so the router wire contract can express the hint and clients that omit it are
  unaffected.

## Scope

Scoped to the router per the discussion on #10572. The `nvext` field, the
`x-dynamo-request-priority-load-shed-percent` header, the preprocessor lift, the EPP/GAIE
path, the C bindings, and user-facing docs are ready on a follow-up branch and will be
opened once the core behavior is agreed. Until then every call site outside the router
passes `0`, so behavior is unchanged.

`lib/kv-router/src/scheduling/policy_queue.rs` — `boosted_limit`, the `queue_rejection`
call, and the new tests. That is the entire behavior change.

## Tests

- A non-priority request keeps the base cap.
- A priority request with `0` percent keeps the base cap.
- A priority request with a positive percent queues past the base cap and is still
  rejected past the boosted cap.
- The same for the raw ISL token and cached token tiers.
- `boosted_limit` rounding and saturation.
- `RouterRequest::New` round-trip, including an omitted field deserializing to `0`.

Verified locally: `cargo fmt` clean; `cargo clippy --no-deps --all-targets -- -D warnings`
clean at the workspace root and in `lib/bindings/python`; dynamo-kv-router 856 passed,
standalone-selection 49, dynamo-llm 1785, dynamo-ext-proc 81. CUDA/KVBM targets rely on
CI.

## Related Issues

Part of #10572. The follow-up PR completes the `nvext` and header surface described there.


ZhuohaoNi:
Rescoped to the router per @michaelfeil's feedback on #10572, and force-pushed as a
single commit.

- This PR: 17 files, +453/-3 — the queue-cap boost, `RouterRequest::New`, and tests.
- Follow-up (branch ready, not opened yet): 13 files, +253/-43 — `nvext`, the header, the
  preprocessor lift, ext-proc, the C bindings, and docs.

Call sites outside the router pass `0` until the follow-up lands, so behavior is
unchanged.

This also folds in the earlier fixes: the standalone selection core had test
constructions missing the new field, which is feature-gated and so invisible to
`cargo test -p dynamo-kv-router` but not to the workspace build, plus two ext-proc clippy
lints that only surface under `-D warnings`. @coderabbitai's note about the `epp.rs`
header override now belongs to the follow-up, where that path is actually wired.

The Pre Merge suite still needs a maintainer to approve the run for this commit.

PR #12214, opened 2026-07-27
feat(vllm): support multimodal sidecar requests

connorcarpenter15:
## Summary

Part 4 of a 4-PR stack.

- Convert Dynamo multimodal inputs into vLLM `MediaItem` messages.
- Map canonical `mm_hashes` to `MediaItem.uuid` so repeated images use the same keys as vLLM KV events.
- Carry multimodal prompt token IDs through the private preprocessor-to-sidecar handoff.
- Support aggregate and disaggregated multimodal generation without duplicating media processing.
- Keep this layer free of mocker changes.

## Stack

1. [#12734 — split gRPC services and Control discovery](https://github.com/ai-dynamo/dynamo/pull/12734)
2. [#12736 — preserve decode handoff on cancellation](https://github.com/ai-dynamo/dynamo/pull/12736)
3. [#12735 — deterministic DP and KV routing](https://github.com/ai-dynamo/dynamo/pull/12735)
4. [#12214 — multimodal sidecar requests](https://github.com/ai-dynamo/dynamo/pull/12214)

**Base:** [#12735](https://github.com/ai-dynamo/dynamo/pull/12735)

## Validation

- `cargo test -p dynamo-sidecar-common -p dynamo-vllm-sidecar -p dynamo-vllm-mocker` — 38 tests passed
- `cargo clippy -p dynamo-sidecar-common -p dynamo-vllm-sidecar -p dynamo-vllm-mocker --all-targets -- -D warnings`
- The cumulative implementation was previously exercised across the planned 2×H100 aggregate, cancellation, DP2/KV-routing, disaggregated, KVBM, multimodal, and recovery matrix. This stack split preserves that implementation; the rebase added only an unrelated upstream documentation commit.

### DIFF
--- a/lib/sidecar/vllm/src/convert.rs
+++ b/lib/sidecar/vllm/src/convert.rs
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use dynamo_backend_common::{
-    DisaggregationMode, DynamoError, GuidedDecodingOptions, LLMEngineOutput, PrefillResult,
-    PreprocessedRequest, StopReason, TopLogprob, usage,
+    DisaggregationMode, DynamoError, GuidedDecodingOptions, LLMEngineOutput, MultimodalData,
+    PrefillResult, PreprocessedRequest, StopReason, TopLogprob, usage,
 };
 
 use crate::client;
@@ -11,6 +11,9 @@ use crate::json::{json_to_struct, struct_to_json};
 use crate::proto as pb;
 
 const VLLM_LOGPROB_FLOOR: f64 = -9999.0;
+const MULTIMODAL_PROMPT_TOKEN_IDS_KEY: &str = "_dynamo_sidecar_multimodal_prompt_token_ids";
+const MM_HASHES_KEY: &str = "mm_hashes";
+// Must match DYNAMO_CACHE_SALT_PREFIX in lib/kv-router/src/zmq_wire/extra_keys.rs.
 const DYNAMO_CACHE_SALT_PREFIX: &str = "dynamo-cache-salt:";
 
 pub(crate) fn build_generate_request(
@@ -20,6 +23,26 @@ pub(crate) fn build_generate_request(
 ) -> Result<pb::GenerateRequest, DynamoError> {
     validate_request(&request, mode)?;
 
+    let has_media = request
+        .multi_modal_data
+        .as_ref()
+        .is_some_and(|media| media.values().any(|items| !items.is_empty()));
+    // Decode reuses the prefill-expanded tokens without reprocessing media.
+    let forwarded_mm_uuids = if has_media && !mode.is_decode() {
+        forwarded_mm_uuids(&request)?
+    } else {
+        None
+    };
+    let media = if mode.is_decode() {
+        Vec::new()
+    } else {
+        build_media(&request, forwarded_mm_uuids.as_deref())?
+    };
+    let mut prefill_result = request.prefill_result;
+    let mut token_ids = request.token_ids;
+    if mode.is_decode() && has_media {
+        token_ids = take_multimodal_prompt_token_ids(&mut prefill_result)?;
+    }
     let prompt_logprobs = request.output_options.prompt_logprobs;
     let output_logprobs = request.output_options.logprobs;
     let max_new_tokens = if mode.is_prefill() {
@@ -45,13 +68,19 @@ pub(crate) fn build_generate_request(
     let stop_conditions = request.stop_conditions;
     let mut extra_args = request.extra_args;
     consume_redundant_nvext(&mut extra_args, cache_salt.as_deref())?;
-    let kv = build_kv_parameters(extra_args, request.prefill_result, cache_salt, mode)?;
+    if has_media && let Some(serde_json::Value::Object(extra)) = extra_args.as_mut() {
+        // These fields are already represented by token_ids and media.
+        extra.remove("messages");
+        extra.remove("formatted_prompt");
+        extra.remove(MM_HASHES_KEY);
+    }
+    let kv = build_kv_parameters(extra_args, prefill_result, cache_salt, mode)?;
 
     Ok(pb::GenerateRequest {
         request_id,
         model: String::new(),
         prompt: Some(pb::generate_request::Prompt::TokenIds(pb::TokenIds {
-            ids: request.token_ids,
+            ids: token_ids,
         })),
         temperature: sampling.temperature,
         sampling: Some(pb::RandomSampling {
@@ -81,7 +110,7 @@ pub(crate) fn build_generate_request(
             ignore_eos: stop_conditions.ignore_eos.unwrap_or(false),
         }),
         response: Some(pb::ResponseOptions {
-            prompt_token_ids: prompt_logprobs.is_some(),
+            prompt_token_ids: prompt_logprobs.is_some() || (has_media && mode.is_prefill()),
             prompt_logprobs: prompt_logprobs.is_some(),
             prompt_candidates: prompt_logprobs.map(top_n_candidates).transpose()?,
             output_text: Some(true),
@@ -93,7 +122,7 @@ pub(crate) fn build_generate_request(
         truncate_prompt_tokens: 0,
         priority,
         session_id: None,
-        media: Vec::new(),
+        media,
     })
 }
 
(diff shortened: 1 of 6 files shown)


lib/mocker/servers/vllm/src/server_request.rs
```diff
@@ -383,6 +384,19 @@ impl KvTransferRole {
...
+    hasher.update(&seed.to_le_bytes());
+    hasher.update(request_id.as_bytes());
+    let mut bytes = [0u8; 16];
+    bytes.copy_from_slice(&hasher.finalize().as_bytes()[..16]);
+    // Mark the stable digest as an RFC 4122 variant/version-4 UUID. It remains
+    // deterministic; these bits only make diagnostics parse cleanly.
+    bytes[6] = (bytes[6] & 0x0f) | 0x40;
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
+    Uuid::from_bytes(bytes)
+}
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `bdc2a85898`. The vLLM mock Control Abort path now imports and calls `dynamo_mocker::live::stable_request_uuid`, and the duplicate local implementation was removed. No new UUID test was added because request admission already exercises the shared helper and the compiler now enforces the single implementation. The focused sidecar/mocker suite passes (38 tests), and Clippy passes with warnings denied.

lib/mocker/servers/vllm/src/server.rs
```diff
@@ -84,8 +86,57 @@ impl VllmMockerService {
...
+                .unwrap_or_default(),
+            max_batched_tokens: engine_args
+                .max_num_batched_tokens
+                .map(u64::try_from)
+                .transpose()
+                .map_err(|_| {
+                    anyhow::anyhow!("max_num_batched_tokens exceeds the Control API range")
+                })?
+                .unwrap_or_default(),
+            supports_explicit_data_parallel_rank: true,
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `bdc2a85898`. `PreparedRequest::new` now rejects any supplied `data_parallel_rank` other than the mock server rank instead of silently scheduling it on rank zero. The new wire-admission regression passes along with all 38 focused sidecar/mocker tests.

lib/sidecar/vllm/src/convert.rs
```diff
@@ -38,18 +72,27 @@ pub(crate) fn build_generate_request(
...
+    if has_media && let Some(serde_json::Value::Object(extra)) = extra_args.as_mut() {
+        // The frontend retains the original chat messages and rendered prompt
+        // for multimodal bookkeeping after it has already produced token_ids
+        // and multi_modal_data. The gRPC request carries those preprocessed
+        // fields directly; these values are not vLLM engine options.
+        extra.remove("messages");
+        extra.remove("formatted_prompt");
+        extra.remove(MM_HASHES_KEY);
+    }
+    let kv = build_kv_parameters(extra_args, prefill_result, cache_salt, mode)?;
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed the valid cache-namespace case in `bdc2a85898` without silently discarding unsupported semantics. The sidecar now validates that `extra_args.nvext.cache_salt` matches the canonical `routing.cache_namespace`, removes that redundant frontend envelope, and continues to reject remaining `nvext`, sampling, or reasoning passthrough fields that the pinned gRPC protocol cannot preserve. Matching text and multimodal requests pass; mismatches fail before RPC submission.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -146,87 +153,158 @@ impl LLMEngine for VllmSidecarEngine {
...
-                    biased;
-                    _ = cancellation.as_mut() => {
-                        yield Ok(cancelled(&state));
-                        break;
+                let message = if request_cancelled {
+                    tokio::select! {
+                        biased;
+                        _ = shutdown_cancellation.as_mut() => None,
+                        message = stream.message() => Some(message),
+                    }
```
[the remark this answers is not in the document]
connorcarpenter15: Verified the liveness concern, but the proposed bounded stream drop is unsafe. Before the first decode output, a timeout cannot distinguish a completed NIXL receive from an in-flight handoff. Dropping the stream would make vLLM auto-abort and can strand remote KV, violating the explicit prefill-router and backend-common cancellation invariant. The current completion-gated behavior remains covered by `decode_cancellation_waits_for_submission_and_first_token`; safe bounded cancellation needs an authoritative transfer-complete signal or guaranteed server-side cleanup.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -86,22 +87,24 @@ impl VllmSidecarEngine {
...
-        let engine = Self::new(endpoint, model.clone(), mode, transport);
-        let (tool_call_parser, reasoning_parser) = if mode.is_prefill() {
-            (None, None)
-        } else {
-            (
-                args.sidecar.common.dyn_tool_call_parser,
-                args.sidecar.common.dyn_reasoning_parser,
-            )
-        };
+        let engine = Self::new(endpoint, model.clone(), mode, transport, startup_deadline);
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `bdc2a85898`. The engine no longer stores the parse-time `Instant`: bootstrap uses one deadline, and `start()` derives a fresh absolute deadline from `self.transport.startup_deadline`, shared across connection, health, and Control discovery. The focused fresh-budget regression and full sidecar/mocker suites pass.

lib/sidecar/vllm/proto/control.proto
```diff
@@ -0,0 +1,78 @@
...
+service Control {
+  rpc GetServerInfo (GetServerInfoRequest) returns (ServerInfo) {}
+  rpc GetModelInfo (GetModelInfoRequest) returns (ModelInfo) {}
+  rpc Abort (AbortRequest) returns (AbortResponse) {}
+  rpc GetKvEventSources (GetKvEventSourcesRequest) returns (GetKvEventSourcesResponse) {}
+}
+
+message GetServerInfoRequest {}
+
+message ServerInfo {
```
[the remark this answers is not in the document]
connorcarpenter15: Verified byte-level provenance; no change is needed. `cmp` reports the vendored file and pinned vLLM source as identical, and both SHA-256 values are `cd4e7a8043f19d05929a2f59f5a5442894a037ef2d65832d3f7099992b1f1dbd`, matching `proto/README.md`. The blank line exists in both files.

lib/sidecar/vllm/src/model.rs
```diff
@@ -1,20 +1,125 @@
...
+        }
+        let reasoning_parser = nonempty(model.reasoning_parser);
+        let tool_call_parser = nonempty(model.tool_call_parser);
+        let identity = ModelIdentity {
+            source: source.clone(),
+            served_name: served_name.clone(),
+            aliases: model.served_model_aliases,
+            reasoning_parser: reasoning_parser.clone(),
+            tool_call_parser: tool_call_parser.clone(),
+        };
```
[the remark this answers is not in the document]
connorcarpenter15: No change is needed for the current producer. vLLM returns aliases from its configured `Vec` via `served_model_names().iter().skip(1)`, so the two Control calls preserve deterministic order; no unordered collection is involved. Exact identity comparison therefore detects real configuration changes. Canonicalizing aliases could be a future protocol choice, but it does not fix a current failure and would weaken the existing identity check.

lib/sidecar/vllm/tests/executable.rs
```diff
@@ -19,14 +19,17 @@ fn executable_exposes_native_grpc_configuration() {
...
         "--grpc-connect-attempt-timeout-secs",
         "--grpc-retry-interval-secs",
         "--grpc-startup-deadline-secs",
     ] {
         assert!(stdout.contains(flag), "missing {flag} in help output");
     }
+    assert!(
+        !stdout.contains("--model-path"),
+        "model identity must be discovered through vllm.Control"
+    );
```
connorcarpenter15: remove this

lib/sidecar/vllm/Dockerfile
connorcarpenter15, line 1: remove changes here

lib/mocker/servers/vllm/src/server_tests.rs
```diff
@@ -58,6 +58,15 @@ fn preparation_is_deterministic() {
...
 
+#[test]
+fn preparation_rejects_an_unconnected_data_parallel_rank() {
+    let mut wrong_rank = request("wrong-rank");
+    wrong_rank.data_parallel_rank = Some(DP_RANK + 1);
+    let error = PreparedRequest::new(wrong_rank, &MockerServerConfig::default()).unwrap_err();
+    assert_eq!(error.code(), tonic::Code::InvalidArgument);
+    assert!(error.message().contains("data_parallel_rank"));
+}
+
```
connorcarpenter15: remove

lib/sidecar/vllm/src/tests.rs
```diff
@@ -161,6 +191,105 @@ impl pb::generate_server::Generate for FakeGenerate {
...
+        .expect_err("DP routing accepted a server that silently discards the rank");
+    assert!(error.to_string().contains("explicit data-parallel rank"));
+
+    let mut single_rank = server_info();
+    single_rank.supports_explicit_data_parallel_rank = false;
+    single_rank.parallelism.as_mut().unwrap().data_parallel_size = 1;
+    DiscoveredModel::from_proto(model_info(), single_rank)
+        .expect("single-rank serving does not require explicit rank routing");
+}
+
```
connorcarpenter15: remove

lib/sidecar/vllm/src/tests.rs
```diff
@@ -397,19 +544,47 @@ fn request() -> PreprocessedRequest {
...
 fn engine(endpoint: &str, mode: DisaggregationMode, connections: usize) -> VllmSidecarEngine {
+    engine_with_model(endpoint, mode, connections, model_info())
+}
+
+fn engine_with_model(
+    endpoint: &str,
+    mode: DisaggregationMode,
+    connections: usize,
+    model: pb::ModelInfo,
+) -> VllmSidecarEngine {
```
connorcarpenter15: Why the split here? Is the `engine()` function used anywhere `engine_with_model()` cannot be used?

lib/sidecar/vllm/src/tests.rs
```diff
@@ -474,16 +676,237 @@ async fn aggregated_generation_converts_request_stream_and_usage() {
...
+    let mut changed = model_info();
+    changed.model_id = "different-model".to_string();
+    *server.service.model_info_override.lock().await = Some(changed);
+
+    let error = engine
+        .start(0)
+        .await
+        .expect_err("startup accepted a changed model identity");
+    assert!(error.to_string().contains("model identity changed"));
+}
```
connorcarpenter15: remove

lib/sidecar/vllm/src/tests.rs
```diff
@@ -474,16 +676,237 @@ async fn aggregated_generation_converts_request_stream_and_usage() {
...
+        .health
+        .set_service_status(INFERENCE_SERVICE, HealthServingStatus::Serving)
+        .await;
+
+    tokio::time::timeout(std::time::Duration::from_secs(2), start)
+        .await
+        .expect("startup did not observe SERVING")
+        .expect("startup task")
+        .expect("start");
+}
```
connorcarpenter15: remove

lib/sidecar/vllm/src/tests.rs
```diff
@@ -474,16 +676,237 @@ async fn aggregated_generation_converts_request_stream_and_usage() {
...
+        DisaggregationMode::Aggregated,
+        transport,
+    );
+    tokio::time::sleep(std::time::Duration::from_millis(550)).await;
+
+    tokio::time::timeout(std::time::Duration::from_secs(2), engine.start(0))
+        .await
+        .expect("startup did not receive a fresh deadline")
+        .expect("start");
+}
```
connorcarpenter15: remove

lib/sidecar/vllm/src/tests.rs
```diff
@@ -474,16 +676,237 @@ async fn aggregated_generation_converts_request_stream_and_usage() {
...
+    let error = tokio::time::timeout(
+        std::time::Duration::from_millis(250),
+        client.wait_for_services(&[CONTROL_SERVICE], deadline, transport.retry_interval),
+    )
+    .await
+    .expect("missing health registration was retried")
+    .expect_err("missing health registration was accepted");
+    assert!(error.to_string().contains("standard gRPC health API"));
+}
+
```
connorcarpenter15: remove

lib/sidecar/vllm/tests/executable.rs
connorcarpenter15: Removed the negative `--model-path` help assertion in 9b7346ba3f.

lib/mocker/servers/vllm/src/server_tests.rs
connorcarpenter15: Removed the mocker's unconnected data-parallel-rank test in 9b7346ba3f.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Removed the sidecar explicit-rank capability unit test in 9b7346ba3f.

lib/sidecar/vllm/Dockerfile
connorcarpenter15: Reverted the Dockerfile comment change in 9b7346ba3f; it again points to the retained aggregate deployment example.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Consolidated the helpers in 9b7346ba3f. There is now one `engine(endpoint, mode, connections, model)` helper, and each caller supplies the discovered model explicitly.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Removed the startup health-gating test and its now-unused mutable health scaffolding in 9b7346ba3f.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Removed the startup model-identity-change test in 9b7346ba3f.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Removed the missing health-registration test in 9b7346ba3f.

lib/sidecar/vllm/src/tests.rs
connorcarpenter15: Removed the fresh startup-deadline test in 9b7346ba3f.

tanmayv25 (approved):

lib/sidecar/vllm/src/convert.rs
```diff
@@ -20,6 +23,26 @@ pub(crate) fn build_generate_request(
 ) -> Result<pb::GenerateRequest, DynamoError> {
     validate_request(&request, mode)?;
 
+    let has_media = request
```
tanmayv25, line 26: **[AI-generated review — please verify before acting]**

Nit / simplification: this `has_media` predicate — `multi_modal_data.as_ref().is_some_and(|m| m.values().any(|i| !i.is_empty()))` — is duplicated verbatim in three places: here, in `ResponseState::new` further down this file, and in the capability gate in `engine.rs`. All three must stay in lockstep — if the notion of "has media" ever changes (a new modality, or empty-vs-present rules), the engine's gate can silently drift from what `convert.rs` actually forwards and from what `ResponseState` expects.

Consider extracting a single `fn has_media(request: &PreprocessedRequest) -> bool` and calling it at all three sites.

<sub>🤖 Generated by an AI reviewer.</sub>

lib/sidecar/vllm/src/engine.rs
```diff
@@ -186,6 +186,17 @@ impl LLMEngine for VllmSidecarEngine {
         request: dynamo_backend_common::PreprocessedRequest,
         ctx: GenerateContext,
     ) -> Result<BoxStream<'static, Result<LLMEngineOutput, DynamoError>>, DynamoError> {
+        if request
```
tanmayv25, line 189: **[AI-generated review — please verify before acting]**

Test-coverage gap: now that `validate_request` no longer rejects `multi_modal_data`, this guard is the only thing preventing media from being forwarded to a text-only backend. If it ever regresses, a multimodal request silently reaches a non-multimodal engine instead of failing before the RPC.

`unsupported_features_fail_before_rpc_submission` already runs with a `supports_multimodal = false` model, but only feeds `mm_processor_kwargs` (still routed through `validate_request`) — it never exercises this guard. Consider adding a request with `multi_modal_data` set to that test so both the "does not advertise multimodal support" rejection and the "no RPC submitted" invariant are covered.

<sub>🤖 Generated by an AI reviewer.</sub>

lib/sidecar/vllm/src/convert.rs
tanmayv25, line 26: **Simplicity: the "has media" predicate is duplicated verbatim three times in this diff.**

`request.multi_modal_data.as_ref().is_some_and(|m| m.values().any(|items| !items.is_empty()))` appears in `build_generate_request` (convert.rs:26), `ResponseState::new` (convert.rs:644), and `VllmSidecarEngine::generate` (engine.rs:189). All three gate multimodal behavior (media forwarding, prompt-token accounting, model-capability reject) and must stay in lockstep — if the definition of "has media" changes in one place and not the others, behavior diverges silently (e.g. engine forwards media but `ResponseState` treats the request as text-only, producing prompt-token-count mismatch protocol errors on the expanded media tokens).

**Suggestion:** extract one `pub(crate) fn request_has_media(request: &PreprocessedRequest) -> bool` in convert.rs and call it from all three sites.

_— from an automated multi-agent code-review pass (expert lenses + adversarial verification over this PR's diff)._

lib/sidecar/vllm/src/engine.rs
```diff
@@ -186,6 +186,17 @@ impl LLMEngine for VllmSidecarEngine {
...
+            .multi_modal_data
+            .as_ref()
+            .is_some_and(|media| media.values().any(|items| !items.is_empty()))
+            && !self.model.supports_multimodal
```
tanmayv25, line 193: **Test coverage: the new `supports_multimodal` rejection guard has no coverage.**

`validate_request` no longer rejects `multi_modal_data` (that branch was removed in convert.rs), so this guard is now the *only* thing stopping media from being forwarded to a text-only vLLM model. Every test either sends no media or overrides `supports_multimodal = true` (the multimodal test sets `discovered.supports_multimodal = true`; `model_info()` defaults to `false` but no media test uses it). If this guard regresses, media silently reaches a non-multimodal engine with no test catching it.

**Suggestion:** add a negative case (e.g. extend `unsupported_features_fail_before_rpc_submission`, which already uses `model_info()` with `supports_multimodal = false`) that sends a request with `multi_modal_data` and asserts it is rejected before RPC submission with the "does not advertise multimodal support" error.

_— from an automated multi-agent code-review pass (expert lenses + adversarial verification over this PR's diff)._

tanmayv25 (commented):
Automated multi-agent code-review pass over this PR's diff (framing + expert lenses, each finding adversarially verified). Findings below are all **minor / non-blocking**.

furionw (commented):
Do we have e2e test for sidecar agg, P/D in general?

can we also have ones for multimodal agg and multimodal P/D ?

connorcarpenter15:
We don’t have any tests for sidecar yet (including E2E). We have some launch scripts and deploy examples, but only for regular agg/disagg. CI tests and more extensive launch examples will be added prior to 1.5.0 once we work it out with ops. We will add multimodal support at that time.

furionw (approved):
Thank you

PR #12226, opened 2026-07-27
fix(vllm): checkpoint communicator state for snapshots

galletas1712:
## Summary

A process-level checkpoint captures a vLLM worker while its communicators
(FlashInfer, NCCL) hold live state that CRIU cannot restore meaningfully, so the
restored process comes back with communicators pointing at a world that no
longer exists.

This wraps the existing snapshot sleep/wake with vLLM's communicator
checkpoint hooks:

```text
sleep                  native vLLM: pauses and drains scheduling
checkpoint_prepare     communicators tear down
  << CRIU dump / restore >>
checkpoint_restore     communicators rebuild
wake_up                native vLLM: remaps memory, resumes scheduling
```

The ordering matters. `sleep()` already owns the scheduler pause, so preparation
runs on a quiesced engine, and `wake_up()` is not reached until communicators are
back. That is also why this does not need vllm-project/vllm#50036.

Only the snapshot pause controller passes `prepare_for_process_checkpoint=True`.
Ordinary admin sleep/wake and the GMS shadow-mode failover path construct their
own controller without it, so their behaviour is unchanged.

Failures fail closed: if `checkpoint_prepare()` or `checkpoint_restore()` raises,
the exception propagates and the process dies rather than continuing on
half-rebuilt communicators.

## Changes

- `vllm/handlers.py` — `VllmEnginePauseController` gains the opt-in flag and
  calls the two hooks around the existing sleep/wake
- `vllm/snapshot.py` — the snapshot controller opts in
- `vllm/tests/test_vllm_sleep_wake_handlers.py` — call ordering, plus both
  failure directions

## vLLM requirement

Needs `AsyncLLM.checkpoint_prepare()` and `checkpoint_restore()` from
vllm-project/vllm#46877 (merged upstream 2026-07-26 as `b5b61c62`).

**Minimum vLLM is 0.27.0**, and every runtime image in `container/context.yaml`
clears it: CUDA and CPU are on `v0.28.0`, XPU on `v0.27.1`. The hooks are absent
from 0.26.0, which is what an earlier revision of this branch pinned:

```python
# vllm/v1/engine/async_llm.py @ v0.28.0
async def checkpoint_prepare(self) -> None:
    await self.collective_rpc("checkpoint_prepare")

async def checkpoint_restore(self) -> None:
    await self.collective_rpc("checkpoint_restore")
```

An earlier revision of this PR was based on a tree pinning 0.26.0, where the
hooks were genuinely missing; it has been rebased onto `main`.

## Validation

- Rebased onto `main`; patch applies with no conflicts
- `black`, `isort`, `flake8` clean at the versions pinned in
  `.pre-commit-config.yaml`
- Hook presence and signatures verified against the upstream `v0.27.0` and
  `v0.28.0` tags
- Focused pytest collection is unavailable locally because the compiled
  `dynamo._core` extension is not installed; CI's backend image provides it

### DIFF
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -398,9 +398,12 @@ class VllmEnginePauseController:
                     "Failed to resume generation after native vLLM sleep failure"
                 )
             raise
-        self._is_paused = True
+        # Prepare before recording the pause: a failed prepare must not leave
+        # the controller claiming paused, or the next resume would issue a
+        # checkpoint_restore with no matching prepare.
         if self._prepare_for_process_checkpoint:
             await self._engine_client.checkpoint_prepare()
+        self._is_paused = True
         return True
 
     async def resume(self, tags: list[str] | None = None) -> bool:
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -358,8 +358,14 @@ async def _deferred_abort_guard(
 
 
 class VllmEnginePauseController:
-    def __init__(self, engine_client: Any):
+    def __init__(
+        self,
+        engine_client: Any,
+        *,
+        prepare_for_process_checkpoint: bool = False,
+    ):
         self._engine_client = engine_client
+        self._prepare_for_process_checkpoint = prepare_for_process_checkpoint
         self._is_paused = False
         self._generation_paused = False
 
@@ -393,6 +399,8 @@ class VllmEnginePauseController:
                 )
             raise
         self._is_paused = True
+        if self._prepare_for_process_checkpoint:
+            await self._engine_client.checkpoint_prepare()
         return True
 
     async def resume(self, tags: list[str] | None = None) -> bool:
@@ -400,6 +408,8 @@ class VllmEnginePauseController:
             return False
 
         if self._is_paused:
+            if self._prepare_for_process_checkpoint:
+                await self._engine_client.checkpoint_restore()
             if tags is None:
                 await self._engine_client.wake_up()
             else:
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -329,8 +329,14 @@ async def _deferred_abort_guard(
 
 
 class VllmEnginePauseController:
-    def __init__(self, engine_client: Any):
+    def __init__(
+        self,
+        engine_client: Any,
+        *,
+        prepare_for_process_checkpoint: bool = False,
+    ):
         self._engine_client = engine_client
+        self._prepare_for_process_checkpoint = prepare_for_process_checkpoint
         self._is_paused = False
         self._generation_paused = False
 
@@ -364,6 +370,8 @@ class VllmEnginePauseController:
                 )
             raise
         self._is_paused = True
+        if self._prepare_for_process_checkpoint:
+            await self._engine_client.checkpoint_prepare()
         return True
 
     async def resume(self, tags: list[str] | None = None) -> bool:
@@ -371,6 +379,8 @@ class VllmEnginePauseController:
             return False
 
         if self._is_paused:
+            if self._prepare_for_process_checkpoint:
+                await self._engine_client.checkpoint_restore()
             if tags is None:
                 await self._engine_client.wake_up()
             else:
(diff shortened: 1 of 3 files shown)


components/src/dynamo/vllm/handlers.py
```diff
@@ -371,13 +377,17 @@ async def pause(self, *args: object) -> bool:
                 )
             raise
         self._is_paused = True
+        if self._prepare_for_process_checkpoint:
+            await self._engine_client.checkpoint_prepare()
```
matteso1, line 405: On a vLLM without the vllm-project/vllm#46877 hooks this is a bare AttributeError at pause time. Worth a hasattr check with an error naming the minimum vLLM?

components/src/dynamo/vllm/handlers.py
```diff
@@ -371,13 +377,17 @@ async def pause(self, *args: object) -> bool:
...
         return True
 
     async def resume(self, tags: list[str] | None = None) -> bool:
         if not self._is_paused and not self._generation_paused:
             return False
 
         if self._is_paused:
+            if self._prepare_for_process_checkpoint:
+                await self._engine_client.checkpoint_restore()
```
matteso1, line 415: If this or wake_up() raises, the restored process dies instead of replaying communicator steps, which matches the fail-closed claim. Can we pin that with a test? checkpoint_restore raising, controller stays paused, exception propagates.

matteso1 (commented):
Traced the full path across both repos (run_lifecycle sentinel flow, the pause controller, and the vLLM side down to the FlashInfer workspace functions). The ordering holds without vllm-project/vllm#50036, the fail-closed behavior is real, and the path is dormant without DYN_SNAPSHOT_CONTROL_DIR, so merge risk looks low to me. Two small asks inline; the pending dependency-complete E2E reads as the real acceptance bar.

components/src/dynamo/vllm/handlers.py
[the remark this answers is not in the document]
mohammedabdulwahhab: Fixed. `is_paused` relocated to after the call to checkpoint_prepare in latest

mohammedabdulwahhab: Yep, test added!

components/src/dynamo/vllm/handlers.py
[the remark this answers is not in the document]
mohammedabdulwahhab: vllm version in Dynamo is now upto date and has these hooks

mohammedabdulwahhab: Yep, check added!

components/src/dynamo/vllm/snapshot.py
```diff
@@ -42,7 +42,10 @@ async def prepare_snapshot_engine(
     gc.collect()
     snapshot_controller = EngineSnapshotController(
         engine=engine,
-        pause_controller=VllmEnginePauseController(engine[0]),
+        pause_controller=VllmEnginePauseController(
+            engine[0],
+            prepare_for_process_checkpoint=True,
+        ),
```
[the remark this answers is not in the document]
mohammedabdulwahhab: stale

mohammedabdulwahhab (approved):

components/src/dynamo/vllm/snapshot.py
```diff
@@ -39,10 +39,26 @@ async def prepare_snapshot_engine(
...
+    # after sleep() had already released the engine's memory.
+    checkpoint_hooks = all(
+        hasattr(engine[0], hook)
+        for hook in ("checkpoint_prepare", "checkpoint_restore")
+    )
+    if not checkpoint_hooks:
+        logger.warning(
+            "This vLLM build has no AsyncLLM.checkpoint_prepare/checkpoint_restore; "
+            "snapshotting without communicator checkpointing. "
+            "Requires vLLM 0.27.0 or newer."
```
GuanLuo, line 52: If it will be available soon, can we monkey-patch engine with no-op `checkpoint_prepare` and `checkpoint_restore` if undefined? Then in the handler these two functions can be called unconditionally. We will clean up the check-and-patch here once we advance the vLLM version

GuanLuo (approved):

components/src/dynamo/vllm/snapshot.py
mohammedabdulwahhab: It's already available. No monkey patch needed!

biswapanda:
lgtm 👍

PR #12228, opened 2026-07-27
feat(recipes): publish Kimi-K3 to main

BenHamm:
## Summary

The Kimi-K3 recipes shipped only on `release/1.4.0-kimi-k3-dev.1`, so `recipes/` on `main` shows no Kimi-K3 support and the Available Recipes table has no row for it. The Fern docs page is currently the only discoverable entry point. The runtime containers are published, so nothing gates the source.

This PR:

1. Brings `recipes/kimi-k3/**` onto `main`, byte-identical to the release branch (7 files: README, model-cache assets, 4 deploy manifests).
2. Adds the missing **Available Recipes** row in `recipes/README.md`, per `recipes/CONTRIBUTING.md`.
3. Repoints the docs page Source links from the release branch back to `main`, now that the assets live there, and drops the branch-pinning note.

**Scope note:** this is recipe manifests and docs only. It touches no Rust and no `Cargo.toml`, so it does **not** bring the release branch's `frontend-crates` `feat/Kimi-K3` git dependency onto `main`. Same shape as GLM-5.2, whose recipe sits on `main` pinning a dev-tagged runtime image. Publishing the Rust crates and repointing the release branch is a separate follow-up.

## Validation

- `python3 docs/fern/recipes/_catalog/validate.py` — **`OK: all catalog checks passed`**. First fully clean run; the four `recipes/kimi-k3/**` asset paths previously could not resolve because the tree was not on `main`.
- `fern check` — 0 errors.
- Recipe files verified identical to `release/1.4.0-kimi-k3-dev.1` (no content drift).

Note: `lychee` may flag the repointed `blob/main/recipes/kimi-k3/...` links until this merges, since those paths only exist on this branch. They resolve on merge.


BenHamm:
## Review triage

Verified every bot finding against the manifests. Three were real and are fixed in cd0d741df9; the rest are either incorrect or belong to engineering rather than this publish PR.

### Fixed

1. **`$(MODEL_PATH)` unset in the disagg-gb200 frontend** (flagged by both bots, correctly). The frontend container defines no env, so Kubernetes passes the placeholder through literally. `agg-gb200`, `agg-gb300`, and `disagg-gb300` all omit `--model-path` entirely, as does the GLM-5.2 disaggregated recipe. Dropped the argument to match.
2. **Broken Kubernetes guide link.** `../../docs/kubernetes/README.md` was moved by the docs restructure; repointed to `../../docs/fern/kubernetes/quickstart.mdx`. This one only breaks on `main` — the path is still valid on the release branch, so it is a genuine artifact of publishing here.
3. **Quick Start described the model-cache PVC flow as universal.** It is GB300-only: the GB200 manifests mount node-local storage and serve `--model /models/model_weight` with `HF_HUB_OFFLINE=1`, so steps 2-3 alone leave them undeployable. Split per SKU, with both GB200 options spelled out. The Fern page carries the same correction.

### Not applied — incorrect

- **"Recipe README needs Fern frontmatter and no body H1."** Recipe READMEs live outside `docs/` and use the HTML-comment SPDX form; `glm-5.2`, `kimi-k2.6`, and `inkling` all have no frontmatter and a body H1. Applying this would make Kimi-K3 the only inconsistent one.

### Not applied — engineering decisions, flagged rather than changed

Both are real asymmetries, but each changes routing behaviour on manifests that have already been QA'd, so they should be eng calls with a test behind them, not drive-by edits in a publish PR:

- **GB300 frontends omit `--router-kv-events`** while both GB200 frontends pass it, even though the GB300 workers publish KV events. If intentional, worth a comment in the manifest; if not, it means GB300 falls back to approximate routing.
- **Disaggregated decode workers never set `--kv-events-config`** (prefill does, in both disagg profiles), so the router cannot see decode-side cache state.

cc @NVShreyas @chw001 — the two items above are yours to confirm. The `$(MODEL_PATH)` fix should also be applied to `release/1.4.0-kimi-k3-dev.1`, since the bug ships there too.

### Not applied — hardening, out of scope

`activeDeadlineSeconds` on the download Job, `automountServiceAccountToken: false`, and HF-token placement. All reasonable, none a defect; happy to take them in a follow-up if the recipe owners want them.

nealvaidya (approved):

jthomson04 (approved):

PR #11664, opened 2026-07-14
feat: add GMS shadow-failover observability (metrics + dashboard)

mohammedabdulwahhab:
Adds observability for GMS shadow-engine failover (intra-pod).

Main additions:
- Introduces `dynamo_component_engine_failover_*` metrics: Counters for failover attempts/successes + gauges to record current state and state transition timings. Metrics registry wired into existing Dynamo metrics via `register_prometheus_expfmt_callback`
- PodMonitor configuration to scrape metrics on failover setup. Previously misconfigured for failover pods.
- Grafana dashboard to inspect current state of the DGD

## Failover Metrics Introduced

| Metric | Type | Extra labels | Meaning |
|---|---|---|---|
| `_state` | gauge | `state` | 1-hot current state (init/standby/waking/active) |
| `_state_entered_timestamp_seconds` | gauge | | when the current state was entered (age = `time() - value`) |
| `_last_state_duration_seconds` | gauge | `state` | last completed duration of each state (`state="waking"` is the switch time) |
| `_transitions_total` | counter | `from_state`, `to_state` | state transitions |
| `_switch_attempts_total` | counter | | real failover promotions attempted (contended-lock wins; bootup excluded) |
| `_switch_success_total` | counter | | promotions that completed and began serving |

`_switch_attempts_total` / `_switch_success_total` are write-through persisted, so a caught failure survives the process death (derived failures = attempts − success).
  
  ## Grafana Dashboard
  
image

PR #12229, opened 2026-07-27
fix(vllm): pin gms_shadow_mode=False in worker-factory test mock

zhongdaor-nv:
## Summary

All six prefill-path tests in `components/src/dynamo/vllm/tests/test_vllm_worker_factory.py` (`TestPrefillRegistrationContract::test_prefill_registers_with_tokens[*]` and `test_prefill_serves_lora_lifecycle_endpoints_when_enabled[*]`) fail with:

```
ModuleNotFoundError: No module named 'gpu_memory_service'
  worker_factory.py:1163 in _create_prefill_worker
  worker_factory.py:767 in _maybe_create_failover_metrics
```

in any environment that does not bundle the optional `gpu-memory-service` wheel 

Root cause: `_make_config` returns a bare `Mock`, so `config.gms_shadow_mode` auto-creates as a **truthy** Mock attribute. The guard added in #11664 —

```python
if not config.gms_shadow_mode:
    return None
from gpu_memory_service.failover_lock.failover_metrics import ...
```

— therefore never takes its early return in these tests, and the lazy import of the optional package runs unconditionally. Standard CI images install the GMS wheel (`ENABLE_GPU_MEMORY_SERVICE=true`), which is why this never surfaced here; the tests were silently exercising the shadow-mode-on path either way.

Fix: pin `gms_shadow_mode: False` in `_make_config` defaults, matching the real `Config` default and the explicit `gms_shadow_mode=False` already used in `test_vllm_unit.py`. Tests that want the shadow-mode path can still opt in via `_make_config(gms_shadow_mode=True)`.

cc @mohammedabdulwahhab (#11664)

## Validation

- Reproduced nightly in dynamo-ci (image without the GMS wheel): 6/6 prefill-path factory tests fail with the traceback above; with this pin the guard takes the early return, so the optional import is no longer reached.
- Every config in this test file is built through `_make_config`, so the single default covers all affected tests.
- `python3 -m py_compile` passes; no production code touched.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12229" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Tests**
  * Improved test stability by explicitly disabling shadow mode in test configuration.
  * Added clarification to prevent unintended test behavior in environments without the GPU memory service.


furionw:
Thanks for working on this. How is the error reproduced?

asking because, IIUC, both PR and nightly build enable GMS. so I am wondering how you encountered the error.

furionw (approved):

PR #12233, opened 2026-07-27
perf(mocker): streamline native block pool hot paths

jthomson04:
## Overview

Streamline native-G1 mocker block-pool hot paths while preserving physical-capacity, prefix-reuse, and eviction semantics.

## Summary

- replace the timestamp-keyed inactive `BTreeMap` with an intrusive oldest-to-newest LRU
- resolve and reserve a destination's resident prefix in one block-pool traversal
- derive inactive membership from zero references and zero pins, avoiding a separate membership discriminant
- keep constant-time invariant checks in debug builds and exhaustive structural validation in unit tests
- validate intrusive-list updates before mutation so failed invariants cannot leave a partially updated list

## Details

The native block pool now stores optional predecessor and successor IDs directly in each cached copy. Inactive membership is derived from `refs == 0 && pins == 0`; active entries carry empty links. Pinning removes an inactive entry before incrementing its pin count, and insertion/removal validates the target, neighbors, endpoints, and length before changing the list.

`reserve_resident_prefix` walks candidates until the first partial block or cache miss, pins the resident prefix, and reserves the fresh suffix atomically. Its hit vector is preallocated from the layout length.

Ordinary debug builds perform only constant-time occupancy and endpoint checks. The full LRU/hash-index sweep is compiled only for tests and runs once per logical operation rather than inside eviction loops or twice during cached allocation.

### Performance evidence

The performance run predates the review-fix commit. It compared control `329aba5d0c91c4520c5ce8a1707310fa68fccc69` with performance commit `97f68ec59d9085d18944ffd9bfe2e7c9bd9e981f`. The follow-up commit `c87c93a5a38101dca8f4e938b041ee025d54c9ac` was not rebenchmarked.

Both arms used one native-G1 mocker with 32 workers on CPUs `12-23`, frontend/AIPerf on CPUs `0-11`, 128,000 blocks per worker, block size 16, concurrency 512, all 336 Weka trace entries, a 120-second measured phase, and basic `taskset` isolation only.

| Metric | Control | Performance commit | Delta |
|---|---:|---:|---:|
| Mocker CPU ms/request | 23.5446 | 22.4133 | **-4.80%** |
| Mocker CPU s / 1M modeled tokens | 0.301957 | 0.286989 | **-4.96%** |
| Throughput (req/s) | 255.429 | 258.795 | **+1.32%** |
| Peak RSS (MiB) | 7107.2 | 7183.8 | +1.08% |

Both arms registered all 32 workers and completed with zero request errors or cancellations. This was intentionally one accepted run per arm; unrelated host activity remained on the mocker CPU mask.

The aggregated-worker workload exercises the intrusive LRU but not disaggregated destination reservation, so the measured CPU improvement should primarily be attributed to the LRU change. Prefix fusion is correctness-tested but not performance-qualified by this run.

The measured RSS value reflects the earlier `Option<InactiveLinks>` representation. The review follow-up removes that outer discriminant, eliminating up to 31.25 MiB of embedded cached-copy state at the benchmark's aggregate 4,096,000-block capacity, but that reduction was not measured in a new E2E run.

1. `lib/mocker/src/cache/vllm_block_pool.rs`: cached-copy state, intrusive-list validation/mutation, and invariant-check boundaries.
2. `lib/mocker/src/kv_manager/vllm_backend.rs`: destination prefix reservation and the native-manager coverage for disabled prefix caching and partial-block termination.

## Related Issues

## Validation

- `cargo test -p dynamo-mocker` — 641 passed, 1 ignored
- pre-review release-like Python binding build with DWARF; build ID `7c1d0bf8bf0ea539407a6355e6f174c9b79de257`
- pre-review E2E gates: 32/32 workers, enforced affinities, zero AIPerf errors, zero cancellations, and full process-group teardown

### DIFF
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -16,7 +16,7 @@ new_key_type! {
     pub(crate) struct BlockCopyId;
 }
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 struct InactiveLinks {
     previous: Option<BlockCopyId>,
     next: Option<BlockCopyId>,
@@ -29,7 +29,7 @@ enum CopyState {
         hash: SequenceHash,
         refs: usize,
         pins: usize,
-        inactive: Option<InactiveLinks>,
+        inactive: InactiveLinks,
     },
 }
 
@@ -136,9 +136,9 @@ impl VllmBlockPool {
         candidates: impl IntoIterator<Item = SequenceHash>,
         total: usize,
     ) -> Option<ReserveOutcome> {
-        let mut hits = Vec::new();
+        let mut hits = Vec::with_capacity(total);
         for hash in candidates {
-            assert!(hits.len() < total, "prefix candidates exceed layout");
+            debug_assert!(hits.len() < total, "prefix candidates exceed layout");
             let Some(id) = self.first_copy(hash) else {
                 break;
             };
@@ -232,16 +232,20 @@ impl VllmBlockPool {
     }
 
     pub(crate) fn allocate_private(&mut self, reservation: &mut BlockReservation) -> BlockCopyId {
+        let id = self.allocate_private_unchecked(reservation);
+        self.debug_assert_invariants();
+        id
+    }
+
+    fn allocate_private_unchecked(&mut self, reservation: &mut BlockReservation) -> BlockCopyId {
         assert!(reservation.fresh > 0, "reservation has no fresh capacity");
         assert!(self.reserved > 0, "pool reserved-capacity underflow");
         reservation.fresh -= 1;
         self.reserved -= 1;
 
-        let id = self.copies.insert(BlockCopy {
+        self.copies.insert(BlockCopy {
             state: CopyState::Private,
-        });
-        self.debug_assert_invariants();
-        id
+        })
     }
 
     /// Allocate a transferred/computed full block directly into the cache.
@@ -251,14 +255,21 @@ impl VllmBlockPool {
         reservation: &mut BlockReservation,
         hash: SequenceHash,
     ) -> (BlockCopyId, bool) {
-        let id = self.allocate_private(reservation);
-        let became_visible = self.cache_private(id, hash);
+        let id = self.allocate_private_unchecked(reservation);
+        let became_visible = self.cache_private_unchecked(id, hash);
+        self.debug_assert_invariants();
         (id, became_visible)
     }
 
     /// Make a request-private computed full block available for prefix reuse.
     /// Returns whether this is the first resident physical copy of `hash`.
     pub(crate) fn cache_private(&mut self, id: BlockCopyId, hash: SequenceHash) -> bool {
+        let became_visible = self.cache_private_unchecked(id, hash);
+        self.debug_assert_invariants();
+        became_visible
+    }
+
+    fn cache_private_unchecked(&mut self, id: BlockCopyId, hash: SequenceHash) -> bool {
         let became_visible = !self.by_hash.contains_key(&hash);
         let Some(copy) = self.copies.get_mut(id) else {
             panic!("attempted to cache an unknown block copy")
@@ -271,10 +282,9 @@ impl VllmBlockPool {
             hash,
             refs: 1,
             pins: 0,
-            inactive: None,
+            inactive: InactiveLinks::default(),
         };
         self.by_hash.entry(hash).or_default().push(id);
-        self.debug_assert_invariants();
         became_visible
     }
 
@@ -360,25 +370,31 @@ impl VllmBlockPool {
         matches!(
             &self.copies[id].state,
             CopyState::Cached {
-                inactive: Some(_),
+                refs: 0,
+                pins: 0,
                 ..
             }
         )
     }
 
     fn pin(&mut self, id: BlockCopyId) {
-        let was_inactive = {
-            let CopyState::Cached { pins, inactive, .. } = &mut self.copies[id].state else {
+        let (was_inactive, new_pins) = {
+            let CopyState::Cached { refs, pins, .. } = &self.copies[id].state else {
                 panic!("prefix hash points to a private copy")
             };
-            *pins = pins
-                .checked_add(1)
-                .unwrap_or_else(|| panic!("pin count overflow"));
-            inactive.is_some()
+            (
+                *refs == 0 && *pins == 0,
+                pins.checked_add(1)
+                    .unwrap_or_else(|| panic!("pin count overflow")),
+            )
         };
         if was_inactive {
             self.remove_inactive(id);
         }
+        let CopyState::Cached { pins, .. } = &mut self.copies[id].state else {
+            unreachable!()
+        };
+        *pins = new_pins;
     }
 
     fn activate_pin(&mut self, id: BlockCopyId, expected_hash: SequenceHash) {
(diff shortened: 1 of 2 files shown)


lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -419,11 +538,107 @@ impl VllmBlockPool {
...
+            self.debug_assert_invariants();
             Some(hash)
         } else {
+            self.debug_assert_invariants();
             None
         }
     }
+
+    #[cfg(debug_assertions)]
+    fn debug_assert_invariants(&self) {
```
jthomson04, line 621: **The sweep makes the standard dev build quadratic.**

This is O(copies + hash-index + LRU) and allocates *two* `FxHashSet`s per call, and it runs on every `allocate_private`, `cache_private`, `release`, `cancel`, `activate_prefix`, and `evict_one`.

The documented onboarding build is `maturin develop --uv` with no `--release` (`AGENTS.md:97`, `README.md:210`, `.devcontainer/post-create.sh:105`), so `debug_assertions` is on. Default `num_gpu_blocks` is `16384`.

Concrete cost: a 4096-token request at `block_size=16` is 256 blocks -> 256 sweeps x ~33k operations, i.e. ~8M hash-set inserts for one request, roughly 150ms of pure assertion work. At the 128,000-block pool this PR's own benchmark uses, it's ~1.3s **per request**.

Every existing `debug_assert!` in this crate is O(1) (`radix_cache.rs:480`, `sglang_backend.rs:284`, `kvbm_backend.rs:826`), so this is a new cost class for the codebase — and an awkward one to ship inside a perf PR.

Suggestion: gate the full sweep behind an opt-in feature (`pool-invariants`) or `#[cfg(test)]`, and keep only O(1) `debug_assert!`s on the mutation paths (e.g. `inactive_head.is_none() == inactive_tail.is_none()`). Every unit test here uses a pool of 1-3 blocks, so `#[cfg(test)]` alone retains the full value at zero cost.

lib/mocker/src/kv_manager/vllm_backend.rs
```diff
@@ -346,13 +346,21 @@ impl VllmKvManager {
...
+                if self.enable_prefix_caching {
+                    self.pool.reserve_resident_prefix(
+                        blocks.iter().map_while(|block| match block {
+                            UniqueBlock::FullBlock(hash) => Some(*hash),
+                            UniqueBlock::PartialBlock(_) => None,
+                        }),
+                        blocks.len(),
+                    )
+                } else {
+                    self.pool.reserve(&[], blocks.len())
```
jthomson04, line 358: **`reserve_destination_at` has no automated coverage on the native path.**

This is the only caller of `reserve_resident_prefix` and the only behavior change on the disagg path, but nothing in the Rust suite reaches it:

- `vllm_backend.rs`'s own test module has no `reserve_destination_at` test.
- `scheduler/vllm/tests.rs`'s `destination_lifecycle` module never sets `.g1_backend(...)`, and `G1Backend`'s `#[default]` is `Kvbm` — so those tests exercise the KVBM path.
- `replay/offline/disagg_tests.rs` likewise never sets `Native`.

The two branches introduced here with no test behind them: this `enable_prefix_caching == false` arm, and a layout where a `PartialBlock` interrupts the full-block prefix. That `map_while` used to live in the named `resident_prefix` helper; inline in the acquire path it's easier to break silently.

The description already notes this path isn't perf-qualified — it isn't correctness-qualified either. Two `VllmKvManager`-level tests would close it.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -419,11 +538,107 @@ impl VllmBlockPool {
         };
         if remove_hash {
             self.by_hash.remove(&hash);
+            self.debug_assert_invariants();
```
jthomson04: **The sweep runs inside the eviction loop.**

`evict_one` asserts on both return paths, and it's called in a loop from `reserve_hits`/`reserve_fresh`, which then sweep again on completion — so a reservation needing *k* evictions pays *k+1* full sweeps.

Even if the sweep stays as-is, I'd hoist this out to the callers' tails; the intermediate states aren't independently interesting. Same pattern in `allocate_cached`, which sweeps twice (via `allocate_private` then `cache_private`) for one logical operation.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -120,6 +126,33 @@ impl VllmBlockPool {
...
+    }
+
+    /// Resolve the longest resident prefix and reserve its remaining fresh
+    /// suffix in one pool traversal.
+    pub(crate) fn reserve_resident_prefix(
+        &mut self,
+        candidates: impl IntoIterator<Item = SequenceHash>,
+        total: usize,
+    ) -> Option<ReserveOutcome> {
+        let mut hits = Vec::new();
```
jthomson04: `Vec::with_capacity(total)` would be better here. The old `reserve` collected from `prefix.iter().map(...)`, an `ExactSizeIterator`, so it made one exact allocation; this loop grows by pushes and pays log(n) reallocs on a long prefix.

`total` is a known upper bound and the caller already holds a `Vec<UniqueBlock>` of that length, so it's a free win in a change whose stated goal is removing a temporary vector.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -18,14 +16,20 @@ new_key_type! {
     pub(crate) struct BlockCopyId;
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct InactiveLinks {
```
jthomson04, line 20: On the 40 -> 48 byte growth and the follow-up you flagged: the fix is small enough that it may be worth folding in rather than touching this code twice.

LRU membership is exactly `refs == 0 && pins == 0`, so storing `InactiveLinks` unconditionally (16 bytes, both fields niche-packed, no outer discriminant) recovers the original 40 and drops the `.is_some()` / `Some(links)` matching throughout `pin`, `is_inactive`, `insert_inactive`, and `remove_inactive`.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +415,116 @@ impl VllmBlockPool {
...
+        let previous = self.inactive_tail;
+        if let Some(previous) = previous {
+            let CopyState::Cached {
+                inactive: Some(links),
+                ..
+            } = &mut self.copies[previous].state
+            else {
+                panic!("inactive tail is not linked")
+            };
+            assert!(links.next.replace(id).is_none());
```
jthomson04: Minor: both `insert_inactive` and `remove_inactive` mutate neighbours before their validating asserts — here the predecessor's `next` is written before line 435 checks the node wasn't already linked, and `remove_inactive` patches both neighbours before clearing the node.

If either assert fires the list is left structurally corrupted rather than merely inconsistent, and under pyo3 these become Python exceptions rather than aborts, so a caught panic leaves a live corrupt pool. The `BTreeMap` version was closer to atomic.

All "cannot happen" paths, so low severity — but validating first and mutating second costs nothing here.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +415,116 @@ impl VllmBlockPool {
...
-            inactive_at,
+            inactive,
         } = copy.state
         else {
             panic!("inactive LRU points to a private copy")
         };
         assert_eq!(refs, 0);
         assert_eq!(pins, 0);
-        assert_eq!(inactive_at, Some(timestamp));
+        assert_eq!(inactive, None);
```
jthomson04: This assert is now vacuous — `remove_inactive` did `inactive.take()` two lines up, so it can never fail.

The assertion it replaced (`inactive_at == Some(timestamp)`) was a real cross-structure check between the map key and the node. `remove_inactive` carries the meaningful version now (`assert_eq!(inactive.take(), Some(links))`), so I'd drop this one rather than leave it reading as protection that isn't there.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -120,6 +126,33 @@ impl VllmBlockPool {
...
+    /// Resolve the longest resident prefix and reserve its remaining fresh
+    /// suffix in one pool traversal.
+    pub(crate) fn reserve_resident_prefix(
+        &mut self,
+        candidates: impl IntoIterator<Item = SequenceHash>,
+        total: usize,
+    ) -> Option<ReserveOutcome> {
+        let mut hits = Vec::new();
+        for hash in candidates {
+            assert!(hits.len() < total, "prefix candidates exceed layout");
```
jthomson04: `debug_assert!` would be more consistent here — this runs on every iteration including ones that immediately `break`, and it can only fire if a caller passes more candidates than `total`, which the sole caller structurally cannot.

The cost is negligible either way; it just sits oddly against the rest of the PR moving invariants to debug-only.

jthomson04 (commented):
## Review

Two independent changes here: the intrusive LRU replacing the timestamp `BTreeMap`, and the fused `reserve_resident_prefix`. Both refactors look faithful to me.

I traced the prefix semantics specifically, since that's the one behavior change on the disagg path. The old `resident_prefix` used `map_while(FullBlock(h) if prefix_hit(h).is_some())`, terminating on a partial block *or* a miss. The new split — caller-side `map_while(FullBlock => Some, Partial => None)` composed with the pool-side `break` on a `first_copy` miss — is exactly equivalent, and `first_copy` is the same predicate `prefix_hit` wraps. LRU ordering (monotonic timestamp -> tail append), the `inactive_hits` dedup for duplicate hashes, pin-before-evict ordering, and the leaf-before-parent reversal in `cancel` are all preserved.

The tests are well targeted. `intrusive_lru_unlinks_middle_and_requeues_it_as_newest` hits the exact case a naive intrusive list corrupts, and `fused_prefix_reservation_stops_at_first_cache_miss` proves stop-at-miss indirectly but convincingly — hash `9` gets *evicted* rather than pinned, which only happens if the walk stopped at `8`.

The 40 -> 48 byte note in the description checks out: `Option<BlockCopyId>` is niche-packed to 8 bytes by slotmap's `NonZeroU32` version field, but that consumes the only niche, so `Option<InactiveLinks>` is 24 vs. the old `Option<u64>` at 16. The suggested follow-up is right — details inline.

Findings below, most significant first. Two I'd want addressed before merge (the debug-invariant cost, and the missing coverage on `reserve_destination_at`); the rest are cleanups that can ride along or follow.

One further note without a clean anchor: `pin()` increments `pins` to 1 *before* calling `remove_inactive`, so the "inactive => pins == 0" invariant is transiently violated inside that function. It's correct today only because `remove_inactive` doesn't run the sweep. Worth a comment, since the invariant work makes it tempting to add one there.

No security surface, no `unsafe`, and the intrusive list is entirely index-based through the slotmap.

*Note: I reviewed against the head tree rather than rebuilding, so the reported `fmt`/`clippy`/`test` results are taken as given.*

PR #12242, opened 2026-07-28
perf(mocker): drive only ready offline worker groups

jthomson04:
## Summary

- replace ordered maps for monotonic offline worker and rank IDs with direct-indexed, tombstoned vectors
- maintain an ordered readiness frontier so `drive_ready` visits only logical workers that can execute, while preserving lowest-worker-ID ordering and coordinator-turn retries for effect-free passes
- replace the temporary per-DP-pass `BTreeMap` with an inline `SmallVec`
- make zero-duration re-arming self-contained, assert the frontier against a debug-only full scan, and centralize group tombstoning with an O(1) live-group count
- pin rank/group storage invariants and cover ordering, startup activation, and both removal paths directly
- keep KVBM refresh behavior and the DP>1 `SmallVec<[...; 1]>` behavior unchanged pending dedicated profiles

Builds on #12240, which is now merged. This branch has been rebased onto its merge commit and contains only the readiness-frontier change.

Start in `lib/mocker/src/replay/offline/components/engine.rs` at `group_is_ready`, `refresh_group`, and `drive_ready` for the readiness-frontier invariant and zero-duration re-arming. Then review `tombstone_group` and `live_group_count` for lifecycle bookkeeping, followed by the focused tests at the bottom of the same file.

## Performance

AgentX + native G1 + KV routing, 128,000 blocks per worker, release build with DWARF and normal compiler inlining, pinned to CPU 0.

| Workload | #12240 control | Candidate | Change |
| --- | ---: | ---: | ---: |
| 16 workers / 16 sessions replay time | 8.9098 s | 8.1361 s | 1.095x speedup |
| 16 workers / 16 sessions requests/s | 360.4 | 394.6 | +9.51% |
| 16 workers / 16 sessions tokens/s | 54.26M | 59.41M | +9.51% |
| 16 workers / 16 sessions max RSS | 242,528 KiB | 247,356 KiB | +1.99% |

The completed A/B used one run per arm, candidate first. Both arms completed the same 3,211 requests, 479,843,008 input tokens, 3,552,472 output tokens, and total simulated duration.

A matched 60-second `perf record -e cpu_core/cycles/u -F 99 --call-graph dwarf,16384` comparison at 128 workers / 128 AgentX sessions shows the targeted cost collapsing. The candidate profile below was repeated after the review fixes:

| `EngineComponent::drive_ready` | Existing #12240 profile | Reviewed candidate profile | Change |
| --- | ---: | ---: | ---: |
| Self CPU | 49.65% | 1.06% | -48.59 pp |
| Sampled cycles | 160.86B | 3.46B | -97.85% / 46.5x less |

The two captures observed effectively identical total cycles over 60 seconds (323.98B vs 324.91B), and both lost zero samples. The large workload was stopped after each capture; the profile comparison is targeted CPU attribution, not a completed 128-worker throughput claim.

## Validation

- `cargo test -p dynamo-mocker replay::offline::components::engine::tests` (15 passed)
- `cargo test -p dynamo-mocker same_timestamp_destination_activation_precedes_next_decode_drive` (1 passed)
- `cargo test -p dynamo-mocker replay::offline` (174 passed)
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline` (177 passed)
- `git diff --check`
- exact KV replay parity at `08a1bb313b` versus #12240: PASS for vLLM aggregated/disaggregated KVBM and SGLang aggregated/disaggregated; not rerun for the mechanical `f8df6564ba` cleanup
- completed 16-worker / 16-session AgentX native-G1 A/B
- repeated 60-second 128-worker / 128-session candidate perf capture after the review fixes: 5,941 samples, zero lost samples, `drive_ready` at 1.06% self CPU

## Related Issues

- [x] No related issue

## Summary by CodeRabbit

* **Improvements**
  * Improved offline replay worker scheduling and readiness handling.
  * Enhanced consistency when workers are drained, removed, or scaled down.
  * Preserved deterministic scheduling order for ready worker groups.
  * Updated readiness and offload behavior for more reliable replay execution.

### DIFF
--- a/lib/mocker/src/replay/offline/components/engine.rs
+++ b/lib/mocker/src/replay/offline/components/engine.rs
@@ -49,10 +49,6 @@ where
     ready_groups: BTreeSet<usize>,
     /// Ready groups that made no observable progress during the prior drive.
     deferred_ready_groups: BTreeSet<usize>,
-    /// Counter for generating the next stable scheduler ID.
-    next_id: usize,
-    /// Counter for generating the next stable mocker worker ID.
-    next_worker_id: usize,
     /// Mocker workers marked for removal — skipped by round-robin, removed when drained.
     pending_removal: BTreeSet<usize>,
     /// Mocker workers still starting up — excluded from active set until ready.
@@ -91,8 +87,6 @@ where
             live_group_count: count,
             ready_groups: BTreeSet::new(),
             deferred_ready_groups: BTreeSet::new(),
-            next_id: count,
-            next_worker_id: count,
             pending_removal: BTreeSet::new(),
             pending_startup: BTreeSet::new(),
             args: MockEngineArgs::default(),
@@ -139,8 +133,6 @@ where
             live_group_count: num_workers,
             ready_groups: BTreeSet::new(),
             deferred_ready_groups: BTreeSet::new(),
-            next_id: num_workers.saturating_mul(dp_size),
-            next_worker_id: num_workers,
             pending_removal: BTreeSet::new(),
             pending_startup: BTreeSet::new(),
             args,
@@ -243,12 +235,10 @@ where
     /// Add a new mocker worker and all of its DP-rank schedulers, returning
     /// the stable mocker worker ID.
     pub(in crate::replay::offline) fn add_worker(&mut self) -> usize {
-        let worker_id = self.next_worker_id;
-        self.next_worker_id += 1;
+        let worker_id = self.worker_groups.len();
         let mut rank_ids = Vec::with_capacity(self.args.dp_size.max(1) as usize);
         for dp_rank in 0..self.args.dp_size.max(1) {
-            let rank_id = self.next_id;
-            self.next_id += 1;
+            let rank_id = self.workers.len();
             let worker = OfflineWorkerState::new_with_rank(
                 rank_id,
                 worker_id as u64,
@@ -746,7 +736,7 @@ where
 
     #[cfg(test)]
     pub(in crate::replay::offline) fn rank_id_capacity(&self) -> usize {
-        self.next_id
+        self.workers.len()
     }
 
     #[cfg(feature = "kvbm-offload")]
@@ -1133,6 +1123,15 @@ mod tests {
         assert!(newly_marked.is_empty());
         assert_eq!(engine.active_worker_ids().len(), 2);
         assert_eq!(engine.worker_count(), 2);
+
+        // Tombstoned stable IDs are never reused. The next worker and rank
+        // come from the append-only vector tails.
+        let (added, newly_marked, removed) = engine.apply_target_count(3);
+        assert_eq!(added, vec![4]);
+        assert!(newly_marked.is_empty());
+        assert!(removed.is_empty());
+        assert_eq!(engine.worker_groups[4].as_deref(), Some(&[4][..]));
+        assert_eq!(engine.rank_id_capacity(), 5);
     }
 
     #[test]
--- a/lib/mocker/src/replay/offline/components/engine.rs
+++ b/lib/mocker/src/replay/offline/components/engine.rs
@@ -1,10 +1,11 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeSet;
 use std::marker::PhantomData;
 
 use anyhow::bail;
+use smallvec::SmallVec;
 
 use super::super::core::{EngineEventBatch, EngineProgress, NoEngineEvents, WorkerTopology};
 use super::super::events::SimulationWorkerStage;
@@ -20,8 +21,7 @@ use super::{
 };
 use crate::common::protocols::{DirectRequest, ForwardPassSnapshot, MockEngineArgs};
 use crate::replay::TraceCollector;
-use crate::scheduler::RouterEventVisibility;
-use crate::scheduler::SchedulerCommand;
+use crate::scheduler::{EnginePassResult, RouterEventVisibility, SchedulerCommand};
 
 fn fpm_has_scheduled_work(snapshot: &ForwardPassSnapshot) -> bool {
     snapshot.num_prefill_requests > 0 || snapshot.num_decode_requests > 0
@@ -34,9 +34,13 @@ where
     stage: SimulationWorkerStage,
     pass_mode: EnginePassMode,
     /// DP-rank schedulers keyed by stable ID (monotonic, never reused).
-    workers: BTreeMap<usize, OfflineWorkerState>,
+    workers: Vec<Option<OfflineWorkerState>>,
     /// Mocker worker IDs mapped to their per-rank scheduler IDs.
-    worker_groups: BTreeMap<usize, Vec<usize>>,
+    worker_groups: Vec<Option<Vec<usize>>>,
+    /// Logical workers that can start a pass, ordered by stable worker ID.
+    ready_groups: BTreeSet<usize>,
+    /// Ready groups that made no observable progress during the prior drive.
+    deferred_ready_groups: BTreeSet<usize>,
     /// Counter for generating the next stable scheduler ID.
     next_id: usize,
     /// Counter for generating the next stable mocker worker ID.
@@ -62,13 +66,15 @@ where
         workers: Vec<OfflineWorkerState>,
     ) -> Self {
         let count = workers.len();
-        let map: BTreeMap<usize, OfflineWorkerState> = workers.into_iter().enumerate().collect();
-        let worker_groups = (0..count).map(|id| (id, vec![id])).collect();
-        Self {
+        let workers = workers.into_iter().map(Some).collect();
+        let worker_groups = (0..count).map(|id| Some(vec![id])).collect();
+        let mut component = Self {
             stage,
             pass_mode,
-            workers: map,
+            workers,
             worker_groups,
+            ready_groups: BTreeSet::new(),
+            deferred_ready_groups: BTreeSet::new(),
             next_id: count,
             next_worker_id: count,
             pending_removal: BTreeSet::new(),
@@ -76,7 +82,9 @@ where
             args: MockEngineArgs::default(),
             capture_raw: Observation::CAPTURE_RAW,
             observation: PhantomData,
-        }
+        };
+        component.refresh_all_groups();
+        component
     }
 
     /// Build one scheduler core per DP rank while retaining the live mocker's
@@ -88,31 +96,32 @@ where
         num_workers: usize,
     ) -> Self {
         let dp_size = args.dp_size.max(1) as usize;
-        let mut workers = BTreeMap::new();
-        let mut worker_groups = BTreeMap::new();
+        let mut workers = Vec::with_capacity(num_workers.saturating_mul(dp_size));
+        let mut worker_groups = Vec::with_capacity(num_workers);
         for worker_id in 0..num_workers {
             let mut rank_ids = Vec::with_capacity(dp_size);
             for dp_rank in 0..dp_size {
                 let rank_id = worker_id * dp_size + dp_rank;
-                workers.insert(
+                debug_assert_eq!(rank_id, workers.len());
+                workers.push(Some(OfflineWorkerState::new_with_rank(
                     rank_id,
-                    OfflineWorkerState::new_with_rank(
-                        rank_id,
-                        worker_id as u64,
-                        dp_rank as u32,
-                        args.clone(),
-                        Observation::CAPTURE_RAW,
-                    ),
-                );
+                    worker_id as u64,
+                    dp_rank as u32,
+                    args.clone(),
+                    Observation::CAPTURE_RAW,
+                )));
                 rank_ids.push(rank_id);
             }
-            worker_groups.insert(worker_id, rank_ids);
+            debug_assert_eq!(worker_id, worker_groups.len());
+            worker_groups.push(Some(rank_ids));
         }
         Self {
             stage,
             pass_mode,
             workers,
             worker_groups,
+            ready_groups: BTreeSet::new(),
+            deferred_ready_groups: BTreeSet::new(),
             next_id: num_workers.saturating_mul(dp_size),
             next_worker_id: num_workers,
             pending_removal: BTreeSet::new(),
@@ -133,6 +142,52 @@ where
         self.capture_raw = capture_raw;
     }
 
+    fn worker(&self, rank_id: usize) -> Option<&OfflineWorkerState> {
+        self.workers.get(rank_id)?.as_ref()
+    }
+
+    fn worker_mut(&mut self, rank_id: usize) -> Option<&mut OfflineWorkerState> {
+        self.workers.get_mut(rank_id)?.as_mut()
+    }
+
+    fn group_is_ready(&self, worker_id: usize) -> bool {
+        let Some(rank_ids) = self.worker_groups.get(worker_id).and_then(Option::as_ref) else {
+            return false;
+        };
+        !rank_ids.iter().any(|&rank_id| {
+            self.worker(rank_id)
+                .is_some_and(OfflineWorkerState::is_busy)
+        }) && rank_ids.iter().any(|&rank_id| {
+            self.worker(rank_id)
+                .is_some_and(OfflineWorkerState::is_ready)
+        })
+    }
+
+    fn refresh_group(&mut self, worker_id: usize) {
+        self.deferred_ready_groups.remove(&worker_id);
+        if self.group_is_ready(worker_id) {
+            self.ready_groups.insert(worker_id);
+        } else {
+            self.ready_groups.remove(&worker_id);
+        }
+    }
+
+    fn refresh_rank(&mut self, rank_id: usize) {
+        let Some(worker_id) = self
+            .worker(rank_id)
+            .and_then(|worker| usize::try_from(worker.rank_identity().0).ok())
+        else {
+            return;
+        };
+        self.refresh_group(worker_id);
+    }
+
+    fn refresh_all_groups(&mut self) {
+        for worker_id in 0..self.worker_groups.len() {
+            self.refresh_group(worker_id);
+        }
+    }
+
     /// Add a new mocker worker and all of its DP-rank schedulers, returning
     /// the stable mocker worker ID.
     pub(in crate::replay::offline) fn add_worker(&mut self) -> usize {
@@ -149,10 +204,12 @@ where
                 self.args.clone(),
                 self.capture_raw,
             );
-            self.workers.insert(rank_id, worker);
+            debug_assert_eq!(rank_id, self.workers.len());
+            self.workers.push(Some(worker));
             rank_ids.push(rank_id);
         }
-        self.worker_groups.insert(worker_id, rank_ids);
+        debug_assert_eq!(worker_id, self.worker_groups.len());
+        self.worker_groups.push(Some(rank_ids));
         worker_id
     }
 
@@ -167,11 +224,14 @@ where
     /// Remove all marked workers that have fully drained, returning their IDs.
     pub(in crate::replay::offline) fn try_remove_drained(&mut self) -> Vec<usize> {
         let mut removed = Vec::new();
+        let worker_groups = &self.worker_groups;
+        let workers = &self.workers;
         self.pending_removal.retain(|&id| {
-            if let Some(rank_ids) = self.worker_groups.get(&id) {
+            if let Some(rank_ids) = worker_groups.get(id).and_then(Option::as_ref) {
                 if rank_ids.iter().all(|rank_id| {
-                    self.workers
-                        .get(rank_id)
+                    workers
+                        .get(*rank_id)
+                        .and_then(Option::as_ref)
                         .is_none_or(OfflineWorkerState::is_drained)
                 }) {
                     removed.push(id);
@@ -184,9 +244,13 @@ where
             true // keep in pending set
         });
         for &id in &removed {
-            if let Some(rank_ids) = self.worker_groups.remove(&id) {
+            self.ready_groups.remove(&id);
+            self.deferred_ready_groups.remove(&id);
+            if let Some(rank_ids) = self.worker_groups.get_mut(id).and_then(Option::take) {
                 for rank_id in rank_ids {
-                    self.workers.remove(&rank_id);
+                    if let Some(worker) = self.workers.get_mut(rank_id) {
+                        worker.take();
+                    }
                 }
             }
         }
@@ -234,9 +298,13 @@ where
                 .collect();
             for &id in &to_cancel {
                 self.pending_startup.remove(&id);
-                if let Some(rank_ids) = self.worker_groups.remove(&id) {
+                self.ready_groups.remove(&id);
+                self.deferred_ready_groups.remove(&id);
+                if let Some(rank_ids) = self.worker_groups.get_mut(id).and_then(Option::take) {
                     for rank_id in rank_ids {
-                        self.workers.remove(&rank_id);
+                        if let Some(worker) = self.workers.get_mut(rank_id) {
+                            worker.take();
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -522,6 +620,9 @@ where
             if !effects.is_empty() {
                 return Ok(effects);
             }
+            if self.group_is_ready(worker_id) {
+                self.deferred_ready_groups.insert(worker_id);
+            }
```
jthomson04, line 683: **Correctness / robustness — the one I'd most like fixed.**

On this path the group has been popped from `ready_groups` and is added to *neither* set before returning. If the pass was zero-duration and the group still has queued work, it is now absent from the frontier entirely.

It only comes back because `agg.rs:705` / `disagg.rs:1958` feed every payload through `on_scheduled_completion`, which calls `refresh_rank`. I verified that holds today — non-empty effects always imply a payload for this group — so this is **not** a live bug. But it is a cross-module invariant with no comment and no assertion. A caller that ever consumes `immediate_completions` without round-tripping them, or a future `EngineEffects` variant not paired with a payload, silently drops the group forever. That manifests as a hung replay, not a panic.

Cheap fix that makes the component self-contained and idempotent with the caller's later refresh:

```suggestion
            if !effects.is_empty() {
                // Re-arm here rather than relying on the caller round-tripping
                // every payload through `on_scheduled_completion`.
                self.refresh_group(worker_id);
                return Ok(effects);
            }
            if self.group_is_ready(worker_id) {
                self.deferred_ready_groups.insert(worker_id);
            }
```

Cost is O(dp_size) on a path that already did O(dp_size) work. No livelock risk: re-inserting does not change whether effects are produced, so the `drive_ready_workers` loop termination is still governed entirely by `made_progress`.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -365,62 +451,74 @@ where
...
-        // the loop body only mutates `self.workers`, never the group map itself.
-        for rank_ids in self.worker_groups.values() {
+        // The serial coordinator may make another component observable at the
+        // same virtual timestamp. Match the former full scan by retrying
+        // effect-free groups on the next coordinator drive, but not again
+        // within this call.
+        let deferred = std::mem::take(&mut self.deferred_ready_groups);
+        for worker_id in deferred {
+            self.refresh_group(worker_id);
+        }
```
jthomson04, line 514: **Highest-leverage suggestion in this review.**

Right after this loop, `ready_groups` should be *exactly* the set of ready groups — that is precisely the invariant the whole PR rests on, and a violation of it is silent. Asserting it here makes all 173 existing `replay::offline` tests validate the frontier against the pre-PR full-scan semantics for free:

```rust
debug_assert_eq!(
    self.ready_groups,
    (0..self.worker_groups.len())
        .filter(|&id| self.group_is_ready(id))
        .collect::<BTreeSet<_>>(),
    "readiness frontier drifted from a full scan",
);
```

Given that every finding below is some flavour of "a missed refresh would be invisible", this converts the central invariant from *holds by inspection of two other files* into something the test suite enforces on every run.

jthomson04 (commented):
## Review pass (Claude Code)

**Overall: approve with changes.** The design is sound and the perf case is convincing. I traced every mutation path into `OfflineWorkerState` reachable from `agg.rs`/`disagg.rs` — `dispatch`, `apply_command`, `on_scheduled_completion`, `tick_offload_engines`, and the two removal paths — and each has a matching refresh. I could not construct a reproducible lost wakeup. My concerns are about the **fragility of the new invariant**, not a bug in the current code.

Things I checked and believe are correct:

- **Ordering is preserved.** `pop_first` on a `BTreeSet` is equivalent to the old ascending scan from 0, and groups re-enter the frontier before the next call, so a coordinator turn still restarts from the lowest ready worker.
- **`deferred_ready_groups` faithfully replicates the old same-timestamp retry.** Because `drive_ready_workers` loops until effects are empty, and the deferred set is drained at the top of *every* `drive_ready` call, an effect-free group is still re-scanned on each iteration within the same coordinator turn — matching the old full scan. No new livelock (deferral returns empty effects, so `changed` stays false) and no new dead-end.
- **Tombstoned `Vec` indexing is safe.** Both ID counters are already monotonic and never reused, and removal never shortens the vectors, so `next_id == workers.len()` holds at every `push`. The `debug_assert_eq!` guards pin it.
- **`effects.is_empty() == false` always implies a payload exists for that group.** `admissions`, `pass_start_events`, and `progress.made_progress` each imply `made_progress`, which implies the rank's payload is pushed to `immediate_completions` or `scheduled_completions`. This is what makes the caller round-trip in finding 1 actually work today.
- **The `zip` rewrite is sound** — the first loop pushes exactly one entry per rank in `rank_ids` order, including `None` for not-ready ranks, so lengths always match.

Two items I'd want before merge (~4 lines total, both inline below): the caller dependency in `drive_ready`'s effects-return path, and a `debug_assert` pinning `ready_groups` against a full scan so the existing 173 tests enforce the invariant. The rest are follow-ups.

### Test coverage

`attention_dp_empty_rank_blocks_mid_epoch_arrival` happens to exercise the DP round trip well (dispatch-to-busy-group leaves the frontier empty, completions re-arm it, next epoch fires), and the new assertions in `queued_only_zero_duration_pass_does_not_report_progress` pin the deferral semantics nicely. Gaps:

- **No scale-down x frontier test.** A group sitting in `ready_groups` when `try_remove_drained` or the pending-startup cancel path tombstones it is exactly where a stale entry would hit the `.unwrap()` in `drive_ready`. Both cleanup sites are new code with no direct coverage.
- **`ready_groups_preserve_lowest_worker_id_order` stops one assertion short.** A second `drive_ready` asserting worker 1 runs would pin "the un-popped remainder stays in the frontier", which is the other half of what this change does.
- **No `mark_worker_ready` frontier test** — a pending-startup worker that received a dispatch before its ready event.
- Nothing covers the caller dependency (finding 1) directly; it is only reachable through the `agg`/`disagg` callers.

### Security

Nothing to flag — simulation-only code, no I/O, no untrusted input. The new direct indexing panics rather than corrupting state, which is the right failure mode here.

### On the benchmark

The single-run-per-arm throughput A/B is noise-prone, but the PR body says so, and the 49.65% -> 1.45% self-CPU attribution over matched 60s captures with zero lost samples is a much stronger signal on its own. No objection to the methodology as presented.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -133,6 +142,52 @@ where
...
+        let Some(rank_ids) = self.worker_groups.get(worker_id).and_then(Option::as_ref) else {
+            return false;
+        };
+        !rank_ids.iter().any(|&rank_id| {
+            self.worker(rank_id)
+                .is_some_and(OfflineWorkerState::is_busy)
+        }) && rank_ids.iter().any(|&rank_id| {
+            self.worker(rank_id)
+                .is_some_and(OfflineWorkerState::is_ready)
+        })
```
jthomson04: **Inconsistent tolerance for missing ranks.**

`group_is_ready` uses `is_some_and`, treating a `None` rank slot as neither busy nor ready. But `drive_ready` uses `self.workers[rank_id].as_ref().unwrap()` (lines 472 / 478 / 486). So if the "a group's ranks are tombstoned together with the group" invariant ever breaks, this returns `true` and `drive_ready` **panics** instead of skipping.

Unreachable today — both removal sites take the group and all its ranks atomically — but the two functions should agree. Either make `drive_ready` tolerant, or make this one assert.

Minor readability point while you're here: the `!any(busy) && any(ready)` chain across two closures joined by `&&` is dense. Two `let` bindings would read better, and it's worth noting in a comment that `is_ready()` already implies `!busy`, so the first clause is specifically the *sibling* barrier, not redundant.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -62,21 +66,25 @@ where
...
             next_id: count,
             next_worker_id: count,
             pending_removal: BTreeSet::new(),
             pending_startup: BTreeSet::new(),
             args: MockEngineArgs::default(),
             capture_raw: Observation::CAPTURE_RAW,
             observation: PhantomData,
-        }
+        };
+        component.refresh_all_groups();
```
jthomson04, line 96: Two small things here.

**1. New undocumented precondition on `new`.** `refresh_rank` maps rank -> group via `worker.rank_identity().0`, while `new` builds `worker_groups[i] = vec![i]`. Nothing enforces `workers[i].rank_identity().0 == i`. It holds because `disagg.rs:996`/`1009` construct via `OfflineWorkerState::new(worker_idx, ..)`, which sets `worker_id = worker_idx` (`state.rs:269-271`) — but that's incidental, and a mismatch would silently refresh the *wrong* group. Worth pinning:

```rust
debug_assert!(
    workers.iter().enumerate().all(|(i, w)| w.rank_identity().0 == i as u64),
    "EngineComponent::new assumes worker_id == index",
);
```

**2. Asymmetry with `new_ranked`,** which doesn't call `refresh_all_groups`. Both constructors start with empty cores, so this is a no-op either way — but pick one so the next reader doesn't have to work out that it doesn't matter.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -34,9 +34,13 @@ where
...
     /// DP-rank schedulers keyed by stable ID (monotonic, never reused).
-    workers: BTreeMap<usize, OfflineWorkerState>,
+    workers: Vec<Option<OfflineWorkerState>>,
     /// Mocker worker IDs mapped to their per-rank scheduler IDs.
-    worker_groups: BTreeMap<usize, Vec<usize>>,
+    worker_groups: Vec<Option<Vec<usize>>>,
+    /// Logical workers that can start a pass, ordered by stable worker ID.
+    ready_groups: BTreeSet<usize>,
+    /// Ready groups that made no observable progress during the prior drive.
+    deferred_ready_groups: BTreeSet<usize>,
```
jthomson04, line 51: **Scalability note worth a comment on these fields.**

IDs are monotonic and slots are never reclaimed, so `workers.len()` equals *total workers ever created*, not live workers. Each dead slot still occupies `size_of::<OfflineWorkerState>()` — `EngineCore` inlines `MockEngineArgs` + `SchedulerState` + `KvManager` by value, so that is not small.

Consequence: every remaining full scan is now O(ever-created) rather than O(live):

- `in_flight()` — reached per coordinator iteration via `cluster_in_flight()`
- `is_drained()` — `agg.rs:448`/`457`
- `worker_count()`, `active_group_ids()`, `refresh_all_groups()`

The 16-worker benchmark can't see this (no churn). The `global_planner` autoscaling workloads are exactly the case that would: a planner cycling between 4 and 16 workers over a long replay accumulates a tombstone per scale-up forever.

Not a blocker, and the +1.99% RSS on a no-churn run says nothing either way. But it should at least be documented on these fields, since it's a real behavioural difference from the `BTreeMap` this replaces. If it turns out to matter, `Vec<Option<Box<OfflineWorkerState>>>` bounds the per-slot cost, or generation-tagged slot reuse bounds the count.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -610,13 +715,15 @@ where
                 .lifecycle_events
                 .extend(worker_effects.lifecycle_events);
         }
+        self.refresh_all_groups();
```
jthomson04, line 776: **This reintroduces the O(G) scan the PR removes, for `kvbm-offload` builds.**

`refresh_all_groups()` re-evaluates every group (including tombstones) on every offload tick. The refresh is genuinely needed — `tick_offload_only` can promote swap-ins and flip `core.is_empty()` — but the loop directly above already visits each worker, so it can be narrowed:

```rust
for (rank_id, worker) in self.workers.iter_mut().enumerate().filter(...) {
    ...
}
// then refresh only the ranks whose effects were non-empty
```

Worth noting the perf A/B was a default build, so this path wasn't measured. `cargo test --features kvbm-offload` passing tells us it's correct, not that it's fast.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -554,22 +653,27 @@ where
...
-        self.workers.values().all(OfflineWorkerState::is_drained)
+        self.workers
+            .iter()
+            .filter_map(Option::as_ref)
+            .all(OfflineWorkerState::is_drained)
     }
 
     pub(in crate::replay::offline) fn worker_count(&self) -> usize {
-        self.worker_groups.len()
+        self.worker_groups.iter().flatten().count()
```
jthomson04: `worker_count()` goes O(1) -> O(n), and it's called per accounting tick to compute worker-seconds (`agg.rs:877`, `disagg.rs:2328-2329`).

The semantics are right (counting `Some` == live groups, matching the old `BTreeMap::len`). Trivial cost today, but combined with the unbounded tombstone growth noted on the field declarations it drifts. A maintained counter incremented in `add_worker` / decremented in the two removal sites would be free and restore O(1).

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -365,62 +451,74 @@ where
...
                 .iter()
-                .any(|rank_id| self.workers.get(rank_id).unwrap().is_ready())
+                .any(|&rank_id| self.workers[rank_id].as_ref().unwrap().is_ready())
             {
                 continue;
             }
 
-            let mut executed_by_rank = BTreeMap::new();
+            let mut executed_by_rank: SmallVec<[Option<EnginePassResult>; 1]> =
+                SmallVec::with_capacity(rank_ids.len());
```
jthomson04, line 541: Inline capacity 1 means every `dp_size > 1` group heap-allocates here, once per group per drive — so for attention-DP deployments this is the same allocation the `BTreeMap` was doing, just cheaper.

Two options worth measuring rather than guessing (`Option<EnginePassResult>` is a wide inline element, so bumping the inline size isn't free either):

- Raise the inline size to cover common `dp_size` values (2/4/8).
- Hoist `executed_by_rank` out of the `while` loop and `clear()` it per iteration, so the allocation amortizes across all groups in a drive regardless of inline size.

The second is probably the better trade — it keeps the inline element count at 1 for the dp=1 fast path while eliminating the repeated allocation for dp>1.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -315,19 +394,26 @@ where
...
     ) -> anyhow::Result<()> {
-        let worker = self
-            .workers
-            .get_mut(&worker_id)
-            .ok_or_else(|| anyhow::anyhow!("offline replay selected unknown worker {worker_id}"))?;
-        worker.receive_request(request);
+        self.worker_mut(worker_id)
+            .ok_or_else(|| anyhow::anyhow!("offline replay selected unknown worker {worker_id}"))?
+            .receive_request(request);
+        self.refresh_rank(worker_id);
```
jthomson04: Naming nit, but a real one for readability: `dispatch`, `apply_command`, and `worker_is_busy` all name their parameter `worker_id` when it is in fact a **rank ID**. This PR correctly renames it to `rank_id` in `on_scheduled_completion` — worth finishing the job in the functions you're already touching.

As written, `self.refresh_rank(worker_id)` reads like a bug until you go check that `dispatch`'s `worker_id` indexes `self.workers` and not `self.worker_groups`.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. `drive_ready` now re-arms a productive pass itself only when `group_end_ms <= now_ms`; scheduled passes remain completion-driven while their ranks are busy. `productive_zero_duration_pass_reaches_terminal_completion` now asserts the frontier is seeded before caller completion processing. Validation: 15 focused engine tests, 173 native replay tests, 176 KVBM replay tests, clippy, fmt, and diff checks all pass. The repeated 128x128 capture has zero lost samples and `drive_ready` at 1.06% self CPU.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. A debug-only `debug_assert_ready_frontier` recomputes the full-scan set and checks exact equality immediately after deferred groups are refreshed, so the native and KVBM replay suites exercise the invariant. Validation: 173 native and 176 KVBM replay tests pass, along with clippy and fmt.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. `group_is_ready` now uses named `any_busy` / `any_ready` checks, documents the sibling barrier, and shares checked `required_worker` access with `drive_ready`. A broken live-group/rank invariant now fails with a clear message instead of being tolerated in one path and unqualified-unwrapped in another. Native/KVBM replay suites and clippy pass.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. `EngineComponent::new` now debug-asserts that each worker's logical ID matches its vector index, and `new_ranked` now initializes symmetrically through `refresh_all_groups`. The focused engine tests plus full native/KVBM replay suites pass.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Addressed in `08a1bb313b`. The fields now explicitly document that stable-ID tombstones are unreclaimed and make storage/scans scale with total-ever-created ranks. `worker_count` is restored to O(1), and both removal paths share `tombstone_group`. Slot reuse/boxing is intentionally deferred until an autoscaling-churn profile justifies that larger design change. Native/KVBM replay suites and clippy pass.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Keeping this unchanged in `08a1bb313b`. `tick_offload_only` can change readiness through swap-in progress without necessarily producing an outward effect, so narrowing refreshes based only on non-empty effects would risk a lost wakeup. This PR's CPU evidence is native-G1 only; KVBM refresh narrowing needs a dedicated KVBM profile plus an explicit readiness-transition signal rather than a speculative change. The existing behavior is covered by 176 passing KVBM replay tests.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. `live_group_count` is initialized by both constructors, incremented by `add_worker`, and decremented centrally by `tombstone_group`, so `worker_count()` is O(1) again. Both drained removal and startup cancellation now have direct cleanup/count coverage; all native/KVBM replay tests and clippy pass.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Keeping `SmallVec<[...; 1]>` unchanged in `08a1bb313b`. The measured AgentX/native-G1 path is DP=1, and `drive_ready` returns at the first productive group, limiting within-call amortization from hoisting the scratch vector. Choosing persistent scratch or a larger inline capacity should follow an attention-DP allocation/CPU profile rather than extending this PR without evidence. Attention-DP correctness remains covered by the native and KVBM replay suites.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `08a1bb313b`. The rank-indexed parameters in `dispatch`, `apply_command`, and `worker_is_busy` are now consistently named `rank_id`, including their error messages and refresh calls. Clippy and both replay suites pass.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -1011,6 +1231,8 @@ mod tests {
 
         let first = engine.drive_ready(0.0, Some(&mut collector)).unwrap();
         assert!(first.is_empty());
+        assert!(engine.ready_groups.is_empty());
+        assert_eq!(engine.deferred_ready_groups, BTreeSet::from([0]));
```
PeaBrane, line 1234: Non-blocking: this test confirms an effect-free group remains deferred, but it never exercises the reason for deferral—a same-timestamp handoff/KVBM event making that group runnable. A coordinator-level case that inserts an `ActivateDestination` or offload completion between drives and asserts eventual terminal completion would guard against early quiescence or a lost wakeup. I’d also run the exact base/head KV replay parity campaign, since matching aggregate totals does not prove scheduling-order parity; this can be follow-up work.

PeaBrane (approved):
✅ Looks good overall. The ordering and frontier refresh paths check out. One non-blocking validation follow-up inline.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -33,10 +33,22 @@ where
...
+    workers: Vec<Option<OfflineWorkerState>>,
     /// Mocker worker IDs mapped to their per-rank scheduler IDs.
-    worker_groups: BTreeMap<usize, Vec<usize>>,
+    ///
+    /// Removed groups also leave tombstones to preserve stable IDs.
+    worker_groups: Vec<Option<Vec<usize>>>,
+    /// Number of non-tombstoned worker groups.
+    live_group_count: usize,
+    /// Logical workers that can start a pass, ordered by stable worker ID.
+    ready_groups: BTreeSet<usize>,
```
PeaBrane, line 49: Non-blocking follow-up: `EngineComponent` now owns a fairly large cluster of fields that form one lifecycle invariant—rank/group storage, live count, ready/deferred membership, IDs, startup/removal state, and worker-construction config. My preference would be to keep this PR focused and extract that cluster afterward into a private same-file `OfflineWorkerPool`, moving the existing helpers mechanically and using this head as the replay-parity baseline. That said, I’m also fine with folding it into this PR if you’re confident the move remains mechanical and replay parity stays clean. Either way, it reduces the flat state; I would avoid extracting `ReadyFrontier` alone because readiness depends directly on worker storage.

PeaBrane (commented):
Non-blocking follow-up suggestion; my approval stands.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -33,10 +33,22 @@ where
...
+    /// Ready groups that made no observable progress during the prior drive.
+    deferred_ready_groups: BTreeSet<usize>,
     /// Counter for generating the next stable scheduler ID.
     next_id: usize,
```
PeaBrane: Could we remove both `next_id` and `next_worker_id` in this PR? They were useful with the live-only `BTreeMap`s, but the append-only tombstoned vectors now make `workers.len()` and `worker_groups.len()` the next stable rank and worker IDs. The push-site `debug_assert_eq!` checks already encode that invariant. Removing the counters would eliminate two synchronized fields; `rank_id_capacity()` can return `workers.len()`.

PeaBrane (commented):
Small cleanup request on the new storage representation; my approval stands.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Addressed in `f8df6564ba`. Tracing the native-VLLM coordinator showed that zero-delay `ReserveDestination` and `ActivateDestination` are drained back-to-back by `drive_pending_actions` before the next decode drive, so production does not create a deferred-frontier interval between those commands. I added `same_timestamp_destination_activation_precedes_next_decode_drive` to pin the actual invariant: no quiescent decode drive between reservation and activation, activation/admission before the next quiescent drive, identical reservation/activation timestamps, and terminal completion. The exact four-row KV replay parity campaign at `08a1bb313b` versus #12240 already passed; per the agreed scope it was not rerun for this mechanical counter cleanup. Final validation: 174 native replay tests and 177 KVBM replay tests pass, plus clippy, fmt, and diff checks.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Agreed. I kept this PR focused and did not extract `OfflineWorkerPool`. That mechanical same-file extraction is better handled as a follow-up based on this reviewed head, and I agree that extracting `ReadyFrontier` alone would split an invariant that depends directly on worker storage.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `f8df6564ba`. `next_id` and `next_worker_id` are removed; `add_worker` derives stable IDs from `workers.len()` / `worker_groups.len()`, and `rank_id_capacity()` now returns `workers.len()`. The startup-tombstone test now scales back up and verifies that worker/rank ID 4 is appended rather than reusing tombstones 2 or 3. All 174 native and 177 KVBM replay tests pass, along with clippy, fmt, and diff checks.

PR #12244, opened 2026-07-28
perf(mocker): use intrusive inactive block LRU

jthomson04:
## Summary

- replace the native `VllmBlockPool` inactive-copy `BTreeMap` with an intrusive
  doubly linked LRU stored directly in each cached copy
- make release-to-tail, pin/unlink, and head eviction constant-time without
  allocating or rebalancing tree nodes
- preserve the existing ordinary-LRU ordering and add coverage for unlinking a
  pinned copy from the middle of the inactive list

This builds on #12242, which is now merged. The branch is rebased onto current
`main` and contains only the block-pool commit.

## Performance

Native-G1 AgentX offline replay, KV routing, CPU 0, stock allocator.

The 128-worker / 128-session candidate `perf` capture used the same workload,
CPU, event, and 60-second window as the prior #12242 profile. Total sampled
cycles differed by 0.055%, and both captures lost zero samples.

- selected direct inactive-LRU/tree self-time: 14.66% -> 1.52% (-89.7%)
- `VllmBlockPool::insert_inactive`: 5.02% -> 0.30% (-93.9% sampled cycles)
- primary `BTreeMap::remove`: 4.95% -> 0.03% (-99.3% sampled cycles)
- replacement `VllmBlockPool::unlink_inactive`: 1.06%

Completed 16-worker / 16-session A/B, one arm at a time:

| Metric | #12242 control | Intrusive LRU | Change |
| --- | ---: | ---: | ---: |
| Replay time | 8.0715 s | 5.8251 s | -27.83% / 1.386x |
| Host request rate | 397.82 req/s | 551.24 req/s | +38.56% |
| Host token rate | 59.889M tok/s | 82.985M tok/s | +38.56% |
| Peak RSS | 248,680 KiB | 233,752 KiB | -6.00% |

Both arms completed exactly 3,211 requests and 483,395,480 processed tokens.
The simulated-duration difference was 0.112 ppm.

Post-review candidate sanity on #12242's reviewed head (`08a1bb31`) before it
merged:

- 5.5281 s replay time, 580.85 host req/s, and 87.443M host tok/s
- 236,152 KiB peak RSS
- exact request and token totals; simulated-duration drift of 0.449 ppm
- 5.1% faster than the original candidate run

This is a regression sanity check rather than a fresh matched A/B because it
also includes #12242's review fixes. The completed table above remains the
matched end-to-end comparison.

## Validation

- `cargo test -p dynamo-mocker cache::vllm_block_pool::tests` (7 passed)
- `cargo test -p dynamo-mocker
  native_g1_runs_through_offline_replay_entrypoint` (2 passed: vLLM and
  TensorRT-LLM)
- `cargo test --release -p dynamo-mocker
  cache::vllm_block_pool::tests::eviction_rejects_head_with_predecessor` (1
  passed)
- `cargo test -p dynamo-mocker replay::offline` (174 passed)
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline` (177
  passed)
- focused middle-node ordering and activated-prefix reinsertion regression tests
- 128-worker / 128-session `perf record`: 5,941 samples, zero lost

### DIFF
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -7,8 +7,6 @@
 //! the pool models occupancy, reference/pin state, and LRU eviction without
 //! reproducing vLLM's numeric block IDs or null block.
 
-use std::collections::BTreeMap;
-
 use dynamo_tokens::SequenceHash;
 use rustc_hash::{FxHashMap, FxHashSet};
 use slotmap::{SlotMap, new_key_type};
@@ -21,11 +19,15 @@ new_key_type! {
 #[derive(Debug)]
 enum CopyState {
     Private,
+    /// A cached copy is linked into the inactive LRU if and only if both
+    /// `refs` and `pins` are zero. Any future cached sub-state must preserve or
+    /// explicitly revise that membership invariant.
     Cached {
         hash: SequenceHash,
         refs: usize,
         pins: usize,
-        inactive_at: Option<u64>,
+        inactive_prev: Option<BlockCopyId>,
+        inactive_next: Option<BlockCopyId>,
     },
 }
 
@@ -66,9 +68,10 @@ pub(crate) struct VllmBlockPool {
     capacity: usize,
     copies: SlotMap<BlockCopyId, BlockCopy>,
     by_hash: FxHashMap<SequenceHash, SmallVec<[BlockCopyId; 1]>>,
-    /// Ordinary LRU: lower timestamp is evicted first.
-    inactive: BTreeMap<u64, BlockCopyId>,
-    next_lru: u64,
+    /// Intrusive ordinary LRU: head is evicted first, tail was released last.
+    inactive_head: Option<BlockCopyId>,
+    inactive_tail: Option<BlockCopyId>,
+    inactive_len: usize,
     reserved: usize,
 }
 
@@ -79,8 +82,9 @@ impl VllmBlockPool {
             capacity,
             copies: SlotMap::with_key(),
             by_hash: FxHashMap::default(),
-            inactive: BTreeMap::new(),
-            next_lru: 0,
+            inactive_head: None,
+            inactive_tail: None,
+            inactive_len: 0,
             reserved: 0,
         }
     }
@@ -128,7 +132,7 @@ impl VllmBlockPool {
                 .filter_map(|(_, id)| self.is_inactive(*id).then_some(*id))
                 .collect::<FxHashSet<_>>()
                 .len();
-            let evictable_after_pins = self.inactive.len().saturating_sub(inactive_hits);
+            let evictable_after_pins = self.inactive_len.saturating_sub(inactive_hits);
             if needed_evictions > evictable_after_pins {
                 return None;
             }
@@ -158,7 +162,7 @@ impl VllmBlockPool {
     fn reserve_fresh(&mut self, fresh: usize) -> Option<ReserveOutcome> {
         let free = self.free_capacity();
         let needed_evictions = fresh.saturating_sub(free);
-        if needed_evictions > self.inactive.len() {
+        if needed_evictions > self.inactive_len {
             return None;
         }
 
@@ -231,7 +235,8 @@ impl VllmBlockPool {
             hash,
             refs: 1,
             pins: 0,
-            inactive_at: None,
+            inactive_prev: None,
+            inactive_next: None,
         };
         self.by_hash.entry(hash).or_default().push(id);
         became_visible
@@ -278,7 +283,7 @@ impl VllmBlockPool {
     }
 
     pub(crate) fn num_active(&self) -> usize {
-        self.copies.len() - self.inactive.len() + self.reserved
+        self.copies.len() - self.inactive_len + self.reserved
     }
 
     pub(crate) fn num_active_refs(&self) -> usize {
@@ -292,7 +297,7 @@ impl VllmBlockPool {
     }
 
     pub(crate) fn num_inactive(&self) -> usize {
-        self.inactive.len()
+        self.inactive_len
     }
 
     pub(crate) fn capacity(&self) -> usize {
@@ -313,31 +318,24 @@ impl VllmBlockPool {
     }
 
     fn is_inactive(&self, id: BlockCopyId) -> bool {
-        matches!(
-            &self.copies[id].state,
-            CopyState::Cached {
-                inactive_at: Some(_),
-                ..
-            }
-        )
+        let CopyState::Cached { refs, pins, .. } = &self.copies[id].state else {
+            return false;
+        };
+        *refs == 0 && *pins == 0
     }
 
     fn pin(&mut self, id: BlockCopyId) {
-        let inactive_at = {
-            let CopyState::Cached {
-                pins, inactive_at, ..
-            } = &mut self.copies[id].state
-            else {
-                panic!("prefix hash points to a private copy")
-            };
-            *pins = pins
-                .checked_add(1)
-                .unwrap_or_else(|| panic!("pin count overflow"));
-            inactive_at.take()
-        };
-        if let Some(timestamp) = inactive_at {
-            assert_eq!(self.inactive.remove(&timestamp), Some(id));
+        // Must unlink before bumping pins: list membership is derived from
+        // refs and pins.
+        if self.is_inactive(id) {
+            self.unlink_inactive(id);
         }
+        let CopyState::Cached { pins, .. } = &mut self.copies[id].state else {
+            panic!("prefix hash points to a private copy")
+        };
+        *pins = pins
+            .checked_add(1)
+            .unwrap_or_else(|| panic!("pin count overflow"));
     }
 
     fn activate_pin(&mut self, id: BlockCopyId, expected_hash: SequenceHash) {
@@ -374,38 +372,209 @@ impl VllmBlockPool {
     }
 
     fn insert_inactive(&mut self, id: BlockCopyId) {
-        let timestamp = self.next_lru;
-        self.next_lru = self
-            .next_lru
+        debug_assert!(
+            self.is_inactive(id),
+            "only an unreferenced, unpinned cached copy can enter the inactive LRU"
+        );
+        // A singleton has no links, so head membership is its only
+        // double-insertion signal.
+        debug_assert_ne!(
+            self.inactive_head,
+            Some(id),
+            "copy is already in the inactive LRU"
+        );
+        let previous_tail = self.inactive_tail;
+        {
+            let (prev, next) = self.inactive_links_mut(id);
+            debug_assert!(
+                prev.is_none() && next.is_none(),
+                "copy entering the inactive LRU still has list links"
+            );
+            *prev = previous_tail;
+        }
+        if let Some(tail) = previous_tail {
+            let (_, next) = self.inactive_links_mut(tail);
+            let old_next = next.replace(id);
+            debug_assert!(
+                old_next.is_none(),
+                "inactive LRU tail already has a successor"
+            );
+        } else {
+            let old_head = self.inactive_head.replace(id);
+            debug_assert!(old_head.is_none(), "empty inactive LRU still has a head");
+        }
+        self.inactive_tail = Some(id);
+        self.inactive_len = self
+            .inactive_len
             .checked_add(1)
-            .unwrap_or_else(|| panic!("LRU timestamp overflow"));
-        let CopyState::Cached { inactive_at, .. } = &mut self.copies[id].state else {
-            panic!("private copy cannot enter inactive LRU")
+            .unwrap_or_else(|| panic!("inactive block count overflow"));
+    }
+
+    fn inactive_links_mut(
+        &mut self,
+        id: BlockCopyId,
+    ) -> (&mut Option<BlockCopyId>, &mut Option<BlockCopyId>) {
+        let CopyState::Cached {
+            inactive_prev,
+            inactive_next,
+            ..
+        } = &mut self.copies[id].state
+        else {
+            panic!("inactive LRU link target is not a cached copy")
         };
-        assert!(inactive_at.replace(timestamp).is_none());
-        assert!(self.inactive.insert(timestamp, id).is_none());
+        (inactive_prev, inactive_next)
+    }
+
+    fn unlink_inactive(&mut self, id: BlockCopyId) {
+        debug_assert!(
+            self.is_inactive(id),
+            "only an unreferenced, unpinned cached copy can leave the inactive LRU"
+        );
+        let (previous, next) = {
+            let (previous, next) = self.inactive_links_mut(id);
+            (previous.take(), next.take())
+        };
+
+        if let Some(previous) = previous {
+            let (_, previous_next) = self.inactive_links_mut(previous);
+            let old_next = std::mem::replace(previous_next, next);
+            debug_assert_eq!(
+                old_next,
+                Some(id),
+                "inactive LRU predecessor does not point to the removed copy"
+            );
+        } else {
+            let old_head = std::mem::replace(&mut self.inactive_head, next);
+            debug_assert_eq!(
+                old_head,
+                Some(id),
+                "inactive LRU head does not match the removed copy"
+            );
+        }
+
+        if let Some(next) = next {
+            let (next_previous, _) = self.inactive_links_mut(next);
+            let old_previous = std::mem::replace(next_previous, previous);
+            debug_assert_eq!(
+                old_previous,
+                Some(id),
+                "inactive LRU successor does not point to the removed copy"
+            );
+        } else {
+            let old_tail = std::mem::replace(&mut self.inactive_tail, previous);
+            debug_assert_eq!(
+                old_tail,
+                Some(id),
+                "inactive LRU tail does not match the removed copy"
+            );
+        }
+
+        self.inactive_len = self
+            .inactive_len
+            .checked_sub(1)
+            .unwrap_or_else(|| panic!("inactive block count underflow"));
+    }
+
+    #[cfg(test)]
+    fn assert_lru_consistent(&self) {
+        let mut linked = FxHashSet::default();
+        let mut previous = None;
+        let mut cursor = self.inactive_head;
+
+        while let Some(id) = cursor {
+            assert!(linked.insert(id), "inactive LRU contains a cycle");
+            let Some(copy) = self.copies.get(id) else {
+                panic!("inactive LRU points to a missing copy")
+            };
+            let CopyState::Cached {
+                refs,
+                pins,
+                inactive_prev,
+                inactive_next,
+                ..
+            } = &copy.state
+            else {
+                panic!("inactive LRU contains a private copy")
+            };
+            assert_eq!(
+                (*refs, *pins),
+                (0, 0),
+                "active copy is linked into the inactive LRU"
+            );
+            assert_eq!(
+                *inactive_prev, previous,
+                "inactive LRU contains a broken back-pointer"
+            );
+            previous = Some(id);
+            cursor = *inactive_next;
+        }
+
+        assert_eq!(
+            linked.len(),
+            self.inactive_len,
+            "inactive LRU length does not match its reachable copies"
+        );
+        assert_eq!(
+            self.inactive_tail, previous,
+            "inactive LRU tail does not match its final reachable copy"
+        );
+        assert_eq!(
+            self.inactive_head.is_none(),
+            self.inactive_tail.is_none(),
+            "inactive LRU head and tail emptiness disagree"
+        );
+
+        for (id, copy) in self.copies.iter() {
+            let CopyState::Cached {
+                refs,
+                pins,
+                inactive_prev,
+                inactive_next,
+                ..
+            } = &copy.state
+            else {
+                assert!(!linked.contains(&id), "private copy is in the inactive LRU");
+                continue;
+            };
+            let should_be_linked = *refs == 0 && *pins == 0;
+            assert_eq!(
+                linked.contains(&id),
+                should_be_linked,
+                "cached copy membership disagrees with its refs and pins"
+            );
+            if !should_be_linked {
+                assert!(
+                    inactive_prev.is_none() && inactive_next.is_none(),
+                    "active cached copy retains inactive LRU links"
+                );
+            }
+        }
     }
 
     /// Evict one physical copy. A hash is returned only on its final copy.
     fn evict_one(&mut self) -> Option<SequenceHash> {
-        let Some((timestamp, id)) = self.inactive.pop_first() else {
+        let Some(id) = self.inactive_head else {
             panic!("prechecked inactive capacity disappeared")
         };
+        let CopyState::Cached { inactive_prev, .. } = &self.copies[id].state else {
+            panic!("inactive LRU points to a private copy")
+        };
+        assert!(
+            inactive_prev.is_none(),
+            "inactive LRU head has a predecessor"
+        );
+        self.unlink_inactive(id);
         let Some(copy) = self.copies.remove(id) else {
             panic!("inactive LRU points to a missing copy")
         };
         let CopyState::Cached {
-            hash,
-            refs,
-            pins,
-            inactive_at,
+            hash, refs, pins, ..
         } = copy.state
         else {
             panic!("inactive LRU points to a private copy")
         };
-        assert_eq!(refs, 0);
-        assert_eq!(pins, 0);
-        assert_eq!(inactive_at, Some(timestamp));
+        assert_eq!(refs, 0, "evicted cached copy still has references");
+        assert_eq!(pins, 0, "evicted cached copy is still pinned");
 
         let remove_hash = {
             let Some(copies) = self.by_hash.get_mut(&hash) else {
@@ -450,6 +619,7 @@ mod tests {
         pool.release(first_id);
         pool.release(second_id);
         assert_eq!(pool.num_inactive(), 2);
+        pool.assert_lru_consistent();
     }
 
     #[test]
@@ -463,6 +633,7 @@ mod tests {
         assert!(pool.reserve(&[9], 1).is_none());
         assert_eq!(pool.num_active(), 0);
         assert_eq!(pool.num_inactive(), 1);
+        pool.assert_lru_consistent();
     }
 
     #[test]
@@ -484,6 +655,7 @@ mod tests {
         let second_eviction = reserve(&mut pool, &[], 2);
         assert_eq!(second_eviction.removed, vec![3]);
         pool.cancel(second_eviction.reservation);
+        pool.assert_lru_consistent();
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -25,7 +23,8 @@ enum CopyState {
         hash: SequenceHash,
         refs: usize,
         pins: usize,
-        inactive_at: Option<u64>,
+        inactive_prev: Option<BlockCopyId>,
+        inactive_next: Option<BlockCopyId>,
```
jthomson04, line 30: The invariant `refs == 0 && pins == 0` <=> linked into the inactive list is the load-bearing fact of this whole design, and it is written down nowhere. Please document it here (or on `is_inactive`).

Concretely: a future `CopyState::Cached` sub-state that legitimately has zero refs and zero pins while deliberately *out* of the list — an in-flight transfer, a quiesced copy — would make `is_inactive` lie, and `unlink_inactive`'s assertions would start firing on a state that is not actually corrupt. The old `inactive_at: Option<u64>` encoded membership directly and could not drift; the derived predicate can.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -313,31 +315,22 @@ impl VllmBlockPool {
...
-                panic!("prefix hash points to a private copy")
-            };
-            *pins = pins
-                .checked_add(1)
-                .unwrap_or_else(|| panic!("pin count overflow"));
-            inactive_at.take()
-        };
-        if let Some(timestamp) = inactive_at {
-            assert_eq!(self.inactive.remove(&timestamp), Some(id));
+        if self.is_inactive(id) {
```
jthomson04, line 330: Correct, and subtle enough to deserve a comment: the `is_inactive` check has to happen *before* `pins` is incremented, because `unlink_inactive` asserts on the derived predicate. Swapping these two statements would make the unlink silently never happen, leaving a pinned copy linked in the LRU and evictable — a real corruption, with no test that obviously points at the cause.

Something like `// Must unlink before bumping pins: list membership is derived from refs/pins.`

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +367,93 @@ impl VllmBlockPool {
     }
 
     fn insert_inactive(&mut self, id: BlockCopyId) {
-        let timestamp = self.next_lru;
-        self.next_lru = self
-            .next_lru
+        assert!(self.is_inactive(id));
```
jthomson04: Bare, message-less assertions here and at `unlink_inactive` are the only ones in a file that otherwise consistently carries messages (`"capacity must be > 0"`, `"cached-copy reference underflow"`, `"prefix pin underflow"`).

Also the perf angle: this re-indexes `self.copies[id]` immediately before `inactive_links_mut` does the same lookup. `debug_assert!` here (matching the `debug_assert_*` convention in `scheduler/vllm/core.rs`) would drop it from release builds entirely.

jthomson04 (commented):
## Review

Traced every state transition that can touch the list (`cache_private`, `release`, `pin`, `activate_pin`, `unpin`, `evict_one`, `cancel`). The load-bearing invariant holds: a `Cached` copy is linked iff `refs == 0 && pins == 0`. Ordering is preserved (append-to-tail / pop-head is equivalent to monotonic timestamp / `pop_first`), duplicate prefix hashes still behave (the second `pin()` sees `pins == 1` and skips the unlink), and there is no `unsafe` — using slotmap keys instead of raw pointers avoids the usual aliasing hazards of intrusive lists. Good call.

**No live correctness bug found.** Two robustness issues worth fixing before merge, both 1-3 lines; details inline:

1. The actual link writes live *inside* `assert!`/`assert_eq!` arguments. Correct today, but converting them to `debug_assert!` — the obvious next step for a perf PR, and a pattern this crate already uses in `scheduler/vllm/core.rs:1219` — would silently delete the list maintenance in release builds.
2. `insert_inactive`'s double-insert guard has a hole for a single-element list, which the old `inactive_at.replace()` assertion caught unconditionally.

### Performance

Direction is clearly right and the numbers are internally consistent. Worth recording: the struct does not grow. `Option<u64>` (16 B incl. padding) to two `Option<BlockCopyId>` (8 B each, via slotmap's `NonZeroU32` version niche) is 40 B of payload either way — so the -6% peak RSS is attributable purely to eliminating B-tree node allocation, which corroborates the measurement.

One caveat on the wall-clock claim: 1.386x exceeds what the profile share alone predicts (13.1 pp of self-time removed implies ~1.15x by Amdahl). It is explainable — the captures are at different scales (128/128 profile vs 16/16 A/B) and allocator/cache effects do not show up as LRU self-time — but with a single run per arm, a repeat would firm it up. Not blocking.

Missed optimization, given this is a perf PR: the hot paths do redundant slotmap indexing on the same key. `pin()` on an inactive copy performs five `self.copies[id]` lookups (`is_inactive`, `unlink_inactive`'s assert, `inactive_links_mut`, the neighbor, the final `let-else`). Making the `is_inactive` asserts `debug_assert!` and hoisting the destructure would cut most of that.

### Tests

`pinning_middle_inactive_copy_preserves_lru_order` is a good test — hand-traced, and it genuinely validates both neighbor repair *and* `inactive_tail` repair (the later re-insert of `middle` only lands correctly if `tail` still points at `3`). Combined with `canceled_prefix_evicts_leaf_before_parent`, which pins the tail and then a sole node, the head/tail/middle/single-element unlink branches are all exercised.

The gap is that nothing validates list structure directly — every assertion is indirect via eviction order. See the inline suggestion for a `assert_lru_consistent` helper.

### Security

Nothing of concern. Simulator-internal, no untrusted input, no `unsafe`, no new dependencies.

### Verdict

LGTM with nits — the two assertion issues are the only things I would ask to be addressed. (Posting as a comment rather than an approval since it is your own PR.)

On the stacking: `lib/mocker/src/cache/vllm_block_pool.rs` is byte-identical on `main` and on #12242's head (blob `5256c002`), so #12242 does not touch this file and the retarget to `main` will be a no-op rebase.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +367,93 @@ impl VllmBlockPool {
...
+        let previous_tail = self.inactive_tail;
+        {
+            let (prev, next) = self.inactive_links_mut(id);
+            assert!(prev.is_none() && next.is_none());
```
jthomson04: **Double-insert guard has a hole for a single-element list.**

This catches re-inserting a head-of-N node (`next` is `Some`) or a tail node (`prev` is `Some`). But if `id` is the *only* node in the list — `head == tail == id`, both links `None` — every assertion in this function passes:

- line 374: both `None`, passes
- `previous_tail` is `Some(id)`, so we take the `if let Some(tail)` branch with `tail == id`
- line 379: `next.replace(id)` returns `None`, passes

Result: `id.prev == id`, `id.next == id`, `head == tail == id`, `inactive_len == 2`. Self-loop, silently corrupt. The old `assert!(inactive_at.replace(timestamp).is_none())` caught this unconditionally regardless of list position.

One line closes it:

```suggestion
        debug_assert_ne!(
            self.inactive_head,
            Some(id),
            "copy is already in the inactive LRU"
        );
        assert!(prev.is_none() && next.is_none());
```

I could not find a reachable path via the public API — `release` asserts `*refs > 0` and `unpin` asserts `*pins > 0` before either can reach `insert_inactive` twice — so this is defense-in-depth, not a live bug.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +367,93 @@ impl VllmBlockPool {
...
+            *prev = previous_tail;
+        }
+        if let Some(tail) = previous_tail {
+            let (_, next) = self.inactive_links_mut(tail);
+            assert!(next.replace(id).is_none());
+        } else {
+            assert!(self.inactive_head.replace(id).is_none());
+        }
```
jthomson04, line 405: **Structural mutation inside assertion macros.**

`next.replace(id)` and `self.inactive_head.replace(id)` are not checks — they *are* the link writes. Same pattern at lines 414/416/421/424 in `unlink_inactive`.

This is correct today, since `assert!` is never compiled out. The problem is that it is a trap for the obvious follow-up: this is a perf PR, these are hot-path assertions, and the crate already has a `debug_assert_*` convention (`scheduler/vllm/core.rs:1219`). Anyone converting `assert!` -> `debug_assert!` here deletes the entire list maintenance from release builds while debug builds and all tests keep passing.

Split the effect from the check:

```rust
if let Some(tail) = previous_tail {
    let (_, next) = self.inactive_links_mut(tail);
    let old = next.replace(id);
    debug_assert!(old.is_none());
} else {
    let old = self.inactive_head.replace(id);
    debug_assert!(old.is_none());
}
```

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +367,93 @@ impl VllmBlockPool {
...
+
+        if let Some(next) = next {
+            let (next_previous, _) = self.inactive_links_mut(next);
+            assert_eq!(std::mem::replace(next_previous, previous), Some(id));
+        } else {
+            assert_eq!(
+                std::mem::replace(&mut self.inactive_tail, previous),
+                Some(id)
+            );
+        }
```
jthomson04, line 470: Same issue as `insert_inactive`: all four `std::mem::replace` calls in this block are the actual link/head/tail updates, sitting inside `assert_eq!` arguments. Hoist each into a `let old = ...` and assert on `old`, so the unlink survives any future `assert!` -> `debug_assert!` pass.

The list logic itself is right — I checked all four branch combinations (head, tail, middle, sole node) and neighbor plus head/tail repair is correct in each.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +367,93 @@ impl VllmBlockPool {
...
+            inactive_prev,
+            inactive_next,
         } = copy.state
         else {
             panic!("inactive LRU points to a private copy")
         };
         assert_eq!(refs, 0);
         assert_eq!(pins, 0);
-        assert_eq!(inactive_at, Some(timestamp));
+        assert!(inactive_prev.is_none() && inactive_next.is_none());
```
jthomson04: This assertion is now a tautology — `unlink_inactive` on line 440 `take()`s both links immediately above, so it cannot fail and verifies nothing.

Worth noting what was lost: the old `assert_eq!(inactive_at, Some(timestamp))` cross-checked that the copy the LRU pointed at agreed it was in the LRU at that position. There is no equivalent check now. Either drop this to `..` in the destructure, or replace it with something that can actually fail (e.g. assert `self.inactive_head == Some(id)` before the unlink on line 440).

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -510,4 +558,27 @@ mod tests {
         assert!(pool.prefix_hit(8).is_none());
         pool.cancel(pressure.reservation);
     }
+
+    #[test]
+    fn pinning_middle_inactive_copy_preserves_lru_order() {
```
jthomson04, line 688: Good test — I hand-traced it and it validates more than it looks like it does. After `pin(middle)` unlinks from the middle of `[1, 2, 3]`, the later `cancel` re-inserts `middle` at the tail, which only lands correctly if `inactive_tail` still points at `3`. So neighbor repair *and* tail repair are both covered, and combined with `canceled_prefix_evicts_leaf_before_parent` (which pins the tail, then a sole node) every unlink branch is exercised.

The gap: every assertion here is indirect, via eviction order. Nothing checks list structure. A test-only validator would turn "the eviction order happened to come out right" into "the list is well-formed", and would catch the single-element double-insert hole flagged on line 374:

```rust
#[cfg(test)]
impl VllmBlockPool {
    fn assert_lru_consistent(&self) {
        let mut walked = 0;
        let mut previous = None;
        let mut cursor = self.inactive_head;
        while let Some(id) = cursor {
            let CopyState::Cached { refs, pins, inactive_prev, inactive_next, .. } =
                &self.copies[id].state
            else {
                panic!("inactive LRU contains a private copy")
            };
            assert_eq!((*refs, *pins), (0, 0), "active copy in inactive LRU");
            assert_eq!(*inactive_prev, previous, "broken back-pointer");
            walked += 1;
            assert!(walked <= self.inactive_len, "inactive LRU contains a cycle");
            previous = cursor;
            cursor = *inactive_next;
        }
        assert_eq!(walked, self.inactive_len);
        assert_eq!(self.inactive_tail, previous);
    }
}
```

Calling it at the end of each test in this module is cheap and high value for an intrusive-list change.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Documented the iff membership invariant on CopyState::Cached, including the requirement for future cached substates to preserve or revise it.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added the ordering comment and restructured pin: active copies increment through the initial borrow, while inactive copies unlink before pins transitions from zero to one.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Converted the insert and unlink preconditions to message-bearing debug assertions, so release builds no longer pay the redundant is_inactive lookup.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added the inactive-head guard before mutation, closing the sole-node reinsertion hole. The consistency validator also detects cycles and membership drift.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Hoisted the tail and head replacements into standalone writes and now debug-check their previous values separately.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Hoisted all four neighbor, head, and tail replacements out of assertion macros, so the structural mutations remain in release builds.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Removed the post-unlink link assertion and destructure the cleared links with ..; the message-bearing hard checks for references and pins remain.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added a test-only consistency validator for cycles, forward and backward links, head and tail, length, and refs/pins membership. It runs in all five tests and at the middle-node test's intermediate transitions.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +386,200 @@ impl VllmBlockPool {
...
+        }
     }
 
     /// Evict one physical copy. A hash is returned only on its final copy.
     fn evict_one(&mut self) -> Option<SequenceHash> {
-        let Some((timestamp, id)) = self.inactive.pop_first() else {
+        let Some(id) = self.inactive_head else {
             panic!("prechecked inactive capacity disappeared")
         };
+        self.unlink_inactive(id);
```
jthomson04, line 566: **Release-mode invariant checks were dropped.** Every structural check in the new code is `debug_assert` (`insert_inactive`, `unlink_inactive`, and here). The old code asserted `inactive_at == Some(timestamp)` at this point, and asserted list/flag agreement on insert, in release builds too.

This matters because the mocker runs in release for exactly the benchmarks this PR optimizes. In release now:

- a double-insert of the current tail sets that node's `next` to itself, producing a self-loop;
- `unlink_inactive` on an unlinked-but-`is_inactive` copy takes the `previous == None` branch and silently clears `inactive_head`/`inactive_tail` for the whole list.

Both used to be immediate panics; now they are silent corruption that surfaces later as wrong eviction order or a confusing `inactive block count underflow`. The `checked_add`/`checked_sub` on `inactive_len` and the `refs`/`pins` asserts just below do survive in release, which limits the blast radius.

Suggest keeping one O(1) release-mode structural check here, e.g. assert the head's `inactive_prev.is_none()` before unlinking. That is a single field read on a cache line you are already touching.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -313,31 +318,38 @@ impl VllmBlockPool {
...
+            } else {
+                *pins = pins
+                    .checked_add(1)
+                    .unwrap_or_else(|| panic!("pin count overflow"));
+                false
+            }
         };
-        if let Some(timestamp) = inactive_at {
-            assert_eq!(self.inactive.remove(&timestamp), Some(id));
+        if !was_inactive {
```
jthomson04: The compute-flag / early-return / re-borrow shape exists only to satisfy the borrow checker, and it splits the overflow check across two paths (the inactive path skips `checked_add` because it goes 0 -> 1, which is correct but subtle). This is equivalent and reads better:

```rust
fn pin(&mut self, id: BlockCopyId) {
    if self.is_inactive(id) {
        self.unlink_inactive(id);
    }
    let CopyState::Cached { pins, .. } = &mut self.copies[id].state else {
        panic!("prefix hash points to a private copy")
    };
    *pins = pins.checked_add(1).unwrap_or_else(|| panic!("pin count overflow"));
}
```

`is_inactive` returns `false` for `Private`, so the let-else still produces the same panic. Costs one extra slotmap index on the inactive path; `pin` is not the hot spot the profile called out (`insert_inactive` at 5.02% was).

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +386,200 @@ impl VllmBlockPool {
...
-            .next_lru
+        debug_assert!(
+            self.is_inactive(id),
+            "only an unreferenced, unpinned cached copy can enter the inactive LRU"
+        );
+        debug_assert_ne!(
+            self.inactive_head,
+            Some(id),
+            "copy is already in the inactive LRU"
+        );
```
jthomson04, line 385: This looks redundant next to the `prev.is_none() && next.is_none()` check just below, but it is in fact the only one of the two that catches re-inserting the sole member of a singleton list (where both links are already `None`). That is non-obvious enough to deserve a one-line comment, or the two checks could be folded into a single membership assertion.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +386,200 @@ impl VllmBlockPool {
...
+        &mut self,
+        id: BlockCopyId,
+    ) -> (&mut Option<BlockCopyId>, &mut Option<BlockCopyId>) {
+        let CopyState::Cached {
+            inactive_prev,
+            inactive_next,
+            ..
+        } = &mut self.copies[id].state
+        else {
             panic!("private copy cannot enter inactive LRU")
```
jthomson04: This message is accurate on the insert path but misleading when it fires while walking a predecessor or successor from `unlink_inactive`. Something neutral like `"inactive LRU link target is not a cached copy"` covers both callers. Same applies to the `unreachable!("inactive LRU contains a private copy")` in `pin`.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +386,200 @@ impl VllmBlockPool {
...
+            let old_head = self.inactive_head.replace(id);
+            debug_assert!(old_head.is_none(), "empty inactive LRU still has a head");
+        }
+        self.inactive_tail = Some(id);
+        self.inactive_len = self
+            .inactive_len
             .checked_add(1)
-            .unwrap_or_else(|| panic!("LRU timestamp overflow"));
-        let CopyState::Cached { inactive_at, .. } = &mut self.copies[id].state else {
+            .expect("inactive block count overflow");
```
jthomson04: Style nit: the rest of this file consistently uses `.unwrap_or_else(|| panic!(...))` — `free_capacity`, `activate_pin`, and the `insert_inactive` this replaces. `expect` here (and in `unlink_inactive`) breaks that uniformity.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -374,38 +386,200 @@ impl VllmBlockPool {
...
+        }
+
+        self.inactive_len = self
+            .inactive_len
+            .checked_sub(1)
+            .expect("inactive block count underflow");
+    }
+
+    #[cfg(test)]
+    fn assert_lru_consistent(&self) {
```
jthomson04, line 479: Consider making this `pub(crate)` (still under `#[cfg(test)]`). As a module-private fn it can only be called from the five unit tests below, all of which use capacity <= 3. The 173 `replay::offline` tests exercise this pool orders of magnitude harder, and being able to assert consistency from there is probably the cheapest large coverage win available on this change.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -509,5 +686,34 @@ mod tests {
         assert!(pool.prefix_hit(7).is_some());
         assert!(pool.prefix_hit(8).is_none());
         pool.cancel(pressure.reservation);
+        pool.assert_lru_consistent();
+    }
+
+    #[test]
+    fn pinning_middle_inactive_copy_preserves_lru_order() {
```
jthomson04, line 688: Good test — it covers middle unlink and pins down the re-entry-at-tail ordering that the timestamp scheme used to give for free.

Coverage gap worth closing alongside it: no unit test in this module reaches the LRU through `activate_prefix`/`activate_pin`. All five go via `cancel`/`unpin`. The `activate_pin` -> `release` path is where a copy takes `refs`, later drops to zero, and re-enters through `insert_inactive` — and it is exactly the path that breaks if `activate_pin` ever leaves stale links behind. The `debug_assert` in `insert_inactive` would catch it, but only if a test walks that path. Something like `reserve(&[h], 0)` -> `activate_prefix` -> `release` -> `assert_lru_consistent` would do it.

jthomson04 (commented):
## Review — intrusive inactive block LRU

Scope: the commit on this branch (`9e45ba9`) touches only `vllm_block_pool.rs`. The `engine.rs` hunks in the PR diff are #12242's content resurfacing because that PR was squash-merged; a rebase onto current `main` will drop them.

**The data-structure change is correct.** I traced the membership invariant (`linked <=> Cached && refs == 0 && pins == 0`) through every transition — `cache_private` (`refs: 1`), `release`, `pin`, `activate_pin`, `unpin`, `evict_one` — and it holds. Ordering is exactly equivalent to the old timestamp scheme: append-at-tail is the old "new timestamp", head eviction is `pop_first`. Specific cases checked:

- **Repeated hash in one `reserve` prefix** — `hits` holds the same id twice; the first `pin` unlinks and sets `pins = 1`, the second takes the `else` branch to `pins = 2`, and `cancel` unpins in reverse and re-inserts once. Same as before.
- **`inactive_hits` pre-check** — `is_inactive` is now derived from `refs`/`pins` rather than list membership, but those were equivalent under the old code, so `evictable_after_pins` is unchanged.
- **`activate_pin`** transiently has `pins == 0 && refs == 0` between the decrement and the increment, but no list operation observes it. Fine single-threaded; worth knowing it is load-bearing.

No API surface changed — every modified item is private to the module, so nothing outside this file can be affected. `Option<BlockCopyId>` is 8 bytes (slotmap's version field is `NonZeroU32`, so the niche applies), so `CopyState::Cached` is the same size as before: the two link fields exactly replace the 16-byte `Option<u64>`.

### The one thing I would change before merge

All structural invariant checks are now `debug_assert`. The old code carried **release-mode** assertions (`assert!(inactive_at.replace(timestamp).is_none())`, `assert!(self.inactive.insert(timestamp, id).is_none())`, `assert_eq!(inactive_at, Some(timestamp))`). See the inline comment on `evict_one` — one O(1) release-mode check restores most of the safety net for free.

### Test coverage

`assert_lru_consistent` is a good addition, and `pinning_middle_inactive_copy_preserves_lru_order` is well chosen: it covers middle unlink *and* verifies the unpinned copy returns as the newest LRU entry, which is the property most likely to silently regress. Together with `canceled_prefix_evicts_leaf_before_parent_under_pressure` (tail unlink, then singleton unlink) and eviction-driven head unlink, the structural cases are covered. Two gaps noted inline.

Everything else below is polish.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added an O(1) release-mode check before head eviction that requires the head's `inactive_prev` to be `None`. I also added a corruption regression test and ran that test under `cargo test --release`; it passes with the expected panic.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Simplified `pin` to this shape: inactive copies unlink first, then every cached copy uses the same checked pin increment. The private-copy panic and unlink-before-increment invariant remain explicit; all 7 focused tests pass.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added a comment explaining the non-obvious case: a singleton has no links, so head membership is its only double-insertion signal.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Changed the shared helper panic to the neutral `inactive LRU link target is not a cached copy`. The `pin` simplification also removed the old misleading `unreachable!` entirely.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Replaced both `inactive_len` `expect` calls with the file's established `unwrap_or_else(|| panic!(...))` style.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: I kept the validator module-private to preserve the PR's one-file scope: replay-wide access would require test-only plumbing through the native manager, scheduler core, worker state, and runtimes. Instead I added the missing `activate_prefix`/`activate_pin` lifecycle test directly at the pool boundary. The broader offline suites still pass (174 standard, 177 with `kvbm-offload`).

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Added `activated_prefix_reenters_inactive_lru_on_release`, which exercises reserve-prefix → `activate_prefix`/`activate_pin` → release → inactive reinsertion → eviction, with consistency checks at the active and reinserted states. The focused suite now passes 7/7.

PeaBrane (approved):
✅

jthomson04:
/ok to test [45ab61d](https://github.com/ai-dynamo/dynamo/commit/45ab61ddca6cc9d6560fbcd92cd234a2f9848efd)

PR #12247, opened 2026-07-28
fix(sidecar): make vLLM disaggregation work end-to-end

tanmayv25:
## Summary

Two fixes to the vLLM sidecar's disaggregated path so prefill→decode actually works. Both are in the sidecar — no upstream vLLM change and no engine patching.

- **`backend-common` + `engine.rs`** — register prefill/decode under distinct discovery components (`prefill` vs `backend`) so the frontend can route the handoff; previously both used the same operator-injected component (`worker`) and requests went straight to decode → `BackendInvalidArgument: decode request is missing the prefill_result KV payload`. The mapping lives on `DisaggregationMode::discovery_component` in `backend-common` (zero-duplication rule); the **SGLang sidecar's identical local `component_for_mode` was removed** in favor of it. Per review, **aggregated workers keep their operator-configured `--component`/`DYN_COMPONENT`** — only the disagg roles override to the fixed names (a `from_args` test covers this).

- **`convert.rs`** — send `remote_port` in `kv_transfer_params` as a string. vLLM's NIXL connector f-strings the port into the side-channel ZMQ URL (`make_zmq_path` → `f"tcp://{host}:{port}"`), but the value crosses vLLM's gRPC contract as a `google.protobuf.Struct`, whose numbers are always `double`. An integer port therefore arrived as `5600.0`, producing `tcp://<ip>:5600.0`, which fails to parse → KV load failure → **zero generated tokens**. As a string vLLM renders `5600`; it only uses this value to build the URL (verified against vLLM `main`, which does not int-cast it in `make_zmq_path` or `nixl/metadata.py`). Includes a unit test; marked removable once vLLM coerces the port itself.

## Validation

Validated end-to-end on an amd64 GPU cluster (co-located prefill+decode, NIXL/UCX over EFA, no RDMA):

- Aggregated: worker `2/2`, registers `backend.generate`, `/v1/models` + `/v1/chat/completions` return normally (no regression from the component change).
- Disaggregated: prefill + decode `2/2`, register `prefill.generate` / `backend.generate` with **no `--component` flag**, engine left **unpatched**, decode log shows `Transfer plan: TransferTopology(...)`, end-to-end completion returned (`completion_tokens > 0`). Before this change the same setup returned 0 tokens.
- Full `cargo test -p dynamo-vllm-sidecar` green (14 tests), incl. the updated `prefill_decode_handoff_is_opaque_and_repeatable` boundary test (expects only `remote_port` stringified) and the new `component_honors_config_for_aggregated_but_fixes_disagg_roles`.

Companion deploy example: ai-dynamo/dynamo#12238 (uses `kv_role=kv_producer`/`kv_consumer`).

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12247" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved distributed inference configuration by ensuring remote port values are transmitted in the expected format.
  * Corrected worker routing so requests are directed to the appropriate prefill or backend component based on the selected processing mode.

### DIFF
--- a/lib/sidecar/vllm/src/convert.rs
+++ b/lib/sidecar/vllm/src/convert.rs
@@ -217,15 +217,19 @@ fn build_kv_parameters(
             );
             Some(serde_json::Value::Object(params))
         }
-        DisaggregationMode::Decode => Some(
-            prefill_result
+        DisaggregationMode::Decode => {
+            let mut params = prefill_result
                 .ok_or_else(|| {
                     client::invalid_argument(
                         "decode request is missing the prefill_result KV payload",
                     )
                 })?
-                .disaggregated_params,
-        ),
+                .disaggregated_params;
+            // The prefill handoff carries the NIXL side-channel port; normalize it
+            // so vLLM builds a valid ZMQ URL (see stringify_remote_port).
+            stringify_remote_port(&mut params);
+            Some(params)
+        }
         DisaggregationMode::Encode => {
             return Err(client::invalid_argument(
                 "encode mode is not supported by the vLLM sidecar",
@@ -238,6 +242,34 @@ fn build_kv_parameters(
         cache_salt: cache_salt.unwrap_or_default(),
         kv_transfer_params: kv_transfer_params.map(json_to_struct).transpose()?,
     })
+}
+
+/// vLLM's NIXL connector builds the decode->prefill side-channel address by
+/// f-stringing `remote_port` (`f"tcp://{host}:{port}"`). The value reaches the
+/// engine through a `google.protobuf.Struct`, whose numbers are always `double`,
+/// so an integer port arrives as `5600.0` and the ZMQ URL fails to parse. Send it
+/// as a string instead — vLLM renders `5600` and only ever uses it to build that
+/// URL. Remove once vLLM coerces the port itself.
+fn stringify_remote_port(params: &mut serde_json::Value) {
+    let serde_json::Value::Object(map) = params else {
+        return;
+    };
+    // struct_to_json recovers a whole port as an integer; coerce a whole float
+    // defensively so a bare `5600.0` never slips through unmodified.
+    let port = map.get("remote_port").and_then(|value| {
+        value.as_u64().or_else(|| {
+            value
+                .as_f64()
+                .filter(|f| f.fract() == 0.0 && *f >= 0.0)
+                .map(|f| f as u64)
+        })
+    });
+    if let Some(port) = port {
+        map.insert(
+            "remote_port".to_string(),
+            serde_json::Value::String(port.to_string()),
+        );
+    }
 }
 
 fn bool_extra(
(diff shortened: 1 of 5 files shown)


lib/sidecar/vllm/src/engine.rs
```diff
@@ -104,7 +104,9 @@ impl VllmSidecarEngine {
         };
         let config = WorkerConfig {
             namespace: args.sidecar.common.namespace,
-            component: args.sidecar.common.component,
+            // Derive the discovery component from the disaggregation role so the
+            // frontend can route prefill separately from decode.
+            component: mode.discovery_component().to_string(),
```
connorcarpenter15: Not sure what we want to do here, but we probably shouldn't just silently override the `component` arg.

> This silently ignores `--component` / `DYN_COMPONENT` for every vLLM mode. `CommonArgs` still exposes that operator contract, and `WorkerConfig.component` selects the actual discovery endpoint, so aggregated or custom-topology deployments can register under `backend`/`prefill` instead of the configured path and become undiscoverable. Please preserve explicit configuration where it is supported, or reject/remove the option when fixed role names are required; do not parse and silently discard it. A `from_args` test with a non-default `--component` should assert the returned `WorkerConfig.component`.

lib/sidecar/vllm/src/convert.rs
```diff
@@ -217,15 +217,19 @@ fn build_kv_parameters(
...
                     client::invalid_argument(
                         "decode request is missing the prefill_result KV payload",
                     )
                 })?
-                .disaggregated_params,
-        ),
+                .disaggregated_params;
+            // The prefill handoff carries the NIXL side-channel port; normalize it
+            // so vLLM builds a valid ZMQ URL (see stringify_remote_port).
+            stringify_remote_port(&mut params);
```
connorcarpenter15, line 230: This normalization conflicts with the existing `prefill_decode_handoff_is_opaque_and_repeatable` test: its fake prefill emits numeric `remote_port: 20097`, but after this call the recorded decode request contains `"20097"`, while `tests.rs:528` still asserts the full decoded payload equals the original numeric handoff. Please update that existing boundary test to expect only `remote_port` to be stringified while preserving every other field. That test uniquely protects the gRPC/protobuf contract, so the new private-helper-only unit test can be removed.

connorcarpenter15 (commented):
Two small comments.

lib/sidecar/vllm/src/engine.rs
tanmayv25: Good call — fixed. Aggregated now keeps the operator-configured component (`--component` / `DYN_COMPONENT`); only prefill/decode override to the fixed `prefill`/`backend` names the frontend routing requires. Added a `from_args` test asserting a non-default `--component` is preserved for aggregated, and that the disagg roles override.

lib/sidecar/vllm/src/convert.rs
tanmayv25: Fixed — updated `prefill_decode_handoff_is_opaque_and_repeatable` to expect only `remote_port` stringified while every other field round-trips unchanged, and removed the redundant private-helper unit test. Full `cargo test -p dynamo-vllm-sidecar` is green now (I'd missed this by filtering to just my test — thanks for catching it).

connorcarpenter15 (approved):

PR #12248, opened 2026-07-28
perf(mocker): flatten native G1 sequences

jthomson04:
## Summary

- keep each VLLM/TRT-LLM native-G1 request in one flat token representation
- build native block hashes, sequence hashes, lineage, and identities directly,
  avoiding `TokenBlockSequence` and per-block token allocations
- retain complete token history only when token-ID KV events require it;
  otherwise retain at most the current partial block plus one token
- bound eager flat-storage capacity to output realizable under the model and
  physical KV limits while preserving the logical request maximum
- preserve `TokenBlockSequence` for KVBM and leave SGLang unchanged

The optimized path applies to both offline replay and the live mocker, with
prefix caching enabled or disabled and with or without ZMQ KV-event output.
The public surface changes add the canonical `compute_block_hash_for_tokens`
helper and remove the unused `ActiveSequence::tokens()` getter. In addition,
`block_token_ids()` rejects native flat sequences whose token history was
intentionally discarded; its only production caller is the KVBM-only swap-in
path. Flat `pop()` also requires a retained freshly generated token and panics
on contract violation, while legacy storage preserves its historical no-op.
CLI options, event formats, and serialization formats are unchanged.

### Measured main versus PR: 16-worker / 16-session AgentX

Exact main control `a96a2a8c48` and measured PR build `8ac610d6f6` were built
from the same lockfile with identical release optimization, DWARF,
`x86-64-v3`, frame pointers, and the stock allocator. The PR arm ran first,
followed by main, with full teardown between arms and CPU 0 affinity. Current
head `366ff0ef52` includes subsequent review fixes, including bounded eager
capacity for overdeclared or rejected requests. The table therefore describes
the measured `8ac610d6f6` build rather than a rerun of the final head.

| Metric | Main control `a96a2a8c48` | PR build `8ac610d6f6` | Change |
| --- | ---: | ---: | ---: |
| Replay time | 8.108080 s | 5.881516 s | **-27.46% / 1.379x** |
| Host request throughput | 396.02 req/s | 545.95 req/s | **+37.86%** |
| Host processed-token throughput | 59.62M tok/s | 82.19M tok/s | **+37.86%** |
| `mprof` peak RSS | 253.590 MiB | 219.797 MiB | **-13.33%** |

Both arms completed exactly 3,211 requests, 479,843,008 input tokens,
3,552,472 output tokens, and 483,395,480 processed tokens. Deterministic report
metrics matched exactly (0.0 ppm).

### 128-worker / 128-session CPU profile

The established main/#12244 capture and the latest equivalent PR-patch capture
used the same 128-worker / 128-session AgentX workload:

| Targeted inclusive attribution | Main baseline | PR branch | Change |
| --- | ---: | ---: | ---: |
| `TokenBlockSequence::split_tokens` | 13.989% | 0.000% | **eliminated** |
| `ActiveSequence` construction | 15.187% legacy | 2.158% flat | **-85.79%** |
| Flat hashing | n/a | 0.320% | below 5% gate |
| Flat token-vector growth | n/a | 0.000% | no measured growth |
| Native metadata-vector growth | n/a | 0.000% | no measured growth |
| Flat tail promotion | n/a | 0.084% | below 5% gate |
| Flat tail compaction | n/a | 0.000% | no measured compaction |
| Token-ID materialization | n/a | 0.000% | no measured materialization |
| Unresolved leaf cycles | 0.419% | 0.302% | usable symbols |

The PR capture contains 5,940 samples, zero lost samples, and usable DWARF
symbols. No hashing, metadata-growth, tail-compaction, or token-materialization
path exceeds 5% self CPU. These fixed-window shares are CPU attribution rather
than a completed-throughput comparison; the completed A/B is reported above.

> **Follow-up:** Native G1’s retained-token buffer now overlaps with SGLang’s flat sequence storage. A separate PR can evaluate extracting shared token-storage and canonical hashing primitives while preserving the distinct native-G1 and SGLang KV-cache lifecycles. This refactor is intentionally out of scope here.

## Validation

- focused flat/legacy sequence parity and native-path selection tests
- eager and push-driven promotion parity, including bounded retained capacity
- eager-capacity regression for model-capped and post-materialization-rejected requests
- native ZMQ store/promotion token-ID payload parity
- prefix-caching-on hash/lineage parity and prefix-caching-off random identity
- canonical token-hash helper and shared-salt parity
- bounded retained-token storage and stable token/metadata decode capacities
- chunked token-ID allocation with a nonzero start offset
- partial-parent, promotion-position, reset, and rollback invariants
- native-G1 rejection of the legacy KVBM swap-in lifecycle
- `cargo test -p dynamo-mocker flat_`
- `cargo test -p dynamo-mocker scheduler::vllm`
- `cargo test -p dynamo-mocker replay::offline`
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline`
- locked metadata and clippy in the root, Python, runtime-example, and KVBM workspaces
- `cargo clippy -p dynamo-tokens -p dynamo-mocker --all-targets -- -D warnings`
- `git diff --check`
- exact main-control versus PR 16-worker / 16-session comparison with `mprof`
- 128-worker / 128-session AgentX `perf` capture

### DIFF
--- a/lib/mocker/src/common/sequence.rs
+++ b/lib/mocker/src/common/sequence.rs
@@ -4,9 +4,95 @@
 use crate::common::protocols::MoveBlock;
 use derive_getters::Getters;
 use dynamo_tokens::blocks::UniqueBlock;
-use dynamo_tokens::{BlockHash, PositionalLineageHash, TokenBlockSequence, Tokens};
+use dynamo_tokens::{
+    BlockHash, PositionalLineageHash, SaltHash, TokenBlockSequence, Tokens,
+    compute_block_hash_for_tokens, compute_next_sequence_hash,
+};
 use rand::random;
 use validator::Validate;
+
+const MOCKER_SALT_HASH: SaltHash = 1337;
+
+#[derive(Debug)]
+struct FlatTokens {
+    retained: Vec<u32>,
+    retained_start: usize,
+}
+
+impl FlatTokens {
+    fn new(
+        mut tokens: Vec<u32>,
+        output_capacity_hint: usize,
+        block_size: usize,
+        retain_history: bool,
+    ) -> Self {
+        if retain_history {
+            tokens.reserve_exact(output_capacity_hint);
+            return Self {
+                retained: tokens,
+                retained_start: 0,
+            };
+        }
+
+        let retained_start = tokens.len() - (tokens.len() % block_size);
+        // Lazy promotion runs after the next block's first token is pushed, so
+        // the retained window must hold one complete block plus that token.
+        let mut retained = Vec::with_capacity(
+            block_size
+                .checked_add(1)
+                .expect("flat token tail capacity overflow"),
+        );
+        retained.extend_from_slice(&tokens[retained_start..]);
+        Self {
+            retained,
+            retained_start,
+        }
+    }
+
+    fn len(&self) -> usize {
+        self.retained_start
+            .checked_add(self.retained.len())
+            .expect("flat token length overflow")
+    }
+
+    fn push(&mut self, token: u32) {
+        self.retained.push(token);
+    }
+
+    fn pop(&mut self) -> Option<u32> {
+        self.retained.pop()
+    }
+
+    fn complete_block(&self, position: usize, block_size: usize) -> Option<&[u32]> {
+        let start = position.checked_mul(block_size)?;
+        debug_assert!(
+            start >= self.retained_start,
+            "promoted block precedes retained flat-token window"
+        );
+        let end = start.checked_add(block_size)?;
+        let local_start = start.checked_sub(self.retained_start)?;
+        let local_end = end.checked_sub(self.retained_start)?;
+        self.retained.get(local_start..local_end)
+    }
+
+    fn discard_through(&mut self, absolute_end: usize) {
+        let local_end = absolute_end
+            .checked_sub(self.retained_start)
+            .expect("promoted block precedes retained flat-token window");
+        assert!(
+            local_end <= self.retained.len(),
+            "promoted block extends beyond retained flat-token window"
+        );
+        self.retained.drain(..local_end);
+        self.retained_start = absolute_end;
+    }
+}
+
+#[derive(Debug)]
+enum SequenceTokens {
+    Legacy(TokenBlockSequence),
+    Flat(FlatTokens),
+}
 
 /// Create unique blocks, block hashes, and positional-lineage hashes from a
 /// [`TokenBlockSequence`].
@@ -44,6 +130,53 @@ fn create_sequence_cache(
     (unique_blocks, block_hashes, plhs)
 }
 
+/// Build the native-G1 block metadata directly over a flat token vector.
+fn create_flat_sequence_cache(
+    tokens: &[u32],
+    block_size: usize,
+    enable_prefix_caching: bool,
+    completion_blocks: usize,
+) -> (Vec<UniqueBlock>, Vec<BlockHash>, Vec<PositionalLineageHash>) {
+    let mut unique_blocks = Vec::with_capacity(completion_blocks);
+    let mut block_hashes = if enable_prefix_caching {
+        Vec::with_capacity(completion_blocks)
+    } else {
+        Vec::new()
+    };
+    let mut plhs = Vec::with_capacity(completion_blocks);
+    let mut parent_hash = None;
+
+    for (position, block) in tokens.chunks_exact(block_size).enumerate() {
+        if enable_prefix_caching {
+            let block_hash = compute_block_hash_for_tokens(block, MOCKER_SALT_HASH);
+            let sequence_hash = parent_hash
+                .map(|parent| compute_next_sequence_hash(parent, block_hash))
+                .unwrap_or(block_hash);
+            block_hashes.push(block_hash);
+            unique_blocks.push(UniqueBlock::FullBlock(sequence_hash));
+            plhs.push(PositionalLineageHash::new(
+                sequence_hash,
+                parent_hash,
+                position as u64,
+            ));
+            parent_hash = Some(sequence_hash);
+        } else {
+            unique_blocks.push(UniqueBlock::FullBlock(random::<u64>()));
+            plhs.push(PositionalLineageHash::new(
+                random::<u64>(),
+                None,
+                position as u64,
+            ));
+        }
+    }
+
+    if !tokens.len().is_multiple_of(block_size) {
+        unique_blocks.push(UniqueBlock::default());
+    }
+
+    (unique_blocks, block_hashes, plhs)
+}
+
 /// A sequence that is actively being built, with the ability to add tokens and commit to hashes
 /// TODO: reuse tokens
 #[derive(Debug, Getters, Validate)]
@@ -52,7 +185,8 @@ pub struct ActiveSequence {
     block_hashes: Vec<BlockHash>,
     plhs: Vec<PositionalLineageHash>,
 
-    tokens: TokenBlockSequence,
+    #[getter(skip)]
+    tokens: SequenceTokens,
 
     #[getter(copy)]
     #[validate(range(min = 2))]
@@ -107,28 +241,69 @@ impl ActiveSequence {
             return None;
         };
 
-        let last_complete = self.tokens.last_complete_block().unwrap_or_else(|| {
-            panic!("partial sequence tail cannot be promoted without a complete token block")
-        });
-        let last_seq_hash = if self.enable_prefix_caching {
-            last_complete.sequence_hash()
-        } else {
-            random::<u64>()
-        };
-        let last_block_hash = self
-            .enable_prefix_caching
-            .then(|| last_complete.block_hash());
-        // With prefix caching off, the sequence hash and PLH must both remain
-        // request-unique so another identical prompt cannot reuse this slot.
-        let last_plh = if self.enable_prefix_caching {
-            last_complete.positional_lineage_hash()
-        } else {
-            PositionalLineageHash::new(random::<u64>(), None, self.plhs.len() as u64)
-        };
-        let promote_token_ids = if self.emit_token_ids {
-            Some(last_complete.tokens().to_vec())
-        } else {
-            None
+        let parent_hash = self.unique_blocks[..self.unique_blocks.len() - 1]
+            .last()
+            .map(|block| match block {
+                UniqueBlock::FullBlock(hash) => *hash,
+                UniqueBlock::PartialBlock(_) => panic!("partial block cannot be a parent"),
+            });
+        let position = self.plhs.len();
+        debug_assert_eq!(position + 1, self.len() / self.block_size);
+        let (last_seq_hash, last_block_hash, last_plh, promote_token_ids) = match &self.tokens {
+            SequenceTokens::Legacy(tokens) => {
+                let last_complete = tokens.last_complete_block().unwrap_or_else(|| {
+                    panic!(
+                        "partial sequence tail cannot be promoted without a complete token block"
+                    )
+                });
+                let last_seq_hash = if self.enable_prefix_caching {
+                    last_complete.sequence_hash()
+                } else {
+                    random::<u64>()
+                };
+                let last_block_hash = self
+                    .enable_prefix_caching
+                    .then(|| last_complete.block_hash());
+                // With prefix caching off, the sequence hash and PLH must both remain
+                // request-unique so another identical prompt cannot reuse this slot.
+                let last_plh = if self.enable_prefix_caching {
+                    last_complete.positional_lineage_hash()
+                } else {
+                    PositionalLineageHash::new(random::<u64>(), None, position as u64)
+                };
+                let promote_token_ids = if self.emit_token_ids {
+                    Some(last_complete.tokens().to_vec())
+                } else {
+                    None
+                };
+                (last_seq_hash, last_block_hash, last_plh, promote_token_ids)
+            }
+            SequenceTokens::Flat(tokens) => {
+                let complete = tokens
+                    .complete_block(position, self.block_size)
+                    .unwrap_or_else(|| {
+                    panic!(
+                        "partial flat sequence tail cannot be promoted without a complete token block"
+                    )
+                });
+                let last_block_hash = self
+                    .enable_prefix_caching
+                    .then(|| compute_block_hash_for_tokens(complete, MOCKER_SALT_HASH));
+                let last_seq_hash = last_block_hash
+                    .map(|block_hash| {
+                        parent_hash
+                            .map(|parent| compute_next_sequence_hash(parent, block_hash))
+                            .unwrap_or(block_hash)
+                    })
+                    .unwrap_or_else(random::<u64>);
+                let last_plh = if self.enable_prefix_caching {
+                    PositionalLineageHash::new(last_seq_hash, parent_hash, position as u64)
+                } else {
+                    PositionalLineageHash::new(random::<u64>(), None, position as u64)
+                };
+                let promote_token_ids = self.emit_token_ids.then(|| complete.to_vec());
+                (last_seq_hash, last_block_hash, last_plh, promote_token_ids)
+            }
         };
         if let Some(last_block_hash) = last_block_hash {
             self.block_hashes.push(last_block_hash);
@@ -136,13 +311,19 @@ impl ActiveSequence {
         self.plhs.push(last_plh);
         self.unique_blocks.pop();
 
-        // After pop, the last element is the parent block.
-        let parent_hash = self.unique_blocks.last().map(|block| match block {
-            UniqueBlock::FullBlock(hash) => *hash,
-            UniqueBlock::PartialBlock(_) => panic!("partial block cannot be a parent"),
-        });
         self.unique_blocks
             .push(UniqueBlock::FullBlock(last_seq_hash));
+
+        if !self.emit_token_ids {
+            let promoted_end = position
+                .checked_add(1)
+                .and_then(|blocks| blocks.checked_mul(self.block_size))
+                .expect("promoted flat-token boundary overflow");
+            if let SequenceTokens::Flat(tokens) = &mut self.tokens {
+                tokens.discard_through(promoted_end);
+            }
+        }
+        self.debug_assert_flat_token_invariants();
 
         Some(MoveBlock::Promote(
             uuid,
@@ -183,7 +364,7 @@ impl ActiveSequence {
         let block_size = block_size.unwrap_or(64);
         let num_input_tokens = tokens.len();
 
-        let tokens = Tokens::from(tokens).into_sequence(block_size as u32, Some(1337));
+        let tokens = Tokens::from(tokens).into_sequence(block_size as u32, Some(MOCKER_SALT_HASH));
         let (unique_blocks, block_hashes, plhs) =
             create_sequence_cache(&tokens, block_size, enable_prefix_caching);
 
@@ -191,7 +372,7 @@ impl ActiveSequence {
             unique_blocks,
             block_hashes,
             plhs,
-            tokens,
+            tokens: SequenceTokens::Legacy(tokens),
             block_size,
             max_output_tokens,
             generated_tokens: 0,
@@ -205,16 +386,67 @@ impl ActiveSequence {
         seq
     }
 
+    /// Build a native-G1 sequence directly over the request's flat token vector.
+    ///
+    /// `output_capacity_hint` controls eager allocation only. The logical
+    /// generation limit remains `max_output_tokens`, and the vectors can grow
+    /// if the scheduler realizes more output than the hint.
+    pub(crate) fn new_flat_with_planned_output_ids(
+        tokens: Vec<u32>,
+        max_output_tokens: usize,
+        output_capacity_hint: usize,
+        block_size: usize,
+        enable_prefix_caching: bool,
+        emit_token_ids: bool,
+        planned_output_ids: Option<Vec<u32>>,
+    ) -> Self {
+        let num_input_tokens = tokens.len();
+        let emit_token_ids = emit_token_ids && enable_prefix_caching;
+        let output_capacity_hint = output_capacity_hint.min(max_output_tokens);
+        let completion_blocks = num_input_tokens
+            .checked_add(output_capacity_hint)
+            .expect("native sequence completion length overflow")
+            .div_ceil(block_size);
+        let (unique_blocks, block_hashes, plhs) = create_flat_sequence_cache(
+            &tokens,
+            block_size,
+            enable_prefix_caching,
+            completion_blocks,
+        );
+        let tokens = FlatTokens::new(tokens, output_capacity_hint, block_size, emit_token_ids);
+
+        let seq = Self {
+            unique_blocks,
+            block_hashes,
+            plhs,
+            tokens: SequenceTokens::Flat(tokens),
+            block_size,
+            max_output_tokens,
+            generated_tokens: 0,
+            planned_output_ids,
+            num_input_tokens,
+            num_allocated_tokens: 0,
+            enable_prefix_caching,
+            emit_token_ids,
+        };
+        seq.validate().expect("invalid flat ActiveSequence");
+        seq.debug_assert_flat_token_invariants();
+        seq
+    }
+
     pub fn extra_tokens(&self) -> u32 {
         (self.len() % self.block_size) as u32
     }
 
     pub fn len(&self) -> usize {
-        self.tokens.total_tokens()
+        match &self.tokens {
+            SequenceTokens::Legacy(tokens) => tokens.total_tokens(),
+            SequenceTokens::Flat(tokens) => tokens.len(),
+        }
     }
 
     pub fn is_empty(&self) -> bool {
-        self.tokens.total_tokens() == 0
+        self.len() == 0
     }
 
     /// Current known sequence footprint in blocks: prompt plus generated tokens.
@@ -258,12 +490,7 @@ impl ActiveSequence {
         let plhs = self.plhs[plh_start..plh_end].to_vec();
 
         let token_ids = if self.emit_token_ids && hash_start < hash_end {
-            Some(
-                self.tokens.blocks()[hash_start..hash_end]
-                    .iter()
-                    .map(|b| b.tokens().to_vec())
-                    .collect(),
-            )
+            Some(self.block_token_ids_in(hash_start, hash_end))
         } else {
             None
         };
@@ -282,12 +509,37 @@ impl ActiveSequence {
         &self.plhs
     }
 
+    fn block_token_ids_in(&self, start: usize, end: usize) -> Vec<Vec<u32>> {
+        match &self.tokens {
+            SequenceTokens::Legacy(tokens) => tokens.blocks()[start..end]
+                .iter()
+                .map(|block| block.tokens().to_vec())
+                .collect(),
+            SequenceTokens::Flat(tokens) => {
+                assert!(
+                    self.emit_token_ids && tokens.retained_start == 0,
+                    "flat sequences retain full token history only when token-ID events are enabled"
+                );
+                tokens
+                    .retained
+                    .chunks_exact(self.block_size)
+                    .skip(start)
+                    .take(end - start)
+                    .map(<[u32]>::to_vec)
+                    .collect()
+            }
+        }
+    }
+
+    /// Materialize every complete block's token IDs.
+    ///
+    /// # Panics
+    ///
+    /// Panics for native flat sequences that were created without token-ID
+    /// event emission, because those sequences intentionally discard completed
+    /// prompt and decode blocks.
     pub fn block_token_ids(&self) -> Vec<Vec<u32>> {
-        self.tokens
-            .blocks()
-            .iter()
-            .map(|block| block.tokens().to_vec())
-            .collect()
+        self.block_token_ids_in(0, self.len() / self.block_size)
     }
 
     /// Commit a successful allocation by advancing `num_allocated_tokens`.
@@ -328,8 +580,14 @@ impl ActiveSequence {
     /// Push a token to the sequence
     #[cfg_attr(feature = "profile", inline(never))]
     pub fn push(&mut self, token: u32) -> Option<Vec<MoveBlock>> {
-        self.tokens.append(token).expect("Token push failed.");
+        match &mut self.tokens {
+            SequenceTokens::Legacy(tokens) => {
+                tokens.append(token).expect("Token push failed.");
+            }
+            SequenceTokens::Flat(tokens) => tokens.push(token),
+        }
         self.generated_tokens += 1;
+        self.debug_assert_flat_token_invariants();
 
         if self.len() % self.block_size != 1 {
             return None;
@@ -354,6 +612,7 @@ impl ActiveSequence {
             None,
             None,
         ));
+        self.debug_assert_flat_token_invariants();
         Some(signals)
     }
 
@@ -450,27 +709,90 @@ impl ActiveSequence {
     /// must be in the current partial block, so we only need to drop the trailing
     /// partial `UniqueBlock` when the sequence length returns to an exact block
     /// boundary. Using this to unwind arbitrary prompt history would be incorrect.
+    ///
+    /// If this contract is violated in release builds, legacy token storage
+    /// preserves its historical no-op on an empty buffer, while flat storage
+    /// panics to surface the invalid rollback.
     pub fn pop(&mut self) {
-        self.tokens.pop();
+        debug_assert!(
+            self.generated_tokens > 0,
+            "sequence rollback requires a freshly generated token"
+        );
+        match &mut self.tokens {
+            SequenceTokens::Legacy(tokens) => {
+                tokens.pop();
+            }
+            SequenceTokens::Flat(tokens) => {
+                debug_assert!(
+                    !tokens.retained.is_empty(),
+                    "flat rollback token must be retained"
+                );
+                tokens.pop().expect("flat rollback token must be retained");
+            }
+        }
         self.generated_tokens = self.generated_tokens.saturating_sub(1);
 
         // Reverts to the last full block
-        if self.tokens.total_tokens().is_multiple_of(self.block_size) {
+        if self.len().is_multiple_of(self.block_size) {
             self.unique_blocks.pop();
         }
+        self.debug_assert_flat_token_invariants();
+    }
+
+    fn debug_assert_flat_token_invariants(&self) {
+        if let SequenceTokens::Flat(tokens) = &self.tokens {
+            debug_assert_eq!(
+                tokens.len(),
+                self.num_input_tokens + self.generated_tokens,
+                "flat retained-token window must cover the logical sequence suffix"
+            );
+            if self.emit_token_ids {
+                debug_assert_eq!(
+                    tokens.retained_start, 0,
+                    "token-ID events require complete flat-token history"
+                );
+            } else {
+                debug_assert!(
+                    tokens.retained.len() <= self.block_size + 1,
+                    "non-emitting flat sequences retain at most one block plus the next token"
+                );
+            }
+        }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn uses_flat_tokens(&self) -> bool {
+        matches!(self.tokens, SequenceTokens::Flat(_))
+    }
+
+    #[cfg(test)]
+    pub(crate) fn flat_storage_capacities(&self) -> Option<(usize, usize, usize, usize)> {
+        let SequenceTokens::Flat(tokens) = &self.tokens else {
+            return None;
+        };
+        Some((
+            tokens.retained.capacity(),
+            self.unique_blocks.capacity(),
+            self.block_hashes.capacity(),
+            self.plhs.capacity(),
+        ))
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use dynamo_tokens::SequenceHash;
 
     fn block_hashes_from_tokens(seq: &ActiveSequence) -> Vec<BlockHash> {
-        seq.tokens
-            .blocks()
-            .iter()
-            .map(|block| block.block_hash())
-            .collect()
+        match &seq.tokens {
+            SequenceTokens::Legacy(tokens) => tokens
+                .blocks()
+                .iter()
+                .map(|block| block.block_hash())
+                .collect(),
+            SequenceTokens::Flat(_) => seq.block_hashes().clone(),
+        }
     }
 
     fn assert_cached_hashes_match_promoted_blocks(seq: &ActiveSequence) {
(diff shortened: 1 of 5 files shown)


lib/mocker/src/common/sequence.rs
```diff
@@ -2,12 +2,24 @@
...
 use dynamo_tokens::blocks::UniqueBlock;
-use dynamo_tokens::{BlockHash, PositionalLineageHash, TokenBlockSequence, Tokens};
+use dynamo_tokens::{
+    BlockHash, PositionalLineageHash, TokenBlockSequence, Tokens, compute_block_hash,
+    compute_next_sequence_hash,
+};
 use rand::random;
 use validator::Validate;
 
+const NATIVE_HASH_SEED: u64 = 1337;
```
jthomson04: **Misnamed, and duplicates a magic number that must not drift.**

`compute_block_hash`'s second parameter is a `SaltHash`, not a seed (`lib/tokens/src/lib.rs:148`). More importantly this value **must** equal the literal `1337` at `sequence.rs:282` (`Tokens::from(tokens).into_sequence(block_size as u32, Some(1337))`) or the two arms produce different hashes for identical prompts — a silent, hard-to-diagnose KV-index corruption with no compile error and no test failure outside the parity test.

Make it one constant used in both places, and name it for what it is:

```rust
const MOCKER_SALT_HASH: SaltHash = 1337;
```

The `NATIVE_` prefix is also misleading — it implies the value is native-specific when the entire point is that it *must not* be.

lib/mocker/src/common/sequence.rs
```diff
@@ -44,6 +56,52 @@ fn create_sequence_cache(
...
+    let mut block_hashes = Vec::new();
+    if enable_prefix_caching {
+        block_hashes.reserve(full_blocks);
+    }
+    let mut plhs = Vec::with_capacity(full_blocks);
+    let mut parent_hash = None;
+
+    for (position, block) in tokens.chunks_exact(block_size).enumerate() {
+        if enable_prefix_caching {
+            let block_hash = compute_block_hash(cast_slice(block), NATIVE_HASH_SEED);
```
jthomson04: **Silent MM divergence, and an avoidable dependency.**

`commit_current` has an MM-aware branch (`lib/tokens/src/lib.rs:1341-1345`) that hashes substituted bytes via `compute_block_bytes_with_mm`. This path unconditionally uses `cast_slice(block)`. The mocker never sets `mm_runs` today so there's no live bug, but if multimodal support is added to the mocker the native-G1 hashes diverge from legacy silently.

Consider pushing the helper into `dynamo-tokens` instead: `TokenBlockChunk::from_tokens` (`lib/tokens/src/lib.rs:1334`) is already exactly `compute_block_hash(cast_slice(tokens), salt)` but is private. Exposing

```rust
pub fn compute_block_hash_for_tokens(tokens: &[Token], salt: SaltHash) -> BlockHash
```

would drop the new `bytemuck` dependency from `lib/mocker/Cargo.toml`, centralize the token→bytes encoding, and give the MM divergence a single place to be fixed later. (`bytemuck = "1.22"` as a non-workspace dep does have precedent in `lib/tokens/Cargo.toml`, so the current form isn't wrong — just avoidable.)

lib/mocker/src/common/sequence.rs
```diff
@@ -52,7 +110,8 @@ pub struct ActiveSequence {
     block_hashes: Vec<BlockHash>,
     plhs: Vec<PositionalLineageHash>,
 
-    tokens: TokenBlockSequence,
+    #[getter(skip)]
```
jthomson04, line 188: Minor: the PR body says "Public APIs ... are unchanged", but `#[getter(skip)]` removes the derived `pub fn tokens(&self) -> &TokenBlockSequence` from a `pub` struct.

I grepped and found no in-tree consumers, so this is harmless in practice — just worth stating accurately rather than claiming no change.

lib/mocker/src/common/sequence.rs
```diff
@@ -107,40 +166,77 @@ impl ActiveSequence {
...
-        } else {
-            None
+        let parent_hash = self
+            .unique_blocks
+            .iter()
+            .rev()
+            .find_map(|block| match block {
+                UniqueBlock::FullBlock(hash) => Some(*hash),
+                UniqueBlock::PartialBlock(_) => None,
+            });
```
jthomson04, line 249: **This silently drops an invariant guard.**

The old code was `pop(); last()` with `PartialBlock(_) => panic!("partial block cannot be a parent")`. The new `.rev().find_map()` *skips* partial blocks instead.

Today at most one partial exists and it's at the tail, so the result is identical — I verified that. But if that invariant ever breaks, the old code failed loudly while this silently picks an older full block and emits a **wrong lineage chain** into the router's KV index.

Restore the guard while keeping the pre-pop ordering:

```rust
let parent_hash = self.unique_blocks[..self.unique_blocks.len() - 1]
    .last()
    .map(|block| match block {
        UniqueBlock::FullBlock(hash) => *hash,
        UniqueBlock::PartialBlock(_) => panic!("partial block cannot be a parent"),
    });
```

`len() >= 1` is guaranteed here — the `let ... else` on `last()` above already matched.

lib/mocker/src/common/sequence.rs
```diff
@@ -107,40 +166,77 @@ impl ActiveSequence {
...
+                };
+                let promote_token_ids = if self.emit_token_ids {
+                    Some(last_complete.tokens().to_vec())
+                } else {
+                    None
+                };
+                (last_seq_hash, last_block_hash, last_plh, promote_token_ids)
+            }
+            SequenceTokens::Flat(tokens) => {
+                let start = position * self.block_size;
```
jthomson04: **Worth pinning this invariant with a cheap assert.**

The two arms derive the promote target differently: legacy from the *token sequence* (`last_complete_block()`), flat from *metadata* (`plhs.len()`). They agree only while `plhs.len() == committed_blocks`.

A `pop()` that crossed back over an already-promoted block would desync them, and the arms would then promote **different blocks** — legacy would silently regress a block, flat would read past the end and hit the `unwrap_or_else(panic!)` below with a message describing a different failure than the actual one. Not reachable under `pop()`'s documented one-token-rollback contract, but it's the kind of thing worth making loud:

```rust
debug_assert_eq!(position, self.len() / self.block_size - 1);
```

Holds at both entry points (`len = N*bs` and `len = N*bs + 1` both give `len/bs - 1 == N-1`).

jthomson04 (commented):
## Review summary

Solid, well-benchmarked change. The `SequenceTokens` enum is cleanly encapsulated behind `len()`/`push()`/`pop()`/`block_token_ids_in()`, only one construction site changed, and the `promote_last_partial` refactor keeps both arms driven by the same `unique_blocks` partial-tail guard so they stay in lockstep. The `signal_shape`/`assert_sequence_parity` harness is genuinely strong — it compares concrete hashes, PLHs and token-ID payloads rather than just structure.

I traced these equivalences rather than trusting the parity test, and they hold:

- `position = plhs.len()` always equals `blocks.len() - 1` at both promote entry points (`promote_computed_tail` at `len ≡ 0 mod bs`, `push` at `len ≡ 1 mod bs`).
- `block_token_ids()` switching from `tokens.blocks().len()` to `len() / block_size` is safe — `TokenBlockSequence::extend` commits eagerly on every boundary (`lib/tokens/src/lib.rs:1298-1304`; the "AND there are still more tokens" comment there is stale, the code commits unconditionally), so `blocks.len() == total_tokens / block_size` always holds.
- `chunks_exact().skip(start).take(end-start)` is index-equivalent to `blocks()[start..end]`, and `hash_end <= block_hashes.len() <= len()/block_size` in both arms, so the slice can't be short.
- The flat hashing faithfully replicates `TokenBlockChunk::new` + `TokenBlock::from_chunk` for the non-MM path, and the legacy salt really is `1337`.

### What I'd want before merge

1. **`reserve_exact(max_output_tokens)` on the flat vec** — likely free perf and lower RSS at benchmark prompt sizes (details inline).
2. **Chunked `prepare_allocation` parity test** — `block_token_ids_in` with `start > 0` on the flat arm is production-reachable via `core.rs:1847` but has zero coverage.
3. **Single shared salt constant**, correctly named.
4. **Restore the partial-block-parent panic** and add a `debug_assert_eq!` on the flat block index.

Items 3 and 4 are cheap and guard against exactly the failure mode this change makes possible: two hashing implementations silently drifting apart.

### Notes

- This PR is stacked — its base `e7d8764` is not in `main` and `G1Backend` doesn't exist there yet. Review assumes the parent lands first.
- No security concerns: no new unsafe, no I/O, no deserialization. `bytemuck::cast_slice` `&[u32] -> &[u8]` is infallible (alignment only ever relaxes), and hash values are internal cache keys, not authentication material.

lib/mocker/src/common/sequence.rs
```diff
@@ -205,16 +301,50 @@ impl ActiveSequence {
...
+    ) -> Self {
+        let num_input_tokens = tokens.len();
+        let (unique_blocks, block_hashes, plhs) =
+            create_flat_sequence_cache(&tokens, block_size, enable_prefix_caching);
+
+        let seq = Self {
+            unique_blocks,
+            block_hashes,
+            plhs,
+            tokens: SequenceTokens::Flat(tokens),
```
jthomson04, line 422: **The flat vec has no reserved decode headroom — this is a large, avoidable realloc.**

`request.tokens` is moved in with capacity ≈ prompt length, so the first decode token triggers a doubling.

At this PR's own benchmark scale — 479,843,008 input tokens / 3,211 requests ≈ **149,437 tokens/request** — that's a 598 KB → 1.20 MB reallocation plus a 598 KB memcpy, with both blocks live during the copy. Meanwhile mean output is only 3,552,472 / 3,211 ≈ **1,106 tokens**, so:

```rust
let num_input_tokens = tokens.len();
let mut tokens = tokens;
tokens.reserve_exact(max_output_tokens);
```

costs ~4.4 KB and eliminates *every* reallocation for the sequence's lifetime (`generate_token`'s assert bounds pushes at `max_output_tokens`).

This likely explains both the reported 1.197% "flat token-vector growth" CPU share and a good chunk of the 8,804 KiB peak-RSS gap vs. full-prehash.

Caveat worth measuring: with short prompts and a large `max_output_tokens` the tradeoff inverts, so re-run the 16-worker arm before committing to it.

lib/mocker/src/common/sequence.rs
```diff
@@ -282,12 +407,23 @@ impl ActiveSequence {
...
+            SequenceTokens::Legacy(tokens) => tokens.blocks()[start..end]
+                .iter()
+                .map(|block| block.tokens().to_vec())
+                .collect(),
+            SequenceTokens::Flat(tokens) => tokens
+                .chunks_exact(self.block_size)
+                .skip(start)
+                .take(end - start)
+                .map(<[u32]>::to_vec)
+                .collect(),
```
jthomson04: **`start > 0` on this arm is production-reachable but has zero test coverage.**

Every test reaches `block_token_ids_in` with `start == 0`:

- `assert_sequence_parity` → `block_token_ids()` → `block_token_ids_in(0, ..)`
- `take_creation_signal()` → `prepare_allocation(len)` with `num_allocated_tokens == 0` → `hash_start = 0`

But `core.rs:1847` calls `prepare_allocation(allocation_target)` inside the chunked-prefill loop, so with `enable_chunked_prefill` + `zmq_kv_events_port` set — a real config — `prev_blocks > 0` and this `chunks_exact().skip(start)` arithmetic runs unverified. That's precisely the "native ZMQ store token-ID payload parity" the PR body claims to have validated, and I don't see a test for it in the diff.

(Perf note: `Skip` on `ChunksExact` does specialize through `nth`, so the `skip` itself is O(1) — the concern here is coverage, not cost.)

lib/mocker/src/common/sequence.rs
```diff
@@ -775,4 +932,212 @@ mod tests {
...
+                .collect::<Vec<_>>()
+        );
+        assert_eq!(legacy.generated_tokens(), flat.generated_tokens());
+        assert_eq!(legacy.num_input_tokens(), flat.num_input_tokens());
+        assert_eq!(legacy.num_allocated_tokens(), flat.num_allocated_tokens());
+        assert_eq!(legacy.block_token_ids(), flat.block_token_ids());
+    }
+
+    #[test]
+    fn flat_sequence_matches_legacy_across_prompt_and_output_boundaries() {
```
jthomson04, line 1220: Two gaps worth closing in this harness:

1. **Chunked allocation is never exercised.** Suggest walking `prepare_allocation`/`commit_allocation` in `block_size`-sized chunks and asserting `signal_shape` parity at each step — that covers the `block_token_ids_in(start > 0)` path flagged above.

2. **`sequence_pair` hardcodes `enable_prefix_caching = true`**, so legacy-vs-flat parity with caching *off* is never compared. Values are random by design there, but the structural parity (block counts, `unique_blocks` shape, promote/deref signal sequencing) is still comparable and worth asserting.

Separately: `flat_tokens_cover_every_native_g1_configuration` only asserts *which* representation is selected — it never drives the scheduler forward. Fine as a selection test, just not behavioral coverage.

lib/mocker/src/common/sequence.rs
jthomson04, line 422: ### Fix 1 — exact-size the token vector (biggest single memory lever)

Expanding on the CPU-side comment I left on this line: `reserve_exact` is actually a bigger **memory** win than a CPU one.

`request.tokens` arrives with capacity ≈ `len` and no headroom, so the first decode token doubles it. At benchmark scale:

- needed: 149,437 + 1,106 = 150,543 tokens = **602 KB**
- actual after doubling: 298,874 capacity = **1,195 KB**
- **593 KB per request of capacity that is never touched**

On top of that, the doubling copy briefly holds the old 598 KB buffer and the new 1,195 KB buffer simultaneously — a ~1.8 MB transient per request, which is exactly the kind of thing that sets peak RSS rather than average.

```rust
let num_input_tokens = tokens.len();
let mut tokens = tokens;
tokens.reserve_exact(max_output_tokens);
```

Costs 4.4 KB, eliminates both the slack and the transient, and guarantees zero reallocations for the sequence's lifetime (`generate_token`'s assert bounds pushes at `max_output_tokens`).

Worth measuring the inverse case before committing: with short prompts and a large `max_output_tokens` the tradeoff flips, so re-run the 16-worker arm.

lib/mocker/src/common/sequence.rs
```diff
@@ -44,6 +56,52 @@ fn create_sequence_cache(
...
+    block_size: usize,
+    enable_prefix_caching: bool,
+) -> (Vec<UniqueBlock>, Vec<BlockHash>, Vec<PositionalLineageHash>) {
+    let full_blocks = tokens.len() / block_size;
+    let mut unique_blocks = Vec::with_capacity(full_blocks + 1);
+    let mut block_hashes = Vec::new();
+    if enable_prefix_caching {
+        block_hashes.reserve(full_blocks);
+    }
+    let mut plhs = Vec::with_capacity(full_blocks);
```
jthomson04: ### Fix 2 — pre-size the metadata vectors to the completion footprint

All three vectors are sized for the **prompt** (`full_blocks + 1`), but they keep growing through decode as blocks are promoted — so each doubles once and lands at ~2× the capacity it ever uses. At `block_size = 64` and benchmark scale that's ~112 KB of untouched slack per request across the three.

`to_completion_blocks()` already computes exactly the right number: `(num_input_tokens + max_output_tokens).div_ceil(block_size)`. Cleanest fix is to pass it in rather than reserving after the fact, which avoids any `reserve(additional)` underflow arithmetic:

```rust
fn create_flat_sequence_cache(
    tokens: &[u32],
    block_size: usize,
    enable_prefix_caching: bool,
    completion_blocks: usize,   // (num_input + max_output).div_ceil(block_size)
) -> (Vec<UniqueBlock>, Vec<BlockHash>, Vec<PositionalLineageHash>) {
    let full_blocks = tokens.len() / block_size;
    let mut unique_blocks = Vec::with_capacity(completion_blocks + 1);
    let mut block_hashes = Vec::new();
    if enable_prefix_caching {
        block_hashes.reserve(completion_blocks);
    }
    let mut plhs = Vec::with_capacity(completion_blocks);
```

Note `create_sequence_cache` above has the identical pattern, so the same change there recovers the same ~112 KB/request on the legacy and KVBM paths — i.e. this is a win for `main`, not just for native G1. If you'd rather keep this PR scoped to the flat path, splitting the legacy half into its own commit would make that easy to land independently.

Unrelated and definitely out of scope, but worth filing while it's visible: `UniqueBlock` is 24 B to carry an 8 B payload, because the `PartialBlock(Uuid)` variant widens every element — and there is at most one partial block per sequence. Hoisting that UUID out-of-line would make this a `Vec<u64>` and save another ~37 KB/request on both arms.

lib/mocker/src/common/sequence.rs
```diff
@@ -2,12 +2,24 @@
...
+};
 use rand::random;
 use validator::Validate;
 
+const NATIVE_HASH_SEED: u64 = 1337;
+
+#[derive(Debug)]
+enum SequenceTokens {
+    Legacy(TokenBlockSequence),
+    Flat(Vec<u32>),
```
jthomson04: ### Fix 3 — don't retain the prompt when `emit_token_ids == false` (follow-up PR)

Larger than fixes 1 and 2, but it closes the gap *completely* rather than halving it, so it's worth writing down while the design is fresh.

The flat token vector is read in exactly three places:

1. hashing at construction — consumed immediately, then dead
2. `compute_block_hash(cast_slice(complete))` in `promote_last_partial` — needs only the ≤`block_size` tokens of the in-flight tail block
3. `block_token_ids_in` — the only reader that needs arbitrary history

And (3) only runs when `emit_token_ids`, i.e. `zmq_kv_events_port.is_some() && enable_prefix_caching`. **So with ZMQ KV events off, the entire 602 KB prompt is dead weight the moment the constructor returns.**

Sketch, composing with fix 1's split of prompt from decode tail:

```rust
enum SequenceTokens {
    Legacy(TokenBlockSequence),
    Flat { prompt: Option<Vec<u32>>, decoded: Vec<u32>, decoded_base: usize },
}
```

- `prompt: None` when `!emit_token_ids`; seed `decoded` with the trailing `prompt.len() % block_size` tokens so the straddling block can still be completed and hashed, and set `decoded_base` to `prompt.len() - remainder` for indexing.
- After each promote, drain `decoded` back to the current partial block — its tokens are never read again. That bounds token storage at **O(block_size), ~260 bytes per request, independent of both prompt and output length.**
- `len()` doesn't need `prompt` at all: `num_input_tokens + decoded.len()`.
- `block_token_ids_in` asserts `prompt.is_some()`.
- `pop()`'s one-token-rollback contract guarantees `decoded` is non-empty — worth a `debug_assert`.

**One precondition to verify before relying on this.** `block_token_ids()` has a production caller that is *not* gated on `emit_token_ids`: `core.rs:1193` in `complete_swap_in`. It's behind `#[cfg(feature = "kvbm-offload")] `and reached through KVBM swap-in coordination, so under `G1Backend::Kvbm` it should always take the `Legacy` arm — but I haven't traced `AwaitingSwapIn` construction to confirm a Native-G1 request can never reach it. If that holds, an `expect("flat sequences retain prompt tokens only when emit_token_ids")` in the flat arm of `block_token_ids_in` turns the assumption into an enforced invariant instead of a latent panic-or-wrong-answer.

Net effect if it works out: the flat path matches full-prehash on memory for the non-ZMQ case while keeping the single unified code path — i.e. the flat design would no longer be trading RSS for maintainability, it'd just win on both.

jthomson04 (commented):
## Follow-up: alleviating the peak-RSS gap

Framing first, so the win in this PR doesn't get lost: **against `main` this change *reduces* peak RSS** (237,744 → 233,156 KiB, −1.93%). The +8,804 KiB is only relative to the tail-only full-prehash prototype. The three fixes below close that gap without giving back the reduction.

### Where the bytes go

At this PR's own benchmark scale — 479,843,008 / 3,211 ≈ **149,437 prompt tokens** and 3,552,472 / 3,211 ≈ **1,106 output tokens** per request:

| Allocation | Live need | Actual after growth | Slack |
| --- | ---: | ---: | ---: |
| Flat token `Vec<u32>` | 602 KB | **1,195 KB** | **593 KB** |
| `unique_blocks` (24 B/blk) | 56 KB | 112 KB | 56 KB |
| `plhs` (16 B/blk) | 38 KB | 75 KB | 37 KB |
| `block_hashes` (8 B/blk) | 19 KB | 37 KB | 19 KB |

(metadata rows at `block_size = 64`; they scale inversely with block size)

Every one of these vectors is sized for the **prompt only** and then grows geometrically during decode, so each ends up at roughly 2× the capacity it ever uses. **About half of what the flat path holds is untouched slack.** At ~1,195 KB per in-flight sequence the 8,804 KiB gap is on the order of 7–8 concurrent requests at peak, which is consistent with a 16-worker run.

### Three inline comments below

1. **Exact-size the token vector** — few lines, ~half the gap.
2. **Pre-size the metadata vectors** — few lines, and a net win for `main`/KVBM too since `create_sequence_cache` has the identical pattern.
3. **Drop the prompt when `emit_token_ids == false`** — larger refactor, but closes the gap entirely and makes token storage O(block_size) regardless of prompt length. This recovers the full-prehash memory profile *through* the unified flat path, which was the whole argument for preferring flat over full-prehash.

1 and 2 are safe to land here. 3 is probably its own PR.

lib/mocker/src/common/sequence.rs
jthomson04: Addressed in `8edeb9b33a`: both legacy `into_sequence` construction and flat hashing now use the single `MOCKER_SALT_HASH: SaltHash`. The legacy/flat parity tests compare the resulting block and sequence hashes.

lib/mocker/src/common/sequence.rs
jthomson04: Added the canonical `dynamo_tokens::compute_block_hash_for_tokens` helper and use it from both `TokenBlockChunk` and mocker. The direct mocker `bytemuck` dependency and its lockfile edges are gone; the focused helper test verifies canonical byte-encoding parity, and the API docs keep multimodal hashing routed through `compute_block_bytes_with_mm`.

lib/mocker/src/common/sequence.rs
jthomson04: Corrected the PR body: the public API changes are the new canonical token-hash helper and removal of the unused `ActiveSequence::tokens()` getter. CLI, event, and serialization formats remain unchanged.

lib/mocker/src/common/sequence.rs
jthomson04: Restored the immediate-parent guard: promotion now inspects the element before the tail and panics on a partial parent. `flat_promotion_rejects_partial_parent` pins the invariant.

lib/mocker/src/common/sequence.rs
jthomson04: Added `debug_assert_eq!(position, self.len() / self.block_size - 1)` at promotion, with the legacy/flat boundary and multi-block promotion tests covering both valid entry shapes.

lib/mocker/src/common/sequence.rs
jthomson04: Addressed for both storage modes: event-emitting flat sequences call `reserve_exact(max_output_tokens)`, while non-emitting sequences use a `block_size + 1` retained window. The capacity test proves no token-vector growth through maximum decode; the 128x128 attribution fell from 1.197% to 0.000%.

lib/mocker/src/common/sequence.rs
jthomson04: Added `flat_sequence_matches_chunked_token_id_allocations`, which walks `prepare_allocation`/`commit_allocation` through 4, 8, and 12 tokens and checks the exact payload for each nonzero start offset.

lib/mocker/src/common/sequence.rs
jthomson04: Closed both gaps: the chunked allocation test exercises `start > 0`, and `flat_sequence_matches_uncached_legacy_structure` compares caching-off block shape plus use/promote/deref sequencing. The path-selection test remains separate and intentionally minimal.

lib/mocker/src/common/sequence.rs
jthomson04: Implemented exact reservation for event-retaining sequences and the smaller bounded tail window for non-emitting sequences. The clean 16x16 pair improved replay time from 6.114684 s to 5.910230 s and `mprof` peak RSS from 232.484 MiB to 214.367 MiB.

lib/mocker/src/common/sequence.rs
jthomson04: Native `unique_blocks`, `block_hashes`, and lineage vectors are now pre-sized to the exact completion footprint, and the decode-capacity test proves they never grow. Per the agreed scope, legacy/KVBM capacity behavior and `UniqueBlock` layout are unchanged; candidate metadata-growth attribution is 0.000%.

lib/mocker/src/common/sequence.rs
jthomson04: Implemented the bounded retained window in this PR: non-emitting native sequences keep at most `block_size + 1` tokens and discard each promoted block, while event-emitting sequences retain full history. Token materialization now rejects discarded-history state explicitly. I also traced and enforced the swap-in precondition: native G1 reports no offload engine, rejects attachment, `complete_swap_in` asserts KVBM, and the feature-gated test covers it. The new 214.367 MiB peak is 2.16% below the historical full-prehash arm.

lib/mocker/src/common/sequence.rs
```diff
@@ -107,43 +235,90 @@ impl ActiveSequence {
...
-        } else {
-            None
+        let parent_hash = self.unique_blocks[..self.unique_blocks.len() - 1]
+            .last()
+            .map(|block| match block {
+                UniqueBlock::FullBlock(hash) => *hash,
+                UniqueBlock::PartialBlock(_) => panic!("partial block cannot be a parent"),
+            });
+        let position = self.plhs.len();
+        debug_assert_eq!(position, self.len() / self.block_size - 1);
```
jthomson04: ### Blocking: this assert can underflow-panic on its own arithmetic

I suggested this line last round and got the form wrong — sorry.

It sits on the shared path, so it applies to `Legacy` too, and when `len < block_size` it evaluates `0 - 1` and panics in debug builds *on the assert itself* rather than on the condition it is checking.

I traced it as currently unreachable: it needs a `PartialBlock` tail at `len <= block_size`, which only arises at `len == 1` with an empty prompt — and there `unique_blocks.last()` is `None`, so the `let ... else` above has already returned. But a debug assert that can panic on its own arithmetic is worse than no assert, and the underflow-free form is free and equivalent:

```rust
debug_assert_eq!(position + 1, self.len() / self.block_size);
```

lib/mocker/src/common/sequence.rs
```diff
@@ -4,10 +4,90 @@
...
+            return Self {
+                retained: tokens,
+                retained_start: 0,
+            };
+        }
+
+        let retained_start = tokens.len() - (tokens.len() % block_size);
+        let mut retained = Vec::with_capacity(
+            block_size
+                .checked_add(1)
```
jthomson04, line 42: The `+ 1` here is load-bearing and, as far as I can tell, unexercised — see my comment on `flat_sequence_matches_legacy_across_prompt_and_output_boundaries`.

It exists for the case where `promote_last_partial` fires from `push` at `len % block_size == 1`: `retained` reaches exactly `block_size + 1` before `discard_through` drains it back to 1. Every current test promotes via `promote_computed_tail` instead, so `retained` never exceeds `block_size` and `with_capacity(block_size)` would pass the whole suite while silently reintroducing a realloc per block.

Worth a short comment on the line recording *why* it is `+ 1`, independent of whether you add the test.

lib/mocker/src/common/sequence.rs
```diff
@@ -775,4 +1068,442 @@ mod tests {
...
+            };
+            assert!(
+                tokens.retained.len() <= flat.block_size() + 1,
+                "non-emitting flat storage exceeded one block plus one token"
+            );
+        }
+    }
+
+    #[test]
+    fn flat_sequence_matches_legacy_across_prompt_and_output_boundaries() {
```
jthomson04, line 1220: ### Coverage gap: push-driven promotion with discard is never exercised

Every flat test calls `promote_computed_tail` at each block boundary, so `push`'s call to `promote_last_partial` always early-returns on a `Full` tail:

- **this test** — promotes at every boundary (the `generated_tokens() < max_output_tokens()` guard only skips the final token)
- `flat_sequence_capacities_do_not_grow_during_decode` — same; its `retained.len() <= BLOCK_SIZE + 1` assertion therefore never observes the `+ 1`
- `flat_sequence_matches_legacy_reset_and_one_token_rollback` — prompt is an exact multiple of `BLOCK_SIZE`, so the tail is already `Full` at the first push
- `flat_sequence_preserves_uncached_random_identity_behavior` — uses `promote_computed_tail`

So the path where `retained` hits its ceiling and `discard_through` leaves exactly one token is untested, and `Vec::with_capacity(block_size)` (no `+ 1`) would pass everything.

This is a real production path — the `promote_last_partial` doc says as much: *"The scheduler may already have promoted this block at its computed boundary. Retain this fallback for callers that have not."*

Cheapest fix is a third axis on this test's existing loop — something like `for promote_eagerly in [true, false]`, skipping the `promote_computed_tail` call when `false` and letting `push` drive promotion. That keeps the legacy-parity assertions doing the verification work rather than hand-writing expected values.

lib/mocker/src/common/sequence.rs
```diff
@@ -107,43 +235,90 @@ impl ActiveSequence {
...
+                    None
+                };
+                (last_seq_hash, last_block_hash, last_plh, promote_token_ids)
+            }
+            SequenceTokens::Flat(tokens) => {
+                let complete = tokens
+                    .complete_block(position, self.block_size)
+                    .unwrap_or_else(|| {
+                    panic!(
+                        "partial flat sequence tail cannot be promoted without a complete token block"
```
jthomson04, line 286: This message now covers two quite different failures.

`complete_block`'s `checked_sub(self.retained_start)` returns `None` both when the block is genuinely incomplete **and** when it was already discarded (`start < retained_start`). With the sliding window in place the second is the more likely cause in practice, but the message only describes the first — so the panic would point a debugger at the wrong invariant.

Either split the check:

```rust
debug_assert!(
    position * self.block_size >= tokens.retained_start,
    "promoted block precedes retained flat-token window"
);
```

or have `complete_block` return a two-variant error so the two cases can carry their own messages.

lib/mocker/src/common/sequence.rs
```diff
@@ -451,26 +698,72 @@ impl ActiveSequence {
...
+        match &mut self.tokens {
+            SequenceTokens::Legacy(tokens) => {
+                tokens.pop();
+            }
+            SequenceTokens::Flat(tokens) => {
+                debug_assert!(
+                    !tokens.retained.is_empty(),
+                    "flat rollback token must be retained"
+                );
+                tokens.pop().expect("flat rollback token must be retained");
```
jthomson04, line 730: Two small things about `pop()`:

**1. This is a release-mode panic where the legacy arm silently no-ops.** `TokenBlockSequence::pop()` returns `Option` and the legacy arm discards it; this `expect` aborts. Flat's behaviour is the better one — it converts silent corruption into a loud failure — but it does mean a rollback pattern that "worked" (incorrectly) on KVBM now hard-fails on native G1. Worth a line in the `pop()` doc comment noting the arms deliberately differ, so it does not read as an oversight later.

For what it's worth I traced the double-pop case and it is safe: after a promote the discard leaves the post-boundary tokens in `retained`, so `test_promote_uses_previous_full_block_as_parent`-style usage does not trip it.

**2.** The `debug_assert!` at line 703 says `"flat rollback requires a freshly generated token"` but sits above the `match`, so it fires for `Legacy` sequences too. Trivial, but it will misdirect whoever hits it.

lib/mocker/src/common/sequence.rs
```diff
@@ -282,12 +497,37 @@ impl ActiveSequence {
...
+    }
+
+    /// Materialize every complete block's token IDs.
+    ///
+    /// # Panics
+    ///
+    /// Panics for native flat sequences that were created without token-ID
+    /// event emission, because those sequences intentionally discard completed
+    /// prompt and decode blocks.
     pub fn block_token_ids(&self) -> Vec<Vec<u32>> {
```
jthomson04, line 541: Good call documenting the panic. One thing worth surfacing outside rustdoc: `ActiveSequence` is `pub`, so this is a deliberate change to a public method's contract — it now panics for a whole class of sequences rather than returning data.

I verified the only production caller is `core.rs:1193` in `complete_swap_in`, which is `#[cfg(feature = "kvbm-offload")]` and unreachable for native G1 (`try_park_for_swap_in` early-returns on `!has_offload_engine()`, and it is the only site constructing `AwaitingSwapIn`). So there is no live risk — I would just mention it in the PR body, since the previous revision's "Public APIs are unchanged" claim is now doubly inaccurate (this, plus the dropped `tokens()` getter).

The layering here is right, incidentally: `debug_assert_eq!` on `g1_backend` in `complete_swap_in` catches it early in debug, and this `assert!` is a real release-mode backstop.

lib/mocker/src/kv_manager/g1_manager.rs
```diff
@@ -488,6 +488,9 @@ impl G1Manager {
     }
 
     #[cfg(feature = "kvbm-offload")]
+    /// Returns whether the legacy KVBM backend has a lower-tier offload
+    /// engine. Native G1 always returns `false`, so native requests cannot
+    /// enter the KVBM swap-in lifecycle.
```
jthomson04, line 492: Nit: the doc comment sits *after* `#[cfg(feature = "kvbm-offload")]`. That is legal and rustdoc handles it, but convention is doc comment first, then attributes — and rustfmt will not reorder it for you.

Content-wise the comment is accurate and load-bearing: I traced `try_park_for_swap_in` (`core.rs:1258`) early-returning on `!has_offload_engine()`, and it is the only site constructing `AwaitingSwapIn`, so "native requests cannot enter the KVBM swap-in lifecycle" holds. Worth keeping.

jthomson04 (commented):
## Re-review of `8ac610d`

All four items from my previous round landed, plus more than I asked for — `bytemuck` is gone from the mocker in favour of `dynamo_tokens::compute_block_hash_for_tokens`, and `FlatTokens` implements the full sliding-window design (`retained` + `retained_start` + `discard_through`), not just the two cheap fixes. The `debug_assert_flat_token_invariants` helper is the right way to keep the two arms honest.

**This is static analysis only — I did not run the test suite.**

### Traced and believed correct

- **`retained.len() <= block_size + 1` holds** across every promote/push/pop interleaving I could construct, including promote→push→pop at a boundary and the early-return case where the tail is already `Full`.
- **`Vec::with_capacity(completion_blocks)` (no `+1`) is exactly right.** `div_ceil(n+m, bs)` covers full blocks plus a trailing partial in both the exact-multiple and remainder cases, and mid-flight `unique_blocks.len() == div_ceil(len, bs) <= div_ceil(n+m, bs)`.
- **`discard_through` arithmetic cannot fail:** `promoted_end = (position+1)*bs == (len/bs)*bs <= len`, so `local_end <= retained.len()` always.
- **`retain_history` correctly uses the ANDed `emit_token_ids`.** Shadowing it before both `FlatTokens::new` and the struct field is what makes this safe — the dangerous direction (`retain_history == false` while the field is `true`) is unreachable.
- **Preemption is correctly coupled.** `reset_with_signal` → re-admission replays `Use` from block 0 including `token_ids`, but discard only runs when `emit_token_ids == false`, in which case `token_ids` is `None`.
- **The swap-in gate is real.** `try_park_for_swap_in` early-returns on `!has_offload_engine()` (`core.rs:1258`) and is the only site constructing `AwaitingSwapIn`, so `complete_swap_in` → `block_token_ids()` is unreachable for native G1. The new doc comment is load-bearing and accurate.
- **The new `debug_assert!(generated_tokens > 0)` in `pop()` does not break existing legacy tests** — I checked all five pre-existing `pop()` call sites (lines 502, 517, 518, 525, 664 on `main`), including the double-pop in `test_promote_uses_previous_full_block_as_parent`; `generated_tokens` is ≥1 at each.
- `compute_block_bytes_with_mm` is `pub`, so the new intra-doc link resolves. The `"legacy kvbm-offload cannot be attached to native G1"` message the new test expects pre-exists at `g1_manager.rs:485`.

### Test coverage

Substantially stronger. Everything I asked for landed: chunked `prepare_allocation` parity with explicit token-ID content assertions, prefix-caching-off structural parity, `should_panic` coverage for both new asserts, native-vs-KVBM separation at both construction and the swap-in boundary, and a `dynamo-tokens` helper test pinned to the existing `HASH_1_4` constant.

`flat_sequence_capacities_do_not_grow_during_decode` is a nice touch — a regression guard on the perf claim itself.

The most valuable property: the `emit_token_ids = false` arm of the boundaries test runs with prefix caching **on**, so block hashes and PLHs are compared against legacy *while the discard path is active*. That's exactly the risky new interaction. One hole remains, flagged inline.

### Verdict

Blocking: **the `debug_assert_eq!` at line 245 can panic on its own arithmetic** — my suggestion, my correction.

Strongly recommended: **the `block_size + 1` capacity is load-bearing and unexercised** by any current test.

Everything else is diagnostics and docs.

lib/mocker/src/kv_manager/g1_manager.rs
jthomson04, line 492: Nit: the doc comment sits *after* `#[cfg(feature = "kvbm-offload")]`. That is legal and rustdoc handles it, but convention is doc comment first, then attributes — and rustfmt will not reorder it for you.

Content-wise the comment is accurate and load-bearing: I traced `try_park_for_swap_in` (`core.rs:1258`) early-returning on `!has_offload_engine()`, and it is the only site constructing `AwaitingSwapIn`, so "native requests cannot enter the KVBM swap-in lifecycle" holds. Worth keeping.

lib/mocker/src/common/sequence.rs
jthomson04: Added a message-bearing debug assertion in `FlatTokens::complete_block` before retained-window subtraction, distinguishing an already-discarded block from an incomplete block.

lib/mocker/src/common/sequence.rs
```diff
@@ -205,16 +386,61 @@ impl ActiveSequence {
...
+        tokens: Vec<u32>,
+        max_output_tokens: usize,
+        block_size: usize,
+        enable_prefix_caching: bool,
+        emit_token_ids: bool,
+        planned_output_ids: Option<Vec<u32>>,
+    ) -> Self {
+        let num_input_tokens = tokens.len();
+        let emit_token_ids = emit_token_ids && enable_prefix_caching;
+        let completion_blocks = num_input_tokens
```
PeaBrane, line 406: **AI-generated review comment — please have a human evaluate and address this concern.**

This appears to be a resource-allocation regression. The flat path eagerly sizes all three metadata vectors for `num_input_tokens + max_output_tokens`, and the token-ID event path also reserves the entire declared output token capacity. For vLLM, that declared maximum is not clamped here, while actual generation is separately limited by `max_model_len`; construction also happens before waiting-admission rejection. A request that can produce only a few tokens—or will be rejected—can therefore reserve for an arbitrarily large declared completion, unlike the legacy path that grew with realized decode.

Please use a separate capacity hint bounded by the realizable output ceiling (for example, `min(max_output_tokens, max_model_len - prompt_len)`) while preserving the logical requested maximum.

PeaBrane (approved):
Approved.

**AI-generated review note:** Please have a human reviewer evaluate the inline concern and address it before merge.

PR #12253, opened 2026-07-28
refactor(mocker): migrate online replay to LiveEngine [DIS-2478] [DYN-3852]

PeaBrane:
## Summary

- migrate online replay from its replay-specific demux to request-scoped `LiveEngine` handles and a production `OnlineTraceRecorder`
- make replay orchestration fail fast: return the first internal error, surface task panics as `JoinError`, propagate router bookkeeping failures, and discard the replay-owned runtime without producing a partial report
- retain one injectable cancellation token at the replay runtime boundary and observe it from workload waits, dispatch, and request setup without using cancellation as internal-error cleanup
- preserve successful shutdown ordering: requests complete, engines shut down, admission drains, the router shuts down, and the recorder finalizes
- add opt-in per-request JSONL, online SLA/goodput metrics, fixed worker/GPU capacity accounting, and agentic trace mode without concurrency
- timestamp streamed output at the `LiveEngine` dispatcher boundary so request-channel backpressure does not inflate ITL, E2E, or SLA observations

Online G1/G2 offload remains supported and is exercised through the public replay path. Online G3/G4 now fail validation with a clear error before runtime construction: the equivalent multi-worker panic reproduces on exact `origin/main`, and this PR does not change shared KVBM lifecycle behavior to approximate a fix.

## Non-goals

- no shared `LiveEngine` lifecycle simplification
- no Ctrl+C or Python-signal integration
- no Mooncake-delta online mode; correct session accumulation needs generated-token feedback and is not a mechanical constructor change
- no online G3/G4 lifecycle fix

## Validation

- `cargo test -p dynamo-mocker --no-default-features` — 649 passed
- `cargo test -p dynamo-mocker --no-default-features --features kvbm-offload` — 747 passed, 1 ignored doctest
- rebuilt the PyO3 extension with the project `.venv`, then ran replay smoke and CLI tests — 77 passed, 8 expected online-disaggregated skips
- repeated final online replay 20 times — 20/20 passed
- SC-01 four-worker G1/G2 soak: 1,000/1,000 requests, exact 9,792,030 input and 199,084 output tokens, concurrency 32, 17 preemptions, complete per-request records, finite ordered timestamps, worker assignments 0–3, and nonzero goodput/capacity metrics
- exact-source SC-01 matched performance, three warmups plus ten alternating pairs:
  - one worker/concurrency 8: 699.430 ms baseline, 775.659 ms candidate, 1.1086 paired-median ratio
  - four workers/concurrency 32: 280.544 ms baseline, 306.368 ms candidate, 1.0942 paired-median ratio

The remaining profiled overhead is in the shared `LiveEngine` per-output routing/channel path; the recorder is below 1%. Further reduction would require shared output-delivery changes or restoring a replay-specific demux, so it is outside this replay-orchestration change.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12253" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Online and synthetic replay now support Goodput SLA thresholds and include Goodput results in reports.
  * Online replay can export per-request JSONL records, including timing, capacity, and performance metrics.
  * Added support for agentic trace replay in online mode.
  * Replay workflows now support improved request tracking, session metadata, and cancellation handling.

* **Bug Fixes**
  * Improved replay shutdown, request cleanup, event ordering, and error reporting.
  * Added validation for unsupported online KV offload configurations.


lib/mocker/src/replay/entrypoints.rs
```diff
@@ -214,6 +249,8 @@ pub fn simulate_loaded_trace_live_with_router_mode(
         trace,
         num_workers,
         router_mode,
+        false,
+        online_replay_options(record_per_request, sla),
```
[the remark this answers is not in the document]
PeaBrane: Fixed in `1dc601fe87`: loaded online traces now derive `emit_session_metadata` from `!trace.is_single_turn()` after normalization and validation, matching the offline contract. I also added a public-entrypoint regression that replays a two-turn session with per-request recording and verifies both `session_id` and `turn_index`; the targeted test and the full online replay test set pass.

lib/bindings/python/rust/llm/replay.rs
```diff
@@ -1230,15 +1218,31 @@ fn run_loaded_dynamo_request_trace(
...
+                    num_workers,
+                    router_mode,
+                    record_per_request,
+                    sla,
+                ),
+                other => anyhow::bail!(
+                    "replay_mode must be either 'offline' or 'online', got '{}'",
+                    other
+                ),
+            }
```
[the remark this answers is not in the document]
PeaBrane: No change in this PR: this is an existing offline limitation. At the PR base (`11c1d7cd25`), the agentic arm already called `simulate_agentic_trace_workload_with_router_mode` without a per-request flag and the JSONL writer ran afterward. This PR only adds the online agentic dispatch and threads `record_per_request` through that new online path. Extending the offline agentic API/reporting contract is outside this online-replay change and should be handled separately.

lib/mocker/src/replay/online/tests.rs
```diff
@@ -177,6 +231,107 @@ fn test_online_trace_workload_completes_multiturn_sessions() {
...
+    );
+    assert_eq!(report.goodput.unwrap().completed_requests, 3);
+    assert_eq!(report.throughput.prefill_worker_seconds, 0.0);
+    assert_eq!(report.throughput.decode_gpus_per_worker, 2);
+    let expected_worker_seconds = 2.0 * report.throughput.duration_ms / 1000.0;
+    assert!(
+        (report.throughput.decode_worker_seconds - expected_worker_seconds).abs() < f64::EPSILON
+    );
+    let expected_gpu_hours = expected_worker_seconds * 2.0 / 3600.0;
+    assert!((report.throughput.gpu_hours - expected_gpu_hours).abs() < f64::EPSILON);
```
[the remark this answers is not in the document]
PeaBrane: No change: the accumulation premise does not apply to online replay. `OnlineTraceRecorder` configures `TraceCollector` with `set_static_worker_count(0, num_workers)`, and `TraceCollector::finish` computes decode worker-seconds once as `decode_count * duration_s`; GPU-hours are then derived directly from those worker-seconds and GPUs per worker. The repeated `add_worker_seconds` path is used only when no static count is configured. The test is intentionally checking this closed-form online capacity contract, so loosening it for incremental-summation error would mask the wrong failure mode.

PeaBrane:
The three review-body nits are addressed in `1dc601fe87`: shutdown now preserves a scheduler/dispatcher failure ahead of the secondary active-route assertion; the preemption total is an explicit `snapshot` input instead of a patched zero; and the legacy output-only admission no-op is documented. This keeps LiveEngine lifecycle behavior unchanged. Validation: the new multi-turn regression, 27 online replay tests, 13 LiveEngine tests, 13 scheduler tests, Mocker Clippy, formatting, and diff checks all pass.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -546,16 +561,19 @@ impl LiveEffectsPublisher {
...
+        let Some(tx) = self.event_tx.as_ref() else {
+            return Ok(());
         };
         if !admissions.is_empty() {
             self.record(PublishedEffect::Admissions);
         }
-        for admission in admissions {
-            let _ = tx.send(admission.clone());
+        if let Err(SchedulerEventSendError::OrderedLaneClosed) =
+            tx.send_admissions(admissions.to_vec()).await
```
jthomson04: **[P2] Avoid cloning discarded admissions**

Every nonempty admission batch is copied before the sender variant is known. Existing output-only `create_engine` callers use `SchedulerEventSender::Outputs`, which immediately discards this `Vec`, adding a heap allocation and copy to the live scheduler hot path. Accept a slice and clone only in the `Ordered` branch.

lib/mocker/src/replay/online/recorder.rs
```diff
@@ -0,0 +1,197 @@
...
+
+pub(super) struct OnlineTraceRecorder {
+    tx: mpsc::UnboundedSender<RecorderEvent>,
+    task: tokio::task::JoinHandle<Result<TraceSimulationReport>>,
+    capture_per_request: bool,
+}
+
+impl OnlineTraceRecorder {
+    pub(super) fn start(options: OnlineRecorderOptions) -> Self {
+        let (tx, rx) = mpsc::unbounded_channel();
```
jthomson04, line 112: **[P2] Bound recorder queues**

The new recorder queue is unbounded, as is the admission-forwarding queue in `live_runtime.rs:188`. Same-timestamp trace bursts can enqueue arrivals, admissions, and terminals without backpressure and grow until OOM. Use bounded channels with cancellation-aware sends, or document a defensible hard bound.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -99,9 +100,21 @@ pub(crate) struct LiveSchedulerState {
     cancellation_tx: mpsc::Sender<SchedulerCancellationEnvelope>,
     lifecycle_rx: Arc<Mutex<Option<mpsc::Receiver<SchedulerLifecycleEvent>>>>,
     metrics_rx: watch::Receiver<MockerMetrics>,
+    actor_handle: Arc<Mutex<Option<tokio::task::JoinHandle<anyhow::Result<()>>>>>,
```
jthomson04: **[P3] Return the actor handle directly**

This handle has one owner and is immediately taken by the engine factory. `Arc<Mutex<Option<JoinHandle>>>` adds synchronization and introduces production `expect()` paths here and in `engine.rs:86/100`. Have `spawn_live_scheduler` return the state and actor handle together; `LiveEngine` can retain it and legacy constructors can explicitly detach it.

lib/mocker/src/live.rs
```diff
@@ -417,6 +497,50 @@ impl LiveEngine {
     pub fn active_request_count(&self) -> usize {
         self.inner.routes.by_client.len()
     }
+
+    pub(crate) async fn shutdown(&self) -> anyhow::Result<()> {
+        self.inner.cancel.cancel();
+        shutdown_routes(&self.inner.routes);
+        let (scheduler_actor, dispatcher_supervisor) = {
+            let mut tasks = self.inner.tasks.lock().unwrap();
```
jthomson04, line 505: **[P3] Do not panic during shutdown**

A poisoned task mutex now panics in production cleanup. Propagate an `anyhow` error or use task ownership that does not require a poisonable lock; shutdown is exactly where another panic is least useful.

lib/mocker/src/replay/online/tests.rs
```diff
@@ -612,3 +716,138 @@ fn test_online_trace_replay_kv_router_marks_prefill_and_free_once() {
...
+    assert!(
+        (1..=3).contains(&stats.vllm_preemptions_total),
+        "fixture should cross the capacity edge without preemption cycling: {}",
+        stats.vllm_preemptions_total
+    );
+}
+
+#[test]
+fn test_online_replay_four_worker_stress_cleans_every_lifecycle() {
+    let requests = (0..256)
```
jthomson04: **[P3] Keep the unit test small**

This 256-request stress case only asserts lifecycle counts. A handful of requests covering all four workers exercises the same behavior; keep the 1,000-request soak in external validation rather than the unit suite.

lib/mocker/src/replay/online/task.rs
```diff
@@ -5,31 +5,33 @@ use std::future::Future;
...
     pub(super) stats: Arc<SharedLiveRuntimeStats>,
     pub(super) workload: Option<Arc<WorkloadDispatchState>>,
+    pub(super) cancel: CancellationToken,
+    pub(super) start: Instant,
 }
 
 /// Releases a `WorkloadDriver` cap slot on drop if `mark_completed` was not called.
-/// Preserves the drop-safety of the old `OwnedSemaphorePermit` so a cancelled or
-/// panicking request task can't leak capacity.
+/// Preserves the drop-safety of the old `OwnedSemaphorePermit` so a cancelled
```
jthomson04: **[P3] Remove implementation history**

The first sentence already states the invariant. The `OwnedSemaphorePermit` comparison records implementation history rather than explaining current behavior; delete it.

jthomson04 (commented):
Static review only; no tests run.

Scope is too broad: the `LiveEngine` migration is mixed with online agentic mode, JSONL reporting, SLA/goodput, capacity accounting, and G3/G4 validation. These should be separate review units if practical.

lib/mocker/src/scheduler/live_boundary.rs
PeaBrane: Fixed in `c691c71ac3`: `send_admissions` now accepts a slice, the output-only branch returns without allocating, and only the ordered lane materializes an owned batch. The online replay, LiveEngine, and scheduler-boundary suites pass.

lib/mocker/src/replay/online/recorder.rs
PeaBrane: No change in this PR after comparing the buffering topology with the merge-base. The prior demux already used unbounded arrival, admission, and output queues; this PR replaces those demux inputs and adds a bounded scheduler-to-`LiveEngine` ordered event lane. Replay also necessarily retains O(requests) trace and collector state. Making only the recorder sends bounded would backpressure the observation path and can alter measured replay timing, so a safe change needs a replay-wide buffering/timing contract rather than a local channel swap. Resolving this as declined for this PR.

lib/mocker/src/scheduler/live_boundary.rs
PeaBrane: Fixed in `c691c71ac3`: `spawn_live_scheduler` now returns the scheduler state and actor `JoinHandle` directly. `LiveEngine` retains that handle, while legacy constructors explicitly detach it. The `Arc<Mutex<Option<JoinHandle>>>`, extraction helpers, and production `expect()` paths are gone.

lib/mocker/src/live.rs
PeaBrane: No change after tracing the lock ownership. `inner.tasks` has one lock site, and the guard spans only two `Option::take()` operations with no await or panicking call. Scheduler and dispatcher panics occur outside that critical section and surface through their `JoinHandle`s, so supported execution cannot poison this mutex. I am avoiding shared `LiveEngine` recovery machinery for an unreachable state and resolving this as declined.

lib/mocker/src/replay/online/tests.rs
PeaBrane: Fixed in `c691c71ac3`: the unit fixture now uses 16 requests instead of 256 and explicitly proves that worker IDs 0–3 were all exercised, in addition to preserving the completion, prefill-marked, and freed lifecycle assertions. The focused test passes; scale remains covered by the documented 1,000-request soak.

lib/mocker/src/replay/online/task.rs
PeaBrane: Fixed in `c691c71ac3`: removed the `OwnedSemaphorePermit` implementation-history sentence and kept only the current `InFlightGuard` drop invariant.

PeaBrane:
@jthomson04 On the review-body scope note: I am keeping the current units together. JSONL, SLA/goodput, and capacity accounting are direct consumers of the `OnlineTraceRecorder` that replaces the replay demux; the G3/G4 validation is the fail-closed boundary for a panic reproduced on `origin/main`, without changing shared KVBM lifecycle; and online agentic mode reuses the same `WorkloadDriver` path. Splitting at this validated point would not isolate those observation and entrypoint changes cleanly. I agree that future unrelated replay features should land separately.

jthomson04 (approved):

PR #4037, opened 2025-11-01
fix: qwen3 recipe yaml

sozercan:
#### Overview:

fixes the yaml syntax

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- closes GitHub issue: #xxx

## Summary by CodeRabbit

* **Chores**
  * Fixed container configuration structure to align with deployment standards.

PR #12255, opened 2026-07-28
fix(operator): gate readiness on webhook server

sttts:
## Summary

This PR closes the operator-owned admission-webhook false-ready windows exposed by the shared vCluster CI:

- Dynamo reports Ready only after controller-runtime starts its webhook server.
- Dynamo's Helm and kustomize readiness probes use a 5-second timeout because `/readyz` now performs a TLS connection.
- The bundled Grove operator enables its existing health probes, and CI waits for both operator containers before starting deploy tests.

The PR does not patch Pod status, Services, or EndpointSlices. The separate vCluster host/tenant version-skew failure remains tracked in [#12266](https://github.com/ai-dynamo/dynamo/issues/12266).

### Dynamo webhook readiness

Dynamo previously registered `healthz.Ping` for `/readyz`. Controller-runtime can start its health server before the webhook runnable, so Kubernetes could mark the Pod Ready before the TLS listener on port 9443 accepted connections.

The readiness check now uses `webhookServer.StartedChecker()`. Liveness remains an independent ping.

### Grove webhook readiness

After the Dynamo fix, the [vLLM aggregated deployment job](https://github.com/ai-dynamo/dynamo/actions/runs/30353383283/job/90257708438) exposed the same local false-ready window in bundled Grove:

```text
failed calling webhook "pcs.defaulting.webhooks.grove.io":
Post "https://grove-operator.default.svc:9443/webhooks/default-podcliqueset":
connect: connection refused
```

Grove already gates `/readyz` on certificate and webhook-server readiness, but its chart disables health probes by default. The Platform chart now enables them and checks the rendered pinned Grove Deployment for a readiness probe.

The existing host-side wait covers both operators and compares ready with total container counts instead of assuming `1/1`.

## Validation

- `go test ./cmd/...` in `deploy/operator`
- `pre-commit run --files` for every changed file
- Platform Helm chart lint: 1 chart linted, 0 failures
- Platform Helm chart tests: 7 suites and 31 tests passed
- Rendered manifests verified:
  - Grove liveness `/healthz` and readiness `/readyz` on port 9444
  - Dynamo readiness `timeoutSeconds: 5` in Helm and kustomize
  - no `publishNotReadyAddresses`
  - no Pod-status workaround

## Related

- [#12266](https://github.com/ai-dynamo/dynamo/issues/12266) — vCluster Pod-status corruption under host/tenant Kubernetes version skew
- [DYN-3452](https://linear.app/nvidia/issue/DYN-3452/expand-operator-live-cluster-e2e-coverage-in-ci-using-the-shared)

### DIFF
--- a/deploy/operator/cmd/main.go
+++ b/deploy/operator/cmd/main.go
@@ -510,7 +510,7 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("webhook-server", webhookServer.StartedChecker()); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
--- a/deploy/helm/charts/platform/components/operator/templates/deployment.yaml
+++ b/deploy/helm/charts/platform/components/operator/templates/deployment.yaml
@@ -130,6 +130,7 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
+          timeoutSeconds: 5
         resources: {{- toYaml .Values.controllerManager.manager.resources | nindent
           10 }}
         securityContext: {{- toYaml .Values.controllerManager.manager.containerSecurityContext
--- a/deploy/operator/config/manager/manager.yaml
+++ b/deploy/operator/config/manager/manager.yaml
@@ -99,6 +99,7 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
+          timeoutSeconds: 5
         resources:
           limits:
             cpu: 1024m


tmonty12 (approved):

julienmancuso (approved):

deploy/helm/charts/platform/components/operator/templates/webhook-service.yaml
```diff
@@ -25,6 +25,9 @@ metadata:
     app.kubernetes.io/part-of: dynamo-operator
   {{- include "dynamo-operator.labels" . | nindent 4 }}
 spec:
+  {{- if .Values.webhook.service.publishNotReadyAddresses }}
```
sttts: eh, no. We don't want this.

deploy/helm/charts/platform/components/operator/values.yaml
```diff
@@ -189,6 +189,11 @@ checkpoint:
 
 # Webhook configuration
 webhook:
+  service:
+    # Publish not-ready webhook endpoints when an environment cannot reliably
+    # propagate virtual Pod readiness to the Service endpoint controller.
+    publishNotReadyAddresses: false
```
sttts, line 196: We don't want this.

deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
```diff
@@ -96,6 +96,15 @@ tests:
...
           count: 1
 
+  - it: publishes not-ready webhook endpoints when explicitly enabled
+    template: charts/dynamo-operator/templates/webhook-service.yaml
+    set:
+      dynamo-operator.webhook.service.publishNotReadyAddresses: true
+    asserts:
+      - equal:
+          path: spec.publishNotReadyAddresses
+          value: true
```
sttts: We don't want this.

deploy/helm/charts/platform/values.yaml
```diff
@@ -218,6 +218,12 @@ dynamo-operator:
 
   # Webhook configuration for admission control and validation
   webhook:
+    service:
+      # -- Publish not-ready webhook endpoints. Keep disabled for regular
+      # clusters; vCluster installations may need it when Pod readiness
+      # propagation to Service endpoints is unreliable.
+      publishNotReadyAddresses: false
```
sttts, line 225: We don't want this. This hides problems.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -434,13 +434,19 @@ runs:
...
+        # endpoint, which gates on certificates and the webhook listener.
         KUBECONFIG=${{ github.workspace }}/.kubeconfig-vcluster \
         helm upgrade --install dynamo-platform . --namespace default \
           "${IMAGE_ARGS[@]}" \
           --set dynamo-operator.gpuDiscovery.enabled=false \
           "${CHECKPOINT_HELM_ARGS[@]}" \
           --set global.grove.install=true \
           --set global.kai-scheduler.install=false \
+          --set dynamo-operator.webhook.service.publishNotReadyAddresses=true \
+          --set grove.config.server.healthProbes.enable=true \
```
sttts: shouldn't this always be set? Can we default it?

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -434,13 +434,19 @@ runs:
...
+        # Grove disables health probes by default, so enable its readiness
+        # endpoint, which gates on certificates and the webhook listener.
         KUBECONFIG=${{ github.workspace }}/.kubeconfig-vcluster \
         helm upgrade --install dynamo-platform . --namespace default \
           "${IMAGE_ARGS[@]}" \
           --set dynamo-operator.gpuDiscovery.enabled=false \
           "${CHECKPOINT_HELM_ARGS[@]}" \
           --set global.grove.install=true \
           --set global.kai-scheduler.install=false \
+          --set dynamo-operator.webhook.service.publishNotReadyAddresses=true \
```
sttts: we don't want this

deploy/helm/charts/platform/components/operator/templates/webhook-service.yaml
sttts: Removed in 083e55b65a. The webhook Service template is back to normal readiness-filtered endpoints; it no longer renders publishNotReadyAddresses.

deploy/helm/charts/platform/components/operator/values.yaml
sttts: Removed in 083e55b65a. No publishNotReadyAddresses value or public API remains in the operator subchart.

deploy/helm/charts/platform/tests/namespace_restriction_webhooks_test.yaml
sttts: Removed in 083e55b65a together with the unsupported value and template behavior.

deploy/helm/charts/platform/values.yaml
sttts: Removed in 083e55b65a. Normal Kubernetes endpoint readiness filtering remains intact, so this PR no longer publishes unready endpoints or hides that failure mode.

.github/actions/setup-dynamo-operator/action.yml
sttts: Done in 083e55b65a. grove.config.server.healthProbes.enable is now true by default in the Dynamo Platform chart, and the CI-only --set was removed. The rendered Grove Deployment has /readyz on port 9444.

.github/actions/setup-dynamo-operator/action.yml
sttts: Removed in 083e55b65a. The CI-only Dynamo publishNotReadyAddresses flag and the Service patches are gone; the fix now relies on real webhook readiness.

deploy/helm/charts/platform/values.yaml
```diff
@@ -286,6 +286,10 @@ grove:
   tolerations: []
   # -- Affinity for Grove pods
   affinity: {}
+  config:
+    server:
+      healthProbes:
+        enable: true
```
sttts: for the reviewer: this is one of two actual changes

deploy/operator/cmd/main.go
```diff
@@ -510,7 +510,7 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("webhook-server", webhookServer.StartedChecker()); err != nil {
```
sttts, line 513: for the reviewer: this is one of two actual changes

deploy/helm/charts/platform/values.yaml
sttts: Confirmed. This remains one of the two functional changes: the bundled Grove dependency now enables its existing health and readiness probes by default.

deploy/operator/cmd/main.go
sttts: Confirmed. This is the other functional change: Dynamo readiness now uses webhookServer.StartedChecker() instead of an unconditional ping.

deploy/helm/charts/platform/values.yaml
dmitry-tokarev-nv: **Verified the path is right, but this can regress silently.**

I pulled `grove-charts v0.1.0-alpha.11`: the default is `config.server.healthProbes.enable: false` with `port: 9444`, so this override is correct today and the rendered Deployment does gain `/healthz` + `/readyz`.

Two things worth handling:

1. **Helm does not error on unknown subchart value paths.** If a future Grove bump renames `config.server.healthProbes`, this silently becomes a no-op: the Deployment renders with *no* `readinessProbe`, the pod reports `1/1` as soon as it is Running, and the "Wait for operator pods to be ready" loop passes with no gate at all. A silent green is the worst outcome for a CI gate. Since `grove-charts` is a pinned alpha, a cheap assertion (`helm template ... | grep -q readinessProbe`, or a helm-unittest on the rendered Grove Deployment) would pin it at bump time.
2. This is now a default for **every** bundled-Grove user, not the CI-only `--set` it replaced. Grove ships probes disabled; a line in the PR body on why flipping that default is safe for everyone would help reviewers.

Separately, every other key in the `grove:` block carries a `# --` helm-docs comment and a `README.md` row. This is the only one that changes behaviour and it has neither:

```suggestion
  config:
    server:
      healthProbes:
        # -- Enable Grove's health and readiness probes on port 9444. Grove's /readyz
        # gates on certificate readiness and its webhook listener, so PodCliqueSet
        # admission is not attempted before Grove can serve it.
        enable: true
```

(needs a `helm-docs` regen for the new README row)

deploy/operator/cmd/main.go
dmitry-tokarev-nv, line 513: This change is correct — I checked the wiring rather than assuming it. `webhookServer` is always constructed (`:247`) and assigned to `mgrOpts.WebhookServer` (`:267`), and `registerWebhookHandlers` (`:524`) unconditionally calls `mgr.GetWebhookServer()`, which is what lazily adds the server as a manager runnable. If webhook registration were conditional this check would never pass and the pod would never go Ready. It is not, so this is safe. `DefaultServer.Start` also only holds `s.mu` briefly to set `started = true`, so the checker's `s.mu.Lock()` cannot block on it.

One consequence worth a follow-up in the chart: `/readyz` now performs a real TLS dial and handshake instead of returning instantly, but `components/operator/templates/deployment.yaml:127-132` still leaves `timeoutSeconds` at the Kubernetes default of **1s**, on a container with a `cpu: 1024m` limit. Grove's own probe — which per your description gates on the same `WebhookServer.StartedChecker()` — uses 5s:

```yaml
readinessProbe:
  httpGet: {path: /readyz, port: 9444}
  initialDelaySeconds: 10
  timeoutSeconds: 5
```

Now that `publishNotReadyAddresses` is gone (good call), a readiness flap empties the endpoint set directly — which is the failure signature under investigation. Matching Grove seems worthwhile:

```yaml
        readinessProbe:
          httpGet:
            path: /readyz
            port: 8081
          initialDelaySeconds: 5
          periodSeconds: 10
          timeoutSeconds: 5
```

`deploy/operator/config/manager/manager.yaml:98` carries the same probe for kustomize users.

.github/actions/dynamo-deploy-test/action.yml
```diff
@@ -141,6 +141,48 @@ runs:
           --alluredir=test-results/allure-results \
           --log-cli-level=INFO
 
+    - name: Debug admission failure
```
dmitry-tokarev-nv: **Right instinct, but I think these diagnostics are attached to the wrong failure boundary for the scenario in the description.**

`setup-dynamo-operator`'s "Debug deployment failure" step is `if: failure()` on *that* action. In run 30350064178 the admission smoke test **succeeded** after 31 connection-refused retries — so setup passed, and the richest new diagnostics in this PR (host-side Service/Endpoints, vCluster syncer logs) would never have fired. Only this step runs, and it has just the vCluster kubeconfig: no host state, no syncer logs.

Compounding that, the condition is transient and self-healing, so a one-shot dump *after* the step fails will most likely show a fully recovered Service and prove nothing.

The place that actually holds state at the moment of refusal is the smoke-test retry loop in `setup-dynamo-operator/action.yml`, which is exactly the code path that degraded-but-passed last time. Two lines next to the existing `::warning::... retrying in 2s`:

```bash
for KC in "--kubeconfig=${VKUBECONFIG} -n default" "-n ${NAMESPACE}"; do
  kubectl ${KC} get endpointslices \
    -l 'kubernetes.io/service-name in (dynamo-platform-dynamo-operator-webhook-service,grove-operator)' \
    -o custom-columns=NAME:.metadata.name,ADDR:.endpoints[*].addresses,READY:.endpoints[*].conditions.ready \
    --no-headers 2>&1 || true
done
```

That splits the fork the entire investigation turns on: **empty endpoints** (readiness/propagation — a ClusterIP Service with no ready endpoints gets a kube-proxy `REJECT` rule, surfacing as exactly `connect: connection refused`) versus **populated endpoints, dead listener** (which is what `StartedChecker()` fixes). Neither dump currently distinguishes them, so "next time we will know" is not yet true for the case described.

Two cheap additions in the same spirit:

- Pod `.status.conditions[?(@.type=="Ready")].lastTransitionTime` on both sides — a readiness flap shows up there directly and separates a flap from pure syncer lag.
- "Zero restarts" does not rule out a *replaced* pod: `restartCount` is per-container-within-pod, so a pod UID / `creationTimestamp` comparison is worth capturing too.

.github/actions/dynamo-deploy-test/action.yml
```diff
@@ -141,6 +141,48 @@ runs:
...
+        echo "::group::Grove operator logs"
+        GROVE_POD=$(kubectl get pods -n "${NAMESPACE}" -o name 2>/dev/null \
+          | grep 'grove-operator' \
+          | head -1 || true)
+        if [ -n "${GROVE_POD}" ]; then
+          kubectl logs "${GROVE_POD}" -n "${NAMESPACE}" --tail=200 2>&1 || true
+        else
+          echo "Grove operator pod not found"
+        fi
+        echo "::endgroup::"
```
dmitry-tokarev-nv: Minor robustness on the log capture. This action runs against the vCluster, where pods keep their original labels — unlike the host side, where vCluster rewrites them and `grep`-by-name is genuinely the right call (worth a one-line comment there saying so, since the two blocks otherwise look gratuitously different).

- label selectors instead of `grep | head -1`, which also covers more than one replica
- `--previous`, which would have directly confirmed or refuted the "zero restarts" reading
- `--all-containers` for Grove, currently fetched without `-c` and so failing silently through `|| true` on a multi-container pod

```suggestion
        echo "::group::Operator manager logs"
        kubectl logs -n "${NAMESPACE}" -l control-plane=controller-manager \
          -c manager --tail=200 --prefix 2>&1 || echo "Operator pod not found"
        kubectl logs -n "${NAMESPACE}" -l control-plane=controller-manager \
          -c manager --tail=200 --prefix --previous 2>&1 || true
        echo "::endgroup::"

        echo "::group::Grove operator logs"
        kubectl logs -n "${NAMESPACE}" -l app.kubernetes.io/name=grove-operator \
          --all-containers --tail=200 --prefix 2>&1 || echo "Grove operator pod not found"
        kubectl logs -n "${NAMESPACE}" -l app.kubernetes.io/name=grove-operator \
          --all-containers --tail=200 --prefix --previous 2>&1 || true
        echo "::endgroup::"
```

Labels verified: `control-plane: controller-manager` on the operator pod template, `app.kubernetes.io/name: grove-operator` on the rendered Grove pod template.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -474,30 +474,34 @@ runs:
...
-          STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
             | grep 'dynamo-platform-dynamo-operator-controller-manager' \
             | awk '{print $2}' | head -1)
-          if [ "${STATUS}" = "1/1" ]; then
-            echo "Operator pod Ready"
+          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+            | grep 'grove-operator' \
+            | awk '{print $2}' | head -1)
+          if [ "${DYNAMO_STATUS}" = "1/1" ] && [ "${GROVE_STATUS}" = "1/1" ]; then
```
dmitry-tokarev-nv: Not a bug today — I rendered `grove-charts v0.1.0-alpha.11` and it produces exactly one Deployment, one container, `replicaCount: 1`, no Jobs and no init containers, so `1/1` and `head -1` are accurate. This is purely future-proofing against a pinned **alpha** dependency:

- if Grove ever adds a sidecar the READY column becomes `2/2` and this loop burns the full 1800s on every deploy job before failing
- if a pod whose name merely starts with `grove-operator-` ever appears (a cert or migration Job), `head -1` can pin the status at `0/1` forever

Comparing ready-vs-total and requiring `Running` costs nothing and removes both:

```suggestion
          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /dynamo-platform-dynamo-operator-controller-manager/ && $3 == "Running" {print $2; exit}')
          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /grove-operator/ && $3 == "Running" {print $2; exit}')
          if [ -n "${DYNAMO_STATUS}" ] && [ -n "${GROVE_STATUS}" ] \
             && [ "${DYNAMO_STATUS%%/*}" = "${DYNAMO_STATUS##*/}" ] \
             && [ "${GROVE_STATUS%%/*}" = "${GROVE_STATUS##*/}" ]; then
```

dmitry-tokarev-nv (commented):
Reviewed at `083e55b65a`. The trimmed diff is a clear improvement — dropping the `publishNotReadyAddresses` chart value was the right call, and *"We don't want this. This hides problems."* is the correct read: it would have masked the failure mode in the one environment meant to regression-test it, and it expanded the chart's public API for a vCluster quirk.

**The Go change is correct.** I verified the wiring rather than assuming it: `webhookServer` is always non-nil (`main.go:247`, assigned at `:267`), `registerWebhookHandlers` (`:524`) unconditionally triggers `mgr.GetWebhookServer()` — which is what adds the server as a manager runnable — `AddReadyzCheck` precedes `mgr.Start`, and controller-runtime's `Start` only holds `s.mu` briefly to set `started = true`, so `StartedChecker()` cannot deadlock the probe.

**On "not sure this is actually the root cause" — the code agrees with you, fairly strongly.** `main.go:559` blocks on `certMgr.WaitForMountedCertificate` *before* `mgr.Start`, and the health server on `:8081` is itself started by `mgr.Start`. So nothing answers `/readyz` until the certs are already on disk, and the webhook listener comes up milliseconds later — while `readinessProbe.initialDelaySeconds: 5` already covers most of that window. The race is real and worth closing, but it cannot produce 60+ seconds of connection-refused *after* the workflow already waited for `1/1`.

**On "webhooks in general are flaky" — that is the most informative datum in the whole investigation**, and it points away from either operator. Two independent operators, independent charts, independent cert management, failing the same way in the same environment implicates the shared layer. `connect: connection refused` — as opposed to `no endpoints available`, which your retry loop already distinguishes at `setup-dynamo-operator/action.yml:528-530` — combined with a Service **DNS URL** in the Grove error (`https://grove-operator.default.svc:9443/...`, i.e. the ClusterIP path) is the signature of a ClusterIP with zero ready endpoints: kube-proxy installs a `REJECT --reject-with icmp-port-unreachable` rule for endpoint-less Services, and the socket layer surfaces that as `ECONNREFUSED`.

That single mechanism fits everything observed: both operators at once, `1/1` with no crash, self-healing after ~60s, and only in vCluster — where this action's own comment already documents that the syncer stalls pod-Ready propagation on AWS.

If that is right, the current diff no longer touches the observed symptom, which matches your own read. The value then rests entirely on the diagnostics, and that is where my main comment is.

Three things I would want before merge, none large: a guard against the Grove value silently becoming a no-op, `timeoutSeconds` on the Dynamo readiness probe now that it does real I/O, and an endpoint snapshot inside the smoke-test retry loop. Details inline.

<sub>Housekeeping: all three approvals are on `62bb8bd3e6` or earlier while head is `083e55b65a`, so the current diff — including the one thing that is new rather than removed — has not been reviewed by anyone but you. The three `fixup!` commits also want squashing unless this repo squash-merges.</sub>

deploy/operator/cmd/main.go
sttts: 5s makes more sense.

deploy/helm/charts/platform/values.yaml
sttts: Addressed in c31ee028de. Added the helm-docs comment and regenerated the README row. The setup action now renders the pinned Grove Deployment after helm dependency build and fails if no readinessProbe is present, so a renamed alpha-chart value cannot become a silent no-op. I also expanded the PR body with why enabling the existing Grove readiness gate is safe for all bundled-Grove users. Helm lint and all 31 chart tests pass, and the rendered Grove probe is /readyz on port 9444.

deploy/operator/cmd/main.go
sttts: Addressed in c31ee028de. Both the Helm Deployment and deploy/operator/config/manager/manager.yaml now set readinessProbe.timeoutSeconds to 5. The rendered Helm manager probe and the kustomize manifest were both verified at 5 seconds.

.github/actions/dynamo-deploy-test/action.yml
sttts: Agreed and addressed in c31ee028de. I removed the large post-failure diagnostic additions from both actions. The only new observation is now inside the existing admission retry, at the moment it sees the refusal. It records host and vCluster EndpointSlice address/Ready state plus pod UID, creation timestamp, Ready status/lastTransitionTime, and restart counts. It does not add retries or extend the existing deadline.

.github/actions/dynamo-deploy-test/action.yml
sttts: The downstream Debug admission failure step was removed entirely in c31ee028de, so the grep-based and incomplete log capture no longer exists. The focused retry snapshot does not fetch container logs.

.github/actions/setup-dynamo-operator/action.yml
sttts: Applied in c31ee028de. The host-side wait now requires Running, compares ready versus total container counts for both operators, and no longer assumes 1/1. This keeps the original 1800-second deadline unchanged while handling future sidecars.

.github/actions/dynamo-deploy-test/action.yml
sttts: Follow-up in 348b6b2e23: removed this retry snapshot again after the full run showed it was not reliable enough to keep. The vCluster EndpointSlice output confirmed stale `ready:false`, but the host selectors did not match vCluster-rewritten resource names and the custom Pod Ready columns rendered empty; repeating that partial snapshot on every retry added noise without resolving the issue. The retry count, 120-second deadline, and failure behavior remain unchanged. The PR description now documents both the observed stale EndpointSlice and removal of all PR-added diagnostics.

sttts:
Follow-up audit of [review 4798611212](https://github.com/ai-dynamo/dynamo/pull/12255#pullrequestreview-4798611212) against current head `348b6b2e23`:

### Retained in the current diff

- **Grove value regression guard:** after `helm dep build`, CI renders the pinned Grove Deployment and fails if it has no `readinessProbe`.
- **Grove default rationale and documentation:** the PR body explains why enabling Grove's existing probe is safe for bundled-Grove users; `values.yaml` has a helm-docs comment and the generated README contains the new row.
- **Dynamo probe timeout:** `timeoutSeconds: 5` is present in both the Helm Deployment and the kustomize manager manifest.
- **Future-proof operator wait:** the host-side loop requires `Running` and compares ready/total container counts for both Dynamo and Grove instead of assuming `1/1`.
- **No readiness bypass:** `publishNotReadyAddresses` and the temporary Service patches remain absent.

### Superseded rather than retained

- The requested **retry-time EndpointSlice/Pod snapshot** was implemented in `c31ee028de`, then removed in `348b6b2e23` after the full run showed that it was not reliable enough to keep: host selectors missed vCluster-rewritten resources, the Pod condition columns were incomplete, and the repeated output added noise. The virtual EndpointSlice result did establish `ready:false`, but the snapshot did not provide a dependable host/virtual comparison.
- The downstream log-capture robustness suggestions became moot because that entire PR-added diagnostic block was removed.

The retry behavior itself is unchanged: no additional attempts, no longer deadline, no skipped tests, and no endpoint-publication workaround.

### Housekeeping / merge readiness

- All five inline threads from the review have replies and are resolved.
- The current head still has no non-author approval; GitHub reports `REVIEW_REQUIRED`.
- The repository supports squash merge and disables merge commits, so the `fixup!` commit history is safe only if this PR is squash-merged.
- The latest full-CI `deploy-operator` job is still queued, so this audit is not claiming that the PR is currently green.

So: every review point has been explicitly handled, but it would be inaccurate to say every requested diagnostic remains implemented. One request was deliberately retired with the reason documented above and in its review thread.

sttts:
I investigated the latest failure live on the shared CI cluster. The evidence now identifies a separate vCluster/version-skew bug behind the persistent two-minute failures; it is not a remaining Dynamo or Grove listener/certificate race.

Tracking issue with the complete reproducer, evidence, upstream references, proposed remediation, and acceptance criteria: https://github.com/ai-dynamo/dynamo/issues/12266

Internal cluster, namespace, Pod, node, registry, and Service-IP identifiers are intentionally omitted or replaced here.

## Environment

```text
Host Kubernetes:        v1.34.9
kubectl client:          v1.33.2
vCluster:                v0.33.0
vCluster Kubernetes:     v1.32.13
vCluster image:          loft-sh/vcluster-pro:0.33.0
Tenant image:            loft-sh/kubernetes:v1.32.13
```

The tenant version comes from the current `vcluster_k8s_version: v1.32.13` default in `.github/actions/setup-dynamo-operator/action.yml`.

## Exact timeline from the latest failed job

Job: https://github.com/ai-dynamo/dynamo/actions/runs/30373470631/job/90327563378

1. `15:49:42`: the physical Dynamo operator Pod was created.
2. `15:49:49`: the `manager` container started.
3. `15:49:52`: certificates were ready, the health server was listening on 8081, and controller-runtime logged that the webhook server was serving on `0.0.0.0:9443`.
4. `15:50:06`: the host-side CI loop declared Dynamo and Grove Ready because `kubectl get pods` displayed all containers ready (`1/1`).
5. `15:50:08`: the first server-side admission smoke request failed.
6. The request was retried approximately every 2.8 seconds.
7. `15:52:08`: the final attempt exhausted the existing 120-second outer deadline:

```text
v1alpha1 admission smoke test did not become reachable within 120s

Error from server (InternalError):
conversion webhook for nvidia.com/v1alpha1, Kind=DynamoGraphDeployment failed:
Post "https://dynamo-platform-dynamo-operator-webhook-service.default.svc:443/convert?timeout=30s":
dial tcp <service-cluster-ip>:443: connect: connection refused
```

So the observed two minutes are exactly the pre-existing smoke-test retry deadline. This PR did not lengthen it. It is also not cert-manager startup time: the certificate and listener were ready before the first request.

## Smoking gun: ready container, non-ready Pod

While the admission requests were failing, the translated physical operator Pod simultaneously reported:

```text
status.containerStatuses[manager].ready = true
status.qosClass                         = Burstable
status.observedGeneration               = 1

Pod condition Ready                     = False (ContainersNotReady)
Pod condition ContainersReady           = False (ContainersNotReady)
Pod condition Initialized               = False
Pod condition PodReadyToStartContainers = False
```

The stale `False` conditions had a transition time from before the manager started. The Pod `resourceVersion` continued changing every 10–15 seconds while these conditions remained false. I caught one read where `Ready=True`; the immediately following full read had reverted to the original stale `Ready=False` state. This is active status oscillation/overwriting, not a slow probe.

The same stale physical-Pod condition pattern was visible on Grove. Since Dynamo and Grove have independent readiness implementations and certificate handling, identical corruption implicates the shared vCluster layer.

## Why `1/1 Running` was misleading

The `READY` column from `kubectl get pods` counts ready containers. It does **not** report the Pod `Ready` condition:

```text
containerStatuses[*].ready = true  -> kubectl displays 1/1
Pod Ready condition         = false -> Service endpoint remains unavailable
```

That means the current host-side loop in this PR produced a false positive. Comparing the two halves of `1/1` is sidecar-safe, but it still does not establish Pod readiness. The comment that host Pod status is “always accurate” is not true for this vCluster/version-skew failure.

The wait should inspect `status.conditions[type=Ready].status`, use `kubectl wait --for=condition=Ready`, or gate on a ready address in the webhook Service's EndpointSlice. Under the broken vCluster combination, that truthful wait should fail rather than allowing the admission smoke test to proceed.

## vCluster evidence and networking mechanism

The syncer logged errors from the same failure family:

```text
Failed to get pods assigned to node: Index with name spec.nodeName does not exist
Can't get CPU or zone information for node
Operation cannot be fulfilled on pods: the object has been modified
```

Preceding reproductions also showed the tenant webhook EndpointSlice stuck at:

```yaml
conditions:
  ready: false
  serving: false
```

The operator process is listening on 9443, but the Service has no usable ready endpoint. Requests through the ClusterIP are rejected before TLS or HTTP reaches the operator, which explains `connect: connection refused` rather than an x509/TLS error.

## Root cause

This closely matches the upstream host/tenant Kubernetes version-skew bug:

- https://github.com/loft-sh/vcluster/issues/3578
- https://github.com/loft-sh/vcluster/issues/3560
- proposed fix: https://github.com/loft-sh/vcluster/pull/4037

Kubernetes 1.34 writes Pod-status fields such as `observedGeneration` that the 1.32 tenant does not preserve in the same representation. vCluster sees a persistent difference and reconciles stale tenant status back into the physical object, fighting the host kubelet and overwriting readiness conditions. The continuously advancing `resourceVersion` and observed `True -> False` reversion are consistent with that mechanism.

As of today, both upstream issues remain open. PR #4037 remains open, blocked, and has changes requested, so upgrading only the vCluster binary is not yet a dependable remediation.

`qosClass` belongs to the same upstream failure family and was `Burstable` here, but there were no `status.qosClass is immutable` errors in this run. The active evidence is stale/oscillating Pod conditions plus `observedGeneration` under version skew.

## What this rules out

- **Dynamo crash:** manager stayed running with zero restarts and no panic/fatal logs.
- **Dynamo webhook startup race:** webhook was serving approximately 16 seconds before the smoke test began.
- **Certificate delay:** certificate readiness preceded the listener; the failure was connection refusal, not TLS validation.
- **Dynamo-only behavior:** Grove showed the same physical status corruption.
- **A longer wait introduced by this PR:** the failing duration is the unchanged 120-second admission deadline.
- **An active qosClass immutable-field error:** none appeared in this run.

## Recommended follow-up

The primary fix is to align the tenant minor version with the host:

```yaml
vcluster_k8s_version:
  default: v1.34.9
```

I verified that `ghcr.io/loft-sh/kubernetes:v1.34.9` exists for both `linux/amd64` and `linux/arm64`. This should first be validated as an explicit action input on the shared CI cluster, then made the default if the deploy-test clients support it.

Separately, the host wait should test the actual Pod Ready condition or the webhook EndpointSlice, not ready-container counts. That will make failures occur at the broken layer and prevent the misleading “operators Ready” message.

`publishNotReadyAddresses: true` would probably route around this by publishing the endpoint despite the false readiness state, but that defeats admission-webhook readiness semantics and hides the vCluster defect. It should remain absent except for a controlled diagnostic.

## Effect on this PR

The operator-side changes remain valid:

- Dynamo's `StartedChecker()` closes the real local listener false-ready window.
- The 5-second probe timeout is appropriate now that `/readyz` performs a TLS connection.
- Enabling Grove's existing readiness probe closes Grove's independent local false-ready window.

However, those changes cannot repair vCluster's host/tenant status synchronization. The latest red deploy job is environmental/shared-layer evidence, and the current `kubectl get pods` ready-count loop should not be treated as proof that the Pod or webhook Service is actually ready.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -474,34 +482,59 @@ runs:
...
-          if [ "${STATUS}" = "1/1" ]; then
-            echo "Operator pod Ready"
+          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+            | awk '$1 ~ /dynamo-platform-dynamo-operator-controller-manager/ && $3 == "Running" {print $2; exit}')
+          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+            | awk '$1 ~ /grove-operator/ && $3 == "Running" {print $2; exit}')
+          if [ -n "${DYNAMO_STATUS}" ] && [ -n "${GROVE_STATUS}" ] \
+             && [ "${DYNAMO_STATUS%%/*}" = "${DYNAMO_STATUS##*/}" ] \
+             && [ "${GROVE_STATUS%%/*}" = "${GROVE_STATUS##*/}" ]; then
+            echo "Dynamo and Grove operator pods Ready"
```
dmitry-tokarev-nv: Per #12266, this comment is now wrong on both counts and this check is the false positive that let the `15:50:06` gate pass:

> The comment that the host Pod status is "always accurate" should be corrected.

The `qosClass`/NATS rationale is also superseded — #12266 found no `status.qosClass is immutable` errors in that run, and attributes the corruption to host/tenant version skew instead.

My earlier suggestion made this sidecar-safe, which was worth doing, but as #12266 notes it "still does not establish Pod readiness": the `READY` column counts `containerStatuses[*].ready`, not `status.conditions[type=Ready]`, and under this bug those disagree. So the loop still cannot distinguish a healthy operator from the failure being worked around.

I would keep this as an explicit container-level *precondition* and move the authoritative gate into the step below, where it can verify the patch actually took effect. That reads honestly and needs no reordering:

```suggestion
        # Poll the host namespace instead of the vCluster API: kubectl wait against
        # the virtual cluster does not return reliably here. This is a container-level
        # precondition only -- it counts ready containers, not the Pod Ready condition,
        # and under the version skew tracked in
        # https://github.com/ai-dynamo/dynamo/issues/12266 those two disagree. The
        # authoritative readiness gate is the endpoint check in the next step.
        end=$((SECONDS + 1800))
        while true; do
          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /dynamo-platform-dynamo-operator-controller-manager/ && $3 == "Running" {print $2; exit}')
          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /grove-operator/ && $3 == "Running" {print $2; exit}')
          if [ -n "${DYNAMO_STATUS}" ] && [ -n "${GROVE_STATUS}" ] \
             && [ "${DYNAMO_STATUS%%/*}" = "${DYNAMO_STATUS##*/}" ] \
             && [ "${GROVE_STATUS%%/*}" = "${GROVE_STATUS##*/}" ]; then
            echo "Dynamo and Grove operator pods have all containers ready"
```

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -474,34 +482,59 @@ runs:
...
+          dynamo-platform-dynamo-operator-controller-manager \
+          grove-operator; do
+          pod=$(kubectl --kubeconfig="${VKUBECONFIG}" get pods -n default -o name \
+            | grep "${pod_pattern}" | head -1)
+          kubectl --kubeconfig="${VKUBECONFIG}" patch "${pod}" \
+            --namespace=default \
+            --subresource=status \
+            --type=strategic \
+            --patch='{"status":{"conditions":[{"type":"PodReadyToStartContainers","status":"True"},{"type":"Initialized","status":"True"},{"type":"ContainersReady","status":"True"},{"type":"Ready","status":"True"}]}}'
+        done
```
dmitry-tokarev-nv: **This is the one I would hold on.** Your own evidence in #12266 says the status is actively oscillating:

> I caught one read where `Ready=True`; the immediately following full read had reverted to the original stale `Ready=False` state. This is active status oscillation/overwriting, not a slow probe.

A single unverified patch against a controller observed reverting `Ready=True` between two consecutive reads is a coin flip, and the smoke test fires ~1-2 seconds later — before the endpointslice controller and kube-proxy have necessarily caught up even in the good case. There is no signal either way: if it holds, green with nothing learned; if it is reverted, the failure lands in the six downstream deploy-test jobs, which is where this started, now without the diagnostics.

Patching the tenant so the value vCluster keeps rewriting is the correct one is the right lever, and I think the direction is right. Verifying it is what turns that from a hypothesis into a tested one — and the verification doubles as the truthful readiness gate #12266 asks for, so it also covers the false positive in the step above.

Three smaller things in the same step:

- `pod=$(... | grep ... | head -1)` under `set -euo pipefail` with no `|| true`. Every other `grep | head -1` in this file has `|| true` for exactly this reason: a non-match, or a `SIGPIPE` when `head` closes the pipe, aborts the step with a cryptic failure.
- This runs once, in setup. The deploy-test jobs run up to 25 minutes against the same vCluster; a rescheduled or restarted operator Pod gets no patch.
- The comment does not link #12266. For a deliberate temporary workaround against a tracked upstream bug, that is the pointer a future reader needs most.

```suggestion
    - name: Restore vCluster operator pod readiness
      shell: bash
      run: |
        set -euo pipefail
        VKUBECONFIG=${{ github.workspace }}/.kubeconfig-vcluster

        # Workaround for https://github.com/ai-dynamo/dynamo/issues/12266: under
        # host/tenant Kubernetes version skew, vCluster v0.33.0 reconciles stale
        # tenant Pod status onto the physical Pod and overwrites the kubelet-owned
        # readiness conditions. Correcting the tenant Pod makes the value vCluster
        # keeps rewriting the correct one. Status was observed oscillating, so patch
        # and verify until the webhook Service actually publishes a ready endpoint,
        # and fail at the broken layer if it never does.
        end=$((SECONDS + 300))
        while true; do
          for pod_pattern in \
            dynamo-platform-dynamo-operator-controller-manager \
            grove-operator; do
            pod=$(kubectl --kubeconfig="${VKUBECONFIG}" get pods -n default -o name 2>/dev/null \
              | grep "${pod_pattern}" | head -1 || true)
            [ -n "${pod}" ] || continue
            kubectl --kubeconfig="${VKUBECONFIG}" patch "${pod}" \
              --namespace=default \
              --subresource=status \
              --type=strategic \
              --patch='{"status":{"conditions":[{"type":"PodReadyToStartContainers","status":"True"},{"type":"Initialized","status":"True"},{"type":"ContainersReady","status":"True"},{"type":"Ready","status":"True"}]}}' >/dev/null || true
          done

          READY=$(kubectl --kubeconfig="${VKUBECONFIG}" get endpointslices -n default \
            -l kubernetes.io/service-name=dynamo-platform-dynamo-operator-webhook-service \
            -o jsonpath='{.items[*].endpoints[*].conditions.ready}' 2>/dev/null || true)
          if grep -q true <<<"${READY}"; then
            echo "Webhook Service publishes a ready endpoint"
            break
          fi
          if [ "${SECONDS}" -ge "${end}" ]; then
            echo "::error::Webhook EndpointSlice never published a ready address; see https://github.com/ai-dynamo/dynamo/issues/12266"
            kubectl --kubeconfig="${VKUBECONFIG}" get endpointslices -n default -o wide 2>&1 || true
            exit 1
          fi
          sleep 5
        done
```

Worth extending the same endpoint check to `grove-operator`, since #12266 shows Grove had identical corruption — I left it out of the suggestion to keep it reviewable.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -399,6 +399,14 @@ runs:
         cd deploy/helm/charts/platform/
         helm dep build .
 
+        GROVE_RENDER=$(helm template dynamo-platform . \
+          --set global.grove.install=true \
+          --show-only charts/grove/templates/deployment.yaml)
+        if ! grep -q '^          readinessProbe:' <<<"${GROVE_RENDER}"; then
```
dmitry-tokarev-nv: Good guard — this is exactly what closes the silent-no-op hole.

One nit: anchoring on ten leading spaces couples the assertion to Grove's current template indentation, so a reindent in the pinned alpha chart trips it with a "does not render a readiness probe" error while the probe is present. `--show-only` already scopes the match to the Grove Deployment, so the anchor is not buying much:

```suggestion
        if ! grep -q 'readinessProbe:' <<<"${GROVE_RENDER}"; then
```

dmitry-tokarev-nv (commented):
Follow-up at `0b2dd25575`. All five threads from [review 4798611212](https://github.com/ai-dynamo/dynamo/pull/12255#pullrequestreview-4798611212) are resolved — the Grove regression guard, helm-docs comment and README row, `timeoutSeconds: 5` in both the Helm Deployment and the kustomize manifest, and the sidecar-safe wait are all in and correct. Thanks for the fast turnaround.

The investigation in #12266 is excellent and I think the root cause holds up. One detail there resolves something that had not added up: the `READY` column counts ready *containers*, while the EndpointSlice controller keys off the Pod `Ready` *condition*, and under this bug those two disagree. That closes the loop on `1/1 Running` coexisting with `ready:false, serving:false`, and confirms the ClusterIP path end to end — healthy listener, no ready backend published, connection refused before TLS.

Two things I would gate on, both in the new workaround step and both anchored inline.

**1. The diff contradicts its own issue.** #12266 states plainly that the host-side wait is a false positive in this failure mode and that *"the comment that the host Pod status is 'always accurate' should be corrected."* That comment is still in the diff, and the `qosClass`/NATS rationale next to it is superseded by the version-skew finding (#12266: no `qosClass is immutable` errors appeared in this run). My earlier suggestion made that check sidecar-safe, which #12266 correctly notes still "does not establish Pod readiness" — so the loop that produced the `15:50:06` false positive is unchanged.

**2. The patch is unverified against a system your own evidence describes as oscillating.** From #12266: *"I caught one read where `Ready=True`; the immediately following full read had reverted to the original stale `Ready=False` state. This is active status oscillation/overwriting."* A single `kubectl patch` with no confirmation, followed ~1-2s later by the smoke test, is a coin flip. If it holds, everything is green and nobody learns anything; if it is reverted, the failure resurfaces in the six downstream deploy-test jobs — the original situation, now with the diagnostics removed.

Patching the *tenant* so that the value vCluster keeps rewriting is the correct one is a genuinely clever inversion, and I think it is right. But it is a hypothesis about the bug's direction, and a verify loop is what turns it into a tested one. It also composes with point 1: patch, then gate on a ready endpoint, and you get both a truthful readiness check and a durable detector for recurrence — which is the piece that goes missing now that the diagnostics are gone.

**On removing the diagnostics** — reasonable now that the cause is known and documented. Worth noting for the record that both stated failures were fixable rather than fundamental: the host selectors missed vCluster-rewritten names, and `-o custom-columns` does not reliably evaluate JSONPath filter expressions like `[?(@.type=="Ready")]` (`-o jsonpath` does). The virtual half did its job — it produced the `ready:false` that pointed at the right layer.

**One question worth answering explicitly in the PR body:** #12266 names aligning `vcluster_k8s_version` with the host as the primary fix and suggests validating it as an explicit action input first. Is shipping the status-patching band-aid preferable to doing that? The band-aid fights a known upstream bug at a layer that can overwrite you, and #4037 being blocked upstream means it is not short-lived.

<sub>Housekeeping: seven commits now including a `fixup! fixup!`, and the three approvals are still pinned to `62bb8bd3e6` — the diff has changed twice since, including the addition of a step that writes to Pod status subresources.</sub>

.github/actions/setup-dynamo-operator/action.yml
sttts: Addressed in f7bf8e352b. The host-side loop is now explicitly documented and logged as a container-level precondition only; the stale qosClass/NATS and “always accurate” claims are gone. The authoritative readiness gate is now the EndpointSlice verification in the following step.

.github/actions/setup-dynamo-operator/action.yml
sttts: Addressed in f7bf8e352b. The workaround now links #12266, tolerates missing pods and pipefail/SIGPIPE, repeatedly patches the tenant status, and verifies ready EndpointSlice addresses for both the Dynamo and Grove webhook Services before continuing. It fails in setup with the EndpointSlices printed if readiness never publishes. I kept the bound at 120s—the existing admission-failure window investigated here—instead of extending it to 300s. The PR body now also explains why tenant-version alignment remains the preferred follow-up but is a broader compatibility change than this PR.

.github/actions/setup-dynamo-operator/action.yml
sttts: Addressed in f7bf8e352b. The Grove render assertion now matches `readinessProbe:` without coupling the guard to the chart indentation.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -474,34 +482,59 @@ runs:
...
+            echo "::error::Operator pods not Ready after 1800s (Dynamo: ${DYNAMO_STATUS:-not found}, Grove: ${GROVE_STATUS:-not found})"
             exit 1
           fi
-          echo "Waiting for operator pod... status=${STATUS:-not found}"
+          echo "Waiting for operator pods... Dynamo=${DYNAMO_STATUS:-not found}, Grove=${GROVE_STATUS:-not found}"
           sleep 15
         done
         echo "::endgroup::"
 
+    - name: Restore vCluster operator pod readiness
```
sttts: this is dirty, not fixing the root cause. I don't think we want this.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -474,30 +482,34 @@ runs:
...
-          if [ "${STATUS}" = "1/1" ]; then
-            echo "Operator pod Ready"
+          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+            | awk '$1 ~ /dynamo-platform-dynamo-operator-controller-manager/ && $3 == "Running" {print $2; exit}')
+          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
+            | awk '$1 ~ /grove-operator/ && $3 == "Running" {print $2; exit}')
+          if [ -n "${DYNAMO_STATUS}" ] && [ -n "${GROVE_STATUS}" ] \
+             && [ "${DYNAMO_STATUS%%/*}" = "${DYNAMO_STATUS##*/}" ] \
+             && [ "${GROVE_STATUS%%/*}" = "${GROVE_STATUS##*/}" ]; then
+            echo "Dynamo and Grove operator pods Ready"
```
dmitry-tokarev-nv: This is the substantive one. Fixed in `f7bf8e352b`, lost in the force-push — at head these lines are back to the pre-`f7bf8e352b` text.

Both claims here are contradicted by #12266, which you filed two hours before this landed:

> The comment that the host Pod status is "always accurate" should be corrected.

and the `qosClass`/NATS rationale is superseded — #12266 found no `status.qosClass is immutable` errors in that run and attributes the corruption to host/tenant version skew.

It matters more now than when I first raised it: with the workaround step removed, this loop is the only readiness gate left in the action, and #12266 documents this exact check as the false positive that let the `15:50:06` gate pass. Shipping it with a comment asserting the opposite is what the next person debugging this will read first.

To be clear on scope: I am **not** asking for the truthful `status.conditions[type=Ready]` gate here. That is #12266 remediation work, and the check itself is pre-existing — this diff only made it sidecar-safe. Just the comment.

```suggestion
        # Poll the host namespace instead of the vCluster API: kubectl wait against
        # the virtual cluster does not return reliably here. This is a container-level
        # precondition only -- it counts ready containers, not the Pod Ready condition,
        # and under the version skew tracked in
        # https://github.com/ai-dynamo/dynamo/issues/12266 those two disagree.
        end=$((SECONDS + 1800))
        while true; do
          DYNAMO_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /dynamo-platform-dynamo-operator-controller-manager/ && $3 == "Running" {print $2; exit}')
          GROVE_STATUS=$(kubectl get pods -n "${NAMESPACE}" --no-headers 2>/dev/null \
            | awk '$1 ~ /grove-operator/ && $3 == "Running" {print $2; exit}')
          if [ -n "${DYNAMO_STATUS}" ] && [ -n "${GROVE_STATUS}" ] \
             && [ "${DYNAMO_STATUS%%/*}" = "${DYNAMO_STATUS##*/}" ] \
             && [ "${GROVE_STATUS%%/*}" = "${GROVE_STATUS##*/}" ]; then
            echo "Dynamo and Grove operator pods have all containers ready"
```

.github/actions/setup-dynamo-operator/action.yml
dmitry-tokarev-nv: Same story — de-anchored in `f7bf8e352b`, back at head.

`--show-only` already scopes the match to the Grove Deployment, so the ten leading spaces only couple the guard to the pinned alpha chart's current indentation:

```suggestion
        if ! grep -q 'readinessProbe:' <<<"${GROVE_RENDER}"; then
```

dmitry-tokarev-nv (commented):
Re-flagging two small items that regressed in the last force-push — not a new review, just recovering work that was already done.

`f7bf8e352b` (16:57Z) fixed both, but it is a **sibling** of the current head rather than an ancestor:

```text
compare f7bf8e352b...1d47a21e87  ->  status: diverged, ahead 1, behind 1
```

so its changes are not in `1d47a21e87`. The corresponding threads are marked resolved against that commit, and both are also `outdated`, so this is invisible from the review UI. Verified against head rather than inferred from the diff.

**No action needed on the patch step** — removing it was the right call, and I agree with the reasoning. This is only the collateral.

Nothing else in the current diff is open from my side: the `main.go` change, `timeoutSeconds: 5` in both manifests, the Grove value with its helm-docs comment and README row, and the sidecar-safe wait all check out.

<sub>Also still worth doing before merge: squash the `fixup!` / `fixup! fixup!` commits, and note that the three approvals are pinned to `62bb8bd3e6`, five commits back.</sub>

.github/actions/setup-dynamo-operator/action.yml
sttts: Addressed in bd442fd389. The Grove render guard now matches readinessProbe: without depending on chart indentation.

.github/actions/setup-dynamo-operator/action.yml
sttts: Resolved. The Pod-status patch step remains removed at bd442fd389; there is no replacement workaround.

.github/actions/setup-dynamo-operator/action.yml
sttts: Addressed in bd442fd389. The comment now describes this as a container-level precondition, explicitly distinguishes ready containers from the Pod Ready condition, links #12266, and removes the superseded NATS/qosClass and host-status-accuracy claims. The log message now says that all containers are ready.

PR #12277, opened 2026-07-28
fix: forward logprob_token_ids through the OpenAI frontend

AineshSootha:
vLLM has supported `logprob_token_ids` on `SamplingParams` since v0.20 (vllm-project/vllm#34539) and exposed it on its own OpenAI HTTP schema in v0.26.0 (vllm-project/vllm#43463). It pins the set of vocab ids whose logprobs appear in the response, independent of top-k rank — the primitive used for classifier-style scoring, where a small fixed label set must be read at the first decoded position even when a label would not surface in top-k.

Dynamo does not serve vLLM's OpenAI app, so the field landed in `unsupported_fields` and was rejected with `400 Unsupported parameter(s)` before reaching the worker. Two allowlists gate it:

- `PASSTHROUGH_EXTRA_FIELDS` (validate.rs), checked at the HTTP layer
- `sampling_passthrough_args` (preprocessor.rs), which forwards into `extra_args.sampling_options`

Add the field to both. `build_sampling_params` already applies these generically via setattr, so no worker-side change is needed.

Verified end-to-end against a Dynamo image built on vllm/vllm-openai:v0.23.0 serving Qwen3-0.6B: a request pinning three class tokens returns logprobs for exactly those ids, and the values are discriminative across prompts.

Note for callers: vLLM's `SamplingParams.verify()` requires `logprobs == len(logprob_token_ids)` when both are set, so `top_logprobs` must match the number of pinned ids. vLLM's own frontend sidesteps this by nulling `logprobs`; this change forwards `top_logprobs` as-is rather than changing existing semantics.

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for `logprob_token_ids` in sampling options.
  * Requests using `logprob_token_ids` are now accepted and correctly forwarded for backend processing.
  * Preserved compatibility with existing sampling parameters and validation behavior.

### DIFF
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -798,7 +798,14 @@ def build_sampling_params(
     # Apply output_options (logprobs, prompt_logprobs, etc.)
     output_options = request.get("output_options", {}) or {}
     logprobs, prompt_logprobs = _shared_logprobs.parse_logprob_options(output_options)
-    if logprobs is not None:
+    # Explicit `logprob_token_ids` replace vLLM's natural top-k selection, so the
+    # requested width no longer applies. vLLM's own OpenAI adapters null `logprobs`
+    # in this case and let `num_logprobs` derive the width from the id list; mirror
+    # that here, otherwise `SamplingParams.verify()` rejects the pair unless the
+    # caller happens to set `top_logprobs == len(logprob_token_ids)`.
+    if getattr(sampling_params, "logprob_token_ids", None):
+        sampling_params.logprobs = None
+    elif logprobs is not None:
         sampling_params.logprobs = logprobs
     if prompt_logprobs is not None:
         sampling_params.prompt_logprobs = prompt_logprobs
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -788,7 +788,14 @@ def build_sampling_params(
     # Apply output_options (logprobs, prompt_logprobs, etc.)
     output_options = request.get("output_options", {}) or {}
     logprobs, prompt_logprobs = _shared_logprobs.parse_logprob_options(output_options)
-    if logprobs is not None:
+    # Explicit `logprob_token_ids` replace vLLM's natural top-k selection, so the
+    # requested width no longer applies. vLLM's own OpenAI adapters null `logprobs`
+    # in this case and let `num_logprobs` derive the width from the id list; mirror
+    # that here, otherwise `SamplingParams.verify()` rejects the pair unless the
+    # caller happens to set `top_logprobs == len(logprob_token_ids)`.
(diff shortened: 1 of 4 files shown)


jh-nv (approved):
Acknowledge the limitation on the top_logprobs. It is ok that caller needs to set it properly for now, but the semantic is bit strange. Could you please make a ticket to record this so we come back to it?

rmccorm4:
/ok to test [4b5f203](https://github.com/ai-dynamo/dynamo/pull/12277/commits/4b5f203fb4e2ba7b933a009111b50415f19b2f2c)

rmccorm4:
vLLM deliberately sets logprobs=None when explicit IDs exists, let's follow the same pattern: https://github.com/vllm-project/vllm/blob/v0.26.0/tests/entrypoints/openai/chat_completion/test_logprob_token_ids.py#L56-L71.

Something like this:

```python
output_options = request.get("output_options", {}) or {}
logprobs, prompt_logprobs = _shared_logprobs.parse_logprob_options(output_options)

# Explicit IDs replace natural top-k selection in vLLM.
if getattr(sampling_params, "logprob_token_ids", None):
    sampling_params.logprobs = None
elif logprobs is not None:
    sampling_params.logprobs = logprobs

if prompt_logprobs is not None:
    sampling_params.prompt_logprobs = prompt_logprobs
```

The Rust passthrough and `output_options` should remain unchanged: they preserve the client's request to return logprobs. Only the vLLM engine parameter needs normalization. With `SamplingParams.logprobs=None`, vLLM's `num_logprobs` derives the width from `len(logprob_token_ids)`, matching its OpenAI adapters; an empty ID list falls through to the existing top-k behavior.

A focused test in `components/src/dynamo/vllm/tests/test_vllm_tito_parity.py` should cover `output_options.logprobs=5` plus `logprob_token_ids=[5000]` and assert:

```python
assert sp.logprob_token_ids == [5000]
assert sp.logprobs is None
assert sp.num_logprobs == 1
```

rmccorm4 (approved):
One small comment to address: https://github.com/ai-dynamo/dynamo/pull/12277#issuecomment-5149370632

Otherwise LGTM, thanks for the contribution @AineshSootha !

rmccorm4:
/ok to test [6835b5c](https://github.com/ai-dynamo/dynamo/pull/12277/commits/6835b5c1d203b173d5404382d88fed70c82ea111)

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `75d49d7` | unsigned |
| `868371c` | unsigned |
| `480453f` | unsigned |

rmccorm4:
/ok to test [2fce060](https://github.com/ai-dynamo/dynamo/pull/12277/commits/2fce0601750de038a9ef35fddc6a6d906a5fd0e0)

karen-sy (approved):

PR #12279, opened 2026-07-28
ci(deploy-test): align vCluster Kubernetes with host and remove readiness workaround

### LINKED ISSUE #12266: vCluster Kubernetes version skew corrupts Pod readiness and breaks operator webhooks in deploy CI
## Summary

The shared deploy CI currently runs a Kubernetes `v1.32.13` vCluster tenant on a Kubernetes `v1.34.9` host. In this combination, vCluster can continuously overwrite a physical Pod's kubelet-owned status with stale tenant status. The operator container is healthy and serving its webhook, but the Pod `Ready` condition remains or repeatedly becomes `False`. Service discovery then keeps the webhook endpoint unavailable, so conversion/admission requests fail with `connection refused` until the existing 120-second smoke-test deadline expires.

This is a shared vCluster/version-skew problem, not a Dynamo webhook startup or certificate race. It can affect any webhook-backed component in the tenant; the same run showed stale conditions on both Dynamo and Grove operator Pods.

Related PR: https://github.com/ai-dynamo/dynamo/pull/12255
Failed job: https://github.com/ai-dynamo/dynamo/actions/runs/30373470631/job/90327563378

Internal CI cluster, namespace, Pod, node, registry, and Service-IP identifiers are intentionally omitted or replaced below.

## Affected environment

```text
Host:                   shared CI Kubernetes cluster
Host Kubernetes:        v1.34.9
kubectl client:          v1.33.2
vCluster:                v0.33.0
vCluster Kubernetes:     v1.32.13
vCluster image:          loft-sh/vcluster-pro:0.33.0
Tenant image:            loft-sh/kubernetes:v1.32.13
```

The tenant version is currently the default in `.github/actions/setup-dynamo-operator/action.yml`:

```yaml
vcluster_k8s_version:
  description: Kubernetes version for the vCluster control plane
  default: v1.32.13
```

and is passed as:

```bash
vcluster create "${VCLUSTER_NAME}" \
  --set controlPlane.distro.k8s.enabled=true \
  --set controlPlane.distro.k8s.version=${{ inputs.vcluster_k8s_version }}
```

## Exact failure timeline

From public run `30373470631`:

1. `15:49:42`: the physical operator Pod was created.
2. `15:49:49`: the Dynamo `manager` container started.
3. `15:49:52`: certificates were ready, the health server was listening on 8081, and controller-runtime logged that the webhook server was serving on `0.0.0.0:9443`.
4. `15:50:06`: the host-side CI loop declared both operators Ready because `kubectl get pods` displayed all containers as ready (`1/1`).
5. `15:50:08`: the first server-side admission smoke request failed.
6. Attempts continued approximately every 2.8 seconds.
7. `15:52:08`: the final attempt exhausted the existing 120-second outer deadline:

```text
v1alpha1 admission smoke test did not become reachable within 120s

Error from server (InternalError):
conversion webhook for nvidia.com/v1alpha1, Kind=DynamoGraphDeployment failed:
Post "https://dynamo-platform-dynamo-operator-webhook-service.default.svc:443/convert?timeout=30s":
dial tcp <service-cluster-ip>:443: connect: connection refused
```

The observed two minutes are therefore exactly the smoke test's existing retry deadline. They are not time spent waiting for cert-manager or for the operator process to start.

## Live physical-Pod evidence

While the job was failing, the translated physical operator Pod simultaneously reported:

```text
status.containerStatuses[manager].ready = true
status.qosClass                         = Burstable
status.observedGeneration               = 1

Pod condition Ready                     = False (ContainersNotReady)
Pod condition ContainersReady           = False (ContainersNotReady)
Pod condition Initialized               = False
Pod condition PodReadyToStartContainers = False
```

The stale `False` conditions had a `lastTransitionTime` from before the manager container started. The Pod `resourceVersion` continued changing every 10–15 seconds while these conditions remained false. One observation caught `Ready=True`; the immediately following read had reverted it to the original stale `Ready=False` state. This is status oscillation/overwriting, not a slow readiness probe.

The physical Pod therefore had a ready manager container and a false Pod readiness condition at the same time. `kubectl get pods` still displayed `1/1 Running` because its `READY` column counts ready containers; it does not display the Pod `Ready` condition.

The same stale false-condition pattern was visible on the physical Grove operator Pod, which independently implements its own certificate and webhook readiness checks. This points to the shared virtualization layer rather than either operator.

## vCluster syncer evidence

The vCluster syncer logged errors from the same failure family:

```text
Failed to get pods assigned to node: Index with name spec.nodeName does not exist
Can't get CPU or zone information for node
Operation cannot be fulfilled on pods: the object has been modified
```

Tenant-side observations from preceding reproductions also showed the operator EndpointSlice stuck at:

```yaml
conditions:
  ready: false
  serving: false
```

This explains the networking symptom: the API server reaches the webhook through the Service ClusterIP, but no ready backend is published, so the connection is rejected before TLS or HTTP reaches the healthy listener.

## Likely mechanism

This closely matches the upstream host/tenant version-skew bug. Kubernetes 1.34 writes Pod-status fields such as `observedGeneration` that the Kubernetes 1.32 tenant does not preserve in the same representation. vCluster repeatedly sees a difference and reconciles stale tenant status back into the physical object, fighting the host kubelet and overwriting readiness conditions.

Relevant upstream reports:

- https://github.com/loft-sh/vcluster/issues/3578 — sync failures with Kubernetes 1.34/1.35 hosts and older guests
- https://github.com/loft-sh/vcluster/issues/3560 — Pod readiness and EndpointSlice synchronization failure, including the missing `spec.nodeName` index
- https://github.com/loft-sh/vcluster/pull/4037 — proposed fix for Pod health under host/tenant version skew

As of 2026-07-28, both issues remain open. PR #4037 is open, blocked, and has changes requested, so upgrading only the vCluster binary is not currently a dependable fix.

`status.qosClass` belongs to the same upstream failure family and was `Burstable` in this run, but there were no `status.qosClass is immutable` errors. The active evidence here is stale/oscillating Pod conditions plus `observedGeneration` under version skew.

## CI false positive

The current host-side operator wait in `.github/actions/setup-dynamo-operator/action.yml` parses column 2 of `kubectl get pods`, for example `1/1`, and treats equal ready/total container counts as Pod readiness. This is insufficient in this failure mode:

```text
containerStatuses[*].ready = true  -> kubectl displays 1/1
Pod Ready condition         = false -> Service endpoint remains unavailable
```

The comment that the host Pod status is “always accurate” should be corrected. A wait should inspect `status.conditions[type=Ready].status`, or wait for the webhook EndpointSlice to contain a ready endpoint. With the current version-skew bug, that wait should fail rather than incorrectly allowing the admission smoke test to proceed.

### Primary fix

Align the vCluster tenant Kubernetes minor version with the host:

```yaml
vcluster_k8s_version:
  default: v1.34.9
```

`ghcr.io/loft-sh/kubernetes:v1.34.9` was verified to exist for both `linux/amd64` and `linux/arm64`.

This should first be validated as an explicit action input on the shared CI cluster, then made the default if the deploy-test Python/Kubernetes clients are compatible.

### Make the readiness wait truthful

Replace the `kubectl get pods` ready-container-count check with an actual Pod-condition check, for example:

```bash
kubectl wait \
  --for=condition=Ready \
  pod/<translated-physical-pod> \
  -n "${NAMESPACE}" \
  --timeout=...
```

Alternatively, gate the admission smoke test on a ready address in the webhook Service's EndpointSlice. This does not repair vCluster, but it prevents the misleading `1/1` false positive and fails at the broken layer.

### Not recommended as a fix

`publishNotReadyAddresses: true` restores traffic by bypassing readiness filtering and is documented as an upstream workaround. It would, however, route to admission webhooks before Kubernetes considers them ready and hide the underlying synchronization bug. It should be used only for a controlled diagnostic, not as the CI or chart default.

## Ruled out

- **Dynamo process crash:** manager remained running with zero restarts and no fatal/panic logs.
- **Webhook startup race:** webhook was serving on 9443 approximately 16 seconds before the admission test began.
- **Certificate provisioning delay:** certificates were ready before the webhook listener started; the final error was connection refusal, not an x509 or TLS error.
- **Smoke test extending the wait:** the retry deadline was already 120 seconds; no new delay was introduced.
- **Dynamo-only bug:** Grove exhibited the same physical Pod-condition corruption.
- **Active qosClass rejection in this run:** no immutable-qosClass errors appeared in the syncer logs.

## Acceptance criteria

- [ ] A vCluster tenant on the shared CI cluster uses a Kubernetes version compatible with the host 1.34 minor version.
- [ ] Physical operator Pod conditions no longer oscillate or revert after kubelet marks the container ready.
- [ ] Both Dynamo and Grove webhook EndpointSlices publish ready endpoints.
- [ ] The host-side CI wait checks the actual Pod Ready condition or ready EndpointSlice state, not only ready-container counts.
- [ ] The v1alpha1 and v1beta1 server-side admission smoke tests succeed without `publishNotReadyAddresses`.
- [ ] The existing 120-second admission deadline is unchanged.

sara4dev:
## Summary

- Derive the vCluster guest Kubernetes version from the host API at setup time, normalizing provider-suffixed versions to the upstream `vMAJOR.MINOR.PATCH` image tag.
- Remove the temporary `publishNotReadyAddresses` overrides for the Dynamo and Grove webhook Services, including the chart value and its dedicated unit test.
- Keep the readiness gates, admission smoke tests, Grove health probes, and failure diagnostics added by #12278 and #12368.
- Preserve #12399's removal of the brittle Grove defaulting-mutation assertion.

Fixes #12266.

## Root cause

The deploy-test vCluster guest was pinned to Kubernetes `v1.32.13` while the host had moved to `v1.34.9`. That skew triggers loft-sh/vcluster#3578: the syncer can fail to write Pod status because `status.qosClass` is immutable, leaving guest Pods stale even when their host Pods are running.

The old static pin could drift again on the next host upgrade. This change reads the host's `serverVersion.gitVersion` and uses its upstream semantic version for the vCluster control-plane image.

## Validation

- Rebased onto `4830f0c7ad`, which includes #12399.
- DCO-signed commit; signature verified locally.
- Targeted pre-commit hooks passed, including YAML, whitespace, conflict, and codespell checks.
- Composite action parses successfully with all 18 steps intact and no stale `vcluster_k8s_version` input.
- Version normalization verified for upstream, EKS-suffixed, and build-suffixed Kubernetes versions.
- Operator Helm chart lints and renders successfully; the rendered Service omits `publishNotReadyAddresses`.
- Full CI passed on `187aacbf02`: `deploy-operator`, both operator jobs, and all eight vLLM/SGLang/TRT-LLM deploy variants are green.

The first full-CI attempt hit the self-hosted runner's 30-minute cutoff in an unrelated TRT-LLM parallel GPU test; the same job failed concurrently on PR #12363. A failed-jobs rerun passed without code changes.

The full platform Helm dependency build could not complete locally because the external NATS chart download stalled; the component chart used by this change was linted and rendered directly.

## Out of scope

- The shared-vCluster bootstrap race between parallel matrix jobs.
- The 1800-second deployment wait running under a 1200-second pytest timeout.


.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -475,30 +478,30 @@ runs:
...
       shell: bash
       env:
         NAMESPACE: ${{ steps.resolve-names.outputs.namespace }}
       run: |
-        echo "::group::Wait for operator manager container readiness (host namespace)"
-        # Poll the host namespace instead of the vCluster API because Pod status
-        # propagation can lag in either direction. The READY column confirms
-        # container readiness only; the admission smoke test below is the
-        # authoritative end-to-end readiness gate.
+        echo "::group::Wait for operator pod Ready (host namespace)"
```
sttts: in this loop we should also wait for ready Grove operator (or just use helm's `--wait`).

sara4dev: okay, can we add this as follow up PR?

dagil-nvidia (approved):
Approving. Green across 81 checks, and this removes the CI workaround rather than adding one.

The chart-surface removal is safe: `dynamo-operator.webhook.service.publishNotReadyAddresses` was introduced yesterday in #12278 and is not in `release/1.3.0`, so it never shipped and nothing can depend on it. Better to take this before the v1.4.0 cut than to ship a production chart value that exists only to paper over a CI environment bug.

One follow-up, not a blocker: the previous code carried an explicit constraint that the vCluster guest version must be supported by the kr8s Python library, deliberately independent of the host. Deriving it from the host couples them, so a future host upgrade past kr8s support will surface as confusing deploy-test failures rather than an unsupported-version error. Worth confirming kr8s covers 1.34, and considering an upper bound or supported-version allowlist.

.github/actions/setup-dynamo-operator/action.yml
sara4dev: rebased from the main and kept grove operator readiness check

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -47,9 +47,9 @@ inputs:
     required: false
     default: ''
   vcluster_k8s_version:
-    description: 'Kubernetes version for the vCluster control plane (must be supported by kr8s)'
+    description: 'Kubernetes version for the vCluster control plane. Keep aligned with the host cluster version, and it must be supported by kr8s.'
     required: false
-    default: 'v1.32.13'
+    default: 'v1.34.9'
```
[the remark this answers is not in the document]
sara4dev: used dynamic k8s version now

PR #12281, opened 2026-07-28
refactor(preprocessor)!: let workers declare their per-image token

furionw:
## Summary

Kimi-K3's prompt carries one placeholder per image, and the serving engines disagree on which token belongs there: vLLM builds the native media sequence from `<|kimi_image_placeholder|>`, while SGLang only repeats `<|media_pad|>` up to the feature count.

`expands_image_pad_token` encoded that as a mode flag naming how SGLang's processor works internally. The preprocessor used it to pick between two renderer-supplied strings and rewrite the rendered segments after the fact. The flag was only ever a proxy for the string, and the rewrite only existed because rendering ran before the engine's preference was consulted.

This carries the string instead. `image_placeholder_token: Option<String>` is read from the processor the worker actually loaded, so it states that engine's real contract rather than an assumption about it, and the formatter renders it directly. `substitute_image_pad_token` and `swap_marker_segments` are gone, and so is the notion of pad expansion.

## Why this is safe for non-K3 models

SGLang now declares a token for every multimodal model it serves, not just K3. Containment is structural: the token is threaded only into `kimi_k3_formatter_for`, and no Jinja-templated family has a parameter to receive one. For Qwen3-VL, `is_kimi_k3()` returns false, execution falls through to `PromptFormatter::from_parts`, and the field is never read by any code path.

That is stronger than what it replaces. The old rewrite ran for every model on every request and was inert only because both trait accessors had to be `Some` and Jinja prompts carry no segments.

Workers that declare nothing are unchanged: vLLM keeps rendering the checkpoint default, and `skip_serializing_if` keeps them byte-identical on the wire.

## Follow-up commit

Review found that `allow_special: true` on a rendered segment only selects tiktoken's special-aware encode method — it is not a membership check. An unregistered token falls through to ordinary BPE and silently shatters, so one image becomes several ids. The preprocessor now checks the declaration once at construction and warns. It sits in `new_with_parts` because the discovery path builds the formatter and preprocessor separately and never calls `new`.

## Breaking changes

- `ModelRuntimeConfig.expands_image_pad_token` is replaced by `ModelRuntimeConfig.image_placeholder_token`.
- Rolling upgrade is a silent downgrade: `ModelRuntimeConfig` has no `deny_unknown_fields`, so an old worker publishing `expands_image_pad_token` at a new frontend deserializes to `None` and reverts to the default with no error.

## Known limitations (pre-existing, not introduced here)

- Only the first-registered worker's declaration takes effect per model — `mdcsum` does not cover `runtime_config`, so a mixed vLLM+SGLang pool for one model is first-card-wins. Same shape as the flag it replaces; now recorded in the field doc.
- mm-routing's `routing_image_token_id` is still resolved from `config.json` and is deliberately not wired to the declared token. Verified behaviour-preserving: token ids are byte-identical before and after for both engines, and no family in the lightseek registry was ever affected. Wiring it needs a `kimi_k3` registry entry and a fix to the `mm_hashes` gate first.

## Test plan

- `cargo test -p dynamo-llm --no-default-features --features mm-routing` — preprocessor suite 15/15
- `cargo clippy -p dynamo-llm --no-default-features --features mm-routing --all-targets -- -D warnings` — clean
- `cargo check` on `lib/bindings/python` — clean
- `pre-commit run` on all changed files — clean
- New `declared_token_does_not_change_a_jinja_rendered_prompt`, the regression guard for the containment claim above
- New SGLang accessor tests: text-only worker, token read from the loaded processor, multiple spellings declined, missing `mm_tokens` logged distinctly, engine without a tokenizer manager
- Built and tested on macOS with `--no-default-features`; `block-manager` is Linux-only. The 9 ZMQ/event-plane lib-test failures in that environment were confirmed pre-existing by reproducing the identical set on the unmodified base. Linux CI is the real gate.

## Merge order

Requires ai-dynamo/frontend-crates#155 to merge into `feat/Kimi-K3` first. **CI on this PR will be red until then** — the branch calls `kimi_k3_formatter_for` with four arguments and the pinned renderer still takes three.

After that lands, `cargo update -p dynamo-renderer` is needed for **both** manifests — the root `Cargo.toml` and `lib/bindings/python/Cargo.toml`, which carries its own `[patch.crates-io]` block because the bindings crate is outside the workspace.


lib/bindings/python/src/dynamo/_core.pyi
```diff
@@ -864,6 +864,7 @@ class ModelRuntimeConfig:
     data_parallel_start_rank: int
     data_parallel_size: int
     enable_local_indexer: bool
+    image_placeholder_token: str | None
```
rmccorm4: ### Proposed Kimi-K3 placeholder handling

Standardize Dynamo’s frontend contract on one `<|media_pad|>` special token per image.

- SGLang consumes this directly.
- On the raw-media path, the vLLM adapter converts each media-pad ID into the checkpoint-native `<|kimi_image_placeholder|>` token sequence before constructing `TokensPrompt`.
- vLLM’s existing Kimi-K3 processor then performs its normal media expansion.
- Both values come from checkpoint metadata, so no new model-registration field or SGLang processor introspection is required.

The conversion should live in vLLM’s `build_tokens_prompt()` path and run only when raw image data is present. It must not touch already-processed inputs.

Future frontends that preprocess through Rust frontend preprocessing (similar to SMG frontend) should use the explicit pre-rendered contract:

- `expanded_token_ids`
- Multimodal placeholder ranges
- Processed multimodal inputs/`mm_kwargs`
- Multimodal hashes

Those requests bypass the raw vLLM adapter. The request should retain the canonical one-pad-per-image `token_ids` as a fallback if transferring the expanded inputs fails.

For rollout, deploy a vLLM adapter that accepts both the old native marker and the new media-pad form, then switch the frontend to always emit `<|media_pad|>`. Reject mixed or mismatched placeholder counts rather than guessing.

rmccorm4: TLDR; I'd rather not start with registering a custom field here via MDC/discovery if there is a suitable alternative, because these types of changes become very sticky and hard to remove later on without breaking the public python bindings API surface.

My proposal above is to always send the same thing from the frontend (`<|media_pad|>`) and lightly handle it on vllm backend (adapter). This is an easier thing to go back and change later if we decide we want to do it another way, because we can just handle different paths based on the input in the adapter logic here without updating our python API surface. I'm also picturing we'll probably need similar things for video token, etc. in future, and prefer to not overload the runtime config / register method with too many params if we can avoid it for now.

Let me know what you think.

rmccorm4: This is an example adapter fix in build_tokens_prompt() in vllm backend multimodal utils:
```python
if hf_config.model_type == "kimi_k3":
    pad_id = hf_config.media_placeholder_token_id
    native_ids = tokenizer.encode(
        hf_config.image_placeholder,
        add_special_tokens=False,
    )

    # Replace exactly one structural pad ID per supplied image.
    token_ids = replace_each(token_ids, pad_id, native_ids)
```

NOTE: I chose above to always send `<|media_pad|>` from frontend instead of always send `<|kimi_image_placeholder|>` because `<|media_pad|>` is a literal token ID in the tokenizer vocab (`hf_config.media_placeholder_token_id`), so a single consistent thing to replace -- whereas `<|kimi_image_placeholder|>` is a string and not a single token ID (`hf_config.image_placeholder`) and relies on string parsing, so a bit harder to replace consistently - as far as I understand it.

furionw:
Superseded by the smaller draft stack:\n\n- #12396 — native Kimi-K3 frontend (base)\n- #12394 — vLLM media-pad expansion (stacks on #12396)\n- #12393 — SGLang ServerArgs override compatibility (stacks on #12396)\n- #12395 — SGLang Kimi-K3 multimodal alignment (stacks on #12393)\n\nThis preserves the agreed lineage while allowing the vLLM and SGLang work to review and land independently.

PR #12263, opened 2026-07-28
refactor(operator): extract Grove workload rendering from graph controller

julienmancuso:
## Summary

This PR introduces a dedicated `groveWorkloadRenderer` and moves Grove-specific workload rendering out of the common `DynamoGraphDeployment` controller.

It advances the composition-first design by giving the Grove pathway a clearer internal structure:

1. `groveProgram` orchestrates the complete Grove pathway.
2. `groveWorkloadRenderer` resolves render inputs and constructs the desired `PodCliqueSet`.
3. `groveProgram` synchronizes that object and aggregates its readiness.
4. The common controller no longer owns Grove rendering or compatibility transformations.

## Changes

- Add `groveWorkloadRenderer` with explicit dependencies:
  - Kubernetes client
  - operator configuration
  - runtime configuration
  - Docker secret retriever
- Introduce explicit `groveRenderInputs` containing:
  - the DGD used for rendering
  - the existing `PodCliqueSet`, when present
- Move Grove render preparation and compatibility handling into the renderer:
  - legacy worker-selector compatibility
  - restart annotation preservation
  - topology-constraint upgrades
  - clique and group ordering preservation
  - managed replica preservation
- Make `groveProgram` responsible for:
  - resolving Grove render inputs
  - requesting the desired `PodCliqueSet`
  - synchronizing the `PodCliqueSet`
  - wrapping the synchronized object for readiness aggregation
- Remove the corresponding Grove-specific rendering methods from the common graph controller.
- Update existing characterization and program tests to exercise the new boundary.

## Design notes

This is deliberately an intermediate boundary.

The existing `GenerateGrovePodCliqueSet` implementation still performs some Kubernetes-backed input resolution. The new renderer therefore does not claim to be a pure rendering function yet. Instead, it makes those dependencies and responsibilities explicit so resolution and pure rendering can be separated cleanly in a follow-up.

The renderer does not reconcile resources, register watches, manage finalizers, or write status. Those orchestration responsibilities remain with `groveProgram` and the top-level reconciler.

## Behavior

This PR is intended to preserve existing behavior. In particular, it retains:

- legacy worker identity compatibility
- native Grove worker identity behavior
- topology-constraint repair during upgrades
- existing `PodCliqueSet` ordering
- replica counts managed by rolling updates
- restart annotations
- existing readiness aggregation and error handling

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12263" target="_blank">
  <picture>
[...]

PR #12283, opened 2026-07-28
refactor(operator): implement composition-first DGD reconciliation

julienmancuso:
## Summary

Complete the composition-first refactor of `DynamoGraphDeployment` reconciliation while preserving the existing workload pathways and Kubernetes behavior.

This pull request consolidates the full series into one reviewable change. Component/DCD and Grove remain separate, complete workload programs, but both now compose focused reconcilers with explicit dependencies instead of delegating back to the all-capable `DynamoGraphDeploymentReconciler`.

The outer DGD reconciler is reduced to lifecycle ownership: load and finalize the DGD, select one workload program, invoke it once, persist its authoritative status, and emit queued status-transition events after persistence succeeds.

### Establish a small program contract

- Introduce a workload-program request containing only the mutable DGD.
- Return `ctrl.Result`, complete desired status, and queued status-transition events in the program result.
- Preserve intentional non-status mutations on the request DGD while accumulating status separately.
- Keep the outer reconciler as the only DGD status-subresource writer.
- Preserve meaningful partial status when a later reconciliation step fails.
- Emit status-transition events only after status persistence succeeds and direct resource-mutation events only after the corresponding semantic mutation succeeds.

### Compose complete, separate workload programs

- Keep Component/DCD and Grove as separate programs that own their own order and timing.
- Keep the existing one-method program-selection boundary.
- Avoid a shared provider lifecycle, renderer lifecycle, rollout abstraction, or mutable facts/request bag.
- Pass concrete domain inputs and typed observations directly between collaborators.

The Component program owns DCD generation and synchronization, managed worker rollout, restart progress, checkpoint startup policy, readiness aggregation, and component-path cleanup.

The Grove program owns PodCliqueSet rendering and synchronization, Grove scaling, stable services and ingress resources, restart progress, coherent readiness observation, topology-condition projection, and Grove child watches.

### Decompose shared DGD reconciliation

Replace the former monolithic program-input flow with focused reconcilers for:

- RBAC;
- PVCs;
- discovery resources;
- GMS ResourceClaimTemplates;
- checkpoints;
- EPP resources;
- the wait-for-leader ConfigMap;
- MPI SSH keys;
- scaling adapters.

The shared-resources reconciler preserves the existing execution order and returns focused, typed checkpoint observations for later workload steps.

### Narrow rendering and provider boundaries

- Make DCD and Grove workload rendering read-only through `client.Reader` dependencies.
- Move checkpoint-storage writes out of rendering and into dedicated reconciliation.
- Replace Grove-wide render-facts, workload-model, and render-request bags with direct inputs and local temporary values.
- Isolate Grove rendering, scaling, stable-resource reconciliation, readiness interpretation, topology projection, and watch setup behind focused collaborators.
- Preserve Grove ordering, topology migration, restart annotations, and scale-subresource-owned replica values.
- Preserve scaling-adapter-owned replicas after adapter creation while allowing the DGD replica count to seed new adapters.

### Make source and test ownership explicit

- Keep program code focused on construction, sequencing, and result projection.
- Place focused tests alongside the reconcilers they exercise while retaining end-to-end controller coverage.
- Add scoped agent guidance documenting the composition, status, event, capability, and ownership boundaries future changes must preserve.

## Status and event ownership

The selected program returns the complete desired status for the reconciliation attempt, including:

- state and component status;
- checkpoint status;
- restart and rolling-update status;
- the final `Ready` condition;
- `ObservedGeneration`;
- provider-specific conditions such as Grove topology availability.

The outer reconciler persists that status once. Status-transition events are queued by the program and emitted only after persistence succeeds. Events for child-resource mutations are emitted only when the corresponding create, update, patch, or delete succeeds and produces a semantic change.

## Behavior and compatibility

This is an internal architecture refactor. It does not intentionally change:

- DGD APIs, CRDs, or RBAC permissions;
- workload-path selection or feature-gate behavior;
- DCD, LWS, or Grove workload semantics;
- rendered names, labels, annotations, selectors, or ownership;
- managed rolling-update eligibility or phase behavior;
- restart and checkpoint startup policies;
- service, ingress, and virtual-service generation;
- Grove readiness, topology, scaling, cleanup, or significant-watch behavior;
- the single event-driven reconciliation model.

Component and Grove remain separate complete workload programs. No cross-provider lifecycle or capability-driven rollout algorithm is introduced.

## Review guidance

A useful review order is:

1. the workload-program request/result boundary and outer status/event ownership;
2. Component and Grove program construction and sequencing;
3. shared-resource reconciliation and its focused child reconcilers;
4. Component workload, restart, and rollout collaborators;
5. Grove workload, scaling, readiness, topology, and watch collaborators;
6. focused tests and the scoped architectural agent guidance.

## Review history

This pull request contains the cumulative work previously presented through:

- #12115 — read-only DCD rendering boundary;
- #12120 — Component-program ownership;
- #12121 — authoritative program status and event contract;
- #12261 — complete Grove-program ownership;
- #12263 — read-only Grove rendering boundary;
- #12283 — Grove collaborators and ownership corrections;
- #12502 — focused DGD reconcilers and complete removal of all-capable program calls.

GitHub marked #12502 merged after its head became fully contained by this pull request.

## Validation

Combined head `6773306ee9` passed:

```bash
GOCACHE=/tmp/dynamo-go-cache go test ./internal/controller -count=1

docker buildx build --platform linux/arm64 --target linter --progress=plain --build-context snapshot=../snapshot .

docker buildx build --platform linux/arm64 --target tester --progress=plain --build-context snapshot=../snapshot --build-context operator-chart=../helm/charts/platform/components/operator .
```

The ARM64 tester runs the complete operator `make test` target, including formatting, vetting, generated-resource checks, envtest-backed tests, and coverage. Controller-package coverage is 73.9%.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12283" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added automatic checkpoint storage provisioning when checkpointing is enabled.
  - Improved checkpoint restoration, including PVC-backed restores and GPU-related configuration.
  - Added enhanced Grove workload scaling, readiness tracking, topology status, and resource preservation.
  - Added support for more consistent workload rendering across component and Grove deployments.

- **Bug Fixes**
  - Improved reconciliation behavior for missing or temporarily unavailable workload resources.
  - Preserved existing replica counts, ordering, restart settings, and legacy deployment compatibility during updates.

- **Tests**
  - Expanded coverage for checkpoint restoration, scaling, readiness, upgrades, rollouts, and status reporting.


deploy/operator/internal/controller/dynamographdeployment_grove_status.go
```diff
@@ -0,0 +1,219 @@
...
+		condition = metav1.Condition{
+			Type:    nvidiacomv1beta1.ConditionTypeTopologyLevelsAvailable,
+			Status:  metav1.ConditionTrue,
+			Reason:  nvidiacomv1beta1.ConditionReasonAllTopologyLevelsAvailable,
+			Message: "All required topology levels are available in the cluster topology",
+		}
+	}
+
+	meta.SetStatusCondition(&status.Conditions, condition)
+}
```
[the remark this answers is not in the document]
julienmancuso: Confirmed: DynamoGraphDeployment.AddStatusCondition is only a wrapper around meta.SetStatusCondition(&dgd.Status.Conditions, condition), so the old and new implementations have identical condition-level ObservedGeneration behavior. Neither implementation populated it, and there are no downstream consumers relying on it for TopologyLevelsAvailable. The DGD-level ObservedGeneration remains managed by the outer reconciler. The stale-condition removal when topology constraints are removed is intentional.

sttts (commented):
**Disclosure: This review was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

This is an update to [the architectural review on #12263 at `fc2e40a`](https://github.com/ai-dynamo/dynamo/pull/12263#pullrequestreview-4828802055), after reading the incremental `b09d873` commit in this PR.

## What this PR improves

This is a meaningful step in the proposed direction:

- `groveProgram` now owns concrete renderer, scaler, and status collaborators.
- Grove watch setup is separated from the common controller.
- Read-only paths use `client.Reader` instead of `client.Client`.
- Readiness is observed once from a coherent post-scaling snapshot.
- Grove topology status is accumulated in the returned program status; the outer reconciler remains the status writer.
- Kubernetes-backed retrieval and PodCliqueSet construction are now mechanically separated.

The new `groveScaler`, `groveStatusResolver`, and `groveWatchSetup` demonstrate the useful part of the approach: concrete responsibilities with narrower capabilities.

## 1. Keep the retrieval/rendering boundary, but avoid facts and request bags

The capability boundary is good. The resulting data abstraction is not.

`GrovePodCliqueSetRenderFactsRequest`, `GrovePodCliqueSetRenderFacts`, `GrovePodCliqueSetRenderRequest`, and `groveRenderInputs` divide one concrete operation into several structureless carrier types. They make the call graph more abstract without giving the values a stronger domain invariant.

`checkpoint.ResolvedPodSpecRestore` is different: it represents one cohesive, validated domain result and prevents callers from depending on checkpoint-storage internals. That is a useful resolved type. A generic collection of unrelated “facts” is not.

I would keep `groveWorkloadRenderer` as the concrete boundary, inject `client.Reader`, retrieve the required objects directly, keep temporary collections local, and pass concrete values to pure leaf functions:

```go
type groveWorkloadRenderer struct {
    reader                client.Reader
    config                *configv1alpha1.OperatorConfiguration
    runtimeConfig         *commoncontroller.RuntimeConfig
    dockerSecretRetriever dockerSecretRetriever
}

func (r *groveWorkloadRenderer) Render(
    ctx context.Context,
    dgd *nvidiacomv1beta1.DynamoGraphDeployment,
    restart *dynamo.RestartState,
    checkpoints map[string]*checkpoint.CheckpointInfo,
) (*grovev1alpha1.PodCliqueSet, error) {
    existing := &grovev1alpha1.PodCliqueSet{}
    key := types.NamespacedName{
        Name:      dynamo.PCSNameForDGD(dgd.Name, dgd.Spec.Components),
        Namespace: dgd.Namespace,
    }
    if err := r.reader.Get(ctx, key, existing); err != nil {
        if !apierrors.IsNotFound(err) {
            return nil, fmt.Errorf("get PodCliqueSet %s: %w", key, err)
        }
        existing = nil
    }

    renderDGD := dgd.DeepCopy()
    applyGroveCompatibility(renderDGD, existing)

    // Resolve queue, topology, and checkpoint inputs here. Keep collections
    // local unless an individual value has a cohesive domain contract.
    // Then call pure rendering helpers with the actual inputs.
    // ...
}
```

The goal is a hard read/write capability boundary, not purity at the price of argument-bag abstractions. There is also no need for a custom one-method reader interface or a helper that merely renames `reader.Get`.

## 2. Continue replacing method-name composition with struct composition

The Grove-specific extractions are good examples, but the program still retains the complete `DynamoGraphDeploymentReconciler` and still composes large responsibilities through it:

```go
p.reconciler.reconcileProgramInputs(...)
p.reconciler.resolveProgramRestartState(...)
p.reconciler.reconcileProgramResult(...)
p.reconcileStableResources(...)
```

I would continue until the struct graph itself explains the composition:

```go
type groveProgram struct {
    rollout         *dgdUnsupportedRolloutReconciler
    sharedResources *dgdSharedResourcesReconciler
    restart         *dgdRestartReconciler
    workloads       *groveWorkloadReconciler
    stableResources *dgdStableResourcesReconciler
    topology        *dgdGroveTopologyConditionReconciler
}
```

Each should be independently constructible and testable with only its own fixtures. Use short standardized method names such as `Reconcile`, `Resolve`, and `Render`; let the concrete type identify the purpose. Separate files should use a consistent DGD prefix, for example `dgd_checkpoints_reconciler.go` and `dgd_restart_reconciler.go`.

In particular, `reconcileProgramInputs` still effectfully reconciles RBAC, PVCs, discovery, GMS claims, checkpoints, EPP resources, the wait-for-leader ConfigMap, and MPI SSH keys. These are the next concrete reconcilers to extract. The criterion is cohesive resource ownership and an independent fixture contract, not mechanical fragmentation.

## 3. Finish the Request/Result and status-ownership boundary

The empty `resolvedFacts` and unused `Request.Facts` still exist. Remove them until a real common input appears.

`workloadProgramResult` also still splits `Status` from `ReadyReason` and `ReadyMessage`, while the outer reconciler computes and overrides Ready based on rollout state. That leaves condition ownership shared.

The contract should remain:

```go
type Request struct {
    // Mutable primary object. Programs may persist non-status mutations.
    DGD *nvidiacomv1beta1.DynamoGraphDeployment
}

type Result struct {
    ctrl.Result
    Status nvidiacomv1beta1.DynamoGraphDeploymentStatus
    Events []Event
}
```

The selected program should return the complete desired status, including the final Ready condition, rollout/restart state, and `ObservedGeneration`. The outer reconciler should persist that status unchanged.

This PR moves topology projection in that direction. The same ownership rule should now be applied to Ready instead of returning a reason/message that another layer later interprets and overrides.

## 4. Return status-transition events with the status

`groveStatusResolver` now writes into the returned status, but it also owns an `EventRecorder` and emits the topology warning before the outer `Status().Update` succeeds.

That event describes a DGD status transition. It should therefore be queued in `Result.Events` and emitted only after the corresponding status update succeeds and actually changes persisted status. Otherwise an update failure can produce an event for an edge that was not persisted, and retries can emit it again.

Direct event creation remains fine after a locally detected semantic resource mutation successfully completes. The distinction is where the edge becomes real:

- resource mutation edge: emit after the successful `Create`, `Update`, `Patch`, or `Delete`, and only when something actually changed;
- DGD status edge: return it with the status, persist the status, then emit it.

## Direction

This PR materially improves Grove isolation and read-side dependency boundaries. It does not yet complete the program abstraction described in the earlier review.

The main remaining work is:

- remove the empty common facts container;
- avoid Grove-wide facts/request bags;
- make the returned status authoritative, including Ready;
- queue status-transition events and emit them after persistence;
- replace the remaining all-powerful reconciler calls with concrete nested reconcilers;
- decompose `reconcileProgramInputs` into cohesive DGD reconcilers.

DCD and Grove should remain separate workload programs by design. No renderer or provider lifecycle composition between them is being proposed. The shared abstraction is the small program Request/Result contract; each program continues to own its own order and timing.

deploy/operator/internal/controller/dynamographdeployment_grove_watches.go
```diff
@@ -0,0 +1,240 @@
...
+		Watches(
+			&grovev1alpha1.PodCliqueScalingGroup{},
+			handler.EnqueueRequestsFromMapFunc(s.mapPodCliqueScalingGroupToRequests),
+			builder.WithPredicates(pcsgEventPredicates()),
+		)
+}
+
+func podCliqueEventPredicates() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(event.CreateEvent) bool { return true },
```
sttts: **Disclosure: This comment was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

Compared with #12263, this changes the PodClique and PodCliqueScalingGroup create/delete predicates from ignoring those events to enqueueing the DGD. In particular, creation now provides a new retry edge after scaling observes a child as not found, while deletion can cause immediate recreation. This is observable controller behavior rather than only an extraction. Is that change intentional?

deploy/operator/internal/controller/dynamographdeployment_grove_watches.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// podCliqueStatusChangeIsSignificant mirrors every PodClique field consumed by
+// Grove readiness and capacity classification.
+func podCliqueStatusChangeIsSignificant(
+	oldPodClique *grovev1alpha1.PodClique,
+	newPodClique *grovev1alpha1.PodClique,
+) bool {
+	return oldPodClique.Status.ReadyReplicas != newPodClique.Status.ReadyReplicas ||
+		oldPodClique.Generation != newPodClique.Generation ||
```
sttts: **Disclosure: This comment was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

This adds generation-only PodClique and PodCliqueScalingGroup updates to the significant-event set. That broadens the controller’s wake-up behavior compared with #12263 even when none of the previously observed readiness fields has changed. Is this intentional?

deploy/operator/internal/controller/dynamographdeployment_grove_status.go
```diff
@@ -0,0 +1,219 @@
...
+func (r *groveStatusResolver) projectTopologyCondition(
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+	status *nvidiacomv1beta1.DynamoGraphDeploymentStatus,
+) {
+	if status == nil {
+		return
+	}
+	if !dgd.HasAnyTopologyConstraint() {
+		meta.RemoveStatusCondition(
```
sttts: **Disclosure: This comment was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

Previously, removing topology constraints made topology projection a no-op and could leave the existing `TopologyLevelsAvailable` condition in status. This now actively removes that condition, so it is a persisted status-semantics change rather than only moving the old code. Is that intentional?

deploy/operator/internal/controller/dynamographdeployment_grove_program.go
```diff
@@ -106,39 +118,69 @@ func (p *groveProgram) Reconcile(
...
+	stableResources, err := p.reconcileStableResources(
+		ctx,
+		dynamoDeployment,
+		renderInputs.DGD,
+	)
+	if err != nil {
+		return ReconcileResult{}, err
+	}
+
+	grovePodCliqueSetAsResource, groveReadiness, err := p.observePodCliqueSetReadiness(
```
sttts: **Disclosure: This comment was written by Codex as its interpretation of a discussion with Stefan; it is not Stefan’s text.**

This changes observation semantics: #12263 evaluated Grove resource readiness and its Ready-reason classification through separate reads, whereas this PR uses one snapshot taken after PodCliqueSet synchronization, scaling, and stable-resource reconciliation. Under concurrent child updates, the resulting status may therefore differ from the previous implementation. Is that intentional?

deploy/operator/internal/controller/dynamographdeployment_grove_watches.go
julienmancuso: No, that behavior change was not intentional. I restored the PodClique and PodCliqueScalingGroup create/delete predicates to false, preserving the previous behavior, and added predicate coverage.

deploy/operator/internal/controller/dynamographdeployment_grove_watches.go
julienmancuso: No, generation-only wakeups were not intentional. I removed the object generation comparisons for both resource types. Status observed-generation and actual readiness changes remain significant, and the tests now assert that generation-only updates are filtered.

deploy/operator/internal/controller/dynamographdeployment_grove_status.go
julienmancuso: No, removing the existing condition was not intended in this structural refactor. Topology projection is again a no-op when no topology constraint is present, preserving the previous condition, with regression coverage.

deploy/operator/internal/controller/dynamographdeployment_grove_program.go
julienmancuso: Yes, this one is intentional. Grove readiness is observed once after PodCliqueSet synchronization, scaling, and stable-resource reconciliation so that readiness classification, the Ready condition, and component status all come from one coherent snapshot rather than potentially inconsistent concurrent observations.

sttts (approved):

galletas1712 (approved):

PR #12329, opened 2026-07-28
perf(mocker): move request KV ownership into backend leases

jthomson04:
## Summary

- introduce move-only, backend-specific request leases: native vLLM/TRT owns logical and physical block state in `BlockRequestLease`, while SGLang owns page-native state and incremental page hashes in `RadixRequestLease`
- remove native `MoveBlock`, manager UUID block-table lookups, native PLH construction, partial-block UUID tracking, and SGLang page-to-token expansion from the new hot paths
- keep request token/progress state lightweight, preserve bounded vector capacity across lease construction, and reserve vLLM prefix-hit/SGLang page-hash storage up front
- make raw-sink token-ID behavior consistent across native G1 and KVBM, retain SGLang page hashes across retract/re-admit, and harden request lifecycle invariants
- bound SGLang eager token/hash storage by remaining physical KV capacity and, for unplanned outputs, `clip_max_new_tokens`; planned DynoSim outputs retain their known-size reservation within the physical limit
- remove token-rescanning SGLang test adapters, forbid native rollback, avoid disabled-prefix-caching finalization work, and add focused prefix-reuse/admission regressions
- remove cold-prefix reservation over-allocation, make native finalization checks incremental in debug builds, and cover Native/KVBM preemption and disaggregated ownership lifecycles symmetrically
- keep legacy KVBM behavior isolated and preserve aggregate/disaggregated request lifecycle semantics

This PR is directly based on refreshed `main`. The five pre-existing PR commits were rebased onto `d2e3f2f6a4`; `git range-diff` verified that all five remained patch-identical before the PeaBrane review-fix commit was added.

The larger `G1Manager` dispatcher split remains intentionally deferred to #12340.

### Compatibility

The shared vLLM/TRT-LLM scheduler now requires `block_size >= 2` for Native, explicit KVBM, and automatically resolved KVBM. SGLang continues to accept block size one. Supporting KVBM block size one is intentionally deferred rather than expanding legacy `ActiveSequence` in this PR.

## Validation

- `cargo test -p dynamo-mocker --lib` — 683 passed
- `cargo test -p dynamo-mocker --all-features --lib` — 812 passed
- `cargo clippy -p dynamo-mocker --lib --all-targets -- -D warnings`
- `git diff --check`
- focused regressions cover block-size-two Native/KVBM request construction for vLLM and TRT-LLM, block-size-one normalization rejection for every shared-scheduler backend configuration, zero-capacity cold/empty prefix reservations, partial/full prefix reuse, Native/KVBM preemption and re-admission, the full source-hold/destination reserve→activate→admit→release lifecycle, and exact raw-sink payload `[[0, 1, 2, 3]]` from an unaligned prompt

### DynoSim correctness parity against refreshed main

Correctness-only parity compared baseline `d2e3f2f6a405657adec175e5dc63610abe233033` with candidate `0baf13eaa8cf69d3f51547162e7db79bfe4c748d` on the fixed contiguous 5,000-request Mooncake corpus. Every row ran two independent processes per revision. Both runs within each revision were byte-stable, baseline and candidate canonical digests matched exactly, all requests completed, and lifecycle evidence matched.

| Backend / topology | Canonical SHA-256 | Handoffs | Preemptions | G1→G2 reserved/completed | G2→G1 reserved/completed |
|---|---|---:|---:|---:|---:|
| SGLang aggregate | `e522f4835ac9df7f982def3c05a1a79b3e4e94dd3a727b6ed08e36688ff4d87a` | 0 | 0 | 0 / 0 | 0 / 0 |
| SGLang disaggregated | `40f414ca264abb065b33ed51b10370e5786eb37ff1aa4a4a77b27aca3e506b0f` | 5,000 | 0 | 0 / 0 | 0 / 0 |
| vLLM aggregate, KVBM stress | `dd7af53c6fa32fab6ba44510c0a7587175de1cd20715f526c33585e3e06d309a` | 0 | 3 | 6,479 / 6,479 | 135 / 135 |
| vLLM disaggregated, KVBM stress | `9e89fa875e0689ecc24a9b0dd6894bda989e67799588d42dd0f76a165a71051d` | 5,000 | 2 | 16,459 / 16,459 | 307 / 307 |

No performance benchmark or profile was rerun for this review pass. The production hot-path change removes unnecessary cold-prefix allocation, while the invariant optimization affects debug/test builds only. The existing isolated ownership-refactor A/B below remains the relevant performance measurement.

### DynoSim aggregate benchmark

AgentX replay, 16 workers / 16 sessions, block size 16, CPU 0 affinity. Baseline and candidate were freshly built from the same runner, lockfile, configs, toolchain, and release flags (`x86-64-v3`, frame pointers, DWARF enabled).

The 16/16 A/B parity gate compared the complete DynoSim reports, not only replay time and memory. Both arms processed exactly 3,211 requests, 479,843,008 input tokens, 3,552,472 output tokens, and 483,395,480 total tokens, and every simulated/non-wall-clock report field was identical.

The benchmark was collected before the history-only rebase: baseline `9e0a8cc6d6` was rebased #12248 plus the native resolved-backend fix, and measured candidate `897efbec73` contains the same three request-lease commits ending at `db98b92fdf`. Those ownership-refactor patches remain unchanged in the current history, as verified by `git range-diff`; the subsequent review-fix commits were not part of the measured candidate.

| Backend | Baseline replay | Candidate replay | Replay delta | Baseline peak RSS | Candidate peak RSS | RSS delta |
|---|---:|---:|---:|---:|---:|---:|
| vLLM | 3.971660 s | 3.029880 s | -23.71% | 215.961 MiB | 205.500 MiB | -4.84% |
| SGLang | 3.039690 s | 2.370373 s | -22.02% | 173.023 MiB | 164.773 MiB | -4.77% |

Candidate `.text` size is 2,647,557 bytes versus 2,652,773 bytes for baseline (-0.20%).

These are deliberately one-shot results per arm; no CPU profiles were collected.

### DIFF
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -114,17 +114,7 @@ impl VllmBlockPool {
             return self.reserve_fresh(fresh);
         }
 
-        let hits = prefix
-            .iter()
-            .map(|hash| {
-                let Some(id) = self.first_copy(*hash) else {
-                    panic!("authorized prefix hash {hash} is no longer resident")
-                };
-                (*hash, id)
-            })
-            .collect::<Vec<_>>();
-
-        self.reserve_hits(hits, fresh)
+        self.reserve_exact_prefix(prefix.iter().copied(), prefix.len() + fresh)
     }
 
     /// Resolve and pin the longest resident prefix from `candidates`, then
@@ -134,7 +124,16 @@ impl VllmBlockPool {
         candidates: impl IntoIterator<Item = SequenceHash>,
         total: usize,
     ) -> Option<ReserveOutcome> {
-        let mut hits = Vec::with_capacity(total);
+        let mut candidates = candidates.into_iter();
+        let Some(first_hash) = candidates.next() else {
+            return self.reserve_fresh(total);
+        };
+        assert!(total > 0, "prefix candidates exceed layout");
+        let Some(first_id) = self.first_copy(first_hash) else {
+            return self.reserve_fresh(total);
+        };
+
+        let mut hits = vec![(first_hash, first_id)];
         for hash in candidates {
             assert!(hits.len() < total, "prefix candidates exceed layout");
             let Some(id) = self.first_copy(hash) else {
@@ -151,12 +150,27 @@ impl VllmBlockPool {
     /// Unlike [`Self::reserve_resident_prefix`], every candidate must still be
     /// resident. A missing hash means the caller's synchronous prefix
     /// authorization changed before allocation committed.
-    pub(crate) fn reserve_exact_prefix(
+    pub(crate) fn reserve_exact_prefix<I>(
         &mut self,
-        candidates: impl IntoIterator<Item = SequenceHash>,
+        candidates: I,
         total: usize,
-    ) -> Option<ReserveOutcome> {
-        let mut hits = Vec::with_capacity(total);
+    ) -> Option<ReserveOutcome>
+    where
+        I: IntoIterator<Item = SequenceHash>,
+        I::IntoIter: ExactSizeIterator,
+    {
+        let mut candidates = candidates.into_iter();
+        let candidate_count = candidates.len();
+        let Some(first_hash) = candidates.next() else {
+            return self.reserve_fresh(total);
+        };
+        assert!(candidate_count <= total, "prefix candidates exceed layout");
+        let Some(first_id) = self.first_copy(first_hash) else {
+            panic!("authorized prefix hash {first_hash} is no longer resident")
+        };
+        let mut hits = Vec::with_capacity(candidate_count);
+        hits.push((first_hash, first_id));
+
         for hash in candidates {
             assert!(hits.len() < total, "prefix candidates exceed layout");
             let Some(id) = self.first_copy(hash) else {
@@ -659,6 +673,30 @@ mod tests {
     }
 
     #[test]
+    fn empty_exact_prefix_reserves_fresh_without_prefix_storage() {
+        let mut pool = VllmBlockPool::new(3);
+        let outcome = pool
+            .reserve_exact_prefix(std::iter::empty(), 3)
+            .expect("fresh capacity should fit");
+
+        assert_eq!(outcome.reservation.prefix.capacity(), 0);
+        assert_eq!(outcome.reservation.fresh_len(), 3);
+        pool.cancel(outcome.reservation);
+    }
+
+    #[test]
+    fn cold_resident_prefix_reserves_fresh_without_prefix_storage() {
+        let mut pool = VllmBlockPool::new(3);
+        let outcome = pool
+            .reserve_resident_prefix([7, 8, 9], 3)
+            .expect("fresh capacity should fit");
+
+        assert_eq!(outcome.reservation.prefix.capacity(), 0);
+        assert_eq!(outcome.reservation.fresh_len(), 3);
+        pool.cancel(outcome.reservation);
+    }
+
+    #[test]
     fn duplicate_hashes_consume_distinct_capacity_but_share_visibility() {
         let mut pool = VllmBlockPool::new(2);
         let mut first = reserve(&mut pool, &[], 1).reservation;
@@ -704,6 +742,7 @@ mod tests {
         assert!(outcome.removed.is_empty());
         assert_eq!(outcome.reservation.len(), 2);
         assert_eq!(outcome.reservation.fresh_len(), 1);
+        assert_eq!(outcome.reservation.prefix.capacity(), 1);
         pool.cancel(outcome.reservation);
         assert_eq!(pool.num_inactive(), 1);
         pool.assert_lru_consistent();
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -691,6 +691,25 @@ mod tests {
     }
 
     #[test]
+    fn resident_prefix_stops_at_first_miss_and_reserves_fresh_suffix() {
+        let mut pool = VllmBlockPool::new(2);
+        let mut seed = reserve(&mut pool, &[], 1).reservation;
+        let id = pool.allocate_private(&mut seed);
+        assert!(pool.cache_private(id, 7));
+        pool.release(id);
+
+        let outcome = pool
+            .reserve_resident_prefix([7, 9], 2)
+            .expect("one resident prefix plus one fresh block should fit");
+        assert!(outcome.removed.is_empty());
+        assert_eq!(outcome.reservation.len(), 2);
+        assert_eq!(outcome.reservation.fresh_len(), 1);
+        pool.cancel(outcome.reservation);
+        assert_eq!(pool.num_inactive(), 1);
+        pool.assert_lru_consistent();
+    }
+
+    #[test]
     fn removal_is_reported_only_for_the_last_physical_copy() {
         let mut pool = VllmBlockPool::new(2);
         let mut first = reserve(&mut pool, &[], 1).reservation;
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -146,6 +146,28 @@ impl VllmBlockPool {
         self.reserve_hits(hits, fresh)
     }
 
+    /// Pin an already-authorized prefix and reserve the remaining entries.
+    ///
+    /// Unlike [`Self::reserve_resident_prefix`], every candidate must still be
+    /// resident. A missing hash means the caller's synchronous prefix
+    /// authorization changed before allocation committed.
+    pub(crate) fn reserve_exact_prefix(
+        &mut self,
+        candidates: impl IntoIterator<Item = SequenceHash>,
+        total: usize,
+    ) -> Option<ReserveOutcome> {
+        let mut hits = Vec::with_capacity(total);
+        for hash in candidates {
+            assert!(hits.len() < total, "prefix candidates exceed layout");
+            let Some(id) = self.first_copy(hash) else {
+                panic!("authorized prefix hash {hash} is no longer resident")
+            };
+            hits.push((hash, id));
+        }
+        let fresh = total - hits.len();
+        self.reserve_hits(hits, fresh)
+    }
+
     fn reserve_hits(
         &mut self,
         hits: Vec<(SequenceHash, BlockCopyId)>,
@@ -627,6 +649,13 @@ mod tests {
     fn reserve(pool: &mut VllmBlockPool, prefix: &[u64], fresh: usize) -> ReserveOutcome {
         pool.reserve(prefix, fresh)
             .unwrap_or_else(|| panic!("unexpected capacity exhaustion"))
+    }
+
+    #[test]
+    #[should_panic(expected = "authorized prefix hash 9 is no longer resident")]
+    fn exact_prefix_reports_missing_hash() {
+        let mut pool = VllmBlockPool::new(1);
+        let _ = pool.reserve_exact_prefix([9], 1);
     }
 
     #[test]
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -134,7 +134,7 @@ impl VllmBlockPool {
         candidates: impl IntoIterator<Item = SequenceHash>,
         total: usize,
     ) -> Option<ReserveOutcome> {
-        let mut hits = Vec::new();
+        let mut hits = Vec::with_capacity(total);
         for hash in candidates {
             assert!(hits.len() < total, "prefix candidates exceed layout");
             let Some(id) = self.first_copy(hash) else {
@@ -746,7 +746,10 @@ mod tests {
         pool.release(id);
 
         let mut activation = reserve(&mut pool, &[7], 0).reservation;
-        assert_eq!(pool.activate_prefix(&mut activation), vec![id]);
+        assert_eq!(
+            pool.activate_prefix(&mut activation).collect::<Vec<_>>(),
+            vec![(7, id)]
+        );
         pool.cancel(activation);
         assert_eq!(pool.num_inactive(), 0);
         pool.assert_lru_consistent();
--- a/lib/mocker/src/cache/vllm_block_pool.rs
+++ b/lib/mocker/src/cache/vllm_block_pool.rs
@@ -124,6 +124,33 @@ impl VllmBlockPool {
             })
             .collect::<Vec<_>>();
 
+        self.reserve_hits(hits, fresh)
+    }
+
+    /// Resolve and pin the longest resident prefix from `candidates`, then
+    /// reserve the remaining entries as fresh capacity in one traversal.
+    pub(crate) fn reserve_resident_prefix(
+        &mut self,
+        candidates: impl IntoIterator<Item = SequenceHash>,
+        total: usize,
+    ) -> Option<ReserveOutcome> {
+        let mut hits = Vec::new();
+        for hash in candidates {
+            assert!(hits.len() < total, "prefix candidates exceed layout");
+            let Some(id) = self.first_copy(hash) else {
+                break;
+            };
+            hits.push((hash, id));
+        }
+        let fresh = total - hits.len();
+        self.reserve_hits(hits, fresh)
+    }
+
+    fn reserve_hits(
+        &mut self,
+        hits: Vec<(SequenceHash, BlockCopyId)>,
+        fresh: usize,
+    ) -> Option<ReserveOutcome> {
         let free = self.free_capacity();
         let needed_evictions = fresh.saturating_sub(free);
         if needed_evictions > 0 {
@@ -187,14 +214,12 @@ impl VllmBlockPool {
     pub(crate) fn activate_prefix(
         &mut self,
         reservation: &mut BlockReservation,
-    ) -> Vec<BlockCopyId> {
+    ) -> std::vec::IntoIter<(SequenceHash, BlockCopyId)> {
         let prefix = std::mem::take(&mut reservation.prefix);
-        let mut ids = Vec::with_capacity(prefix.len());
-        for (hash, id) in prefix {
+        for &(hash, id) in &prefix {
             self.activate_pin(id, hash);
-            ids.push(id);
         }
-        ids
+        prefix.into_iter()
     }
 
     pub(crate) fn allocate_private(&mut self, reservation: &mut BlockReservation) -> BlockCopyId {
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -1014,8 +1014,8 @@ fn validate_mock_engine_args(args: &MockEngineArgs) -> Result<(), ValidationErro
         return Err(mock_engine_args_validation_error(
             "native_g1_block_size_too_small",
             format!(
-                "native G1 requires block_size to be at least 2 for engine_type={:?}, got block_size={}",
-                args.engine_type, args.block_size
+                "native G1 requires block_size to be at least 2 for engine_type={:?}, got block_size={}; set g1_backend=kvbm to keep block_size=1",
+                args.engine_type, args.block_size,
             ),
         ));
     }
@@ -2181,21 +2181,27 @@ mod tests {
     #[test]
     fn test_normalized_native_g1_rejects_block_size_one() {
         for engine_type in [EngineType::Vllm, EngineType::Trtllm] {
-            let error = MockEngineArgs::builder()
+            let explicit = MockEngineArgs::builder()
                 .engine_type(engine_type)
                 .g1_backend(G1Backend::Native)
                 .block_size(1)
                 .build()
-                .unwrap()
-                .normalized()
-                .unwrap_err();
+                .unwrap();
+            let mut unset = explicit.clone();
+            unset.g1_backend = None;
 
-            assert!(
-                error
-                    .to_string()
-                    .contains("native G1 requires block_size to be at least 2"),
-                "engine_type={engine_type:?}, error={error:#}"
-            );
+            for args in [explicit, unset] {
+                let error = args.normalized().unwrap_err();
+                let message = error.to_string();
+                assert!(
+                    message.contains("native G1 requires block_size to be at least 2"),
+                    "engine_type={engine_type:?}, error={error:#}"
+                );
+                assert!(
+                    message.contains("set g1_backend=kvbm to keep block_size=1"),
+                    "engine_type={engine_type:?}, error={error:#}"
+                );
+            }
         }
     }
 
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -1007,6 +1007,19 @@ fn validate_mock_engine_args(args: &MockEngineArgs) -> Result<(), ValidationErro
         ));
     }
 
+    if matches!(args.engine_type, EngineType::Vllm | EngineType::Trtllm)
+        && args.resolved_g1_backend() == G1Backend::Native
+        && args.block_size < 2
+    {
+        return Err(mock_engine_args_validation_error(
+            "native_g1_block_size_too_small",
+            format!(
+                "native G1 requires block_size to be at least 2 for engine_type={:?}, got block_size={}",
+                args.engine_type, args.block_size
+            ),
+        ));
+    }
+
     if args.g1_backend == Some(G1Backend::Native) && args.requires_kvbm_g1() {
         return Err(mock_engine_args_validation_error(
             "native_g1_legacy_offload_conflict",
@@ -2163,6 +2176,44 @@ mod tests {
             .unwrap();
 
         assert_eq!(args.block_size, 1);
+    }
+
+    #[test]
+    fn test_normalized_native_g1_rejects_block_size_one() {
+        for engine_type in [EngineType::Vllm, EngineType::Trtllm] {
+            let error = MockEngineArgs::builder()
+                .engine_type(engine_type)
+                .g1_backend(G1Backend::Native)
+                .block_size(1)
+                .build()
+                .unwrap()
+                .normalized()
+                .unwrap_err();
+
+            assert!(
+                error
+                    .to_string()
+                    .contains("native G1 requires block_size to be at least 2"),
+                "engine_type={engine_type:?}, error={error:#}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_normalized_kvbm_accepts_block_size_one() {
+        for engine_type in [EngineType::Vllm, EngineType::Trtllm] {
+            let args = MockEngineArgs::builder()
+                .engine_type(engine_type)
+                .g1_backend(G1Backend::Kvbm)
+                .block_size(1)
+                .build()
+                .unwrap()
+                .normalized()
+                .unwrap();
+
+            assert_eq!(args.block_size, 1, "engine_type={engine_type:?}");
+            assert_eq!(args.resolved_g1_backend(), G1Backend::Kvbm);
+        }
     }
 
     #[test]
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -1007,14 +1007,11 @@ fn validate_mock_engine_args(args: &MockEngineArgs) -> Result<(), ValidationErro
         ));
     }
 
-    if matches!(args.engine_type, EngineType::Vllm | EngineType::Trtllm)
-        && args.resolved_g1_backend() == G1Backend::Native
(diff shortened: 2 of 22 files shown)


jthomson04 (commented):
## Review

**Scope:** this covers only the commits unique to this PR — `f9ffa9a9b7` (move KV ownership into request leases), `d97c842436`, and `897efbec73` (preserve request lease capacity). The flattening work in `b8790f15c0` / `9e0a8cc6d6` is #12248 and I've left it for that PR's review. Where a finding is a *consequence* of the ownership refactor landing on top of #12248, I say so.

---

### Overview

Replaces the `MoveBlock`-signal protocol between the schedulers and their KV managers with move-only, backend-specific request leases:

- **Native vLLM/TRT** — `BlockRequestLease` (owner + `Vec<BlockLeaseEntry>` + `allocated_tokens`) holds logical identity *and* the physical `BlockCopyId` per block. `VllmKvManager` drops `request_blocks: FxHashMap<Uuid, Vec<OwnedBlock>>` and `partial_uuids: FxHashSet<Uuid>` entirely, along with `VllmBlockLayout`, `process_use`/`process_deref`/`process_promote`, and the PLH-alignment validators.
- **SGLang** — `ActiveKvLease` → `RadixRequestLease`, storing `Vec<KvPageId>` + `materialized_tokens` instead of a flattened `Vec<usize>` of token slots, plus incrementally-computed `page_hashes` that survive retraction so re-admission never rescans the prompt.
- `RequestKvState` in the new `scheduler/vllm/request.rs` selects storage once at admission; the `PolicySequence` trait keeps `policy.rs` generic over both.

The core idea is sound and the execution is careful. Allocation is genuinely atomic (`reserve_resident_prefix` → `commit_lease_range` → `cancel`, with the lease untouched on `CapacityExhausted` — nicely pinned by `duplicate_full_hashes_consume_physical_capacity` and `page_native_extension_oom_is_atomic`). Removing page↔index expansion from `collect_path_pages` and `canonicalize_unfinished_pages` is a real structural win, and the page-hash caching is the right fix for the O(n) rescan on retract/re-admit. Consolidating #12248's three `completion_blocks`-sized vectors (`unique_blocks`, `block_hashes`, `plhs`) into one `entries` vector should also account for a chunk of the reported RSS drop.

---

### Issues I'd want addressed before merge

**1. `emit_token_ids` now diverges between native and KVBM** — `scheduler/vllm/core.rs:497`

```rust
let emit_token_ids = kv_event_publishers.raw_enabled() || args.zmq_kv_events_port.is_some();
```

but the KVBM branch still builds `ActiveSequence` with `self.args.zmq_kv_events_port.is_some()`. With a raw sink attached and no ZMQ port, native G1 now emits `block_token_ids` in Stored events and KVBM doesn't — same config, different router-visible payload. If the widening is deliberate (the reworked `vllm_backend` event test asserts `event.block_token_ids == Some(...)`), apply it to both branches so the backends stay observationally equivalent. Worth noting it also flips `FlatTokens` into full-history retention, so #12248's memory win disappears whenever a raw sink is present.

**2. `assert!(block_size >= 2)` is a new runtime panic for a config `MockEngineArgs` accepts** — `common/sequence.rs:141`

`MockEngineArgs` validation only requires `block_size > 0` (`common/protocols.rs:791`), and #12248's `FlatTokens` carried no such constraint. With native G1 (the default for vLLM/TRT) and `block_size: 1`, the first `receive()` panics instead of failing validation. The constraint itself is defensible — at `block_size == 1` every block is full, so `NativeBlockIdentity::partial()` is never constructed while `generate_token()` reports `opened_partial` on every token — but it belongs in `MockEngineArgs::normalized()` so it surfaces as a config error rather than mid-run.

**3. `activate_destination_lease` passes unbounded `page_hashes`** — `kv_manager/sglang_backend.rs:547`

```rust
let new_last_node = self.cache_unfinished_hashes(
    lease.page_hashes(),   // every other call site uses page_hashes_through(n, page_size)
    &mut prefix_pages, last_node, prefix_len,
);
```

Correct today only because `SglangRequest::prepare()` populates hashes through the prompt and `token_count == prompt_len` at activation. But `cache_unfinished_hashes` derives `complete_len = page_hashes.len() * page_size`, so if the lease ever carries hashes past `token_count` this either over-inserts or trips the `complete_len / block_size <= pages.len()` assert. Since the whole point of `page_hashes` is that they now outlive retraction, the bounded accessor is the safer default here — `lease.page_hashes_through(token_count, page_size)`, matching `finish` and `extend_allocation`.

---

### Design

**`G1Manager` has become a 15-`panic!` dispatcher.** `g1_manager.rs` lines 175, 194, 207, 214, 225, 238, 293, 326, 338, 376, 405, 416, 486, 495, 587 are all backend-mismatch panics, and six methods now carry `_owner` / `_reusable_prefix_blocks` / `_sequence` params kept only for signature compatibility. `RequestKvState` already forces every caller to branch on the backend, so these encode an invariant the type system could hold — splitting into `NativeG1`/`KvbmG1` behind the existing enum, or moving dispatch into `RequestKvState`, would delete most of them. Fine as a follow-up, but the current shape is a trap for the next call site.

**Latent trap in the three preemption-cleanup loops** — `core.rs:1997-2006`, `2267-2276`, `2395-2404`. These still call `process_for_request(preempted.uuid, &signal, 0)` unconditionally. Safe today *only* because `reset_legacy_with_signal()` returns `Vec::new()` for `Native`, so the body never executes. Anything that later emits a signal on the native path turns into `panic!("native requests must use direct block-lease operations")` at runtime. A `debug_assert!` on emptiness next to each would make the coupling explicit and cheap to keep true.

**`SglangRequest::prepare()` is a convention-only invariant.** Both `From<DirectRequest>` sites (`scheduler/sglang/core.rs:246-247`, `370-371`) immediately call `prepare()`. If a third construction path forgets it, `reserve_destination_lease` sees empty `page_hashes` → `prefix_len == 0`, and `activate_destination_lease` still sets `cached_tokens = token_count / page_size * page_size` — which makes `release_active_lease` compute `first_owned_page` past the real pages and skip freeing them. Folding `prepare` into a constructor that takes `block_size` (and dropping the `From` impl) makes it unrepresentable.

**Minor:** `activate_destination_lease` (vLLM) doesn't re-check `resident_block_count() == 0` the way `reserve_destination_lease` does; the old code asserted `!request_blocks.contains_key()` on both sides. `commit_lease_range`'s per-entry `assert!(entry.copy.is_none())` covers it, but the explicit precondition read better.

---

### Dead code left behind

The page-native rewrite orphans the flattening helpers — after this PR they have no production callers in the crate:

- `PagePool::expand_pages`, `PagePool::free_indices`, `PagePool::allocate_indices_into` (reachable only through the `#[cfg(test)]` `allocate`/`free` wrappers)
- `RadixCache::insert`, `insert_from_node`, `page_ids` — every remaining caller is inside `#[cfg(test)]`
- `SglangKvManager::allocate_decode_token` / `DecodeTokenReservation::take`, superseded by `take_page`

They survive `-D warnings` only because `cache::radix_cache` is a `pub mod`. The `PagePool` doc comment was rewritten to justify keeping them — *"Compatibility helpers can still flatten page IDs into token indices for existing callers"* — when there are no such callers left. Either delete them or gate them `#[cfg(test)]`; otherwise the next reader reasonably assumes two live representations coexist.

---

### Documentation drift

Both texts were accurate in #12248, where `ActiveSequence` had a flat variant. Making it KVBM-only here is what stales them:

- `common/sequence.rs:599-601` — *"Panics for native flat sequences that were created without token-ID event emission"* on `ActiveSequence::block_token_ids`. That method is never flat now; the caveat belongs on `RequestSequence::block_token_ids`, which is where the `retained_start == 0` assert actually lives.
- `common/sequence.rs:767-769` — *"legacy token storage preserves its historical no-op on an empty buffer, while flat storage panics"* on `ActiveSequence::pop`. There is no flat storage in `ActiveSequence`; the panicking path is `RequestSequence::pop_generated_token`.

Also `NativeBlockIdentity::partial()` is defined but `BlockRequestLease::append_partial` re-inlines the struct literal, because the helper is private to `sequence.rs`. Make it `pub(crate)` and call it.

---

### Test coverage

Good additions: `page_native_extension_oom_is_atomic`, `active_kv_lease_retains_one_id_per_physical_page` (replacing a `size_of` assertion with a behavioural page-count test is a clear improvement), `native_g1_cannot_enter_kvbm_swap_in`, and the widened `flat_storage_capacity_is_bounded_by_realizable_output` case in `897efbec73`. Removing `vllm_boundary_rejects_misaligned_lineage_metadata` is correct now that the PLH validator is gone.

Gaps:

- **Lost KVBM coverage.** `policy/tests.rs::preempted_non_aligned_prompt_uses_complete_known_context` was `for backend in [Kvbm, Native]` and is now Native-only. Non-aligned prompt + retained generation across preemption is exactly the scenario where the two paths could drift; please restore the KVBM arm via `RequestKvState::kvbm`.
- **No retract → re-admit test for SGLang page-hash reuse.** That `page_hashes` survive `release_active_lease` and short-circuit `ensure_page_hashes` on re-admission is the headline SGLang win, and nothing currently pins it — a regression there would be silently slow, not wrong.
- No `#[should_panic]` coverage for any of the 15 backend-mismatch panics. If the claim is that they're unreachable, that's better enforced structurally than left untested.

---

### Smaller notes

- `commit_lease_range` builds `stores` when `cache_fresh && materialize_store_events`, and `activate_destination_lease` now passes `cache_fresh: true` unconditionally (previously `self.enable_prefix_caching`). With prefix caching off you allocate and walk a vec of `None`s for nothing. Pass `self.enable_prefix_caching`, or fold the check into the `stores` guard.
- `insert_page_suffix` asserts `page_ids.len() >= expected_pages`, then `insert_page_hash_suffix` re-asserts `page_ids.len() >= page_keys.len()` where `page_keys.len() == expected_pages` by construction. One of the two is redundant.
- `allocate_lease`'s `assert_eq!(reservation.len() - reservation.fresh_len(), reusable_prefix_blocks, "authorized native prefix changed during atomic allocation")` replaces `reserve()`'s `panic!("authorized prefix hash {hash} is no longer resident")`. Same strictness, but the new message drops the offending hash — worth keeping it for debuggability.
- `preempt_uuid`'s `num_allocated_tokens == 0` debug assertion is now cfg-gated to KVBM only, because `preempt_native` zeroes `allocated_tokens` *after* the state transition. Reordering so the invariant holds for both backends would restore the check.
- `allocate_for_request_lease` uses `debug_assert_eq!(pages.len(), prefix_len / page_size)` on the collected path while `reserve_destination_lease` defensively `truncate`s. Both are fine given `match_prefix_hashes` splits nodes to land exactly on the match, but the two call sites should agree on which contract they're relying on.

---

### Verdict

Structurally this is a good refactor and the perf story is credible. I'd like items **1-3** resolved before merge; the `G1Manager` panic-dispatcher shape and the orphaned flattening helpers are the significant non-blocking ones. The doc drift, restored KVBM test arm, and `NativeBlockIdentity::partial()` are cheap and should ride along.

One request on the validation: both benchmark arms are single-shot with no CPU profiles, and the candidate is the full stack including #12248. Since the two PRs will presumably be judged separately, a `9e0a8cc6d6` → `897efbec73` delta would isolate what the ownership refactor itself buys.

jthomson04:
## Review: `perf(mocker): move request KV ownership into backend leases`

Traced the vLLM allocate → finalize → preempt → finish lifecycle and the SGLang allocate → extend → cache → retract lifecycle end to end. The refactor is sound; I didn't find a correctness break in either path. One memory regression is worth fixing before merge, plus a few follow-up-worthy items.

**Scope note:** commit `be8be336ef` ("resolve G1 backend before comparison") is #12336's fix, but `baseRefName` is `main`, so merging this merges that too. That two-line change is load-bearing here — `resolved_g1_backend()` now selects the *request representation*, not just the manager — so it's worth calling out explicitly or retargeting the base.

---

### Verified correct

Recording these so they don't get re-derived:

- `required_pages = len.div_ceil(ps) - prefix_len/ps` is algebraically identical to the old `(len - prefix_len).div_ceil(ps)` given page-aligned `prefix_len`.
- `extend_decode`'s `materialized_tokens.is_multiple_of(page_size)` matches the old `take(last_index)` boundary, and the new `page_count * block_size == allocated_tokens` debug-assert holds across prefill (`ceil_to_block(chunk_end)`), decode (`crossing_page_boundary`), and destination activation.
- Retained `page_hashes` are safe across retract/re-admit only because `reset_for_retract` doesn't truncate `sequence_tokens` — so the hashes are always a valid prefix. `retract_and_readmit_reuses_cached_page_hashes` pins this.
- `retain_local_hashes` ⊇ the publishing branch of `materialize_store_events()`: `publish_kv_event` returns before the `local_hashes.len() == full_blocks.len()` assert when publishers are empty, so the trace-only path (`KV_CACHE_TRACE_ENABLED`, no sinks) can't emit misaligned events.
- `duplicate_prefix_len` is already page-rounded (`div_ceil(page_size) * page_size`), so `policy.rs`'s `.expect("…at least one page")` can't fire and `[..duplicate_prefix_len / page_size]` can't go out of bounds.
- Precedence in `debug_assert_invariants`' closure (`&&` binds before `||`) gives the intended "only the final block may be partial".
- `reserve_exact_prefix`'s panic matches the pre-existing `VllmBlockPool::reserve` behavior — not a new failure mode.

---

### 1. `SglangRequest::new` eagerly reserves unbounded output capacity

`lib/mocker/src/scheduler/sglang/request.rs`

```rust
let output_capacity = max_output_tokens;
sequence_tokens.reserve_exact(output_capacity);
let completion_pages = (len + output_capacity) / block_size;
kv_lease.reserve_page_hashes(completion_pages);
```

`max_output_tokens` falls back to the caller-supplied `DirectRequest::max_output_tokens` whenever `output_token_ids` is `None`, with no clamp. The vLLM path deliberately doesn't do this — `output_capacity_hint()` clamps by `num_gpu_blocks * block_size - prompt_len` and `max_model_len`, and `flat_storage_capacity_is_bounded_by_realizable_output` guards it. `SglangRequest::new` only receives `block_size`, so the clamp isn't even expressible today.

Concretely: a queue of 1,000 SGLang requests declaring `max_output_tokens: 100_000` now eagerly commits ~400 MB of token buffers plus page-hash storage, on a worker whose KV pool may only admit a few thousand tokens. Previously zero.

The DynoSim numbers can't catch this — AgentX replay supplies `output_token_ids`, so `max_output_tokens == output_token_ids.len()` (realized output). Given this PR is measured on RSS, worth closing before merge: thread an `output_capacity_hint` equivalent (KV pool size, and `max_model_len` if SGLang has one) through `SglangCore::submit`/`from_handoff`, or drop the `reserve_exact` and let the `Vec` grow amortized.

### 2. `G1Manager` panic surface grew substantially

Seven methods now `panic!` on `G1ManagerBackend::Native` (`process_for_request`, `finalize_computed_prefix`, `process_decode_signal_for_request`, `reserve_destination_at`, `activate_destination`, `get_prefill_cost`, `request_block_count`), plus `RequestKvState`'s four `#[cfg(feature = "kvbm-offload")]` accessors on the Native variant.

Deferring the dispatcher split is reasonable for review size — but this is the change that makes it necessary, and messages like `"native requests must use direct block-lease operations"` are documentation of an invariant the compiler could enforce. Suggest at minimum a tracking issue referenced from a module-level comment on `g1_manager.rs`, so the deferral is visible to the next reader.

### 3. Test-only shims decouple tests from the production path

`#[cfg(test)]` now gates `RadixCache::{insert, insert_from_node, page_ids, insert_page_suffix}`, `KvPageId::{from_token_index, first_token_index}`, `PagePool::{allocate_indices_into, expand_pages, free_indices}`, `SglangKvManager::{cache_finished_req, cache_unfinished_req, publish_stored_event, allocate_decode_token, reserve_destination, allocate_for_request, retract}`, and `DecodeTokenReservation::take`.

Several shims do exactly the work production now avoids:

```rust
#[cfg(test)]
fn cache_unfinished_req(&mut self, token_ids: &[u32], pages: &mut [KvPageId], ...) -> NodeId {
    let page_hashes = self.cache.page_hashes(token_ids);   // production uses lease-owned hashes
    self.cache_unfinished_hashes(&page_hashes, pages, last_node, first_new_token)
}
```

So a passing test no longer proves the "hashes are owned and incrementally extended by the lease" contract that production depends on — which is what `ensure_page_hashes` / `page_hashes_through` exist for, and the thing most likely to rot. Suggest converting the highest-value tests (`test_duplicate_completion_releases_unretained_pages_and_removes_on_eviction`, `cache_unfinished_*`, the canonicalization tests) onto the hash API and deleting the corresponding shims, rather than carrying both APIs.

### 4. `rollback_partial` would panic if the Native `pop()` path became reachable

```rust
pub(crate) fn rollback_partial(&mut self) {
    let entry = self.entries.pop().expect("native rollback requires a partial tail");
    assert!(entry.identity.sequence_hash.is_none() && entry.copy.is_none(), ...);
}
```

`RequestKvState::pop()`'s Native arm is unreachable today only because native `generate_token()` returns `Vec::new()`, so `effects.immediate` is empty and the non-speculative decode loop always breaks on `G1Acquire::Ready(())` before reaching `sequence.pop()`. That's an accidental invariant several call frames away. In the *speculative* path, `use_native_decode_reservation` sets `entry.copy = Some(..)` on the partial tail immediately after `append_partial`, so if a rollback ever reached that entry the `copy.is_none()` assert would fire.

Either make it explicit (`unreachable!("native decode never rolls back a sampled token")` in `RequestKvState::pop`) or have `rollback_partial` release the copy through the pool.

### 5. Stale `prefill_cost` → `reserve_exact_prefix` panic

`schedule_request` passes `cached_prefix_tokens / block_size` as `reusable_prefix_blocks`, where `cached_prefix_tokens` may come from a `prefill_cost` computed earlier in the admission pass. `reserve_exact_prefix` panics on any candidate `first_copy()` can't find. Same behavior as the old `VllmBlockPool::reserve`, so no regression — but the new message is now the only thing documenting the requirement. Worth a test that admits N requests in one pass where an earlier one's allocation evicts a later one's authorized prefix, to confirm the invariant actually holds rather than being untested.

### 6. `finalize_lease_computed_prefix` with prefix caching off

Dropping the `if !self.enable_prefix_caching { return; }` early-return is correct (identities and `discard_completed_block` are needed regardless), but it now reaches `publish_store_sequence(vec![None; n])` on every finalize when publishers exist and caching is off. That's a no-op only because `flush_store_group` returns on `None` — easy to break later. A one-line comment or an `if stores.iter().any(Option::is_some)` guard would make the intent explicit.

### 7. `block_size == 1` is now a hard config error for native G1

Since `resolved_g1_backend()` defaults to Native for vLLM/TRT-LLM without KVBM tiers, any existing config with `block_size: 1` now fails validation at startup. Justified (`RequestSequence::new` asserts `block_size >= 2`), but user-facing and should be release-noted. The message would also be more useful if it named the workaround (`set g1_backend: kvbm to keep block_size 1`), which `test_normalized_kvbm_accepts_block_size_one` shows is supported.

### 8. Minor / style

- `sglang/policy.rs`: the fold accumulates into `LocalBlockHash`, but the accumulated value is semantically a *sequence* hash — `LocalBlockHash(compute_next_seq_hash(parent.0, local))`. Folding into a plain `u64` and using that as the `FxHashSet` key would read better. Also `dynamo_kv_router::protocols::LocalBlockHash` is fully qualified inline while `compute_next_seq_hash` is imported at the top.
- `DecodeTokenReservation::page_size` is `#[cfg(test)]` inside a `pub` struct, so the layout differs between test and non-test builds. Harmless (private field, internal construction), but unusual enough to warrant a comment.
- `RequestSequence::complete_block_identity` panics via `complete_block(...)` *before* the `!enable_prefix_caching` early return, then discards the tokens. Deliberate invariant check — say so, or a future reader will "optimize" it away.
- `debug_assert_invariants(&self, _identities: impl ExactSizeIterator<…>)`: the underscore-prefixed parameter is genuinely used inside the `cfg(debug_assertions)` block. Necessary to avoid an unused warning in release, but a short comment would help since it reads as dead at every call site.

---

### Test coverage

Good additions: `page_native_extension_oom_is_atomic`, `retract_and_readmit_reuses_cached_page_hashes` (including the pointer/capacity check), `destination_activation_bounds_hashes_to_materialized_tokens`, `active_kv_lease_retains_one_id_per_physical_page`, `exact_prefix_reports_missing_hash`, and parameterizing the raw-sink test over both backends with an exact `vec![vec![0,1,2,3], vec![4,5,6,7]]` assertion instead of `!is_empty()`.

Gaps:

- **No test that an unset `g1_backend` produces `RequestKvState::Native`.** Every new test passes `.g1_backend(...)` explicitly. This is precisely what `be8be336ef` fixes and it now decides the request representation — a regression there sends default-config vLLM requests down the KVBM branch against a native manager.
- `reserve_resident_prefix`'s partial-hit path (some candidates resident, some not) has no direct unit test — only the `exact` variant's panic is covered.
- `allocate_lease`'s `"only a request's first allocation may reuse a prefix"` and `assert!(reusable_prefix_blocks <= target_blocks - previous_blocks)` are untested.
- `RequestKvState::pop()` / `rollback_partial` untested (see #4).

---

**Bottom line:** item 1 is the one I'd fix before merge, since it's a memory regression the benchmark methodology structurally can't observe. Items 2–4 are follow-up-worthy; the rest are polish.

lib/mocker/src/common/protocols.rs
```diff
@@ -1007,6 +1007,19 @@ fn validate_mock_engine_args(args: &MockEngineArgs) -> Result<(), ValidationErro
...
     }
 
+    if matches!(args.engine_type, EngineType::Vllm | EngineType::Trtllm)
+        && args.resolved_g1_backend() == G1Backend::Native
+        && args.block_size < 2
+    {
+        return Err(mock_engine_args_validation_error(
+            "native_g1_block_size_too_small",
+            format!(
+                "native G1 requires block_size to be at least 2 for engine_type={:?}, got block_size={}; set g1_backend=kvbm to keep block_size=1",
```
PeaBrane: AI review: this advertised workaround is not executable. `g1_backend=kvbm, block_size=1` passes normalization, but the first vLLM/TRT request constructs `ActiveSequence`, whose `block_size` validator still requires `>= 2`; `validate().expect(...)` then panics. Offload configurations automatically resolved to KVBM hit the same path. Please reject size one for both shared-scheduler backends unless legacy KVBM is fully updated, and cover actual request construction rather than normalization alone.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -124,6 +124,55 @@ impl VllmBlockPool {
...
+    }
+
+    /// Resolve and pin the longest resident prefix from `candidates`, then
+    /// reserve the remaining entries as fresh capacity in one traversal.
+    pub(crate) fn reserve_resident_prefix(
+        &mut self,
+        candidates: impl IntoIterator<Item = SequenceHash>,
+        total: usize,
+    ) -> Option<ReserveOutcome> {
+        let mut hits = Vec::with_capacity(total);
```
PeaBrane: AI review: `total` is the full requested layout, but this vector stores only resident prefix hits. With no hits it allocates `O(total)` unused capacity; `reserve_exact_prefix` does the same, adding an allocation/free to fresh native growth and leaving cold destination holds with prompt-sized unused capacity. Fast-path zero-hit reservations through `reserve_fresh(total)` and size non-empty storage according to actual hits.

lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -1724,6 +1770,26 @@ impl VllmCore {
         preempted
     }
 
+    fn process_preemption_cleanup(&mut self, preempted: &PreemptedRequest) {
```
PeaBrane, line 1773: AI review: the new backend-specific lifecycle dispatch lacks an explicit KVBM regression for ordinary preemption and the complete disaggregated reserve → activate → admit → release path. Native tests do not exercise KVBM's `reset_legacy_with_signal`, deferred cleanup, activation-time `commit_allocation`, or held-source dereference. Please add backend-parameterized lifecycle coverage; otherwise wrong/empty cleanup signals can leak blocks or strand re-admission without existing tests failing.

lib/mocker/src/scheduler/vllm/tests.rs
```diff
@@ -2381,19 +2424,22 @@ mod live_scheduler {
         assert_scheduler_idle(&metrics);
     }
 
+    #[rstest]
+    #[case::kvbm(G1Backend::Kvbm)]
+    #[case::native(G1Backend::Native)]
     #[tokio::test]
-    async fn test_live_scheduler_forwards_buffered_kv_token_ids() {
+    async fn raw_sink_emits_token_ids_for_both_g1_backends(#[case] backend: G1Backend) {
```
PeaBrane: AI review: this covers raw token-ID parity only for a block-aligned prompt. It misses the new retained-history edge where decode completes an unaligned prompt block. A parameterized three-token prompt with planned output `3` should assert that the newly completed Stored payload is exactly `[0, 1, 2, 3]` for both Native and KVBM.

lib/mocker/src/common/sequence.rs
```diff
@@ -88,10 +84,251 @@ impl FlatTokens {
...
+    ) {
+        // The underscore keeps release builds warning-free; debug builds use
+        // the iterator for the full logical-identity consistency check.
+        #[cfg(debug_assertions)]
+        {
+            let identity_count = _identities.len();
+            let aligned = self.len().is_multiple_of(self.block_size);
+            debug_assert_eq!(identity_count, self.current_known_blocks());
+            debug_assert!(
+                _identities
```
PeaBrane: AI review: this walks the entire lease entry list after every completed-block finalization. In debug/test builds, decoding a `B`-block request becomes `O(B²)` invariant work, and chunked prefill repeatedly rescans the prompt. Restrict the check to the changed range plus the final-tail invariant.

PeaBrane (commented):
AI-assisted review of `8c8c1e2`. The request-lease refactor looks structurally sound across KV-event ordering, disaggregated handoff ownership, KVBM/offload isolation, and SGLang retract/re-admit. I found one supported-configuration crash, one hot-path allocation regression, and several focused regression-coverage/performance concerns below.

PeaBrane (approved):
✅

lib/mocker/src/common/protocols.rs
jthomson04: Addressed in the latest review-fix commit. Normalization now rejects `block_size == 1` for vLLM/TRT-LLM regardless of Native, explicit KVBM, an unset backend, or auto-resolved KVBM. Added normalization coverage for all four configurations plus request-construction coverage at block size two for both vLLM/TRT-LLM × Native/KVBM.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Addressed. `reserve` delegates non-empty prefixes to `reserve_exact_prefix`; empty exact prefixes and zero-hit resident prefixes now take `reserve_fresh(total)` without allocating prefix storage. Non-empty exact storage is sized from the candidate count, and resident-prefix storage grows only after the first hit. New pool tests assert zero prefix-vector capacity for cold/empty cases and preserve partial-hit behavior.

lib/mocker/src/scheduler/vllm/core.rs
jthomson04: Added backend-parameterized Native/KVBM coverage for ordinary preemption/re-admission and the full source-hold plus destination reserve → activate → admit → release lifecycle. The tests assert zero residency after preemption and final release, deferred source cleanup, activation ownership, and event ordering.

lib/mocker/src/scheduler/vllm/tests.rs
jthomson04: Added the parameterized unaligned-prompt regression for both backends. It uses prompt `[0, 1, 2]`, materializes token `3` on the following decode forward, and asserts that the only newly completed Stored payload is exactly `[[0, 1, 2, 3]]`. The extra planned token is needed because the terminal sampled token itself never becomes computed KV.

lib/mocker/src/common/sequence.rs
jthomson04: Addressed. Post-finalization validation now calls `debug_assert_finalized_range` over only the newly finalized entries plus the final tail. The one-time full construction scan, identity-count check, and bounded token-storage assertions remain.

jthomson04:
/ok to test [e7e9baf](https://github.com/ai-dynamo/dynamo/commit/e7e9baf5e66a232d10e851e870208a45d8ea2c4e)

jthomson04:
/ok to test [03d309a](https://github.com/ai-dynamo/dynamo/commit/03d309ad8ef6470eda344b3e10034f9a8d28935b)

dreamtalen (approved):

jthomson04:
/ok to test [e85cfc4](https://github.com/ai-dynamo/dynamo/commit/e85cfc465e6873deba4f0d7c31da9659d9febfff)

jthomson04:
/ok to test [6992129](https://github.com/ai-dynamo/dynamo/commit/699212977a16f852232f9dc23d29c08adabacb9b)

PR #12331, opened 2026-07-29
fix(frontend): track Codex subagent threads

ishandhanani:
Codex child threads now execute through Dynamo's Responses frontend while retaining thread and parent-thread lineage. With `dynamo-protocols` 5.3.1 now supplied by `main`, Codex agent messages use the normal typed parser instead of Dynamo's raw-body fallback.

### How This Was Implemented

- Map Codex `thread-id` and `x-codex-parent-thread-id` into Dynamo's agent context; the run-wide `session-id` remains unused.
- Flatten namespace function tools for the chat backend and restore their namespace when the flattened name has one unambiguous origin.
- Use Axum's standard typed Responses JSON extractor and the released protocol parser for Codex agent messages.
- Require Codex V2 to call `spawn_agent` then `wait_agent`; its exec smoke uses exactly `{\"cmd\":\"ls\"}` so it cannot request unavailable approvals.

<details>

Codex's `thread-id` identifies an agent; `x-codex-parent-thread-id` links a child to its immediate parent. The root-run `session-id` is deliberately not used for Dynamo session affinity.

```mermaid
sequenceDiagram
    participant Parent as Codex parent
    participant Dynamo as Dynamo Responses frontend
    participant Model as Local model
    participant Child as Codex child
    Parent->>Dynamo: thread-id + collaboration tools
    Dynamo->>Model: flattened function tools
    Model-->>Dynamo: collaboration.spawn_agent
    Dynamo-->>Parent: namespaced tool call
    Parent->>Child: spawn_agent(task_name, message)
    Child->>Dynamo: agent_message + child/parent thread headers
    Dynamo->>Model: typed dynamo-protocols request
    Dynamo-->>Child: response; trace child parent context
```

Cross-origin name collisions remain rejected because a flat backend function call cannot be routed safely; duplicate top-level functions and duplicate functions from one namespace remain valid. `dynamo-protocols` owns Codex `agent_message` deserialization, so Dynamo no longer carries a divergent raw parser.

</details>

### Validation

- `tests/frontend/test_frontend_api_surface_compliance.py::test_frontend_api_surface_compliance` against local `Qwen/Qwen3-VL-2B-Instruct` (1 passed in 51.14s).
- `cargo test -p dynamo-llm protocols::openai::responses --lib` (96 passed).
- `cargo clippy -p dynamo-llm --lib --tests -- -D warnings`.
- `cargo fmt --all -- --check` and `git diff --check`.

## Summary by CodeRabbit

- **New Features**
  - Improved Codex thread tracking, including support for parent and child session relationships.
  - Child-thread requests now preserve parent session context for more accurate tracing.

- **Bug Fixes**
  - Corrected Codex session identification to use thread-specific headers instead of the run-wide session identifier.

- **Documentation**
  - Updated agent harness and session ID guidance to reflect the revised Codex header mappings.

### DIFF
--- a/lib/llm/src/protocols/agents.rs
+++ b/lib/llm/src/protocols/agents.rs
@@ -5,8 +5,6 @@
 
 use axum::http::HeaderMap;
 
-use crate::session_affinity::MAX_SESSION_AFFINITY_ID_BYTES;
-
 pub(crate) const HEADER_CLAUDE_CODE_SESSION_ID: &str = "x-claude-code-session-id";
 pub(crate) const HEADER_CLAUDE_CODE_AGENT_ID: &str = "x-claude-code-agent-id";
 pub(crate) const HEADER_CLAUDE_CODE_PARENT_AGENT_ID: &str = "x-claude-code-parent-agent-id";
@@ -64,12 +62,8 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
 
     if let Some(session_id) = header_value(headers, HEADER_DYNAMO_SESSION_ID) {
         return Some(AgentContextHeaderValues {
-            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID).filter(
-                |parent_session_id| {
-                    parent_session_id != &session_id
-                        && parent_session_id.len() <= MAX_SESSION_AFFINITY_ID_BYTES
-                },
-            ),
+            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID)
+                .filter(|parent_session_id| parent_session_id != &session_id),
             session_id,
             session_final,
         });
@@ -86,16 +80,13 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
         let parent_session_id = mapping
             .parent_session_header
             .and_then(|parent_header| header_value(headers, parent_header))
+            .filter(|parent_session_id| parent_session_id != &session_id)
             .filter(|_| {
                 !mapping.infer_parent_from_session_for_child || session_id != root_session_id
             })
             .or_else(|| {
                 (mapping.infer_parent_from_session_for_child && session_id != root_session_id)
                     .then(|| root_session_id.clone())
-            })
-            .filter(|parent_session_id| {
-                parent_session_id != &session_id
-                    && parent_session_id.len() <= MAX_SESSION_AFFINITY_ID_BYTES
             });
 
         return Some(AgentContextHeaderValues {
--- a/lib/llm/src/protocols/agents.rs
+++ b/lib/llm/src/protocols/agents.rs
@@ -5,6 +5,8 @@
 
 use axum::http::HeaderMap;
 
+use crate::session_affinity::MAX_SESSION_AFFINITY_ID_BYTES;
+
 pub(crate) const HEADER_CLAUDE_CODE_SESSION_ID: &str = "x-claude-code-session-id";
 pub(crate) const HEADER_CLAUDE_CODE_AGENT_ID: &str = "x-claude-code-agent-id";
 pub(crate) const HEADER_CLAUDE_CODE_PARENT_AGENT_ID: &str = "x-claude-code-parent-agent-id";
@@ -62,8 +64,12 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
 
     if let Some(session_id) = header_value(headers, HEADER_DYNAMO_SESSION_ID) {
         return Some(AgentContextHeaderValues {
-            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID)
-                .filter(|parent_session_id| parent_session_id != &session_id),
+            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID).filter(
+                |parent_session_id| {
+                    parent_session_id != &session_id
+                        && parent_session_id.len() <= MAX_SESSION_AFFINITY_ID_BYTES
+                },
+            ),
             session_id,
             session_final,
         });
@@ -80,13 +86,16 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
         let parent_session_id = mapping
             .parent_session_header
             .and_then(|parent_header| header_value(headers, parent_header))
-            .filter(|parent_session_id| parent_session_id != &session_id)
             .filter(|_| {
                 !mapping.infer_parent_from_session_for_child || session_id != root_session_id
             })
             .or_else(|| {
                 (mapping.infer_parent_from_session_for_child && session_id != root_session_id)
                     .then(|| root_session_id.clone())
+            })
+            .filter(|parent_session_id| {
+                parent_session_id != &session_id
+                    && parent_session_id.len() <= MAX_SESSION_AFFINITY_ID_BYTES
             });
 
         return Some(AgentContextHeaderValues {
--- a/lib/llm/src/protocols/agents.rs
+++ b/lib/llm/src/protocols/agents.rs
@@ -8,7 +8,8 @@ use axum::http::HeaderMap;
 pub(crate) const HEADER_CLAUDE_CODE_SESSION_ID: &str = "x-claude-code-session-id";
 pub(crate) const HEADER_CLAUDE_CODE_AGENT_ID: &str = "x-claude-code-agent-id";
 pub(crate) const HEADER_CLAUDE_CODE_PARENT_AGENT_ID: &str = "x-claude-code-parent-agent-id";
-pub(crate) const HEADER_CODEX_SESSION_ID: &str = "session-id";
+pub(crate) const HEADER_CODEX_THREAD_ID: &str = "thread-id";
+pub(crate) const HEADER_CODEX_PARENT_THREAD_ID: &str = "x-codex-parent-thread-id";
 pub(crate) const HEADER_OPENCODE_SESSION_ID: &str = "x-session-id";
 pub(crate) const HEADER_OPENCODE_PARENT_SESSION_ID: &str = "x-parent-session-id";
 pub(crate) const HEADER_DYNAMO_SESSION_ID: &str = "x-dynamo-session-id";
@@ -31,9 +32,9 @@ const AGENT_HEADER_MAPPINGS: &[AgentHeaderMapping] = &[
         infer_parent_from_session_for_child: true,
     },
     AgentHeaderMapping {
-        root_session_header: HEADER_CODEX_SESSION_ID,
+        root_session_header: HEADER_CODEX_THREAD_ID,
         child_session_header: None,
-        parent_session_header: None,
+        parent_session_header: Some(HEADER_CODEX_PARENT_THREAD_ID),
         infer_parent_from_session_for_child: false,
     },
     AgentHeaderMapping {
--- a/lib/llm/src/protocols/agents.rs
+++ b/lib/llm/src/protocols/agents.rs
@@ -62,8 +62,9 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
 
     if let Some(session_id) = header_value(headers, HEADER_DYNAMO_SESSION_ID) {
         return Some(AgentContextHeaderValues {
+            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID)
+                .filter(|parent_session_id| parent_session_id != &session_id),
             session_id,
-            parent_session_id: header_value(headers, HEADER_DYNAMO_PARENT_SESSION_ID),
             session_final,
         });
     }
@@ -79,6 +80,7 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
         let parent_session_id = mapping
             .parent_session_header
             .and_then(|parent_header| header_value(headers, parent_header))
+            .filter(|parent_session_id| parent_session_id != &session_id)
             .filter(|_| {
                 !mapping.infer_parent_from_session_for_child || session_id != root_session_id
             })
--- a/lib/llm/src/protocols/openai/chat_completions/delta.rs
+++ b/lib/llm/src/protocols/openai/chat_completions/delta.rs
@@ -128,6 +128,7 @@ impl DeltaGenerator {
                     dynamo_protocols::types::ChatCompletionTokenLogprob {
                         token: token_str.clone(),
                         logprob: lp,
+                        token_id: None,
                         bytes: token_to_utf8_bytes(&token_str),
                         top_logprobs: converted,
                     }
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -3,7 +3,7 @@
 
 pub mod stream_converter;
 
-use std::{collections::HashMap, sync::OnceLock};
+use std::collections::HashMap;
 
 use dynamo_protocols::types::responses::{
     AssistantRole, FunctionCallOutput, FunctionToolCall, IncludeEnum, IncompleteDetails,
@@ -1013,7 +1013,6 @@ pub struct ResponseParams {
     pub prompt_cache_key: Option<String>,
     pub prompt_cache_retention: Option<PromptCacheRetention>,
     pub safety_identifier: Option<String>,
-    pub(crate) function_namespaces: OnceLock<HashMap<String, Option<String>>>,
 }
 
 impl ResponseParams {
@@ -1025,38 +1024,28 @@ impl ResponseParams {
     }
 
     fn namespace_for_function(&self, name: &str) -> Option<String> {
-        self.function_namespaces
-            .get_or_init(|| {
-                let mut origins = HashMap::new();
-                for tool in self.tools.as_deref().unwrap_or_default() {
-                    match tool {
-                        Tool::Function(function) => {
-                            origins.insert(function.name.clone(), None);
-                        }
-                        Tool::Namespace(namespace) => {
-                            for tool in &namespace.tools {
-                                let NamespaceToolParamTool::Function(function) = tool else {
-                                    continue;
-                                };
-                                let origin = Some(namespace.name.clone());
-                                origins
-                                    .entry(function.name.clone())
-                                    .and_modify(|previous| {
-                                        if previous != &origin {
-                                            *previous = None;
-                                        }
-                                    })
-                                    .or_insert(origin);
-                            }
-                        }
-                        _ => {}
-                    }
-                }
-                origins
-            })
-            .get(name)
-            .cloned()
-            .flatten()
+        let tools = self.tools.as_deref()?;
+        if tools
+            .iter()
+            .any(|tool| matches!(tool, Tool::Function(function) if function.name == name))
+        {
+            return None;
+        }
+
+        let mut namespaces = tools.iter().filter_map(|tool| {
+            let Tool::Namespace(namespace) = tool else {
+                return None;
+            };
+            namespace
+                .tools
+                .iter()
+                .any(|tool| matches!(tool, NamespaceToolParamTool::Function(function) if function.name == name))
+                .then_some(namespace.name.as_str())
+        });
+        let namespace = namespaces.next()?;
+        namespaces
+            .all(|other_namespace| other_namespace == namespace)
+            .then(|| namespace.to_owned())
     }
 }
 
@@ -2834,14 +2823,11 @@ mod tests {
             ..Default::default()
         };
 
-        assert!(params.function_namespaces.get().is_none());
-
         let response = chat_completion_to_response(chat_resp, &params, None).unwrap();
         let OutputItem::FunctionCall(call) = &response.inner.output[0] else {
             panic!("expected function call");
         };
         assert_eq!(call.namespace.as_deref(), Some("agents"));
-        assert!(params.function_namespaces.get().is_some());
     }
 
     #[test]
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -3,7 +3,7 @@
 
 pub mod stream_converter;
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::OnceLock};
 
 use dynamo_protocols::types::responses::{
     AssistantRole, FunctionCallOutput, FunctionToolCall, IncludeEnum, IncompleteDetails,
@@ -661,9 +661,11 @@ fn convert_tools(tools: &[Tool]) -> anyhow::Result<Vec<ChatCompletionTool>> {
      -> anyhow::Result<()> {
         if let Some(previous_namespace) = origins.get(name) {
             if previous_namespace.as_deref() != namespace {
-                return Err(anyhow::anyhow!(
-                    "Responses function tool name '{name}' is ambiguous after namespace flattening"
-                ));
+                return Err(ResponsesConversionError::InvalidArgument(
+                    "Responses function tool names are ambiguous after namespace flattening"
+                        .to_string(),
+                )
+                .into());
             }
         } else {
             origins.insert(name.to_owned(), namespace.map(str::to_owned));
@@ -698,6 +700,7 @@ fn convert_tools(tools: &[Tool]) -> anyhow::Result<Vec<ChatCompletionTool>> {
                     }
                 }
             }
+            // Only function tools are forwarded to Chat Completions.
             _ => {}
         }
     }
@@ -1010,6 +1013,7 @@ pub struct ResponseParams {
     pub prompt_cache_key: Option<String>,
     pub prompt_cache_retention: Option<PromptCacheRetention>,
     pub safety_identifier: Option<String>,
+    pub(crate) function_namespaces: OnceLock<HashMap<String, Option<String>>>,
 }
 
 impl ResponseParams {
@@ -1021,28 +1025,38 @@ impl ResponseParams {
     }
 
     fn namespace_for_function(&self, name: &str) -> Option<String> {
-        let tools = self.tools.as_deref()?;
-        if tools
-            .iter()
-            .any(|tool| matches!(tool, Tool::Function(function) if function.name == name))
-        {
-            return None;
-        }
-
-        let mut namespaces = tools.iter().filter_map(|tool| {
-            let Tool::Namespace(namespace) = tool else {
-                return None;
-            };
-            namespace
-                .tools
-                .iter()
-                .any(|tool| matches!(tool, NamespaceToolParamTool::Function(function) if function.name == name))
-                .then_some(namespace.name.as_str())
-        });
-        let namespace = namespaces.next()?;
-        namespaces
-            .all(|other_namespace| other_namespace == namespace)
-            .then(|| namespace.to_owned())
+        self.function_namespaces
+            .get_or_init(|| {
+                let mut origins = HashMap::new();
+                for tool in self.tools.as_deref().unwrap_or_default() {
+                    match tool {
+                        Tool::Function(function) => {
+                            origins.insert(function.name.clone(), None);
+                        }
+                        Tool::Namespace(namespace) => {
+                            for tool in &namespace.tools {
+                                let NamespaceToolParamTool::Function(function) = tool else {
+                                    continue;
+                                };
+                                let origin = Some(namespace.name.clone());
+                                origins
+                                    .entry(function.name.clone())
+                                    .and_modify(|previous| {
+                                        if previous != &origin {
+                                            *previous = None;
+                                        }
+                                    })
+                                    .or_insert(origin);
+                            }
+                        }
+                        _ => {}
+                    }
+                }
+                origins
+            })
+            .get(name)
+            .cloned()
+            .flatten()
     }
 }
 
@@ -1060,6 +1074,17 @@ pub(super) fn normalize_tools(tools: Vec<Tool>) -> Vec<Tool> {
                     ft.strict = Some(true);
                 }
                 Tool::Function(ft)
+            }
+            Tool::Namespace(mut namespace) => {
+                for tool in &mut namespace.tools {
+                    let NamespaceToolParamTool::Function(function) = tool else {
+                        continue;
+                    };
+                    if function.strict.is_none() {
+                        function.strict = Some(true);
+                    }
+                }
+                Tool::Namespace(namespace)
             }
             other => other,
         })
@@ -2573,8 +2598,12 @@ mod tests {
         let error = NvCreateChatCompletionRequest::try_from(req).unwrap_err();
         assert_eq!(
             error.to_string(),
-            "Responses function tool name 'lookup' is ambiguous after namespace flattening"
+            "Responses function tool names are ambiguous after namespace flattening"
         );
+        assert!(matches!(
+            error.downcast_ref::<ResponsesConversionError>(),
+            Some(ResponsesConversionError::InvalidArgument(_))
+        ));
     }
 
     #[test]
@@ -2596,7 +2625,7 @@ mod tests {
         let error = NvCreateChatCompletionRequest::try_from(req).unwrap_err();
         assert_eq!(
             error.to_string(),
-            "Responses function tool name 'lookup' is ambiguous after namespace flattening"
+            "Responses function tool names are ambiguous after namespace flattening"
         );
     }
 
@@ -2635,6 +2664,20 @@ mod tests {
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
         assert_eq!(chat_req.inner.tools.unwrap().len(), 2);
+    }
+
+    #[test]
+    fn test_normalize_tools_sets_namespace_function_strict() {
+        let tools = serde_json::from_value(serde_json::json!([{
+            "type": "namespace",
+            "name": "agents",
+            "description": "Subagent tools",
+            "tools": [{"type": "function", "name": "spawn_agent"}]
+        }]))
+        .unwrap();
+
+        let normalized = serde_json::to_value(normalize_tools(tools)).unwrap();
+        assert_eq!(normalized[0]["tools"][0]["strict"], true);
     }
 
     #[allow(deprecated)]
@@ -2791,11 +2834,14 @@ mod tests {
             ..Default::default()
         };
 
+        assert!(params.function_namespaces.get().is_none());
+
         let response = chat_completion_to_response(chat_resp, &params, None).unwrap();
         let OutputItem::FunctionCall(call) = &response.inner.output[0] else {
             panic!("expected function call");
         };
         assert_eq!(call.namespace.as_deref(), Some("agents"));
+        assert!(params.function_namespaces.get().is_some());
     }
 
     #[test]
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -6,10 +6,10 @@ pub mod stream_converter;
 use std::collections::HashMap;
 
 use dynamo_protocols::types::responses::{
-    AssistantRole, FunctionCallOutput, FunctionToolCall, IncludeEnum, IncompleteDetails,
-    InputContent, InputItem, InputOutputMessageContent, InputParam, InputRole, InputTokenDetails,
-    Instructions, Item, MessageItem, NamespaceToolParamTool, OutputItem, OutputMessage,
-    OutputMessageContent, OutputStatus, OutputTextContent, OutputTokenDetails,
+    AssistantRole, CreateResponse, FunctionCallOutput, FunctionToolCall, IncludeEnum,
+    IncompleteDetails, InputContent, InputItem, InputOutputMessageContent, InputParam, InputRole,
+    InputTokenDetails, Instructions, Item, MessageItem, NamespaceToolParamTool, OutputItem,
+    OutputMessage, OutputMessageContent, OutputStatus, OutputTextContent, OutputTokenDetails,
     PromptCacheRetention, Reasoning, ReasoningItem, Response, ResponseTextParam, ResponseUsage,
     Role as ResponseRole, ServiceTier, Status, SummaryPart, SummaryTextContent,
     TextResponseFormatConfiguration, Tool, ToolChoiceOptions, ToolChoiceParam, Truncation,
@@ -37,25 +37,13 @@ use super::chat_completions::{NvCreateChatCompletionRequest, NvCreateChatComplet
 use super::{OpenAISamplingOptionsProvider, OpenAIStopConditionsProvider};
 use crate::protocols::common::extensions::{NvExt, NvExtProvider};
 
-/// Request body for `POST /v1/responses`. Uses a plain
-/// `#[derive(Deserialize)]` — the relaxed input shapes are handled by
-/// Dynamo-owning the input chain in `dynamo_protocols::types::responses`,
-/// not by a custom pre-parse JSON patcher.
-/// An earlier iteration of this type carried a hand-written `impl Deserialize`
-/// that walked `serde_json::Value` to inject synthetic defaults for missing
-/// `id` / `status` / `annotations`; that was replaced by typed ownership for
-/// correctness and to avoid the double-deserialize cost.
-#[derive(ToSchema, Serialize, Deserialize, Validate, Debug, Clone)]
+/// Request body for `POST /v1/responses`.
+#[derive(ToSchema, Serialize, Validate, Debug, Clone)]
 pub struct NvCreateResponse {
     /// Flattened CreateResponse fields (model, input, temperature, etc.).
     ///
-    /// `CreateResponse` and its `input` chain (`InputParam`, `InputItem`,
-    /// `Item`, `MessageItem`, `InputOutputMessage`, `InputOutputMessageContent`,
-    /// `InputOutputTextContent`) are Dynamo-owned in `dynamo-protocols`. They
-    /// mirror upstream async-openai but accept the relaxed shapes real clients
-    /// emit (optional `id` / `status` / `content` on assistant messages,
-    /// optional `annotations` on `output_text` parts). See
-    /// `dynamo_protocols::types::responses` for the full rationale.
+    /// `CreateResponse` and its input chain are Dynamo-owned in
+    /// `dynamo-protocols` and mirror upstream async-openai's types.
     #[serde(flatten)]
     #[schema(value_type = Object)]
     pub inner: dynamo_protocols::types::responses::CreateResponse,
@@ -63,6 +51,98 @@ pub struct NvCreateResponse {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Object)]
     pub nvext: Option<NvExt>,
+}
+
+#[derive(Deserialize)]
+struct NvCreateResponseWire {
+    #[serde(flatten)]
+    inner: CreateResponse,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    nvext: Option<NvExt>,
+}
+
+impl<'de> Deserialize<'de> for NvCreateResponse {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let mut value = serde_json::Value::deserialize(deserializer)?;
+        normalize_codex_agent_messages(&mut value).map_err(serde::de::Error::custom)?;
+        let wire = NvCreateResponseWire::deserialize(value).map_err(serde::de::Error::custom)?;
+        Ok(Self {
+            inner: wire.inner,
+            nvext: wire.nvext,
+        })
+    }
+}
+
+/// Convert Codex's task envelope into a standard input message.
+fn normalize_codex_agent_messages(body: &mut serde_json::Value) -> Result<(), String> {
+    let Some(items) = body
+        .get_mut("input")
+        .and_then(serde_json::Value::as_array_mut)
+    else {
+        return Ok(());
+    };
+
+    for item in items {
+        let Some(message) = item.as_object() else {
+            continue;
+        };
+        if message.get("type").and_then(serde_json::Value::as_str) != Some("agent_message") {
+            continue;
+        }
+
+        let author = message
+            .get("author")
+            .and_then(serde_json::Value::as_str)
+            .unwrap_or("unknown agent");
+        let recipient = message
+            .get("recipient")
+            .and_then(serde_json::Value::as_str)
+            .unwrap_or("unknown recipient");
+        let content = message
+            .get("content")
+            .and_then(serde_json::Value::as_array)
+            .map(|parts| {
+                parts
+                    .iter()
+                    .map(
+                        |part| match part.get("type").and_then(serde_json::Value::as_str) {
+                            Some("input_text") => part
+                                .get("text")
+                                .and_then(serde_json::Value::as_str)
+                                .ok_or_else(|| {
+                                    "Codex agent_message input_text missing text".to_string()
+                                }),
+                            Some("encrypted_content") => {
+                                Err("Codex agent_message with encrypted content is unsupported"
+                                    .to_string())
+                            }
+                            Some(kind) => Err(format!(
+                                "Codex agent_message contains unsupported content type {kind}"
+                            )),
+                            None => Err("Codex agent_message content missing type discriminator"
+                                .to_string()),
+                        },
+                    )
+                    .collect::<Result<Vec<_>, _>>()
+                    .map(|parts| parts.join("\n"))
+            })
+            .transpose()?
+            .unwrap_or_default();
+
+        *item = serde_json::json!({
+            "type": "message",
+            "role": "user",
+            "content": [{
+                "type": "input_text",
+                "text": format!("Message from {author} to {recipient}:\n{content}"),
+            }],
+        });
+    }
+
+    Ok(())
 }
 
 #[derive(ToSchema, Deserialize, Validate, Debug, Clone)]
@@ -3029,6 +3109,52 @@ thinking
             }
             other => panic!("expected MessageItem::Output, got {:?}", other),
         }
+    }
+
+    #[test]
+    fn test_nvcreate_response_normalizes_codex_agent_message() {
+        let body = serde_json::json!({
+            "model": "m",
+            "input": [{
+                "type": "agent_message",
+                "author": "/root",
+                "recipient": "/root/dynamo_subagent_smoke",
+                "content": [
+                    {"type": "input_text", "text": "First."},
+                    {"type": "input_text", "text": "Second."},
+                ],
+            }],
+        });
+
+        let req: NvCreateResponse = serde_json::from_value(body).unwrap();
+        let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
+        let [ChatCompletionRequestMessage::User(message)] = &chat_req.inner.messages[..] else {
+            panic!("expected one user message");
+        };
+        assert_eq!(
+            message.content,
+            ChatCompletionRequestUserMessageContent::Text(
+                "Message from /root to /root/dynamo_subagent_smoke:\nFirst.\nSecond.".to_string()
+            )
+        );
+    }
+
+    #[test]
+    fn test_nvcreate_response_rejects_encrypted_codex_agent_message() {
+        let body = serde_json::json!({
+            "model": "m",
+            "input": [{
+                "type": "agent_message",
+                "content": [{"type": "encrypted_content", "encrypted_content": "AB=="}],
+            }],
+        });
+
+        let error = serde_json::from_value::<NvCreateResponse>(body).unwrap_err();
+        assert!(
+            error
+                .to_string()
+                .contains("encrypted content is unsupported")
+        );
     }
 
     #[test]
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -793,19 +793,22 @@ fn convert_input_items_to_messages(
 /// namespace to restore on the response path.
 fn convert_tools(tools: &[Tool]) -> anyhow::Result<Vec<ChatCompletionTool>> {
     let mut converted = Vec::new();
-    let mut origins = HashMap::<String, String>::new();
+    let mut origins = HashMap::<String, bool>::new();
     let mut push_function = |name: &str,
                              description: &Option<String>,
                              parameters: &Option<serde_json::Value>,
                              strict: Option<bool>,
-                             origin: String|
+                             is_namespaced: bool|
      -> anyhow::Result<()> {
-        if let Some(previous_origin) = origins.get(name) {
+        if origins
+            .get(name)
+            .is_some_and(|previous_is_namespaced| *previous_is_namespaced || is_namespaced)
+        {
             return Err(anyhow::anyhow!(
-                "Responses function tool name '{name}' is ambiguous after namespace flattening: declared by {previous_origin} and {origin}"
+                "Responses function tool name '{name}' is ambiguous after namespace flattening"
             ));
         }
-        origins.insert(name.to_owned(), origin);
+        origins.entry(name.to_owned()).or_insert(is_namespaced);
         converted.push(ChatCompletionTool {
             r#type: ChatCompletionToolType::Function,
             function: FunctionObject {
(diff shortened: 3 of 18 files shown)


docs/fern/agents/agent-harnesses.mdx
```diff
@@ -36,7 +36,7 @@ bash examples/backends/sglang/launch/agg_agent.sh
     codex -m zai-org/GLM-4.7-Flash -c model_provider=dynamo
     ```
 
-    Codex sends a `session-id` header that Dynamo maps to `session_id`.
+    Dynamo maps Codex's `thread-id` header to `session_id`. A spawned child thread also sends `x-codex-parent-thread-id`, which Dynamo maps to `parent_session_id`; Codex's run-wide `session-id` is not used.
```
ishandhanani: Remove `codex's run-wide `session-id` is not used`

docs/fern/agents/session-ids.mdx
```diff
@@ -39,7 +39,7 @@ Custom clients should send the canonical Dynamo headers. When `X-Dynamo-Session-
 
 ## Native Agent Headers
 
-Dynamo also recognizes the current stable identity headers emitted by the following coding agents. The [frontend API surface compliance test](https://github.com/ai-dynamo/dynamo/blob/main/tests/frontend/test_frontend_api_surface_compliance.py) catches header changes as coding agents evolve. See [Agent Harnesses](agent-harnesses.mdx) for packaged setup.
```
ishandhanani, line 42: Bring the CI link back. No need to remove

docs/fern/agents/session-ids.mdx
```diff
@@ -50,7 +50,12 @@ Dynamo also recognizes the current stable identity headers emitted by the follow
     - `x-claude-code-parent-agent-id` for nested children; top-level children fall back to `x-claude-code-session-id`
   </Tab>
   <Tab title="Codex">
-    Dynamo maps `session-id` to `session_id`.
+    Dynamo maps Codex's per-agent thread headers:
+
+    - `thread-id` to `session_id`
+    - `x-codex-parent-thread-id` to `parent_session_id` for child threads
+
+    Codex's `session-id` identifies the overall run and is not used as an agent session ID.
```
ishandhanani: Remove this sentance

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -930,6 +965,31 @@ def _run_codex_exec_smoke(
         )
 
 
+def _run_codex_child_header_smoke(
```
ishandhanani: Do we need this? Shoudl we just follow how claude does it?

docs/fern/agents/agent-harnesses.mdx
ishandhanani: Fixed in 87d7f4d17f: removed the run-wide `session-id` sentence.

docs/fern/agents/session-ids.mdx
ishandhanani: Fixed in 87d7f4d17f: restored the frontend compliance-test link.

docs/fern/agents/session-ids.mdx
ishandhanani: Fixed in 87d7f4d17f: removed the sentence.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -615,9 +688,25 @@ fn convert_tools(tools: &[Tool]) -> Vec<ChatCompletionTool> {
                     strict: f.strict,
                 },
             }),
-            _ => None, // Only function tools are forwarded to chat completions
-        })
-        .collect()
+            Tool::Namespace(namespace) => {
```
rmccorm4, line 690: I just ran into this issue trying to spawn subagents with self-hosted kimi k3 on dynamo + codex v0.146.0

It works on codex v0.132.0 but after that point they made it namespace tool which dynamo doesn't currently support

Let's get this fix in

rmccorm4:
Some hardening after local testing: https://github.com/ai-dynamo/dynamo/pull/12476

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -461,11 +464,11 @@ def _opencode_cli(_tools_cache, _node_bin) -> Path:
...
 # Budget: tool-install fixtures (~30-60s first session run, near-zero on
 # cache hit) + sglang cold start (30-60s) + bun compliance (up to 180s) +
-# codex exec (up to 180s) + claude exec (up to 180s) + optional claude
-# subagent probe (up to 45s) + opencode run (up to 180s) + inter-suite
-# health checks + teardown. 975s leaves headroom for CI variance without
-# masking real hangs.
-@pytest.mark.timeout(975)
+# codex exec (up to 180s) + Codex child process (up to 180s) + claude exec
+# (up to 180s) + optional claude subagent probe (up to 45s) + opencode run
+# (up to 180s) + inter-suite health checks + teardown. 975s leaves headroom
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 90980d7758: the timeout-budget comment now matches the 1095-second decorator.

lib/llm/src/protocols/agents.rs
```diff
@@ -31,9 +32,9 @@ const AGENT_HEADER_MAPPINGS: &[AgentHeaderMapping] = &[
         infer_parent_from_session_for_child: true,
     },
     AgentHeaderMapping {
-        root_session_header: HEADER_CODEX_SESSION_ID,
+        root_session_header: HEADER_CODEX_THREAD_ID,
         child_session_header: None,
-        parent_session_header: None,
+        parent_session_header: Some(HEADER_CODEX_PARENT_THREAD_ID),
         infer_parent_from_session_for_child: false,
```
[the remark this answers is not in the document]
ishandhanani: Verified against the current Codex client: Responses requests carry both `session-id` and `thread-id`, and Dynamo intentionally derives agent lineage only from `thread-id`. 90980d7758 adds a regression proving a `session-id`-only request produces no agent context.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -461,11 +464,11 @@ def _opencode_cli(_tools_cache, _node_bin) -> Path:
...
+# for CI variance without masking real hangs.
+@pytest.mark.timeout(1095)
 @pytest.mark.frontend_api_surface_compliance
 @pytest.mark.pre_merge
 @pytest.mark.flaky(reruns=2, only_rerun=["did not report the marker file"])
```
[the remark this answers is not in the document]
ishandhanani: Keeping the subagent flow as a required, non-retried gate per the intended CI contract. It is isolated to trace records emitted after the Codex invocation; no flaky failure has been reproduced.

lib/llm/src/protocols/openai/responses/mod.rs
ishandhanani: Implemented in 90980d7758: namespace functions are flattened for the Chat backend and their namespace is restored in unary and streaming Responses output. The targeted namespace tests now pass.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -738,7 +758,8 @@ def _assert_agent_context_in_trace(
     ]
     pytest.fail(
         f"request trace did not contain agent_context after {source_label!r} "
-        f"within {timeout_s}s; saw {seen}"
+        f"within {timeout_s}s; expected session={expected_session_id!r}, "
+        f"parent={expected_parent_session_id!r}; saw {seen}"
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 90980d7758: removed the unused expected session/parent arguments and the corresponding dead filtering.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -930,6 +962,56 @@ def _run_codex_exec_smoke(
...
+        cwd=str(cwd),
+    )
+    if result.returncode != 0:
+        pytest.fail(
+            f"codex subagent smoke failed (exit={result.returncode}).\n"
+            f"stdout:\n{result.stdout}\n\nstderr:\n{result.stderr}"
+        )
+    _assert_agent_parent_context_in_trace(
+        request_trace_path, "codex subagent", trace_start_index
+    )
```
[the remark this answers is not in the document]
ishandhanani: The trace start index is captured immediately before the Codex invocation, so only newly emitted records are inspected. The assertion requires distinct child and parent lineage; absent subagent execution fails the gate.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -58,13 +51,118 @@ pub struct NvCreateResponse {
...
+                        None => {
+                            return Err("Codex agent_message content missing type discriminator"
+                                .to_string());
+                        }
+                    }
+                }
+                Ok(text.join("\n"))
+            })
+            .transpose()?
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 6d364d53ea: `agent_message.content` now accepts a plain string and preserves it as child-task text. The regression test covers that path; other non-string/non-array shapes are rejected explicitly.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -65,6 +53,98 @@ pub struct NvCreateResponse {
...
+                            )),
+                            None => Err("Codex agent_message content missing type discriminator"
+                                .to_string()),
+                        },
+                    )
+                    .collect::<Result<Vec<_>, _>>()
+                    .map(|parts| parts.join("\n"))
+            })
+            .transpose()?
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
ishandhanani: Confirmed the current client-visible contract: `responses_router` applies `smart_json_error_middleware`, which converts the extractor 422 into a structured Dynamo `ErrorMessage` with HTTP 400. This validation error follows the same malformed-request path as other JSON extraction failures.

tests/frontend/test_frontend_api_surface_compliance.py
ishandhanani: Implemented as a blocking Codex child-agent probe: it starts trace capture immediately before invocation and requires newly emitted child/parent lineage, mirroring the Claude coverage model.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -65,6 +53,99 @@ pub struct NvCreateResponse {
...
+    {
+        let mut value = serde_json::Value::deserialize(deserializer)?;
+        normalize_codex_agent_messages(&mut value).map_err(serde::de::Error::custom)?;
+        let wire = NvCreateResponseWire::deserialize(value).map_err(serde::de::Error::custom)?;
+        Ok(Self {
+            inner: wire.inner,
+            nvext: wire.nvext,
+        })
+    }
+}
```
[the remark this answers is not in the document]
ishandhanani: Agreed. The root fix is tracked in [ai-dynamo/frontend-crates#165](https://github.com/ai-dynamo/frontend-crates/pull/165), which adds a typed `agent_message` input variant. Once that crate is released, this PR can remove the request-wide normalizer; a git/path dependency here would make Dynamo CI non-reproducible. Leaving this thread open until the released dependency is consumed.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -615,9 +696,25 @@ fn convert_tools(tools: &[Tool]) -> Vec<ChatCompletionTool> {
...
+                                name: f.name.clone(),
+                                description: f.description.clone(),
+                                parameters: f.parameters.clone(),
+                                strict: f.strict,
+                            },
+                        });
+                    }
+                }
+            }
+            _ => {}
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 7b30bc20ef: the flat Chat bridge now rejects duplicate function names across namespaces or between namespace and top-level declarations before inference. The two collision cases are covered by unit tests.

tests/frontend/agent_smoke_inputs.py
```diff
@@ -19,6 +19,11 @@ def write_codex_config(codex_home: Path, frontend_port: int) -> None:
         f"""
 model_max_output_tokens = 4096
 
+[features.multi_agent_v2]
+enabled = true
+max_concurrent_threads_per_session = 2
+non_code_mode_only = false
```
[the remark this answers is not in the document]
ishandhanani: Fixed in babc26241c: the baseline marker-file smoke runs with the default tool set. The multi-agent feature is enabled only immediately before the dedicated subagent probe.

lib/llm/src/protocols/agents.rs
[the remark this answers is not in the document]
ishandhanani: Intentional: Dynamo uses Codex `thread-id` as the stable per-agent identity and deliberately ignores run-wide `session-id`; the regression test locks that contract. Current Codex Requests emit both headers, so this does not affect supported clients.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -65,6 +53,99 @@ pub struct NvCreateResponse {
...
+
+        *item = serde_json::json!({
+            "type": "message",
+            "role": "user",
+            "content": [{
+                "type": "input_text",
+                "text": format!("Message from {author} to {recipient}:\n{content}"),
+            }],
+        });
+    }
```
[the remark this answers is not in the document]
ishandhanani: Verified against the Codex multi-agent fixture: `agent_message` is receiver-directed communication (parent→child or child→parent), so it is correctly represented as a user turn for the recipient. The emitting agent own prior output is a normal assistant message, not an `agent_message` item.

lib/llm/src/protocols/openai/responses/mod.rs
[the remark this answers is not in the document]
ishandhanani: Fixed in 4b96cc46a4: the endpoint now reads request bytes and uses the typed `Json<NvCreateResponse>` parse directly. It only builds a `Value` after that parse rejects a real Codex `agent_message`, so ordinary Responses requests do not allocate or walk a generic JSON tree. `cargo clippy -p dynamo-llm --lib --tests -- -D warnings` and 95 Responses tests pass.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -601,23 +682,66 @@ fn convert_input_items_to_messages(
...
+                             parameters: &Option<serde_json::Value>,
+                             strict: Option<bool>,
+                             origin: String|
+     -> anyhow::Result<()> {
+        if let Some(previous_origin) = origins.get(name) {
+            return Err(anyhow::anyhow!(
+                "Responses function tool name '{name}' is ambiguous after namespace flattening: declared by {previous_origin} and {origin}"
+            ));
+        }
+        origins.insert(name.to_owned(), origin);
```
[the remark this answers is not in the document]
ishandhanani: Fixed in f453f4aa5a: duplicate top-level function names retain the prior pass-through behavior. Only a collision involving a namespace is rejected, and all three cases are covered by unit tests.

lib/llm/src/protocols/agents.rs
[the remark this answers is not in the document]
ishandhanani: Fixed in 26b8b2716f: Dynamo now drops a parent session ID equal to the session ID for all recognized agent headers. Current Codex fixtures already omit the parent header for root requests and use a distinct child thread ID.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -601,23 +729,58 @@ fn convert_input_items_to_messages(
...
+                             is_namespaced: bool|
+     -> anyhow::Result<()> {
+        if origins
+            .get(name)
+            .is_some_and(|previous_is_namespaced| *previous_is_namespaced || is_namespaced)
+        {
+            return Err(anyhow::anyhow!(
+                "Responses function tool name '{name}' is ambiguous after namespace flattening"
+            ));
+        }
```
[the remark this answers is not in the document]
ishandhanani: Keeping the rejection. The backend receives only a flat function name, and `ResponseParams::namespace_for_function` deliberately returns `None` for this collision; accepting it would emit an unroutable function call rather than preserve either provider. This is an unsupported Responses-to-Chat conversion, so the existing 501 conversion error is intentional.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -65,6 +58,141 @@ pub struct NvCreateResponse {
...
+                            }
+                            Some(kind) => Err(format!(
+                                "Codex agent_message contains unsupported content type {kind}"
+                            )),
+                            None => Err("Codex agent_message content missing type discriminator"
+                                .to_string()),
+                        },
+                    )
+                    .collect::<Result<Vec<_>, _>>()
+                    .map(|parts| parts.join("\n"))?,
```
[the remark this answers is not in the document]
ishandhanani: Keeping the explicit compatibility contract. Current Codex serializes agent messages as `input_text`, optionally followed by `encrypted_content` (`codex-rs/protocol/src/protocol.rs:814-836`); no alternate plaintext discriminator exists. We will extend the normalizer only when Codex introduces a concrete wire format.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -601,23 +589,58 @@ fn convert_input_items_to_messages(
...
+            r#type: ChatCompletionToolType::Function,
+            function: FunctionObject {
+                name: name.to_owned(),
+                description: description.clone(),
+                parameters: parameters.clone(),
+                strict,
+            },
+        });
+        Ok(())
+    };
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 61adbd522e. Duplicate names from the same namespace now remain forwarded; cross-origin collisions remain rejected because their response namespace is ambiguous.

lib/llm/src/protocols/openai/chat_completions/delta.rs
```diff
@@ -128,6 +128,7 @@ impl DeltaGenerator {
                     dynamo_protocols::types::ChatCompletionTokenLogprob {
                         token: token_str.clone(),
                         logprob: lp,
+                        token_id: None,
```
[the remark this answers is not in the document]
ishandhanani: Confirmed `dynamo-protocols` 5.3.1 marks `token_id` with `skip_serializing_if = "Option::is_none"`, so this PR does not emit `token_id: null`. Populating the available backend ID is isolated in #12444 to keep this Codex parser/dependency update narrow.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -930,6 +948,56 @@ def _run_codex_exec_smoke(
...
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 92ea7d8ce9. The Codex child-agent path remains blocking: launch and parent-context trace failures now use the existing bounded two-retry policy for model-mediated variance, then still fail the suite. The live compliance probe passed (1 passed, 52.46s).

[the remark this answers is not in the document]
ishandhanani: [AI] Updated in fd8889eac6 to match the Claude path: Codex child-agent invocation and missing parent trace are now best-effort and logged. The inline TODO names both Codex and Claude probes as future blocking checks once their model-mediated execution is stable. Validation: direct nonzero-result check plus the local frontend compliance test (1 passed, 61.60s).

tmonty12 (approved):

PeaBrane (approved):

dagil-nvidia (approved):

nv-tusharma (approved):

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -575,6 +579,19 @@ def test_frontend_api_surface_compliance(
...
+        _try_run_codex_subagent_smoke(
+            _codex_cli,
+            _node_bin,
+            codex_home,
+            agent_cwd,
+            request_trace_path,
+            codex_subagent_trace_start,
+        )
+        _wait_for_frontend_healthy(frontend_port)
         claude_trace_start = _request_trace_record_count(request_trace_path)
```
dmitry-tokarev-nv: **BLOCKING** — the gating Claude assertion can be satisfied by a leftover Codex record.

`_trace_contains_agent_context` (line 787) takes no source argument and returns `True` for *any* record after `start_index` with a non-empty `agent_context.session_id`. `_assert_agent_context_in_trace` accepts `source_label` but only interpolates it into the failure message.

That was safe before this PR: the step ahead of `claude_trace_start` hard-failed on a non-zero exit, so you could never reach here with Codex traffic still in flight. `_try_run_codex_subagent_smoke` now returns silently on timeout *and* on non-zero exit while its child threads (`max_concurrent_threads_per_session = 2`) may still have requests outstanding, and `_wait_for_frontend_healthy` is a `/v1/models` poll, not a quiesce.

Net effect: `_assert_agent_context_in_trace(request_trace_path, "claude_code", claude_trace_start)` can pass on a Codex record. The Claude header mapping could regress to emitting no `agent_context` at all and this suite stays green.

Cheapest fix is to make the matcher source-aware — pass `source_label` through to `_trace_contains_agent_context` and filter on the request path or the header family that produced the context, instead of using it only in the error string.

Second, independent route to the same bug: `_read_request_trace_records_since` slices by count of *successfully parsed* records, while `_read_request_trace_records` silently drops torn JSONL lines (`except json.JSONDecodeError: ... continue`). With `DYN_REQUEST_TRACE_FILE_FLUSH_INTERVAL_MS=10`, a half-flushed final line at the snapshot on line 595 leaves `start_index` permanently one too low.

lib/llm/src/protocols/agents.rs
```diff
@@ -78,6 +80,7 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
         let parent_session_id = mapping
             .parent_session_header
             .and_then(|parent_header| header_value(headers, parent_header))
+            .filter(|parent_session_id| parent_session_id != &session_id)
             .filter(|_| {
                 !mapping.infer_parent_from_session_for_child || session_id != root_session_id
             })
```
dmitry-tokarev-nv, line 86: > [!WARNING]
> **Withdrawn — I tested this and the conclusion was wrong.** See [the correction below](https://github.com/ai-dynamo/dynamo/pull/12331#discussion_r3761704246). The current behavior is correct and the suggested fix in this comment **would be a regression** — please don't apply it.
>
> What survives is non-blocking: a test-table row for the Claude mapping, and the `session-ids.mdx` drift in the last paragraph. Original text kept below for the record.

---

**BLOCKING** — this guard (added in 26b8b2716f for the self-parenting thread above) is correct for Codex but inverts the intent for Claude Code.

It sits *before* the existing `.filter(...)` / `.or_else(...)` pair. For the one mapping with `infer_parent_from_session_for_child: true` (Claude Code), turning `Some(parent)` into `None` doesn't drop the parent — it hands control to the `.or_else` on line 87, which then fabricates one.

Concretely, for headers `x-claude-code-session-id: S`, `x-claude-code-agent-id: A`, `x-claude-code-parent-agent-id: A`:

- before: `.filter(|_| !true || A != S)` is `true`, so `parent_session_id = Some(A)`
- after: this filter drops it to `None`, the next filter is a no-op on `None`, and `.or_else(|| (true && A != S).then(|| S))` yields `Some(S)`

So a self-referential parent isn't ignored — it's re-parented to the root. Downstream, `_trace_contains_agent_parent_context` and `_trace_contains_claude_subagent_context` both `continue` when `parent == session`, so a record that used to be skipped as self-referential now counts as a genuine parent/child edge.

The new `agent_context_ignores_self_parent_header` test only covers Codex, where `infer_parent_from_session_for_child: false` makes the `.or_else` dead — so the `infer = true` branch of this filter has no coverage at all, while the test name asserts a contract that holds for only two of the three mappings.

Applying the guard *after* `.or_else` is correct for all three:

```rust
        let parent_session_id = mapping
            .parent_session_header
            .and_then(|parent_header| header_value(headers, parent_header))
            .filter(|_| {
                !mapping.infer_parent_from_session_for_child || session_id != root_session_id
            })
            .or_else(|| {
                (mapping.infer_parent_from_session_for_child && session_id != root_session_id)
                    .then(|| root_session_id.clone())
            })
            .filter(|parent_session_id| parent_session_id != &session_id);
```

Worth extending the test table to the Claude mapping either way.

Minor, same change: the identical guard on lines 65-66 narrows the documented `X-Dynamo-Parent-Session-ID` contract, but `docs/fern/pages/use-cases/agents/session-ids.mdx` (edited in this PR) still says it "Maps to `agent_context.parent_session_id`" unconditionally — only the Codex tab was updated.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -461,11 +465,11 @@ def _opencode_cli(_tools_cache, _node_bin) -> Path:
...
-# codex exec (up to 180s) + claude exec (up to 180s) + optional claude
-# subagent probe (up to 45s) + opencode run (up to 180s) + inter-suite
-# health checks + teardown. 975s leaves headroom for CI variance without
-# masking real hangs.
-@pytest.mark.timeout(975)
+# codex exec (up to 180s) + optional Codex child process (up to 210s) + claude
+# exec (up to 180s) + optional Claude subagent probe (up to 45s) + opencode run
+# (up to 180s) + inter-suite health checks + teardown. 1095s leaves headroom
+# for CI variance without masking real hangs.
+@pytest.mark.timeout(1095)
```
dmitry-tokarev-nv, line 471: **BLOCKING** — building on Devin's thread at this line, with two things it doesn't cover.

**1. There is no headroom at all**, not merely "not enough". The comment's own eight ceilings sum to exactly the new cap:

```
60 + 60 + 180 + 180 + 210 + 180 + 45 + 180 = 1095
```

That leaves 0s for the "inter-suite health checks + teardown" the same comment lists but doesn't number. Before this PR the equivalent sum was 885 against 975, i.e. 90s of real slack. "1095s leaves headroom for CI variance" is no longer true.

**2. A pytest-timeout kill is not retried.** `@pytest.mark.flaky(reruns=2, only_rerun=["did not report the marker file"])` on line 475 matches on failure text; pytest-timeout raises `Failed: Timeout >1095.0s`, which doesn't match. So the timeout path is a straight red on whatever PR happens to be running, attributed to the whole compliance suite rather than to the optional probe.

**Counter-pressure on the suggested fix** (raise the marker to ~1185): this marker is also the GPU orchestrator's only a-priori duration estimate. `tests/utils/pytest_parallel_gpu.py:70-77` defines `est_duration = timeout / 3.0` — *"Repo convention sets `@pytest.mark.timeout` to ~3x a test's measured runtime, so `timeout / 3` is the best a-priori duration estimate"* — and `_priority_key` (line 331) sorts longest-first (LPT) to minimize makespan. Raising the marker to a hard worst-case cap tells the scheduler this test is ~395s when its real runtime is heading toward ~1000s, so the longest test in the sglang parallel stage gets scheduled as a mid-sized job and can become the makespan tail.

The two consumers can't both be served by one number. Shrinking the work satisfies both — see the note on line 474.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -930,6 +950,66 @@ def _run_codex_exec_smoke(
...
+            result.stdout,
+            result.stderr,
+        )
+        return
+    try:
+        _assert_agent_parent_context_in_trace(
+            request_trace_path, "codex subagent", trace_start_index
+        )
+    except pytest.fail.Exception as error:
+        logger.warning("Optional Codex subagent smoke did not trace: %s", error)
```
dmitry-tokarev-nv: **BLOCKING** — the PR's headline behavior has no gating coverage, but the docstring says it does.

All three exits of `_try_run_codex_subagent_smoke` are `logger.warning(...); return`: `TimeoutExpired` (988), non-zero exit (998), and this handler. Revert `parent_session_header: Some(HEADER_CODEX_PARENT_THREAD_ID)` in `agents.rs` and this suite still passes.

Making the probe best-effort is a reasonable call for model-mediated behavior — it's the change made in response to Devin's earlier "gates the suite on model-mediated behavior" thread. The problem is that the promise wasn't updated with it:

- the rewritten module docstring, item 3, says this check "**requires** a real Codex child request through `/v1/responses`", with no "Optional" — unlike item 6 for the Claude probe
- `docs/fern/pages/use-cases/agents/session-ids.mdx:42`, edited in this PR, tells readers "The frontend API surface compliance test catches header changes as coding agents evolve"

Minimum to unblock: fix the docstring and that docs sentence so neither claims coverage that doesn't exist.

Worth doing as a follow-up: `x-codex-parent-thread-id` to `parent_session_id` is deterministic and needs no model. A test that POSTs `/v1/responses` with both headers set explicitly and asserts the trace record would gate the actual contract in about a second — no CLI, no GPU. That would also let this probe leave `pre_merge` entirely.

Also: the `TODO` on line 584 has no issue link, against the repo's `TODO(DIS-1679)` convention, and this is now the second permanently-optional probe carrying the same promise.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -657,23 +645,63 @@ fn convert_input_items_to_messages(
...
+                             description: &Option<String>,
+                             parameters: &Option<serde_json::Value>,
+                             strict: Option<bool>,
+                             namespace: Option<&str>|
+     -> anyhow::Result<()> {
+        if let Some(previous_namespace) = origins.get(name) {
+            if previous_namespace.as_deref() != namespace {
+                return Err(anyhow::anyhow!(
+                    "Responses function tool name '{name}' is ambiguous after namespace flattening"
+                ));
```
dmitry-tokarev-nv: **BLOCKING** — two problems with this error, both independent of whether rejecting the collision is the right policy (that's settled in the thread above).

**1. It returns 500, not 400.** Every sibling rejection in this file uses the typed error — `ResponsesConversionError::InvalidArgument`, e.g. lines 277/283/297/319/327/334 — and gets a client error. A bare `anyhow` fails the `downcast_ref::<ResponsesConversionError>()` in `http/service/openai.rs` and falls through to `internal_server_error_with_details`. The repo's own `untyped_responses_conversion_errors_remain_internal` test pins that outcome: 500, `message: "Failed to convert responses request"`, `details: None`, `ErrorType::Internal`.

So the caller never learns which tool collided, the frontend error-rate SLO absorbs a client-input error, and SDKs or gateways that retry 5xx-but-not-4xx will retry a deterministic failure.

**2. It interpolates an unbounded client-supplied string into ERROR-level logs.** `FunctionToolParam.name` comes straight from the JSON body with no length or charset validation, and `{name}` is echoed into `tracing::error!(request_id, error = %e, ...)` and again on the error-response path. Header-sourced values are protected by `HeaderValue::to_str()` rejecting control bytes; a body-sourced name has no such guard, so newlines in a tool name reach the structured log verbatim. On `main` this path logged nothing, because the tool was silently dropped.

The typed error fixes (1). For (2), dropping the name is simplest — the collision is deterministic and reproducible from the request body. Keep it if you want the debuggability, but truncate and quote it.

```suggestion
                return Err(ResponsesConversionError::InvalidArgument(
                    "Responses function tool names are ambiguous after namespace flattening"
                        .to_string(),
                )
                .into());
```

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -461,11 +465,11 @@ def _opencode_cli(_tools_cache, _node_bin) -> Path:
...
-# health checks + teardown. 975s leaves headroom for CI variance without
-# masking real hangs.
-@pytest.mark.timeout(975)
+# codex exec (up to 180s) + optional Codex child process (up to 210s) + claude
+# exec (up to 180s) + optional Claude subagent probe (up to 45s) + opencode run
+# (up to 180s) + inter-suite health checks + teardown. 1095s leaves headroom
+# for CI variance without masking real hangs.
+@pytest.mark.timeout(1095)
 @pytest.mark.frontend_api_surface_compliance
 @pytest.mark.pre_merge
```
dmitry-tokarev-nv, line 473: **NON-BLOCKING** — but this is the fix that resolves the tension on line 472.

`sglang-test` (`.github/workflows/pr.yaml:847`) does not set `gpu_test_timeout_minutes`, so it inherits `shared-test.yml`'s default of **30 minutes** per GPU stage — the sibling `vllm-multi-gpu-test`, `sglang-multi-gpu-test` and `trtllm-multi-gpu-test` jobs all set 60 explicitly. This test carries `profiled_vram_gib(14.2)` so it runs in that stage, against a `gpu_parallel_max_vram_gib: '24'` cap whose own comment says it yields "one auto slot today".

One attempt at the new cap is 1095s = 61% of the 1800s stage (was 54%). Two attempts is 2190s = 122%, and a GitHub step kill lands before pytest writes JUnit XML — so the whole sglang GPU-parallel stage reports zero results with nothing but "exceeded the maximum execution time". `pytest_parallel_gpu.py` can also relaunch the child up to `_MAX_RETRIES` more times on `_RETRYABLE_INIT_MARKERS`, and there's no orchestrator-level wall-clock kill.

Against `.ai/pytest-guidelines.md`: *"Tests averaging over 60 seconds should default to `post_merge`"*, *"Prefer `post_merge` for E2E tests"*, and `pre_merge` is for tests that are *"absolutely critical"*. This one is `e2e` + `gpu_1` + `pre_merge`, and per the TODO on line 584 the newly added path produces no signal by design.

Suggest keeping the deterministic parts in `pre_merge` and moving the two model-mediated subagent probes to `post_merge` — or replacing them with the header-level test described on line 1009. Either way the line 472 budget stops being a problem.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -575,6 +579,19 @@ def test_frontend_api_surface_compliance(
...
+        # TODO: Make the Codex and Claude model-mediated subagent probes blocking
+        # once they are stable in CI.
+        _try_run_codex_subagent_smoke(
+            _codex_cli,
+            _node_bin,
+            codex_home,
+            agent_cwd,
+            request_trace_path,
+            codex_subagent_trace_start,
+        )
```
dmitry-tokarev-nv: **NON-BLOCKING** — the probe shares a mutable working directory with a gating assertion.

`agent_cwd` holds the single `marker_{uuid}.txt` that `_run_codex_exec_smoke` (already run) and `_run_claude_exec_smoke` (line 596) both assert on. Unlike every earlier step, this one runs `codex exec --dangerously-bypass-approvals-and-sandbox` *and* spawns a child agent with its own shell tool — while swallowing every failure.

If the child does anything to `agent_cwd` the prompt didn't ask for (a cleanup `rm`, a `mv`, a stray redirect over the marker filename), the probe reports nothing and the *next* gating step fails with "claude -p did not report the marker file" — which **does** match `only_rerun`, so pytest burns two more full replays before reporting, and the log points at Claude and `/v1/messages` when the cause was three subprocesses upstream.

`codex_subagent_prompt()` doesn't reference anything in the cwd, so isolating it is free:

```suggestion
        codex_subagent_cwd = agent_cwd / "codex_subagent"
        codex_subagent_cwd.mkdir(exist_ok=True)
        _try_run_codex_subagent_smoke(
            _codex_cli,
            _node_bin,
            codex_home,
            codex_subagent_cwd,
            request_trace_path,
            codex_subagent_trace_start,
        )
```

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -930,6 +950,66 @@ def _run_codex_exec_smoke(
...
+            cmd,
+            env=_agent_subprocess_env(extra_env, path_prepend=[node_bin]),
+            cwd=cwd,
+            capture_output=True,
+            text=True,
+            timeout=180,
+        )
+    except subprocess.TimeoutExpired:
+        logger.warning("Optional Codex subagent smoke timed out")
+        return
```
dmitry-tokarev-nv: **NON-BLOCKING** — the timeout path discards exactly the diagnostics the TODO on line 584 needs.

This returns before `_attach_subprocess_log`, throwing away the `stdout`/`stderr` that `subprocess.run` has already attached to the exception. That's the 180s-burn case — the one you need evidence from to ever make this probe blocking — and today it yields a context-free warning and no artifact.

```suggestion
    except subprocess.TimeoutExpired as error:
        logger.warning(
            "Optional Codex subagent smoke timed out after 180s; stdout=%r stderr=%r",
            error.stdout,
            error.stderr,
        )
        return
```

Separately, "best-effort" only holds for the two exception types explicitly caught; everything else propagates and hard-fails the suite. `_attach_subprocess_log` does `import allure` (whose own comment notes allure "is only available inside the test image") and sits between the two `try` blocks; `subprocess.run` raises `OSError` if the npm shim is a broken symlink, which `_install_npm_cli`'s `.installed` sentinel doesn't rule out; and `text=True` decodes with `errors='strict'`, so one invalid UTF-8 byte from the Codex TUI raises `UnicodeDecodeError`. The sibling `_try_run_claude_subagent_smoke` wraps its whole body instead — worth matching for consistency.

(The `except pytest.fail.Exception` below is correct, for the record: `_pytest.outcomes.Failed` derives from `BaseException`, not `Exception`, so a blanket `except Exception` would not have caught it.)

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -657,23 +645,63 @@ fn convert_input_items_to_messages(
...
+                            &f.name,
+                            &f.description,
+                            &f.parameters,
+                            f.strict,
+                            Some(&namespace.name),
+                        )?;
+                    }
+                }
+            }
+            _ => {}
```
dmitry-tokarev-nv, line 704: **NON-BLOCKING** — the explanatory comment was dropped in the rewrite. It's the only thing telling a reader that landing here is intentional rather than an unhandled case:

```suggestion
            // Only function tools are forwarded to chat completions
            _ => {}
```

Related, and worth a look before merge: `NamespaceToolParamTool::Custom` is silently dropped by the `if let` on line 690. Combined with the caller's `.filter(|t| !t.is_empty())`, a request whose namespace contains only custom tools reaches the backend as `tools: None` — the model answers in prose instead of calling a tool, with nothing in the logs. That mirrors pre-existing behavior for top-level custom tools, but this PR is what makes namespaced tools reach this code at all.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -990,6 +1019,31 @@ impl ResponseParams {
...
+                .then_some(namespace.name.as_str())
+        });
+        let namespace = namespaces.next()?;
+        namespaces
+            .all(|other_namespace| other_namespace == namespace)
+            .then(|| namespace.to_owned())
+    }
 }
 
 /// Normalize tools so that `FunctionTool.strict` is always set.
```
dmitry-tokarev-nv, line 1052: **NON-BLOCKING** — `normalize_tools` enforces this invariant only for top-level `Tool::Function`; `Tool::Namespace` falls through the `other => other` arm on line 1064.

The doc comment right here explains why that matters: `strict` is `skip_serializing_if = "Option::is_none"`, so `None` omits the field, and *"Schema validators (Zod, etc.) expect `strict` to always be present"*. Namespace-nested `FunctionToolParam.strict` is never filled — which is exactly the shape of this PR's own fixture, `{"type": "function", "name": "spawn_agent", "parameters": {...}}` with no `strict`.

`normalize_tools` runs on the echoed response body in both the unary and streaming paths, so after this PR a namespace-tool request comes back with `strict` present on top-level tools and absent on nested ones. Dormant on `main`, because namespace tools never reached the response echo.

The harness that would catch it is the bun/OpenResponses zod validator this very test file runs, and it surfaces as an opaque schema error rather than pointing here. None of the four new namespace tests assert on `strict`.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -990,6 +1019,31 @@ impl ResponseParams {
...
+                .tools
+                .iter()
+                .any(|tool| matches!(tool, NamespaceToolParamTool::Function(function) if function.name == name))
+                .then_some(namespace.name.as_str())
+        });
+        let namespace = namespaces.next()?;
+        namespaces
+            .all(|other_namespace| other_namespace == namespace)
+            .then(|| namespace.to_owned())
+    }
```
dmitry-tokarev-nv, line 1049: **NON-BLOCKING** — this is recomputed per tool call *and* per streaming delta, for a value that is immutable for the whole request.

`ResponseParams.tools` is cloned once when the request is accepted and never mutated, but `stream_converter.rs` calls this on every name-carrying delta with no "name unchanged" guard, and each hit allocates a fresh `String`. A routine shape — 50 tools x 10 subtools x 3 calls x 20 chunks — is ~33k string compares and ~60 allocations per response.

`convert_tools` already builds `origins: HashMap<String, Option<String>>`, which is precisely this lookup, and discards it. Carrying that map on `ResponseParams` (or keying it on `&str` borrowed from `tools`, whose lifetime outlives it) removes both this scan and the per-request owned-key allocation that happens today even when no namespace is present.

Worth noting the input isn't bounded: `validate_response_unsupported_fields` checks `background`, `previous_response_id`, `prompt` and `max_tool_calls`, but never the size of `tools`. The only ceiling is the request body limit, and a minimal namespaced entry is ~30 bytes, so the scan length is client-controlled on an endpoint that is frequently unauthenticated. Memoizing makes that moot.

lib/llm/src/protocols/openai/responses/stream_converter.rs
```diff
@@ -1362,6 +1368,35 @@ mod tests {
...
+            ..default_params()
+        };
+        let mut conv = ResponseStreamConverter::new("test-model".into(), params);
+        let _ = conv.process_chunk(&tool_call_chunk(
+            0,
+            Some("call-1"),
+            Some("spawn_agent"),
+            Some(r#"{"agent_type":"worker"}"#),
+        ));
+        let _ = conv.process_chunk(&finish_chunk(FinishReason::ToolCalls));
```
dmitry-tokarev-nv: **NON-BLOCKING** — this test asserts on the aggregated accessor only, so it covers neither of the two SSE sites this PR touches.

Both chunks' events are discarded (`let _ = conv.process_chunk(...)`) and the assertion reads `conv.completed_output()[0]`. The PR threads `namespace` through three independent construction sites: the `ResponseOutputItemAdded` event, the `output_item.done` event in `emit_completed_items`, and `completed_output`. Reverting the first two to `namespace: None` leaves this test green.

Those two are the ones that matter for the wire format: a client watching the stream would see `output_item.added` and `output_item.done` with no namespace while the terminal aggregated `Response.output` has one. A streaming agent harness acts on the streamed items, which is the mode Codex is actually in.

Asserting `namespace` on the `output_item.added` and `output_item.done` events from the returned event lists would close this with no new fixtures.

lib/llm/src/protocols/agents.rs
```diff
@@ -31,9 +32,9 @@ const AGENT_HEADER_MAPPINGS: &[AgentHeaderMapping] = &[
         infer_parent_from_session_for_child: true,
     },
     AgentHeaderMapping {
-        root_session_header: HEADER_CODEX_SESSION_ID,
+        root_session_header: HEADER_CODEX_THREAD_ID,
         child_session_header: None,
-        parent_session_header: None,
+        parent_session_header: Some(HEADER_CODEX_PARENT_THREAD_ID),
```
dmitry-tokarev-nv, line 37: **NON-BLOCKING** — two small follow-ups on the header swap, both additive to the threads above.

**1. `parent_session_id` has no length cap.** `AffinityCoordinator` enforces `MAX_SESSION_AFFINITY_ID_BYTES` (256) on `session_id` only (`session_affinity/coordinator.rs:407` and `:444`), and only when affinity is enabled. `x-codex-parent-thread-id` is new and unauthenticated, and flows into `agent_context` and the request trace with no bound at all. Worth capping it the same way at ingest.

**2. The `session-id` removal has no deprecation entry.** I take the point from the thread above that current Codex emits both headers, so supported clients are unaffected. That leaves pinned or older builds, which fail silently — no error, no warn log, no metric — and CI can't see it because `_install_npm_cli` installs `@openai/codex` at `@latest`. `docs/fern/pages/reference/general/releases/deprecations.mdx` has an open `## v1.4.0 (upcoming)` section already carrying Behavioral/Removed entries; one line there would cover it.

Also flagging that Devin's "session affinity granularity changes from per-run to per-thread" note on lines 105-113 looks unanswered. That one has a real KV-cache-hit-rate consequence for multi-agent runs and is worth an explicit yes/no before merge.

dmitry-tokarev-nv (commented):
Review scoped to CI, testing, security and maintainability, as requested. 13 items: **4 blocking**, 9 non-blocking. Posting as a comment rather than Request Changes — the blocking labels below are my assessment, not a formal gate.

I've skipped anything an existing thread already covers. In particular I am **not** re-raising "namespace collisions turn a previously-served request into a hard failure" (Devin, `responses/mod.rs`, Aug 8) — that's been answered and partly addressed. My comment on that same code is only about the error *type* and what gets logged, which is unaddressed. Likewise for the `session-id` removal: the "current Codex emits both headers" answer holds, so all I've left is a request for a deprecations entry.

### Blocking

| Where | What |
|---|---|
| `test_frontend_api_surface_compliance.py:594` | The gating Claude trace assertion can be satisfied by a leftover **Codex** record — the matcher ignores its `source_label`. The Claude mapping could break completely and stay green. |
| `test_frontend_api_surface_compliance.py:472` | The budget comment's own ceilings now sum to *exactly* the cap (1095), and a pytest-timeout kill doesn't match `only_rerun`, so there is no retry. |
| `test_frontend_api_surface_compliance.py:1009` | The header to `parent_session_id` mapping this PR exists to add has no gating coverage, while the docstring and `session-ids.mdx` both tell readers it does. |
| `responses/mod.rs:664` | Client input returns HTTP 500 instead of 400, and an unbounded client-supplied tool name is interpolated into ERROR-level logs. |

**Correction (after testing):** `agents.rs:83` was originally listed here as blocking. I ran it against the real crate and withdrew it — the behavior is correct, and my suggested fix there would have been a regression. Details in [that thread](https://github.com/ai-dynamo/dynamo/pull/12331#discussion_r3761704246); it's now a non-blocking coverage note.

Two of the remaining items are verified by execution rather than reading: `responses/mod.rs:664` reproduces as **500 / `ErrorType::Internal`** and the suggestion turns it into **400 / `Validation`**; and `stream_converter.rs:1386` stays green with both SSE sites reverted to `namespace: None`. Evidence is in those threads.

### Non-blocking

CI budget vs the 30-minute sglang stage cap (`:474`), probe/gating-step `cwd` isolation (`:586`), discarded timeout diagnostics (`:987`), the dropped `_ => {}` comment plus silently-dropped `Custom` tools (`mod.rs:701`), `normalize_tools` skipping namespace-nested `strict` (`mod.rs:1049`), the per-delta rescan in `namespace_for_function` (`mod.rs:1023`), the new stream-converter test covering neither SSE site (`stream_converter.rs:1386`), and a missing `parent_session_id` length cap plus a deprecations entry (`agents.rs:35`).

Four carry `suggestion` blocks that can be committed directly from the UI.

Checked and found clean, in case it saves anyone time: the namespace flatten/restore pair is symmetric, the generated `config.toml` is valid TOML in both branches, and `_SUBPROCESS_ENV_ALLOWLIST` keeps secrets out of the Allure artifact.

jthomson04 (approved):

lib/llm/src/protocols/agents.rs
dmitry-tokarev-nv: **Correction — I was wrong on this one. Downgrading from BLOCKING to a minor test-coverage note, and please do not apply my suggested fix.**

I ran it against the real crate on a Linux box (`cargo test -p dynamo-llm --lib`, PR head 369b46d3). The *mechanism* I described is exactly right, but my conclusion from it was not. Four cases through `agent_context_from_headers`:

| headers | `session_id` | `parent_session_id` |
|---|---|---|
| session=S, agent=A, parent=A (self-parent) | `A` | `Some("S")` |
| session=S, agent=A (no parent header) | `A` | `Some("S")` |
| session=S, agent=S, parent=S (root turn) | `S` | `None` |
| session=S, agent=B, parent=A (real parent) | `B` | `Some("A")` |

Row 1 and row 2 **converge**. That is the point I missed: for the `infer_parent_from_session_for_child: true` mapping, "ignore the self-referential header" correctly falls through to the designed inference, and lands on exactly the same answer as a request that never sent the header at all. Row 3 confirms there's no false lineage on root turns, and row 4 confirms real parents are untouched.

So this is not a fabricated edge. It is strictly better than the old behavior, which produced `Some("A")` — a self-loop that a lineage walker would have to defend against. My suggested chain (trailing `.filter` after `.or_else`) would turn row 1 into `None` while leaving row 2 as `Some("S")`, splitting two cases that should agree and *losing* lineage the design intends to infer. Applying it would be a regression. Sorry for the noise.

What I think does survive, as a **non-blocking** note: `agent_context_ignores_self_parent_header` asserts `parent_session_id == None`, and that contract is only true for the two `infer = false` mappings. The Claude Code mapping returns `Some(root)` for the same input shape and has no coverage. A one-line addition to the test table pinning row 1 to `Some("S")` would document the intended difference and stop a future refactor from silently changing it.

The `session-ids.mdx` note in my last paragraph still stands on its own.

lib/llm/src/protocols/openai/responses/mod.rs
dmitry-tokarev-nv: **Verified end-to-end on the PR head** (`cargo test -p dynamo-llm --lib`, Linux). I fed the PR's own collision fixture (`crm.lookup` + `billing.lookup`) through `NvCreateChatCompletionRequest::try_from` and passed the resulting error to the real `responses_conversion_error_response`:

```
PROBE raw error: Responses function tool name 'lookup' is ambiguous after namespace flattening
PROBE collision -> status=500 message="Failed to convert responses request" type=Internal
```

So both halves are confirmed: client input returns **500 / `ErrorType::Internal`** with no detail reaching the caller, and the client-supplied name (`'lookup'`) is in the error string that gets logged.

Applying the `suggestion` above and re-running the same probe:

```
PROBE collision -> status=400 message="Failed to convert responses request: Responses function tool names are ambiguous after namespace flattening" type=Validation
```

**400 / `Validation`**, and the tool name no longer appears. `cargo check -p dynamo-llm` is clean with it applied — no errors, no unused-variable warning (`name` is still used by the `origins.insert` below).

lib/llm/src/protocols/openai/responses/stream_converter.rs
dmitry-tokarev-nv: **Verified by mutation on the PR head.** I reverted both SSE construction sites to `namespace: None` — line 491 (`ResponseOutputItemAdded`) and line 631 (the `output_item.done` event in `emit_completed_items`) — leaving only `completed_output` (line 697) threading the namespace, then ran this test:

```
test protocols::openai::responses::stream_converter::tests::test_function_call_preserves_namespace ... ok
test result: ok. 1 passed; 0 failed
```

Green with both streaming paths broken. The test covers one of the three sites the PR touches.

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -712,22 +721,47 @@ def _read_request_trace_records(path: Path) -> list[dict]:
...
 
 
 def _read_request_trace_records_since(path: Path, start_index: int) -> list[dict]:
-    return _read_request_trace_records(path)[start_index:]
+    return _parse_request_trace_records(_read_request_trace_lines(path)[start_index:])
+
+
+@pytest.mark.unit
+@pytest.mark.gpu_0
+@pytest.mark.pre_merge
```
dmitry-tokarev-nv, line 745: **NON-BLOCKING, but this test never runs.** The regression test added here for the cursor off-by-one is missing the backend marker, so no CI job selects it.

Every marker expression in `pr.yaml` requires one — `pre_merge and sglang and gpu_0` (line 860), and the same shape for vllm/trtllm/planner. This test has `unit + gpu_0 + pre_merge` and no `sglang`. The sibling unit test in this same file (`test_patch_openresponses_suite_adds_output_cap`, line 351) carries `sglang` and `core` for exactly that reason.

Verified with pytest, using the two markers sets copied verbatim into an isolated file:

```
$ pytest --collect-only -q -m "pre_merge and sglang and gpu_0"
test_markers.py::test_patch_openresponses_suite_adds_output_cap

1/2 tests collected (1 deselected) in 0.00s
```

The sibling is collected; this one is deselected. So the fix is real but unguarded — a future refactor of `_request_trace_line_count` back to record-counting would reintroduce the off-by-one with nothing to catch it.

```suggestion
@pytest.mark.unit
@pytest.mark.sglang
@pytest.mark.core
@pytest.mark.gpu_0
@pytest.mark.pre_merge
```

dmitry-tokarev-nv (commented):
Reviewed `399e72e9f0`. This addresses every item from my review, and in several places with a better fix than I proposed. I re-ran things rather than reading the diff — results below.

**Verified by execution** (`cargo test -p dynamo-llm --lib` on Linux at this head):

- **2003 passed, 0 failed.** The `OnceLock` field on `ResponseParams` doesn't break any derive, and the only external construction site (`openai.rs:3079`) is updated.
- **The SSE gap is genuinely closed.** I re-applied the exact mutation that passed before — `namespace: None` at lines 491 and 631 — and `test_function_call_preserves_namespace` now fails with `response.output_item.added did not preserve the namespace`. Before this commit that mutation was green.
- **The 975 budget is arithmetically sound.** `60+60+180+180+35+180+45+180 = 920`, so the 55s of claimed headroom is exactly right. That also resolves the scheduler tension I raised on line 472: the marker goes back to 975, so `est_duration` returns to 325s while real runtime drops by ~175s. Both consumers are now served by the same number.

**The best change here is `_run_codex_parent_header_check`.** Replacing the model-mediated probe with an explicit `/v1/responses` POST that asserts both `expected_session_id` and `expected_parent_session_id` turns the PR's headline behavior from untested into a deterministic gate, and drops 210s of GPU `pre_merge` time to 35s in the process. It also makes the `session-ids.mdx:42` claim true rather than aspirational. That one change closes four of my findings at once.

Also confirmed fixed: the typed `InvalidArgument` (with the client-supplied name no longer interpolated) plus a test pinning 400/`Validation`; line-based trace cursor; `MAX_SESSION_AFFINITY_ID_BYTES` on `parent_session_id` in both branches; namespace-nested `strict`; memoized `function_namespaces`; the restored `_ => {}` comment; the deprecations entry; and the `X-Dynamo-Parent-Session-ID` doc text. The affinity-granularity note in `session-ids.mdx` also answers Devin's open thread.

**One new issue**, inline: the added regression test carries no backend marker, so no CI job collects it.

**One question, on `agents.rs`.** You moved the self-parent filter after `.or_else` — which is the chain I proposed and then [withdrew](https://github.com/ai-dynamo/dynamo/pull/12331#discussion_r3761704246) as a regression. I still think my withdrawal was right on the facts, and you've since made the contract uniform and tested it across both mappings, which is a coherent answer. Just confirming one consequence is intended: a Claude Code request with `parent-agent-id == agent-id` now yields `parent_session_id = None`, while the *same request omitting the header* yields `Some(root)` via inference. So a malformed parent header is now worse than no parent header. Fine if that's the intended reading of "ignore"; worth a line in the test if so.

**Not addressed, and still fine to defer:** `NamespaceToolParamTool::Custom` is still silently dropped at `mod.rs:690`, so a namespace containing only custom tools still reaches the backend as `tools: None`.

tests/frontend/test_frontend_api_surface_compliance.py
dmitry-tokarev-nv: Closing this — both halves are handled.

The concrete half, the cursor off-by-one from torn JSONL lines, is fixed in the right shape: `_request_trace_line_count` counts lines and `_read_request_trace_records_since` slices lines *before* parsing, so a half-flushed final line no longer shifts the cursor.

The source-blind half is closed in practice rather than in the matcher. `_run_codex_parent_header_check` is synchronous and waits for its own trace record before returning, so there is no longer another traffic source in flight when `claude_trace_start` is taken. `_trace_contains_agent_context` still matches any record by default, but for the CLI-driven steps the session IDs are generated inside the agent process and aren't available to assert on, so that looks inherent rather than something to fix here.

One caveat tracked separately: the regression test added for this carries no backend marker, so CI never collects it — [details here](https://github.com/ai-dynamo/dynamo/pull/12331#discussion_r3761987162).

lib/llm/src/protocols/openai/responses/mod.rs
dmitry-tokarev-nv: Comment restored — thanks.

Leaving this thread open for the second half only: `NamespaceToolParamTool::Custom` is still silently dropped by the `if let` on line 690, so a namespace containing only custom tools still reaches the backend as `tools: None` and the model answers in prose instead of calling a tool.

Entirely fine as a follow-up issue rather than something for this PR — I just don't want it to disappear along with the comment fix.

dmitry-tokarev-nv (approved):
Thank you for addressing comments. remaining issues are non-blocking

tests/frontend/test_frontend_api_surface_compliance.py
```diff
@@ -789,7 +853,7 @@ def _trace_contains_agent_parent_context(records: list[dict]) -> bool:
         session_id = agent_context.get("session_id")
         if not session_id:
             continue
-        if parent_session_id == session_id:
+        if parent_session_id == session_id or parent_session_id not in session_ids:
             continue
```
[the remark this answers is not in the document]
ishandhanani: Fixed in a8a5d6b417: the observed-parent requirement now applies only to the Codex child assertion; OpenCode keeps the existing distinct-parent check. Verified with Ruff, py_compile, and the changed-file pre-commit hooks.

PR #12332, opened 2026-07-29
fix(llm): normalize tool_call arguments from JSON string to object before template render

peilii:
## Summary

GLM-5.2's Jinja chat template calls `.items()` on `tool_calls[*].function.arguments`, expecting a dict. The OpenAI wire format carries arguments as a JSON string. In multi-turn conversations where the assistant's previous turn contained tool calls, the template renderer receives a string where it expects a dict, producing empty or garbled argument output.

**Before** (arguments as JSON string passed to the template):
```
template sees: tc.function.arguments = '{"path": "/tmp/foo"}'
.items() on a string → empty iteration
rendered:      <tool_call>read_file</tool_call>
```

**After** (arguments normalized to a dict before the template):
```
template sees: tc.function.arguments = {"path": "/tmp/foo"}
.items() works correctly
rendered:      <tool_call>read_file<arg_key>path</arg_key><arg_value>/tmp/foo</arg_value></tool_call>
```

This matches TRT-LLM's behavior: `chat_utils.py` always calls `json.loads(arguments)` before `apply_chat_template` when the value is a string. No template detection or thread-local plumbing is needed — `normalize_tool_call_arguments()` is called unconditionally in `messages()` for all three `OAIChatLikeRequest` implementors (`NvCreateChatCompletionRequest`, `UnifiedRequest`, `SpeculativePrefillRequest`). Non-object and malformed-JSON values are replaced with `{}` with a warning (byte length only — no raw content in logs).

## Validation

- `cargo test -p dynamo-llm`: all tests pass
- `cargo clippy --no-deps -p dynamo-llm -- -D warnings`: clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for tool calls by consistently normalizing function arguments before messages are processed.
  * Invalid or incorrectly formatted tool-call arguments are now safely converted to an empty object instead of causing processing issues.
  * Applied the same handling across standard, speculative prefill, and unified chat requests.


lib/llm/src/preprocessor/prompt.rs
```diff
@@ -36,13 +36,63 @@ pub trait MediaRequestExt {
...
+
 impl OAIChatLikeRequest for NvCreateChatCompletionRequest {
     fn model(&self) -> String {
         self.inner.model.clone()
     }
 
     fn messages(&self) -> Value {
-        let messages_json = serde_json::to_value(&self.inner.messages).unwrap();
+        let mut messages_json = serde_json::to_value(&self.inner.messages).unwrap();
+        normalize_tool_call_arguments(&mut messages_json);
```
rmccorm4: **[P1] Shared normalization breaks native DeepSeek V3.2 rendering.** The new mutation in [OAIChatLikeRequest::messages()](https://github.com/ai-dynamo/dynamo/blob/6889de80f26dcbd40705cc9a7cad5118f0e17e49/lib/llm/src/preprocessor/prompt.rs#L93-L96) converts arguments to objects for every renderer, but DeepSeek V3.2 requires the original JSON string. Both local testing and [GitHub Actions](https://github.com/ai-dynamo/dynamo/actions/runs/30650810134/job/91223385279) fail the same two existing tests with `Missing or invalid 'arguments' field`. Move normalization to the MiniJinja boundary.

rmccorm4 (commented):
See inline finding.

rmccorm4 (commented):
See three inline findings.

lib/llm/src/preprocessor/prompt.rs
```diff
@@ -36,13 +36,63 @@ pub trait MediaRequestExt {
...
+            .get_mut("tool_calls")
+            .and_then(serde_json::Value::as_array_mut)
+        else {
+            continue;
+        };
+        for tc in tool_calls.iter_mut() {
+            let Some(args_str) = tc.pointer("/function/arguments").and_then(|v| v.as_str()) else {
+                continue;
+            };
+            let value = match serde_json::from_str::<serde_json::Value>(args_str) {
```
rmccorm4: **[P1] Valid large integers are silently corrupted.** This parses valid JSON through the default `serde_json::Value` number representation, but `serde_json/arbitrary_precision` is not enabled. A focused exact-head probe changed `123456789012345678901234567890` to `1.2345678901234568e+29`, silently changing tool-call history before it reaches the model. Please preserve the original numeric value/spelling during template normalization and add a regression case.

lib/llm/src/protocols/unified.rs
```diff
@@ -463,7 +463,8 @@ impl OAIChatLikeRequest for UnifiedRequest {
     }
 
     fn messages(&self) -> minijinja::value::Value {
-        let messages_json = serde_json::to_value(&self.inner.inner.messages).unwrap();
+        let mut messages_json = serde_json::to_value(&self.inner.inner.messages).unwrap();
+        crate::preprocessor::prompt::normalize_tool_call_arguments(&mut messages_json);
```
rmccorm4: **[P2] Malformed Responses history is fabricated as `{}`.** This applies normalization after conversion, but the converted request does not run the direct Chat validator that rejects malformed or non-object argument strings. A focused Responses probe confirmed that malformed `function_call.arguments` reaches this path and is silently replaced with `{}`. That also differs from TensorRT-LLM, which rejects invalid arguments. Please validate the converted request or return an invalid-request error instead of inventing empty arguments.

lib/llm/src/preprocessor/prompt.rs
```diff
@@ -36,13 +36,63 @@ pub trait MediaRequestExt {
...
+                Ok(_) => {
+                    // Scalar or array — GLM's .items() would panic at render time.
+                    tracing::warn!(
+                        args_len = args_str.len(),
+                        "tool_call arguments parsed to a non-object; \
+                         substituting {{}} for template safety"
+                    );
+                    serde_json::Value::Object(serde_json::Map::new())
+                }
+                Err(_) => {
```
rmccorm4: **[P2] Valid no-argument calls generate warning spam.** [`validate_json_object_string`](https://github.com/ai-dynamo/dynamo/blob/6889de80f26dcbd40705cc9a7cad5118f0e17e49/lib/llm/src/protocols/openai/validate.rs#L461-L496) explicitly permits `arguments: ""`, but that valid no-argument representation enters this error arm and emits `warn!` for every occurrence. Please special-case empty input without warning, or make validation and normalization enforce the same contract.

lib/llm/src/preprocessor/prompt.rs
peilii: Fixed — moved `normalize_tool_call_arguments` out of `NvCreateChatCompletionRequest::messages()` (prompt.rs) and `SpeculativePrefillRequest::messages()` (speculative_prefill.rs). Normalization now lives only in `UnifiedRequest::messages()` (unified.rs), which is the MiniJinja / GLM rendering path. The DeepSeek formatter uses the `NvCreateChatCompletionRequest` path and receives arguments as the original JSON string. All 12 DeepSeek V3.2 tests pass.

lib/llm/src/preprocessor/prompt.rs
peilii: Fixed — introduced `parse_args_object()` which uses `serde_json::value::RawValue` to detect precision loss. For each numeric value, it compares the serde_json round-tripped representation against the raw token from the source string; if they differ (i.e. a >15-digit integer was corrupted to f64), the value is kept as `Value::String` with the original spelling. This avoids enabling `arbitrary_precision` crate-wide (which broke the logprob SSE test). Regression test `normalize_preserves_large_integer` added.

lib/llm/src/protocols/unified.rs
peilii: Fixed — `normalize_tool_call_arguments()` now returns `anyhow::Result<()\>`. On malformed JSON it returns `Err` instead of substituting `{}`. The caller in `UnifiedRequest::messages()` handles the error by logging at `error!` level and returning the un-normalized messages (arguments remain as the original JSON string), so the template fails explicitly on `.items()` rather than silently using empty args. Regression test `normalize_malformed_json_returns_error` added.

lib/llm/src/preprocessor/prompt.rs
peilii: Fixed — empty string is now special-cased before the `serde_json::from_str` call: it normalises silently to `{}` without entering the `Err` arm or emitting a warning. Regression test `normalize_empty_string_becomes_empty_object` added.

indrajit96:
E2E tests caught this a multiturn failure impacting GPT OSS below is the flow for the failing test (repro shared offline)
 1. User asks for weather in San Francisco and New York.
 2. GPT-OSS returns a structured get_weather call for San Francisco.
 3. The test executes the mock tool and adds the San Francisco result to the conversation.
  4. GPT-OSS attempts to call get_weather for New York.
  5. Dynamo incorrectly places that call’s Harmony markup in reasoning_content instead of returning a structured tool call.
  6. The test cannot execute the New York tool and receives no final summary, so it fails.


Why this PR could have caused this?

 Possible GPT-OSS regression from the unconditional normalization in UnifiedRequest::messages() (https://github.com/ai-dynamo/dynamo/blob/420d11ab03a5b3df7db2f07cf1af67ccb483bec0/lib/llm/src/protocols/unified.rs#L465-L469). The normalizer converts historical function.arguments from a JSON string into an object at prompt.rs:188–228 (https://github.com/ai-dynamo/dynamo/blob/420d11ab03a5b3df7db2f07cf1af67ccb483bec0/lib/llm/src/preprocessor/prompt.rs#L188-L228).

This affects every MiniJinja model, including GPT-OSS, rather than only GLM 5.2. Our first tool call succeeds, but once that call is present in conversation history, the follow-up Harmony call is returned inside reasoning_content instead of as a structured tool call. This reproduced in 19/20 requests. 
Could this normalization be opt-in for templates that require object arguments, with a GPT-OSS multi-turn regression test added?

Recommended Fix : 

Make the change OPT-IN
I recommend moving this normalization out of `UnifiedRequest::messages()` and making it an explicit template capability. 
Preserve JSON-string arguments by default and opt GLM 5.2 into object arguments. 
This avoids changing the rendered conversation history for every MiniJinja model, including GPT-OSS. A model-name check here would be fragile because served-model aliases may differ; the renderer/model-card configuration is the safer boundary (But I will leave that to you)

Agent instructions :

  1. Remove the unconditional call to normalize_tool_call_arguments() from UnifiedRequest::messages().
  2. Add a template-level option such as:
```
  enum ToolCallArgumentsFormat {
      JsonString, // default, backward compatible
      JsonObject,
  }
```
  3. Normalize arguments immediately before rendering only when the selected template opts into JsonObject.
  4. Enable that option for GLM 5.2; leave GPT-OSS and other models unchanged.
  5. Add two multi-turn prompt tests:
      - GLM 5.2: historical arguments render as an object.
      - GPT-OSS: a previous tool call can be followed by another structured Harmony call.

peilii:
Thanks for the clear root-cause analysis. Fixed in b359482cdf.

`normalize_tool_call_arguments()` is removed from `UnifiedRequest::messages()` — it no longer runs for GPT-OSS or any other MiniJinja model by default.

The opt-in is a new `ToolCallArgumentsFormat { JsonString (default), JsonObject }` field in `ModelRuntimeConfig`. GLM-5.2 opts in automatically when `tool_call_parser = "glm47"` is set (no config change needed for existing deployments), or explicitly via `tool_call_arguments_format: json_object`. The normalization is applied through a `NormalizedArgsRequest<R>` wrapper inside `apply_template` so the path is isolated to models that need it.

Added two regression tests: `glm52_historical_arguments_are_object` (JsonObject mode) and `gptoss_historical_arguments_remain_string` (JsonString default, Harmony path unaffected).

indrajit96:
/ok to test [b501d2b](https://github.com/ai-dynamo/dynamo/pull/12332/commits/b501d2b161091eb67bd7c706adc7129373deea1c)

indrajit96:
/ok to test [2580a36](https://github.com/ai-dynamo/dynamo/pull/12332/commits/2580a363e8616aff50c1d53750fc24a9c85c09e9)

indrajit96 (approved):
LGTM!
Thanks for scoping the changes to GLM only!

KrishnanPrash (approved):

PR #12333, opened 2026-07-29
fix(llm): recover truncated glm47 <tool_call> blocks on finish_reason=length

peilii:
## Summary

GLM-5.2 emits tool calls as XML (`<tool_call>fn<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>`). When `max_tokens` truncates the response inside a `<tool_call>` block, the partial block is silently dropped and the client receives an incomplete turn. This PR recovers the truncated block as raw content in both the non-streaming and streaming paths.

### Non-streaming (`aggregator.rs`)

Before calling the parser, `rfind("<tool_call>")` locates the last opened block. If no `</tool_call>` follows it on a `finish_reason=length` response, the tail is saved. After parsing (which may extract a complete first tool call), the tail is appended to `choice.text` if not already present. The first complete tool call is returned as a structured `tool_calls` entry; the truncated second block is returned as raw content.

### Streaming (`preprocessor.rs`)

A per-choice `ChoiceRecovery` struct buffers raw input text in the jail-input stream. On `finish_reason=length`, the same `rfind` logic finds the last incomplete block and emits it as a synthetic content delta before the finish chunk. Using `rfind` means already-parsed complete blocks are never duplicated as raw content.

### Note on fix location

The recovery lives in the frontend stream rather than inside `dynamo-parsers` because the parser does not see SSE protocol events like `finish_reason`. By the time `finish_reason=length` arrives, the parser's internal state for the partial block has already been discarded. Pushing this into the parser would require a new `flush_on_truncation()` interface in `dynamo-parsers` and threading the truncation signal into the jail — a larger change deferred to a follow-up. Deferring to @keivenc for parser-level feedback.

## Validation

- `cargo test -p dynamo-llm`: all tests pass
- `cargo clippy --no-deps -p dynamo-llm -- -D warnings`: clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GLM-47 handling of truncated tool calls in streaming responses.
  * Preserved incomplete tool-call content when responses end due to length limits.
  * Prevented duplicated content and ensured finish indicators appear only once.
  * Improved handling of truncated responses containing multilingual text and emoji.

* **Tests**
  * Added coverage for streaming recovery, multiple tool calls, clean completions, and UTF-8 boundaries.


lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3012,59 @@ impl OpenAIPreprocessor {
...
+            let mut recoveries: Vec<Annotated<NvCreateChatCompletionStreamResponse>> = Vec::new();
+            if is_glm47 && let Some(ref data) = nv_chunk.data {
+                let mut cr = choice_recovery.lock().expect("choice recovery poisoned");
+                for choice in &data.inner.choices {
+                    let state = cr.entry(choice.index).or_default();
+                    if matches!(
+                        choice.finish_reason,
+                        Some(dynamo_protocols::types::FinishReason::Length)
+                    ) {
+                        let recovered = state.input_text.rfind("<tool_call>").and_then(|start| {
```
keivenchang: i think this could double-emit — theres no already-recovered guard here, but the aggregator path has one (`!choice.text.contains(&tail)` at `aggregator.rs:458`).

it'll bite when trailing prose and the truncated block land in the same jailed buffer. glm47 isnt in `suppresses_tool_call_markup()` (thats hermes/qwen25/jamba only), so the jail just releases the whole thing verbatim. ran it on 5.1.3:

```
in:   <tool_call>a<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x
jail: content="tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x"     <- already out
this: content="<tool_call>b<arg_key>k</arg_key><arg_value>x"                <- second copy
```

so the client gets the markup twice.

how about tracking what the jail actually sent per choice, and skipping if the tail's already in there?

```rust
#[derive(Default)]
struct ChoiceRecovery {
    input_text: String,
    emitted_text: String,
    recovered: bool,
}
```

one gotcha — push before you check, not after. the verbatim release lands on the same chunk that carries the finish reason, so checking first misses it:

```rust
if let Some(ChatCompletionMessageContent::Text(t)) = &choice.delta.content {
    state.emitted_text.push_str(t);
}
```

then gate on both:

```rust
if state.recovered || state.emitted_text.contains(&recovered) {
    continue;
}
state.recovered = true;
```

the latch and the `contains` are doing different jobs — one stops a second `length` chunk re-firing, the other stops duplicating what the jail already sent, so i think you want both. nice side effect is `emitted_text` ends up being the streaming mirror of the aggregator's `choice.text`, so the two guards cant really drift apart.

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3012,59 @@ impl OpenAIPreprocessor {
...
+                            }
+                        });
+                        if let Some(recovered) = recovered {
+                            tracing::warn!(
+                                choice_index = choice.index,
+                                recovered_bytes = recovered.len(),
+                                "glm47 streaming: partial <tool_call> emitted as content \
+                                 on length finish"
+                            );
+                            let mut rec = nv_chunk.clone();
```
keivenchang: small one — the clone also drags along `logprobs`, `event`, `comment` and `error`. `DeltaAggregator` extends content logprobs so they end up counted twice, and a cloned `error` gets surfaced to the client twice.

you can just null them next to the ones you already do:

```rust
rec.event = None;
rec.comment = None;
rec.error = None;
```

plus `rc.logprobs = None;` inside the `for rc in &mut rd.inner.choices` loop.

`id` i'm honestly not sure about — its the SSE annotation id and duplicating it is probably harmless. i'd lean toward nulling it too, just because a synthesized chunk claiming to be the same event as a real one feels off, but happy to defer on that one.

lib/llm/src/preprocessor.rs
```diff
@@ -2938,13 +2938,41 @@ impl OpenAIPreprocessor {
         let pending = Arc::new(Mutex::new(PendingMetrics::default()));
         let pending_in = Arc::clone(&pending);
 
+        // Per-choice recovery state — allocated only for glm47 since only that
+        // parser emits <tool_call> XML that can be truncated at max_tokens.
+        // Buffers raw input per choice.index so n > 1 is handled correctly.
+        #[derive(Default)]
+        struct ChoiceRecovery {
+            input_text: String,
```
keivenchang, line 3368: this holds the whole generation per choice and never clears, when the only thing you ever read is the tail after the last `<tool_call>`. on a long response thats a full second copy of the output hanging around for the life of the stream.

how about dropping everything before the last marker as you append?

```rust
const START: &str = "<tool_call>";
let entry = cr.entry(choice.index).or_default();
entry.input_text.push_str(content);
let mut keep_from = match entry.input_text.rfind(START) {
    Some(pos) => pos,
    // no marker yet — keep just enough tail to match one split across chunks
    None => entry.input_text.len().saturating_sub(START.len() - 1),
};
while keep_from > 0 && !entry.input_text.is_char_boundary(keep_from) {
    keep_from -= 1;
}
entry.input_text.drain(..keep_from);
```

one warning though, and its the reason i wrote out the whole snippet: `drain(..n)` panics if n isnt a char boundary, and that `None` arm slices by raw byte count. so a multibyte char landing in the last 10 bytes takes down the frontend on completely ordinary output. thats what the `is_char_boundary` walk-back is for. could you add a test with a CJK char or emoji straddling that boundary?

and honestly if the walk-back doesnt make it in, i'd rather you just left the buffer unbounded — the memory isnt worth a panic.

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3012,59 @@ impl OpenAIPreprocessor {
...
+                                    rc.finish_reason = None;
+                                }
+                                recoveries.push(rec);
+                            }
+                        }
+                    }
+                }
             }
+
+            let mut out = Vec::with_capacity(recoveries.len() + 1);
```
keivenchang: `Vec::with_capacity(recoveries.len() + 1)` now runs on every chunk of every tool-calling stream, glm47 or not, since `.map` became `.flat_map`. an empty `Vec` doesnt actually allocate in rust, so `recoveries` is free — this line is the whole cost.

you can drop these three lines and chain instead:

```rust
futures::stream::iter(recoveries.into_iter().chain(std::iter::once(nv_chunk)))
```

no `is_glm47` fast path needed once `out` is gone.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -388,6 +388,25 @@ impl DeltaAggregator {
...
                 // experimental_v2_batch_eligible — see tool_parser_v2::batch_tool_choice_eligible)
                 // keep the v1 finalize path.
+                // Extract the truncated tail BEFORE parsing so a second <tool_call>
+                // truncated after a complete first one is not silently dropped.
+                let glm47_truncated_tail = if parser == "glm47"
+                    && matches!(
+                        choice.finish_reason,
+                        Some(dynamo_protocols::types::FinishReason::Length)
+                    ) {
+                    choice.text.rfind("<tool_call>").and_then(|start| {
```
keivenchang: these are hardcoded here and again over in `preprocessor.rs`, but `Glm47ParserConfig` (`config.rs:249-250`) makes them serde-configurable. so if anyone overrides them, these two sites just quietly stop matching — no error, recovery silently stops firing.

you can pull them off the config instead. its reachable even though the type isnt re-exported at `tool_calling`:

```rust
use dynamo_parsers::tool_calling::config::Glm47ParserConfig;

let cfg = Glm47ParserConfig::default();
// cfg.tool_call_start, cfg.tool_call_end
```

another way to look at it is this is really the same change as the shared helper i mentioned up top — helper reads the tokens once, both sites use it. probably worth doing as one commit, since fixing the literals on their own still leaves two copies to keep in sync.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -416,9 +435,38 @@ impl DeltaAggregator {
...
+                {
+                    tracing::warn!(
+                        parser,
+                        "glm47: partial <tool_call> returned as content on length finish"
+                    );
+                }
+
+                // Recover any tail saved before parsing — the parser drops a truncated
+                // second block silently when an earlier complete block was already parsed.
+                if let Some(tail) = glm47_truncated_tail
```
keivenchang, line 445: i think streaming and batch disagree here. `rfind` starts at the marker, so batch drops the prose between the last `</tool_call>` and the truncated block, but on the streaming side the jail already emitted it:

```
in:     <tool_call>a<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x

batch:  content="<tool_call>b<arg_key>k</arg_key><arg_value>x"
stream: content="tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x"
```

same bytes in, different `content` out.

whichever way we go it should be identical in both, and i think it falls out of the shared helper for free — as long as streaming emits the helper's output rather than letting the jail's verbatim release through (the `contains` guard on the other comment is what suppresses that copy).

if it helps, my vote is batch's version, marker onward only. glm47's parser drops post-call prose on purpose — *"Only surface normal text that precedes the first parsed call ... matches vLLM's glm47_moe_tool_parser"* — so streaming keeping it isnt really a decision anyone made, its fallout from glm47 not being in `suppresses_tool_call_markup()`. would rather not lock that accident in as the contract, especially since a parity fixture will freeze whatever shape we pick.

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3012,59 @@ impl OpenAIPreprocessor {
...
                 comment: a.comment,
                 error: a.error.map(DynamoError::msg),
+            };
+
+            // glm47: on finish_reason=length, recover the last incomplete <tool_call>
+            // block. rfind skips complete blocks, so earlier parsed tool calls are
+            // never duplicated as raw content. Recovered content WILL contain raw
+            // <tool_call> markup — callers that require strict "no tool tags in content"
+            // must filter on finish_reason=length.
+            let mut recoveries: Vec<Annotated<NvCreateChatCompletionStreamResponse>> = Vec::new();
```
keivenchang: not asking you to fix this one — glm47 isnt in `V2_FAMILIES` yet so nothing's broken today, and hoisting it is a bigger change than this PR should carry. but could you leave a TODO so it doesnt get lost?

heres what'll happen. this recovery lives inside `apply_tool_calling_jail`, and the v2 streaming path skips the jail entirely (`preprocessor.rs:2413`):

```rust
if should_jail && use_parsers_v2 {
    Box::pin(tool_parser_v2::apply_stream(...))     // no jail
} else if should_jail {
    Box::pin(Self::apply_tool_calling_jail(...))    // <- this code
}
```

so the day someone adds `"glm47"` to `V2_FAMILIES` (`tool_parser_v2.rs:42`, currently just `["qwen3_coder", "deepseek_v4"]`), streaming quietly stops recovering while the `aggregator.rs` copy keeps going — that one's computed outside the v2/v1 branch at `:393`. nothing fails to compile, nothing goes red.

something like:

```rust
// TODO: this recovery runs inside apply_tool_calling_jail, which the v2 path
// bypasses (see the use_parsers_v2 branch in preprocessor_stream). Adding
// "glm47" to V2_FAMILIES in tool_parser_v2.rs silently disables it for
// streaming, while the aggregator.rs copy keeps running. Hoist this above the
// jail/v2 branch at that point — it only needs the buffered input text plus
// finish_reason, both available there.
```

two reasons i think it needs to be a comment and not just a ticket. v2 doesnt make the problem go away — `tool_parser_v2.rs:11` says it *"drops a parameter value truncated at EOF rather than guessing it"*, same policy, so the gap just reopens. and that module's own doc says adding a family is *"a one-line change here plus support in that crate"* — whoevers doing that one-liner has no reason to ever open this file.

keivenchang (commented):
hey Pei, thanks for this — the bug is real, I reproduced the drop on `dynamo-parsers` 5.1.3:

```
in:  <tool_call>get_weather<arg_key>city</arg_key><arg_value>Bos   (finish_reason=length)
out: finish=length  tool_calls=None  content=""
```

empty turn, no way for the client to tell "model said nothing" from "we ate 58 bytes". Fine by me to keep the fix in `lib/llm`.

One ask though: route both paths through a refactored single helper instead of the two `rfind` copies (`preprocessor.rs:3031`, `aggregator.rs:398`). They've already diverged — same input, different `content`:

```
in:     <tool_call>a<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x

batch:  content="<tool_call>b<arg_key>k</arg_key><arg_value>x"
stream: content="tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x"
```

one `fn(&str, Option<FinishReason>) -> Option<String>` with both call sites on it makes that unrepresentable, and gets the token literals off hardcoded strings — `Glm47ParserConfig` (`config.rs:249-250`) makes them serde-configurable.

Worth a comment on that helper pointing at `parsers.rs:186` (`// GLM-4.7 intentionally omitted: match upstream vLLM/SGLang behavior`) so the next person doesn't read the two repos as accidentally contradicting each other.

Also no tests — 130 lines of new conditional behavior, none of it reachable from the conformance fixtures. Inline comments have the rest; the first is a confirmed double-emit.

Looking forward to getting this in.

peilii:
Thanks for the thorough review @keivenchang, all good catches.

**Double-emit + prose consistency**: Added `emitted_text` per choice to track what the jail releases verbatim. On the finish chunk, pass 2 now strips from the `<tool_call>` marker onwards before the recovery chunk is emitted, so streaming produces just the marker-onwards tail — same as the aggregator path. The `recovered` latch stops a second length chunk re-firing.

**Clone side effects**: Nulled `id`, `event`, `comment`, `error`, and `logprobs` on the synthetic chunk.

**Bounded buffer**: Draining `input_text` back to the last marker after each append. Added the `is_char_boundary` walk-back you suggested — good catch on the panic path.

**Vec → chain**: Done.

**Config tokens**: Both sites now read `Glm47ParserConfig::default().tool_call_start/end` so they stay in sync if the defaults ever move.

**V2 TODO**: Added both at `V2_FAMILIES` and inline near the streaming recovery block.

One thing I want to double-check: in pass 2 I strip from `t.find(pr.tail)` which removes the marker-onwards portion from the jail's content chunk. If there's prose before the marker in that final chunk (e.g. "tail prose "), it survives as content unless the whole string is empty. Is that acceptable, or should I strip everything from the last `</tool_call>` onwards to silence all post-call prose in the finish chunk?

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3042,133 @@ impl OpenAIPreprocessor {
...
+                        .iter_mut()
+                        .filter(|c| c.index == pr.choice_idx)
+                    {
+                        if let Some(ChatCompletionMessageContent::Text(ref mut t)) =
+                            rc.delta.content
+                        {
+                            // Strip from the marker onwards (and any post-call
+                            // prose that precedes it in this chunk).
+                            if let Some(marker_pos) = t.find(pr.tail.as_str()) {
+                                t.truncate(marker_pos);
```
keivenchang: i think this doesnt quite land where the comment says it does. `truncate(marker_pos)` keeps everything *before* the marker, so the post-call prose survives — but the comment above says this is "matching the non-streaming path which uses rfind and produces marker-only content (no post-call prose)". measured on 5.1.3, same input both ways:

```
in:     <tool_call>a<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>tail prose <tool_call>b<arg_key>k</arg_key><arg_value>x

stream: "tail prose " + "<tool_call>b<arg_key>k</arg_key><arg_value>x"
batch:  "<tool_call>b<arg_key>k</arg_key><arg_value>x"
```

so they still disagree. how about either cutting the prose too so it really matches batch, or leaving the code alone and dropping the parity claim from the comment? i lean toward cutting it — glm47's parser drops post-call prose deliberately (*"Only surface normal text that precedes the first parsed call ... matches vLLM's glm47_moe_tool_parser"*), so keeping it in streaming is fallout rather than a decision. either way i just dont want the comment promising something the code isnt doing, since thats what the next person will trust.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -1735,4 +1788,119 @@ mod tests {
         );
         assert!(json.get("reasoning_content").is_some());
     }
+
+    // --- glm47 truncation recovery tests ---
```
keivenchang, line 1867: these are good and they pin the original bug nicely — but all three go through `DeltaAggregator`, so theyre all batch. nothing here covers the streaming path, and thats where everything new in this revision lives: the three passes, the `recovered` latch, the strip.

could you add a couple over `apply_tool_calling_jail`? the two that would've caught real things:

- separate terminal chunk (content chunks, then a data-less chunk with `finish_reason=length`) — the jail emits two finish=length chunks in that shape, so this is the one that proves the latch
- prose + truncated block in the same jailed buffer — the case where `tail_already_emitted` is true and pass 2 has to strip

you can drive it straight through `apply_tool_calling_jail` with a `stream::iter` of chunks and collect the output, no engine needed.

keivenchang (commented):
thanks Pei, this addressed most of it — the latch is doing real work, i checked on 5.1.3 and the realistic shape (backend sends a separate terminal chunk) actually produces TWO finish=length chunks, so without `recovered` you'd have double-emitted there. good catch putting the TODO at `V2_FAMILIES` too, thats where someone will actually see it.

two things left, inline. also one correction thats my fault: my earlier comment said adding `"glm47"` to `V2_FAMILIES` disables streaming recovery, and your TODO picked that up — it also needs `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` on, plus tool_choice auto/unset. worth a word in the TODO so nobody reads it as "one line breaks it".

lib/llm/src/preprocessor.rs
keivenchang: ok looks like this is handled — `emitted_text` plus the `recovered` latch. i re-checked on 5.1.3 and the realistic shape (backend sends a data-less terminal chunk) actually produces TWO finish=length chunks, so the latch isnt belt-and-braces, its doing real work there.

lib/llm/src/preprocessor.rs
keivenchang: gotcha, all of them nulled including `id`. thanks for taking that one.

lib/llm/src/preprocessor.rs
keivenchang: I see you kept the `is_char_boundary` walk-back — that was the bit i was actually worried about, since `drain` would've panicked on ordinary CJK output. thanks.

lib/llm/src/preprocessor.rs
keivenchang: yep, `chain` does it. no vec on the common path now.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
keivenchang: ok, both sites read from `Glm47ParserConfig::default()` now. one leftover though — the warn branch a few lines down still has a literal `"<tool_call>"` in `choice.text.contains(...)`, worth pulling off the config too while youre in there.

lib/llm/src/preprocessor.rs
keivenchang: I see you put one at `V2_FAMILIES` as well — thats better than what i asked for, since thats the line someone actually edits. one tweak: the wording should mention `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2`, adding the family alone doesnt flip it. my fault, my original comment said it wrong.

keivenchang:
Just 2 more to address, inlined. By the way, can you reply "done" or "addressed" in the future, and then mark "Resolve conversation"? Thanks!

keivenchang:
@coderabbitai Resolve conversation, if you confirmed that it's addressed

lib/llm/src/preprocessor.rs
peilii: Fixed in the latest push — pass 2 now sets `rc.delta.content = None` entirely instead of truncating at the marker. That drops both the prose and the tail from the finish chunk, letting the recovery chunk carry just the marker-onwards tail. The comment is updated to match.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Added two streaming tests through `apply_tool_calling_jail` in `test_streaming_tool_parsers.rs`: one for the data-less terminal chunk shape (proves the latch fires and doesn't re-fire), and one for prose + truncated block in the same jailed buffer (proves `tail_already_emitted` triggers pass 2 suppression and the recovery chunk appears exactly once with no duplicate). One note on the prose test: the jail actually splits the content into two chunks (`"tail prose "` on one, the truncated block on the finish chunk), so the assertion checks that `<tool_call>get_time` appears exactly once rather than asserting the prose is absent — the prose from the earlier chunk can't be suppressed retroactively.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Fixed — `glm47_cfg` is now constructed once before the tail extraction block and reused in the warn branch's `choice.text.contains(glm47_cfg.tool_call_start.as_str())`.

lib/llm/src/preprocessor.rs
peilii: Updated — the TODO now reads "when glm47 is added here AND `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` is set" so it's clear both conditions are needed.

lib/llm/src/preprocessor.rs
peilii: Good catch — added two tests in the latest commit: one with CJK (3-byte chars) and one with emoji (4-byte chars), each chosen so the initial `keep_from` lands mid-char before the walk-back corrects it.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Addressed in the latest push — pass 2 now sets `rc.delta.content = None` on the finish chunk when the jail released the tail verbatim, dropping both the post-call prose and the tail. The recovery chunk then carries only the marker-onwards tail, matching batch.

lib/llm/src/preprocessor.rs
peilii: Done.

lib/llm/src/preprocessor.rs
peilii: Done, nulled all of them including `id`.

lib/llm/src/preprocessor.rs
peilii: Done — `emitted_text`, `recovered` latch, and the pass-2 prose suppression are all in. Good to know the latch is doing real work on the two-chunk shape.

keivenchang (approved):
approved — thanks Pei, all of it addressed and the streaming tests cover both shapes. nice work.

lib/llm/src/preprocessor.rs
```diff
@@ -2984,7 +3042,127 @@ impl OpenAIPreprocessor {
...
+                let mut cr = choice_recovery.lock().expect("choice recovery poisoned");
+                nv_chunk
+                    .data
+                    .iter()
+                    .flat_map(|data| data.inner.choices.iter())
+                    .filter_map(|choice| {
+                        let state = cr.entry(choice.index).or_default();
+                        if let Some(ChatCompletionMessageContent::Text(t)) = &choice.delta.content {
+                            state.emitted_text.push_str(t);
+                        }
```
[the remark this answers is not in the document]
peilii: Fixed — applied the same rfind/char-boundary-walkback/drain pattern as `input_text`. The append and drain are also skipped once `state.recovered` is true since no further detection is needed after recovery fires.

lib/llm/src/preprocessor.rs
```diff
@@ -2938,13 +2938,71 @@ impl OpenAIPreprocessor {
...
+        // Per-choice recovery state — allocated only for glm47 since only that
+        // parser emits <tool_call> XML that can be truncated at max_tokens.
+        // Buffers raw input and tracks what the jail emitted per choice.index
+        // so n > 1 is handled correctly and double-emit is avoided.
+        #[derive(Default)]
+        struct ChoiceRecovery {
+            input_text: String,
+            emitted_text: String,
+            recovered: bool,
+        }
```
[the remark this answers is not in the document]
peilii: Fixed — applied the same rfind/char-boundary-walkback/drain pattern as `input_text`. The append and drain are also skipped once `state.recovered` is true since no further detection is needed after recovery fires.

lib/llm/src/preprocessor.rs
peilii: Fixed in 380ed26acd (now in this branch via main merge). The `tail_already_emitted` path is covered by the new streaming test `test_glm47_streaming_prose_plus_truncated_block_tail_already_emitted`: prose + truncated block in the same jailed buffer — the jail splits the output, Pass 2 suppresses the finish chunk's content, and the recovery chunk carries the tail exactly once.

lib/llm/src/preprocessor.rs
peilii: Fixed — `rec.event = None`, `rec.comment = None`, `rec.error = None` are cleared on lines 3475–3478 and `rc.logprobs = None` per choice at line 3487 of the current code.

lib/llm/src/preprocessor.rs
peilii: Fixed — `emitted_text` is bounded the same way `input_text` is: after each append we drain everything before the last `<tool_call>` marker (lines 3389–3400), with a char-boundary walk-back to prevent panic on multi-byte chars.

lib/llm/src/preprocessor.rs
peilii: Done — `Vec::with_capacity(recoveries.len() + 1)` was replaced with `recovery_chunks: Vec<_> = recoveries.into_iter().filter_map(…).collect()` chained with `iter::once(nv_chunk)`. In the common (no-recovery) path the filter_map yields nothing and the Vec is empty. Ownership constraints prevent a fully allocation-free chain without an extra clone.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Fixed — both sites now read from `Glm47ParserConfig::default()` (lines 395 and 3270 respectively) so the strings stay in sync. The warn branch literal in aggregator.rs was also moved to the config in 380ed26acd.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Acknowledged — this is an inherent behavioral difference: in streaming the jail already emitted any prose before the marker as earlier content, so the client sees it. The batch rfind-based tail starts at the marker and excludes that prose. Both paths match on the marker-onwards content itself. Noted in a comment.

lib/llm/src/preprocessor.rs
peilii: Fixed — the TODO is present at lines 3362–3368, now clarified (in 380ed26acd) to note that adding glm47 to `V2_FAMILIES` alone is insufficient; `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` must also be set.

lib/llm/src/preprocessor.rs
peilii: Fixed in 380ed26acd — `truncate(marker_pos)` was replaced with `rc.delta.content = None` (line 3455). The comment now correctly says this suppresses the entire finish chunk content, dropping post-call prose and matching batch behavior (glm47's parser drops post-call prose deliberately).

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
peilii: Fixed — Pass 2 now sets `rc.delta.content = None` unconditionally when `tail_already_emitted`, matching batch which uses rfind and produces marker-only content.

peilii: Added in 380ed26acd: `test_glm47_streaming_truncated_data_less_terminal_chunk` (data-less terminal chunk, recovered latch) and `test_glm47_streaming_prose_plus_truncated_block_tail_already_emitted` (prose + truncated block in same buffer, no double-emit). All 4 streaming glm47 tests pass.

PR #12337, opened 2026-07-29
fix(mocker): compare optional G1 backends correctly

rmccorm4:
## Summary

- Compare the optional G1 backend against `Some(G1Backend::Native)` when selecting native flat sequence storage.
- Apply the same type-correct comparison to the feature-gated KVBM swap-in assertion.

## Root cause

The native G1 sequence optimization in #12248 compared `Option<G1Backend>` fields against bare `G1Backend` values. The first mismatch fails the `lib/bindings/kvbm` Clippy matrix with `E0308`; the second would fail the subsequent feature-gated KVBM check.

## Validation

- `(cd lib/bindings/kvbm && cargo clippy --locked --no-deps --all-targets -- -D warnings)`
- `cargo clippy --locked --no-deps -p dynamo-bench --test mooncake_trace --features mocker-kvbm-offload -- -D warnings`
- `cargo test --locked -p dynamo-mocker flat_ -- --nocapture` — 10 passed

## Related Issues

- Closes [DIS-2542](https://linear.app/nvidia/issue/DIS-2542/fix-g1-backend-type-mismatches-in-mocker)

### DIFF
--- a/lib/mocker/src/scheduler/vllm/core.rs
+++ b/lib/mocker/src/scheduler/vllm/core.rs
@@ -888,7 +888,7 @@ impl VllmCore {
         // unchanged here. The waiting-admission policy owns terminal rejection
         // because that path can emit the lifecycle signal.
         let output_capacity_hint = self.output_capacity_hint(prompt_len, max_output_tokens);
-        let sequence = if self.args.g1_backend == G1Backend::Native {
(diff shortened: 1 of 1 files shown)


rmccorm4:
dupe of https://github.com/ai-dynamo/dynamo/pull/12336

PR #12341, opened 2026-07-29
perf(mocker): optimize offline replay hot paths

jthomson04:
## Overview

This PR reduces offline-replay coordinator, cache-index, and completion-event
overhead without changing public APIs, configuration, replay formats, admission,
scaling, or KV-event formats.

#12329 has merged; this branch is based on `main`.

## Summary

- maintain aggregate offline worker in-flight ownership so coordinator checks are
  O(1), with debug recomputation only at mutation boundaries
- replace SGLang's hash-set-wide oldest-leaf scan with an ordered
  `(last_access_time, NodeId)` evictable-leaf index
- skip ordered-index searches for SGLang nodes that cannot be unlocked leaves
- execute and lower singleton offline worker groups without attention-DP staging
  while preserving token-time alignment
- store native vLLM's primary cached-copy ID directly, with duplicate copies in a
  rare overflow queue
- batch each positive-duration attention-DP epoch into one ordered completion
  heap event while preserving rank-level processing

LRU behavior is preserved except that equal SGLang eviction timestamps now use
`NodeId` as a deterministic tie-breaker instead of hash-set iteration order.

## Details

- `EngineComponent::in_flight()` returns the tracked aggregate directly; mutation
  wrappers retain debug validation against the full worker sum.
- SGLang leaf insertion, removal, touch, split, promotion, and partial eviction
  update one ordered evictable-leaf index. `split_node` keeps the original child
  as the suffix node, including its timestamp/index identity.
- Singleton execution avoids grouped staging but shares the grouped effects and
  refresh/defer tail.
- Scheduled completion batches allocate lazily on the first payload using a
  caller-provided capacity hint. Sequence allocation still advances by payload
  count so later events retain their pre-batching sequence identities.
- Disaggregated immediate completions use the same worker-accounting, FPM,
  lifecycle, and deferred-action path as scheduled completions.
- Native vLLM lookup keeps the common primary copy inline and uses the duplicate
  queue only when a hash has multiple physical copies.

Start with:

1. `lib/mocker/src/replay/offline/components/engine.rs` for aggregate ownership
   and singleton/grouped execution.
2. `lib/mocker/src/cache/radix_cache.rs` for the ordered SGLang eviction index.
3. `lib/mocker/src/cache/vllm_block_pool.rs` for the compact native vLLM copy
   index.
4. `lib/mocker/src/replay/offline/events.rs` and
   `lib/mocker/src/replay/offline/runtime_utils.rs` for attention-DP completion
   batching.

## Related Issues

No related issue.

## Exact-head 128/128 Results

Frozen ordinary-release binaries were built with
`RUSTFLAGS="-C target-cpu=x86-64-v3"` from:

- P0 `97f38efdfc`: rebased state before the later singleton, cache-index, review,
  and attention-DP commits
- pre-fix `61b7fb4034`: prior PR head before the latest SGLang review fix
- final `f4c80b5986`: current PR head including all review fixes

AgentX 128 sessions; 128 workers; KV routing/events; native G1; 16-token engine
blocks/pages; 64-token trace blocks; engine speedup 20; seed 42; CPU 0. Runs were
sequential in final-vLLM, final-SGLang, pre-fix-SGLang, P0-vLLM, P0-SGLang
order, with no warm-ups. These are one-point measurements, not statistical
estimates.

| Backend | P0 processed tok/s | Final processed tok/s | Gain | P0 wall ms | Final wall ms | Wall reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| vLLM | 128,714,377 | 134,904,746 | +4.81% | 36,453.293 | 34,780.562 | 4.59% |
| SGLang | 161,273,456 | 176,451,658 | +9.41% | 29,093.833 | 26,591.209 | 8.60% |

Normalized reports are exactly equal after removing only `wall_time_ms`,
`processed_tokens_per_s`, and `processed_output_tokens_per_s`. Every run
completed all 38,114 requests and processed 4,692,062,939 tokens. Prefix-cache
reuse was identical within each backend: 97.369893% for vLLM and 97.354061% for
SGLang. CPU-0 preflight was clean for every cell (98.02-100% idle).

The isolated SGLang review fix improved processed-token throughput from
176,236,918 to 176,451,658 tok/s (+0.12%) and reduced wall time from
26,623.610 to 26,591.209 ms (0.12%). The normalized reports were exactly equal.

The two optimizations already present in P0 retain their earlier isolated
SGLang measurements: +4.98% for aggregate in-flight accounting and +2.82% for
ordered leaves. Those historical one-point deltas are not compounded with the
P0-to-final results above.

The untracked harness preserves commit, binary/config/trace hashes, build IDs,
CPU preflight, environment capture, run order, and raw reports.

## Attention-DP Speedup

Exact `99e975ee98` baseline versus the attention-DP completion batching commit,
using ordinary release builds and the AgentX setup with 32 logical workers,
DP4, 128 ranks and sessions, native G1, KV routing/events, block size 16,
trace block size 64, speedup 20, seed 42, and CPU 0.

One matched pair was collected per backend with no warm-ups:
vLLM baseline then candidate, followed by SGLang candidate then baseline.
Every CPU-0 preflight reported 100% idle.

| Backend | Baseline processed tok/s | Candidate processed tok/s | Gain |
| --- | ---: | ---: | ---: |
| vLLM | 97,292,732 | 115,701,733 | +18.92% |
| SGLang | 112,896,058 | 144,153,017 | +27.69% |

Normalized reports are byte-identical after removing wall-clock time and its
derived throughput fields. Each run completed 38,114 requests and processed
4,692,062,939 tokens.

One no-warm-up candidate profile per backend recorded zero lost samples.
Against the existing pre-change DP4 structural profiles, completion push/pop
helper samples fell 82-96%, heap push samples fell 72-78%, and aggregate heap
pop samples fell 60-70%. `on_scheduled_completion`, token-time alignment, and
KV-event capture remain visible as expected.

### Full correctness parity

**PASS.** The full correctness-parity skill compared the exact PR merge-base
`677f92cc38` with head `f4c80b5986` using identical temporary lifecycle
instrumentation and two independent processes per revision across SGLang and
vLLM aggregate/disaggregated rows (16 canonical reports total).

Every report completed all 5,000 requests. Results were deterministic within
each revision and byte-identical across revisions after excluding only wall
time and derived throughput. Coverage included KV-overlap-sensitive requests,
immediate and queued admission, 5,000 handoffs in each disaggregated row,
vLLM preemptions, and balanced G1→G2/G2→G1 reservation/completion lifecycles.
The Mooncake corpus SHA-256 was
`3892ae19ae480b643155f0c6b9d798591cbe2e73bec6a0fa5ae3d3bc0332fb8a`.

- `cargo test -p dynamo-mocker cache::radix_cache` (15 passed)
- `cargo test -p dynamo-mocker cache::vllm_block_pool` (12 passed)
- `cargo test -p dynamo-mocker replay::offline` (180 passed)
- `cargo test -p dynamo-mocker` (693 passed)
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline`
  (183 passed)
- `git diff --check`
- exact normalized 128/128 report parity for both backends
- every commit retains a DCO signoff

## Summary by CodeRabbit

* **New Features**
  * Added batched worker-completion events for offline replay.
  * Preserved completion ordering and metadata across multi-worker simulations.
  * Improved cache eviction behavior with consistent least-recently-used ordering.
  * Improved handling of duplicate physical cache copies during lookup, reservation, and eviction.

* **Bug Fixes**
  * Fixed cache index consistency after partial eviction, node promotion, and unlock operations.

* **Documentation**
  * Updated offline replay documentation for batched completion events and current module responsibilities.

### DIFF
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -278,7 +278,7 @@ impl RadixCache {
             .filter(|(id, node)| *id != self.root && node.lock_ref == 0 && node.children.is_empty())
             .map(|(id, node)| (node.last_access_time, id))
             .collect::<BTreeSet<_>>();
-        debug_assert_eq!(
+        assert_eq!(
             self.evictable_leaves, expected,
             "SGLang evictable-leaf index drifted from radix nodes"
         );
@@ -642,6 +642,8 @@ impl RadixCache {
     }
 
     fn split_node(&mut self, child_id: NodeId, split_pos: usize) -> NodeId {
+        // Keep child_id as the suffix node so its last-access timestamp and
+        // evictable-index key remain valid across the split.
         let (child_parent, original_ck, prefix_key, prefix_value, suffix_ck, lock_ref, accessed) = {
             let child = &mut self.nodes[child_id];
             let child_parent = child.parent;
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -242,19 +242,21 @@ impl RadixCache {
         (self.nodes[id].last_access_time, id)
     }
 
+    fn is_evictable_leaf(&self, id: NodeId) -> bool {
+        id != self.root && self.nodes[id].lock_ref == 0 && self.is_leaf(id)
+    }
+
     fn insert_evictable_leaf(&mut self, id: NodeId) {
-        debug_assert_ne!(id, self.root, "root must not enter the eviction index");
-        debug_assert!(self.is_leaf(id), "only leaves are directly evictable");
-        debug_assert_eq!(
-            self.nodes[id].lock_ref, 0,
-            "locked leaf must not enter the eviction index"
+        debug_assert!(
+            self.is_evictable_leaf(id),
+            "only unlocked non-root leaves are directly evictable"
         );
         let inserted = self.evictable_leaves.insert(self.evictable_key(id));
         debug_assert!(inserted, "evictable leaf was already indexed");
     }
 
     fn remove_evictable_leaf(&mut self, id: NodeId) -> bool {
-        if id == self.root {
+        if !self.is_evictable_leaf(id) {
             return false;
         }
         self.evictable_leaves.remove(&self.evictable_key(id))
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -6,8 +6,9 @@
 //! Reference: sglang/python/sglang/srt/mem_cache/radix_cache.py
 
 use dynamo_kv_router::protocols::{BlockHashOptions, LocalBlockHash, compute_block_hash_for_seq};
-use rustc_hash::{FxHashMap, FxHashSet};
+use rustc_hash::FxHashMap;
 use slotmap::{SlotMap, new_key_type};
+use std::collections::BTreeSet;
 use std::time::Instant;
 
 new_key_type! {
@@ -192,8 +193,10 @@ pub struct RadixCache {
     root: NodeId,
     pub page_pool: PagePool,
     page_size: usize,
+    /// Unlocked leaves ordered by their last access time for O(log n) updates
+    /// and O(1) oldest-victim lookup.
+    evictable_leaves: BTreeSet<(Instant, NodeId)>,
     /// Total token count in evictable nodes.
-    pub evictable_leaves: FxHashSet<NodeId>,
     pub evictable_size: usize,
     /// Total token count in protected (locked) nodes.
     pub protected_size: usize,
@@ -216,7 +219,7 @@ impl RadixCache {
             root,
             page_pool: PagePool::new(total_tokens, page_size),
             page_size,
-            evictable_leaves: FxHashSet::default(),
+            evictable_leaves: BTreeSet::new(),
             evictable_size: 0,
             protected_size: 0,
         }
@@ -233,6 +236,55 @@ impl RadixCache {
     }
     pub fn num_nodes(&self) -> usize {
         self.nodes.len()
+    }
+
+    fn evictable_key(&self, id: NodeId) -> (Instant, NodeId) {
+        (self.nodes[id].last_access_time, id)
+    }
+
+    fn insert_evictable_leaf(&mut self, id: NodeId) {
+        debug_assert_ne!(id, self.root, "root must not enter the eviction index");
+        debug_assert!(self.is_leaf(id), "only leaves are directly evictable");
+        debug_assert_eq!(
+            self.nodes[id].lock_ref, 0,
+            "locked leaf must not enter the eviction index"
+        );
+        let inserted = self.evictable_leaves.insert(self.evictable_key(id));
+        debug_assert!(inserted, "evictable leaf was already indexed");
+    }
+
+    fn remove_evictable_leaf(&mut self, id: NodeId) -> bool {
+        if id == self.root {
+            return false;
+        }
+        self.evictable_leaves.remove(&self.evictable_key(id))
+    }
+
+    fn touch_node(&mut self, id: NodeId, now: Instant) {
+        let was_indexed = self.remove_evictable_leaf(id);
+        self.nodes[id].last_access_time = now;
+        if was_indexed {
+            self.insert_evictable_leaf(id);
+        }
+    }
+
+    #[cfg(debug_assertions)]
+    fn debug_assert_evictable_index(&self) {
+        let expected = self
+            .nodes
+            .iter()
+            .filter(|(id, node)| *id != self.root && node.lock_ref == 0 && node.children.is_empty())
+            .map(|(id, node)| (node.last_access_time, id))
+            .collect::<BTreeSet<_>>();
+        debug_assert_eq!(
+            self.evictable_leaves, expected,
+            "SGLang evictable-leaf index drifted from radix nodes"
+        );
+    }
+
+    #[cfg(test)]
+    fn evictable_leaf_is_indexed(&self, id: NodeId) -> bool {
+        self.evictable_leaves.contains(&self.evictable_key(id))
     }
 
     pub(crate) fn page_hashes(&self, tokens: &[u32]) -> Vec<LocalBlockHash> {
@@ -306,9 +358,11 @@ impl RadixCache {
 
             matched_pages += common_len;
             current = child_id;
-            self.nodes[current].last_access_time = now;
+            self.touch_node(current, now);
         }
 
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
         (matched_pages * self.page_size, current)
     }
 
@@ -359,7 +413,10 @@ impl RadixCache {
             value.len()
         );
         let page_ids = self.page_ids(&value[..aligned_len], aligned_len / self.page_size);
-        self.insert_page_suffix(self.root, 0, key, &page_ids, false)
+        let inserted = self.insert_page_suffix(self.root, 0, key, &page_ids, false);
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
+        inserted
     }
 
     /// Insert only the suffix after a retained, page-aligned prefix.
@@ -394,7 +451,10 @@ impl RadixCache {
             &value[prefix_len..aligned_len],
             (aligned_len - prefix_len) / self.page_size,
         );
-        self.insert_page_suffix(prefix_node, prefix_len, key, &page_ids, true)
+        let inserted = self.insert_page_suffix(prefix_node, prefix_len, key, &page_ids, true);
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
+        inserted
     }
 
     /// Insert page identities already materialized by the request lease.
@@ -422,12 +482,15 @@ impl RadixCache {
             "prefix pages {prefix_pages} exceed hashed pages {}",
             page_keys.len()
         );
-        self.insert_page_hash_suffix(
+        let inserted = self.insert_page_hash_suffix(
             prefix_node,
             &page_keys[prefix_pages..],
             &pages[prefix_pages..page_keys.len()],
             true,
-        )
+        );
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
+        inserted
     }
 
     #[cfg(test)]
@@ -525,7 +588,7 @@ impl RadixCache {
             if common_len == child_len {
                 key_offset += common_len;
                 current = child_id;
-                self.nodes[current].last_access_time = now;
+                self.touch_node(current, now);
             } else {
                 if common_len > 0 {
                     let intermediate = self.split_node(child_id, common_len);
@@ -549,8 +612,9 @@ impl RadixCache {
     fn touch_path(&mut self, node_id: NodeId, now: Instant) {
         let mut current = Some(node_id);
         while let Some(id) = current {
-            self.nodes[id].last_access_time = now;
-            current = self.nodes[id].parent;
+            let parent = self.nodes[id].parent;
+            self.touch_node(id, now);
+            current = parent;
         }
     }
 
@@ -561,12 +625,12 @@ impl RadixCache {
         value: &[KvPageId],
         now: Instant,
     ) -> NodeId {
+        self.touch_node(node_id, now);
         let node = &mut self.nodes[node_id];
         debug_assert!(node.children.is_empty());
         debug_assert!(node.lock_ref <= 1);
         node.key.extend_from_slice(key);
         node.value.extend_from_slice(value);
-        node.last_access_time = now;
         if node.lock_ref == 0 {
             self.evictable_size += key.len() * self.page_size;
         } else {
@@ -639,11 +703,11 @@ impl RadixCache {
         let ck = key[0];
         let new_id = self.nodes.insert(new_node);
 
-        self.evictable_leaves.remove(&parent_id);
+        self.remove_evictable_leaf(parent_id);
 
         self.nodes[parent_id].children.insert(ck, new_id);
 
-        self.evictable_leaves.insert(new_id);
+        self.insert_evictable_leaf(new_id);
         self.evictable_size += key.len() * self.page_size;
 
         new_id
@@ -659,16 +723,21 @@ impl RadixCache {
             if id == self.root {
                 break;
             }
+            let was_unlocked = self.nodes[id].lock_ref == 0;
+            if was_unlocked {
+                self.remove_evictable_leaf(id);
+            }
             let node = &mut self.nodes[id];
             let tokens = node.key.len() * self.page_size;
             node.lock_ref += 1;
-            if node.lock_ref == 1 {
-                self.evictable_leaves.remove(&id);
+            if was_unlocked {
                 self.evictable_size -= tokens;
                 self.protected_size += tokens;
             }
             current = self.nodes[id].parent;
         }
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
     }
 
     pub fn dec_lock_ref(&mut self, node_id: NodeId) {
@@ -688,11 +757,13 @@ impl RadixCache {
                 self.protected_size -= tokens;
                 self.evictable_size += tokens;
                 if self.is_leaf(id) {
-                    self.evictable_leaves.insert(id);
+                    self.insert_evictable_leaf(id);
                 }
             }
             current = self.nodes[id].parent;
         }
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
     }
 
     /// Evict tokens from the cache by LRU order, rounding partial leaves to full pages.
@@ -702,15 +773,13 @@ impl RadixCache {
         let mut evicted_indices =
             Vec::with_capacity(num_tokens.min(self.evictable_size).div_ceil(self.page_size));
         while evicted < num_tokens {
-            let victim = self
-                .evictable_leaves
-                .iter()
-                .min_by_key(|&&id| self.nodes[id].last_access_time)
-                .copied();
-
-            let Some(victim_id) = victim else {
+            let Some((last_access_time, victim_id)) = self.evictable_leaves.pop_first() else {
                 break;
             };
+            debug_assert_eq!(
+                last_access_time, self.nodes[victim_id].last_access_time,
+                "eviction index timestamp drifted from radix node"
+            );
 
             let victim_pages = self.nodes[victim_id].key.len();
             let victim_tokens = victim_pages * self.page_size;
@@ -732,6 +801,7 @@ impl RadixCache {
 
                 self.evictable_size -= eviction_len;
                 evicted += eviction_len;
+                self.insert_evictable_leaf(victim_id);
                 continue;
             }
 
@@ -742,7 +812,6 @@ impl RadixCache {
             let tokens = victim_node.key.len() * self.page_size;
             let parent_id = victim_node.parent;
 
-            self.evictable_leaves.remove(&victim_id);
             self.evictable_size -= tokens;
             evicted += tokens;
 
@@ -756,10 +825,12 @@ impl RadixCache {
                     && self.nodes[pid].children.is_empty()
                     && self.nodes[pid].lock_ref == 0
                 {
-                    self.evictable_leaves.insert(pid);
+                    self.insert_evictable_leaf(pid);
                 }
             }
         }
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
         (evicted, evicted_indices)
     }
 
@@ -978,7 +1049,7 @@ mod tests {
         assert_eq!(cache.protected_size, 7); // 2+2+3
 
         cache.dec_lock_ref(node_a);
-        assert!(cache.evictable_leaves.contains(&node_a));
+        assert!(cache.evictable_leaf_is_indexed(node_a));
         cache.dec_lock_ref(node_b);
         assert_eq!(cache.protected_size, 0);
     }
@@ -1031,6 +1102,64 @@ mod tests {
             "should evict unlocked [4,5,6] indices"
         );
         assert_eq!(cache.match_prefix(&[1, 2, 3]).0, 3);
+    }
+
+    #[test]
+    fn touching_an_evictable_leaf_updates_order() {
+        let mut cache = RadixCache::new(100, 1);
+        cache.insert(&[1, 2, 3], &[0, 1, 2]);
+        std::thread::sleep(std::time::Duration::from_millis(1));
+        cache.insert(&[4, 5, 6], &[3, 4, 5]);
+        std::thread::sleep(std::time::Duration::from_millis(1));
+        assert_eq!(cache.match_prefix(&[1, 2, 3]).0, 3);
+
+        assert_eq!(cache.evict(3).0, 3);
+        assert_eq!(cache.match_prefix(&[1, 2, 3]).0, 3);
+        assert_eq!(cache.match_prefix(&[4, 5, 6]).0, 0);
+    }
+
+    #[test]
+    fn unlocking_preserves_last_access_order() {
+        let mut cache = RadixCache::new(100, 1);
+        cache.insert(&[1, 2, 3], &[0, 1, 2]);
+        let (_, older) = cache.match_prefix(&[1, 2, 3]);
+        cache.inc_lock_ref(older);
+
+        std::thread::sleep(std::time::Duration::from_millis(1));
+        cache.insert(&[4, 5, 6], &[3, 4, 5]);
+        let (_, newer) = cache.match_prefix(&[4, 5, 6]);
+        cache.inc_lock_ref(newer);
+        cache.dec_lock_ref(newer);
+        cache.dec_lock_ref(older);
+
+        assert_eq!(cache.evict(3).0, 3);
+        assert_eq!(cache.match_prefix(&[1, 2, 3]).0, 0);
+        assert_eq!(cache.match_prefix(&[4, 5, 6]).0, 3);
+    }
+
+    #[test]
+    fn partial_eviction_and_parent_promotion_keep_index_consistent() {
+        let mut cache = RadixCache::new(100, 1);
+        cache.insert(&[1, 2, 3, 4], &[0, 1, 2, 3]);
+        let (_, leaf) = cache.match_prefix(&[1, 2, 3, 4]);
+
+        assert_eq!(cache.evict(2).0, 2);
+        assert_eq!(cache.node(leaf).key.len(), 2);
+        assert!(cache.evictable_leaf_is_indexed(leaf));
+
+        let mut branched = RadixCache::new(100, 1);
+        branched.insert(&[10, 11, 12], &[0, 1, 2]);
+        branched.insert(&[10, 11, 13], &[0, 1, 3]);
+        let parent = *branched.nodes[branched.root]
+            .children
+            .values()
+            .next()
+            .unwrap();
+        assert!(!branched.is_leaf(parent));
+
+        assert_eq!(branched.evict(2).0, 2);
+        assert!(branched.is_leaf(parent));
+        assert!(branched.evictable_leaf_is_indexed(parent));
     }
 
     #[test]
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -193,6 +193,8 @@ pub struct RadixCache {
     root: NodeId,
     pub page_pool: PagePool,
     page_size: usize,
+    #[cfg(test)]
+    test_now: Option<Instant>,
     /// Unlocked leaves ordered by their last access time for O(log n) updates
     /// and O(1) oldest-victim lookup.
     evictable_leaves: BTreeSet<(Instant, NodeId)>,
@@ -219,6 +221,8 @@ impl RadixCache {
             root,
             page_pool: PagePool::new(total_tokens, page_size),
             page_size,
+            #[cfg(test)]
+            test_now: None,
             evictable_leaves: BTreeSet::new(),
             evictable_size: 0,
             protected_size: 0,
@@ -236,6 +240,19 @@ impl RadixCache {
     }
     pub fn num_nodes(&self) -> usize {
         self.nodes.len()
+    }
+
+    fn now(&self) -> Instant {
+        #[cfg(test)]
+        if let Some(now) = self.test_now {
+            return now;
+        }
+        Instant::now()
+    }
+
+    #[cfg(test)]
+    fn set_test_now(&mut self, now: Instant) {
+        self.test_now = Some(now);
     }
 
     fn evictable_key(&self, id: NodeId) -> (Instant, NodeId) {
@@ -325,11 +342,31 @@ impl RadixCache {
 
     /// Match a prefix using page hashes already owned by the request.
     pub(crate) fn match_prefix_hashes(&mut self, page_keys: &[LocalBlockHash]) -> (usize, NodeId) {
-        let now = Instant::now();
+        self.match_prefix_hashes_impl(page_keys, false)
+    }
+
+    /// Match and protect a prefix using page hashes already owned by the request.
+    pub(crate) fn match_prefix_hashes_and_lock(
+        &mut self,
+        page_keys: &[LocalBlockHash],
+    ) -> (usize, NodeId) {
+        self.match_prefix_hashes_impl(page_keys, true)
+    }
+
+    fn match_prefix_hashes_impl(
+        &mut self,
+        page_keys: &[LocalBlockHash],
+        lock_match: bool,
+    ) -> (usize, NodeId) {
+        let now = self.now();
         self.nodes[self.root].last_access_time = now;
 
         let mut current = self.root;
         let mut matched_pages: usize = 0;
+        // Delay the most recently matched node's touch until we know whether it
+        // is the terminal match. The match-and-lock path can then remove an
+        // unlocked terminal leaf once without reinserting it before locking.
+        let mut pending_touch = None;
 
         while matched_pages < page_keys.len() {
             let child_id = match self.nodes[current]
@@ -340,6 +377,10 @@ impl RadixCache {
                 Some(id) => id,
                 None => break,
             };
+
+            if let Some(id) = pending_touch.take() {
+                self.touch_node(id, now);
+            }
 
             let (common_len, child_len) = {
                 let child_key = &self.nodes[child_id].key;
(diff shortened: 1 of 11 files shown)


lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -537,6 +685,48 @@ where
...
+                Self::lower_executed_pass(
+                    &mut self.workers,
+                    self.stage,
+                    rank_id,
+                    PassBoundary {
+                        start_ms: now_ms,
+                        end_ms: group_end_ms,
+                    },
+                    executed,
+                    None,
```
jthomson04: **The fast path silently drops `align_pass_token_times`.**

The general path calls `collector.align_pass_token_times(&output_signals, boundary.end_ms)` for every visible pass (line 583). Passing `None` here means it never runs for single-rank groups.

Today this is a no-op in default builds, because every core records tokens at exactly the `end_ms` it returns:

- `sglang/core.rs:688` `on_token(…, decode.end_ms)` vs `sglang/core.rs:768` `end_ms: decode.end_ms`
- `vllm/core.rs:2199,2438` `decode_end_ms = decode_start_ms + decode_time` vs `vllm/core.rs:1620` `end_ms = decode_start_ms + decode_time * 1000`

The exception is the `kvbm-offload` stall-advance at `vllm/core.rs:1628-1634`, which bumps `end_ms` to an offload deadline *after* `on_token` has already fired. Pre-PR, a single-rank visible pass in that state had its token timestamps rewritten to the advanced deadline; now it keeps the pre-advance value — a TTFT/ITL difference in recorded traces. `kvbm-offload` is non-default, so CI won't catch it.

Either pass the collector through here, or add a comment stating the cross-module invariant the omission depends on ("cores record token times at their returned `end_ms`, so alignment is identity when the group is a single rank") and call out the offload exception. Right now that invariant lives in two other modules with nothing pointing at it.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -499,6 +566,87 @@ where
         Ok(worker.is_busy())
     }
 
+    #[inline(always)]
```
jthomson04: This is the **only `#[inline(always)]` in the entire crate**, on an ~80-line function called from two sites. It also defeats the crate's own profiling convention: there are 9 `#[cfg_attr(feature = "profile", inline(never))]` markers on hot functions precisely so they stay resolvable in profiles, and this one will be invisible.

Suggest dropping it or downgrading to plain `#[inline]` — the function is generic over `Observation` so it's already inlinable at both call sites, and LLVM has the size heuristics.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -537,6 +685,48 @@ where
...
+                        let Some(collector) = collector.as_deref_mut() else {
+                            bail!("offline replay visible engine pass requires a collector");
+                        };
+                        Self::required_worker_mut(&mut self.workers, rank_id)
+                            .execute_pass(collector, now_ms)
+                    }
+                    EnginePassMode::Hidden => Self::required_worker_mut(&mut self.workers, rank_id)
+                        .execute_hidden_pass(now_ms),
+                };
+                let group_end_ms = executed.end_ms;
```
jthomson04: Minor: the general path computes `group_end_ms` as `fold(now_ms, f64::max)`, i.e. `max(now_ms, end_ms)`, while this uses `executed.end_ms` raw.

I checked all three consumers — `wall_time_secs()` clamps at `0.0`, the `boundary.end_ms > boundary.start_ms` gate, and the `group_end_ms <= now_ms` re-arm — and behavior is identical either way. Still worth matching the general path so the two branches read as the same computation.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -475,10 +537,15 @@ where
...
     ) -> anyhow::Result<ObservedCommandEffects<Observation::Batch>> {
-        let mut effects = self
-            .worker_mut(rank_id)
-            .ok_or_else(|| anyhow::anyhow!("offline replay selected unknown rank {rank_id}"))?
-            .apply_command(command)?;
+        let (mut effects, before, after) = {
+            let worker = self
+                .worker_mut(rank_id)
+                .ok_or_else(|| anyhow::anyhow!("offline replay selected unknown rank {rank_id}"))?;
+            let before = worker.in_flight();
```
jthomson04: `total_in_flight` is only correct because `receive_request` / `apply_command` / `mark_completed` are all funneled through tracked wrappers like this one. But `worker_mut` (line 177) still hands out a raw `&mut OfflineWorkerState`, and `test_mark_worker_ready_activates_pending` (line ~1355) already uses it to call `receive_request` directly, desyncing the aggregate. That test happens not to call `in_flight()` afterward, so `debug_assert_total_in_flight` never fires.

Consider renaming to `worker_mut_untracked` to make the hazard visible at the call site, or routing that test through `engine.dispatch`.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -306,9 +360,11 @@ impl RadixCache {
 
             matched_pages += common_len;
             current = child_id;
-            self.nodes[current].last_access_time = now;
+            self.touch_node(current, now);
         }
 
+        #[cfg(debug_assertions)]
+        self.debug_assert_evictable_index();
```
jthomson04, line 419: `debug_assert_evictable_index` allocates a fresh `BTreeSet` over **all** radix nodes, and it's called on every `match_prefix`, insert, lock/unlock, and evict (lines 367, 420, 458, 494, 742, 768, 835).

`debug_assert_ready_frontier` sets the precedent for full-scan invariant checks, but its N is the worker count (single digits) while this one is the node count (tens of thousands). Test suites are unaffected — the mocker suite still runs in 1.84s — but a dev-profile replay becomes O(nodes) per admission, which is a footgun for anyone profiling without `--release`.

Consider a dedicated cargo feature, or hoisting the check to `#[cfg(test)]`.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -702,15 +775,13 @@ impl RadixCache {
...
             Vec::with_capacity(num_tokens.min(self.evictable_size).div_ceil(self.page_size));
         while evicted < num_tokens {
-            let victim = self
-                .evictable_leaves
-                .iter()
-                .min_by_key(|&&id| self.nodes[id].last_access_time)
-                .copied();
-
-            let Some(victim_id) = victim else {
+            let Some((last_access_time, victim_id)) = self.evictable_leaves.pop_first() else {
```
jthomson04, line 845: Behavior note for the PR description: this changes the **eviction tie-break order**. `min_by_key` over an `FxHashSet` picked an arbitrary node among equal `last_access_time` values; ordering on `(Instant, NodeId)` now breaks ties by slot key.

That's an improvement (deterministic), but the PR body claims scheduler policy is preserved, and this is a case where it isn't strictly. Low impact given `Instant` resolution, but worth a line in the summary.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -36,6 +36,62 @@ struct BlockCopy {
...
+    }
+
+    fn push(&mut self, id: BlockCopyId) {
+        self.duplicates
+            .get_or_insert_with(|| Box::new(VecDeque::new()))
+            .push_back(id);
+    }
+
+    /// Remove `id`, returning whether the hash no longer has any copies.
+    fn remove(&mut self, id: BlockCopyId) -> Option<bool> {
```
jthomson04: `Option<bool>` nests two meanings here: `None` = "invariant violated, caller panics", `Some(true)` = "hash is now empty", `Some(false)` = "copies remain". Both call sites end up reading awkwardly (`let Some(remove_hash) = copies.remove(id) else { panic!(…) }`).

A small enum would document itself:

```rust
enum Removed { Last, Remaining, Missing }
```

The logic itself is correct — I verified the `duplicates.is_some() => non-empty` invariant holds (`push` never leaves it empty, and both removal branches reset to `None` when it drains), so the `pop_front().expect(…)` is sound.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -299,8 +354,13 @@ impl VllmBlockPool {
             inactive_prev: None,
             inactive_next: None,
         };
-        self.by_hash.entry(hash).or_default().push(id);
-        became_visible
+        if let Some(copies) = self.by_hash.get_mut(&hash) {
```
jthomson04: Nit: `get_mut` followed by `insert` hashes twice on the miss path. `match self.by_hash.entry(hash)` gets it to one:

```rust
match self.by_hash.entry(hash) {
    Entry::Occupied(mut e) => { e.get_mut().push(id); false }
    Entry::Vacant(e) => { e.insert(HashCopies::new(id)); true }
}
```

Still a net win over the old `contains_key` + `entry().or_default()`, so this is purely incremental.

jthomson04 (commented):
## Review: offline replay hot-path optimizations

**Verified locally on `bffc991`:** `cargo clippy -p dynamo-mocker --all-targets` clean, `cargo test -p dynamo-mocker` — **688 passed / 0 failed**.

The refactoring is careful and the invariant-checking discipline (`debug_assert_evictable_index`, `debug_assert_total_in_flight`, `assert_hash_index_consistent`) matches the existing `debug_assert_ready_frontier` convention. I traced every index-mutation site and the accounting holds:

- `inc_lock_ref` removes from the index *before* incrementing (required — the reverse would strand entries).
- The partial-eviction path re-inserts the truncated victim after `pop_first`, and can't loop forever because `eviction_len >= remaining`.
- `create_child` removes the parent before wiring the child in, so `is_leaf` still reflects pre-insert state.
- `split_node` preserves both nodes' index membership (intermediate is never a leaf; the child keeps its timestamp and lock state).
- `HashCopies` preserves the `SmallVec`'s FIFO `first_copy` semantics on primary promotion.
- Nothing outside `radix_cache.rs` touches `last_access_time`, so there's no stale-index-key hazard.
- `immediate_completions` route through `on_scheduled_completion` in both `agg.rs` and `disagg.rs`, so `mark_completed` is always accounted.

One substantive finding (the fast path dropping `align_pass_token_times`) plus some quality nits, all inline.

### Test coverage

The new tests are well targeted — the three radix index-consistency tests and `evicting_non_primary_duplicate_preserves_primary_lookup` each cover a distinct hazard the refactor introduces, and Datadog reports 100% patch coverage. Two gaps:

1. **The Validation section only cites `cargo test -p dynamo-mocker replay::offline`**, but two of the three changed files live under `cache::` and are entirely outside that filter. `cache::` alone is 27 tests. Worth citing the full-suite run in the PR body instead.
2. **Nothing pins the single-rank fast path against the general path.** Existing single-worker visible-pass tests now route through the fast path and still pass, which is good evidence the `align_pass_token_times` skip is currently benign — but a test that runs the same workload through a 1-rank and a 2-rank group and compares token timelines would lock it down.

### Performance

The claimed asymptotics are right: eviction victim lookup goes O(L) → O(log L), `in_flight()` goes O(W) → O(1), and the `is_evictable_leaf` guard keeps `touch_path`'s ancestor walk O(1) per internal node (only the terminal leaf pays for the BTreeSet round-trip).

One trade worth measuring when you refresh the numbers: `touch_node` is now O(log L) with two BTreeSet operations where it used to be a single field write, and `match_prefix_hashes` pays that on every matched leaf. For a small cache under heavy prefix-match traffic with no eviction pressure this could be a small regression, so cover both regimes (eviction-heavy and eviction-free), not just the 128/128 shape.

### Security

Nothing of concern — simulator-internal, no I/O or untrusted input. The new `expect`/panic paths (`checked_add`/`checked_sub` on in-flight counts, `"duplicate-copy queue must not be empty"`) all guard genuine invariants and are preferable to silent wraparound.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `99e975ee98`. Singleton visible passes now pass the collector into `lower_executed_pass`, so token timestamps align to any KVBM stall-advanced boundary. I also added one-rank versus two-rank visible token-timeline regression coverage.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `99e975ee98`. Replaced `#[inline(always)]` with the crate convention `#[cfg_attr(feature = "profile", inline(never))]`, leaving normal LLVM inlining decisions unconstrained while retaining a resolvable profile symbol.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `99e975ee98`. The singleton branch now computes `group_end_ms` as `executed.end_ms.max(now_ms)`, matching the general grouped path.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `99e975ee98`. The raw mutable accessor was replaced with a closure-based `update_worker` helper that snapshots ownership before and after every mutation. Dispatch, commands, completions, and test-only mutations use it, and the pending-startup test now directly asserts aggregate `in_flight`.

lib/mocker/src/cache/radix_cache.rs
jthomson04: Fixed in `99e975ee98`. The full-node `debug_assert_evictable_index` scan and all of its call sites are now `#[cfg(test)]`; the cheap local debug assertions remain available in development builds.

lib/mocker/src/cache/radix_cache.rs
jthomson04: Updated the PR summary to call out the deterministic `NodeId` tie-break for equal timestamps explicitly, and clarified that LRU semantics are otherwise preserved.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Fixed in `99e975ee98`. `HashCopies::remove` now returns an explicit `CopyRemoval::{Last, Remaining, Missing}` enum, and eviction handles each state directly.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Fixed in `99e975ee98`. `cache_private` now uses `FxHashMap::entry`, avoiding the second hash on insertion misses.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -728,11 +868,9 @@ where
     }
 
     pub(in crate::replay::offline) fn in_flight(&self) -> usize {
-        self.workers
-            .iter()
-            .filter_map(Option::as_ref)
-            .map(OfflineWorkerState::in_flight)
-            .sum()
+        #[cfg(debug_assertions)]
+        self.debug_assert_total_in_flight();
```
jthomson04: `in_flight()` is a **read**, not a mutation boundary — but this makes it O(workers) in every debug build.

`record_in_flight_change` already asserts on every mutation, so this call adds no coverage. Meanwhile `cluster_in_flight()` → `in_flight()` runs from `is_workload_done` in the coordinator loop, so debug builds now pay *more* than they did before the change (the same O(n) sum, plus a comparison). The PR description says "debug recomputation at mutation boundaries", which is exactly right — dropping the assert here would make the code match it.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -235,6 +238,57 @@ impl RadixCache {
...
+
+    #[cfg(test)]
+    fn debug_assert_evictable_index(&self) {
+        let expected = self
+            .nodes
+            .iter()
+            .filter(|(id, node)| *id != self.root && node.lock_ref == 0 && node.children.is_empty())
+            .map(|(id, node)| (node.last_access_time, id))
+            .collect::<BTreeSet<_>>();
+        debug_assert_eq!(
```
jthomson04: This silently checks nothing under `cargo test --release`.

`cb1bae3bf8` narrowed the gate from `cfg(debug_assertions)` to `cfg(test)`, which is the right call for debug-build speed — but the body still uses `debug_assert_eq!`, so in a release-mode test run the function compiles and executes while the comparison is stripped. Since it's already test-only, plain `assert_eq!` costs nothing extra and keeps the check alive in both profiles.

lib/mocker/src/replay/offline/components/types.rs
```diff
@@ -111,26 +111,60 @@ pub(in crate::replay) struct ObservedOffloadEffects<Events: EngineEventBatch> {
...
+    pub(in crate::replay::offline) fn prepare_scheduled_completion(
+        &mut self,
+        at_ms: f64,
+        capacity: usize,
+    ) {
+        assert!(
+            self.scheduled_completion.is_none(),
+            "offline replay engine effects already contain a scheduled completion"
+        );
+        self.scheduled_completion = Some(ScheduledWorkerCompletions {
```
jthomson04: This makes an empty batch representable, and the invariant that rules it out is spread across three functions.

Once this installs `Some(ScheduledWorkerCompletions { payloads: vec![] })`, if the rank loop pushed nothing then `is_empty()` returns `false` (it only checks `is_none()`), the effects escape `drive_ready`, and `push_worker_completions` panics on its `payload_count > 0` assert.

It's unreachable today — `group_end_ms > now_ms` implies at least one rank scheduled — but it only holds by reasoning across `drive_ready`, `is_empty`, and `push_worker_completions`. Reserving lazily instead (a capacity hint on `schedule_completion`, or `payloads.reserve(n)` on first push) makes the empty state unrepresentable and deletes this function.

lib/mocker/src/replay/offline/runtime_utils.rs
```diff
@@ -89,36 +96,39 @@ pub(super) fn push_worker_completion<Events: EngineEventBatch>(
...
+        }
+    };
+    events.push(SimulationEvent {
+        at_ms,
+        seq_no: *next_event_seq,
+        kind,
     });
-    *next_event_seq += 1;
+    *next_event_seq = next_event_seq
+        .checked_add(payload_count)
```
jthomson04, line 113: Advancing by `payload_count` rather than 1 is functionally inert — worth a comment saying why, or simplifying.

Relative order of every later push is identical whether this adds 1 or N, since nothing is ever pushed between the payloads of one batch. If the intent is to keep `seq_no` values comparable with pre-batching runs, that's a reasonable goal but non-obvious enough to deserve a line of comment. Otherwise `+= 1` also removes the `u64::try_from(payloads.len()).expect(...)` above, which is infallible on any 64-bit target.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -549,6 +686,57 @@ where
                 continue;
             }
 
+            let mut effects = EngineEffects::default();
+            if rank_ids.len() == 1 {
```
jthomson04: The tail of this branch is byte-identical to the grouped path's tail:

```rust
if !effects.is_empty() {
    if group_end_ms <= now_ms { self.refresh_group(worker_id); }
    return Ok(effects);
}
if self.group_is_ready(worker_id) { self.deferred_ready_groups.insert(worker_id); }
```

Since the only real differences are skipping the `SmallVec` staging buffer, the max-fold, and `prepare_scheduled_completion`, restructuring so the singleton branch just produces `(effects, group_end_ms)` and falls through would drop ~10 duplicated lines and one `continue` — and would keep the two paths from drifting, which is precisely what happened with the `.max(now_ms)` clamp and the alignment call in `cb1bae3bf8`.

lib/mocker/src/replay/offline/events.rs
```diff
@@ -15,6 +15,20 @@ pub(crate) enum SimulationWorkerStage {
     Decode,
 }
 
+#[derive(Debug)]
+pub(crate) struct WorkerCompletionPayload<Events: EngineEventBatch = ()> {
```
jthomson04: Follow-up perf idea, not for this PR: now that `WorkerCompletionPayload` lives here, `SimulationEventKind::WorkerCompletion` could hold `Box<WorkerCompletionPayload>` instead of its 11 flattened fields.

That would delete the large destructure/rebuild in `pop_ready_worker_completions` (including the `fpm.map(Box::new)` / `fpm.map(|fpm| *fpm)` round-trip) and shrink `SimulationEvent` from ~100+ bytes — it's currently sized by that variant — to roughly 32. That's exactly the `memmove` this PR reports cutting from 3.73% → 1.09%, so there may be more left on the table.

The `#[allow(clippy::large_enum_variant)]` note on `ReadyWorkerCompletions` reasons about the transient *pop result*, where boxing would add an allocation per DP1 completion. It doesn't apply to the heap-resident event, which is memmoved on every sift.

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -764,12 +865,43 @@ mod tests {
...
+        pool.cancel(promoted.reservation);
+
         let second_eviction = reserve(&mut pool, &[], 2);
         assert_eq!(second_eviction.removed, vec![3]);
         pool.cancel(second_eviction.reservation);
         pool.assert_lru_consistent();
     }
 
+    #[test]
+    fn evicting_non_primary_duplicate_preserves_primary_lookup() {
```
jthomson04: Coverage gap: this only ever exercises `position() == 0`.

With exactly one duplicate, `HashCopies::remove`'s linear search over `duplicates` always hits the first element. A case with three or more copies of a hash, evicting a middle one, would cover the `VecDeque::remove(position)` path properly — that's the only branch in the new index where the search can actually walk.

jthomson04 (commented):
## Review notes

Traced the invariants these four optimizations depend on against the head at `450ed370d0`. Correctness looks sound; the blocking item is that the headline numbers were measured on pre-fix binaries.

### Blocking: P2/P3 benchmark numbers predate `cb1bae3bf8`

The review-fix commit changed the singleton fast path in two ways:

```diff
-                let group_end_ms = executed.end_ms;
+                let group_end_ms = executed.end_ms.max(now_ms);
+                let align_collector = if self.pass_mode == EnginePassMode::Visible { ... };
                     executed,
-                    None,
+                    align_collector,
```

The frozen binaries at `278dbee099` (P2) and `4e515437f0` (P3) were built *before* that fix, i.e. with `align_pass_token_times` skipped for singleton groups. The PR description's own profile evidence confirms it — "singleton `align_pass_token_times`: 3.16% vLLM and 3.66% SGLang self to no samples" — and the shipped code calls it again.

So P1→P2 (+8.31% SGLang / +7.06% vLLM) and the combined P0→P3 figures (+17.38% / +10.60%) overstate what the merged code delivers. The identical-reuse / identical-token-total checks don't cover this either, since alignment affects latency attribution rather than reuse. Please re-run P0→P3 against the current head. The fast path may well still be a win from dropping the `SmallVec` staging alone, but the number should come from the final code.

### Verified invariants

These all hold, but several are implicit and worth recording:

- `split_node` retains `child_id` as the **suffix** node with its original `last_access_time`, so a split never invalidates an evictable-index key and needs no index update. The entire `evictable_leaves` design depends on this silently — worth a one-line comment at `radix_cache.rs:644` so a future refactor that flips the split direction doesn't quietly corrupt the index.
- Every write to `last_access_time` goes through `touch_node` except `create_child`/`split_node` (fresh nodes) and `radix_cache.rs:329` (root, never indexed). No stale index keys are reachable.
- `inc_lock_ref` correctly removes from the index *before* incrementing `lock_ref`; the `was_unlocked` restructure is equivalent to the old `lock_ref == 1` check.
- `OfflineWorkerState::in_flight` is only mutated by `receive_request`, `apply_command`, and `mark_completed`, all now inside `update_worker`. `try_execute_pass`/`try_execute_hidden_pass`, still reached through the raw `required_worker_mut` split-borrow helper, don't touch it — so the aggregate can't drift.
- `only_idle_events_remain` (`agg.rs:462`, `disagg.rs:1676`) allow-lists `WorkerReady | ScalingTick`, so `WorkerCompletionBatch` correctly counts as pending work; `ordering_rank`'s `_ => 0` gives it the same tie-break rank as `WorkerCompletion`.
- Batching preserves same-timestamp ordering: the N payloads already got consecutive `seq_no`s and nothing was ever pushed between them.

The `#[cfg(test)]` full-index recomputation at every `radix_cache` mutation boundary is the strongest part of this PR — it makes index-drift bugs essentially unshippable.

### Undocumented behavior change

Eviction tie-breaking when two leaves share a `last_access_time` moved from `FxHashSet` iteration order to `NodeId` order. That's strictly better (deterministic and reproducible), but it can change victim selection and isn't mentioned in the description or the `97f38efdfc` commit message.

### Smaller items

- `WorkerCompletionPayload` is now `pub(crate)` with bare `pub` fields; the `runtime_utils` version it replaced was `pub(super)`. `pub(in crate::replay::offline)` would match its sibling `ScheduledWorkerCompletions`.
- README: the `runtime_utils.rs` bullet no longer mentions the type it now owns (`ReadyWorkerCompletions`).
- `HashCopies` `push_back`/`pop_front` reproduces the old `SmallVec` `first()`/`remove(pos)` promotion order exactly, and the 16-byte value vs. SmallVec's 24 checks out.

### Security / risk

No security surface: simulation crate, no I/O, no untrusted input. Public APIs, scheduler policy, and KV-event formats are preserved. `RadixCache::evictable_leaves` going `pub` → private is a good tightening with no external users (verified by grep across `lib/`), and it incidentally fixes the misplaced `/// Total token count in evictable nodes.` doc comment, which was sitting on the wrong field.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `61b7fb4034`. `in_flight()` now returns the tracked aggregate directly; full recomputation remains only at tracked mutation boundaries.

lib/mocker/src/cache/radix_cache.rs
jthomson04: Fixed in `61b7fb4034`. The test-only full index check now uses `assert_eq!`, so it remains active under release-mode tests.

lib/mocker/src/replay/offline/components/types.rs
jthomson04: Fixed in `61b7fb4034`. `prepare_scheduled_completion` is gone; `schedule_completion` creates the batch lazily on its first payload using a capacity hint, so empty scheduled batches are no longer representable.

jthomson04:
Addressed the remaining review feedback in `61b7fb4034`.

In addition to the seven inline threads, the follow-up:

- documents the radix split suffix-node timestamp/index identity
- narrows `WorkerCompletionPayload` and related event fields to the offline replay module
- routes disaggregated immediate completions through `apply_worker_completion`
- updates the offline README for `ReadyWorkerCompletions`
- relies on the existing `single_rank_visible_timeline_matches_grouped_path` parity test instead of adding redundant test coverage

Validation is complete: the full mocker suite passed (693 tests), as did the
focused cache/offline suites, KVBM-offload offline suite (183 tests), clippy,
formatting, and `git diff --check`.

The refreshed exact-head 128/128 comparison is now in the PR body. Against P0,
the final head measured +3.72% for vLLM and +7.19% for SGLang, with exact
normalized report parity. These are one-point measurements, not statistical
claims.

The boxed singleton heap-event suggestion is intentionally deferred to a
separate measured optimization.

lib/mocker/src/replay/offline/runtime_utils.rs
jthomson04: Documented in `61b7fb4034`: advancing by payload count preserves the sequence values that later events would have received before completion batching.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Fixed in `61b7fb4034`. The singleton and grouped branches now produce a common `(effects, group_end_ms)` result and share the refresh/defer tail; the singleton fast path still preserves token-time alignment.

lib/mocker/src/replay/offline/events.rs
jthomson04: Agreed; I am leaving boxed singleton heap events for a separate measured follow-up. It changes the DP1 allocation tradeoff and is outside this review-fix commit.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Fixed in `61b7fb4034`. The test now creates a primary plus three duplicates, releases and evicts a middle duplicate (`position() == 1`), and verifies primary lookup and index consistency.

lib/mocker/src/replay/offline/disagg.rs
```diff
@@ -1955,23 +1970,10 @@ where
...
-                    .insert(payload.worker_idx, fpm, self.now_ms);
-            }
-            self.process_prefill_pass(
-                payload.worker_idx,
-                payload.completed_requests,
-                payload.output_signals,
-                payload.lifecycle_events,
-                payload.engine_events,
-            )?;
+            self.apply_worker_completion(payload)?;
```
jthomson04, line 1973: Routing immediate completions through `apply_worker_completion` is fine, but the description oversells what it changes.

`on_scheduled_completion`, FPM buffering, and `process_prefill_pass`/`process_decode_pass` were already identical on both paths. The only actual delta is the added `wake_deferred_actions(stage, worker_idx)` — and that call can never wake anything here:

- All four `ActionExecution::Deferred` sites (`ActivateDestination`, `ReleaseSource`, `CancelSource`, `CancelDestination`) are gated on `worker_is_busy(worker_idx)`.
- `mark_busy()` is reachable only from `lower_executed_pass` when `end_ms > start_ms` (`engine.rs:631`) and from the empty-rank barrier when `group_end_ms > now_ms` (`engine.rs:769`); `mark_idle()` only from `on_scheduled_completion` (`engine.rs:846`).
- An immediate completion comes from a pass with `end_ms <= now_ms`, which never calls `mark_busy`, and `drive_ready` refuses to execute a group with any busy rank.

So at an immediate completion the worker is provably idle, and its deferred map was already drained by its previous scheduled completion. The wake is inert.

Defensive symmetry between the two paths is worth having, so I'd keep the change. But the Details bullet — "Disaggregated immediate completions use the same worker-accounting, FPM, lifecycle, and deferred-action path as scheduled completions" — reads as though a gap was closed, when three of those four were already shared and the fourth is a no-op. Suggest restating it as symmetry-for-safety rather than a fix, and adding a short comment here noting that it holds only because a zero-duration pass never marks the worker busy. Otherwise a future reader will assume this call is load-bearing and design around it.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -306,9 +360,11 @@ impl RadixCache {
 
             matched_pages += common_len;
             current = child_id;
-            self.nodes[current].last_access_time = now;
+            self.touch_node(current, now);
```
PeaBrane: [AI review][P2] Avoid the remove → insert → remove cycle on matched leaves.

`touch_node` removes and reinserts an evictable terminal leaf, but both production callers immediately call `inc_lock_ref`, which removes it again. That turns a common reusable-prefix hit into three `O(log L)` ordered-tree mutations, plus reinsertion on unlock. The reported benchmark shows a net win for its workload, but a high-hit, low-eviction shape may regress. Could the match-and-lock path update the timestamp without reinserting the terminal leaf, or could an eviction-free benchmark bound this tradeoff?

lib/mocker/src/cache/vllm_block_pool.rs
```diff
@@ -36,6 +36,71 @@ struct BlockCopy {
...
+        }
+    }
+
+    fn push(&mut self, id: BlockCopyId) {
+        self.duplicates
+            .get_or_insert_with(|| Box::new(VecDeque::new()))
+            .push_back(id);
+    }
+
+    fn remove(&mut self, id: BlockCopyId) -> CopyRemoval {
```
PeaBrane, line 67: [AI review][P2] Cover primary promotion while multiple overflow copies remain.

This branch promotes `pop_front()` and must retain the rest of the duplicate queue. The tests cover primary promotion with one duplicate and middle-duplicate removal while the primary remains, but not primary eviction with several duplicates. A regression that drops the remaining tail would pass both tests. Please add a primary-plus-three case that evicts the primary, verifies the promoted lookup, and confirms removal is reported only for the final physical copy.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -1033,6 +1108,64 @@ mod tests {
         assert_eq!(cache.match_prefix(&[1, 2, 3]).0, 3);
     }
 
+    #[test]
+    fn touching_an_evictable_leaf_updates_order() {
+        let mut cache = RadixCache::new(100, 1);
+        cache.insert(&[1, 2, 3], &[0, 1, 2]);
+        std::thread::sleep(std::time::Duration::from_millis(1));
```
PeaBrane: [AI review][P3] Make the LRU-order tests independent of wall-clock resolution.

These 1 ms sleeps assume `Instant::now()` produces distinct timestamps; a coarse or virtualized clock can collapse them and let the `NodeId` tie-break choose the victim. The partial-eviction/parent-promotion test also checks membership but not the resulting age. A controllable test clock or explicit timestamps, with a competitor leaf and next-victim assertions after extension and promotion, would make the actual LRU contract deterministic.

PeaBrane (approved):
## AI-generated review

Static multi-agent review against `61b7fb4034`. The inline findings are non-blocking; no correctness or lifecycle blockers were found. No tests were run as part of this review.

✅

lib/mocker/src/cache/radix_cache.rs
jthomson04: Fixed in `f4c80b5986`. Both production SGLang lease paths now use `match_prefix_hashes_and_lock`: the matcher defers the terminal touch, removes an unlocked terminal leaf once, updates its timestamp, and locks without reinserting and removing it again. Ordinary matching and partial-split timestamp semantics are unchanged. The isolated 128/128 SGLang point improved from 176.237M to 176.452M processed tok/s (+0.12%), with exact normalized report parity.

lib/mocker/src/cache/vllm_block_pool.rs
jthomson04: Fixed in `f4c80b5986`. The last-copy test now creates a primary plus three duplicates, evicts the primary, verifies lookup resolves to the promoted first duplicate while two overflow copies remain, then evicts each remaining copy and reports hash removal only for the final physical copy.

lib/mocker/src/cache/radix_cache.rs
jthomson04: Fixed in `f4c80b5986`. Radix tests now use a test-only controllable clock instead of sleeps and assert the next victim after touch and leaf extension, unlock, partial eviction, and unlocked-parent promotion. Production builds have no clock field or branch.

PR #12357, opened 2026-07-29
fix: pin uv to one version across all container stages

nv-anants:
`vllm-runtime` builds began failing on 2026-07-29 with:

```
× Failed to read `ai-dynamo-runtime @ file:///opt/dynamo/wheelhouse/ai_dynamo_runtime-1.4.0-cp310-abi3-manylinux_2_28_x86_64.whl`
├─▶ Failed to deserialize cache entry
╰─▶ array had incorrect length, expected 4
```

uv 0.12.0 (published 2026-07-28 18:58 UTC) added a `size` field to its cached `Archive` and
`Revision` entries without bumping the cache bucket version. A newer uv still reads older
entries, but an **older uv fails on newer ones** — the message is `rmp_serde` finding a
5-element array where the older struct expects 4.

Every stage shares one BuildKit uv cache mount. The mounts carried no `id=`, so BuildKit keyed
them by target path alone: one `/root/.cache/uv` shared across every stage and every image built
on that builder. Meanwhile the stages ran *different* uv versions:

| uv source | version |
|---|---|
| `dynamo_base`, `wheel_builder` via `ghcr.io/astral-sh/uv:latest` | 0.12.0 (as of 2026-07-28) |
| `manylinux_2_28` bundled, at `/usr/local/bin` | 0.11.32 |
| `vllm/vllm-openai` bundled, at `/usr/local/bin` | 0.11.32 |
| `vllm-ci-test-repo` (XPU) and `vllm-openai-cpu` bundled, at `/root/.local/bin` | 0.11.x |

As soon as any 0.12.0 stage wrote to the shared cache, every older reader failed. It looked
intermittent because it depended on which image last rebuilt its uv layer on a given builder.

Note the CUDA `pre_runtime` never ran our uv at all: it copied uv into `/bin`, which the vLLM
image's `/usr/local/bin/uv` shadows on PATH. Pinning alone would not have fixed this.

## Details:

1. **Single pinned version.** `uv_version` in `container/context.yaml`, consumed by every
   template. Replaces four `ghcr.io/astral-sh/uv:latest` references and two stragglers pinned
   to `0.10.7`.

2. **Dedicated install directory ahead of base-image copies.** uv installs to `/opt/uv/bin`
   with `ENV PATH=/opt/uv/bin:${PATH}`. Prepending `/usr/local/bin` instead would reorder every
   binary in it; `/opt/uv/bin` holds only `uv` and `uvx`, so nothing else changes resolution.
   Two stages (`planner_builder`, distroless `planner`) set an absolute PATH with no `${PATH}`,
   so the directory is spliced in explicitly there.

3. **Version-keyed cache mount IDs.** `id=uv-root-${uv_version}` and `id=uv-dynamo-${uv_version}`,
   so a future uv bump rotates to a fresh cache instead of mixing formats. This also means no
   builder cache purge is needed to land the fix — the poisoned default-ID cache is abandoned.

closes: OPS-7955

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized Docker builds on a pinned `uv` version for more predictable tooling behavior.
  * Isolated build caches by `uv` version to prevent incompatible cached entries from being reused.
  * Updated `uv` and `uvx` locations across container stages for consistent runtime availability.
* **Documentation**
  * Updated container documentation to reflect the revised `uv` versioning, paths, and caching behavior.

### DIFF
--- a/container/Dockerfile.docs
+++ b/container/Dockerfile.docs
@@ -22,8 +22,8 @@ ENV DYNAMO_COMMIT_SHA=$DYNAMO_COMMIT_SHA
 ARG DYNAMO_DOCS_VERSION=dev
 ENV DYNAMO_DOCS_VERSION=$DYNAMO_DOCS_VERSION
 
-# TODO: Pin uv image to a specific version tag for reproducibility (e.g. ghcr.io/astral-sh/uv:0.10.7)
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+# Keep in sync with container/context.yaml `uv_version`.
+COPY --from=ghcr.io/astral-sh/uv:0.12.0 /uv /uvx /usr/local/bin/
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
--- a/container/Dockerfile.docs
+++ b/container/Dockerfile.docs
@@ -23,7 +23,8 @@ ARG DYNAMO_DOCS_VERSION=dev
 ENV DYNAMO_DOCS_VERSION=$DYNAMO_DOCS_VERSION
 
 # Keep in sync with container/context.yaml `uv_version`.
-COPY --from=ghcr.io/astral-sh/uv:0.12.0 /uv /uvx /usr/local/bin/
+COPY --from=ghcr.io/astral-sh/uv:0.12.0 /uv /uvx /opt/uv/bin/
+ENV PATH=/opt/uv/bin:${PATH}
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
--- a/container/context.yaml
+++ b/container/context.yaml
@@ -34,6 +34,9 @@ dynamo:
   # it so planner NOTICES attribute only what the builder adds on top.
   planner_baseline_sbom: python@043085f9
   python_version: "3.12"
+  # Every stage must use the same uv: stages share one BuildKit uv cache mount,
+  # and uv rejects cache entries written by a newer version.
+  uv_version: "0.12.0"
 
   nats_version: v2.10.28
   etcd_version: v3.5.30
--- a/container/templates/dev.Dockerfile
+++ b/container/templates/dev.Dockerfile
@@ -359,13 +359,13 @@ COPY --from=wheel_builder --chown=dynamo:0 --chmod=775 /workspace/.venv/bin/matu
 # SGLang XPU: conda env from framework stage; install uv and maturin.
 # uv doesn't natively recognize conda envs (no pyvenv.cfg), so we use
 # --python to target the conda interpreter explicitly.
-COPY --from=ghcr.io/astral-sh/uv:0.10.7 /uv /tmp/uv-binary
+COPY --from=ghcr.io/astral-sh/uv:{{ context.dynamo.uv_version }} /uv /tmp/uv-binary
 RUN cp /tmp/uv-binary ${VIRTUAL_ENV}/bin/uv && \
     chmod +x ${VIRTUAL_ENV}/bin/uv && \
     pip install maturin[patchelf]
 {% else %}
 # SGLang CUDA: Create venv with --system-site-packages to inherit runtime packages
-COPY --from=ghcr.io/astral-sh/uv:0.10.7 /uv /tmp/uv-binary
+COPY --from=ghcr.io/astral-sh/uv:{{ context.dynamo.uv_version }} /uv /tmp/uv-binary
 RUN mkdir -p /opt/dynamo/venv && \
     python3 -m venv --system-site-packages /opt/dynamo/venv && \
     cp -r /usr/local/lib/python${PYTHON_VERSION}/dist-packages/* \
@@ -414,7 +414,7 @@ ARG FRAMEWORK
 RUN --mount=type=bind,source=./container/deps/requirements.dev.txt,target=/tmp/requirements.dev.txt \
     --mount=type=bind,source=./container/deps/requirements.test.txt,target=/tmp/requirements.test.txt \
     # Cache uv downloads; uv handles its own locking for this cache.
-    --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=cache,id=uv-root-{{ context.dynamo.uv_version }},target=/root/.cache/uv \
     export UV_CACHE_DIR=/root/.cache/uv UV_GIT_LFS=1 UV_HTTP_TIMEOUT=300 UV_HTTP_RETRIES=5 && \
     # Git LFS init (needed for requirements with lfs=true); folded in to save a layer.
     git lfs install && \
--- a/container/templates/dev.Dockerfile
+++ b/container/templates/dev.Dockerfile
@@ -382,7 +382,7 @@ RUN mkdir -p /opt/dynamo/venv && \
 # CUDA: Runtime uses system Python, so --system-site-packages correctly inherits packages.
 RUN mkdir -p /opt/dynamo/venv && \
     python3 -m venv --system-site-packages /opt/dynamo/venv && \
-    ln -sf /usr/local/bin/uv /opt/dynamo/venv/bin/uv
+    ln -sf /opt/uv/bin/uv /opt/dynamo/venv/bin/uv
 {% else %}
 # CPU/XPU: Runtime uses /opt/venv from upstream vLLM-CPU image. Reuse it directly
 # instead of creating /opt/dynamo/venv, since --system-site-packages points to UV Python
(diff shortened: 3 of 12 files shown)


dmitry-tokarev-nv (approved):
LGTM.

container/context.yaml
```diff
@@ -34,6 +34,9 @@ dynamo:
   # it so planner NOTICES attribute only what the builder adds on top.
   planner_baseline_sbom: python@043085f9
   python_version: "3.12"
+  # Every stage must use the same uv: stages share one BuildKit uv cache mount,
+  # and uv rejects cache entries written by a newer version.
+  uv_version: "0.12.0"
```
dmitry-tokarev-nv, line 39: Right place for the pin — but the single source of truth stops at the templated stages. `container/Dockerfile.test:15` still mounts the unversioned cache:

```dockerfile
--mount=type=cache,target=/root/.cache/uv,sharing=shared \
```

That one isn't dead code: `shared-build-image.yml:637` and `xpu-ci.yaml:123` build it on the same builders immediately after the framework image, as `USER root`, using the runtime image's uv. So it's exercised for vllm / trtllm / dynamo-runtime / frontend / planner (sglang falls through to the `pip --break-system-packages` branch and is unaffected).

Two consequences:

- The poisoned default-ID cache isn't actually abandoned — it just loses every writer except this one, so it stays alive and keeps growing on each builder.
- It re-opens the same cross-version mixing at the next bump: pin to 0.13 here and `Dockerfile.test` will still read whatever some other build left in the unnamed `/root/.cache/uv`.

Cheapest fix, and it also covers the `Dockerfile.docs` duplication: `ARG UV_VERSION` in both non-templated Dockerfiles with `id=uv-root-${UV_VERSION}`, value passed from the two workflows. Dropping the mount from `Dockerfile.test` is also defensible — it's one thin layer of test deps.

container/Dockerfile.docs
```diff
@@ -22,8 +22,9 @@ ENV DYNAMO_COMMIT_SHA=$DYNAMO_COMMIT_SHA
 ARG DYNAMO_DOCS_VERSION=dev
 ENV DYNAMO_DOCS_VERSION=$DYNAMO_DOCS_VERSION
 
-# TODO: Pin uv image to a specific version tag for reproducibility (e.g. ghcr.io/astral-sh/uv:0.10.7)
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+# Keep in sync with container/context.yaml `uv_version`.
+COPY --from=ghcr.io/astral-sh/uv:0.12.0 /uv /uvx /opt/uv/bin/
+ENV PATH=/opt/uv/bin:${PATH}
```
dmitry-tokarev-nv, line 27: Second source of truth for the version, with a comment as the only sync mechanism. I checked `.pre-commit-config.yaml` — no hook covers it, so this will drift.

Worth weighing when you pick a fix: this file has **no uv cache mount at all**, so it isn't in the failure class the PR describes. The pin here is consistency-only, and the "keep in sync" line creates a standing maintenance obligation that buys nothing operationally.

Three ways out, roughly in order of effort:

1. Drop the sync comment and state that the docs image is deliberately independent of the shared-cache constraint. Zero machinery, honest about what this pin is.
2. Add a two-line pre-commit check that greps this file against `container/context.yaml`. Actually enforceable.
3. Render this file through `render.py` like the rest.

I'd skip a bare `ARG UV_VERSION=0.12.0` unless something really passes it — with only a default, it's the same drift risk plus a layer of indirection.

container/templates/dev.Dockerfile
```diff
@@ -414,7 +414,7 @@ ARG FRAMEWORK
 RUN --mount=type=bind,source=./container/deps/requirements.dev.txt,target=/tmp/requirements.dev.txt \
     --mount=type=bind,source=./container/deps/requirements.test.txt,target=/tmp/requirements.test.txt \
     # Cache uv downloads; uv handles its own locking for this cache.
-    --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=cache,id=uv-root-{{ context.dynamo.uv_version }},target=/root/.cache/uv \
```
dmitry-tokarev-nv, line 417: Now that the ID is explicit and deliberate, `sharing=` across the `uv-root-*` group is three different things:

- `sharing=shared` — `wheel_builder.Dockerfile:227`, `:514`, `:538`, `:622`, `:714`, `:738`
- `sharing=locked` — `vllm_runtime.Dockerfile:128`, `:141`, `:153`, `:184`, `:201`, `:269`; `trtllm_runtime.Dockerfile:119`
- unspecified (→ shared) — this line

Pre-existing, since the default target-path ID grouped exactly the same set. But the grouping used to be accidental and is now intentional, which makes it worth resolving rather than inheriting. Every comment in the tree asserts uv handles its own locking for this cache — including line 416 directly above — and if that's right, the `locked` mounts are serializing concurrent builds for no benefit.

At minimum, make this one explicit so the inconsistency is visible rather than implied:

```suggestion
    --mount=type=cache,id=uv-root-{{ context.dynamo.uv_version }},target=/root/.cache/uv,sharing=shared \
```

dmitry-tokarev-nv (commented):
Reviewed the full stage graph, not just the diff. Root cause and fix are both right — everything below is follow-up material, not a blocker, and a separate PR is fine.

### What I verified holds

I traced `PATH` through every stage that invokes uv, including the `ENV`-doesn't-cross-`FROM` trap:

- `dynamo_base:40` → `dynamo_runtime:10` (`FROM dynamo_base`) → `dev:177` (`FROM runtime`) inherit correctly.
- `frontend:106`, `vllm_runtime:74`, `wheel_builder:210` re-declare after their COPY.
- `trtllm_runtime` re-adds `/opt/uv/bin` in **both** ENV blocks of the rebase `pre_runtime` stage (`:209`, `:218`) — the stage that re-`FROM`s upstream and would otherwise silently lose it.
- `planner` splices it into the two stages that set an absolute PATH with no `${PATH}` (`:54`, `:119`).
- SGLang never runs uv in its runtime stage (all `pip`, no `VIRTUAL_ENV`); its dev stage copies the pinned binary into the venv (`dev:362`, `dev:368`).

`/opt/uv/bin` holding only `uv`/`uvx` is what makes the prepend safe — including in `frontend.Dockerfile`, where it lands ahead of the venv `bin` and still can't reorder anything. Also confirmed no in-repo consumer (tests, `deploy/`, `recipes/`, workflows, `container/*.sh`) references the old `/bin/uv`, `/usr/bin/uv`, or `/usr/local/bin/uv`, so the path move isn't user-visible here.

The four inline bot findings are stale against HEAD — `74a3cfefd` / `6ab039f24` / `78c80ccc8` fixed the `/usr/local/bin` iteration, the usr-merge comment, and the TRT-LLM rebase PATH respectively.

### Coverage gap worth an XPU run

65 checks pass and every container build is green (vllm / sglang / trtllm / planner / frontend / dynamo-runtime, multi-arch, EFA). But `xpu-ci-vllm` and `xpu-status-check` are skipping, and there's no `vllm-runtime cpu` build in the run — and those are exactly the images your table identifies as bundling uv at `/root/.local/bin`, i.e. where the `/opt/uv/bin` prepend is load-bearing rather than incidental. On CUDA the shadow is `/usr/local/bin` and the prepend clears it comfortably; the `/root/.local/bin` case is the one CI hasn't exercised. The mechanism is unconditional, so I expect it works.

Related, and I'd keep your choice: pinning **up** to 0.12.0 rather than **down** to 0.11.32 (what every base bundles) means any stage where a bundled older uv still wins the PATH race now hard-fails against the versioned cache instead of degrading quietly. That's the better trade for a failure this hard to notice — it's just an argument for running the XPU/CPU matrix rather than skipping it.

### Doc drift

`container/README.md:169` still documents `--mount=type=cache,target=/root/.cache/uv` with no `id=`; line 27 was updated but this one wasn't. Since the version-keyed ID is the whole mechanism, that line is the natural place for a sentence on why it's keyed.

### Nit

`local_dev.Dockerfile:24` re-prepends `/usr/local/bin` ahead of `/opt/uv/bin` (which sits at the tail inside `${PATH}`). Not a build-time problem — local-dev runs no uv — and `${VIRTUAL_ENV}/bin/uv` is the pinned copy or symlink in every framework branch, so the pinned uv still wins at runtime. But it's the one place that re-creates the shadowing this PR removes, and it survives only because of that symlink. A comment, or putting `/opt/uv/bin` in front explicitly, would keep the invariant local instead of load-bearing two files away.

container/templates/dev.Dockerfile
nv-anants: sharing we have tried before and it was leading to some cache issues from what i remember, will defer to later

container/Dockerfile.docs
nv-anants: docs are quite independent, okay for now. There is a comment to for manual sync, hopefully that keeps it close

container/context.yaml
nv-anants: dropped the mount for now 83e6ccd75829907728c84399c8f4b04e729f8830

PR #12359, opened 2026-07-29
ci(deploy): use runtime-versioned image tags

sttts:
## Summary

- tag CI deployment images as `<dynamo-version>-ci-<sha>-<image>` when copying them to ACR
- keep the immutable SHA tag as the ECR build artifact
- pass the resulting ACR tag through deploy, checkpoint, and GAIE workflows
- remove test-side `runtimeVersionOverride` injection and version parsing

## Dependency

Prefactor: #12362 introduces the reusable CI image-tagging changes against `main`. This stacked PR currently carries those workflow changes for end-to-end validation. After #12362 merges, this PR will be rebased onto it and retain only the #10494-specific test cleanup.

## Why

#10494 validates runtime compatibility from the main container image tag. CI currently replaces those images with SHA-tagged artifacts, forcing generic test utilities to add an override. Giving the CI artifacts an accurate SemVer-compatible tag lets admission use the normal production path.

## Validation

- pre-commit hooks
- `git diff --check`

Stacked on #10494.


sttts:
Closing this follow-up because its final changes are now folded directly into #10494 after rebasing #10494 onto the prefactor in #12362. The workflow portion is provided by #12362, and the remaining test-helper cleanup is already present on current main.

PR #12362, opened 2026-07-29
refactor(ci): prefactor runtime-version image tags as semver for #10494

sttts:
## Prefactor for #10494

This PR prepares the CI image-tagging workflow required by #10494. It is intentionally separate so the runtime-version PR can stay focused on operator behavior.

## Summary

- Publish every image copied through `shared-copy.yml` to ACR as `<dynamo-version>-ci-<sha>-<image>`.
- Keep immutable SHA-keyed source artifacts in ECR unchanged.
- Expose the exact ACR target tag to deploy, checkpoint, and GAIE workflows.
- Remove caller-side opt-ins and assumptions that ACR tags always use `<sha>-<suffix>`.

## Impact

- No operator or runtime behavior changes.
- No custom-image behavior changes.
- ECR build tags remain unchanged.
- All ACR tags produced by the shared copy workflow become SemVer-compatible.

## Validation

- Pre-commit hooks
- `git diff --check`
- Workspace version extraction (`1.4.0`)
- Tests not run (workflow-only change)

### DIFF
--- a/lib/llm/src/engines.rs
+++ b/lib/llm/src/engines.rs
@@ -10,6 +10,7 @@ use async_stream::stream;
 use async_trait::async_trait;
 
 use dynamo_runtime::engine::{AsyncEngine, AsyncEngineContextProvider, ResponseStream};
+use dynamo_runtime::error::{DynamoError, ErrorType};
 #[cfg(test)]
 use dynamo_runtime::pipeline::RequestStream;
 use dynamo_runtime::pipeline::{Error, ManyIn, ManyOut, SingleIn};
@@ -433,7 +434,11 @@ where
 
         // Validate the request first
         if let Err(validation_error) = request.validate() {
-            return Err(anyhow::anyhow!("Validation failed: {}", validation_error));
+            return Err(DynamoError::builder()
+                .error_type(ErrorType::InvalidArgument)
+                .message(format!("Validation failed: {validation_error}"))
+                .build()
+                .into());
         }
 
         // Forward to inner engine if validation passes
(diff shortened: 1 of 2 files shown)


julienmancuso (approved):

.github/workflows/shared-copy.yml
```diff
@@ -67,7 +82,23 @@ jobs:
...
+              Cargo.toml)
+            DYNAMO_VERSION=${DYNAMO_VERSION%%[-+]*}
+            if [[ ! "${DYNAMO_VERSION}" =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
+              echo "::error::Invalid Dynamo workspace version: ${DYNAMO_VERSION}"
+              exit 1
+            fi
+            TARGET_TAG="${DYNAMO_VERSION}-ci-${IMAGE_TAG}"
+          fi
+
           echo "image_tag=${IMAGE_TAG}" >> $GITHUB_OUTPUT
```
nv-tusharma: maintaining 2 seperate variables target_tag & image_tag might get confusing in the future, anyway we can consolidate?

.github/workflows/shared-copy.yml
```diff
@@ -24,6 +24,18 @@ on:
         required: false
         type: string
         default: ''
+      runtime_versioned_target_tag:
```
nv-tusharma: If this is truly required, we can just default to this & not require additional input?

.github/workflows/shared-copy.yml
sttts: Done. I removed the ambiguous `image_tag` name. The two remaining values now represent the distinct registry contracts explicitly: `source_tag` is the immutable ECR SHA tag, while `target_tag` is the SemVer-compatible ACR tag. They cannot be one value because the copy intentionally retags the artifact.

sttts: Correction: I kept the existing `image_tag` name for the immutable ECR source tag to minimize churn. `target_tag` is the only new value and is the SemVer-compatible ACR destination tag. They cannot be consolidated into one value because this workflow intentionally retags the artifact while copying it.

.github/workflows/shared-copy.yml
sttts: Done. The SemVer-compatible ACR tag is now the default for every `shared-copy.yml` invocation. I removed the boolean input and all caller-side opt-ins.

.github/workflows/shared-copy.yml
sttts: Implemented the consolidation: `source_tag` is now the internal immutable ECR source tag, and `image_tag` is the SemVer-compatible ACR tag exposed by the reusable workflow and consumed by downstream jobs. The only remaining `target_tag` is the existing input name of the `skopeo-copy` action.

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/pull/12362/commits/28c8d6207cac098e07801d6a7ebe68d3e84446bb

PR #12368, opened 2026-07-29
fix(ci): harden Grove webhook readiness in vCluster

lavanyavijayk:
## Overview:

  Hardens Grove webhook readiness in vCluster CI against stale host Pod readiness caused by
  Kubernetes version skew. This complements #12255 by enabling Grove’s webhook-aware probes and adds
  a temporary CI-only Service-routing workaround.

  ## Details:

  - Enable Grove’s existing /healthz and /readyz probes.
  - Set publishNotReadyAddresses: true on the Grove Service only in short-lived vCluster CI.
  - Wait for both Dynamo and Grove operator containers.
  - Verify the setting on the virtual and translated host Services.
  - Run a server-side PodCliqueSet admission smoke test before deploy tests.
  - Add Grove-specific failure diagnostics.
  - Update the generated Helm documentation.

  Validated with Helm lint, 33 Helm tests, pre-commit checks, rendered manifest inspection, and
  embedded Bash syntax validation.

  Start with .github/actions/setup-dynamo-operator/action.yml, particularly the Grove Service patch,
  host-side verification, and PodCliqueSet admission smoke test. Then review deploy/helm/charts/
  platform/values.yaml for the Grove health-probe configuration.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12368" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable Grove health probes, enabled by default.
  * Grove readiness now reflects certificate and webhook server availability.

* **Bug Fixes**
  * Improved operator readiness checks and admission webhook validation.
  * Added coverage for Grove admission testing alongside existing Dynamo checks.
  * Enhanced deployment failure diagnostics with additional Grove service, endpoint, pod, and configuration details.

* **Documentation**
  * Documented the new Grove health probe configuration and default behavior.


.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -548,14 +579,49 @@ runs:
...
+            if [ "${HOST_GROVE_PUBLISH_NOT_READY}" = "true" ]; then
+              break
+            fi
+          fi
+          if [ "${SECONDS}" -ge "${end}" ]; then
+            echo "::error::vCluster did not preserve publishNotReadyAddresses on the host Grove webhook Service"
+            exit 1
+          fi
+          sleep 2
+        done
```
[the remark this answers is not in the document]
lavanyavijayk: Confirmed on the AKS CI cluster that the translated Grove Service preserves app.kubernetes.io/name=grove-operator alongside the injected vCluster labels. The selector is intentionally strict, so future label changes will fail verification instead of silently passing.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -460,47 +472,64 @@ runs:
...
+            echo "Waiting for ${display_name} containers... status=${status:-not found}"
+            sleep 15
+          done
+        }
+
+        wait_for_containers \
+          '^dynamo-platform-dynamo-operator-controller-manager-' \
+          'Dynamo operator'
+        wait_for_containers '^grove-operator-' 'Grove operator'
         echo "::endgroup::"
```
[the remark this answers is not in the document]
lavanyavijayk: The wait checks container readiness from the READY column, not the stale Pod-level Ready condition. In the reported failure the container was 1/1 and serving, while the Pod condition remained false or missing; publishNotReadyAddresses specifically bypasses that incorrect Pod-level condition.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -590,23 +656,42 @@ runs:
...
+              spec:
+                roleName: smoke
+                replicas: 1
+                minAvailable: 1
+                podSpec:
+                  containers:
+                  - name: smoke
+                    image: registry.k8s.io/pause:3.10
+        EOF
+        }
```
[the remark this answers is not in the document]
lavanyavijayk: Confirmed against the pinned CRD: cliqueStartupType has an OpenAPI default, while headlessServiceConfig and terminationDelay are optional. The follow-up commit also verifies both Grove webhook configurations before the dry-run, so a missing webhook can no longer produce a false pass.

deploy/helm/charts/platform/values.yaml
```diff
@@ -292,6 +292,12 @@ grove:
   tolerations: []
   # -- Affinity for Grove pods
   affinity: {}
+  config:
+    server:
+      healthProbes:
+        # -- Enable Grove's webhook-aware liveness and readiness probes.
+        # The readiness endpoint stays false until certificates and the webhook server are ready.
+        enable: true
```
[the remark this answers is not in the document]
lavanyavijayk: Confirmed against the pinned Grove chart v0.1.0-alpha.12-rc1: config.server.healthProbes.enable is the exact value used by the Deployment template. Rendering the chart with it enabled produces the expected /healthz and /readyz probes.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -451,6 +451,18 @@ runs:
...
+        # version skew can leave the translated host Pod's Ready condition stale
+        # after Grove is serving its webhook. Grove's pinned chart does not
+        # expose publishNotReadyAddresses, so patch only the short-lived CI
+        # Service. The readiness wait and admission smoke test below still prove
+        # that the actual webhook process is healthy before deploy tests start.
+        kubectl --kubeconfig=${{ github.workspace }}/.kubeconfig-vcluster \
+          patch service grove-operator \
+          --namespace default \
+          --type merge \
+          --patch '{"spec":{"publishNotReadyAddresses":true}}'
```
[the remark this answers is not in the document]
lavanyavijayk: Confirmed against the pinned chart: service.name defaults to grove-operator, and the Service is rendered directly by Helm, so it exists when helm upgrade --install completes.

sara4dev (approved):
if this unblocks the CI for now, lets go with this.


But we need to revert these once we have updated vcluster version and simplify this.

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -460,47 +472,85 @@ runs:
...
         VKUBECONFIG=${{ github.workspace }}/.kubeconfig-vcluster
-        CRD=dynamographdeployments.nvidia.com
-        end=$((SECONDS + 1800))
-        until kubectl --kubeconfig="${VKUBECONFIG}" get "crd/${CRD}" >/dev/null 2>&1; do
-          if [ $SECONDS -ge $end ]; then
-            echo "::error::CRD ${CRD} was not created after 1800s"
-            exit 1
-          fi
-          sleep 5
+        for CRD in dynamographdeployments.nvidia.com podcliquesets.grove.io; do
```
nv-tusharma, line 475: We're doing this twice? https://github.com/ai-dynamo/dynamo/blob/01a9ef7909eec042d8bcf90eceeec3cc6c656f0e/.github/actions/setup-dynamo-operator/action.yml#L485

https://github.com/ai-dynamo/dynamo/blob/01a9ef7909eec042d8bcf90eceeec3cc6c656f0e/.github/actions/setup-dynamo-operator/action.yml#L560

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -548,14 +600,49 @@ runs:
...
+          echo "::error::Grove webhook Service does not publish not-ready addresses in the vCluster"
+          exit 1
+        fi
+
+        # The injected vCluster labels identify the translated Service without
+        # depending solely on application labels being preserved by the syncer.
+        GROVE_HOST_SELECTOR="vcluster.loft.sh/managed-by=${VCLUSTER_NAME},vcluster.loft.sh/namespace=default,app.kubernetes.io/name=grove-operator"
+        HOST_GROVE_SERVICE=
+        HOST_GROVE_PUBLISH_NOT_READY=
+        end=$((SECONDS + 120))
```
nv-tusharma, line 581: do we need this verification, does smoke_grove already address it?

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -618,19 +724,70 @@ runs:
...
+
+          configuration=$(kubectl --kubeconfig="${VKUBECONFIG}" \
+            get "${resource_kind}/${configuration_name}" \
+            -o json)
+          if ! jq -e --arg webhook_name "${webhook_name}" '
+            any(.webhooks[];
+              .name == $webhook_name
+              and .failurePolicy == "Fail"
+              and any(.rules[];
+                (.apiGroups | index("grove.io")) != null
```
nv-tusharma, line 730: I don't think these assertions are necessary since smoke_grow already validates thios

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -460,47 +472,85 @@ runs:
...
-          sleep 15
-        done
+        # container readiness reported by each webhook-aware probe; the
+        # admission smoke tests below are the authoritative end-to-end gate.
+        wait_for_containers() {
+          local pod_pattern="$1"
+          local display_name="$2"
+          local -a pod_statuses
+          local pod_status
+          local status
```
nv-tusharma: can we just update this to be 2 pods, 1 container each instead of writing this function? We likely don't need this flexibility to handle any pods x any containers since we know the amount expected.

nv-tusharma:
Summarizing for myself: 

The Grove Service `publishNotReadyAddresses: true` patch overrides the readiness signal that `grove.config.server.healthProbes.enable: true` just made meaningful, so the two read as redundant and someone will eventually delete one. Worth a short comment noting that the probe here buys the readiness wait-gate and remains the correct behavior after the temporary Service patch is removed.

.github/actions/setup-dynamo-operator/action.yml
lavanyavijayk: smoke_grove alone can succeed if either webhook configuration is absent or does not match the request, because Kubernetes would not invoke that webhook. These checks confirm that the expected mutating and validating webhook entries are installed, match CREATE requests for grove.io/v1alpha1 PodCliqueSet, and use failurePolicy: Fail. The subsequent smoke request then verifies that the configured fail-closed admission path is reachable, so I’d prefer to keep these assertions.

.github/actions/setup-dynamo-operator/action.yml
lavanyavijayk: Done — simplified the readiness gate to wait for exactly the two expected operator Pods, with each required to report 1/1. This removes the generic any-Pod/any-container helper. Addressed in f1952ccb8c.

.github/actions/setup-dynamo-operator/action.yml
lavanyavijayk: Removing the second wait.

.github/actions/setup-dynamo-operator/action.yml
lavanyavijayk: I think we should keep this verification. smoke_grove proves the webhook is reachable in the current run, but it can still pass when the Pod happens to be Ready even if vCluster failed to propagate publishNotReadyAddresses to the host Service. Since the readiness bug is intermittent, explicitly checking the translated host Service proves that the workaround itself is installed and will remain effective when stale Pod readiness occurs.

nv-tusharma (approved):
approving to unblock CI

sara4dev:
I am getting closer on this one - https://github.com/ai-dynamo/dynamo/pull/12279, can we wait for this checks to finish and merge that instead of this workaround?

sttts (approved):

PR #12380, opened 2026-07-29
feat(mocker): add typed LiveEngine handoff control

connorcarpenter15:
## Overview:

Implements the first half of DIS-2481 by adding the typed `LiveEngine` boundary needed to remove duplicated Mocker scheduler plumbing from `lib/llm`. This PR is additive: `LiveEngine::start` remains unchanged, and replay-only controls remain private.

## Details:

- Add `LiveEngineConfig` and `LiveEngine::start_with_config` for configured KV-event and FPM publishers.
- Add prepared request registration with route ownership, client/scheduler ID isolation, bounded output, duplicate-ID rejection, and stream-drop cancellation.
- Add typed handoff control and normalized lifecycle events while keeping scheduler IDs and command types private.
- Move lifecycle demultiplexing, dispatcher supervision, route cleanup, and idempotent graceful shutdown into `LiveEngine`.
- Prevent stale handoff controls from mutating replacement registrations and serialize admission acknowledgements with cleanup commands.
- Avoid forwarding admission events when no observer is configured.
- This is the first PR in the stack; it is followed by #12381.

### Validation

- `cargo test -p dynamo-mocker` — 691 passed
- `git diff --check`

- `lib/mocker/src/live.rs` for engine startup, dispatcher supervision, and shared shutdown behavior.
- `lib/mocker/src/live/request.rs` for request registration, submission, cancellation, and route ownership.
- `lib/mocker/src/live/handoff.rs` for the typed handoff API and generation-safe command routing.
- `lib/mocker/src/live/tests.rs` for lifecycle, route reuse, shutdown, and handoff regressions.
- `lib/mocker/src/scheduler/mod.rs` for the optional admission-forwarding path.

## Related Issues

- Tracked in Linear as DIS-2481

### DIFF
--- a/lib/mocker/src/live.rs
+++ b/lib/mocker/src/live.rs
@@ -19,6 +19,7 @@ use tokio::sync::{mpsc, oneshot, watch};
 use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
+use crate::common::handoff::{HandoffId, HandoffTransferTiming};
 use crate::common::protocols::{
     DirectRequest, FpmPublisher, KvEventPublishers, MockEngineArgs, OutputSignal,
 };
@@ -26,6 +27,7 @@ use crate::engine::{LiveEngineScheduler, create_engine_with_event_sender};
 use crate::scheduler::{
     LiveEngineEvent, MockerMetrics, SchedulerCancellationEnvelope, SchedulerCommand,
     SchedulerCommandEnvelope, SchedulerCommandResult, SchedulerEventSender, SchedulerHandle,
+    SchedulerLifecycleEvent,
 };
 
 #[derive(Default)]
@@ -38,6 +40,62 @@ type Routes = Arc<RequestRoutes>;
 
 const SCHEDULER_EVENT_CAPACITY: usize = 8;
 const DEFAULT_REQUEST_OUTPUT_CAPACITY: usize = 8;
+const HANDOFF_EVENT_CAPACITY: usize = 8;
+
+/// Runtime publishers used by one live Mocker scheduler.
+#[derive(Clone, Default)]
+pub struct LiveEngineConfig {
+    pub kv_event_publishers: KvEventPublishers,
+    pub fpm_publisher: FpmPublisher,
+}
+
+#[derive(Default)]
+struct HandoffRoutes {
+    by_id: DashMap<HandoffId, Arc<HandoffRoute>>,
+}
+
+type SharedHandoffRoutes = Arc<HandoffRoutes>;
+
+struct HandoffRoute {
+    handoff_id: HandoffId,
+    event_tx: Mutex<Option<mpsc::Sender<LiveHandoffEvent>>>,
+}
+
+impl HandoffRoute {
+    fn new(handoff_id: HandoffId, event_tx: mpsc::Sender<LiveHandoffEvent>) -> Self {
+        Self {
+            handoff_id,
+            event_tx: Mutex::new(Some(event_tx)),
+        }
+    }
+
+    async fn send(&self, event: LiveHandoffEvent, cancel: &CancellationToken) -> bool {
+        let event_tx = self.event_tx.lock().unwrap().as_ref().cloned();
+        let Some(event_tx) = event_tx else {
+            return false;
+        };
+        tokio::select! {
+            biased;
+            _ = cancel.cancelled() => false,
+            result = event_tx.send(event) => result.is_ok(),
+        }
+    }
+
+    fn shutdown(&self) {
+        self.event_tx.lock().unwrap().take();
+    }
+}
+
+/// Scheduler lifecycle events normalized for a registered disaggregated handoff.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LiveHandoffEvent {
+    SourceHeld {
+        transfer_timing: HandoffTransferTiming,
+    },
+    DestinationReserved {
+        transferable_prompt_tokens: usize,
+    },
+}
 
 pub(crate) struct ObservedAdmission {
     pub(crate) event: crate::scheduler::AdmissionEvent,
@@ -245,6 +303,7 @@ struct LiveEngineInner {
     command_tx: mpsc::Sender<SchedulerCommandEnvelope>,
     cancellation_tx: mpsc::Sender<SchedulerCancellationEnvelope>,
     routes: Routes,
+    handoff_routes: SharedHandoffRoutes,
     metrics_rx: tokio::sync::watch::Receiver<MockerMetrics>,
     request_output_capacity: Option<NonZeroUsize>,
     allow_zero_output: bool,
@@ -259,12 +318,31 @@ struct LiveEngineInner {
 struct LiveEngineTasks {
     scheduler_actor: Option<tokio::task::JoinHandle<anyhow::Result<()>>>,
     dispatcher_supervisor: Option<tokio::task::JoinHandle<anyhow::Result<()>>>,
+    lifecycle_supervisor: Option<tokio::task::JoinHandle<anyhow::Result<()>>>,
 }
 
 impl LiveEngine {
     /// Start one live scheduler at `dp_rank`.
     pub fn start(args: MockEngineArgs, dp_rank: u32) -> anyhow::Result<Self> {
         Self::start_internal(args, dp_rank, LiveEngineOptions::default(), None)
+    }
+
+    /// Start one live scheduler with runtime-owned KV and FPM publishers.
+    pub fn start_with_config(
+        args: MockEngineArgs,
+        dp_rank: u32,
+        config: LiveEngineConfig,
+    ) -> anyhow::Result<Self> {
+        Self::start_internal(
+            args,
+            dp_rank,
+            LiveEngineOptions {
+                kv_event_publishers: config.kv_event_publishers,
+                fpm_publisher: config.fpm_publisher,
+                ..LiveEngineOptions::default()
+            },
+            None,
+        )
     }
 
     pub(crate) fn start_with_options(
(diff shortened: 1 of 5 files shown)


lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,405 @@
...
+        let route = routes
+            .by_id
+            .get(&handoff_id)
+            .map(|entry| Arc::clone(entry.value()));
+        if let Some(route) = route
+            && !route.send(event, &cancel).await
+        {
+            remove_handoff_route(&routes, &route);
+        }
+    }
```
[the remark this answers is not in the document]
connorcarpenter15: Verified against the current scheduler contract. Each engine emits at most one routed lifecycle event per handoff (`SourceHeld` on the source or `DestinationReserved` on the destination), while the per-route buffer holds eight. The shared dispatcher therefore cannot block in the supported handoff flow. Switching to an overflow/drop policy would weaken lifecycle delivery without fixing a reachable failure mode, so I left this unchanged.

lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,405 @@
...
+        ))
+    }
+}
+
+/// Typed scheduler controls for one disaggregated handoff.
+#[derive(Clone)]
+pub struct LiveHandoffControl {
+    engine: LiveEngine,
+    route: Arc<HandoffRoute>,
+}
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in 6a340ddec5. `LiveHandoffControl` now stores a `Weak<LiveEngineInner>`, upgrades it for each command, and returns `live Mocker engine no longer exists` after owner teardown. The new `retained_handoff_control_does_not_keep_engine_alive` regression verifies that a retained control no longer pins scheduler tasks or event routes.

lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,405 @@
...
+        let command_guard = self.route.command_lock.clone().lock_owned().await;
+        self.ensure_generation(false)?;
+        self.engine
+            .submit_prepared(
+                registration,
+                PreparedSubmission::Destination(self.route.handoff_id),
+                Some(command_guard),
+            )
+            .await
+    }
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in 6a340ddec5. Request routes now record destination-aware cancellation after reservation admission, so `LiveRequest` drop/cancel and ID-based cancellation issue `CancelDestination` instead of `CancelRequest`. The regression covers both vLLM and SGLang and verifies that dropping a reserved destination returns scheduler capacity to idle.

lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,405 @@
...
+        };
+        let route = routes
+            .by_id
+            .get(&handoff_id)
+            .map(|entry| Arc::clone(entry.value()));
+        if let Some(route) = route
+            && !route.send(event, &cancel).await
+        {
+            remove_handoff_route(&routes, &route);
+        }
```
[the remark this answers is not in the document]
connorcarpenter15: Verified; this duplicates the dispatcher concern above. Current schedulers emit only one relevant lifecycle event per handoff on a given engine, versus route capacity eight, so a supported handoff cannot fill this channel. I kept reliable delivery instead of adding an unreachable overflow/drop path.

lib/mocker/src/live.rs
```diff
@@ -832,16 +846,23 @@ fn spawn_cancellation(
...
+            RequestCancellation::Request => {
+                cancel_request(
+                    &cancellation_tx,
+                    route.scheduler_id,
+                    abandon_stream,
+                )
+                .await
+            }
+            RequestCancellation::Destination(handoff_id) => {
+                cancel_destination(&command_tx, handoff_id).await
```
PeaBrane: **[P1] Preserve the handoff generation during request cancellation**

Destination cancellation retains only the reusable `HandoffId`. If generation A is canceled, generation B reuses that ID, and A’s `LiveRequest` is subsequently dropped, this sends a raw `CancelDestination` that can cancel B. Please retain the generation-bound route/control on destination requests and cancel under its command lock after verifying current ownership.

lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,414 @@
...
+        engine: &LiveEngineInner,
+        allow_unregistered: bool,
+    ) -> anyhow::Result<()> {
+        match engine.handoff_routes.by_id.get(&self.route.handoff_id) {
+            Some(current) if Arc::ptr_eq(current.value(), &self.route) => Ok(()),
+            Some(_) => bail!(
+                "handoff {:?} control belongs to an earlier registration",
+                self.route.handoff_id
+            ),
+            None if allow_unregistered => Ok(()),
```
PeaBrane: **[P1] Reject stale unregistered controls**

An empty `by_id` entry does not establish that this control is the latest generation. If B’s event receiver has closed, `by_id` can be empty while `last_by_id` identifies B; a retained generation-A control then passes this branch and can operate on B’s scheduler state. Please validate the control’s generation against `last_by_id` before permitting unregistered cleanup.

lib/mocker/src/live.rs
```diff
@@ -411,42 +302,62 @@ impl LiveEngine {
...
+            command_tx: self.inner.command_tx.clone(),
             cancellation_tx: self.inner.cancellation_tx.clone(),
             runtime: self.inner.runtime.clone(),
         };
+        let registration = LiveRequestRegistration {
+            engine: Arc::downgrade(&self.inner),
+            routes: Arc::clone(&self.inner.routes),
+            prepared: Some(PreparedRequest { request, route }),
+        };
+        Ok((registration, live))
```
PeaBrane, line 325: **[P2] Recheck shutdown after publishing the prepared route**

Shutdown can cancel and sweep the maps after the initial cancellation check but before these insertions, allowing this function to return a route that the sweep missed after its dispatcher has exited. Please recheck cancellation after publishing the route and synchronously remove and close it on failure. `register_handoff()` needs the equivalent post-insertion protection.

PeaBrane (dismissed):
The typed handoff boundary introduces generation ownership, but two cleanup paths bypass that guarantee and allow state from an older handoff generation to affect a later reuse of the same `HandoffId`. Prepared request and handoff registration can also race the shutdown sweep. Requesting changes on those lifecycle boundaries. The stack-level architectural context is covered on #12381.

PeaBrane (commented):
Leaving this as a neutral comment review. The three inline threads are the intended feedback: two cleanup paths bypass the typed handoff generation guarantee, and prepared request/handoff registration can race the shutdown sweep. The stack-level architectural context is covered on #12381.

lib/mocker/src/live.rs
connorcarpenter15: Fixed in 1f30c077e7. Destination request cancellation now retains a generation-bound handle, takes that generation command lock, validates latest ownership, and treats stale late cleanup as a safe no-op. The new late_request_cancellation_cannot_cancel_a_reused_handoff_id regression covers both vLLM and SGLang; the full dynamo-mocker suite passes (694 tests).

lib/mocker/src/live/handoff.rs
connorcarpenter15: Fixed in 1f30c077e7. Unregistered cleanup is now accepted only when the retained route generation matches last_by_id. The stale-control regression now drops the replacement event stream before the old control attempts cleanup, covering the previously unchecked empty-by_id branch.

lib/mocker/src/live.rs
connorcarpenter15: Fixed in 1f30c077e7. prepare_request and register_handoff now recheck engine cancellation after publishing all route-map entries; on shutdown they synchronously close and pointer-conditionally remove the exact route before returning an error. Rustfmt, Clippy, and all 694 dynamo-mocker tests pass.

lib/mocker/src/live.rs
```diff
@@ -317,24 +183,33 @@ impl LiveEngine {
...
             options.kv_event_publishers,
             Some(cancel.clone()),
             options.fpm_publisher,
         );
         let command_tx = scheduler.command_sender();
         let cancellation_tx = scheduler.cancellation_sender();
         let metrics_rx = scheduler.metrics_receiver();
+        let lifecycle_rx = scheduler
+            .take_lifecycle_receiver()
+            .expect("new live scheduler must expose one lifecycle receiver");
```
jthomson04, line 206: **[P3] Propagate the missing lifecycle lane instead of panicking.** start_internal already returns Result, so this expect should become a contextual error. A panic here bypasses the normal startup error path and loses useful caller context.

lib/mocker/src/live/handoff.rs
```diff
@@ -0,0 +1,461 @@
...
+            handoff_id,
+            generation: Uuid::new_v4(),
+            routes,
+            event_tx: Mutex::new(Some(event_tx)),
+            command_lock: Arc::new(tokio::sync::Mutex::new(())),
+        }
+    }
+
+    async fn send(&self, event: LiveHandoffEvent, cancel: &CancellationToken) -> bool {
+        let event_tx = self.event_tx.lock().unwrap().as_ref().cloned();
```
jthomson04, line 61: **[P3] Avoid poisonable mutex unwraps in this runtime boundary.** This mutex is only protecting a sender clone or take; parking_lot::Mutex would remove the poison and production-panic path. The extracted request route has the same pattern for its output sender.

lib/mocker/src/live.rs
connorcarpenter15: Thanks for flagging this. I traced the factory path: both supported backends are created through spawn_live_scheduler with lifecycle_rx initialized to Some, and start_internal owns the fresh handle and is the first consumer. None therefore represents an internal constructor invariant violation rather than a recoverable startup condition, so I am retaining the expect and its fail-fast diagnostic.

lib/mocker/src/live/handoff.rs
connorcarpenter15: I checked both sender locks and am leaving them unchanged. Their guards only clone, take, or synchronously try_send a Tokio sender; they contain no await, callback, or application code that can plausibly panic while holding the lock. Switching dynamo-mocker to parking_lot would add a direct dependency without removing a reachable failure path, while poisoning preserves fail-fast behavior after an unexpected internal panic.

PeaBrane (approved):
✅

PR #12381, opened 2026-07-29
refactor(llm): migrate Mocker execution to LiveEngine

connorcarpenter15:
## Overview:

Completes DIS-2481 by migrating the `lib/llm` Mocker adapter to the typed `LiveEngine` API introduced in #12380 and deleting its duplicated scheduler bridge. The supported `make_mocker_engine` contract and existing scheduling, output, metrics, KV/FPM, replay, and bootstrap behavior are preserved.

## Details:

- Collapse the legacy Mocker wrappers into a private annotated execution context backed by one `LiveEngine` per DP rank.
- Start each rank with configured KV/FPM publishers while preserving bootstrap timing, metrics publication, capacity limits, relay lifetimes, and shutdown ordering.
- Route aggregated requests through `LiveEngine::submit` and disaggregated requests through prepared registrations plus typed handoff controls/events.
- Replace raw scheduler-channel handoff plumbing with a private semantic control boundary and boxed event stream.
- Propagate startup failures to waiting requests and clean up prepared bootstrap resources and already-started engines on partial initialization failure.
- Bound adapter output buffering and cancel scheduler state when the context stops or the response stream is dropped.
- Delete duplicated request routing, scheduler construction, output/lifecycle demultiplexing, command sender cells, task ownership, and the stale DIS-2478 TODO.
- This PR is stacked on #12380.

### Validation

- `cargo test -p dynamo-mocker` — 691 passed
- `cargo test -p dynamo-llm --lib mocker` — 18 passed
- CPU Mocker E2E soak: `python -m pytest -vv -s tests/router/test_router_e2e_with_mockers.py::test_mocker_router_soak` — 6 variants passed and 6,144 requests completed successfully (1,024 each across KV, round-robin, and random routing over NATS and TCP) in 103.25s; no error-level logs or panics
- `git diff --check`
- Scoped legacy-symbol and `TODO(DIS-2478)` search is clean

- `lib/llm/src/mocker.rs` for startup, per-rank engine ownership, request generation, response cancellation, metrics, and shutdown.
- `lib/llm/src/mocker/handoff.rs` for the semantic adapter over typed handoff control/events.
- `lib/llm/src/mocker/handoff_tests.rs` for acknowledgment ordering, cleanup, cancellation, timing, KV publication, and engine coverage.

## Related Issues

- Tracked in Linear as DIS-2481

### DIFF
--- a/lib/llm/src/mocker.rs
+++ b/lib/llm/src/mocker.rs
@@ -1296,11 +1296,6 @@ mod tests {
         let engine = MockerExecutionContext::new(args);
 
         assert!(engine.prepare_bootstrap().await.is_err());
-        assert!(engine.engines.get().is_none());
-        assert!(engine.handoff_session_permits.get().is_none());
-        assert!(engine._relay_publishers.get().is_none());
-        assert!(!engine.handoff_shutdown.is_cancelled());
-        assert!(!engine.metrics_shutdown.is_cancelled());
 
         drop(occupied);
         let prepared = engine
--- a/lib/llm/src/mocker.rs
+++ b/lib/llm/src/mocker.rs
@@ -39,15 +39,15 @@ use dynamo_runtime::metrics::MetricsHierarchy;
 use dynamo_runtime::protocols::annotated::Annotated;
 use dynamo_runtime::{
     component::Endpoint,
-    engine::AsyncEngineContextProvider,
+    engine::{AsyncEngineContext, AsyncEngineContextProvider},
     pipeline::{AsyncEngine, Error, ManyOut, ResponseStream, SingleIn, async_trait},
     traits::DistributedRuntimeProvider,
 };
 use futures::StreamExt;
 use rand::Rng;
 use serde::Deserialize;
-use tokio::sync::{Notify, OnceCell, Semaphore, mpsc, oneshot};
-use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio::sync::{OnceCell, Semaphore, mpsc, oneshot, watch};
+use tokio_stream::wrappers::ReceiverStream;
 use tokio_util::sync::CancellationToken;
 use tokio_util::task::TaskTracker;
 use uuid::Uuid;
@@ -59,6 +59,7 @@ use self::handoff::{
 use self::metrics::NativeMockerMetrics;
 
 pub const MOCKER_COMPONENT: &str = "mocker";
+const RESPONSE_STREAM_CAPACITY: usize = 8;
 
 #[derive(Debug, Clone, Deserialize)]
 struct ResponseReplayTraceRow {
@@ -218,10 +219,26 @@ fn no_bootstrap_handoff_delay(
     Some(Duration::from_secs_f64(delay_ms.max(0.0) / 1000.0))
 }
 
+async fn send_response(
+    stream_tx: &mpsc::Sender<LLMEngineOutput>,
+    output: LLMEngineOutput,
+    context: &Arc<dyn AsyncEngineContext>,
+) -> bool {
+    tokio::select! {
+        biased;
+        _ = stream_tx.closed() => false,
+        _ = context.stopped() => {
+            let _ = stream_tx.try_send(LLMEngineOutput::cancelled());
+            false
+        }
+        result = stream_tx.send(output) => result.is_ok(),
+    }
+}
+
 struct MockerExecutionContext {
     engines: OnceCell<Vec<LiveEngine>>,
     handoff_session_permits: OnceCell<Vec<Arc<Semaphore>>>,
-    engines_ready: Notify,
+    startup_state: watch::Sender<StartupState>,
     engine_args: MockEngineArgs,
     response_replay_table: Option<ResponseReplayTable>,
     unset_dp_rank_counter: AtomicU32,
@@ -242,10 +259,36 @@ struct MockerExecutionContext {
 struct PreparedBootstrap {
     server: Arc<BootstrapServer>,
     max_sessions: usize,
+    cancel: Option<CancellationToken>,
+}
+
+impl PreparedBootstrap {
+    async fn shutdown(mut self) {
+        if let Some(cancel) = self.cancel.take() {
+            cancel.cancel();
+        }
+        self.server.wait_closed().await;
+    }
+}
+
+impl Drop for PreparedBootstrap {
+    fn drop(&mut self) {
+        if let Some(cancel) = self.cancel.take() {
+            cancel.cancel();
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+enum StartupState {
+    Starting,
+    Ready,
+    Failed(Arc<str>),
 }
 
 impl MockerExecutionContext {
     fn new(engine_args: MockEngineArgs) -> Self {
+        let (startup_state, _) = watch::channel(StartupState::Starting);
         let native_metrics = NativeMockerMetrics::new(engine_args.engine_type, engine_args.dp_size)
             .expect("mocker native metrics collectors should be valid");
         let response_replay_table = engine_args
@@ -268,7 +311,7 @@ impl MockerExecutionContext {
         Self {
             engines: OnceCell::new(),
             handoff_session_permits: OnceCell::new(),
-            engines_ready: Notify::new(),
+            startup_state,
             engine_args,
             response_replay_table,
             unset_dp_rank_counter: AtomicU32::new(0),
@@ -304,9 +347,10 @@ impl MockerExecutionContext {
             .effective_handoff_capacity()
             .checked_mul(self.engine_args.dp_size as usize)
             .expect("mocker handoff session limit overflow");
+        let cancel = self.handoff_shutdown.child_token();
         let server = BootstrapServer::start(
             port,
-            self.handoff_shutdown.clone(),
+            cancel.clone(),
             BootstrapServerConfig {
                 max_pending_connections: max_sessions,
                 ..BootstrapServerConfig::default()
@@ -316,14 +360,15 @@ impl MockerExecutionContext {
         Ok(Some(PreparedBootstrap {
             server,
             max_sessions,
+            cancel: Some(cancel),
         }))
     }
 
-    fn commit_bootstrap(&self, prepared: PreparedBootstrap) {
-        let PreparedBootstrap {
-            server,
-            max_sessions,
-        } = prepared;
+    fn commit_bootstrap(&self, mut prepared: PreparedBootstrap) {
+        prepared.cancel.take();
+        let server = Arc::clone(&prepared.server);
+        let max_sessions = prepared.max_sessions;
+        drop(prepared);
         let incoming_rx = server
             .take_incoming_receiver()
             .expect("new bootstrap server must own its incoming receiver");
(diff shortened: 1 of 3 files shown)


lib/llm/src/mocker.rs
```diff
@@ -826,11 +857,7 @@ impl AsyncEngine<SingleIn<PreprocessedRequest>, ManyOut<LLMEngineOutput>, Error>
             ..Default::default()
         };
 
-        let (request_tx, mut request_rx) = mpsc::unbounded_channel::<OutputSignal>();
-        self.active_requests.insert(request_uuid, request_tx);
-
-        // Create a simple channel for the stream
-        let (stream_tx, stream_rx) = mpsc::unbounded_channel::<LLMEngineOutput>();
+        let (stream_tx, stream_rx) = mpsc::channel::<LLMEngineOutput>(RESPONSE_STREAM_CAPACITY);
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in 0d18b51e72. The adapter again uses a lossless unbounded outer response stream while `LiveEngine` retains its bounded internal route. This keeps the pump draining `LiveRequest` and prevents downstream lag from triggering internal overflow cancellation. `unread_response_completes_without_overflow_cancellation` holds a 32-token response unread until generation finishes, then verifies every token and the terminal finish.

lib/llm/src/mocker.rs
```diff
@@ -221,42 +219,76 @@ fn no_bootstrap_handoff_delay(
...
+) -> bool {
+    tokio::select! {
+        biased;
+        _ = stream_tx.closed() => false,
+        _ = context.stopped() => {
+            let _ = stream_tx.try_send(LLMEngineOutput::cancelled());
+            false
+        }
+        result = stream_tx.send(output) => result.is_ok(),
+    }
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in 0d18b51e72 with the response-stream correction. Cancellation terminals now use reliable sends on the unbounded adapter stream, so a full bounded queue can no longer drop `FinishReason::Cancelled`. Receiver-drop detection remains intact through `UnboundedSender::closed`.

lib/llm/src/mocker/handoff_tests.rs
```diff
@@ -626,1457 +335,288 @@ async fn source_held_waits_for_submit_outcome_before_progressing() {
...
+async fn live_handoff_preserves_timing_kv_and_cleanup_for_supported_engines() {
+    for engine_type in [EngineType::Vllm, EngineType::Sglang] {
+        for transfer_timing_mode in [
+            KvTransferTimingMode::FullPrompt,
+            KvTransferTimingMode::DestinationMissing,
+        ] {
+            let (source_engine, mut source_kv) =
+                start_live_engine(engine_type, WorkerType::Prefill, transfer_timing_mode);
+            let (destination_engine, mut destination_kv) =
+                start_live_engine(engine_type, WorkerType::Decode, transfer_timing_mode);
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in 0d18b51e72. The typed `LiveEngine` fixture now builds and validates the complete normalized conformance report, then compares it exactly with offline replay for vLLM and SGLang, both transfer-timing modes, and both participant arrival orders. It covers lifecycle ordering, token/completion counts, KV publication timing, repeated activation hashes, capacity cleanup, permits, and route reuse without restoring raw scheduler-channel plumbing.

lib/llm/src/mocker/handoff.rs
```diff
@@ -59,99 +57,129 @@ pub(crate) fn order_for_engine(engine_type: EngineType) -> Result<HandoffOrder>
...
+            HandoffControlAction::CancelSource => {
+                self.control.cancel_source().await?;
+                Ok(HandoffActionOutcome::Applied)
+            }
+            HandoffControlAction::ActivateDestination => {
+                self.control.activate_destination().await?;
+                Ok(HandoffActionOutcome::Applied)
+            }
+            HandoffControlAction::CancelDestination => {
+                self.control.cancel_destination().await?;
```
PeaBrane, line 134: **[P1] Abandon an unsubmitted registration before destination cleanup**

`CancelDestination` can execute before `ReserveDestination` consumes this registration. The outer cleanup then receives `Noop` and calls `LiveRequest::cancel()`, which waits for the route to leave `Submitting`, while this adapter still owns the registration whose drop would release that state. Please take and drop any remaining registration under this mutex before awaiting `control.cancel_destination()`.

lib/llm/src/mocker/handoff_tests.rs
```diff
@@ -626,1457 +340,424 @@ async fn source_held_waits_for_submit_outcome_before_progressing() {
...
+    .unwrap_or(false);
+    drop(probe);
+    wait_for_idle(engine).await;
+    completed && engine.active_request_count() == 0
+}
 
-    transport_shutdown.cancel();
-    server.wait_closed().await;
-    scheduler_shutdown.cancel();
+#[tokio::test]
```
PeaBrane, line 848: **[P2] Restore abnormal handoff coverage**

This matrix gives strong successful-handoff coverage, but it replaces tests for active cancellation, lifecycle loss, participant disconnects, lost acknowledgements, deadlines, active shutdown, and rendezvous retirement. Please restore representative cases through the typed boundary and assert cleanup of held KV, destination state, permits, and lifecycle routes.

lib/llm/src/mocker/handoff_tests.rs
```diff
@@ -291,31 +207,22 @@ async fn destination_reservation_ack_precedes_an_early_lifecycle_fact() {
...
-    envelope
-        .reply
-        .send(Ok(SchedulerCommandEffects {
-            result: SchedulerCommandResult::DestinationAccepted { request_id },
-            lifecycle_events: Vec::new(),
-            kv_events: Vec::new(),
-        }))
         .unwrap();
+    assert!(
+        tokio::time::timeout(Duration::from_millis(20), source.recv())
```
PeaBrane, line 242: **[P2] Add a deterministic lifecycle-consumption barrier**

A 20 ms absence assertion can pass merely because the task consuming the injected lifecycle event was not scheduled in time. Please have the fixture acknowledge that the event was consumed before asserting that no outbound fact appears ahead of the command acknowledgement.

PeaBrane (commented):
Reviewing #12380 and #12381 as one stack, the decode-cancellation deadlock here and the generation-bypass cleanup paths in #12380 share one ownership gap: the prepared registration, request stream, typed control, lifecycle route, and scheduler state do not carry one indivisible generation identity.

Cleanup should atomically abandon an unsubmitted registration, while every destination command should retain the generation-bound route and verify ownership before reaching the scheduler.

The lifecycle dispatcher’s `HandoffId`-only routing is related but predates this stack, so I am not counting it as a newly introduced regression. It nevertheless shows that the generation-safety boundary remains incomplete. The invariant should be that a command or event can affect only the registration that created its internal request/generation identity, regardless of logical `HandoffId` reuse.

PeaBrane (commented):
Non-blocking follow-up: the no-bootstrap P/D path is intended to model vLLM, whose handoff order is source-first. Now that the Mocker adapter uses `LiveEngine`’s typed handoff boundary, could we make this two-request compatibility path explicitly exercise that ordering rather than only delaying terminal prefill completion? The modeled sequence should be source submitted and held, destination reserved, transfer delay applied, source released, then destination activated. This would preserve no-bootstrap coverage while making its behavior and cleanup more faithful to vLLM.

lib/llm/src/mocker/handoff.rs
connorcarpenter15: Fixed in f421152fee. CancelSource and CancelDestination now take and drop any unsubmitted registration before awaiting scheduler cleanup, without holding the mutex across the await. destination_cleanup_abandons_an_unsubmitted_registration verifies cancellation completes, capacity returns idle, and both request and handoff IDs are reusable.

lib/llm/src/mocker/handoff_tests.rs
connorcarpenter15: Addressed with focused typed-boundary coverage in f421152fee. The new tests cover pre-reservation cancellation and shutdown after SourceHeld plus DestinationReserved, asserting engine idle state, zero active decode blocks, permit return, route reuse, and successful drain probes. Existing tests retain source cancellation, rendezvous retirement, premature cleanup, and vLLM/SGLang success coverage. The scoped Mocker suite passes (21 tests).

lib/llm/src/mocker/handoff_tests.rs
connorcarpenter15: Fixed in f421152fee. Semantic lifecycle events now carry a per-event consumption acknowledgement, and both destination-reservation and source-held ordering tests await that acknowledgement before asserting no fact precedes the scheduler command acknowledgement. This removes the scheduler-timing false positive.

lib/llm/src/mocker/handoff.rs
```diff
@@ -1471,147 +1462,76 @@ pub(crate) async fn run_destination_session(
...
-            SchedulerCommand::ActivateDestination { handoff_id },
-            SchedulerCommandResultKind::Applied,
-        ),
-        HandoffAction::CancelDestination { handoff_id } => (
-            SchedulerCommand::CancelDestination { handoff_id },
-            SchedulerCommandResultKind::AppliedOrNoop,
-        ),
+) -> std::result::Result<HandoffControlAction, HandoffActionOutcome> {
+    let action = match action.action {
+        HandoffAction::ReserveDestination { .. } => HandoffControlAction::ReserveDestination,
```
jthomson04: **[P2] Validate the action handoff ID before erasing it.** The destination validates handoff IDs on facts, but these patterns discard the ID. The typed control then targets the local route, so a mismatched bootstrap action is applied and acknowledged instead of rejected. Pass the connection handoff ID into this conversion and reject mismatches.

lib/llm/src/mocker.rs
```diff
@@ -221,42 +218,76 @@ fn no_bootstrap_handoff_delay(
...
-    request_senders: OnceCell<Vec<mpsc::UnboundedSender<DirectRequest>>>,
-    command_senders: OnceCell<Vec<mpsc::Sender<SchedulerCommandEnvelope>>>,
-    // TODO(DIS-2478): Store the scheduler cancellation senders and connect the AsyncEngine
-    // context stop/response-stream drop path to targeted cancellation. Until then, lib/mocker
-    // cancellation is not wired end to end through MockEngine.
+async fn send_response(
+    stream_tx: &mpsc::UnboundedSender<LLMEngineOutput>,
+    output: LLMEngineOutput,
+    context: &Arc<dyn AsyncEngineContext>,
+) -> bool {
```
jthomson04, line 225: **[P2] Do not report context cancellation as a closed output receiver.** send_response returns false for both receiver closure and context shutdown, while each false call site logs an error claiming the receiver closed. Return the reason or treat both outcomes as normal cleanup.

lib/llm/src/mocker/handoff.rs
connorcarpenter15: Fixed in 3aca78eb2b. prepare_destination_action now receives the connection handoff ID and rejects reserve, activate, or cancel actions whose embedded ID differs before invoking typed control. destination_rejects_an_action_for_another_handoff verifies a mismatched action receives Failed and never reaches scheduler control. The scoped Mocker suite passes (22 tests).

lib/llm/src/mocker.rs
connorcarpenter15: Fixed in 3aca78eb2b. The three checked send_response call sites now treat false uniformly as normal producer termination and no longer emit a receiver-closed error during context shutdown. Cleanup behavior is unchanged; the misleading error reporting is removed.

PeaBrane (approved):
✅

PR #12387, opened 2026-07-29
fix(ci): stop vCluster Grove admission startup failures

sttts:
## Summary

**STOP THE BLEEDING** for vCluster operator CI:

- keep Dynamo's existing `publishNotReadyAddresses=true` workaround;
- do not add a Grove health-probe dependency in this PR;
- wait for the Grove CRD, then retry a server-side `PodCliqueSet` dry-run for up to 120 seconds; and
- require Grove's defaulting mutation to prove the webhook handled the request.

## Why

vCluster can leave the synchronized Pod `Ready=False` even after the physical operator container and webhook are ready. Dynamo therefore needs:

```yaml
spec:
  publishNotReadyAddresses: true
```

The pinned Grove chart does not expose that Service option. Enabling Grove readiness probes here would introduce the same unroutable-endpoint failure. Instead, this PR leaves the probes disabled and makes admission itself the readiness gate:

```yaml
apiVersion: grove.io/v1alpha1
kind: PodCliqueSet
metadata:
  name: ci-webhook-smoke-grove
spec:
  replicas: 0
  template:
    cliques:
    - name: worker
      spec:
        roleName: worker
        replicas: 1
        minAvailable: 1
        podSpec:
          containers:
          - name: worker
            image: busybox:latest
```

The command uses `kubectl create --dry-run=server -o json` and retries only transient webhook transport failures. It succeeds only when Grove injects `headlessServiceConfig`.

Original failure: [deploy-operator job 90710737439](https://github.com/ai-dynamo/dynamo/actions/runs/30490931478/job/90710737439?pr=12387).

Longer-term Grove health probes remain separate: [ai-dynamo/grove#743](https://github.com/ai-dynamo/grove/pull/743).

## Validation

- `pre-commit run --files .github/actions/setup-dynamo-operator/action.yml deploy/helm/charts/platform/README.md deploy/helm/charts/platform/values.yaml`
- YAML and whitespace checks passed.

### DIFF
--- a/.github/actions/setup-dynamo-operator/action.yml
+++ b/.github/actions/setup-dynamo-operator/action.yml
@@ -438,7 +438,6 @@ runs:
         helm upgrade --install dynamo-platform . --namespace default \
           "${IMAGE_ARGS[@]}" \
           --set dynamo-operator.gpuDiscovery.enabled=false \
-          --set dynamo-operator.webhook.service.publishNotReadyAddresses=true \
           "${CHECKPOINT_HELM_ARGS[@]}" \
           --set global.grove.install=true \
           --set global.kai-scheduler.install=false \
@@ -510,6 +509,10 @@ runs:
           wait --for=condition=established \
           crd/dynamographdeployments.nvidia.com \
           --timeout=120s
+        kubectl --kubeconfig=${{ github.workspace }}/.kubeconfig-vcluster \
+          wait --for=condition=established \
+          crd/podcliquesets.grove.io \
+          --timeout=120s
 
     - name: Verify operator image and admission webhooks
       shell: bash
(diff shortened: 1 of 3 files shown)


dmitry-tokarev-nv (approved):

julienmancuso (approved):

deploy/helm/charts/platform/values.yaml
```diff
@@ -292,6 +292,13 @@ grove:
   tolerations: []
   # -- Affinity for Grove pods
   affinity: {}
+  config:
+    server:
+      healthProbes:
+        # -- Enable Grove's health and readiness probes on port 9444. Grove's /readyz
+        # gates on certificate readiness and its webhook listener, so PodCliqueSet
+        # admission is not attempted before Grove can serve it.
+        enable: true
```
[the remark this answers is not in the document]
sttts: we know

julienmancuso (approved):

.github/actions/setup-dynamo-operator/action.yml
```diff
@@ -629,8 +675,9 @@ runs:
           done
         }
 
-        run_admission_smoke_test v1alpha1 smoke_v1alpha1
-        run_admission_smoke_test v1beta1 smoke_v1beta1
+        run_admission_smoke_test v1alpha1 dynamo smoke_v1alpha1
+        run_admission_smoke_test v1beta1 dynamo smoke_v1beta1
+        run_admission_smoke_test PodCliqueSet grove smoke_grove
```
dmitry-tokarev-nv, line 680: **The retry classifier treats timeouts as fatal, and #12266 documents the exact condition that produces them.**

`action.yml:659-661` is a whitelist, so any error string it does not match returns immediately and fails the whole setup action with no retry:

```
'no endpoints available|connection refused|connection reset by peer|TLS handshake timeout|service .* not found|unexpected EOF|^EOF$'
```

`connection refused` covers the startup window this PR targets. I traced Grove's `main.go` to check: `RegisterControllersAndWebhooks` blocks on `certsReadyCh` before registering handlers, so during that window :9443 simply is not listening. Your reasoning for leaving the probes disabled holds up, and there is no x509/caBundle window to worry about either, since the CA is injected before the listener comes up.

What the list does not cover is the *other* failure mode #12266 records in this same cluster:

```
```

A stale EndpointSlice points the Service at an address that no longer answers, which yields `dial tcp <ip>:9443: i/o timeout` rather than `connection refused`. Add `--request-timeout=15s` plus the webhook's own `timeoutSeconds: 10` and `context deadline exceeded` becomes reachable too. All transient, all currently fatal on the first occurrence.

`TLS handshake timeout` already being in the list suggests this family was on your mind; this just completes it:

```bash
            if ! grep -Eiq \
              'no endpoints available|connection refused|connection reset by peer|TLS handshake timeout|service .* not found|unexpected EOF|^EOF$|i/o timeout|context deadline exceeded|request did not complete within requested timeout' \
              <<<"${output}"; then
```

For a PR whose job is to stop the bleeding, a transient timeout that hard-fails the action is the remaining hole.

dmitry-tokarev-nv (commented):
Re-reviewed after the re-scope to CI-only (I had started on the earlier 3-file version). The narrowed approach is the right call and the smoke test itself is well built. Two things, one inline.

<details>
<summary>Verified against the pinned Grove chart, no action needed</summary>

- `headlessServiceConfig` has **no CRD-level default** in `v0.1.0-alpha.12-rc1`, and `defaultPodCliqueSetTemplateSpec` sets it unconditionally. So the grep really does prove the mutating webhook handled the request, rather than catching something the API server would have synthesized on its own. Good choice of assertion.
- The manifest is schema- and webhook-valid against the pinned CRD: `spec.replicas: 0` passes `ValidateNonnegativeField`, the clique supplies all three required fields (`roleName`, `replicas`, `podSpec`), `minAvailable: 1 <= replicas: 1`, and both PCS webhooks are `sideEffects: None`, so server-side dry-run is permitted rather than rejected.
- Your reason for not enabling the Grove probes holds up in the source, as noted inline.
</details>

### This collides with #12368

@lavanyavijayk's #12368 ("fix(ci): harden Grove webhook readiness in vCluster") is open, touches the same `action.yml`, and takes the opposite approach: enable Grove's `/healthz` and `/readyz` **and** patch `publishNotReadyAddresses: true` onto the Grove Service for short-lived CI only, which answers the objection in this description that "the pinned Grove chart does not expose that Service option."

Both designs are defensible, but they are not composable, and whichever lands second gets a conflict right in the smoke-test region. Neither description references the other. Worth an explicit maintainer decision before either merges. If this one wins on stop-the-bleeding grounds, saying so in the body would let #12368 be closed or rebased down to whatever survives.

.github/actions/setup-dynamo-operator/action.yml
sttts: we will not get a timeout here if no endpoint exists. Why care?

sttts: Becuase there is no health check, the pod will be ready. The only risk is that the webhook is actually not up yet, which leads to connection refused and that the loop will cope with.

sttts: "EndpointSlice points the Service at an address that no longer answers" means that the pods must have been replaced. Yes, this could happen, but it is a different error mode.

PR #12388, opened 2026-07-29
fix(docs): stop comment examples from triggering Fern's rolldown bundling

dagil-nvidia:
## Summary

Docs previews fail intermittently on PRs that touch `docs/**`, with two variants of the same error:

```
rolldown exited with code 127
sh: 1: rolldown: not found
```

```
rolldown exited with code 1
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'/home/runner/.npm/_npx/3f406405b888ccef/node_modules/rolldown/dist/shared/bindingify-input-options-BsDhfYrS.mjs'
```

Both are the same root cause, and it is not an infrastructure problem. **A comment is being read as an import.**

## Root cause

`fern-api@5.80.2` — the version pinned in `docs/fern/fern.config.json` — scans every `.js/.jsx/.ts/.tsx` file under the `experimental.mdx-components` directory with this regex:

```
/(?:^|[^\w.])(?:import|export)\s+(?:[\w*\s{},$]*?from\s+)?["']([^"'\n]+)["']
 |import\(\s*["']([^"'\n]+)["']\s*\)
 |require\(\s*["']([^"'\n]+)["']\s*\)/gm
```

Any specifier that is neither relative (`./`, `../`) nor in its allowlist — `["react","react-dom","@mdx-js/react","next"]` — is treated as a third-party dependency, and Fern shells out to:

```
npx --quiet --yes rolldown@1.1.4 -c <config>
```

once per offending file. That is a registry download on every docs build. When the fetch fails outright you get exit 127; when it leaves a half-written npx cache you get `ERR_MODULE_NOT_FOUND`.

**The scan does not strip comments.** Three docblock usage examples matched it:

| File | Line | Matched text |
|---|---|---|
| `docs/fern/components/RecipeStyles.tsx` | 17 | `import { RecipeStyles } from "@/components/RecipeStyles";` |
| `docs/fern/components/ReferenceStyles.tsx` | 23 | `import { ReferenceStyles } from "@/components/ReferenceStyles";` |
| `docs/fern/components/TerminalDemo.tsx` | 15, 28 | `` `import "asciinema-player"` `` and the `@/components/TerminalDemo` example |

No component actually depends on anything outside Fern's allowlist. `TerminalDemo` deliberately loads asciinema-player from jsDelivr at runtime (line 58) precisely to avoid a bundled import; its only real import is `react`. So all three rolldown fetches were pure waste on top of being a flake.

## What changed

1. Quotes the specifiers in those examples with backticks instead of double quotes, and says why in the surrounding comment so the next author does not revert it. Backticks are already the code-quoting convention in these docblocks.
2. Adds `docs/fern/scripts/check_component_imports.py`, which mirrors `fern-api`'s own regex and allowlist verbatim, wired as a pre-commit hook scoped to `docs/fern/components/**`. It runs locally for authors and in the existing `pre-commit` pre-merge job.

No behavioral change to any component. The diff to the three `.tsx` files is comments only.

## Verification

- `check_component_imports.py` exits 1 naming all three files against `main` as it stands, and 0 after the change — red baseline before green.
- An independently written JS replica of Fern's detector, using the regex and allowlist extracted from the `fern-api@5.80.2` tarball, agrees: three hits before, zero after.
- `fern check`: 0 errors, 1292 warnings both before and after — baseline unchanged.
- `pre-commit run --files <changed>`: all applicable hooks pass, including the new one.

## Follow-up

#12330 adds `BlogStyles.tsx` and `LandingStyles.tsx` following the same `RecipeStyles` docblock convention, so it inherits the trap and would take the build from three rolldown fetches to five. Once this lands, that branch needs the same one-line treatment; the new pre-commit hook will catch it.

If a component ever needs a genuine third-party dependency, this check has to be revisited alongside a `package.json` and `node_modules` for the docs project — which is exactly what Fern's error message asks for. That is documented in the script's docstring.

The upstream bug is that Fern's detector does not skip comments. Worth reporting separately.

## Update: merged main, and the pre-commit failure is fixed

The failing `pre-commit` check was not this PR's code. The
`check-published-styles-selftest` hook was invoking the script with the
filename appended, and that script's argparse takes no positional arguments:

```
check_published_styles.py: error: unrecognized arguments:
docs/fern/scripts/check_published_styles.py
```

`main` had already fixed this by adding `pass_filenames: false` to the hook.
This branch was 52 commits behind and predated that fix, so it kept running
the broken form.

Merging `main` alone did not bring it. This branch inserts
`check-component-imports` directly after that hook, so the auto-merge resolved
the hunk to the branch side and dropped `main`'s added line silently. Restored
explicitly; the hook block is now byte-identical to `main`'s.

Validation: the published-styles self-test passes 7/7, the component-import
scanner reports no unbundleable specifiers, and its self-test passes 16/16.
The original failure was reproduced by invoking the script with a filename
argument, confirming `pass_filenames: false` is what prevents it.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12388" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Documentation**
  * Clarified component import formatting requirements for documentation examples.
  * Updated guidance on runtime loading and bundling behavior for terminal demos.
  * Revised reference component usage snippets for consistent formatting.

* **Chores**
  * Added automated validation to detect unsupported third-party imports in documentation components.
  * The validation now runs automatically before relevant checks and reports files requiring attention.


.pre-commit-config.yaml
```diff
@@ -86,6 +86,12 @@ repos:
       language: system
       files: ^docs/fern/(main\.css|components/CustomFooter\.tsx|scripts/sync_site_css\.py)$
       pass_filenames: false
+    - id: check-component-imports
+      name: Check Fern components trigger no rolldown bundling
+      entry: python3 docs/fern/scripts/check_component_imports.py
+      language: system
+      files: ^docs/fern/(components/.*\.(js|jsx|ts|tsx)|scripts/check_component_imports\.py)$
```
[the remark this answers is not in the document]
dagil-nvidia: Addressed in `050f7c8` — the hook now selects `docs/fern/fern.config.json`, so a version bump forces `check_derivation()` to run and blocks the commit until `DERIVED_FROM_FERN` (and, when the upstream regex or allowlist changed, `SPECIFIER` and `ALLOWLIST`) are re-derived from the new Fern release.

Regex before:
```
^docs/fern/(components/.*\.(js|jsx|ts|tsx)|scripts/check_component_imports\.py)$
```

Regex after:
```
^docs/fern/(components/.*\.(js|jsx|ts|tsx)|scripts/check_component_imports\.py|fern\.config\.json)$
```

Validated: `docs/fern/fern.config.json` now matches; unrelated paths (`components/README.md`, `fern.config.jsonx`) still do not. Self-test 16/16, full scan exit 0, and `pre-commit run check-component-imports --files docs/fern/fern.config.json` now selects the file and passes.

docs/fern/scripts/check_component_imports.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
dagil-nvidia: Half of this was already fixed, half was real. Fixed in 2bc794c.

The copyright range was addressed in 4955d76 on this branch, before the thread was filed. The header has read `2025-2026` since then, so "uses `2026` alone" no longer describes the file.

The Apache block was still outstanding, and you are right about it. `documentation-style-guide.md` is explicit that code and config files -- `.py`, `.sh`, `.yaml`, `Dockerfile` -- take the full Apache-2.0 header block, not just the two SPDX lines. The sibling `docs/fern/scripts/sync_site_css.py` on main has the same gap, but an existing gap is not licence to add another, so the new script now carries the full block.

One unrelated defect surfaced while validating this, worth naming since it lived in the same file. `REPO_ROOT` was resolved through a slice of `Path.parents`, and `PurePath.parents` only became sliceable in 3.10. The hook is `language: system`, so it runs under whichever `python3` is on `PATH`, and under 3.9 both hooks died in a `TypeError` at import before reaching a single check -- a gate whose entire purpose is to fail loudly instead failing as a traceback. `.parent.parent` is equivalent for every path at or below the repo root, never raises, and needs no minimum version. That went in on the same commit.

Validation: `pre-commit run --files docs/fern/scripts/check_component_imports.py` is green on both hooks under the 3.9 interpreter pre-commit resolves here, where it previously raised `TypeError` twice. The self-test is 16/16 and the tree scan exits 0.

docs/fern/scripts/check_component_imports.py
```diff
@@ -0,0 +1,247 @@
...
+`export const KIND = "all"`, which is why the real one insists on `from` or a
+directly quoted specifier.
+
+Being a transcription, it is only correct for one release. DERIVED_FROM_FERN
+records which, and the scan refuses to report a clean tree once
+docs/fern/fern.config.json moves past it — a stale pattern fails closed but
+silently, which is the worst outcome for a gate. To re-derive after a bump:
+
+    npm pack fern-api@<version> && tar xzf fern-api-<version>.tgz
+    grep -o 'TZu=\\[[^]]*\\]' package/cli.cjs     # the allowlist
```
dmitry-tokarev-nv: **Blocking — the documented recovery procedure does not work today, and this gate blocks the whole repo.**

I ran both greps verbatim against the current release, `fern-api@5.92.0`:

    $ grep -o 'TZu=\[[^]]*\]' package/cli.cjs   -> no match
    $ grep -o 'SGm="[^"]*"'   package/cli.cjs   -> no match

`TZu` is now `Vtl`, and `SGm` is gone entirely (the docstring also asks for its value, which the script has no field to store). These are minifier-assigned names; they churn on essentially every release. The repo pins 5.80.2 and npm latest is already 5.92.0.

Why that is blocking rather than a stale comment: the `pre-commit` job (`pre-merge.yml:57`) has no `needs: changed-files` and no `if:`, and `pre-commit/action` defaults to `--all-files`, so `fern.config.json` is always in scope — and `pre-commit` is a hard dep of `pre-merge-status-check`. The day a routine Fern bump merges, every open PR goes red, Rust-only ones included, and the only documented way out returns nothing. The path of least resistance under that pressure is to bump `DERIVED_FROM_FERN` without re-deriving anything, which converts this into a silent no-op — precisely the "fails closed but silently ... the worst outcome for a gate" that line 40 warns about.

Value-anchored greps survive renaming where symbol-anchored ones do not:

    $ grep -o '.\{6\}=\["react","react-dom","@mdx-js/react","next"\]' package/cli.cjs
    $/,Vtl=["react","react-dom","@mdx-js/react","next"]

Suggest anchoring the recipe on the literal allowlist array and the literal `/\.(js|jsx|ts|tsx)$/`, and gating this staleness check on the docs filter so a Fern bump cannot red unrelated PRs.

docs/fern/scripts/check_component_imports.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        print(USAGE, file=sys.stderr)
+        return 2
+    if args == ["--test"]:
+        return run_tests()
+    stale = check_derivation()
+    if stale:
+        return stale
+    failures = 0
+    for path in sorted(COMPONENTS.rglob("*")):
```
dmitry-tokarev-nv, line 277: **Blocking — fails open.** `COMPONENTS.rglob("*")` on a missing or renamed directory yields nothing and raises nothing, so `failures` stays 0 and `main()` returns 0. The scanner reports a clean tree it never opened.

    $ mv docs/fern/components docs/fern/components_RENAMED
    $ python3 docs/fern/scripts/check_component_imports.py; echo $?
    0

The two halves of this PR disagree about this exact case. The new CI step guards it explicitly:

    if [ "$COUNT" -eq 0 ]; then
      echo "::error::No component sources found under docs/fern/components"; exit 1
    fi

and `check_derivation()` guards the analogous missing-`fern.config.json` case with a bespoke "the script has moved" message. Only the components directory is left unguarded — so renaming it, or repointing `docs.yml`'s `experimental.mdx-components`, leaves this hook green forever while every new third-party import goes unchecked.

`if not COMPONENTS.is_dir(): return 1`, plus a count of files actually scanned, closes it — and would give `--test` something meaningful to assert about `main()`.

.github/workflows/pre-merge.yml
```diff
@@ -82,6 +82,32 @@ jobs:
...
+            \( -name '*.tsx' -o -name '*.ts' -o -name '*.jsx' -o -name '*.js' \) "$@"
+        }
+        COUNT=$(find_components | wc -l)
+        if [ "$COUNT" -eq 0 ]; then
+          echo "::error::No component sources found under docs/fern/components"
+          exit 1
+        fi
+        echo "Parsing $COUNT component sources"
+        find_components -print0 \
+          | xargs -0 npx --yes esbuild@0.24.0 --outdir="$(mktemp -d)"
```
dmitry-tokarev-nv: **Blocking — a swallowed failure on this line, plus a reliability concern about the step.**

1. `--outdir="$(mktemp -d)"` is a word on the `xargs` command line, not an assignment, so `set -e` never sees the command substitution's exit status. If `mktemp` fails this expands to `--outdir=` and esbuild runs anyway (reproduced locally). Hoisting it to `OUTDIR=$(mktemp -d)` on its own line makes the failure fatal as intended.

2. `npx --yes esbuild@0.24.0` re-fetches 4.35 MB from the registry on every run: no lockfile, no integrity pin, and no cache (`actions/setup-node` is used here with only `node-version`; there is no `cache:` key anywhere in this file). `fern-check` is a hard dep of `pre-merge-status-check`, so a registry 5xx reds the gate on a PR that only touched an `.mdx` page.

3. The step rides the repo-wide `docs` filter (`docs/**`, `**/*.md`, `.agents/**`, ...). Measured on `main` since 2026-06-01: 458 of 1172 commits match that filter, but only 21 touched a `docs/fern/components/*.{ts,tsx,js,jsx}` file — so roughly 95% of runs do no useful work. A narrower filter for just this step would pay for itself.

Minor, but it will bite later: all entry points share one `--outdir` keyed by basename, so a future `Foo.ts` + `Foo.tsx` pair aborts the whole invocation with esbuild's "Two output files share the same path" — a red gate whose message has nothing to do with component syntax. `--outbase=.` avoids it.

Also worth knowing: `.github/workflows/**` is not in the `docs` filter (`.github/filters.yaml:26-42` lists only `fern-docs.yml`), so a follow-up PR touching only this step will not run `fern-check`, and the change would merge without ever being exercised.

docs/fern/scripts/check_component_imports.py
```diff
@@ -0,0 +1,247 @@
...
+        if not path.is_file() or path.suffix not in SUFFIXES:
+            continue
+        found = offenders(path.read_text())
```
dmitry-tokarev-nv: **Blocking — crashes on the very files this PR edits, under a non-UTF-8 locale.** `path.read_text()` has no `encoding=`, so decoding follows the locale, while Fern always decodes UTF-8.

    $ LC_ALL=C LANG=C python3 -X utf8=0 ...
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 ...

28 of the 33 component `.ts`/`.tsx` files contain non-ASCII, including all three this PR touches — the em dashes in the reworded comments are what trip it. The failure surfaces as a pathlib traceback rather than the gate's carefully written message. Under a latin-1-family locale it does not crash at all; it silently scans mojibake. Windows contributors get cp1252 by default.

Line 103 is the worse instance: `UnicodeDecodeError` subclasses `ValueError`, so `pinned_fern_version()` catches it and misreports a perfectly good file as "could not read a version".

Four sibling scripts in this directory already pass `encoding="utf-8"`.

docs/fern/scripts/check_component_imports.py
```diff
@@ -0,0 +1,247 @@
...
+
+# Verbatim from fern-api's cli.cjs. re.ASCII is not cosmetic: JavaScript's \w is
+# ASCII-only, Python's is Unicode-aware, so without it a Unicode letter sitting
+# directly against `import` is a word character here but not there — Fern would
+# bundle the file and this check would stay silent.
+SPECIFIER = re.compile(
+    r"""(?:^|[^\w.])(?:import|export)\s+(?:[\w*\s{},$]*?from\s+)?["']([^"'\n]+)["']"""
+    r"""|import\(\s*["']([^"'\n]+)["']\s*\)"""
+    r"""|require\(\s*["']([^"'\n]+)["']\s*\)""",
+    re.MULTILINE | re.ASCII,
```
dmitry-tokarev-nv: **Blocking — a Unicode space defeats the scanner.** `re.ASCII` is applied to the whole pattern. It correctly narrows `\w`, as the comment above explains, but it also narrows `\s` — and JavaScript's `\s` *is* Unicode-aware, so the two diverge in the opposite direction.

Against this exact file at HEAD:

    offenders('import { X } from "@scope/pkg";')        -> ['@scope/pkg']   # U+0020 space
    offenders('import { X } from "@scope/pkg";')   -> []               # U+00A0 space

ECMA-262 `\s` matches U+00A0, U+2000–200A, U+2028/29, U+3000 and U+FEFF; Python's `\s` under `re.ASCII` is only `[ \t\n\r\f\v]`. So Fern's regex matches, Fern shells out to rolldown, and the preview fails with the same `ERR_MODULE_NOT_FOUND` this PR exists to prevent — while both new gates stay green, since NBSP is valid ECMAScript whitespace and esbuild compiles it happily too.

NBSP is also the most likely stray character to arrive from the copy-out-of-the-rendered-README workflow this PR institutes, which makes it a live risk rather than a theoretical one.

Spelling `[^\w.]` as `[^A-Za-z0-9_.]` and dropping the global `re.ASCII` flag keeps the `\w` fidelity without narrowing `\s`.

docs/fern/components/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+### TerminalDemo
+
+Props are documented in the component's own header.
+
+```mdx
+import { TerminalDemo } from "@/components/TerminalDemo";
+
+<TerminalDemo
+  src="/dynamo/assets/dynamo-demo.cast"
```
dmitry-tokarev-nv: **Blocking — the canonical example ships the exact string the repo's asset gate exists to reject.**

`check_asset_paths.py`'s docstring names this verbatim as its anti-pattern:

    src="/dynamo/assets/dynamo-demo.cast"          in TSX
    ... is rewritten by nothing and reaches the browser verbatim, where it 404s.

Three compounding problems:

- That gate never fires here: `DEFAULT_GLOBS` covers `components/**/*.tsx` and `components/**/*.ts`, but not `components/**/*.md`.
- `dynamo-demo.cast` does not exist. The only `.cast` in the tree is `assets/hero-demo-25.cast`.
- The correct form was deleted from `TerminalDemo.tsx` in this same diff — `src="../../assets/dynamo-demo.cast"   // relative, so Fern rewrites it` — which was the only place in the repo stating the relative-path constraint. The rewritten USAGE block does not mention `src` at all, and the `TerminalDemoProps` TSDoc says only "a same-origin asset path (recommended) or an https URL", never "relative".

Since this README is now the declared source of truth for usage, the first person to wire up `TerminalDemo` copy-pastes a 404 — the same shape as the #12373 regression that blanked the Home hero mark.

Suggest restoring the relative form both here and in the `TerminalDemo.tsx` header, and adding `components/**/*.md` to `DEFAULT_GLOBS` so this file is actually covered.

dmitry-tokarev-nv (commented):
Review of the new component-import gate — six blocking findings inline, each reproduced against head `3629cdf`.

The direction is right: `fern check` genuinely does not validate component syntax, and transcribing Fern's scanner is a reasonable way to close that. The findings are about the gate's own failure modes — two ways it silently passes files Fern will reject, one way it reds the entire repo, and one crash path.

Non-blocking items I left out of the inline comments, happy to file separately: `offenders()` flags quoted filenames in ordinary prose (`' * We import the styles from "main.css".'` -> `['main.css']`), and the one CASES entry meant to pin prose safety is quote-free so it can never catch that; `--test` exercises only `offenders()`, leaving `main()`, `check_derivation()` and file discovery uncovered; `{js,jsx,ts,tsx}` and `components` are hand-copied across three files while `docs.yml`'s `experimental.mdx-components` (the actual source of truth) is read by none of them; `LandingStyles.tsx` and `BlogStyles.tsx` still carry backtick-quoted docblock examples pointing at a `RecipeStyles.tsx` passage this diff replaced; and a few README specifics (`welcome.mdx` / `community/README.mdx` do not exist; the "no quoted non-relative specifier" rule is contradicted by the react allowlist).

harryskim (approved):
LGTM

dagil-nvidia:
@dmitry-tokarev-nv all six are addressed in `04391df`. Each was reproduced before and after, against fern-api 5.92.0 where you ran your greps.

**Unicode whitespace.** You were right that `re.ASCII` is not selective. Verified across the ECMA-262 set: the old pattern caught only U+0020 and missed U+00A0, U+2009, U+2028 and U+3000. Both character classes now spell out ASCII explicitly and the flag is gone.

Your list turned up one more. Python's `\s` omits **U+FEFF**, which ECMA-262 includes, so neither the old pattern nor the straightforward fix caught it. It is matched by hand now. All six code points verified, with the relative and allowlisted cases still clean.

**Fails open.** Reproduced exactly: renaming `components/` gave exit 0. It now fails closed, and separately fails when the directory exists but holds no component sources, since `docs.yml` still points Fern at it either way. The scan prints a file count so a silent pass is visible.

**Locale decoding.** Both reads pin `encoding="utf-8"`. Your point about line 103 was the sharper one: `UnicodeDecodeError` subclasses `ValueError`, so a good file was reported as "could not read a version". `LC_ALL=C` no longer traces back.

**The re-derivation recipe.** Confirmed broken: `TZu=` and `SGm=` both return nothing on 5.92.0. The recipe is now value-anchored and I ran each command verbatim as a reader would copy it, so the docstring is tested rather than asserted.

Worth recording from that exercise: at 5.92.0, twelve minor releases past the pin, the allowlist, the suffix list and the regex are all unchanged, and the regex is byte-identical to `SPECIFIER`. Only the binding name moved, `TZu` to `Vtl` to `ZJm`. So a Fern bump is normally a verification, not a rewrite, which is the argument for keeping the gate fail-closed now that recovery is two minutes instead of a dead end.

**The workflow.** `OUTDIR=$(mktemp -d)` is hoisted; I reproduced both forms, and the inline one does silently continue with `--outdir=`. `--outbase=.` added for the `Foo.ts` plus `Foo.tsx` collision.

On the filter: the step now rides a narrow `fern_components` filter rather than the repo-wide `docs` one, wired through the changed-files action and documented in FILTERS.md. Your 458-of-1172 measurement is what made the case.

**The README example.** All three of your points were correct, including that `dynamo-demo.cast` does not exist. The example uses the relative form with the real asset, `components/**/*.md` is in `DEFAULT_GLOBS` so the gate actually covers this file, and the relative-path constraint is restored to `TerminalDemo`'s `src` TSDoc, which this PR had removed as the only place stating it.

One I did **not** fix, flagging rather than omitting: the esbuild step still fetches 4.35 MB via `npx` with no lockfile, integrity pin, or cache, so a registry 5xx can still red a docs PR. Narrowing the filter reduces how often that exposure is taken but does not remove it. Happy to add a cached, pinned install here or take it as a follow-up, whichever you prefer.

Validation: import scanner 33 sources clean, self-test 16/16, asset gate 392 files clean, ruff clean, both workflow YAML files parse.

dagil-nvidia:
@dmitry-tokarev-nv the `npx` point is fixed too, in `fcb7e8d` — that was the one I had left as a follow-up.

esbuild is now installed with `--no-save` and invoked from `node_modules/.bin`, with the npm store cached and keyed on the pinned version. A warm run needs no network, and a version bump misses the cache deliberately rather than serving a stale tarball for a new pin. `node_modules/` and `package-lock.json` are already gitignored at any depth, so the tree stays clean and no `package.json` appears beside `fern/`.

That closes all six of your findings plus the reliability concern. Ready for another look.

docs/fern/scripts/check_component_imports.py
dmitry-tokarev-nv: Verified on `fcb7e8d`: renaming `components/` now exits 1 with the repointing message, and the empty-directory case you added on top of it exits 1 as well — that one was not in my report. A normal run prints `checked 33 component source(s)`, so a silent pass is visible. Resolving.

.github/workflows/pre-merge.yml
dmitry-tokarev-nv: Verified on `fcb7e8d`: `OUTDIR` is hoisted, `--outbase=.` is in, and `npm install --no-save` invoked from `node_modules/.bin` with the store cached on the pinned version replaces the `npx` fetch.

The `fern_components` filter also lists `.github/workflows/pre-merge.yml`, so a follow-up touching only this step now exercises it — that closes the last part of this comment too. Resolving.

docs/fern/scripts/check_component_imports.py
dmitry-tokarev-nv: Verified on `fcb7e8d`: both reads pin `encoding="utf-8"`, and `LC_ALL=C LANG=C python3 -X utf8=0` now runs clean over all 33 real components and exits 0 instead of raising `UnicodeDecodeError`. Resolving.

docs/fern/scripts/check_component_imports.py
dmitry-tokarev-nv: Verified on `fcb7e8d`, and the fix is broader than the six code points in your note. I ran the full ECMA-262 WhiteSpace plus LineTerminator set — 23 code points, including U+1680, U+202F, U+205F, U+2029, U+000B, U+000C and all of U+2000 to U+200A — and none are missed. The relative, allowlisted and backtick cases stay clean.

Good catch on U+FEFF: Python's `\s` omits it, and I had missed that in my own report. Resolving.

docs/fern/components/README.md
dmitry-tokarev-nv: Verified on `fcb7e8d`: the example is the relative form pointing at `hero-demo-25.cast`, which exists, and the `src` TSDoc states the constraint explicitly again.

I checked that the new glob does real work rather than just being listed — re-injecting `/dynamo/assets/dynamo-demo.cast` into README.md makes `check_asset_paths.py` exit 1, and restoring the relative form exits 0. Resolving.

dmitry-tokarev-nv (approved):
Thank you for addressing comments. LGTM

PR #12391, opened 2026-07-29
ci: regroup buildkit cache pools by base image tag

nv-anants:
BuildKit pods were split into vllm / sglang / general-trt-combined, giving each group a third of the fleet. TRT-LLM and the general builds together account for the large majority of build time, so that third was oversubscribed while the vllm and sglang pools sat idle.

Group by the dynamo_base image tag instead, which is where the rendered Dockerfiles diverge: manylinux, wheel_builder_base and runtime_wheel_builder are shared by every build kind. TRT-LLM is the only build on a cuda13.1 base and now gets its own pool. The cuda13.0 family splits into the dynamo images and the vLLM/SGLang frameworks, which share a pool because their build filters differ and a dedicated pool for either would idle whenever only the other changed.

Pool sizing is unchanged at ceil(active pods / 3).

closes: OPS-8013

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated BuildKit cache routing to align with the current image groups.
  * Improved build distribution for framework, general, and TRT-LLM workloads.
  * Ensured unknown workload types use the general cache group.


.github/scripts/route_buildkit.sh
```diff
@@ -187,17 +197,18 @@ get_active_indices() {
   echo "${active_indices[@]}"
 }
 
-GROUP_KEYS=("vllm" "sglang" "general-trt-combined")
+GROUP_KEYS=("frameworks" "dynamo" "trtllm")
```
dmitry-tokarev-nv: I'd rename "dynamo" into "non-frameworks" OR "general" to make it more meaningful semanitcally

.github/scripts/route_buildkit.sh
```diff
@@ -11,10 +11,20 @@
 # Routing is optimized for Docker layer caching, linear scaling, and
 # 100% pod utilization across any number of BuildKit pods.
 #
-# CACHE GROUPS (3 distinct groups, one cache domain per framework):
-#   - Group 0 (vllm):                 vLLM
-#   - Group 1 (sglang):               SGLang
-#   - Group 2 (general-trt-combined): TRT-LLM & General Builds
+# CACHE GROUPS (3 distinct groups, one cache domain per dynamo_base image tag):
+#   - Group 0 (frameworks): vLLM & SGLang                    (cuda13.0 base)
+#   - Group 1 (dynamo):     dynamo runtime, planner, frontend,
```
dmitry-tokarev-nv: dynamo >> general or non-frameworks

.github/scripts/route_buildkit.sh
nv-anants: done

.github/scripts/route_buildkit.sh
```diff
@@ -187,17 +197,18 @@ get_active_indices() {
   echo "${active_indices[@]}"
 }
 
-GROUP_KEYS=("vllm" "sglang" "general-trt-combined")
+GROUP_KEYS=("frameworks" "general" "trtllm")
```
dillon-cullinan: Maybe don't change these, and just update the `if` elsewhere in the code to have both entries. Its unintuitive having `trtllm` separate from `frameworks`.

dillon-cullinan (changes requested):

.github/scripts/route_buildkit.sh
nv-anants: updated in https://github.com/ai-dynamo/dynamo/pull/12391/changes/cf0604543ad69a0745b13e78eb30bd5a6b53c374

dillon-cullinan (approved):

PR #12392, opened 2026-07-29
feat(gms): add V1 weight artifact save and load

galletas1712:
## Summary

Stacked directly on #12011 in stack #12691.

Adds V1 exact-ID weight artifact save/load and keeps the implementation under `gpu_memory_service.v1.snapshot`:

- save committed rank-local weight allocation bytes;
- record exact allocation IDs, aligned sizes, shard paths, and offsets; and
- recreate and load those exact allocations in a fresh rank-local V1 server through the existing snapshot transfer backends.

The existing snapshot CLI wrappers select V1 with `--use-v1`.

## Scope

V1 retains its own artifact manifest, metadata, publication, failure cleanup, validation, and session policy. It reuses only a narrow device-shard writing helper from `snapshot.disk`; V0's storage client is a thin adapter to that helper, with its established external behavior unchanged.

The V1 implementation supports the existing `nixl`, `nixl-gds`, and `sharded-ssd` transfer backends. It restores physical weight backing only; restoring the engine process state and CUDA VA reservations remains a separate concern.

External sharded-SSD directories are manifest-owned generations. Coordinated deletion and garbage collection across external SSD roots, and pagination of allocation enumeration, are out of scope for this temporary snapshot path.

This cleanup intentionally removes the weight CLI tests. `test_v1_weight_artifact.py` contains only two behavior-level tests: exact-ID/size/byte round trip, and failure atomicity with replacement-writer recovery. It contains no small monkeypatch or plumbing tests.

## Validation

Current head: `cfddfbc2431182cdbc2f05ec780d81b042cac28e`

- V1 weight artifact behavior tests: `2 passed`;
- existing V0 snapshot regression suite: `12 passed`;
- complete available GMS suite: `66 passed, 8 skipped`;
- exact-head V0 shadow-engine failover on an nscale B200: vLLM `1 passed in 179.78s`, SGLang `1 passed in 118.85s`; TensorRT-LLM remains source-skipped for OPS-4450;
- direct V0 shard-placement parity checks for relative and absolute paths, boundaries, and offsets;
- clean wheel and sdist build with the new module locations and no stale old module paths;
- Ruff, compileall, pre-commit, and `git diff --check` passed; and
- every incremental changed path resolves only to `@ai-dynamo/dynamo-gms-codeowners`.

### DIFF
--- a/lib/gpu_memory_service/cli/snapshot/loader.py
+++ b/lib/gpu_memory_service/cli/snapshot/loader.py
@@ -12,6 +12,7 @@ restored engine on weight load.
 from __future__ import annotations
 
 import argparse
+import importlib
 import logging
 import os
 import time
@@ -158,8 +159,16 @@ def _list_checkpoint_devices(
 
 
 def main(argv: list[str] | None = None) -> None:
+    selector = argparse.ArgumentParser(add_help=False, allow_abbrev=False)
+    selector.add_argument("--use-v1", action="store_true")
+    options, remaining = selector.parse_known_args(argv)
+    if options.use_v1:
+        v1_loader = importlib.import_module("gpu_memory_service.v1.snapshot.loader")
+        v1_loader.main(remaining)
+        return
+
     parser = _build_parser()
-    args = parser.parse_args(argv)
+    args = parser.parse_args(remaining)
     if not args.checkpoint_dir:
         parser.error(
             f"--checkpoint-dir is required for --transfer-backend={args.transfer_backend}"
--- a/lib/gpu_memory_service/v1/server/rpc.py
+++ b/lib/gpu_memory_service/v1/server/rpc.py
@@ -23,6 +23,7 @@ from gpu_memory_service.v1.protocol import (
     REQUEST_TYPES,
     AbortRequest,
     AllocateRequest,
+    AllocationRecord,
     CommitRequest,
     ErrorResponse,
     ExportRequest,
@@ -30,8 +31,11 @@ from gpu_memory_service.v1.protocol import (
     FreeRequest,
     HandshakeRequest,
     HandshakeResponse,
+    ListAllocationsRequest,
+    ListAllocationsResponse,
     Message,
     Request,
+    Response,
     SuccessResponse,
     receive_message,
     send_message,
@@ -234,7 +238,8 @@ class GMSServerMemoryManager:
             raise ValueError("GPU UUID must not be empty")
         self._identity = (str(uuid4()), gpu_uuid)
         self._allocations = GMSAllocationManager(vmm, device)
-        self._sessions = GMSSessionManager(self._allocations.clear)
+        self._allocation_sizes: dict[str, int] = {}
+        self._sessions = GMSSessionManager(self._clear_allocations)
 
     @property
     def identity(self) -> tuple[str, str]:
@@ -249,10 +254,11 @@ class GMSServerMemoryManager:
 
     def handle_request(
         self, session: ServerSession, request: Request
-    ) -> tuple[SuccessResponse | ExportResponse, int]:
+    ) -> tuple[Response, int]:
         if isinstance(request, AllocateRequest):
             self._require_rw(session)
             self._allocations.allocate(request.allocation_id, request.aligned_size)
+            self._allocation_sizes[request.allocation_id] = request.aligned_size
             return SuccessResponse(), -1
         if isinstance(request, ExportRequest):
             self._require_active(session)
@@ -260,7 +266,15 @@ class GMSServerMemoryManager:
         if isinstance(request, FreeRequest):
             self._require_rw(session)
             self._allocations.free(request.allocation_id)
+            del self._allocation_sizes[request.allocation_id]
             return SuccessResponse(), -1
+        if isinstance(request, ListAllocationsRequest):
+            self._require_active(session)
+            allocations = tuple(
+                AllocationRecord(allocation_id, aligned_size)
+                for allocation_id, aligned_size in self._allocation_sizes.items()
+            )
+            return ListAllocationsResponse(allocations), -1
         if isinstance(request, CommitRequest):
             self._require_rw(session)
             self._sessions.commit(session)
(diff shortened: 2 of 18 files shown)


lib/gpu_memory_service/v1/weight_artifact.py
```diff
@@ -0,0 +1,545 @@
...
+        shard_index: int,
+        group: list[tuple[AllocationRecord, LocalMapping, int]],
+    ) -> None:
+        shard_name = f"shard_{shard_index:04d}.bin"
+        shard_path = shard_dirs[shard_index % len(shard_dirs)] / shard_name
+        manifest_path = (
+            str(shard_path)
+            if external_attempts
+            else os.path.join(_SHARDS_DIR, shard_name)
+        )
```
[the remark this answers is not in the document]
galletas1712: Verified against the current location at `v1/snapshot/weight_artifact.py`; the behavior remains. On success, the local manifest is the publication authority and the UUID-qualified external directories become manifest-owned data. No repository janitor scans `*.attempt`. Renaming those directories alone would not solve deletion and would add partial multi-filesystem publication, rollback, and collision cases. Coordinated publication/deletion/crash-GC across external roots is intentionally out of scope for this temporary snapshot path and is now stated explicitly in the PR scope.

lib/gpu_memory_service/cli/snapshot/loader.py
```diff
@@ -158,8 +159,16 @@ def _list_checkpoint_devices(
 
 
 def main(argv: list[str] | None = None) -> None:
+    selector = argparse.ArgumentParser(add_help=False, allow_abbrev=False)
+    selector.add_argument("--use-v1", action="store_true")
+    options, remaining = selector.parse_known_args(argv)
+    if options.use_v1:
+        v1_loader = importlib.import_module("gpu_memory_service.v1.loader")
+        v1_loader.main(remaining)
+        return
```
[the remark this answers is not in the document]
galletas1712: Verified. The deferred import is intentional: the V1 loader and saver reuse the existing CLI `_build_parser`, so importing them from these wrappers at module scope creates a cycle. Removing the dispatch would require extracting or duplicating the parser and expanding this narrow PR. We are keeping the explicit `--use-v1` dispatch so V0 remains untouched.

lib/gpu_memory_service/v1/server/rpc.py
```diff
@@ -249,18 +254,27 @@ def acquire(
...
             self._allocations.free(request.allocation_id)
+            del self._allocation_sizes[request.allocation_id]
             return SuccessResponse(), -1
+        if isinstance(request, ListAllocationsRequest):
+            self._require_active(session)
+            allocations = tuple(
+                AllocationRecord(allocation_id, aligned_size)
+                for allocation_id, aligned_size in self._allocation_sizes.items()
+            )
+            return ListAllocationsResponse(allocations), -1
```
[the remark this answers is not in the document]
galletas1712: Confirmed the 1 MiB bound. These records correspond to CUDA MemPool allocator segments, not individual model tensors, and supported target models are expected to remain far below the roughly 17k-allocation threshold. Pagination would add protocol state and recovery surface without an observed workload need. We are retaining the frame safety limit and will add pagination only if measured allocation cardinality approaches it.

lib/gpu_memory_service/v1/snapshot/loader.py
```diff
@@ -0,0 +1,83 @@
...
+def main(argv: list[str] | None = None) -> None:
+    parser = _build_parser()
+    parser.description = "Load exact GMS V1 weights into a fresh rank-local server."
+    parser.add_argument(
+        "--device",
+        type=int,
+        default=0,
+        help="CUDA-visible rank-local device ordinal.",
+    )
+    parser.add_argument(
```
mohammedabdulwahhab, line 45: Should this be moved into build_parser so all args in one place?

lib/gpu_memory_service/v1/snapshot/loader.py
```diff
@@ -0,0 +1,83 @@
...
+        "--posix-backend-param",
```
mohammedabdulwahhab, line 46: An example of what this could look like would be helpful.

lib/gpu_memory_service/v1/snapshot/saver.py
```diff
@@ -0,0 +1,57 @@
...
+    parse_sharded_ssd_roots,
+)
+from gpu_memory_service.v1.device import get_socket_path
+from gpu_memory_service.v1.snapshot.weight_artifact import save_weights
+
+
+def main(argv: list[str] | None = None) -> None:
+    parser = _build_parser()
+    parser.description = "Save exact committed GMS V1 weight allocations."
+    parser.add_argument(
```
mohammedabdulwahhab, line 23: Why is this not already in build parser?

lib/gpu_memory_service/v1/snapshot/weight_artifact.py
```diff
@@ -0,0 +1,496 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Exact-allocation cold storage for experimental GMS V1 weights."""
```
mohammedabdulwahhab: I think this comment needs to be more descriptive in that it contains the load-bearing implementation of save and load.

mohammedabdulwahhab (approved):
Some nits, but overall looks good.

lib/gpu_memory_service/v1/snapshot/loader.py
galletas1712: Moved the V1 loader arguments into a local `_build_parser()`. It extends the V0 parser, sets the V1 description, and adds both V1-only arguments in one place; `main()` now only builds and parses it.

lib/gpu_memory_service/v1/snapshot/loader.py
galletas1712: Added a concrete help example using a POSIX parameter already exercised by GMS: `--posix-backend-param ios_pool_size=64`.

lib/gpu_memory_service/v1/snapshot/saver.py
galletas1712: Moved `--device` into a V1-local `_build_parser()` that extends the V0 saver parser. `main()` now only builds and parses it.

lib/gpu_memory_service/v1/snapshot/weight_artifact.py
galletas1712: Expanded the module docstring to identify this as the load-bearing save/load implementation and list the responsibilities it owns: validation, exact-ID reconstruction, device mappings, transfer, cleanup, and publication.

Issue #11762, opened 2026-07-16
[BUG]: /v1/messages returns 500 for Claude Code ≥2.1.154 — mid-conversation `role:"system"` turns fail chat-template rendering


bloodeagle40234:
### Describe the Bug

With `--enable-anthropic-api`, **Claude Code ≥ 2.1.154** sends turns with `role:"system"` **inside** the `messages` array (the Anthropic Messages API only defines `system` at top level). The `AnthropicCreateMessageRequest -> NvCreateChatCompletionRequest` conversion (`lib/llm/src/protocols/anthropic/types.rs`) emits them in place, producing an ordering like `[System, User, …, System]`. Chat templates that require a leading/single system message (Qwen, Llama) then raise during rendering, and the frontend returns **HTTP 500**. Every Claude Code request fails.

This is **not** the deserialize bug from [#10108](<https://github.com/ai-dynamo/dynamo/issues/10108>) (that already made `role:"system"` deserialize succeed). The failure is at chat-template application.

It is the Anthropic `/v1/messages` counterpart of ai-dynamo/dynamo#10206 (Codex first-turn "System message must be at the beginning" on `/v1/responses`, fixed by [#10283](<https://github.com/ai-dynamo/dynamo/issues/10283>) which merges a *leading* run of system messages). This case differs in two ways: (1) the Anthropic path, and (2) the failing system turn appears **mid-conversation** (after user/assistant turns, and it toggles under tool use) — which a leading-run merge does not cover.

### Steps to Reproduce

Version boundary (bisected on real hardware): **Claude Code 2.1.150 → OK**, **2.1.154 → fails**.

Minimal repro (fails on current `main` with a Qwen backend + `--enable-anthropic-api`):

```bash
curl -sS "$FRONTEND/v1/messages?beta=true" \
  -H 'content-type: application/json' -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"<qwen-model>","max_tokens":64,"stream":true,
       "system":[{"type":"text","text":"You are Claude Code."}],
       "messages":[{"role":"user","content":"hello"},
                   {"role":"system","content":[{"type":"text","text":"mid-conversation context"}]}]}'
```

Control (no message-level system) returns 200, confirming the system turn's position is the sole cause.

What Claude Code actually sends (captured, shape only):

* **Turn 1:** `messages = [user, system]` + top-level `system`; the `role:"system"` block at index 1 is the agent-types/MCP/skills list, byte-stable across turns.
[...]

PR #12404, opened 2026-07-29
fix(frontend): return 400 instead of 500 when chat template rendering fails

KrishnanPrash:
#### Summary

When a chat template refuses to render a request (for example via `raise_exception`), the frontend returns HTTP 500 `"Failed to generate completions"` and drops the template's own error message. The failure is caused by the request, so it should be a client error.

This PR classifies every chat template render failure as HTTP 400 and forwards the template's message to the client. vLLM does the same for template render failures. The fix is @rmccorm4's, taken over from `rmccormick/template-error-400`: `OpenAIPreprocessor::map_prompt_render_error` now maps all render failures to `ErrorType::InvalidArgument`, which the HTTP layer already turns into 400.

**Before / after** (model whose template rejects a history with no user message):

```json
POST /v1/chat/completions
{"model": "<model>", "max_tokens": 1,
 "messages": [{"role": "assistant", "content": "prefill"}]}
```

Before:

```http
HTTP/1.1 500 Internal Server Error
{"message":"Failed to generate completions","type":"Internal Server Error","code":500}
```

After:

```http
HTTP/1.1 400 Bad Request
{"message":"invalid operation: No user query found in messages. (in default:1)","type":"Bad Request","code":400}
```

Notes:

- `stream: true` behaves the same: a JSON 400, not an SSE stream. Rendering fails before the response stream opens.
- Requests the template accepts are unchanged. An assistant-only history against a template that renders it still returns 200. The earlier revision of this PR rejected user-less requests at the frontend; that was dropped because vLLM and SGLang both accept such requests when the model's template does. Whether a history is servable is the template's decision; the frontend's job is to report it with the right status code.
- The fix covers `/v1/chat/completions`, `/v1/responses`, and `/v1/messages`, which all use the same renderer.
- Same bug class as #11762 (mid-conversation `role: "system"` turns make Qwen and Llama templates raise, surfacing as 500).
- Render failures are logged at `debug` level, so a server-side template misconfiguration that fails at render time stays diagnosable.

#### Validation

```bash
cargo test -p dynamo-llm --test chat_template_render_errors
cargo test -p dynamo-llm --lib
cargo fmt --all --check && cargo clippy -p dynamo-llm --all-targets
```

- New end-to-end test (`lib/llm/tests/chat_template_render_errors.rs`): drives real HTTP through a real preprocessor and template. A raising template returns 400 with a JSON body for both `stream: false` and `stream: true`; an accepting template returns 200 for the same assistant-only request.
- Unit tests pin the error classification directly.
- Reverting the fix makes the e2e test fail with `left: 500, right: 400` on both streaming modes, so the test covers the regression, not the implementation.

### DIFF
--- a/lib/llm/src/preprocessor.rs
+++ b/lib/llm/src/preprocessor.rs
@@ -1461,17 +1461,17 @@ impl OpenAIPreprocessor {
         hidden_eos_token_ids.len() != before
     }
 
+    /// Rendering is driven by the request, so its failures are reported as 400 rather than
+    /// 500, matching vLLM. A misconfigured template can also fail here, for instance a
+    /// `chat_template` map that omits the `tool_use` key, so log the cause chain before it
+    /// is flattened into the client-facing message.
     fn map_prompt_render_error(error: anyhow::Error) -> anyhow::Error {
-        if let Some(PromptRenderError::InvalidRequest(message)) =
-            error.downcast_ref::<PromptRenderError>()
-        {
-            return DynamoError::builder()
-                .error_type(ErrorType::InvalidArgument)
-                .message(message.clone())
-                .build()
-                .into();
-        }
-        error
+        tracing::debug!(?error, "Chat template rendering failed");
+        let message = match error.downcast_ref::<PromptRenderError>() {
+            Some(PromptRenderError::InvalidRequest(message)) => message.clone(),
+            None => format!("{error:#}"),
+        };
+        invalid_argument_error(message)
     }
 
     pub fn apply_template<
@@ -4233,13 +4233,16 @@ mod tests {
     }
 
     #[test]
-    fn ordinary_prompt_error_remains_internal() {
+    fn ordinary_prompt_error_maps_to_invalid_argument() {
         let mapped = OpenAIPreprocessor::map_prompt_render_error(anyhow::anyhow!(
             "template configuration failed"
         ));
+        let mapped = mapped
+            .downcast_ref::<DynamoError>()
+            .expect("any prompt render failure should map to a DynamoError");
 
-        assert!(mapped.downcast_ref::<DynamoError>().is_none());
-        assert_eq!(mapped.to_string(), "template configuration failed");
+        assert!(matches!(mapped.error_type(), ErrorType::InvalidArgument));
+        assert_eq!(mapped.message(), "template configuration failed");
     }
 
     fn url_entry(u: &str) -> MultimodalData {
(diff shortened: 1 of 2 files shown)


rmccorm4:
Native vLLM 0.26.0 does **not** reject the assistant-only request in this PR. I served `Qwen/Qwen3-0.6B` directly and received HTTP 200 with generated output. The streaming form also returned HTTP 200, normal SSE deltas, and `[DONE]`.

<details>
<summary>Example vLLM 0.26.0 serve command</summary>

```bash
CUDA_VISIBLE_DEVICES=0 vllm serve Qwen/Qwen3-0.6B   --host 127.0.0.1   --port 18105   --dtype half   --max-model-len 4096   --gpu-memory-utilization 0.35   --enforce-eager
```

</details>

Request:

```bash
curl -i   -X POST http://127.0.0.1:18105/v1/chat/completions   -H 'Content-Type: application/json'   --data-binary '{
    "model": "Qwen/Qwen3-0.6B",
    "max_tokens": 32,
    "messages": [
      {
        "role": "assistant",
        "content": "I am an assistant."
      }
    ]
  }'
```

Response (abridged):

```http
HTTP/1.1 200 OK
content-type: application/json

{
  "model": "Qwen/Qwen3-0.6B",
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": "<think>\nOkay, the user says, \"I am an assistant.\" ..."
      },
      "finish_reason": "length"
    }
  ],
  "system_fingerprint": "vllm-0.26.0-nohash",
  "usage": {
    "prompt_tokens": 13,
    "completion_tokens": 32,
    "total_tokens": 45
  }
}
```

This means the proposed user-role validation is stricter than native vLLM 0.26.0 for this model/template. It would be a Dynamo API policy rather than vLLM parity.

rmccorm4:
Native SGLang does **not** reject the default assistant-only request in this PR. I served `Qwen/Qwen3-0.6B` directly using `lmsysorg/sglang:nightly-dev-cu13-20260729-16a52bff`, which reports SGLang package version `0.0.0.dev1+gcb12a1547`, and received HTTP 200 with generated output.

<details>
<summary>Example SGLang serve command</summary>

```bash
docker run --rm   --name sglang-qwen3   --gpus 'device=0'   --shm-size 8g   -p 127.0.0.1:18106:30000   -v "$HOME/.cache/huggingface:/root/.cache/huggingface"   lmsysorg/sglang:nightly-dev-cu13-20260729-16a52bff   sglang serve   --model-path Qwen/Qwen3-0.6B   --host 0.0.0.0   --port 30000   --dtype float16   --context-length 4096   --mem-fraction-static 0.35   --cuda-graph-backend-decode disabled   --cuda-graph-backend-prefill disabled   --disable-radix-cache
```

</details>

Request:

```bash
curl -i   -X POST http://127.0.0.1:18106/v1/chat/completions   -H 'Content-Type: application/json'   --data-binary '{
    "model": "Qwen/Qwen3-0.6B",
    "max_tokens": 32,
    "messages": [
      {
        "role": "assistant",
        "content": "I am an assistant."
      }
    ]
  }'
```

Response (abridged):

```http
HTTP/1.1 200 OK
content-type: application/json

{
  "model": "Qwen/Qwen3-0.6B",
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": "<think>\nOkay, the user said, \"I am an assistant.\" Let me think about how to respond. ..."
      },
      "finish_reason": "length"
    }
  ],
  "usage": {
    "prompt_tokens": 13,
    "completion_tokens": 32,
    "total_tokens": 45,
    "reasoning_tokens": 0
  },
  "metadata": {
    "weight_version": "default"
  }
}
```

This is SGLang's default behavior with `continue_final_message` omitted. The proposed user-role validation is therefore also stricter than native SGLang's default path for this model/template.

lib/llm/src/http/service/openai.rs
```diff
@@ -2247,6 +2247,18 @@ pub fn validate_chat_completion_required_fields(
         }));
     }
 
+    if !inner
+        .messages
+        .iter()
+        .any(|message| matches!(message, ChatCompletionRequestMessage::User(_)))
+    {
+        return Err(ErrorMessage::from_http_error(HttpError {
```
rmccorm4: I don't think this fix is at the right layer, this will reject assistant-only messages for all models, but only certain model templates reject it while other accept it. It's likely an issue with funneling back the right error code 400 instead of 500 when template rendering fails or something along those lines.

What do you think of a change like this instead: https://github.com/ai-dynamo/dynamo/compare/main...rmccormick/template-error-400? (If you agree, please just take it over in this PR)

rmccorm4 (approved):

PR #8783, opened 2026-04-28
feat: Rust ext_proc Server for EPP. Fixes [DYN-2909]

atchernych:
#### Overview:

Fixes [DYN-2909]

This PR creates an Ext_proc service for EPP written in rust. It is meant to replace the llm-d implementation. The ext_proc service will be called by a gateway like Ai Envoy to make a routing decision. The ext_proc server will call our router to make a routing decision and will put the worker ids into the request header as well as tokens into the request body.

Ext protocol is stable so there should not be much maintenance to this code: https://github.com/envoyproxy/envoy/blob/main/api/envoy/service/ext_proc/v3/external_processor.proto  

The current GO setup has **three language boundaries**:
(Envoy C++ → Go gRPC → C FFI → Rust) 
In this implementation there is **one** boundary (Envoy C++ → Rust gRPC)

This is per last part of the proposal: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2790 

This is a RUST implementation of the Light Weight EPP https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2842/changes:
```
Go LW-EPP (GAIE #2842).         | Rust ext-proc (Dynamo #8783)        
pkg/lwepp/handlers/server.go    | src/server.rs                                       
pkg/lwepp/handlers/request.go | src/server.rs (request handlers)      
pkg/lwepp/handlers/response.go | src/server.rs + src/envoy_helpers.rs 
pkg/lwepp/datastore/datastore.go (Endpoint) | src/picker.rs 
EndpointPicker / RoundRobinPicker (in server.go) | src/picker.rs 
```

**Differences**
- The Light weight EPP go implementation does not yet handle the handling the request body.
When endpoint selection happens, Go LW-EPP Picks endpoint in the header phase (handleRequestHeaders). Does not look at the body at all. Rust implementation defers picking to the body phase (handle_request_body) so it can tokenize the body for KV-aware routing. 

[...]

PR #12443, opened 2026-07-30
build(container): install libprotobuf-dev in the dev image

MatejKosec:
## Overview:

The dev image has `protoc` but not the protobuf well-known types, so `cargo check --workspace` cannot run inside it.

## Details:

`container/templates/dev.Dockerfile` installs `protobuf-compiler`. That package ships the `protoc` binary and none of the `.proto` files — `google/protobuf/struct.proto` and friends live in `libprotobuf-dev`. On Ubuntu noble, `protobuf-compiler` contains zero files under `/usr/include/google/protobuf/` and `libprotobuf-dev` contains eleven.

`dynamo-ext-proc` is a root workspace member, and its `build.rs` compiles four vendored Envoy protos, two of which import `google/protobuf/struct.proto` and `duration.proto`. So a workspace build inside the dev image fails with:

```
protoc failed: google/protobuf/struct.proto: File not found
```

Crate-scope builds are unaffected, which is why this went unnoticed — only a `--workspace` build reaches it.

This is not a new requirement, and nothing about the change is speculative. ai-dynamo/dynamo#8783 added `ext-proc` to the workspace, and it already installs both packages in its own image, with a comment saying why:

```dockerfile
# protobuf-compiler provides protoc; libprotobuf-dev provides the well-known
# .proto files (google/protobuf/struct.proto etc.) that our vendored Envoy
```

The crate's `build.rs` says the same thing:

```rust
// Well-known Google protobuf types (struct.proto, duration.proto, etc.)
// live in /usr/include when libprotobuf-dev is installed.
if std::path::Path::new("/usr/include/google/protobuf").exists() {
```

So the dependency was known and handled in `ext-proc/Dockerfile`; the shared dev image was simply never updated to match. This PR closes that gap.

That `exists()` guard is also why the failure is hard to read. With the directory absent, the build script quietly drops `/usr/include` from the include path and carries on, so the error surfaces later as an unresolvable import instead of a clear "libprotobuf-dev is missing".

### An alternative worth considering

`container/templates/wheel_builder.Dockerfile` already solves this differently: it extracts both `bin/protoc` and `include/*` from the upstream protoc 25.3 release and sets `PROTOC`. That pins binary and includes to one version, where two apt packages stay in step only by coming from the same suite. The dev image copies just rustup, cargo and maturin from that stage, so it never inherited the includes.

I went with the apt package because it matches how this file installs everything else, and because `build.rs` probes `/usr/include` specifically — the `wheel_builder` route lands them in `/usr/local/include`, so it would also need `PROTOC` set or that guard widened. Happy to switch if the container owners prefer the consistency.

The whole change is six lines in `container/templates/dev.Dockerfile` — `libprotobuf-dev` added to the existing apt list, plus a comment.

For whether it is the right fix, the two files that already document the dependency are `deploy/inference-gateway/ext-proc/Dockerfile` (lines 33-40) and `deploy/inference-gateway/ext-proc/build.rs`.

## How this surfaced

An automated agent working on ai-dynamo/dynamo#12288 ran `cargo check --workspace --all-targets` in the dev image, hit the error above, and confirmed it reproduced on the unmodified tree before stopping as blocked. Its own change was healthy — 1846 crate-scope tests passed.

On that base commit the failing crate was `lib/sidecar/vllm`, which no longer exists on `main`; `ext-proc` is the crate with this requirement today. Same failure, same cause.

## Validation

* Ubuntu noble package file lists confirm the split: `protobuf-compiler` ships no `/usr/include/google/protobuf/*.proto`, `libprotobuf-dev` ships eleven including `struct.proto`. Checked against the published lists, not from memory.
* `Cargo.toml` lists `deploy/inference-gateway/ext-proc` in `[workspace] members` and the crate inherits `version.workspace`, so `cargo check --workspace` genuinely builds it.
* `container/render.py` renders the `dev` target for vllm, sglang and trtllm with the package present. trtllm needs `--cuda-version 13.1`, as its `--help` documents; that is unrelated to this change and behaves identically without it.
* No rendered Dockerfile is committed, so the template is the whole change.
* Pre-commit hooks pass.

I have not built the image locally, and — correcting something I wrote earlier in this description — **PR CI does not build it either**. The `vllm-dev` / `sglang-dev` / `trtllm-dev` jobs live in `post-merge-ci.yml`, which triggers only on push to `main` or `release/*`, so they report as `skipping` here. Every check on this PR is green, but none of them has built the file this PR changes.

If you want the image built before merge rather than after, `post-merge-ci.yml` also accepts `workflow_dispatch`, so it can be run against the `pull-request/12443` branch that copy-pr-bot created at `b0ebc58f`. I don't have permission to dispatch it. Otherwise the first real build of this change happens post-merge, which is worth knowing given the change is a package install that has to resolve on both the amd64 and arm64 bases.

## Related Issues


container/templates/dev.Dockerfile
```diff
@@ -84,6 +84,12 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         pybind11-dev \
         pkg-config \
         protobuf-compiler \
+        # protobuf-compiler ships protoc but none of the well-known types.
+        # libprotobuf-dev provides /usr/include/google/protobuf/*.proto, which
+        # the vendored envoy protos under deploy/inference-gateway/ext-proc
+        # import. Without it `cargo check --workspace` fails with
+        # "google/protobuf/struct.proto: File not found".
```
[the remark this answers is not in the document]
MatejKosec: `Cargo.toml` lists `"deploy/inference-gateway/ext-proc"` in `[workspace] members`, and the crate inherits `version.workspace = true` / `edition.workspace = true` rather than declaring its own `[workspace]` table. 

The build script is actually the strongest evidence for this PR — it already names the package:

```rust
// Well-known Google protobuf types (struct.proto, duration.proto, etc.)
// live in /usr/include when libprotobuf-dev is installed.
if std::path::Path::new("/usr/include/google/protobuf").exists() {
    include_dirs.push("/usr/include".to_string());
}
```

That guard is why the failure looks the way it does. When the directory is absent the script silently omits `/usr/include` from the include path and proceeds, so the error arrives later as an unresolvable import rather than as a clear "libprotobuf-dev is missing". Installing the package makes the guard true.

One correction I owe you, since you were right to poke at the reproduction path. The run that originally hit this was on base `ddf2b8e2a`, where the failing crate was `lib/sidecar/vllm` (`dynamo-vllm-sidecar`), not `ext-proc`. That crate no longer exists on `main`. On current `main` the crate with this requirement is `ext-proc`. Both are workspace members and both fail the same way, but the PR body describes today's tree, which is the one that matters for the fix. I'll make that explicit in the description.

On `wheel_builder`'s approach: I agree it's the more consistent fix and I'd be glad to switch. It pins protoc and its includes to one version, whereas two apt packages are only kept in step by being drawn from the same suite. I went with the smaller change because it matches how this file already installs everything else, and because `build.rs` looks in `/usr/include` specifically — the `wheel_builder` route puts them in `/usr/local/include`, so it would also need `PROTOC` set or the guard widened. Happy to do that version if the container owners prefer it.

connorcarpenter15: AI slop comment

connorcarpenter15 (approved):

nv-anants (approved):

container/templates/dev.Dockerfile
nv-anants: +1

MatejKosec: Removed in ff14e5d. Just the package line now.

Issue #12441, opened 2026-07-30
DEP: decouple Spica and Replay from Dynamo


tedzhouhk:
## Summary

Refactor Spica and Replay into Dynamo-independent AISim core packages with optional Dynamo Planner and Router adapters. Backend-only simulation and sweep must install and run without `ai-dynamo`; Dynamo-specific search and runtime behavior is loaded only when explicitly selected.

## Motivation

Spica currently mixes generic sweep/backend logic with Dynamo Planner, Router, Mocker, Replay, and KVBM details. This prevents the simulation stack from becoming a standalone backend capability, makes Dynamo an unconditional dependency, and couples the future AISim repository extraction to Dynamo implementation details. Replay is undergoing the same separation in parallel, so both workstreams need a stable shared contract.

## Proposal

- Keep the current Spica sweep, backend materialization, scoring, and parallel orchestration in `aisimulate`, but remove all direct Dynamo dependencies.
- Move Planner and Router search-space validation, preparation, and candidate-specific configuration generation into adapters under their corresponding Dynamo components.
- Discover selected search adapters through the `aisimulate.adapters` Python entry-point group; do not load Dynamo when no Dynamo adapter is selected.
- Replace the direct Dynamo replay evaluator with an injected `RunnerFactory` consuming a versioned, serializable `ReplaySpec` and returning `ReplayReport`.
- Develop Replay in parallel as two compositions over one event loop:
  - AISim Replay is Dynamo-independent and provides built-in neutral placement.
  - Dynamo Replay reuses AISim Replay and injects Dynamo Router `PlacementPolicy` and Planner `ScalingPolicy` runtime adapters.
- Version `ReplaySpec`, runtime-hook descriptors, runner capabilities, and canonical fixtures so the Spica and Replay workstreams can develop independently and fail early on capability mismatches.
- Remove KVBM search/configuration fields without a compatibility layer. Native G2 support is a separate follow-up.
- Preserve behavioral parity for all existing non-KVBM Spica search, scoring, Planner, Router, concurrency, timeout, and replay-report behavior.

## Alternate Solutions

- Keep lazy hard-coded Dynamo imports in `aisimulate`. This avoids an adapter registry but leaves the core aware of Dynamo module paths and requires another cleanup during repository extraction.
- Put Python `module:function` paths in user configuration. This removes hard-coded imports but exposes internal Dynamo module layout as a public configuration contract.
[...]

PR #12446, opened 2026-07-30
refactor(spica): decouple and rename AI Simulate Sweeper

tedzhouhk:
## Summary

- rename the experimental `aisimulate.spica` package to `aisimulate.sweeper`
- replace the public `run_smart_search` function with the reusable `Sweeper` execution class; every `run()` call owns fresh studies, caches, runners, and worker pools
- keep the Sweeper core Dynamo-free behind versioned `ReplaySpec` / `RunnerFactory` contracts
- rename the extension ABI to `SweepConfigProvider`, its input model to `AdapterSearchConfig`, and its entry-point group to `aisimulate.sweep_config_providers`
- keep Dynamo-owned Planner and Router providers under their components as `DynamoPlannerSweepConfigProvider` and `DynamoRouterSweepConfigProvider`
- deprecate KVBM sweep fields because native G2 replaces that path
- add a bounded `post_merge` integration test that executes full `Sweeper` orchestration through real `DynamoReplayRunnerFactory` instances in two spawned workers
- make `aisimulate/docs/sweeper` and `aisimulate/examples/sweeper` the canonical backend-neutral documentation and example locations; Fern consumes checked-in synchronized copies and keeps Dynamo-specific integration guidance separate

Replay is being refactored in parallel. Backend-only Sweeper will compose with Dynamo-free Replay, while Sweeper with Dynamo hooks will compose with Dynamo Replay. This PR keeps `DynamoReplayRunnerFactory` as the transitional boundary until that split lands.

- public execution API: `aisimulate/src/aisimulate/sweeper/search.py`
- provider ABI and discovery: `aisimulate/src/aisimulate/sweeper/provider.py` and `discovery.py`
- Dynamo providers: `components/src/dynamo/planner/simulation/provider.py` and `components/src/dynamo/router/simulation/provider.py`
- replay composition and spawned-process coverage: `components/src/dynamo/replay/simulation.py` and `tests/test_simulation_integration.py`
- canonical docs and Fern synchronization: `aisimulate/docs/sweeper` and `docs/fern/scripts/sync_aisimulate_docs.py`

## Related Issues

- Related DEP: #12441
- Design plan: #12273

## Install and run a Dynamo sweep

The supported prebuilt environment is the `dynamo-planner` image. Its build creates the `aisimulate` wheel from the same commit as Dynamo and installs both wheels into the image. The `aisimulate` wheel remains image-local and is not staged or published as a release artifact.

This PR does **not** add an `ai-dynamo[simulation]` extra.

> [!NOTE]
> **TODO:** Add `ai-dynamo[simulation]` only after the complete AI Simulate and Replay refactor has a supported distribution path.

For source development:

```bash
python -m pip install pip "maturin[patchelf]"
cd lib/bindings/python
maturin develop --uv --release --features aic-forward-pass
cd ../../..
python -m pip install --no-deps -e .
python -m pip install -e ./aisimulate
python -m pip install -r container/deps/requirements.planner.txt
```

Run the Dynamo integration example:

```bash
python aisimulate/examples/sweeper/tools/run_sweep.py \
  --config aisimulate/examples/sweeper/configs/smart_sweep.yaml
```

Or compose it directly:

```python
from aisimulate.sweeper import SmartSearchConfig, Sweeper
from dynamo.replay.simulation import DynamoReplayRunnerFactory

config = SmartSearchConfig.from_yaml("smart_sweep.yaml")
candidates = Sweeper(
    runner_factory=DynamoReplayRunnerFactory(),
).run(config)
```

There is no implicit Dynamo runtime or `--enable-dynamo` flag. `python -m aisimulate.sweeper --config ...` validates the backend-neutral config and explains that an application must inject a runner.

## Configuration format

Core backend fields stay under `search_space`. Each optional adapter receives a complete adapter-owned **search space**, not one concrete Planner or Router config:

```yaml
search_space:
  deployment_mode: [disagg, agg]
  model_name: deepseek-ai/DeepSeek-V3
  hardware_sku: h200_sxm
  gpu_budget: 32
  backend: [vllm]

adapters:
  dynamo.router:
    search_space:
      mode: [kv_router, round_robin]
      overlap_score_credit: [0.0, 0.5, 1.0]
  dynamo.planner:
    search_space:
      scaling_policy: [disabled, load_180_5]
      fpm_sampling: [default]
      load_sensitivity: [default]

workload:
  trace_path: /data/replay/traffic.jsonl
  trace_format: mooncake

goal:
  target: goodput_per_gpu
  sla: {ttft_ms: 2000, itl_ms: 30}

sweep:
  max_rounds: 40
  parallel_evals: 16
```

The `ai-dynamo` distribution registers only the selected provider implementations. Omitting an adapter disables that feature and avoids importing its Dynamo code.

## Validation

- `282 passed` — full Sweeper suite plus Planner/Router providers, Dynamo Replay unit tests, and real replay integration (including `parallel_evals=2` spawned workers carrying active Planner and Router runtime hooks)
- fixed-SHA parity audit: the legacy Spica baseline and new Sweeper produced byte-identical canonical candidate JSON across two runs per revision (`92884f9a…72f22`), and Planner/Router search plans plus replay specs also matched across two runs per revision (`faed4cdf…0434`)
- backend-neutral and Dynamo integration examples now share the canonical `aisimulate/examples/sweeper` tree; the neutral example executes successfully and all moved tool entry points import correctly
- CODEOWNERS coverage remains 100%; `aisimulate/` resolves to `@ai-dynamo/dynamo-planner-codeowners`, `@jasonqinzhou`, and `@dreamtalen`, and all 86 CODEOWNERS tests pass
- all changed-file pre-commit hooks pass, including the canonical AI Simulate → Fern sync check
- canonical documentation relative links, Fern YAML, and Fern asset paths validate locally
- backend-neutral example executes successfully without importing Dynamo
- branch merged with current `origin/main` without conflicts

## Compatibility notes

- public execution is now `Sweeper(runner_factory=...).run(config)`; `run_smart_search` is removed
- package imports move from `aisimulate.spica` to `aisimulate.sweeper`
- Planner and Router inputs remain under `adapters.<name>.search_space`
- KVBM fields fail with an explicit deprecation error and have no adapter migration
- `AISIMULATE_SWEEPER_VIZIER_ALGO` is the new experimental sampler override; `SPICA_VIZIER_ALGO` remains a deprecated fallback during migration
- adapter namespacing can change Vizier's exact finite-round suggestion sequence; parity targets search domains, scoring, replay behavior, diagnostics, and candidate semantics rather than an identical optimizer trajectory


components/src/dynamo/planner/simulation/adapter.py
```diff
@@ -0,0 +1,452 @@
...
+            sla=state["sla"] if isinstance(state["sla"], dict) else None,
+            min_endpoint=space.min_endpoint,
+        )
+        hook = RuntimeHookSpec(
+            provider=_HOOK.provider,
+            kind=_HOOK.kind,
+            api_version=_HOOK.api_version,
+            config={"planner_config": planner_config},
+        )
+        return AdapterReplaySpec(config=planner_config, runtime_hooks=(hook,))
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb: AdapterReplaySpec.config now preserves the raw scaling_policy selection for candidate output, while the runtime hook still receives only valid PlannerConfig fields. Regression coverage includes preset and custom policies.

aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -66,51 +80,301 @@ def evaluate(
...
+            )
+            local_ranges = plan.fragment.float_ranges_by_branch.get(
+                branch.deployment_mode, {}
+            )
+            overlap = set(local_choices).intersection(local_ranges)
+            if overlap:
+                raise ValueError(
+                    f"adapter {name!r} defined parameters as both categorical and "
+                    f"continuous: {sorted(overlap)}"
+                )
```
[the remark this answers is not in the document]
tedzhouhk: No code change: omission from SearchSpaceFragment means that an adapter contributes no optimizer dimensions for that branch; it does not declare the topology unsupported. materialize_replay is deliberately called for every configured candidate and must accept an empty selection when it declares no knobs. Silently skipping it would disable a requested adapter.

aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -142,27 +406,123 @@ def _evaluate_one(
...
             # concurrency and one derived from kv_load_ratio.
             sample["concurrency"] = concurrency
-        plan = build_deployment(
-            sample,
-            backend_version=backend_version,
-            optimization_target=goal.target.planner_optimization_target,
-            planner_sla=goal.sla,
+        backend_deployment = build_backend_deployment(
+            sample, backend_version=backend_version
+        )
```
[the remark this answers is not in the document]
tedzhouhk: Acknowledged; no code change. Materialization remains on the main process by ABI design so adapter implementations and Dynamo imports do not cross the worker boundary. Replay remains worker-only. Large-sweep profiling can be handled separately if this becomes a bottleneck.

.github/workflows/release.yml
```diff
@@ -811,15 +812,22 @@ jobs:
...
           shopt -u nullglob
 
           if [ ${#AI_DYNAMO_WHEELS[@]} -eq 0 ] || [ ${#AI_DYNAMO_RUNTIME_WHEELS[@]} -eq 0 ]; then
             echo "::error::Missing ai_dynamo or ai_dynamo_runtime wheel in ${UPLOAD_DIR}"
             exit 1
           fi
+          if [ ${#AISIMULATE_WHEELS[@]} -ne 1 ]; then
+            echo "::error::Expected exactly one aisimulate wheel in ${UPLOAD_DIR}, found ${#AISIMULATE_WHEELS[@]}"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb: release staging now reads wheel METADATA and verifies that the aisimulate wheel Version matches the exact aisimulate pin in every ai_dynamo Requires-Dist entry before upload. The workflow block was exercised against locally built wheels.

aisimulate/src/aisimulate/spica/__init__.py
```diff
@@ -20,95 +34,92 @@
...
+    "resolve_adapters",
+    "BackendDeploymentSpec",
+    "HookCapability",
+    "ReplayReport",
+    "ReplaySpec",
+    "Runner",
+    "RunnerCapabilities",
+    "RunnerFactory",
+    "canonical_json",
+    *_LAZY_EXPORTS,
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb: every lazy export is listed explicitly in sorted __all__, with a regression assertion that the lazy-export registry remains a subset of the public API.

aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -142,27 +406,123 @@ def _evaluate_one(
...
+        and "goodput_output_throughput_tok_s" not in report
+    ):
+        return (
+            None,
+            None,
+            "failed",
+            "runner contract violation: goodput objective requires "
+            "goodput_output_throughput_tok_s; aggregate latency cannot be used "
+            "as a fallback",
+        )
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb by parenthesizing the failure message while preserving the four-tuple and exact runtime text.

components/src/dynamo/planner/simulation/presets.py
```diff
@@ -0,0 +1,92 @@
...
+def throughput_intervals(policies: list[str | dict[str, Any]]) -> list[int]:
+    """Return distinct intervals for policies using predictive throughput scaling."""
+
+    intervals: set[int] = set()
+    for entry in policies:
+        fields = scaling_fields(entry)
+        interval = fields["throughput_adjustment_interval_seconds"]
+        if fields["enable_throughput_scaling"] and interval is not None:
+            intervals.add(int(interval))
+    return sorted(intervals)
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb: custom throughput intervals are canonicalized to integer keys before predictor lookup, while load intervals remain unchanged. A 180.0 regression verifies winner lookup and materialization.

components/src/dynamo/planner/tests/integration/test_spica_simulation_parity.py
```diff
@@ -0,0 +1,249 @@
...
+from dynamo.planner.simulation import create_adapter
+from dynamo.replay.simulation import DynamoReplayRunnerFactory
+
+pytestmark = [
+    pytest.mark.gpu_0,
+    pytest.mark.integration,
+    pytest.mark.planner,
+    pytest.mark.pre_merge,
+    pytest.mark.filterwarnings("ignore:invalid escape sequence.*:SyntaxWarning"),
+]
```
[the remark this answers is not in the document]
tedzhouhk: The Spica replay parity suite was removed in 9a57231abb as requested, so this marker concern is obsolete.

components/src/dynamo/planner/tests/integration/test_spica_simulation_parity.py
```diff
@@ -0,0 +1,249 @@
...
+
+_REPO_ROOT = Path(__file__).resolve().parents[6]
+TRACE = str(_REPO_ROOT / "aisimulate/tests/spica/data/mooncake_tiny.jsonl")
```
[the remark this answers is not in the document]
tedzhouhk: The Spica replay parity suite was removed in 9a57231abb as requested, so this trace-packaging concern is obsolete.

components/src/dynamo/planner/tests/integration/test_spica_simulation_parity.py
```diff
@@ -0,0 +1,249 @@
...
+    ) == pytest.approx(expected)
+
+
+def test_real_static_path_preserves_goodput() -> None:
+    report, adapter_spec = _run_one("disabled")
+
+    assert adapter_spec.runtime_hooks == ()
+    assert report.metrics["goodput_output_throughput_tok_s"] > 0.0
+    assert report.metrics["gpu_hours"] > 0.0
+
```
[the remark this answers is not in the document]
tedzhouhk: The Spica replay parity suite was removed in 9a57231abb as requested, so these timeout concerns are obsolete.

components/src/dynamo/replay/simulation.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @staticmethod
+    def _synthetic_kwargs(spec: ReplaySpec) -> dict[str, JSONValue]:
+        workload = spec.workload
+        request_rate = workload.get("request_rate")
+        arrival_interval_ms = (
+            1000.0 / _float_value(request_rate, "request_rate")
+            if request_rate is not None
+            else None
+        )
```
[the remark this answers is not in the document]
tedzhouhk: Fixed in 9a57231abb: synthetic replay rejects zero and negative request_rate values with an explicit ValueError before division. Both cases have regression coverage.

components/src/dynamo/router/tests/test_router_sweep_config_provider.py
```diff
@@ -0,0 +1,141 @@
...
+from aisimulate.spica.adapter import CandidateContext, SweepContext
+from aisimulate.spica.replay import BackendDeploymentSpec
+from dynamo.router.simulation import create_adapter
+
+pytestmark = [
+    pytest.mark.pre_merge,
+    pytest.mark.unit,
+    pytest.mark.gpu_0,
+    pytest.mark.planner,
+]
```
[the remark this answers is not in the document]
tedzhouhk: No marker change: this unit test is intentionally copied into the Planner image and selected by the existing pre_merge and planner and gpu_0 CPU job. Changing it to router would remove it from CI because there is no corresponding Router CPU test job. The Planner marker is the scheduling owner here, not a path-derived component label.

components/src/dynamo/replay/simulation.py
```diff
@@ -0,0 +1,345 @@
...
+    benchmark_granularity: int = 8
+
+    def run(self, spec: ReplaySpec) -> ReplayReport:
+        """Execute one trace or synthetic replay with requested Dynamo hooks."""
+
+        self.capabilities.require_compatible(spec)
+        planner_config, router_mode, router_config = self._resolve_hooks(
+            spec.runtime_hooks
+        )
+        common: dict[str, Any] = {
```
dreamtalen, line 95: trace_format is part of ReplaySpec.workload, but it is not forwarded to run_trace_replay, so non-Mooncake input is silently interpreted using the default Mooncake format. Could we pass it through?

components/src/dynamo/replay/tests/test_simulation.py
```diff
@@ -0,0 +1,327 @@
...
+        backend_version="0.11.0",
+        agg_engine_args={"engine_type": "vllm", "max_num_seqs": 256},
+        num_workers=3,
+    )
+
+
+def test_trace_runner_preserves_current_replay_arguments(monkeypatch) -> None:
+    seen = {}
+
+    def fake_run_trace_replay(**kwargs):
```
dreamtalen, line 78: This replaces the real `test_replay_integration.py` coverage with tests mock run_trace_replay, so nothing now exercises the new adapter → ReplaySpec → DynamoReplayRunner → Rust replay boundary. Could we retain at least the former static and Planner integration cases through DynamoReplayRunnerFactory? The KVBM case can remain removed.

dreamtalen (approved):

components/src/dynamo/replay/simulation.py
tedzhouhk: Thanks for catching this. Fixed in `7efffce160`: the Dynamo runner now validates and forwards `ReplaySpec.workload.trace_format` to `run_trace_replay` for both aggregated and disaggregated deployments. The unit regression uses `trace_format: dynamo` and asserts that it reaches the Replay API.

components/src/dynamo/replay/tests/test_simulation.py
tedzhouhk: Thanks, agreed. Added focused real-Replay integration coverage in `7efffce160` without restoring the removed KVBM/full parity suite. The static and `load_180_5` cases now exercise Planner adapter materialization → `ReplaySpec` → `DynamoReplayRunnerFactory` → the Rust replay boundary. Both cases pass locally against bindings rebuilt from the merged source; the Planner case also asserts a real planner tick, goodput, and GPU-hours.

components/src/dynamo/replay/tests/test_simulation_integration.py
```diff
@@ -0,0 +1,163 @@
...
+                    "scaling_policy": [scaling_policy],
+                    "fpm_sampling": ["default"],
+                    "load_sensitivity": ["default"],
+                    "load_predictor_candidates": ["constant_last"],
+                }
+            }
+        },
+        # Slow the trace enough for load_180_5 to execute a Planner tick.
+        workload={"trace_path": _TRACE, "arrival_speedup_ratio": 0.5},
+        sweep={"max_rounds": 1, "candidates_per_round": 1, "parallel_evals": 1},
```
jasonqinzhou: Thanks for restoring the real static and Planner Replay paths. One part of the deleted parity gate is still missing: this helper bypasses `run_smart_search`, while the removed test exercised the full search loop with `parallel_evals=2`. The current process-pool tests replace `ProcessPoolExecutor`, so they do not validate factory/spec pickling, worker initialization, or real Replay execution in spawned processes. Could we retain one bounded `post_merge` test through `run_smart_search(..., runner_factory=DynamoReplayRunnerFactory())`, with an explicit timeout?

jasonqinzhou (commented):
This PR establishes a versioned cross-package ABI and moves ownership across AI Simulate, Replay, Planner, and Router. The related DEP #12441 is still `dep:draft`, while #12273 remains an open draft collaboration document marked “[not for merge].” Can we advance the DEP or otherwise record the accepted ABI and ownership decision before merging? I’d like to avoid the implementation becoming the de facto architecture before the design has been accepted.

PeaBrane (approved):

PeaBrane (approved):

aisimulate/src/aisimulate/spica/search.py
```diff
@@ -190,127 +552,76 @@ def _evaluate_one(
...
+    _WORKER_CTX.update(runner=runner)
 
 
-def _worker_eval(selection: dict[str, Any], parallel_config: Any) -> _EvalResult:
-    return _evaluate_one(selection, parallel_config, **_WORKER_CTX)
+def _worker_eval(spec: ReplaySpec) -> _ReplayResult:
+    return _run_replay(spec, _WORKER_CTX["runner"])
 
 
 def run_smart_search(
```
jasonqinzhou: `run_smart_search()` now owns substantial orchestration state and lifecycle: adapter preparation, sampler studies, caching, process-pool creation/replacement, timeouts, runner cleanup, callbacks, and result aggregation. Would it make sense to expose a `Sweeper` class that receives `RunnerFactory` and optional adapters, with `run(config)` managing one isolated sweep? The existing function could remain as a compatibility wrapper. This would give the SDK a clearer composition object and let the pool/cache/branch logic be decomposed into testable methods without changing the neutral Replay boundary.

aisimulate/src/aisimulate/spica/config.py
```diff
@@ -722,12 +486,20 @@ class SweepConfig(BaseModel):
     max_eval_seconds: float | None = Field(default=600.0, gt=0)
 
 
+class AdapterRequest(BaseModel):
```
jasonqinzhou: `AdapterRequest` only wraps an adapter-owned `search_space`; it does not identify, load, or execute an adapter, and the adapter name already comes from the enclosing `adapters` mapping key. Could we rename this to `AdapterSearchConfig` and update `SmartSearchConfig.adapters` accordingly? `Request` sounds like a runtime or lifecycle request, while this model is specifically the search configuration supplied to an adapter.

aisimulate/src/aisimulate/sweeper/__init__.py
```diff
@@ -1,16 +1,30 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Experimental replay-backed smart sweeper for Dynamo deployments.
+"""Experimental, backend-neutral deployment simulation sweeper.
```
jasonqinzhou, line 4: can we rename spica to sweeper and rename the folder too?

aisimulate/src/aisimulate/spica/adapter.py
```diff
@@ -0,0 +1,141 @@
...
+@dataclass(frozen=True)
+class AdapterReplaySpec:
+    """Concrete adapter configuration and hooks for one replay candidate."""
+
+    config: dict[str, JSONValue] = field(default_factory=dict)
+    runtime_hooks: tuple[RuntimeHookSpec, ...] = ()
+
+
+@runtime_checkable
+class SimulationAdapter(Protocol):
```
jasonqinzhou: `SimulationAdapter` is a search-time configuration provider: it validates component-owned sweep configuration, contributes optimizer dimensions, and materializes the selected configuration into runtime hooks; it does not execute the simulation itself. Could we rename this contract to `SweepConfigProvider` (and the concrete implementations to `DynamoPlannerSweepConfigProvider` / `DynamoRouterSweepConfigProvider`)? This matches the standalone roadmap terminology, makes the search-time/runtime-time split clearer alongside `ReplayProvider`, and avoids overloading “adapter” across the sweep, replay, Router, and Planner layers.

aisimulate/README.md
```diff
@@ -12,32 +12,48 @@ SPDX-License-Identifier: Apache-2.0
...
+    runner_factory=DynamoReplayRunnerFactory(),
+)
+```
+
+The standalone module validates the backend-neutral core schema but intentionally has no implicit
+replay runtime. Adapter-owned search spaces are validated when the selected adapters are resolved
+by `run_smart_search`.
+KVBM sweep fields have been removed; native G2 is their replacement.
+
 Read the [Spica documentation](../docs/fern/pages/developer-guide/knowledge-base/modular-components/ai-simulate-experimental/spica-experimental/overview.md)
```
jasonqinzhou: Could we make `aisimulate/` the canonical home for the Dynamo-neutral documentation instead of linking the package README back into the Dynamo Fern tree? I suggest moving the AI Simulate overview, quickstart, architecture, and the Sweeper documentation (tutorial, configuration/search space, optimization goals, traffic, results, and a `SweepConfigProvider` extension guide) under `aisimulate/docs/`, with neutral examples under `aisimulate/examples/`. Fern can remain the publishing/navigation layer, but it should consume or sync from the package-owned source rather than becoming a second source of truth. The Dynamo docs should retain only Dynamo-owned integration material such as the Planner/Router providers, `DynamoReplayRunnerFactory`, Planner-image installation, and Dynamo-specific examples. Please include Fern navigation updates, link checking, redirects for the current URLs, and doc-build/sync validation as part of the move. This will make eventual extraction into a standalone repository mostly mechanical.

pyproject.toml
```diff
@@ -89,6 +89,10 @@ vllm_tests = "dynamo.vllm.tests.conftest"
 trtllm_tests = "dynamo.trtllm.tests.conftest"
 sglang_tests = "dynamo.sglang.tests.conftest"
 
+[project.entry-points."aisimulate.adapters"]
```
jasonqinzhou: Since the roadmap includes a separate `dynamo-sim` wheel and Rust crate, could we make that package the long-term owner of these provider registrations instead of the root `ai-dynamo` wheel? The component-local adapters are reasonable for this PR as an interim step, but I suggest capturing this target layout:

```text
dynamo-sim/
├── pyproject.toml
├── src/dynamo_sim/
│   ├── providers/
│   │   ├── planner.py
│   │   └── router.py
│   ├── replay.py
│   ├── discovery.py
│   └── _native/
├── tests/
└── docs/

lib/dynamo-sim/
├── Cargo.toml
└── src/
    ├── lib.rs
    ├── planner.rs
    └── router.rs

components/src/dynamo/planner/   # Planner models and behavior
components/src/dynamo/router/    # Router models and behavior
```

The component directories would continue to own Planner/Router configuration, policies, and implementation. `dynamo-sim` would own the thin `SweepConfigProvider` and `ReplayProvider` wrappers, runtime-hook translation, entry-point registration, native bindings, and integration tests. `aisimulate` would own only the neutral contracts and would never import `dynamo-sim`. This gives the wheel/crate one coherent build and release boundary, avoids multiple wheels installing overlapping `dynamo.*` files, and still lets the component teams own the relevant provider files through `CODEOWNERS`. Would you be open to tracking this as the target packaging layout or a follow-up to this PR?

jasonqinzhou:
The agreed artifact model introduces both a `dynamo-sim` Python wheel and a `dynamo-sim` Rust crate. Could this PR create the `dynamo-sim` wheel project and move the new Python Planner, Router, Replay-provider code, tests, and provider registration into it?

For the Rust crate, if Dynamo-specific native adapter implementation is already in scope, could we add it under `lib/dynamo-sim/` as well? If no native implementation is ready yet, please create a concrete follow-up issue rather than adding an empty crate.

pyproject.toml
tedzhouhk: based on discussion, we will not have a separate `dynamo-sim` wheel.

aisimulate/src/aisimulate/spica/search.py
tedzhouhk: Thanks — implemented. `Sweeper` owns the injected `RunnerFactory` and optional providers, while each `run(config)` call creates fresh studies, caches, runners, and worker-pool state. Based on the follow-up API decision, the redundant `run_smart_search` wrapper was removed and the class is the sole public execution interface.

aisimulate/src/aisimulate/sweeper/__init__.py
tedzhouhk: Done — the Python package and on-disk folder are now `aisimulate.sweeper`. Imports, CLI, tests, examples, documentation, and wheel packaging were migrated; the packaging test also verifies that the wheel contains no `aisimulate.spica` package.

aisimulate/README.md
tedzhouhk: Thanks — implemented. Backend-neutral Sweeper docs and examples now live canonically under `aisimulate/docs/sweeper` and `aisimulate/examples/sweeper`. Fern keeps checked-in synchronized copies, while Dynamo-owned integration and Planner/Router example pages remain in the Dynamo docs tree. The sync check validates the canonical set, Fern copies, and navigation registration. Published Fern paths were preserved, so the existing URLs do not require redirects.

jasonqinzhou (approved):

jasonqinzhou:
Approved. I do not think the remaining review follow-ups should expand this already-large PR, and none of them need to be fixed in #12446.

I split the technical follow-ups into draft PR #12849, stacked on this PR's exact head:

- map Pareto goals containing goodput to the Planner's SLA optimization target while preserving default Pareto behavior
- verify the built wheel contains no legacy `aisimulate.spica` package
- make the existing real two-worker Replay integration test gate pre-merge CI as well as run post-merge

The DEP status and outstanding CODEOWNER review requirements remain merge-process items for #12446 rather than code changes for the follow-up. No further code action is requested on this PR for the three items above.

dagil-nvidia (approved):

tests/report_pytest_markers.py
```diff
@@ -283,6 +283,9 @@
     "aiconfigurator_core.sdk",
     "aiconfigurator_core.sdk.engine",
     "aiconfigurator_core.sdk.memory",
+    "aiconfigurator_core.sdk.models",
+    "aiconfigurator_core.sdk.perf_database",
+    "aiconfigurator_core.sdk.utils",
```
dmitry-tokarev-nv, line 288: **These three also need to be in `FORCE_STUB_MODULES`.**

`ensure_available()` (L445) returns the *real* module whenever `importlib.import_module()` succeeds, so a `STUB_MODULES`-only entry is inert against an older-but-importable `aiconfigurator_core`. That is exactly the version-drift case the comment at L298-300 describes, and exactly why `.engine` and `.memory` are force-stubbed.

Two importers make it reachable:

- `aisimulate/src/aisimulate/sweeper/kv_estimate.py:30` --- `from aiconfigurator_core.sdk.perf_database import get_latest_database_version`
- `aisimulate/src/aisimulate/sweeper/model_hw.py:26` --- `from aiconfigurator_core.sdk.models import check_is_moe`

Both raise `ImportError` from *inside* the importing module, which the stubber never sees.

Reproduced with a synthetic older AIC on `PYTHONPATH`: 10 collection errors, `Interrupted: 10 errors during collection`, exit 2 --- i.e. the `pytest-marker-report` pre-commit hook fails. With the three entries added to `FORCE_STUB_MODULES`: exit 0, and 258 tests collected instead of 150.

`FORCE_STUB_MODULES` (L301-306) is outside this diff so GitHub will not take a suggestion there. The fix:

```python
FORCE_STUB_MODULES = {
    "aiconfigurator.sdk.task_v2",
    "aiconfigurator.cli.main",
    "aiconfigurator_core.sdk.engine",
    "aiconfigurator_core.sdk.memory",
    "aiconfigurator_core.sdk.models",
    "aiconfigurator_core.sdk.perf_database",
    "aiconfigurator_core.sdk.utils",
}
```

pyproject.toml
```diff
@@ -370,6 +374,13 @@ ignore_errors = true
 module = ["dynamo.planner.manual.*"]
 ignore_errors = true
 
+[[tool.mypy.overrides]]
+# AI Simulate is a separate, optional distribution. Check Dynamo's adapters,
+# but do not recursively analyze the standalone package from this mypy target.
+module = ["aisimulate", "aisimulate.*"]
+follow_imports = "skip"
```
dmitry-tokarev-nv: **`follow_imports = "skip"` gives the new adapter ABI zero static checking** --- the opposite of what L378-379 says this override does.

Every `aisimulate` symbol becomes `Any`. Minimal repro under this exact override: a call with wrong arity **and** wrong argument type **and** a nonexistent attribute on the return value produces `Success: no issues found`. Changing only `skip` to `silent` catches all three.

So `AdapterSearchPlan`, `AdapterReplaySpec`, `SweepContext`, `CandidateContext`, `RuntimeHookSpec`, `SearchSpaceFragment` and `JSONValue` (imported at `components/src/dynamo/planner/simulation/provider.py:16-24` and `components/src/dynamo/router/simulation/provider.py:13`) are all unchecked. Combined with duck-typed providers (no ABC base) and string-resolved entry points, an ABI rename lands green and surfaces only at runtime.

`ignore_missing_imports = true` already covers CI, where aisimulate is not on the mypy job's `MYPYPATH`. So `skip` only takes effect in a planner dev venv where aisimulate *is* installed --- precisely where you want it checking.

```suggestion
follow_imports = "silent"
```

container/templates/planner.Dockerfile
```diff
@@ -80,12 +80,23 @@ RUN --mount=type=bind,source=./container/deps/requirements.planner.txt,target=/t
...
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation.py /workspace/components/src/dynamo/replay/tests/test_simulation.py
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation_integration.py /workspace/components/src/dynamo/replay/tests/test_simulation_integration.py
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/router/tests/test_router_sweep_config_provider.py /workspace/components/src/dynamo/router/tests/test_router_sweep_config_provider.py
+COPY --chmod=775 --chown=dynamo:0 aisimulate/examples /workspace/aisimulate/examples
 COPY --chmod=775 --chown=dynamo:0 aisimulate/tests /workspace/aisimulate/tests
+RUN chmod 775 \
+    /workspace/components/src/dynamo/replay \
+    /workspace/components/src/dynamo/replay/tests \
+    /workspace/components/src/dynamo/router \
+    /workspace/components/src/dynamo/router/tests
```
dmitry-tokarev-nv: **`--chmod=664` also applies to the parent directories BuildKit creates implicitly, and the `RUN chmod` below is a hardcoded patch for that.**

Verified with real BuildKit builds: these three COPYs produce `drw-rw-r-- 1000:0` on `.../replay` and `.../replay/tests` --- mode 0664 on a *directory*, so no search bit. The image is correct today: ownership is `1000:0`, the `RUN chmod` under `USER dynamo` succeeds, and I confirmed the final stage's `COPY --chown=1000:0 --from=planner_builder /workspace /workspace` preserves modes.

The risk is that the RUN is load-bearing with a hardcoded path list. In the same experiment, placing a `--chmod=664` COPY first produced `drw-rw-r--` on `/workspace/components`, `/workspace/components/src` and `/workspace/components/src/dynamo`, and a later `--chmod=775` COPY did **not** repair its own parents. So an alphabetical sort of this block, or a fourth 664 file COPY under a new subpackage, silently leaves unsearchable directories: no build step fails, and it surfaces only as mass pytest collection errors (EACCES) in the shipped image --- including for arbitrary OpenShift UIDs with gid 0.

Copying these at 775 like every sibling line removes the failure mode and the patch together. The `router` path copied at 775 in the same experiment produced `drwxrwxr-x` parents and was readable with no RUN at all.

```suggestion
COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation.py /workspace/components/src/dynamo/replay/tests/test_simulation.py
COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation_integration.py /workspace/components/src/dynamo/replay/tests/test_simulation_integration.py
COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/router/tests/test_router_sweep_config_provider.py /workspace/components/src/dynamo/router/tests/test_router_sweep_config_provider.py
COPY --chmod=775 --chown=dynamo:0 aisimulate/examples /workspace/aisimulate/examples
COPY --chmod=775 --chown=dynamo:0 aisimulate/tests /workspace/aisimulate/tests
```

pyproject.toml
```diff
@@ -89,6 +89,10 @@ vllm_tests = "dynamo.vllm.tests.conftest"
 trtllm_tests = "dynamo.trtllm.tests.conftest"
 sglang_tests = "dynamo.sglang.tests.conftest"
 
+[project.entry-points."aisimulate.sweep_config_providers"]
+"dynamo.planner" = "dynamo.planner.simulation:create_provider"
+"dynamo.router" = "dynamo.router.simulation:create_provider"
```
dmitry-tokarev-nv, line 95: **These entry points expose a module that imports an undeclared dependency.**

`dynamo.planner.simulation` imports `tqdm` at module scope --- `components/src/dynamo/planner/simulation/provider.py:14` and `load_predictor.py:13` --- but `tqdm` appears in neither `[project] dependencies` nor `container/deps/requirements.planner.txt`. I checked the built wheel's METADATA: no `Requires-Dist: tqdm`.

It resolves today only because `aisimulate/pyproject.toml` pins `tqdm>=4.0.0` and aisimulate is co-installed in the planner image --- an implicit cross-distribution contract with nothing enforcing it. `pip install ai-dynamo` followed by `from dynamo.planner.simulation import create_provider` fails with `ModuleNotFoundError: No module named 'tqdm'`, before reaching the intended "aisimulate not installed" error. Any future reordering that installs ai-dynamo without aisimulate breaks the same way.

`[project] dependencies` (L15-27) is outside this diff, so as a plain patch:

```toml
    "typing_extensions>=4.10.0",
    "tqdm>=4.0.0",
]
```

.github/codeowners/areas.yaml
```diff
@@ -559,6 +558,11 @@ areas:
   - tests/serve/multimodal_profiles/vllm_xpu.py
   - tests/serve/test_vllm_xpu.py
 shared:
+- glob: aisimulate/
+  owners:
+  - planner
+  - '@jasonqinzhou'
+  - '@dreamtalen'
```
dmitry-tokarev-nv: **This is the only `shared:` entry that grants ownership to individuals by copying an area's globs.**

`.github/codeowners/README.md:54-56`: "An external individual who has earned ownership of an area is granted it by attaching them to that area's **label** in `external_contributors.yaml` --- never by copying its globs". And L102-104: "Base area owners are GitHub **teams**. The one exception is `external_contributors.yaml`, which appends named individuals as area-scoped co-owners." Every other shared entry pairs two area labels.

Three consequences:

1. Neither individual reaches `CONTRIBUTORS.md`, which the emitter generates from `external_contributors.yaml` (currently `contributors: []`).
2. They do not inherit future `planner` globs, so every `aisimulate/` path split has to re-copy the glob by hand.
3. `shared:` is "any one approves" and CODEOWNERS is last-match-wins, so `CODEOWNERS:599` now lets a single individual approve the entire `aisimulate/` tree with no `planner-codeowners` member --- and the team-only rule at `CODEOWNERS:164` no longer applies to anything. Confirmed three times with `who_owns.py`.

To be clear, this is a policy-shape issue and not drift: the generated file is byte-identical to `emit_codeowners.py` output and `build_codeowners.py --strict` reports 5224/5224.

.github/filters.yaml
```diff
@@ -211,7 +211,6 @@ planner:
   - 'components/src/dynamo/profiler/**'
   - 'components/src/dynamo/global_router/**'
   - 'aisimulate/**'
-  - 'examples/aisimulate/**'
   # Power-aware planner (PR #9369): per-node power agent component
   - 'deploy/power-agent/**'
   - 'deploy/helm/charts/power-agent/**'
```
dmitry-tokarev-nv, line 216: **`planner:` is the only filter without the `!**/*.md` / `!**/*.rst` negation, and this PR moves 9 markdown docs into `aisimulate/`.**

After this PR, a one-word typo fix in `aisimulate/docs/sweeper/traffic.md` sets `planner == 'true'`, gating `planner-build` amd64+arm64 (`pr.yaml:623-633`), `planner-test` (847-866), planner Compliance, and the profiler deploy job (656-666).

Measured on this PR's own run 31210466394: 17m02s multi-arch build + 2m57s/3m46s CPU test + 42s/55s compliance, so roughly **25.5 runner-minutes per docs-only push** --- on the builder pool that is already the known contention point. The coverage gain is zero: the `docs:` filter (L30, `'**/*.md'`) already matches those files.

`core`, `operator`, `vllm`, `sglang`, `trtllm` and `frontend` all carry the negation. The engine agrees it would behave identically here: `pr.yaml` uses `tj-actions/changed-files@v42.1.0`, whose list-filter form is micromatch's accumulate-then-omit, left to right, so a trailing negation works exactly as in those sibling blocks.

The new docs-sync hook makes this bite more often, since it forces a paired Fern-copy edit in the same commit.

```suggestion
  - 'aisimulate/**'
  # Power-aware planner (PR #9369): per-node power agent component
  - 'deploy/power-agent/**'
  - 'deploy/helm/charts/power-agent/**'
  - '!**/*.md'
  - '!**/*.rst'
```

container/templates/wheel_builder.Dockerfile
```diff
@@ -596,9 +596,9 @@ RUN --mount=type=secret,id=aws-web-identity-token,target=/run/secrets/aws-token
 {% endif %}    /tmp/use-sccache.sh show-stats "Dynamo Runtime"
 
 {% if target == "planner" %}
-# AI Simulate is a separate Python distribution. Build it only for the planner
-# image, after the Dynamo wheels so Python-only changes do not invalidate the
-# expensive Rust build layers above.
+# AI Simulate is a separate Python distribution used by the planner image. Build
+# it after the Dynamo wheels so Python-only changes do not invalidate the
+# expensive Rust build layers above. This wheel remains an image-local artifact.
 COPY aisimulate/ /opt/dynamo/aisimulate/
```
dmitry-tokarev-nv, line 602: **This COPY now drags the new `aisimulate/docs/` and `aisimulate/examples/` trees into the layer that gates the wheel build**, which undercuts the ordering rationale in the comment directly above.

`uv build --wheel` consumes only `src/aisimulate` plus pyproject/LICENSE/README (`aisimulate/pyproject.toml:58-59`), and `.dockerignore` excludes nothing under `aisimulate/`. So editing any of the 9 canonical markdown files busts this layer and everything downstream: the `uv build` RUN, the cargo-registry license harvest RUN, `COPY container/compliance`, the `bundle_wheel_notices` RUN, and then both compliance stages --- `container/render.py:237-245` returns `planner_builder` as `compliance_base_stage` for `target == "planner"`, and `compliance.Dockerfile:64/171` are `FROM {{ compliance_base_stage }}`. That whole cascade re-runs on both architectures for a documentation typo.

Adding `aisimulate/docs/` to `.dockerignore` is enough. Note that `aisimulate/examples/` has to stay in the build context --- `planner.Dockerfile:93` copies it.

.pre-commit-config.yaml
```diff
@@ -80,6 +80,12 @@ repos:
 # and give it visiblity into all the local and third_party packages expected.
 - repo: local
   hooks:
+    - id: sync-aisimulate-docs
+      name: Check canonical AI Simulate docs are synced to Fern
+      entry: python3 docs/fern/scripts/sync_aisimulate_docs.py --check
```
dmitry-tokarev-nv, line 85: **Running this script to satisfy the hook silently discards any edit made to the Fern copy.**

A docs contributor fixes a typo in `docs/fern/pages/developer-guide/knowledge-base/modular-components/ai-simulate-experimental/sweeper-experimental/results.md`, the hook fails with "AI Simulate Fern copies are out of sync" plus "Run `python3 docs/fern/scripts/sync_aisimulate_docs.py`", they run it, and their staged edit is replaced by the canonical `aisimulate/docs/sweeper/results.md` content. The loss is invisible in the diff they staged.

Nothing in the destination directory distinguishes the 9 auto-synced files from the 4 hand-owned ones: neither copy carries a "generated --- do not edit" banner, and the only discriminator is the `DOCUMENTS` / `DYNAMO_OWNED_DOCUMENTS` tuples in a script three directories away. At minimum a generated-file banner on the synced copies would make this self-evident at the point of editing.

Worth noting the repo's existing mechanism for this shape is a symlink (`skills/` to `.agents/skills/`) rather than a byte-copy plus checker, and the documentation style guide says not to duplicate content across pages.

.pre-commit-config.yaml
```diff
@@ -80,6 +80,12 @@ repos:
...
+      language: system
+      files: ^(aisimulate/docs/sweeper/|docs/fern/pages/developer-guide/knowledge-base/modular-components/ai-simulate-experimental/sweeper-experimental/.*\.md$|docs/fern/(index\.yml|scripts/sync_aisimulate_docs\.py)$)
```
dmitry-tokarev-nv: **This `files:` regex can never fire on a deletion, which is the main way the sync drifts.**

pre-commit builds its file list with `git diff --staged --diff-filter=ACMRTUXB` (D excluded), so a commit that only deletes a file matches nothing and the hook is skipped. `git rm aisimulate/docs/sweeper/results.md && git commit` lands with an orphaned Fern copy plus a live `path:` entry in `docs/fern/index.yml` --- exactly the state `_integrity_errors()` rejects with "Fern documents are neither canonical copies nor Dynamo-owned". Same for removing a `DYNAMO_OWNED_DOCUMENTS` entry.

Low severity: CI runs `pre-commit/action` with `--all-files` (`pre-merge.yml:64`), so this is a local-commit-only gap --- the contributor just gets the failure later than they should.

container/templates/planner.Dockerfile
```diff
@@ -80,12 +80,23 @@ RUN --mount=type=bind,source=./container/deps/requirements.planner.txt,target=/t
...
-# runtime code comes from the wheel installed above; copy only its tests.
+# runtime code comes from the wheel installed above. The Router adapter and
+# replay bridge unit tests also run here because this image installs that wheel;
+# copy the AI Simulate examples exercised by those tests as test fixtures.
 COPY --chmod=664 --chown=dynamo:0 pyproject.toml /workspace/pyproject.toml
 COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/planner /workspace/components/src/dynamo/planner
 COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/profiler /workspace/components/src/dynamo/profiler
 COPY --chmod=775 --chown=dynamo:0 components/src/dynamo/global_planner /workspace/components/src/dynamo/global_planner
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation.py /workspace/components/src/dynamo/replay/tests/test_simulation.py
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/replay/tests/test_simulation_integration.py /workspace/components/src/dynamo/replay/tests/test_simulation_integration.py
```
dmitry-tokarev-nv: **The test this line copies is `post_merge`-only, so the one new code path that spawns worker processes never runs pre-merge.**

`test_sweeper_runs_real_dynamo_replay_in_spawned_workers` is marked `post_merge`, while the PR-lane filter is `pre_merge and planner and gpu_0`. I grepped both planner CPU Test job logs for this SHA (jobs 92979337378 amd64, 92979337273 arm64): zero occurrences --- it did not run, while the two `pre_merge` siblings did and passed.

So the only new path that forks a `ProcessPoolExecutor` (`parallel_evals=2`, `max_eval_seconds: 240`, `@pytest.mark.timeout(300)`) ships validated only by a local run. The post-merge planner job uses `cpu_parallel_mode: none` with a 30-minute cap on both arches, and this repo's orchestrator does not hard-kill hung children, so a hang there burns the full cap twice. Marking it `pre_merge` as well would be cheap insurance.

container/templates/planner.Dockerfile
```diff
@@ -80,12 +80,23 @@ RUN --mount=type=bind,source=./container/deps/requirements.planner.txt,target=/t
...
+COPY --chmod=664 --chown=dynamo:0 components/src/dynamo/router/tests/test_router_sweep_config_provider.py /workspace/components/src/dynamo/router/tests/test_router_sweep_config_provider.py
+COPY --chmod=775 --chown=dynamo:0 aisimulate/examples /workspace/aisimulate/examples
```
dmitry-tokarev-nv: **The whole 8-file examples tree is copied for one fixture, and it ships in the released image.**

Only `aisimulate/examples/sweeper/tools/run_sweep.py` (38 lines) is a fixture (`test_simulation.py:29-31`). Note there is a second, different `aisimulate/examples/sweeper/run_sweep.py` (58 lines): editing the wrong one silently has no effect on the test, and a bad edit to the `tools/` one becomes a *collection* error rather than a test failure, because it is `exec_module`'d at import time.

The other seven files come along, and L119 `COPY --chown=1000:0 --from=planner_builder /workspace /workspace` ships `/workspace` wholesale into the runtime image --- so example configs and READMEs land in the released planner image and in the compliance/SBOM scan at L106. `--chmod=775` also makes them world-executable (verified: `-rwxrwxr-x 1000:0 .../examples/README.md`). Touching any of them invalidates this layer and the four COPYs after it.

I did not verify whether `aisimulate/tests` (L94) reads anything else under `examples/`, so I am not attaching a narrowing suggestion --- but if it does not, copying just the one fixture file would be tighter.

dmitry-tokarev-nv (commented):
## Scoped review: `container/`, `.github/`, `tests/` and top-level files

This review deliberately covers only the CI, container and packaging surfaces of this PR. I did not review `lib/`, `components/`, `docs/`, `deploy/`, `examples/` or `recipes/`, so treat it as partial.

13 findings, each reproduced rather than inferred. 11 are inline; the two below anchor to lines outside the diff, so they live here.

The two I would treat as blocking are both inline: the missing `FORCE_STUB_MODULES` entries in `tests/report_pytest_markers.py`, and `follow_imports = "skip"` in the new mypy override. Everything else is cleanup, CI efficiency, or a latent landmine.

---

### `pyproject.toml:125` --- the wheel ships a test module that fails at import time

`[tool.hatch.build.targets.wheel]` has no `exclude`, so `ai_dynamo-1.4.0-py3-none-any.whl` contains 351 files under `tests/`, now including the new `dynamo/replay/tests/test_simulation.py`.

Shipping tests is pre-existing and deliberate (`[project.entry-points.pytest11]` references `dynamo.vllm.tests.conftest` and friends). What is new is that this is the first shipped test module doing unguarded filesystem work at import time: `components/src/dynamo/replay/tests/test_simulation.py:28-35` runs `spec_from_file_location(...)` then `loader.exec_module(...)` at module scope, on `Path(__file__).resolve().parents[5] / "aisimulate/examples/sweeper/tools/run_sweep.py"`.

I extracted the built wheel into a site-packages layout: `parents[5]` resolves to `<venv>/lib`, the fixture is not there, and `exec_module` raises `FileNotFoundError` --- not `ImportError`, so it escapes the `pytest.importorskip("aisimulate.sweeper")` guard on L19 and becomes a hard collection error. The `assert _RUN_SWEEP_SPEC is not None and _RUN_SWEEP_SPEC.loader is not None` on L32 gives false assurance: `spec_from_file_location` never stats the path, and returns a non-None spec with a non-None loader for a path that does not exist.

Not reachable in this repo's CI --- there is no `--pyargs` anywhere, and pytest always runs from `/workspace` where `parents[5]` is correct. It hits downstream consumers of the published wheel. Either exclude tests from the wheel, or make the fixture-path resolution tolerate absence.

### `.github/filters.yaml:130` --- `core:` lists a path that does not exist

`'scripts/report_pytest_markers.py'`. The real file is `tests/report_pytest_markers.py` --- the one this PR modifies. `git ls-tree` finds no `scripts/` copy, and the root-level `'*.py'` on L128 does not cross `/` under micromatch, so this entry matches nothing.

Pre-existing and harmless today, since coverage survives via `'tests/**'` on L113. Flagging it only because it sits in the filter file this diff touches and names the very script this diff edits, so it will silently stop covering it if `tests/**` is ever narrowed.

---

<details>
<summary>Checked and refuted --- deliberately not reported</summary>

- The `RUN chmod` in `planner.Dockerfile` failing with EPERM (ownership is `1000:0`, chmod succeeds, the build is green)
- `sys.path` shadowing of the partial `replay/` and `router/` trees in the planner image
- Missing fixtures or collection errors in the planner image as actually built
- `tqdm` or stub gaps breaking the marker hook in the *current* environment
- CODEOWNERS generator drift (output byte-identical, `--strict` 5224/5224)
- A wheel-glob collision between `aisimulate*.whl` and `ai_dynamo*any.whl`
- Any leak of the unpublished aisimulate wheel into release artifacts
- Stale `examples/aisimulate` references after the move
- New pytest module-name collisions in the planner image

</details>

<details>
<summary>How this was verified</summary>

Verified against `71f7c0cfa8`. All 7 files referenced above are byte-identical at current head `c541201` --- the only change since is the merge of `main`.

Built both wheels and inspected contents, METADATA and entry points; reproduced the `FORCE_STUB` failure with a synthetic older AIC on `PYTHONPATH`; reproduced the mypy blind spot with a minimal case; reproduced BuildKit's `--chmod`-on-implicit-parent-directories behaviour across 4 real container builds; reproduced the site-packages `parents[5]` breakage; ran `emit_codeowners.py`, `build_codeowners.py --strict`, `who_owns.py`, `report_pytest_markers.py` (exit 0, 4752 tests) and `sync_aisimulate_docs.py --check` (exit 0); resolved the pre-commit regex against 10 representative paths; and read both planner CPU Test job logs (1715 passed / 4 skipped per arch), confirming which new tests actually executed rather than skipped.

</details>

---

*Automated review (Claude Code), scoped to CI/container/packaging files. Findings were reproduced locally, but please push back on anything that does not match your intent.*

tests/report_pytest_markers.py
tedzhouhk: Thanks for reproducing the version-drift case. Fixed in 1dd57b17d4 by adding models, perf_database, and utils to FORCE_STUB_MODULES. The Report pytest markers pre-commit hook now passes with the expanded forced stubs.

pyproject.toml
tedzhouhk: Thanks, agreed. Changed follow_imports from skip to silent in 1dd57b17d4 and updated the comment to describe the behavior accurately. Mypy passes for the Planner and Router provider modules with the current AI Simulate source on MYPYPATH.

container/templates/planner.Dockerfile
tedzhouhk: Thanks for the BuildKit reproduction. Changed all three single-file test COPY instructions to mode 775 and removed the load-bearing hardcoded chmod block in 1dd57b17d4.

pyproject.toml
tedzhouhk: Thanks, fixed in 1dd57b17d4 by declaring tqdm>=4.0.0 in the ai-dynamo project dependencies. I built the wheel and confirmed its METADATA contains Requires-Dist: tqdm>=4.0.0.

.github/codeowners/areas.yaml
tedzhouhk: Thanks for pointing to the ownership policy. Removed the individual handles from shared ownership and registered Jason Zhou and Yongming Ding through external_contributors.yaml on the Planner area, then regenerated CODEOWNERS and CONTRIBUTORS.md. The strict ownership check reports 5226/5226 files explicitly owned.

.github/filters.yaml
tedzhouhk: Thanks, fixed in 1dd57b17d4 by adding the Markdown and RST negations at the end of the Planner filter. I verified with micromatch that an AI Simulate Markdown file no longer selects Planner while an AI Simulate Python source file still does.

container/templates/wheel_builder.Dockerfile
tedzhouhk: Thanks. Added aisimulate/docs/ to .dockerignore in 1dd57b17d4 so canonical documentation changes no longer invalidate the AI Simulate wheel and compliance layers. The Planner test fixture copy is also narrowed to the one run_sweep.py file it uses.

.pre-commit-config.yaml
tedzhouhk: Thanks, fixed in 1dd57b17d4. The sync generator now inserts a visible HTML comment after frontmatter in every generated Fern copy, naming the canonical AI Simulate source and instructing contributors to edit it instead.

.pre-commit-config.yaml
tedzhouhk: Thanks, fixed in 1dd57b17d4 by making the sync hook always_run and removing the misleading files regex. I verified the hook runs and passes even when invoked with only an unrelated README.md path.

container/templates/planner.Dockerfile
tedzhouhk: Thanks. This was addressed by the earlier follow-up commit already on the current head: test_sweeper_runs_real_dynamo_replay_in_spawned_workers now has both pre_merge and post_merge markers. The current targeted suite, including this test, passes.

container/templates/planner.Dockerfile
tedzhouhk: Thanks, fixed in 1dd57b17d4. The Planner image now copies only aisimulate/examples/sweeper/tools/run_sweep.py instead of the full eight-file examples tree, while preserving the path expected by the replay test.

tedzhouhk:
@dmitry-tokarev-nv Thank you for the detailed, reproduced CI/container/packaging review. I addressed the full set in 1dd57b17d4 and replied to/resolved each inline thread.

The two review-body findings are also fixed:

- The shipped dynamo/replay/tests/test_simulation.py now skips module collection when the repo-only AI Simulate example fixture is absent. I built and extracted ai_dynamo-1.4.0-py3-none-any.whl and verified the installed-layout module reaches that clean skip instead of FileNotFoundError.
- The core filter now names tests/report_pytest_markers.py rather than the nonexistent scripts path.

Validation completed:

- 284 AI Simulate Sweeper, Planner/Router provider, and Dynamo Replay tests passed.
- 86 CODEOWNERS tests passed; strict ownership is 5226/5226.
- Planner and Router provider mypy checks passed with the current AI Simulate sources on MYPYPATH.
- All touched-file pre-commit hooks passed, including marker reporting and the always-run docs sync check.
- The built ai-dynamo wheel declares tqdm>=4.0.0.
- Micromatch verification confirms AI Simulate Markdown no longer triggers Planner CI while Python changes still do.

One policy consequence is now explicit: moving Jason and Yongming into external_contributors.yaml attaches them to the Planner area, so the generated policy grants Planner-area co-ownership rather than only the AI Simulate glob. This follows the repository ownership model described in the review.

PR #12447, opened 2026-07-30
refactor(runtime): extract PushRouter transport seam behind StreamingDispatch trait

tanmayv25:
## Summary

Extract the final-hop transport in `PushRouter` behind a new
`StreamingDispatch<T, U>` trait. After `PushRouter` selects a worker, the
final hop goes through this trait object instead of a fixed
`AddressedPushRouter`. `PushRouter` keeps instance selection, occupancy, fault
detection, and migration; only the transport below the seam can now be swapped.
`AddressedPushRouter` (the request plane) is the default implementation and its
behavior is unchanged.

This is a **behavior-preserving refactor with no new functionality** — it makes
the final-hop transport pluggable behind a single typed seam. No alternate
transport implementation is included here.

## Changes

- New `StreamingDispatch<T, U>` trait (`generate`, `generate_bidirectional`,
  `on_instance_removed` / `on_instance_added`), with `AddressedPushRouter` as the
  default impl delegating to its existing `AsyncEngine` path.
- `PushRouter.addressed`: `Arc<AddressedPushRouter>` -> `Arc<dyn StreamingDispatch<T, U>>`.
- `spawn_instance_removal_watcher` is generic and drives discovery-removal
  cleanup through the trait.
- New `PushRouter::from_client_with_dispatch` to inject a custom dispatch
  (unused in this PR).
- Rename inherent `AddressedPushRouter::generate_bidirectional` ->
  `dispatch_bidirectional` (disambiguates from the trait method); make
  `AddressedRequest::into_parts` public.

## Validation

- `cargo build -p dynamo-runtime` and `-p dynamo-llm` (downstream) compile.
- `cargo test -p dynamo-runtime` egress suite: **67 passed, 0 failed** — the
  request-plane path is exercised unchanged.

---

> Enables pluggable final-hop transports behind the router seam for future work.

### DIFF
--- a/lib/runtime/src/pipeline/network/egress/push_router.rs
+++ b/lib/runtime/src/pipeline/network/egress/push_router.rs
@@ -13,7 +13,7 @@ use crate::{
     engine::{AsyncEngine, AsyncEngineContext, Data},
     metrics::frontend_perf::{STAGE_DURATION_SECONDS, STAGE_ROUTE},
     pipeline::{
-        AddressedPushRouter, AddressedRequest, Error, ManyIn, ManyOut, SingleIn,
+        AddressedPushRouter, AddressedRequest, Error, ManyIn, ManyOut, SingleIn, StreamingDispatch,
         error::{PipelineError, PipelineErrorExt},
     },
     protocols::{EndpointId, maybe_error::MaybeError},
@@ -149,9 +149,10 @@ where
     /// Number of round robin requests handled. Used to decide which server is next.
     round_robin_counter: Arc<AtomicU64>,
 
-    /// The next step in the chain. PushRouter (this object) picks an instances,
-    /// addresses it, then passes it to AddressedPushRouter which does the network traffic.
-    addressed: Arc<AddressedPushRouter>,
+    /// The final hop: after selecting an instance, `PushRouter` hands it to this
+    /// `StreamingDispatch` (the request-plane `AddressedPushRouter` by default).
+    /// A trait object so an alternate transport can swap it out.
+    addressed: Arc<dyn StreamingDispatch<T, U>>,
 
     /// When false, `generate_with_fault_detection` skips fault detection logic:
     /// it won't call `report_instance_down` on errors, and it uses the raw discovery
@@ -320,11 +321,14 @@ static ENDPOINT_CACHE_INDEXER_WATCHER_ACTIVE: std::sync::OnceLock<
 /// a migratable `Disconnected` error. Uses raw `list_and_watch` events
 /// (not a coalesced snapshot diff) so a rapid remove→re-add of the same
 /// identity is not silently swallowed. Keyed by full `EndpointInstanceId`.
-fn spawn_instance_removal_watcher(
+fn spawn_instance_removal_watcher<T, U>(
     endpoint: Endpoint,
-    addressed: Arc<AddressedPushRouter>,
+    dispatch: Arc<dyn StreamingDispatch<T, U>>,
     cancel_token: tokio_util::sync::CancellationToken,
-) {
+) where
+    T: Data + Serialize + 'static,
+    U: Data + for<'de> Deserialize<'de> + MaybeError + 'static,
+{
     use crate::discovery::{
         DiscoveryEvent, DiscoveryInstance, DiscoveryInstanceId, DiscoveryQuery,
     };
@@ -388,23 +392,12 @@ fn spawn_instance_removal_watcher(
                         match event {
                             Some(Ok(DiscoveryEvent::Removed(id))) => {
                                 if let DiscoveryInstanceId::Endpoint(eid) = &id {
-                                    let n = addressed.cancel_instance_streams(eid).await;
-                                    if n > 0 {
-                                        tracing::warn!(
-                                            namespace = %eid.namespace,
-                                            component = %eid.component,
-                                            endpoint = %eid.endpoint,
-                                            instance_id = eid.instance_id,
-                                            cancelled = n,
-                                            "Cancelled pending response streams for removed \
-                                             instance (discovery-driven cleanup)"
-                                        );
-                                    }
+                                    dispatch.on_instance_removed(eid).await;
                                 }
                             }
                             Some(Ok(DiscoveryEvent::Added(DiscoveryInstance::Endpoint(inst)))) => {
                                 let eid: EndpointInstanceId = inst.endpoint_instance_id();
-                                addressed.clear_instance_tombstone(&eid).await;
+                                dispatch.on_instance_added(&eid).await;
                             }
                             Some(Ok(_)) => {}
                             Some(Err(e)) => {
@@ -558,7 +551,8 @@ where
             None
         };
 
-        // Cancel orphaned pending response streams when workers die.
+        // Type-erase to the seam so discovery-removal cleanup runs through it.
+        let addressed: Arc<dyn StreamingDispatch<T, U>> = addressed;
         spawn_instance_removal_watcher(
             client.endpoint.clone(),
             addressed.clone(),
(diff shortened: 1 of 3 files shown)


lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -651,6 +650,54 @@ where
...
+            Some(get_or_create_routing_occupancy_state(&client.endpoint).await)
+        } else {
+            None
+        };
+
+        spawn_instance_removal_watcher(
+            client.endpoint.clone(),
+            dispatch.clone(),
+            client.endpoint.drt().primary_token(),
+        );
```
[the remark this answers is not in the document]
tanmayv25: Verified — not a live defect in this PR, so no logic change. `AddressedPushRouter` is the only `StreamingDispatch` impl here, and its `on_instance_removed`/`on_instance_added` act on the shared per-DRT `TcpStreamServer` (`tcp_server()` is a process-wide `OnceCell`), so whichever router wins the `ENDPOINT_WATCHER_ACTIVE` race, cleanup hits the same shared state. The dedup is also pre-existing — this PR only made the watcher generic. It becomes a real bug only once a `StreamingDispatch` holds *per-instance* state, which doesn't exist in this PR.

Captured as a trait invariant on `StreamingDispatch` (commit 1d99755): a stateful dispatch must share per-endpoint state, or the dedup key must gain dispatch identity. The fan-out/subscriber-set fix belongs with the PR that introduces such a dispatch.

lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -320,11 +323,14 @@ static ENDPOINT_CACHE_INDEXER_WATCHER_ACTIVE: std::sync::OnceLock<
...
+fn spawn_instance_removal_watcher<T, U>(
     endpoint: Endpoint,
-    addressed: Arc<AddressedPushRouter>,
+    dispatch: Arc<dyn StreamingDispatch<T, U>>,
     cancel_token: tokio_util::sync::CancellationToken,
-) {
+) where
+    T: Data + Serialize + 'static,
+    U: Data + for<'de> Deserialize<'de> + MaybeError + 'static,
+{
```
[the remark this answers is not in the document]
tanmayv25: Verified — not a live defect in this PR, so no logic change. `AddressedPushRouter` is the only `StreamingDispatch` impl here, and its `on_instance_removed`/`on_instance_added` act on the shared per-DRT `TcpStreamServer` (`tcp_server()` is a process-wide `OnceCell`), so whichever router wins the `ENDPOINT_WATCHER_ACTIVE` race, cleanup hits the same shared state. The dedup is also pre-existing — this PR only made the watcher generic. It becomes a real bug only once a `StreamingDispatch` holds *per-instance* state, which doesn't exist in this PR.

Captured as a trait invariant on `StreamingDispatch` (commit 1d99755): a stateful dispatch must share per-endpoint state, or the dedup key must gain dispatch identity. The fan-out/subscriber-set fix belongs with the PR that introduces such a dispatch.

lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -651,6 +650,54 @@ where
...
+            router_mode,
+            round_robin_counter: Arc::new(AtomicU64::new(0)),
+            fault_detection_enabled: true,
+            response_timeout: response_inactivity_timeout(),
+            occupancy_state,
+            multimodal_cache_indexer: None,
+            multimodal_cache_key_extractor: None,
+            _phantom: PhantomData,
+        })
+    }
```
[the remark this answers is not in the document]
tanmayv25: Correct, and unused in this PR. Rather than plumb speculative `WorkerLoadMonitor` / `MultimodalCacheIndex` parameters for a caller that doesn't exist yet, I documented the constructor's scope (commit 1d99755): it wires frontend-local occupancy only — no load monitor and no multimodal cache indexer, so `RouterMode::DeviceAwareWeighted` is non-functional and a caller needing either must extend it.

lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -388,23 +392,12 @@ fn spawn_instance_removal_watcher(
...
-                                            namespace = %eid.namespace,
-                                            component = %eid.component,
-                                            endpoint = %eid.endpoint,
-                                            instance_id = eid.instance_id,
-                                            cancelled = n,
-                                            "Cancelled pending response streams for removed \
-                                             instance (discovery-driven cleanup)"
-                                        );
-                                    }
+                                    dispatch.on_instance_removed(eid).await;
```
PeaBrane, line 395: AI-assisted review — please evaluate: `ENDPOINT_WATCHER_ACTIVE` is keyed only by `EndpointId`, so these lifecycle callbacks go only to the first dispatch registered for an endpoint. The current default path is safe because `AddressedPushRouter` instances share the runtime TCP server, but `from_client_with_dispatch` makes heterogeneous dispatches reachable: default-first means the custom transport misses cleanup, while custom-first means request-plane stream cancellation and tombstone clearing are skipped. The documented “share state per endpoint” constraint only covers instances of the same implementation. Could we either fan events out to all registered dispatches or explicitly enforce one dispatch implementation per endpoint?

lib/runtime/src/pipeline/network/egress/push_router.rs
```diff
@@ -651,6 +646,49 @@ where
...
     }
 
+    /// Like the other constructors but with a caller-supplied [`StreamingDispatch`]
+    /// as the final hop. Fault detection is on, so the dispatch's `ErrorType`
+    /// mapping drives report-down / overload / migration as usual.
+    ///
+    /// Wires frontend-local occupancy only — no `WorkerLoadMonitor` and no
+    /// multimodal cache indexer, so `RouterMode::DeviceAwareWeighted` is
+    /// non-functional; a caller needing those must extend it.
+    pub async fn from_client_with_dispatch(
```
PeaBrane, line 656: AI-assisted suggestion — please evaluate adding one focused fake-dispatch integration test for this extension contract. Construct through `from_client_with_dispatch`, verify unary and bidirectional requests reach the supplied dispatch with the selected address/instance, and verify discovery add/remove hooks are delivered. Existing tests exercise the default adapter, but not a supplied alternate dispatch or its lifecycle wiring.

PeaBrane (approved):
✅ AI-assisted review complete. Approving with two non-blocking comments for human evaluation.

lib/runtime/src/pipeline/network/egress/push_router.rs
tanmayv25: Agreed this is the sharp edge. It's latent in this PR: `AddressedPushRouter` is the only `StreamingDispatch` impl and its `on_instance_removed`/`on_instance_added` act on the shared per-DRT `TcpStreamServer`, so cancellation and tombstone-clearing are global regardless of which instance wins the dedup race — the documented per-endpoint sharing constraint holds. It only bites once a *heterogeneous* dispatch shares an endpoint, which first happens in the direct-gRPC follow-up (a mixed pool with some request-plane and some direct workers on the same component). I'll fix it there — most likely by fanning the watcher out to a set of per-endpoint subscribers (weak refs, deregistered on drop) rather than a single owner — keeping this PR behavior-preserving. Tracking it as a prerequisite for that PR.

lib/runtime/src/pipeline/network/egress/push_router.rs
tanmayv25: Good call — added in 87a8ef88d1. A `RecordingDispatch` fake constructed through `from_client_with_dispatch` now asserts unary and bidirectional requests reach the supplied dispatch with the selected address/instance, and that discovery removal/re-addition are delivered to `on_instance_removed`/`on_instance_added`. Runs on the in-process discovery backend, no external deps.

PR #12450, opened 2026-07-30
fix(ci): run dind test container as runner uid to fix shared model cache

nv-tusharma:
## Summary

The dind test container in `pytest/action.yml` was running `docker run` without `--user`, so it wrote `/models/hub/` entries as root. The `pytest-local` path (ARC runner, UID 1001) cannot create new subdirs in those root-owned directories, causing `INTERNALERROR` in conftest's `download_models()` before any tests run.

Add `--user $(id -u):$(id -g)` to the dind `docker run` so both paths write to `/models/hub/` as the same UID and can share the model cache.

## Validation

- dind and `pytest-local` now both run as UID 1001; new model cache entries are writable by both paths
- No other `docker run` in CI mounts `/models`, so no other callers need this change

### DIFF
--- a/.github/actions/pytest-local/action.yml
+++ b/.github/actions/pytest-local/action.yml
@@ -142,10 +142,15 @@ runs:
         REQUESTED_HF_HOME="${{ inputs.hf_home }}"
         FALLBACK_HF_HOME="${WORK_DIR}/.cache/huggingface"
         # Prefer the requested cache path when it is usable in this job container.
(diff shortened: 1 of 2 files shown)


nv-tusharma:
Closing — the fix targets pytest/action.yml which is never invoked by any workflow (dead code). The actual issue is H100 runner-node specific: /models/hub/ has a permission state where UID 1001 can pass the top-level -w check but cannot create new subdirectories, causing INTERNALERROR when conftest tries to download a new model (EAGLE3). This needs to be fixed on the runner infrastructure side, not in CI code.

PR #8197, opened 2026-04-14
feat: integrate cache_salt with KV routing for multi-tenant isolation [DYN-3178]

jthomson04:
## Overview:

Adds cache-salt-aware KV routing and reuse isolation across Dynamo request ingestion, block hashing, router selection, bindings, standalone indexing, vLLM, and TensorRT-LLM KV events.

This work is tracked internally as DYN-3178. It is backward compatible for unsalted requests: when no non-empty salt is supplied, Dynamo retains the existing hash and cache-reuse behavior.

## Details:

- Resolves the public cache-salt API with this precedence:
  1. Non-empty `x-tenant-id` gateway header.
  2. Non-empty `nvext.cache_salt` (recommended request API).
  3. Non-empty top-level `cache_salt` compatibility field.
- Treats empty strings as absent, including falling back from an empty NvExt value to a non-empty top-level compatibility value.
- Carries the resolved namespace through the standalone preprocessor, C binding, Rust EPP, Python binding, router request, and cache-overlap paths. The Rust EPP applies the gateway header before body-derived values.
- Mixes the namespace into block hashing with a dedicated seed while preserving both the existing unsalted hash output and the established LoRA hash formula shared with `dynamo-kv-hashing`.
- Threads cache salt into both legacy and unified vLLM `TokensPrompt` inputs. Dynamo uses an internal tag while the salt is inside vLLM's untyped `extra_keys`, then removes it in the Rust event decoder so arbitrary public values, including 64-character hex salts and salts equal to a LoRA name, round-trip without ambiguity.
- Preserves cache salt on TensorRT-LLM stored-block events, rejects events that combine conflicting block salts, and drops only the malformed event rather than terminating the publisher thread.
- Adds pre-merge real-engine isolation E2Es for unified vLLM and unified TensorRT-LLM; both submit identical tokens under two salts and verify the router indexes each namespace independently.
- Makes the standalone indexer contract explicit:
  - `/query` computes salted hashes from token IDs.
  - `/query_by_hash` requires hashes that were already computed with the intended salt and returns `400 Bad Request` for any non-null `cache_salt`.
- Avoids logging raw cache namespace values in the C-binding and Go EPP paths.
- Documents the API surface, precedence, frontend NvExt gating, empty-value behavior, backend support matrix, isolation usage, and standalone-indexer behavior. End-to-end backend reuse isolation is currently supported for vLLM and TensorRT-LLM, not SGLang.
[...]

PR #12456, opened 2026-07-30
fix(kv-router): fail closed on unrecognized KV event media

Change72:
## Summary

dynamo was **fail-open** on KV-event `medium` strings it doesn't recognize: `convert_event` fell back to the Device tier via `from_kv_medium_or_default`, so an unrecognized medium was indexed on the G1 primary radix tree — and separately flowed through the salted cache-namespace normalizer state and was tracked as G1 by both consolidator ingresses. This PR makes all three paths **fail closed**.

Concrete trigger: vLLM **v0.26.0** (dynamo's current pin) emits `FS` / `OBJ` media (present since vLLM #47923; unified to `STORAGE` only in 0.27+ via #48123). dynamo doesn't recognize `FS` / `OBJ`, so today they hit all three holes. This is the follow-up promised in #11571:

> Hardening dynamo's handling of unrecognized media (fail-closed instead of defaulting to Device) is a separate follow-up PR — it is a pre-existing latent issue on `main`, independent of this change.

## Behavior

| medium | before → after |
|---|---|
| absent / `GPU` / `CPU` / `STORAGE` / other recognized | unchanged |
| unrecognized (e.g. v0.26.0 `FS` / `OBJ`, future strings) | **router**: fell back to Device → indexed on G1 → **dropped in conversion (rate-limited warn), now a backstop**; **normalizer**: polluted salted-namespace state, and a later conversion drop burned a `next_event_id` (false engine-drop metric) → **rejected in preprocess (`unknown_medium` filter reason); no event id burned**; **consolidator**: tracked as G1 → **dropped** |

## Changes

- `convert_event`: drop unrecognized media with a rate-limited warning (was a silent Device fallback). Removes the now-unused `StorageTier::from_kv_medium_or_default`. This drop is kept as a defensive backstop for direct callers that bypass `preprocess`.
- `preprocess_with_reason`: reject unrecognized media with a new `unknown_medium` filter reason instead of letting them reach conversion. Filtering here — exactly like the #11571 locality gate — means the LLM ZMQ listener records an intentional filter rather than accepting the event, incrementing `next_event_id`, and only dropping it in conversion; that id gap is what the event processor misreads as an engine drop (`engines_dropped_events`). Recognized lower-tier media (`STORAGE` → Disk/External) still bypass normalizer state; **Device and HostPinned events (GPU, CPU offload #10368) still flow through the normalizer** so their salted namespaces keep propagating.
- Both consolidator ingresses (`kvbm-consolidator`, LLM `kv_consolidator`): gate unrecognized media out of the G1-only tracker.

The normalizer and consolidator conditions are intentionally different (normalizer keeps CPU on-path for salted propagation; the consolidator is G1-only and drops CPU), so there is no shared predicate.

Out of scope: locality semantics (settled in #11571), the CPU/HostPinned normalizer path (#10368), any `FS` / `OBJ` → tier mapping (dynamo deliberately does not consume them — this only flips them from fail-open to fail-closed), and removal/refcount semantics (#11930).

## Validation

All commands on a Linux workstation, all green:

```
cargo test -p dynamo-kv-router --lib
cargo test -p dynamo-kv-router --features standalone-indexer --lib
cargo test -p dynamo-llm --lib kv_router::publisher
cargo test -p dynamo-llm --lib block_manager::kv_consolidator
cargo test -p kvbm-consolidator
cargo fmt --all -- --check
cargo clippy --no-deps --all-targets -- -D warnings
```

New/updated tests:
- `preprocess_with_reason` rejects `FS` / `OBJ` / `XYZ` as `unknown_medium` (store + remove) while recognized media pass.
- an `FS` event reusing a salted GPU hash is rejected in `preprocess` (so it never touches namespace state) and the salted GPU chain keeps inheriting the namespace.
- listener-level `GPU → FS → GPU` (real ZMQ publisher): the two GPU stores land on consecutive event ids (0, 1), proving the filtered `FS` store burns no `next_event_id` (no false engine drop).
- `convert_event` still drops `FS` / `OBJ` / `XYZ` as a backstop (store + remove, any locality); both consolidator ingresses drop `FS`.

## AI assistance

Implemented with AI assistance (Cursor); design, scoping, and review were human-directed.

## Summary by CodeRabbit

* **New Features**
  * Added support for vLLM disk offloading through the documented storage tiering configuration.
  * Clarified that `STORAGE` events map to Dynamo’s Disk tier.

* **Bug Fixes**
  * Unknown or unsupported storage media are now rejected instead of silently defaulting to device storage.
  * Invalid events no longer consume event IDs or alter namespace state.
  * Updated event filtering to consistently accept only supported device-tier events where required.

* **Documentation**
  * Documented rejection of legacy `FS` and `OBJ` event types.


lib/kv-router/src/zmq_wire/mod.rs
```diff
@@ -139,18 +141,25 @@ impl ZmqEventNormalizer {
...
+        //    listener records an intentional filter. Bypassing to conversion,
+        //    which drops them, would instead accept the event, burn a
+        //    next_event_id, and leave an id gap the event processor mistakes for
+        //    an engine drop -- the same trap the locality gate above avoids.
+        if let Some(m) = raw.medium() {
+            match StorageTier::from_kv_medium(m) {
+                Some(StorageTier::Device | StorageTier::HostPinned) => {}
+                Some(_) => return Ok(raw),
+                None => return Err(ZmqEventFilterReason::UnknownMedium),
+            }
```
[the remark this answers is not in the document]
Change72: Good consistency catch — fixed in cd14d9ff6c. Empty-string `medium` is now normalized to absent at the wire deserialize boundary (mirroring empty `cache_salt`), so an unset medium stays on the device tier and is indexed instead of failing closed. No dynamo/vLLM producer emits `medium: ""` today; this preserves prior behavior for that value and keeps fail-closed scoped to populated-but-unrecognized media. Added a deserialize test (store + remove).

docs/fern/pages/developer-guide/knowledge-base/modular-components/router/offloading-support-matrix.md
```diff
@@ -13,7 +13,7 @@ Legend: ✅ tier-aware routing · 🟡 router-visible, tier-agnostic · 🚧 Dyn
 
 | Framework | Version gates | GPU | CPU RAM | Disk | Shared pool |
 | --- | --- | --- | --- | --- | --- |
-| [**vLLM**](#vllm) | vLLM v0.24.0+; Dynamo v1.3.0+ | ✅ KV events | ✅ `OffloadingConnector` + self-describing KV events (aggregated) | 🚧 vLLM main emits FS/OBJ events; Dynamo tier mapping is in progress | 🚧 vLLM locality events are merged; Dynamo shared-pool indexing is in progress |
+| [**vLLM**](#vllm) | vLLM v0.24.0+; Dynamo v1.3.0+ | ✅ KV events | ✅ `OffloadingConnector` + self-describing KV events (aggregated) | ✅ `TieringOffloadingSpec` + self-describing STORAGE events (aggregated) | 🚧 vLLM locality events are merged; Dynamo shared-pool indexing is in progress |
```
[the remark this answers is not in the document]
Change72: Fixed in cd14d9ff6c — the Disk cell now reads "✅ `TieringOffloadingSpec` + self-describing STORAGE events (vLLM 0.27+; aggregated)" so readers on the pinned 0.26.0 see the gate. The related `native-kv-offloading.md` is already consistent on `main` (Disk ⚠️, STORAGE→Disk mapping, cache-salt caveat), so no change is needed there.

lib/kv-router/src/zmq_wire/mod.rs
```diff
@@ -66,6 +66,7 @@ struct KvCacheGroupMetadata {
 pub enum ZmqEventFilterReason {
     IgnoredEvent,
     NonLocalLocality,
+    UnknownMedium,
```
jthomson04, line 69: P2: `ZmqEventFilterReason` is public and was published in v1.3.1 without `#[non_exhaustive]`. Adding this variant breaks downstream exhaustive matches when they update to 1.4.0. Reuse a compatible existing reason for this release, or handle this as a planned breaking API change.

lib/kv-router/src/zmq_wire/deserialize.rs
```diff
@@ -187,7 +187,7 @@ impl<'de> Visitor<'de> for RawKvEventVisitor {
                     .ok_or_else(|| de::Error::invalid_length(4, &"missing block_size"))?;
                 // Position 5 was lora_id in older formats; consume and discard for compat.
                 let _lora_id: Option<u64> = seq.next_element()?.unwrap_or(None);
-                let medium: Option<String> = seq.next_element()?.unwrap_or(None);
+                let medium: Option<String> = normalize_medium(seq.next_element()?.unwrap_or(None));
```
jthomson04, line 190: P3: `normalize_medium` is added to both map and sequence branches, but the new regression test exercises only map-encoded store/remove events. Add one sequence-encoded empty-medium case here so this compatibility behavior is covered on the legacy tuple/array path.

jthomson04 (commented):
Two source review findings.

lib/kv-router/src/zmq_wire/mod.rs
Change72: v1.3.1 shipped this enum with 5 variants — `NonLocalLocality` ([#11571](https://github.com/ai-dynamo/dynamo/pull/11571)) and `AmbiguousCacheNamespace` ([#8197](https://github.com/ai-dynamo/dynamo/pull/8197)) are both post-1.3.1 on main, so 1.4.0 breaks exhaustive matches on this enum regardless of this PR. Given that, reusing an existing reason wouldn't avoid the break and would misclassify unknown-medium drops in the zmq_filtered_events label. 

Proposal: treat 1.4.0 as the planned breaking release for this enum and make it the last one — keep `UnknownMedium` and add #[non_exhaustive] (happy to do it in this PR or split it out, your call).

lib/kv-router/src/zmq_wire/deserialize.rs
Change72: Added a test in [7e48754](https://github.com/ai-dynamo/dynamo/pull/12456/commits/7e4875482eedfa6ff18c31b7cbf15697f3d1f635)

jthomson04 (approved):

PR #12467, opened 2026-07-31
Proposal: fall back when conditional decode selection fails

MatejKosec:
This is a proposed implementation of the fallback discussed in #11725 and tracked in DYN-3694.

## What goes wrong today

Conditional disaggregation first chooses a decode worker from the scheduler state without reserving that worker or admitting the request. The chosen worker and rank are then written into the request, and the decode router performs the real selection and admission.

That second operation can fail before request tracking or backend dispatch begins. For example, the worker may no longer be eligible or may no longer have capacity. Today that error ends the request even though the ordinary remote-prefill route is still available.

## How the fallback works

For a request carrying the router-owned conditional-bypass marker, `KvPushRouter` returns the original request when `select_with_affinity` fails with a non-cancellation error. At that point the request has not been recorded by `track_selection`, no `RequestGuard` has been created, and backend dispatch has not begun.

`PrefillRouter` recognizes only that ownership-return error. It restores the complete routing value from before the advisory decode pin, removes the internal bypass marker, and continues through the existing remote-prefill path. Restoring the complete value preserves any original worker pins, constraints, priorities, LoRA information, and the distinction between absent and empty routing hints.

Cancellation is not converted into fallback. Errors after selection succeeds—including request tracking, dispatch, backend, and response-stream errors—also continue to propagate without retry because the request may already have recorded state or reached the backend.

## Why the pipeline frontend changes

`next.generate(...)` registers a response sender before forwarding the request downstream. When downstream returns the ownership-return error, no response arrives to remove that sender. The new registration guard removes the exact pending sender whenever `generate` exits before receiving a response, including immediate errors and cancellation. Its identity token prevents an older call from removing a newer registration which happens to use the same request ID.

## Checks run

Focused tests were added for context preservation, cancellation, the post-selection boundary, routing restoration, and response-registration cleanup. The tests have not been run, so this PR remains a draft.

The checks that were run are:

- `rustfmt --edition 2024 --check` on the four changed files
- `git diff --check`

### DIFF
--- a/lib/llm/src/kv_router/prefill_router/mod.rs
+++ b/lib/llm/src/kv_router/prefill_router/mod.rs
@@ -26,7 +26,7 @@ use crate::{
     protocols::common::{
         extensions::{SESSION_AFFINITY_CONTEXT_KEY, SessionAffinityId},
         llm_backend::{LLMEngineOutput, PreprocessedRequest},
-        preprocessor::{BootstrapInfo, PrefillResult, TraceLink},
+        preprocessor::{BootstrapInfo, PrefillResult, RoutingHints, TraceLink},
         timing::{RequestPhase, RequestTracker},
     },
     session_affinity::{AffinityCoordinator, AffinityTarget},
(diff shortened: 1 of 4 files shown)


dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `adca76f` | unsigned |
| `36fb235` | unsigned |

PR #12474, opened 2026-07-31
fix(frontend): support max Responses reasoning effort

rmccorm4:
## Summary

- accept `reasoning.effort = "max"` on the Responses API using Dynamo's shared reasoning-effort type
- preserve the requested effort through Responses-to-Chat conversion and echo it in unary and streaming Responses output
- keep upstream-compatible response serialization while covering both `high` and `max` with focused regression tests

The upstream Responses reasoning enum currently stops at `xhigh`, so it rejects `max` before model-specific handling. Dynamo's shared Chat reasoning-effort enum already supports `max`; this change uses it at the Responses boundary.

Linear: [DIS-2560](https://linear.app/nvidia/issue/DIS-2560/codex-compatibility-support-reasoning-effort-max-for-kimi-k3-style)

## Validation

- `cargo check -p dynamo-llm --tests`
- `cargo test -p dynamo-llm --lib 'protocols::openai::responses::'` (87 passed)
- `git diff --check`
- repository pre-commit hooks

### DIFF
--- a/lib/llm/src/http/service/openai.rs
+++ b/lib/llm/src/http/service/openai.rs
@@ -2534,7 +2534,7 @@ async fn responses(
         tools: request.inner.tools.clone(),
         tool_choice: request.inner.tool_choice.clone(),
         instructions: request.inner.instructions.clone(),
-        reasoning: request.inner.reasoning.clone(),
+        reasoning: request.reasoning.clone(),
         text: request.inner.text.clone(),
         service_tier: request.inner.service_tier,
         include: request.inner.include.clone(),
@@ -3984,6 +3984,7 @@ mod tests {
                 model: Some("test-model".into()),
                 ..Default::default()
             },
+            reasoning: None,
             nvext: None,
         }
     }
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -9,7 +9,7 @@ use dynamo_protocols::types::responses::{
     AssistantRole, FunctionCallOutput, FunctionToolCall, IncludeEnum, IncompleteDetails,
     InputContent, InputItem, InputOutputMessageContent, InputParam, InputRole, InputTokenDetails,
     Instructions, Item, MessageItem, OutputItem, OutputMessage, OutputMessageContent, OutputStatus,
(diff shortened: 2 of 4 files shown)


rmccorm4:
Going for an upstream fix first: https://github.com/64bit/async-openai/pull/581

Can revisit this as a WAR if no traction on upstream fix.

rmccorm4:
`Max` ReasoningEffort should be coming to upstream async-openai typing in next release: https://github.com/64bit/async-openai/pull/573

stu-cao:
The upstream fix landed, but it hasn't reached Dynamo — `max` is still rejected on `/v1/responses` today.

[64bit/async-openai#573](https://github.com/64bit/async-openai/pull/573) merged on 2026-08-18 and shipped in **async-openai 0.42.0** (2026-09-09), adding `Max` to the shared `ReasoningEffort`. The gap is that `dynamo-protocols` still requires `async-openai = "0.41"` — in 5.4.3 (what `main` pins) and in 6.1.0, the latest on crates.io — and `"0.41"` will never resolve to 0.42.

Verified against both published versions:

```
dynamo-protocols 5.4.3  reasoning.effort=max  FAIL -> unknown variant `max`,
    expected one of `none`, `minimal`, `low`, `medium`, `high`, `xhigh`
dynamo-protocols 6.1.0  reasoning.effort=max  FAIL -> (same)
```

I've opened [ai-dynamo/frontend-crates#258](https://github.com/ai-dynamo/frontend-crates/pull/258) for the bump. The `Max` arm is a one-liner; the rest is following three renames in 0.42 (`ToolChoiceParam::Mode` → `::Option`, `responses::ServiceTier` → `ServiceTierResponses`, plus two structs that gained fields). 149 tests pass, clippy `-D warnings` clean.

One thing worth flagging for this side: it's breaking. `CreateResponse.service_tier` changes type, and `dynamo-llm` imports `responses::ServiceTier` in `protocols/openai/responses/mod.rs`, so Dynamo will need a small migration alongside the pin bump once the new `dynamo-protocols` is published. Happy to open that follow-up if the bump looks right to you.

Context for the urgency: this affects Codex and any DeepSeek/GLM-style client that sends `max` by default. Chat Completions accepts the same value, since it goes through Dynamo's own enum — so the same model answers differently depending on which endpoint you call.

PR #12478, opened 2026-07-31
perf(frontend): avoid quadratic batch prompt copies

gitover22:
## Summary

- Move owned completion batch prompts into per-prompt requests instead of cloning the full batch for every child request.
- Clone a payload-free request template while preserving the existing request and context metadata flow.
- Move the sole owned token batch through preprocessing without an additional token-vector copy.

## Details

For a batch of (B) prompts with (S) total prompt data, request splitting previously cloned the full (S)-sized batch for each child request, in addition to cloning each extracted prompt. For equally sized prompts, that makes prompt copying quadratic in the batch size.

This change removes the batch prompt from the request template before cloning it, then moves each owned string or token vector into its child request. Prompt variants, ordering, choice-index remapping, streaming behavior, annotations, and context metadata are unchanged.

A Criterion benchmark with 1,024 token IDs per prompt measured the request-splitting path as follows (request construction excluded from both measurements):

| Batch size | Before | After | Speedup |
| ---: | ---: | ---: | ---: |
| 8 | 12.894 µs | 2.284 µs | 5.65x |
| 32 | 179.377 µs | 7.464 µs | 24.03x |
| 128 | 2.805 ms | 22.793 µs | 123.05x |

At batch size 128, the old split copied approximately 64.5 MiB of prompt-token payload for a 512 KiB input batch; the new split moves that payload without copying it.

Start with `completions_batch` in `lib/llm/src/http/service/openai.rs`, followed by the consuming prompt conversion in `lib/llm/src/protocols/openai/completions.rs`.

## Validation

- `cargo test -p dynamo-llm --lib protocols::openai::completions::tests::test_into_single_prompts_preserves_batch_order`
- `cargo test -p dynamo-llm --test openai_completions`
- `cargo test -p dynamo-llm --test http-service`
- `cargo clippy -p dynamo-llm --lib --tests -- -D warnings`
- `uvx pre-commit run --files lib/llm/src/http/service/openai.rs lib/llm/src/preprocessor.rs lib/llm/src/protocols/openai/completions.rs`

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12478" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized batch completion request processing for improved efficiency and reduced memory overhead.

* **Tests**
  * Added test coverage for prompt batch conversion functionality.

### DIFF
--- a/lib/llm/src/http/service/openai.rs
+++ b/lib/llm/src/http/service/openai.rs
@@ -974,7 +974,8 @@ async fn completions_batch(
     batch_size: usize,
     n: u8,
 ) -> Result<Response, ErrorResponse> {
-    use crate::protocols::openai::completions::extract_single_prompt;
+    use crate::protocols::openai::completions::into_single_prompts;
+    use dynamo_protocols::types::Prompt;
     use futures::stream::{self, StreamExt};
 
     let request_id = request.id().to_string();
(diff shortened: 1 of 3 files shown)


gitover22:
Addressed the error-message wording nit in `c06c66674`: the message now states that exactly one token vector is required, which also covers the empty-batch case accurately. The focused unit test and formatting checks pass.

PR #12493, opened 2026-07-31
fix(dgdr): normalize model cache paths

cpakkamisaac-sae:
## Overview

Fix DGDR model-cache path handling when `spec.modelCache.pvcModelPath` is already the container-visible path under `spec.modelCache.pvcMountPath`.

Without this normalization, the profiler/AIC path flow can generate DGD args such as `/opt/models/opt/models/...`, which points workloads at a non-existent checkpoint path.

## Details

- Add shared DGDR model-cache path helpers:
  - `normalize_model_cache_path(...)` returns the container-visible model path used by Dynamo config modifiers and local model inspection.
  - `model_cache_path_in_pvc(...)` returns the PVC-relative path expected by AIC generator inputs.
- Use the container-visible helper in `resolve_model_path()` and `BaseConfigModifier._normalize_model_path()`.
- Use the PVC-relative helper before passing `k8s_model_path_in_pvc` into rapid and thorough AIC generator paths.
- Preserve existing relative paths and legacy leading-slash inputs such as `/model/qwen`; only paths already under the configured mount are treated as already container-visible.
- Add regression coverage for the helper behavior, rapid/thorough AIC handoff, and backend config modifier DGD arg generation.

## Where should reviewer start?

Start with `components/src/dynamo/profiler/utils/model_cache_paths.py`, then the two call sites in `rapid.py` and `thorough.py`. The regression tests show the intended behavior for legacy PVC-relative paths and already-mounted absolute paths.

## Related Issues

Addresses the DGDR model-cache path item described in #12016.

## Validation

- Reproduced on AKS with Dynamo Platform 1.3.0 using a throwaway `DynamoGraphDeploymentRequest` with `autoApply: false`, `pvcMountPath: /opt/models`, and an already-mounted `pvcModelPath`. The generated selected config contained:
  - `Frontend --model-path /opt/models/opt/models/...`
  - `VllmDecodeWorker --model /opt/models/opt/models/...`
- Validated the AIC handoff in the AKS `dynamo-planner:1.3.0` image: absolute `k8s_model_path_in_pvc` doubled, while the PVC-relative path produced by this fix generated clean YAML with no `/opt/models/opt/models`.
- `PYTHONPATH=/tmp/dynamo-profiler-testdeps:components/src python -m pytest -q components/src/dynamo/profiler/tests/unit/test_model_cache_paths.py`
- `PYTHONPATH=/tmp/dynamo-profiler-testdeps:components/src python -m py_compile ...` for touched profiler modules/tests
- `git diff --check`

Local higher-level profiler test targets were module-skipped in this lightweight checkout because `aiconfigurator` and the built `dynamo.llm` bindings are not installed locally; the added tests are intended to run in the normal CI environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model paths for profiler workflows using mounted PVC storage.
  * Prevented already-mounted absolute paths from being prefixed twice.
  * Improved handling of relative paths, mount roots, nested paths, trailing slashes, and legacy leading-slash formats.
  * Ensured profiling services receive the correct container and Kubernetes model paths.

* **Tests**
  * Added regression coverage across vLLM, SGLang, TRT-LLM, rapid profiling, and thorough profiling workflows.

### DIFF
--- a/components/src/dynamo/profiler/utils/model_cache_paths.py
+++ b/components/src/dynamo/profiler/utils/model_cache_paths.py
@@ -1,0 +1,59 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Path helpers for DGDR model-cache PVC configuration."""
+
+
+def normalize_model_cache_path(
+    pvc_mount_path: str | None,
+    pvc_model_path: str | None,
+) -> str:
+    """Resolve a DGDR model-cache path to the container-visible model path.
+
+    ``pvcModelPath`` is normally relative to the PVC. For compatibility, a leading
+    slash is still treated as PVC-relative unless the value begins with the
+    configured mount path, in which case it is an already container-visible path.
+    To address a PVC directory that happens to match the mount-path prefix, use the
+    unambiguous relative form without a leading slash.
+    """
+    mount = (pvc_mount_path or "").rstrip("/")
+    raw_path = (pvc_model_path or "").strip()
+    if not raw_path:
+        return mount
+
+    absolute_path = raw_path.rstrip("/")
+    if mount and (absolute_path == mount or absolute_path.startswith(f"{mount}/")):
+        return absolute_path
+
+    sub_path = raw_path.lstrip("/")
+    if not sub_path:
+        return mount
+    return f"{mount}/{sub_path}"
+
+
+def model_cache_path_in_pvc(
+    pvc_mount_path: str | None,
+    pvc_model_path: str | None,
+) -> str | None:
+    """Return the model path form expected by PVC-relative generators.
+
+    AIC's Kubernetes generator expects ``k8s_model_path_in_pvc`` to be relative
+    to the PVC mount. Preserve existing relative and legacy leading-slash inputs,
+    but strip the mount prefix from already container-visible paths. A path inside
+    the PVC that matches the mount prefix must use the relative form without a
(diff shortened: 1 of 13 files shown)


components/src/dynamo/profiler/utils/model_cache_paths.py
```diff
@@ -0,0 +1,56 @@
...
+    kept backward compatible and treated as PVC-relative unless the value already
+    includes the configured mount path.
+    """
+    mount = (pvc_mount_path or "").rstrip("/")
+    raw_path = (pvc_model_path or "").strip()
+    if not raw_path:
+        return mount
+
+    absolute_path = raw_path.rstrip("/")
+    if mount and (absolute_path == mount or absolute_path.startswith(f"{mount}/")):
```
tedzhouhk, line 25: [P2] Please document the ambiguous prefix case. `pvcModelPath` is currently defined as a path within the PVC, so with `pvcMountPath: /opt/models`, a legacy leading-slash value such as `/opt/models/checkpoint` can mean the PVC-relative path `opt/models/checkpoint` (previously resolving to `/opt/models/opt/models/checkpoint`). This condition silently reinterprets the same value as an already-mounted container path, leaving no representation for the former meaning. If that compatibility tradeoff is intentional, please update the API/CRD contract and add this collision case to the tests.

tedzhouhk (approved):
The fix covers the affected local-resolution, config-modifier, rapid, and thorough paths. I also verified that the issue is still present on the latest main and that the targeted profiler unit suites pass. Leaving one non-blocking compatibility comment.

hhzhang16 (approved):

PR #12508, opened 2026-07-31
fix(frontend): translate response_format/guided_* on the vllm chat-processor path

yifjiang:
## Problem

Setting a chat processor (`--dyn-chat-processor vllm`) builds a preprocessed pipeline via `build_preprocessed_pipeline`, which has **no preprocessor op in the link chain** — so the Rust `OpenAIPreprocessor` never runs.

That preprocessor is what normally translates `response_format` and the `guided_*` extensions into `sampling_options.guided_decoding`. On this path the translation simply doesn't happen, so **every structured-output request silently loses its constraint**: the worker gets no constraint, the model free-runs, and the client gets unconstrained text back with **HTTP 200**. There is no error anywhere to indicate the request was not honoured.

## Fix

Add `_build_guided_decoding()` and call it when assembling the preprocessed request. It mirrors `CommonExtProvider::get_guided_json` in `lib/llm/src/protocols/openai/chat_completions.rs` so the two paths agree:

| input | result |
|---|---|
| `guided_json` + `response_format` | `guided_json` wins |
| `response_format: json_object` | "any JSON object" schema `{"type": "object"}` |
| `response_format: json_schema` | the inner `schema` |
| `response_format: text` | no constraint |
| `guided_choice: []` | **not** a constraint, matching `GuidedDecodingOptions::from_optional` |
| `guided_whitespace_pattern` alone | **not** a constraint — it only refines one |

### One non-obvious detail

The key is set **only when a constraint is present**, rather than always writing a dict of `None`s. The worker feeds this straight into vLLM's `StructuredOutputsParams`, which rejects a constraint set whose fields are all `None` — so an always-present key would convert "no constraint requested" into a hard error.

## Test

Adds `TestGuidedDecodingTranslation` covering each row above plus the empty-choice and whitespace-only edge cases.

⚠️ I could not execute the full test module locally — `test_vllm_processor_unit.py` imports `dynamo.frontend.prepost`, which imports `vllm`, and I have no local vLLM install. I verified the new function's behaviour by exercising it directly against all 11 cases (all pass). **CI is the real check here** — please treat the test run as unverified by me.

## Notes

Filed as a **draft** — happy to attach a Linear ID and rename the branch to the `<login>/<desc>__<DYN-ID>` convention before review.

## Summary by CodeRabbit

* **New Features**
  * Added support for guided decoding options, including JSON schemas, JSON objects, regular expressions, grammars, choices, and whitespace patterns.
  * Requests now translate response-format and guided-decoding settings into generation constraints before processing.

* **Bug Fixes**
  * Ensured empty constraints are omitted and conflicting settings follow a consistent precedence.
  * Added validation to emit no more than one guided-decoding constraint.

### DIFF
--- a/components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
+++ b/components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
@@ -1104,3 +1104,100 @@ def test_runtime_config_context_length(vllm_processor_module, runtime_config, ex
     mdc = SimpleNamespace(runtime_config=lambda: runtime_config)
 
     assert vllm_processor_module._runtime_config_context_length(mdc) == expected
+
+
+class TestGuidedDecodingTranslation:
+    """response_format / guided_* must reach sampling_options.guided_decoding.
+
+    The Rust OpenAIPreprocessor is bypassed on the vllm chat-processor path, so
+    this file owns the translation the Rust ``get_guided_json`` normally does.
+    The worker feeds the result into vLLM's ``StructuredOutputsParams``, which
+    rejects a constraint set holding zero or more than one constraint — so the
+    key must be absent unless exactly one constraint is present.
+    """
+
+    SCHEMA = {
+        "type": "object",
+        "properties": {"city": {"type": "string"}},
+        "required": ["city"],
+    }
+
+    def _build(self, request):
+        from dynamo.frontend.vllm_processor import _build_guided_decoding
+
+        return _build_guided_decoding({"model": MODEL, **request})
+
+    @pytest.mark.parametrize(
+        "request_extra, expected",
+        [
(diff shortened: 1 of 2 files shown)


rmccorm4:
Thanks for putting this together. I took another look while resolving the current merge conflicts, and #12684 has since landed with the same core goal: building and forwarding guided-decoding constraints on the Python vLLM chat-processor path. It also covers `response_format`, the legacy `guided_*` fields, and their interaction with tool guidance through the shared preprocessing flow, with broader validation and test coverage.

Given that, this PR is now superseded by #12684 and can be closed. Thanks again for identifying the gap and documenting the failure mode so clearly.

rmccorm4:
Closing as superseded by #12684

PR #12510, opened 2026-07-31
fix(frontend): strip unconsumed control markers from parsed content

yifjiang:
> **Context: one of six fixes from bringing up [`thinkingmachines/inkling`](https://huggingface.co/thinkingmachines/inkling) on Dynamo + vLLM** (4×GB200, TP=4). inkling's reasoning parser is vLLM's `InklingParser` (`vllm/parser/inkling.py`, via `InklingParserReasoningAdapter`) — a **text-grammar** parser, which is what surfaced this. The bug itself is general: it hits any model whose parser forces `skip_special_tokens=False`. Sibling PRs: #12541, #12511, #12507, #12543.

## Problem

Models whose reasoning parser is a **text-grammar** parser separate thinking from answer using content-kind control markers in the token stream. Such a parser can only split the stream if those markers survive detokenisation — which is exactly why `ParserEngine.adjust_request` forces `skip_special_tokens = False`.

Nothing then removes the markers the parser did **not** consume, so terminal ones leak into user-visible content:

```
content = "<|end_message|>408<|end_message|>"
```

As shipped the two settings are mutually exclusive, and **neither is correct**:

| client sends | result |
|---|---|
| `skip_special_tokens: true` (default) | markers stripped **before** the parser can split → reasoning separation silently doesn't happen, `reasoning_content` comes back empty |
| `skip_special_tokens: false` | parser works, but unconsumed markers **leak into visible output** |

Neither path reports an error — you get HTTP 200 and wrong content either way.

## Fix

Keep the markers for the parser, then strip the residue afterwards. Applied at three points: `_compose_delta_message`, the `_build_choice` funnel, and the one site that seeds a delta directly from `delta_text`.

**The funnel is the part worth reviewing.** `_build_choice` has eight call sites on current `main` and some seed `delta["content"]` directly without going through the composer — so a strip applied only at the composer still let a leading marker through. Enforcing the invariant once at the last point every choice passes through is what actually closes it. (The `_fast_plain_text` early return also exits through `_build_choice`, so the invariant genuinely holds for every choice.)

### Scope is deliberately narrow

- **Only when a reasoning or tool parser is active** — the only case where `skip_special_tokens` was forced off. Models that never had markers stripped are unaffected.
- **Only the `<|...|>` shape**, anchored on both delimiters, so ordinary prose containing a bare `<|` or `|>` is untouched (`a <| b |> c` and `<|end-message|>` both pass through unchanged — the latter because `-` is not in the character class).
- **`reasoning` is left alone** — it is already clean, and rewriting it would risk mangling legitimate text the model quoted.
- **A parser's own markers are preserved** — see below.

### Why a shape, and not `self._control_markers`?

Fair question, since `StreamingPostProcessor` already collects `tokenizer.all_special_tokens` into `self._control_markers` for `_is_control_only_content`. Two reasons:

1. One of the three strip sites seeds `delta_message` from the raw `delta_text` **before** tool parsing, and its result feeds the `_tool_start_markers()` scan that decides whether to buffer for tool extraction. A blanket special-token strip there would delete a tool-call marker that is also a special token and silently break tool detection.
2. In practice the two are near-identical — for `Qwen/Qwen3-0.6B`, every entry of `all_special_tokens` matches this pattern — so the shape costs nothing while staying unable to reach a marker a parser owns.

For the same reason, markers reported by `_tool_start_markers()` / `_tool_end_markers()` are **preserved** even when they match the shape (`<|tool_call|>` and friends belong to the parser's wire format). That case is covered by `test_a_parsers_own_markers_survive`.

### Relationship to `_is_control_only_content`

`main` already has `_is_control_only_content`, but it is a *suppression predicate* — it drops a delta that is **entirely** markers, and only on the tool-call path (`if self.in_progress_tool_calls`). It does not strip, so mixed content like `<|end_message|>408<|end_message|>` still reaches the client intact. The two are complementary.

## Test

**Now covered** — the earlier revision of this PR had no unit tests because `prepost.py` imports `vllm` and I had no local install. Resolved by running the suite inside a dynamo+vLLM container.

`TestControlMarkerStrip`, 15 cases:

- the reported leak, `<|end_message|>408<|end_message|>` → `408`
- leading-only, mid-string, no-marker passthrough, empty, `None`
- the two near-miss shapes above
- the parser-inactive gate (`test_no_parser_active_leaves_text_untouched`)
- a parser's own `<|...|>`-shaped markers survive, while markers it does not own are still stripped
- both funnel behaviours: strip in place, and **drop the key** when the delta was only markers (rather than emitting an empty string)
- `reasoning` untouched while `content` is stripped
- end to end through `process_output`, and the `_fast_plain_text` path left alone

**14 of the 15 fail without the strip.** (The one that passes is the no-parser gate case, which the patch correctly does not change.)

Verified against current `main` in a dynamo+vLLM container: no regressions in `test_vllm_processor_unit.py` or `tests/frontend/test_prepost.py` versus a pristine-`main` baseline, and no new `ruff` or `mypy` findings.

Also validated end to end on a real deployment before this rewrite: the observed marker-leak count went **11 → 6** (composer only) **→ 0** (with the funnel).

## Notes

**Ordering with #12541.** #12541 makes `skip_special_tokens=False` the effective default for requests with reasoning active — which is exactly the condition that produces this leak. That PR does **not** contain this fix, so merging it alone would put the leak on `main`. This one should land first, or alongside.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unconsumed control markers from user-visible streamed and completed responses when reasoning or tool parsing is active.
  * Preserved markers that are managed by the tool parser.
  * Removed responses containing only control markers while preserving normal plain-text output.

* **Tests**
  * Added coverage for streaming, completed responses, parser-owned markers, empty content, and unconfigured parser behavior.


components/src/dynamo/frontend/prepost.py
```diff
@@ -915,10 +916,53 @@ def _tool_end_markers(self) -> tuple[str, ...]:
...
+    # Why a shape rather than `self._control_markers` (this class already collects
+    # `tokenizer.all_special_tokens` for _is_control_only_content): one of the three
+    # call sites seeds `delta_message` from the raw `delta_text` *before* tool
+    # parsing, and its result reaches the `_tool_start_markers()` scan that decides
+    # whether to buffer for tool extraction. A blanket special-token strip there
+    # would delete a tool-call marker that is also a special token and silently
+    # break tool detection. The two are near-identical in practice -- for
+    # Qwen/Qwen3-0.6B every entry of `all_special_tokens` matches this pattern --
+    # but the shape cannot reach a marker a parser owns.
+    _CONTROL_MARKER_RE = re.compile(r"<\|[A-Za-z0-9_]+\|>")
```
rmccorm4: **[P1] Restrict removal to actual special tokens**

This regex treats the textual shape as proof that content is protocol residue. On the pinned vLLM 0.27.1 Hermes/Qwen path, `<|not_a_real_special_token|>` is not in `all_special_tokens` and encodes to nine normal token IDs, but `process_output()` changes `The literal is <|not_a_real_special_token|>.` to `The literal is .`. That silently corrupts valid generated text on any parser-active request. Please remove only tokenizer special tokens, excluding markers owned by the active parser, and add a real parser-path regression for a non-special literal of this shape.

components/src/dynamo/frontend/prepost.py
```diff
@@ -1124,7 +1179,11 @@ def process_output(self, output: Any) -> dict[str, Any] | None:
         current_text = self.previous_text + delta_text
         current_token_ids = self.previous_token_ids + delta_token_ids
 
-        delta_message: DeltaMessage | None = DeltaMessage(content=delta_text)
+        # Seeds the delta directly rather than through _compose_delta_message, so it
+        # needs the same control-marker strip.
+        delta_message: DeltaMessage | None = DeltaMessage(
```
rmccorm4: **[P2] Avoid stripping every delta twice**

For normal parser output, this call runs here and `_build_choice()` runs the same regex again. The seed also cannot feed the later `_tool_start_markers()` scan as the comment claims: the reasoning branch overwrites `delta_message` before the scan, while the tool branch passes raw `delta_text` directly to the parser. Against the exact PR parent, a passthrough-parser `process_output()` probe increased from 5.58 µs to 10.39 µs per five-character delta; `_build_choice()` increased from 0.379 µs to 2.330 µs. Please strip once after parser output stabilizes and precompute the immutable marker set.

components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
```diff
@@ -1708,6 +1709,188 @@ async def test_root_thinking_suppresses_deployment_default(self, tokenizer):
...
+
+
+class _MarkerOwningToolParser(_PassthroughStreamingToolParser):
+    """A tool parser whose own marker happens to match the control-token shape."""
+
+    tool_call_start_token = "<|tool_call|>"
+    tool_call_end_token = "<|tool_call_end|>"
+
+
+class TestControlMarkerStrip:
```
rmccorm4, line 1745: **[P3] Reduce the regression surface**

This addition is 183 lines / 15 cases, with repeated setup and prose plus overlapping empty/None/leading/middle/helper-level assertions. Keep the few distinct contracts—reported end-to-end leak, parser-owned marker preservation, inactive-parser pass-through, and the missing non-special-literal preservation case—so the tests stay focused on independent regressions.

components/src/dynamo/frontend/prepost.py
yifjiang: Confirmed and fixed in 44cae6f.

Reproduced exactly as described against `Qwen/Qwen3-0.6B`: `<|not_a_real_special_token|>` is absent from `all_special_tokens` (14 entries, all `<|im_*|>`/`<|vision_*|>`-style), and the old code turned

```
The literal is <|not_a_real_special_token|>.  ->  The literal is .
```

Removal is now driven by membership in `tokenizer.all_special_tokens`, minus the markers the active parser owns, matched as escaped literals rather than by shape:

```python
owned = set(self._tool_start_markers()) | set(self._tool_end_markers())
self._marker_strip_re = _compile_marker_strip_re(
    tuple(m for m in self._control_markers if m not in owned)
)
```

Regression added — `test_non_special_literal_of_the_same_shape_survives`. It fails on the previous commit with precisely the corruption above:

```
E  AssertionError: assert 'The literal is .' == 'The literal ...cial_token|>.'
E    - The literal is <|not_a_real_special_token|>.
E    + The literal is .
```

One consequence worth flagging: the tests now use `<|im_end|>` rather than the reported `<|end_message|>`. That marker is special on the affected model's tokenizer, which is why it survived detokenisation, but it is *not* special on Qwen3-0.6B — so under the corrected rule the old assertions were checking a string the code must now leave alone.

components/src/dynamo/frontend/prepost.py
yifjiang: Agreed on both counts, fixed in 44cae6f.

The strip now runs **once**, in `_build_choice` — every choice this class emits passes through it, so the invariant holds at the funnel. `_compose_delta_message` is back to being the `@staticmethod` it was and the delta seed is untouched, so both drop out of the diff entirely.

The marker set is computed once per postprocessor rather than per delta, and the compiled pattern is memoised per `(tokenizer, parser)` — `re.compile` over ~15 escaped literals is ~12 µs and a `StreamingPostProcessor` is built per request, so the first version of this fix just moved the cost into `__init__`. Measured on the passthrough-parser probe, min of 3 runs, five-character delta:

| | main | this PR before | this PR now |
|---|---|---|---|
| `process_output` | 2.63 µs | 4.73 µs | **2.81 µs** |
| `_build_choice` | 0.283 µs | 1.242 µs | **0.428 µs** |
| `__init__` | 3.87 µs | 3.90 µs | **5.48 µs** |

The residual +1.6 µs in `__init__` is the once-per-request marker-set computation that replaced the per-delta work.

One correction on the mechanism, since it changes whether the seed strip was load-bearing: the reasoning branch **merges** rather than overwrites —

```python
delta_message = self._merge_delta_messages(
    delta_message, self._finish_engine_parser(self.reasoning_parser))
...
post_content = (delta_message.content if delta_message else None) or ""
tool_start_markers = self._tool_start_markers()
if post_content and any(marker in post_content for marker in tool_start_markers):
```

so the seed's content *can* reach `post_content` and the scan. Removing the seed strip is still correct, but for a different reason than "it never gets there": parser-owned markers are excluded from the strip set, so the scan sees the same text either way.

components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
yifjiang: Done in 44cae6f — **183 lines / 15 cases → 101 lines / 4 cases**, and the PR overall is now 166+/0- (was 245+/3-).

Kept exactly the four distinct contracts:

- `test_reported_leak_is_removed` — the reported end-to-end leak, through `process_output`
- `test_parser_owned_markers_are_preserved` — plus one special token the parser does *not* own, still removed
- `test_no_parser_active_is_passthrough` — the gate
- `test_non_special_literal_of_the_same_shape_survives` — the missing case from P1

Dropped the repeated setup (one `_post` helper and one `_content` helper now), the empty/None/leading/middle permutations, the two shape-anchoring cases (`a <| b |> c`, `<|end-message|>`) which were asserting on a regex that no longer exists, and the composer/`_build_choice` helper-level tests that were re-checking the same strip from a different entry point now that there is only one strip site.

Verification for the whole change: the full `test_vllm_processor_unit.py` failure set is **identical to main's** apart from the two tests that assert the fix. Same container, `prepost.py` swapped underneath:

```
main: 43 failed, 83 passed
PR:   41 failed, 85 passed
fail under PR but not main:  (none)
fixed by PR:  test_reported_leak_is_removed, test_parser_owned_markers_are_preserved
```

The 41 shared failures are environmental — the container has vLLM 0.26.0 rather than the pinned 0.27.1 — and are unaffected by this change in either direction.

components/src/dynamo/frontend/prepost.py
```diff
@@ -718,6 +736,21 @@ def __init__(
...
         )
 
+        # Markers that may be removed from user-visible content: the tokenizer's
+        # actual special tokens, minus any the active parser owns. Membership in
+        # all_special_tokens is the test -- a literal that merely looks like a
+        # control token (`<|not_a_real_special_token|>`) is ordinary generated
+        # text and must survive. Built once: it cannot change for the lifetime of
+        # this postprocessor, and compiled once per process (see the memo above).
+        self._marker_strip_re: re.Pattern[str] | None = None
+        if self.reasoning_parser is not None or self.tool_parser is not None:
```
rmccorm4: [P1] Gate cleanup on configured reasoning parsers

This misses configured-but-inactive reasoning parsers. The Rust preprocessor selects `skip_special_tokens=false` from the configured parser, while this class sets `self.reasoning_parser=None` when `response_reasoning_ended=True` (or thinking is disabled). On this exact head, a real MiniMax M3 tokenizer/parser probe took that path and returned the special-token payload `]~b]408]~b]` unchanged in `content`. Use `reasoning_parser_class is not None` for this gate (or the resolved decode policy), and add a configured-but-inactive parser regression.

yifjiang: Confirmed and fixed in `a95cc72`.

Reproduced on that exact head with a real `Qwen3ParserReasoningAdapter` and a real tokenizer — both inactivity paths, not just one:

```
response_reasoning_ended=True   reasoning_parser=None  strip_re=None  content='<|im_end|>408<|im_end|>'
enable_thinking=False           reasoning_parser=None  strip_re=None  content='<|im_end|>408<|im_end|>'
active (control)                reasoning_parser=<...>  strip_re=compiled  content=None
```

Same shape as your `]~b]408]~b]`. (I could not reproduce the M3 payload exactly — no M3 tokenizer in the local cache — but the mechanism is identical and this uses a real reasoning-parser class.)

The variable mismatch is the whole bug: what forces `skip_special_tokens=false` is the deployment's configured parser **name** — `preprocessor.rs::parser_requires_special_tokens` (PRE.1) matches static names only, and vLLM's `ParserEngine.adjust_request` likewise. Neither sees `enable_thinking` or `response_reasoning_ended`. But those two are exactly what null `self.reasoning_parser` in this class, so an instance-keyed gate is a no-op precisely on the requests that still receive raw markers.

Gate now reads `reasoning_parser_class`. Working through the four combinations, only one cell changes — class set / instance None / no tool parser — which is the reported bug. In that cell the only thing newly removable is a literal byte-identical to a `tokenizer.all_special_tokens` entry, which is the same trade-off already accepted for every active-parser request and already pinned by `test_non_special_literal_of_the_same_shape_survives`.

On your parenthetical alternative — the resolved decode policy is reachable (`sampling_params.skip_special_tokens`), but it is weaker: on the Python-only path nothing sets it False for a reasoning-only, no-tools request, so that gate would leave this bug unfixed there. Took the class gate.

Regression test is parametrized over both inactivity paths; both fail without the one-line change with `assert '<|im_end|>408<|im_end|>' == '408'`. Also renamed `test_no_parser_active_is_passthrough` → `test_no_parser_configured_is_passthrough` to match the corrected contract.

rmccorm4 (approved):

PR #12512, opened 2026-07-31
refactor(router): use default worker scorer and picker traits

ishandhanani:
Reorganizes the score and pick helpers extracted in #12505 into private `WorkerScorer` and `WorkerPicker` contracts exercised by the production `DefaultWorkerSelector`. It also isolates formula inputs behind private borrowed request and worker-row views so the later public API will not expose scheduler maps or `SchedulingRequest`.

CLOSES: DYN-3710

### Implementation

- Adds a borrowed `WorkerSelectionContext` and stack-built `WorkerSelectionRow` containing only signals consumed by scoring.
- Makes `DefaultWorkerScorer` consume that row contract while preserving the exact cost formula and taint multiplier.
- Makes the concrete default scorer/picker independently constructible from borrowed selector state without allocation or dynamic dispatch.
- Keeps `DefaultWorkerPicker` on the existing eligibility enumeration, temperature lookup, reservoir tie handling, and softmax path.
- Keeps `DefaultWorkerSelector` on concrete static dispatch without candidate vectors or new allocations.

### Walkthrough

```mermaid
flowchart LR
    E["Existing RoutingEligibility"] --> R["Borrowed worker row"]
    R --> S["DefaultWorkerScorer"]
    S --> P["DefaultWorkerPicker"]
    P --> O["Existing WorkerSelectionResult"]
```

Pinned selection validates the worker, builds one stack row, and calls the same formula directly. Nonpinned selection computes the minimum active-prefill load once; the picker then enumerates eligible workers in the same order, builds one stack row at a time, and statically calls the scorer.

The contracts remain private. This PR adds no public API, filters, eligibility changes, candidate vectors, masks, reusable scratch, boxed dispatch, `SelectionService` changes, or policy adapter.

### Validation

- `cargo test -p dynamo-kv-router --features standalone-selection scheduling::selector::tests` — 22 passed.
- `cargo test -p dynamo-kv-router --features standalone-selection` — 908 passed, 2 ignored; 6 standalone HTTP tests passed.
- `cargo clippy -p dynamo-kv-router --features standalone-selection --all-targets -- -D warnings -A dead-code` — passed; `-A dead-code` covers one pre-existing indexer warning.
- `git diff --check`

### Benchmark Results

Criterion A/B and B/A runs compare the prior borrowed-row head (`073c7de8ac`) with the independently constructible stages (`2b9a176e5e`). Deltas are new versus old; no regression repeats across ordering.

| Workers | Temperature | A/B delta | B/A delta |
|---:|---:|---:|---:|
| 2 | 0.0 | -1.03% | -0.37% |
| 32 | 0.0 | +0.13% | +0.16% |
| 1,024 | 0.0 | -2.43% | +0.60% |
| 10,000 | 0.0 | -1.67% | -0.86% |
| 2 | 0.7 | -1.40% | -1.32% |
| 32 | 0.7 | +0.33% | +0.93% |
| 1,024 | 0.7 | +0.81% | +0.42% |
| 10,000 | 0.7 | -0.32% | -0.76% |


lib/kv-router/src/scheduling/selector.rs
```diff
@@ -112,6 +112,32 @@ struct LogitWeights {
...
+}
+
+trait WorkerPicker<C: WorkerConfigLike, S: WorkerScorer<C>> {
+    fn pick(
+        &self,
+        scorer: &S,
+        workers: &HashMap<WorkerId, C>,
+        request: &SchedulingRequest,
+        eligibility: RoutingEligibility<'_>,
+    ) -> (WorkerWithDpRank, f64);
```
PeaBrane: **AI-reviewed; human evaluation needed:** [DYN-3710](https://linear.app/nvidia/issue/DYN-3710/refactor-default-worker-selection-around-scorer-and-picker-traits) says these contracts should be hardened before later exposure to custom crates; [DYN-3711](https://linear.app/nvidia/issue/DYN-3711/add-opt-in-scorer-and-picker-policies-to-selectionservice) and [DYN-3712](https://linear.app/nvidia/issue/DYN-3712/publish-native-worker-scorer-and-picker-api) then plan to publish a shared eligible-worker input view with the scorer/picker APIs. `DefaultWorkerScorer::new` derives request- and candidate-relative state—including `min_active_prefill_tokens`—from one `workers`/`request`/`eligibility` universe, while `WorkerPicker::pick` independently accepts another universe and request. The current caller passes identical values, but the contract permits a scorer prepared against universe A to pick over universe B, or combine A’s scoring configuration with B’s temperature. Could both stages share one borrowed selection context/input view now—without candidate materialization or dynamic dispatch—so the future public invariant is enforced by construction?

PeaBrane (approved):
✅

lib/kv-router/src/scheduling/selector.rs
ishandhanani: [AI] Fixed in 9ab450d3e6. `WorkerSelectionInput` now owns the worker/request/eligibility context and precomputed score state, so the scorer and picker receive the same borrowed input by construction.

lib/kv-router/src/scheduling/selector.rs
```diff
@@ -134,113 +185,82 @@ impl DefaultWorkerSelector {
...
+    fn from_selector(selector: &'a DefaultWorkerSelector) -> Self {
+        Self {
+            kv_router_config: &selector.kv_router_config,
+            worker_type: selector.worker_type,
+        }
+    }
+}
+
+impl<C: Borrow<KvRouterConfig>> DefaultWorkerScorer<C> {
+    fn selection_weights(&self, request: &SchedulingRequest) -> LogitWeights {
```
PeaBrane, line 200: Could we add an observable selection test for the per-request overrides now owned by these extracted stages? Current selector tests set `router_config_override: None`, while score tests pass `LogitWeights` directly, so a regression in `overlap_score_credit`, `prefill_load_scale`, or `shared_cache_multiplier` precedence—and `router_temperature` in the picker—would not fail. A small table where each override reverses the selected worker would lock down the contract without testing implementation details.

PeaBrane (approved):
✅

lib/kv-router/src/scheduling/selector.rs
ishandhanani: [AI] Fixed in 2303907d0f. Added an observable selector test proving per-request overlap credit, prefill scale, shared-cache multiplier, and temperature overrides change the selected worker through the extracted scorer/picker stages.

PR #12522, opened 2026-08-01
feat(router): inject SelectionService worker selectors

ishandhanani:
Adds the smallest `SelectionService` injection seam needed by the native worker-selection work. The default selector remains concrete and inline; only the opt-in branch uses dynamic dispatch.

CLOSES: DYN-3711

### Implementation

- Threads an optional crate-private `WorkerSelector` factory through `SelectionServiceBuilder` and `SelectionCore`.
- Invokes the factory once per `SelectionEntry` and reuses that selector for requests in the routing partition.
- Keeps `DefaultWorkerSelector` inline with static dispatch and no added allocation.
- Allows custom selectors to hold actor-local non-`Sync` state without adding a mutex.
- Delegates transparently to the selected `WorkerSelector`; validation and accounting behavior are unchanged in this PR.

### Walkthrough

```mermaid
flowchart LR
    B["SelectionServiceBuilder"] --> C["SelectionCore"]
    C -->|"once per SelectionEntry"| S["Default or custom WorkerSelector"]
    S --> A["Existing scheduling and reservation flow"]
```

The factory is not called per request. Each scheduler actor owns one selector instance.

### Boundaries

- No scorer or picker adapter.
- No candidate columns, cost scratch, picker-index contract, filters, validation changes, configuration graph, EPP changes, runtime loading, or C ABI.
- Stacked on #12512.

### Validation

- `cargo test -p dynamo-kv-router --features standalone-selection` — 909 passed, 2 ignored; 6 standalone HTTP tests passed.
- `cargo clippy -p dynamo-kv-router --features standalone-selection --all-targets -- -D warnings -A dead-code` — passed; `-A dead-code` covers one pre-existing indexer warning.
- `git diff --check`
- The unchanged default selector benchmark matrix is recorded in #12512.


lib/kv-router/src/services/selection/core/mod.rs
```diff
@@ -492,7 +532,15 @@ impl SelectionCore {
...
         ));
-        let selector = DefaultWorkerSelector::new(Some(self.kv_router_config.clone()), WORKER_TYPE);
+        let selector = self.worker_selector_factory.as_ref().map_or_else(
+            || {
+                SelectionWorkerSelector::Default(DefaultWorkerSelector::new(
+                    Some(self.kv_router_config.clone()),
+                    WORKER_TYPE,
+                ))
+            },
+            |factory| SelectionWorkerSelector::Custom(factory()),
```
PeaBrane: AI review: This factory runs while `ensure_entry` holds the global `entries` write lock (acquired at line 492). A slow factory therefore blocks entry reads and registrations for every partition, and a factory that re-enters a `SelectionCore` path needing `entries` can deadlock on the non-reentrant `parking_lot::RwLock`. Could we construct it outside the global map lock, or use per-key initialization to preserve the once-per-entry guarantee without executing injected code under this lock?

lib/kv-router/src/services/selection/tests.rs
```diff
@@ -130,6 +161,77 @@ async fn register_worker_id(app: Router, worker_id: u64, max_tokens: Option<u64>
...
+    assert_eq!(factory_calls.load(Ordering::Relaxed), 1);
+
+    let reserved = post(
+        app.clone(),
+        "/select_and_reserve",
+        r#"{"model_name":"model","token_ids":[1,2,3,4],"selection_id":"custom-selector"}"#,
+    )
+    .await;
+    assert_eq!(reserved.status(), StatusCode::OK);
+    assert_eq!(response_json(reserved).await["worker_id"], 2);
```
PeaBrane: AI review: This does not reliably prove that the custom selector handled the request. The two workers have equal default scores, and `DefaultWorkerSelector` randomly reservoir-samples tied minima, so a regression that constructs the custom selector but still dispatches to the default can also return worker 2. Could the custom selector increment an invocation counter, or could the fixtures make the default and custom policies deterministically disagree?

PeaBrane (approved):
AI reviewed ✅

PeaBrane (approved):
✅

lib/kv-router/src/services/selection/core/mod.rs
ishandhanani: [AI] Fixed in 0325aff581. Entry slots now use per-partition once-only initialization, so factory code runs once without the global entries lock; the concurrency test proves separate partitions initialize in parallel.

lib/kv-router/src/services/selection/tests.rs
ishandhanani: [AI] Fixed in 0325aff581. The custom selector now exposes a shared invocation count, and the integration test asserts it handled the request while host-owned accounting replaces its fabricated fields.

lib/kv-router/src/services/selection/core/mod.rs
```diff
@@ -46,10 +47,60 @@ use super::types::{
...
+                let worker = selector
+                    .select_worker(workers, request, eligibility, block_size)?
+                    .worker;
+                if eligibility
+                    .pinned_worker()
+                    .is_some_and(|pinned| pinned != worker)
+                    || eligibility.validate_worker_rank(workers, worker).is_err()
+                {
+                    return Err(KvSchedulerError::NoEndpoints);
+                }
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 2b8cf33678. The custom branch now runs the same pinned-worker and overload preflight used by the default selector, preserving PinnedWorkerNotAllowed, PinnedWorkerOverloaded, and AllEligibleWorkersOverloaded. The existing invalid-selection integration test now also verifies a disallowed pin returns HTTP 400 before invoking custom policy code.

lib/kv-router/src/services/selection/tests.rs
```diff
@@ -130,6 +173,128 @@ async fn register_worker_id(app: Router, worker_id: u64, max_tokens: Option<u64>
...
+    });
+    assert_eq!(
+        model_registration.await.unwrap().status(),
+        StatusCode::CREATED
+    );
+    assert_eq!(
+        other_registration.await.unwrap().status(),
+        StatusCode::CREATED
+    );
+    assert_eq!(max_active_factories.load(Ordering::Relaxed), 2);
```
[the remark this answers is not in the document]
ishandhanani: [AI] Fixed in 2b8cf33678. Replaced the sleep/max-concurrency timing assertion with a condition-variable rendezvous: the first factory waits for the second to enter, with a bounded timeout that fails cleanly if factory construction is serialized. The targeted test passed 10/10 runs.

PR #12524, opened 2026-08-01
perf(bindings): NVTX spans around per-response egress GIL holds

tzulingk:
## Overview:

Add two NVTX spans on the Python→Rust response egress path in the Python bindings so its per-response, GIL-held cost is visible in `nsys` profiles. Instrumentation only — no behavior change, and a no-op unless built with `--features nvtx` + `DYN_ENABLE_RUST_NVTX=1`.

## Details:

`lib/bindings/python/rust/engine.rs` had no NVTX annotations, so the per-response GIL-held egress work sat outside every instrumented range and showed up only as unattributed GIL-hold time. This adds two RAII spans via the existing `dynamo_nvtx_range!` macro:

- **`egress.bridge.anext`** — the per-response Python→Rust hand-off (asking the Python async generator for its next item and scheduling it on the event loop).
- **`egress.depythonize`** — the per-response conversion of the yielded Python object into a Rust value.

Both guards are opened **before** `Python::with_gil`, so a profile also captures time spent **waiting** to acquire the GIL (contention), not just the hold. Each is scoped so the guard drops before any `.await`, keeping the thread-local NVTX push/pop balanced.

Same gating as the existing span sites (`preprocessor.rs` / `backend.rs` / routers): no-op unless built with `--features nvtx` and `DYN_ENABLE_RUST_NVTX=1`. Verified with `cargo check` on `dynamo-py3`.

`lib/bindings/python/rust/engine.rs` — the two `dynamo_nvtx_range!` guards in `demand_driven_python_stream` and `process_item`.

## Related Issues


lib/bindings/python/rust/engine.rs
```diff
@@ -113,6 +114,10 @@ fn demand_driven_python_stream(
     let anext = generator.getattr("__anext__")?.unbind();
     let stream = futures::stream::unfold((anext, locals), |(anext, locals)| async move {
         let next = Python::with_gil(|py| {
+            // Per-response bridge crossing: schedule the generator's `__anext__`
+            // on the event loop. GIL-held; this is where the ~37 us egress cost
+            // (and its ~1 ms tail) lives, previously outside every NVTX range.
+            let _nvtx = dynamo_nvtx_range!("egress.bridge.anext");
```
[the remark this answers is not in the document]
tzulingk: Good catch — fixed. Both spans now open **before** `Python::with_gil` so they also cover time spent waiting to acquire the GIL (the contention cost), not just the hold. Each guard is scoped (bridge: an explicit block that drops before `next.await`; depythonize: inside the `spawn_blocking` sync closure) so there's no `.await` between push and pop and the thread-local NVTX ranges stay balanced.

nv-kmcgill53 (changes requested):
A couple of things:
- We need to upgrade to the v3 nvtx api as v2 is deprecated in cuda 13
- I would like to see the entire trtllm worker be captured with the nvtx spans so that we can use it as a template for all other workers/frontends/components in dynamo

tzulingk:
@nv-kmcgill53 thanks for the review — both points are addressed:

### 1. Upgrade to the NVTX v3 API (v2 deprecated in CUDA 13) — `fc05180270`

The runtime NVTX module was calling `cudarc::nvtx::result::*`, which binds the legacy NVTX **v2** API and needs `libnvToolsExt.so` — the shared library CUDA 12.9 / CUDA 13 removed. I switched the backend to the header-only **NVTX v3** `nvtx` crate (the same one `kvbm` already uses); it compiles the vendored `nvtx3/nvToolsExt.h` via `cc`, so there is no shared library to link and it builds/runs on CUDA 13. The public `dynamo_nvtx_range!` macro API and the two-level gating (feature `nvtx` + `DYN_ENABLE_RUST_NVTX`) are unchanged.

### 2. Capture the whole TRT-LLM worker as a template — `fd515018eb`

Extended the *same* unified NVTX system to Python so a worker's own hot path lands on the same Nsight Systems timeline as the Rust core, under one gate (build `--features nvtx` + `DYN_ENABLE_RUST_NVTX=1`):

- New **`dynamo.nvtx.range()`** context manager, backed by NVTX **correlated ranges** (`start`/`end`) exposed from Rust on `dynamo._core`. Correlated ranges — not the thread-local push/pop stack — are required here because the workers are `asyncio` coroutines that interleave on one event-loop thread: a start/end pair stays correct across `await` points, where push/pop would mis-nest.
- Annotated `HandlerBase` with `worker.trtllm.{generate, preprocess, engine_submit}`.
- Per-token egress is already bracketed by the Rust bridge spans (`egress.bridge.anext` / `egress.depythonize`), so each decode step stays marked from the Rust side while the whole-request span bounds the loop.

The template for any other Python worker/frontend/component is just:

```python
from dynamo import nvtx

with nvtx.range("worker.<backend>.<phase>"):
    ...
```

No new dependencies, and it shares the exact feature/env gate as the Rust spans (no-op unless both are on).

**Validation:** `cargo check` + `clippy` with the feature **on and off**, and `cargo fmt`, on `dynamo-runtime` and `dynamo-py3`; pre-commit (isort/black/flake8/ruff); plus a unit test of the `range()` helper covering start/end pairing, the exception path, and the disabled no-op.

lib/bindings/python/src/dynamo/nvtx.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Safe in both synchronous and ``async`` code — the block may ``await`` or
+    ``yield`` while the range is open. A no-op unless NVTX is :func:`enabled`.
+    """
+    range_id = _nvtx_range_start(message)
+    try:
+        yield
+    finally:
+        if range_id:
+            _nvtx_range_end(range_id)
```
dmitry-tokarev-nv, line 60: **[P1] This adds a second, incompatible Python NVTX switch — a capture set up per this PR silently loses every marker the repo already has**

The stated goal is one unified timeline "under one gate". As written the repo ends up with two gates that do not know about each other:

| | module | env gate | call sites after this PR |
|---|---|---|---|
| existing | `dynamo.common.utils.nvtx_utils` | `DYN_NVTX` | 10 modules — `vllm/handlers.py`, `sglang/.../worker_handler.py`, `frontend/vllm_processor.py`, `common/multimodal/{image_loader,audio_loader,embedding_transfer,mm_kwargs_transfer}.py`, … |
| new here | `dynamo.nvtx` | `DYN_ENABLE_RUST_NVTX` + `--features nvtx` | 1 module — `trtllm/request_handlers/handler_base.py` |

`nvtx_utils.py:38` reads `DYN_NVTX` and nothing else; `lib/runtime/src/nvtx.rs`'s `init()` reads `DYN_ENABLE_RUST_NVTX` and nothing else. So a user who sets `DYN_ENABLE_RUST_NVTX=1` and profiles gets the Rust core plus TRT-LLM and **zero** markers from the ten existing modules; a user who sets `DYN_NVTX=1` gets those ten and **zero** from the Rust core and TRT-LLM. Either way the timeline is half-empty with no error, and this PR adds no doc saying to set both.

#13145 resolves exactly this by extending `nvtx_utils` and driving both halves from `DYN_NVTX` — its docstring change reads "The same variable switches on the Rust runtime's markers, so one setting covers a whole capture."

Secondary, on the same file: the docstring says `range` "is a cheap no-op" when disabled, and the PR description says "no-op unless built with `--features nvtx`". Neither holds — `_nvtx_range_start(message)` is an unconditional call into the C extension on every `__enter__`, feature or no feature. Measured on macOS ARM64 / CPython 3.13, compiling this PR's `range_start_impl`/`range_end_impl` and the three `#[pyfunction]`s verbatim with the `nvtx` feature **off** and driving them through this exact `nvtx.py`:

```
       2.6 ns/op   baseline: pass
     470.6 ns/op   with dynamo.nvtx.range('worker.trtllm.preprocess'): pass     <-- this PR
     137.6 ns/op   with nvtx_utils.annotate('worker.trtllm.preprocess'): pass   <-- main today
      20.6 ns/op     of which: the _nvtx_range_start FFI call
      15.1 ns/op     of which: a zero-arg zero-work pyfunction call
```

468 ns vs 135 ns net of loop overhead — 3.5x. At the three per-request sites this PR adds that is ~1.4 us/request, which is noise next to a forward pass, so this is **not** a performance objection to the current diff. It matters because the PR offers `dynamo.nvtx` as "the template for any other Python worker/frontend/component", and the module it is displacing documents the opposite contract on purpose: *"Pure Python no-ops: no C extension calls, no string allocations. The ENV var is read once at import time — no per-call branch overhead."* A template advertised as free that costs an FFI crossing per entry will eventually be put on a per-token path.

Both halves fall out of using `nvtx_utils` for the Python side. If `dynamo.nvtx` has to stay, the cheap version of the second half is to bind the name once at import, the way `nvtx_utils` does, so the disabled path never crosses into Rust:

```python
if _nvtx_enabled():
    @contextmanager
    def range(message: str) -> Iterator[None]:
        range_id = _nvtx_range_start(message)
        try:
            yield
        finally:
            if range_id:
                _nvtx_range_end(range_id)
else:
    class _NoOpRange:
        __slots__ = ()
        def __enter__(self): return self
        def __exit__(self, *exc): return False
    _noop = _NoOpRange()
    def range(message: str):
        return _noop
```

(Not offered as a `suggestion` block: `_nvtx_enabled()` is evaluated at import, so it needs `dynamo._core` to be initialised by then — worth the author checking before adopting. I did not verify that ordering.)

lib/runtime/Cargo.toml
```diff
@@ -23,7 +23,8 @@ compute-validation = [] # Enable validation and timing for compute macros
 tcp-low-latency = [] # Enable Linux-specific TCP optimizations (TCP_QUICKACK, SO_BUSY_POLL)
 # NVTX timeline annotations for Nsight Systems (compile-time off; also gated at runtime by DYN_ENABLE_RUST_NVTX).
 # Overhead: feature off = zero; feature on, env off = ~1ns AtomicBool load; feature on, env on = ~50ns/annotation.
-nvtx = ["dep:cudarc"]
+# Backed by the `nvtx` crate (vendored, header-only NVTX v3 — no libnvToolsExt.so, so it builds on CUDA 13).
+nvtx = ["dep:nvtx"]
```
dmitry-tokarev-nv, line 27: **[P2] Nothing in CI compiles `--features nvtx`, so this rewrite has no automated coverage at all**

`grep -rn nvtx .github/ container/` on this head returns only `codeowners/areas.yaml`, `license_overrides.yaml` and base-image SBOM entries — no workflow, no Dockerfile, no cargo invocation passes the feature. The same grep against `origin/main` gives the same answer. Every line this PR rewrites lives inside `#[cfg(feature = "nvtx")]`, so the green checks on this PR are compiling the *old* code path for these functions; the new `nvtx` crate is never even resolved.

That is what makes the backend swap risky rather than the swap itself: the only evidence it builds is the description's `cargo check`. I re-ran it and it does hold at this head —

```
$ cargo check -p dynamo-runtime --features nvtx --offline
    Checking dynamo-runtime v1.4.0
    Finished `dev` profile in 57.18s
$ find target/debug/build -name libnvtx.a
target/debug/build/nvtx-56f3c610ad00a7e1/out/libnvtx.a
```

— on macOS ARM64, which is not the platform that matters; I did not build it on Linux/CUDA 13, which is the case the swap exists for.

A `cargo check -p dynamo-runtime --features nvtx` (and the `dynamo-py3` equivalent) in the pre-merge lint job would cost seconds and would stop this path from silently rotting again. #13145 adds exactly that (`.github/workflows/pre-merge.yml`, +18); if that PR lands first this is covered, otherwise it is worth pulling the job in here.

dmitry-tokarev-nv (commented):
Reviewed at `fd51501`, checked out locally against merge-base `e6e135b` (the branch is 903 commits behind `main`).

## State: both requested changes are addressed on this head

I checked them against the code, not the description:

1. **"upgrade to the v3 nvtx api"** — done. `lib/runtime/Cargo.toml` swaps `cudarc {features=["nvtx"]}` for `nvtx = "1.3"`, and every call site in `lib/runtime/src/nvtx.rs` goes through it. I built it: `cargo check -p dynamo-runtime --features nvtx` succeeds on macOS ARM64 at this head, and the crate's `cc` step really does emit `libnvtx.a` from the vendored `nvtx3/` headers, so no `libnvToolsExt.so` link remains. The diff's claim that "the `nvtx` crate resolves the current OS thread id internally" is also accurate — `nvtx-sys/export.c`'s `ffi_name_thread` uses `syscall(SYS_gettid)` on Linux, identical to the `cudarc` code it replaces.
2. **"see the entire trtllm worker captured"** — done. `HandlerBase` now carries `worker.trtllm.{generate,preprocess,engine_submit}`.

So the `CHANGES_REQUESTED` state is stale; this is waiting on a re-review, not on the author.

## Verified, no finding

**Span scoping is correct at both Rust sites.** In `demand_driven_python_stream` the guard is confined to an explicit block that ends before `next.await`; in `process_item` it lives entirely inside the `spawn_blocking` sync closure, so push and pop land on the same blocking-pool thread. No `?` escapes either scope, so neither can leave a range unclosed.

**The Rust spans really are zero-cost with the feature off.** I compiled this PR's `NvtxRangeGuard` (feature off) into one crate and `dynamo_nvtx_range!("egress.depythonize")` into a separate consumer crate — `--release`, `lto = false`, so no cross-crate LTO to flatter the result — and disassembled:

```
_with_span:          # let _nvtx = dynamo_nvtx_range!("egress.depythonize"); x*3+7
    add  x8, x0, x0, lsl #1
    add  x0, x8, #0x7
    ret
```

Byte-identical to the same function without the macro: no call, no frame. The two new `engine.rs` spans cost nothing in the default build. The Python half is not free — that is P1 below.

## Before more work goes into this branch: #13145 looks like a superset

#13145 is open, by the same reviewer who requested changes here, and it:

- does the same `cudarc` → `nvtx` v3 swap in `lib/runtime/Cargo.toml` + `lib/runtime/src/nvtx.rs`;
- instruments the same three `lib/bindings/python/rust/engine.rs` sites under different names (`pybridge.invoke_generator` / `pybridge.anext_call` / `pybridge.decode_response` vs `egress.bridge.anext` / `egress.depythonize`);
- extends the existing `components/src/dynamo/common/utils/nvtx_utils.py` and puts both halves behind **one** env var, rather than adding a second Python module;
- adds the CI gate, the docs page and the tests this PR does not have.

The two PRs touch the same four files and will conflict. Worth settling which one lands before rebasing 903 commits.

One detail from that PR is worth carrying either way: it declares its own `ffi_*` symbols instead of using the `nvtx` crate's, because upstream `bindings.rs` declares `fn ffi_range_end(id: c_int) -> c_int` while `nvtx-sys/export.c` defines `void ffi_range_end(int id)`. I read both files in the vendored `nvtx-1.3.0` source and the mismatch is real. This PR's `range_end_impl` calls `nvtx::range_end!` through that declaration. I could **not** execute the consequence — it needs an attached NVTX injection library, which I do not have here — so treat the impact as **unverified**; only the declaration mismatch is confirmed.

## Findings

1 × P1, 2 × P2. Two are inline; the third has nothing in the diff to anchor to:

### [P2] The unit test claimed in the 2026-08-05 comment is not in this PR

That comment's Validation says "plus a unit test of the `range()` helper covering start/end pairing, the exception path, and the disabled no-op." `git diff e6e135b...fd51501 --name-only` returns 10 files and none of them is a test:

```
Cargo.lock
components/src/dynamo/trtllm/request_handlers/handler_base.py
lib/bindings/python/Cargo.lock
lib/bindings/python/rust/engine.rs
lib/bindings/python/rust/lib.rs
lib/bindings/python/rust/nvtx.rs
lib/bindings/python/src/dynamo/_core.pyi
lib/bindings/python/src/dynamo/nvtx.py
lib/runtime/Cargo.toml
lib/runtime/src/nvtx.rs
```

Either it was dropped before the push or it never existed. Flagging it because a reviewer reading that comment would reasonably not go looking for the test themselves. The same comment is also where the PR description stops matching the branch: the description still says "Instrumentation only — no behavior change" and points the reviewer at `engine.rs`, but the head also replaces the NVTX backend for **every existing** annotation site (`preprocessor.rs`, `backend.rs`, the KV-router selection path, the TCP egress routers) and adds a new public Python module to the `ai-dynamo-runtime` wheel. Worth refreshing before the next review pass.

lib/bindings/python/rust/engine.rs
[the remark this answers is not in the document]
dmitry-tokarev-nv: Confirming this thread is resolved on `fd51501`, with the verification it was missing.

Both guards are now opened before `Python::with_gil`, and the scoping holds: in `demand_driven_python_stream` the guard is confined to an explicit block that ends before `next.await`; in `process_item` it lives entirely inside the `spawn_blocking` sync closure, so push and pop run on the same blocking-pool thread. Neither block contains a `?` that could escape and leave a range open.

On the second half of the original comment — I measured what these two guards cost when the `nvtx` feature is off, which is the default build. Compiling this PR's `NvtxRangeGuard` into one crate and `dynamo_nvtx_range!("egress.depythonize")` into a separate consumer crate, `--release`, `lto = false`:

```
_with_span:          # let _nvtx = dynamo_nvtx_range!("egress.depythonize"); x*3+7
    add  x8, x0, x0, lsl #1
    add  x0, x8, #0x7
    ret
```

Identical to the same function without the macro — no call, no stack frame. So the cross-crate guard really is eliminated and these two sites are free in the shipped wheel. (macOS ARM64; I did not repeat it on x86-64.)

lib/bindings/python/src/dynamo/nvtx.py
dmitry-tokarev-nv: **[P1] Still live at `fd51501`.** The two gates are still separate, so a capture set up the way this file documents loses the image-loader markers inside the same TRT-LLM process. `lib/bindings/python/src/dynamo/nvtx.py:60`. Please make `dynamo.nvtx` read `DYN_NVTX` as well, or say in the module docstring that a capture must set both.

<details>
<summary>Marker counts from a capture run both ways, with controls</summary>

The head has not moved. I also merged the current `main` tip, `b5221198016b2cdf5858b7215347b9c45835b103`, which is 989 commits past the merge base, and re-checked on the merged tree. #13145, which would unify the two gates, is still open, so neither outcome that would close this has happened.

Both marker families live in one process. `handler_base.py` now carries 3 `dynamo.nvtx.range` sites. The same worker imports `dynamo.common.multimodal.image_loader` at `trtllm/encode_helper.py:15` and `trtllm/multimodal_processor.py:42`, and that module carries 4 `nvtx_utils` sites at `image_loader.py:115,136,210,240`.

Python half. I drove the real marker names from the 10 existing modules through the real `nvtx_utils` module, with a recording stand-in for the `nvtx` backend package that counts every emission. Three entry points per name: `start_range`, the context manager, and `range_decorator`.

| gate | `nvtx_utils.ENABLED` | emitted, all 41 names | emitted, the 4 image-loader names |
|---|---|---|---|
| `DYN_NVTX=1` (control) | True | 123 | 12 |
| `DYN_ENABLE_RUST_NVTX=1` (this file's setup) | False | 0 | 0 |
| neither (negative control) | False | 0 | 0 |
| both | True | 123 | 12 |

Rust half, which is the gate `dynamo.nvtx` actually rides: `lib/bindings/python/rust/nvtx.rs:34` calls `rt_nvtx::range_start_impl`. Real `nsys` capture, Linux x86_64, `dynamo-runtime` built with `--features nvtx`, 100 correlated ranges plus 100 push and pop ranges per run.

```
DYN_ENABLE_RUST_NVTX=1   nsys stats --report nvtx_sum:
   Instances  Style     Range
         100  StartEnd  :dynamo.nvtx.range/worker.trtllm.preprocess
         100  PushPop   :egress.depythonize
DYN_NVTX=1 only          SKIPPED: does not contain NV Tools Extension (NVTX) data.
neither                  SKIPPED: does not contain NV Tools Extension (NVTX) data.
```

So a capture set up per this docstring records 200 new markers and 0 of the 41 existing ones. A capture set up per the existing gate records the reverse. Nothing warns the user either way.
</details>

lib/runtime/Cargo.toml
dmitry-tokarev-nv: **[P2] Still true at `fd51501`, and now shown by mutation.** No lane compiles `--features nvtx`, so a break in this code path ships green. `lib/runtime/Cargo.toml:27`. Please add `cargo check -p dynamo-runtime --features nvtx` to the pre-merge Rust job. The good news is that the crate does build on Linux, which I could not show last round.

<details>
<summary>Mutation test with the lane staying green, plus the Linux build that was missing last round</summary>

Search at the live head, with the vendored SBOM files under `container/compliance/base_sboms/` excluded. `nvtx` appears 8 times in `.github/` and `container/`: one CODEOWNERS glob, two `dpkg` license overrides, and five lines about the Python `nvtx==0.2.14` pin, which serves `DYN_NVTX` and not this feature. Positive control: the same search for `--features` returns 27 lines, so the search does find things. None of the 27 names `nvtx`. No crate lists `nvtx` in a `default` feature set either, so feature unification never pulls it in.

Mutation test on the merged tree, GPU box, Linux x86_64, rustc 1.96.1.

| step | command | exit | `sha256` of `lib/runtime/src/nvtx.rs` |
|---|---|---|---|
| baseline | `cargo clippy --locked --no-deps -p dynamo-runtime --all-targets -- -D warnings` | 0 | `e50c8a67…` |
| mutated | the same command | 0 | `3a013ade…`, differs, so the edit landed |
| control | `cargo check --locked -p dynamo-runtime --features nvtx` | 101 | `3a013ade…` |
| restored | `git checkout HEAD -- lib/runtime/src/nvtx.rs` | n/a | `e50c8a67…`, matches the baseline |
| restored | `cargo check --locked -p dynamo-runtime --features nvtx` | 0 | `e50c8a67…` |

The mutation adds a call to a symbol that does not exist, inside the `#[cfg(feature = "nvtx")]` arm of `range_start_impl`. The lane passes anyway. With the feature on, the same tree fails:

```
error[E0425]: cannot find function, tuple struct or tuple variant `THIS_SYMBOL_DOES_NOT_EXIST_MUTATION_PROBE` in this scope
error: could not compile `dynamo-runtime` (lib) due to 1 previous error
```

Closing my own gap from last round. `cargo check --locked -p dynamo-runtime --features nvtx` now passes on Linux x86_64 with the pinned toolchain, on the merged tree, in 23.6 s. `cargo fmt -- --check`, `cargo doc --no-deps -p dynamo-runtime`, `cargo machete` and the clippy lane are green there too, and `nvtx 1.3.0` is already in `Cargo.lock` on `main` through the kvbm crates. The swap itself builds. Only the coverage is missing.
</details>

dmitry-tokarev-nv (commented):
Re-reviewed at `fd51501`. The head has not moved since the last round, so this is a re-verification, not a new read. I also merged the current `main` tip locally and re-ran the affected checks on the merged tree: `cargo fmt -- --check`, `cargo doc --no-deps -p dynamo-runtime`, `cargo machete`, the clippy lane, and `cargo check -p dynamo-runtime --features nvtx` are all green on Linux x86_64 with the pinned toolchain.

Two items stay open, both with new evidence on their existing threads.

1. P1 at `lib/bindings/python/src/dynamo/nvtx.py:60`. The two gates are still separate. Measured both ways: 41 existing marker names emit 123 markers under `DYN_NVTX=1` and 0 under `DYN_ENABLE_RUST_NVTX=1`, while an `nsys` capture of the new gate records 200 markers and none of the old ones. A TRT-LLM worker holds both families at once, so one capture always loses half. This one blocks approval.
2. P2 at `lib/runtime/Cargo.toml:27`. A mutation test confirms it: a call to a symbol that does not exist, placed inside the `nvtx` feature arm, leaves the clippy lane at exit 0 and fails only with the feature on.

The thread about guard placement in `engine.rs` still holds. Both guards open before `Python::with_gil`, and the scoping survives the merge with `main`, where `process_item` now calls the `depythonize_annotated` helper. I leave that thread closed.

Not approving yet, on the P1 alone.

PR #12525, opened 2026-08-01
refactor(aisimulate): extract generalized mocker engine and replayer

dreamtalen:
## Summary

This PR extracts Dynamo-neutral simulation execution into AISimulate while keeping Dynamo-specific composition in Dynamo.

* `aisimulate-core` is the public Rust crate. Its `engine` module owns vLLM, SGLang, and TensorRT-LLM scheduler simulation, framework-local G1 KV accounting, preemption, timing, single-rank execution, and attention-DP grouping.
* `aisimulate-core::replay` owns deterministic virtual time, logical-worker lifecycle, aggregated and disaggregated replay, placement/scaling seams, artifacts, canonical reports, and execution evidence.
* The crate root exposes the stable `EngineConfig`, `TimingModel`, `TimingModelConfig`, `ReplaySpec`, `ReplayReport`, and `Replayer` facade. Advanced contracts remain under explicit `engine` and `replay` paths.
* Dynamo owns Router and Planner composition and the Live Mocker driver, including transport, cancellation, publication, and production metrics.
* Live Mocker and offline replay now drive the same attention-DP-aware Generalized Mocker Engine instead of maintaining separate scheduler implementations.
* Python runners and the native `aisimulate._runtime` binding expose engine-only and Dynamo-composed replay to the CLI and Sweeper boundaries.
* The retired KVBM-specific Mocker replay/offload paths are removed. No KVBM compatibility adapter is added; standalone production KVBM crates and runtime functionality are unchanged.

The dependency and module direction is:

```text
Dynamo Router / Planner / Live adapters
                    ↓
        aisimulate-core::replay
                    ↓
        aisimulate-core::engine
```

`aisimulate-core` has no Dynamo, KVBM, or Tokio dependency.

### Review surface

The raw GitHub diff substantially overstates the implementation delta because file extraction is counted on both the deleted and added sides. The following line-level analysis separates pre-existing code, mechanical edits, tests, and the extracted or reworked review surface.

image

## Breaking changes

* The Mocker-only KVBM G1–G4 replay/offload implementation is removed without a compatibility adapter. Production KVBM remains unchanged, but simulated G2–G4 coverage is no longer available in the current replay stack.
* The corresponding `python -m dynamo.mocker` and replay options are removed: `--g1-backend`, `--num-g2-blocks`, `--num-g3-blocks`, `--enable-g4-storage`, `--offload-batch-size`, and the `--bandwidth-g*-to-g*-gbps` options. Framework-native G2 simulation is separate future work.
* The `dynamo-kv-replay-parity` protocol now covers native G1. Historical KVBM parity campaigns must use a historical revision and its matching protocol.
* Disaggregated attention-DP is rejected before materialization until rank-aware handoff identity is implemented in ai-dynamo/dynamo#12965.

## Ownership and entrypoints

| Layer | Ownership |
| -- | -- |
| Stable Rust facade | `aisimulate-core` |
| Scheduler, cache, G1 KV accounting, timing, attention-DP engine | `aisimulate-core::engine` |
| Virtual-time replay, worker lifecycle, Agg/Disagg runtimes, reports/artifacts | `aisimulate-core::replay` |
| Engine runner, traffic materialization, reporting, native PyO3 execution seam | AISimulate Python package |
| Router/Planner policies and replay composition | Dynamo |
| Live driver, transport, cancellation, publication, production metrics | Dynamo Mocker |

Current entrypoints are:

* Engine-only replay: `python -m aisimulate.replay`
* Dynamo-integrated replay: `python -m dynamo.replay`
* Configuration sweeps: `Sweeper(runner_factory=...).run(config)` and the examples under `aisimulate/examples/sweeper/`

The two replay modules share their engine, topology, traffic, replay-mode, SLA, and output argument schema. The Dynamo entrypoint adds optional Router, Planner, AIC, and online integration; its no-adapter form remains compatible with engine-only replay.

## Correctness hardening

The review and latest-main reconciliation preserve one execution path while adding focused safeguards:

* Rebasing onto `29930c4cea` incorporates ai-dynamo/dynamo#13052; exact output-token plans remain authoritative, and disaggregated prefill truncates its local plan without mutating decode.
* Grouped Live Mocker shutdown drains and acknowledges the final pass boundary before cancellation.
* Production `LiveEngine::submit()` bursts are forwarded without limiting admission to one request per pass; the public compatibility request lane remains unbounded while grouped-engine command lanes remain bounded.
* Attention-DP ranks report FPM duration at the modeled shared group boundary.
* Aggregated replay dispatches placement releases from every topology callback.
* SGLang supports disabled prefix caching while rejecting unsupported KV-token-ID and non-chunked-prefill controls.
* vLLM cached-prefix admission state survives preemption and propagates consistently through replay artifacts and Dynamo completion usage.
* SGLang cache metrics remain latched across grouped completion.
* Disaggregated attention-DP and custom replay worker selectors fail early rather than executing unsupported semantics.

## API walkthrough

<details><summary>Expand API walkthrough</summary>

### Stable facade

`aisimulate-core` exposes the common stable surface at the crate root:

```rust
aisimulate_core::{
    EngineConfig,
    TimingModel,
    TimingModelConfig,
    ReplaySpec,
    ReplayReport,
    Replayer,
}
```

Advanced engine and adapter contracts use explicit `aisimulate_core::engine` and `aisimulate_core::replay` paths.

### Overall flow

```text
python -m aisimulate.replay
    ↓
EngineReplayRunner
    ↓
aisimulate._runtime
    ↓
Replayer<RoundRobinComposition>

python -m dynamo.replay / Sweeper
    ↓
DynamoReplayRunner
    ↓
Dynamo Router / Planner composition
    ↓
Replayer<DynamoComposition>
    ↓
GeneralizedMockerEngine<SchedulerRank>
```

Live Mocker does not use the virtual-time Replayer, but shares its grouped engine:

```text
Offline Replayer ── virtual clock ──┐
                                    ├─ Generalized Mocker Engine
Live Mocker ─────── Tokio clock ────┘
```

### Generalized Mocker Engine

Main files:

* `aisimulate/crates/core/src/engine/generalized/contracts.rs`
* `aisimulate/crates/core/src/engine/generalized/engine.rs`
* `aisimulate/crates/core/src/engine/scheduler/`

`SchedulerRank` owns one rank's scheduler, KV accounting, preemption, and timing state. `GeneralizedMockerEngine<C>` owns attention-DP membership, readiness, pass barriers, and modeled group completion. Offline replay supplies virtual time; Live Mocker supplies Tokio/wall-clock events through a Dynamo-owned driver.

### Replayer

Main files:

* `aisimulate/crates/core/src/replay/replayer.rs`
* `aisimulate/crates/core/src/replay/spec.rs`
* `aisimulate/crates/core/src/replay/engine.rs`

`Replayer<C: ReplayComposition>` owns virtual time, the event queue, logical-worker lifecycle, Agg/Disagg runtime state, trace collection, artifacts, and reports.

* `Replayer::new(...)` provides round-robin placement with no scaling.
* `Replayer::with_composition(...)` accepts Dynamo-owned placement and scaling policies.
* `Replayer::run_with_artifacts(...)` uses the shared aggregated runtime with an optional Replay-owned artifact sink.

Placement, scaling, engine, and invariant failures retain typed provenance.

### Python runner and CLI boundary

Main files:

* `aisimulate/src/aisimulate/runner.py`
* `aisimulate/src/aisimulate/replay/`
* `components/src/dynamo/replay/simulation.py`
* `components/src/dynamo/replay/main.py`

`EngineReplayRunner` materializes a serializable `ReplaySpec`, calls `aisimulate._runtime`, and normalizes `ReplayReport`. `DynamoReplayRunner` implements the same Runner contract while constructing Dynamo Router and Planner providers.

`python -m dynamo.replay` composes the shared AISimulate base parser/lowering with Dynamo-only extensions; it does not duplicate the engine-only schema.

### Router, Planner, and Live adapters

Dynamo reuses the existing `PlacementPolicy<Request>` boundary, including round-robin and `KvRouterPlacement`, through `ReplayComposition`. Planner is injected through the scaling-policy seam.

Live Mocker constructs the same grouped engine as offline replay. Dynamo remains responsible for Tokio driving, transport, cancellation, output publication, and production metrics.

</details>

## Validation

Final HEAD: `23518163eb`

### Rust

* `cargo test -p aisimulate-core -p dynamo-mocker -p dynamo-mocker-backend --no-fail-fast`
  * `aisimulate-core`: 449 tests passed
  * `dynamo-mocker`: 194 tests passed
  * `dynamo-mocker-backend`: 20 tests passed
* `cargo clippy -p aisimulate-core -p dynamo-mocker -p dynamo-mocker-backend --all-targets -- -D warnings`
* `cargo fmt --all --check`
* `pre-commit run --all-files`
* `git diff --check`
* Root, Python-binding, and KVBM-binding `cargo metadata --locked` checks passed.
* A tracked-only EPP/ext-proc Docker workspace smoke passed `cargo metadata --locked`; both builders now copy `aisimulate/crates/` alongside `lib/` before Cargo resolves the root workspace.
* A tracked-only Planner wheel context containing every root Cargo workspace member passed `cargo metadata --locked` and built the native AISimulate wheel; a regression test now enforces complete workspace COPY coverage.

Focused regressions cover grouped shutdown, production-lane submit batching, attention-DP FPM normalization and rank identity, vLLM/SGLang MTP lifecycle boundaries, cached-prefix propagation, SGLang cache-metric latching, bounded same-timestamp retry, pass-end KV visibility, topology release evidence, artifact capture, unsupported topology/selector rejection, and SGLang prefix-caching behavior.

### Python and packaging

* The dedicated Planner test image builds for both `linux/amd64` and `linux/arm64`, is readable as the CI runner identity (`1001:1001`), and collects the intended 25 optional AISimulate/Planner tests.
* Native Linux arm64 validation with current AISimulate and `ai_dynamo_runtime` wheels passed the exact Planner selector (`25 passed, 134 deselected`), the online KV Router CLI regressions (`2 passed`), and SGLang AIC parity (`6 passed, 6 deselected`).
* The Planner image's two standalone wheel/build-context source-contract tests passed.
* 173 targeted Runner, replay CLI, traffic, Sweeper, and Dynamo replay tests passed.
* The standalone AISimulate wheel builds and imports `aisimulate._runtime` on Python 3.12.
* Wheel metadata confirms no `ai-dynamo` dependency and no console script.
* CODEOWNERS coverage remains 100%.

### Replay parity

The authoritative native-G1 correctness campaign compared candidate `23518163eb` with pinned merge-base main `975d6a15a6` using the first 5,000 requests from the canonical Mooncake trace. The frozen slice SHA-256 is `3892ae19ae480b643155f0c6b9d798591cbe2e73bec6a0fa5ae3d3bc0332fb8a`, matching the `dynamo-kv-replay-parity` protocol.

The four skill-aligned KV-aware rows used the current qualified internal-polynomial seeds. Each baseline and candidate row ran twice in a separate process; every revision/row pair produced one deterministic digest, completed 5,000/5,000 requests, and matched after accounting for the intentionally retired KVBM configuration schema:

| Engine | Topology | Baseline raw SHA-256 | Candidate raw / normalized semantic SHA-256 | Pressure | Requests with reuse |
| -- | -- | -- | -- | -- | -- |
| vLLM | Aggregated | `cbda006db3b1b4c7fe4eda15412469aa237b5ada44d1aaa39c5261b136ef6180` | `b69f998ec1978f5bf00f172d7648467af730e29cb57d7e42210664fbc220ad27` | 1 preemption | 4,917 |
| vLLM | Disaggregated | `dd693851bdc43cff67fe049ef77cf0f4d41e85c869a0bfd995a85d72147e6ec0` | `b40ab4ef3bae09e54c8c078e387012bc1bae20e926c69d79aadaf56dec3943af` | 14 preemptions | 4,988 |
| SGLang | Aggregated | `362d4d323e5fd8d813d6f437875b722bc259f014aa4bb254d843779d3df6666e` | `681438bff6a6420c59d831fbe672dce428e0cd4ad33a2a918c594e5f3d18e85c` | 1 retraction | 4,840 |
| SGLang | Disaggregated | `c0488fcb82bb5c66b1c0d13b6b0b9c043fdadbf6f90ea171a763c38b70027afd` | `c5289b3b709b46ebe0873efbeb9441ad48a722b5f58f35207a0050e94c03eff1` | 14 retractions | 4,990 |

This result is `PASS_WITH_SEMANTIC_EXCEPTIONS` under the skill terminology. Raw baseline/candidate SHA values differ only because the baseline still serializes the KVBM G1–G4/offload/bandwidth configuration keys that this PR explicitly removes. Removing exactly those baseline-only retired keys produces the normalized semantic SHA shown above; no scheduler, routing, timing, request, lifecycle, evidence, or report-value difference remains.

Every pressure record was fully readmitted. Aggregated rows used all workers 0–3 and recorded 5,000 immediate, zero queued placements. Disaggregated rows used both prefill and decode workers 0–1, recorded 5,000 complete handoffs, and recorded 10,000 immediate, zero queued placements across the two pools. The pinned main baseline requalified vLLM aggregated reuse at 4,917 requests; baseline and candidate agree exactly.

A supplemental 1,000-request preflight also added Round-robin to the four KV Router rows. All eight supplemental rows were internally deterministic and matched across revisions after the same retired-schema normalization. The updated 5,500-block vLLM disaggregated seed exposed a missing deferred-ready fairness path during extraction; commit `23518163eb` restores it, and both the 5,000-row authoritative row and the supplemental Round-robin/KV Router rows now match.

This section reports the skill's authoritative correctness stage. The separate Linux 60-pair performance campaign was not run as part of this PR validation.

Dependency-firewall tests confirm that `aisimulate-core` contains no Dynamo or KVBM dependency. Existing Mooncake/KV Router benchmark callers retain their replay/artifact entrypoints.

## Related issues

* ai-dynamo/dynamo#12446 established the Sweeper boundary and is included in `main`.
* ai-dynamo/dynamo#12965 tracks rank-aware handoff identity for disaggregated attention-DP.

---

Open in Devin Review

## Summary by CodeRabbit

* **New Features**
  * Added an engine-neutral replay system supporting aggregated and disaggregated simulations across vLLM, SGLang, and TensorRT-LLM.
  * Added Python and Sweeper runner integration, configurable timing models, workload and trace loading, scaling policies, handoff simulation, KV-cache events, and optional per-request or raw reports.
  * Added attention data-parallel routing and grouped live-engine support.
* **Bug Fixes**
  * Added validation for invalid configurations, malformed traces, unsupported topologies, and runtime errors.
* **Documentation**
  * Updated replay, engine, compatibility, and configuration guidance.


tedzhouhk (changes requested):
I finished reviewing the public API. The engine/replayer extraction and the Engine/Dynamo composition boundaries look good to me; I do not have API concerns with those parts.

I am requesting changes specifically for the user-facing CLI. The proposed `aisimulate predict` interface mixes several abstraction levels and is difficult to reason about. A few examples:

- `--isl` and `--osl` are top-level CLI flags, while `concurrency`, `request_count`, `request_rate`, and arrival controls for the same traffic model live in YAML, sometimes at different YAML levels.
- Trace input comes through a separate `--traffic` YAML, while `config.workload` can override it. This gives one workload multiple input sources with non-obvious precedence.
- Deployment is split between CLI flags (`model`, `backend`, `system`, `tp-size`, and `replicas`) and config fields (`num_workers`, `engine_args`, and `prefill`/`decode`), with multiple aliases for some fields.
- `concurrency` appears both inside `workload` and as a top-level `ReplaySpec` field. The latter is understandable as a Sweeper-resolved value for KV-load candidates, but exposing that intermediate contract through a concrete one-shot CLI is confusing.
- The existing replay CLI's domain split—traffic, Router, prefill/decode engines, and Planner—is easier for users to understand.

Please defer the CLI changes from this PR for now. I suggest landing the engine/replayer/API extraction without establishing this new public CLI contract. Keep single-run replay and Sweeper as two separate user-facing paths for the moment; do not partially combine them through `aisimulate predict` until we have discussed and agreed on a coherent, typed CLI/config model. We can introduce the consolidated `predict`/`sweep` surface in a follow-up after that design discussion.

lib/mocker/src/generalized_live.rs
```diff
@@ -0,0 +1,1246 @@
...
+    }
+
+    async fn serve_pass_boundary(
+        &mut self,
+        mut requests: mpsc::Receiver<BoundaryRequest>,
+    ) -> Result<bool> {
+        loop {
+            let request = tokio::select! {
+                biased;
+                _ = self.cancel_token.cancelled() => return Ok(false),
```
tedzhouhk, line 427: **[P1] Wait for the final grouped boundary before cancelling the actor**\n\nTerminal output delivery does not mean the completion dispatcher has sent `Finish` yet. `LiveRunSession` immediately calls `engine.shutdown()`, which cancels this token; because this select is biased, the actor can drop the boundary receiver before `dispatch_pass_completion` calls `boundary.finish()`, turning a successful replay into `grouped live pass boundary is closed` and losing the report. An 8-thread stress test with fresh runtime/engine state reproduced this at iteration 9,814/10,000. Please add an orderly boundary-drain/ack handshake before cancellation plus a deterministic regression test.

aisimulate/crates/engine/src/scheduler/rank.rs
```diff
@@ -0,0 +1,840 @@
...
+            pending: PendingPass {
+                effects: completion_effects,
+            },
+        })
+    }
+
+    fn complete_pass(
+        &mut self,
+        mut pending: Self::PendingPass,
+        _end_ms: f64,
```
tedzhouhk: **[P1] Normalize active-rank FPM to the modeled group barrier**\n\nIgnoring the group completion time leaves active ranks with rank-local durations, while idle ranks use the shared barrier and offline lowering also rewrites every rank to that barrier. A DP2 repro produced a 40 ms group boundary but live FPMs of `(rank0=20 ms, rank1=40 ms)` versus offline `(40 ms, 40 ms)`. Planner regresses directly against this wall time, so live/offline scaling inputs diverge. Please report the modeled shared boundary for active ranks as well, without including a late wall-clock driver wakeup, and cover unequal active-rank latencies.

aisimulate/src/aisimulate/runner.py
```diff
@@ -0,0 +1,742 @@
...
+                raise ValueError(
+                    f"engine provider {role} aic_nextn={nextn} conflicts with "
+                    f"timing_model.config.nextn={configured_nextn!r}"
+                )
+            timing_config["nextn"] = nextn
+            timing_model["config"] = timing_config
+            rank["timing_model"] = timing_model
+
+    return {
+        "dp_size": dp_size,
```
tedzhouhk, line 720: **[P2] Reject disaggregated attention-DP before materializing the engine**\n\nThe public `Replayer` and `aisimulate --stack engine` currently accept disaggregated DP2 even though ranked disaggregated handoff is documented as unsupported. A minimal one-logical-worker run reported scheduler rank 1 as logical worker 1, and a custom scaling composition lost the rank-1 FPM when pruning against active logical worker IDs. The Dynamo-backed wrapper still rejects this, but the new public engine runner silently returns invalid results. Please reject prefill/decode `dp_size != 1` in the replay core, with an early runner error, until the identity and handoff model is implemented.

aisimulate/crates/replay/src/agg.rs
```diff
@@ -0,0 +1,1066 @@
...
+                WorkerTopology {
+                    worker_id: id,
+                    scheduler_ids: Vec::new(),
+                },
+                self.now_ms,
+            )?;
+            self.dispatch_placements(placements)?;
+        }
+        let placements = self.placement.topology_settled(self.now_ms)?;
+        let released = placements
```
tedzhouhk: **[P2] Record releases returned by every topology callback**\n\nThe placements returned by `worker_ready`, `worker_draining`, and `worker_removed` are dispatched above, but only the IDs returned by `topology_settled` are accumulated here. A custom public composition repro completed the released request while recording `topology_released_request_uuids=[]`; the merge-base implementation accumulated all four callback batches. Please extend `released` before dispatching each topology callback and restore a custom-policy regression test. Current built-in KV Router happens to release only from `topology_settled`, so that does not make the public evidence contract safe.

aisimulate/crates/engine/src/config.rs
```diff
@@ -0,0 +1,713 @@
...
+        ensure!(
+            self.kv_bytes_per_token.is_none_or(|bytes| bytes > 0),
+            "kv_bytes_per_token must be positive"
+        );
+        ensure!(
+            self.kv_transfer_bandwidth
+                .is_none_or(|bandwidth| bandwidth.is_finite() && bandwidth >= 0.0),
+            "kv_transfer_bandwidth must be finite and non-negative"
+        );
+        if self.backend == Backend::Sglang {
```
tedzhouhk: **[P2] Reject or implement the generic SGLang controls**\n\nThe new public `EngineConfig` accepts `emit_kv_token_ids=true`, `enable_prefix_caching=false`, and `enable_chunked_prefill=false` for SGLang, but the backend does not honor them. A focused run was accepted by the factory, still chunked an 8-token prompt at the SGLang chunk size, reused the repeated prefix, and emitted `stored_token_ids=[None]`. The backend limitation predates this extraction, but the newly documented strict API should not silently accept unsupported settings. Please implement them or reject/namespace them during SGLang validation.

tedzhouhk (changes requested):
Independent adversarial verification confirmed the two blocking P1s and the three scoped P2s below. I am not including the previously considered mid-pass CancelDestination issue because both shipped drivers defer that command until the pass has completed.

dreamtalen:
@tedzhouhk Thanks, removed the new `aisimulate predict `user-facing CLI and single-run replay remains `python -m dynamo.replay` in this PR.

lib/mocker/src/generalized_live.rs
dreamtalen: Thanks, fixed in 40c1791816.
`LiveRunSession` now waits for an explicit completion-boundary drain before orderly shutdown. Added a regression test `terminal_delivery_waits_for_grouped_completion_boundary_before_shutdown`, which pauses the dispatcher immediately before finish().

aisimulate/crates/engine/src/scheduler/rank.rs
dreamtalen: Thanks, fixed in 40c1791816.
Both active and idle ranks now use max(rank end_ms). Added a regression test `attention_dp_active_rank_fpm_uses_the_modeled_shared_boundary`.

aisimulate/src/aisimulate/runner.py
dreamtalen: Fixed in 40c1791816.
The replay core now rejects prefill or decode dp_size != 1 before runtime/topology materialization.

aisimulate/crates/replay/src/agg.rs
dreamtalen: Thanks, fixed in 40c1791816.
Aggregated replay now accumulates request IDs and dispatches placements returned by every topology callback: worker_ready, worker_draining, worker_removed, and topology_settled. Added a regression test `aggregated_topology_callbacks_dispatch_and_record_every_released_request`

aisimulate/crates/engine/src/config.rs
dreamtalen: Fixed in 40c1791816 by explicitly rejecting the unsupported combinations rather than silently accepting them.
SGLang validation now reports field-specific errors for:
- emit_kv_token_ids=true
- enable_prefix_caching=false
- enable_chunked_prefill=false

Added a regression test `sglang_rejects_unsupported_generic_controls_at_validation_and_factory_boundaries`

aisimulate/crates/engine/src/common/sequence.rs
```diff
@@ -0,0 +1,467 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::hashing::{
+    BlockHash, SequenceHash, compute_block_hash_for_tokens, compute_next_sequence_hash,
+};
+use rand::random;
+
+const MOCKER_SALT_HASH: u64 = 1337;
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. Removed the duplicated `MOCKER_SALT_HASH` and now use `common::hashing::XXH3_SEED` at both sequence-hash call sites. The Engine suite and strict Clippy pass.

aisimulate/crates/core/src/engine/config.rs
```diff
@@ -0,0 +1,750 @@
...
+            "max_model_len is supported only for backend=vllm"
+        );
+        ensure!(
+            self.speedup_ratio.is_finite() && self.speedup_ratio >= 0.0,
+            "speedup_ratio must be finite and non-negative"
+        );
+        ensure!(
+            self.decode_speedup_ratio.is_finite() && self.decode_speedup_ratio >= 0.0,
+            "decode_speedup_ratio must be finite and non-negative"
+        );
```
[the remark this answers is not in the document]
dreamtalen: No code change here: zero is an intentional, tested sentinel in the existing contract. `modeled_duration_ms` divides only when the ratio is greater than zero; zero preserves the unscaled duration. Transfer timing similarly returns no modeled delay when bandwidth is zero, before any division. Rejecting zero would change established behavior, so this finding is not applicable to this extraction.

aisimulate/crates/core/src/engine/kv_manager/vllm_backend.rs
```diff
@@ -673,30 +669,25 @@ impl VllmKvManager {
...
-            KvCacheEventData::Removed(KvCacheRemoveData {
-                block_hashes: full_blocks
-                    .into_iter()
-                    .map(ExternalSequenceBlockHash)
-                    .collect(),
-            })
+            KvEventData::Removed {
+                block_hashes: full_blocks,
+            }
         };
```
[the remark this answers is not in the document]
dreamtalen: No code change here after auditing construction. Every production Stored-event path retains local hashes whenever KV publication is enabled: `VllmCore` sets `retain_local_hashes` when publishers are present, and publication returns before constructing an event when no publisher exists. Prefix-caching-off construction also removes the publishers. The fallback is therefore unreachable under the enforced production invariant and predates this extraction.

aisimulate/src/aisimulate/runner.py
```diff
@@ -0,0 +1,760 @@
...
+                f"{deployment_backend_version!r}"
+            )
+        timing_model = rank.get("timing_model")
+        uses_aic_timing = timing_model is None or (
+            isinstance(timing_model, dict)
+            and timing_model.get("type") == "external"
+            and timing_model.get("provider") == "aic"
+        )
+        if uses_aic_timing:
+            aic_timing_overrides["backend_version"] = deployment_backend_version
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. Explicit AIC `timing_model.config.backend_version` is accepted when it matches `BackendDeploymentSpec.backend_version` and rejected with a field-specific conflict when it differs. Added `test_runner_accepts_matching_backend_version_in_explicit_aic_timing` and `test_runner_rejects_conflicting_backend_version_in_explicit_aic_timing`; Runner tests pass 22/22.

lib/backend-common/examples/mocker/src/engine.rs
```diff
@@ -444,7 +445,7 @@ impl LLMEngine for MockerBackend {
             },
         );
 
-        if request_tx.send(direct).is_err() {
+        if request_tx.send(direct).await.is_err() {
             self.active.remove(&uuid);
             return Err(engine_shutdown("scheduler is not accepting requests"));
         }
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. The bounded enqueue now uses a cancellation-aware biased select for request cancellation, backend shutdown, and queue send. An RAII active-request guard cleans active/KV state on cancellation, send failure, or a dropped future. Added deterministic blocked-request, blocked-backend, and closed-queue tests; `dynamo-mocker-backend` passes 22/22.

lib/mocker/src/generalized_live.rs
```diff
@@ -0,0 +1,1357 @@
...
+        tokio::select! {
+            biased;
+            result = self.event_tx.send(event) => {
+                result.map_err(|_| anyhow!("grouped live engine event lane is closed"))
+            },
+            _ = self.cancel_token.cancelled() => {
+                bail!("grouped live engine stopped while publishing effects")
+            },
+        }
+    }
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. Cancellation while blocked on event publication is now a typed orderly shutdown, while an unexpectedly closed event lane remains an error. Added `cancellation_while_blocked_publishing_is_orderly` and `unexpectedly_closed_event_lane_remains_an_error`; the Mocker suite passes 189/189.

lib/mocker/src/grouped_scheduler/compatibility.rs
```diff
@@ -0,0 +1,260 @@
...
+                    entry.destination_request = None;
+                }
+                (entry.source_request.is_none() && entry.destination_request.is_none())
+                    .then_some(*id)
+            })
+            .collect::<Vec<_>>();
+        for id in ids {
+            self.remove_if_finished(id);
+        }
+    }
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. `cancel_request` and `finish_destination_request` now remove only entries that matched the target request and became empty. Added multi-handoff regressions proving unrelated empty entries survive both operations; the Mocker suite passes 189/189.

lib/mocker/src/live.rs
```diff
@@ -737,22 +887,20 @@ async fn run_event_dispatcher(
...
+                let Some(failed) = dispatch_output_batch(signals, &routes, &cancel) else {
                     return Ok(());
-                }
+                };
+                // This is the route-delivery boundary, not merely the
+                // scheduler-to-dispatcher enqueue boundary. Failed routes are
+                // returned to the grouped dispatcher, which applies native
+                // cleanup synchronously through `GroupedPassBoundary`.
+                let _ = delivered.send(failed);
             }
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. Ordered send/ack now propagates a typed cancellation outcome instead of `OrderedLaneClosed`. The completion dispatcher handles that as orderly shutdown, while unexpected lane closure remains an error. Boundary finish uses a biased ready-error-first wait, covered by `ready_boundary_error_wins_over_cancellation`; dropped-ack cancellation and non-cancellation regressions are also covered.

lib/mocker/src/replay/offline/extensions/kv_router/mod.rs
```diff
@@ -629,7 +601,31 @@ impl OfflineReplayRouter {
...
+                    removed.block_hashes.first(),
+                    removed.block_hashes.last(),
+                ),
+                dynamo_kv_router::protocols::KvCacheEventData::Cleared => "cleared".to_string(),
+            };
+            self.indexer.apply_event(event).with_context(|| {
+                format!(
+                    "failed to apply replay KV event worker={worker_id} dp_rank={dp_rank} event_id={event_id} data={data}"
+                )
+            })?;
```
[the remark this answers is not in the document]
dreamtalen: Fixed in `0181825317`. KV-event diagnostics now capture only a Copy `KvEventSummary` on the success path and perform formatting lazily inside `with_context`, while preserving the stored/removed/cleared diagnostic details.

aisimulate/src/aisimulate/runner.py
```diff
@@ -0,0 +1,783 @@
...
+            deployment.backend_version,
+            deployment.parallel_config,
+            raw_decode,
+            "decode",
+        )
+        for role, role_config in (("prefill", prefill), ("decode", decode)):
+            if role_config["dp_size"] != 1:
+                raise ValueError(
+                    "disaggregated engine replay requires "
+                    f"{role} dp_size=1; attention-DP handoff identity is not implemented"
```
tedzhouhk, line 241: **disaggregated attention-DP limitation**

Before this PR, legacy `dynamo-replay` already rejected disaggregated DP>1 because rank-aware prefill→decode handoff identity was not implemented, so this underlying feature gap is pre-existing. 
If fix is deferred, can we add an explicit P0 TODO and fix ASAP?

aisimulate/crates/engine/src/config.rs
```diff
@@ -0,0 +1,750 @@
...
+                "aic_nextn_accept_rates requires aic_nextn"
+            );
+        }
+        if self.backend == Backend::Sglang {
+            ensure!(
+                !self.emit_kv_token_ids,
+                "emit_kv_token_ids=true is not supported for backend=sglang"
+            );
+            ensure!(
+                self.enable_prefix_caching,
```
tedzhouhk: **SGLang prefix_caching=false limitation**

Before this PR, the SGLang simulator could not disable prefix caching: `enable_prefix_caching=false` was accepted but silently behaved as `true`, so the fidelity gap is pre-existing. 

After this PR, rejecting the unsupported value makes the API honest. If fix is deferred, can we add an explicit P0 TODO and fix ASAP?

tedzhouhk (commented):
Two pre-existing replay capability gaps are now exposed as user-facing Runner/Sweeper integration failures by this refactor. Please either close them in this PR or make the limitations explicit P0 TODOs at the capability boundary and prevent unsupported candidates from being generated.

aisimulate/README.md
```diff
@@ -16,10 +16,14 @@ AI Simulate is a standalone Python distribution in the Dynamo repository. Its fi
 specifications. The package does not depend on `ai-dynamo`.
 
 Sweeper accepts a replay `RunnerFactory` through its Python API. Optional feature adapters own
-their search spaces and runtime hooks. A backend-only sweep can use a Dynamo-free replay runner;
+their search spaces and runtime hooks. An engine-only sweep can use a Dynamo-free replay runner;
 a sweep configured with Dynamo Planner or Router adapters uses Dynamo's runner composition.
 
-Install AI Simulate by itself for backend-only development:
+For a Dynamo-integrated single replay run, use `python -m dynamo.replay`. For configuration
```
tedzhouhk: **[CLI API] Split the single-run CLI at the same package boundary as the implementation**

Removing the unified `aisimulate predict/sweep` CLI addresses the earlier confusing surface, but the current result swings too far: the standalone AISimulate Engine/Replayer has no single-run CLI, while `dynamo.replay` owns both engine-only fields and Dynamo adapter fields. Please make AISimulate own an engine-only simulation CLI and its base config schema (engine/topology, traffic, replay mode/output), then have `python -m dynamo.replay` reuse/extend that base with only Dynamo-owned adapter fields such as Router and Planner configuration (plus any genuinely Dynamo-only modes). This should be composition of one parser/schema, not two copied CLIs that can drift. The dependency direction already supports this: AISimulate remains Dynamo-free, while Dynamo may extend AISimulate. Please add parity tests showing that the shared engine-only arguments lower identically through both entry points.

tedzhouhk (commented):
One additional CLI package-boundary concern after removing the unified predict/sweep surface.

aisimulate/src/aisimulate/runner.py
jasonqinzhou: Adding the Sweeper impact: this is no longer only a documented runtime limitation. [parallel_enum.py](https://github.com/ai-dynamo/dynamo/blob/46edbec13267ea7d105691ed9836eca64bf8823c/aisimulate/src/aisimulate/parallel_enum.py#L132-L177) still enumerates MoE layouts with attention_dp greater than 1, and the Sweeper tests materialize a disaggregated SGLang decode worker with attention_dp=8. EngineReplayRunner rejects that topology only after candidate generation, so a supported-looking sweep can fail at execution.

Please filter or prune this topology at the search-space/capability boundary, with a test proving it is never scheduled. The full rank-aware handoff implementation can remain in #12965; a TODO alone leaves a user-visible failure mode in this PR.

aisimulate/crates/engine/src/config.rs
jasonqinzhou: Adding the Sweeper impact: honest validation is an improvement, but the current defaults make the unsupported combination reachable. [SearchSpace](https://github.com/ai-dynamo/dynamo/blob/46edbec13267ea7d105691ed9836eca64bf8823c/aisimulate/src/aisimulate/sweeper/config.py#L345-L398) defaults decode_enable_prefix_caching to false, while [deployment materialization](https://github.com/ai-dynamo/dynamo/blob/46edbec13267ea7d105691ed9836eca64bf8823c/aisimulate/src/aisimulate/sweeper/deploy.py#L18-L46) passes the selected backend and this value through unchanged. Selecting disaggregated SGLang therefore produces a candidate rejected here.

Please reject or prune this combination during search-space validation or candidate generation, and add a test that it never reaches the runner. A TODO alone still allows a sweep to fail on a candidate it generated itself.

jasonqinzhou:
**Merge readiness: approve the DEP and run the new Python/package surface in CI**

This PR makes an architectural ownership change and deliberately removes Mocker KVBM G2-G4 replay behavior. The governing [DEP #12441](https://github.com/ai-dynamo/dynamo/issues/12441) is still open with the dep:draft label and is not linked in the Related issues section. Please get the DEP accepted, or record explicit maintainer approval, and link it before merge so this breaking deletion has a clear architectural decision and migration record.

Also, on exact HEAD 46edbec13267ea7d105691ed9836eca64bf8823c, the visible checks exercise Rust, clippy, docs, and CodeQL, but do not show the Planner CPU/Python suite promised by [aisimulate/tests/conftest.py](https://github.com/ai-dynamo/dynamo/blob/46edbec13267ea7d105691ed9836eca64bf8823c/aisimulate/tests/conftest.py#L3-L8). The PR body reports 20 local runner tests, but this boundary change adds substantial runner, traffic, Sweeper, PyO3, and packaging surface. Please run or add exact-head CI coverage for those Python suites and the standalone wheel/import boundary, and attach the result before approval.

aisimulate/src/aisimulate/runner.py
dreamtalen: Addressed in 05a30ee1db. `RunnerCapabilities` now advertises whether a runner supports disaggregated attention-DP, and the full rank-aware handoff implementation remains tracked in #12965.

aisimulate/crates/engine/src/config.rs
dreamtalen: Fixed this in 3251e40db0 by implementing the enable_prefix_caching=false for SGLang: the engine skips radix-prefix matching and retention while preserving request-local KV lifecycle and chunked execution.

aisimulate/README.md
dreamtalen: Addressed in b10d20596c. AISimulate now owns the engine-only `python -m aisimulate.replay` entrypoint and the shared base parser/config/reporting schema. `python -m dynamo.replay` composes that same parser through `add_base_replay_arguments()` and `parse_base_replay_config()`, then adds only Dynamo-owned Router/Planner/AIC adapter options.

jasonqinzhou:
**Packaging/API boundary: make a single public `aisimulate-core` crate**

The `aisimulate-engine` and `aisimulate-replay` split is useful for internal implementation boundaries. For the external Rust API, however, I recommend introducing one public `aisimulate-core` Cargo package and keeping Engine and Replay as internal modules behind that facade.

This matters if we intend to publish on crates.io: a published `aisimulate-core` facade cannot depend on unpublished path-only `aisimulate-engine` and `aisimulate-replay` packages. Cargo replaces the local path with registry resolution during publication, so those dependencies would also have to be published and versioned independently. See the [Cargo dependency documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations).

My preferred target is:

```text
aisimulate-core                         public Rust crate
├── engine                              internal module
├── replay                              internal module
├── stable EngineConfig/ReplaySpec/ReplayReport facade
└── TimingModel provider boundary
    └── optional aiconfigurator-core integration
```

This gives us one public semver surface and avoids version skew across implementation crates. AIC should remain a one-way timing-provider dependency through the existing `TimingModel` boundary rather than leaking AIC types into Engine or Replay. If we intentionally keep Engine and Replay as separate Cargo packages instead of modules, then all three crates would need an explicit publication and compatible-version policy.

Could this PR establish or document that intended boundary, and mark the implementation crates `publish = false` until the public `aisimulate-core` package is introduced?

tedzhouhk (approved):

PeaBrane (approved):

dreamtalen:
@jasonqinzhou Done in 27c39e106d. We now have a single public aisimulate-core Rust crate, with Engine and Replay organized as modules.

jasonqinzhou:
Documentation follow-up: the docs should mirror the ownership boundary established by this PR.

- **AI Simulate should own the canonical documentation for the shared engine and offline Replayer**: the Generalized Mocker Engine, scheduler/KV accounting/timing, deterministic virtual time, event queue, logical workers, `ReplaySpec`, `ReplayReport`, and the base `python -m aisimulate.replay` interface.
- **Dynamo should continue to own Live Mocker and Dynamo integration documentation**: `python -m dynamo.mocker`, service discovery, transport, KV-event/metrics publication, Router/Planner composition, online replay, and deployment workflows.
- The current mixed pages should be split accordingly:
  - `dynosim-architecture.md`: neutral replay architecture under AI Simulate; Router/Planner/online composition remains under Dynamo.
  - `mocker-engine-architecture.md`: shared engine internals under AI Simulate; Live Mocker integration remains under Dynamo.
  - `dynosim-replay-cli-reference.mdx`: base AI Simulate arguments and report schema separated from the Dynamo Router/Planner/online extensions.
- `lib/mocker/README.md` should be updated because it still describes `dynamo-mocker` as owning the generic engine and replay modules.

One publishing constraint: the Fern workflow currently publishes only `docs/fern/pages/`, so moving canonical Markdown under `aisimulate/docs/` alone would make it disappear from the published site. The follow-up should keep thin Fern pages or add a one-way synchronization step rather than maintaining two hand-edited copies.

I will open a separate follow-up PR for this documentation restructuring so this already-large refactor does not absorb another broad content move.

Follow-up implementation: https://github.com/ai-dynamo/dynamo/pull/13051

jasonqinzhou (approved):

dreamtalen:
Will rebase on main once https://github.com/ai-dynamo/dynamo/pull/13052 is merged, as it's a straightforward fix

saturley-hall (changes requested):
This breaks the `ai-dynamo` wheel.

`python -m dynamo.replay` imports `aisimulate` at module, but neither `ai-dynamo` nor `ai-dynamo-runtime` declares that dependency, and we are intentionally not publishing it for the moment.

This will only work in the planner container.

dreamtalen:
@saturley-hall Thanks! As we synced offline, I added the placeholder dependency in 427aa6e927: ai-dynamo now declares aisimulate==0.1.0, and nightly scripts update the AISimulate package version. Please feel free to replace this with the actual dev0 artifact.

Follow-up from Linux nightly validation: ai-dynamo supports Python 3.13, while AISimulate’s current AIC dependency stack supports Python 3.10–3.12. I addressed this in d31968b by making the AISimulate dependency conditional on `Python <3.13` and adding a clear early error when dynamo.replay is invoked on Python 3.13.

nv-anants:
/ok to test [5db74a1](https://github.com/ai-dynamo/dynamo/pull/12525/commits/5db74a158a1daf7c7a05824e6af4f0053fe33d4c)

saturley-hall (approved):
With the passing PR workflow I think this satisfies the minimal ops requirements.

Note we will need to file a follow-up PR to enable `aisimulate` to go out with the nightly dynamo build and update the dependency to that nightly version so that the latest work is verifiable.

leigao97:
Hi @dreamtalen,

I've been using the kvbm offloading feature in DynoSim. Since this feature was removed in this refactor, would it be added back later?

Thanks.

dreamtalen:
Hi @leigao97, we plan to restore kv offload simulation in the AISimulate repo https://github.com/ai-dynamo/aisimulate, but not through KVBM, since KVBM is being sunset from Dynamo.

Could you share which specific KVBM offloading features or workflows you have been using? That would help us prioritize the replacement. Thanks!

leigao97:
Thank you, @dreamtalen, for the quick reply. I was primarily using the KVBM offloading feature to simulate three-tiered (HBM/DRAM/SSD) prefix cache offloading with configurable capacity and bandwidth, similar to the HiCache design in SGLang. Having an accurate simulation of these multi-tiered offloading behaviors in AISimulate would be incredibly valuable for my workflows.

PR #12527, opened 2026-08-01
perf(kv-router): remove router hint ownership scan [DYN-3849]

PeaBrane:
## Summary

- Remove the single-threaded radix indexer's per-request scan of every router-hint candidate owner's external-hash table. The scan treated an engine hash invariant violation as a normal routing state and added `O(candidate owners × matched prefix)` hash probes to the lookup hot path.
- Restore router-hint chain retention for the concurrent indexer, which does not need that validation either.
- Document that, within an indexer/hash domain, engine-published local block chains and external sequence-hash chains have a deterministic one-to-one mapping. Producer, configuration, or protocol violations must not be revalidated on request lookup or scheduling paths.

This is a narrow stacked change against #11695. It does not add a replacement runtime validation path.

## Validation

- `cargo test -p dynamo-kv-router find_match_details_can_retain_router_hint_root_candidates`
- `cargo test -p dynamo-llm --no-default-features router_hint_chain_retention_requires_event_driven_primary`
- Commit hooks, including codespell and repository hygiene checks

### DIFF
--- a/lib/kv-router/CLAUDE.md
+++ b/lib/kv-router/CLAUDE.md
@@ -18,3 +18,13 @@ when one exists.
   paths.
 - Do not use `FxHashMap` / `FxHashSet` for text keys or externally controlled
   values such as `request_id`; use the standard hash collections there.
+
+## Engine Hash Invariant
+
+- Within one indexer/hash domain, engine-published local block hashes and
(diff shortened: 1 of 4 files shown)


oandreeva-nv (approved):
lgtm

PR #12529, opened 2026-08-01
docs: restructure the compatibility page and correct platform data

dagil-nvidia:
## Summary

Two things the current support matrix cannot express, and one it states wrongly.

**`status` conflated two different facts.** Ubuntu 24.04 and Ubuntu 22.04 both read "Supported", but only 24.04 is a shipped container base (`cuda-dl-base 25.11-cuda13.x-devel-ubuntu24.04`, per `container/context.yaml`). 22.04 works because the wheels are `manylinux_2_28` — there is no container for it. A reader choosing a host OS cannot tell those apart. Replaced `status` with `scope`: "Containers and wheels" vs "Wheels only".

**The wheel-validation claim was inaccurate.** It said wheels are "validated on CentOS Stream 9 and Ubuntu 22.04/24.04". They are built in a `manylinux_2_28` environment (AlmaLinux 8, glibc 2.28+) and validated on Ubuntu 22.04 and 24.04 only. Corrected, and the CentOS Stream 9 row dropped rather than left claiming a validation that does not happen.

**67 near-identical rows behind one accordion.** The generated table repeated Dynamo/Backend/CUDA/Driver/Note per release, with `Note` empty on almost every row. The signal a reader wants — CUDA 12 vs 13 — was a number buried mid-row. `ReleaseSupportMatrix` makes the CUDA major the visual anchor and groups by minor line.

## What changed

- `ReleaseSupportMatrix.tsx` — new component, rows derived from `CUDA_HISTORY` and `RELEASES` in `releases.data.ts`; nothing transcribed. Native `<details>`/`<summary>` for disclosure, so no client JS.
- `releases.data.ts` — `status` → `scope`, CentOS row removed, wheel note corrected.
- `gen_llms_tables.py` — the human-facing accordion generator is removed; the `<llms-only>` twin still emits the same rows as a flat markdown table, so the agent-facing export loses nothing.
- `FeatureInteractions.tsx` — new component rendering the pairwise matrices in the `FeatureHeatmap` visual vocabulary, from `FEATURE_INTERACTIONS` in `releases.data.ts`.
- `releases.data.ts` — gains `FEATURE_INTERACTIONS`: the matrices used to live only in hand-authored JSX, so no pairwise status reached Fern's agent exports. `gen_llms_tables.py` now emits them into the `<llms-only>` twin from the same source.
- `InteractionStatus.tsx` and its `.dynref-interaction*` rules in `ReferenceStyles.tsx` are removed, superseded by the above.

## Verification

- `gen_llms_tables.py --check`: all outputs `unchanged`.
- `fern check`: 0 errors, 1292 warnings (matches the pre-existing baseline).
- `pre-commit`: all applicable hooks pass.
- `fern docs broken-links`: all checks passed.
- The 165 authored interaction cells were migrated mechanically and diffed against the original JSX: 18 differences, all of them the intended LoRA fix.

**One fix worth calling out.** The component originally used `<style>{RSM_CSS}</style>`. A text child is escaped on render, so the `>` child combinator in its CSS would have become `&gt;` and been dropped silently in production — the failure that took the Home page down in #12402. Switched to `dangerouslySetInnerHTML`, matching `RecipeStyles.tsx`, with the reason recorded in the header. `ReferenceStyles` keeps the text-child form because it still has zero child combinators.

## Notes

Rebased across #12373, which renamed every docs path. Five conflict hunks, all resolved toward the new paths.

## Update: absorbed #12241

#12241 was a second PR editing `compatibility.mdx` and `gen_llms_tables.py`,
and the two conflicted in both files. Landing them in sequence would have
meant resolving that conflict blind on the second rebase, so it was folded
here instead and #12241 is closed.

Carried over: the driver-floor table, which inverts the support matrix so a
reader can start from a driver they already have. `<ReleaseSupportMatrix />`
renders only the forward view, so this stays a generated span, now placed
beside the component it inverts. Also the SGLang multimodal and router prose
corrections, the v1.3.0 release-note and release-history fixes, the
`UpgradePanel` change, `test_gen_llms_tables.py` with its `pytest.ini` and
pre-commit hook, and the CUDA compatibility link now resolving without a
redirect hop.

Dropped as superseded: the generated `support-matrix` accordion and its
renderer, replaced by `<ReleaseSupportMatrix />`, and the raw
feature-interaction markdown tables, replaced by `<FeatureInteractions />`.
Their per-cell notes live in `releases.data.ts`.

**The fold exposed a factual error.** `FEATURE_INTERACTIONS` marked the SGLang
Multimodal x KV-Aware Routing cell `"no"` ("This feature combination is not
supported"), while the merged feature table on the same page described it as
supported with a fallback. The page contradicted itself.
`multimodal-kv-routing.md` is the authority and says supported: hash
forwarding is upstream in SGLang 0.5.13+ and Dynamo pins 0.5.16. The cell is
now `"yes"`, noting that a custom build without the patch degrades to
text-prefix routing.

Validation: `gen_llms_tables.py --check` clean; `check_agent_twins.py` clean on
all three governed pages with its self-test at 14/14; `test_gen_llms_tables.py`
14 passed; ruff clean. Confirmed the corrected note reaches the agent twin.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12529" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added a release support matrix with CUDA, driver, toolkit, backend, and related compatibility details.
  - Added release statistics for versions v0.6.0–v0.9.0 and expanded release history information.
  - Added interactive, keyboard-accessible tooltips for feature and interaction support details.
- **Documentation**
  - Clarified Ubuntu container and wheel support, manylinux requirements, and validated environments.
  - Updated compatibility guidance and related links.
  - Refined TensorRT-LLM known-issue documentation and cancellation guidance.
- **Bug Fixes**
  - Removed outdated CentOS Stream and AWS/Amazon Linux platform references.


docs/fern/pages/reference/general/compatibility.mdx
```diff
@@ -224,72 +103,72 @@ TensorRT-LLM delivers maximum inference performance and optimization, with full
...
+| **Tool Calling** | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="na" /> | |
+| **Speculative Decoding** | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="na" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="na" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="na" /> | <Tooltip tip={<>Eagle3 support is documented. <a href="../features/speculative-decoding/speculative-decoding-vllm.md">Source</a></>}><InteractionStatus status="yes" noteLabel="Eagle3 support" /></Tooltip> | <InteractionStatus status="na" /> |
```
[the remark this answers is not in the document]
dagil-nvidia: Real on every count, and fixed in `4fa08e8b33f`. I resolved all ten targets against the tree and every one was missing.

The markdown links took corrected relative paths:

| was | now |
| --- | --- |
| `../getting-started/local-installation.mdx` | `../../cli/installation/install-dynamo.mdx` |
| `../backends/trtllm/trtllm-known-issues.md` | `../../developer-guide/knowledge-base/modular-components/backends/tensorrt-llm/known-issues.md` |
| `releases-data.mdx` | `releases-machine-readable.mdx` |

Your closing point about the raw `<a href>` elements was the important one, and it turned out to be the reason this shipped in the first place. Those anchors sit inside a JSX attribute expression (`tip={<>…</>}`), not in markdown, so Fern's relative-link rewriting never touches them — and `fern docs broken-links` does not report them either. That is why `check_reference.sh` gate 5 was green with ten dead links on the page. It *does* validate site-absolute hrefs, so the eight tooltip anchors are now absolute, matching the pattern `ReleaseHeader.tsx`, `CommunityLanding.tsx`, and `WelcomeHero.tsx` already use.

Worth recording for whoever writes the next one: Fern's slug is a kebab of the **nav title**, and it splits on case. `SGLang` → `sg-lang`, `vLLM` → `v-llm`, `TensorRT-LLM` → `tensor-rt-llm`. My first pass used `sglang`/`vllm`/`tensorrt-llm` and gate 5 rejected all five. A comment on the page now records that.

Gate 5 after the fix: `total site errors: 0 … in reference/: 0`.

docs/fern/pages/reference/general/compatibility.mdx
```diff
@@ -8,142 +8,21 @@ subtitle: Hardware, platform, and feature support for Dynamo backends
...
 
 <Note>
-See [Release Artifacts](release-artifacts.mdx) for the full artifact inventory — container images, wheels, Helm charts, and crates — and [Model Early Access Builds](model-early-access-builds.mdx) for per-model early access container builds.
+See [Release Artifacts](release-artifacts.mdx) for the full artifact inventory — container images, wheels, Helm charts, and crates — [Local Installation](../getting-started/local-installation.mdx) for host OS and architecture requirements, and [Model Early Access Builds](model-early-access-builds.mdx) for per-model early access container builds. For backend-specific runtime workarounds — including the local `docker run --network host` hang first reported on Amazon Linux 2023 — see [TensorRT-LLM Known Issues](../backends/trtllm/trtllm-known-issues.md).
```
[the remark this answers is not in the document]
dagil-nvidia: Same finding as the Devin thread above and correct — fixed in `4fa08e8b33f`. Both intro links, the eight `<Tooltip>` anchors, and the `releases-data.mdx` link near the release matrix were dead; I verified each target against the tree before changing it.

One correction to the diagnosis: the tooltip anchors were not failing because of a wrong path *style*. They are raw `<a>` elements inside a JSX attribute expression, which Fern neither rewrites nor link-checks, so no relative path could have worked there. They are site-absolute now. `fern docs broken-links` does check absolute hrefs, which is what caught my own first attempt at the slugs — Fern kebabs the nav title and splits on case, so it is `backends/sg-lang/…`, `backends/v-llm/…`, `backends/tensor-rt-llm/…`.

`check_reference.sh` gate 5 now reports 0 errors in `pages/reference/`; `fern check` is 0 errors / 1292 warnings, unchanged from baseline.

docs/fern/components/FeatureHeatmap.tsx
```diff
@@ -148,21 +155,91 @@ const HEAT_CSS = `
...
+}
+
+.dynref-heat-cell[data-backend="SGLang"] .dynref-heat-tooltip {
+    left: 0;
+    transform: translate(0, 3px);
+}
+
+.dynref-heat-cell[data-backend="vLLM"] .dynref-heat-tooltip {
+    right: 0;
+    left: auto;
```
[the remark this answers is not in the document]
dagil-nvidia: Good hypothesis, and you were right to say it should be confirmed empirically rather than assumed. It does not hold — the repo already contains the experiment.

Two shipped components inject their CSS as a `<style>` text child *and* depend on double-quoted attribute selectors:

- `ArtifactBrowser.tsx` — 14 of them, including `[for="ab-all"]`, `[for="ab-container"]`, `[for="ab-wheel"]`
- `PinnedEnvironment.tsx` — 12, including `[for="pe-sglang"]`, `[for="pe-trtllm"]`, `[for="pe-vllm"]`

Both render on `pages/reference/general/release-artifacts.mdx`, and those `[for=…]` rules are what drive the radio-backed selector UI. If `"` were escaped to `&quot;` on render, both selectors would be visibly broken in production today. `RunsWhereWizard.tsx`, `TagLookup.tsx`, and `UpgradeSelector.tsx` are in the same position. `ReferenceStyles.tsx`'s `content: "⧉"` is the copy glyph on every reference page and renders.

So the #12402 failure was specific to `>`; quotes survive. That also matches the shape of the mitigation: `ReleaseSupportMatrix.tsx` switched to `dangerouslySetInnerHTML` because its CSS uses `>` child combinators, and `ReferenceStyles.tsx` kept the text-child form because it has none. I checked `HEAT_CSS` for the same hazard — 218 lines, zero `>` characters — so `FeatureHeatmap` is safe where it is, and `content: ""` is fine for the same reason as `content: "⧉"`.

Leaving both files as they are. Normalising them to one form would lose the distinction that is doing the work.

docs/fern/components/ReleaseSupportMatrix.tsx
```diff
@@ -0,0 +1,573 @@
...
+    <div className="dynref-rsm-backend-row">
+      <span className="dynref-rsm-backend-name">{line.backend}</span>
+      <span className="dynref-rsm-backend-cell">
+        <span className="dynref-rsm-cell-label">Engine</span>
+        <span className="dynref-rsm-cell-value dynref-mono">{line.engine ?? "—"}</span>
+      </span>
+      <span className="dynref-rsm-backend-cell">
+        <span className="dynref-rsm-cell-label">Toolkit</span>
+        <span className="dynref-rsm-cell-value dynref-mono">{line.toolkit}</span>
+      </span>
```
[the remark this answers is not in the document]
dagil-nvidia: Real, and fixed in `5b1e0506982`. Confirmed the exact pairing you named: `CUDA_HISTORY` gives v0.7.1 SGLang/12.8 → `570.xx+` and vLLM/12.9 → `575.xx+`, with v0.7.0 reversing it, and `driverSummary()` collapsed the block to `570.xx+ or 575.xx+` while `BackendRow` rendered only Engine / Toolkit / NIXL.

Took the first of your two options — `BackendLine` already carried `minDriver`, it just was not rendered. Each backend row now has a `Driver` cell (grid widened from four columns to five; the mobile `1fr 1fr` fallback is unchanged), and the block header keeps the summary so the CUDA chip still reads at a glance. Splitting blocks by toolkit/driver would have fragmented the CUDA-major grouping that is the point of the layout.

docs/fern/components/ReleaseSupportMatrix.tsx
```diff
@@ -0,0 +1,573 @@
...
+  );
+}
+
+function MinorLineSummary({ group }: { group: MinorGroup }) {
+  const releaseCount = group.releases.length;
+  return (
+    <summary className="dynref-rsm-line-summary">
+      <span className="dynref-rsm-line-title">{group.minor}</span>
+      <span className="dynref-rsm-line-meta">
+        {releaseCount} release{releaseCount === 1 ? "" : "s"}
```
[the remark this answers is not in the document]
dagil-nvidia: Real, and fixed in `5b1e0506982`. Your numbers check out exactly — counted against `releases.data.ts`:

```
0.8.x  RELEASES=5  cards=2
0.7.x  RELEASES=3  cards=2
1.3.x  RELEASES=1  cards=1   (all other lines already matched)
```

Rather than relabel the count, I made the page show what it was counting. `releaseCount` now comes from `RELEASES` (stable + patch in the minor line), and each release card lists the post trains that resolve to it — "Post trains v0.8.1.post1, v0.8.1.post2, v0.8.1.post3 share this CUDA support". Post trains carry no `CUDA_HISTORY` row on purpose (`CUDA_NOTES`: they have the same CUDA support as their base), so they belong on the base card rather than getting one of their own; that keeps the count honest and the reader still sees all five.

Sister PR #12241 hits the same inheritance gap in `gen_llms_tables.py`'s driver-floor table and fixes it the same way.

docs/fern/components/InteractionStatus.tsx
```diff
@@ -0,0 +1,53 @@
...
+export function InteractionStatus({ status, noteLabel }: InteractionStatusProps) {
+  const label = noteLabel ? `${STATUS_LABEL[status]} — ${noteLabel}` : STATUS_LABEL[status];
+
+  return (
+    <span
+      className={`dynref-interaction dynref-interaction--${status}`}
+      role="img"
+      aria-label={label}
+      tabIndex={noteLabel ? 0 : undefined}
+    >
```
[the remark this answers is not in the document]
dagil-nvidia: Partly addressed in `5b1e0506982`, with the part I could verify separated from the part I could not.

What I could verify: `ReferenceStyles.tsx` had no `:focus-visible` rule for `.dynref-interaction` at all. So regardless of how Fern's `<Tooltip>` renders its trigger, the `tabIndex={0}` stop was invisible — focus landed on a cell with no ring and no state change. Added the same ring the matrix summaries use:

```css
.dynref-interaction:focus-visible {
    outline: 2px solid var(--nv-color-green, #76B900);
    outline-offset: 2px;
    border-radius: 3px;
}
```

What I could not verify: whether Fern's `<Tooltip>` renders its own focusable trigger, which would make two tab stops per noted cell. That needs the built page, and both stops land on the same cell either way, so I left `tabIndex` in place — dropping it would lose keyboard reach to the tip under any implementation that uses `asChild`. Flagging it for Dan rather than guessing at a third-party component's markup.

`aria-label` is unchanged, so the screen-reader path you noted is unaffected.

docs/fern/components/releases.data.ts
```diff
@@ -954,17 +954,24 @@ export const PLATFORM_PREVIEW_COVERAGE: Record<string, Coverage> = {
...
+ *     are manylinux_2_28 (glibc 2.28+), so `pip install ai-dynamo` runs on it.
+ *     Getting Started's Local Installation lists Ubuntu 22.04 for that reason.
+ * A future host that carries neither status would take a third scope value. */
 export const PLATFORM = {
   gpus: ["Blackwell", "Hopper", "Ada Lovelace", "Ampere"],
   os: [
-    { name: "Ubuntu", version: "24.04", arch: "x86_64, ARM64", status: "Supported", chip: "ubuntu" },
-    { name: "Ubuntu", version: "22.04", arch: "x86_64", status: "Supported", chip: "ubuntu" },
-    { name: "CentOS Stream", version: "9", arch: "x86_64", status: "Experimental", chip: "centos" },
+    { name: "Ubuntu", version: "24.04", arch: "x86_64, ARM64", scope: "Containers and wheels", chip: "ubuntu" },
```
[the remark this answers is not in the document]
dagil-nvidia: Not taking this one, and the suggested fix would make the data worse.

`status: "Supported"` cannot be kept alongside `scope` without asserting something that is no longer true. The whole point of the rename is that the two OS rows are not the same kind of supported: Ubuntu 24.04 is the shipped container base (`container/context.yaml` — every CUDA runtime image is `cuda-dl-base …-ubuntu24.04`), while Ubuntu 22.04 is wheels-only. Re-adding a flat `"Supported"` to both would restore exactly the ambiguity the field split removes, and a consumer reading the legacy key would get the pre-correction answer.

On the schema being stable: `gen_llms_tables.py` does describe `releases.json` as a stable-schema serialization, but there is no `schemaVersion` field, no documented consumer, and `grep -rn "releases.json" .github/ scripts/ container/ benchmarks/` finds nothing that reads it. It is a generated mirror of `releases.data.ts`, regenerated wholesale on every release bump.

The larger schema change in this PR is the removal of the `platform.csp` key, not the `status` → `scope` rename. That one is a product-support question rather than a mechanical one, so I have left it open for Dan on the Devin thread about Amazon Linux 2023.

docs/fern/components/releases.data.ts
```diff
@@ -954,17 +954,24 @@ export const PLATFORM_PREVIEW_COVERAGE: Record<string, Coverage> = {
...
+    { name: "Ubuntu", version: "22.04", arch: "x86_64", scope: "Wheels only", chip: "ubuntu" },
   ],
   arch: ["x86_64", "ARM64 (Ubuntu 24.04 only)"],
   wheelsNote:
-    "Wheels are built in a manylinux_2_28-compatible environment and validated on CentOS Stream 9 and Ubuntu 22.04/24.04. Other Linux distributions are expected to work but are not officially verified.",
-  csp: [{ provider: "AWS", os: "Amazon Linux 2023", arch: "x86_64", status: "Supported" }],
+    "Wheels are built in a manylinux_2_28 environment (AlmaLinux 8, glibc 2.28+) and validated on Ubuntu 22.04 and 24.04. They install on any Linux distribution with glibc 2.28+ (Debian 11+, RHEL 9, etc.), but only Ubuntu 22.04/24.04 are officially verified.",
```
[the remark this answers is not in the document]
dagil-nvidia: Confirmed, and leaving this open for @dagil-nvidia to call.

The facts: `PLATFORM.csp` is gone from `releases.data.ts`, so `platform.csp` is gone from `docs/fern/assets/releases.json` and the CSP bullet is gone from `platform_lines()` in `gen_llms_tables.py`. AL2023 now appears nowhere as a supported platform on the human page or in the machine-readable exports — its only remaining mention is the `docker run --network host` workaround, which the intro `<Note>` links to.

There is a reading under which the de-listing is right: the new `scope` model asks whether an OS is a shipped container base or a wheels-only host, and AL2023 is neither verified nor a container base. The new `wheelsNote` already covers it generically — wheels install on any glibc 2.28+ distribution, only Ubuntu 22.04/24.04 are officially verified.

But that is a change to what Dynamo publicly claims to support on AWS, and the PR description only justifies dropping the CentOS Stream 9 row and rewriting the wheels note. Dan: is the de-listing deliberate? If so the PR description should say why; if not, AL2023 wants an OS row with `scope: "Wheels only"`. Not resolving this one.

The related `status` → `scope` schema question is answered on the dynamo-review-agent thread.

docs/fern/pages/reference/general/compatibility.mdx
```diff
@@ -224,72 +103,72 @@ TensorRT-LLM delivers maximum inference performance and optimization, with full
...
-
-> **Notes:**
-> 1. **Multimodal + KV-Aware Routing**: Image-aware KV routing is supported in the documented vLLM paths. The default Rust frontend path supports model families handled by `llm-multimodal`; the Python chat-processor path delegates to vLLM's multimodal processor. ([Source][mm-kv-routing])
-> 2. **KV-Aware LoRA Routing**: vLLM supports routing requests based on LoRA adapter affinity.
-> 3. **Audio Support**: vLLM supports audio models like Qwen2-Audio (experimental). ([Source][mm-vllm])
-> 4. **Video Support**: vLLM supports video input with frame sampling. ([Source][mm-vllm])
-> 5. **Speculative Decoding**: Eagle3 support documented. ([Source][vllm-spec])
-
-#### SGLang Feature Interactions
+| **Disaggregated Serving** | <InteractionStatus status="na" /> | | | | | | | | | |
```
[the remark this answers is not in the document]
dagil-nvidia: Verified, and you are right — but the fix needs a call from @dagil-nvidia, so leaving this open.

What I checked, per the rule that a claimed export loss gets verified against the generator first:

- The **release support matrix** is *not* lost. `render_compatibility()` still emits `cuda_table(data)` under "CUDA toolkit and minimum driver per Dynamo release", so the accordion → `<ReleaseSupportMatrix />` swap keeps its agent twin. That part of the restructure is sound.
- The **interaction matrices** are a different story, and your reading is correct. They were literal `✅`/`🚧` glyphs in plain markdown before this PR, so they survived any export by construction. `gen_llms_tables.py` has no source for pairwise interaction data — it lives in the MDX, not in `releases.data.ts` — so the `llms-tables` twin carries the aggregate feature table only, and nothing reproduces the 3 × 11 × 11 grid. Under the project's own stated assumption ("component output may be dropped from the agent-facing exports", the reason `<llms-only>` exists at all), agents can now get rows with blank status cells.

Why I did not just fix it: both of your options — a textual fallback per cell, or an `<llms-only>` twin of the three matrices — add roughly 40 lines to `compatibility.mdx`, which is the file #12241 and this PR already conflict on. Growing it before that merge is resolved trades a docs problem for a merge problem. A third option is to restore the footnote list under each table, which would also give keyboard and no-hover readers the note text that is currently tooltip-only.

Dan: which of the three, and land it here or as a follow-up after #12241 → #12529 merges? Leaving unresolved.

docs/fern/components/releases.data.ts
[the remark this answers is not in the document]
dagil-nvidia: Resolved — AL2023 was restored in `37867ac` ("restore the AWS platform row"). Verified on the current head: `PLATFORM.csp` is back in `releases.data.ts`, AL2023 appears in `releases.json` and twice on `compatibility.mdx`, and `gen_llms_tables.py` carries 4 `csp` references, so the machine-readable exports have it too.

harryskim (approved):
approved

.pre-commit-config.yaml
```diff
@@ -105,12 +105,36 @@ repos:
...
+    - id: check-agent-twins
+      name: Check component-rendered data has an agent-readable twin
+      entry: python3 docs/fern/scripts/check_agent_twins.py
+      language: system
+      # releases.data.ts is the reason this runs at all: it is where a new
+      # release appears, and the twin then has to carry it. It is .ts, not
+      # .tsx, so a components/*.tsx pattern misses it entirely. Component
+      # files matter for the undeclared-component tripwire, and only the
+      # governed page matters among the pages.
+      files: ^docs/fern/(components/releases\.data\.ts|components/.*\.tsx|pages/reference/general/compatibility\.mdx|scripts/check_agent_twins\.py)$
```
dmitry-tokarev-nv: **The hook is scoped to 1 of the 3 pages it governs.** `files:` matches only `compatibility.mdx`, but `IN_SCOPE` in `check_agent_twins.py` (L61-67) governs three pages: `compatibility.mdx`, `release-artifacts.mdx`, `model-early-access-builds.mdx`.

Two consequences:

1. A commit that deletes the `<llms-only>` twin from `release-artifacts.mdx` or `model-early-access-builds.mdx` runs no hook at all — which is precisely the regression this gate exists to catch.
2. `main()` hard-fails (`::error::in-scope page not found`, `return 1`) if one of those pages is renamed or moved. Since the renaming commit doesn't match `files:` either, that failure surfaces on the next unrelated PR that happens to touch any `components/*.tsx`.

```suggestion
      # releases.data.ts is the reason this runs at all: it is where a new
      # release appears, and the twin then has to carry it. It is .ts, not
      # .tsx, so a components/*.tsx pattern misses it entirely. Component
      # files matter for the undeclared-component tripwire, and the page
      # list here has to stay in step with IN_SCOPE in check_agent_twins.py.
      files: ^docs/fern/(components/releases\.data\.ts|components/.*\.tsx|pages/reference/general/(compatibility|release-artifacts|model-early-access-builds)\.mdx|scripts/check_agent_twins\.py)$
```

.pre-commit-config.yaml
```diff
@@ -105,12 +105,36 @@ repos:
...
       files: ^docs/fern/scripts/check_published_styles\.py$
       pass_filenames: false
+    - id: gen-llms-tables
+      name: Test the generated docs tables (gen_llms_tables.py)
+      entry: pytest -c docs/fern/scripts/pytest.ini -p no:cacheprovider docs/fern/scripts/test_gen_llms_tables.py
+      language: python
+      files: ^docs/fern/(scripts/(gen_llms_tables|test_gen_llms_tables)\.py|scripts/pytest\.ini|components/releases\.data\.ts)$
+      pass_filenames: false
+      additional_dependencies:
+        - pytest
```
dmitry-tokarev-nv, line 153: **This hook never checks staleness.** It triggers on `components/releases.data.ts` but only runs `test_gen_llms_tables.py`; it never invokes `gen_llms_tables.py --check`. I grepped `.github/` and the rest of this config — the only other reference to the script is `auto-dep-upgrade-trigger.yml`, which *regenerates* rather than checks.

So editing `releases.data.ts` and forgetting to regenerate leaves `docs/fern/assets/releases.json` and every generated span stale, with no gate anywhere in the repo. The hook's name implies a check it doesn't perform.

`main()` already returns 1 with `check failed: N output(s) stale` (L1344-1349), so a second `language: system` hook covers it:

```suggestion
    - id: gen-llms-tables
      name: Test the generated docs tables (gen_llms_tables.py)
      entry: pytest -c docs/fern/scripts/pytest.ini -p no:cacheprovider docs/fern/scripts/test_gen_llms_tables.py
      language: python
      files: ^docs/fern/(scripts/(gen_llms_tables|test_gen_llms_tables)\.py|scripts/pytest\.ini|components/releases\.data\.ts)$
      pass_filenames: false
      additional_dependencies:
        - pytest
    - id: gen-llms-tables-check
      name: Check the generated docs tables are up to date
      entry: python3 docs/fern/scripts/gen_llms_tables.py --check
      language: system
      files: ^docs/fern/(scripts/gen_llms_tables\.py|components/releases\.data\.ts|assets/releases\.json|pages/reference/general/.*\.mdx)$
      pass_filenames: false
```

.pre-commit-config.yaml
```diff
@@ -105,12 +105,36 @@ repos:
       entry: python3 docs/fern/scripts/check_style_components.py
       language: system
       files: ^docs/fern/(components/.*Styles\.tsx|scripts/check_style_components\.py)$
```
dmitry-tokarev-nv: **The two new components are outside this hook's reach.** `files:` matches `components/.*Styles\.tsx` only, and `check_style_components.py` defaults to `glob("*Styles.tsx")` — but this PR adds two large CSS template literals outside that naming convention: `RSM_CSS` (`ReleaseSupportMatrix.tsx:49`) and `FI_CSS` (`FeatureInteractions.tsx:26`).

A stray backtick or `${` in either — the exact defect this hook was written for — breaks the Fern build with nothing catching it.

The script already accepts explicit paths (L70) and returns `[]` for files with no template literal (L49-50), so widening the pattern is sufficient; I ran it against both new components and it passes cleanly.

```suggestion
      files: ^docs/fern/(components/.*\.tsx|scripts/check_style_components\.py)$
```

Worth widening the argless default at `check_style_components.py:71` to match, so a manual full run covers the same set.

.pre-commit-config.yaml
```diff
@@ -105,12 +105,36 @@ repos:
...
+      # .tsx, so a components/*.tsx pattern misses it entirely. Component
+      # files matter for the undeclared-component tripwire, and only the
+      # governed page matters among the pages.
+      files: ^docs/fern/(components/releases\.data\.ts|components/.*\.tsx|pages/reference/general/compatibility\.mdx|scripts/check_agent_twins\.py)$
+      pass_filenames: false
+    - id: check-agent-twins-selftest
+      name: Self-test the agent-twin matcher
+      entry: python3 docs/fern/scripts/check_agent_twins.py --test
+      language: system
+      files: ^docs/fern/scripts/check_agent_twins\.py$
```
dmitry-tokarev-nv, line 138: Nit: this hook omits `pass_filenames: false`, unlike its sibling `check-published-styles-selftest` immediately below. Pre-commit therefore appends the script path after `--test`.

Harmless today only because `main()` short-circuits on `"--test" in sys.argv` before parsing args — it's one refactor away from the self-test silently running against a file argument.

```suggestion
    - id: check-agent-twins-selftest
      name: Self-test the agent-twin matcher
      entry: python3 docs/fern/scripts/check_agent_twins.py --test
      language: system
      files: ^docs/fern/scripts/check_agent_twins\.py$
      pass_filenames: false
```

docs/fern/scripts/check_agent_twins.py
```diff
@@ -0,0 +1,435 @@
...
+            continue
+        # Word-boundary, so "1.3.0rc19" does not satisfy "1.3.0". A twin of
+        # release candidates would otherwise score full marks while carrying
+        # no released row at all.
+        missing = sorted(
+            i
+            for i in spec
+            if not re.search(rf"(?<![\w.]){re.escape(i)}(?![\w.])", blob, re.I)
+        )
+        covered = 1 - len(missing) / len(spec)
```
dmitry-tokarev-nv, line 310: **The release-coverage search is unscoped, so unrelated version strings satisfy it.** This is the same hazard you deliberately guarded against for `FeatureInteractions` via `_interactions_segment` (L156-164) — the release check needs the same treatment.

The backend-pins table lives in the same twin and lists bare NIXL/UCX/engine versions. NIXL `1.3.2` is already present there as a bare string, so when Dynamo ships v1.3.2 and someone adds the `CUDA_HISTORY` rows without regenerating the twin, `ReleaseSupportMatrix` scores 100% and passes at its `1.0` floor.

Measured on the current page: deleting the entire CUDA table drops coverage to **40%**, not 0% — 6 of 15 release strings are already satisfied by text elsewhere in the twin.

A `_releases_segment` alongside `_interactions_segment`, bounded by the next `**...**` heading (`gen_llms_tables.py:643` emits `**CUDA toolkit and minimum driver per Dynamo release**`, followed by `**Feature support by backend ...**` at L649):

```python
RELEASES_HEADING = r"\*\*CUDA toolkit and minimum driver per Dynamo release\*\*"


def _releases_segment(blob: str) -> str:
    """The release-table section of the twin, or empty if absent.

    Scoped for the same reason the interactions section is: the backend-pins
    table in this twin lists bare NIXL/UCX/engine versions, so an unscoped
    search lets an unrelated "1.3.2" satisfy the release row for 1.3.2.
    """
    m = re.search(RELEASES_HEADING, blob)
    if not m:
        return ""
    rest = blob[m.end() :]
    nxt = re.search(r"\n\*\*", rest)
    return rest[: nxt.start()] if nxt else rest
```

and the call site:

```suggestion
        if not spec:
            continue
        # Word-boundary, so "1.3.0rc19" does not satisfy "1.3.0". A twin of
        # release candidates would otherwise score full marks while carrying
        # no released row at all.
        #
        # Scoped to the release section for the same reason the interactions
        # rows are: version strings from the backend-pins table in this same
        # twin would otherwise satisfy release rows that were never written.
        hay = _releases_segment(blob) if component == "ReleaseSupportMatrix" else blob
        missing = sorted(
            i
            for i in spec
            if not re.search(rf"(?<![\w.]){re.escape(i)}(?![\w.])", hay, re.I)
        )
        covered = 1 - len(missing) / len(spec)
```

If the section is missing entirely this now fails loudly rather than scoring on stray matches, which reads like the intent.

docs/fern/scripts/check_agent_twins.py
```diff
@@ -0,0 +1,435 @@
...
+
+    INTERACTION_FEATURES also carries status literals ("yes", "wip"); those
+    start lowercase and are not row labels, so they are dropped.
+    """
+    try:
+        blk = source[source.index("export const INTERACTION_FEATURES") :]
+    except ValueError:
+        return set()
+    blk = blk.split("\nexport const ", 1)[0]
+    return {n for n in re.findall(r'"([^"]+)"', blk) if n[:1].isupper()}
```
dmitry-tokarev-nv: The slice at L152 splits on the next `\nexport const `, but `releases.data.ts` has two intervening `export` declarations — `export type InteractionState` (L997) and `export interface BackendInteractions` (L1006) — before the next `export const FEATURE_INTERACTIONS` (L1012). The block therefore swallows both.

The uppercase filter then admits `"SGLang"` and `"TensorRT-LLM"` from `backend: "SGLang" | "TensorRT-LLM" | "vLLM"`, but not `"vLLM"` — so it's inconsistent as well. Verified: returns 12 names, 2 of which are backends rather than features.

Harmless today because those strings appear in the section anyway, but the "missing N of M row labels" message reports the wrong denominator, and the docstring's "row labels" claim isn't what the function returns. Bounding on the array literal instead:

```suggestion
def _interaction_features(source: str) -> set[str]:
    """Row labels in the pairwise matrices.

    INTERACTION_FEATURES also carries status literals ("yes", "wip"); those
    start lowercase and are not row labels, so they are dropped.
    """
    try:
        blk = source[source.index("export const INTERACTION_FEATURES") :]
    except ValueError:
        return set()
    # Bounded by the array literal, not by the next `export const`: two type
    # declarations sit in between, and BackendInteractions.backend would
    # otherwise contribute "SGLang" and "TensorRT-LLM" as row labels.
    blk = blk.split("]", 1)[0]
    return {n for n in re.findall(r'"([^"]+)"', blk) if n[:1].isupper()}
```

docs/fern/scripts/check_agent_twins.py
```diff
@@ -0,0 +1,435 @@
...
+    ]
+    tmp = Path("/tmp/_agent_twin_case.mdx")
+    passed = 0
+    for name, body, expect_ok in cases:
+        tmp.write_text(body, encoding="utf-8")
+        if (not check(tmp, exp)) == expect_ok:
+            passed += 1
+        else:
+            print(f"  FAIL {name}: expected {'pass' if expect_ok else 'fail'}")
+    tmp.unlink(missing_ok=True)
```
dmitry-tokarev-nv: Nit: hardcoded `/tmp` path. Two concurrent runs (parallel pre-commit invocations on a shared runner) collide on the same file, and a redirected or read-only `TMPDIR` fails the hook outright — reproduced locally as `PermissionError: [Errno 1] Operation not permitted: '/tmp/_agent_twin_case.mdx'`.

Needs `import tempfile` at the top.

```suggestion
    passed = 0
    with tempfile.TemporaryDirectory() as td:
        tmp = Path(td) / "_agent_twin_case.mdx"
        for name, body, expect_ok in cases:
            tmp.write_text(body, encoding="utf-8")
            if (not check(tmp, exp)) == expect_ok:
                passed += 1
            else:
                print(f"  FAIL {name}: expected {'pass' if expect_ok else 'fail'}")
```

docs/fern/scripts/gen_llms_tables.py
```diff
@@ -928,6 +1063,25 @@ def render_releases_data(data: dict) -> str:
...
+    ``platform.os[].status``. So the JSON carries both — ``scope`` as the
+    precise field, ``status`` re-emitted with the value those rows always
+    carried. ``status`` is deprecated; read ``scope``.
+    """
+    plat = {k: copy.deepcopy(v) for k, v in data["PLATFORM"].items()}
+    for key in ("os", "csp"):
+        for row in plat.get(key, []):
+            if "scope" in row and "status" not in row:
+                row["status"] = "Supported"
+    return plat
```
dmitry-tokarev-nv, line 1076: **The docstring's premise doesn't hold, and this publishes an unconditional support claim.** "`status` re-emitted with the value those rows always carried" — but the diff removes a row that carried `status: "Experimental"`:

```
-    { name: "CentOS Stream", version: "9", arch: "x86_64", status: "Experimental", chip: "centos" },
```

So the rows did not always carry `Supported`. Nothing is misrepresented today (that row is gone, and both surviving scopes are container/wheel scopes), but the next row added with a preview or restricted scope gets published to `docs/fern/assets/releases.json` asserting `"status": "Supported"` — a published, stably-schema'd asset claiming support that doesn't exist, with no gate to catch it.

An explicit map that fails on an unmapped scope, matching how `feature_cell` handles an unknown status (L402-403):

```suggestion
    plat = {k: copy.deepcopy(v) for k, v in data["PLATFORM"].items()}
    for key in ("os", "csp"):
        for row in plat.get(key, []):
            if "scope" in row and "status" not in row:
                scope = row["scope"]
                if scope not in SCOPE_STATUS:
                    raise TSParseError(
                        f"unknown platform scope {scope!r}; add it to SCOPE_STATUS "
                        f"so releases.json does not claim support it never had"
                    )
                row["status"] = SCOPE_STATUS[scope]
    return plat
```

with, near the other module constants:

```python
# Legacy ``status`` values for the scopes that replaced them. Deliberately
# exhaustive: a new scope must state its deprecated status rather than
# defaulting to "Supported" in a published asset.
SCOPE_STATUS = {
    "Containers and wheels": "Supported",
    "Wheels only": "Supported",
}
```

docs/fern/scripts/gen_llms_tables.py
```diff
@@ -443,6 +441,52 @@ def feature_table(data: dict) -> str:
...
+                    grid[c][r] = cell  # mirror, so both directions read alike
+        body: list[list] = []
+        for r, feature in enumerate(features):
+            row = [feature]
+            for c in range(len(features)):
+                cell = grid[r][c]
+                if cell is None:
+                    row.append(None)
+                    continue
+                text = INTERACTION_GLYPH[cell["status"]]
```
dmitry-tokarev-nv: Nit: an unmapped status here raises a bare `KeyError`. It is caught by `except (TSParseError, KeyError)` in `main()`, so no traceback — but the operator sees `error: rendering failed: KeyError('preview')` with no indication of which map or how to fix it. Every other parse failure in this module raises `TSParseError` with an actionable message, including the directly analogous `feature_cell` at L402-403.

```suggestion
                status = cell["status"]
                if status not in INTERACTION_GLYPH:
                    raise TSParseError(
                        f"unknown interaction status {status!r} for "
                        f"{matrix['backend']}; add it to INTERACTION_GLYPH"
                    )
                text = INTERACTION_GLYPH[status]
```

dmitry-tokarev-nv (commented):
Reviewed with a DevOps focus — the pre-commit wiring and the two scripts it gates. **No blockers.** Everything below is a gap in the new gates rather than something that breaks the build, and the four `.pre-commit-config.yaml` comments are all the same shape: the hooks are scoped more narrowly than the code they are meant to guard.

The two I'd most want addressed before merge:

- `check-agent-twins` fires on 1 of the 3 pages its own `IN_SCOPE` list governs, so two of the three twins this PR is built to protect are ungated — and a rename of either throws the hard-fail onto an unrelated PR later.
- `gen-llms-tables` runs the tests but never `--check`, so `releases.json` and every generated span can silently go stale. Nothing else in the repo checks it.

One item that has no diff line to hang off: **`minor_line` (L690) and `group_by_minor_line` (L701) in `gen_llms_tables.py` are now dead** — their only caller, `render_support_matrix`, is deleted in this PR. Nothing under `docs/` or `.github/` references them.

### Validated locally at `de97c69`

- `check_agent_twins.py` → exit 0; `--test` → 14/14 passed
- `gen_llms_tables.py --check` → all 7 outputs `unchanged`
- `pytest -c docs/fern/scripts/pytest.ini ... test_gen_llms_tables.py` → 14 passed
- `driver_floor_table` output hand-checked against `CUDA_HISTORY`: the `None` cells for TensorRT-LLM at the 570/575 floors and the `0.7.0.post1` post-train substitution are both correct
- `check_style_components.py` run manually against `ReleaseSupportMatrix.tsx` and `FeatureInteractions.tsx` → clean (so the scoping comment is about coverage, not a live defect)

The 40% coverage figure in the `check_agent_twins.py` comment is measured, not estimated: I deleted the CUDA table from the twin and re-ran the check.

Nice work on the twin gate generally — the `_interactions_segment` scoping and the "cases that have actually failed here" self-test are the right instincts. Most of this is asking for those same instincts applied one level wider.

.pre-commit-config.yaml
dagil-nvidia: Fixed. The pattern now covers all three IN_SCOPE pages, with a note that it has to stay in step with IN_SCOPE — verified each of the three matches.

.pre-commit-config.yaml
dagil-nvidia: Fixed. Added a `gen-llms-tables-check` hook running `--check`, scoped to the generator, `releases.data.ts`, the reference pages and `releases.json`.

.pre-commit-config.yaml
dagil-nvidia: Fixed. `pass_filenames: false` added, matching its sibling below.

docs/fern/scripts/check_agent_twins.py
dagil-nvidia: Fixed. Uses `tempfile.TemporaryDirectory()` with `import tempfile` added.

docs/fern/scripts/check_agent_twins.py
dagil-nvidia: Fixed. Bounded on the array literal instead of the next `export const`. Now returns 10 row labels with no backends; it was 12 with SGLang and TensorRT-LLM included and vLLM dropped.

docs/fern/scripts/check_agent_twins.py
dagil-nvidia: Fixed. Added `RELEASES_HEADING` and `_releases_segment`, mirroring `_interactions_segment`. Reproduced your measurement: gutting the release rows scored 40% before and now reports 0% and fails. Added as a self-test case, and the release fixtures gained the heading since one without it models a page that scores zero.

docs/fern/scripts/gen_llms_tables.py
dagil-nvidia: Fixed. Added `SCOPE_STATUS` mapping both current scopes explicitly and raising `TSParseError` on an unmapped one, matching how `feature_cell` handles an unknown status.

docs/fern/scripts/gen_llms_tables.py
dagil-nvidia: Fixed. Raises `TSParseError` naming the status, the backend, and the map to add it to.

.pre-commit-config.yaml
dagil-nvidia: Covered, but scoped differently — `components/.*\.tsx` fails on three existing components.

I ran the check against each: `FeatureHeatmap` (a label template at :345), `ModelEABuildCards` (composes CSS via `${EA_DOTS_CSS}`, and a `pullCommand`) and `TerminalDemo` (an `asciinema-player@${PLAYER_VERSION}` URL) all exit 1. `ReleaseSupportMatrix` and `FeatureInteractions` exit 0, as you found.

The cause is that `LITERAL` matches every `const X = ` + backtick literal, not only CSS ones, and the check forbids interpolation — correct for a pure CSS literal, wrong for ordinary TSX.

So the two new components are named explicitly instead, in both the hook pattern and the argless default. Your concern is covered without reddening three files that were never in scope. Argless run now reports 6 components clean.

A rule that actually distinguishes CSS literals from ordinary ones would let the glob widen properly, but that is a change to the check rather than its scope, and I did not want to smuggle it in here.

dmitry-tokarev-nv (commented):
Re-reviewed at `98425e2`. All nine are addressed and the threads are already resolved, so nothing is outstanding from the first pass — this is three new items, none of them blockers.

### On the one you adjusted

Your pushback on the style-hook scoping is right, and my suggestion was wrong: `LITERAL` matches every `const X = \`…\`` literal rather than only CSS ones, and forbidding `${` is correct for a pure CSS literal and wrong for ordinary TSX. Naming the two components is the right call.

One correction to the count, which matters for the follow-up you floated: it is **four** components, not three. `TagLookup.tsx:118` also fails — and it is the interesting one, because `TG_CSS` *is* a real CSS literal that interpolates legitimately (`${ENTRIES.map((_, i) => …)}`). So a rule that distinguishes CSS literals from ordinary template literals would still not let the glob widen; `TagLookup` would need the interpolation rule relaxed, not the CSS detection improved. Worth knowing before that follow-up gets scoped.

### Verified at `98425e2`

- `check_agent_twins.py` → exit 0 on all three pages; `--test` → 15/15
- `gen_llms_tables.py --check` → all 7 outputs unchanged; `test_gen_llms_tables.py` → 14 passed
- `_releases_segment` re-tested destructively against the real page, not just the fixture: gutting the release table under the heading now reports **0% of what ReleaseSupportMatrix renders, below 100%** and exits 1. That was 40% and passing before, so the scoping does what it claims on the live twin, not only in the self-test.
- `_interaction_features` → 10 row labels, no backends
- Style hook covers 6 components clean; the argless default matches the hook pattern
- `minor_line` / `group_by_minor_line` gone, no residual references

The self-test case you added for the unscoped search — versions in a backend-pins position, heading present, rows deleted — is exactly the shape that was slipping through. Good addition.

docs/fern/scripts/check_style_components.py
```diff
@@ -68,7 +69,15 @@ def check(path: Path) -> list[str]:
 
 def main() -> int:
     args = [Path(a) for a in sys.argv[1:]]
```
dmitry-tokarev-nv, line 76: **P2 / Medium — this hook cannot report a defect.** Pre-existing, but the file is touched here and its scope just widened, so it is worth closing now.

`check()` builds its diagnostic with `path.relative_to(REPO)` (L63), unguarded. Pre-commit passes repo-relative paths, so the moment the check actually finds something, that call raises `ValueError` before the message prints. Reproduced from the repo root at `98425e2`:

```
$ python3 docs/fern/scripts/check_style_components.py docs/fern/components/FeatureHeatmap.tsx
Traceback (most recent call last):
  ...
ValueError: 'docs/fern/components/FeatureHeatmap.tsx' is not in the subpath of
'/…/dynamo' OR one path is relative and the other is absolute
```

The gate still blocks (exit 1), so nothing bad merges — but the "raw backtick inside `RSM_CSS` / A raw backtick ends the template literal and breaks the build" message, which is the entire value of this check, never reaches the author. They get a `pathlib` traceback instead. Clean files are unaffected, which is why it has stayed hidden: the failure only appears on the one run that matters.

`check_agent_twins.py` guards the identical call in both places (`path.relative_to(REPO) if path.is_relative_to(REPO) else path`, L247 and L256), so the pattern is already established in this PR.

Resolving at the entry point fixes it for every caller:

```suggestion
    args = [Path(a).resolve() for a in sys.argv[1:]]
```

Verified with that one line applied — the same command now prints all four diagnostics and exits 1. Worth also guarding L63 the way `check_agent_twins.py` does, for a path passed from outside the repo.

.pre-commit-config.yaml
```diff
@@ -104,13 +104,57 @@ repos:
...
+        - pytest
+    - id: gen-llms-tables-check
+      name: Check the generated docs tables are current
+      entry: python3 docs/fern/scripts/gen_llms_tables.py --check
+      language: system
+      # The hook above proves the generator works; this one proves it was
+      # run. Without it, editing releases.data.ts and forgetting to
+      # regenerate leaves releases.json and every generated span stale,
+      # with nothing else in the repo checking.
+      files: ^docs/fern/(scripts/gen_llms_tables\.py|components/releases\.data\.ts|pages/reference/general/.*\.mdx|assets/releases\.json)$
```
dmitry-tokarev-nv: **P3 / Low — `releases-atom.xml` is generated but not gated.** The pattern lists `assets/releases\.json` but not `assets/releases-atom\.xml`, and `--check` covers both (`ATOM_PATH` at `gen_llms_tables.py:83`, `ASSET_OUTPUTS`). Its own run confirms it: `releases-atom.xml: unchanged` is line 7 of the output.

So a hand-edit to the Atom feed is the one generated artifact that can drift without this hook noticing — the same shape as the gap this hook was added to close, and inconsistent with including `releases.json` right next to it.

```suggestion
      files: ^docs/fern/(scripts/gen_llms_tables\.py|components/releases\.data\.ts|pages/reference/general/.*\.mdx|assets/releases(-atom\.xml|\.json))$
```

docs/fern/scripts/gen_llms_tables.py
```diff
@@ -928,6 +1048,31 @@ def render_releases_data(data: dict) -> str:
...
+def platform_payload(data: dict) -> dict:
+    """PLATFORM with the retired ``status`` key restored for consumers.
+
+    releases.data.ts renamed ``status`` to ``scope`` because "Supported" read
+    the same on a shipped container base and on a wheels-only host. That rename
+    is right for the page, but releases.json is a published asset with a stable
+    schema: dropping a key breaks anything already reading
+    ``platform.os[].status``. So the JSON carries both — ``scope`` as the
+    precise field, ``status`` re-emitted with the value those rows always
+    carried. ``status`` is deprecated; read ``scope``.
```
dmitry-tokarev-nv: **P3 / Low — the docstring still carries the premise the fix disproved.** "`status` re-emitted with the value those rows always carried" was the claim the CentOS Stream row (`status: "Experimental"`, removed in this PR) contradicted. `SCOPE_STATUS` is the right fix, but the prose above it now describes neither the old behaviour accurately nor the new behaviour at all — and it is the first thing the next person reading this function sees.

```suggestion
    ``platform.os[].status``. So the JSON carries both — ``scope`` as the
    precise field, ``status`` derived from it through SCOPE_STATUS. Deriving
    rather than defaulting is deliberate: the rows have not all carried
    "Supported" (the CentOS Stream row was "Experimental"), so an unmapped
    scope fails closed instead of publishing a support claim. ``status`` is
    deprecated; read ``scope``.
```

docs/fern/scripts/check_style_components.py
dagil-nvidia: Fixed. Reproduced exactly — the ValueError fired before the message printed. Guarded with `path.relative_to(REPO) if path.is_relative_to(REPO) else path`, matching both call sites in `check_agent_twins`. The diagnostic now prints for relative and absolute paths alike; verified with both.

.pre-commit-config.yaml
dagil-nvidia: Fixed. Pattern now matches `assets/releases(-atom\.xml|\.json)`. Confirmed it matches both generated assets and not an unrelated file under `assets/`.

docs/fern/scripts/gen_llms_tables.py
dagil-nvidia: Fixed, taking your wording. The docstring now describes `SCOPE_STATUS` and why deriving fails closed, instead of the premise the CentOS Stream row disproved.

docs/fern/pages/reference/general/compatibility.mdx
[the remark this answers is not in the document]
dagil-nvidia: Addressed by the restructure. `InteractionStatus` is gone — the matrices render through `FeatureInteractions` now, and `check_agent_twins.py` gates the twin: the interaction row labels must appear inside the twin's interactions section, scoped so the feature table above cannot satisfy them. Deleting the matrix rows fails the check.

docs/fern/components/ReleaseSupportMatrix.tsx
```diff
@@ -0,0 +1,620 @@
...
+  }
+  return { release, version, blocks, postTrains };
+}
+
+function groupByMinorLine(): MinorGroup[] {
+  const releases = releaseIndex();
+  const supported = supportedReleases();
+  const groups = new Map<string, MinorGroup>();
+  const seenVersions = new Set<string>();
+  for (const row of CUDA_HISTORY) {
```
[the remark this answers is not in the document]
dagil-nvidia: Still held — fixed in `632fc97`. Verified: v0.6.0 and v0.6.1 are stable releases with no `CUDA_HISTORY` row, plus five post-releases across v0.6.1, v0.7.0 and v0.8.1, so the matrix omitted seven releases the prose promised.

Fixed the claim rather than the data, since deriving from `CUDA_HISTORY` is what the rows are. The section now says it covers releases with recorded CUDA data and names the ones predating that tracking.

docs/fern/components/ReleaseSupportMatrix.tsx
```diff
@@ -0,0 +1,620 @@
...
+    );
+    const block = buildReleaseBlock(version, release, versionRows, postTrains);
+    const group = groups.get(line) ?? {
+      minor: line,
+      releases: [],
+      majors: [],
+      // Counted from RELEASES, not from the cards: a post train shares its
+      // base's card, so counting cards would report 0.8.x as two releases
+      // when the line actually shipped five.
+      releaseCount: supported.filter((v) => minorLine(v) === line).length,
```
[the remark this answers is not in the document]
dagil-nvidia: Same finding as the `ReleaseSupportMatrix.tsx:431` thread — fixed in `632fc97`. The prose now states the matrix covers releases with recorded CUDA data and names the seven it omits, instead of claiming every stable release and patch.

docs/fern/pages/reference/general/compatibility.mdx
```diff
@@ -118,83 +103,95 @@ TensorRT-LLM delivers maximum inference performance and optimization, with full
...
+| **LoRA** | | | | <InteractionStatus status="wip" /> | | | | <InteractionStatus status="na" /> | | |
+| **Tool Calling** | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="wip" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | <InteractionStatus status="yes" /> | | <InteractionStatus status="na" /> | |
+| **Speculative Decoding** | <Tooltip tip="Code hooks exist, but examples and documentation are not yet available."><InteractionStatus status="wip" noteLabel="Limited integration" /></Tooltip> | <InteractionStatus status="wip" /> | <InteractionStatus status="na" /> | <InteractionStatus status="wip" /> | <InteractionStatus status="na" /> | <InteractionStatus status="wip" /> | <InteractionStatus status="na" /> | | <InteractionStatus status="wip" /> | <InteractionStatus status="na" /> |
+
```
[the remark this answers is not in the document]
dagil-nvidia: Addressed by the restructure. The SGLang LoRA row is no longer blank cells — every pairing carries an explicit `status: "no"` with the note "SGLang does not support LoRA in Dynamo, so every LoRA pairing is unsupported" and a source link, so there is nothing left to infer from a blank.

docs/fern/components/ReleaseSupportMatrix.tsx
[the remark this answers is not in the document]
dagil-nvidia: You are right and my earlier fix was wrong — reverted in `b874ce5`.

I compared `RELEASES` against `CUDA_HISTORY` at the data level, found seven stable and patch releases with no CUDA rows, and concluded the matrix dropped them. It does not: `ReleaseSupportMatrix` iterates `RELEASES` filtered by kind (L218) and `releaseLines` falls back to `unrecordedLines` (L328), so those seven render with toolkit and driver marked "Not recorded". Confirmed all seven are kind stable or patch and the matrix renders 22 releases.

So the original claim was accurate and is restored. I kept one sentence naming the "Not recorded" treatment, since the prose never explained it and it is what a reader meeting those rows would wonder about.

My error was reading the data instead of the rendering path. Thanks for catching it.

docs/fern/pages/reference/general/compatibility.mdx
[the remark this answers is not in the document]
dagil-nvidia: Thanks for confirming.

docs/fern/components/ReleaseSupportMatrix.tsx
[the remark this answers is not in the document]
dagil-nvidia: Correction to my reply above: this finding was already fixed, and my "still held" assessment was wrong.

`ReleaseSupportMatrix` iterates `RELEASES` filtered by kind and falls back to `unrecordedLines` for releases with no CUDA rows, so v0.6.0 and v0.6.1 do appear, marked "Not recorded". The restructure that introduced `unrecordedLines` closed this. My prose change is reverted in `b874ce5`.

docs/fern/components/ReleaseSupportMatrix.tsx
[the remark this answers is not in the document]
dagil-nvidia: Correction to my reply above: same as the sibling thread — this was already fixed by `unrecordedLines`, not still outstanding. My prose change is reverted in `b874ce5`.

dmitry-tokarev-nv (approved):
Thank you for addressing comments.
Docs changes are low risk.
LGTM from DevOps perspective.

dmitry-tokarev-nv:
Approved. One correction to something I asserted earlier, since it is wrong on the record and is the kind of claim that gets reused.

On the `check-agent-twins` scoping thread I wrote that a rename of an `IN_SCOPE` page would throw the script's hard-fail "onto the next unrelated PR that touches any `components/*.tsx`", and the commit message repeats that framing. That scenario does not occur. `pre-merge.yml` runs `pre-commit/action` with no arguments, which is `run --all-files`, so every hook in this config executes on every PR and every push to `main` regardless of its `files:` pattern. I confirmed the behaviour rather than inferring it — `pre-commit run check-agent-twins-selftest --all-files` reports `Passed`, not `Skipped`, despite its pattern naming a single file.

The practical difference: a `files:` pattern does not scope which PRs a hook can break. It scopes which files get passed to the hook, and which local `git commit` runs invoke it. So a rename trips the check on the renaming PR itself, with `::error::in-scope page not found` naming the file — it fails closed and self-reports to the person who caused it, which is the right behaviour and not the deferred landmine I described.

Widening the pattern to all three `IN_SCOPE` pages was still worth doing: it is what makes the hook fire on a local `git commit` that touches `release-artifacts.mdx` or `model-early-access-builds.mdx`, and it keeps the pattern honest about what the script governs. The fix stands; only my rationale for it was wrong.

For the record, on the CI and security surface, since this config gates both PR and post-merge:

- The `pre-commit` job on `7dc4547` took 130s against its `timeout-minutes: 3` cap. Recent `main` runs: 134s, 98s, 92s, 99s. The three added hooks run in under 0.1s each, so this sits inside main's existing variance. (That the cap is already at ~74% on main is worth its own look, unrelated to this PR.)
- All new hooks are under the existing `repo: local` block running repo-local scripts; no new external pre-commit repo or third-party Action, and `pre-commit/action` stays SHA-pinned. No `eval`, `exec`, `subprocess`, or network call in any of the three scripts.
- Nothing outside `.pre-commit-config.yaml` and `docs/fern/**` is touched, so no core code or functional test is in scope.

No action needed on any of this.

dagil-nvidia:
Thanks — confirmed your correction independently before relying on it. `pre-merge.yml` passes no `extra_args` to `pre-commit/action@v3.0.1`, so it is `run --all-files`, and `pre-commit run check-agent-twins-selftest --all-files` reports `Passed` here too, not `Skipped`. A `files:` pattern scopes which filenames reach a hook and which local `git commit` invokes it — not which PRs it can break. Nothing in the tree encodes the old framing; it lived only in that thread and a commit message.

While checking that, I found a worse error of my own in the same area, so putting it on the record too.

The merge commit `7dc4547` claims it fixed a pre-existing defect: `gen-llms-tables-check` missing `pass_filenames: false`. That is wrong. The hook **already had** the flag at `b874ce5`. My conflict resolution dropped it — the conflict boundary cut through that hook's definition, and the union I applied kept HEAD's block only up to `files:`, leaving the flag behind. I then hit the failure locally, diagnosed it as a branch defect, and "fixed" what I had just broken one step earlier.

Two things follow. The file is correct now, so there is nothing to re-fix — `pass_filenames: false` is present and the full pre-commit run passes. But the commit message misattributes the cause, and `pre-commit` was **green on `b874ce5`**, which is the evidence I should have checked before claiming a pre-existing defect rather than after.

The lesson generalises past this hook: a regex union across conflict markers is not safe when the boundary falls inside a block, because it silently truncates the block rather than failing. Verifying the resolved file against the pre-merge version — not just that it parses — is what would have caught it.

No action needed from you; your approval stands on the file state, which is right.

PR #10921, opened 2026-06-24
feat(runtime): add TLS support to TCP request plane

walkoss:
## Overview

Adds opt-in TLS encryption to the TCP request plane (frontend ↔ worker) using [rustls](<https://github.com/rustls/rustls>) with the `ring` crypto provider. When no TLS env vars are set the transport behaves exactly as before — fully backward compatible.

## Details

**New shared helper** (`lib/runtime/src/tls_utils.rs`):

* `server_tls_config(cert, key)` — builds a `rustls::ServerConfig`
* `client_tls_config(ca, insecure)` — builds a `rustls::ClientConfig`; emits a prominent warning when `insecure=true`

**TCP server** (`tcp/server.rs`):

* Reads `DYN_TCP_TLS_CERT_PATH` + `DYN_TCP_TLS_KEY_PATH` at startup to optionally build a `TlsAcceptor`
* TLS handshake is spawned per-connection (non-blocking accept loop), bounded by a 10s timeout
* `BoxRead`/`BoxWrite` (`Box<dyn AsyncRead/AsyncWrite + Unpin + Send>`) unify TLS and plaintext streams through all inner functions with no branching after accept
* `CallHomeHandshake` read and response-stream prologue read are also bounded by 10s timeouts

**TCP client** (`tcp/client.rs`):

* `OnceLock<Option<TlsConnector>>` — connector built once from env on first connection, `Arc`-cloned per connection
* `connect_and_split()` upgrades to TLS when configured, with a 10s handshake timeout
* IPv6-safe SNI parsing with `DYN_TCP_TLS_SERVER_NAME` override for IP-addressed servers

**Environment variables** (all optional, plaintext when unset):

| Variable | Side | Purpose |
| -- | -- | -- |
| `DYN_TCP_TLS_CERT_PATH` | Server | PEM server certificate |
| `DYN_TCP_TLS_KEY_PATH` | Server | PEM server private key |
| `DYN_TCP_TLS_CA_CERT_PATH` | Client | CA cert to verify server |
| `DYN_TCP_TLS_INSECURE` | Client | Skip cert verification (dev only) |
| `DYN_TCP_TLS_SERVER_NAME` | Client | SNI override for IP-addressed servers |

## Where should the reviewer start?

1. `lib/runtime/src/tls_utils.rs` — new file, self-contained rustls helpers
2. `lib/runtime/src/pipeline/network/tcp/server.rs` — `build_tls_acceptor()` and the accept loop spawn
[...]

Issue #10809, opened 2026-06-20
[CONTRIBUTION]: Add opt-in TLS and mTLS support to NATS and TCP transports


walkoss:
## Type of Change

New feature

## Problem Statement

All service-to-service communication in Dynamo is currently plaintext. In environments where traffic traverses untrusted or shared networks — such as multi-tenant Kubernetes clusters — there is no encryption in transit and no mutual authentication between components. Specifically:

- The **NATS transport** (`async-nats` client) connects to the NATS server without TLS, exposing event-plane traffic.
- The **TCP request plane** (frontend ↔ worker inference routing) uses raw `TcpStream` connections with no encryption or identity verification.

This leaves inference request payloads, KV cache metadata, and control messages unprotected on the wire.

## Proposed Solution

Add opt-in TLS and mutual TLS (mTLS) to both transports, controlled entirely by environment variables so deployments without TLS are unaffected.

A new shared `tls_utils` module loads PEM certificates via `rustls` and provides `server_tls_config` and `client_tls_config` helpers used by both transports.

**NATS** — env vars: `NATS_TLS_CA_CERT_PATH`, `NATS_TLS_CLIENT_CERT_PATH`, `NATS_TLS_CLIENT_KEY_PATH`, `NATS_TLS_INSECURE`. When a CA cert or client cert is configured, the NATS client upgrades to TLS and optionally presents a client certificate for mTLS.

**TCP request plane** — env vars: `DYN_TCP_TLS_CERT_PATH`, `DYN_TCP_TLS_KEY_PATH`, `DYN_TCP_TLS_CA_CERT_PATH`, `DYN_TCP_TLS_CLIENT_CERT_PATH`, `DYN_TCP_TLS_CLIENT_KEY_PATH`, `DYN_TCP_TLS_CLIENT_CA_CERT_PATH`, `DYN_TCP_TLS_SERVER_NAME`, `DYN_TCP_TLS_INSECURE`. The server listener optionally wraps accepted connections in TLS (with a per-connection handshake timeout to avoid blocking the accept loop), and can require client certificates for mTLS. Clients optionally upgrade connections and present a certificate.

All internal handler functions are updated to work with both plaintext and TLS streams via boxed `AsyncRead`/`AsyncWrite` trait objects — no generics proliferation, no new dependencies. All 440 existing tests continue to pass unchanged.

## Estimated PR Size

XL (501–1000 lines)

## Files/Components Affected

[...]

PR #12533, opened 2026-08-01
feat(runtime): encrypt TCP request dispatch (frontend → worker)

walkoss:
## Overview

Extends TLS encryption to the TCP request plane (`egress/tcp_client` / `ingress/shared_tcp_endpoint`), which carries user prompts from frontend to worker. This closes the security gap where inference request payloads were transmitted in plaintext while only the response stream was encrypted (#10921).

## Details

**Server side** (`ingress/shared_tcp_endpoint.rs`):
- Builds `TlsAcceptor` from `DYN_TCP_TLS_CERT_PATH`/`KEY_PATH` at startup
- Per-connection TLS handshake with configurable timeout
- Panics on invalid TLS config (fail-closed, not silent fallback)
- `BoxRead`/`BoxWrite` unify TLS and plaintext through `read_loop`/`write_loop`

**Client side** (`egress/tcp_client.rs`):
- `OnceLock`-cached `TlsConnector` from `DYN_TCP_TLS_CA_CERT_PATH`/`INSECURE`
- Honors `DYN_TCP_TLS_SERVER_NAME` for SNI override
- TLS handshake with timeout on each pooled connection
- `BoxRead`/`BoxWrite` through `writer_task`/`reader_task`

Reuses the same `DYN_TCP_TLS_*` environment variables as the call-home transport (#10921). No new env vars needed.

1. `lib/runtime/src/pipeline/network/ingress/shared_tcp_endpoint.rs` - TLS acceptor in `new()` + `accept_loop()`
2. `lib/runtime/src/pipeline/network/egress/tcp_client.rs` - TLS connector in `connect()` + `get_request_plane_tls_connector()`

## Related Issues

- Depends on #10921 (TCP call-home TLS)
- Relates to #10809 (TLS contribution request)

## Summary by CodeRabbit

- **New Features**
  - Added optional TLS encryption for TCP request and response traffic.
  - Supports configurable certificates, CA validation, server names, insecure mode, and handshake timeouts.
  - TLS connections are securely established per connection.

- **Bug Fixes**
  - Invalid or incomplete TLS settings now prevent startup with clear configuration errors.
  - Failed or timed-out TLS handshakes are safely closed.

- **Documentation**
  - Expanded TLS guidance with supported traffic flows and configuration behavior.

### DIFF
--- a/docs/fern/pages/reference/components/tls-configuration.mdx
+++ b/docs/fern/pages/reference/components/tls-configuration.mdx
@@ -100,7 +100,7 @@ python -m dynamo.frontend \
 ## Kubernetes deployment
 
 In Kubernetes, TLS certificates are typically delivered by a certificate
-management system (e.g., cert-manager) and mounted into pods. Set the
+management system and mounted into pods. Set the
 environment variables on each component's pod template in the
 `DynamoGraphDeployment` spec:
 
@@ -153,11 +153,16 @@ When TLS is configured, the following frontend↔worker streams are encrypted:
 
 ## Design notes
 
+- Server certificates hot-reload: the request-plane and response-stream servers
+  serve their leaf cert/key through a resolver that re-reads the files from disk
+  when their contents change (detected by a content hash, so rotations done by
+  an atomic symlink swap are handled too), so certificate rotation
+  takes effect **without a process restart**. The check is rate-limited (at most
+  once every 30s, sooner after a failed reload) and never blocks a handshake; a
+  failed reload keeps serving the last valid certificate. Client trust anchors
+  (the CA) are still loaded once, so rotating the CA itself requires a restart.
 - Client TLS connectors are built once and cached via `OnceCell` on the first
-  outbound connection; the request-plane and response-stream servers build their
-  TLS acceptor eagerly at startup. Because the config is cached, certificate
-  rotation currently requires a process restart — hot-reloading certificates
-  from disk via a rustls certificate resolver is a possible future enhancement.
+  outbound connection.
 - The TLS handshake is spawned per-connection on both the request plane and
   response stream servers so the accept loop is never blocked.
 - Invalid TLS configuration on the request plane (e.g. bad cert path) prevents
--- a/docs/fern/pages/reference/components/tls-configuration.mdx
+++ b/docs/fern/pages/reference/components/tls-configuration.mdx
@@ -2,18 +2,22 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 title: TCP TLS
-subtitle: Encrypt all TCP connections between frontend and workers
+subtitle: Encrypt the TCP request and response streams between frontend and workers
 ---
 
-Dynamo supports opt-in TLS encryption on all TCP transports between frontends
-and workers. When enabled, both the **request plane** (frontend → worker
-inference requests) and the **response stream** (worker → frontend inference
-output) are encrypted using [rustls](https://github.com/rustls/rustls) with the
-`ring` cryptographic provider. When no TLS configuration is provided, all
-transports operate in plaintext exactly as before.
+Dynamo supports opt-in TLS encryption on the TCP request and response streams
+between frontends and workers. When enabled, both the **request plane**
+(frontend → worker inference requests) and the **response stream** (worker →
+frontend inference output) are encrypted using
+[rustls](https://github.com/rustls/rustls) with the `ring` cryptographic
+provider. When no TLS configuration is provided, these streams operate in
+plaintext exactly as before.
 
 The same `DYN_TCP_TLS_*` environment variables apply to both transports — a
-single set of certificates encrypts all TCP traffic between Dynamo components.
+single set of certificates encrypts the frontend↔worker TCP request and
+response streams. This does **not** cover the KV event plane, which is carried
+over ZMQ or NATS Core depending on setup and is a separate transport; those
+paths are not encrypted by this configuration.
 
 ## Environment variables
 
@@ -139,8 +143,7 @@ server and client depending on the stream direction.
 
 ## Encrypted paths
 
-When TLS is configured, all TCP communication between frontend and worker is
-encrypted:
+When TLS is configured, the following frontend↔worker streams are encrypted:
 
 | Path | Direction | Data | Transport |
 |---|---|---|---|
@@ -152,8 +155,9 @@ encrypted:
 
 - Client TLS connectors are built once and cached via `OnceCell` on the first
   outbound connection; the request-plane and response-stream servers build their
-  TLS acceptor eagerly at startup. Either way, certificate rotation requires a
-  process restart.
+  TLS acceptor eagerly at startup. Because the config is cached, certificate
+  rotation currently requires a process restart — hot-reloading certificates
+  from disk via a rustls certificate resolver is a possible future enhancement.
 - The TLS handshake is spawned per-connection on both the request plane and
   response stream servers so the accept loop is never blocked.
 - Invalid TLS configuration on the request plane (e.g. bad cert path) prevents
--- a/docs/fern/pages/reference/components/tls-configuration.mdx
+++ b/docs/fern/pages/reference/components/tls-configuration.mdx
@@ -2,16 +2,18 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 title: TCP TLS
-subtitle: Encrypt TCP streaming connections between frontend and workers
+subtitle: Encrypt all TCP connections between frontend and workers
 ---
 
-Dynamo supports opt-in TLS encryption on the TCP call-home streaming transport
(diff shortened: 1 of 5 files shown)


docs/fern/pages/reference/components/tls-configuration.mdx
```diff
@@ -135,12 +137,27 @@ server and client depending on the stream direction.
...
 ## Design notes
 
-- TLS configuration is cached after the first TCP connection via `OnceCell`.
-  Certificate rotation requires a process restart.
-- The TLS handshake is spawned per-connection on the server side so the accept
-  loop is never blocked by a slow handshake.
+- Client TLS connectors are built once and cached via `OnceCell` on the first
+  outbound connection; the request-plane and response-stream servers build their
+  TLS acceptor eagerly at startup. Either way, certificate rotation requires a
+  process restart.
```
sttts: this surprises. Why is that? Shouldn't be hard to catch up a changed cert from disk and serve it when a request comes in. I bet rutls has that built in.

docs/fern/pages/reference/components/tls-configuration.mdx
```diff
@@ -2,16 +2,18 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 title: TCP TLS
-subtitle: Encrypt TCP streaming connections between frontend and workers
+subtitle: Encrypt all TCP connections between frontend and workers
```
sttts: This is not yet true. The KV cache events go through ZMQ or NATS depending on setup. And they are kind of sensitive as well because of potential timing and dictionary attacks.

walkoss: Good catch, you're right! the `DYN_TCP_TLS_*` certs now explicitly cover only the frontend↔worker TCP request/response streams, with a note that the KV event plane (ZMQ or NATS) is a separate transport not encrypted by this configuration. Agreed the event plane is sensitive and NATS TLS is a follow-up PR in this stack, and ZMQ has no TLS story yet, so I'll keep that scoped and tracked separately rather than implied here. https://github.com/ai-dynamo/dynamo/issues/10809#issuecomment-5176618852

docs/fern/pages/reference/components/tls-configuration.mdx
walkoss: Yes this not a rustls limitation, rustls does support hot-reload via a dynamic cert resolver (ResolvesServerCert) that re-reads from disk per handshake. I kept the cache-once approach to keep this PR minimal and softened the doc note accordingly
Should I file a follow-up issue to add on-the-fly cert reload so rotation doesn't need a restart or should I fix that in this PR. Thanks!

docs/fern/pages/reference/components/tls-configuration.mdx
```diff
@@ -2,16 +2,21 @@
...
 
-Dynamo supports opt-in TLS encryption on the TCP call-home streaming transport
-(implemented by `TcpStreamServer` and `TcpClient`, handling both response
-streams and request streams between frontends and workers). When enabled, all TCP
-connections on this path are upgraded to TLS using
-[rustls](https://github.com/rustls/rustls) with the `ring` cryptographic
-provider. When no TLS configuration is provided, the transport operates in
-plaintext exactly as before.
+Dynamo supports opt-in TLS encryption on all TCP transports between frontends
+and workers. When enabled, both the **request plane** (frontend → worker
```
sttts: here we still have the "all" which is slightly misleading.

walkoss: Sorry, fixed in https://github.com/ai-dynamo/dynamo/pull/12533/changes/d4786f488c0c6eb2e7af26e5493b071aa07a11e8

docs/fern/pages/reference/components/tls-configuration.mdx
walkoss: Fixed in https://github.com/ai-dynamo/dynamo/pull/12533/changes/d4786f488c0c6eb2e7af26e5493b071aa07a11e8

sttts (approved):
For docs.

docs/fern/pages/reference/components/tls-configuration.mdx
sttts: Yes, would follow-up with the reload. Without reload this feels incomplete.

sttts: Especially because without, and expiry, this leads to unexpected outages.

lib/runtime/src/pipeline/network/egress/tcp_client.rs
```diff
@@ -1713,6 +1774,66 @@ mod tests {
         (addr, conn_count)
     }
 
+    /// Full request-plane TLS handshake + encrypted round-trip using the real
+    /// `tls_utils` config builders (the same ones egress/ingress use). Generates
+    /// a self-signed cert, trusts it as the CA, and drives a verified handshake
+    /// with SNI `localhost` (mirroring `connector.connect(server_name, stream)`
+    /// in `TcpConnection::connect`), then echoes bytes over the encrypted stream.
+    #[tokio::test]
+    async fn request_plane_tls_handshake_roundtrip() {
```
jthomson04: [P2] Exercise the request-plane TLS path

This test constructs `TlsAcceptor` and `TlsConnector` directly, so it cannot catch broken env parsing, `OnceCell`/SNI wiring, boxed I/O, or request-plane reader/writer integration. Please drive one encrypted request through the actual request-plane client/server path; otherwise this is only another `tls_utils` round-trip test.

lib/runtime/src/pipeline/network/ingress/shared_tcp_endpoint.rs
```diff
@@ -150,6 +154,8 @@ pub struct SharedTcpServer {
     /// Overflow-queue capacity; `read_loop` compares against it to tell whether
     /// the queue is empty for the FIFO direct-dispatch rule.
     queue_capacity: usize,
+    /// Optional TLS acceptor for encrypting request plane connections.
+    tls_acceptor: Option<Arc<TlsAcceptor>>,
```
jthomson04: [P3] Drop the redundant `Arc`

`TlsAcceptor` is already cheaply `Clone` and internally holds an `Arc<ServerConfig>`. Store `Option<TlsAcceptor>` here; the outer `Arc` adds another pointer and atomic refcount without adding useful ownership semantics.

lib/runtime/src/pipeline/network/ingress/shared_tcp_endpoint.rs
```diff
@@ -395,8 +458,38 @@ impl SharedTcpServer {
...
+                                            tls.accept(stream),
+                                        )
+                                        .await
+                                        {
+                                            Ok(Ok(tls_stream)) => {
+                                                let (r, w) = tokio::io::split(tls_stream);
+                                                (Box::new(r), Box::new(w))
+                                            }
+                                            Ok(Err(e)) => {
+                                                tracing::warn!(
```
jthomson04, line 476: [P3] Use structured tracing fields

Please keep the peer and error queryable: `tracing::warn!(peer_addr = %peer_addr, error = %e, "Request-plane TLS handshake failed")`. The timeout warning immediately below should likewise record `peer_addr` as a field instead of formatting it into the message.

jthomson04 (approved):
The transport implementation follows #10921 closely. I left targeted follow-ups on test coverage and small Rust/tracing cleanup.

docs/fern/pages/reference/components/tls-configuration.mdx
walkoss: Added fixes for cert rotation

walkoss: server certs now hot-reload from disk on content change https://github.com/ai-dynamo/dynamo/pull/12533/commits/a85e95009843aa3bc28475f70081303f9ce79bf0

lib/runtime/src/pipeline/network/egress/tcp_client.rs
walkoss: Done. Replaced the direct-acceptor/connector test with request_plane_tls_end_to_end, which drives the real client path - TcpConnection::connect + TLS handshake (SNI from env) + boxed I/O + reader/writer + send_request framing - against a TLS-wrapped echo server built from the production server_tls_config. Added a separate request_plane_tls_connector_from_env_parses test for the env parsing.

lib/runtime/src/pipeline/network/ingress/shared_tcp_endpoint.rs
walkoss: Done - field is now `Option<TlsAcceptor>`.

lib/runtime/src/pipeline/network/ingress/shared_tcp_endpoint.rs
walkoss: Done - both the handshake-failed and handshake-timeout warnings now use `peer_addr = %peer_addr, error = %e` structured fields instead of interpolating into the message.

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/pull/12533/commits/c4a66a43576c4f42cc58e226cb005a10517c7fcb

sttts:
/ok to test https://github.com/ai-dynamo/dynamo/commit/e94247a882d03a2dd7bcbbc34060f79ecc0fccb9

Issue #8636, opened 2026-04-23
Streaming drops parallel tool_calls on vLLM path when enable_thinking=False


keivenchang:
### Summary
Streaming chat completions with parallel tool calls silently drop every `tool_call` on the vLLM `enable_thinking=False` path. Clients see `reasoning_content` spam and `finish_reason=stop` instead of `tool_calls`.

### Root cause
Qwen3's `enable_thinking=False` chat template puts an empty `<think>\n\n</think>` block into the **prompt**, so the generated **output** never contains `</think>`. `StreamingPostProcessor` only exits "reasoning mode" when it sees `</think>` in a chunk — since that never arrives, it stays in reasoning mode for the whole stream. Every chunk is routed to the reasoning parser and labeled `reasoning_content`; the tool parser is never called.

### Repro
Streaming chat-completions request with `tool_choice=auto`, `parallel_tool_calls=true`, `chat_template_kwargs={"enable_thinking": false}`, and a prompt that elicits multiple tool calls (e.g. "What's the weather in Tokyo and Paris?"). Observe:
- Zero `tool_calls` deltas in the SSE stream
- `finish_reason=stop` on the terminal frame (should be `tool_calls` per OpenAI ChatCompletion enum)
- Tool-call markup appears in `reasoning_content` deltas

Single-tool requests on the same stack work fine — the bug only surfaces on parallel tool calls.

### Scope
- Affects: vLLM backend, frontend post-processor (`components/src/dynamo/frontend/prepost.py`)
- Unaffected: SGLang (different architecture, already remaps `finish_reason` inline)

PR #12541, opened 2026-08-02
fix(frontend): call the reasoning parser's adjust_request during preprocessing

yifjiang:
> **Context: this is one of six fixes from bringing up [`thinkingmachines/inkling`](https://huggingface.co/thinkingmachines/inkling) on Dynamo + vLLM** (4×GB200, TP=4). inkling's reasoning parser is vLLM's `InklingParser` (`vllm/parser/inkling.py`, via `InklingParserReasoningAdapter`) — a **text-grammar** parser. The bugs it exposed are general: they hit any model whose parser overrides sampling params through `adjust_request()`. Sibling PRs: #12510, #12511, #12507, #12543.

## Problem

vLLM reasoning parsers expose `adjust_request()` so they can declare the sampling params they need. Parsers that split thinking from the answer on **text markers** require `skip_special_tokens=False` — otherwise the markers are stripped during detokenisation before the parser ever sees them, and the split silently does not happen.

`_prepare_request` calls the **tool** parser's `adjust_request`:

```python
if tool_parser_class and (has_tools or enable_auto_tool_choice):
    if request_for_sampling.tool_choice != "none":
        tool_parser = tool_parser_class(tokenizer, request_for_sampling.tools)
        request_for_sampling = tool_parser.adjust_request(request_for_sampling)
```

There is **no equivalent for the reasoning parser** — `_prepare_request` does not even receive `reasoning_parser_class`, because the reasoning parser is constructed later, in `StreamingPostProcessor.__init__`, long after the sampling request is finalised.

**Net effect:** `reasoning_content` comes back empty unless the *client* sends `skip_special_tokens: false`. Ordinary OpenAI-compatible clients do not, so for such models the reasoning channel silently does nothing on normal traffic. Note the tool path is additionally gated on `tool_choice != "none"`, so a plain tools-free chat request currently gets no `adjust_request` call from **either** parser.

## Fix

Mirror the tool path: construct the reasoning parser during request preparation and let it adjust the request.

### The gating is the load-bearing part

The gate is now single-sourced as `_reasoning_parser_enabled()`, a `TypeGuard` shared by `_prepare_request`, `preprocess_chat_request` and `StreamingPostProcessor`. The three must agree — if they drift, the sampling params get adjusted for a parser that then ends up `None`, and the model's control markers reach the client as **visible text**. The call sits after `chat_template_kwargs` is resolved, and is skipped when thinking is disabled (#8636).

This is a **no-op for parsers that do not need it** — the base `ReasoningParser.adjust_request` is `return request`.

### ⚠️ Two files are required

`prepost.py` alone is **inert**: `vllm_processor.py` must pass `reasoning_parser_class` into `preprocess_chat_request`, otherwise the new parameter defaults to `None` and the gate never fires. Worth knowing for review — a grep-based check on `prepost.py` cannot distinguish the patched-and-wired case from the patched-but-inert one.

## Review feedback (@rmccorm4) — addressed

> This can change more than `skip_special_tokens`. I reproduced a parser converting `response_format` to a structural tag here: the request carried the tag afterward, but `preprocess_chat_request()` still forwarded the pre-adjustment JSON schema because `assistant_guided_decoding` was captured earlier and adjusted guidance is only considered when `tool_parser` is set. Could we recompute and forward the reasoning parser's guidance independently of the tool parser?

Confirmed and fixed in `9432889`. Two things were wrong:

1. `parser_guided_decoding` was gated on `tool_parser is not None`, so a reasoning-parser rewrite was never even recomputed.
2. Even recomputed, it could not surface: `parser_guided_decoding` is consumed only by `build_tool_call_guided_decoding`, which returns early when the request has no tools.

Now the guidance is recomputed when **either** parser is active, and a rewrite that only the reasoning parser can have made — no tool parser in the request, so it is attributable — takes precedence over the caller's pre-adjustment copy.

**Scoped deliberately.** When a tool parser is *also* active the change is not attributable to either parser, and parser-vs-assistant precedence is already settled there: the caller's explicit constraint wins, pinned by `test_assistant_guidance_takes_precedence_over_auto_tool_guidance` and the `response-format-precedence` row of `TOOL_GUIDANCE_PARITY_CASES`. I tried the unified precedence first and it regressed exactly those three cases, which read as a deliberate policy rather than an oversight — happy to revisit it in a follow-up if you'd rather it change.

## Test

`TestReasoningParserAdjustRequest` — five cases using a stub parser, no real model needed:

- `adjust_request` is called and the mutation reaches the sampling request
- **not** called when `enable_thinking: False` (the gate that protects against marker leakage)
- called when `enable_thinking: True`
- no parser class ⇒ request untouched
- a pass-through `adjust_request` leaves the request unchanged (default-parser behaviour)

`TestReasoningParserGuidanceForwarding` — four cases for the review fix: the rewrite with no tool parser, a pass-through parser, no parser at all, and the thinking-disabled gate. **The first fails without this change** with the reported symptom — `{"json": {"type": "object"}}` instead of the structural tag.

Run against `main` in a dynamo+vLLM container: no regressions in `test_vllm_processor_unit.py` or `tests/frontend/test_prepost.py` versus a pristine-`main` baseline, and no new `ruff`/`mypy` findings.

**Verified end to end on a real deployment**, on a model whose reasoning parser sets `skip_special_tokens=False`. A 9-probe functional battery, run twice — once with the client sending `skip_special_tokens: false` and once without:

| pass | before | after |
|---|---|---|
| client sends `skip_special_tokens: false` | 9/9, `reasoning_content` 53 chars | 9/9 (no regression) |
| **client sends nothing** (ordinary request) | **7/9, `reasoning_content` 0 chars** | **9/9, `reasoning_content` 53 chars** |

Thinking-off requests were checked separately and are unaffected: clean visible text, empty `reasoning_content`, no marker leakage.

It also fixed a worse symptom on the same path: on the default pass, visible content was losing its **first token** when thinking was on (`408` → `''`, `top-left` → `-left`, `{"city":…` → `city":…`), scoring 2/9. That does not reproduce after this change, across a 3-cell disambiguation that separately exonerated CUDA graphs and MTP. Note this makes the buggy path unreachable rather than repairing the boundary handling itself — happy to file that separately.

## Notes

The behaviour change is server-visible and intentional: `skip_special_tokens=False` becomes the effective default for requests with reasoning active on models whose parser asks for it, matching what vLLM's own OpenAI server does. Happy to gate it behind a flag instead if that is preferred.

**Ordering with #12510.** This PR makes `skip_special_tokens=False` the effective default, which is exactly the condition under which unconsumed control markers leak into visible content — the leak #12510 fixes. #12510 is **not** in this branch's base. Merging this one alone puts that leak on `main`, so #12510 should land first or alongside. (`main` today has `_control_markers`/`_is_control_only_content`, but they are used only in the tool-call path as a suppression predicate for all-marker deltas, so mixed content like `<|end_message|>408<|end_message|>` still leaks.)

**Known follow-up, not in this PR:** the parser is constructed per request and discarded — measured p50 83.8 ms on a live deployment, which is noise at low rate but a ~40% TTFT hit on short prompts. The cleaner shape is to construct once and thread it through `PreprocessResult` into `StreamingPostProcessor` instead of `reasoning_parser_class`; that touches the postprocessor's signature and all its callers, so I kept it out of this change. Say the word and I'll fold it in.


yifjiang:
Additional evidence: this also fixes a **content-truncation** bug, not just empty `reasoning_content`.

On a model whose reasoning parser sets `skip_special_tokens=False` via `adjust_request`, requests that arrive **without** the client setting that field (i.e. ordinary traffic) and with thinking enabled were dropping the **first token of the visible answer**:

| expected | got | lost |
|---|---|---|
| `408` | `''` | all |
| `top-left` | `-left` | `top` |
| `{"city":"Reykjavik",…` | `city":"Reykjavik",…` | `{"` |

`finish_reason` was `stop` throughout and nothing leaked — content was being *removed*. It appeared only when thinking was ON; a `reasoning_effort: "none"` request was byte-clean.

**Attributed with a single-variable control** — two images differing only by this patch, same node, same flags, same model, same checkpoint:

| build | default pass (client sends nothing) | with `skip_special_tokens: false` |
|---|---|---|
| without this patch | **4/8** — chat, multimodal, reasoning, thinking_toggle all fail | 8/8 |
| with this patch | **8/8** | 8/8 |

Two further cells with the reasoning path exercised but speculative decoding disabled, across two different `cudagraph_mode` settings, were clean — so this is not a CUDA-graph or speculative-decoding artefact.

**Why this is consistent with the mechanism:** the drop only occurred on the `skip_special_tokens=true` path, which is exactly where the content-kind markers are stripped before the text-grammar parser can split on them. The boundary handling there is already doing something delicate, and it loses a token. This patch means that path is no longer taken.

**Worth being precise about scope:** this makes the buggy path *unreachable* for models whose parser wires `adjust_request`; it does not repair the boundary handling itself. Any model in that configuration without the wiring would still hit it — which is an argument for fixing it here in `prepost.py` rather than per-deployment. If a maintainer would rather I also chase the underlying truncation at the thinking→content boundary, happy to open a separate issue with the reproduction.

components/src/dynamo/frontend/prepost.py
```diff
@@ -491,6 +492,26 @@ def _prepare_request(
...
+    #
+    # ReasoningParser.adjust_request is `return request` by default, so this is a no-op
+    # for parsers that do not need it.
+    if reasoning_parser_class is not None and (
+        chat_template_kwargs.get("enable_thinking") is not False
+    ):
+        request_for_sampling = reasoning_parser_class(
+            tokenizer,
+            chat_template_kwargs=chat_template_kwargs,
+        ).adjust_request(request_for_sampling)
```
rmccorm4, line 570: This can change more than `skip_special_tokens`. I reproduced a parser converting `response_format` to a structural tag here: the request carried the tag afterward, but `preprocess_chat_request()` still forwarded the pre-adjustment JSON schema because `assistant_guided_decoding` was captured earlier and adjusted guidance is only considered when `tool_parser` is set. Could we recompute and forward the reasoning parser’s guidance independently of the tool parser?

rmccorm4 (commented):
One issue with reasoning-parser guidance propagation.

yifjiang:
Moved into the review thread it answers, so the discussion stays in one place: https://github.com/ai-dynamo/dynamo/pull/12541#discussion_r3865392614

components/src/dynamo/frontend/prepost.py
yifjiang: Thanks — reproduced and fixed in 9432889. You were right, and there were two layers to it:

1. `parser_guided_decoding` was gated on `tool_parser is not None`, so a reasoning-parser rewrite was never recomputed at all.
2. Even once recomputed it could not surface: `parser_guided_decoding` is consumed only by `build_tool_call_guided_decoding`, which returns early when the request has no tools. So with a reasoning parser and no tools there was no path to `guided_decoding` whatsoever.

Now the guidance is recomputed when **either** parser is active, and a rewrite that only the reasoning parser can have made — no tool parser in the request, so it is attributable — takes precedence over the caller's pre-adjustment copy.

One judgement call I'd like your read on. I first made the parser's rewrite win over `assistant_guided_decoding` unconditionally, which is the more obviously correct rule. It regressed exactly three cases — `test_assistant_guidance_takes_precedence_over_auto_tool_guidance` (both params) and the `response-format-precedence` row of `TOOL_GUIDANCE_PARITY_CASES` — which pin "the caller's explicit constraint beats the tool parser's rewrite". That reads as deliberate policy, not an oversight, so I scoped the new precedence to the reasoning-parser-only case and left the tool path exactly as it was. That does leave one gap: when both parsers are active the change isn't attributable to either, and a reasoning rewrite still loses to the caller's copy. Happy to widen it in a follow-up if you think the pinned tool-path policy should change too.

New coverage is `TestReasoningParserGuidanceForwarding` (4 cases). The first one fails without this change with exactly the symptom you described — `{"json": {"type": "object"}}` instead of the structural tag.

Also rebased onto latest `main`, which clears the `Generated API References` failure (that gate has since moved to write-mode generation, and both generators exit 0 here). Verified no regressions in `test_vllm_processor_unit.py` or `tests/frontend/test_prepost.py` against a pristine-`main` baseline, and no new `ruff`/`mypy` findings.

tests/frontend/test_prepost.py
```diff
@@ -2076,3 +2076,105 @@ def test_streaming_parallel_tool_calls_no_think(
...
+# Reasoning-parser adjust_request wiring
+# ---------------------------------------------------------------------------
+class _RecordingReasoningParser:
+    """Stand-in for a reasoning parser whose adjust_request mutates the request.
+
+    Real example: a parser that splits thinking from the answer on text markers needs
+    skip_special_tokens=False so those markers survive detokenisation.
+    """
+
+    instances: list["_RecordingReasoningParser"] = []
```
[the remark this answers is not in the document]
yifjiang: Addressed in 9432889 — the shared registry is gone.

`_RecordingReasoningParser.instances` is no longer a class attribute at module scope. The list is now owned by a per-test fixture and closed over by the parser class, which is the pattern suggested here:

```python
@pytest.fixture
def recording_reasoning_parser():
    """Returns ``(cls, instances)``. The list is owned by the fixture rather than
    living on the class, so each test starts clean without depending on a shared
    registry being cleared (see .ai/pytest-guidelines.md, "Hermetic Testing")."""
    instances = []

    class _RecordingReasoningParser:
        def __init__(self, tokenizer, chat_template_kwargs=None):
            ...
            instances.append(self)

    return _RecordingReasoningParser, instances
```

Each test takes the fixture and gets a fresh list, so there is no `instances.clear()` to forget and no order dependence. With the class attribute gone, the `RUF012` finding no longer applies either.

tests/frontend/test_prepost.py
```diff
@@ -2076,3 +2076,105 @@ def test_streaming_parallel_tool_calls_no_think(
...
+
+    def __init__(self, tokenizer, chat_template_kwargs=None):
+        self.tokenizer = tokenizer
+        self.chat_template_kwargs = chat_template_kwargs
+        self.called = False
+        type(self).instances.append(self)
```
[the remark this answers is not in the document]
yifjiang: Addressed in 9432889 — the shared registry is gone.

`_RecordingReasoningParser.instances` is no longer a class attribute at module scope. The list is now owned by a per-test fixture and closed over by the parser class, which is the pattern suggested here:

```python
@pytest.fixture
def recording_reasoning_parser():
    """Returns ``(cls, instances)``. The list is owned by the fixture rather than
    living on the class, so each test starts clean without depending on a shared
    registry being cleared (see .ai/pytest-guidelines.md, "Hermetic Testing")."""
    instances = []

    class _RecordingReasoningParser:
        def __init__(self, tokenizer, chat_template_kwargs=None):
            ...
            instances.append(self)

    return _RecordingReasoningParser, instances
```

Each test takes the fixture and gets a fresh list, so there is no `instances.clear()` to forget and no order dependence. With the class attribute gone, the `RUF012` finding no longer applies either.

components/src/dynamo/frontend/prepost.py
yifjiang: Update: the tradeoff turned out to be avoidable, so there is nothing here to settle. Pushed in 2b89671 — please read this instead of my previous reply.

The problem was that attribution was **inferred** from `tool_parser is None`. Both parsers rewrite the same request inside `_prepare_request` — tool at [`prepost.py:481`](https://github.com/ai-dynamo/dynamo/blob/9432889/components/src/dynamo/frontend/prepost.py#L481), reasoning at [`:537`](https://github.com/ai-dynamo/dynamo/blob/9432889/components/src/dynamo/frontend/prepost.py#L537) — and the code diffed the guidance across *both* calls at once. One diff cannot say who moved it, hence the proxy, hence the choice between regressing the pinned tool policy and leaving the both-parsers gap.

Taking the reading *between* the two calls makes it exact:

```
client            -> after_tool_parser    changed by the tool parser
after_tool_parser -> final                changed by the reasoning parser
```

Each rewrite then carries its own precedence, which is what was wanted:

- **reasoning rewrote it** → wins, whether or not a tool parser is active. Gap closed.
- **only the tool parser rewrote it** → caller's explicit constraint still wins. `test_assistant_guidance_takes_precedence_over_auto_tool_guidance` (both params) and the `response-format-precedence` row are untouched — that path is unchanged.
- **both rewrote it** → nothing to arbitrate. The reasoning parser runs *last, on the already-tool-adjusted request*, so the final value already **is** the composition. Forwarding it is just forwarding what the request holds — the invariant you flagged in the first place.

So the "who wins when both changed it" question was an artifact of collapsing two sequential rewrites into one before/after diff, not a real policy question.

Validated against the previous commit, same container, `prepost.py` swapped underneath:

```
test_vllm_processor_unit.py   before: 11 failed / 117 passed
                              after:  10 failed / 118 passed
regressions: none — the only delta is the new gap test
tests/frontend/test_prepost.py  15 passed both
```

The new `test_rewrite_is_forwarded_with_a_tool_parser_also_active` fails on the previous commit with exactly the symptom you reported — `{'json': {'type': 'object'}}` instead of the structural tag — and `test_tool_parser_rewrite_still_loses_to_the_caller` passes under both, pinning the half that must not move.

Two implementation notes: `_prepare_request` takes an optional write-only `guidance_snapshots` dict rather than a sixth return value, because the return is unpacked at 22 call sites and widening it would have churned all of them; and the snapshot uses `request_for_sampling.structured_outputs`, the same accessor as `adjusted_structured_guidance`, since `extract_structured_outputs()` is a different view and would have made the diff spurious. `reasoning_only_guided_decoding` and its scoping rationale are gone.

components/src/dynamo/frontend/prepost.py
```diff
@@ -495,6 +535,24 @@ def _prepare_request(
...
+    # only works when the *client* happens to send skip_special_tokens=false.
+    #
+    # Gated identically to StreamingPostProcessor's own parser construction below: if
+    # the sampling params were adjusted for a parser that then does not exist, the
+    # markers reach the client as visible text.
+    #
+    # ReasoningParser.adjust_request is `return request` by default, so this is a no-op
+    # for parsers that do not need it.
+    if _reasoning_parser_enabled(reasoning_parser_class, chat_template_kwargs):
+        request_for_sampling = reasoning_parser_class(
```
rmccorm4, line 566: [P1] This constructs the parser without `model_config`. The Cohere parser in the pinned vLLM release gates its `response_format` to `structural_tag` rewrite on `self._model_config.architecture`, so this new `adjust_request()` path is still a no-op for the real parser. I reproduced Dynamo keeping both the JSON schema and `response_format`; passing `model_config` produced the structural tag and cleared `response_format`. Can we thread the processor model config through here and keep parser construction consistent in postprocessing?

components/src/dynamo/frontend/prepost.py
```diff
@@ -457,6 +481,22 @@ def _prepare_request(
...
             request_for_sampling = tool_parser.adjust_request(request_for_sampling)
 
+    # Both parsers rewrite the same request in place, tool first and reasoning
+    # below, so a single before/after diff cannot say which one moved the
+    # guidance. Reading it here splits that into two attributable diffs and is the
+    # only reason the caller can apply one precedence rule to a reasoning rewrite
+    # and a different one to a tool rewrite. Optional and write-only, so the
+    # existing call sites are unaffected.
+    if guidance_snapshots is not None:
+        guidance_snapshots["after_tool_parser"] = deepcopy(
```
rmccorm4: [P2] We now deep-copy the full structured output on every chat request because `preprocess_chat_request()` always passes a snapshots dict, even when no reasoning parser runs. With a 1.49 MB schema in the pinned vLLM release, this added 34.646 ms median locally (0.003 ms to 34.648 ms). Can we only request this snapshot when the reasoning parser is enabled? The tool-only path does not need the intermediate copy.

components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -542,6 +542,7 @@ async def _generator_inner(
                 tokenizer=self.tokenizer,
                 renderer=self.input_processor.renderer,
                 tool_parser_class=self.tool_parser_class,
+                reasoning_parser_class=self.reasoning_parser_class,
```
rmccorm4, line 554: [P2] Please add one test through `VllmProcessor._generator_inner()` that proves this argument reaches `preprocess_chat_request()`. I removed this line locally and both changed test files still passed 143/143, so the load-bearing production wiring can regress without a failure. One of the overlapping helper-level pass-through or no-parser cases could be replaced with this boundary check.

components/src/dynamo/frontend/prepost.py
yifjiang: Confirmed and fixed in `9692a98`.

Verified against real vLLM before changing anything. `ReasoningParser.__init__` reads `model_config` out of `**kwargs`, and `BaseCohereCommandReasoningParser.adjust_request` has `if model_architecture is None: return request` — so the construction here landed in that branch every time. Three arms with the real `CohereCommand3ReasoningParser` and a real `ChatCompletionRequest`:

| construction | `_model_config` | `response_format` after | `structured_outputs` after |
|---|---|---|---|
| `Cls(tok, chat_template_kwargs={})` — what this PR shipped | None | still the json_schema | None |
| `Cls(tok, chat_template_kwargs={}, model_config=mc)` | set | None | structural_tag |
| `Cls(tok, mc, chat_template_kwargs={})` — positional | None | still the json_schema | None |

Arm 1 reproduces exactly what you saw. Arm 3 is worth knowing: a positional second argument lands in `*args` and is silently dropped, so it has to be the keyword.

No plumbing needed — `InputProcessor` already sets `self.model_config` from `vllm_config`, so `self.input_processor.model_config` was reachable from `_generator_inner` as-is. Threaded to all three construction sites: `_prepare_request`, `StreamingPostProcessor`, and `_build_reasoning_parser_metadata`. That third one does not read it, but constructing one of the three differently is what let this hide, which I take to be your "keep parser construction consistent" point.

Deliberately **not** added to `parser_kwargs` — that dict goes on the wire as `dynamo_preproc["reasoning_parser_kwargs"]` and `ModelConfig` is not serializable; the worker builds its own.

Also added `_ModelConfigGatedReasoningParser`, a fake that mirrors Cohere's actual contract (no `model_config` ⇒ no rewrite). The existing fakes ignored `model_config` entirely, which is the reason a no-op passed them. Removing the kwarg now fails `test_model_config_reaches_the_parser`.

components/src/dynamo/frontend/prepost.py
yifjiang: Fixed in `9692a98`, and it was worse than the numbers suggested.

You are right that the only reader is behind the same `_reasoning_parser_enabled()` test and short-circuits, so the snapshot is needed if and only if a reasoning parser actually runs. Moved it inside that gate.

I also dropped the `deepcopy` rather than just gating it. `adjust_request()` **replaces** `request.structured_outputs` with a fresh object rather than mutating the one the snapshot names, so a plain reference still denotes the pre-adjustment value afterwards — the copy was not load-bearing against any parser that exists.

That turned out to matter twice. Measured here on a 1.49 MB schema (162k nodes), median of 60:

| | median |
|---|---|
| copy, as shipped | 56.081 ms |
| no copy | 0.0001 ms |
| the `!=` it feeds, deepcopied | 0.7155 ms |
| the `!=` it feeds, aliased | 0.0001 ms |

The second pair is the non-obvious one: `dict.__eq__` short-circuits on identity, so aliasing makes the unchanged case O(1) while the copy forced a full structural walk on the comparison as well. Tool-only and no-parser requests now pay neither.

One thing worth flagging separately: `client_structured_guidance` one line up pays the same ~56 ms on the same request, and it is not from this PR — it came in with #12684. The same "parsers replace, never mutate" argument likely applies, but its blast radius spans both parsers so I left it alone rather than widening this PR. Happy to file it.

components/src/dynamo/frontend/vllm_processor.py
yifjiang: You are right, and I reproduced it before fixing: copied the worktree, deleted that line, ran both changed files — **143 passed**, zero failures. Exactly your count.

The reason is structural: `_generator_inner` is called from only two places in the test tree and both replace `preprocess_chat_request` with a mock, so nothing ever reads the real call's kwargs. Every test I added calls `preprocess_chat_request` / `_prepare_request` directly and passes `reasoning_parser_class` itself, so the production wiring was untested by construction.

Added `test_generator_inner_forwards_reasoning_parser_and_model_config` in `9692a98`. It drives the real `_generator_inner` with a spy that captures kwargs and raises to stop the generator at the boundary, then asserts identity against a sentinel class. It checks `model_config` alongside, since after the fix above that line is equally load-bearing and equally invisible.

Verified both directions — deleting either wiring line fails only this test:

```
assert None is <class ..._SentinelReasoningParser>
assert None is namespace(architecture='SentinelForCausalLM')
```

Traded out `test_no_reasoning_parser_keeps_the_caller_constraint` as you suggested — it overlapped the helper-level no-parser case that was already covered.

components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
```diff
@@ -1709,6 +1777,189 @@ async def test_root_thinking_suppresses_deployment_default(self, tokenizer):
...
+    from a no-op, which is how the missing kwarg went unnoticed.
+    """
+
+    def adjust_request(self, request):
+        if self._model_config is None:
+            return request
+        return super().adjust_request(request)
+
+
+class TestReasoningParserGuidanceForwarding:
```
rmccorm4, line 1820: Could we trim this suite to the cases that prove distinct behavior? The PR adds 13 tests and 374 test lines for this path; the pass-through, no-parser, explicit-true, and negative no-model-config controls overlap the processor-boundary, thinking-disabled, conversion, and mixed-parser precedence cases. Keeping the failure-distinct cases would make the regression much easier to maintain.

rmccorm4 (approved):

components/src/dynamo/frontend/tests/test_vllm_processor_unit.py
rmccorm4: not blocking, can be a follow up

components/src/dynamo/frontend/prepost.py
```diff
@@ -610,11 +697,45 @@ async def preprocess_chat_request(
             "tool_choice forces a tool call and cannot be combined with an "
             "explicit guided_* or structured_outputs constraint."
         )
-    guided_decoding = (
-        assistant_guided_decoding
-        if assistant_guided_decoding is not None and not is_forced_tool_choice
-        else tool_guided_decoding
+    # Which parser moved the guidance, measured rather than inferred. Both run
```
rmccorm4, line 700: @yifjiang please adjust your workflow to generally produce fewer comments and much less verbose comments for future PRs - many of these comments are overkill

PR #12542, opened 2026-08-02
feat(snapshot): add ns-bind-mount C helper and Go mount package

oleg-kushniriov:
## Overview                                                                                                                            
                                                                                                                                         
  **Add foundational packages for cross-namespace bind mounting**, required to stage the agent binary bundle inside a placeholder        
  container's mount namespace before CRIU restore runs.                                                                                  
                                                                                                                                         
  Go's multithreaded runtime prohibits `setns(CLONE_NEWNS)`, so the implementation is split across a C helper binary and two thin Go     
  packages. This PR is additive only — no existing files are modified and the new packages are not yet wired into the restore path.      
                                                                                                                                         
  ## Details                                                                                                                             
                                                                                                                                         
  Three layers are introduced:                                                                                                           
                                                                                                                                         
  - **`cmd/ns-bind-mount/main.c`** — C helper that performs bind mount (`open_tree(OPEN_TREE_CLONE)` + `move_mount`) and lazy unmount    
  (`umount2(MNT_DETACH)`) across mount namespaces. Requires Linux 5.2+.                                                                  
  - **`internal/nsbindmount/nsbindmount.go`** — Go `Mounter` interface backed by `ExecMounter`, which shells out to the C helper and     
  returns a cleanup func.                                                                                                                
  - **`internal/injection/injector.go` + `paths.go`** — `Injector` struct that uses `Mounter` to bind-mount the agent binary bundle      
  read-only into a target namespace given a PID.                                                                                         
  - **`internal/injection/injector_test.go`** — unit tests via `mockMounter`; no cluster or Linux privileges required.                   
                                                                                                                                         
  The Dockerfile build stage for the C binary and the wiring of `Injector` into `executor/restore.go` are intentionally deferred to a    
  follow-up PR.                                                                                                                          
                                                                                                                                         
  ### 1. `cmd/ns-bind-mount/main.c`                                                                                                      
  Most security-sensitive code. Verify `open_tree`/`move_mount` ordering, `enter_mnt_ns`, and error paths (fd cleanup on failure, `rmdir`
  on `move_mount` failure).                                                                                                              
                                                                                                                                         
  ### 2. `internal/nsbindmount/nsbindmount.go`                                                                                           
  Check the `Mounter` interface contract and cleanup closure (`exec.Command` without context is intentional — cleanup must complete even 
  after the parent context is cancelled).                                                                                                
                                                                                                                                         
  ### 3. `internal/injection/injector.go`                                                                                                
  Straightforward once the above two are understood.                                                                                     
                                                                                                                                         
  ## Related Issues                                                                                                                      
                                                                                                                                         
  **SOLVES issue #12579**                                                                                               

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added support for mounting agent binaries into a target process’s mount namespace.
  * Added read-only mount options and reliable cleanup, including lazy unmounting.
  * Added validation and configurable source and destination paths for injected binaries.
  * Added clear error reporting and logging for mount and namespace operations.

* **Tests**
  * Added coverage for configuration defaults, validation, mounting, cleanup, and failure handling.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


galletas1712:
Generally looks good to me! Just a small nit:

Could we consolidate the Go code under something like `internal/nsmountinjector`?

I would still keep the C helper as cmd/ns-bind-mount, since it is a generic namespace bind-mount utility rather than an asset-specific injector. This may avoid introducing internal/nsbindmount as a one-consumer abstraction.

In general, we just want to consolidate the go code so we don't introduce too many one-off files/paths. For example, maybe paths.go doesn't need to exist separately?

deploy/snapshot/cmd/ns-bind-mount/main.c
```diff
@@ -0,0 +1,275 @@
...
+    }
+    if (!S_ISDIR(st.st_mode)) {
+      fprintf(stderr, "dst %s exists but is not a plain directory\n", dst);
+      close(tree_fd);
+      return 1;
+    }
+  }
+
+  /* Move the cloned mount into the target namespace at dst. */
+  if (sys_move_mount(tree_fd, "", AT_FDCWD, dst, MOVE_MOUNT_F_EMPTY_PATH) < 0) {
```
Ronkahn21, line 426: **[High] The bundle is exposed writable before the read-only remount.** `move_mount()` attaches the detached bind mount here with its original writable attributes; `MS_RDONLY` is applied only in the later `mount()` call. A root process in the target workload can spin on `dst` and write, truncate, or replace bundle files during that interval, modifying the underlying source and potentially poisoning a later privileged `nsrestore` execution. Please apply `MOUNT_ATTR_RDONLY | MOUNT_ATTR_NOSUID | MOUNT_ATTR_NODEV` to `tree_fd` with `mount_setattr()` *before* this call, or otherwise guarantee and verify that the source mount is intrinsically read-only before attachment. This may require updating the stated minimum kernel version.

deploy/snapshot/cmd/ns-bind-mount/main.c
```diff
@@ -0,0 +1,275 @@
...
+
+  /* Clone the source mount tree before entering the target namespace. */
+  int tree_fd = sys_open_tree(AT_FDCWD, src, OPEN_TREE_CLONE | O_CLOEXEC);
+  if (tree_fd < 0) {
+    fprintf(stderr, "open_tree %s: %s\n", src, strerror(errno));
+    return 1;
+  }
+
+  /* Enter the target process's mount namespace. */
+  if (enter_mnt_ns(pid) < 0) {
```
Ronkahn21, line 414: **[Medium] The target filesystem root is not pinned.** `setns(CLONE_NEWNS)` changes mount-namespace membership but does not change the helper's filesystem root or working directory. The subsequent absolute `mkdir`, `move_mount`, remount, and unmount paths therefore remain rooted through the helper's original filesystem root; on a normal container layout this can act on the wrong filesystem or fail because the target path is not in the entered namespace. Please open and pin `/proc/<pid>/root` together with the namespace fd, then use `fchdir`/`chroot` or fd-relative target resolution.

deploy/snapshot/cmd/ns-bind-mount/main.c
```diff
@@ -0,0 +1,275 @@
...
+/* Returns 0 if path is absolute and begins with allowed_prefix, -1 otherwise. */
+static int
+check_path_prefix(const char* path, const char* allowed_prefix, const char* label)
+{
+  if (path[0] != '/') {
+    fprintf(stderr, "%s must be an absolute path: %s\n", label, path);
+    return -1;
+  }
+  size_t plen = strlen(allowed_prefix);
+  if (strncmp(path, allowed_prefix, plen) != 0 || (path[plen] != '\0' && path[plen] != '/')) {
```
Ronkahn21, line 85: **[Medium] This string-prefix allowlist is bypassable by pathname resolution.** For example, `/snapshot-binaries/../etc` passes this check but resolves to `/etc`; `/tmp/snapshot-binaries/../../etc` similarly escapes the destination prefix. Intermediate symlinks are also not constrained, and the Go `Config.Validate` layer currently checks only for empty values despite the comment above saying both layers enforce these prefixes. Please use canonical, race-resistant fd-based resolution—preferably `openat2()` with `RESOLVE_BENEATH`/`RESOLVE_IN_ROOT` and `RESOLVE_NO_MAGICLINKS`—rather than treating a textual prefix as the security boundary.

deploy/snapshot/internal/nsbindmount/nsbindmount.go
```diff
@@ -0,0 +1,155 @@
...
+	if opts.ReadOnly {
+		args = append(args, "ro")
+	}
+	out, err := exec.CommandContext(ctx, m.binaryPath, args...).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("ns-bind-mount %s -> %s: %w\noutput: %s", src, dst, err, strings.TrimSpace(string(out)))
+	}
+	m.log.Info("mounted into namespace", "src", src, "dst", dst, "readonly", opts.ReadOnly, "pid", pid)
+
+	nsFd, err := os.Open(fmt.Sprintf("/proc/%d/ns/mnt", pid))
```
Ronkahn21: **[Low] The namespace identity is captured only after the mount helper returns.** If the target exits and its host PID is reused in this interval, the mount can occur in namespace A while this handle pins namespace B; cleanup can then affect B and leave A's mount behind. Please open the namespace fd before launching the helper, pass that same fd through `ExtraFiles` for the mount operation, and retain the parent copy for cleanup.

deploy/snapshot/cmd/ns-bind-mount/main.c
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  if (umount2(dst, MNT_DETACH) < 0) {
+    if (errno != ENOENT && errno != EINVAL) {
+      fprintf(stderr, "umount2 %s: %s\n", dst, strerror(errno));
+      return 1;
+    }
+  }
+
+  rmdir(dst);
```
Ronkahn21: **[Low] Cleanup removes `dst` even when this helper did not create it.** This unconditional `rmdir()` also runs after `umount2()` returned `EINVAL`. A pre-existing empty destination can therefore be deleted, and a namespace-identity mismatch can make that deletion target an unrelated workload. Track whether the directory was created by this mount operation and propagate that ownership into the cleanup handle, or leave pre-existing targets untouched.

deploy/snapshot/internal/injection/injector.go
```diff
@@ -0,0 +1,108 @@
...
+	return &injectionHandle{mount: handle}, nil
+}
+
+// injectionHandle is the concrete InjectionHandle returned by Inject.
+type injectionHandle struct {
+	mount nsbindmount.MountHandle
+}
+
+func (h *injectionHandle) BinPath(name string) string {
+	return filepath.Join(h.mount.TargetPath(), name)
```
Ronkahn21: **[Low] `BinPath` permits traversal outside the injected directory.** `filepath.Join` cleans `..`, so a name such as `../../bin/tool` escapes `TargetPath()`. If this argument ever becomes nonconstant, it can redirect the later privileged execution to a workload-controlled path. Restrict `name` to a single basename with no separators or dot components, or expose a fixed `NSRestorePath()` instead.

Ronkahn21 (commented):
Security re-review of head `76598765`: the update improves defense in depth (`0700` destination mode, final-component type checking, path-prefix checks, and an explicit child-fd constant), but it does not resolve the central writable-attachment race or the target root/namespace pinning problems. The new string-prefix validation is also bypassable through pathname resolution. Recommendation remains **REQUEST CHANGES before this helper is wired into the privileged restore path**.

deploy/snapshot/cmd/ns-bind-mount/main.c
oleg-kushniriov: PID-reuse is fully eliminated — the Go side opens` /proc/<pid>/ns/mnt` before launching the helper and passes it via ExtraFiles. We didn't add `fchdir/chroot` because after `setns(CLONE_NEWNS)` the kernel resolves absolute paths through the target namespace's own mount table, which is the correct behavior for cross-namespace bind mounting. Let me know if you see a case where this assumption breaks.

galletas1712 (approved):
Let's have @Ronkahn21 approve this before merging

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package nsmountinjector stages the agent's restore binaries inside a
+// placeholder container's mount namespace before CRIU restore can run.
+//
+// The agent's binary bundle is bind-mounted read-only from agentBinDir into
+// SnapshotBinDir inside the target namespace, via the ns-bind-mount C helper
+// (cmd/ns-bind-mount). The placeholder image ships none of these binaries.
+package nsmountinjector
```
Ronkahn21: can we change the name of package to just nsmount

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// hasAllowedPrefix reports whether path equals prefix or begins with prefix+"/".
+func hasAllowedPrefix(path, prefix string) bool {
+	if !strings.HasPrefix(path, prefix) {
+		return false
+	}
+	rest := path[len(prefix):]
+	return rest == "" || rest[0] == '/'
+}
```
Ronkahn21:
```suggestion

func hasAllowedPrefix(path, prefix string) bool {
    rest, found := strings.CutPrefix(path, prefix)
    return found && (rest == "" || rest[0] == '/')
}
```

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+		return errors.New("nsmountinjector.Config: DestinationDir must not be empty")
+	}
+	if !hasAllowedPrefix(c.SourceDir, agentBinDir) {
+		return fmt.Errorf("nsmountinjector.Config: SourceDir must start with %s: %s", agentBinDir, c.SourceDir)
+	}
+	if !hasAllowedPrefix(c.DestinationDir, SnapshotBinDir) {
+		return fmt.Errorf("nsmountinjector.Config: DestinationDir must start with %s: %s", SnapshotBinDir, c.DestinationDir)
+	}
+	return nil
+}
```
Ronkahn21: this config should be siting in the agent config and not here
this is not a standalone package, the packge should just get a simple parameters to use

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+	// Cleanup unmounts the injected directory from the target namespace.
+	// Idempotent — safe to call from a defer even if Inject partially failed.
+	Cleanup(ctx context.Context) error
+}
+
+// Injector mounts the agent's restore binaries into a placeholder container's
+// mount namespace and returns a handle for cleanup.
+type Injector interface {
+	Inject(ctx context.Context, pid int) (Handle, error)
+}
```
Ronkahn21: can we just rename inject to mount

Ronkahn21: as go best practice interface should sit at the caller side not here

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+// WithDefaults returns a copy of c with empty fields filled from package defaults.
+func (c Config) WithDefaults() Config {
+	if c.SourceDir == "" {
+		c.SourceDir = agentBinDir
+	}
+	if c.DestinationDir == "" {
+		c.DestinationDir = SnapshotBinDir
+	}
+	return c
+}
```
Ronkahn21: all the config should sit as part of the agent config not here

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+	cfg = cfg.WithDefaults()
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
+	return &NSMountInjector{cfg: cfg, mounter: m, log: log}, nil
+}
+
+// Inject bind-mounts cfg.SourceDir into cfg.DestinationDir inside the mount
+// namespace of pid. The caller must call Handle.Cleanup when done.
+func (i *NSMountInjector) Inject(ctx context.Context, pid int) (Handle, error) {
```
Ronkahn21:
```suggestion
func (nsm *NSMountInjector) Inject(ctx context.Context, pid int) (Handle, error) {
```

deploy/snapshot/internal/nsmountinjector/injector.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	i.log.Info("agent bundle mounted", "pid", pid, "dst", handle.TargetPath())
+	return &injectionHandle{mount: handle}, nil
+}
+
+// injectionHandle adapts a MountHandle to the bundle-oriented Handle surface.
+type injectionHandle struct {
```
Ronkahn21: I am not sure I understand the doc, or this strct

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,160 @@
...
+	// TargetPath returns the dst path as seen inside the target namespace.
+	TargetPath() string
+}
+
+// mounter mounts src at dst inside the mount namespace identified by pid.
+// It exists so tests can substitute the ns-bind-mount subprocess; production
+// callers get an execMounter via New and never name this type.
+type mounter interface {
+	Mount(ctx context.Context, pid int, src, dst string, opts MountOptions) (MountHandle, error)
+}
```
Ronkahn21, line 57: why Unmount is not part of this interface ?

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+// MountOptions configures a single namespace-aware mount operation.
+type MountOptions struct {
+	ReadOnly bool // mount with MS_RDONLY
+}
+
+// MountHandle represents an active mount inside a foreign namespace.
+// The owner must call Unmount when the mount is no longer needed.
+type MountHandle interface {
```
Ronkahn21: the name Mount handle is not clear to me

deploy/snapshot/internal/nsmountinjector/mount.go
```diff
@@ -0,0 +1,160 @@
...
+	return &execMounter{binaryPath: path, log: log}, nil
+}
+
+// mountHandle is the concrete MountHandle returned by execMounter.Mount.
+// It holds an open /proc/<pid>/ns/mnt fd captured before the mount helper runs
+// so that Unmount can re-enter the correct namespace even after the target
+// process has exited and its PID been recycled by the kernel.
+type mountHandle struct {
+	binaryPath string
+	nsFd       *os.File // /proc/<pid>/ns/mnt opened before Mount runs
```
Ronkahn21: the doc is misleading, I would just keep the stcts handle

deploy/snapshot/internal/nsmount/mount.go
oleg-kushniriov: MountHandle follows the Go `<Noun>Handle` convention for types representing a live resource that requires cleanup (cf. sql.Tx, os.File). Open to ActiveMount or NSMount as alternatives

deploy/snapshot/internal/nsmount/mount.go
oleg-kushniriov: Unmount is intentionally absent from `mounter` because the `mounter` is stateless - its only job is to create a mount. Cleanup state (the pinned namespace fd, whether the helper created dst) is captured inside the `MountHandle` returned by Mount. 
Putting Unmount on mounter would force the caller to pass all of that state back in externally, which the handle already encapsulates. The handle pattern keeps each concern in one place.

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+	// agentBinDir is the host-side directory in the agent image holding the
+	// binaries to inject.
+	agentBinDir = "/snapshot-binaries"
+
+	// SnapshotBinDir is where that bundle appears inside the placeholder
+	// namespace. This is the bundle-layout contract: nsrestore resolves criu,
+	// the criu plugin directory, and the cuda-checkpoint helpers underneath it,
+	// and deploy/snapshot/Dockerfile builds the bundle to match.
+	SnapshotBinDir = "/tmp" + agentBinDir
+)
```
Ronkahn21: this package provide functionality only so we should remove any opinionated consts, it should sit in the business logic consts

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+
+// hasAllowedPrefix reports whether path equals prefix or begins with prefix+"/".
+func hasAllowedPrefix(path, prefix string) bool {
+	rest, found := strings.CutPrefix(path, prefix)
+	return found && (rest == "" || rest[0] == '/')
+}
+
+// Handle represents a live binary mount into a placeholder namespace.
+// The caller must call Cleanup after nsrestore returns.
+type Handle interface {
```
Ronkahn21: the name handler does not explain to me anything, is very generic
change the name to MoundPoint

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+	// BinPath returns the in-namespace absolute path to the named binary.
+	// name must be a single path element with no separators or dot-dot components.
+	// Example: handle.BinPath("nsrestore") → "/tmp/snapshot-binaries/nsrestore", nil
+	BinPath(name string) (string, error)
```
Ronkahn21:
```suggestion
	Path(name string) (string, error)
```

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Cleanup unmounts the mounted directory from the target namespace.
+	// Idempotent — safe to call from a defer even if Mount partially failed.
+	Cleanup(ctx context.Context) error
```
Ronkahn21:
```suggestion
	Unmount(ctx context.Context) error
```

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+// by the C helper. It errors if the paths are invalid or the helper binary is
+// missing, so a misconfigured node fails at startup rather than at the first restore.
+// Requires Linux 5.12+ (mount_setattr; open_tree/move_mount need only 5.2).
+func New(src, dst string, log logr.Logger) (*NSMounter, error) {
+	if !hasAllowedPrefix(src, agentBinDir) {
+		return nil, fmt.Errorf("nsmount: src must start with %s: %s", agentBinDir, src)
+	}
+	if !hasAllowedPrefix(dst, SnapshotBinDir) {
+		return nil, fmt.Errorf("nsmount: dst must start with %s: %s", SnapshotBinDir, dst)
+	}
```
Ronkahn21: why do your care if its subset of certain opinited consts
you provide functionality, the buisense logic would need to validate this

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+type bundleHandle struct {
+	mount MountHandle
+}
+
+// BinPath returns the in-namespace absolute path to the named binary.
+// name must be a single path element: callers pass literals, and anything
+// else could redirect a privileged exec outside the mounted bundle.
+func (h *bundleHandle) BinPath(name string) (string, error) {
+	if name == "" || name == "." || name == ".." || strings.ContainsRune(name, os.PathSeparator) {
+		return "", fmt.Errorf("nsmount: invalid binary name %q", name)
```
Ronkahn21, line 86: check if there is build-in logic for this
if not export this to seprate functionlity

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,161 @@
...
+		cmd.ExtraFiles = []*os.File{h.nsFd}
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			h.log.Error(err, "failed to unmount from namespace", "dst", h.dst, "output", strings.TrimSpace(string(out)))
+			h.unmountErr = fmt.Errorf("ns-bind-mount umount-fd %s: %w\noutput: %s", h.dst, err, strings.TrimSpace(string(out)))
+			return
+		}
+		h.log.Info("unmounted from namespace", "dst", h.dst)
+	})
+	return h.unmountErr
```
Ronkahn21, line 115: make sure that unmount will lead to restart continner so we will never leaving the mount open

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,161 @@
...
+		args := []string{"umount-fd", strconv.Itoa(nsFdChildNum), h.dst}
+		if h.createdDst {
+			args = append(args, "created")
+		}
+		cmd := exec.CommandContext(ctx, h.binaryPath, args...)
+		cmd.ExtraFiles = []*os.File{h.nsFd}
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			h.log.Error(err, "failed to unmount from namespace", "dst", h.dst, "output", strings.TrimSpace(string(out)))
+			h.unmountErr = fmt.Errorf("ns-bind-mount umount-fd %s: %w\noutput: %s", h.dst, err, strings.TrimSpace(string(out)))
```
Ronkahn21, line 110: no need to return and log

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Mount bind-mounts src (in the current namespace) to dst inside the mount
+// namespace of pid. It opens /proc/<pid>/ns/mnt *before* launching the helper
+// so the namespace is pinned against PID reuse, then passes the fd to the
+// mount-fd subcommand via ExtraFiles. The fd is retained in the returned handle
+// so Unmount can re-enter the namespace without relying on the PID.
+func (m *execMounter) Mount(ctx context.Context, pid int, src, dst string, opts MountOptions) (MountHandle, error) {
+	// Pin the namespace fd before calling the helper so mount and cleanup
+	// provably act on the same namespace regardless of PID reuse.
+	nsFd, err := os.Open(fmt.Sprintf("/proc/%d/ns/mnt", pid))
```
Ronkahn21: const

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	nsm.log.Info("agent bundle mounted", "pid", pid, "dst", handle.TargetPath())
+	return &bundleHandle{mount: handle}, nil
+}
+
+// bundleHandle wraps a MountHandle to expose the bundle-oriented Handle
+// surface: BinPath resolves binary names relative to the mounted directory,
+// and Cleanup delegates to the underlying mount's Unmount.
+type bundleHandle struct {
```
Ronkahn21: again here, Handle is not a good name

deploy/snapshot/internal/nsmount/injector.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return newWithMounter(src, dst, m, log)
+}
+
+// newWithMounter is the test seam: it takes an arbitrary mounter so tests can
+// exercise Mount without the ns-bind-mount subprocess.
+func newWithMounter(src, dst string, m mounter, log logr.Logger) (*NSMounter, error) {
+	return &NSMounter{src: src, dst: dst, mounter: m, log: log}, nil
+}
+
```
Ronkahn21, line 59: redundant

deploy/snapshot/internal/nsmount/mount.go
```diff
@@ -0,0 +1,162 @@
...
+		defer func() {
+			if err := h.nsFd.Close(); err != nil {
+				h.log.Error(err, "close ns fd", "dst", h.dst)
+			}
+		}()
+		// Fresh context with a hard timeout. The parent context is intentionally
+		// not forwarded: cleanup must complete even if the caller's context is
+		// already cancelled.
+		ctx, cancel := context.WithTimeout(context.Background(), unmountTimeout)
+		defer cancel()
```
Ronkahn21, line 99: `sync.Once` invalidates this mount reference after the first unmount attempt, even when that attempt fails. Ignoring subsequent calls is correct only after a successful unmount. If `umount-fd` fails, keep the namespace FD open and allow another `Unmount` attempt; mark the reference unmounted and close the FD only after success.

Ronkahn21 (commented):
One concern with the unmount lifecycle:

deploy/snapshot/internal/nsmount/injector.go
oleg-kushniriov: It will be used in next PR

deploy/snapshot/internal/nsmount/mount.go
oleg-kushniriov: retry doesn't change the outcome. `umount2(MNT_DETACH)` has exactly three results: 
it succeeds
it returns ENOENT/EINVAL (already treated as success by the helper)
it returns something else - and whatever kernel condition caused that error will still be present on the second call. There is no transient failure mode where running the same command a second time produces a different result.

Beyond the logic, the practical failure rate of umount2(MNT_DETACH) is essentially zero. It is a lazy detach that succeeds even when the path is busy; the only realistic errno is EPERM on a misconfigured node, which a retry won't fix.

docs/fern/pages/community/contributing/documentation/building-and-publishing.md
```diff
@@ -360,7 +360,7 @@ workflows. Authors never need to run it manually.
 ## Running Locally
 
 You can preview the documentation site on your machine using the
-[Fern CLI](https://buildwithfern.com/learn/docs/getting-started/overview/). This is useful
```
Ronkahn21: revert

Ronkahn21 (approved):

PR #12543, opened 2026-08-02
feat(request-trace): opt-in media redaction for request_payload records

yifjiang:
> **Context: one of six fixes from bringing up [`thinkingmachines/inkling`](https://huggingface.co/thinkingmachines/inkling) on Dynamo + vLLM** (4×GB200, TP=4, arm64). Sibling PRs from the same bring-up: #12507, #12510, #12541, #12511. Found while auditing payload traces on a multimodal deployment that exports to OTLP. The leak is general — it affects any deployment with `DYN_REQUEST_TRACE_RECORDS=request_payload` serving image or video input. Pairs with #12511, which emits these records on the BYO chat-processor path.

## Problem

`request_payload` records carry the full inbound request. For multimodal requests that includes base64 data URIs and media URLs, and they are written **verbatim** to every configured sink — file, stderr, and OTLP. On a deployment exporting to an OTLP collector, raw image/video bytes leave the process.

Observed on a multimodal deployment: a 1×1 PNG sent as an `image_url` part appears unredacted in the record.

```json
{"type":"image_url","image_url":{"url":"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA…"}}
```

## Fix

Add an **opt-in** redaction pass, `DYN_REQUEST_TRACE_REDACT_MEDIA` (**off by default**), that replaces media parts with a text placeholder **before the pristine snapshot is taken**:

```
{"type":"image_url","image_url":{...}}  ->  {"type":"text","text":"[image_url omitted by audit]"}
```

### Configuration

| | |
|---|---|
| env var | `DYN_REQUEST_TRACE_REDACT_MEDIA` |
| default | **off** — records carry inbound media verbatim, as they do today |
| when set truthy | media parts become `[<kind> omitted by audit]` before any sink sees them |

Off by default so that enabling request-trace payload records does not silently change
what those records contain for deployments already consuming them, and so multimodal
debugging / replay workflows that need the media keep working. The resolved value is
logged on the existing `Request trace initialized` line next to `records` and `sinks`,
so which mode a deployment is in is readable straight from its startup logs.

⚠️ Worth stating plainly for reviewers: with the default, a deployment exporting
`request_payload` records to an OTLP collector **does** ship raw base64 media off-box.
That is the pre-existing behaviour, not a regression introduced here — this PR adds the
means to stop it, and leaves the choice to the operator.

Placed inside `create_handle_with_config` rather than at the call sites, which means:

- **both producers are covered by construction** — the Rust preprocessor (`preprocessor.rs`) and the chat-processor wrapper;
- **every sink is covered**, because they all serialise the same snapshot;
- no sink can observe media even transiently.

Only the trace snapshot is rewritten — the request forwarded to the engine is untouched.

### Tool messages are scanned, not just user messages

This is the non-obvious part. `ChatCompletionRequestToolMessageContentPart` carries the same `ImageUrl`/`VideoUrl`/`AudioUrl` variants as the user enum, and `preprocessor.rs` converts tool media parts to user parts for multimodal processing. **Scanning only user messages would leak media supplied in tool results.** One of the tests fails specifically on that case.

### Cost

Pure-text requests pay a **scan, not a clone** — the request is cloned only when at least one media part is present. That matches the previous behaviour, which cloned unconditionally, so this is not a regression for text and is one clone (not two) for media.

The match arms are exhaustive for the enums as defined today; the trailing catch-all is kept under `#[allow(unreachable_patterns)]` as a fail-closed net, so a newly added media variant is redacted as `unknown` rather than leaked.

## Test

Eleven unit tests, covering the flag in **both states at both layers**.

**Flag ON/OFF at the snapshot hook** (`create_handle_with_config`, `payload.rs`):

| flag | request | expected |
|---|---|---|
| on | has media | redacted to placeholders |
| off | has media | **preserved verbatim** |
| on | text-only | **byte-identical** to the inbound request |

That last row guards the `redact_media` → `None` fallback: a regression there would
corrupt or drop content on effectively *all* traffic, not just multimodal.

**Flag ON/OFF at config load** (`load_from_env`, `config.rs`):

| env | expected |
|---|---|
| unset | off |
| truthy (`1`) | on |
| falsey (`0`, `false`, `off`, `no`, empty) | **off** |

The falsey row pins that this reads through `lib/truthy` rather than checking presence —
under an `env::var(..).is_ok()` check, writing `DYN_REQUEST_TRACE_REDACT_MEDIA=0` would
silently mean ON. (The full accepted-spelling matrix is owned by `lib/truthy`, which also
carries a workspace test forbidding hand-rolled bool parsers.)

**Redaction helper itself:**

- pure-text (plain string) → `None`, no clone
- text-only parts array → `None`
- user `image_url` → replaced; **text sibling survives**; **original request not mutated**
- **tool-message media → replaced** (fails if only user messages were scanned)
- `video_url` / `audio_url` → distinct `[video_url …]` / `[audio …]` placeholder kinds
- **`/v1/responses` `input_image` → replaced** (see below)

### Responses API coverage

`/v1/responses` carries media as `{"type": "input_image", "image_url": "…"}` — a shape `redact_media` never sees directly. It **is** covered, but only *by construction*: the handler converts `NvCreateResponse` → `UnifiedRequest` (`unified.rs:193`) → `NvCreateChatCompletionRequest` before the preprocessor takes the snapshot, and `responses/mod.rs:278` maps `InputContent::InputImage` onto `ChatCompletionRequestUserMessageContentPart::ImageUrl`.

Nothing pinned that. Every other test here starts from an already chat-shaped request, so a future direct Responses path — or a change to that conversion — could bypass redaction with all of them still passing. The sixth test drives a real `/v1/responses` payload through the conversion to hold that composition in place.

(`InputVideo` / `InputAudio` are not supported on the Responses path at all today — there is an explicit TODO in `responses/mod.rs`, and `InputFile` errors.)

✅ **Rebased onto `main` (`a97a2b8`) and CI is green** — 21 checks pass, 0 fail; `rust-tests (.)` passes in 29m55s with all six `request_trace::payload::tests::redact_media_*` cases green.

The earlier CI red was **not** a test failure: `cargo fmt --check` objected to a single stray blank line before the closing brace of `mod tests`. Worth flagging that the Datadog comment on this PR reports it as *"expected string to contain the video URL placeholder but it contained sensitive information"* — that summary is inaccurate. It was generated from source lines that merely appeared in the fmt diff; no assertion failed and nothing leaked.

## Relationship to earlier drafts

This supersedes the redaction half of #10415, which was written against the old `lib/llm/src/audit/` module before it became `request_trace`, and applied at `AuditHandle::set_request`. Two substantive differences beyond the relocation:

1. #10415 scans only user messages, with the rationale that tool contents are "text/refusal/tool-call shapes only" — that is not true of the current enum, hence the tool coverage here.
2. #10415 matches an `InputAudio` variant that does not exist at this commit; it would not compile as-is.

Happy to close #10415/#10416 in favour of this, or rebase them, whichever maintainers prefer.

## Notes

Filed as a **draft** — happy to attach a Linear ID and rename the branch to the `<login>/<desc>__<DYN-ID>` convention before review.

The placeholder string `[<kind> omitted by audit]` is deliberately kept verbatim from the earlier draft: downstream log filters and verification probes match on `omitted by audit`.

## Summary by CodeRabbit

* **New Features**
  * Added media redaction for request traces, replacing images, audio, and video with descriptive placeholders.
  * Redaction applies to both user and tool messages before request snapshots are stored.

* **Bug Fixes**
  * Preserved the original request content while creating redacted traces.
  * Ensured text-only requests remain unchanged.
  * Added support for image conversion in Responses API traces.

### DIFF
--- a/lib/llm/src/request_trace/config.rs
+++ b/lib/llm/src/request_trace/config.rs
@@ -103,6 +103,9 @@ pub struct RequestTracePolicy {
     pub s3_prefix: Option<String>,
     pub s3_roll_uncompressed_bytes: u64,
     pub s3_flush_interval_ms: u64,
+    /// Redact media request parts before snapshotting (`DYN_REQUEST_TRACE_REDACT_MEDIA`).
+    /// Defaults to `false`: media is recorded verbatim unless explicitly opted in.
+    pub redact_media: bool,
 }
 
 impl RequestTracePolicy {
@@ -223,6 +226,9 @@ fn load_from_env() -> RequestTracePolicy {
             .filter(|value| !value.is_empty())
             .unwrap_or_else(|| DEFAULT_TOOL_EVENTS_TOPIC.to_string())
     });
+    // Off by default: enabling request-trace payload records must not silently change
+    // what those records contain for existing deployments.
+    let redact_media = env_is_truthy(env_request_trace::DYN_REQUEST_TRACE_REDACT_MEDIA);
     let s3_bucket = env_trimmed(env_request_trace::DYN_REQUEST_TRACE_S3_BUCKET);
     let s3_region = env_trimmed(env_request_trace::DYN_REQUEST_TRACE_S3_REGION);
     let s3_prefix = env_trimmed(env_request_trace::DYN_REQUEST_TRACE_S3_PREFIX);
@@ -256,6 +262,7 @@ fn load_from_env() -> RequestTracePolicy {
         s3_prefix,
         s3_roll_uncompressed_bytes,
         s3_flush_interval_ms,
+        redact_media,
     }
 }
 
@@ -442,6 +449,7 @@ mod tests {
         env_request_trace::DYN_REQUEST_TRACE_FILE_FORMAT,
         env_request_trace::DYN_REQUEST_TRACE_CAPACITY,
         env_request_trace::DYN_REQUEST_TRACE_RECORDS,
+        env_request_trace::DYN_REQUEST_TRACE_REDACT_MEDIA,
         env_request_trace::DYN_REQUEST_TRACE_NATS_SUBJECT,
         env_request_trace::DYN_REQUEST_TRACE_OTEL_MAX_PAYLOAD_BYTES,
         env_request_trace::DYN_REQUEST_TRACE_FILE_BUFFER_BYTES,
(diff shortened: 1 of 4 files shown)


dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `0c36fa7` | unsigned |

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `3aee381` | unsigned |

yifjiang:
Both CodeRabbit nitpicks addressed in `3aee3818`. Verified each against the current tree before applying rather than taking the suggested diffs as-is.

**1. Audio coverage in `redact_media_labels_video_and_audio`** — valid; the test name promised audio and only exercised video. Added an `audio_url` part with its own secret. Confirmed the wire shape `{"type": "audio_url", "audio_url": {"url": …}}` against `lib/llm/tests/preprocessor.rs:553` rather than assuming it. Note the placeholder is `[audio omitted by audit]`, not `[audio_url …]` — `AudioUrl` maps to the bare kind `audio`, so a copy-paste of the video arm would have asserted the wrong string.

**2. Test the snapshot hook directly** — the more valuable of the two, and a real gap: every existing test called `redact_media` directly, so a bypass introduced at the ingress hook would have left all of them green while raw media reached the sinks. Added `create_handle_stores_redacted_request`, which drives `create_handle_with_config` and asserts on the request actually stored on the handle.

Redaction tests are now seven:

| test | guards |
|---|---|
| `redact_media_none_for_text_only` | pure text → `None`, no clone |
| `redact_media_none_for_text_parts_array` | text-only parts array → `None` |
| `redact_media_replaces_user_image_url` | user media replaced; text sibling survives; original not mutated |
| `redact_media_replaces_tool_media` | tool-message media replaced |
| `redact_media_labels_video_and_audio` | distinct `video_url` / `audio` placeholder kinds |
| `redact_media_covers_responses_api_input_image` | `/v1/responses` `input_image` survives the conversion into a redactable part |
| `create_handle_stores_redacted_request` | the snapshot hook itself cannot store raw media |

Unrelated but worth flagging for anyone reading the earlier CI history: the Datadog comment above reports the previous failure as *"expected string to contain the video URL placeholder but it contained sensitive information."* That is inaccurate — no assertion failed and nothing leaked. The actual failure was `cargo fmt --check` objecting to a stray blank line before the closing brace of `mod tests`, and the quoted lines merely appeared in the fmt diff.

dynamo-ops:
## Unsigned Commits Detected

| Commit | Reason |
|--------|--------|
| `5b0bef3` | unsigned |

PR #12550, opened 2026-08-03
feat(router): support native policies in Dynamo frontends

ishandhanani:
Adds generic construction seams for statically linked native worker-selection policies in Dynamo's default HTTP frontend and Kubernetes EPP. Customers can rebuild either production entrypoint with a custom `WorkerSelectionPolicy` while retaining the same discovery, scheduling, host validation, accounting, reservations, and dispatch as the concrete default.

CLOSES: DYN-3725

### How This Was Implemented
- Added `HttpFrontend::default()` as the shared HTTP runner; an optional startup factory selects `WorkerSelectionPolicy`, while the default remains concrete `DefaultWorkerSelector`.
- Carried the selector type through the existing watcher, decode router, prefill router, and scheduler without request-path type erasure.
- Added `EppRouter::from_selection_service` as the complete public EPP seam around a caller-built `SelectionService`; the lower-level selector wrapper remains crate-private.
- Reused the existing EPP discovery and reconciliation path and validates queue capacity and replica-sync capability before Kubernetes discovery starts.
- Kept policy configuration and examples out of this PR; callers provide a normal statically linked Rust factory.

<details>

#### Mental model

Both supported entrypoints make one startup choice, then enter Dynamo's existing routing machinery. The default choice stays statically dispatched; only an explicitly supplied policy uses the published scorer/picker composition.

```mermaid
flowchart LR
    DF[Default HTTP frontend] --> H[HttpFrontend]
    CF[Custom HTTP image] --> H
    CE[Custom EPP image] --> E[EppRouter]
    H --> S[SelectionService and scheduler]
    E --> S
    S --> G[Host eligibility]
    G --> P[Concrete default or custom score/pick]
    P --> V[Host validation and accounting]
    V --> D[Existing dispatch and reservation lifecycle]
```

#### State and ordering

The HTTP factory runs when each decode or prefill worker set is constructed, not per request. Without a factory, `DefaultWorkerSelector` remains concrete throughout the watcher and routers and does not construct custom candidate columns or a trait-object policy.

The EPP constructor accepts an already built `SelectionService`. It checks that queue-enabled models have `max_num_batched_tokens` and that peer discovery is paired with replica sync before starting cluster discovery.

#### Request lifecycle

Discovery creates or reconciles workers through the existing entrypoint-specific code. Requests then enter the existing policy-class scheduler, Dynamo computes host eligibility, the selected strategy scores and picks, and Dynamo validates the returned row before the normal accounting, reservation, dispatch, completion, cancellation, and drop cleanup paths.

This PR exposes construction only. It adds no policy registry, runtime loading, configuration graph, filters, C ABI, example policy, or alternate worker-selection engine.

</details>

### Validation
- `cargo test -p dynamo-ext-proc --lib` (117 passed)
- `cargo check -p dynamo-llm --tests`
- `cargo clippy -p dynamo-ext-proc --all-targets -- -D warnings`
- `cargo clippy -p dynamo-llm --tests -- -D warnings`
- `git diff --check`


ishandhanani:
Folded this generic HTTP frontend and EPP construction work into #12526 as requested. #12536 remains stacked directly on #12526 and contains only the standalone SGLang cache/load policy crate plus required ownership metadata.

PR #12554, opened 2026-08-03
feat(observability): add request-trace export sampling

### LINKED ISSUE #12553: feat(observability): add deterministic sampling for request-trace exports
## Why this matters

Request-trace records can fan out to file, stderr, NATS, S3, and OTLP. At production request rates, every selected request, payload, and tool row is offered to every configured sink. Operators need a volume control for those records without changing runtime span collection or ordinary application logging.

## Scope

Add `DYN_REQUEST_TRACE_SAMPLE_RATIO`, a finite ratio in `[0.0, 1.0]`, evaluated before request-trace fan-out.

* Unset or `1.0` retains every selected request-trace record; `0.0` retains none.
* For a fractional ratio, choose deterministically from the agent session ID for agent request-end, payload, and tool rows; otherwise use the request ID. Thus, all rows in one agent trajectory have one result.
* Make the payload decision before creating the payload-capture handle, so rejected payloads do not incur header/body capture work.
* A retained row is offered to every configured sink; a dropped row is offered to none.
* Export `dynamo_frontend_request_trace_sampling_records_total{decision,event_type}` at the frontend `/metrics` endpoint. It measures the pre-fan-out decision, not queue pressure or delivery success.
* `DYN_AUDIT_FORCE_LOGGING` still selects legacy payload records, but those selected records are also subject to this ratio.

## Out of scope

* Runtime OpenTelemetry span sampling (`OTEL_TRACES_SAMPLE_RATIO`).
* Sampling ordinary Rust/Python structured logs: ai-dynamo/dynamo#13152.
* Forward Pass Metrics sampling and interval behavior.
* Sink retries, queue capacity, delivery guarantees, or per-sink sampling.

## Definition of done

* Unit tests cover unset, `0`, `1`, fractional, and invalid/non-finite/out-of-range values; invalid values retain records and emit a configuration warning.
* Tests prove deterministic decisions, agent-session coherence, pre-capture payload suppression, and pre-fan-out all-sink/none-sink behavior.
* Tests verify legacy forced payloads are sampled rather than bypassing the ratio.
* The metric uses only `decision={retained,dropped}` and the bounded request-trace event-type label.
* Docs distinguish this setting from `OTEL_TRACES_SAMPLE_RATIO` and document the metric semantics.

## Implementation

Implemented by [#12554](<https://github.com/ai-dynamo/dynamo/issues/12554>); close this issue when that PR merges.

kristinapathak:
## Overview

Add deterministic `DYN_REQUEST_TRACE_SAMPLE_RATIO` sampling and expose each pre-fan-out decision through a frontend Prometheus counter.

## Details

- Agent request-end, payload, and tool rows use their session ID; other rows use the request ID.
- `dynamo_frontend_request_trace_sampling_records_total{decision,event_type}` distinguishes retained and dropped rows from sink delivery health.
- `OTEL_TRACES_SAMPLE_RATIO` remains limited to runtime OpenTelemetry span sampling.
- `DYN_AUDIT_FORCE_LOGGING` selects legacy payload records but does not bypass request-trace sampling.

- `lib/llm/src/request_trace/config.rs` for deterministic selection.
- `lib/llm/src/request_trace/metrics.rs` and `lib/llm/src/http/service/service_v2.rs` for metric registration and exposure.

## Validation

- `cargo check -p dynamo-llm --lib --no-default-features`
- `git diff --check`
- Regenerated `lib/bindings/python/src/dynamo/prometheus_names.py`

## Related Issues

- Closes #12553

### DIFF
--- a/lib/llm/src/fpm_trace/config.rs
+++ b/lib/llm/src/fpm_trace/config.rs
@@ -28,13 +28,18 @@ pub struct FpmTracePolicy {
     pub enabled: bool,
     pub output_path: String,
     pub mode: FpmTraceMode,
-    pub sample_ratio: Option<f64>,
     pub sample_interval_ms: u64,
     pub jsonl_gz_roll_bytes: u64,
     pub max_segments: usize,
 }
 
-static POLICY: OnceLock<FpmTracePolicy> = OnceLock::new();
+#[derive(Clone, Debug)]
+struct FpmTraceRuntimePolicy {
+    policy: FpmTracePolicy,
+    sample_ratio: Option<f64>,
+}
+
+static POLICY: OnceLock<FpmTraceRuntimePolicy> = OnceLock::new();
 
 impl Default for FpmTracePolicy {
     fn default() -> Self {
@@ -42,7 +47,6 @@ impl Default for FpmTracePolicy {
             enabled: false,
             output_path: DEFAULT_OUTPUT_PATH.to_string(),
             mode: FpmTraceMode::Sampled,
-            sample_ratio: None,
             sample_interval_ms: DEFAULT_SAMPLE_INTERVAL_MS,
             jsonl_gz_roll_bytes: DEFAULT_JSONL_GZ_ROLL_BYTES,
             max_segments: DEFAULT_MAX_SEGMENTS,
@@ -67,7 +71,7 @@ where
     Ok(parsed)
 }
 
-fn load_enabled_policy() -> anyhow::Result<FpmTracePolicy> {
+fn load_enabled_runtime_policy() -> anyhow::Result<FpmTraceRuntimePolicy> {
     let output_path = match std::env::var(env_fpm_trace::DYN_FPM_OUTPUT_PATH) {
         Ok(value) if value.trim().is_empty() => {
             anyhow::bail!("{} must not be empty", env_fpm_trace::DYN_FPM_OUTPUT_PATH)
@@ -83,49 +87,69 @@ fn load_enabled_policy() -> anyhow::Result<FpmTracePolicy> {
         Err(_) => FpmTraceMode::Sampled,
     };
 
-    Ok(FpmTracePolicy {
-        enabled: true,
-        output_path,
-        mode,
+    Ok(FpmTraceRuntimePolicy {
+        policy: FpmTracePolicy {
+            enabled: true,
+            output_path,
+            mode,
+            sample_interval_ms: positive_integer_from_env(
+                env_fpm_trace::DYN_FPM_SAMPLE_INTERVAL_MS,
+                DEFAULT_SAMPLE_INTERVAL_MS,
+            )?,
+            jsonl_gz_roll_bytes: positive_integer_from_env(
+                env_fpm_trace::DYN_FPM_JSONL_GZ_ROLL_BYTES,
+                DEFAULT_JSONL_GZ_ROLL_BYTES,
+            )?,
+            max_segments: positive_integer_from_env(
+                env_fpm_trace::DYN_FPM_MAX_SEGMENTS,
+                DEFAULT_MAX_SEGMENTS,
+            )?,
+        },
         sample_ratio: trace_sample_ratio_from_env(),
-        sample_interval_ms: positive_integer_from_env(
-            env_fpm_trace::DYN_FPM_SAMPLE_INTERVAL_MS,
-            DEFAULT_SAMPLE_INTERVAL_MS,
-        )?,
-        jsonl_gz_roll_bytes: positive_integer_from_env(
-            env_fpm_trace::DYN_FPM_JSONL_GZ_ROLL_BYTES,
-            DEFAULT_JSONL_GZ_ROLL_BYTES,
-        )?,
-        max_segments: positive_integer_from_env(
-            env_fpm_trace::DYN_FPM_MAX_SEGMENTS,
-            DEFAULT_MAX_SEGMENTS,
-        )?,
     })
 }
 
-fn load_from_env() -> FpmTracePolicy {
+fn disabled_runtime_policy() -> FpmTraceRuntimePolicy {
+    FpmTraceRuntimePolicy {
+        policy: FpmTracePolicy::default(),
+        sample_ratio: None,
+    }
+}
+
+fn load_runtime_policy_from_env() -> FpmTraceRuntimePolicy {
     let enabled = match std::env::var(env_fpm_trace::DYN_FPM_TRACE) {
-        Err(_) => return FpmTracePolicy::default(),
+        Err(_) => return disabled_runtime_policy(),
         Ok(value) => match parse_bool(&value) {
             Ok(enabled) => enabled,
             Err(error) => {
                 tracing::warn!(%error, "invalid FPM trace configuration; tracing disabled");
-                return FpmTracePolicy::default();
+                return disabled_runtime_policy();
             }
         },
     };
     if !enabled {
-        return FpmTracePolicy::default();
+        return disabled_runtime_policy();
     }
 
-    load_enabled_policy().unwrap_or_else(|error| {
+    load_enabled_runtime_policy().unwrap_or_else(|error| {
         tracing::warn!(%error, "invalid FPM trace configuration; tracing disabled");
-        FpmTracePolicy::default()
+        disabled_runtime_policy()
     })
 }
 
+#[cfg(test)]
+fn load_from_env() -> FpmTracePolicy {
+    load_runtime_policy_from_env().policy
+}
+
 pub fn policy() -> &'static FpmTracePolicy {
-    POLICY.get_or_init(load_from_env)
+    &POLICY.get_or_init(load_runtime_policy_from_env).policy
+}
+
+pub(super) fn sample_ratio() -> Option<f64> {
+    POLICY
+        .get_or_init(load_runtime_policy_from_env)
+        .sample_ratio
 }
 
 pub fn is_enabled() -> bool {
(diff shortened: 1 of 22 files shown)


lib/llm/src/request_trace/config.rs
```diff
@@ -139,6 +183,7 @@ fn load_from_env() -> RequestTracePolicy {
     let enabled = !records.is_empty();
     let (sinks, legacy_file_format, legacy_audit_sinks_selected) =
         load_sinks(enabled, legacy_audit_sinks.as_deref());
+    let sample_ratio = enabled.then(trace_sample_ratio_from_env).flatten();
```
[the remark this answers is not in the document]
kristinapathak: Fixed in 6743ce3. `DYN_AUDIT_FORCE_LOGGING` payload records now bypass `OTEL_TRACES_SAMPLE_RATIO`, preserving the legacy capture contract. Normal request-trace and FPM exports remain sampled, and the docs state the exception.

lib/llm/src/request_trace/config.rs
```diff
@@ -126,6 +129,47 @@ impl RequestTracePolicy {
...
+                .agent_context
+                .as_ref()
+                .map(|context| context.session_id.as_str())
+        });
+    let Some(key) = key else {
+        return false;
+    };
+
+    let threshold = (ratio * ((u64::MAX as f64) + 1.0)) as u64;
+    xxh3_64(key.as_bytes()) < threshold
```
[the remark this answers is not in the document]
kristinapathak: Fixed in 6743ce3. Agent context now takes priority, so agent request-end and harness tool rows use the same session ID. The new test covers the case where the request ID would otherwise make a different decision.

lib/llm/src/request_trace/mod.rs
```diff
@@ -102,7 +102,10 @@ pub(crate) async fn start_tool_event_ingest_from_policy(
 }
 
 pub fn publish(record: RequestTraceRecord) {
-    BUS.publish(record);
+    // Sample before the broadcast so every configured sink writes the same records.
+    if policy().should_sample(&record) {
+        BUS.publish(record);
+    }
```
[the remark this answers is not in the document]
kristinapathak: Confirmed. This change shares the configured ratio but intentionally makes independent record-level decisions because request-trace and FPM records do not carry the OpenTelemetry trace ID. The PR description now documents that boundary. Correlated sampling requires separate trace-context propagation, so it is outside this change.

lib/llm/src/fpm_trace/config.rs
```diff
@@ -27,6 +28,7 @@ pub struct FpmTracePolicy {
     pub enabled: bool,
     pub output_path: String,
     pub mode: FpmTraceMode,
+    pub sample_ratio: Option<f64>,
```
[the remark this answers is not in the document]
kristinapathak: Fixed in c142edd. `FpmTracePolicy` is restored to its previous public shape. A private runtime-policy wrapper now carries the sample ratio from environment loading through worker initialization.

lib/llm/src/request_trace/config.rs
```diff
@@ -86,6 +87,8 @@ pub struct RequestTracePolicy {
     pub enabled: bool,
     pub records: Vec<RequestTraceRecordKind>,
     pub sinks: Vec<RequestTraceSinkKind>,
+    pub sample_ratio: Option<f64>,
```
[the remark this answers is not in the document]
kristinapathak: Fixed in c142edd. `RequestTracePolicy` is restored to its previous public shape. A private runtime-policy wrapper carries the sample ratio and audit state; sampling is exposed only through a parent-module helper used by `publish`.

tmonty12:
I think we should decouple this from `OTEL_TRACES_SAMPLE_RATIO`:

- In Dynamo, it controls runtime OTel span sampling.
- Request traces are separate records. Their `otel` sink exports OTLP logs, but they also fan out to file, stderr, NATS, and S3.
- Reusing the OTel-named setting is confusing: this change actually controls every request-trace sink, not just OTel output.

Could we:

- Rename the new knob to `DYN_REQUEST_TRACE_SAMPLE_RATIO`.
- Remove the FPM portion; its existing `sampled` mode already controls local trace volume.
- Make one deterministic keep/drop decision keyed by agent session ID, then propagate it to the payload, request-end, and tool-event paths.

That gives a coherent outcome: every record in an agent trajectory is retained together or dropped together, rather than sampled independently.

kristinapathak:
Implemented in 9caadaf.

- Added `DYN_REQUEST_TRACE_SAMPLE_RATIO`; `OTEL_TRACES_SAMPLE_RATIO` remains runtime span-only.
- Removed the FPM sampling changes.
- Agent payload capture receives the session ID through a private wrapper, and request-end and tool records use the same deterministic session key. Non-agent records use the request ID.
- Preserved the `DYN_AUDIT_FORCE_LOGGING` payload bypass and public request-trace APIs.
- Removed the stale unused profiler-test import reported by pre-commit.

Focused request-trace tests pass with 38 tests total. `cargo clippy -p dynamo-llm --no-default-features --lib -- -D warnings` and the repository-pinned Ruff hook pass.

lib/llm/src/request_trace/config.rs
```diff
@@ -128,17 +129,110 @@ impl RequestTracePolicy {
...
+            eprintln!(
+                "WARNING: invalid DYN_REQUEST_TRACE_SAMPLE_RATIO '{}'; expected a number between 0.0 and 1.0, keeping all request trace records",
+                raw
+            );
+            None
+        }
+    }
+}
+
+static POLICY: OnceLock<RequestTraceRuntimePolicy> = OnceLock::new();
```
tmonty12, line 220: The focused request-trace suite is now parallel-racy. Emitters initialize this global `OnceLock` while the config tests temporarily set `DYN_REQUEST_TRACE_SAMPLE_RATIO=0`; an unmarked emitter test can cache that value and then wait forever for a record that was sampled out. I reproduced two failures plus one >60s hang with the repository's default parallel test execution; each affected test passes alone, and all 67 pass with `--test-threads=1`. Can we avoid the global environment policy in these unit tests, or serialize every test that reads/initializes `POLICY` with the environment-mutating tests?

lib/llm/src/preprocessor.rs
```diff
@@ -3812,10 +3812,18 @@ impl
         } else {
             None
         };
-        let payload_handle = crate::request_trace::payload::create_handle(
+        let payload_agent_context = context
```
tmonty12: This lookup runs on every chat request, including the default tracing-disabled path. Missing agent context is normal, but `Context::get` formats an error `String` when the key is absent, so this adds another hash lookup plus heap allocation per request. Can we use `get_optional` and reuse the result for the later agent-context attachment, or gate the lookup behind payload capture?

lib/llm/src/request_trace/config.rs
```diff
@@ -128,17 +129,110 @@ impl RequestTracePolicy {
...
+    ///
+    /// The agent session ID keeps agent request and tool records together. Other
+    /// request-end and request-payload records use the request ID as their stable key.
+    fn should_sample(&self, record: &RequestTraceRecord) -> bool {
+        if self.legacy_audit_force_logging
+            && record.event_type == super::RequestTraceEventType::RequestPayload
+        {
+            return true;
+        }
+        request_trace_key(record).is_some_and(|key| self.should_sample_key(key))
```
tmonty12: This changes the no-op/default behavior of the public `publish()` API: a record with no `agent_context`, `request`, or `payload` is dropped even when the ratio is unset/invalid or exactly `1.0`, whereas pre-PR `publish()` broadcast unconditionally. Can we short-circuit the identity cases before requiring a key, then define keyless behavior only for fractional sampling?

lib/llm/src/preprocessor.rs
```diff
@@ -3812,10 +3812,18 @@ impl
...
+            .get::<crate::protocols::common::extensions::AgentContext>(
+                crate::protocols::common::extensions::AGENT_CONTEXT_CONTEXT_KEY,
+            )
+            .ok();
+        let payload_handle = crate::request_trace::payload::create_handle_with_agent_session(
```
tmonty12: The payload keep/drop decision happens here, after the HTTP path has already scanned, joined, and cloned the allowlisted headers into a `BTreeMap` and inserted it into the context. With a ratio of `0.0`, every payload record is suppressed but all header-capture costs are still paid. Since the request ID and agent context are already available in the HTTP path, can we decide whether to retain the payload before capturing headers?

lib/llm/src/request_trace/config.rs
```diff
@@ -128,17 +129,110 @@ impl RequestTracePolicy {
...
+    legacy_audit_force_logging: bool,
+}
+
+impl RequestTraceRuntimePolicy {
+    /// Applies the configured ratio before request-trace records fan out to sinks.
+    ///
+    /// The agent session ID keeps agent request and tool records together. Other
+    /// request-end and request-payload records use the request ID as their stable key.
+    fn should_sample(&self, record: &RequestTraceRecord) -> bool {
+        if self.legacy_audit_force_logging
```
tmonty12: Could `DYN_REQUEST_TRACE_SAMPLE_RATIO` apply to payloads selected by `DYN_AUDIT_FORCE_LOGGING` too? `DYN_AUDIT_FORCE_LOGGING` can remain the legacy record-selection switch, while an unset/`1.0` ratio still keeps every audit payload and an explicitly configured ratio controls export volume consistently. That lets us remove both audit-specific sampling bypasses and the `legacy_audit_force_logging` field from `RequestTraceRuntimePolicy`; docs and tests would need to reflect the new precedence.

tmonty12 (commented):
Deep review: five actionable inline findings.

lib/llm/src/request_trace/config.rs
kristinapathak: Fixed in 067c057c6. The emitter test that initializes the global request-trace policy is now serialized with the environment-mutating configuration tests, so it cannot cache a temporary sampling ratio.

lib/llm/src/preprocessor.rs
kristinapathak: Fixed in 067c057c6. This now uses `get_optional`, and the result is reused when attaching agent context to the preprocessed request, avoiding the missing-key error allocation and duplicate lookup.

lib/llm/src/request_trace/config.rs
kristinapathak: Fixed in 067c057c6. Unset and `1.0` ratios now preserve the public `publish()` no-op behavior for keyless records. Only fractional ratios require a stable record key; the new test covers all identity and keyless cases.

lib/llm/src/preprocessor.rs
kristinapathak: Fixed in 067c057c6. Payload capture and sampling are decided before the allowlisted-header context lookup, so sampled-out payloads skip header-map retrieval and capture work.

lib/llm/src/request_trace/config.rs
kristinapathak: Fixed in 067c057c6. `DYN_AUDIT_FORCE_LOGGING` now selects legacy payload records without bypassing `DYN_REQUEST_TRACE_SAMPLE_RATIO`. The audit-specific runtime field is removed, and the tests and reference docs cover the precedence.

tmonty12 (approved):

PR #12556, opened 2026-08-03
fix(sglang): restore TP non-leader readiness

### LINKED ISSUE #12555: SGLang TP non-leader nodes become ready too early and publish KV events
## Problem

When SGLang runs under Dynamo with multi-node tensor parallelism, non-leader nodes (`node_rank >= 1`) rely on Dynamo system health instead of SGLang's native dummy health server because Dynamo sets `SGLANG_BLOCK_NONZERO_RANK_CHILDREN=0`.

That changes two TP-only behaviors:

1. The Dynamo system port can be bound before the SGLang non-leader scheduler has completed initialization, so a TCP port readiness check can report ready too early.
2. In TP-only multi-node mode, non-leader ranks should not publish KV events under the leader worker id. Only DP-attention mode has distinct per-node DP rank slices that need relaying.

## Expected behavior

For TP-only multi-node SGLang:

- rank1+ should become ready only after the SGLang engine constructor has passed the scheduler-ready barrier, matching native SGLang dummy health server semantics.
- rank1+ should remain resident for local metrics, but should not resolve the leader worker id or initialize KV event publishing.

## Impact

Kubernetes readiness based on Dynamo's bound port can admit a TP non-leader before NCCL/SGLang scheduler readiness. KV event publishing from TP non-leaders can also create duplicate or incorrectly attributed KV event sources.

AllenXu93:
## Summary

- restore SGLang non-leader rank readiness by marking Dynamo system health ready only after the Engine has passed the scheduler-ready barrier
- keep non-leader SGLang nodes resident for metrics and clear readiness during cleanup/error paths
- avoid KV event publishing from TP-only non-leader nodes by using the existing publishes_kv_events gate

Fixes #12555

## Validation

- `python3 -m py_compile components/src/dynamo/sglang/publisher.py components/src/dynamo/sglang/init_llm.py components/src/dynamo/sglang/init_diffusion.py components/src/dynamo/sglang/tests/test_sglang_publisher.py components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py`
- `python3 -m pytest components/src/dynamo/sglang/tests/test_sglang_publisher.py components/src/dynamo/sglang/tests/test_sglang_local_dp_ranks.py -q` *(not run: local Python environment has no pytest module)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved readiness reporting for distributed inference workers.
  * Worker health is now marked ready only after local scheduler initialization completes.
  * Health status is reliably cleared during shutdown, cancellation, or initialization failures.

* **Tests**
  * Added coverage for readiness transitions during successful, delayed, and failed worker initialization.

### DIFF
--- a/components/src/dynamo/sglang/init_diffusion.py
+++ b/components/src/dynamo/sglang/init_diffusion.py
@@ -73,7 +73,7 @@ async def init_llm_diffusion(
     assert publisher is not None, "setup_sgl_metrics returned None on chat path"
 
     if server_args.node_rank >= 1:
-        await handle_non_leader_node(engine, publisher, metrics_task)
+        await handle_non_leader_node(runtime, engine, publisher, metrics_task)
         return
 
     ready_event = asyncio.Event()
--- a/components/src/dynamo/sglang/init_llm.py
+++ b/components/src/dynamo/sglang/init_llm.py
@@ -104,7 +104,7 @@ async def init_decode(
     logging.debug(f"SGLang model load time: {load_time:.2f}s")
 
     if server_args.node_rank >= 1:
-        await handle_non_leader_node(engine, publisher, metrics_task)
+        await handle_non_leader_node(runtime, engine, publisher, metrics_task)
         return
 
     ready_event = asyncio.Event()
@@ -259,7 +259,7 @@ async def init_prefill(
     publisher.component_gauges.set_model_load_time(load_time)
 
     if server_args.node_rank >= 1:
-        await handle_non_leader_node(engine, publisher, metrics_task)
+        await handle_non_leader_node(runtime, engine, publisher, metrics_task)
         return
 
     try:
--- a/components/src/dynamo/sglang/publisher.py
+++ b/components/src/dynamo/sglang/publisher.py
@@ -579,8 +579,9 @@ async def handle_non_leader_node(
     native rank-local readiness contract.
     """
     logging.info(
-        f"Non-leader node detected (node_rank={engine.server_args.node_rank}). "
-        "Running rank-local readiness and metrics."
+        "Non-leader node detected (node_rank=%s). "
+        "Running rank-local readiness and metrics.",
+        engine.server_args.node_rank,
     )
 
     runtime.set_health_status(True)
@@ -604,6 +605,7 @@ async def handle_non_leader_node(
                 await metrics_task
             except asyncio.CancelledError:
                 pass
-            publisher.cleanup()
+            finally:
+                publisher.cleanup()
         finally:
             runtime.set_health_status(False)
--- a/components/src/dynamo/sglang/publisher.py
+++ b/components/src/dynamo/sglang/publisher.py
@@ -557,6 +557,7 @@ async def setup_sgl_metrics(
 
 
 async def handle_non_leader_node(
+    runtime,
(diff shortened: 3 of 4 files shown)


jthomson04:
@AllenXu93, could you clarify which deployed readiness/probe contract this fixes and add validation at that boundary before closing #12555?

The current change marks Dynamo HTTP health ready after the Engine constructor returns, which is useful, but it does not change when the status-server TCP listener binds. A TCP-only readiness probe can therefore succeed independently of `set_health_status(True)`. Also, the current operator removes startup/readiness/liveness probes from SGLang non-leader worker containers, so please identify the actual manifest/probe used in the original reproduction rather than assuming this HTTP health change covers every deployment path.

Please validate a delayed engine initialization through the real construction/probe path: HTTP remains NotReady before the scheduler-ready barrier, becomes Ready afterward, and returns to NotReady on failure/cleanup. If the acceptance case still uses a TCP-only probe, explain how that probe will be corrected; toggling HTTP health alone does not address its early success.

Please also narrow the issue/PR completion claim: TP-only non-leader KV-event publishing was already gated by #12185, so the remaining readiness behavior is the important acceptance target here. The current head does preserve publisher cleanup when the metrics task fails; that earlier finding is addressed.

Finally, please run the affected tests in the repository environment and obtain current-head CI; the PR description still says pytest was not available locally.

AllenXu93:
Thanks @jthomson04 — agreeing on the acceptance target and the TCP vs HTTP distinction.

**Probe / readiness contract this PR intends to fix**  
This change restores the Dynamo **HTTP / system-health** readiness signal for SGLang TP non-leaders (`node_rank >= 1`): `runtime.set_health_status(True)` only after `handle_non_leader_node` (i.e. after Engine ctor / scheduler-ready barrier, with `SGLANG_BLOCK_NONZERO_RANK_CHILDREN=0`), and we clear health in `finally`.

It does **not** change when the status-server **TCP listener binds**. A TCP-only readiness probe can still succeed independently of `set_health_status(True)`. I also hear that the operator strips probes from SGLang non-leader containers — so this only helps deployments that actually probe Dynamo HTTP health.

**Original reproduction / which probe**  
The original #12555 repro used **Dynamo HTTP health** (not TCP-only on the status port). So this PR’s `set_health_status` gating matches the acceptance path we hit in practice.

**TCP-only acceptance case**  
If some other deployment still uses TCP-only probes, this PR alone would not fully close that path; that would need delaying TCP bind or switching the probe to HTTP — treating as a separate follow-up. For the HTTP-health repro, this change is the intended fix.

**Scope vs #12185**  
Agreed: TP-only non-leader KV publishing is already gated by #12185. Acceptance here is **readiness**.

**Validation**  
I re-ran the FakeRuntime readiness lifecycle coverage for `handle_non_leader_node` locally (ready on enter → cleared on cancel/error; cleanup still runs if metrics fail during cancel) — 8/8 assertions passed. Full in-repo `pytest` for `test_sglang_publisher.py` still needs an environment with `sglang` + compiled `dynamo._core` (this Mac doesn’t have that); I’ll rely on current-head CI for the full suite and will update the PR description to remove the old “pytest not available” note.

Happy to adjust further if you want an extra real HTTP NotReady→Ready→NotReady check against a live status endpoint.

PR #12558, opened 2026-08-03
fix(container): prevent mixed UCX linkage in SGLang arm64 runtime

Chokoyo:
## Overview:
Fixes a UCX library conflict in recent SGLang runtime (`nvcr.io/nvstaging/ai-dynamo/sglang-runtime:1.4.0-rc.0`, upstream `lmsysorg/sglang:v0.5.16-cu130-runtime`) that causes multimodal encode-to-PD NIXL reads to fail with `NIXL_ERR_BACKEND` on GB200.

Verified on GB200: the upstream SGLang container resolves NVSHMEM’s generic UCX dependencies to Ubuntu UCX 1.16 while NIXL loads its bundled UCX 1.21, creating an in-process UCX conflict.

## Details:

- Discover the installed `nixl-cu*` wheel and its auditwheel-renamed UCX libraries.
- Add generic `libucp.so.*` and `libucs.so.*` aliases beside NIXL’s private libraries.
- Expose that directory through a stable runtime path while preserving UCX’s `$ORIGIN` dependency and module lookup.
- Validate that NVSHMEM’s UCX transport and NIXL’s CUDA modules resolve without missing or distro UCX dependencies.
- Leave XPU images unchanged and avoid modifying existing wheel files or ELF metadata.
## Validation

Validated on GB200 (aarch64) with `Qwen/Qwen3-VL-2B-Instruct`, multimodal E/P/D in `--single-gpu` mode, the default `nixl-read` embedding-transfer path, and the video from DYN-3695.

With this PR's compatibility layout:

- Two consecutive cache-enabled video requests returned HTTP 200.
- The second request logged `Embedding cache hit for VIDEO URL index 0`, covering both the initial population and cache-hit paths.
- A separate cache-disabled control also returned HTTP 200.
- The encode process still loaded `nvshmem_transport_ucx.so.3`, but mapped only NIXL's UCX 1.21 core libraries and CUDA modules.
- No system UCX mappings, UCX CUDA-support warnings, remote-key unpack failures, or `NIXL_ERR_BACKEND` errors were observed.

- `container/deps/sglang/install_nixl_ucx_compat.sh` — installs the aliases and validates the resulting dependency graph.
- `container/deps/sglang/discover_nixl_ucx_layout.py` — discovers the NIXL and NVSHMEM wheel layout without hard-coding CUDA versions or library paths.
- `container/templates/sglang_runtime.Dockerfile` — wires the compatibility layout into non-XPU SGLang images.

## Related Issues

- close DYN-3695

## Summary by CodeRabbit

* **New Features**
  * Added automatic NIXL and UCX compatibility setup for CUDA-based runtime images.
  * Exposes required UCX libraries and transport plugins through a stable runtime path.
  * Supports compatible NVSHMEM transport discovery when available.

* **Bug Fixes**
  * Added validation for missing, ambiguous, inconsistent, or unresolved CUDA and UCX components.
  * Provides descriptive installation errors and warnings when required components are unavailable.
  * XPU runtime images remain unchanged.


container/deps/sglang/install_nixl_ucx_compat.sh
```diff
@@ -0,0 +1,112 @@
...
+    ln -s "${NIXL_LIB_DIR}" "${OUTPUT_DIR}"
+fi
+
+# Do not inherit the image's ambient NIXL paths here: they would mask a broken
+# $ORIGIN closure in the compatibility layout.
+for plugin in "${PLUGINS[@]}"; do
+    resolved="$(env -u LD_LIBRARY_PATH LD_LIBRARY_PATH="${OUTPUT_DIR}" ldd "${plugin}")"
+    if grep -F 'not found' <<<"${resolved}" >/dev/null; then
+        die "${plugin}: unresolved dependency after installing UCX aliases"
+    fi
```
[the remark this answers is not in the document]
Chokoyo: fixed in b9f10fa9da804f4520b2c240201ea27eb9d218d7.

container/templates/sglang_runtime.Dockerfile
```diff
@@ -162,6 +162,16 @@ RUN --mount=type=bind,source=./container/deps/requirements.sglang.txt,target=/tm
...
         --requirement /tmp/requirements.sglang.txt
 
+{% if device != "xpu" %}
+# Add generic UCX aliases beside NIXL's private libraries so native consumers
+# use the same implementation without breaking UCX's $ORIGIN-based core and
+# module lookup. No existing wheel file or ELF metadata is modified.
+RUN --mount=type=bind,source=./container/deps/sglang/install_nixl_ucx_compat.sh,target=/tmp/install_nixl_ucx_compat.sh,readonly \
+    --mount=type=bind,source=./container/deps/sglang/discover_nixl_ucx_layout.py,target=/tmp/discover_nixl_ucx_layout.py,readonly \
+    bash /tmp/install_nixl_ucx_compat.sh /opt/dynamo/nixl-ucx-compat
+ENV LD_LIBRARY_PATH=/opt/dynamo/nixl-ucx-compat${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
```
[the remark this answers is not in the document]
Chokoyo: fixed in b9f10fa9da804f4520b2c240201ea27eb9d218d7. Updated the gate to exclude dev and local-dev.

container/deps/sglang/discover_nixl_ucx_layout.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def main() -> None:
+    """Print the discovered layout as tab-separated records."""
+    nixl_matches = cuda_distributions("nixl")
+    if len(nixl_matches) != 1:
+        names = sorted(name for name, _, _ in nixl_matches)
+        raise SystemExit(
+            f"nixl-ucx-compat: expected one nixl-cu* distribution, found {names}"
+        )
```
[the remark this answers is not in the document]
Chokoyo: The fail-fast behavior is intentional. Silently skipping an unknown wheel layout would let the image build successfully while reintroducing the mixed-UCX failure this check is meant to prevent.

container/deps/sglang/install_nixl_ucx_compat.sh
```diff
@@ -0,0 +1,112 @@
...
+        die "${module_path}: unresolved private UCX dependency"
+    fi
+    if grep -E 'libuc[mpst][.]so[^=]*=> /(usr/)?lib(64)?/' <<<"${resolved}" >/dev/null; then
+        die "${module_path}: a distro UCX core library remains in its dependency graph"
+    fi
+done
+
+if [[ "${#PLUGINS[@]}" -eq 0 ]]; then
+    echo "nixl-ucx-compat: no NVSHMEM UCX transport found; skipped consumer validation"
+fi
```
dmitry-tokarev-nv, line 139: This is the one fail-open path in an otherwise strictly fail-fast script, and it guards the thing the PR exists to fix.

`PLUGINS` is populated only from `nvidia-nvshmem-cu*` **pip distributions** (`discover_nixl_ucx_layout.py:40-45`). If NVSHMEM ships in the base image outside pip — common in vendor CUDA images — or the wheel is ever renamed, the array is empty, the entire validation loop at lines 81-93 is skipped, and line 112 still prints `installed 2 aliases ...`. A skimmed build log reads as success while nothing has confirmed the aliases do anything.

That contrasts with every other path here: unknown wheel layout dies, missing UCX lib dies, alias that fails to resolve dies. You defended that fail-fast stance in the thread above and I agree with it — this is the one place it isn't applied.

It matters more than it looks because there is no runtime backstop: `pr.yaml` pins GPU lanes to `["amd64"]` ("No ARM GPUs available"), so the arm64 path this fixes has no CI lane and this build-time check is the only guard.

```suggestion
if [[ "${#PLUGINS[@]}" -eq 0 && "${NIXL_UCX_COMPAT_ALLOW_NO_CONSUMER:-0}" != "1" ]]; then
    die "no NVSHMEM UCX transport found; the aliases were installed but nothing validated them (set NIXL_UCX_COMPAT_ALLOW_NO_CONSUMER=1 to accept this)"
fi
```

container/deps/sglang/install_nixl_ucx_compat.sh
```diff
@@ -0,0 +1,112 @@
...
+    if grep -E 'libuc[mpst].*=> not found' <<<"${resolved}" >/dev/null; then
+        die "${plugin}: unresolved UCX dependency after installing UCX aliases"
+    fi
+    grep -F "libucp.so." <<<"${resolved}" | grep -F "${OUTPUT_DIR}/libucp.so." >/dev/null || \
+        die "${plugin}: libucp did not resolve through the compatibility directory"
+    grep -F "libucs.so." <<<"${resolved}" | grep -F "${OUTPUT_DIR}/libucs.so." >/dev/null || \
+        die "${plugin}: libucs did not resolve through the compatibility directory"
+    if grep -E 'libuc[mpst][.]so[^=]*=> /(usr/)?lib(64)?/' <<<"${resolved}" >/dev/null; then
+        die "${plugin}: a distro UCX core library remains in its dependency graph"
+    fi
```
dmitry-tokarev-nv: This enumerates distro paths rather than asserting the invariant, and the enumeration has gaps: `/(usr/)?lib(64)?/` matches `/lib*` and `/usr/lib*` but not `/usr/local/ucx/lib` — which is exactly where this repo installs its own UCX in the XPU branch of this same file (line 65) and in `dev.Dockerfile:274` — nor `/opt/hpcx/ucx/lib`, which several NVIDIA base images carry.

**Careful with the obvious fix, though:** simply adding `/usr/local/lib` to the denylist would break the build, because the NIXL wheel's own libraries live under `/usr/local/lib/python3.x/dist-packages/.nixl_cu*.mesonpy.libs/`. Your current pattern happens to be safe against that only because `local` follows `/usr/`. That trap is a good argument for inverting the check rather than extending it.

The allowlist form below covers every path at once and subsumes the `not found` check at lines 83-85 (an unresolved line yields `not`, which is not under either prefix). It accepts `${NIXL_LIB_DIR}` as well as `${OUTPUT_DIR}` because the `$ORIGIN`-resolved hashed dependencies may print through the real directory rather than the stable symlink, depending on how the loader canonicalized the path — worth confirming against your GB200 `ldd` output before taking it.

```suggestion
    while read -r soname arrow resolved_path _; do
        [[ "${arrow}" == "=>" ]] || continue
        [[ "${resolved_path}" == "${OUTPUT_DIR}/"* || "${resolved_path}" == "${NIXL_LIB_DIR}/"* ]] || \
            die "${plugin}: ${soname} resolved outside NIXL's UCX: ${resolved_path}"
    done < <(grep -E '(^|[[:space:]/])libuc[mpst][-.]' <<<"${resolved}")
```

The same gap applies to the module loop at line 104.

Separately: only `libucp` and `libucs` get aliases. That is fine today, but if a consumer ever needs a generic `libuct.so.0`/`libucm.so.0` the build dies with *"unresolved UCX dependency after installing UCX aliases"*, which doesn't hint that the alias set is what needs extending. Either alias all four in `discover_nixl_ucx_layout.py:48`, or say in a comment why two suffice.

dmitry-tokarev-nv (commented):
Reviewed the current head (`b9f10fa`) alongside DYN-3695 and the existing bot threads. The Devin findings on the `not found` scope and the dev/local-dev gate are already resolved, so the inline comments below are only the points not raised elsewhere.

The mechanism looks right to me: resolving the generic SONAME to the file NIXL already mapped lets glibc dedupe by `(st_dev, st_ino)` and reuse the existing mapping instead of creating a second UCX, and keeping the alias in the *same* directory to preserve `$ORIGIN` is the correct call — the comment explaining that is the comment a future reader needs.

Three PR-level asks beyond the inline notes:

**1. Validation against the ticket's own acceptance criterion.** DYN-3695 names `test_sglang_serve_deployment[multimodal_epd_video_embedding_cache]` (dynamo-qa) as "the only failing case in the 1.4.0-rc.0 sanity run without an owning bug." This PR verifies the *linkage* (NVSHMEM resolving to Ubuntu UCX 1.16 while NIXL loads its bundled 1.21), but doesn't state that the repro now returns 200 or that the QA case passes. Could you add a `Validation` section (also required by `AGENTS.md`) with that result?

**2. Flag-specificity isn't explained by the stated mechanism.** Mixed linkage is a static property of the image — present on every launch. DYN-3695 documents the identical launch *without* `--multimodal-embedding-cache-capacity-gb` passing on the same GB200 node. There is a plausible reconciliation — the cache path registers the domain that becomes `md[1]`, and only that domain needs the CUDA support the mis-linked UCX lacks — but the ticket explicitly leaves root cause unestablished and prescribes the diagnostic: rerun both launches with `UCX_LOG_LEVEL=debug` / `NIXL_LOG_LEVEL=DEBUG` and diff the registered memory domains. That should settle it in one run.

Relatedly, the `ucx_utils.cpp:444] UCX version is less than 1.19, CUDA support is limited` line from the ticket is the strongest single piece of evidence for this fix — it is exactly what you would see if NIXL's runtime UCX were the distro 1.16 rather than its bundled 1.21. It currently lives only in Linear; it belongs in the PR body.

**3. Nothing in CI will exercise this.** The repro is already in-repo as `tests/serve/test_sglang.py::video_e_pd_qwen` (`multimodal_epd.sh` + `--multimodal-embedding-cache-capacity-gb 0.1`, `pre_merge`), and it runs against the runtime image this PR fixes (`shared-test.yml:142` uses `target_tag_plain`). But `pr.yaml:702,741,780` pin the GPU lanes to `["amd64"]` — *"No ARM GPUs available"* — and the arm64 lanes are CPU-only with single-GPU tests skipped. So the failing configuration has no lane, and the build-time `ldd` check is the only standing regression guard. That is why the fail-open path flagged inline matters more than it first looks.

Also: DYN-3695 is tagged `[release/1.4]` against `1.4.0-rc.0` and this targets `main` — is a `release/1.4.0` cherry-pick tracked anywhere?

container/deps/sglang/install_nixl_ucx_compat.sh
```diff
@@ -0,0 +1,112 @@
...
+    esac
+done
+
+[[ -n "${NIXL_PACKAGE}" ]] || die "failed to identify the NIXL CUDA wheel"
+[[ -d "${NIXL_LIB_DIR}" ]] || die "failed to locate the NIXL private library directory"
+[[ "${#ALIASES[@]}" -eq 2 ]] || die "failed to locate both NIXL UCX core libraries"
+
+# Keep the generic names beside NIXL's private libraries. Loading libucp or
+# libucs through a separate symlink directory changes $ORIGIN and prevents UCX
+# from finding both its hashed core dependencies and its ucx/ module directory.
```
dmitry-tokarev-nv, line 88: The PR description says this avoids "modifying existing wheel files or ELF metadata" — accurate, but worth being precise about what it *does* do: these are new files inside a pip-managed install tree that the wheel's `RECORD` doesn't list.

Two consequences, both silent:
- `pip uninstall nixl-cu*` removes only RECORD-listed files, leaving the two symlinks behind as strays.
- Any later reinstall or upgrade of NIXL drops the aliases while `LD_LIBRARY_PATH` keeps pointing at a directory that no longer has them — no build failure, just the original bug back.

Nothing in the current template reinstalls NIXL after this step, so this is future-proofing rather than a live problem. A note in the existing comment block makes the ordering constraint explicit for whoever adds the next `pip install` to this file:

```suggestion
# Keep the generic names beside NIXL's private libraries. Loading libucp or
# libucs through a separate symlink directory changes $ORIGIN and prevents UCX
# from finding both its hashed core dependencies and its ucx/ module directory.
#
# These symlinks land inside the wheel's install tree but are absent from its
# RECORD: they survive `pip uninstall nixl-cu*` as strays, and any reinstall or
# upgrade of NIXL drops them silently while LD_LIBRARY_PATH keeps pointing at
# the directory. This step must run after the last install that can touch NIXL.
```

container/deps/sglang/install_nixl_ucx_compat.sh
```diff
@@ -0,0 +1,112 @@
...
+    fi
+    if grep -E 'libuc[mpst][.]so[^=]*=> /(usr/)?lib(64)?/' <<<"${resolved}" >/dev/null; then
+        die "${module_path}: a distro UCX core library remains in its dependency graph"
+    fi
+done
+
+if [[ "${#PLUGINS[@]}" -eq 0 ]]; then
+    echo "nixl-ucx-compat: no NVSHMEM UCX transport found; skipped consumer validation"
+fi
+echo "nixl-ucx-compat: installed ${#ALIASES[@]} aliases for ${NIXL_PACKAGE} in ${NIXL_LIB_DIR}"
```
dmitry-tokarev-nv: Consider recording which UCX everything now shares, in the build log.

This repo already has a live sensitivity to the bundled UCX version: `trtllm_runtime.Dockerfile:45-46` LD_PRELOADs TRT-LLM's own libnixl specifically to *avoid* nixl-cu13's bundled UCX 1.20.0 (ai-dynamo/nixl#1668 — "hangs with two agents/host"), and `aws.Dockerfile:51` repeats the caveat. This change moves NVSHMEM off the distro UCX and onto whatever UCX the SGLang image's `nixl-cu*` wheel happens to bundle. You state 1.21 in the description, presumably past #1668 — but nothing here pins or asserts it, so a future NIXL bump could quietly land the image on a regressed UCX with no signal.

No clean suggestion from me: the auditwheel filename carries the SONAME version (`.so.0.0.0`), not the UCX release, so this needs whatever method you used to establish 1.21. Even echoing the version alongside the alias count would make a future regression visible in the build log. A cross-reference to #1668 in the comment at lines 25-27 would also help the next person see that the two workarounds pull in opposite directions.

container/deps/sglang/discover_nixl_ucx_layout.py
dmitry-tokarev-nv: Minor, and fully in keeping with the fail-fast stance you defended below — just make the failure diagnosable.

`distributions()` yields one entry per `sys.path` location, so a package visible from two dist-packages directories surfaces twice. This Dockerfile installs with `--break-system-packages` repeatedly, so that isn't purely hypothetical, and today the message would read `found ['nixl-cu13', 'nixl-cu13']` — which gives no way to tell a genuine ambiguity (two CUDA majors) from the same install seen twice. Including the install root distinguishes them at no behavioural cost:

```suggestion
    if len(nixl_matches) != 1:
        found = sorted(
            f"{name} ({Path(dist.locate_file('')).resolve()})"
            for name, _, dist in nixl_matches
        )
        raise SystemExit(
            f"nixl-ucx-compat: expected one nixl-cu* distribution, found {found}"
        )
```

(The zero-match case still prints `[]`, which reads a little oddly for "expected one, found none" — `install_nixl_from_wheel.sh`'s "upstream NIXL wheel layout changed" phrasing would be both clearer and consistent with the sibling script.)

container/deps/sglang/discover_nixl_ucx_layout.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Discover NIXL's private UCX libraries and native UCX consumers."""
```
dmitry-tokarev-nv, line 3: Worth cross-referencing `container/deps/vllm/install_nixl_from_wheel.sh`, which already solves the adjacent half of this problem: it discovers the same `.nixl_cu*.mesonpy.libs` directory (by glob rather than `importlib.metadata`) and already creates a stable `/opt/dynamo/nixl` symlink to it, with its own "upstream NIXL wheel layout changed" failure messages.

This PR adds a second discovery mechanism and a second stable prefix (`/opt/dynamo/nixl-ucx-compat`) pointing at the same directory, under a framework-specific path. Since both are pinned to the upstream wheel's internal layout — which you have reasonably chosen to fail fast on — a future NIXL repackage becomes two fixes in two places with two different techniques.

Not blocking, and I don't think the UCX-alias logic belongs in that script as-is. But either hoisting the shared discovery to `container/deps/` or just naming the other script in a comment here would keep the coupling visible.

container/templates/sglang_runtime.Dockerfile
```diff
@@ -162,6 +162,16 @@ RUN --mount=type=bind,source=./container/deps/requirements.sglang.txt,target=/tm
...
         --requirement /tmp/requirements.sglang.txt
 
+{% if device != "xpu" and target not in ("dev", "local-dev") %}
+# Add generic UCX aliases beside NIXL's private libraries so native consumers
+# use the same implementation without breaking UCX's $ORIGIN-based core and
+# module lookup. No existing wheel file or ELF metadata is modified.
+RUN --mount=type=bind,source=./container/deps/sglang/install_nixl_ucx_compat.sh,target=/tmp/install_nixl_ucx_compat.sh,readonly \
+    --mount=type=bind,source=./container/deps/sglang/discover_nixl_ucx_layout.py,target=/tmp/discover_nixl_ucx_layout.py,readonly \
+    bash /tmp/install_nixl_ucx_compat.sh /opt/dynamo/nixl-ucx-compat
+ENV LD_LIBRARY_PATH=/opt/dynamo/nixl-ucx-compat${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
```
dmitry-tokarev-nv, line 175: Separate from the dev/local-dev thread above: this prepends the directory to `LD_LIBRARY_PATH` for **every** process in the image, and that directory holds more than UCX.

The NIXL wheel's `.nixl_cu*.mesonpy.libs` also carries generically-named project libraries — `install_nixl_from_wheel.sh:102` requires `libnixl.so libnixl_build.so libnixl_common.so libserdes.so libstream.so` in exactly that directory. The auditwheel-vendored dependencies are hash-mangled and so can't shadow anything, but those aren't; any binary in the image that needs a generic `libstream.so` or `libserdes.so` now resolves to NIXL's copy first. Low probability, but it is a global change made to solve a two-library problem, and it also applies to the `licenses`/compliance stage and to the EFA stage when `make_efa` is set.

Options, roughly in order of how much I'd argue for them:
- Keep it and say so in the comment above — "deliberately global; the wheel dir is on the default search path for every process" — so the blast radius is a decision rather than a side effect.
- Drop a `/etc/ld.so.conf.d/nixl-ucx-compat.conf` + `ldconfig` instead. Lower precedence than `LD_LIBRARY_PATH` and it doesn't propagate into every child environment. It would not override a consumer whose `DT_RPATH` points elsewhere — your existing `ldd` check would catch that at build time.
- Scope the variable to the worker entrypoint rather than the image.

Also a nit: `/opt/dynamo/nixl-ucx-compat` is now written three times (script default at `install_nixl_ucx_compat.sh:7`, the argument on line 171, and this `ENV`). Dropping the script default or the explicit argument would leave one source of truth.

container/deps/sglang/install_nixl_ucx_compat.sh
Chokoyo: Fixed in 69c5b315. A missing NVSHMEM UCX transport now fails by default, with NIXL_UCX_COMPAT_ALLOW_NO_CONSUMER=1 as the explicit escape hatch. I also checked the existing multi-arch runtime and EFA build logs: both amd64 and arm64 currently discover the transport.

container/deps/sglang/install_nixl_ucx_compat.sh
Chokoyo: Fixed in 69c5b315. Validation now allowlists only the stable compatibility path or the real NIXL library directory for every libuc[mpst] entry, for both the NVSHMEM plugin and CUDA modules.

container/deps/sglang/install_nixl_ucx_compat.sh
Chokoyo: Fixed in 69c5b315. The discovery helper now loads the exact wheel-private libucp and calls ucp_get_version(), so the build log reports the actual bundled release.

container/deps/sglang/install_nixl_ucx_compat.sh
Chokoyo: Added in 69c5b315.

container/deps/sglang/discover_nixl_ucx_layout.py
Chokoyo: Fixed in 69c5b315. The zero-match case now has a dedicated error, and an ambiguous match includes each distribution’s resolved installation root.

container/deps/sglang/discover_nixl_ucx_layout.py
Chokoyo: Checked the installed wheel layout and these are different directories. `/opt/dynamo/nixl targets .nixl_cu*.mesonpy.libs`, which contains the NIXL project libraries. The new `/opt/dynamo/nixl-ucx-compat` symlink targets nixl_cu*.libs, which contains the auditwheel-vendored UCX dependency closure and `ucx/ modules`.

The separate stable prefix is intentional: reusing /opt/dynamo/nixl would prevent the generic UCX aliases from finding their hash-mangled libuct/libucm dependencies and module directory. It is only a symlink and does not duplicate the libraries, so I kept the two mechanisms separate.

container/templates/sglang_runtime.Dockerfile
Chokoyo: Verified the directory contents, the compatibility link points to nixl_cu13.libs, not .nixl_cu13.mesonpy.libs. In the current wheel, that directory contains hash-mangled auditwheel dependencies; the only generic names are the two UCX aliases added here, so libstream.so/libserdes.so are not exposed through this path.

KrishnanPrash (approved):

container/deps/sglang/discover_nixl_ucx_layout.py
dmitry-tokarev-nv: You're right and I was wrong here — thanks for going and checking the actual layout.

I'd extrapolated from `install_nixl_from_wheel.sh:95`'s `.nixl_cu${cuda_major}.mesonpy.libs` and assumed the compat link targeted that same directory. It doesn't, and the two are distinct by construction rather than by coincidence: meson-python emits `.<name>.mesonpy.libs/` for the project's own libraries, auditwheel emits `<name>.libs/` for the vendored dependency closure. The `69c5b315` build log confirms your reading — `... in /usr/local/lib/python3.12/dist-packages/nixl_cu13.libs`.

Keeping the two prefixes separate is correct, and the `$ORIGIN` reason you give for not reusing `/opt/dynamo/nixl` is the right one. Withdrawing the consolidation suggestion.

container/templates/sglang_runtime.Dockerfile
dmitry-tokarev-nv: Same correction as the other thread — my premise was wrong, and your check settles it.

`libstream.so` / `libserdes.so` live in `.nixl_cu13.mesonpy.libs`, not in the `nixl_cu13.libs` that this `LD_LIBRARY_PATH` entry points at, so they were never exposed by it. The build log confirms the target directory.

The comment you added reads accurately, and the property is structural rather than incidental: auditwheel hash-mangles everything it vendors, and the `ucx/` modules that *do* carry generic names sit in a subdirectory, which `LD_LIBRARY_PATH` doesn't search. Nothing further needed from me on this one.

container/deps/sglang/install_nixl_ucx_compat.sh
dmitry-tokarev-nv: Confirmed on the `69c5b315` run, and the ordering change makes it self-proving: because the gate now runs *before* alias installation, `sglang-runtime / Build multi-arch cuda13.0` passing is itself evidence the transport was discovered. Both architectures logged it:

```
#132 [linux/amd64 pre_runtime 16/29] ...
#132 0.733 nixl-ucx-compat: installed 2 aliases for nixl-cu13 UCX 1.21.0 in /usr/local/lib/python3.12/dist-packages/nixl_cu13.libs
#160 [linux/arm64 pre_runtime 16/29] ...
#160 0.805 nixl-ucx-compat: installed 2 aliases for nixl-cu13 UCX 1.21.0 in /usr/local/lib/python3.12/dist-packages/nixl_cu13.libs
```

That is the fail-closed behaviour doing its job on a path with no arm64 GPU lane to back it up. Thanks for taking it.

container/deps/sglang/install_nixl_ucx_compat.sh
dmitry-tokarev-nv: The `ctypes` / `ucp_get_version` probe works, and it answers the `nixl#1668` question directly — both architectures on `69c5b315` printed **UCX 1.21.0**, so the SGLang wheel is clear of the 1.20.0 hang that `trtllm_runtime.Dockerfile` works around. Much better to have that in the build log than in a PR description.

Worth noting for the record that it loads cleanly in a build container with no driver present, which was the risk I'd have raised against a `dlopen` at build time. The multi-arch build passing covers it.

One optional increment, now that the version is in hand: assert a floor rather than only printing it. 1.19 is the principled threshold — it is exactly where DYN-3695's `ucx_utils.cpp:444] UCX version is less than 1.19, CUDA support is limited` fires, i.e. below it the failure this PR fixes comes back. One line, and it turns a log entry into a build failure. Entirely your call; not blocking.

dmitry-tokarev-nv:
Follow-up after `69c5b315`. All seven threads addressed, and the multi-arch build log settles the substantive ones. Recording the evidence in one place, since it is doing a lot of work:

```
#132 [linux/amd64 pre_runtime 16/29] ...
#132 0.733 nixl-ucx-compat: installed 2 aliases for nixl-cu13 UCX 1.21.0 in /usr/local/lib/python3.12/dist-packages/nixl_cu13.libs
#160 [linux/arm64 pre_runtime 16/29] ...
#160 0.805 nixl-ucx-compat: installed 2 aliases for nixl-cu13 UCX 1.21.0 in /usr/local/lib/python3.12/dist-packages/nixl_cu13.libs
```

That single line confirms three things at once: the new `ctypes` version probe works in a driver-less build container; the bundled UCX is 1.21.0, clear of `nixl#1668`'s 1.20.0 hang; and — because the no-consumer gate now precedes alias installation — the NVSHMEM transport is discovered on **both** architectures.

I also had two things wrong, corrected in the threads above: the compat link targets `nixl_cu13.libs` (auditwheel's vendored closure), not `.nixl_cu13.mesonpy.libs` (meson-python's project libraries), so neither the `LD_LIBRARY_PATH` exposure nor the duplicate-prefix concern applies.

On the new code itself, nothing blocking. Two things I checked that are easy to get wrong and that you got right: `while read … done < <(…)` keeps the loop in the current shell, so `die`'s `exit 1` actually exits — a pipe there would have silently continued past a validation failure; and `((dependency_count += 1))` is safe under `set -e` because `+= 1` returns the new value. (`((count++))` in that slot *would* abort on the first iteration — worth knowing for whoever edits it next.)

Three things still open, none of them code:

1. **`Validation` section.** `AGENTS.md` requires `Summary` and `Validation` in the description, and DYN-3695 names `test_sglang_serve_deployment[multimodal_epd_video_embedding_cache]` as its acceptance case. The description still verifies the *linkage* rather than the *outcome* — worth stating plainly that the repro now returns 200 and that the QA case passes.

2. **Flag-specificity.** Still unexplained by the stated mechanism: mixed linkage is a static property of the image, present on every launch, yet DYN-3695 records the identical launch *without* `--multimodal-embedding-cache-capacity-gb` passing on the same GB200 node. The ticket's own suggested diagnostic — rerun both with `UCX_LOG_LEVEL=debug` / `NIXL_LOG_LEVEL=DEBUG` and diff the registered memory domains — would settle it in one run, and would make the fix's scope defensible for the release rather than inferred.

3. **`release/1.4.0` cherry-pick.** DYN-3695 is filed against `1.4.0-rc.0` and this targets `main`. Is that tracked anywhere?

Process note: `69c5b315` had a full green run including `sglang-runtime / Build multi-arch cuda13.0` and `sglang-runtime / Test cuda13.0, arm64`, but the head is the subsequent main merge (`003d4b6`), whose checks are mostly skipped — it will need a fresh `/ok to test 003d4b6` before merge.

dmitry-tokarev-nv (approved):
posted follow ups which IMO are non-blocking. approved

Chokoyo:
Thanks @dmitry-tokarev-nv, added verification in PR description and opened a CP to release 1.4.0 at https://github.com/ai-dynamo/dynamo/pull/12558

PR #12574, opened 2026-08-03
fix(runtime): propagate discovery instance updates

### LINKED ISSUE #12566: [BUG]: Frontend never updates a worker's transport when it changes without a readiness dip — `list_and_watch` diffs by key only, causing permanent `Connection refused`
### Describe the Bug

On the Kubernetes discovery backend, `KubeDiscoveryClient::list_and_watch`<br>(`lib/runtime/src/discovery/kube.rs`) diffs successive discovery snapshots **by instance key****<br>****only**: it emits `DiscoveryEvent::Added` for keys that appear and `DiscoveryEvent::Removed`<br>for keys that disappear, with **no "modified" path**. The key is<br>`DiscoveryInstanceId::Endpoint { namespace, component, endpoint, instance_id }` — it does not<br>include the transport, and `instance_id = hash(pod_name)` is stable across container restarts,<br>while the worker's shared-TCP server rebinds to a **new random port** on each restart.

So when a worker's transport (`ip:port`) changes while its instance key stays present in the<br>snapshot, the client emits nothing and the frontend keeps routing to the **old, now-dead**<br>address indefinitely. Every request fails with<br>`HTTP 500 "Failed to generate completions: Connection refused (os error 111)"` (≈35 ms,<br>`output_tokens=0`). Discovery is correct throughout — the `DynamoWorkerMetadata` CR holds the<br>new address and it is reachable; only the frontend's in-memory instance set is stale. A<br>frontend restart fixes it, because the initial `list` reseeds the client from the current CR.

**When it triggers.** A normal worker restart flips pod readiness Ready→NotReady→Ready; the<br>NotReady phase removes the instance from the snapshot, so the client sees a `Removed` then a<br>fresh `Added` with the new port (self-heals). The bug requires the transport to change<br>**without the client observing a** `Removed` — a container restart fast enough (or a<br>readinessProbe lenient enough) that the endpoint never leaves the ready set, or a watch<br>reconnect that re-LISTs in one step. In that case the stale address is permanent.

**Secondary effect.** The SLA Planner reads isl/osl from the frontend's success-only<br>histograms; while every request fails these freeze, so the planner logs `isl: nan osl: nan`,<br>`throughput_reason=n/a`, and stops scaling.

**Root cause.**

* `discovery/kube.rs` `KubeDiscoveryClient::list_and_watch` (\~L417–482): builds `current`/`known`<br>maps keyed by `DiscoveryInstanceId`, computes `added = current \ known` and<br>`removed = known \ current`, emits `Added`/`Removed`, then sets `known = current`. No branch<br>compares the `DiscoveryInstance` value for keys present in both, so a transport change on a<br>surviving key yields no event.
* `discovery/mod.rs` `DiscoveryInstance::id()` → `DiscoveryInstanceId::Endpoint` is<br>`{ namespace, component, endpoint, instance_id }` — transport excluded; `instance_id` stable<br>across restarts.
* The daemon publishes the corrected transport on the CR generation change<br>(`daemon.rs` + `MetadataSnapshot::has_changes_from`); it reaches the client's watch and is<br>discarded at the key-only diff above.

### Steps to Reproduce

The trigger is "transport changes for a still-present instance key." To reproduce, keep<br>the worker pod continuously Ready and change its CR transport.

```bash
kubectl apply -f - <<'YAML'
kind: DynamoGraphDeployment
metadata: { name: stale-fe-mocker, namespace: dynamo-demo }
spec:
  components:
    - name: Frontend
      type: frontend
      replicas: 1
      podTemplate:
        spec:
          containers:
            - name: main
              image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.2.1
    - name: VllmWorker
      type: decode
      replicas: 1
      podTemplate:
        spec:
          containers:
            - name: main
              image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.2.1
              workingDir: /workspace/examples/backends/vllm
              command: ["python3", "-m", "dynamo.mocker"]
              args: ["--model-path", "Qwen/Qwen3-0.6B"]
YAML
# wait until both pods are 1/1 Ready (~30s), then:

WK=$(kubectl -n dynamo-demo get pods -o name | grep mocker-vllmworker | sed 's#pod/##')
FE(){  kubectl -n dynamo-demo get pods -o name | grep mocker-frontend | sed 's#pod/##'; }
REQ(){ kubectl -n dynamo-demo exec "$(FE)" -c main -- curl -s -m15 -o/dev/null -w '%{http_code}' \
  -X POST localhost:8000/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"Qwen/Qwen3-0.6B","messages":[{"role":"user","content":"hi"}],"max_tokens":4}'; }
PORT(){ kubectl -n dynamo-demo get dynamoworkermetadata "$WK" -o json | python3 -c \
  'import sys,json;e=json.load(sys.stdin)["spec"]["data"]["endpoints"];k=next(x for x in e if "/generate/" in x);print(e[k]["transport"]["tcp"].split("/")[0].rsplit(":",1)[1])'; }
# rewrite the CR's generate-endpoint transport port (bumps generation; never touches the pod)
SET(){ python3 - "$WK" "$1" <<'PY'
import json, subprocess, sys
ns, name, new = "dynamo-demo", sys.argv[1], sys.argv[2]
o = json.loads(subprocess.check_output(["kubectl","-n",ns,"get","dynamoworkermetadata",name,"-o","json"]))
e = o["spec"]["data"]["endpoints"]
addr = next(v["transport"]["tcp"].split("/")[0] for k, v in e.items() if "/generate/" in k)
ip = addr.rsplit(":", 1)[0]
for v in e.values():
    t = v.get("transport", {})
    if t.get("tcp", "").startswith(addr + "/"):
        t["tcp"] = ip + ":" + new + "/" + t["tcp"].split("/", 1)[1]
o["metadata"].pop("managedFields", None)
subprocess.run(["kubectl","replace","-f","-"], input=json.dumps(o), text=True, check=True)
PY
}

PA=$(PORT)                               # the worker's real port
REQ                                      # 200  (baseline)
SET 44999                                # move CR transport to a dead port; pod stays Ready
sleep 4; REQ                             # 200  <-- BUG: client ignored the change, still on $PA
kubectl -n dynamo-demo delete pod "$(FE)"; sleep 8   # restart frontend -> re-lists current CR
REQ                                      # 500  (fresh client adopted the dead port)
SET "$PA"                                # restore CR to the real port; worker + discovery healthy
sleep 5; REQ                             # 500  <-- still stuck: running client frozen on 44999
kubectl -n dynamo-demo delete pod "$(FE)"; sleep 8
REQ                                      # 200  (recovered)
```

### Expected Behavior

When a snapshot reports a different transport for an instance key the client already knows,<br>`list_and_watch` should emit an update (a `Modified`, or a `Removed`+`Added`) so the client<br>re-resolves the address — i.e. diff full `DiscoveryInstance` values, not just<br>`DiscoveryInstanceId` keys. (Defense in depth: a request failing `Connection refused` against<br>an instance could trigger re-resolution + retry.)

### Actual Behavior

```
baseline:                          transport=IP:44707  req=200
CR transport 44707 -> 44999 (dead), generation bumped, pod still Ready:
  probe:                           transport=IP:44999  req=200   <-- client still dials 44707
  restart frontend:                transport=IP:44999  req=500   <-- fresh client adopts dead port
CR transport restored to 44707 (worker + discovery healthy):
  probe:                           transport=IP:44707  req=500   <-- running client stuck on 44999
  restart frontend:                                    req=200   <-- recovered
```

Real-world symptom: every request `HTTP 500 ... Connection refused (os error 111)`,<br>`elapsed_ms≈34`, `output_tokens=0`; `dynamo_frontend_requests_total{status="error"}` advances<br>at the offered rate while `{status="success"}` and the isl/osl histograms are frozen; the<br>worker's re-registered transport is present in the CR and TCP-reachable from the frontend pod<br>the entire time; `kubectl delete pod <frontend>` restores service on the first request.

### Environment

* Dynamo 1.2.1 (operator + `vllm-runtime:1.2.1`), `DYN_DISCOVERY_BACKEND=kubernetes`
* Kubernetes v1.36.1, Calico CNI
* Reproduces with the Mocker backend (no GPU) and round-robin routing; also observed on a<br>disaggregated vLLM graph with KV routing. The bug is in the shared discovery→client path,<br>independent of backend and router mode.

### Additional Context

Not a duplicate of [#732](<https://github.com/ai-dynamo/dynamo/issues/732>) (transient "no responders", closed), ai-dynamo/dynamo#9200 (0 instances at startup),<br>or ai-dynamo/dynamo#11855 (ready-before-routable startup window).

### Screenshots

*No response*

hhzhang16:
## Overview:

Update Kubernetes discovery to compare same-ID endpoint updates and emit `Added` as an upsert when an existing instance changes. Propagate same-worker metadata changes like new transport addresses instead of only detecting added or removed worker IDs.

## Details:

Reproduced + verified bug ai-dynamo/dynamo#12566.

* persistent memory increases per watch stream, while per-update temporary allocation decreases
* runs on Kubernetes discovery updates so update frequency should be low
* Before, each watch retained a HashSet. Now, each watch retains a HashMap<DiscoveryInstanceId, DiscoveryInstance>. For endpoints/event channels, that is likely a few hundred extra bytes per matched instance.

## Related Issues

* Closes ai-dynamo/dynamo#12566

---

Open in Devin Review

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated endpoint discovery to recognize changed instances and replace outdated connection data.
  * Prevented stale connections from removing newer replacements during cleanup.
  * Improved handling of endpoint additions, changes, removals, and shutdowns.
* **Tests**
  * Added coverage for endpoint updates, replacement cancellation, unchanged instances, and safe connection cleanup.
* **Documentation**
  * Clarified that discovery events may represent newly added or changed instances.


lib/runtime/src/component/client.rs
```diff
@@ -804,8 +807,38 @@ impl Client {
...
+            &mut instances,
+            DiscoveryEvent::Added(DiscoveryInstance::Endpoint(original)),
+        );
+        apply_endpoint_discovery_event(
+            &mut instances,
+            DiscoveryEvent::Added(DiscoveryInstance::Endpoint(updated.clone())),
+        );
+
+        assert_eq!(instances, HashMap::from([(1, updated)]));
+    }
```
connorcarpenter15: Is this test really necessary? It only really tests the new `apply_endpoint_discovery_event` helper, which is identical code from before. Could probably be removed.

connorcarpenter15 (commented):
Overall LGTM, just a couple comments from me and AI.

lib/runtime/src/discovery/kube.rs
```diff
@@ -438,20 +451,10 @@ impl Discovery for KubeDiscoveryClient {
...
-                        let known_keys: HashSet<&DiscoveryInstanceId> = known.iter().collect();
-
-                        let added: Vec<&DiscoveryInstanceId> =
-                            current_keys.difference(&known_keys).copied().collect();
-
-                        let removed: Vec<DiscoveryInstanceId> = known_keys
-                            .difference(&current_keys)
-                            .map(|&id| id.clone())
-                            .collect();
+                        let (upserted, removed) = diff_discovery_instances(&known, &current);
```
connorcarpenter15: **[P1] Fast CR recreation can still bypass the fix.** The new full-instance comparison only runs after metadata_watch.changed() at [kube.rs:454](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kube.rs#L454), but the daemon first suppresses snapshots whose {instance_id: generation} map is unchanged at [metadata.rs:426](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/metadata.rs#L426-L437). A delete/recreate within the 500 ms debounce window at [daemon.rs:207](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kube/daemon.rs#L207-L225) can reuse the stable instance ID and same numeric generation while changing transport. Kubernetes assigns distinct UIDs specifically to distinguish historical occurrences of same-named objects, so excluding UID makes this gate unsound. [Kubernetes object identity documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids).

lib/runtime/src/discovery/mod.rs
```diff
@@ -1096,7 +1096,9 @@ impl DiscoveryInstanceId {
 /// Events emitted by the discovery watch stream
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DiscoveryEvent {
-    /// A new instance was added
+    /// An instance was added or its data changed.
+    ///
+    /// Consumers must replace any existing state with the same [`DiscoveryInstanceId`].
     Added(DiscoveryInstance),
```
connorcarpenter15, line 1104: **[P1] A same-ID EventChannel update permanently disables KVBM direct-ZMQ recovery.** The new contract says Added must replace same-ID state at [mod.rs:1095](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/mod.rs#L1095-L1102). However, [direct_zmq.rs:345](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/llm/src/kv_router/indexer/recovery/direct_zmq.rs#L345-L406) treats an endpoint change as an immutable-channel violation, fences it, and ignores every later upsert until a Removed arrives—which this PR intentionally does not emit for updates. Either handle it as a generation-safe replacement or explicitly narrow the upsert contract so EventChannel endpoints cannot change under the same ID.

lib/runtime/src/discovery/kube.rs
```diff
@@ -388,8 +404,8 @@ impl Discovery for KubeDiscoveryClient {
                 }
             }
 
-            // Track known instances by their unique ID
-            let mut known: HashSet<DiscoveryInstanceId> = initial.into_keys().collect();
+            // Retain the full instance state so same-ID updates are detectable.
```
connorcarpenter15: **[P2] Per-watch memory now scales with complete discovery payload size.** Each Kubernetes watch permanently retains owned DiscoveryInstance values at [kube.rs:371](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kube.rs#L371-L408), then rebuilds and deep-compares another complete map on every snapshot at [kube.rs:439](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kube.rs#L439-L504). This includes model-card and event-source JSON, not merely the few-hundred-byte endpoint records described in the PR. Prefer compact owner revisions such as (uid, generation) or shared instance storage, and measure broad model-watch memory at fleet scale.

lib/runtime/src/discovery/kube.rs
```diff
@@ -21,10 +21,27 @@ use crate::discovery::{
 use anyhow::Result;
 use async_trait::async_trait;
 use kube::{Api, Client as KubeClient, api::DeleteParams};
-use std::collections::HashSet;
+use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
+fn diff_discovery_instances(
```
connorcarpenter15: **[P3] The shared upsert policy is duplicated across backends.** Equivalent diff logic now exists in [Kubernetes](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kube.rs#L28-L43), [mock](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/mock.rs#L249-L268), and [KV resync](https://github.com/ai-dynamo/dynamo/blob/caae60f1782e4ecb9ad7771f892fd7c4882f8e1d/lib/runtime/src/discovery/kv_store.rs#L311-L334). Extract a discovery-core helper so the newly documented contract cannot drift again.

connorcarpenter15 (commented):
Inline findings:

connorcarpenter15 (approved):

PR #12576, opened 2026-08-03
feat(llm): stream guided tool calls by default and route Qwen3 to v2

keivenchang:
#### Overview:

Guided tool calls now stream as they are generated on the default path, with no experimental flag. This affects named and required tool choices whose backend installs guided JSON. Previously the client received one terminal tool-call chunk only after the full call had been generated. `tool_choice="auto"` stays on native parsing and is unchanged.

Qwen3 also routes through the v2 unified parser under `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2`, so reasoning that follows a tool call stays after it.

> [!IMPORTANT]
> Scope changed from the original version of this PR. Guided streaming is not gated by `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` and is not Qwen3-only. Guided calls outside the Qwen3 parser-v2 route run through the v1 jail; Qwen3 under parser v2 runs through the unified parser. Both paths now honor `DYN_ENABLE_GUIDED_TOOL_STREAMING`; setting it to `0` buffers guided calls to completion without reverting code.

#### Example (before -> after):

Streaming request, `tool_choice="required"`, flag unset. Measured against a served model, long payload:

```
before -- 0 tool-call chunks, one terminal burst at the end of the stream

after  -- 169 chunks; first at 9.2% of the stream; argument bytes spread
          across every decile of stream time
```

Arrival histograms over deciles of stream time. A buffering path reads `[0,0,0,0,0,0,0,0,0,1]`:

| tool_choice | flag | deltas | first delta | delivery span | histogram |
|---|---|---|---|---|---|
| required | off | 169 | 9.2% in | 90.2% | `[3,20,21,23,20,16,17,16,16,17]` |
| named | off | 181 | 1.0% in | 98.7% | `[18,17,15,23,20,15,16,21,18,18]` |
| required | on | 229 | 19.9% in | 79.4% | `[0,2,36,29,27,26,31,23,25,30]` |
| named | on | 260 | 1.4% in | 98.4% | `[25,26,24,26,28,27,25,29,24,26]` |

Every reassembled payload parses as valid JSON, exactly once.

#### Details:

- `postprocessor_parsing_stream_with_constraint` reads the installed constraint and `DYN_ENABLE_GUIDED_TOOL_STREAMING` once per request, then passes the same decision to the jail or unified-v2 path. Before this wiring, three v2 cases produced the same 28 / 23 / 96 argument-fragment chunks with the flag unset and set to `0`. Unified rollback selects `RecoverAsText`, which buffers to completion, rather than `Reject`, which would return a typed error for malformed output.
- `apply_tool_choice_guided_decoding` reports the constraint it installed instead of a bool, so the postprocessor does not infer guided JSON from `tool_choice` alone. That keeps native fallback working when a backend ignores its grammar and prevents a forced request with no installed JSON schema from streaming native markup into `function.arguments`.
- `parallel_tool_calls=false` now constrains generation with `maxItems: 1`; the HTTP filter remains as defense in depth. Guided-streaming support uses published `dynamo-parsers 8.1.0` and `dynamo-parsers-v2 0.3.2` releases, with no frontend-crates path dependency.

#### Known issues:

- `DIS-2779`: a zero-argument tool under guided decoding generates whitespace until `max_tokens` and returns no usable call. Main reproduces this; this PR does not cause it.
- `DIS-2780`: a tool call cut off mid-string by `max_tokens` can return `arguments` that do not parse. This PR causes that behavior. On main, the same input drops the tool call and leaks raw tool markup into `content`.

#### Verification:

The streaming measurements above cover named and required guided calls. Tests cover the shared rollback decision and the unified buffer-to-completion path.

#### Where:

- `lib/llm/src/preprocessor.rs`, `lib/llm/src/preprocessor/{tool_choice,structural_tag}.rs`
- `lib/llm/src/protocols/openai/chat_completions/{tool_parser_v2,unified_parser,aggregator}.rs`
- `lib/llm/src/protocols/openai/tools.rs`
- `lib/llm/src/http/service.rs`, `lib/llm/src/http/service/{openai,anthropic}.rs`
- `lib/llm/src/discovery/worker_set.rs`
- `lib/llm/tests/{postprocessor_parsing_stream,tool_choice,aggregators}.rs`

## Summary by CodeRabbit

- **New Features**
  - Improved structured tool-call handling for named, required, optional, and disabled tool choices.
  - Added better support for unified reasoning and tool parsing across supported models.
  - Improved streaming behavior, including response ordering, usage updates, and deferred content handling.
  - Added support for limiting required tool calls when parallel calls are disabled.

- **Bug Fixes**
  - Invalid tool configurations now return clear request validation errors.
  - Improved recovery when responses contain native tool-call markup or truncated content.
  - Prevented tool calls from appearing when tools are unavailable or disabled.

- **Documentation**
  - Clarified experimental parser configuration and its batch and streaming behavior.

### DIFF
--- a/lib/llm/src/http/service.rs
+++ b/lib/llm/src/http/service.rs
@@ -81,7 +81,8 @@ fn apply_request_tool_call_parsing_options(
     )?;
     Ok(parsing_options
         .with_guided_tool_constraint(guided_tool_constraint)
-        .with_tool_call_parsing_enabled(tool_call_parsing_enabled))
+        .with_tool_call_parsing_enabled(tool_call_parsing_enabled)
+        .with_tools(converted_tools))
 }
 
 #[cfg(test)]
--- a/lib/llm/src/http/service.rs
+++ b/lib/llm/src/http/service.rs
@@ -43,23 +43,447 @@ pub use metrics::Metrics;
 
 use crate::{
     preprocessor::OpenAIPreprocessor,
-    protocols::openai::{
-        ParsingOptions, chat_completions::NvCreateChatCompletionRequest,
-        chat_completions::tool_parser_v2::batch_tool_choice_eligible,
-    },
+    protocols::openai::{ParsingOptions, chat_completions::NvCreateChatCompletionRequest},
 };
+
+use dynamo_protocols::types::ChatCompletionToolChoiceOption;
+use dynamo_runtime::error::DynamoError;
 
 /// Apply the request-level tool-call gates shared by the HTTP protocol handlers.
 fn apply_request_tool_call_parsing_options(
     parsing_options: ParsingOptions,
     request: &NvCreateChatCompletionRequest,
-) -> ParsingOptions {
+) -> Result<ParsingOptions, DynamoError> {
     let tool_call_parsing_enabled = OpenAIPreprocessor::tool_call_parsing_enabled(request);
-    parsing_options
-        .with_experimental_v2_batch_eligible(batch_tool_choice_eligible(
-            request.inner.tool_choice.as_ref(),
-        ))
-        .with_tool_call_parsing_enabled(tool_call_parsing_enabled)
+    let tool_choice = request
+        .inner
+        .tool_choice
+        .as_ref()
+        .unwrap_or(&ChatCompletionToolChoiceOption::Auto);
+    let converted_tool_choice = crate::preprocessor::tool_choice::convert_tool_choice(tool_choice);
+    let tools = request.inner.tools.as_deref().unwrap_or(&[]);
+    let converted_tools = crate::preprocessor::tool_choice::convert_tools(tools);
+    let uses_structural_tag = crate::preprocessor::structural_tag::structural_tag_decision(
+        parsing_options.tool_call_parser.as_deref(),
+        &converted_tool_choice,
+        &converted_tools,
+        request.inner.parallel_tool_calls,
+        parsing_options.structural_tag_mode,
+        parsing_options.structural_tag_scope,
+        parsing_options.exclude_tools_when_tool_choice_none,
+    )?
+    .is_required();
+    let guided_tool_constraint = crate::preprocessor::tool_choice::guided_tool_constraint(
+        request,
+        parsing_options.tool_call_parser.as_deref(),
+        parsing_options.reasoning_parser.as_deref(),
+        uses_structural_tag,
+    )?;
+    Ok(parsing_options
+        .with_guided_tool_constraint(guided_tool_constraint)
+        .with_tool_call_parsing_enabled(tool_call_parsing_enabled))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::protocols::openai::GuidedToolConstraint;
+    use serde_json::{Value, json};
+
+    fn request(tool_choice: Value) -> NvCreateChatCompletionRequest {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": tool_choice,
+            "tools": [{
+                "type": "function",
+                "function": {
+                    "name": "get_weather",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"location": {"type": "string"}},
+                        "required": ["location"]
+                    }
+                }
+            }]
+        });
+        serde_json::from_value(value).expect("request must deserialize")
+    }
+
+    // A Kimi K2 pair with a forced/named tool_choice must resolve to the real
+    // structural-tag decision computed by the single shared owner,
+    // `structural_tag::structural_tag_decision`, not to a second, independently
+    // reconstructed decision. `apply_tool_choice_structural_tag` is the in-process
+    // preprocessing owner that consults the same shared function; this HTTP-layer
+    // helper must agree with it, including when the parser registry does or does not
+    // actually have a builder for the parser (see `structural_tag_decision`'s own
+    // doc comment for why eligibility and registry-availability are bundled into one
+    // value rather than checked independently).
+    #[test]
+    fn kimi_k2_required_resolves_to_the_real_structural_tag_decision() {
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k2".to_string()),
+            ..Default::default()
+        };
+        let result =
+            apply_request_tool_call_parsing_options(parsing_options, &request(json!("required")))
+                .expect("kimi_k2 + required must resolve a constraint");
+        assert_eq!(
+            result.guided_tool_constraint,
+            GuidedToolConstraint::StructuralTag,
+            "kimi_k2 + required must use the intrinsic structural tag, not a reconstructed JSON schema"
+        );
+    }
+
+    #[test]
+    fn kimi_k2_named_resolves_to_the_real_structural_tag_decision() {
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k2".to_string()),
+            ..Default::default()
+        };
+        let named = json!({"type": "function", "function": {"name": "get_weather"}});
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request(named))
+            .expect("kimi_k2 + named must resolve a constraint");
+        assert_eq!(
+            result.guided_tool_constraint,
+            GuidedToolConstraint::StructuralTag,
+            "kimi_k2 + a named tool choice must use the intrinsic structural tag, not a reconstructed JSON schema"
+        );
+    }
+
+    // Non-structural-tag parsers must be unaffected by this fix: they still get a
+    // reconstructed JSON-schema constraint for a forced tool_choice.
+    #[test]
+    fn non_structural_tag_parser_still_gets_guided_json_required() {
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("hermes".to_string()),
+            ..Default::default()
+        };
+        let result =
+            apply_request_tool_call_parsing_options(parsing_options, &request(json!("required")))
+                .expect("hermes + required must resolve a constraint");
+        assert_eq!(
+            result.guided_tool_constraint,
+            GuidedToolConstraint::GuidedJsonRequired
+        );
+    }
+
+    // The registry-builder gap this test suite exists to close: a non-Kimi parser
+    // (qwen3_coder, which the parser registry does register a structural-tag builder
+    // for) with a forced tool_choice must still resolve to `StructuralTag` once the
+    // operator has globally enabled structural-tag mode — matching the real
+    // preprocessing path (`apply_tool_choice_structural_tag`'s
+    // `structural_tag_mode != Off` branch), not the narrower Kimi-only reconstruction
+    // this helper used before this fix.
+    #[test]
+    fn operator_enabled_mode_resolves_structural_tag_for_a_non_kimi_parser() {
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("qwen3_coder".to_string()),
+            structural_tag_mode: crate::local_model::runtime_config::StructuralTagMode::On,
+            ..Default::default()
+        };
+        let result =
+            apply_request_tool_call_parsing_options(parsing_options, &request(json!("required")))
+                .expect("qwen3_coder + required must resolve a constraint");
+        assert_eq!(
+            result.guided_tool_constraint,
+            GuidedToolConstraint::StructuralTag,
+            "an operator-enabled structural_tag_mode must apply to any registry-supported \
+             parser, not only the Kimi-intrinsic case"
+        );
+    }
+
+    // With the operator default (`structural_tag_mode = Off`), the same non-Kimi
+    // parser must NOT get a structural tag — confirms the mode gate above is real,
+    // not a permanently-on regression.
+    #[test]
+    fn operator_default_mode_off_does_not_resolve_structural_tag_for_a_non_kimi_parser() {
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("qwen3_coder".to_string()),
+            ..Default::default()
+        };
+        let result =
+            apply_request_tool_call_parsing_options(parsing_options, &request(json!("required")))
+                .expect("qwen3_coder + required (mode off) must resolve a constraint");
+        assert_ne!(
+            result.guided_tool_constraint,
+            GuidedToolConstraint::StructuralTag
+        );
+    }
+
+    // `tool_choice: none` must never leave the client able to observe a tool call or
+    // a tool-call finish reason, regardless of the operator's structural-tag mode or
+    // exclusion setting. The real preprocessing path may install a ban tag at
+    // generation time for `none` (which is a request-time generation constraint on
+    // the engine, tracked separately from this field), but that must not surface as
+    // a guided-tool-output-parsing constraint here: `with_tool_call_parsing_enabled`
+    // (called unconditionally by this function, below) always suppresses tool-call
+    // output and resets `guided_tool_constraint` to `None` whenever tool-call parsing
+    // is disabled, which `tool_choice: none` always is. Covering both the ban-enabled
+    // and default-exclusion configurations proves this safety invariant holds either
+    // way, not just in the configuration that happens not to trigger the ban path.
+    #[test]
+    fn tool_choice_none_never_exposes_tool_calls_regardless_of_ban_tag_configuration() {
+        for exclude_tools_when_tool_choice_none in [true, false] {
+            let parsing_options = ParsingOptions {
+                tool_call_parser: Some("qwen3_coder".to_string()),
+                structural_tag_mode: crate::local_model::runtime_config::StructuralTagMode::On,
+                exclude_tools_when_tool_choice_none,
+                ..Default::default()
+            };
+            let result =
+                apply_request_tool_call_parsing_options(parsing_options, &request(json!("none")))
+                    .expect("tool_choice=none must resolve a constraint");
+            assert_eq!(
+                result.guided_tool_constraint,
+                GuidedToolConstraint::None,
+                "tool_choice=none must never report a guided-tool-output constraint \
+                 (exclude_tools_when_tool_choice_none={exclude_tools_when_tool_choice_none})"
+            );
+            assert!(
+                result.suppress_tool_calls,
+                "tool_choice=none must suppress tool calls \
+                 (exclude_tools_when_tool_choice_none={exclude_tools_when_tool_choice_none})"
+            );
+        }
+    }
+
+    // `guided_tool_constraint` must validate the forced choice against the real
+    // `tools` list via `get_json_schema_from_tools` (the same function
+    // `apply_tool_choice_guided_decoding` calls at request-preprocessing time)
+    // instead of unconditionally installing a JSON-schema constraint. A
+    // `tool_choice: "required"` request with an empty `tools` list has nothing to
+    // constrain against and must be rejected, not silently resolved to
+    // `GuidedJsonRequired`.
+    #[test]
+    fn required_tool_choice_with_empty_tools_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": "required",
+            "tools": []
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("hermes".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "tool_choice=required with no tools must not resolve to a bogus GuidedJsonRequired"
+        );
+    }
+
+    // A named tool_choice that references a tool absent from `tools` is equally
+    // unvalidatable and must also be rejected rather than resolved to a
+    // constraint for a tool the request never declared.
+    #[test]
+    fn named_tool_choice_for_absent_tool_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": {"type": "function", "function": {"name": "does_not_exist"}},
+            "tools": [{
+                "type": "function",
+                "function": {
+                    "name": "get_weather",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"location": {"type": "string"}},
+                        "required": ["location"]
+                    }
+                }
+            }]
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("hermes".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "a named tool_choice for a tool absent from `tools` must not resolve to a \
+             bogus constraint"
+        );
+    }
+
+    // Regression for the CodeRabbit finding on PR #12576 (review comment 3836534587,
+    // filed against this file): `structural_tag_decision` returned `Required`
+    // without checking `tools`, so `guided_tool_constraint` skipped
+    // `get_json_schema_from_tools` entirely for the STRUCTURAL-TAG branch (it
+    // short-circuits to `Ok(StructuralTag)` whenever `uses_structural_tag` is true,
+    // without building or validating anything). The two `hermes` tests above only
+    // ever exercised the non-structural-tag JSON-schema fallback, which already
+    // validated correctly — they never covered the actual bug. Kimi K2 is the
+    // structural-tag-intrinsic parser, so this reproduces the real gap.
+    #[test]
+    fn kimi_k2_required_with_empty_tools_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": "required",
+            "tools": []
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k2".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "kimi_k2 + required with no tools must not resolve to StructuralTag"
+        );
+    }
+
+    #[test]
+    fn kimi_k2_named_tool_choice_for_absent_tool_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": {"type": "function", "function": {"name": "does_not_exist"}},
+            "tools": [{
+                "type": "function",
+                "function": {
+                    "name": "get_weather",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"location": {"type": "string"}},
+                        "required": ["location"]
+                    }
+                }
+            }]
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k2".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "kimi_k2 + a named tool_choice for a tool absent from `tools` must not \
+             resolve to StructuralTag"
+        );
+    }
+
+    #[test]
+    fn kimi_k3_named_tool_choice_for_absent_tool_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": {"type": "function", "function": {"name": "does_not_exist"}},
+            "tools": [{
+                "type": "function",
+                "function": {
+                    "name": "get_weather",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"location": {"type": "string"}},
+                        "required": ["location"]
+                    }
+                }
+            }]
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k3".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "kimi_k3 + a named tool_choice for a tool absent from `tools` must not \
+             resolve to StructuralTag"
+        );
+    }
+
+    #[test]
+    fn kimi_k3_required_with_empty_tools_is_rejected() {
+        let value = json!({
+            "model": "test-model",
+            "messages": [{"role": "user", "content": "test"}],
+            "tool_choice": "required",
+            "tools": []
+        });
+        let request: NvCreateChatCompletionRequest =
+            serde_json::from_value(value).expect("request must deserialize");
+        let parsing_options = ParsingOptions {
+            tool_call_parser: Some("kimi_k3".to_string()),
+            ..Default::default()
+        };
+        let result = apply_request_tool_call_parsing_options(parsing_options, &request);
+        assert!(
+            result.is_err(),
+            "kimi_k3 + required with no tools must be rejected before the XTML path"
+        );
+    }
+
+    // Same gap, operator-enabled path: qwen3_coder only gets a structural tag when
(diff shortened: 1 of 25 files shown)


nv-kmcgill53:
I have a newbie question to make sure I understand the scope of this change:

The reason for not touching the `batch` branch, and only the streaming branch, is that we would otherwise have to breakup a single response back to the user into `N` different responses depending on the sequence of thinking and tool calling from the model.

>Model emits: <think>A</think> → tool_call(get_weather) → <think>B</think> → "Done."

As stated in your PR, if this is in batch mode then we still get `reasoning="AB"` because we would have to split this into at least 3 different messages back to the client for them to recover the sequence: A -> tool -> B+content.

keivenchang:
@nv-kmcgill53 yep, that's it.

For example, if the model produces reasoning A, then a tool call, then reasoning B, batch can only return one message with `reasoning="AB"` and the tool call beside it. There's nowhere in that message to record that the call happened between A and B.

Streaming preserves the order because it returns chunks in sequence. Preserving that order in batch would require a different API response: multiple messages or an ordered parts list. That's outside this parser PR.

So batch still returns `reasoning="AB"`. This PR didn't regress that; the comment above `parse_complete` in `lib/llm/src/protocols/openai/chat_completions/unified_parser.rs:470` documents the existing limit.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -400,27 +520,74 @@ impl DeltaAggregator {
...
+                            } else {
+                                tracing::warn!(
+                                    family,
+                                    why = "guided_json_produced_zero_calls_under_tool_choice_required",
+                                    original_bytes = choice.text.len(),
+                                    "guided-JSON parse returned zero tool calls; preserving original text"
+                                );
                             }
-                            choice.text = content;
                         }
```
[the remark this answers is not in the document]
keivenchang: not fixing — unreachable by construction. `suppress_tool_calls` only ever flips true inside `with_tool_call_parsing_enabled(false)` (openai.rs:457), and that same call resets `guided_tool_constraint` to `None` right below it, so `installs_guided_json()` can't be true when suppression is on.

lib/llm/src/preprocessor/tool_choice.rs
```diff
@@ -189,9 +195,92 @@ fn convert_tools(tools: &[ChatCompletionTool]) -> Vec<ToolDefinition> {
...
 }
 
+/// The guided-tool constraint a request implies, given what the structural-tag stage
+/// already decided.
+///
+/// Direct postprocessor callers and remote frontends use this compatibility helper
+/// when they did not run request preprocessing. The worker streaming path carries the
+/// enum returned by `apply_tool_choice_guided_decoding`; topology-B aggregation may
+/// reconstruct a structural-tag request as guided JSON, so its complete-output parser
+/// also checks the generated wire shape.
```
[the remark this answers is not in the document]
keivenchang: not fixing — this is a permanent dual-path helper for topology-B/remote frontends that skip request preprocessing entirely (see the doc comment right above it), not a version-migration shim. Nothing plans to retire that call path, so there's no removal condition to state.

lib/llm/src/preprocessor.rs
```diff
@@ -4900,6 +5054,198 @@ impl OpenAIPreprocessor {
...
+                            }
+                        }
+                        if let Some(reasoning) = reasoning {
+                            choice
+                                .delta
+                                .reasoning_content
+                                .get_or_insert_default()
+                                .push_str(&reasoning);
+                        }
+                    }
```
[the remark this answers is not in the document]
keivenchang: not fixing — verified unreachable on this code path. `carries_parts` sets `saw_visible_output=true` before the guarded `state.drain()` runs, so the `release_split()` block a few lines up already drains `pending_reasoning`/`pending_content` into a prefix chunk first; by the time the cited `state.drain()` call would run, both fields are already empty. Confirmed by diffing full stream output byte-for-byte with and without the guard — identical. The sibling v1 guard you're citing as precedent (line ~4925) is real because that function has no equivalent release-on-Parts step; this one already does.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -400,27 +520,74 @@ impl DeltaAggregator {
...
+                            if !calls_is_empty {
+                                choice.text = content;
+                            } else {
+                                tracing::warn!(
+                                    family,
+                                    why = "guided_json_produced_zero_calls_under_tool_choice_required",
+                                    original_bytes = choice.text.len(),
+                                    "guided-JSON parse returned zero tool calls; preserving original text"
+                                );
                             }
```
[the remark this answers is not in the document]
keivenchang: fixed in 9534dfc7 — plain Muse zero-call responses now return stripped content instead of raw control markup

lib/llm/src/http/service.rs
```diff
@@ -43,23 +43,273 @@ pub use metrics::Metrics;
...
+        parsing_options.structural_tag_scope,
+        parsing_options.exclude_tools_when_tool_choice_none,
+    )
+    .is_required();
+    let guided_tool_constraint = crate::preprocessor::tool_choice::guided_tool_constraint(
+        request,
+        parsing_options.tool_call_parser.as_deref(),
+        parsing_options.reasoning_parser.as_deref(),
+        uses_structural_tag,
+    )?;
```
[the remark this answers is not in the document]
keivenchang: fixed in 9534dfc7 — forced choices are validated against the request tools before structural-tag or XTML routing

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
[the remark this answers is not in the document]
keivenchang: fixed in 9534dfc7 — native Muse fallback now keeps recovered reasoning and stripped content when it finds zero calls

lib/llm/tests/aggregators.rs
```diff
@@ -424,3 +746,138 @@ async fn test_muse_unified_batch_finalize_suppresses_calls_when_not_eligible() {
         );
     }
 }
+
+// ===================================
+// DIS-2544 stage5b regression tests (Fix A/B/C)
+// ===================================
```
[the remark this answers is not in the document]
keivenchang: fixed in 9534dfc7 — removed the internal tracker reference from the public source comment. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

nv-kmcgill53:
@keivenchang Thanks for the confirmation. I don't see anything structurally bad with this PR while it's in draft. I think the architecture of how everything is laid out is fine. I'll revisit this when you mark this PR "ready for review"

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
```diff
@@ -0,0 +1,3098 @@
...
+    tool_call_parser: Option<&str>,
+    reasoning_parser: Option<&str>,
+) -> Option<&'static str> {
+    // A request that silently fell back to the split path is otherwise
+    // indistinguishable from one the unified parser handled — the two produce the
+    // same shape of response. One INFO line per stream lets an operator answer
+    // "did v2 actually run?" from the log instead of inferring it from a build
+    // pin, which is exactly the ambiguity that cost real debugging time here.
+    let configured = configured_family(tool_call_parser, reasoning_parser);
+    tracing::info!(
```
rmccorm4, line 123: This runs on every non-Muse stream, including requests with no Qwen parser pair, so the default INFO level creates one routine log record per request. At serving rates this adds avoidable log volume and subscriber/formatting work on the request hot path. Please make this DEBUG/TRACE, or emit a one-time INFO event when the model parser configuration is registered.

lib/llm/src/preprocessor.rs
```diff
@@ -3238,21 +3336,43 @@ impl OpenAIPreprocessor {
...
+                    })
+                    .collect()
+            });
             let unified: Pin<Box<dyn Stream<Item = _> + Send>> =
-                Box::pin(tool_parser_v2::apply_unified_stream(
+                Box::pin(unified_parser::apply_stream_with_constraint(
                     stream,
                     tool_definitions,
+                    guided_tool_constraint,
+                    unified_parser::stream_prefill(family, prompt_injected_reasoning),
```
KrishnanPrash, line 3381: *AI Generated Review Comment*

**Guided JSON is swallowed as reasoning under an injected-reasoning prompt**

BEFORE:
```rust
unified_parser::stream_prefill(family, prompt_injected_reasoning),
```
AFTER:
```rust
// Guided JSON forbids `</think>`; never start inside reasoning.
let prefill = if guided_tool_constraint.installs_guided_json() {
    UnifiedParserStartingState::None
} else {
    unified_parser::stream_prefill(family, prompt_injected_reasoning)
};
```

Guided decoding forbids the model to emit `</think>`. When the rendered prompt ends inside a thought, this call starts the unified parser in the `Reasoning` state. The parser then waits for a closer that cannot arrive, and the whole guided payload lands in `reasoning_content`. I reproduced this at the PR head: `tool_choice="required"`, prefill `Reasoning`, input `[{"name":"get_weather","parameters":{"city":"Paris"}}]` produced zero tool calls. Qwen3-Thinking style templates append `<think>` to the generation prompt, so this combination is reachable under the flag. The v1 path handled this shape with `bypass_bare_guided_json`. The override above stops the swallow, but it classifies a `reasoning</think>JSON` turn as content. A fuller fix detects the output shape as v1 does, or teaches the guided cursor to accept a `Reasoning` start.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
```diff
@@ -0,0 +1,3098 @@
...
+///
+/// Reasoning spans are concatenated because the non-streaming message schema has ONE
+/// `reasoning_content` string — the ordering the unified parser recovered survives only
+/// on the streaming path, which is where a client can act on it.
+pub(crate) fn parse_complete(
+    family: &str,
+    content: &str,
+    guided_tool_constraint: &GuidedToolConstraint,
+) -> anyhow::Result<CompleteOutput> {
+    let mut parser = create_unified_parser_for_family(family, &[])?;
```
KrishnanPrash: *AI Generated Review Comment*

**Batch parsing drops the tool schemas, so argument types diverge from streaming**

BEFORE:
```rust
let mut parser = create_unified_parser_for_family(family, &[])?;
```
AFTER:
```rust
pub(crate) fn parse_complete(
    family: &str,
    content: &str,
    tools: &[Tool],
    constraint: &GuidedToolConstraint,
) -> anyhow::Result<CompleteOutput> {
    let mut parser = create_unified_parser_for_family(family, tools)?;
```

The streaming path gives the parser the request tools, so the XML scanner types each argument from its schema. The batch path constructs the parser with an empty tool list, because `ParsingOptions` does not carry tools. The same output then types differently. I ran `<parameter=count>42</parameter>` through `parse_complete` at the PR head and got `{"count":"42"}`; the streaming path returns `{"count":42}`. This breaks the module's stated goal that stream/batch parity is structural. Thread the request tools from the HTTP handlers into `parse_complete`; the unary handlers hold the full request. The muse branch (`parse_complete_unified` with `None`) has the same gap, but it predates this PR.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
```diff
@@ -0,0 +1,3098 @@
...
+
+/// Which channel the prompt opened, inferred from complete output text.
+///
+/// The batch path has no prompt in hand, only what the model produced, so the channel
+/// state is read back off the text: a `<think>` opener means the model opened reasoning
+/// itself; a bare `</think>` with no opener means the prompt had already opened it; and
+/// neither marker means reasoning never ran for this turn.
+fn detect_prefill(family: &str, content: &str) -> anyhow::Result<UnifiedParserStartingState> {
+    match family {
+        QWEN3_UNIFIED_FAMILY => Ok(if contains_unquoted_marker(content, "<think>") {
```
KrishnanPrash: *AI Generated Review Comment*

**`detect_prefill` compares marker presence, not marker position**

BEFORE:
```rust
Ok(if contains_unquoted_marker(content, "<think>") {
    UnifiedParserStartingState::None
} else if contains_unquoted_marker(content, "</think>") {
    UnifiedParserStartingState::Reasoning
```
AFTER (sketch; `first_unquoted` is a position-returning variant of the scanner):
```rust
let opener = first_unquoted(content, "<think>");
let closer = first_unquoted(content, "</think>");
Ok(match (opener, closer) {
    (Some(o), Some(c)) if c < o => UnifiedParserStartingState::Reasoning,
    (Some(_), _) => UnifiedParserStartingState::None,
    (None, Some(_)) => UnifiedParserStartingState::Reasoning,
    (None, None) => UnifiedParserStartingState::Response,
})
```

The function tests whether an opener exists anywhere before it tests for a closer. It does not compare positions. For `secret</think>answer<think>second thought</think>tail` it returns `None`, so the parser treats the leading prompt-opened reasoning as visible content, and reasoning text reaches the client. Compare the first unquoted positions instead: a closer before the first opener means the prompt opened reasoning. I confirmed the wrong result with a test at the PR head.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
```diff
@@ -0,0 +1,3098 @@
...
+            if escaped {
+                escaped = false;
+            } else if character == '\\' {
+                escaped = true;
+            } else if character == active_quote {
+                quote = None;
+            }
+            continue;
+        }
+        let apostrophe = character == '\''
```
KrishnanPrash, line 295: *AI Generated Review Comment*

**A possessive apostrophe opens a phantom quote span and hides real markers**

BEFORE:
```rust
let apostrophe = character == '\''
    && content[..offset]
        .chars()
        .next_back()
        .is_some_and(char::is_alphanumeric)
    && content[offset + character.len_utf8()..]
        .chars()
        .next()
        .is_some_and(char::is_alphanumeric);
```
AFTER:
```rust
// A real opening quote never directly follows a letter or a digit.
let apostrophe = character == '\''
    && content[..offset]
        .chars()
        .next_back()
        .is_some_and(char::is_alphanumeric);
```

The exemption requires an alphanumeric character on both sides of the apostrophe. A possessive that ends a word, such as `James'`, fails the test, so the scan opens a quote span there. Any later apostrophe, such as the one in `isn't`, closes that span, and every marker between the two is hidden. For `James' secret </think>answer isn't final` the scanner reports no `</think>`, `detect_prefill` selects `Response`, and the reasoning is served as content. I confirmed this with a test at the PR head. A real opening quote does not directly follow a letter, so it is safe to exempt every `'` that follows an alphanumeric character.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
```diff
@@ -0,0 +1,3098 @@
...
+                    }
+                    emitted.push(original);
+                    continue;
+                }
+
+                let state = match states.entry(original.index) {
+                    std::collections::hash_map::Entry::Occupied(entry) => entry.into_mut(),
+                    std::collections::hash_map::Entry::Vacant(entry) => {
+                        let mode = tool_output_mode(&guided_tool_constraint);
+                        match ChoiceState::new(family, &tools, prefill, mode) {
```
KrishnanPrash: *AI Generated Review Comment*

**The already-parsed detour rebuilds the parser in the request-level prefill state**

BEFORE:
```rust
let mode = tool_output_mode(&guided_tool_constraint);
match ChoiceState::new(family, &tools, prefill, mode) {
```
AFTER (sketch):
```rust
let detoured = records.contains_key(&original.index);
let resume = if detoured {
    // Reasoning may already be over.
    UnifiedParserStartingState::Response
} else {
    prefill
};
match ChoiceState::new(family, &tools, resume, mode) {
```

An already-parsed chunk removes the live `ChoiceState`. A later raw chunk rebuilds the parser here with the request-level `prefill`, which is `Reasoning` for a prompt-injected request. If the first raw run already left reasoning, the resumed run classifies all later visible text as reasoning, and no closer arrives to stop it. Every detour test passes the `None` prefill, so this combination has no coverage. The sketch above resumes in `Response`; a sharper rule derives the state from the detour chunk itself, for example `Reasoning` when it carried `reasoning_content`. If the mixed-stream shape is out of scope, document the limit and add a test that pins the current behavior.

lib/llm/src/preprocessor.rs
```diff
@@ -4062,14 +4196,35 @@ impl OpenAIPreprocessor {
...
+        let jailed: Pin<
+            Box<
+                dyn Stream<
+                        Item = JailAnnotated<
+                            dynamo_protocols::types::CreateChatCompletionStreamResponse,
+                        >,
+                    > + Send,
+            >,
+        > = if guided_streaming {
+            Box::pin(apply_tool_calling_jail_with_guided_streaming(
```
KrishnanPrash, line 4275: *AI Generated Review Comment*

**The jail path passes a terminal error through and still finalizes**

```rust
Box::pin(apply_tool_calling_jail_with_guided_streaming(
```

This PR gives both unified adapters an error short-circuit: a terminal error ends the stream and suppresses end-of-stream recovery. The jail path has no such guard. The wrapper above maps the error into the jail's `Annotated` (the `jail_input` closure a few lines up), the jail passes data-less chunks through, and its end-of-stream finalize can still emit data and a synthesized `tool_calls` finish after the error. The behavior predates this PR, but guided streaming raises the cost, because partial arguments are already on the wire when the error arrives. Give the jail wrapper the same contract as the unified adapters: yield the error, end the stream, and do not run the finalize.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
```diff
@@ -31,6 +31,62 @@ fn contains_harmony_protocol(text: &str) -> bool {
...
+    constraint: &crate::protocols::openai::GuidedToolConstraint,
+) -> anyhow::Result<(
+    Vec<dynamo_parsers::tool_calling::ToolCallResponse>,
+    Option<String>,
+)> {
+    if constraint.installs_guided_json() {
+        match super::tool_parser_v2::parse_complete_guided_json(content, constraint) {
+            Ok(calls) => return Ok((calls, Some(String::new()))),
+            Err(guided_error) => {
+                if !contains_native_tool_call_marker(content, parser) {
```
KrishnanPrash, line 96: *AI Generated Review Comment*

**The native fallback ignores the forced tool name**

```rust
Err(guided_error) => {
    if !contains_native_tool_call_marker(content, parser) {
        return Err(guided_error);
    }
```

When a `GuidedJsonNamed` parse fails and native markup is present, the fallback parses natively and returns whatever tool the markup names. Nothing compares the recovered calls to the forced name, so a client that pinned `tool_choice` to `get_weather` can receive a different tool. The v1 finalize had the same tolerance, so this is not a regression, but the typed constraint is now in scope and the filter is one comparison. After a native fallback under a named constraint, drop or reject each call whose name differs from `tool_name`. The muse and qwen unified batch branches need the same filter.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
KrishnanPrash, line 123: *AI Generated Review Comment*

**`selected_family` writes one INFO line per request on the hot path**

BEFORE:
```rust
tracing::info!(
    target: "dynamo_unified",
    ?tool_call_parser,
    ?reasoning_parser,
    ?configured,
    flag_on = experimental_parsers_v2_enabled(),
    "unified parser path decision"
);
```
AFTER:
```rust
if configured.is_some() {
    tracing::info!(
        target: "dynamo_unified",
        // ...
        "unified parser path decision"
    );
}
```

`selected_family` runs once per streaming chat request for every model family; only the muse arm returns before it. Each request writes one INFO line, also on deployments that never configure the qwen3 pair. Keep the diagnostic value at lower cost: log only when `configured` is `Some`, or demote the line to `debug!`, or log once per process.

lib/llm/src/preprocessor.rs
```diff
@@ -3436,11 +3563,16 @@ impl OpenAIPreprocessor {
...
+                // A forced tool_choice installed a JSON grammar, so the jail may release
+                // calls as they arrive instead of buffering to the closing brace. The
+                // jail keeps its own native fallback, so a backend that ignores the
+                // grammar (MiniMax M2 emits XML under `required`) still parses normally.
                 Box::pin(Self::apply_tool_calling_jail(
                     effective_tool_call_parser,
                     request.inner.tool_choice.clone(),
                     tool_definitions,
                     uses_tool_call_structural_tag,
+                    guided.installs_guided_json(),
```
KrishnanPrash: *AI Generated Review Comment*

**Do we need a rollback lever for default-on guided streaming?**

```rust
guided.installs_guided_json(),
```

Suggestion, feel free to resolve. Incremental release changes client-visible semantics for every forced tool call on the default path, and the release mechanics live in the published `dynamo-parsers` crate. If a backend half-honors the grammar in production, the only rollback is a dependency repin and a rebuild. Is that acceptable for the first release, or do we want a one-release env escape hatch here, for example `guided.installs_guided_json() && !env_is_truthy(DYN_DISABLE_GUIDED_TOOL_STREAMING)`? If the repin path is acceptable, resolve this comment.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
rmccorm4: +1 to demoting this to DEBUG instead of INFO: https://github.com/ai-dynamo/dynamo/pull/12576#discussion_r3846414971

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
keivenchang: fixed in 094a8c97ff — apostrophe exemption now requires alphanumeric on either side instead of both, so trailing possessives don't open a phantom quote span

lib/llm/src/preprocessor.rs
keivenchang: fixed in 094a8c97ff — ported v1's `bypass_bare_guided_json` byte-shape check into the unified parser: the starting state now comes from the first content chunk's shape instead of committing blindly, so bare guided JSON no longer gets swallowed as reasoning while the genuine reason-then-JSON case still splits correctly

lib/llm/src/preprocessor.rs
keivenchang: fixed in 094a8c97ff — the jail wrapper now latches a terminal upstream error and stops the stream there, instead of still running the jail's own finalize afterward

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
keivenchang: fixed in 094a8c97ff — `detect_prefill` now compares first marker positions instead of testing presence

lib/llm/src/preprocessor.rs
keivenchang: added in 094a8c97ff — `DYN_ENABLE_GUIDED_TOOL_STREAMING`, on by default, set it falsy to fall back to buffer-to-completion at runtime without a repin

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
keivenchang: fixed in 094a8c97ff — added `tools` to `ParsingOptions`, threaded real request tools through `parse_complete` and its callers, so batch argument types now match streaming

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
keivenchang: fixed in 094a8c97ff — native fallback results are now filtered against the forced tool name at all 3 sites (this one, the muse batch branch, and the qwen unified batch path) instead of being returned unfiltered

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
keivenchang: fixed in 094a8c97ff — tracks per-choice detour state so a raw run resumed after an already-parsed interruption starts at `Response`, not the original request-level prefill

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
keivenchang: fixed in 094a8c97ff — same fix as the sibling comment on this line, gated to only log when the qwen3 pair is configured

keivenchang: fixed in 094a8c97ff — `selected_family`'s INFO log now only fires when the qwen3 pair is actually configured, instead of once per request regardless

KrishnanPrash (approved):
Stamped to unblock

lib/llm/src/preprocessor.rs
KrishnanPrash: *AI Generated Review Comment*

**The shape check commits on the first delta, so the swallow is still reachable**

```rust
bare_guided_json_prefill(original.delta.content.as_ref())
```
(unified_parser.rs:1114)

The shape check runs once, on the first delta of the choice. A role-only delta (`content: null`) or a whitespace-only chunk returns `Reasoning`, and the parser keeps that state forever. Most streams open with a role-only delta, so the bug from this thread still occurs. I reproduced `calls=0` at this head with the chunks `[" ", "[{...get_weather...}]"]`. Defer the decision until the first non-whitespace byte arrives, as the v1 code does: it re-evaluates on every chunk. The new regression test starts with `[` in the first chunk, so it cannot catch this.

lib/llm/src/protocols/openai/chat_completions/aggregator.rs
KrishnanPrash: *AI Generated Review Comment*

**A name-filter mismatch leaks the wrong tool's raw markup as content**

```rust
Err(error) => {
    tracing::debug!(
        error = %error,
        parser,
        "failed to parse aggregated chat tool calls"
    );
    continue;
}
```
(aggregator.rs:717)

When the filter drops every mismatched call, `parse_complete_tool_output` returns an error, and this `continue` leaves `choice.text` unchanged. The client that pinned `tool_choice` to one tool then receives a different tool's raw native markup as content. This is the default batch path, and it is new in this commit: the previous code replaced the text with parsed content. Strip the markup on this path (use the parse's normal text, or blank the text), and extend the mismatch tests to assert on content, not only on empty `tool_calls`.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
KrishnanPrash: *AI Generated Review Comment*

**The either-side apostrophe rule disables single-quote spans**

BEFORE (either side, condensed):
```rust
character == '\'' && (prev_is_alphanumeric || next_is_alphanumeric)
```
AFTER (left side only):
```rust
character == '\'' && prev_is_alphanumeric
```
(unified_parser.rs:295)

A normal opening quote is followed by a letter (`'quoted text'`), so the either-side rule exempts it too, and single-quote spans no longer hide markers. I reproduced it at this head: `note 'a </think> tag' more text` now returns `Reasoning`; before this commit it returned `Response`, which is correct. Exempt on the left side only. That fixes `James'` and `isn't` equally, keeps quoted spans working, and `'twas` stays covered by the unclosed-quote rule. Add a test that pins the quoted-marker case.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
KrishnanPrash: *AI Generated Review Comment*

**A reasoning-only detour must resume at `Reasoning`, and the doc claim is false**

```rust
let choice_prefill = if records
    .get(&original.index)
    .is_some_and(|record| record.detoured)
{
    UnifiedParserStartingState::Response
```
(unified_parser.rs:1108)

The `ChoiceRecord` doc says a detour "only happens after any reasoning phase would have concluded". A detour that carries only `reasoning_content` disproves that: the worker was inside reasoning. I reproduced the leak at this head: raw `"secret-open "`, a reasoning-only detour, then raw `"more-secret</think>answer"` returns `more-secret</think>answer` as visible content, with the raw marker included. Resume at `Reasoning` when the detour chunk carried `reasoning_content`; keep `Response` for the other triggers. Correct the doc sentence as well; the regression test closes reasoning before its detour, so it cannot catch this.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
KrishnanPrash: *AI Generated Review Comment*

**Tools reach one batch site of three**

BEFORE:
```rust
super::tool_parser_v2::parse_complete(content, None, parser)
```
AFTER:
```rust
super::tool_parser_v2::parse_complete(content, Some(&parsing_options.tools), parser)
```
(aggregator.rs:111)

The tools now reach `unified_parser::parse_complete`, which fixes the qwen3 pair. But this site and both muse sites (lines 593 and 630, `parse_complete_unified(..., None, ...)`) still pass no tools. A `qwen3_coder`-only config with the v2 flag takes this arm, so the typing divergence from this thread survives there. Pass the tools at all three sites.

lib/llm/src/protocols/openai/chat_completions/unified_parser.rs
KrishnanPrash: *AI Generated Review Comment*

**The log gate from this thread is not in the commit**

```rust
tracing::info!(
    target: "dynamo_unified",
    ...
    "unified parser path decision"
);
```
(unified_parser.rs:123)

The reply above says the log now fires only when the qwen3 pair is configured, fixed in `094a8c97ff`. The code at that commit still logs unconditionally on every streaming request, and no hunk in the fix commit touches this function. The gate probably got lost in the rebase; please re-apply it.

keivenchang:
## Why this can land with DIS-2779 and DIS-2780 open

The Qwen3.6-27B E2E run found two issues. Neither needs to block this PR, for different reasons.

**DIS-2779 is below this PR.** With guided decoding, a zero-argument tool generates whitespace until `max_tokens`. Main reproduces it, and it accounts for both of main's custom-suite failures. I also reproduced it locally in three configurations, including with this PR's feature disabled: 21,427 characters each time, with one non-whitespace character. The failure is in guided decoding, below the code this PR changes.

**DIS-2780 includes one pre-existing behavior and one regression from this PR.**

The wrong `finish_reason` is pre-existing. `impl From<DeltaChoice> for ChatChoice` in `aggregator.rs` changes `Length` to `ToolCalls` whenever tool calls are present. That block is byte-identical on main, its TODO asks this exact question, and main has a test pinning the behavior. This PR makes the path reachable because main does not produce a tool call for this input.

The truncated `arguments` behavior is caused by this PR. A tool call cut off mid-string is returned with `arguments` that do not parse.

**Why I still think this is better than main.** With the feature off, the same request drops the tool call and writes 130 characters of raw tool markup into `content`, so the caller receives it as assistant text. I measured that locally, and it matches main's harness output. This PR returns a named tool call with truncated arguments; main silently drops the tool call and leaks raw markup into visible `content`. Both responses are wrong. The contract for a call interrupted by `max_tokens` still needs to be defined.

**Suite score.** This PR is 44/48; main is 46/48. The two extra failures are the same truncation case in stream and nonstream. Main passes because the harness checks the tool-call fields and not the `content` leak. The two-point gap measures the current assertion, not a better response from main.

**Rollback.** `DYN_ENABLE_GUIDED_TOOL_STREAMING=0` restores main's behavior byte-for-byte for this input, including the 130-character leak. A downstream user who hits the regression can disable the feature without reverting code.

DIS-2780 stays open to define what a tool call interrupted by `max_tokens` should return and apply that decision at both override sites, so streaming and non-streaming agree.

nv-tusharma (approved):

nv-anants (approved):
dependency updates look good

jthomson04 (approved):

keivenchang:
## Correction: the rollback claim above was wrong

I wrote above that `DYN_ENABLE_GUIDED_TOOL_STREAMING=0` restores the previous behavior and gives a downstream user a way out without reverting code. That was wrong for the configuration this PR makes the default. With `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2=1`, the flag does not restore the previous behavior.

The E2E harness runs with `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2=1`. For the Qwen3 family, `postprocessor_parsing_stream_with_constraint` returns early into `unified_parser::apply_stream_with_constraint`. The code reads `DYN_ENABLE_GUIDED_TOOL_STREAMING` further down on the jail branch, so the v2 path never reaches that read.

Here is the live-server measurement for three streaming forced-tool-choice cases. The table counts SSE chunks carrying argument fragments:

| config | case 1 | case 2 | case 3 |
|---|---|---|---|
| v2 on, flag unset | 28 | 23 | 96 |
| v2 on, flag `0` | 28 | 23 | 96 |
| v2 off, flag unset | 28 | 23 | 96 |
| v2 off, flag `0` | 1 | 1 | 1 |

The flag works in the last row. The second row shows no change when v2 is on.

My earlier test changed both variables in the same restart: it unset `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` and set the guided flag to `0`. I attributed the result to the guided flag, but v2 turning off caused it.

Fixed in `044c0fd17f127a640173e051beb2550c3e61c49c`: the v2 path now reads the same rollback decision the jail path already used, so the flag governs both.

Separately, the same run showed that the feature works end to end. Arguments start arriving within the first three SSE events, the JSON parses, the values match the request, and nothing leaks into `content`.

PR #12583, opened 2026-08-03
fix(mocker): use logical KV tokens for decode timing [DYN-3851]

PeaBrane:
## Summary

- Align Mocker decode timing with the scheduled logical KV-token semantics used by profiler-derived performance data.
- Stop using distinct physically resident blocks as a decode-compute proxy, which undercounts heterogeneous batches when prefix caching deduplicates shared blocks.
- Keep the full logical sum uncapped for interpolation while saturating only the legacy polynomial model's utilization ratio at 1.0.

This does not change KV allocation, prefix-cache ownership, routing, or direct AIC decode prediction.

## Details

- vLLM ordinary and speculative decode now pass the summed context lengths of the requests scheduled for decode.
- SGLang uses the same uncapped logical-KV meaning.
- Removed reservation-size accessors that were only needed by the physical-block timing proxy.

Start with `lib/mocker/src/scheduler/vllm/core.rs` for the semantic change and `lib/mocker/src/common/perf_model.rs` for the timing-model boundary.

## Related Issues

## Validation

- `cargo test -p dynamo-mocker test_fpm_prefill_and_decode_mixed_batch -- --nocapture` (vLLM and SGLang cases passed locally and on a clean Linux checkout)
- `cargo test -p dynamo-mocker --features kvbm-offload speculative_decode_reservation_waits_without_preempting -- --nocapture`
- `cargo fmt --all -- --check`, commit hooks, and `git diff --check`
- ComputeLab SC-01: single H100, Qwen3-1.7B, vLLM TP1, 38,863 runtime GPU KV blocks, four prefix groups at 68.68-74.85% measured prefix-cache token hit rate, three trials each at concurrency 8 and 128, zero request errors or output-length mismatches.
- At concurrency 128, candidate median ITL was 21.69 ms versus live 22.12 ms; baseline was 8.98 ms. Candidate output-throughput error was -19.6% versus baseline +40.8%.
- Paired trace at scheduler batch 106/context 4,103 showed candidate `active_kv_tokens=434,972` (106.01 logical contexts) versus baseline `143,968` (35.09 contexts), confirming the corrected timing input tracks the scheduled batch under prefix dedup.

- Lower-level attribution on the exact live stack (H100 80GB, pinned vLLM 0.23.1 development image, Qwen3-1.7B BF16 TP1, FlashAttention 3) reproduced the scheduler shape at batch 106 with 434,976 logical KV tokens. At fixed batch and logical total, the actual ragged request vector was within 0.5% of the homogeneous attention control.
- The old batch-35 proxy underpriced direct paged FlashAttention by 58.2% and steady-state engine decode by 48.6% versus the actual batch-106 workload. This directly validates removing the dedup-aware scheduler cap.
- Physical prefix-block aliasing reduced attention time by 35.7% versus an unaliased replay. That is a separate attention-model fidelity opportunity; it does not make deduplicated physical blocks a valid proxy for the scheduled decode batch.

The live image reports vLLM `0.23.1rc1.dev1502`; the available AIC interpolation data is for vLLM `0.24.0`. Candidate mean and upper-tail simulated ITL remained conservative, so the campaign validates the accounting semantics and center-of-distribution parity rather than exact end-to-end calibration across those versions.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12583" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/lib/mocker/src/scheduler/vllm/core.rs
+++ b/lib/mocker/src/scheduler/vllm/core.rs
@@ -2200,8 +2200,8 @@ impl VllmCore {
         let (decode_time, decode_end_ms) = if self.args.worker_type == WorkerType::Prefill {
             (Duration::ZERO, decode_start_ms)
         } else {
-            let active_kv_tokens = self.kv_manager.num_active_blocks() * self.args.block_size;
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
+            let active_kv_tokens = total_length.min(total_kv_tokens);
             let context_length = total_length / ready.len();
             let decode_ms = self.args.perf_model.predict_decode_time(
                 ready.len(),
@@ -2435,12 +2435,8 @@ impl VllmCore {
         let (decode_time, decode_end_ms) = if self.args.worker_type == WorkerType::Prefill {
             (Duration::ZERO, decode_start_ms)
         } else {
-            let active_kv_tokens = self
-                .kv_manager
-                .num_active_blocks()
-                .saturating_sub(reservation.len())
-                * self.args.block_size;
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
+            let active_kv_tokens = total_length.min(total_kv_tokens);
             let context_length = total_length / ready.len();
             let decode_ms = self.args.perf_model.predict_decode_time(
                 ready.len(),
--- a/lib/mocker/src/scheduler/vllm/core.rs
+++ b/lib/mocker/src/scheduler/vllm/core.rs
@@ -2201,7 +2201,7 @@ impl VllmCore {
             (Duration::ZERO, decode_start_ms)
         } else {
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
-            let active_kv_tokens = total_length.min(total_kv_tokens);
+            let active_kv_tokens = total_length;
             let context_length = total_length / ready.len();
             let decode_ms = self.args.perf_model.predict_decode_time(
                 ready.len(),
@@ -2436,7 +2436,7 @@ impl VllmCore {
             (Duration::ZERO, decode_start_ms)
         } else {
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
-            let active_kv_tokens = total_length.min(total_kv_tokens);
+            let active_kv_tokens = total_length;
             let context_length = total_length / ready.len();
             let decode_ms = self.args.perf_model.predict_decode_time(
                 ready.len(),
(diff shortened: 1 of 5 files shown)


lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -2435,12 +2435,8 @@ impl VllmCore {
         let (decode_time, decode_end_ms) = if self.args.worker_type == WorkerType::Prefill {
             (Duration::ZERO, decode_start_ms)
         } else {
-            let active_kv_tokens = self
-                .kv_manager
-                .num_active_blocks()
-                .saturating_sub(reservation.len())
-                * self.args.block_size;
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
+            let active_kv_tokens = total_length.min(total_kv_tokens);
```
dreamtalen: Non-blocking: do we have to cap the active_kv_tokens at total_kv_tokens here? Is it because AIC doesn't allow a value larger than total_kv_tokens?

dreamtalen (approved):

lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -2200,8 +2200,8 @@ impl VllmCore {
         let (decode_time, decode_end_ms) = if self.args.worker_type == WorkerType::Prefill {
             (Duration::ZERO, decode_start_ms)
         } else {
-            let active_kv_tokens = self.kv_manager.num_active_blocks() * self.args.block_size;
             let total_kv_tokens = self.args.num_gpu_blocks * self.args.block_size;
+            let active_kv_tokens = total_length.min(total_kv_tokens);
```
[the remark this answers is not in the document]
PeaBrane: No fix for this concern: prefill work is timed separately in a mixed iteration, and decode active KV/FPMs describe only requests actually scheduled for decode. KV resident for requests still chunk-prefilling or blocked is capacity state, not decode compute in this pass. The follow-up keeps that distinction while making the scheduled logical sum uncapped.

lib/mocker/src/scheduler/vllm/core.rs
PeaBrane: You are right: this should not be capped at the scheduler. Prefix dedup means the scheduled logical KV sum can exceed physical KV capacity. Commit 86e3f556 updates both vLLM paths and SGLang to pass the uncapped logical sum. Interpolated timing consumes that value, direct AIC ignores it, and only the legacy polynomial model clamps its utilization ratio to 1.0.

jthomson04 (approved):

ishandhanani (approved):

PR #12421, opened 2026-07-30
feat(operator): add runtime feature gates

xianlubird:
## Summary

Add explicit, per-feature gates whose rendered defaults depend only on the resolved Dynamo runtime compatibility version. This is the runtime-gate foundation for #11083.

## Details

- add `internal/features/runtime` with a small `Gate` type and centralized feature definitions
- define `CanaryHealthChecks` as introduced by Dynamo runtime 1.4.0
- add normalized runtime-version comparison through `runtimeversion.Version.AtLeast`
- keep rendering stable across operator-only upgrades; changing runtime version is the trigger for adopting new defaults
- keep explicit user configuration authoritative for older-runtime escape hatches
- leave the existing origin-based vLLM compatibility gate unchanged and out of scope

## Where should reviewers start?

- `deploy/operator/internal/features/runtime/gate.go`
- `deploy/operator/internal/features/runtime/gates.go`

## Related Issues

- Follow-up implementation: #11083
- Runtime version API: #10494

PR #12587, opened 2026-08-03
feat(operator): hash effective runtime rendering profiles

tmonty12:
## Summary

- resolve the effective runtime version from `runtimeVersionOverride` or the main image tag
- construct a behavior-oriented `RuntimeProfile` and thread it through `ComponentContext`
- include only non-empty profiles in the v2 worker hash while continuing to strip the raw override
- keep the legacy alpha hash unchanged and regenerate the affected CRD descriptions

## Stack

Depends on #12421.

This branch is one commit directly on top of #12421 commit `8fa7ea8765c48e941c6a60047e3574ad2960a411`. The dependency PR head is in a fork, so GitHub cannot use that branch as the base of an upstream PR; until #12421 merges, the Files changed view also includes its commits.

## Compatibility contract

- unknown runtimes and resolved versions through 1.4 use the empty legacy profile, which is omitted from v2 hash serialization
- resolved versions 1.5 and newer use the first non-empty versioned rendering profile
- equivalent implicit and explicit versions hash identically
- version changes that produce the same effective profile do not change the hash
- profile changes create a new worker generation
- the legacy alpha hash remains frozen

## Draft note

This intentionally makes 1.5 the profile-adoption boundary even though the canary capability gate has a 1.4 minimum. That preserves established 1.4 v2 hashes. The profile is threaded to renderers here but is not yet consumed by an injection renderer; the first profile-consuming renderer and this hash behavior should land together so the hash continues to represent rendered output.

## Validation

- full non-e2e operator Go test suite with envtest assets
- `go test ./internal/runtimeversion ./internal/features/runtime ./internal/dynamo`
- `make generate`
- `make vet`
- `make lint`
- regenerated operator manifests


tmonty12:
Superseded by the clean stacked draft in xianlubird/dynamo#2, based directly on the #12421 head branch.

PR #11083, opened 2026-06-30
feat(operator): enable worker canary health checks

xianlubird:
## Overview

Enable canary worker health checks by default for Kubernetes operator-managed workers. This brings the operator back in line with the published health check documentation, which states that Kubernetes deployments default `DYN_HEALTH_CHECK_ENABLED` to `true`: https://docs.nvidia.com/dynamo/user-guides/observability-local/health-checks#worker-liveness-and-health-check

## Details

- Set the worker base container default `DYN_HEALTH_CHECK_ENABLED` value to `true`.
- Update Grove, DCD, and controller test expectations for newly rendered worker specs.
- Keep upgrade-test persisted child fixtures at the old `false` value and explicitly assert that upgrading across this default flip changes the child spec hash, while still preserving legacy worker identity labels.

The worker default was disabled in #5038 as a temporary mitigation for DIS-1185 races involving discovery propagation, NATS subscription readiness, and health check requests. The underlying issues have since been addressed:

- #4876 added discovery readiness waiting for canary health checks.
- #8165 made the canary the sole authority for endpoint readiness when enabled, avoiding the eager Ready -> NotReady race that could trigger Kubernetes restarts.
- #8294 moved canary health checks to the local endpoint registry, removing the fragile discovery / PushRouter / instance-id path for self-checks.

With those fixes in place, Kubernetes workers can now use the documented default and require the canary path to validate readiness.

## Where should the reviewer start?

Start with `deploy/operator/internal/dynamo/component_worker.go`, then review the test expectation updates in `deploy/operator/internal/dynamo/graph_test.go`, `deploy/operator/internal/controller/dynamocomponentdeployment_controller_test.go`, and `deploy/operator/internal/controller/upgrade_test.go`.

## Related Issues

- DIS-1185
- Follow-up to #5038
- Related fixes: #4876, #8165, #8294

## Summary

- Enable `DYN_HEALTH_CHECK_ENABLED=true` by default for operator-managed worker components in Kubernetes.
- Align the operator behavior with the published Kubernetes health check default.
- Document and test that existing persisted worker children with the old default will roll out when reconciled with the new default.

## Validation

- `go test ./internal/dynamo ./internal/controller` from `deploy/operator`
- `git diff --check`

PR #12617, opened 2026-08-04
feat(operator): hash effective runtime profiles

xianlubird:
## Summary

- resolve each component's effective Dynamo runtime compatibility version from `runtimeVersionOverride` or the main image tag
- evaluate the centralized runtime feature gates into a canonical `RuntimeProfile`
- include non-empty runtime profiles in the v2 DGD worker hash while continuing to exclude the raw override value
- thread the profile through `ComponentContext` so rendering and hashing can consume the same gate decisions
- preserve the frozen legacy v1alpha1 hash and omit empty profiles to avoid changing existing pre-gate hashes
- clarify that changing `runtimeVersionOverride` can alter runtime-gated Pod rendering and trigger a rollout

## Why

#10494 intentionally excluded `runtimeVersionOverride` from worker hashing because it did not affect rendered resources at that time. After #12421 introduced runtime-version feature gates, changing only the override can change the rendered `PodSpec`.

Hashing the raw override would be too sensitive: `1.4.0` and `2.0.0` currently enable the same gates and should not create different worker generations. This change hashes the effective rendering profile instead:

| Change | Effective profile | Worker hash |
| --- | --- | --- |
| unknown runtime → `1.3.0` | empty → empty | unchanged |
| `1.3.0` → `1.4.0` | canary disabled → enabled | changed |
| `1.4.0` → `2.0.0` | enabled → enabled | unchanged |
| image tag `1.5.0` → override `1.5.0` | equivalent | unchanged |

This ensures changes that affect rendered worker behavior use the managed rolling-update path without introducing rollouts for compatibility-version changes that produce identical rendering.

This is an upstream-focused version of Thomas Montfort's follow-up proposal in https://github.com/xianlubird/dynamo/pull/2, with the unrelated `graph_test.go` consolidation left out to keep the review surface focused.

## Validation

- `make test`
- `make manifests`
- `make vet`
- `go test ./internal/runtimeversion ./internal/features/runtime ./internal/dynamo`

`make lint` was also attempted locally, but the repository-pinned golangci-lint v1.64.8 failed while loading the Go 1.26 module with `no go files to analyze`; it did not report a source finding.

## Related Issues

- Builds on #12421
- Supports the runtime-gated canary implementation in #11083

## Summary by CodeRabbit

* **New Features**
  * Runtime compatibility versions now determine applicable runtime feature behavior during deployment rendering.
  * Runtime version overrides take precedence over image-derived versions, including for custom images.
  * Runtime-gated changes, such as canary health checks, are reflected in deployment updates and rollouts.

* **Documentation**
  * Clarified that changing the runtime compatibility version can affect rendered Pods and trigger a rollout, without changing the container image.


deploy/operator/internal/dynamo/hash.go
```diff
@@ -60,9 +61,10 @@ func ComputeDGDWorkersSpecHash(dgd *v1beta1.DynamoGraphDeployment) (string, erro
 	}
 
 	type workerTemplate struct {
-		Labels      map[string]string                     `json:"labels,omitempty"`
-		Annotations map[string]string                     `json:"annotations,omitempty"`
-		Spec        v1beta1.DynamoComponentDeploymentSpec `json:"spec"`
+		Labels         map[string]string                     `json:"labels,omitempty"`
+		Annotations    map[string]string                     `json:"annotations,omitempty"`
+		RuntimeProfile *runtimefeatures.RuntimeProfile       `json:"runtimeProfile,omitempty"`
```
sttts, line 66: This needs thought why it is correct to add the gates to the hash. And further, what it means when gates changes, or gates are introduced in a version, but are false (then the hash should not change?).

And we need a plan for when gates become true, whether we tolerate automatic rollouts, in general or only by opt-in.

tmonty12:
Closing in favor of https://github.com/ai-dynamo/dynamo/pull/12633

Issue #12050, opened 2026-07-23
DEP: Balancing / Syncing Worker Taints for Routing


nnshah1:
*Note* this was split from #10703 and originally authored by @michaelfeil 

## Area
router / dynamo-runtime / deployment::k8s

## Status
draft

## Summary
Given the ability to update a worker's taints after registration (DEP #10703, Part 1),
this DEP defines how a set of interchangeable workers on the same endpoint converge
toward a *desired taint distribution* (e.g. 50% `fast` / 50% `slow`) and keep that
assignment consistent enough for routing — across router replicas, scale up/down, and
node loss — without restarting workers.

Part 1 gives the *mechanism* (update a taint). This DEP is about the *policy and
coordination*: who decides which worker is `fast`, and how the decision stays stable
and consistent.

## Motivation / Use case
Operators want to carve one homogeneous endpoint into policy pools for SLA / customer
routing (premium traffic requires/prefers `fast`), with ratios that can change at
runtime (50/50 -> 40/60), on shared multi-cluster capacity and NVL72 topologies —
*without* hard-partitioning into separate deployments/namespaces. Hard partitions
fragment capacity, and moving a GPU between pools costs a full drain (~4 min) + cold
start (up to ~23 min for large models w/ JIT+CUDA-graph warmup), which under load
causes SLA misses and cascading failures.

Explicitly out of scope for the interim: anything requiring a new k8s operator. The
first shippable version should be operator-free, small (~hundreds of LOC), and
backportable.

## Scope
In scope:
- Deriving a per-worker policy taint (e.g. `fast`/`slow`) from a target ratio + worker
  membership, and keeping it eventually consistent.
- A dynamically-watchable source for the desired ratio (ConfigMap / CRD field / DGD).
- Convergence and anti-oscillation behavior.

Out of scope:
- The taint update mechanism itself (Part 1 / #10703).
- Infra/topology-derived taints (`dynamo.topology/*`) — authoritative, never balanced.
- Reconfiguring worker *behavior* on taint change (TP/DP attn, speculation length,
  etc.) — some of these can't happen without a restart; tracked separately.

## The core problem: stable identity + cross-replica consistency
Two hard constraints surfaced in discussion:

1. **No stable hash basis in k8s.** With workers {A,B,C} and 40% `fast`, hashing to pick
   the top-K gives {A,B}; a new worker D arrives -> {A,D}; A restarts -> {B,D}. Pod
[...]

Issue #10703, opened 2026-06-14
DEP: Support dynamic worker taint updates in ModelRuntimeConfig discovery


michaelfeil:
### Area

router

## Summary

  Workers can advertise routing taints through `ModelRuntimeConfig.taints` in the Model Deployment Card (MDC), and the router already evaluates request `required_taints` / `preferred_taints` against the current worker runtime config.

  However, taints are effectively static today. There is no supported post-registration API to update a worker’s
  taints, and discovery propagation differs between etcd and Kubernetes.

  ## Use Case

  I want workers on the same endpoint to dynamically advertise pool membership, for example:

  - 50% of replicas advertise `fast`
  - 50% advertise `slow`
  - ratios may be changed dynamically, e.g. 40/60
  - workers rebalance eventually after scale-up, scale-down, or node loss
  - routers should use these taints for SLA or customer-specific routing constraints

  Example request behavior:
  - premium traffic requires or prefers `fast`
  - default traffic can go anywhere
  - pool membership should converge without restarting workers

  ## Current Behavior

  ### Router side

  The router already reads worker taints from the current `ModelRuntimeConfig`:

  - `ModelRuntimeConfig.taints` implements `WorkerConfigLike::taints`
  - scheduler selection borrows the current `workers_with_configs`
  - required taints filter workers
  - preferred taints bias worker scoring

  So the router can use updated taints if the `RuntimeConfigWatch` receives an updated config.

  One caveat: queued requests may not be re-evaluated immediately on config-only changes. New routing decisions see
  the current config, but pending queue wakeups currently appear tied to remote state / periodic rechecks, not worker-
  config changes.

  ### etcd / KV discovery path

  MDCs are registered under keys like:

  `v1/mdc/{namespace}/{component}/{endpoint}/{instance_id}[/{lora_suffix}]`

  The KV watch path maps `WatchEvent::Put` to `DiscoveryEvent::Added`, and `watch_and_extract_field` overwrites the
  existing runtime config for that worker id. This means the consumer/watch side can handle update events.

  But current registration is create-oriented:
  - `KVStoreDiscovery::register_internal` calls `bucket.insert(..., revision=0)`
  - etcd treats revision `0` as create
  - if the key already exists, the store returns `Exists` and does not update the value
  - therefore calling `register_model` again for the same worker does not update taints

  ### Kubernetes discovery path

  Kubernetes discovery stores model cards in metadata and persists that metadata to the CR.

[...]

PR #12620, opened 2026-08-04
feat(router): add dynamic worker taint updates

tmonty12:
## Summary

* add the narrow `update_model_taints(endpoint, taints)` Python API for base-model worker routing metadata
* replace only caller-managed taints, reject the reserved `dynamo.topology/*` namespace, and regenerate topology taints from immutable topology metadata
* add existing-record replacement for KV discovery and same-ID value-change propagation for Kubernetes and mock discovery
* wake and re-check pending scheduler requests when worker configs change
* register `POST /engine/update/model_taints` across vLLM, SGLang, and TensorRT-LLM worker modes, including realtime, embedding, multimodal, Omni, and diffusion variants
* register the same route in the shared Rust common worker so TokenSpeed and future unified backends receive it without engine-specific opt-in
* document the Kubernetes port-forward/curl flow
* enforce the immutable model-card boundary and verify taint-only changes do not alter `mdcsum()`

This is Part 1 of ai-dynamo/dynamo#10703. Endpoint-local pool balancing remains deferred to ai-dynamo/dynamo#12050. LoRA MDC runtime configs are not consumed by router runtime-config discovery, so the update API intentionally targets only the base-model MDC.

## Validation

* `cargo test -p dynamo-runtime model_taint_update_tests`
* `cargo test -p dynamo-runtime model_taint_updates_replace_existing_value_and_emit_added`
* `cargo test -p dynamo-runtime changed_model_value_is_an_added_upsert`
* `cargo test -p dynamo-kv-router worker_taint_update_wakes_constrained_pending_request`
* `cargo test -p dynamo-llm runtime_taints_do_not_change_mdcsum`
* `cargo test -p dynamo-backend-common` (137 passed)
* `cargo test -p dynamo-backend-common --features integration handoff_precedes_registration_and_populates_namespaced_registry`
* `cargo test -p dynamo-backend-common --features integration model_taint_update_route_updates_registered_base_model`
* `cargo check -p dynamo-llm`
* `cargo check --manifest-path lib/bindings/python/Cargo.toml`
* `cargo clippy -p dynamo-runtime -p dynamo-kv-router -p dynamo-llm --lib --tests -- -D warnings`
* `cargo clippy -p dynamo-backend-common --lib --tests -- -D warnings`
* `.venv/bin/python -m pytest components/src/dynamo/common/tests/test_model_taints.py` (5 passed)
* Ruff check across all changed backend modules
* Python bytecode compilation across all changed backend modules
* `cargo fmt --all -- --check`
* `cargo fmt --manifest-path lib/bindings/python/Cargo.toml -- --check`

The focused SGLang route-registration regression is included but could not run in the local lightweight environment because `torch` is not installed; CI has the backend dependencies.

## Draft follow-ups

* exercise the update path against live etcd and Kubernetes discovery deployments
* confirm the mutable-boundary enforcement location and mixed-version behavior with reviewers
* align semantic request-validation failures with a reusable typed `/engine/*` 4xx error path

## Summary by CodeRabbit

* **New Features**
  * Added an HTTP API and Python interface for replacing worker routing taints.
  * Worker taint changes now propagate through discovery and routing, enabling dynamic worker selection without restarting services.
  * Supported workers across language, embedding, multimodal, image, video, and realtime workloads.
* **Bug Fixes**
  * Prevented malformed, duplicate, or reserved topology taints from being applied.
  * Improved routing updates so newly eligible workers receive pending requests.
* **Documentation**
  * Added deployment guidance for updating taints, rollout ordering, and security considerations.


lib/runtime/src/storage/kv.rs
```diff
@@ -431,6 +431,9 @@ pub trait Bucket: Send + Sync {
     /// The Key should be the name of the item, not including the bucket name.
     async fn get(&self, key: &Key) -> Result<Option<bytes::Bytes>, StoreError>;
 
+    /// Replace an existing item without creating a missing key.
+    async fn replace(&self, key: &Key, value: bytes::Bytes) -> Result<StoreOutcome, StoreError>;
```
tmonty12: **[P1] Make this a true compare-and-replace operation.** The contract says a missing key must not be created, but the new implementations are check-then-write: etcd ultimately performs an unconditional put and treats a missing `prev_key` as `Created(1)`; NATS `resync_update` calls `create()` when the entry disappeared; and the file backend checks `exists()` before a later rename. If worker unregister/unload deletes the MDC between those steps, a concurrent taint update recreates stale metadata and the router can rediscover a worker/model that is gone. Please use a versioned CAS/transaction where deletion wins with `MissingKey`—never create on a wrong/missing revision—and add deterministic delete-vs-update tests for each backend.

examples/backends/vllm/deploy/README.md
```diff
@@ -207,6 +212,30 @@ After deployment, forward the frontend service to access the API:
...
+In another terminal, replace the caller-managed routing taints for that worker:
+
+```bash
+curl --fail-with-body \
+  -X POST http://localhost:9090/engine/update/model_taints \
+  -H 'Content-Type: application/json' \
+  -d '{"taints":["capacity/fast"]}'
+```
+
+The `taints` array is a replacement, not a merge. Send an empty array to clear caller-managed taints. Dynamo preserves generated `dynamo.topology/` taints and rejects callers that use that reserved prefix. The request updates only the selected worker pod; repeat it for each target worker.
```
tmonty12, line 244: **Please document the mixed-version compatibility boundary here.** Same-key taint changes are only observed by frontend/router binaries containing the new value-aware watcher. With a new worker and any older frontend replica, this request can return success while that replica keeps the previous taints indefinitely. Please state that `update_model_taints` is supported only when every frontend/router consumer and the target worker run a version containing this API; mixed-version use is unsupported. The safe rollout is frontends/routers first, then workers, then enable taint updates.

lib/runtime/src/discovery/mod.rs
```diff
@@ -1243,6 +1468,42 @@ pub trait Discovery: Send + Sync {
...
+            anyhow::bail!("model taint updates are supported only for base model cards")
+        }
+        let query = DiscoveryQuery::EndpointModels {
+            namespace: id.namespace.clone(),
+            component: id.component.clone(),
+            endpoint: id.endpoint.clone(),
+        };
+        let candidate_id = DiscoveryInstanceId::Model(id);
+        let existing = self
+            .list(query)
```
tmonty12: **[P1] Do not validate a mutation through the eventually consistent discovery list.** Kubernetes `list()` reads `metadata_watch`, which can lag this worker's authoritative `self.metadata` and the CR write. An immediate post-registration update can report “not registered”; a rapid `A → B → A` sequence can read stale `A` on the second request and return success as a no-op while authoritative state remains `B`. Please move the authoritative read, taint-only mutation, and persistence into each backend's atomic primitive: the Kubernetes `self.metadata` write lock, versioned CAS for KV stores, and the backend map lock for memory/mock. Keep `list()` as an observation API.

lib/runtime/src/discovery/kv_store.rs
```diff
@@ -247,8 +247,23 @@ impl KVStoreDiscovery {
 
                 match Self::parse_instance(kv.value()) {
                     Ok(instance) => {
-                        known_instances.insert(instance.id(), instance.clone());
-                        vec![DiscoveryEvent::Added(instance)]
+                        let id = instance.id();
+                        match classify_discovery_change(known_instances.get(&id), &instance) {
```
tmonty12, line 290: **[P1] This new state-bearing event can still be dropped before it reaches this classifier.** `KeyValueStoreManager::watch` sends incremental `WatchEvent`s through a 16,384-entry channel with a one-second `send_timeout`; on timeout it logs and continues. If the final taint Put for a worker is dropped there, `known_instances` and router state remain stale indefinitely because this local drop does not trigger a backend reconnect or `Resync`. Please use cancellable blocking sends for incremental state events, or force an authoritative resync/restart on timeout. Add a saturated-channel test proving the final taint state is eventually delivered.

lib/kv-router/src/scheduling/local.rs
```diff
@@ -221,16 +226,18 @@ where
...
 
                     let current_workers = monitor_rx.borrow_and_update().clone();
-                    Self::reconcile_worker_configs(
+                    if Self::reconcile_worker_configs(
                         &slots_monitor,
                         current_workers,
                         &mut last_workers,
-                    );
+                    ) {
+                        queue_config_updates.update().await;
```
tmonty12, line 251: **[P2] A one-worker taint change takes the full fleet/queue fallback here.** The path clones the complete config, joined-ready, and worker maps, reconciles every DP range, then `queue.update()` invokes the all-workers pending recheck. Updating N workers sequentially therefore produces quadratic map/topology work plus repeated full pending scans, even though a taint-only change affects one worker and does not change DP topology. Please preserve the changed worker ID, skip topology reconciliation for taint-only changes, and call `queue.update_worker` for its affected DP ranks; batch/coalesce dirty workers for fleet updates. A large-fleet burst benchmark would guard this path.

tmonty12 (commented):
Scoped review: five remaining concerns. All other previously discussed findings are intentionally omitted.

lib/runtime/src/discovery/kv_store.rs
tmonty12: [AI] Fixed in d72d7a529d. All manager watch events now use cancellable blocking sends, and a saturated-channel regression proves the final taint `Put` is delivered rather than dropped.

lib/kv-router/src/scheduling/local.rs
tmonty12: [AI] Deferred to #12050. PR #12620 is explicitly Part 1 (single-worker mutation and correct propagation); endpoint-local balancing and fleet bursts are Part 2. The follow-up should require that a burst of N taint-only updates does not trigger N full topology reconciliations or all-worker queue rechecks, preserves/coalesces dirty worker IDs safely across `watch` coalescing, and includes a large-fleet burst benchmark. Leaving this thread unresolved to track that work.

lib/runtime/src/discovery/mod.rs
tmonty12: [AI] Fixed in d72d7a529d. `update_model_taints` now delegates the authoritative read, taint-only mutation, and persistence to each backend's lock/CAS primitive instead of validating through `list()`.

examples/backends/vllm/deploy/README.md
tmonty12: [AI] Fixed in d72d7a529d. Documented the unsupported mixed-version boundary and the safe frontends/routers-first, workers-second rollout order.

lib/runtime/src/storage/kv.rs
tmonty12: [AI] Fixed in d72d7a529d. Replaced check-then-write with atomic compare-and-replace: etcd uses a `mod_revision` transaction, NATS uses revision CAS and maps tombstones to `MissingKey`, and memory/FileStore serialize compare and delete; live race regressions pass for all four backends.

components/src/dynamo/sglang/request_handlers/handler_base.py
```diff
@@ -883,6 +884,7 @@ def register_engine_routes(self, runtime: DistributedRuntime) -> None:
                     "with a built-in route"
                 )
 
+        register_model_taint_route(runtime, self.generate_endpoint)
```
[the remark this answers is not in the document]
tmonty12: Fixed in b560a63b51. SGLang now rejects configured routes whose normalized path equals MODEL_TAINT_ROUTE before runtime registration, and the unit test covers the reserved-route collision.

lib/runtime/src/discovery/mod.rs
```diff
@@ -1216,6 +1186,224 @@ fn find_conflicting_model_name(
...
+            },
+        )));
+    }
+
+    if matches!(candidate, DiscoveryInstance::Endpoint(_)) {
+        Ok(Some(DiscoveryEvent::Added(candidate.clone())))
+    } else {
+        Ok(None)
+    }
+}
```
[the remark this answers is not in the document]
tmonty12: No code change here. EventChannel and EventSource registration is create-only: same-ID registration validates and returns the existing record, so supported APIs cannot produce a same-ID replacement for these variants. The shared classifier retains endpoint upsert and model-taint update semantics without inventing a channel/source replacement event contract.

lib/runtime/src/storage/kv/nats.rs
```diff
@@ -247,9 +287,8 @@ impl NATSBucket {
     ) -> Result<StoreOutcome, StoreError> {
         match self.nats_store.entry(key).await {
             Ok(Some(entry)) => {
-                // Re-try the update with new version number
-                let next_rev = entry.revision + 1;
-                match self.nats_store.update(key, value, next_rev).await {
+                // Re-try against the current stream revision.
+                match self.nats_store.update(key, value, entry.revision).await {
                     Ok(correct_revision) => Ok(StoreOutcome::Created(correct_revision)),
```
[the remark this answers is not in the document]
tmonty12: Removed in 958ff92ac0. The NATS resync path is back to the pre-PR entry.revision + 1 behavior, so this PR no longer changes the general Manager::publish conflict semantics.

components/src/dynamo/common/model_taints.py
```diff
@@ -0,0 +1,44 @@
...
+            )
+
+        unique_taints = set(taints)
+        await update_model_taints(endpoint, unique_taints)
+        return {
+            "status": "ok",
+            "taints": sorted(unique_taints),
+        }
+
+    runtime.register_engine_route(MODEL_TAINT_ROUTE, _update_model_taints)
```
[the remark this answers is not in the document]
tmonty12: No code change. /engine/* is the existing trusted control-plane surface for engine routes and currently has no per-route authorization hook. Dynamo deployments must restrict the system port through the trusted network boundary; adding taint-only authorization here would be inconsistent with every other engine route. Generic system-server authentication should be handled separately.

components/src/dynamo/sglang/request_handlers/handler_base.py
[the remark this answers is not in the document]
tmonty12: No code change. SGLang registers this through the same shared /engine/* control-plane route mechanism as the other backends. The system port is expected to be network-restricted today; a generic system-server authentication layer is separate work and should cover all engine routes rather than only model taints.

lib/runtime/src/storage/kv/file.rs
```diff
@@ -5,9 +5,9 @@ use std::cmp;
 use std::collections::HashSet;
 use std::ffi::OsString;
 use std::fmt;
-use std::fs;
-use std::fs::OpenOptions;
+use std::fs::{self, File, OpenOptions};
 use std::io::{ErrorKind, Write};
+use std::os::fd::AsRawFd;
```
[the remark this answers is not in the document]
tmonty12: No code change. The Unix-only FileStore implementation predates this PR: OsStrExt, AsRawFd, and flock were already present on main. Windows FileStore support is therefore a pre-existing portability concern, not a regression introduced by model-taint updates.

tmonty12:
Responses to the review-body-only items (the corresponding inline items 1–5 were answered in their threads):

6. No change. `counter_worker.py` keeps the Dynamo imports inside `main()` after `CUDA_VISIBLE_DEVICES` is derived from `device_type`, so both direct CLI and process-wrapper paths preserve the required import order.

7. Addressed in `958ff92ac0`. The `Bucket::compare_and_replace` contract now documents that successful revisions are backend-specific and neither cross-backend comparable nor globally monotonic.

8. Addressed in `958ff92ac0`. Kubernetes discovery logs only the event kind and instance ID at info level; the full event is emitted at debug level.

9. Deferred. The current two-round-trip etcd CAS is correct, and collapsing it to one transaction is a performance-only optimization unrelated to the model-taint MVP. This should be handled as a focused follow-up.

10. Addressed in `958ff92ac0`. The NATS integration test now deletes its per-run JetStream bucket after assertions; the focused NATS test passed against the local service.

11. Deferred. Consolidating the duplicated FileStore temp-file cleanup is cosmetic and independent of the model-taint behavior; keeping it out avoids expanding this PR's refactor surface.

12. Addressed in `958ff92ac0`. Async FileStore mutations now use `LOCK_EX | LOCK_NB` with bounded async retries instead of blocking a Tokio worker indefinitely. A current-thread regression test proves timers continue to make progress while another process holds the mutation lock.

13. Addressed in `958ff92ac0`. Model-taint CAS conflicts are capped at eight attempts and return `StoreError::Retry` after exhaustion. A deterministic conflicting-bucket test verifies the exact attempt bound and returned error.

lib/runtime/src/storage/kv/file.rs
```diff
@@ -203,6 +205,48 @@ pub struct Directory {
     owned_files: Arc<Mutex<HashSet<PathBuf>>>,
 }
 
+struct DirectoryMutationLock(File);
+
+impl DirectoryMutationLock {
```
Aphoh, line 212: Don't re-implement this. Use https://lib.rs/crates/file-lock.

lib/kv-router/src/scheduling/local.rs
```diff
@@ -117,17 +117,18 @@ where
         slots: &ActiveSequencesMultiWorker<P>,
         current_workers: HashMap<WorkerId, C>,
         last_workers: &mut Option<HashMap<WorkerId, C>>,
-    ) {
+    ) -> bool {
```
Aphoh: It's not clear what the semantic meaning of the return value is. Either use a small enum return value or rename the method.

Aphoh (changes requested):
SGLang side looks fine. Not totally sure about the other areas, but definitely switch to using `file-lock` for the file based discovery.

lib/runtime/src/storage/kv/file.rs
tmonty12: Addressed in 75be30fb20. Replaced the direct unsafe libc::flock calls with the existing fs4 dependency, which is flock-backed and preserves both blocking and nonblocking lock behavior plus owned-file RAII. The linked file-lock crate uses fcntl record locks, which would change same-process contention semantics and require a separate writable lock file. All focused FileStore tests pass.

lib/kv-router/src/scheduling/local.rs
tmonty12: Addressed in 75be30fb20. reconcile_worker_configs now returns the private WorkerConfigReconcileOutcome::{Unchanged, Applied, Rejected} enum, and the queue is refreshed and woken only for Applied. The focused worker-taint queue wakeup regression test passes.

Aphoh (approved):
LGTM

lib/runtime/src/discovery/mod.rs
```diff
@@ -1294,6 +1482,27 @@ pub trait Discovery: Send + Sync {
     /// Backend-specific raw registration implementation.
     async fn register_internal(&self, spec: DiscoverySpec) -> Result<DiscoveryInstance>;
 
+    /// Replace the caller-managed taints of this worker's existing base model card.
+    async fn update_model_taints(
+        &self,
+        id: ModelCardInstanceId,
+        taints: HashSet<String>,
+    ) -> Result<()> {
+        if id.model_suffix.is_some() {
```
jthomson04, line 1518: Non-blocking follow-up: could we enforce `id.instance_id == self.instance_id()` here? The method is documented as updating this worker, but any caller can pass another worker ID. Etcd writes the replacement under the calling client lease; FileStore likewise adds the path to the calling store `owned_files`, so the wrong worker can acquire lifecycle ownership and remove a live card on shutdown. A central ownership check plus a wrong-owner test would make the contract explicit.

lib/runtime/src/storage/kv/file.rs
```diff
@@ -387,6 +436,54 @@ impl Bucket for Directory {
...
+    async fn compare_and_replace(
+        &self,
+        key: &Key,
+        expected: bytes::Bytes,
+        value: bytes::Bytes,
+    ) -> Result<StoreOutcome, StoreError> {
+        let full_path = self.p.join(key.url_safe().as_ref());
+        let str_path = full_path.display().to_string();
+        let _mutation_lock = self.lock_mutations().await?;
+        let temp_path = self.write_temp_file(&value)?;
```
jthomson04: Non-blocking: could we move `write_temp_file` until after the target read and expected-value comparison? The mutation lock is already held, so missing-key and conflict paths can return without writing and deleting the full candidate payload. This also removes the cleanup branches for those expected failures.

lib/runtime/src/discovery/mod.rs
```diff
@@ -1224,6 +1194,224 @@ fn find_conflicting_model_name(
...
+    };
+
+    if existing == candidate {
+        return Ok(None);
+    }
+
+    if matches!(existing, DiscoveryInstance::Model { .. })
+        && matches!(candidate, DiscoveryInstance::Model { .. })
+    {
+        validate_model_taint_update(existing, candidate)?;
```
jthomson04: `validate_model_taint_update` already calls `model_card_without_taints` for both values. Could it return the validated taint sets so this path does not clone and parse both complete model cards a second time?

lib/runtime/src/discovery/mod.rs
```diff
@@ -1224,6 +1194,224 @@ fn find_conflicting_model_name(
...
+    let mut next = HashMap::with_capacity(current.len());
+
+    for (id, candidate) in current {
+        match classify_discovery_change(known.get(&id), &candidate) {
+            Ok(Some(event)) => {
+                events.push(event);
+                next.insert(id, candidate);
+            }
+            Ok(None) => {
+                next.insert(id.clone(), known.get(&id).cloned().unwrap_or(candidate));
```
jthomson04: Nit: compute the retained value first, then move `id` into `next.insert` instead of cloning it here.

lib/runtime/src/discovery/kube.rs
```diff
@@ -449,45 +502,40 @@ impl Discovery for KubeDiscoveryClient {
...
-                            if event_tx.send(Ok(DiscoveryEvent::Removed(id))).is_err() {
+                            if event_tx.send(Ok(event)).is_err() {
                                 tracing::debug!(stream_id = %stream_id, "Watch receiver dropped");
                                 return;
                             }
                         }
 
-                        known_endpoints = endpoint_instances(&current);
-                        known_ids = current.into_keys().collect();
+                        // Update known values.
```
jthomson04: Nit: this comment only restates the assignment below, so I would remove it.

jthomson04 (approved):
Approving. I left a few non-blocking correctness, efficiency, and cleanup follow-ups inline; none block this change.

lib/runtime/src/storage/kv/mem.rs
```diff
@@ -157,6 +157,33 @@ impl Bucket for MemoryBucketRef {
...
+            .get_mut(&self.name)
+            .ok_or_else(|| StoreError::MissingBucket(self.name.to_string()))?;
+        let entry = bucket
+            .data
+            .get_mut(&key.0)
+            .ok_or_else(|| StoreError::MissingKey(key.to_string()))?;
+        if entry.1 != expected {
+            return Err(StoreError::Retry);
+        }
+        let next_revision = entry.0.saturating_add(1).max(1);
```
PeaBrane, line 177: [P1] Preserve dynamic taints across same-ID registration replay. This CAS advances Memory entries from registration revision 0 to 1; a later create-only `register(..., revision=0)` then takes the occupied revision-mismatch branch and overwrites the updated card with its original taints. Kubernetes likewise unconditionally replaces the model card on same-ID registration, while Mock appends a duplicate and can mutate a record hidden by the later duplicate. The supported sequence register A → update B → retry register A therefore reverts to A on Memory/Kubernetes and can hide B on Mock, while other backends retain B. Please make same-ID registration preserve the authoritative mutable taints (or reject a non-identical replay) and enforce one Mock record per ID.

lib/runtime/src/discovery/kube.rs
```diff
@@ -37,6 +38,30 @@ fn validate_kubernetes_publisher_id(publisher_id: u64) -> Result<()> {
...
+    // Serialize metadata mutations through persistence, but leave the shared state untouched
+    // until persistence succeeds. Dropping this future at any await point therefore cannot
+    // leave local metadata ahead of the authoritative Kubernetes object.
+    let mut metadata = metadata.write().await;
+    let mut candidate = metadata.clone();
+    if !candidate.update_model_taints(&id, taints)? {
+        return Ok(false);
+    }
+
+    let persisted = persist(candidate).await?;
```
PeaBrane: [P1] Close the remote-commit/local-state cancellation window. `apply_cr` can commit B at the API server and this task can still be cancelled before `persisted` is copied into `self.metadata`, leaving local A while discovery observes B. A later request to restore A then returns early as a local no-op and never repairs the CR. The added cancellation test covers cancellation before any external effect, not this commit/ack ambiguity. Please shield persistence from request cancellation or reconcile the authoritative CR before no-op detection.

lib/runtime/src/storage/kv/file.rs
```diff
@@ -437,13 +535,21 @@ impl Bucket for Directory {
             .watch(&self.p, RecursiveMode::NonRecursive)
             .map_err(to_fs_err)?;
 
+        // Establish the filesystem watcher before reading the snapshot. Mutations that race
+        // with the snapshot are buffered by notify; event values are read when consumed, so
+        // replaying a buffered notification cannot regress an emitted snapshot value.
+        let initial_entries = self.entries().await?;
```
PeaBrane, line 530: [P2] Add a sequence boundary between this snapshot and buffered notify events. If a key is deleted and recreated after the watcher is installed but before `entries()` completes, the snapshot emits the recreated value, then the buffered pre-snapshot `Remove` emits `Delete`, and only afterward does `Create` restore it. Discovery sees Added → Removed → Added and the transient removal can clear router worker/request state. Re-read or suppress `Remove` when the file currently exists, or publish an authoritative resync/cursor so pre-snapshot deletes cannot follow the snapshot.

lib/backend-common/src/worker.rs
```diff
@@ -758,6 +761,17 @@ impl Worker {
         Ok(())
     }
 
+    /// Register the Dynamo-owned model taint update on the runtime system server.
+    ///
+    /// Unlike engine-advertised updates, this mutates the worker's discovery
+    /// metadata and therefore applies uniformly to every engine implementation.
+    fn register_model_taint_update_route(&self, endpoint: &dynamo_runtime::component::Endpoint) {
+        endpoint.drt().engine_routes().register(
```
PeaBrane, line 778: [P2] Reserve `update/model_taints` before registering this built-in route. An engine can advertise an update named `model_taints`; `register_engine_updates` installs `update/model_taints` first, then this call overwrites that callback and startup still succeeds with only a warning. SGLang now rejects the equivalent configured-route collision. Please reject the reserved update name in the common worker as well so engine-declared behavior is not silently discarded.

lib/kv-router/src/scheduling/local.rs
```diff
@@ -222,16 +234,19 @@ where
...
                     let current_workers = monitor_rx.borrow_and_update().clone();
-                    Self::reconcile_worker_configs(
+                    if Self::reconcile_worker_configs(
                         &slots_monitor,
                         current_workers,
                         &mut last_workers,
-                    );
+                    ) == WorkerConfigReconcileOutcome::Applied
+                    {
+                        queue_config_updates.update().await;
```
PeaBrane, line 251: [P2] Avoid the fleet-wide fallback for a one-worker taint change. Every update clones and compares the full worker map, rebuilds all DP ranges, reconciles the complete topology, and then calls the all-workers queue update here. N sequential updates across W workers therefore cost O(N × (W + total DP ranks)); a fleet rollout is quadratic and repeatedly rechecks blocked lanes. Please retain dirty worker IDs, skip topology reconciliation for taint-only changes, call `update_worker` for affected ranks, and coalesce burst updates.

PeaBrane (commented):
AI-assisted router-only review. I left five inline findings covering discovery authority and event ordering, route reservation, and scheduler update amplification.

PeaBrane (approved):
LGTM

PeaBrane (commented):
Compatibility note for posterity and future AI-assisted work: this PR and #12890 are semantically compatible and fit the same facts → derived-state pattern well.

#12890 treats structural model discovery/materialization and E/P/D rendezvous as level-triggered projections of current committed worker facts. This PR treats worker taints as a separate mutable runtime-routing projection: `ModelTaintsUpdated` updates current per-worker configuration without changing `mdcsum()`, WorkerSet materialization identity, or rendezvous topology.

The expected integration is small: the structural controller explicitly ignores `ModelTaintsUpdated`, while the runtime-config watcher consumes it and wakes scheduler re-evaluation. No major semantic conflict resolution is foreseen; the two paths have separate authority domains and compose cleanly.

Cross-reference: https://github.com/ai-dynamo/dynamo/pull/12890

lib/runtime/src/discovery/mod.rs
tmonty12: Fixed in 46db6aba71. `validate_model_taint_update` now returns the validated existing and candidate taint sets, and classification reuses them instead of cloning and parsing both model cards again.

lib/runtime/src/storage/kv/mem.rs
tmonty12: Fixed in 46db6aba71. Revision-zero Memory inserts are now create-only; KV, Kubernetes metadata, and Mock same-ID model registration preserve the authoritative taints; and Mock keeps one record per ID. Regression coverage exercises `register A -> update B -> register A` and verifies B remains authoritative.

lib/runtime/src/discovery/mod.rs
tmonty12: Fixed in 46db6aba71. `Discovery::update_model_taints` now rejects a model ID whose `instance_id` differs from the discovery client owner, with a regression test confirming a foreign worker ID cannot mutate state.

lib/kv-router/src/scheduling/local.rs
tmonty12: Agreed that the fleet-wide fallback should be replaced before bulk taint rollouts. I am deferring this to #12050 because dirty-worker tracking, taint-only versus topology reconciliation, burst coalescing, and rank-targeted queue updates require a broader scheduler update contract than Part 1. Keeping this thread open for that follow-up.

lib/backend-common/src/worker.rs
tmonty12: Fixed in 46db6aba71. The common worker validates all advertised update names before registering any route and rejects `model_taints` as reserved. The integration test also verifies an earlier valid engine update is not partially registered.

hhzhang16 (approved):
approved as GMS codeowner

nv-anants (approved):
reviewed deps changes, lgtm!

KrishnanPrash (approved):
Approved to unblock

PR #12624, opened 2026-08-04
feat(llm): support root-level thinking_token_budget for chat and responses

### LINKED ISSUE #12302: Pass through thinking_token_budget from OpenAI-compatible requests
## Summary

Allow OpenAI-compatible chat/Responses requests to pass a root-level `thinking_token_budget` through preprocessing into backend sampling parameters when the backend supports it.

## Motivation

Some reasoning models expose a per-request budget for generated reasoning tokens. Dynamo already has lower-level support for a `thinking_token_budget`-style sampling field in some paths, but OpenAI-compatible chat/Responses ingress should be able to carry the same request-level control without the frontend validator stripping it as unsupported.

This is useful for generic reasoning-model deployments where operators need to bound runaway reasoning without changing the model's default configuration.

## Proposed API shape

Accept a root-level optional integer on OpenAI-compatible chat requests, for example:

```json
{
  "model": "dummy-reasoning-model",
  "messages": [
    {"role": "user", "content": "Return a short answer for this synthetic prompt."}
  ],
  "max_completion_tokens": 128,
  "thinking_token_budget": 32
}
```

Equivalent Responses API ingress should preserve the field if the request is converted through chat-completion internals.

## Expected behavior

- Omitted field remains `None` / unset, so existing defaults still apply.
- Present field is validated as a non-negative integer within backend-supported bounds.
- The value is forwarded through preprocessing to `SamplingParams.thinking_token_budget` or the equivalent backend request field.
- Unsupported backends should either ignore it explicitly according to existing backend-option policy or return a clear validation error; it should not be silently dropped before backend selection.

## Reproduction scenario

1. Send the dummy request above to a synthetic reasoning-model setup.
2. Inspect the preprocessed/backend request object.
3. Verify `thinking_token_budget: 32` reaches the sampling parameters.
4. Repeat with the field omitted and verify no budget override is set.

The payload is intentionally synthetic. Do not include real prompts, customer names, internal request IDs, URLs, credentials, or proprietary model names in reproductions.

flpanbin:
## Overview:

Implement root-level `thinking_token_budget` support for OpenAI-compatible chat completion and Responses requests, forwarding it to the backend's `thinking_token_budget` / `max_thinking_tokens` sampling parameter while preserving the legacy `nvext.max_thinking_tokens` passthrough. 

## Details:

- Added `thinking_token_budget` to `NvCreateChatCompletionRequest` and `NvCreateResponse`.
- Added `get_thinking_token_budget()` and implemented it for both request types.
- Updated the Python frontend vLLM pre/post processor to read the root-level field.
- Added unit tests for chat completion and Responses conversion.

- `lib/llm/src/protocols/openai/chat_completions.rs` — new field and `OpenAIStopConditionsProvider` impl.
- `lib/llm/src/protocols/openai/responses/mod.rs` — Responses API request support and conversion.
- `lib/llm/src/protocols/openai.rs` — stop-conditions mapping precedence logic.
- `lib/llm/src/protocols/openai/validate.rs` — validation hook.
- `components/src/dynamo/frontend/vllm_processor.py` — frontend passthrough.
- `tests/frontend/test_vllm_prepost_integration.py` — Python integration tests.

### Automated checks

- `cargo fmt --package dynamo-llm`
- `cargo check -p dynamo-llm`
- `cargo test -p dynamo-llm thinking_token_budget` (5 passed)
- `cargo clippy -p dynamo-llm --all-targets`
- `python3 -m pytest tests/frontend/test_vllm_prepost_integration.py`

### Manual E2E (Frontend + vLLM worker, `Qwen/Qwen3-0.6B`)

1. **Root-level `thinking_token_budget` is enforced** — request with `thinking_token_budget: 16` returns `reasoning_tokens: 17`):

```bash
curl -s http://<frontend>:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model "Qwen3-0.6B",
    "messages": [{"role": "user", "content": "What is 37 times 41? Please reason step by step."}],
    "chat_template_kwargs": {"enable_thinking": true},
    "max_tokens": 1000,
    "temperature": 0,
    "thinking_token_budget": 16
  }'
```

Response:

```json
......
"usage": {
        "prompt_tokens": 24,
        "completion_tokens": 1000,
        "total_tokens": 1024,
        "completion_tokens_details": {
            "reasoning_tokens": 17
        }
    }
```

2. **Omitted field keeps existing defaults** — the same request without `thinking_token_budget` completes normally with no budget override applied.

## Related Issues

- Closes #12302 

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12624" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a thinking token budget directly in chat and response requests.
  * The direct budget setting takes precedence over the legacy configuration when both are provided.
  * The setting is preserved across supported request formats and forwarded to generation controls.

* **Tests**
  * Added coverage for propagation, precedence, fallback behavior, and omission of the thinking token budget.

### DIFF
--- a/lib/llm/src/protocols/anthropic/types.rs
+++ b/lib/llm/src/protocols/anthropic/types.rs
@@ -163,6 +163,7 @@ impl TryFrom<AnthropicCreateMessageRequest> for NvCreateChatCompletionRequest {
             media_io_kwargs: None,
             return_tokens_as_token_ids: None,
             unsupported_fields: Default::default(),
+            ..Default::default()
         })
     }
 }
--- a/lib/llm/src/protocols/openai/responses/mod.rs
+++ b/lib/llm/src/protocols/openai/responses/mod.rs
@@ -45,7 +45,7 @@ use crate::protocols::common::extensions::{NvExt, NvExtProvider};
 /// that walked `serde_json::Value` to inject synthetic defaults for missing
 /// `id` / `status` / `annotations`; that was replaced by typed ownership for
 /// correctness and to avoid the double-deserialize cost.
-#[derive(ToSchema, Serialize, Deserialize, Validate, Debug, Clone)]
+#[derive(ToSchema, Serialize, Deserialize, Validate, Debug, Clone, Default)]
 pub struct NvCreateResponse {
     /// Flattened CreateResponse fields (model, input, temperature, etc.).
     ///
@@ -63,6 +63,12 @@ pub struct NvCreateResponse {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Object)]
     pub nvext: Option<NvExt>,
+
+    /// OpenAI-style reasoning token budget: bounds the number of reasoning
+    /// (thinking) tokens generated per request. Forwarded to the backend's
+    /// `thinking_token_budget` sampling parameter when supported.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub thinking_token_budget: Option<u32>,
 }
 
 #[derive(ToSchema, Deserialize, Validate, Debug, Clone)]
@@ -229,6 +235,10 @@ impl OpenAIStopConditionsProvider for NvCreateResponse {
 
     fn nvext(&self) -> Option<&NvExt> {
         self.nvext.as_ref()
+    }
+
+    fn get_thinking_token_budget(&self) -> Option<u32> {
+        self.thinking_token_budget
     }
 }
 
@@ -805,6 +815,7 @@ impl TryFrom<NvCreateResponse> for NvCreateChatCompletionRequest {
             nvext: resp.nvext,
             chat_template_args: None,
             thinking: None,
+            thinking_token_budget: resp.thinking_token_budget,
             media_io_kwargs: None,
             return_tokens_as_token_ids: None,
             unsupported_fields: Default::default(),
@@ -1207,6 +1218,7 @@ mod tests {
     };
 
     use super::*;
+    use crate::protocols::common::StopConditionsProvider;
     use crate::types::openai::chat_completions::NvCreateChatCompletionResponse;
 
     fn make_response_with_input(text: &str) -> NvCreateResponse {
@@ -1224,6 +1236,7 @@ mod tests {
                 annotations: Some(vec!["debug".into(), "trace".into()]),
                 ..Default::default()
             }),
+            ..Default::default()
         }
     }
 
@@ -1310,6 +1323,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1354,6 +1368,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1433,6 +1448,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1474,6 +1490,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1517,6 +1534,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1565,6 +1583,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1604,6 +1623,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
@@ -1638,6 +1658,7 @@ mod tests {
                 ..Default::default()
             },
             nvext: None,
+            ..Default::default()
         };
 
         let chat_req: NvCreateChatCompletionRequest = req.try_into().unwrap();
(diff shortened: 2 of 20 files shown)


pskiran1:
@flpanbin, please resolve conflicts and rebase with main.

flpanbin:
> @flpanbin, please resolve conflicts and rebase with main.

@pskiran1 Done! I've rebased onto main and resolved all conflicts. Please take another look when you have a moment. Thanks!

rmccorm4:
/ok to test [1fe3831](https://github.com/ai-dynamo/dynamo/pull/12624/commits/1fe38310b421e422a8b4911899f10d223004061a)

rmccorm4:
/ok to test [0e29b98](https://github.com/ai-dynamo/dynamo/pull/12624/commits/0e29b98fe165daece9d67cbcc228305f0ff00a61)

flpanbin:
Hi @rmccorm4 thanks for merging main.
I see some jobs are failing with compile errors, likely due to the merge conflict resolution. I’ll investigate and push a fix.

biswapanda (commented):
Review feedback on the current head.

lib/llm/src/protocols/openai/responses/mod.rs
```diff
@@ -3843,4 +3877,57 @@ thinking
         // nvext should be omitted when None
         assert!(json.get("nvext").is_none());
     }
+
+    #[test]
+    fn test_thinking_token_budget_preserved_in_chat_completion_conversion() {
+        let mut req = make_response_with_input("hi there");
```
biswapanda, line 3883: Please add an end-to-end `/v1/responses` integration test here or alongside the frontend integration tests. Post a JSON payload with `thinking_token_budget` and assert that the captured worker request contains `stop_conditions.max_thinking_tokens`; also cover root-level precedence over `nvext.max_thinking_tokens`. These unit tests construct `NvCreateResponse` directly, so they do not exercise JSON deserialization, Responses validation, `UnifiedRequest` conversion, or frontend-to-worker forwarding.

lib/llm/src/protocols/openai/validate.rs
```diff
@@ -223,6 +223,13 @@ pub fn validate_response_format(
     }
 }
 
+/// Validates the `thinking_token_budget` parameter.
+///
+/// It must be a non-negative integer when present.
+pub fn validate_thinking_token_budget(_value: Option<u32>) -> Result<(), anyhow::Error> {
```
biswapanda, line 229: This function never inspects the value and always returns `Ok(())`. Since `Option<u32>` already rejects negative, non-integer, and overflowing JSON values during deserialization, please remove this helper and its call. If there is a backend-independent upper-bound invariant, enforce it here and add boundary tests instead.

lib/llm/src/protocols/anthropic/types.rs
```diff
@@ -174,6 +174,7 @@ impl TryFrom<AnthropicCreateMessageRequest> for NvCreateChatCompletionRequest {
             media_io_kwargs: None,
             return_tokens_as_token_ids: None,
             unsupported_fields: Default::default(),
+            ..Default::default()
```
biswapanda, line 177: Please keep this production conversion exhaustive by setting `thinking_token_budget: None` explicitly instead of using `..Default::default()`. The struct-update default allows future fields added to `NvCreateChatCompletionRequest` to be silently dropped here rather than producing a compiler error that forces an explicit conversion decision.

lib/llm/src/protocols/openai/responses/mod.rs
flpanbin: Integration tests have been added and verified. E2E test validation has also been completed, and the results have been added to the PR description.

lib/llm/src/protocols/openai/validate.rs
flpanbin: Removed the validation.

lib/llm/src/protocols/anthropic/types.rs
flpanbin: Fixed.

lib/llm/src/protocols/openai/responses/mod.rs
biswapanda: Thanks @flpanbin !

biswapanda:
@flpanbin please take a look - some CI tests are failing

lib/llm/src/protocols/unified.rs
```diff
@@ -550,6 +550,7 @@ mod tests {
             media_io_kwargs: None,
             return_tokens_as_token_ids: None,
             unsupported_fields: Default::default(),
```
biswapanda, line 557: lib/llm/src/protocols/unified.rs:R429-431

**Thinking budget is ignored for requests handled through the shared request wrapper**

The newly added thinking budget is not read back (`get_thinking_token_budget` is missing from the `UnifiedRequest` implementation at `lib/llm/src/protocols/unified.rs:397-431`) when a request goes through the shared wrapper used by the Anthropic and Responses entry points, so the budget silently falls back to the legacy value or to no limit at all.
Impact: A user-set thinking budget can be silently dropped on those request paths, and the model keeps reasoning without the requested cap.

<details>
<summary>Trait default masks the new field for UnifiedRequest</summary>

`OpenAIStopConditionsProvider::get_thinking_token_budget` has a default body returning `None` (`lib/llm/src/protocols/openai.rs:92-94`). `NvCreateChatCompletionRequest` (`lib/llm/src/protocols/openai/chat_completions.rs:509-512`) and `NvCreateResponse` (`lib/llm/src/protocols/openai/responses/mod.rs:240-242`) override it, but `UnifiedRequest`'s impl — which otherwise delegates every getter to `self.inner` — does not, so `extract_stop_conditions` on a `UnifiedRequest` yields `max_thinking_tokens` from `nvext` only. Today both HTTP call sites (`lib/llm/src/http/service/openai.rs:2576`, `lib/llm/src/http/service/anthropic.rs:316`) call `into_inner()` before preprocessing, so the gap is latent, but the wrapper is documented as the migration target for these paths and any future consumer using it directly will lose the field.

</details>

```suggestion
    fn get_common_ignore_eos(&self) -> Option<bool> {
        self.inner.common.ignore_eos
    }

    fn get_thinking_token_budget(&self) -> Option<u32> {
        self.inner.thinking_token_budget
    }
```

biswapanda: fixed in [152bba5](https://github.com/ai-dynamo/dynamo/pull/12624/commits/152bba5fd502c9420f027be3a7d7fd6891f41676)

biswapanda (approved):

lib/llm/src/protocols/unified.rs
flpanbin: Thanks for the fix, really appreciate it!

Issue #12035, opened 2026-07-22
DEP: Decouple operator workload pathways from external orchestrators


julienmancuso:
## Summary

Decouple the Dynamo Kubernetes operator from Grove, LeaderWorkerSet (LWS), DisaggregatedSet, and future workload orchestrators without pretending that they implement the same state machine.

This DEP considers two architectural approaches:

1. A **framework-style pathway abstraction**, where common controller logic drives providers through a uniform lifecycle.
2. A **composition-first architecture**, where each pathway owns its complete reconciliation state machine and reuses smaller building blocks with genuinely shared invariants.

The recommended direction is composition-first. Dynamo keeps a thin, sticky graph-level pathway selector and common user-facing intent, but the selected pathway is a complete reconciliation program rather than a collection of callbacks inside a universal rollout and readiness framework. The initial refactor keeps one controller-runtime loop as the owner of DGD finalization and status while composing independently meaningful reconcilers through ephemeral state.

## Motivation

The operator currently integrates external workload APIs at several layers: pathway selection, resource rendering, scaling, readiness, watches, topology handling, status aggregation, rollout handling, and backend launch configuration. Changes to Grove or LWS can therefore affect broad portions of the operator, and adding a workload pathway is expensive to do safely.

The integrations also operate at different abstraction levels:

- Grove realizes a graph through PodCliqueSet-related resources.
- DisaggregatedSet realizes selected multi-role workers through child LeaderWorkerSets while other components may continue through DynamoComponentDeployments.
- The DynamoComponentDeployment path realizes each component through either a Deployment or a LeaderWorkerSet.

These systems use similar terms such as revision, ready, group, and rolling update, but the terms do not necessarily have the same semantics. They differ in rollout atomicity, partial readiness, scaling during rollout, revision ownership, stale-observation detection, and failure handling.

The existing code reflects these differences. The DGD controller owns graph finalization and status, reconciles common resources, and then branches into workload-specific control flow. The DCD controller is already a separate persisted boundary and independently selects Deployment or LWS behavior. Operator-managed rolling update behavior is currently specific to the DCD/Deployment path rather than a universal provider algorithm.

The architectural problem is therefore not only where provider API types live. It is **who owns control flow and the state machine**.

## Goals

- Add or update a workload pathway without spreading provider-specific logic through the DGD controller.
- Keep provider-specific API types, naming rules, watches, runtime variables, compatibility behavior, and status interpretation in focused packages.
[...]

PR #12625, opened 2026-08-04
feat(operator): add DisaggregatedSet workload program

panpan0000:
## Summary

Rebased onto `main` after ai-dynamo/dynamo#12942 and integrated DisaggregatedSet as the third durable DGD workload provider.

* Support durable provider values `component`, `grove`, and `disaggregatedset`.
* Keep Grove as the default when enabled, while DS remains explicit opt-in after Grove is not selected.
* Materialize `nvidia.com/workload-provider` before workload side effects and adopt legacy DGDs from their owned DCD, PodCliqueSet, or DisaggregatedSet resources.
* Treat provider selection as immutable. The enable annotations are creation/adoption-time intent only; existing DGDs do not migrate between providers.
* Report `Ready=False` with `selected_workload_provider_unavailable` when the selected provider is unavailable; there is no provider fallback or cross-provider cleanup.
* Preserve DS reconciliation, readiness, revision-aware rollout, Service management, restart, checkpoint, and LWS v0.10 behavior. DS continues to use the LWS default of `slices=1` until the grouping/cardinality API is available.
* Remove the parallel runtime `Capabilities` concept and update operator tests and multinode documentation for the durable-provider model.

## Review focus

* `deploy/operator/internal/controller/dgd_workload_provider.go`
* `deploy/operator/internal/controller/dgd_disaggregatedset_program.go`
* `deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go`
* `deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go`

## Validation

* `go test ./internal/controller -count=1` — passed.
* `go test ./... -count=1` — all non-e2e operator packages passed; the e2e package was blocked during Docker image metadata resolution because `nvcr.io` and the configured Docker mirror returned HTTP 403.
* `git diff --check` — passed.

## Related Issues

* Relates to ai-dynamo/dynamo#12035.
* Relates to ai-dynamo/dynamo#10205


panpan0000:
Thanks for the detailed checks.

Updates pushed:
- Fixed gofmt indentation in operatorenv CRD paths (commit 0b80a19373).
- DS fallback visibility: record a DGD condition (DisaggregatedSetEligible) and removed the redundant pre-fallback condition write (commits 7b4925ec82, 0b80a19373).
- Stabilized DS component Service cleanup: keep the currently desired DGD-owned Services for selected roles, and only delete stale DGD-owned component Services for selected roles (commit f97e50a6dc).
- Added a nil guard when applying checkpoint startup policies for selected roles (commit f97e50a6dc).
- Added small t.Log headings to multi-step tests for readability (commit f97e50a6dc).

The earlier CI warning about an unused variable was on an older SHA and should be cleared by these commits.

panpan0000:
Replying to CodeRabbit summary (https://github.com/ai-dynamo/dynamo/pull/12625#issuecomment-5175424178):\n\n- Fallback observability: added DisaggregatedSetEligible condition + warning event (7b4925ec82).\n- Service cleanup: no longer deletes active worker endpoints; only removes stale DS-selected services (f97e50a6dc).\n- Fallback safety: DS cleanup deferred until after component fallback reconcile succeeds (f97e50a6dc).\n- Feature gate: DS now honors explicit LWS opt-out (400f7ca51d).\n\nThanks for the review.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,154 @@
...
+			r.Config,
+			r.RuntimeConfig,
+			r.RestConfig,
+			r.DockerSecretRetriever,
+			r.SSHKeyManager,
+			r.RBACManager,
+		),
+		rollout:         rollout,
+		restart:         newDGDRestartReconciler(),
+		workloads:       newDisaggregatedSetWorkloadsReconciler(r, rollout),
```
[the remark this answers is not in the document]
panpan0000: Thanks. Refactored the DS program/workloads to depend on explicit narrow owner interfaces rather than the full DGD reconciler. See 400f7ca51d.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,154 @@
...
+			return programResult, retErr
+		}
+		if err := p.workloads.owner.restoreDisaggregatedSetServiceOwnershipToDCDs(ctx, req.DGD); err != nil {
+			retErr = fmt.Errorf("failed to restore DisaggregatedSet service ownership: %w", err)
+			return programResult, retErr
+		}
+		if err := p.workloads.owner.deleteDisaggregatedSetIfExists(ctx, req.DGD); err != nil {
+			retErr = fmt.Errorf("failed to delete stale DisaggregatedSet: %w", err)
+			return programResult, retErr
+		}
```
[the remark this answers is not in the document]
panpan0000: Good catch. DS cleanup (service ownership restore + DS delete) now happens only after the fallback component program reconciles successfully, to avoid downtime. See f97e50a6dc.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func newDisaggregatedSetWorkloadsReconciler(
+	owner *DynamoGraphDeploymentReconciler,
+	rollout *dgdWorkerRolloutReconciler,
+) *disaggregatedSetWorkloadsReconciler {
+	return &disaggregatedSetWorkloadsReconciler{
+		owner:   owner,
+		rollout: rollout,
+	}
```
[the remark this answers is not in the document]
panpan0000: Refactored workloads reconciler to depend on a small owner interface, reducing coupling to the outer controller. See 400f7ca51d.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_test.go
```diff
@@ -0,0 +1,197 @@
...
+	require.Contains(t, reason, "decode")
+	require.Equal(t, int32(1), ptr.Deref(statuses["decode"].ReadyReplicas, 0))
+
+	ds.Object["status"].(map[string]any)["roleStatuses"] = []any{
+		map[string]any{"name": "prefill", "replicas": int64(2), "updatedReplicas": int64(2), "readyReplicas": int64(2)},
+		map[string]any{"name": "decode", "replicas": int64(2), "updatedReplicas": int64(2), "readyReplicas": int64(2)},
+	}
+	ready, _, _ = checkDisaggregatedSetReadiness(ds, selection)
+	require.True(t, ready)
+}
```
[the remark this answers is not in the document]
panpan0000: Added step headings (t.Log) for the test flow to improve readability. See f97e50a6dc.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1488 @@
...
+	}
+
+	if err := r.reconcileDisaggregatedSetSideResources(ctx, dgd, dcds, selection); err != nil {
+		return ReconcileResult{}, err
+	}
+	dsReady, dsReason, dsStatuses, err := r.checkDisaggregatedSetReadiness(ctx, syncedDS, selection)
+	if err != nil {
+		return ReconcileResult{}, err
+	}
+	dsReady = dsReady && !dsModified && !checkpointGated
```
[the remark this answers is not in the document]
panpan0000: Fixed. Stale Service cleanup is now limited to DS-selected component services that are no longer desired (no longer deletes active endpoints). See f97e50a6dc.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1488 @@
...
+			gateSelectedRoles = true
+			break
+		}
+	}
+	for componentName := range selection.componentToRole {
+		if gateSelectedRoles {
+			dcds[componentName].Spec.Replicas = ptr.To(int32(0))
+		}
+		selection.desiredReplicas[componentName] = desiredComponentReplicas(&dcds[componentName].Spec.DynamoComponentDeploymentSharedSpec)
+	}
```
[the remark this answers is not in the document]
panpan0000: Fixed. Added a nil guard before dereferencing generated DCDs when applying checkpoint startup policies. See f97e50a6dc.

deploy/operator/internal/testing/operatorenv/env.go
```diff
@@ -493,6 +493,7 @@ func crdDirectoryPaths(opts Options) []string {
 	root := operatorRoot()
 	return []string{
 		filepath.Join(root, "config", "crd", "bases"),
+                filepath.Join(root, "internal", "controller", "testing", "disaggregatedset"),
```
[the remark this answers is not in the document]
panpan0000: Fixed gofmt indentation (tab) in env.go. See 0b80a19373.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1488 @@
...
+			return ReconcileResult{}, err
+		}
+	}
+
+	result := r.checkResourcesReadiness(resources)
+	if result.State == nvidiacomv1beta1.DGDStateSuccessful {
+		if err := r.deleteStaleDisaggregatedSetComponentServices(ctx, dgd, dcds); err != nil {
+			return ReconcileResult{}, err
+		}
+	}
```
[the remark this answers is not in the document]
panpan0000: Fixed. DS pathway no longer deletes desired worker services every time the deployment becomes ready; cleanup now targets only stale services. See f97e50a6dc.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,154 @@
...
+		return programResult, nil
```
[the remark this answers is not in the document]
panpan0000: Fixed. DS fallback no longer tears down running multinode workers before replacements are ready; DS cleanup is deferred until after the component fallback reconcile succeeds. See f97e50a6dc.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+}
```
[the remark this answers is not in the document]
panpan0000: Refactored. DS workloads reconciler now depends on narrow interfaces rather than the full graph reconciler. See 400f7ca51d.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,154 @@
...
+	}
```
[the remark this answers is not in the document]
panpan0000: Kept intentionally. When DS is requested but not eligible, DS program invokes the component program to keep the reconcile entrypoint consistent. Added a DGD condition (DisaggregatedSetEligible) + warning event for observability, and deferred DS cleanup until fallback succeeds. See 7b4925ec82 and f97e50a6dc.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1488 @@
...
+}
+
+func disaggregatedSetDesiredStateEqual(a, b *unstructured.Unstructured) bool {
+	if a == nil || b == nil {
+		return a == b
+	}
+	return equality.Semantic.DeepEqual(a.GetLabels(), b.GetLabels()) &&
+		equality.Semantic.DeepEqual(a.GetOwnerReferences(), b.GetOwnerReferences()) &&
+		equality.Semantic.DeepEqual(a.Object["spec"], b.Object["spec"])
+}
```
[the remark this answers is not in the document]
panpan0000: Acknowledged. We currently compare labels/ownerRefs/spec with semantic equality. I haven't observed update loops in envtest; if we see churn due to CRD defaulting we can follow up with normalization / SSA.

deploy/operator/internal/features/gates.go
```diff
@@ -198,6 +212,10 @@ func New(ctx context.Context, mgr ctrl.Manager, config *configv1alpha1.OperatorC
 
 	lwsAvailable := detectAPIGroup(ctx, mgr, "leaderworkerset.x-k8s.io", "")
 	volcanoAvailable := detectAPIGroup(ctx, mgr, "scheduling.volcano.sh", "")
+	disaggregatedSetAvailable := detectAPIGroup(ctx, mgr, "disaggregatedset.x-k8s.io", "v1")
+	// The DS pathway lists and watches the LWS children created by the DS
+	// controller. Do not register those watches when the LWS API is absent.
+	gates.DisaggregatedSet = lwsAvailable && disaggregatedSetAvailable
```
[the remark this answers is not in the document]
panpan0000: Fixed. DisaggregatedSet gate now honors explicit LWS opt-out (config.orchestrators.lws.enabled=false) in addition to API detection. See 400f7ca51d.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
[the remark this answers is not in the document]
panpan0000: You were right — the cleanup was gated only on a nil error, not on the fallback result state. Fixed in d9ed7a5261: the fallback branch now returns early unless programResult.Status.State == DGDStateSuccessful, matching the legacyCleanup gating in the component/grove programs. Also added an envtest that keeps the DS annotation on, forces fallback (scalingAdapter), and asserts the DisaggregatedSet survives until the replacement DCDs are ready.

julienmancuso:
@coderabbitai full review

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1519 @@
...
+		return fmt.Errorf("failed to update owner references: %w", err)
+	}
+	return nil
+}
+
+func (r *DynamoGraphDeploymentReconciler) restoreDisaggregatedSetServiceOwnershipToDCDs(
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) error {
+	if r.RuntimeConfig == nil || !r.RuntimeConfig.Gate.Enabled(features.DisaggregatedSet) {
```
julienmancuso: **P1: Cleanup capability disappears with the selection gate**

`features.DisaggregatedSet` is being used both as the pathway-selection gate and as the signal that the API is available for compatibility cleanup. Those are different concerns. With the APIs installed but this gate disabled, `restoreDisaggregatedSetServiceOwnershipToDCDs` returns here and `cleanupDisaggregatedSetOnLegacyPath` also skips deletion. A DGD that removes the annotation or selects Grove can therefore leave its old DisaggregatedSet behind. There is a second unsafe case: if the annotation remains but the DisaggregatedSet program falls back to components, the later direct delete can still run after this ownership restoration returned, deleting the wrapper while its Services remain DGD-owned. Please separate API availability/cleanup capability from pathway selection and cover component, Grove, and annotated-fallback transitions with the API present but selection disabled.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1519 @@
...
+			dgd.Namespace,
+			dgd.Name,
+		)
+	}
+
+	desired.SetResourceVersion(current.GetResourceVersion())
+	desired.SetUID(current.GetUID())
+	desired.SetCreationTimestamp(current.GetCreationTimestamp())
+	if !disaggregatedSetDesiredStateEqual(current, desired) {
+		if err := r.Update(ctx, desired); err != nil {
```
julienmancuso: **P1: Full-object DisaggregatedSet updates may never converge against the real CRD**

This compares a freshly rendered object with the live object, including the full label set, owner references, and spec, then updates using the fresh desired object while preserving only a few identity fields. Server-side/defaulted fields in the real LWS-backed CRD can therefore make every reconcile look modified, while unmanaged labels, annotations, finalizers, and non-controller owner references can be dropped. Because `dsModified` suppresses readiness, that can prevent the workload from ever becoming Ready. The schemaless test fixture does not exercise structural CRD defaulting. Please restore dry-run/default normalization (as in the earlier implementation), mutate only fields owned by this controller on a copy of the live object, preserve unrelated metadata, and add a convergence test using the real CRD schema/defaulting behavior.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,175 @@
...
+		reason := reasonFailedToReconcileResources
+		if classified, ok := workloadProgramFailureReason(retErr); ok {
+			reason = classified
+		}
+		programResult.Fail(req.DGD.Generation, reason, retErr)
+	}()
+
+	useDS, fallbackReason := p.owner.shouldUseDisaggregatedSet(req.DGD)
+	if !useDS {
+		fallbackProgram := p.owner.newComponentProgram()
```
julienmancuso: **P2: A complete DisaggregatedSet program invokes the complete component program**

This violates the composition-first boundary: the outer reconciler must select exactly one complete workload program, and one program must not invoke another. The outer selector currently considers only the annotation, while eligibility is decided inside this program and fallback is implemented by calling the component program. Please move `shouldUseDisaggregatedSet` into outer program selection. On fallback, select and run only the component program, with any DisaggregatedSet diagnostic or compatibility cleanup supplied as a focused collaborator. A test should assert that one outer reconcile invokes exactly one program.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,175 @@
...
+			r.Config,
+			r.RuntimeConfig,
+			r.RestConfig,
+			r.DockerSecretRetriever,
+			r.SSHKeyManager,
+			r.RBACManager,
+		),
+		rollout:         rollout,
+		restart:         newDGDRestartReconciler(),
+		workloads:       newDisaggregatedSetWorkloadsReconciler(r, rollout),
```
julienmancuso: **P2: Narrow owner interfaces still hide the all-capable outer reconciler**

The constructor passes the outer `DynamoGraphDeploymentReconciler` into these interfaces, and the large DisaggregatedSet file implements all provider lifecycle operations as methods on that reconciler. This is static interface narrowing, but not dependency or lifecycle isolation; it also leads to ad hoc construction of component/rolling-update helpers and leaves unrelated fields coupled to the program. Please extract concrete collaborators for rendering, DisaggregatedSet synchronization, mixed DCD reconciliation, Service ownership, readiness, restart progress, compatibility cleanup, and watch setup, each with explicit dependencies. The outer reconciler should orchestrate and persist results, not implement the provider lifecycle behind a broad façade.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1519 @@
...
+	for i := range serviceList.Items {
+		service := &serviceList.Items[i]
+		if !isControlledByBetaDGD(service, dgd) {
+			continue
+		}
+		componentName := service.Labels[consts.KubeLabelDynamoComponent]
+		if componentName == "" {
+			continue
+		}
+		if _, selected := selection.componentToRole[componentName]; !selected {
```
julienmancuso: **P2: Roles leaving DisaggregatedSet never return their Services to DCD ownership**

Consider a DGD where three roles were selected and an update makes one role ineligible while two remain selected. The program still runs; the ineligible role is reconciled as a DCD, but its existing Service remains controller-owned by the DGD. Cleanup skips it because it is no longer selected, and normal DCD synchronization does not repair the owner of an already-existing Service. That leaves the DCD controller without ownership/watch semantics and can prevent recreation after deletion. Please reconcile Service ownership from an explicit desired map on every pass: selected roles are DGD-owned, non-selected roles are DCD-owned, removed roles are deleted, with model-service sharing handled deliberately. Add coverage for partial selection changes and role removal.

deploy/operator/internal/controller/dgd_component_program.go
```diff
@@ -133,6 +137,11 @@ func (p *componentProgram) Reconcile(
 			return programResult, fmt.Errorf("failed to reconcile scaling adapters: %w", err)
 		}
 	}
+	if result.State == nvidiacomv1beta1.DGDStateSuccessful && p.legacyCleanup != nil {
```
julienmancuso: **P2: Annotation-only pathway switches preserve a current-looking DisaggregatedSet condition**

Annotations do not increment `.metadata.generation`. The component and Grove paths currently carry existing conditions forward, so after compatibility cleanup deletes the DisaggregatedSet, `DisaggregatedSetEligible=True` with reason `Selected` can remain at the current observed generation. That makes the returned status non-authoritative. The selected replacement program should explicitly remove or replace the provider-specific condition after cleanup; fallback warnings should also be emitted only on a persisted transition. Please add annotation-only DisaggregatedSet-to-component and DisaggregatedSet-to-Grove status tests.

julienmancuso (commented):
I reviewed this against the composition-first DGD rules in `deploy/operator/internal/AGENTS.md` and the behavior carried forward from the earlier DisaggregatedSet implementation.

The third-program direction is right, but I found two P1 correctness blockers and four P2 correctness/architecture blockers. In particular, the existing feedback about program nesting and all-capable-controller coupling is still valid: narrowing the static interface does not resolve it while the implementation remains `DynamoGraphDeploymentReconciler`.

Details are inline. I would address the P1s first, then restructure the DisaggregatedSet program around concrete collaborators before merge.

julienmancuso:
also, additional scope gaps:
- The user-facing documentation changes from #11645 were not ported.
- Most of #11645’s focused unit coverage was dropped, including the real upstream-CRD test and detailed ownership, restart, defaulting, idempotence, and readiness cases.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. API capability detection is now independent from the DisaggregatedSet selection gate, so Component/Grove cleanup remains available when the API exists. Covered by TestDisaggregatedSetCleanupUsesAPICapabilityWhenSelectionIsDisabled and the operator test suite.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. DS sync patches the live object, preserves unmanaged metadata/ownerReferences, and normalizes through server-side dry-run. Added pinned real-CRD convergence coverage: TestDisaggregatedSetRealLWSCRDValidationAndConvergence.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Eligibility is decided by the outer selector; it runs exactly one complete workload program, with no DS-to-Component program invocation. Covered by TestDGDWorkloadProgramSelection.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. DS lifecycle logic now lives on concrete workload/cleanup collaborators with explicit client/config/renderer dependencies; the outer reconciler is no longer retained as the DS owner.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Service ownership is reconciled from an explicit desired map for selected and non-selected roles, including handoff back to DCD ownership. Covered by the DS envtest ownership assertions.

deploy/operator/internal/controller/dgd_component_program.go
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Component/Grove remove the stale DS condition after successful cleanup, and fallback warnings are emitted only on condition transitions. Covered by the fallback status/error-path tests.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,175 @@
...
+	workloads       *disaggregatedSetWorkloadsReconciler
+	scalingAdapters *dgdScalingAdaptersReconciler
+}
+
+type disaggregatedSetProgramOwner interface {
+	shouldUseDisaggregatedSet(dgd *nvidiacomv1beta1.DynamoGraphDeployment) (bool, string)
+	newComponentProgram() *componentProgram
+	restoreDisaggregatedSetServiceOwnershipToDCDs(ctx context.Context, dgd *nvidiacomv1beta1.DynamoGraphDeployment) error
+	deleteDisaggregatedSetIfExists(ctx context.Context, dgd *nvidiacomv1beta1.DynamoGraphDeployment) error
+}
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. The outer selector now chooses one complete program; fallback/legacy cleanup is a focused collaborator and DS no longer constructs or invokes componentProgram. go test ./... -count=1 passes in deploy/operator.

deploy/operator/internal/controller/dgd_disaggregatedset_program.go
```diff
@@ -0,0 +1,175 @@
...
+				"DisaggregatedSetFallback",
+				"DisaggregatedSet requested but falling back to DynamoComponentDeployments: %s",
+				fallbackReason,
+			)
+		}
+		if retErr != nil {
+			return programResult, retErr
+		}
+		setDisaggregatedSetEligibilityCondition(&programResult, req.DGD.Generation, metav1.ConditionFalse, "FallbackToComponentProgram", fallbackReason)
+		if programResult.Status.State != nvidiacomv1beta1.DGDStateSuccessful {
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. The fallback eligibility condition is set before downstream errors and retained in the authoritative failure status. Covered by TestComponentFallbackPreservesEligibilityConditionOnError.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go
```diff
@@ -0,0 +1,494 @@
...
+			Expect(owner.Kind).To(Equal("DynamoComponentDeployment"))
+			Expect(owner.Name).To(Equal(dcd.Name))
+		}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: modelServiceName, Namespace: current.Namespace}, modelService)).To(Succeed())
+		Expect(modelService.UID).To(Equal(modelServiceUID))
+		modelOwner := metav1.GetControllerOf(modelService)
+		Expect(modelOwner).NotTo(BeNil())
+		Expect(modelOwner.Kind).To(Equal("DynamoComponentDeployment"))
+		Expect(fallbackDCDNames).To(HaveKey(modelOwner.Name))
+	})
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Envtest now asserts controller owner identity during pending fallback and after DS handoff, not only Service UID preservation.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go
```diff
@@ -0,0 +1,494 @@
...
+	ctx context.Context,
+	reconciler *DynamoGraphDeploymentReconciler,
+	name string,
+	namespace string,
+) (workloadProgramResult, *nvidiacomv1beta1.DynamoGraphDeployment) {
+	current := &nvidiacomv1beta1.DynamoGraphDeployment{}
+	Expect(k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, current)).To(Succeed())
+	result, err := reconciler.selectWorkloadProgram(current).Reconcile(ctx, workloadProgramRequest{DGD: current})
+	Expect(err).NotTo(HaveOccurred())
+	return result, current
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Added outer status persistence/event-order coverage while retaining program-only tests; the outer controller remains the sole DGD status writer.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go
```diff
@@ -0,0 +1,494 @@
...
+) {
+	for _, dcd := range ownedEnvtestCutoverDCDs(ctx, dgd) {
+		replicas := ptr.Deref(dcd.Spec.Replicas, int32(1))
+		dcd.Status.ObservedGeneration = dcd.Generation
+		dcd.Status.Conditions = []metav1.Condition{{
+			Type:               nvidiacomv1beta1.DynamoComponentDeploymentConditionTypeAvailable,
+			Status:             metav1.ConditionTrue,
+			Reason:             "CutoverTestReady",
+			LastTransitionTime: metav1.Now(),
+		}}
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Synthetic Available conditions now carry the expected ObservedGeneration; covered by the DS envtest readiness assertions.

deploy/operator/internal/features/gates.go
```diff
@@ -198,6 +212,11 @@ func New(ctx context.Context, mgr ctrl.Manager, config *configv1alpha1.OperatorC
 
 	lwsAvailable := detectAPIGroup(ctx, mgr, "leaderworkerset.x-k8s.io", "")
 	volcanoAvailable := detectAPIGroup(ctx, mgr, "scheduling.volcano.sh", "")
+	disaggregatedSetAvailable := detectAPIGroup(ctx, mgr, "disaggregatedset.x-k8s.io", "v1")
+	// The DS pathway lists and watches the LWS children created by the DS
+	// controller. Do not register those watches when the LWS API is absent.
+	lwsOptedOut := config.Orchestrators.LWS.Enabled != nil && !*config.Orchestrators.LWS.Enabled
+	gates.DisaggregatedSet = !lwsOptedOut && lwsAvailable && disaggregatedSetAvailable
```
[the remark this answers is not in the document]
panpan0000: Fixed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. DS selection is gated by the explicit pathway gate while API capability remains available for cleanup; covered by the capability/selection tests.

panpan0000:
Addressed in 5e4cda220f5eb72329cc1f7dc06076cd5780a54c. Ported the DisaggregatedSet routing/installation guidance to the Fern LWS, install, and multinode deployment docs. Added focused coverage for API capability vs selection, service ownership handoff, status/ObservedGeneration, metadata preservation, and pinned real-CRD defaulting/convergence. `cd deploy/operator && go test ./... -count=1` and targeted `go vet` pass; Fern CLI was not available locally.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1592 @@
...
+		return nil, err
+	}
+
+	for _, componentName := range sortedSelectionComponentNames(selection) {
+		dcd := dcds[componentName]
+		if dcd == nil {
+			return nil, fmt.Errorf("generated DynamoComponentDeployment missing for selected component %q", componentName)
+		}
+		_, syncedService, err := commoncontroller.SyncResource(ctx, r, dgd, func(context.Context) (*corev1.Service, bool, error) {
+			return r.renderer.generateService(ctx, dcd)
```
yankay: Hi @panpan0000 

Could we make discovery revision-aware here? This Service is created before `dsReady` and can expose both old and new DS revisions. Matching fingerprints may mix revisions in routing, while differing fingerprints put the discovery group into `Conflict` and withdraw the WorkerSet. Could discovery use LWS's revision-scoped `*-prv` Services, or otherwise expose only one revision at a time?

deploy/operator/internal/controller/dgd_workload_program.go
```diff
@@ -78,7 +78,15 @@ func (r *DynamoGraphDeploymentReconciler) selectWorkloadProgram(
 	if r.isGrovePathway(dgd) {
 		return r.newGroveProgram()
 	}
-	return r.newComponentProgram()
+	useDisaggregatedSet, fallbackReason := shouldUseDisaggregatedSet(dgd, r.RuntimeConfig)
```
yankay: Thanks for keeping pathway dispatch thin here. Could we also keep pathway selection sticky? Today, adding a `ScalingAdapter` makes an existing opted-in DS deployment switch to `ComponentProgram` and eventually deletes the DS, implicitly creating a DS-to-DCD migration. Since the DEP excludes generic pathway migration, unsupported DS intent should probably surface as a condition rather than change providers; only an explicit opt-out should initiate a switch.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go
```diff
@@ -0,0 +1,519 @@
+//go:build !clustertest
```
yankay, line 1: Could we move the shared helpers out of this `!clustertest` file? Untagged tests still reference `newEnvtestDSHappyPathDGD` and `setupProductionWebhooks`; therefore `go test -tags=clustertest ./... -run '^$'` fails with five `undefined` errors before any Kind test can run. Moving these helpers to an untagged `_test.go` file, or tagging all consumers consistently, would keep both suites buildable.

deploy/operator/internal/controller/dgd_workload_program_test.go
```diff
@@ -135,6 +180,49 @@ func TestNewWorkloadProgramResultCopiesStatus(t *testing.T) {
 	assert.Equal(t, nvidiacomv1beta1.RollingUpdatePhaseInProgress, dgd.Status.RollingUpdate.Phase)
 }
 
+func TestComponentFallbackPreservesEligibilityConditionOnError(t *testing.T) {
+	reconcileErr := errors.New("worker hash update failed")
+	dgd := newEnvtestDSHappyPathDGD("fallback-error")
+	dgd.Namespace = "default"
```
yankay: Could this use the existing `testNamespace` constant? Operator CI currently fails `goconst` because `"default"` occurs three times while that constant already exists, blocking the pre-merge check.

deploy/operator/internal/controller/dgd_workload_program_test.go
```diff
@@ -135,6 +180,49 @@ func TestNewWorkloadProgramResultCopiesStatus(t *testing.T) {
...
+			Update: func(context.Context, client.WithWatch, client.Object, ...client.UpdateOption) error {
+				return reconcileErr
+			},
+		}).
+		Build()
+	operatorConfig := &configv1alpha1.OperatorConfiguration{}
+	operatorConfig.Namespace.Restricted = dgd.Namespace
+	reconciler := &DynamoGraphDeploymentReconciler{
+		Client:      kubeClient,
+		Recorder:    record.NewFakeRecorder(1),
```
yankay: This path writes more than one event before anything drains the recorder, so the second `FakeRecorder.writeEvent` blocks. `TestComponentFallbackPreservesEligibilityConditionOnError` consistently reaches the 15-second timeout. Could we drain the recorder or provide enough buffer capacity so the test completes deterministically?

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1592 @@
...
+	return controllerOwner != nil &&
+		controllerOwner.APIVersion == nvidiacomv1beta1.GroupVersion.String() &&
+		controllerOwner.Kind == dynamoGraphDeploymentKind &&
+		controllerOwner.Name == dgd.Name
+}
+
+func checkDisaggregatedSetReadiness(
+	ds *unstructured.Unstructured,
+	selection disaggregatedSetSelection,
+) (bool, string, map[string]nvidiacomv1beta1.ComponentReplicaStatus) {
```
yankay, line 1060: Could readiness also account for owned children whose role is absent from the current selection? This loop only visits `selection.componentToRole`, so an old LWS for a removed role can retain replicas and still be ignored. With the real DS/LWS CRDs in Kind, this returned `ready=true` while the removed-role LWS still had one replica. Readiness should remain false until every owned non-target child is scaled to zero or deleted.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in c64f3a301a. Dynamo component Services now select DS pods by DisaggregatedSet name, role, and revision. If an existing DS-scoped Service is serving an active revision during rollout, the selector is preserved until the target revision is ready; when there is no active DS revision yet, the Service selects only the target revision. Added unit and envtest coverage to prevent discovery from mixing revisions.

deploy/operator/internal/controller/dgd_workload_program.go
panpan0000: Reviewed and left unchanged in this PR. A selected ScalingAdapter makes the DS request ineligible, and the current contract intentionally falls back to the standard DCD pathway while retaining the DS until the fallback workloads are ready; this is covered by the existing envtest. Making pathway selection sticky would be a separate behavior/DEP decision.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset_envtest_test.go
panpan0000: Fixed in c64f3a301a. Added clustertest-tagged companion helpers for the envtest DGD fixture, pod template, and production webhook setup, so both normal and clustertest builds have the required symbols. `go test -tags=clustertest ./internal/controller -run ^` passes.

deploy/operator/internal/controller/dgd_workload_program_test.go
panpan0000: Fixed in c64f3a301a by using the existing `testNamespace` constant.

deploy/operator/internal/controller/dgd_workload_program_test.go
panpan0000: Fixed in c64f3a301a. Increased the fake recorder capacity so the test cannot block on its second event. I also corrected the fixture to start from the legacy worker hash, which ensures the injected Update error path is actually exercised. The targeted test now passes deterministically.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Good catch — this is a valid issue. The readiness check currently only evaluates roles in the current selection, so an old owned LWS for a removed role can be ignored while it still has replicas. I will address this by requiring every owned non-target child LWS to be scaled to zero or deleted before reporting the DS ready. I am leaving this as a follow-up change rather than mixing it into the current review-fix commit.

panpan0000: Re-evaluating this: this does not require a code change in this PR. Readiness is intentionally evaluated against the current target role selection; an LWS for a removed role is transitional cleanup owned by the DS/LWS lifecycle rather than part of the target readiness gate. I am withdrawing the follow-up commitment from my previous reply.

panpan0000:
TODO after #12942:
@julienmancuso introduces durable, immutable workload-provider selection for each DGD. 

After 12942 merged, we need to rebase this PR,  integrate DisaggregatedSet as the third provider,
(including selection, adoption, validation, and ProviderUnavailable handling). 

and remove the current annotation-driven fallback and cross-provider migration logic while preserving the DS-specific reconciliation, readiness, revision rollout, Service management, restart, and checkpoint implementation.

yankay:
Thanks, the revision-aware rollout behavior looks good to me, with no blockers. One small non-blocking tip: the current [name budget](https://github.com/ai-dynamo/dynamo/blob/5f4acf910a85a64be9ff57e0c224570ffb922134/deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go#L64) allows `<ds>-<revision>-<role>` to reach 63 characters. The LWS v0.9.0 DisaggregatedSet controller then adds the `-prv` suffix [here](https://github.com/kubernetes-sigs/lws/blob/v0.9.0/pkg/controllers/disaggregatedset/service_manager.go#L217-L219), making the worst-case Service name 67 characters, which Kubernetes rejects. Could we reserve four more characters and add coverage, perhaps in a follow-up?

panpan0000:
TODO for LWS v0.10 slicing:

- [x] upgrade the operator dependency and docs to LWS v0.10;
- [x] key child readiness by `(slice, role, revision)`, aggregate replica status across slices, and keep label-less legacy children as slice 0;
- [x] keep the stable Dynamo component Service across all slices, while LWS-owned `*-prv` Services stay slice-local;
- [x] reserve the slice segment in the name budget and cover two slices, missing/duplicate slice children, legacy slice 0, and slice watch mapping;
- [ ] after the grouping/cardinality API lands, map Dynamo-owned group cardinality to `DisaggregatedSet.spec.slices`.

This PR still emits no slice cardinality and relies on the LWS default `slices=1`. It does not add a transitional annotation or expose `slices` through `providerOverride`.

@julienmancuso @yankay

panpan0000:
@yankay good catch — fixed in e7a1f0cefe. I reserved the four characters used by the LWS v0.9 `-prv` suffix and extended the boundary test to call LWS's `GenerateServiceName`, so it now checks the actual generated Service name rather than only the child LWS name.

`go test ./internal/controller -count=1` and `go vet ./internal/controller` pass. The additional slice segment introduced in LWS v0.10 remains part of the slicing follow-up above.

panpan0000:
@julienmancuso @yankay I updated the LWS dependency to v0.10 in `905e9eac4e`. The DS readiness and child bookkeeping now understand multiple slices and the legacy slice-0 shape. The stable Dynamo Service still covers all slices, and LWS keeps its `*-prv` Services slice-local.

I kept the API boundary unchanged: this pathway still defaults to one slice. The future grouping/cardinality API can own `spec.slices` without another temporary annotation or `providerOverride` field.

yankay:
> @julienmancuso @yankay I updated the LWS dependency to v0.10 in `905e9eac4e`. The DS readiness and child bookkeeping now understand multiple slices and the legacy slice-0 shape. The stable Dynamo Service still covers all slices, and LWS keeps its `*-prv` Services slice-local.
> 
> I kept the API boundary unchanged: this pathway still defaults to one slice. The future grouping/cardinality API can own `spec.slices` without another temporary annotation or `providerOverride` field.

Thanks, the latest LWS v0.10 and name-budget changes look good. I do not see blockers in the DS-specific logic.

panpan0000:
@julienmancuso @sschimanski

I rebased this PR onto `main` after #12942 and updated it to follow the durable workload-provider model. The latest head is `4538221b52`.

The implementation now supports `disaggregatedset` as the third durable provider, adopts legacy DS-owned DGDs, treats provider selection as immutable, reports `selected_workload_provider_unavailable` without fallback, and removes the old cross-provider cleanup/migration logic. I also updated the provider/admission/envtest coverage and multinode documentation.

`go test ./internal/controller -count=1` and all non-e2e operator packages pass. The e2e Docker build remains environment-blocked by HTTP 403 responses from the configured image registries.

Please review the latest head, especially the provider selection/adoption and DS unavailable-provider behavior.

panpan0000:
doc CI waits for https://github.com/ai-dynamo/dynamo/pull/13333 fix

deploy/operator/internal/controller/dgd_workload_provider.go
```diff
@@ -96,12 +98,16 @@ func providerFromOwnedWorkloads(
...
+	hasDisaggregatedSet, err := hasOwnedDisaggregatedSet(ctx, reader, dgd)
+	if err != nil {
+		return "", false, err
+	}
 
-	// Adopt one unambiguous family and fail closed when both families exist.
+	// Adopt one unambiguous family and fail closed when multiple families exist.
 	switch {
-	case hasComponents && hasGrove:
+	case hasComponents && hasGrove || hasComponents && hasDisaggregatedSet || hasGrove && hasDisaggregatedSet:
```
julienmancuso: [P1] Adopt DisaggregatedSet when auxiliary DCDs exist

A DisaggregatedSet provider legitimately owns the DS plus DCDs for non-selected components such as the global frontend. This condition classifies that normal composite shape as a provider conflict, so a legacy unannotated DGD can remain stuck before its provider is persisted. DCDs may likewise coexist with Grove; the genuinely ambiguous case is DS plus PodCliqueSet.

Please use the provider-owned resources as precedence: DS + PCS fails closed; otherwise DS selects DisaggregatedSet, PCS selects Grove, and DCD-only selects Component. Add adoption cases for DS + auxiliary DCDs, PCS + optional DCDs, DS + PCS, and DCD-only.

deploy/operator/internal/controller/dgd_grove_stable_resources_reconciler.go
```diff
@@ -132,6 +136,14 @@ func (r *groveStableResourcesReconciler) reconcileComponentService(
 		logger.Error(err, "failed to generate the main component service")
 		return nil, fmt.Errorf("failed to generate the main component service: %w", err)
 	}
+	existingService := &corev1.Service{}
+	existingServiceErr := r.Get(ctx, client.ObjectKeyFromObject(service), existingService)
+	if client.IgnoreNotFound(existingServiceErr) != nil {
+		return nil, fmt.Errorf("failed to get existing main component service: %w", existingServiceErr)
+	}
+	if existingServiceErr == nil && !targetReady && isDisaggregatedSetServiceSelector(existingService) {
```
julienmancuso: [P2] Remove cross-provider selector handoff

This makes the Grove pathway understand DisaggregatedSet-native selectors and keep DS serving until Grove is ready, which is partial DS-to-Grove migration logic. The DCD controller contains the equivalent DS-to-Component handoff. The durable workload provider is immutable and this PR states that provider switching and cross-provider cleanup are unsupported, so Grove and DCD should not depend on DS labels or transition semantics.

Please remove the cross-provider selector preservation from Grove and DCD. Keep only DS-internal revision handoff inside the DS program: retaining an old DS revision until a new DS revision is ready is rolling update behavior, not provider switching.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type disaggregatedSetWorkloadsReconciler struct {
+	client.Client
+	// reader defaults to the cached client. The graph reconciler replaces it
+	// with the uncached API reader because compatibility cleanup must act on
+	// the authoritative DisaggregatedSet at a pathway switch even when the
+	// informer has not yet observed a write from a previous reconcile.
+	reader                   client.Reader
```
julienmancuso: [P2] Remove the dead uncached-reader dependency

This reader is replaced with the outer APIReader but is never consumed by the workloads reconciler; the LWS watch mapper separately receives mgr.GetAPIReader. Besides being dead state, the uncached mapping read conflicts with the controller rules: dependency watches and cached, level-driven convergence should handle informer lag unless a concrete correctness requirement proves otherwise.

Please remove this field and the new DGD APIReader plumbing, and use the cached manager client for the watch mapper. A child event that arrives before its parent is cache-visible should remain pending and converge through subsequent watched updates.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1596 @@
...
+
+		if !dsReady {
+			logger.V(1).Info("DisaggregatedSet component not fully updated", "componentName", componentName, "reason", dsReason)
+			updatedInProgress = append(updatedInProgress, componentName)
+		}
+	}
+	return updatedInProgress
+}
+
+func (r *disaggregatedSetWorkloadsReconciler) buildRollingUpdateContext(
```
julienmancuso: [P2] Decompose the DS implementation and remove forwarding helpers

These final methods merely rename one collaborator call, which AGENTS.md explicitly prohibits. More broadly, this 1,596-line file mixes rendering, synchronization, Service ownership, readiness, watches, restart progress, and manual unstructured parsing even though the PR pins and imports the typed DisaggregatedSet API. ResolveRestart also receives a restart reconciler and discards it.

Please register and use the typed DS API, remove artificial uniformity and unused dependencies, and extract only cohesive collaborators such as DS rendering/synchronization, stable Services, readiness, and watch mapping. Keep the program control flow explicit rather than hiding it behind forwarding methods.

deploy/operator/internal/controller/dgd_workload_provider.go
panpan0000: Fixed in 62867368e4. DS now wins when it exists with helper DCDs. Only DS + Grove is ambiguous. I added the four adoption cases.

deploy/operator/internal/controller/dgd_grove_stable_resources_reconciler.go
panpan0000: Fixed in 62867368e4. Grove and Component no longer keep DS selectors. DS still keeps the old selector only during its own rollout.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
panpan0000: Fixed in 62867368e4. I removed the unused reader and APIReader plumbing. I kept the APIReader for the watch mapper because cache timing still needs it.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed the concrete cleanup in 62867368e4: the unused restart dependency and forwarding helpers are gone. I left the larger typed-API/file split for a separate follow-up.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1547 @@
...
+	return nil
+}
+
+func (r *disaggregatedSetWorkloadsReconciler) listOwnedSelectedDCDs(
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+	selection disaggregatedSetSelection,
+) ([]nvidiacomv1beta1.DynamoComponentDeployment, error) {
+	dcdList := &nvidiacomv1beta1.DynamoComponentDeploymentList{}
+	if err := r.List(ctx, dcdList, client.InNamespace(dgd.Namespace), client.MatchingLabels{
```
julienmancuso: [P1] Discover selected DCD cleanup candidates by ownership

This cleanup applies the mutable DGD-name label selector before checking the controller reference. That recreates the ownership bug fixed in provider adoption: a selected DCD owned by this DGD but carrying a missing or stale label is invisible here, survives after the DisaggregatedSet becomes ready, and can continue reconciling a second workload family.

Please discover candidates from the namespace cache and treat the controller owner reference as authoritative, or use an owner-UID cache index if profiling justifies it. Add a regression proving that an owned selected DCD without the graph label is deleted while foreign labeled and unlabeled DCDs are ignored.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
julienmancuso: Thanks for removing the unused workloads-reader field, but this thread is not fully addressed. The watch setup still passes `mgr.GetAPIReader()` to `newDisaggregatedSetWatchMapper`, so the child-to-parent lookup remains an uncached API-server read.

The cache-timing concern is exactly what the controller rules ask level-driven reconciliation and dependency watches to handle: if the child event arrives before the parent DS is cache-visible, return no request and let the subsequent watched DS/LWS update drive convergence. Please pass the cached manager client instead and remove the DGD `APIReader` plumbing introduced solely for this mapper, unless there is a concrete correctness invariant that cannot be represented with cached state and watches.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 090d03491c. Cleanup now lists all DCDs in the namespace and uses the DGD controller owner reference instead of the mutable graph label. Added regression coverage for owned unlabeled/stale-label DCDs and foreign labeled/unlabeled DCDs.

deploy/operator/internal/controller/dgd_disaggregatedset_workloads_reconciler.go
panpan0000: Fixed in 9c4101f7e0. The DS child watch mapper now uses the cached manager client (`mgr.GetClient()`), and the unused DGD APIReader field/setup plumbing is removed. If the parent DS is not in cache yet, the mapper returns no request and later watched events drive reconciliation.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Thanks. The concrete cleanup is fixed: the unused restart dependency and forwarding helpers are gone in 090d03491c. The remaining typed-API registration and full file split are a separate refactor; keeping that larger change out avoids changing reconciliation behavior in this provider rollout. I am treating this as follow-up work, not a blocker for this PR.

julienmancuso (approved):

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1545 @@
...
+		return ReconcileResult{}, fmt.Errorf("failed to get existing restart annotations: %w", err)
+	}
+	existingDSRestartAnnotations, err := r.getExistingRestartAnnotationsDisaggregatedSet(ctx, dgd, selection)
+	if err != nil {
+		logger.Error(err, "failed to get existing DisaggregatedSet restart annotations")
+		return ReconcileResult{}, fmt.Errorf("failed to get existing DisaggregatedSet restart annotations: %w", err)
+	}
+	maps.Copy(existingRestartAnnotations, existingDSRestartAnnotations)
+
+	dcds, err := dynamo.GenerateDynamoComponentsDeployments(
```
sttts: This rendering shape feels backwards: the DS path first expands the DGD into DCD objects, then uses the selected DCDs as an intermediate representation and takes them apart again to render DisaggregatedSet roles.

Grove has the same graph-level rendering problem, but does not go through DCDs. Its provider renderer works directly from the shared component spec and lowers components into Cliques/PCSGs while reusing the lower-level pod rendering.

Could DS follow the same shape and introduce the common rendering boundary explicitly?

```text
DGD component
    -> shared component renderer
        -> DCD/LWS lowering
        -> Grove Clique/PCSG lowering
        -> DisaggregatedSet Role lowering
```

This would keep provider reconciliation, rollout, readiness, and persistence separate while sharing the rendering that is actually common. DS would then not need to construct `componentWorkloadsReconciler` as a helper container or generate selected DCDs only to extract their PodTemplates again.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1545 @@
...
+		return ReconcileResult{}, err
+	}
+	resources = append(resources, syncedDSResource)
+
+	nonSelectedResources, err := r.reconcileDisaggregatedSetNonSelectedDCDs(ctx, dgd, dcds, selection)
+	if err != nil {
+		return ReconcileResult{}, err
+	}
+	resources = append(resources, nonSelectedResources...)
+	desiredServiceNames, err := r.reconcileDisaggregatedSetServiceOwnership(ctx, dgd, dcds, selection, selectedServiceNames)
```
sttts: I think we should enforce a stronger controller invariant here: unless reconciliation is deliberately implementing a persisted consistency protocol across requests — for example, prepare A, write B, then mark A done — one object should have one complete desired state, one persistence operation per reconciliation pass, and one managed-fields owner / field manager.

The same Service currently goes through several persistence paths:

```text
SyncResource(Service)
    -> ensureControlledByDGD(Service)       // separate patch
    -> reconcileDisaggregatedSetServiceOwnership()
        -> Get(Service)
        -> ensureControlledByDGD/DCD()      // potentially another patch
```

Model Services can additionally go through the post-`SyncResource` annotation `Update` in `ReconcileModelServicesForComponents`. This means spec, annotations, and ownership are not composed into one authoritative desired Service before writing it, and we may immediately reread stale cache state after our own write.

Could we have one Service reconciler render the complete desired object — spec, labels, annotations, and controller reference — and persist it once through one field manager? A multi-step persistence protocol should only exist where the intermediate state is intentional, durable, and required for cross-request consistency.

sttts:
I think this exposes a more fundamental problem with the shape of this change.

For every managed object, we should have:

1. one rendering pipeline producing its complete desired state,
2. one reconciler responsible for persisting that state,
3. one stable controller owner for the lifetime of the object, and
4. one managed-fields owner / field manager.

Changing a Service controller reference from DCD to DGD, and potentially back again, violates that model. The owner reference should express stable lifecycle ownership, not which workload provider currently supplies the Service backend. For graph-level Services, the DGD should likely remain the owner, while the DCD, Grove, or DisaggregatedSet rendering determines what backs the Service.

The intended flow should look roughly like this:

    DGD + provider state
            |
            v
      Service renderer
            |
            v
    complete desired Service
            |
            v
      Service reconciler
       (one persistence path,
        one field manager,
        stable ownership)

The size and scope of the implementation are another warning sign. The DisaggregatedSet code currently combines rendering/lowering, side-resource reconciliation, ownership migration, readiness evaluation, rollout coordination, and deletion of transitional resources in one roughly 1,500-line file. These are separate responsibilities, and putting them together makes the lifecycle and ownership invariants difficult to see and enforce.

I think we should first define the stable ownership and reconciliation model for every generated object, then split the implementation along those boundaries. Provider support should add a provider-specific rendering/lowering path, not introduce ownership hand-offs or additional persistence protocols for existing objects.

panpan0000:
wait @julienmancuso and @sttts 's response in slack first.

panpan0000:
reworked based on @sttts's feedback.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1439 @@
...
+	}
+	modelNamesSorted := make([]string, 0, len(modelNames))
+	for modelName := range modelNames {
+		modelNamesSorted = append(modelNamesSorted, modelName)
+	}
+	sort.Strings(modelNamesSorted)
+	for _, modelName := range modelNamesSorted {
+		annotations := maps.Clone(dgd.Spec.Annotations)
+		service := dynamo.GenerateModelServiceForModel(dgd.Namespace, modelName, annotations)
+		if err := r.syncDGDStableService(ctx, dgd, service); err != nil {
```
sttts, line 734: **[P1] Do not give a namespace-shared model Service a DGD controller owner**

`GenerateModelServiceForModel` derives the Service name and selector only from the model name. Two DGDs in the same namespace that reference the same model therefore render the same Service. The first reconciliation makes its DGD the controller owner; the second then fails the owner check below, and deleting the first DGD can garbage-collect an endpoint that the second DGD still needs.

Cleanup has the same identity mismatch: the generated model Service does not carry the graph-name label used by `deleteStaleDisaggregatedSetServices`; the cleanup test adds that label manually.

Please align ownership with the namespace-shared identity, for example through shared lifecycle management, or make the Service identity graph-private. This needs coverage for two DGDs sharing a model, deleting either DGD, and removing the final model reference.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1439 @@
...
+
+	// These fields are allocated by the apiserver and must survive a desired
+	// Service render.
+	updated.Spec.ClusterIP = existing.Spec.ClusterIP
+	updated.Spec.ClusterIPs = existing.Spec.ClusterIPs
+	updated.Spec.IPFamilies = existing.Spec.IPFamilies
+	updated.Spec.IPFamilyPolicy = existing.Spec.IPFamilyPolicy
+	updated.Spec.HealthCheckNodePort = existing.Spec.HealthCheckNodePort
+
+	if equality.Semantic.DeepEqual(existing.Spec, updated.Spec) &&
```
sttts, line 799: **[P1] Normalize API defaults before deciding that this Service is current**

`CopySpec` replaces the typed Service spec through unstructured conversion, so omitted fields become Go zero values. The live Service, however, contains API-server defaults such as `type: ClusterIP`, `sessionAffinity: None`, and `internalTrafficPolicy: Cluster`. This code restores allocated fields, but not those defaults.

Consequently this comparison remains different, the controller updates the Service, the API server restores the defaults, and the Service watch requeues the DGD into the same update loop. Fake-client tests do not apply these defaults.

Please normalize the complete desired Service before comparison, while preserving server-allocated fields, and add an envtest proving that a second reconciliation performs no Update.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -266,6 +294,7 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			UpdateFunc:  func(de event.UpdateEvent) bool { return true },
 			GenericFunc: func(ge event.GenericEvent) bool { return true },
 		})).
+		Owns(&corev1.Service{}, builder.WithPredicates(dgdOwnedServiceEventPredicate())).
```
sttts: **[P3] Consolidate the duplicate Service owner watch**

The same controller builder already registers `Owns(&corev1.Service{})` at line 315 with the same effective create/update/delete behavior. Registering two sources for the same owned Services can enqueue the same DGD twice for one event. Please consolidate the predicates into the existing Service watch and remove the duplicate registration/helper.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 9d97a67dcb. The DisaggregatedSet path now gives each graph a graph-scoped model Service name (model hash plus graph hash), includes the graph identity in its label and selector, and keeps the DGD as the controller owner. Stale model-Service cleanup lists namespace Services and treats the DGD owner reference as authoritative. The envtest covers two DGDs sharing a model and removing one graph without affecting the other, while the legacy model-only helper remains unchanged for DCD/Grove.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed in 9d97a67dcb. Stable Service reconciliation now copies the typed desired spec, applies the Kubernetes Service defaults before comparing, preserves server-allocated fields, and performs no write when the result is unchanged. The envtest reconciles the same graph twice against a real API server and asserts that the Service resourceVersion does not change on the second pass.

deploy/operator/internal/controller/dynamographdeployment_controller.go
panpan0000: Fixed in 9d97a67dcb: removed the duplicate Owns(&corev1.Service{}) registration and its redundant predicate, while keeping the existing Service watch for component and elastic-EP leader Service updates/deletes.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: The current head now renders DisaggregatedSet roles directly from normalized shared component specs; selected components are no longer materialized as DCDs only to extract PodTemplates, while non-selected components still lower to real DCDs. The DS rendering tests cover this split. The original line is outdated, so I am resolving this conversation.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: The current DS path composes each stable Service spec, labels, annotations, and controller owner before a single create/update in syncDGDStableService; it no longer uses the old multi-step Service persistence sequence. The API-defaulting envtest also checks no-op idempotence on a second reconcile. This original line is outdated, so I am resolving the conversation.

panpan0000:
@sttts , Dr.Stefan, thank you a lot , comments fixed again :-)

deploy/operator/internal/controller/dynamocomponentdeployment_controller.go
```diff
@@ -910,7 +935,6 @@ func (r *DynamoComponentDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
 			UpdateFunc:  func(de event.UpdateEvent) bool { return true },
 			GenericFunc: func(ge event.GenericEvent) bool { return true },
 		})).
-		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
sttts, line 913: Removing this watch regresses Service self-healing for the component-provider and standalone-DCD paths. Those Services are still rendered and owned by the DCD when `isDisaggregatedSetBacked` is false, so deleting one no longer enqueues the DCD. The new DGD Service watch only maps Services directly controlled by a DGD and therefore cannot replace this watch.

Please keep both watches; the stable controller owner will route each Service to the correct reconciler. A focused test that deletes a DCD-owned component or model Service and verifies that it is recreated would cover this.

panpan0000: Thank you , and just fixed:  I restored the DCD Service watch and added an envtest that deletes a DCD-owned Service and verifies it is recreated.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
```diff
@@ -108,6 +108,131 @@ func (r *dcdWorkloadRenderer) renderMultinodePodTemplateSpecs(
...
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+	component *nvidiacomv1beta1.DynamoComponentDeploymentSharedSpec,
+	componentName string,
+	workloadName string,
+	dynamoNamespace string,
+	backendFramework dynamo.BackendFramework,
+	checkpointInfo *checkpoint.CheckpointInfo,
+) (*corev1.PodTemplateSpec, *corev1.PodTemplateSpec, error) {
+	podLabels := dynamo.GetDGDComponentResourceLabels(dgd, componentName, component)
+	podAnnotations := dynamo.GetDGDComponentResourceAnnotations(dgd, componentName, component)
```
julienmancuso: [P1] Apply the KV-transfer topology annotations on the direct DisaggregatedSet path. This renderer starts from GetDGDComponentResourceAnnotations but never calls applyKvTransferPolicyTopologyAnnotations, while the DCD and Grove paths do. With spec.experimental.kvTransferPolicy.labelKey, the generated pods therefore lack the topology-source annotation: the topology-label controller has nothing to project, and DYN_TOPOLOGY_ENABLED startup can time out on an empty Downward API file. Please route both adapters through the shared component renderer, or at minimum apply the same topology step here, and add a regression covering labelKey.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
```diff
@@ -0,0 +1,1585 @@
...
+		}
+		if !isControlledByBetaDGD(currentOwner, dgd) {
+			return fmt.Errorf("Service %s/%s is controlled by unrelated DynamoComponentDeployment %s", existing.Namespace, existing.Name, currentOwner.Name)
+		}
+	}
+
+	updated := existing.DeepCopy()
+	updated.Spec = *desired.Spec.DeepCopy()
+	normalizeDGDStableServiceSpec(&updated.Spec)
+	updated.Labels = maps.Clone(existing.Labels)
```
julienmancuso: [P2] Reconcile controller-managed metadata authoritatively. Starting from the existing labels and annotations and overlaying desired values means any propagated DGD/component/model/discovery key removed from the desired Service remains forever. Please define managed-key ownership, or use an SSA field manager, so formerly managed keys are pruned while genuinely external metadata is preserved. Add regressions for both managed-key removal and external-key preservation.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
```diff
@@ -108,6 +108,131 @@ func (r *dcdWorkloadRenderer) renderMultinodePodTemplateSpecs(
 	return leaderPodTemplateSpec, workerPodTemplateSpec, nil
 }
 
+// renderMultinodePodTemplateSpecsForDGDComponent renders a selected DS role
+// directly from the normalized DGD component instead of materializing a DCD.
+func (r *dcdWorkloadRenderer) renderMultinodePodTemplateSpecsForDGDComponent(
```
julienmancuso, line 113: [P2] Please establish one shared component-rendering boundary rather than maintaining a second DCD-like pipeline here. This path duplicates label/annotation assembly, GPU resolution, component typing, checkpoint metadata, and role rendering; the missing KV-topology behavior is already one concrete divergence. DCD and DisaggregatedSet should derive provider-specific inputs and then call the same renderer, with parity tests for topology, checkpoints, discovery metadata, labels/annotations, and worker hashes. Prefer explicit domain inputs over a generic request or state bag.

julienmancuso:
Review summary for e05df2226f72317ca7ac7e1d48a8e13d75674ce6:

The latest DCD-owned Service watch restoration looks correct. It restores self-healing for standalone/component DCDs and does not introduce a second Service writer for DS-backed helper DCDs, because their DCD reconciler skips Service reconciliation once the graph provider is frozen to DisaggregatedSet.

The remaining non-line-local composition-first concern is that dynamographdeployment_disaggregatedset.go is now 1,585 lines and still combines rollout projection, DS rendering and persistence, stable Services, readiness, cleanup, and watch mapping. The new ten-argument worker-hash helper moves a block without creating a cohesive reconciler boundary and makes the sequence harder to follow. Please split this by concrete responsibility/resource family while keeping the selected DS program as the sole owner of its ordering and status result. This should complement the inline shared-renderer finding, not introduce a generic pipeline or state bag.

Validation at this exact head: required linux/arm64 linter passed, required linux/arm64 tester passed, and git diff --check passed.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
panpan0000: Fixed.

deploy/operator/internal/controller/dynamographdeployment_disaggregatedset.go
panpan0000: Fixed.

deploy/operator/internal/controller/dynamocomponentdeployment_renderer.go
panpan0000: Fixed.

panpan0000:
Thank you for your continuous patient help !! @julienmancuso @sttts 


Code updated

image


```
DGD Controller
├── Watch Setup
└── Selected DS Program
    ├── rollout / shared resources / restart
    ├── DS Workloads Reconciler
    │   ├── renderer → DS resource reconciler
    │   ├── readiness resolver
    │   ├── stable resources reconciler
    │   ├── auxiliary DCD reconciler
    │   └── worker-hash commit / cleanup
    ├── scaling adapters
    └── return final status
         ↓
DGD Controller persists status/events
```

PR #12627, opened 2026-08-04
feat(vllm): sample variable-length prefill batches in the self-benchmark

YijiaZhao:
## Overview:

The self-benchmark only ever measures batches whose requests are all the same
length, so it cannot see that two batches with the same token totals and
different spreads do not cost the same. Attention work is quadratic in the
new-token count, and with a sparse indexer the cost also turns on which
requests sit above `topk`. On GLM-5-NVFP4 (B200, TP8) the gap reaches 17% of
step time.

This adds variable-length prefill sampling behind a flag, off by default.

## Details:

**Manifest schema v3.** Prefill points may carry explicit per-request
`[new_tokens, kv_read_tokens]` rows, validated to conserve both totals exactly
— the label is only the cost of the spread if the totals hold. A shape
parameter cannot express these rows: they are solved from an inequality on
every request, and rounding one into a fraction moves a row across `topk` and
silently changes which kernel the batch measures.

**`--benchmark-collect-imbalanced`**, off by default.

- With an explicit manifest, its imbalanced points are measured; without the
  flag they are skipped and logged.
- With no manifest, the generated sweep is expanded: each coordinate
  additionally gets the spreads planned around it, the resulting manifest is
  written next to `--benchmark-output-path`, and that file is what runs.
  Keeping the rows only in memory would leave the run unreproducible — the
  constraints they were solved from do not survive in the results file.

Two properties worth naming: the uniform batch is emitted for every
coordinate, so turning the flag on never loses one the off run would have had;
and coordinates around which no spread can be built are reported rather than
dropped, since a silently missing cell reads as a complete run.

**Metrics.** `ScheduledRequestMetrics` gains the second moment of per-request
KV reads and the batch's order statistics — moments alone cannot tell a batch
straddling `topk` from one entirely above or below it.

**Verified.** 34 unit tests. On GPU: the flag parses, `topk` is probed from the
model config, and the constructed manifest reaches disk (10 sweep coordinates →
26 forward passes). End-to-end collection was not exercised — the image's vLLM
returns two values from `find_longest_cache_hit` where main unpacks three,
which predates this change.

`benchmark_points.py` — the schema and its validators; the totals-conservation
check is the one that matters.

Then `benchmark_imbalance.py`, specifically `build_manifest` and `plan_cell`:
how the batches are built so both totals stay exact and each lands in the
regime it claims.

`instrumented_scheduler.py` — `_bench_build_imbalanced_grid` (plan → write →
read back through the normal loader) and `_bench_point_is_imbalanced`, which
tests whether the rows actually differ rather than whether rows are present;
the uniform reference batch carries rows too.

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12627" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added support for imbalanced prefill benchmarking with configurable repeats, top-k selection, and manifest output.
  - Added partitioned and explicit per-request benchmark shapes, including schema version 3 support.
  - Added automatic planning and validation for uniform, pure, and mixed workload distributions.
  - Expanded benchmark metrics with KV-token variance, minimum KV length, and maximum prompt length.
- **Compatibility**
  - Improved support across scheduler versions with and without prefill throttling.
- **Tests**
  - Added coverage for workload planning, manifest validation, token conservation, and benchmark shape handling.


components/src/dynamo/vllm/benchmark_imbalance.py
```diff
@@ -0,0 +1,609 @@
...
+    for batch_size, total_new, total_kv in cells:
+        s_bar, p_bar = total_new // batch_size, total_kv // batch_size
+        if s_bar < 1:
+            continue
+        # The sweep's own coordinate goes in first and unconditionally. It is
+        # emitted even when no spread can be built around it, because this
+        # manifest REPLACES the generated grid: a cell dropped here is a
+        # coordinate the switch-off run would have measured and this one
+        # silently would not.
+        emit(uniform_rows(batch_size, s_bar, p_bar))
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874.

`reference_rows` replaces the floor-divided `uniform_rows` call in `build_manifest`. It spreads the remainder one unit per row, in whole blocks on the KV axis, so both totals hold exactly and every row's prefix is a whole block; where the totals divide it is still the equal batch. The coordinate is therefore the one the sweep chose, not one up to `b - 1` tokens below it.

A cell whose KV total is not a whole number of blocks cannot be expressed by any split into whole blocks, so it is appended to `notes` and skipped rather than emitted for the loader to reject.

The unit mismatch you flagged at the end is real and fixed separately: `build_manifest` is now called with `max(1, self._bench_hash_block_size)` — the unit `_bench_prefill_kv_read_lengths` validates against — instead of `cache_config.block_size`.

Test: `test_reference_batch_conserves_both_totals_in_whole_blocks` asserts conservation, block alignment, and that the batch stays within one token / one block of equal, over cells that do not divide (including your `(3, 3072, 256)` example). `test_reference_batch_reports_a_cell_it_cannot_express` covers the skip path.

components/src/dynamo/vllm/args.py
```diff
@@ -364,6 +364,12 @@ def update_engine_config_with_dynamo(
             "warmup_iterations": dynamo_config.benchmark_warmup_iterations,
             "output_path": dynamo_config.benchmark_output_path,
             "timeout": dynamo_config.benchmark_timeout,
+            "collect_imbalanced": dynamo_config.benchmark_collect_imbalanced,
+            "imbalance_repeats": dynamo_config.benchmark_imbalance_repeats,
+            "imbalance_topk": dynamo_config.benchmark_imbalance_topk,
+            "imbalance_manifest_path": (
+                dynamo_config.benchmark_imbalance_manifest_path
+            ),
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874. `imbalance_repeats` joined `_INT_FIELDS`. `imbalance_topk` could not: `None` is a value there, not a missing key — it selects the model-config probe — so it is coerced separately, only when it is not `None`.

`collect_imbalanced` was the same class of latent bug with a worse failure mode, since `"false"` is a non-empty string and would have turned collection on. It is now parsed against a truthy set rather than by truthiness.

components/src/dynamo/vllm/benchmark_imbalance.py
```diff
@@ -0,0 +1,609 @@
...
+    b = len(rows)
+    x_idx = sum(idx_work(s, p, topk) for s, p in rows if runs_sparse(s, p, topk))
+    x_sp = sum(mla_work(s, p, topk) for s, p in rows if runs_sparse(s, p, topk))
+    x_dn = sum(mla_work(s, p, topk) for s, p in rows if not runs_sparse(s, p, topk))
+    if runs_sparse(s_bar, p_bar, topk):
+        x_idx -= b * idx_work(s_bar, p_bar, topk)
+        x_sp -= b * mla_work(s_bar, p_bar, topk)
+    else:
+        x_dn -= b * mla_work(s_bar, p_bar, topk)
+    return x_idx, x_sp, x_dn
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874 — the docstring reflected the intended model and the column did not.

`work_columns` now sums `idx_work` over every row and subtracts `b * idx_work(s_bar, p_bar, topk)` unconditionally. Gating it reintroduced exactly the error the `idx_work` docstring records, and did it where it hurts most: a mixed batch is mostly short rows.

Worth naming for anyone reading the results: this changes fitted coefficients, so data collected before this commit is not comparable with data collected after it.

Test: `test_indexer_column_carries_rows_below_the_budget` — a batch entirely below the budget, where the gated column returned exactly zero.

components/src/dynamo/vllm/backend_args.py
```diff
@@ -406,6 +406,58 @@ def add_arguments(self, parser) -> None:
...
+            env_var="DYN_BENCHMARK_IMBALANCE_TOPK",
+            default=None,
+            arg_type=int,
+            help=(
+                "Index budget above which a request takes the sparse-attention "
+                "path, which is what decides the regime each constructed batch "
+                "measures. Default probes the model config; a model with no "
+                "such bound needs nothing here."
+            ),
+        )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874, on both layers as suggested.

`_validate_benchmark_sampling` now requires `benchmark_imbalance_repeats >= 1` and `benchmark_imbalance_topk >= 1` when set, so the operator gets a startup error.

`_bench_init` enforces the same bounds after `BenchmarkConfig` is built, since the env and `additional_config` paths never reach the CLI validator. `imbalance_repeats` was added to `_INT_FIELDS`; `imbalance_topk` is coerced separately because `None` is a value there rather than a missing key. `collect_imbalanced` is now parsed as a boolean rather than by truthiness, where `"false"` read as true.

components/src/dynamo/vllm/benchmark_imbalance.py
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874 — took the second option, keeping the totals exact.

`reference_rows` spreads the remainder one unit per row rather than truncating it, in whole blocks on the KV axis so each row is also block-aligned. Both totals are conserved exactly, so the manifest remains a superset of the sweep it replaces and the docstring claim holds. Where the totals divide, the batch is unchanged.

components/src/dynamo/vllm/benchmark_points.py
```diff
@@ -20,16 +20,93 @@ class _PointCandidate(BaseModel):
...
+                )
+        if self.partition is not None:
+            if self.batch_size < 2:
+                raise ValueError("partition requires batch_size >= 2")
+            if self.partition.high_count >= self.batch_size:
+                raise ValueError("partition.high_count must be less than batch_size")
+            if self.partition.axis in ("kv", "both") and self.total_kv_read_tokens == 0:
+                raise ValueError(
+                    'partition axis "kv" requires total_kv_read_tokens > 0'
+                )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874 — the message now interpolates `self.partition.axis`, so an `axis="both"` manifest reports `"both"`.

components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -2010,6 +2180,94 @@ def _bench_raise_explicit_infeasible(
...
+            return
+
+        topk = self._bench_imbalance_topk()
+        manifest, notes = benchmark_imbalance.build_manifest(
+            cells,
+            topk,
+            repeats=self._bench_config.imbalance_repeats,
+            max_model_len=self.vllm_config.model_config.max_model_len,
+            kv_block=self.cache_config.block_size,
+        )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874 — `build_manifest` is now called with `max(1, self._bench_hash_block_size)`.

The planner has to work in the unit its own reader validates against: a row it emits is read back through `_bench_prefill_kv_read_lengths`, which rejects KV lengths that are not whole hash blocks. Planning in `cache_config.block_size` let it build rows that loader refuses.

components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -2010,6 +2180,94 @@ def _bench_raise_explicit_infeasible(
...
+        logger.info(
+            "benchmark imbalance: %d sweep coordinates -> %d forward passes "
+            "(topk=%s, %d repeats), manifest written to %s",
+            len(cells),
+            len(manifest["prefill"]),
+            topk,
+            self._bench_config.imbalance_repeats,
+            path,
+        )
+        self._bench_build_explicit_grid(load_benchmark_points_file(path))
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 9248874, along the lines you suggested.

`_bench_build_explicit_grid` takes a `generated` flag, threaded into both materializers. A planned point is tagged `"imbalance"` instead of `"explicit"` via `_bench_origin_reason`, and an infeasible one returns `None` and is skipped with a logged count rather than raising through `_bench_raise_explicit_infeasible`. Because the tag is what `_bench_skip_point` keys on, a run-time failure on a planned point now skips as the ordinary generated grid does.

`"explicit"` stays reserved for what it means downstream: the operator asked for this exact point, so an infeasible one is an error. A point this code planned itself carries no such request and should not borrow that promise.

components/src/dynamo/vllm/benchmark_imbalance.py
```diff
@@ -0,0 +1,659 @@
...
+    def emit(rows) -> None:
+        key = tuple(map(tuple, rows))
+        if key in seen:
+            return
+        seen.add(key)
+        point = _point(rows)
+        prefill.extend(dict(point) for _ in range(repeats))
+
+    for batch_size, total_new, total_kv in cells:
+        s_bar, p_bar = total_new // batch_size, total_kv // batch_size
```
tedzhouhk: [P1] Keep planned shapes on the reference coordinate

`reference_rows` preserves the original `total_new` / `total_kv`, but `plan_cell` receives floored `s_bar` / `p_bar`, and every planned shape conserves `b * s_bar` / `b * p_bar`. For `(b=4, total_new=4097, total_kv=0)`, the reference totals 4097 while every imbalance shape totals 4096, so `T_shape - T_reference` includes the cost of a token-count change and is not a pure spread label. These non-divisible points occur naturally at the deliberately sampled `C+1` CUDA-graph boundaries. Please either plan against the exact totals/reference rows or avoid emitting imbalance shapes for non-divisible cells.

components/src/dynamo/vllm/instrumented_scheduler.py
```diff
@@ -1507,23 +1629,36 @@ def _extract_scheduled(self, output: SchedulerOutput) -> ScheduledRequestMetrics
...
 
         return ScheduledRequestMetrics(
             num_prefill_requests=num_prefill,
             sum_prefill_tokens=sum_prefill_tokens,
             var_prefill_length=prefill_lengths.variance(),
-            sum_prefill_kv_tokens=sum_prefill_kv_tokens,
+            sum_prefill_kv_tokens=prefill_kv.s,
+            var_prefill_kv_tokens=prefill_kv.variance(),
+            min_prefill_kv_tokens=min_prefill_kv or 0,
+            max_prefill_length=max_prefill_len,
```
tedzhouhk: [P1] Use the current prefill end length for the work-delta metrics

The identity documented for this feature requires `l_i = s_i + p_i` for the current forward pass. Here `var_prefill_length` and `max_prefill_length` come from `prompt_len`, whose existing contract is the final/full prompt length, while `p_i` is the current `num_computed_tokens` and `s_i` is the current scheduled chunk. For chunked prefills, `prompt_len != p_i + s_i`, so both the variance correction and the `topk` regime can be wrong. Please preserve the existing full-prompt field semantics and compute separate current-end-length statistics from `num_computed_tokens + num_scheduled_tokens`.

components/src/dynamo/vllm/backend_args.py
```diff
@@ -406,6 +406,58 @@ def add_arguments(self, parser) -> None:
...
+                "runs. Off by default -- the spreads exist to calibrate an "
+                "intra-batch work-delta correction and cost several forward "
+                "passes per coordinate."
+            ),
+        )
+        add_argument(
+            g,
+            flag_name="--benchmark-imbalance-repeats",
+            env_var="DYN_BENCHMARK_IMBALANCE_REPEATS",
+            default=5,
```
tedzhouhk: Please justify `5` as the default repeat count with measurement data. Repeats apply to the uniform reference as well as every selected imbalance shape, so one original coordinate can become roughly 15–35 prefill forwards; the PR's 10-coordinate → 26-forward smoke run at `repeats=1` would become 130 forwards with this default. Also, identical repeats are emitted contiguously, so thermal/drift correlation may reduce the benefit. Could we show coefficient or confidence-interval stability for repeats 1/3/5 and choose the lowest sufficient default, or use a lower default and let dedicated calibration runs opt into 5?

tedzhouhk (commented):
Three inline comments on calibration correctness and the default repeat cost.

components/src/dynamo/vllm/benchmark_imbalance.py
YijiaZhao: Fixed in 94a63bb — you were right that the reference and the shapes had drifted apart, and thank you for naming where it bites: the C+1 coordinates are exactly the ones this got wrong, and there the missing token is a capture-size jump rather than a token of work.

Took the first option, planning against the exact totals. `plan_cell` now takes the cell's conserved pair and threads it into `pure_rows` / `mixed_rows`; `_settle` already existed to push a remainder into a row the regime leaves headroom on, so conservation came almost for free. The KV axis is dealt in whole blocks (`_block_split`) so no row goes ragged, and a shape whose totals drift is dropped rather than emitted.

`work_columns` needed the matching half of it: the subtrahend is now the reference batch's own rows rather than `b` copies of the average request. That matters beyond bookkeeping — a reference carrying its remainder on one row can sit on the other side of `topk` from its neighbours, which `b` copies of the average cannot express, and crediting it per row is what keeps the sparse and dense columns clean in that case.

Test: `test_every_shape_conserves_the_cell_s_exact_totals` over `(4, 4097, 0)`, `(8, 32769, 8192)` and `(4, 4095, 256)`.

components/src/dynamo/vllm/instrumented_scheduler.py
YijiaZhao: Fixed in 94a63bb, along the lines you asked for: the existing full-prompt fields keep their contract and the current-end-length statistics are separate.

`var_prefill_length` is untouched — it is documented as the FULL prompt length and other consumers read it. `max_prefill_length` becomes `max_prefill_end_length` and a new `var_prefill_end_length` joins it, both computed from `num_computed_tokens + num_scheduled_tokens`. The rename is not a semantic change of an established field: that field is new in this PR and its own comment already said "KV read + freshly computed", so the implementation now matches what it claimed rather than the other way round.

Worth stating why this is not only a variance correction, since that was the sharper half of your point: the regime is decided by `s + p`, so the first 2k chunk of a 16k prompt was classified sparse on a pass that runs entirely dense, and the label went to the wrong price.

Verified per row that `end == p + s` for both the new-request and the cached-continuation paths, with `p` and `s` being the same quantities the existing code already feeds `sum_prefill_kv_tokens` and `sum_prefill_tokens`:

| case | p | s | end | old length |
|---|---|---|---|---|
| chunk 1/8 of a 16k prompt | 0 | 2048 | 2048 | 16384 |
| chunk 3/8 | 4096 | 2048 | 6144 | 16384 |
| cached continuation, chunk 4/8 | 6144 | 2048 | 8192 | 16384 |
| prefix hit, no chunking | 3072 | 1024 | 4096 | 4096 |

Tests: a chunked prefill whose end length is not its prompt length, and two requests with identical prompts at different chunk offsets — `var_prefill_length` reads 0.0 there while `var_prefill_end_length` reads 16.7M.

components/src/dynamo/vllm/backend_args.py
YijiaZhao: You are right that 5 was not justified, and I do not have the stability data to justify it, so the default is 1 in 94a63bb.

The argument for 1 is not only the absence of data: the ordinary generated grid measures every coordinate exactly once, and this manifest replaces that grid. Defaulting to 5 sampled this path five times denser than the rest of the benchmark on a claim nothing had established. A dedicated calibration run can still ask for more with `--benchmark-imbalance-repeats`.

On the correlation point — also right, and it applied to the layout rather than the count. Repeats are now emitted as whole passes over the sweep instead of back to back per shape, so identical shapes are separated by a full pass rather than sitting inside one thermal and clock excursion. That costs nothing and is what makes the extra forwards worth their time when someone does opt in.

The reference batch and the added shapes are sampled alike. Weighting the shared subtrahend differently would be a sampling policy the rest of the benchmark does not have, and it is not this PR's place to introduce one.

Test: `test_repeats_are_laid_out_as_passes_not_back_to_back`.

components/src/dynamo/vllm/benchmark_imbalance.py
tedzhouhk: Thanks for fixing the reference/shape subtraction, but exact KV totals are still rejected when only the floored per-request average is not block-aligned. `plan_cell` returns early on `p_bar % kv_block`, even though the new `_block_split` can distribute a globally block-aligned exact total. For example, `b=3, total_new=3072, total_kv=256, kv_block=64` has the valid reference `[(1024,128), (1024,64), (1024,64)]` and multiple valid UNSAT spread shapes, but `p_bar=85` causes this cell to emit only the reference. Please remove the average-based guard and rely on the existing `total_kv % kv_block` checks in `reference_rows` / `pure_rows` / `mixed_rows` so these exact-total cells actually sample variance.

components/src/dynamo/vllm/backend_args.py
tedzhouhk: The effective CLI default is still 5. This `add_argument` call still uses `default=5` (and advertises 5 in the help text); `ConfigBase.from_cli_args` copies the argparse namespace before considering class defaults, so changing `DynamoVllmConfig.benchmark_imbalance_repeats` to 1 does not affect normal startup. A no-flag parse currently yields `argparse=5` and `from_cli_args=5`. `build_manifest` also still has `repeats: int = 5`. Please change the parser/help and helper defaults to 1 and add a parse-level regression test, otherwise enabling imbalance collection still incurs five full passes by default.

components/src/dynamo/vllm/benchmark_imbalance.py
YijiaZhao: Fixed in f0d3e7c — the guard predates the block deal and I left it in place when the deal moved to the exact total, so it was rejecting cells on a quantity nothing is computed from any more. Removed; `total_kv % kv_block` in `reference_rows` / `pure_rows` / `mixed_rows` is what decides expressibility now, and it decides it on the quantity actually being dealt.

Your example is what the regression test uses. It now plans:

```
b=3, total_new=3072, total_kv=256, kv_block=64   (p_bar=85)
reference  [(1024,128), (1024,64), (1024,64)]
spreads    [(1294,128), (1358,64), (420,64)]
           [(1429,128), (1493,64), (150,64)]
           [(1920,128), (576,64),  (576,64)]
```

Both totals hold exactly on every row and every prefix is a whole block. Before this the cell emitted the reference alone — a forward pass that pins no coefficient, which is the worst of both: it costs collection time and contributes no variance.

Test: `test_cells_sample_variance_when_only_the_average_prefix_is_ragged`, which asserts the average is ragged and the exact total is not, so it fails if the guard comes back.

components/src/dynamo/vllm/backend_args.py
YijiaZhao: You are right, and thank you for tracing it to `from_cli_args` — I changed the class default and assumed that was the effective one, which it is not. Fixed in f0d3e7c.

All four defaults now read 1, since any one of them left at 5 reinstates the cost:

- `add_argument(--benchmark-imbalance-repeats, default=...)` in `backend_args.py`, the one a no-flag startup actually gets
- `DynamoVllmConfig.benchmark_imbalance_repeats`
- `BenchmarkConfig.imbalance_repeats`, which the env and `additional_config` paths land on
- `build_manifest(repeats=...)`

The help text now also says what 1 means rather than just stating it: the density the generated grid measures a coordinate at, which is the grid this manifest replaces.

Test: `test_imbalance_repeats_defaults_to_one_at_the_parser` asserts both layers of the path you named — the no-flag namespace and the config built from it — and that an explicit `--benchmark-imbalance-repeats 5` still arrives as 5.

One disclosure: that test lives in `test_backend_args.py`, which does not import in my environment (no built `dynamo._core`), so CI is its first real run. I verified the change by exhausting the other direction instead — there is no remaining `5` bound to this option anywhere under `components/src/dynamo/`.

tedzhouhk (approved):

PR #12633, opened 2026-08-04
fix(operator): hash resolved runtime versions

tmonty12:
## Summary

Include the canonical resolved runtime version in the v2 worker hash.

- Resolve the runtime version from `runtimeVersionOverride` when present, otherwise from the main image tag.
- Hash the resolved version only for versions `>= 1.5.0`.
- Continue excluding the raw `runtimeVersionOverride`, so an image-derived version and an equivalent explicit override produce the same hash.
- Preserve existing hash behavior for unresolved and pre-1.5 versions.
- Keep v1/v2 migration and annotation lifecycle unchanged; this uses the existing v2 rollout path.

## Behavior

The resolved runtime version becomes part of the v2 worker-generation fingerprint.

- Changing an override from `1.5.0` to `1.5.1` changes the v2 hash and triggers the normal managed rollout.
- Setting an explicit `1.5.0` override when the image already resolves to `1.5.0` leaves the v2 hash unchanged and does not roll workers.

This PR does not add migration state, target-hash annotations, or worker-rollout restructuring.

## Validation

- `go test -count=1 -vet=off ./internal/dynamo ./internal/runtimeversion`
- `go test -count=1 -vet=off ./internal/controller -run '^TestShouldTriggerRollingUpdate_UsesResolvedRuntimeVersion$'`
- `git diff --check`


xianlubird (commented):
I like this approach better than the RuntimeProfile one — hashing the resolved runtime version is simpler, and it handles image-derived and explicit versions consistently.

One small thought: `1.5.0` is defined both as the hash floor and as the first runtime gate version. Maybe we can keep that in one place, or add a test to make sure they don't drift later.

It might also be worth adding a quick test or note for `v1.5.0` prerelease/nightly images, since they may roll once when the operator starts using version hashing.

Otherwise, looks good to me.

deploy/operator/internal/dynamo/hash.go
```diff
@@ -25,10 +25,14 @@ import (
 
 	"github.com/ai-dynamo/dynamo/deploy/operator/api/v1alpha1"
 	"github.com/ai-dynamo/dynamo/deploy/operator/api/v1beta1"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/runtimeversion"
 )
 
 const dgdWorkerHashPlaceholderValue = "worker-hash-placeholder"
 
+// Preserve hashes for runtimes released before version hashing.
+var minimumHashedRuntimeVersion = runtimeversion.Version{Major: 1, Minor: 5, Patch: 0}
```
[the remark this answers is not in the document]
tmonty12: Fixed in `0d6f4b6515`. Raised `CanaryHealthChecks.MinRuntimeVersion` to `1.5.0`, updated its threshold test, and documented the rationale in the PR description. This aligns the first downstream gated `PodSpec` rendering with the version-hash floor and avoids retroactively changing released 1.4.0 workloads during an operator-only upgrade.

deploy/operator/internal/runtimeversion/runtimeversion_test.go
```diff
@@ -104,3 +104,42 @@ func TestParseImageVersion(t *testing.T) {
...
+			got, err := Resolve(tt.image, tt.override)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("Resolve(%q, %q) error = %v, wantErr %t", tt.image, tt.override, err, tt.wantErr)
+			}
+			if !tt.wantErr && got != tt.want {
+				t.Fatalf("Resolve(%q, %q) = %+v, want %+v", tt.image, tt.override, got, tt.want)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
tmonty12: Fixed in `0d6f4b6515`. Added `t.Log` story headings for setup, execution, and verification across the new runtime-version table tests and the resolved-runtime worker-hash test.

deploy/operator/internal/dynamo/hash.go
```diff
@@ -101,8 +107,49 @@ func workerHashSpec(dcd *v1beta1.DynamoComponentDeployment) v1beta1.DynamoCompon
...
+	image := ""
+	if main := GetMainContainer(component); main != nil {
+		image = main.Image
+	}
+	version, err := runtimeversion.Resolve(image, component.RuntimeVersionOverride)
+	if err != nil || version.Compare(minimumHashedRuntimeVersion) < 0 {
+		return ""
+	}
+
+	return version.String()
```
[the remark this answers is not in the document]
tmonty12: Confirmed. This compatibility guarantee applies to released runtime versions. Nightly and prerelease images such as `v1.5.0-*` are outside that contract, so a one-time managed rollout when upgrading to the first operator with version hashing is acceptable.

tmonty12:
> I like this approach better than the RuntimeProfile one — hashing the resolved runtime version is simpler, and it handles image-derived and explicit versions consistently.
> 
> One small thought: `1.5.0` is defined both as the hash floor and as the first runtime gate version. Maybe we can keep that in one place, or add a test to make sure they don't drift later.

It doesn't make sense to define the canary health check like the following:

```golang
        CanaryHealthChecks = Gate{
		Name:              "CanaryHealthChecks",
		MinRuntimeVersion: minimumHashedRuntimeVersion
	}
```

@sttts is helping add golden test manifests for previous runtime versions so we can prevent gates that would change rendering of older runtime version PodSpecs.

I've also added a comment in gates.go to be explicit that a gate can only be defined for the current version or newer. I think this is sufficient.

> 
> It might also be worth adding a quick test or note for `v1.5.0` prerelease/nightly images, since they may roll once when the operator starts using version hashing.
> 
> Otherwise, looks good to me.

https://github.com/ai-dynamo/dynamo/blob/main/docs/fern/pages/reference/general/release-artifacts.mdx#early-access-artifacts

The release artifacts documentation is explicit that the prerelease/nightlies are purely experimental and not for production use. We are only concerned about preventing rollouts on production releases. I don't think this requires further documentation.

xianlubird:
@tmonty12 Any update on this

deploy/operator/internal/dynamo/hash.go
```diff
@@ -76,9 +81,10 @@ func ComputeDGDWorkersSpecHash(dgd *v1beta1.DynamoGraphDeployment) (string, erro
 				return "", fmt.Errorf("duplicate generated worker DCD component name %q", componentName)
 			}
 			workerDCDs[componentName] = workerTemplate{
-				Labels:      GetDCDKubeLabels(dcd),
-				Annotations: GetDCDKubeAnnotations(dcd),
-				Spec:        workerHashSpec(dcd),
+				Labels:         GetDCDKubeLabels(dcd),
+				Annotations:    GetDCDKubeAnnotations(dcd),
+				RuntimeVersion: resolvedRuntimeVersionForHash(&dcd.Spec.DynamoComponentDeploymentSharedSpec),
```
sttts, line 86: **[P1] V1-only migration can certify the new runtime hash without rolling**

Deployments that already carry `nvidia.com/current-worker-hash-v2` are unaffected. For a DGD with only the legacy `nvidia.com/current-worker-hash`, however, this new v2-only input can be recorded as current before rollout detection runs:

- the legacy v1 hash deliberately clears `RuntimeVersionOverride`, so an override-only change still matches v1;
- `reconcileWorkerRollout` calls `migrateCurrentWorkerHashIfNeeded` before `shouldTriggerRollingUpdate`;
- migration sees matching v1 plus missing v2 and writes `next.v2 = desired.v2` without checking the active worker DCDs;
- the later rollout comparison now sees current v2 equal to desired v2 and creates no new generation.

If an override is changed to a 1.5+ runtime before the first v2 migration reconcile, the operator can therefore mark the new version-gated `PodSpec` identity as active while the old worker generation is still running. This silently defeats the rollout guarantee this change is intended to provide.

Please derive the v2-only inputs from the active worker generation before backfilling, or leave v2 unset when equivalence cannot be proven so the normal rollout path runs. A regression test should cover both the changed-override case (must roll) and an unchanged v1-only migration (must not roll).

sttts (commented):
One blocking rollout-compatibility issue is called out inline.

deploy/operator/internal/dynamo/hash.go
tmonty12: Addressed in f2e38486da. The active-DCD migration regression is now table-driven: unchanged runtimeVersionOverride (1.5.0 -> 1.5.0) backfills v2 and does not roll; changed override (1.5.0 -> 1.5.1) backfills the active v2 and triggers the desired v2 rollout. Focused controller/hash tests pass.

deploy/operator/internal/controller/dgd_worker_rollout_reconciler.go
```diff
@@ -354,6 +358,45 @@ func (r *dgdWorkerRolloutReconciler) migrateCurrentWorkerHashIfNeeded(
...
+		dcd := &dcdList.Items[i]
+		if dynamo.IsWorkerComponent(string(dcd.Spec.ComponentType)) {
+			activeWorkerDCDs = append(activeWorkerDCDs, dcd)
+		}
+	}
+	if len(activeWorkerDCDs) == 0 {
+		return "", false, nil
+	}
+
+	activeV2Hash, err := dynamo.ComputeDCDWorkersSpecHash(activeWorkerDCDs)
```
sttts: **[P1] Persisted checkpoint state makes unchanged old DGDs roll**

`ComputeDCDWorkersSpecHash` is applied to persisted worker DCDs, but those are not in the same canonical shape as the DCDs produced by `ComputeDGDWorkersSpecHash`. In particular, `applyCheckpointStartupPolicy` materializes `checkpointRef`, startup policy, and restore-candidate pod-template annotations before storing a DCD. Those fields remain in `workerHashSpec`/`GetDCDKubeAnnotations`; the new normalization removes only worker-hash labels and `restartAt`.

An unchanged v1-only DGD with an existing checkpoint can therefore produce `activeV2 != desiredV2` after an operator upgrade and get an unexpected managed worker rollout. This is a real legacy upgrade shape because checkpoint materialization predates lazy v2-hash convergence.

Please reconstruct only the new runtime-version identity on top of inputs already proven equal by v1, rather than hashing the complete persisted DCD. Add a regression test with a v1-labeled DCD passed through `applyCheckpointStartupPolicy`: the unchanged DGD must not roll, while the changed-override case must still roll.

deploy/operator/internal/controller/dgd_worker_rollout_reconciler_envtest_test.go
```diff
@@ -0,0 +1,100 @@
...
+	ctx := context.Background()
+	env := sharedEnv.ForTest(t)
+	kubeClient := env.Client()
+
+	t.Log("create a normal DGD and record its completed restart")
+	dgd := createTestDGD("restart-hash-parity", map[string]*nvidiacomv1alpha1.DynamoComponentDeploymentSharedSpec{
+		"worker": {ComponentType: consts.ComponentTypeWorker},
+	})
+	dgd.Namespace = env.Namespace()
+	dgd.Spec.BackendFramework = "vllm"
```
sttts: **[P2] Use the existing vLLM backend constant**

This is the third `"vllm"` literal in the package, so `goconst` fails at this line in the required operator pre-merge job and the later test/generated-file stages are skipped. Please use `string(dynamo.BackendFrameworkVLLM)` here, as the neighboring controller tests do.

sttts (commented):
Two follow-up findings from the rereview are called out inline.

deploy/operator/internal/consts/consts.go
```diff
@@ -265,6 +265,11 @@ const (
 	// AnnotationCurrentWorkerHashV2 stores the active v2 worker generation hash.
 	AnnotationCurrentWorkerHashV2 = "nvidia.com/current-worker-hash-v2"
 
+	// AnnotationTargetWorkerHashV2 stores a v2 generation selected for a
+	// managed rollout. Unlike AnnotationCurrentWorkerHashV2, it does not mean
+	// that workers in the generation are fully active yet.
```
sttts: Afaik this was not the case with current either. Current = we have reconciled the children to that hash. It says nothing about workers. They can lag behind or never get updated.

deploy/operator/internal/controller/dgd_component_program.go
```diff
@@ -202,21 +193,21 @@ func (p *componentProgram) recordRollingUpdateTransition(
...
+			desiredWorkerDCDHash(dgd, hashes),
 		)
 	case current == nvidiacomv1beta1.RollingUpdatePhaseCompleted && previous != current:
-		currentHashes := currentWorkerHashes(dgd)
-		workerHash := currentHashes.v2
+		current := currentWorkerGenerationState(dgd)
+		workerHash := current.currentV2
 		if workerHash == "" {
-			workerHash = currentHashes.v1
+			workerHash = current.currentV1
```
sttts: current.current?

sttts: Seems like currentWorkerGenerationState is misnamed now.

deploy/operator/internal/controller/dgd_worker_rollout_state.go
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+package controller
+
+import "github.com/ai-dynamo/dynamo/deploy/operator/internal/consts"
+
+// calculatedWorkerHashes are the two independently calculated identities for
+// the desired DGD. They are deliberately not derived from current annotations:
+// legacyV1 is L in the state machine and v2 is V.
+type calculatedWorkerHashes struct {
```
sttts: what does calculated mean? Do we need that qualifier?

deploy/operator/internal/controller/dgd_worker_rollout_state.go
```diff
@@ -0,0 +1,154 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controller
+
+import "github.com/ai-dynamo/dynamo/deploy/operator/internal/consts"
+
+// calculatedWorkerHashes are the two independently calculated identities for
```
sttts: not really identities. They don't identify DGDs. They are hashes over relevant worker properties that signal a required rollout.

deploy/operator/internal/controller/dgd_worker_rollout_state.go
```diff
@@ -0,0 +1,154 @@
...
+		return s.currentV1
+	}
+	return s.currentV2
+}
+
+func (s workerGenerationState) isLegacySentinel() bool {
+	return s.currentV1 == consts.LegacyWorkerHash
+}
+
+func (s workerGenerationState) isV1Only() bool {
```
sttts: by now we can exclude this case. All DGDs have a v2 hash, don't they?

deploy/operator/internal/dynamo/hash.go
tmonty12: After further discussion, we've decided this case does not need to be addressed. This already exists in the 1.4.0 operator release where we will set the v2 fingerprint hash without considering changes that only the v2 hash could detect.

deploy/operator/internal/controller/dgd_worker_rollout_reconciler.go
tmonty12: No longer relevant as we are no longer calculating v2 hash on active DCDs

deploy/operator/internal/controller/dgd_worker_rollout_reconciler_envtest_test.go
tmonty12: This test has since been removed, so no longer relevant

xianlubird (commented):
The latest shape looks good overall, and I do not see a new functional blocker beyond the accepted v1-only migration limitation. I have two follow-ups before approval:

1. The latest force-push reintroduced the test-style issues that were fixed earlier. TestComputeBetaDGDWorkersSpecHash_UsesResolvedRuntimeVersion uses a local newDGD closure, and the new runtime-version, hash, and rollout tests are missing the required t.Log story headings. Please restore the named helper and test narration required by deploy/operator/AGENTS.md.

2. Please make the public field documentation explicit that only resolved runtime versions 1.5.0 or newer are included in the worker hash. The current wording, changing it may trigger a rollout, does not explain that unresolved and pre-1.5.0 versions remain excluded.

tmonty12:
Addressed in `6af032da97`:

1. Restored the named `betaDGDWithRuntimeVersion` helper and added the required `t.Log` story headings to the new runtime-version, hash, and rollout tests.
2. Updated the public `RuntimeVersionOverride` documentation to state that setting or changing an override which resolves to `1.5.0` or later may trigger a rollout.

I intentionally kept the public field documentation at that user-visible behavior; it does not expose the worker-hash implementation or enumerate the unresolved/pre-`1.5.0` cases.

sttts (approved):

PR #12638, opened 2026-08-04
fix(llm): skip redundant tool ban for none choice

rmccorm4:
## Summary

- Skip structural-tag tool-call bans for `tool_choice: none` when the prompt formatter already excludes tools.
- Retain the existing ban path when tools remain available.
- Add focused policy regression coverage.

The redundant AnyTokens exclusion could be rejected by vLLM when token-string exclusions lacked tokenizer metadata, even though no tools were present in the formatted prompt.

## Validation

- `cargo test -p dynamo-llm --lib tool_choice_none_skips_structural_tag_when_tools_are_excluded`

## Tracking

Internal: [DIS-2623](https://linear.app/nvidia/issue/DIS-2623/skip-redundant-structural-tag-ban-when-tool-choice-none-excludes-tools)

## Summary by CodeRabbit

* **Bug Fixes**
  * Avoided applying redundant tool restrictions when tools are already excluded by configuration.
  * Preserved existing behavior for other tool-selection settings.

* **Tests**
  * Added coverage for configurations where the restriction should be skipped.

## Why vLLM 0.26.0 crashes

vLLM uses `auto` as its default structured-output backend. For this request:

1. XGrammar first validates the canonical structural tag generated by Dynamo.
2. Resolving `exclude_tokens: ["<tool_call>"]` requires tokenizer information, but the validation call does not supply it. XGrammar rejects the constraint with `InvalidStructuralTagError: Token string resolution requires tokenizer_info`.
3. vLLM catches that validation failure and falls back to llguidance.
4. The llguidance path only understands the legacy structural-tag representation with top-level `triggers` and `structures`. Dynamo supplied the canonical `format` representation, so the fallback raises `KeyError: 'triggers'`.

The request therefore fails before inference and surfaces as an HTTP 500 instead of returning a normal response with no tool call.

### DIFF
--- a/lib/llm/src/preprocessor/structural_tag.rs
+++ b/lib/llm/src/preprocessor/structural_tag.rs
@@ -22,6 +22,10 @@ fn requires_intrinsic_structural_tag(parser_name: Option<&str>, tool_choice: &To
     is_kimi_k3_parser(parser_name) && matches!(tool_choice, ToolChoice::Named(_))
 }
 
+fn should_skip_tool_call_ban(exclude_tools_when_none: bool, tool_choice: &ToolChoice) -> bool {
+    exclude_tools_when_none && matches!(tool_choice, ToolChoice::None)
+}
+
 impl OpenAIPreprocessor {
     /// Apply structural tag guided decoding when enabled for this request.
     pub(super) fn apply_tool_choice_structural_tag(
@@ -50,6 +54,16 @@ impl OpenAIPreprocessor {
         let Some(builder) = Self::structural_tag_builder_for_parser(parser_name) else {
             return Ok(false);
         };
+
+        if should_skip_tool_call_ban(
+            self.runtime_config.exclude_tools_when_tool_choice_none,
+            tool_choice,
+        ) {
+            // The prompt formatter already omits tools for this request. Avoid
+            // sending a redundant AnyTokens structural tag: vLLM cannot
+            // validate token-string exclusions without tokenizer metadata.
+            return Ok(false);
+        }
 
         if matches!(tool_choice, ToolChoice::None) {
             if tools.is_empty() {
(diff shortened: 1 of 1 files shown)


lib/llm/src/preprocessor/structural_tag.rs
```diff
@@ -51,6 +55,16 @@ impl OpenAIPreprocessor {
...
 
+        if should_skip_tool_call_ban(
+            self.runtime_config.exclude_tools_when_tool_choice_none,
+            tool_choice,
+        ) {
+            // The prompt formatter already omits tools for this request. Avoid
+            // sending a redundant AnyTokens structural tag: vLLM cannot
+            // validate token-string exclusions without tokenizer metadata.
+            return Ok(false);
+        }
```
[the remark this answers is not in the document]
rmccorm4: Confirmed against the pinned `dynamo-renderer` 5.0.0 implementation. The native DeepSeek formatter does not receive the configuration boolean because it handles this directly from `OAIChatLikeRequest::tool_choice()`: `inject_tools_and_response_format` sets the tools to `None` for `tool_choice="none"`. Both V3.2 and V4 call that shared function, and the renderer has regression coverage asserting that the tools block and schema are omitted. Therefore the structural-tag skip remains safe for the DeepSeek native path.

lib/llm/src/preprocessor/structural_tag.rs
```diff
@@ -51,6 +55,16 @@ impl OpenAIPreprocessor {
             return Ok(false);
         };
 
+        if should_skip_tool_call_ban(
```
indrajit96: NIT:
Codex recommends moving this block to line 45 so as to fail early (I agree)
This check is independent of parser_name and builder. 
When tool_choice=None and prompt-side tool exclusion is enabled, the request intentionally requires no structural tag. 
Checking the parser first can therefore emit an irrelevant missing/unsupported-parser warning and perform unnecessary parser lookup before eventually returning the same Ok(false).

indrajit96 (approved):
LGTM!
A nit to fail early to avoid confusing warning

lib/llm/src/preprocessor/structural_tag.rs
rmccorm4: Fixed in dbcddd328123be85e698e2bcd5ea16f591c748a1. Moved the `tool_choice=none` prompt-exclusion return ahead of parser and builder lookup, so this path no longer performs unnecessary lookup or emits irrelevant missing/unsupported-parser warnings. Validated with formatting and diff checks plus `cargo check -p dynamo-llm --tests`.

PR #12640, opened 2026-08-04
fix(llm): preserve explicit zero chat temperature

rmccorm4:
## Summary

- Preserve an explicitly supplied `temperature: 0` when applying chat-completions request-template defaults.
- Apply the configured template temperature only when the request omits temperature.
- Extract template application into a focused, unit-testable helper.

The previous `unwrap_or(0.0) == 0.0` check conflated `None` with `Some(0.0)`, changing deterministic greedy requests into sampled requests.

## Validation

- `cargo test -p dynamo-llm --lib test_chat_completions_template_preserves_explicit_zero_temperature`

## Tracking

Internal: [DIS-2621](https://linear.app/nvidia/issue/DIS-2621/preserve-explicit-zero-temperature-in-chat-request-templates)

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat completion requests now correctly preserve an explicitly set temperature of `0.0`.
  * Missing model, temperature, and token-limit settings are automatically filled with the appropriate defaults.


lib/llm/src/http/service/openai.rs
```diff
@@ -1964,6 +1964,23 @@ fn accumulate_reasoning_dispatch(
 ///
 /// Note: For all requests, streaming or non-streaming, we always call the engine with streaming enabled. For
```
indrajit96: Nit : move the func above the comments can cause confusion.

indrajit96 (approved):
LGTM!
Thanks for spoting and fixing this :)

lib/llm/src/http/service/openai.rs
rmccorm4: Done, nice call: 1b1c5de4ff5a91c8eb5e023f863541367c72622f

lib/llm/src/http/service/openai.rs
```diff
@@ -1956,6 +1956,23 @@ fn accumulate_reasoning_dispatch(
...
+            request.model = template.model.clone();
+        }
+        if request.temperature.is_none() {
+            request.temperature = Some(template.temperature);
+        }
+        if request.max_completion_tokens.unwrap_or(0) == 0 {
+            request.max_completion_tokens = Some(template.max_completion_tokens);
+        }
+    }
+}
```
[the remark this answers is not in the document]
rmccorm4: out of scope

PR #12675, opened 2026-08-04
feat(recipes): add Qwen3.6-27B benchmark recipe

ericcurtin:
## Summary

Adds `recipes/qwen3.6-27b`, a 3-way `vllm serve` vs. Dynamo (frontend-decoding,
frontend-decoding + embedding cache) benchmark recipe for `Qwen/Qwen3.6-27B`,
in preparation for `Qwen/Qwen3.8-27B` releasing next week.

Mirrors the sibling [`recipes/qwen3.6-35b`](recipes/qwen3.6-35b), with one
difference: `Qwen3.6-27B` is dense BF16 (no FP8 checkpoint), so it runs
`--tensor-parallel-size 2` instead of `1`. Resources use a `qwen3627-`
prefix so they don't collide with the sibling in a shared namespace.

## Validation

See testing-scope comment below — no GPU cluster available to me, so this
covers schema/render/control-flow checks only, not a real hardware run.

### DIFF
--- a/recipes/qwen3.6-27b/README.md
+++ b/recipes/qwen3.6-27b/README.md
@@ -1,0 +1,226 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Qwen3.6-27B — 3-way `vllm serve` vs Dynamo benchmark
+
+K8s recipe for benchmarking
+[`Qwen/Qwen3.6-27B`](https://huggingface.co/Qwen/Qwen3.6-27B)
+across three configs on the same multi-GPU hardware target:
+
+| Config         | Stack         | Multimodal | Frontend-decoding | Embedding cache |
+|----------------|---------------|------------|-------------------|-----------------|
+| `vllm-serve`   | vanilla vLLM  | n/a        | n/a               | n/a             |
+| `dynamo-fd`    | Dynamo + vLLM | on         | on                | off             |
+| `dynamo-fd-ec` | Dynamo + vLLM | on         | on                | 8 GiB           |
+
+All three configs share one hardware target (H100 or GB200) chosen at
+deploy time via `--hw {h100,gb200}`. See [Hardware targets](#hardware-targets) below.
+
+This recipe mirrors the layout of the sibling
+[`qwen3.6-35b`](../qwen3.6-35b) recipe (which benchmarks
+`Qwen/Qwen3.6-35B-A3B-FP8`), with one key difference: `Qwen/Qwen3.6-27B`
+is a **dense** 27B-param model (~56 GB checkpoint) shipped in **BF16**
+(no FP8 checkpoint yet), so every config here runs
+`--tensor-parallel-size 2` — the pinned node needs **2 free GPUs**, not 1.
+
+## Pre-requisites
+
+1. Kubectl context pointing at a cluster with the right GPUs (2 free
+   GPUs on the pinned node — see point 4).
+2. A namespace you have write access to (`$NAMESPACE` below).
+3. A `shared-model-cache` PVC in that namespace (RWX). If your cluster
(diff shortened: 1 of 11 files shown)


ericcurtin:
**Testing note (no GPU cluster available to me):**
- ✅ All YAML parses; native k8s resources (Job/Pod/Deployment/Service) validated with kubeconform against real k8s 1.29 schema; DGD manifests validated against this repo's own CRD schema
- ✅ Rendered every template through `envsubst` for both hw targets — valid output
- ✅ `run-benchmark.sh`/`run-all-benchmarks.sh` control flow (arg parsing, error paths, full deploy→bench→retrieve→clean) smoke-tested against a mocked `kubectl`
- ❌ Not tested: real pod scheduling, vLLM boot, an actual aiperf run — needs real H100/GB200 hardware

ericcurtin:
Addressed CodeRabbit's 8 comments in 2dd62f2:
- Fixed: missing README SPDX header, 28B→27B param wording, stale `deploy.yaml`/`config.env` refs, hf-token-secret instructions narrowed to the one file that actually has the block, retrieve() doc corrected to the real nested extraction path.
- Fixed (the one functional issue): readinessProbe `failureThreshold` 60→120 and rollout-status timeout 900s→1800s — the old budget (~11 min) was under this BF16 model's documented ~14 min bootstrap.
- Not changed: retrieve()'s tar extraction itself (CodeRabbit's alternate diff) — it already works today (`find` locates the file fine), and I can't verify a change to the tar/untar logic against a real cluster from here, so I fixed the docs to match actual behavior instead of touching the script's proven-in-the-sibling logic.

Note: several of these (missing README SPDX, config.env comment, tar nesting, hf-token-secret text) are copy-paste carryovers from the sibling `qwen3.6-35b` recipe, not new — fixed here since they're low-risk, left the sibling alone since that's out of scope for this PR.

ericcurtin:
Addressed dynamo-review-agent's 2 comments in 0754838:
- `kubectl wait` on a fresh DGD's pod label can race the operator (errors immediately with "no matching resources found" if 0 pods exist yet). Added a bounded `wait_pod_created()` poll before both `kubectl wait --for=condition=Ready` calls in `deploy()`.
- `bench()`'s `kubectl logs -f` was blocking through perf.yaml's post-run `sleep 3600` on every config (up to +3h across a full `run-all-benchmarks.sh` sweep). Piped through `awk '{print} /^Run complete\./{exit}'` so it still streams live but returns as soon as the run finishes, instead of waiting out the idle sleep.

Both smoke-tested against a mocked kubectl (poll-then-succeed race, and early-exit-without-losing-output) since I don't have a real cluster here — see the earlier testing-scope comment.

Same as last round: this exact code (including the race and the blocking log-follow) is copied verbatim from the sibling `qwen3.6-35b` recipe, so it's pre-existing there too — fixed here only, left the sibling out of scope for this PR.

ericcurtin:
Addressed both in 567d73f:
- perf.yaml's `/v1/models` readiness loop had no deadline. Added `MODEL_READY_TIMEOUT_SECONDS` (default 1800s), exits 1 past it.
- `bench()`'s new awk pipe from last round exited 0 unconditionally, masking a run that never printed "Run complete." — now tracks whether the marker was seen and exits 1 if not (dropped the `|| true`), so a real failure aborts the sweep instead of silently continuing to retrieve/clean.

Both isolate-tested (timeout fires correctly; success/failure paths verified against a mocked kubectl).

ericcurtin:
Good catch, fixed in eea5d36: `logs -f | awk{exit}` doesn't return early — bash waits for every pipeline stage, and `kubectl logs -f` won't exit until the container does (no more writes after "Run complete." means no SIGPIPE either). Replaced with a backgrounded `logs -f` piped through a FIFO, read in a foreground loop, explicitly killed once the marker's seen (still fails if the stream ends without it).

Reproduced the exact bug with a stand-in that goes quiet after printing the marker (old code hangs; confirmed via `timeout`), then verified the fix returns immediately, the failure path still aborts, and no orphaned process is left behind.

PR #12678, opened 2026-08-05
fix(vllm): flatten 0-d omni audio payloads before delta comparison

GuanLuo:
## Overview:

`tensor_to_numpy` in the realtime Omni handler only flattened when `ndim > 1`, so a **0-d** array passed through unchanged and later crashed the turn.

`extract_audio_chunks` treats a list payload as "list of waveforms, take the last". If that last entry is a scalar, `np.asarray` produces a 0-d array, which becomes `self.audio_ref`. On the next step, `numpy_audio_prefix_match` reads `prev.shape[0]` and raises:

```
IndexError: tuple index out of range
```

`run_turn`'s blanket handler converts that into an `error` event plus `response.done(status=failed)`, so one scalar step fails the whole turn.

## Details:

Flatten unconditionally. `reshape(-1)` turns a 0-d array into a 1-element 1-d array and is a no-op for the 1-d case that already worked, so every downstream consumer (`shape[0]`, slicing, `concatenate`) gets the 1-d array it already assumes.

Found while fixing an unrelated payload-shape bug in #12667 and split out of it, since this path is reachable on `main` independently of that change. A `graham-code-review` pass flagged mixing the two.

`components/src/dynamo/vllm/omni/realtime_handler.py` — `tensor_to_numpy` is a two-line change.

**Validation.** `pre-commit run --files <changed>` passes clean.

The test could not be executed locally (this checkout has no `vllm_omni` install), so the real module text was run old-vs-new against stubs, driving the exact two-step sequence the new test asserts:

| | `main` | with this fix |
|---|---|---|
| scalar step, then a real delta | `IndexError: tuple index out of range` | `first=[[0.25]]`, `second=[[0.5]]` |

## Related Issues

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12678" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved realtime audio processing for scalar and vector payloads.
  * Prevented previously processed audio samples from being repeated across consecutive updates.

* **Tests**
  * Added coverage for consecutive audio payloads, including list, scalar, and NumPy-based inputs.

### DIFF
--- a/components/src/dynamo/vllm/omni/realtime_handler.py
+++ b/components/src/dynamo/vllm/omni/realtime_handler.py
@@ -462,8 +462,10 @@ def tensor_to_numpy(value: Any) -> np.ndarray | None:
             arr = np.asarray(value)
         except Exception:  # noqa: BLE001 - non-array engine payloads are skipped
             return None
-    if arr.ndim > 1:
-        arr = arr.reshape(-1)
+    # Flatten unconditionally, including 0-d: a list payload yields its last
+    # entry, so a scalar there would reach waveform_to_deltas as a 0-d array and
+    # raise IndexError on the next step's shape[0].
+    arr = arr.reshape(-1)
     return arr.astype(np.float32, copy=False)
 
 
--- a/components/src/dynamo/vllm/tests/omni/test_omni_realtime_handler.py
+++ b/components/src/dynamo/vllm/tests/omni/test_omni_realtime_handler.py
@@ -24,7 +24,7 @@ try:
     # so the fake engine outputs below must use the real type, not a plain dict.
     from vllm_omni.outputs.mm_outputs import MultimodalPayload
 
-    from dynamo.vllm.omni.realtime_handler import RealtimeOmniHandler
+    from dynamo.vllm.omni.realtime_handler import RealtimeOmniHandler, Turn
 except (ImportError, ModuleNotFoundError):
     pytest.skip("vLLM omni dependencies not available", allow_module_level=True)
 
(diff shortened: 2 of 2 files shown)


ptarasiewiczNV (approved):

GuanLuo:
Review context from #12667, where this crash was found.

@rmccorm4 [suggested](https://github.com/ai-dynamo/dynamo/pull/12667#discussion_r3722415911) handling 0-D arrays by **rejecting** them rather than flattening:

```python
if arr.ndim == 0 or arr.dtype.kind not in "biuf":
    return None
```

Both stop the `IndexError`. The difference is what happens to the value:

- **Flatten** (this PR): the scalar becomes a one-sample waveform and is emitted to the client as an audio delta.
- **Reject**: the step is skipped and nothing is emitted.

Reject is arguably more correct — a list payload is a list of *waveforms*, so a bare scalar is malformed rather than a legitimate one-sample chunk, and flattening manufactures an audio delta out of it. Repro against the current code, where `[0.25]` is that fabricated delta:

```
step 1: model_outputs = 0.25 (numeric scalar)
   -> [0.25]   audio_ref.shape = ()
step 2: a normal audio chunk arrives
   -> IndexError: tuple index out of range
```

Flagging rather than switching unilaterally, since either resolves the bug this PR is for. Happy to swap it to the reject form — it is a one-line change plus a test assertion.

furionw:
> Reject is arguably more correct — a list payload is a list of waveforms, so a bare scalar is malformed rather than a legitimate one-sample chunk

my understanding is, valid inputs are

```
  raw_audio = np.array([0.25, 0.50])  # one waveform

  or:

  raw_audio = [
      np.array([0.10, 0.20]),
      np.array([0.25, 0.50]),
  ]  # list of waveforms
```

so `raw_audio=[0.25]` is malformed input that we might just want to reject it instead of making it valid?

furionw (approved):
chatted offline. From Guan --

> I think it is unlikely but the model could spill waveform one by one as scalar, this can be interpreted as normalization

I read
- https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py#L548
- https://github.com/vllm-project/vllm-omni/blob/b4581b293beb837a2936b16c0cad6d2834fed9f6/vllm_omni/entrypoints/openai/realtime_connection.py#L50-L52

I don't think this is true for qwen3-omni, **I am leaning towards failing explicitly**, but I am fine if you want to take another route considering unknown models.

PR #12683, opened 2026-08-05
fix(frontend): report real ISL for requests rejected during preprocessing

GuanLuo:
## Summary

Requests that fail during preprocessing never reach a worker, so no metrics annotation ever comes back to report their input size. Two problems followed from that:

1. `ResponseMetricCollector::drop` recorded `input_tokens` unconditionally from an `isl` field that defaults to `0`. Every such request logged `input_tokens=0` — a fabricated value, indistinguishable from a prompt that genuinely tokenized to nothing, on both the log line and the exported OTel span attribute.
2. Nothing recorded the real ISL at all, even though the frontend knows it: `gather_tokens` computes the token count and hands it straight to the context-length check.

The practical cost is diagnostic. A context-length rejection and, say, a tool-choice conflict produce identical telemetry — both surface as a generic frontend error whose real cause exists only in the client-facing response body. There is no way to tell them apart, or to size either population, from logs or metrics.

### Changes

**`metrics.rs`** — `isl` becomes `Option<usize>`; `Drop` stamps the span field only once the value is known. This matches how every neighbouring field on that path already behaves: the OSL histogram guards on `osl > 0`, and the multimodal counters on `multimodal_counts_observed` with a comment about not stamping a false zero. `input_tokens` was the one that was missed.

**`preprocessor.rs`** — `gather_tokens` records ISL on the enclosing `http-request` span as soon as the token count is known, before the check that can reject the request.

The two are interdependent by design: the first is what stops `Drop` from clobbering the second on the error path.

### Notes

- Preprocessing runs inline on the caller's task — the `spawn_blocking` in the encode path wraps only the `tokenizer.encode()` closure — so the span is current at that point. Outside an inference span, `record` is a no-op.
- **Only the log/OTel span field is affected.** The `input_sequence_length` histogram is published from `observe_response` on the first-token path and is untouched, so ISL percentiles keep their current meaning and no dashboard shifts.
- This is deliberately narrower than the `TODO: publish ISL as soon as the tokenization process completes` at `metrics.rs`. Doing that properly means changing what the ISL histogram measures (it would start including rejected and cancelled requests), which needs a label or a separate metric — a metrics-schema decision, filed separately rather than smuggled in here.
- Scope limit: this covers failures *after* tokenization. Tool-choice and multimodal rejections happen before a token count exists, so those correctly report no ISL. That asymmetry is itself useful — post-deploy, a rejected request carrying a real ISL is a context-length rejection, one without is something else.

### Scope

Stamped **only on the context-length rejection path**. On the success path the response path already stamps the identical count, and a second write is not idempotent — `tracing_subscriber`'s field formatter appends rather than replaces, so recording in both places emits `input_tokens=N input_tokens=N` in the log line. Deleting the response-path write instead would be wrong: where the frontend does not tokenize, the annotation is the only source of ISL.

Preprocessing failures after tokenization for other reasons (tool choice, multimodal parts) still record no ISL, exactly as before this branch — no regression, just not an improvement there. Post-deploy this asymmetry is itself a discriminator: a rejected request carrying an ISL is a context-length rejection; one without is something else.

## Validation

```
cargo fmt --all
cargo clippy -p dynamo-llm --no-deps --all-targets -- -D warnings   # clean
cargo test -p dynamo-llm --lib http::service::metrics::             # 43 passed
cargo test -p dynamo-llm --lib preprocessor::                       # 87 passed
cargo test -p dynamo-llm --test preprocessor_isl_span               # 2 passed
```

**End-to-end, on the real preprocessing reject path** (`tests/preprocessor_isl_span.rs`) — this is the case the change exists for, so it is covered directly rather than argued from code shape:

- `context_length_rejection_still_records_isl` — a request rejected for exceeding the context length must still carry `input_tokens` on the request span. This also pins down the assumption the fix rests on: that no child span sits between the HTTP handler and `gather_tokens`, so `Span::current()` there is the span the frontend logs under.
- `accepted_request_records_the_same_isl` — an accepted request records exactly the count preprocessing produced.

Both fail with `None` if the record in `gather_tokens` is removed. The fixture's tokenizer is a stub whose token count does not track prompt length, so the rejection is forced by lowering the context budget rather than by lengthening the prompt.

**Unit:**

- `drop_stamps_input_tokens_only_once_isl_is_known` — captures real span records through a `tracing_subscriber` layer (reusing the `RequestIdCaptureLayer` pattern from `generate.rs`) and asserts both directions: unset when no response was observed, stamped with the exact count when one was. Confirmed non-vacuous by reverting the guard and observing `left: Some(0), right: None` — the precise false zero being fixed.
- `validate_token_count_rejects_a_prompt_that_fills_the_context` — covers the `>=` boundary (65,535 accepted, 65,536 rejected; the bound is `>=` because `context_length` is the combined input+output budget) and the `context_length == 0` skip. That function previously had no tests.

### DIFF
--- a/lib/llm/src/http/service/metrics.rs
+++ b/lib/llm/src/http/service/metrics.rs
@@ -560,7 +560,10 @@ pub struct ResponseMetricCollector {
     // be computed.
     last_response_time: Option<Duration>,
     osl: usize,
-    isl: usize,
+    // `None` until the input sequence length is known. Requests that fail during
+    // preprocessing never learn one, and stamping a zero for them would be a
+    // fabricated value rather than a measurement.
+    isl: Option<usize>,
     ttft_ms: Option<f64>,
     itl_sum_secs: f64,
     itl_count: u64,
@@ -1608,7 +1611,7 @@ impl ResponseMetricCollector {
             last_response_time: None,
             start_time: Instant::now(),
             osl: 0,
-            isl: 0,
+            isl: None,
             ttft_ms: None,
             itl_sum_secs: 0.0,
             itl_count: 0,
@@ -1779,7 +1782,7 @@ impl ResponseMetricCollector {
         }
 
         // Store ISL for span recording on drop
-        self.isl = isl;
+        self.isl = Some(isl);
 
         // Increment the real-time output tokens counter
         self.output_tokens_counter.inc_by(num_tokens as u64);
@@ -1911,7 +1914,12 @@ impl Drop for ResponseMetricCollector {
         // Record request summary on the enclosing span.
         // InflightGuard::Drop and on_response logs will inherit these.
         let span = tracing::Span::current();
-        span.record("input_tokens", self.isl as u32);
+        // Only stamp ISL once known. Leaving the field empty keeps "we never got
+        // far enough to tokenize" distinguishable from a real zero, and lets a
+        // value recorded earlier in preprocessing survive this drop.
+        if let Some(isl) = self.isl {
+            span.record("input_tokens", isl as u32);
+        }
         span.record("output_tokens", self.osl as u32);
         // Only record once observed, so requests that never carried a metrics
         // annotation (e.g. errored before the first chunk) don't stamp a false zero.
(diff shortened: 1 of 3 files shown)


rmccorm4 (approved):
Seems like /v1/completions is missing the plumbing to propogate ISL information (both before and after this PR) if you want to follow-up with this, but this PR does improve observability for /v1/chat/completions, so LGTM

PR #12684, opened 2026-08-05
fix: Add tool-call constraints to vllm chat processor

indrajit96:
## Overview

Fixes [DIS-2625](https://linear.app/nvidia/issue/DIS-2625).

Tool-calling requests through Dynamo's Python vLLM chat processor could come back as plain text with `tool_calls: null`. The frontend rendered the tools into the prompt but never sent the matching decoding constraint to the backend, so the model generated unconstrained text that often could not be parsed back into a tool call.

The fix builds the tool constraint during preprocessing — through the configured vLLM `ToolParser` where it supports one, otherwise a JSON schema derived from `tools` — and forwards it as `sampling_options.guided_decoding`. Parsers that need their own wire format (those reporting `supports_required_and_named = false`) are left alone, since forcing JSON on them breaks the format they emit.

Production changes are limited to the Python vLLM path.

## Example

```json
{
  "model": "Qwen/Qwen3-0.6B",
  "messages": [{"role": "user", "content": "weather in Paris?"}],
  "tools": [{"type": "function", "function": {                            # <--- without this, no tool guidance is built at all
    "name": "get_weather",
    "parameters": {"type": "object", "properties": {"city": {"type": "string"}}}
  }}],
  "tool_choice": "required"                                               # <--- this is what makes it a forced call
}
```

Before, the response could contain no parsed tool call:

```json
{"message": {"content": "I'll check the weather in Paris.", "tool_calls": null}}
```

After:

```json
{"message": {"content": null, "tool_calls": [
  {"type": "function", "function": {"name": "get_weather", "arguments": "{\"city\":\"Paris\"}"}}
]}}
```

### What controls the behavior

Request fields:

| Field | Effect |
|---|---|
| `tools` | required for any tool guidance; empty or absent means none is built |
| `tool_choice` | `required` or a named tool is "forced"; `auto` and `none` are not |
| `response_format` | dropped for a forced choice, except the `structural_tag` variant |
| `guided_*`, `structured_outputs` | 400 against a forced choice |

Server flags, defined in `components/src/dynamo/common/configuration/groups/runtime_args.py`:

| Flag | Env var | Default | Effect |
|---|---|---|---|
| `--dyn-tool-call-parser <name>` | `DYN_TOOL_CALL_PARSER` | unset | selects the parser, which decides whether a structural tag exists and whether the JSON fallback is safe. With no parser set, only the JSON fallback is available |
| `--dyn-enable-structural-tag` / `--no-dyn-enable-structural-tag` | `DYN_ENABLE_STRUCTURAL_TAG` | `false` | gates the structural tag only. The forced-choice JSON fallback runs regardless, so `required` and named calls are still constrained at the default |
| `--dyn-structural-tag-scope {auto,always}` | `DYN_STRUCTURAL_TAG_SCOPE` | `auto` | `auto`: forced choices, plus `auto` when a tool has `strict: true` or `parallel_tool_calls` is `false`. `always`: also plain `auto` |
| `--dyn-structural-tag-schema {auto,strict}` | `DYN_STRUCTURAL_TAG_SCHEMA` | `auto` | `auto`: real parameter schema only for tools with `strict: true`, others syntactically constrained only. `strict`: real schema for every tool |

The example above assumes defaults — no `--dyn-enable-structural-tag`, so the forced choice takes the JSON-schema fallback. That fallback constrains the entire generation to an array of tool calls, which is why `content` comes back `null`. With structural tags enabled the qwen/hermes-style tag allows free text around the call, so `content` can be non-null there.

## Cases considered

On `main` this path emits no `guided_decoding` at all, so none of the combinations below had a defined outcome — every row here is new behavior, not a changed one.

When a request carries both a tool constraint and an explicit output constraint, only one can occupy the decoder's single grammar slot. Which one wins depends on what the field is scoped to:

| Request | Result (expectation) | Why |
|---|---|---|
| forced `tool_choice` + `response_format: json_object` / `json_schema` | tool constraint wins, `response_format` dropped | `response_format` is scoped to the message the model returns to the user, not to tool calls. `preprocessor/tool_choice.rs` clears `gd.json` for this case, and vLLM's own `ToolParser.adjust_request` sets `response_format = None` |
| forced `tool_choice` + `guided_json` / `guided_regex` / `guided_grammar` / `guided_choice` | 400 | no field scoping — these constrain the same raw token stream the tool grammar needs |
| forced `tool_choice` + `structured_outputs` | 400 | same as above, caller-set constraint over the whole stream |
| forced `tool_choice` + `response_format: structural_tag` | 400 | vLLM normalizes this into `structured_outputs.structural_tag`, a grammar over the whole stream — not a content format |
| forced `tool_choice`, parser reports `supports_required_and_named = false` | no constraint | the parser emits a native wire format; forcing JSON breaks it |
| malformed `tool_choice` (`{}`, `{"type": "function"}` with no name) | not treated as forced | the predicate validates shape, and handles both the raw dict from the `model_construct` fast path and a validated `ChatCompletionNamedToolChoiceParam` |

Two ordering details worth knowing when reading the diff:

- Whether the caller supplied a constraint is read **before** `_prepare_request`. `ToolParser.adjust_request()` mutates that same request object and can add `structured_outputs` itself, so reading it afterwards treats parser-generated guidance as caller-supplied.
- The shared parity table records what SGLang produces today rather than skipping those cases, so that path still executes and any other change in its behavior fails.

## Validation

End-to-end on a live stack (`dynamo.frontend --dyn-chat-processor vllm` + a `dynamo.vllm` worker), same worker and seed each time, only the frontend code swapped. The request has the prompt and `tool_choice` disagree, so the model only complies when constrained:

```json
{"messages": [{"role": "user", "content": "Do NOT call any function. Just reply with the single word: hello"}],
 "tools": [{"type": "function", "function": {"name": "get_weather", "parameters": {"type": "object",
   "properties": {"location": {"type": "string"}}, "required": ["location"]}}}],
 "tool_choice": "required", "temperature": 0, "seed": 42}
```

| Model | Parser | Constraint path | `9a7b29e` (pre-PR) | `d9b81dd` |
|---|---|---|---|---|
| Qwen3-0.6B | `hermes` | JSON-schema fallback | `tool_calls: null` | tool call parsed |
| Mistral-7B-Instruct-v0.3 | `mistral` | parser grammar via `adjust_request` | `tool_calls: null` | tool call parsed |

Both at default flags — no `--dyn-enable-structural-tag`. The two rows exercise different branches of `build_tool_call_guided_decoding`: Qwen takes the JSON-schema fallback, Mistral's parser supplies its own grammar.

The PR author's original scenario matrix, on H100 with 10 deterministic requests per scenario:

| Model | Scenario | Before | After |
|---|---|---:|---:|
| `Qwen/Qwen3-0.6B` | hidden schema, `tool_choice="auto"` | 0/10 | 10/10 |
| `TinyLlama/TinyLlama-1.1B-Chat-v1.0` | hidden schema, `tool_choice="required"` | 0/10 | 10/10 |
| `mistralai/Ministral-3-3B-Instruct-2512` | prompt requests no call while `tool_choice="required"` | 0/10 | 10/10 |

(Ministral-3-3B is `Mistral3ForConditionalGeneration` with a `vision_config`, and under `--dyn-chat-processor vllm` it currently fails to register with `Failed to apply PixtralProcessor on data={'text': '[IMG]'}` — so that row was presumably measured through a different configuration. Tracked in DIS-2625.)

Other checks at `d9b81dd`:

- Processor unit tests: vLLM 97 passed, SGLang 207 passed. The new cases fail against the previous head.
- Shared guidance-policy cases: 8, executed by both backends.
- Frontend-to-backend forwarding integration tests: 2 passed.
- Pre-commit hooks passed.

Note: no test in this repo covers the constrained round trip end to end — coverage is unit tests plus a payload-level integration assertion, so the e2e result above is not regression-protected.

1. `components/src/dynamo/frontend/prepost.py` builds the constraint and decides which one wins.
2. `components/src/dynamo/frontend/vllm_processor.py` reads the runtime settings and forwards it.
3. `components/src/dynamo/frontend/tests/_tool_guidance_parity.py` is the case table both backends run.
4. `tests/frontend/test_vllm_prepost_integration.py` checks the frontend-to-worker boundary.

## Follow-up

The SGLang frontend path has the same class of divergence in its tool-call guided-decoding policy (operator flags never read, `response_format` beating a forced `tool_choice`, legacy `guided_*` dropped). That is tracked separately in [DIS-2643](https://linear.app/nvidia/issue/DIS-2643) and is **not** touched by this PR — `sglang_prepost.py` and `sglang_processor.py` carry only comments here.

## Summary by CodeRabbit

* **New Features**
  * Added structured tool-call guidance for vLLM chat requests.
  * Added configurable structural-tag modes, scopes, and schema handling.
  * Improved support for named tools, response formats, and schema-based argument processing.
* **Bug Fixes**
  * Added validation for conflicting forced-tool and assistant constraints.
  * Preserved appropriate guidance when multiple request constraints are provided.
* **Tests**
  * Expanded coverage across tool choices, schemas, runtime configurations, and vLLM/SGLang behavior.


components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -168,6 +185,70 @@ def _build_reasoning_parser_metadata(
...
+            or request.get("guided_regex") is not None
+            or request.get("guided_grammar") is not None
+            or bool(request.get("guided_choice"))
+        )
+        if has_explicit_guidance:
+            raise ValueError(
+                "guided decoding cannot be used in the same request as "
+                'tool_choice="required" or a named tool_choice.'
+            )
+        return tool_guided_decoding
```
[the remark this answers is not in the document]
indrajit96: Addressed by reducing the scope of the change. _build_guided_decoding, _resolve_guided_decoding, and the custom forced-tool conflict validation have been removed. 
This PR no longer introduces a new ValueError or HTTP error-mapping path; validation and conflict semantics remain owned by vLLM.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,114 @@ class PreprocessResult:
...
+        return False
+    if tool_choice == "required" or _is_named_tool_choice(tool_choice):
+        return True
+    if tool_choice != "auto":
+        return False
+    if structural_tag_scope == "always":
+        return True
+    return request.parallel_tool_calls is False or any(
+        _tool_is_strict(tool) for tool in request.tools
+    )
```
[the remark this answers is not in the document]
indrajit96: Fixed. 
The auto-scope policy now checks that parallel_tool_calls is present in request.model_fields_set before treating False as an explicit single-call request. Added a regression test where the validated field defaults to False but was absent from the client payload.

components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -560,6 +651,18 @@ async def _generator_inner(
...
+        ):
+            tool_guided_decoding = build_tool_call_guided_decoding(
+                request_for_sampling,
+                tool_parser,
+                structural_tag_mode=self.structural_tag_mode,
+                structural_tag_scope=self.structural_tag_scope,
+                structural_tag_schema=self.structural_tag_schema,
+                starts_in_reasoning=True,
+            )
+        guided_decoding = _resolve_guided_decoding(request, tool_guided_decoding)
```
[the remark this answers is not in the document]
indrajit96: Addressed through scope reduction. 
 The starts_in_reasoning argument and the post-preprocessing guidance rebuild have been removed entirely. Dynamo now obtains the parser’s native constraint once and forwards it unchanged; reasoning-specific structural-tag policy remains with vLLM.

components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -168,6 +185,70 @@ def _build_reasoning_parser_metadata(
...
+            request.get("guided_json") is not None
+            or request.get("guided_regex") is not None
+            or request.get("guided_grammar") is not None
+            or bool(request.get("guided_choice"))
+        )
+        if has_explicit_guidance:
+            raise ValueError(
+                "guided decoding cannot be used in the same request as "
+                'tool_choice="required" or a named tool_choice.'
+            )
```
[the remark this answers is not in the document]
indrajit96: Addressed by removing the custom guided-decoding conflict validator. The latest implementation does not raise this ValueError; it only forwards the constraint produced on the validated vLLM request.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,114 @@ class PreprocessResult:
...
+    structured_outputs = request.structured_outputs
+    if structured_outputs is None:
+        return None
+    if structured_outputs.grammar is not None:
+        return {"grammar": structured_outputs.grammar}
+    if structured_outputs.json is not None:
+        return {"json": structured_outputs.json}
+    if structured_outputs.json_object:
+        return {"json": {"type": "object"}}
+    return None
```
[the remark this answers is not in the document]
indrajit96: The `_resolve_guided_decoding` precedence path referenced here has been removed.  The fallback now intentionally treats structured_outputs on the validated vLLM request as an opaque vLLM constraint, regardless of whether it originated from the client or ToolParser.adjust_request(), and forwards it without Dynamo-specific provenance or precedence policy.

components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -168,6 +185,70 @@ def _build_reasoning_parser_metadata(
...
+        tool_choice is not None and not isinstance(tool_choice, str)
+    )
+
+
+def _resolve_guided_decoding(
+    request: dict[str, Any],
+    tool_guided_decoding: dict[str, Any] | None,
+) -> dict[str, Any] | None:
+    assistant_guided_decoding = _build_guided_decoding(request)
+    if tool_guided_decoding is None:
```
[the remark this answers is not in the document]
indrajit96: Addressed by scope reduction. `_resolve_guided_decoding` and Dynamo’s explicit-guidance conflict checks have been removed. The processor now forwards only the constraint present on the validated vLLM preprocessing result.

components/src/dynamo/frontend/vllm_processor.py
```diff
@@ -560,6 +651,18 @@ async def _generator_inner(
             request_for_sampling,
             tokens,
         )
+        if reasoning_ended is False and _is_forced_tool_choice(
```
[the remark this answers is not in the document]
indrajit96: Addressed by removing the reasoning-aware rebuild entirely. The latest implementation does not maintain separate forced/auto reasoning branches; it forwards the configured vLLM parser’s constraint directly.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,114 @@ class PreprocessResult:
...
+    request: ChatCompletionRequest,
+    *,
+    structural_tag_mode: str,
+    structural_tag_scope: str,
+) -> bool:
+    tool_choice = request.tool_choice or "auto"
+    if structural_tag_mode != "on" or not request.tools:
+        return False
+
+    if tool_choice == "none":
```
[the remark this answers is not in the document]
indrajit96: Keeping this out of scope intentionally. _prepare_request does not activate the tool parser for tool_choice="none", so there is no parser-generated constraint to forward. Synthesizing a Dynamo-specific tool-call ban would add new policy beyond this bug; the existing exclude_tools_when_tool_choice_none behavior is unchanged.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,112 @@ class PreprocessResult:
...
+
+
+def _should_build_tool_call_guidance(
+    request: ChatCompletionRequest,
+    *,
+    structural_tag_mode: str,
+    structural_tag_scope: str,
+) -> bool:
+    tool_choice = request.tool_choice or "auto"
+    if structural_tag_mode != "on" or not request.tools:
```
keivenchang: Hey Indrajit, I think this gate leaves the default config still shipping the bug. `structural_tag_mode` defaults to `"off"` (`common/backend/worker.py:136`), so `tool_choice="required"` and named tool choice get no constraint out of the box.

Neither sibling gates the forced case on the mode — Rust falls through to `get_json_schema_from_tools` (`tool_choice.rs:107`), SGLang builds `get_json_schema_constraint` unconditionally (`sglang_prepost.py:580`). As written, the repro in your description would still return unconstrained text on a default deployment.

I think you can either add that fallback outside this check, or say in the description that it's opt-in.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,112 @@ class PreprocessResult:
...
+                    }
+                )
+            }
+        )
+        for tool in request.tools or []
+    ]
+    return request.model_copy(update={"tools": tools})
+
+
+def build_tool_call_guided_decoding(
```
keivenchang, line 187: I think this is a third copy of the policy, not a new one — `sglang_prepost.py:562` already has a `build_tool_call_guided_decoding` with this name and role, plus `_is_named_tool_choice` (`:321`), and `tool_choice.rs` + `structural_tag.rs` is the third.

Worth noting every bot finding here is one copy drifting from another: `parallel_tool_calls` defaults, the `tool_choice="none"` ban, the reasoning flag. I think you can either route the decision through one shared helper, or add a parity test pinning vLLM/SGLang/Rust to the same verdict across the matrix you already wrote.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,112 @@ class PreprocessResult:
...
+                structural_tag.model_dump()
+                if hasattr(structural_tag, "model_dump")
+                else structural_tag
+            )
+            return {"structural_tag": tag_value}
+
+    # Some vLLM parsers build a constraint in adjust_request() rather than
+    # implementing get_structural_tag(). MistralToolParser, for example,
+    # stores a parser-native Lark grammar on the validated request.
+    structured_outputs = request.structured_outputs
```
keivenchang: This fallback got riskier after the force-push. `structured_outputs` is client-settable on vLLM's `ChatCompletionRequest`, so a user-supplied constraint comes back as *tool* guidance — Devin's point above, but `_resolve_guided_decoding` used to sit downstream. With it gone, the value goes straight into `sampling_options.guided_decoding` unchecked. Snapshotting before `adjust_request()` and using only the delta would pin the actual invariant.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +42,112 @@ class PreprocessResult:
...
+    *,
+    structural_tag_schema: str,
+) -> ChatCompletionRequest:
+    strict_schema = structural_tag_schema == "strict"
+    tools = [
+        tool.model_copy(
+            update={
+                "function": tool.function.model_copy(
+                    update={
+                        "strict": strict_schema or tool.function.strict is True,
```
keivenchang: I'm not sure about this one, but it could change auto-mode behaviour. With `structural_tag_schema="auto"`, `strict_schema` is `False`, so this rewrites an unset `strict=None` to `False` before vLLM's parser sees it. Rust keeps it an `Option` and passes `schema_mode` separately (`structural_tag.rs:86`). If `get_structural_tag` distinguishes unset from explicit-false, this changes the tag for every tool omitting `strict` — and the new tests only cover tools that set it.

keivenchang (commented):
hey Indrajit, thanks so much for this — just reviewed it. Note I started against `ba62f13` and you force-pushed `3482d6f` mid-review, so this is verified against the new head.

Two things about that force-push. The description still says the change "preserves response-format precedence and reasoning-aware forced tool calls" — both paths (`_resolve_guided_decoding`, `starts_in_reasoning`) were deleted in `3482d6f`. And dropping `_resolve_guided_decoding` drops the rule Rust enforces at `tool_choice.rs:57-61`: for non-forced `tool_choice`, an explicit `response_format` or `guided_*` suppresses the tool tag. Nothing implements that now, so `structural_tag_mode=on` + tools + `response_format={"type":"json_object"}` emits a tool tag where Rust emits none.

Rest inline.

components/src/dynamo/frontend/prepost.py
indrajit96: Agreed, fixed. 
Forced required and named tool choices are now handled before the `structural_tag_mode` gate. 
When structural tags are off, they fall back to vLLM’s get_json_schema_from_tools, matching the SGLang behavior. Automatic tool guidance remains controlled by the structural-tag mode and scope.
Automatic tool guidance remains controlled by the structural-tag mode and scope.
`test_forced_choice_uses_json_guidance_when_mode_off` covers both required and named choices.

components/src/dynamo/frontend/prepost.py
indrajit96: Agreed on pinning the behavioral contract. 
I kept the backend-specific builders because their output representations differ, but added a shared parity matrix in `_tool_guidance_parity.py` that is consumed by both the vLLM and SGLang processor tests.
It covers 
1. no tools
2. tool_choice="none"
3. required
4. named choice
5. automatic choice, and explicit response-format precedence. 

The expected decisions follow the existing Rust behavior, while allowing each backend to emit JSON schema, grammar, or structural-tag guidance. 
Targeted tests separately cover omitted `parallel_tool_call`s and schema-mode behavior.
 The shared SGLang matrix passed 6/6, the complete SGLang processor suite passed 200/200, and the complete vLLM processor suite passed 81/81.

components/src/dynamo/frontend/prepost.py
indrajit96: Fixed using a before/after snapshot around `ToolParser.adjust_request()`. 
Client structured-output guidance is copied before adjustment, and the adjusted value is treated as parser-generated tool guidance only when it differs from that snapshot.
This prevents client-provided structured_outputs from being reclassified and forwarded as tool guidance, while preserving parser-generated constraints such as Mistral’s grammar. 
Covered by `test_forced_choice_does_not_reuse_client_structured_outputs`and `test_parser_adjust_request_generated_grammar_is_forwarded`

components/src/dynamo/frontend/prepost.py
indrajit96: Good point, fixed. 
In `structural_tag_schema="auto"` mode, the original `tool.function.strict` value is preserved exactly, including None. Only "strict" schema mode overrides it to True.
`test_auto_schema_preserves_unset_strict` covers the omitted value, and `test_vllm_parser_receives_schema_mode` covers mixed explicit True/False tools.

indrajit96:
> hey Indrajit, thanks so much for this — just reviewed it. Note I started against `ba62f13` and you force-pushed `3482d6f` mid-review, so this is verified against the new head.
> 
> Two things about that force-push. The description still says the change "preserves response-format precedence and reasoning-aware forced tool calls" — both paths (`_resolve_guided_decoding`, `starts_in_reasoning`) were deleted in `3482d6f`. And dropping `_resolve_guided_decoding` drops the rule Rust enforces at `tool_choice.rs:57-61`: for non-forced `tool_choice`, an explicit `response_format` or `guided_*` suppresses the tool tag. Nothing implements that now, so `structural_tag_mode=on` + tools + `response_format={"type":"json_object"}` emits a tool tag where Rust emits none.
> 
> Rest inline.

Thanks a lot for your review @keivenchang apprecite it!
I restored response-format precedence without bringing back the broader resolver. 
Assistant guidance is captured before the tool parser mutates the request, tool guidance is built separately, and explicit response_format/guided_* guidance wins for non-forced tool choices. 
Forced required or named choices still use the tool constraint.
I added coverage for both `response_format={"type":"json_object"}` and legacy `guided_regex`, plus the shared response-format precedence case. 
I also removed the stale reasoning-aware claim from the PR description and updated it with latest changes.

indrajit96:
/ok to test [740f3a4](https://github.com/ai-dynamo/dynamo/pull/12684/commits/740f3a4aa5a695e1459e7ecfb18df5dba74a6042)

components/src/dynamo/frontend/prepost.py
```diff
@@ -215,15 +388,26 @@ async def preprocess_chat_request(
...
+    is_forced_tool_choice = _is_forced_tool_choice(validated_request.tool_choice)
     (
         request_for_sampling,
         tool_parser,
         chat_template_kwargs,
         messages,
         chat_params,
     ) = _prepare_request(
-        request,
+        validated_request,
```
rmccorm4: **[P1] Preserve raw request controls during validation**

Passing `validated_request` here makes `_prepare_request()` receive a model instead of the original dictionary, so its dictionary-only handling for Rust-pythonized `chat_template_args` and root `thinking` is unreachable. Local composition probes confirmed that template arguments are dropped and deployment thinking defaults can override an explicit root control.

Please keep `request` as this argument and add an optional already-validated model to `_prepare_request()` (for example, `validated_request: ChatCompletionRequest | None = None`). `_prepare_request()` can use that model for semantic fields while still reading transport-only aliases from the raw dictionary, avoiding duplicate validation. Add regression tests through `preprocess_chat_request()` for both raw fields; direct helper tests do not cover this boundary.

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +43,172 @@ class PreprocessResult:
...
+                if hasattr(structural_tag, "model_dump")
+                else structural_tag
+            )
+            return {"structural_tag": tag_value}
+
+    if parser_guided_decoding is not None:
+        return parser_guided_decoding
+
+    tool_choice = request.tool_choice or "auto"
+    if _is_forced_tool_choice(tool_choice):
```
rmccorm4, line 222: **[P1] Respect parsers that require native tool syntax**

This generic fallback ignores `ToolParser.supports_required_and_named=False`. Gemma4Engine, PoolsideV1, Rust, and other parsers deliberately leave `structured_outputs` unset for required/named choices so the model can emit its native wire format. vLLM documents that forcing JSON for Gemma4 conflicts with that syntax and can crash EngineCore under speculative decoding; local probes with the real Gemma4Engine and PoolsideV1 parsers showed Dynamo adding JSON after both deliberately returned no constraint.

After honoring structural tags and `parser_guided_decoding`, return `None` when `tool_parser is not None and not tool_parser.supports_required_and_named`; only then apply the generic JSON fallback. Add a focused test with a parser whose capability is false and a required choice, while retaining the supported-parser fallback test. The `get_json_schema_from_tools` import should also move to module scope.

rmccorm4 (commented):
Two blocking findings reproduced locally against this exact head.

components/src/dynamo/frontend/prepost.py
keivenchang: ok this is handled — forced tool choice now returns before the mode gate, and the `get_json_schema_from_tools` fallback covers the default config. thanks for taking that one.

components/src/dynamo/frontend/prepost.py
keivenchang: gotcha, the shared parity table does it — both backends read the same case list now.

components/src/dynamo/frontend/prepost.py
keivenchang: yep, deep-copying the client guidance before `adjust_request()` and diffing is exactly the invariant i was after.

components/src/dynamo/frontend/prepost.py
keivenchang: i see you kept unset as unset here — that was the bit i was actually worried about.

components/src/dynamo/frontend/prepost.py
```diff
@@ -232,6 +416,29 @@ async def preprocess_chat_request(
...
+        request_for_sampling,
+        tool_parser,
+        parser_guided_decoding=parser_guided_decoding,
+        structural_tag_mode=structural_tag_mode,
+        structural_tag_scope=structural_tag_scope,
+        structural_tag_schema=structural_tag_schema,
+    )
+    guided_decoding = (
+        assistant_guided_decoding
+        if assistant_guided_decoding is not None and not is_forced_tool_choice
```
keivenchang, line 571: i think this quietly drops a user constraint where the Rust path returns a 400. with `tool_choice="required"` plus `guided_regex`, `is_forced_tool_choice` is true so this picks `tool_guided_decoding` and the regex just vanishes — no error, no annotation. `tool_choice.rs:48-53` raises `InvalidArgument` for exactly that combination, and Devin's earlier comment was about the *type* of that error, not about removing the check. how about restoring the rejection here so the two frontends agree on it being a client mistake?

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +43,172 @@ class PreprocessResult:
...
+        for key, value in (
+            ("json", request_extra.get("guided_json")),
+            ("regex", request_extra.get("guided_regex")),
+            ("grammar", request_extra.get("guided_grammar")),
+            ("choice", request_extra.get("guided_choice") or None),
+        )
+        if value is not None
+    }
+    if legacy_guidance:
+        guided_decoding = {**(guided_decoding or {}), **legacy_guidance}
```
keivenchang: this merge can emit two mutually exclusive constraints at once. i ran it against the PR's own code with `guided_json` + `guided_regex` both set:

```
RESULT: {'json': {'type': 'object'}, 'regex': '\\d+'}
```

`GuidedDecodingOptions::validate()` counts those and errors on `count > 1`, and `from_optional()` is what the Rust HTTP path builds through — so that request 400s there. this path constructs the dict directly and never hits `validate()`, so both keys go to the engine with undefined precedence.

note `_guided_decoding_from_structured_outputs` right above is careful to pick exactly one via the elif chain — the legacy merge is what breaks that invariant. you can either take the same first-wins approach for the legacy keys, or reject multiples the way Rust does.

components/src/dynamo/frontend/tests/_tool_guidance_parity.py
```diff
@@ -0,0 +1,77 @@
...
+class ToolGuidanceParityCase:
+    name: str
+    tool_choice: ToolChoice
+    has_tools: bool
+    has_assistant_constraint: bool
+    expected: GuidanceSource
+
+
+TOOL_GUIDANCE_PARITY_CASES = (
+    ToolGuidanceParityCase("no-tools", "auto", False, False, "none"),
```
keivenchang, line 35: nice that this is shared — one thing missing though: there's no case with a forced tool choice *and* an assistant constraint, which is the one combination the two backends are most likely to drift on. `scope`/`schema` arent dimensions here either, so `auto`+`strict` is pinned but `scope="always"` isnt.

keivenchang (commented):
hey Indrajit, thanks for the rework — just re-reviewed at `740f3a4`.

components/src/dynamo/frontend/prepost.py
krishung5: Fixed in 59dd109f.` _prepare_request` takes the raw request again, plus an optional validated_request: ChatCompletionRequest | None = None reused for semantic fields (no double validation). The dict-only reads for the pythonized chat_template_args alias and root thinking now see the  original dict, so template args aren't dropped and an explicit root thinking isn't overridden by the deployment default. Regression tests through preprocess_chat_request(): test_chat_template_args_survive_preprocess and test_root_thinking_suppresses_deployment_default (both fail on  the prior head, pass now).

components/src/dynamo/frontend/prepost.py
krishung5: Fixed in 59dd109f. The forced-choice branch returns None for parsers that don't advertise the capability, before the JSON fallback:
if tool_parser is not None and not getattr(tool_parser, "supports_required_and_named", True):
    return None

getattr(..., True) keeps real parsers (base default True) unchanged while honoring native-syntax parsers like Gemma4Engine/PoolsideV1 that set it False. get_json_schema_from_tools moved to module scope. Test: test_forced_choice_skips_json_for_native_syntax_parser (required + named). (Module-scope import also required adding vllm.tool_parsers.utils to the pytest-marker-report stub allowlist — ad049059.)

components/src/dynamo/frontend/prepost.py
krishung5: Restored the rejection. A forced tool choice combined with an explicit client constraint (response_format / guided_* / client structured_outputs) now raises InvalidArgument (→400) instead of silently dropping it, matching tool_choice.rs. It keys off the client constraint captured before adjust_request(), so a parser-generated structured_outputs isn't falsely rejected (test_parser_adjust_request_generated_grammar_is_forwarded still forwards it).

The raise propagates cleanly like the SGLang path's InvalidArgument. Tests: test_forced_choice_with_explicit_constraint_is_rejected; the old ..._does_not_reuse_client_structured_outputs is now ..._is_rejected.

components/src/dynamo/frontend/prepost.py
krishung5: Fixed in 2daae405. The legacy guided_* merge now picks exactly one constraint by precedence (json→regex→grammar→choice) rather than merging keys, matching the single-constraint invariant GuidedDecodingOptions::validate() enforces (and the elif chain above it).

Test: 
test_legacy_guided_fields_yield_single_constraint asserts guided_json + guided_regex → a single {"json": …}.

components/src/dynamo/frontend/tests/_tool_guidance_parity.py
krishung5: @keivenchang It seems like currently the two backends don't actually agree here yet.
Rust splits these: forced + guided_* → reject (tool_choice.rs:48-53, your guided_regex example), but forced + response_format → not rejected. It drops response_format ("applies to assistant content, not tool calls") and lets the tool constraint win.

So the forced + response_format case should pin to "tool", and two things need to line up first:
1. vLLM — narrow the rejection I added so it only fires on guided_*/structured_outputs, not response_format. As written it over-rejects forced + response_format (returns 400 where Rust returns tool-wins), so I need to correct that.
2. SGLang — today it lets response_format win over the tool for a forced choice (sglang_prepost.py:816-829); it'd need to switch to tool-wins to match Rust/vLLM.

I'm not super familiar with this so do you think it make sense to change SGLang's behavior here?

components/src/dynamo/frontend/prepost.py
keivenchang: ooops, correcting myself here. i offered first-wins as one of the two options and that was the wrong one — it silently drops a constraint the caller set. rejecting was the only correct one of the two.

the authority i cited isn't clean either, and my repro ran against an older head.

let me make some changes now.

components/src/dynamo/frontend/prepost.py
keivenchang: ok this is handled — the rejection is in at prepost.py:469 now. narrowing it so response_format falls through is the follow-on, tracked in the parity thread.

components/src/dynamo/frontend/prepost.py
keivenchang: got it — `_prepare_request` takes the raw dict again with the validated model as a kwarg, so the dict-only reads for `chat_template_args` and root `thinking` are live again. ran both regression tests here, passing. closing, thanks.

components/src/dynamo/frontend/prepost.py
keivenchang: i see you put the capability gate before the JSON fallback — that was the bit that mattered, native-syntax parsers now keep their own wire format. import is at module scope too. test passes here. closing, thanks.

components/src/dynamo/frontend/tests/_tool_guidance_parity.py
keivenchang: yes you're right, let's update sglang in another PR. here's the linear: [DIS-2643](https://linear.app/nvidia/issue/DIS-2643).

as for the vllm side, i pushed `d9b81dd`. the rule now: `response_format` describes how the answer to the *user* should look. a forced tool call isn't an answer to the user, so it doesn't apply — we drop it and keep the tool call. `guided_*` is different. it describes the raw text, which is the same space the tool call needs, so those two genuinely fight and it stays a 400.

vllm does the same thing internally. i tried all four combos on 0.26.0 and it accepted every one, just clearing `response_format`.

every combination and the reasoning behind it is in the PR description under "cases considered".

keivenchang:
confirmed this end to end on a real worker, not just unit tests.

same request, same worker, same seed — only the frontend code changed. a request where the prompt and `tool_choice` disagree, so the model has to be constrained to comply:

```json
{"messages": [{"role": "user", "content": "Do NOT call any function. Just reply with the single word: hello"}],
 "tools": [{"type": "function", "function": {"name": "get_weather", "parameters": {"type": "object",
   "properties": {"location": {"type": "string"}}, "required": ["location"]}}}],
 "tool_choice": "required", "temperature": 0, "seed": 42}
```

before (`9a7b29e`, pre-PR frontend) — model obeys the prompt and skips the tool:

```
content   : "<think>Okay, the user just said, Do NOT call any function..."
tool_calls: null
```

after (`d9b81dd`):

```
content   : null
tool_calls: [{"type": "function", "function": {"name": "get_weather",
             "arguments": "{\"location\": \"New York\"}"}}]
```

setup: Qwen3-0.6B, `--dyn-chat-processor vllm --tool-call-parser hermes`, worker with `--dyn-tool-call-parser hermes`. **default flags — no `--dyn-enable-structural-tag`**, so this is the JSON-schema fallback path for a forced choice.

three things worth knowing for anyone else reproducing this:

- a cooperative request proves nothing. my first attempt used "what's the weather in SF?" and it passed *before* the fix too — qwen emits hermes-format tool calls unprompted. you need a request where the model would not comply on its own.
- the tool parser has to be declared on the **worker** (`--dyn-tool-call-parser`), not just the frontend. without it the postprocessor has none and `tool_calls` stays null no matter what constraint is sent.
- `Ministral-3-3B` from the linear ticket can't be served on this path at all — it's `Mistral3ForConditionalGeneration` with a `vision_config`, and `chat_engine_factory` fails with `Failed to apply PixtralProcessor on data={'text': '[IMG]'}`. separate issue. mistral also has no structural-tag template (xgrammar only ships `hermes` and `minimax`), so on a mistral model the json fallback is the only route to a constraint.

keivenchang:
@coderabbitai review

@devin-ai-integration review

head is now `d9b81dd` — both of you last reviewed `ba62f13`, so the last two pushes haven't been looked at. the new commit narrows the forced-tool-choice conflict check so `response_format` is dropped rather than rejected, shape-validates the named-tool-choice predicate, and adds parity rows + tests.

rmccorm4 (approved):

components/src/dynamo/frontend/prepost.py
```diff
@@ -42,12 +49,264 @@ class PreprocessResult:
...
+        # forcing a JSON schema conflicts with their wire format and can crash
+        # EngineCore under speculative decoding. Only fall back for parsers that
+        # advertise JSON support (or when no parser is active).
+        if tool_parser is not None and not getattr(
+            tool_parser, "supports_required_and_named", True
+        ):
+            return None
+        json_schema = get_json_schema_from_tools(tool_choice, request.tools)
+        if json_schema is not None:
+            return {"json": json_schema}
```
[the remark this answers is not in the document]
keivenchang: good catch, this is real — verified it against vllm 0.26.0:

```
typed tool_choice -> ChatCompletionNamedToolChoiceParam => schema built
dict  tool_choice -> dict                               => None
```

and end to end through our path: `_is_named_tool_choice` correctly says yes to the dict, so we take the forced branch, and then `get_json_schema_from_tools` returns `None` — a named forced call ends up with no constraint at all, which is exactly the bug this PR exists to fix, reintroduced via the fast path.

this landed after the merge, so the fix is in a follow-up: #12878. it normalizes before the call and adds a regression test that fails without it. sglang hits the same trap and converts for the same reason (`sglang_prepost.py`, `get_json_schema_constraint` branches on `isinstance(tool_choice, ToolChoice)`).

@coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

PR #12687, opened 2026-08-05
fix(preprocessor): bridge thinking_mode for MiniMax M3

ritazh:
## Problem
MiniMax M3's chat template defaults `thinking_mode` to `"adaptive"`, which is incompatible with a schema-constrained (`response_format`) or forced-tool (`tool_choice=required` / named) generation start-state: the model returns `content: null, reasoning: null, finish_reason: stop`. `normalize_thinking_arg` also never read or wrote `thinking_mode`, so a client using `enable_thinking:true` never reached M3's `thinking_mode`-only template, and vice versa.

OpenRouter baseline probe uses `enable_thinking:true`, which M3's template ignores. The alias-loop bridge maps it to `thinking_mode:"enabled"` before rendering.

This PR implements the M3 half of #11554 which shipped M2 parity and its own docs table calls out `thinking_mode=disabled` as a required client setting for M3 tool+reasoning. This turns that manual workaround into automatic behavior.

## Fix

`normalize_thinking_arg`:
  - Adds `thinking_mode` to the alias-lookup loop; `"adaptive"` short-circuits
    to `None` (tri-state must not collapse to a boolean).
  - Writes `thinking_mode ∈ {enabled, disabled}` alongside the boolean aliases
    when the client set one and did not supply `thinking_mode` — bridges
    every template convention. Matches the writeback shape already used by
    `apply_default_thinking_mode_from_runtime_config` and
    `normalize_reasoning_template_args`.
  - When the client set no thinking control AND `reasoning_parser` is
    `minimax_m3` AND (`response_format` present OR `tool_choice` forces a
    call), defaults `thinking_mode` to `"disabled"`.
  - Extracts `has_guided_tool_choice`, replacing the duplicated inline check
    in `guided_output_requires_reasoning`.

## Related Issues

- Closes #XXXX

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12687" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MiniMax M3 compatibility with structured outputs and forced tool calls.
  * Correctly preserves adaptive reasoning settings when explicitly selected.
  * Improved consistency between reasoning controls across supported request formats.
  * Automatically disables reasoning when required for schema-constrained or forced tool generation.


rmccorm4:
/ok to test [3949370](https://github.com/ai-dynamo/dynamo/pull/12687/commits/3949370d63e13cdb262e1c9c980640ad94f1dcb0)

lib/llm/src/preprocessor.rs
```diff
@@ -5746,6 +5876,371 @@ mod tests {
...
+        assert_eq!(
+            args.get("thinking_mode"),
+            Some(&serde_json::json!("adaptive"))
+        );
+        assert!(!args.contains_key("thinking"));
+        assert!(!args.contains_key("enable_thinking"));
+    }
+
+    #[test]
+    fn test_normalize_thinking_arg_operator_default_bypasses_m3_override() {
```
indrajit96: The precendce is always given to client request.
This test suggests that operator setting is given preference over a client request of structured tool call.
This test preserves `thinking_mode="enabled"` for constrained M3 client requests, even though enabled/adaptive thinking is incompatible with JSON-schema and forced-tool generation. 

Suggested fix: track whether the thinking setting came from the client before applying deployment defaults. For constrained M3 requests, force "disabled" unless the client explicitly opted in. 
Update this test to expect "disabled" and reject explicit incompatible settings if enabled thinking cannot be supported.

lib/llm/src/preprocessor.rs
```diff
@@ -798,6 +828,48 @@ impl OpenAIPreprocessor {
                 matches!(reasoning_parser, Some("kimi_k25" | "kimi_k3" | "kimi-k3")).then_some(true)
             });
 
+        // MiniMax M3's chat template defaults `thinking_mode` to "adaptive",
```
indrajit96: Could we extract the MiniMax M3 behavior from `normalize_thinking_arg` into a dedicated function, such as `normalize_minimax_m3_thinking_mode`?
`normalize_thinking_arg` is otherwise a generic compatibility layer for normalizing the public aliases (thinking, enable_thinking, and thinking_mode). 
This block adds several pieces of model-specific policy directly to that generic flow:
 
 Could the flow instead be structured along these lines?
 ```
 fn normalize_thinking_arg(...) {
    let normalized = Self::normalize_thinking_aliases(request, reasoning_parser);
  
       if Self::is_minimax_m3_family(reasoning_parser, tool_call_parser) {
           Self::normalize_minimax_m3_thinking_mode(request, normalized);
           return;
       }
  
       Self::apply_normalized_thinking_aliases(request, normalized);
   }
 ```
 The M3 helper could own the explicit/adaptive detection, the constrained-request override, and thinking_mode canonicalization. 
It would also follow the existing pattern used by `normalize_kimi_k3_named_tool_choice`

indrajit96 (commented):
@ritazh Thanks a lot for adding this!
A couple of comments otherwise looks good.
Were you able to run end to end requests with M3 with this change?
If yes it would be great if you share response before the PR and response after.

ritazh:
Sample request/output below. @indrajit96 Please let me know what you think.

Scenario A: response_format: json_schema with no thinking control

```log
  POST /v1/chat/completions
  {
    "model": "MiniMaxAI/MiniMax-M3",
    "messages": [{"role": "user", "content": "Return a JSON object with a person named Alice, age 30."}],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "person",
        "schema": {
          "type": "object",
          "properties": {"name": {"type": "string"}, "age": {"type": "integer"}},
          "required": ["name", "age"],
          "additionalProperties": false
        }
      }
    },
    "max_tokens": 100
  }
```

Before the PR, template renders with M3's default `thinking_mode="adaptive"`. Reasoning opener + schema-FSM race -> empty output.

```json
  {
    "choices": [{
      "message": {"content": null, "role": "assistant", "reasoning": null},
      "finish_reason": "stop"
    }],
    "usage": {"prompt_tokens": 190, "completion_tokens": 10, "total_tokens": 200}
  }
```

After the PR, `normalize_thinking_arg`'s M3 default block sets `chat_template_kwargs.thinking_mode="disabled"`. Template renders without the reasoning opener; FSM constrains from token 1.

```json
  {
    "choices": [{
      "message": {"content": "{\"name\": \"Alice\", \"age\": 30}", "role": "assistant", "reasoning": null},
      "finish_reason": "stop"
    }],
    "usage": {"prompt_tokens": 177, "completion_tokens": 13, "total_tokens": 190}
  }
```

The prompt_tokens drop from 190 -> 177 (the 13-token reasoning opener is gone). content is now valid schema-matching JSON.

  ---
Scenario B: tool_choice: "required" with no thinking control

Request:

```log
  POST /v1/chat/completions
  {
    "model": "MiniMaxAI/MiniMax-M3",
    "messages": [{"role": "user", "content": "What is the weather in San Francisco?"}],
    "tools": [{"type": "function", "function": {"name": "get_weather", "parameters": {"type": "object", "properties": {"city":
  {"type": "string"}}, "required": ["city"]}}}],
    "tool_choice": "required"
  }
```
Before the PR

```json
  {
    "choices": [{
      "message": {"content": null, "role": "assistant", "reasoning": null, "tool_calls": null},
      "finish_reason": "stop"
    }],
    "usage": {"completion_tokens": 23}
  }
```
tool_choice=required was set but no tool call emitted. finish_reason:"stop" instead of "tool_calls".

After the PR

```json
  {
    "choices": [{
      "message": {
        "content": null,
        "role": "assistant",
        "tool_calls": [{
          "id": "call-5ce06d28-b1bb-474b-8010-a2a29d47a352",
          "type": "function",
          "function": {"name": "get_weather", "arguments": "{\"city\": \"string\"}"}
        }]
      },
      "finish_reason": "tool_calls"
    }],
    "usage": {"completion_tokens": 31}
  }
```

Proper tool call emitted, finish_reason is now "tool_calls".

lib/llm/src/preprocessor.rs
indrajit96: @ritazh Can you have a look at this too?

lib/llm/src/preprocessor.rs
indrajit96: @ritazh Can you have a look at this too?

ritazh:
@indrajit96 I've addressed your comments. PTAL when you get a chance

indrajit96:
/ok to test [90f52fb](https://github.com/ai-dynamo/dynamo/pull/12687/commits/90f52fbeeafb0178acd367c020676846c4766531)

indrajit96 (approved):
LGTM!
Thanks a lot @ritazh for adding this.

PR #12688, opened 2026-08-05
fix(frontend): preserve split UTF-8 characters

xianlubird:
## Summary

The SGLang frontend previously restarted incremental decoding from a fixed token lookback. With byte-fallback tokenizers, that lookback could begin inside a multi-token UTF-8 character, causing replacement characters (`U+FFFD`) to be emitted as output.

This change:

- tracks committed decode context separately from pending generated token IDs
- retains the complete pending region when decoding produces no text or ends in `U+FFFD`
- commits pending IDs only after reaching a clean decode boundary
- initializes decode context from the prompt tail to preserve tokenizer cleanup behavior
- flushes pending output when generation finishes, including a legitimate terminal `U+FFFD`
- passes prompt token IDs through both inline and preprocessing-pool paths
- continues to use the tokenizer's standard `decode()` API without requiring `DecodeStream`

## Validation

- `pytest components/src/dynamo/frontend/tests/test_sglang_processor_unit.py` — 197 passed
- verified randomized prompt/output chunking against full decoding for Qwen3, GPT-2, and TinyLlama — 300 cases per tokenizer
- added a real TinyLlama byte-fallback regression covering the eight-plus-one token split for `🙂🙂`
- added coverage for short split-byte sequences and legitimate terminal `U+FFFD`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming text output for multibyte UTF-8 characters split across token chunks.
  * Prevented incomplete replacement characters from appearing prematurely in intermediate responses.
  * Preserved valid replacement characters when generation completes, including responses without final token IDs.

* **Tests**
  * Added coverage for split UTF-8 sequences and final replacement-character handling.

### DIFF
--- a/components/src/dynamo/frontend/sglang_prepost.py
+++ b/components/src/dynamo/frontend/sglang_prepost.py
@@ -935,16 +935,10 @@ class SglangStreamingPostProcessor:
     """Streaming post-processor using SGLang parsers and HF tokenizer detokenization.
 
     Handles:
-    - Incremental detokenization via sliding-window decode (6-token lookback)
+    - Incremental detokenization across tokenizer-safe boundaries
     - Reasoning content extraction via SGLang ReasoningParser
     - Tool call parsing via SGLang FunctionCallParser or JsonArrayParser
     """
-
-    # Lookback window size for incremental detokenization.  UTF-8 characters
-    # can span up to 4 bytes, each potentially its own token.  A lookback of
-    # 6 covers the worst case (4-token char) plus margin for BPE merges that
-    # cross the old/new boundary.
-    LOOKBACK = 6
 
     def __init__(
         self,
@@ -956,6 +950,7 @@ class SglangStreamingPostProcessor:
         sglang_tools: list[SglangTool] | None = None,
         tool_call_parser_name: str | None = None,
         eos_token_ids: list[int] | None = None,
+        prompt_token_ids: list[int] | None = None,
     ) -> None:
         self.tokenizer = tokenizer
         self.tool_call_parser = tool_call_parser
@@ -978,7 +973,12 @@ class SglangStreamingPostProcessor:
         )
         self._eos_token_ids = set(eos_token_ids or [])
 
-        self._all_token_ids: list[int] = []
+        # Keep a small, known-complete prompt suffix as decode context. Generated
+        # tokens are promoted to context only after they decode without a
+        # trailing replacement character, so the boundary never moves into an
+        # incomplete byte-fallback sequence.
+        self._decode_context_ids = list((prompt_token_ids or [])[-5:])
+        self._pending_decode_ids: list[int] = []
         # Tool call accumulation.  SGLang's streaming parser returns
         # deltas (name in one chunk, argument fragments across subsequent
         # chunks).  However, the base detector processes at most one event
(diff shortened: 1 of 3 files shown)


components/src/dynamo/frontend/sglang_prepost.py
```diff
@@ -1008,7 +1008,9 @@ def _tool_call_id(self, name: str, index: int) -> str:
             self.history_tool_calls_count,
         )
 
-    def _incremental_decode(self, new_token_ids: list[int]) -> str:
+    def _incremental_decode(
```
rmccorm4, line 1019: The U+FFFD handling improves short split-character streams, but the fixed six-token lookback remains unsafe. Once `start` lands inside a byte-tokenized character, `prefix_text` begins with replacement characters. Stripping its trailing U+FFFD characters then treats that decoding damage as pending output and emits it during the finish flush.

For example, TinyLlama encodes `🙂🙂` as nine IDs. Sending the first eight IDs and then the final ID currently produces seven U+FFFD characters with this patch.

Could we replace the moving lookback with a small safe-boundary state machine?

```python
# Initialized in __init__. The prompt tail provides tokenizer cleanup context;
# it is not used as a moving size limit.
self._decode_context_ids = list((prompt_token_ids or [])[-5:])
self._pending_decode_ids: list[int] = []

def _decode_ids(self, token_ids: list[int]) -> str:
    if not token_ids:
        return ""
    return self.tokenizer.decode(
        token_ids,
        skip_special_tokens=self._skip_special_tokens,
    )

def _incremental_decode(
    self,
    new_token_ids: list[int],
    *,
    flush: bool = False,
) -> str:
    self._pending_decode_ids.extend(new_token_ids)

    context_text = self._decode_ids(self._decode_context_ids)
    combined_text = self._decode_ids(
        self._decode_context_ids + self._pending_decode_ids
    )
    delta_text = combined_text[len(context_text):]

    # Empty output may also mean that the tokenizer is withholding an
    # incomplete sequence, so keep those IDs pending as well.
    if not flush and (
        not delta_text or delta_text.endswith("\ufffd")
    ):
        return ""

    # The pending region now ends at a clean decode boundary. Keep it as
    # shared context for the next pair of decodes.
    self._decode_context_ids = self._pending_decode_ids
    self._pending_decode_ids = []
    return delta_text
```

The call remains:

```python
delta_text = (
    self._incremental_decode(
        token_ids,
        flush=finish_reason is not None,
    )
    if token_ids or finish_reason is not None
    else ""
)
```

This approach:

- Never restarts decoding from an arbitrary byte-token boundary.
- Retains incomplete IDs until they decode cleanly.
- Flushes a legitimate terminal U+FFFD.
- Works through the tokenizer’s existing `decode()` API without requiring `DecodeStream` or backend-specific handling.
- Still performs two decoder calls per flush, as the current implementation does. Native SGLang uses the same surrounding/current decode pattern.
- Avoids SGLang’s additional `sent_offset` bookkeeping by emitting nothing from a batch whose tail is incomplete. The tradeoff is at most one additional chunk of latency for that batch.

The regression test should use a real byte-fallback tokenizer and a first batch longer than the old lookback, such as TinyLlama’s eight-plus-one split.

xianlubird: Implemented the safe-boundary state machine in `03ed922a47`. The post-processor now keeps a prompt-tail decode context and a pending generated-token region, and only promotes pending IDs after the combined decode ends cleanly. Both inline and preprocess-pool construction paths pass prompt token IDs, and finish detection uses `finish_reason is not None` so terminal U+FFFD is flushed. I also added the real TinyLlama `🙂🙂` eight-plus-one regression, retained the short split-byte and terminal U+FFFD coverage, and replaced the old lookback-size assertion with a context-state assertion. The TinyLlama tokenizer smoke reproduction and Black/isort/Ruff/py_compile checks pass locally.

rmccorm4 (approved):

PR #12689, opened 2026-08-05
fix(trtllm): preserve zero prompt logprobs

xianlubird:
## Summary

- Preserve `prompt_logprobs=0` when translating Dynamo output options into TensorRT-LLM sampling parameters.
- Add regression coverage that exercises the normal `generate_locally` path and verifies the exact sampling parameters passed to the engine.

## Background

`prompt_logprobs` is an optional integer with two distinct states that matter to the backend:

- `None` means prompt log probabilities were not requested.
- `0` means prompt log probabilities were requested for the actual prompt token only, without additional top-k alternatives.

The TensorRT-LLM handler read the option with `output_options.get("prompt_logprobs")`, but then guarded the assignment with a truthiness check:

```python
if prompt_logprobs_value:
    sampling_params.prompt_logprobs = int(prompt_logprobs_value)
```

Because zero is falsy in Python, an explicit `prompt_logprobs=0` followed the same path as an omitted value. The handler therefore left `SamplingParams.prompt_logprobs` at `None`, and the engine did not receive the caller's request for selected-token prompt log probabilities. Positive values were forwarded correctly, which made the problem specific to the zero-value mode.

## Change

The guard now checks for absence explicitly:

```python
if prompt_logprobs_value is not None:
    sampling_params.prompt_logprobs = int(prompt_logprobs_value)
```

This keeps the existing behavior for omitted values and positive top-k values while preserving zero as a meaningful request value. The change is limited to the TensorRT-LLM request translation path and does not alter defaults or response formatting.

The regression test sends a request with `output_options={"prompt_logprobs": 0}` through the existing mocked `generate_locally` flow. It then inspects the arguments passed to `generate_async` and verifies that the engine-facing sampling parameters contain `prompt_logprobs == 0`.

## Validation

- `python3 -m py_compile components/src/dynamo/trtllm/request_handlers/handler_base.py components/src/dynamo/trtllm/tests/test_trtllm_handler_base.py`
- `git diff --check`
- Targeted pytest was not run locally because the available environment lacks TensorRT-LLM/GPU support and has pytest 7.4.4 while the repository requires pytest 8 or newer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prompt log probability settings so an explicitly configured value of `0` is preserved and applied correctly.
* **Tests**
  * Added regression coverage to verify zero-valued prompt log probability options are forwarded correctly.

### DIFF
--- a/components/src/dynamo/trtllm/request_handlers/handler_base.py
+++ b/components/src/dynamo/trtllm/request_handlers/handler_base.py
@@ -1077,6 +1077,7 @@ class HandlerBase(BaseGenerativeHandler):
         # outputs are interleaved by choice index, so maintain one cursor per
         # choice and emit only the new slice for each Dynamo chunk.
         output_tokens_per_choice: dict[int, int] = {}
+        prompt_logprobs_payload = None
 
         sampling_params = self._override_sampling_params(
             self.default_sampling_params, request
@@ -1292,6 +1293,13 @@ class HandlerBase(BaseGenerativeHandler):
                         if top_logprobs:
                             out["top_logprobs"] = top_logprobs
 
+                        if prompt_logprobs_payload is None:
+                            prompt_logprobs_payload = (
+                                _shared_logprobs.extract_prompt_logprobs_from_completion_output(
+                                    output
+                                )
+                            )
+
                         if output.finish_reason:
                             out["finish_reason"] = output.finish_reason
                         if output.stop_reason:
@@ -1316,6 +1324,11 @@ class HandlerBase(BaseGenerativeHandler):
                                     "Request finished with no finish reason set - "
                                     "this indicates a possible bug"
                                 )
+
+                            if prompt_logprobs_payload is not None:
+                                out.setdefault("engine_data", {})[
+                                    "prompt_logprobs"
+                                ] = prompt_logprobs_payload
 
                             num_input_tokens = len(request.get("token_ids", []))
                             total_completion_tokens = sum(
--- a/components/src/dynamo/trtllm/request_handlers/handler_base.py
+++ b/components/src/dynamo/trtllm/request_handlers/handler_base.py
@@ -1294,10 +1294,8 @@ class HandlerBase(BaseGenerativeHandler):
                             out["top_logprobs"] = top_logprobs
 
                         if prompt_logprobs_payload is None:
-                            prompt_logprobs_payload = (
-                                _shared_logprobs.extract_prompt_logprobs_from_completion_output(
-                                    output
-                                )
+                            prompt_logprobs_payload = _shared_logprobs.extract_prompt_logprobs_from_completion_output(
+                                output
                             )
 
                         if output.finish_reason:
--- a/components/src/dynamo/trtllm/request_handlers/handler_base.py
+++ b/components/src/dynamo/trtllm/request_handlers/handler_base.py
@@ -1096,7 +1096,7 @@ class HandlerBase(BaseGenerativeHandler):
 
             # Handle prompt_logprobs
             prompt_logprobs_value = output_options.get("prompt_logprobs")
-            if prompt_logprobs_value:
+            if prompt_logprobs_value is not None:
                 if hasattr(sampling_params, "prompt_logprobs"):
                     setattr(
                         sampling_params, "prompt_logprobs", int(prompt_logprobs_value)
--- a/components/src/dynamo/trtllm/tests/test_trtllm_handler_base.py
+++ b/components/src/dynamo/trtllm/tests/test_trtllm_handler_base.py
@@ -56,6 +56,7 @@ class MockSamplingParams:
     best_of: int = 1
     ignore_eos: bool = False
     guided_decoding: object | None = None
+    prompt_logprobs: int | None = None
     max_tokens: int | None = None
     min_tokens: int | None = None
     stop_token_ids: list[int] | None = None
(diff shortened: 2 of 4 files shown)


components/src/dynamo/trtllm/request_handlers/handler_base.py
```diff
@@ -1096,7 +1096,7 @@ async def _generate_locally_impl(
 
             # Handle prompt_logprobs
             prompt_logprobs_value = output_options.get("prompt_logprobs")
-            if prompt_logprobs_value:
+            if prompt_logprobs_value is not None:
                 if hasattr(sampling_params, "prompt_logprobs"):
                     setattr(
                         sampling_params, "prompt_logprobs", int(prompt_logprobs_value)
```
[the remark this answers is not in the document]
xianlubird: Confirmed against the pinned TensorRT-LLM 1.3.0rc22 implementation: `SamplingParams(prompt_logprobs=0)` is accepted, and zero means returning only the actual prompt token log probability. I also addressed the response-path gap in fd24f8b88c. The TRT-LLM handler now captures `output.prompt_logprobs`, serializes it through the shared helper, and attaches it to the final chunk under `engine_data.prompt_logprobs`. The regression coverage now checks the real TensorRT-LLM `SamplingParams` contract as well as request forwarding and the returned payload.

components/src/dynamo/trtllm/request_handlers/handler_base.py
```diff
@@ -1292,6 +1293,11 @@ async def _generate_locally_impl(
                         if top_logprobs:
                             out["top_logprobs"] = top_logprobs
 
+                        if prompt_logprobs_payload is None:
```
tanmayv25, line 1296: 🤖 **AI-generated review comment** (automated multi-agent reviewer — please verify before acting)

**Major — spurious `engine_data.prompt_logprobs: []` attached on every TRT-LLM completion**

TRT-LLM's `CompletionOutput.prompt_logprobs` defaults to `[]` (`field(default_factory=list)` in `tensorrt_llm/executor/result.py`), unlike vLLM which defaults to `None`. The shared `extract_prompt_logprobs_from_completion_output` returns `None` **only** when the attribute *is* `None`, so for an unrequested request it returns `[]` here — not `None`.

Consequences:
- `prompt_logprobs_payload` latches `[]`, and the attach guard below (`if prompt_logprobs_payload is not None:`) then writes `out["engine_data"]["prompt_logprobs"] = []` on the final chunk of **every** TRT-LLM completion. TRT-LLM previously emitted no `engine_data` at all, so clients using `nvext extra_fields=["engine_data"]` now receive a spurious empty `prompt_logprobs` on every request.
- The once-only `is None` latch also grabs `[]` on the first chunk and never re-reads. If real prompt logprobs populate on a *later* streaming chunk they'd be missed — including the `prompt_logprobs=0` case this PR is meant to enable.

For comparison, the sibling vLLM handler guards on the **raw engine attribute** (`getattr(res, "prompt_logprobs", None) is not None`), which is safe only because vLLM's sentinel is `None`. This latch copied the idea but not the guard, so TRT-LLM's `[]` sentinel slips through.

Suggested fix (any one):
- Treat empty as absent: `if not prompt_logprobs_payload:` before re-extracting, and attach only when the payload is truthy; **or**
- Make `extract_prompt_logprobs_from_completion_output` return `None` for a falsy/empty input, unifying the vLLM (`None`) and TRT-LLM (`[]`) sentinels — cleanest, fixes all callers at once.

components/src/dynamo/trtllm/tests/test_trtllm_handler_base.py
```diff
@@ -775,13 +777,14 @@ def _make_handler(self) -> HandlerBase:
...
 
-    def _make_mock_generation_result(self):
+    def _make_mock_generation_result(self, prompt_logprobs=None):
         """Mock GenerationResult that yields a single finished token."""
         output = MagicMock()
         output.token_ids = [42]
         output.finish_reason = "stop"
         output.stop_reason = None
         output.request_perf_metrics = None
+        output.prompt_logprobs = prompt_logprobs
```
tanmayv25: 🤖 **AI-generated review comment** (automated multi-agent reviewer — please verify before acting)

**Minor — test mock uses vLLM semantics, hiding the empty-payload behavior**

`_make_mock_generation_result` defaults `prompt_logprobs=None`, so this line sets `output.prompt_logprobs = None` (vLLM's sentinel). The real TRT-LLM `CompletionOutput.prompt_logprobs` defaults to `[]`.

With `None`, every "not requested" test (health check, default-max-tokens, etc.) sees the extractor return `None` and nothing is attached — so the spurious `engine_data.prompt_logprobs: []` produced by the real `[]` default (see the comment on `handler_base.py`) is never exercised, and the behavior change ships with green tests.

Suggested: default the mock's `prompt_logprobs` to `[]` to match TRT-LLM's real `CompletionOutput`, and add a case asserting that **no** `engine_data` / `prompt_logprobs` is attached when the engine returns the empty-list default.

components/src/dynamo/trtllm/request_handlers/handler_base.py
xianlubird: Thanks for catching this. I confirmed that the pinned TensorRT-LLM CompletionOutput uses an empty list by default. In d8034d0af1, the shared extractor now treats both None and an empty list as absent. That keeps the latch open for later streaming chunks and avoids attaching an empty engine_data payload. I also added coverage for both the empty default and prompt logprobs arriving after an initially empty chunk.

components/src/dynamo/trtllm/tests/test_trtllm_handler_base.py
xianlubird: Updated in d8034d0af1. The generation-result mock now matches the TensorRT-LLM empty-list default, and there is a regression test confirming that an empty result does not add prompt_logprobs to engine_data.

tanmayv25 (approved):

PR #12690, opened 2026-08-05
refactor(tests): OPS-7980 make pytests agnostic of how Dynamo is deployed

dmitry-tokarev-nv:
## Summary

Refactors Dynamo's pytest suites so a functional test receives a **URL**, sends a
payload, and asserts on the response — and nothing else. How Dynamo got deployed
(a local subprocess in the same container, or a `DynamoGraphDeployment` on AKS)
becomes a fixture's job, handled in `tests/utils`.

Closes OPS-7980.

### The problem

Verification lived inside the local-process stack. `check_response` was a method
on `EngineProcess`, and `run_serve_deployment` inlined the whole
send/validate/retry loop next to port plumbing and subprocess launch. None of it
was reachable from a test whose Dynamo runs on a cluster, so `tests/deploy`
hand-rolled its own client and its own chat validator — twice, and the two had
already drifted (one required 100 characters of content, the other only
non-empty).

Every deploy test also opened with the same forty lines before its first
assertion: enter `ManagedDeployment`, list frontend pods, take `pods[0]`,
`port_forward`, check it for `None`, f-string a base URL, poll
`wait_for_model_availability`, assert the bool.

### What changed

**New: `tests/utils/inference_endpoint.py`** — `InferenceEndpoint(base_url, model,
headers)` is the whole handle a deployment-agnostic test needs. Works for
`http://localhost:8000`, a port-forwarded frontend, or `https://dynamo.example.com`.
`wait_until_serving()` becomes the single definition of *"deployed and able to
receive inference requests"*: one real chat completion, retried against one
monotonic deadline, raising with the last failure instead of returning a bool
that every caller re-asserts.

**New: `tests/utils/verification.py`** — `run_payloads()` and a free
`check_response()`, with no knowledge of processes or pods. Log assertions take
an explicit `log_source` (the existing `LogReadable` protocol) rather than a
process handle. `tests/serve` and `tests/deploy` now share this one
implementation.

**`ManagedDeployment`** gains `frontend_endpoint()` and `wait_until_serving()`,
so pod selection, the port-forward and URL construction live in utils. The new
`deployed_endpoint` fixture yields a ready URL and tears down after the test.
CI-cluster workarounds (the NIXL skip for vLLM disagg, the MIG
`gpu-memory-utilization` cap) move into the fixture too — they describe the
cluster, not the behaviour under test.

`test_deployment` is now:

```python
async def test_deployment(deployed_endpoint, deployment_target):
    payload = deployment_smoke_chat_payload(model=deployed_endpoint.model)
    run_payloads([payload.bind(deployed_endpoint)], log=logger)
```

Both hand-written validators are deleted. Assertion strength is **preserved, not
merged**: the smoke test keeps its 100-character floor and model/role checks; the
checkpoint test keeps its own prompt, token budget and non-empty-content bar.

**New marker `topology_dependent`** for tests that genuinely assert on *how*
Dynamo is deployed — worker identity, per-worker `/metrics`, log patterns, killed
pods, scaled components. For `tests/serve` it is **derived from each config's
payloads**, so it stays correct as configs change (30 of 146 cases qualify today);
`tests/fault_tolerance`, `tests/router` and `tests/mm_router` apply it
directory-wide.

**CI guard:** in-container jobs select purely by marker with no path argument, so
they collect the whole repo. Cluster tests are excluded today only as a side
effect of every expression carrying a `gpu_N`/framework term. `nightly-ci.yml`
already has a TODO about this, and 205 `k8s`-marked tests *do* carry `gpu_N` —
the protection is one marker edit from lapsing. `and not k8s` is now appended
once where the expression is built, not at 56 caller sites.

### Notable constraint this uncovered

The Kubernetes deploy job runs pytest on a **bare runner** that installs only
`container/deps/requirements.test.txt` — never the `ai-dynamo` wheel. A
module-scope `import dynamo` anywhere in the shared verification layer would make
the deploy tests uncollectable there. `payloads.py`'s `prometheus_names` import is
now lazy, and two tests enforce the rule (one static, one that imports the whole
chain in a subprocess with `dynamo` blocked).

### Scope

In: the seam, the deploy suite, the marker, the CI guard, docs, unit tests.

Deliberately out, and unchanged: migrating `tests/serve`'s ~130 `EngineConfig`s to
the endpoint model (they keep ports; only the shared runner underneath changed);
router and fault-tolerance suites, which need process/pod handles by construction;
multimodal and LoRA fixtures, which are host-bound because the *server* must
resolve a `localhost` pytest-httpserver URL or a `file://` path.

## Validation

Run on macOS/CPU against a locally built `ai-dynamo-runtime` (maturin), plus a
`git worktree` of `origin/main` as the baseline.

**Selection is provably unchanged.** Dumped `nodeid -> markers` for the whole
suite on `main` and on this branch:

- 0 node IDs removed, 0 markers lost.
- 65 node IDs added — 63 new unit tests, 2 from an unrelated untracked local file.
- Only marker delta: `topology_dependent` gained by 1523 intended items
  (`fault_tolerance` 1402, `router` 79, `serve` 30, `mm_router` 11, `deploy` 1).
- No CI expression references `topology_dependent`, so no job changes.

**`and not k8s` drops nothing.** Evaluated all 36 in-container marker expressions
from `pr.yaml`, `post-merge-ci.yml`, `nightly-ci.yml` and `xpu-ci.yaml` with and
without the guard, using pytest's own expression evaluator: **0 tests change
selection**.

**Deploy-job selections are identical:** `-m framework_only` → 33,
`-m framework_with_gaie` → 1, `-m dynamocheckpoint` → 1.

**Tests:** 63 new unit tests pass (~25 s). `pytest -m "pre_merge and parallel and
not (vllm or sglang or trtllm) and gpu_0"` on this branch vs `main`: same 3
failures and same 20 errors (all local-environment gaps — no wheels, no etcd/NATS,
no `aiconfigurator`), **+63 passes**. Full collection: 1961 tests, 1 error
(`torch` not installed on macOS).

`pre-commit run --all-files` passes. All commits GPG-signed and DCO signed off.

**One bug this validation caught:** the first draft of the directory conftests
marked *all 1896 tests* — pytest hands every conftest the whole session item list,
not just its own directory's. Fixed with an explicit path filter; the marker diff
is what surfaced it, since nothing would have failed.

**Not verified locally:** anything needing a GPU or a Kubernetes cluster — the
deploy tests themselves, and the serve E2E paths through the rewired
`run_serve_deployment`. Those need CI.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12690" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added deployment-agnostic inference testing with shared endpoints, readiness checks, payload builders, and response validation.
  * Added support for request headers, model verification, response-length checks, retries, and detailed failure diagnostics.
  * Added deployment smoke-test coverage and endpoint utilities for local and Kubernetes-based environments.

* **Documentation**
  * Expanded testing guidance for deployment-independent tests and topology-dependent scenarios.

* **Test Improvements**
  * Added classification and automatic handling for topology-dependent tests.
  * Added comprehensive coverage for endpoint readiness, payload validation, routing headers, and deployment behavior.


dmitry-tokarev-nv:
## CI results — run 1 (`1faba4c77f`)

**94 passed, 19 skipped, 4 failed.** All 4 failures were pre-existing on `main`.

### The refactor is validated on a real cluster

All 8 Kubernetes deploy tests pass through the new `deployed_endpoint` fixture and the shared `run_payloads` path:

| Job | Result |
|---|---|
| `vllm Deploy Test / agg`, `agg_router`, `disagg`, `disagg_router` | ✅ |
| `sglang Deploy Test / agg`, `agg_router` | ✅ |
| `trtllm Deploy Test / agg`, `agg_router` | ✅ |

From `vllm Deploy Test / agg`, the whole new path executing end to end:

```
inference_endpoint.py:232  Deployment at http://localhost:30931 is serving Qwen/Qwen3-0.6B (attempt 1)
verification.py:182        TESTING: Payload: ChatPayload
test_deploy.py:58          Deployment test PASSED for vllm-agg (... url: http://localhost:30931)
163.02s setup  |  0.84s call  |  4.85s teardown
```

That 163s-setup / 0.84s-call split is the point of the change: deployment mechanics are in the fixture, verification is the test.

The GPU serve jobs confirm the rewired `run_serve_deployment` and the CI guard:

```
Marker expression: ((pre_merge and sglang and gpu_1) and not profiled_vram_gib) and not k8s
4 passed, 3 skipped, 3252 deselected
```

`test_sglang_deployment[video_e_pd_qwen-4]` is a `topology_dependent` config — so a log/metrics-asserting payload went through the new `check_response(..., log_source=server_process)` path and passed. vLLM multi-GPU: 9 passed. TRT-LLM: 159 passed. **0 failures across every serve job.**

### The 4 failures

`{vLLM,SGLang,TRTLLM} DynamoCheckpoint Deploy Test` + the `deploy-status-check` that aggregates them, all with:

```
FileNotFoundError: .../examples/backends/vllm/deploy/v1beta1/agg.yaml
```

Identical failure on `main` (run 30976519632, job 92212852735, `managed_deployment.py:398`). Root cause: #12375 promoted the v1beta1 manifests a directory up and deleted `deploy/v1beta1/`, but `CHECKPOINT_BACKENDS` still names that path. It fails before the deployment is created, so it never reached any of my code.

Fixed in `cd47491a3c` — with `tests/deploy/test_deployment_manifests.py` so this class of breakage is caught in pre-merge instead of only from a cluster job.

## Follow-up commits since run 1

- `16acdeeb88` — `ci:` the `not k8s` guard now falls back to a bare `not k8s` when the caller's expression is empty (`run_gpu_tests` defaults to true while `gpu_test_markers` has no default, so `-m "() and not k8s"` was reachable and pytest rejects it).
- `cd47491a3c` — `fix(tests):` the checkpoint manifest paths above, plus offline regression tests.
- `b4b570ed89` — `test:` closes two holes self-review found in the `topology_dependent` derivation:
  1. `UuidPassthroughChatPayload` fills `expected_log` at run time, so collection-time derivation missed the vLLM `uuid_passthrough` config that greps the worker log. Payloads now answer `declares_log_assertions()`.
  2. Eleven `tests/serve` tests build their `EngineConfig` inline and never reach the derivation; eight are deployment-coupled (LoRA adapter POSTs to a worker system port, per-worker embedding counters, worker `/health` gating). They now carry the marker, and `run_serve_deployment` asserts the marker before launching anything so neither shape can drift again.

  Marker coverage in `tests/serve` goes 30 → 45. The collected-marker diff against `main` is unchanged in shape: no node IDs added or removed among existing tests, no markers lost, `topology_dependent` still the only delta.

dmitry-tokarev-nv:
## Merged current `main` (`7230e7c482`)

Branched off `addcbae786`, which was `main`'s HEAD at the time. `main` has since gained 3 commits; merged them in.

One of them collides with this branch: **#12698 landed the same DynamoCheckpoint manifest fix independently.** The semantic change is identical (drop the `v1beta1/` segment from all three backends), so I resolved the conflict by taking `main`'s side — this branch now adds **no** formatting churn to `CHECKPOINT_BACKENDS`.

What my `cd47491a3c` still contributes over #12698:

- `_checkpoint_manifest_path()` extracted, so the path is derived in one place.
- `tests/deploy/test_deployment_manifests.py` — builds all three checkpoint specs offline and asserts the manifest exists and declares the components the test scales and port-forwards to. #12698 fixed the instance; this makes the class of breakage visible in **pre-merge** instead of only from a cluster job, which is how it stayed red on `main` for a while.

The other two (`dynamo-protocols 5.1.0`, `nightly KV-approximation deploy`) touch no file this branch touches.

Re-verified post-merge: 79 new unit tests pass, full collection clean (1977 tests), and the deploy selections are unchanged — `framework_only` 33, `framework_with_gaie` 1, `dynamocheckpoint` 1.

dmitry-tokarev-nv:
## GPU validation of `eeb77cacfe` (GitHub Actions outage — run on RTX 6000 Ada instead)

Ran on a dev box (RTX 6000 Ada, 48 GB) against the `4a34a46a82-vllm-runtime-test` image, with this branch's worktree mounted over `/workspace` — valid because this PR changes only `tests/`, which come from the mount; the image supplies the engine stack.

| Stage | Selection | Result |
|---|---|---|
| Unit layer | the 4 new test files (endpoint, verification, marker derivation, manifest regression) | **79 passed** in 1.8s |
| Deployment-agnostic serve E2E | `test_serve_deployment[aggregated-2]` — engine launch → `wait_until_serving` → `run_payloads` | **1 passed** in 67s |
| Deployment-coupled serve E2E | `-m 'topology_dependent and gpu_1 and not multimodal and not lmcache'` | **4 passed**, 1 env-error, in 5m05s |

The one error is `test_lora_aggregated`: `MinIO is not available and Docker is not accessible` — MinIO is a CI sidecar service; there is no docker socket inside the nested test container. Environmental, not a test failure; the LoRA path was already green in CI run 1 (`sglang`/`vllm` deploy + serve jobs).

Combined with CI run 1 (94 passed; all 8 k8s deploy tests through the new fixture), both execution modes of the refactor are now validated on real hardware:
- **in-container**: local engine launch → the rewired `run_serve_deployment` → shared `run_payloads`, including `topology_dependent` configs exercising the `expected_log` / worker-`/metrics` seam;
- **Kubernetes**: `deployed_endpoint` fixture → port-forward → same `run_payloads`.

Will re-request `/ok to test` when Actions recovers.

dmitry-tokarev-nv:
## GPU validation, part 2: CI-equivalent suite green on `eeb77cacfe`

`FULL=1` run on the RTX 6000 Ada box — the exact selection the CI GPU stage runs:

```
-m 'pre_merge and vllm and gpu_1'  (plus the central 'and not k8s' guard)
26 passed, 3 skipped, 2003 deselected in 26m32s
```

26 is the same count CI run 1 collected for this expression, and the 3 skips are engine-availability skips (`tensorrt_llm` not in the vLLM image), identical to CI. Zero failures.

With this, every stage this PR touches has passed somewhere real: unit layer (79), k8s deploy fixture (8/8 in CI run 1), deployment-agnostic serve E2E, topology_dependent serve E2E, and now the full pre_merge vLLM GPU suite. Pending only: a formal green CI run on this head once GitHub Actions recovers.

dmitry-tokarev-nv:
## GPU validation, part 3: post-merge suite green on `eeb77cacfe`

Widened to `(pre_merge or post_merge) and vllm and gpu_1` (plus the central `and not k8s`), run detached on the RTX 6000 Ada box with a MinIO sidecar so the LoRA tests execute:

```
65 passed, 102 skipped, 1865 deselected in 3942.84s (1:05:42)   exit 0
```

**Zero failures, zero errors.** All 102 skips are pre-existing and unrelated to this PR:

| n | Reason | Provenance |
|---|---|---|
| 96 | `Flaky: 0% post-merge pass rate ... skipped wholesale until the underlying migration fault is owned and fixed` (all of `tests/fault_tolerance/migration/test_vllm.py`) | on `main` since #9341 (2026-05-09); this PR does not touch `tests/fault_tolerance/migration/` |
| 3 | `test_vllm_omni` — Qwen2.5-Omni-7B needs ~80 GB, Qwen-Image ~40 GB, and a vLLM-Omni `_bookkeeping_sync` signature skew | capacity/version guards in the test file |
| 1 | `vLLM engine core init fails on amd64 post-merge. OPS-4445` | pre-existing tracked issue |
| 2 | collection-time skips (`tensorrt_llm` not in the vLLM image) | framework-marker auto-skip in `tests/conftest.py` |

What actually exercised the refactored path: **36 `test_serve_deployment[...]` configs** — every multimodal variant (agg / b64 / frontend_decoding / video / NVDEC h264 + h265 / router / chat_processor / EPD / E_PD / P_D / uuid_passthrough / custom encoder, across qwen3-vl-2b, qwen2.5-vl-3b, qwen2-vl-2b, qwen3.5-0.8b, gemma4-e2b-it), plus `aggregated`, `aggregated_logprobs`, all three `lmcache` variants, `agg-request-plane-tcp`, three `disaggregated_same_gpu` variants, `completions_only`, `guided_decoding`, `embedding_agg` — and **`test_lora_aggregated` passed** (the MinIO error from part 1 was purely the missing sidecar). Also green: 2 `test_omni_serve_deployment`, the KVBM offload/eviction/determinism set, mm_router and router e2e, frontend pre/post integration, request-cancellation, RL worker discovery, and the self-benchmark.

Note the classification guard raises rather than skips, so 0 failures across 65 executed deployments means every exercised config's `topology_dependent` classification is consistent with its runtime behavior.

dmitry-tokarev-nv:
## Merged main (54 commits) — new head \`30c758a80e\`; and a correction to my earlier GPU-validation claims

### Merge

Main moved 54 commits during the Actions outage and split the module this PR builds on: `tests/utils/managed_deployment.py` → `tests/deploy/dgd_utils.py` + `dgdr_utils.py`, and `tests/deploy/test_deploy.py` → `test_dgd.py`. Four conflicts, all import-blocks, resolved as union + retarget:

| File | Resolution |
|---|---|
| `tests/conftest.py` | both sides added a marker via the *same* `addinivalue_line` call — split into two so `topology_dependent` **and** `token_budget_parity` are both registered |
| `tests/deploy/conftest.py` | keep `InferenceEndpoint`; source `DeploymentSpec`/`ManagedDeployment`/`_get_workspace_dir` from `dgd_utils`; keep main's `dgdr_utils` imports |
| `tests/deploy/test_dgd.py` | keep this PR's endpoint/verification imports, retarget to `dgd_utils`, drop the now-unused `tests.utils.client` import |
| `tests/deploy/test_dynamocheckpoint.py` | same |

This PR's 87-line `frontend_endpoint` addition survived main's rename into `dgd_utils.py` (verified). Two guard tests in this PR caught the rename by failing — `test_deploy.py` was hardcoded in the deploy-module list; updated to `test_dgd.py`. Doc/docstring references to the old module path updated too.

Verified after resolution: **79/79 unit tests pass**, `pre-commit run --all-files` clean (27 hooks), pyflakes clean across `tests/` (no undefined names, no unused imports — which is what confirms dropping the `tests.utils.client` import was correct), full collection unchanged, and no new unmarked inline-config tests introduced by main's 54 commits.

### Correction to the GPU-validation comments above

I ran an adversarial review of my own claims in the two comments above. Two were overstated and I want the record straight:

1. **"the exact selection the CI GPU stage runs" — not accurate.** CI's GPU stage is *two* pytest invocations with different marker expressions (a parallel `profiled_vram_gib` stage and a sequential `not profiled_vram_gib` one, per `shared-test.yml`). A single local 165-item selection isn't equivalent to that split. Concretely, `test_lora_aggregated_router` is `gpu_2` and runs in the sequential stage — my `gpu_1` filter never touched it.
2. **"the skip set is identical to CI's" — not provable.** Several skip gates inside that marker space are environment-dependent rather than repo-declared: the NVDEC `nvdec_available()` driver probe, `test_vllm_omni.py`'s module-level import guard, a `find_spec` framework auto-skip, and a `DYN_HF_GATED_MODELS_ENABLED` skipif. Worth noting this cut *in our favour* here — the `mm_agg_video_nvdec_*_h264/_h265` tests **passed** locally, whereas CI runners typically skip them for lacking the driver `video` capability.

What does hold up: 96 of the 102 skips are one unconditional `@pytest.mark.skip` on the flaky migration test (the full 96-way parametrization), and the remaining ones are static `pytest.mark.skip` entries in config `marks=[...]` lists — none of those are hardware probes. And the `-k 'not spec_decoding'` exclusion I applied cost **nothing**: that config is `nightly`-marked, so the `(pre_merge or post_merge)` expression had already deselected it.

Since the branch has now moved 54 commits, the GPU results above apply to the old head. Re-running the suite against `30c758a80e`.

dagil-nvidia (approved):
Approving: all repo gates green at head ad96ff9 (CODEOWNERS drift 0 across 620 rules, strict coverage, 93 matcher tests, skills validation), 79 new and changed unit tests pass in a head worktree, and the action rewiring is safe for every caller - exactly five call sites consume the changed actions, none passes a k8s-positive expression, and the cluster deploy job never routes through them, so `and not k8s` cannot deselect the deploy suite from its own job. The refactor reproduces the old payload loop exactly, and both policy docs describe code that actually exists (every named fixture, helper, and guard was located). Three non-blocking notes:

1. `tests/utils/inference_endpoint.py:116` - `DeploymentEndpoints` (and `.worker()`) has no producer and no consumer anywhere in the tree; only its own unit test touches it. Either drop it until the topology-dependent migration needs it, or add the `ManagedDeployment` producer here.
2. `tests/deploy/test_dynamocheckpoint.py:527` and `tests/deploy/test_dgd.py:239` - sync `wait_until_serving` (up to 600 s blocking) is called directly inside `async def` tests, blocking the loop that `kubernetes_asyncio` state lives on. `dgd_utils.py` wraps the identical probe in `asyncio.to_thread` with a comment explaining exactly this hazard; these two call sites bypass it. Not a regression, but the PR adds the rationale and applies it inconsistently.
3. `.github/actions/pytest/action.yml:139,248` and `pytest-local/action.yml:190` - the identical 15-line marker-composition block is pasted into three step scripts; a future policy change must hit all three. Compute it once per action and consume the step output.

Pre-existing, noted for follow-up: `xpu-ci.yaml` manual dispatch with empty `pytest_markers` composes `' and xpu_1'`, a pytest parse error (equally broken before this PR).

Interaction with #13845, since both touch the pytest CI surface: the two PRs conflict textually in `tests/README.md` (adjacent rows in the marker table) and likely in `.ai/pytest-guidelines.md`; the compositions are semantically compatible - #13845's selected expressions flow through `pytest-local`, so this PR's central `not k8s` guard wraps them correctly. Recommended order: land this PR first, so the central guard exists before #13845 starts narrowing expressions; two things then belong on the #13845 side - its shadow report models selection without the `not k8s` guard so it can list cluster tests as "selected" for in-container jobs, and its marker-inventory step installs a minimal pip set so the deploy/serve trees fail collection there and this PR's new markers would be missing from the shadow inventory.

PR #12704, opened 2026-08-05
fix(snapshot): preserve restore sentinel for independent readers

galletas1712:
## Summary

- keep `restore-complete` after the restored workload observes it so the independent kubelet startup probe can observe the same marker
- continue removing the capture-only `ready-for-snapshot` and `snapshot-complete` sentinels
- the snapshot agent still removes a leftover `restore-complete` before the next restore (now in [ai-dynamo/snapshot#72](https://github.com/ai-dynamo/snapshot/pull/72); originally the agent-side change on this PR)

The restored workload and kubelet startup probe are independent readers. The workload must not consume completion before kubelet observes it. The snapshot agent remains the owner that invalidates stale completion before a later re-restore.

This PR now carries the Dynamo-side change previously stacked as #13295. The `deploy/snapshot` agent cleanup no longer belongs here after #13177.

## Validation

- same change as #13295: lifecycle keeps `restore-complete` and the unit test expects the stamp to remain

### DIFF
--- a/components/src/dynamo/common/snapshot/lifecycle.py
+++ b/components/src/dynamo/common/snapshot/lifecycle.py
@@ -90,10 +90,11 @@ class SnapshotConfig:
             await asyncio.sleep(SENTINEL_POLL_INTERVAL_SEC)
 
     def _cleanup_ready_and_sentinels(self) -> None:
+        # Keep restore-complete for the kubelet startup probe; the snapshot agent
+        # removes it before restoring the next container incarnation.
         for name in (
             READY_FOR_SNAPSHOT_FILE,
             SNAPSHOT_COMPLETE_FILE,
-            RESTORE_COMPLETE_FILE,
         ):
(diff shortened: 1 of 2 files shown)


hhzhang16 (approved):

galletas1712:
/ok to test [f88964e](https://github.com/ai-dynamo/dynamo/commit/f88964eac3e5f719f09b79bb71741c4d9310324f)

galletas1712:
Removed the unrelated GitHub Actions bootstrap commit from this branch and rebased it onto current `main`. The PR is now lifecycle-only: `components/src/dynamo/common/snapshot/lifecycle.py` plus its focused test. The CI fix now lives in #13452, which should merge first; this branch can then rebase onto it before final CI/merge.

Validation: all applicable pre-commit hooks passed. The focused pytest could not be run locally because `pytest` is not installed in this workspace.

PR #12706, opened 2026-08-05
fix(frontend): preserve completion backend error status

NVShreyas:
## Overview:

Fix /v1/completions returning HTTP 500 when the backend reports a typed client error such as Backend(InvalidArgument).

  The non-streaming completions path now checks for backend errors before response aggregation, preserving the original HTTP 400 status and descriptive message. The existing timeout-aware error-checking helper is generalized to support completion and chat response types.

Adds regression coverage for the SGLang logprobs >= 1 case.

## Related Issues

- Closes #XXXX

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12706" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Typed backend validation errors from completion requests now return HTTP 400 responses with the appropriate error message.
  * Completion error metrics are now recorded consistently when backend errors occur.
* **Tests**
  * Added regression coverage to verify that typed completion errors are surfaced correctly.

### DIFF
--- a/lib/llm/src/http/service/openai.rs
+++ b/lib/llm/src/http/service/openai.rs
@@ -849,6 +849,17 @@ async fn completions_single(
 
         Ok(sse_stream.into_response())
     } else {
+        // Preserve typed backend errors before the completions aggregator turns
+        // them into strings. In particular, Python ValueError/TypeError arrives
+        // as Backend(InvalidArgument) and must remain an HTTP 400.
+        let stream = check_for_backend_error(stream, None)
+            .await
+            .map_err(|error_response| {
+                tracing::error!(request_id, "Backend error detected: {:?}", error_response);
+                inflight_guard.mark_error(extract_error_type_from_response(&error_response));
+                error_response
+            })?;
+
         // Tap the stream to collect metrics for non-streaming requests without altering items
         let mut http_queue_guard = Some(http_queue_guard);
         let stream = stream.inspect(move |response| {
@@ -965,6 +976,28 @@ fn aggregate_batch_completion_usage(
     }
 }
 
+type BoxedCompletionResponseStream =
+    std::pin::Pin<Box<dyn futures::Stream<Item = Annotated<NvCreateCompletionResponse>> + Send>>;
+
+/// Check each prompt stream before merging a non-streaming completion batch.
+///
+/// `select_all` cannot safely provide this check after merging because a normal
+/// event from one prompt may arrive before a typed backend error from another.
+/// Poll all streams concurrently so batch startup is not serialized.
+async fn check_completion_batch_streams<S>(
+    streams: Vec<S>,
+) -> Result<Vec<BoxedCompletionResponseStream>, ErrorResponse>
+where
+    S: futures::Stream<Item = Annotated<NvCreateCompletionResponse>> + Send + 'static,
+{
+    futures::future::try_join_all(
+        streams
+            .into_iter()
+            .map(|stream| check_for_backend_error(stream, None)),
+    )
+    .await
+}
+
 /// Handle batch prompt completions (multiple prompts with n choices each)
 #[tracing::instrument(skip_all)]
 async fn completions_batch(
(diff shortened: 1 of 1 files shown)


rmccorm4:
/ok to test [fa65616](https://github.com/ai-dynamo/dynamo/pull/12706/commits/fa65616474f70ff437a277dffffb4d7ab0012dce)

KrishnanPrash (approved):

lib/llm/src/http/service/openai.rs
```diff
@@ -965,6 +976,28 @@ fn aggregate_batch_completion_usage(
...
+/// `select_all` cannot safely provide this check after merging because a normal
+/// event from one prompt may arrive before a typed backend error from another.
+/// Poll all streams concurrently so batch startup is not serialized.
+async fn check_completion_batch_streams<S>(
+    streams: Vec<S>,
+) -> Result<Vec<BoxedCompletionResponseStream>, ErrorResponse>
+where
+    S: futures::Stream<Item = Annotated<NvCreateCompletionResponse>> + Send + 'static,
+{
+    futures::future::try_join_all(
```
rmccorm4, line 993: _Reviewed locally with Codex:_ try_join_all waits for every prompt stream to produce its first non-annotation frame before any buffered frame reaches select_all. If prompt A produces a token at 10 ms and prompt B at 500 ms, A's metrics aren't observed and http_queue_guard isn't dropped until roughly 500 ms, inflating batch TTFT and queue time to the slowest prompt. Could we preserve the typed error while consuming the merged stream instead of adding this barrier?

lib/llm/src/http/service/openai.rs
```diff
@@ -849,6 +849,17 @@ async fn completions_single(
 
         Ok(sse_stream.into_response())
     } else {
+        // Preserve typed backend errors before the completions aggregator turns
+        // them into strings. In particular, Python ValueError/TypeError arrives
+        // as Backend(InvalidArgument) and must remain an HTTP 400.
+        let stream = check_for_backend_error(stream, None)
```
rmccorm4, line 855: _Reviewed locally with Codex:_ This only checks the first non-annotation frame. If the backend emits a normal completion chunk and then Backend(InvalidArgument), the later error reaches the aggregator, is reduced to a string, and is returned through the generic 500 path. Non-streaming requests should preserve typed errors across the full aggregation.

lib/llm/src/http/service/openai.rs
```diff
@@ -1067,7 +1100,23 @@ async fn completions_batch(
         all_streams.push(remapped_stream);
     }
 
-    // Merge all streams
+    let all_streams: Vec<BoxedCompletionResponseStream> = if streaming {
```
rmccorm4, line 1103: _Reviewed locally with Codex:_ This boxes every prompt stream even when streaming is true, adding one allocation per prompt and dynamic dispatch on every poll of the unchanged streaming path. Can we keep these streams concrete and only box the merged stream if type erasure is needed?

lib/llm/src/http/service/openai.rs
```diff
@@ -5083,6 +5128,82 @@ mod tests {
...
+            error: Some(
+                DynamoError::builder()
+                    .error_type(ErrorType::Backend(BackendError::InvalidArgument))
+                    .message("Dynamo's SGLang backend does not currently support logprobs >= 1")
+                    .build(),
+            ),
+        };
+
+        let error_response =
+            match check_for_backend_error(stream::iter(vec![error_event]), None).await {
```
rmccorm4, line 5150: _Reviewed locally with Codex:_ This test exercises the helper directly, so it still passes if the production call in completions_single is removed; the batch test has the same gap. Please cover the handler/HTTP path so the regression is actually protected.

lib/llm/src/http/service/openai.rs
```diff
@@ -849,6 +849,17 @@ async fn completions_single(
 
         Ok(sse_stream.into_response())
     } else {
+        // Preserve typed backend errors before the completions aggregator turns
+        // them into strings. In particular, Python ValueError/TypeError arrives
+        // as Backend(InvalidArgument) and must remain an HTTP 400.
+        let stream = check_for_backend_error(stream, None)
+            .await
+            .map_err(|error_response| {
+                tracing::error!(request_id, "Backend error detected: {:?}", error_response);
```
rmccorm4, line 858: _Reviewed locally with Codex:_ This formats the response into the log message and logs an expected client-side 400 at error level. Could this use a structured error_response field and a lower level?

lib/llm/src/http/service/openai.rs
rmccorm4: > If prompt A produces a token at 10 ms and prompt B at 500 ms, A's metrics aren't observed and http_queue_guard isn't dropped until roughly 500 ms, inflating batch TTFT and queue time to the slowest prompt.

If this is correct, we need to fix this @KrishnanPrash

lib/llm/src/http/service/openai.rs
rmccorm4: This may impact perf at high concurrency, we should verify

PR #12710, opened 2026-08-05
fix(ci): add a watchdog so one wedged test can't stall the whole GPU job

nv-tusharma:
Linear: [OPS-8235](https://linear.app/nvidia/issue/OPS-8235/gpu-parallel-test-orchestrator-watchdog-so-one-wedged-test-cant-stall)

## Summary

The GPU-parallel orchestrator (`tests/utils/pytest_parallel_gpu.py`) delegates per-test timeouts entirely to the child, which is launched with `--timeout=<test.timeout>`. **Nothing sits behind that.** When pytest-timeout is swallowed by a C-level block, the child runs forever, the orchestrator polls it forever, and every queued test starves until the job's `timeout-minutes` kills the runner.

`start_time` was only ever used for the status display and for computing duration after a child exits — there was no deadline check anywhere in the scheduling loop.

### This is measured, not hypothetical

Post-merge CI, Jul 30 → Aug 5:

| | |
|---|---|
| GPU-test steps hitting the 120-min ceiling | **25 / 464 = 5.4%** |
| Share of all GPU-test-step failures | **25 / 95 = 26%** |
| GPU time burned | **50 GPU-hours in 6 days** |
| Concentrated in | `sglang-runtime / Test cuda13.0, amd64` — **100%** of them (28% of that job's runs) |

Every instance is the same shape: worker `w11` declaring `timeout=320s`, still alive at ~5570s, with `w13, w16, w17` queued behind it and never launched. Representative job: [30957901075](https://github.com/ai-dynamo/dynamo/actions/runs/30957901075/job/92161394649).

That test is currently skipped on `main` (`tests/router/test_router_e2e_with_sglang.py:406`, *"hangs to the job step cap instead of failing"*) — a manual workaround for this exact failure. This is the systemic guard so the next one doesn't cost 50 GPU-hours before someone notices, and it would let that test be un-skipped: it would now fail in ~13 minutes instead of hanging.

## Fix

A parent that kills a child which overruns its budget.

**When.** The deadline is the earlier of two bounds:

- what the test could legitimately need — its own `--timeout` × attempts, plus 120s of grace for interpreter start, collection and teardown;
- what the job has left — the step is torn down at `timeout-minutes` regardless, so a child still running just before that cannot finish.

The second bound is what makes this work at all. `sglang-test` and `trtllm-test` enable the parallel stage without setting `gpu_test_timeout_minutes`, so they inherit `shared-test.yml`'s 30-minute default; the longest `pre_merge` test in each lane wants 34–62 minutes. Unclamped, the runner always died first and the watchdog never got a turn. It also matters for late launches: these lanes run serially, so a routine test starting well into the step has less time left than its own window assumes.

GitHub exposes no variable for a step's own `timeout-minutes`, so `shared-test.yml` passes it through as `GPU_TEST_TIMEOUT_MINUTES`. When it is absent — local runs — there is no job wall and no clamp, rather than a guessed one.

**Attempts** come from `DD_CIVISIBILITY_FLAKY_RETRY_COUNT`, which is a count of *retries*: ddtrace computes `retries_so_far = len(test.test_runs) - 1`, so the shipped `COUNT=2` means three attempts, and `_reset_pytest_timeout()` gives each a fresh window. `DD_CIVISIBILITY_FLAKY_RETRY_ENABLED` is honoured too — `nightly-ci.yml` disables retries on nine jobs, several of which run this stage, where a child makes exactly one attempt.

**How.** Signals are sent without waiting, so the loop keeps its 1s tick: SIGTERM once, so engines can release `/dev/shm` segments, NATS subscriptions and etcd leases; SIGKILL 5s later. Signalling targets the child's process group — `_launch_test` now sets `start_new_session=True` — plus a psutil sweep for descendants that called `setsid` and escaped it. Escalation is timed rather than counted, because re-sending a signal cannot succeed where the last one failed; the kernel already holds it pending.

Nothing reaps the child behind `Popen`'s back. psutil's `wait()` does, which makes the later `poll()` map `ECHILD` to returncode 0 — a killed test reading as `PASSED`. Pass/fail gates on `rc == 0 and watchdog_reason is None`.

**When it can't.** Only uninterruptible sleep survives SIGKILL. Every job enabling this stage runs on a single-GPU runner, so that GPU's memory is not coming back and there is nowhere to put the remaining tests: the run stops with a named cause instead of spinning.

**Bounds on relaunching.** A wedged child's output may contain a retryable-init marker; watchdog kills never relaunch. Separately, each relaunch used to take a fresh `start_time`, so four attempts got four full deadlines — `4 × (2T + 120)` is 88 minutes at `T=600`, past the pre-merge ceiling, with the watchdog silent because its clock kept restarting. Retries are now bounded across all attempts.

**Reporting.** A SIGKILLed child never reaches its own `--junitxml` write, so the orchestrator synthesizes the entry as a `WatchdogTimeout` failure — but only when the child left none, since a wedge *after* `pytest_sessionfinish` leaves a real report on disk. The synthesized `classname` mirrors `_pytest.junitxml.mangle_test_address`, so it lands in the same class a real entry would. Tests the job has no time to start are not launched at all — launching one would kill it within seconds and report it FAILED without having run — and are reported `NOT RUN`, counted against the run so an unfinished suite cannot come out green.

Verified experimentally that the synthesized entry is the *only* record that will exist: with a real `ddtrace` pointed at a stand-in agent, a completed test submits four CI-Visibility events while a SIGKILLed one submits **nothing**.

## Validation

**Unit — 25 tests** in `test_pytest_parallel_gpu.py` (13 pre-existing, 12 new), all pure arithmetic and string handling, no GPU:

| area | covered |
|---|---|
| Attempt count | `COUNT=2` → 3 attempts; retries disabled → 1; unset → shipped default |
| Deadline | full window when the step allows; clamped by the step budget; clamped for a late launch; no clamp at all when there is no job wall |
| Node-id mangling | matches pytest's own output, including a parametrized id containing `::` |
| JUnit synthesis | `<testsuite>` wrapper and counters the aggregator reads; dotted classname; never raises |

**End-to-end — 14 checks** against the real `run_parallel` on an RTX 6000 Ada, using child pytests that block `SIGALRM` so pytest-timeout's signal is undeliverable — the same shape as the C-level block seen in CI — holding `setsid`'d engine-shaped grandchildren:

| check | result |
|---|---|
| Process group reports alive, then dead | pass |
| **Signalling does not block the loop** | pass — ~5ms, vs up to ~15s before |
| **`Popen.poll()` sees the real signal, not a false 0** | pass — `rc=-9` |
| `setsid`'d grandchild swept up | pass — nothing left alive |
| Stuck test killed, recorded `FAILED` with a readable reason | pass |
| Stuck test carrying a retryable-init marker NOT relaunched | pass |
| **Slow-but-within-budget test NOT killed** (false-kill guard) | pass — its own timeout fires first, untouched by the watchdog |
| Passing test untouched | pass |
| Killed tests reach the aggregated JUnit report | pass — dotted classname, `WatchdogTimeout` |
| Run terminates rather than stalling | pass — 16s |
| `run_parallel` reported overall failure | pass — `rc=1` |
| No child processes left alive | pass |

The child in the wedged scenario sets `SIGTERM` to `SIG_IGN`, so the escalation to SIGKILL is genuinely exercised rather than short-circuited. The harness shrinks the grace and teardown windows so a run completes in ~50s; the production values are not exercised at full duration.

Also run: `black` and `ruff` clean.

**Not validated:** the abort path — a process that genuinely ignores SIGKILL (uninterruptible sleep) cannot be created from userspace. And behavior against a real stuck vLLM/SGLang engine holding VRAM: this virtualenv has no torch or CUDA bindings, so what was verified is process-tree death rather than VRAM reclamation.

## Notes

Independent of and non-overlapping with #11970, which reaps orphaned trees when the *orchestrator* exits. That PR does not help here: it reclaims VRAM only at the job ceiling, after the runner slot is already spent. This one prevents the stall.

Not addressed: *why* `test_sglang_indexers_sync` wedges. This is the guardrail, not the root cause.

Also not addressed, and worth its own change: **`combined.xml` has no CI consumer.** Nothing under `.github/` copies, uploads or parses it — the uploaded artifact is the parent pytest's `pytest_test_report.xml`, which receives no reports because `tests/conftest.py`'s `pytest_runtestloop` returns `True`. So the synthesized entries are correct but not yet visible anywhere. Fixing that would also close the existing "239 passed" vs "1 test completed" discrepancy between the orchestrator's own output and the uploaded artifact.

Considered and dropped, recorded so they aren't re-derived:
- **Asking the child to fail itself first** (SIGTERM → plugin raises → pytest writes its own report with a traceback). Prototyped and reverted: it needs a plugin injected into every child plus a two-phase state machine, and would not fire for a signal-deaf child, which is the shape that actually hangs.
- **Modelling GPUs as lanes that retire individually**, so a wedged child costs one lane rather than the run. Built and removed: every job that enables this stage runs on a single-GPU runner, so there is no other lane to schedule onto and the machinery bought nothing.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


nv-tusharma:
@coderabbitai review

Earlier comments were against 5688e41. The never-retried kill, the deadline cap for long-timeout tests, and the missing JUnit entry for killed children were addressed in 7d51ae1 and b886532.

nv-tusharma:
@coderabbitai Not adding these. The watchdog path is validated end-to-end against the real run_parallel (8 checks, 3 of them fault-injection or negative controls); unit-testing it would need process orchestration that is slow and flaky in CI for little return.

Also note the capping you reference was removed in c6c274e, which landed after your review.

nv-tusharma:
/ok to test [c6c274e](https://github.com/ai-dynamo/dynamo/pull/12710/commits/c6c274eb7e4a1f5c10ba817b9a30d5f5fe1190e3)

tests/utils/pytest_parallel_gpu.py
```diff
@@ -267,6 +315,29 @@ def _get_gpu_used_gib(gpu_index: int = 0) -> float:
...
+    more than once and each attempt gets a fresh --timeout window.
+
+    This is deliberately below the theoretical worst case of retries + 1 full
+    attempts: across 376 observed child runs the slowest took 0.78x its declared
+    timeout, so 2x leaves comfortable room while still catching a hang early.
+    """
+    try:
+        return max(1, int(os.environ["DD_CIVISIBILITY_FLAKY_RETRY_COUNT"]))
+    except (KeyError, ValueError):
+        return 2  # matches the shipped retry count
```
dmitry-tokarev-nv: `DD_CIVISIBILITY_FLAKY_RETRY_COUNT` is a count of *retries*, not attempts. ddtrace's `retry_handlers.py` computes `retries_so_far = len(test.test_runs) - 1  # Initial attempt does not count`, so `COUNT=2` means a healthy child can run its test **3** times — and ddtrace calls `_reset_pytest_timeout()` before each attempt, so each one gets a fresh full `--timeout` window instead of sharing one.

That makes the docstring's 0.78x figure a *per-attempt* observation. ART multiplies attempts, so the budget needs to be `attempts x per-attempt max`, not 2x total. As written, worst legitimate runtime ~3T exceeds the 2T+120 deadline for every T > 120s, and every profiled GPU-parallel test declares T >= 180s — so a test that times out on attempts 1 and 2 is SIGKILLed partway through a legitimate attempt 3 and recorded FAILED, with the orchestrator's own relaunch also gated off at L891.

I checked whether `-x` saves this and it does not: ddtrace runs the attempts inside its own `pytest_runtest_protocol` and relabels the reports `rerun`/`dd_atr_attempt_failed`, which never increments `Session._testsfailed`.

Separately, `DD_CIVISIBILITY_FLAKY_RETRY_ENABLED` is never consulted. `shared-test.yml` wires that flag from the `dd_flaky_retry_enabled` input but hardcodes `COUNT: '2'` unconditionally, and `nightly-ci.yml` sets the flag to `'false'` in nine places, several of them on jobs that also set `run_gpu_parallel_tests: true`. With retries off a child makes exactly one attempt, so on nightly the deadline is twice what the stated rationale justifies — on the lane where a stalled slot is most expensive.

```suggestion
    if os.environ.get("DD_CIVISIBILITY_FLAKY_RETRY_ENABLED", "").lower() in (
        "0",
        "false",
    ):
        return 1  # retries off: a healthy child makes exactly one attempt
    try:
        # COUNT is *retries*; a healthy child runs retries + 1 attempts, and
        # ddtrace resets pytest-timeout so each attempt gets a full window.
        return max(1, int(os.environ["DD_CIVISIBILITY_FLAKY_RETRY_COUNT"]) + 1)
    except (KeyError, ValueError):
        return 3  # the shipped retry count of 2, plus the initial attempt
```

The "deliberately below the theoretical worst case" paragraph in the docstring above will need updating to match.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -759,10 +830,44 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
 
+        # Kill anything past its deadline; the completion check below reaps it
+        # this same iteration and frees its GPU budget for the queued tests.
+        # Keyed on poll(), not on watchdog_reason, so a kill that fails is
+        # retried next iteration instead of leaving the child alive forever.
+        for w_id, run_info in list(running.items()):
+            if run_info.proc.poll() is not None:
+                continue
+            elapsed = now - run_info.start_time
+            deadline = run_info.test.timeout * watchdog_multiplier + _WATCHDOG_GRACE_S
```
dmitry-tokarev-nv: Nothing clamps this to the step's own `timeout-minutes`, so the guard is inert exactly where the hang window is widest. The sglang (`pr.yaml:783-788`) and trtllm (`:822-827`) parallel jobs set `run_gpu_parallel_tests: true` without `gpu_test_timeout_minutes`, taking `shared-test.yml:85`'s `default: 30` = 1800s; vllm gets 2700s. A kill can only ever fire when `2T + 120 < cap`, and three pre_merge profiled tests already exceed theirs:

| test | declared | deadline | step cap |
|---|---|---|---|
| `tests/mm_router/test_vllm_mm_router_e2e.py:307` | 1800s | 3720s | 2700s |
| `tests/frontend/test_frontend_api_surface_compliance.py:468` | 975s | 2070s | 1800s |
| `tests/serve/test_trtllm.py:306` | 960s | 2040s | 1800s |

Each is the longest-timeout profiled test in its lane — so the watchdog is dead precisely on the tests with the most room to hang.

The second-order effect is larger. The deadline runs from the *child's* start, not the step's, and `tests/conftest.py:275` calls `auto_worker_count` without `test_profiled_gibs`, so `vram_utils.py:83-89` yields `max(1, int(24 * 0.85 / 24)) = 1` slot and these lanes run strictly serially. A routine `timeout=600` test gets a 1320s deadline with only 480s of slack in a 1800s step, so any child launched more than ~8 minutes in is unguarded whatever its own timeout — under serial execution, nearly every test after the first.

A job-level budget (`deadline = min(deadline, step_deadline - now)`) covers both cases, and from the history it looks like that is what `c6c274eb7e` removed. Without it, the comment at L318-321 — "a test whose deadline lands past the step's own timeout-minutes is simply not guarded, which is where it was before this existed" — is accurate, but describes most of the suite rather than an edge case.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -759,10 +830,44 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
     while pending or running:
         now = time.monotonic()
 
+        # Kill anything past its deadline; the completion check below reaps it
+        # this same iteration and frees its GPU budget for the queued tests.
+        # Keyed on poll(), not on watchdog_reason, so a kill that fails is
+        # retried next iteration instead of leaving the child alive forever.
+        for w_id, run_info in list(running.items()):
+            if run_info.proc.poll() is not None:
+                continue
```
dmitry-tokarev-nv: This contract does not hold, and it inverts: the retry fires only in the case where it cannot help, and is skipped in the case it was written for.

`terminate_process_tree` calls `terminate_process(parent, ...)` — which swallows its own `AccessDenied`/`NoSuchProcess` — and then `parent.wait(timeout=5)` (`managed_process.py:86`), which on POSIX reaps the child via `os.waitpid`. Verified with psutil 7.2.2: after `h.kill(); h.wait(timeout=10)`, `subprocess.Popen.poll()` returns **0**, because CPython's `_internal_poll` maps the resulting `ECHILD` to returncode 0.

So when the parent dies but a grandchild survives and the call raises while killing the survivors — the exact partial-kill case this comment describes — `poll()` is already non-`None`, the `continue` below short-circuits, and the surviving tree is never re-killed. The retry can only actually fire when `parent.wait()` times out on a still-live parent, where re-running the identical call cannot help and instead becomes the unbounded blocking spin flagged at L865. Worth noting L879-881 documents this same psutil-reaps-the-child behaviour for `passed`, so the two comments contradict each other.

There is also a pid-reuse hazard if the retry ever does fire after a successful reap: the pid has been returned to the OS, and `terminate_process_tree(same_pid)` can then signal an unrelated process.

**Related race on these same lines.** Nothing re-polls between the check below and the kill at L865. A child that finishes in that window has already had `watchdog_reason` set at L854, so `passed = rc == 0 and run_info.watchdog_reason is None` (L882) forces `False` — a genuinely passing test is reported FAILED with "killed by the orchestrator", its real JUnit is overwritten at L944, and the relaunch path is suppressed. That window looks like microseconds, but it widens to the full duration of the preceding kills because `elapsed` is measured against the loop-top `now` (next comment). Re-polling immediately before the kill, or setting `watchdog_reason` only once the kill has confirmed it found a live process, closes it.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -759,10 +830,44 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
+            elapsed = now - run_info.start_time
```
dmitry-tokarev-nv: `now` is sampled once at L836, and the kills below block for up to ~15s each. By the time later entries in this same `for` are evaluated, `now` can be stale by 15s x the number of children killed ahead of them — which biases this comparison and widens the exit race above from microseconds to tens of seconds.

```suggestion
            # Re-sample: the kills below block for up to ~15s each, so `now` from
            # the top of the loop is already stale for every later child.
            elapsed = time.monotonic() - run_info.start_time
```

`duration = now - run_info.start_time` at L878 has the same problem, and it feeds three consumers: the `[wN] <name> FAILED [Ns]` line, `_CompletedTest.duration` in the sequential-time/speedup summary, and the `time=` attribute of the synthesized JUnit at L173/L181. `_build_status_lines(now)` and `next_status = now + 10` (L974/L980) are computed from the stale value too, so `next_status` can land in the past and force a status dump on every subsequent iteration.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -759,10 +830,44 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
+                    f"[watchdog] w{w_id} hit its time limit ({limit_note}). Its "
+                    f"own timeout did not stop it — killing the test process and "
+                    f"everything it started."
+                )
+            try:
+                terminate_process_tree(
+                    run_info.proc.pid, immediate_kill=True, timeout=5
+                )
+            except (psutil.Error, OSError) as exc:
+                _print(f"[watchdog] w{w_id} kill failed ({exc}) — retrying")
```
dmitry-tokarev-nv: Three separate things converge on this call.

**1. It can never give up.** `terminate_process_tree(..., timeout=5)` does `parent.wait(5)` + `psutil.wait_procs(all, 5)` + `psutil.wait_procs(alive, 5)` — up to ~15s of blocking (measured) inside a loop meant to tick once a second. If the child does not die — a CUDA/NCCL ioctl in uninterruptible D-state defers even SIGKILL, which is the "swallowed by a C-level block" class the L318 comment names as the motivation — `poll()` stays `None`, the `continue` at L843 never trips, and the kill is re-issued every iteration with no attempt counter and no escalation. The entry never leaves `running`, so `while pending or running:` never terminates: no summary, no `_aggregate_junit_xml`, no non-zero return. And the loop is now blocked ~15 of every 16 seconds, starving every other child's reaping and every queued launch. That is the PR's own motivating scenario, reinstated and made worse.

**2. `immediate_kill=True` skips all cleanup.** The kill-sequence notes in `managed_process.py` (~L517-527) say SIGKILL releases no POSIX shm segments, NATS subscriptions, etcd leases or KV-cache index entries, and leaves ports in TIME_WAIT. A killed vLLM/SGLang tree leaves its EngineCore message queues in `/dev/shm` unlinked, because the `resource_tracker` dies with it. The slot is refilled on the same loop pass, so the next test starts against a shrunken `/dev/shm` and stale distributed state and fails for reasons invisible in its own log — one wedged test becomes a cascade of unrelated red tests.

**3. The GPU budget is released unconditionally.** L958-960 decrement `budget_used`/`running_count` on the same path for a clean exit and for a kill, but `terminate_process_tree` signals nothing back on success, and its descendant snapshot is taken by ppid *before* signalling (`managed_process.py:72`) — so anything spawned after the snapshot, or already reparented to init, survives holding VRAM. `_get_gpu_used_gib` at L989 still sees that memory, so `_select_launches` blocks while the accounting insists the GPU is free.

Worth considering: an attempt cap that escalates to "give up, mark the slot lost, keep scheduling on the other GPUs" rather than spinning; and reusing `ManagedProcess`'s pgid-based kill (`start_new_session=True` + `_terminate_process_group`, `managed_process.py:455`/`498`) instead of ppid walking — `_launch_test`'s `Popen` at L813 sets no session, so there is no process-group fallback today.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -771,12 +876,21 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
                 if run_info.reader_thread is not None:
                     run_info.reader_thread.join(timeout=5)
```
dmitry-tokarev-nv: This `join` gives up silently after 5s, and after a watchdog kill that is the expected case rather than the exception.

`_capture_output` (L342-353) is `for line in iter(pipe.readline, "")` followed by `pipe.close()`, so it only returns at EOF — which requires every holder of the inherited write end to exit. `terminate_process_tree` snapshots descendants by ppid *before* signalling, so anything spawned after the snapshot or already reparented to init survives holding the pipe. `_launch_test`'s `Popen` (L813) sets no `start_new_session`, unlike `ManagedProcess` (`managed_process.py:455`), so there is no process-group fallback either.

When it does time out: `pipe.close()` never runs, so an fd leaks once per kill; and the still-live daemon thread keeps appending to `run_info.captured` while L933-939 iterates and `reversed()`es that same list to pick `fail_reason`. `captured` is also unbounded, so a chatty wedged engine grows it for the run's full duration.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -771,12 +876,21 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
 
                 # Detect retryable init errors (profiling race, OOM at startup)
-                if not passed and test.retries < _MAX_RETRIES:
+                # A stuck test is not a transient startup failure, so never
+                # relaunch one even if its output carries a retryable marker.
+                if (
+                    not passed
+                    and run_info.watchdog_reason is None
+                    and test.retries < _MAX_RETRIES
+                ):
```
dmitry-tokarev-nv: This correctly stops a watchdog kill from being retried — but the retry path itself still has no cumulative bound, so a test can burn the whole step without the watchdog ever getting a turn.

`_launch_test` takes a fresh `start_time=time.monotonic()` on every launch (L820), and `_TestEntry` records `retries` but no first-launch timestamp and no accumulated duration. So each of the up-to-four attempts starts a brand-new `timeout * multiplier + 120` budget.

A test whose engine init is genuinely broken prints an `_RETRYABLE_INIT_MARKERS` line and exits just under its deadline on each attempt: `4 x (2T + 120)` is 88 minutes at T=600 — past the 30-minute pre_merge ceiling — and 4.1 hours at T=1800. The runner then dies at `timeout-minutes` with queued tests never launched, which is the failure this PR exists to prevent; the watchdog simply never fires because the clock keeps restarting.

Carrying a `first_start_time` on `_TestEntry` and checking total elapsed against a per-test ceiling would close it.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -145,6 +150,49 @@ def _fmt_req(test: _TestEntry) -> str:
...
+    ET.SubElement(
+        case, "failure", {"message": reason, "type": "WatchdogTimeout"}
+    ).text = reason
+    try:
+        os.makedirs(_JUNIT_DIR, exist_ok=True)
+        ET.ElementTree(suite).write(
+            _junit_path(test.name), encoding="unicode", xml_declaration=True
+        )
+    except OSError as exc:
+        _print(f"[watchdog] could not write a JUnit entry for {test.name}: {exc}")
```
dmitry-tokarev-nv: Two things in this function body.

**Nothing here is allowed to raise, but most of it can.** Only `os.makedirs` and `.write()` are inside the `try`, and only `OSError` is caught — this `import` and the whole `ET.Element`/`ET.SubElement` construction (L172-186) sit outside it. `.write()` also raises `UnicodeEncodeError` (a `ValueError`, not an `OSError`) for a non-ASCII node id under a non-UTF-8 locale: with `LC_ALL=C` and coercion off, `locale.getpreferredencoding()` returns `US-ASCII` and ElementTree opens the file with it. And the orchestrator parent runs under pyproject's `filterwarnings = ["error"]`, so a first-time import emitting a `DeprecationWarning` becomes an exception — this repo has been bitten by exactly that before (a stray `from ast import Str` silently disabling the NVDEC probe for a whole session).

`while pending or running:` (L835) has no exception handling at all and `running` is a plain local dict, so any escape propagates out of `run_parallel` and out of `pytest_runtestloop`: every child still in `running` is orphaned holding VRAM, and the step burns to `timeout-minutes` anyway. A `try/finally` around the scheduling loop that kills whatever remains in `running` looks worth having independently of this function.

**The synthesized `classname` is a form pytest never emits.** `mangle_test_address` replaces `/` with `.` and strips `.py`, so real pytest writes `classname="tests.serve.test_vllm" name="test_foo[agg]"`. `rpartition("::")` yields `tests/serve/test_vllm.py` instead — names match, classnames never do — and it mis-splits any parametrized id containing `::` (IPv6 literals such as `::1` appear in this repo's ids), moving the trailing fragment into `name`. The repo consumer that keys on this pairing, `.github/workflows/upload_complete_workflow_metrics.py` (`test_full_name = f"{classname}::{name}"`), reads `pytest_test_report.xml` rather than `combined.xml`, so it is masked today by the collection gap noted at L159 — but it goes live the moment that file is wired up, which is the obvious fix for that gap.

Both, together:

```suggestion
    import xml.etree.ElementTree as ET

    try:
        # Mirror pytest's own mangle_test_address so a synthesized entry lines up
        # with a real one: "tests/a/b.py::T::t[p::q]" -> classname "tests.a.b.T",
        # name "t[p::q]".
        path, open_bracket, params = test.name.partition("[")
        names = path.split("::")
        names[0] = names[0].replace("/", ".").removesuffix(".py")
        names[-1] += open_bracket + params
        classname, name = ".".join(names[:-1]), names[-1]
        # _aggregate_junit_xml skips any file whose root is not a <testsuite> and
        # sums these counters off it, so the wrapper is required even for one case.
        # Its own name attribute is discarded there, so it is not worth inventing.
        suite = ET.Element(
            "testsuite", {"tests": "1", "failures": "1", "time": f"{duration:.3f}"}
        )
        case = ET.SubElement(
            suite,
            "testcase",
            {
                "classname": classname or test.name,
                "name": name or test.name,
                "time": f"{duration:.3f}",
            },
        )
        ET.SubElement(
            case, "failure", {"message": reason, "type": "WatchdogTimeout"}
        ).text = reason
        os.makedirs(_JUNIT_DIR, exist_ok=True)
        ET.ElementTree(suite).write(
            _junit_path(test.name), encoding="unicode", xml_declaration=True
        )
    except Exception as exc:  # a report write must never escape into the loop
        _print(f"[watchdog] could not write a JUnit entry for {test.name}: {exc}")
```

tests/utils/pytest_parallel_gpu.py
```diff
@@ -824,6 +937,11 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
                         if stripped and not stripped.startswith("="):
                             fail_reason = stripped
                             break
+                    # Stated outright; after a SIGKILL the child's output is
+                    # only whatever happened to flush.
+                    if run_info.watchdog_reason is not None:
+                        fail_reason = run_info.watchdog_reason
+                        _write_watchdog_junit(test, duration, fail_reason)
```
dmitry-tokarev-nv: This overwrites the child's own report. `_write_watchdog_junit` writes to `_junit_path(test.name)` (L190) and L785 hands the child `--junitxml=_junit_path(test.name)` — same function, same argument, byte-identical path — and the write has no existence check.

pytest emits that file in `LogXML.pytest_sessionfinish`, which `wrap_session` runs in its `finally` *after* `pytest_runtestloop` returns, so fixture-teardown hangs are pre-write and safe. But post-`sessionfinish` wedges — `config._ensure_unconfigure()`, atexit handlers, CUDA/NCCL teardown, a lingering non-daemon engine thread — are exactly the class pytest-timeout cannot see and this watchdog exists to kill. There, a complete report (possibly all-PASSED, or FAILED carrying the real traceback) is already on disk and gets replaced by a synthetic `tests=1 failures=1` suite whose only content is the watchdog string.

```suggestion
                        # Only synthesize what the child never wrote: pytest emits
                        # its --junitxml in pytest_sessionfinish, so a wedge in
                        # post-session teardown leaves a real report on disk that
                        # this would otherwise clobber.
                        if not os.path.exists(_junit_path(test.name)):
                            _write_watchdog_junit(test, duration, fail_reason)
```

tests/utils/pytest_parallel_gpu.py
```diff
@@ -145,6 +150,49 @@ def _fmt_req(test: _TestEntry) -> str:
 _JUNIT_COMBINED = os.path.join(_JUNIT_DIR, "combined.xml")
 
 
+def _junit_path(test_name: str) -> str:
+    """Where a child is told to write its --junitxml, keyed on the node id."""
+    safe_name = test_name.replace("/", "_").replace("::", "__")
+    return os.path.join(_JUNIT_DIR, f"{safe_name}.xml")
+
+
+def _write_watchdog_junit(test: _TestEntry, duration: float, reason: str) -> None:
```
dmitry-tokarev-nv, line 192: Worth resolving before the rest of this function matters: `_JUNIT_DIR` is `/tmp/gpu_parallel_junit/`, which nothing in CI collects — so neither justification in the docstring is reachable through this path.

A whole-repo grep for `gpu_parallel_junit` and `combined.xml` matches only this module; nothing under `.github/` copies, uploads or parses it, and `run_parallel` merely `_print`s the path at L1130. The uploaded artifact is the *parent* pytest's `--junitxml=${TEST_RESULTS_DIR}/pytest_test_report.xml` (`.github/actions/pytest-local/action.yml:226`), and because `tests/conftest.py`'s `pytest_runtestloop` returns `True`, the parent's `LogXML` receives zero `pytest_runtest_logreport` calls. A captured job log in this repo shows the gap directly: the orchestrator prints `239 passed in 1764.06s` right after `JUnit XML: /tmp/gpu_parallel_junit/combined.xml`, while the uploaded artifact for the same job reads `1 tests completed (0 failed, 0 errors)`.

The Datadog half cannot be closed here at all: DD is fed by the in-child ddtrace plugin — there is no `datadog-ci junit upload` anywhere in `.github/` — and a SIGKILLed child submits nothing, irrecoverably, since the data never leaves the dead process.

So the fix sits upstream of this function: either aggregate `combined.xml` into the artifact that actually gets uploaded, or have the orchestrator replay results through the parent's `LogXML`. That seems worth doing on its own merits — it would also make the existing 239-vs-1 discrepancy go away. Until then this function has no observable effect, which is also what makes the `classname` defect above latent rather than harmless.

dmitry-tokarev-nv (commented):
Automated deep review of the watchdog. The idea is right and the failure it targets is real — but most of what follows says the guard does not fire when it needs to, and one says it fires when it should not. Suggestions inline where a concrete patch fits.

**The three that matter most**

1. **`_watchdog_multiplier()` budgets 2 timeout windows; a healthy child gets 3 attempts.** `DD_CIVISIBILITY_FLAKY_RETRY_COUNT=2` is a count of *retries* — ddtrace's `retry_handlers.py` has `retries_so_far = len(test.test_runs) - 1  # Initial attempt does not count` — and ddtrace calls `_reset_pytest_timeout()` before each attempt, so every attempt gets a fresh full `--timeout` window rather than sharing one (measured: `--timeout=3` against a 10s sleep ran 9.02s and produced three separate `Timeout (>3.0s)` failures; `-x` does not suppress the retries, because attempt reports are relabelled `rerun`/`dd_atr_attempt_failed` and never increment `Session._testsfailed`). ART is live in these children — the captured job log shows nine of them printing `Auto Test Retries enabled: True`. So worst legitimate runtime is ~3T against a 2T+120 budget, and 3T > 2T+120 for every T > 120s, while every profiled GPU-parallel test declares T >= 180s.
2. **The deadline is never clamped to the step's own `timeout-minutes`,** so the guard is dead on the longest-timeout test in each lane, and — because these lanes end up with one slot — on nearly every test after the first.
3. **A child that does not die turns the watchdog into a ~15s blocking kill re-issued every iteration forever,** so `while pending or running` never exits and the stage produces no summary and no aggregated JUnit at all.

**Test coverage.** The diff adds 123 lines of scheduling logic and no tests. Commit `5688e41068` removed `test_watchdog_deadline_clears_worst_case_retry_budget` (`assert _watchdog_deadline(declared) > 3 * declared`) and `test_psutil_reap_makes_popen_report_a_false_success`; commit `7d51ae1d3e` then lowered the multiplier from 4 to 2 — which the first of those assertions would have caught. Both guard findings that are still live below, so they look worth restoring.

**Checked and found not to be problems**, in case they come up: `DD_CIVISIBILITY_FLAKY_RETRY_COUNT` *does* reach the orchestrator (the GPU-parallel stage uses `.github/actions/pytest-local`, which runs pytest in the job container with no nested `docker run`), and there is no `timeout(0)`/negative/non-numeric timeout marker anywhere in the repo.

---

tests/utils/pytest_parallel_gpu.py
```diff
@@ -35,10 +35,14 @@
...
 from pathlib import Path
 
+import psutil
+
 _repo_root = str(Path(__file__).resolve().parents[2])
 if _repo_root not in sys.path:
     sys.path.insert(0, _repo_root)
 
+# Below the sys.path insert above, which is what makes these resolvable.
+from tests.utils.managed_process import terminate_process_tree  # noqa: E402
```
dmitry-tokarev-nv: This pulls a fairly heavy chain into a module that the CPU-only lane imports at collection time. `tests/utils/managed_process.py` imports `requests`, `psutil`, `tests.utils.constants`, `tests.utils.port_utils` (which imports `fcntl`) and `tests.utils.test_output` at module scope.

`tests/utils/test_pytest_parallel_gpu.py` is marked `[unit, pre_merge, gpu_0]` and does a module-scope `from tests.utils.pytest_parallel_gpu import ...`, with a docstring stating "No GPU or pynvml required -- the scheduler core is pure arithmetic." That isolation is gone now: the gpu_0 lane would fail at *collection* — not as a test failure — on any image missing `requests`/`fcntl`, for a helper it never calls.

`terminate_process_tree` itself depends on nothing but `psutil`, so moving it to a leaf module (or importing it lazily at the call site) keeps the guard without the dependency chain.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -771,12 +876,21 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
                 if run_info.reader_thread is not None:
                     run_info.reader_thread.join(timeout=5)
                 duration = now - run_info.start_time
-                passed = rc == 0
+                # Not `rc == 0`: terminate_process_tree reaps the child via
+                # psutil, so Popen.poll() hits ChildProcessError and subprocess
+                # falls back to status 0 -- a kill would read as a pass.
+                passed = rc == 0 and run_info.watchdog_reason is None
```
keivenchang: hey Tushar — I think the "needs process orchestration" answer to CodeRabbit only covers the kill path. `_write_watchdog_junit` -> `_aggregate_junit_xml` and `_watchdog_multiplier` are both pure, no processes involved. and the `and` on this line is doing real work — someone "simplifies" it back to `rc == 0` later and every watchdog kill quietly logs as PASSED, which is the exact thing you're fixing. you can either pull those two tests out of 60d90ea, or drop a `# do not simplify:` note so the next person knows.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -267,6 +315,29 @@ def _get_gpu_used_gib(gpu_index: int = 0) -> float:
...
+    Taken straight from DD_CIVISIBILITY_FLAKY_RETRY_COUNT (shared-test.yml),
+    since Datadog Auto Test Retries is what makes a healthy child run its test
+    more than once and each attempt gets a fresh --timeout window.
+
+    This is deliberately below the theoretical worst case of retries + 1 full
+    attempts: across 376 observed child runs the slowest took 0.78x its declared
+    timeout, so 2x leaves comfortable room while still catching a hang early.
+    """
+    try:
+        return max(1, int(os.environ["DD_CIVISIBILITY_FLAKY_RETRY_COUNT"]))
```
keivenchang: dropping the cap is fine today, but it ties this to shared-test.yml. at retry count 2 you get 1800*2+120 = 62min, plenty of room. at 4 it's 122min and blows the step budget with nothing failing to tell you. maybe assert `max_declared_timeout * _watchdog_multiplier() + _WATCHDOG_GRACE_S` stays under budget?

tests/utils/pytest_parallel_gpu.py
```diff
@@ -145,6 +150,49 @@ def _fmt_req(test: _TestEntry) -> str:
...
+def _write_watchdog_junit(test: _TestEntry, duration: float, reason: str) -> None:
+    """Write the JUnit entry a watchdog-killed child never got to write.
+
+    SIGKILL means pytest never reaches its own --junitxml write, so without this
+    the test is absent from the aggregated report entirely -- a silent hole in
+    Datadog test visibility and CI test reports rather than a visible failure.
+    """
+    import xml.etree.ElementTree as ET
+
+    classname, _, name = test.name.rpartition("::")
```
keivenchang: `rpartition("::")` gives you `tests/router/foo.py`, but pytest's junitxml writes `tests.router.foo` — so this entry shows up in its own class in the combined report.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`. `_watchdog_attempts()` returns `COUNT + 1` and returns `1` when `DD_CIVISIBILITY_FLAKY_RETRY_ENABLED` is false, and both cases are pinned by tests. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`: `_watchdog_kill_at()` clamps to `step_deadline - _WATCHDOG_REPORT_RESERVE_S`, fed by the new `GPU_TEST_TIMEOUT_MINUTES` passthrough in `shared-test.yml`, with `None` meaning no wall for local runs. Resolving — though see the separate note on how much of the budget the launch gate now leaves unused.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`. `_signal_process_tree()` deliberately does not wait, so nothing reaps behind `Popen`'s back, and there is now an explicit re-poll before `watchdog_reason` is committed.

Verified on a GPU box (psutil 7.2.2, Python 3.12.3) rather than taken on trust — after killing a `start_new_session=True` child:

```
new _signal_process_tree -> poll() = -9
old terminate_process_tree -> poll() = 0    <- the false success
```

Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`: `now` is re-sampled inside the completion loop after the kills and the pipe join, with a comment naming the four consumers. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`: `_TestEntry.first_start_time` plus the `total_elapsed < retry_budget` and `room_in_step` guards bound the retries as a whole rather than per attempt. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Both addressed in `6dfe26d3c8`. The whole body is inside `try` with `except Exception`, and `encoding="unicode"` became `encoding="utf-8"` — which fixes the `UnicodeEncodeError` at the source rather than only catching it. `_mangle_test_address()` now mirrors `mangle_test_address`, and a test pins the `[::1]` case. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`: guarded by `if not os.path.exists(...)` on both the normal completion path and `_record_unreaped`. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Addressed in `6dfe26d3c8`: the `tests.utils.managed_process` import is gone entirely, and the module now depends only on `psutil` for signalling. The gpu_0 unit lane keeps its isolation — the 25 tests in `test_pytest_parallel_gpu.py` collect and pass with only `pytest` and `psutil` installed. Resolving.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Leaving this open. The unbounded re-signalling is genuinely fixed — `_WATCHDOG_GIVE_UP_S` caps it and the SIGTERM grace addresses the cleanup point — but the outcome I described (the loop never exits, no summary, no aggregated JUnit) still happens, now via a deadlock instead of a spin. Evidence in the reply on `_close_child_pipe`.

tests/utils/pytest_parallel_gpu.py
dmitry-tokarev-nv: Leaving this open: the fd leak is fixed, but `_close_child_pipe` deadlocks in the one case it was written for.

Reproduced on a GPU box against the real `run_parallel` at `e25980bd26`, with a child that swallows its own pytest-timeout and whose signals are dropped (i.e. the uninterruptible-sleep case). The watchdog detects it, escalates, logs *"still alive 8s after SIGKILL ... stopping the run"* — and then hangs. `faulthandler` after 40s:

```
Thread (reader):
  pytest_parallel_gpu.py, line 493 in _capture_output
Thread (main):
  pytest_parallel_gpu.py, line 1000 in _close_child_pipe
  pytest_parallel_gpu.py, line 1013 in _record_unreaped
  pytest_parallel_gpu.py, line 1125 in run_parallel
```

Line 493 is `iter(pipe.readline, "")`; line 1000 is `run_info.proc.stdout.close()`. The reader thread is blocked in `readline()` holding the buffered-reader lock — exactly the survivor case the docstring anticipates — and `close()` from the scheduling thread waits on that same lock forever. Isolated repro, no orchestrator involved:

```python
p = subprocess.Popen([sys.executable, "-c", "import time;time.sleep(60)"],
                     stdout=subprocess.PIPE, text=True)
threading.Thread(target=lambda: [None for _ in iter(p.stdout.readline, "")],
                 daemon=True).start()
time.sleep(0.5)
# close() from another thread never returns
```
→ `stdout.close() returned within 5s: False`

So the run burns to `timeout-minutes` with no summary and no aggregated JUnit — the same end state as before the PR, just with a better log line first. Closing the fd without touching the buffered object (`os.close(run_info.proc.stdout.fileno())`), or simply leaving the fd to process exit, avoids trading a bounded leak for an unbounded hang.

dmitry-tokarev-nv (commented):
Re-reviewed after `6dfe26d3c8` / `e25980bd26`. The response is thorough — 8 of my 11 threads are resolved, and I verified the core fixes on a GPU box (RTX 6000 Ada, Python 3.12.3, psutil 7.2.2, pytest 9.1.1) rather than reading them:

```
new _signal_process_tree   -> poll() = -9      old terminate_process_tree -> poll() = 0
old ppid-walk    -> reparented grandchild alive: True    (missed)
new killpg       -> reparented grandchild alive: False   (reached)
new killpg+psutil -> setsid'd grandchild alive: False    (reached)
_process_group_alive after the child is reaped: True     (survivor seen)
25/25 unit tests pass with only pytest + psutil installed
```

Driving the **real `run_parallel`** at this head then turned up three things, all reproduced rather than inferred. Details inline; the short version:

1. **The give-up path deadlocks**, so the watchdog still hangs the run — same end state as before the PR, with a better log line first.
2. **`launchable` aliases `pending`**, which crashes any local multi-slot run with `IndexError`. CI escapes only because this same commit sets `GPU_TEST_TIMEOUT_MINUTES`.
3. **The NOT RUN gate strands most of the job budget** and then fails the run with zero test failures.

Nothing here undoes the direction of the change — the escalation ladder, the pgid kill, the job-budget clamp and the restored tests are all clearly right. (1) and (2) look like merge blockers; (3) is a judgement call but will be loud on the first run.

---

tests/utils/pytest_parallel_gpu.py
```diff
@@ -758,201 +985,448 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
+        the join times out and the thread never runs its own close(). Closing
+        here keeps a long run from leaking one fd per kill.
+        """
+        if run_info.reader_thread is not None:
+            run_info.reader_thread.join(timeout=5)
+        try:
+            if run_info.proc.stdout is not None:
+                run_info.proc.stdout.close()
+        except OSError:
+            pass
```
dmitry-tokarev-nv, line 1002: **This deadlocks, in the only situation it runs in.** The docstring correctly anticipates the survivor case — but `close()` on a `BufferedReader` whose lock is held by a thread parked in `readline()` waits on that lock forever. It trades a bounded fd leak for an unbounded hang.

Reproduced against the real `run_parallel` at `e25980bd26` on a GPU box, with a child that swallows its own pytest-timeout and whose signals are dropped (the uninterruptible-sleep case this exists for). The watchdog detects it, escalates correctly, logs

```
[watchdog] w1 ran 5s against a 5s limit, from a 3s test timeout x1 attempts — killing ...
[watchdog] w1 still alive 8s after SIGKILL, so it is stuck where signals cannot reach it.
          Its GPU memory is not coming back — stopping the run.
```

…and then never returns. `faulthandler` at 40s:

```
Thread (reader):  pytest_parallel_gpu.py:493  in _capture_output    # iter(pipe.readline, "")
Thread (main):    pytest_parallel_gpu.py:1000 in _close_child_pipe  # proc.stdout.close()
                  pytest_parallel_gpu.py:1013 in _record_unreaped
                  pytest_parallel_gpu.py:1125 in run_parallel
```

The process exits only when the outer `timeout` kills it (`exit=124`). No summary, no `_aggregate_junit_xml`, no return code — the same end state as before this PR.

Isolated repro, no orchestrator involved:

```python
p = subprocess.Popen([sys.executable, "-c", "import time;time.sleep(30)"],
                     stdout=subprocess.PIPE, text=True)
threading.Thread(target=lambda: [None for _ in iter(p.stdout.readline, "")],
                 daemon=True).start()
time.sleep(0.5)
# then, from another thread:
p.stdout.close()                      # returned within 3s: False   <- deadlock
os.close(p.stdout.fileno())           # returned within 3s: True
p.stdout.fileno()                     # returns 3, does not block
```

Dropping the fd underneath the parked thread avoids the lock. Measured on the same box, all three lines above:

```suggestion
        if run_info.reader_thread is not None:
            run_info.reader_thread.join(timeout=5)
            if run_info.reader_thread.is_alive():
                # A survivor still holds the write end, so the thread is parked
                # in readline() holding the buffered-reader lock. close() waits
                # on that same lock and never returns -- the hang this watchdog
                # exists to prevent. Drop the fd underneath it instead; the
                # thread is a daemon and dies with the process.
                try:
                    if run_info.proc.stdout is not None:
                        os.close(run_info.proc.stdout.fileno())
                except (OSError, ValueError):
                    pass
                return
        try:
            if run_info.proc.stdout is not None:
                run_info.proc.stdout.close()
        except OSError:
            pass
```

Separately, once this is unblocked: `_record_unreaped` books the wedged child, but the tests still in `running` when the loop breaks on `abort_reason` are killed in the `finally` and never appended to `completed` — no summary line, no JUnit. Only `pending` gets NOT RUN entries. That is the same reporting hole the PR set out to close, on the abort path. Currently unobservable because the deadlock happens first.

tests/utils/pytest_parallel_gpu.py
```diff
@@ -758,201 +985,448 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
+            # Only offer the scheduler tests the job still has time to finish.
+            # Launching one it does not would put the child past its watchdog
+            # deadline on the first pass: killed within seconds and reported
+            # FAILED without having run.
+            latest_start = _latest_start(now)
+            launchable = (
+                pending
+                if latest_start is None
+                else [t for t in pending if now + t.timeout <= latest_start]
+            )
```
dmitry-tokarev-nv, line 1302: **`launchable` is `pending` here — the same list object, not a copy — and the loop below mutates it while indexing it.** `_select_launches` returns indices into whatever is passed as `pending=`, and lines 1327-1329 do `entry = launchable[launch_idx]` followed by `pending.remove(entry)`. With `latest_start is None` the `remove` shrinks the list the next index addresses.

The comment at L1326-1328 ("take the entries by identity and drop them from `pending` separately") shows the hazard was in mind — this branch just defeats it.

Measured at `e25980bd26` against the real `run_parallel`:

```
no GPU_TEST_TIMEOUT_MINUTES, 4 slots (local)  -> IndexError: list index out of range
no GPU_TEST_TIMEOUT_MINUTES, 1 slot           -> rc=1 (no crash)
GPU_TEST_TIMEOUT_MINUTES=10, 4 slots (CI)     -> rc=1 (no crash)
```

So: every local multi-slot run crashes on its first launch pass. CI is spared only because the `GPU_TEST_TIMEOUT_MINUTES` passthrough added in this same commit makes `_job_budget_s()` non-`None`, so the other branch builds a fresh list. That is a load-bearing coincidence between a workflow env var and a scheduler invariant. The `try/finally` does contain it — children are cleaned up — but the stage dies with a traceback.

```suggestion
            launchable = (
                # list(...), not `pending` itself: the indices _select_launches
                # returns address this list, and the launch loop below removes
                # the chosen entries from `pending` while iterating them. If the
                # two are the same object the second index in a multi-launch
                # pass points past the end.
                list(pending)
                if latest_start is None
                else [t for t in pending if now + t.timeout <= latest_start]
            )
```

tests/utils/pytest_parallel_gpu.py
```diff
@@ -758,201 +985,448 @@ def _launch_test(test: _TestEntry, env_base: dict) -> _RunningTest:
...
+    # Set when the run cannot continue: a child would not die, so its VRAM is
+    # never coming back. Every job that runs this stage has a single GPU, so
+    # there is nowhere else to put the remaining tests.
+    abort_reason: str | None = None
+
+    def _latest_start(now: float) -> float | None:
+        """Last moment a test may be launched and still have time to finish."""
+        if step_deadline is None:
+            return None
+        return step_deadline - _WATCHDOG_REPORT_RESERVE_S
```
dmitry-tokarev-nv, line 1039: Two things about this gate.

**It reserves each test's *declared* timeout plus a 60s tail, so most of the job budget goes unused — and then the run is failed for it.** Measured at `e25980bd26`: 6 tests declaring `timeout=30` that actually take 12s, one slot, `GPU_TEST_TIMEOUT_MINUTES=2` (120s wall):

```
PASSED  [w0] test_s0 [13s/30s]
PASSED  [w1] test_s1 [13s/30s]
PASSED  [w2] test_s2 [13s/30s]
NOT RUN [w3] test_s3 - the job step had less time left than its own 30s timeout needs
NOT RUN [w4] test_s4 - ...
NOT RUN [w5] test_s5 - ...
============== 3 not run, 3 passed in 39.09s (vs 39s seq, 1.0x) ==============
rc=1
```

Zero failures, 81 of 120 seconds unused, job red. The remaining tests would each have finished in 12s with 81s left.

Scaled to the real lanes: `latest_start = step_deadline - 60`, and admission needs `now + t.timeout <= latest_start`. On the 30-minute sglang/trtllm lanes that means nothing may start after roughly `t + 13min` for the longest declared timeouts, and any test declaring >= 1740s could never start at all. Combined with these lanes running one slot (`auto_worker_count` yields 1 at 24 GiB), the tail of a suite that used to just fit now reports NOT RUN — and `return 0 if n_failed == 0 and n_not_run == 0 else 1` turns that red.

The intent ("a suite that did not finish must not come out green") is right, but gating on the declared timeout rather than an expected duration makes it fire far more often than the budget actually requires. Worth checking the real per-lane stage durations before merging, since this flips on the first run rather than degrading gradually.

**Minor:** `now` is accepted and never used — the body only reads `step_deadline` and the constant, so the result is invariant for the whole run and could be computed once outside the loop.

nv-tusharma:
need a better overall solution here, closing stale PR

PR #12718, opened 2026-08-05
test(ci): reclassify CPU-only sglang unit tests from gpu_1 to gpu_0 (OPS-8033)

dmitry-tokarev-nv:
## Summary

The sglang pre_merge amd64 GPU job's parallel stage hit ~29.4 min (the 30-min step cap) while executing **zero seconds of recorded test time**. The wall clock went into 353 items — 9,496 cumulative seconds — trickling through the VRAM-aware scheduler at effective concurrency ~5.4 with the GPU idle (`GPU0: 0.5/22 GiB`, 350+ queued). 89% of that cumulative time is CPU-only **unit** tests marked `gpu_1`, each paying ~24s of per-item scheduling + fresh-interpreter sglang/torch import.

This moves the eight CPU-only files to `gpu_0`, where the loadscope-parallel stage amortizes imports per worker (existing gpu_0 stage: 205 tests in 96s).

Linear OPS-8033

## Details

Observed on PR #12051's run 31036884612 (job 92413246330), from the stage's own telemetry and junit artifacts:

| file | tests | cumulative time in the VRAM-scheduled stage |
|---|---|---|
| frontend `test_sglang_processor_unit.py` | 188 | 4,439s |
| sglang `test_sglang_unit.py` | 82 | 2,235s |
| frontend `test_sglang_tool_calls.py` | 22 | 528s |
| sglang `test_sglang_multimodal_embedding_cache.py` | 17 | 480s |
| frontend `test_sglang_processor_api.py` | 16 | 363s |
| sglang `test_sglang_nvdec_video_decoder.py` | 9 | 261s |
| frontend `test_sglang_multimodal_prepost.py` | 5 | 109s |
| sglang `test_fpm_contract.py` | 2 | 42s |

- The `gpu_1` marks were a proxy for "needs the sglang image" — `test_sglang_unit.py` said so verbatim, and `test_fpm_contract.py` carried `profiled_vram_gib(0)` alongside `gpu_1`. Per `.ai/pytest-guidelines.md`, gpu markers mean GPUs **required**.
- The proxy's fear is empirically obsolete: the **arm64 CPU-only lane already runs `pre_merge and sglang and gpu_0`** (205 passed / 96s in the same run) and imports these same packages (`test_sglang_decode_handler`, `test_sglang_multimodal_video`, …).
- The one CUDA-touching test (`test_frames_are_pinned_on_cuda`) keeps its `skipif not torch.cuda.is_available()` and stays consistent with gpu_0 = "no GPU required": it self-skips on the CPU lane and still guards pinning on the GPU job's gpu_0 stage.
- Serve deployments and router e2e keep `gpu_1` — they are the stage's genuine GPU work (667s + 212s cumulative) and now get the scheduler to themselves. Expected parallel-stage wall drops from ~29 min to well under 10.

Any of the eight `pytestmark` blocks — the change is identical in each; the comment above the mark documents the reasoning in place.

## Validation

- **amd64, GPU masked (`NVIDIA_VISIBLE_DEVICES=void`), run from `/workspace` like CI**: all eight files — **351 passed / 1 skipped / 0 failed** serially, identical result under `-n auto --dist=loadscope` in **~21s wall** (the same tests cost ~8,400 cumulative seconds in the VRAM-scheduled stage). The one skip is `test_frames_are_pinned_on_cuda` skipping exactly as the gpu_0 contract requires. `--collect-only -m "pre_merge and sglang and gpu_0"` selects all 352.
- The no-GPU run surfaced a real latent dependency this PR also fixes (second commit): 14 `test_sglang_unit.py` tests failed on an accelerator-less host because `parse_args` reaches SGLang's `ServerArgs.__post_init__`, whose device auto-detection ends in `NotImplementedError` when no accelerator exists. The arm64 CPU lane never caught it because the old `gpu_1` mark kept the file off that lane. An autouse fixture sets `SGLANG_USE_CPU_ENGINE=1` (SGLang's public CPU knob, verified at v0.5.16) only when CUDA is absent, with `lru_cache` clears on both probes; GPU hosts exercise the real detection path unchanged.
- `ruff` clean; GPU-use audit per file: zero direct CUDA references except the self-skipping pinned test.
- **Measured by this PR's own CI (run 31049499476, green, zero failures)**: the sglang amd64 test job finished in **22 minutes** (was 38, with the parallel stage pinned at the 30-min cap). Stage shape after the move: gpu_0 loadscope stage **557 passed in 87.6s** (absorbing the ~350 moved tests), VRAM-scheduled stage down to **12:09** with only genuine GPU work, serial tail 4:14. The arm64 CPU lane ran the moved files for the first time and passed in ~4 min, exercising the ServerArgs CPU fixture.

## Related Issues

Relates to the timeout observed on #12051.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12718" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Tests**
  * Updated test scheduling markers so CPU-only tests run in the appropriate CPU lane without reserving GPU VRAM.
  * Clarified test annotations for CPU-only execution and CUDA-related behavior.
  * Preserved existing test logic and other execution markers.


components/src/dynamo/sglang/tests/test_sglang_nvdec_video_decoder.py
```diff
@@ -28,7 +28,12 @@
 pytestmark = [
     pytest.mark.unit,
     pytest.mark.sglang,
-    pytest.mark.gpu_1,  # sglang tests run on GPU-enabled workers
+    # gpu_0: CPU-only in the sglang image -- proven by the arm64 CPU test
+    # lane, which already runs `sglang and gpu_0` and imports these same
+    # packages. GPU markers mean GPUs REQUIRED; a gpu_1 mark here routes
+    # the file through the VRAM-scheduled stage at ~24s/test of pure
+    # scheduling+import overhead against an idle GPU.
+    pytest.mark.gpu_0,
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: The premise doesn't hold on this repo's CI shape: the SGLang **GPU job runs the gpu_0 marker group too** (as its first pytest stage, before the gpu_1 stages), on a CUDA-visible runner. Evidence from this PR's green run (31049499476, amd64 test job): `test_frames_are_pinned_on_cuda` **PASSED** on gw4 in the gpu_0 stage — it still executes with CUDA on every PR, and self-skips only on the CPU-only lanes, which is exactly the gpu_0 contract ("no GPU required"). Moving it to a gpu_1 module would route one test back through the VRAM scheduler at ~24s of overhead for coverage it already has. Resolving with no change.

components/src/dynamo/frontend/tests/test_sglang_processor_unit.py
```diff
@@ -63,7 +63,12 @@
 pytestmark = [
     pytest.mark.unit,
     pytest.mark.sglang,
-    pytest.mark.gpu_1,
+    # gpu_0: CPU-only in the sglang image -- proven by the arm64 CPU test
+    # lane, which already runs `sglang and gpu_0` and imports these same
+    # packages. GPU markers mean GPUs REQUIRED; a gpu_1 mark here routes
+    # the file through the VRAM-scheduled stage at ~24s/test of pure
+    # scheduling+import overhead against an idle GPU.
+    pytest.mark.gpu_0,
```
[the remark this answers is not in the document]
dmitry-tokarev-nv: Confirmed and fixed in 1d8afd5891 — both module-scope tokenizer users now carry `pytest.mark.model("Qwen/Qwen3-0.6B")`, so `predownload_tokenizers` warms the HF cache once per session under its file lock instead of every xdist worker fetching over the network. Good catch on the lane-enrollment side effect: the run passed because HF_TOKEN + network were available, but this repo has been bitten by exactly that one-shot tokenizer-download flake before.

components/src/dynamo/frontend/tests/test_sglang_processor_api.py
```diff
@@ -16,7 +16,12 @@
 pytestmark = [
     pytest.mark.unit,
     pytest.mark.sglang,
-    pytest.mark.gpu_1,
+    # gpu_0: CPU-only in the sglang image -- proven by the arm64 CPU test
+    # lane, which already runs `sglang and gpu_0` and imports these same
+    # packages. GPU markers mean GPUs REQUIRED; a gpu_1 mark here routes
+    # the file through the VRAM-scheduled stage at ~24s/test of pure
+    # scheduling+import overhead against an idle GPU.
```
rmccorm4:
```suggestion
    # gpu_0: CPU-only test
```

components/src/dynamo/frontend/tests/test_sglang_processor_unit.py
```diff
@@ -63,7 +63,17 @@
 pytestmark = [
     pytest.mark.unit,
     pytest.mark.sglang,
-    pytest.mark.gpu_1,
+    # gpu_0: CPU-only in the sglang image -- proven by the arm64 CPU test
+    # lane, which already runs `sglang and gpu_0` and imports these same
+    # packages. GPU markers mean GPUs REQUIRED; a gpu_1 mark here routes
+    # the file through the VRAM-scheduled stage at ~24s/test of pure
+    # scheduling+import overhead against an idle GPU.
```
rmccorm4:
```suggestion
    # gpu_0: CPU-only test
```

components/src/dynamo/frontend/tests/test_sglang_processor_unit.py
```diff
@@ -63,7 +63,17 @@
...
+    pytest.mark.gpu_0,
+    # This file builds a real tokenizer at module scope; declare the model so
+    # predownload_tokenizers warms the HF cache once per session instead of
+    # every xdist worker fetching it over the network (flake surface on the
+    # CPU lanes this file now runs on).
```
rmccorm4:
```suggestion
    # Download tokenizer once per module scope
```

rmccorm4: Can you make this change everywhere? Comment is too verbose

components/src/dynamo/frontend/tests/test_sglang_processor_unit.py
rmccorm4: Can you make this change everywhere? Comment is too verbose

components/src/dynamo/frontend/tests/test_sglang_processor_api.py
rmccorm4: Can you make this change everywhere? Comment is too verbose

rmccorm4 (approved):
frontend test changes LGTM

components/src/dynamo/frontend/tests/test_sglang_processor_unit.py
```diff
@@ -63,7 +63,17 @@
...
+    pytest.mark.model("Qwen/Qwen3-0.6B"),
```
rmccorm4, line 69: optional - not sure this is doing what the comment says it is doing:

---

> [P2] Tokenizer predownload is not activated. The [processor fixture (line 84)](/tmp/dynamo-pr12718-review.6qhfpv/components/src/dynamo/frontend/tests/test_sglang_processor_unit.py:84) and [tool-call fixture (line 46)](/tmp/dynamo-pr12718-review.6qhfpv/components/src/dynamo/frontend/tests/test_sglang_tool_calls.py:46) do not request the non-autouse [predownload_tokenizers fixture (line 562)](/tmp/dynamo-pr12718-review.6qhfpv/tests/conftest.py:562). pytest.mark.model only records model names; it does not run the downloader. Add predownload_tokenizers as a dependency of both tokenizer fixtures.

dmitry-tokarev-nv: Half-confirmed — and the comment was claiming the wrong half, so I fixed the comment (e88c0bd426) rather than the code.

What the marker does **not** do: run any downloader (the P2 is right about that). Its suggested fix also cannot work as written: fixture visibility follows the conftest directory chain, and `tests/conftest.py` is not an ancestor of `components/src/dynamo/frontend/tests/`, so requesting `predownload_tokenizers` from these fixtures would fail with "fixture not found".

What it **does** do: `pytest_collection_modifyitems` in `tests/conftest.py` sweeps `model` markers from every collected item in the session into `config.models_to_download`. Several `tests/frontend/*` gpu_0 modules in this same CI stage (`test_request_tracing_logs.py`, `test_self_host_metadata.py`, …) request `predownload_tokenizers`, which downloads that manifest and then leaves `HF_HUB_OFFLINE=1` set for the remainder of that xdist worker's session. On a worker that has flipped offline, this file's module-scope `get_tokenizer(...)` keeps working only because the marker put Qwen3-0.6B into that predownload; without it, a cache miss under offline mode fails — a scheduling-dependent flake.

Both marker comments now state that mechanism (this file and `test_sglang_tool_calls.py`, which still carried the older, wrong wording).

Chokoyo (approved):
LGTM

nv-tusharma (approved):

PR #12723, opened 2026-08-05
test(deploy): retry K8s API calls on vCluster port-forward blips

nv-tusharma:
Linear: [OPS-8003](https://linear.app/nvidia/issue/OPS-8003/ops-support-0803-0807)

## Problem

Deploy tests intermittently fail **after the deployment is already healthy** (`Ready=True`), because a post-Ready Kubernetes-API call loses the local `kubectl port-forward` tunnel to the vCluster API (`127.0.0.1:8443`):

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host 127.0.0.1:8443
ConnectionRefusedError(111, "Connect call failed ('127.0.0.1', 8443)")
```

## Root cause (evidence-backed)

Traced to run `31031856982`, job `deploy-test-sglang / agg` (attempt 1):

- `18:04:02` — deployment reaches `Ready` (the readiness poll went through `127.0.0.1:8443` fine).
- `18:04:09` — the next API call fails: `httpx.RemoteProtocolError('Server disconnected…')` → `ConnectionRefused`.
- Cleanup shows the surviving `kubectl` port-forward is a **new pid (624)**, not the original (252) — i.e. the runner's local `kubectl port-forward` process died mid-test and the 5s watchdog restarted it, but a few seconds too late.

This is a **local, transient** port-forward death, **not** a shared vCluster backend problem: a sibling job (`deploy-test-sglang / agg_router`) on a different runner, pointing at the *same* backend, made a **successful** tunnel call in the *same second* (`18:04:09.235`), and no other concurrent job saw any `8443` refusal. The run passed on attempt 2. So it does not warrant raising vCluster memory or splitting the shared control plane — the fix is to make the test ride out a sub-watchdog-recovery-window blip.

The pre-Ready readiness poll already tolerates these blips (it retries); the gap is the **post-Ready** API calls, which had no retry.

## Fix

Add a small `_retry_api_call` helper and wrap the two unguarded post-Ready K8s-API sites in `ManagedDeployment`:

- **`get_pods()`** — the `kr8s.get("pods", …)` list call (the observed failure site), materialized inside the retry so a drop surfaces as a retryable error rather than partway through the caller's iteration.
- **`port_forward()`** — the initial `pod.portforward()/start()`, which opens a stream through the API, so a tunnel blip retries instead of returning `None`.

It retries **only** connection-level errors and re-raises everything else immediately, so real failures are never masked. The caught set covers all three signatures seen in the failure:

- `ConnectionRefusedError` / `aiohttp.ClientConnectorError` — both `OSError` subclasses.
- `httpx.RemoteProtocolError('Server disconnected')` / `httpx.ConnectError` — both `httpx.TransportError` subclasses.

## Validation

- `py_compile` + `ruff` clean.
- Standalone logic check of `_retry_api_call`: passes success through, retries a connection error then succeeds, re-raises after exhausting attempts, and does **not** retry/mask a non-connection error (e.g. `ValueError`).
- Exception-hierarchy coverage verified against `httpx` 0.28.1 (`RemoteProtocolError`/`ConnectError` are `TransportError` subclasses; `ConnectionRefusedError` is an `OSError`).

## Optional follow-up (not included)

Tightening the watchdog poll (`5s → 1-2s`) in `.github/actions/connect-vcluster/action.yml` would shrink the recovery gap further, but given the rarity the in-test retry is sufficient on its own.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12723" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when connecting to Kubernetes through a vCluster port-forward.
  * Automatically retries transient connection failures, reducing interruptions during pod listing and port-forward setup.
  * Non-connection errors continue to surface immediately.

### DIFF
--- a/tests/utils/managed_deployment.py
+++ b/tests/utils/managed_deployment.py
@@ -37,6 +37,65 @@ def _get_workspace_dir() -> str:
     return os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 
+# --- Resilience for K8s API calls through the vCluster port-forward ----------
+#
+# Deploy tests reach the (v)cluster API over a local ``kubectl port-forward`` at
+# 127.0.0.1:8443. A watchdog restarts that forward when it drops, but the
+# restart takes a few seconds; a K8s API call that lands in that window fails
+# with a connection error. For post-Ready calls (e.g. get_pods()) that error
+# propagates and fails the whole test even though the deployment is healthy.
+# Retry connection-level errors so a transient tunnel blip does not fail the test.
+def _api_connection_errors() -> tuple[type[BaseException], ...]:
+    # ConnectionError/ConnectionRefusedError/TimeoutError and
+    # aiohttp.ClientConnectorError (kubernetes_asyncio) are all OSError subclasses;
+    # httpx.TransportError (kr8s) is not, so add it explicitly. Built defensively
+    # so a missing optional client library never breaks import.
+    errs: list[type[BaseException]] = [OSError]
+    try:
+        import httpx
+
+        errs.append(httpx.TransportError)
+    except Exception:
+        pass
+    return tuple(errs)
+
+
+_API_CONNECTION_ERRORS = _api_connection_errors()
+
+
+def _retry_api_call(
+    fn,
+    *,
+    what: str,
+    logger: logging.Logger,
+    attempts: int = 5,
+    delay: float = 2.0,
+):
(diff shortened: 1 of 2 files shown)


nv-tusharma:
closing stale PR

PR #12735, opened 2026-08-06
feat(vllm): enable deterministic DP and KV routing

connorcarpenter15:
## Overview

Part 3 of a 4-PR stack. This layer enables deterministic vLLM data-parallel rank selection and KV-event routing without extending the vLLM protocol.

## Details

- Forward Dynamo's selected data-parallel rank as per-request `x-data-parallel-rank` gRPC metadata; omit the metadata when no rank is selected so vLLM retains normal load balancing.
- Require vLLM to honor the `x-data-parallel-rank` gRPC metadata for deterministic DP routing. No protocol capability field is used.
- Support one vLLM frontend hosting the complete DP group starting at rank 0; reject nonzero starting ranks because Control does not report local rank count.
- Require one unique, in-range ZMQ KV-event source per DP rank when KV routing is enabled, and warn before skipping unsupported transports.
- Accept vLLM's valid empty KV-event topic, which means subscribe to all topics.
- Consume the preprocessor's redundant `nvext.token_in=true` marker while preserving strict rejection of unsupported nvext semantics.
- Preserve cache-salt propagation for consistent Dynamo/vLLM block hashes.
- Keep the vLLM mocker aligned with the production transport by parsing and validating the same request metadata.

Multimodal request conversion remains in the top PR.

### Stack

1. [#12734 — split gRPC services and Control discovery](https://github.com/ai-dynamo/dynamo/pull/12734)
2. [#12736 — preserve decode handoff on cancellation](https://github.com/ai-dynamo/dynamo/pull/12736)
3. [#12735 — deterministic DP and KV routing](https://github.com/ai-dynamo/dynamo/pull/12735)
4. [#12214 — multimodal sidecar requests](https://github.com/ai-dynamo/dynamo/pull/12214)

Base: [#12736](https://github.com/ai-dynamo/dynamo/pull/12736)

### Validation

- `cargo test -p dynamo-vllm-sidecar` — 18 unit tests and 1 executable integration test passed
- `cargo test -p dynamo-vllm-mocker` — 12 unit tests and 3 sidecar integration tests passed
- `cargo clippy -p dynamo-vllm-sidecar --all-targets -- -D warnings`
- `cargo clippy -p dynamo-vllm-mocker --all-targets -- -D warnings`
- `cargo test -p dynamo-runtime utils::pool::tests::test_sync_pool_blocking_acquire -- --exact` — passed; verifies the prior CI failure was transient
- Changed-file pre-commit hooks

- `lib/sidecar/vllm/src/client.rs` for request-scoped gRPC metadata.
- `lib/sidecar/vllm/src/model.rs` for the supported full-group DP registration contract.
- `lib/sidecar/vllm/src/engine.rs` for rank selection and KV-event source validation.
- `lib/mocker/servers/vllm/src/server.rs` for mock transport parity.

## Related Issues

### DIFF
--- a/lib/sidecar/vllm/README.md
+++ b/lib/sidecar/vllm/README.md
@@ -31,7 +31,7 @@ The protocol does not support multimodal input, LoRA, encode workers, beam searc
 
 ## Run
 
-Start a vLLM build with the split Inference and Control services. Data-parallel routing requires support for the per-request `x-data-parallel-rank` gRPC metadata:
+Start vLLM with its gRPC listener:
 
 ```bash
 vllm-rs serve Qwen/Qwen3-0.6B --host 127.0.0.1 --grpc-port 50051
--- a/lib/sidecar/vllm/README.md
+++ b/lib/sidecar/vllm/README.md
@@ -31,7 +31,7 @@ The protocol does not support multimodal input, LoRA, encode workers, beam searc
 
 ## Run
 
-Start a vLLM build with the split Inference and Control services. Data-parallel routing requires a build containing [vLLM PR #51178](https://github.com/vllm-project/vllm/pull/51178) or a release that includes it:
+Start a vLLM build with the split Inference and Control services. Data-parallel routing requires support for the per-request `x-data-parallel-rank` gRPC metadata:
 
 ```bash
 vllm-rs serve Qwen/Qwen3-0.6B --host 127.0.0.1 --grpc-port 50051
--- a/lib/sidecar/vllm/README.md
+++ b/lib/sidecar/vllm/README.md
@@ -53,7 +53,7 @@ provided through the environment.
 
 The sidecar discovers `model_id`, the served name, context length, KV capacity, scheduler limits, data-parallel topology, and KV-event sources through `vllm.Control`. `model_id` must be readable locally or fetchable by Dynamo for tokenization and chat templates. Parser defaults are not advertised because the current inference protocol cannot preserve all parser-related request semantics.
 
-Control reports the global data-parallel size, the rank hosted by the connected frontend, and whether explicit rank routing is supported. Dynamo forwards the selected rank on each generation request and uses discovered KV-event sources for KV-aware routing. Startup fails for data-parallel deployments when the server does not advertise explicit rank routing.
+The sidecar currently supports one vLLM frontend hosting the complete data-parallel group starting at rank 0. Control reports the global size and whether explicit rank routing is supported; Dynamo forwards the selected rank on each generation request. Partial and hybrid rank ownership are unsupported because the protocol does not report the locally hosted rank count, and a nonzero starting rank is rejected. When KV routing is enabled, Control must return one unique ZMQ event source for every rank in the group.
 
 Aggregated serving is the default. Set the existing `--disaggregation-mode` to `prefill` or `decode` only for non-aggregated deployments; the current Control API does not report engine role.
 
--- a/lib/sidecar/vllm/README.md
+++ b/lib/sidecar/vllm/README.md
@@ -25,13 +25,13 @@ It is a standalone Rust executable.
 - Token and text requests through Dynamo preprocessing
 - Sampling, stop conditions, structured output, logprobs, cache options, and priority
 - Opaque `kv_transfer_params` handoff
+- Data-parallel rank routing and KV-event source discovery
 
-The initial protocol does not support multimodal input, LoRA, KV-aware data
-parallel routing, encode workers, beam search, or `n > 1`.
+The protocol does not support multimodal input, LoRA, encode workers, beam search, or `n > 1`.
 
 ## Run
 
-Start vLLM with its released gRPC listener:
+Start a vLLM build with the split Inference and Control services and explicit data-parallel-rank capability used by the vendored protocol:
 
 ```bash
 vllm-rs serve Qwen/Qwen3-0.6B --host 127.0.0.1 --grpc-port 50051
@@ -51,9 +51,9 @@ dynamo-vllm-sidecar \
 Use `VLLM_GRPC_ENDPOINT` instead of `--vllm-endpoint` when the endpoint is
 provided through the environment.
 
-The sidecar discovers `model_id`, the served name, context length, KV capacity, and scheduler limits through `vllm.Control`. `model_id` must be readable locally or fetchable by Dynamo for tokenization and chat templates. Parser defaults are not advertised because the current inference protocol cannot preserve all parser-related request semantics.
+The sidecar discovers `model_id`, the served name, context length, KV capacity, scheduler limits, data-parallel topology, and KV-event sources through `vllm.Control`. `model_id` must be readable locally or fetchable by Dynamo for tokenization and chat templates. Parser defaults are not advertised because the current inference protocol cannot preserve all parser-related request semantics.
 
-Data-parallel registration is omitted because Control reports global topology, not the rank range hosted by the connected frontend.
+Control reports the global data-parallel size, the rank hosted by the connected frontend, and whether explicit rank routing is supported. Dynamo forwards the selected rank on each generation request and uses discovered KV-event sources for KV-aware routing. Startup fails for data-parallel deployments when the server does not advertise explicit rank routing.
 
 Aggregated serving is the default. Set the existing `--disaggregation-mode` to `prefill` or `decode` only for non-aggregated deployments; the current Control API does not report engine role.
 
--- a/lib/sidecar/vllm/README.md
+++ b/lib/sidecar/vllm/README.md
@@ -31,7 +31,7 @@ The protocol does not support multimodal input, LoRA, encode workers, beam searc
 
 ## Run
 
-Start a vLLM build with the split Inference and Control services and explicit data-parallel-rank capability used by the vendored protocol:
+Start a vLLM build with the split Inference and Control services. Data-parallel routing requires a build containing [vLLM PR #51178](https://github.com/vllm-project/vllm/pull/51178) or a release that includes it:
 
 ```bash
 vllm-rs serve Qwen/Qwen3-0.6B --host 127.0.0.1 --grpc-port 50051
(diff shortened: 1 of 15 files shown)


lib/sidecar/vllm/src/model.rs
```diff
@@ -81,6 +92,10 @@ impl DiscoveredModel {
                 total_kv_blocks: nonzero(self.server.total_kv_blocks),
                 max_num_seqs: nonzero(self.server.max_running_requests),
                 max_num_batched_tokens: nonzero(self.server.max_batched_tokens),
+                data_parallel_size: parallelism
+                    .and_then(|parallelism| nonzero(parallelism.data_parallel_size)),
+                data_parallel_start_rank: parallelism
+                    .map(|parallelism| parallelism.data_parallel_rank),
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `34633780b4` without extending the vLLM protocol. The sidecar now supports the topology exercised by this PR: one vLLM frontend hosting the complete DP group starting at rank 0. Discovery rejects a nonzero reported starting rank, Dynamo registers `[0, global_dp_size)`, and KV routing requires one unique in-range ZMQ source for every rank. The fixture now models DP2 with ranks 0 and 1 instead of advertising the invalid range 2 through 5. Partial and hybrid frontend ownership remain explicitly unsupported until Control can describe local ownership.

lib/sidecar/vllm/src/convert.rs
```diff
@@ -92,10 +102,51 @@ pub(crate) fn build_generate_request(
...
+                ));
+            }
+        }
+        nvext.is_empty()
+    };
+    if remove_nvext {
+        extra.remove("nvext");
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `34633780b4`. Request conversion now consumes and validates `nvext.token_in=true` as redundant preprocessing provenance already represented by `PreprocessedRequest.token_ids`. The existing aggregate conversion test includes that marker and passes. Other nvext fields such as `extra_fields` and `metadata_upload` remain rejected because they request semantics the gRPC path does not preserve.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -291,6 +293,66 @@ impl LLMEngine for VllmSidecarEngine {
...
+                        "GetKvEventSources returned a ZMQ source without an endpoint",
+                    ));
+                }
+                Ok(KvEventSource::Zmq {
+                    endpoint: zmq_connect_endpoint(&source.endpoint, &self.endpoint)?,
+                    topic: source.topic,
+                    dp_rank,
+                })
+            })
+            .collect()
```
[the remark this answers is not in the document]
connorcarpenter15: Fixed in `34633780b4`. KV-event discovery now warns when skipping unsupported transports, rejects duplicate ranks, rejects ranks outside the advertised global DP range, and requires exact source coverage for every DP rank when KV routing is enabled. Empty topics remain valid and are preserved.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -291,6 +293,66 @@ impl LLMEngine for VllmSidecarEngine {
...
+    endpoint: &str,
+    grpc_endpoint: &GrpcEndpoint,
+) -> Result<String, DynamoError> {
+    let port = endpoint
+        .strip_prefix("tcp://*:")
+        .or_else(|| endpoint.strip_prefix("tcp://0.0.0.0:"))
+        .or_else(|| endpoint.strip_prefix("tcp://[::]:"));
+    let Some(port) = port else {
+        return Ok(endpoint.to_string());
+    };
```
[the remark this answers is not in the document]
connorcarpenter15: Confirmed; no code change is appropriate here. The rewrite is intentionally limited to wildcard TCP bind addresses that cannot be used as subscriber connect addresses: `tcp://*:`, `tcp://0.0.0.0:`, and `tcp://[::]:`. Concrete TCP endpoints and IPC/in-process endpoints are explicit topology supplied by vLLM and must be preserved verbatim. `tcp://::` is not a canonical IPv6 endpoint because the address must be bracketed. Rewriting explicit endpoints would corrupt valid same-namespace and operator-configured deployments.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -319,6 +323,88 @@ impl LLMEngine for VllmSidecarEngine {
...
+
+    let grpc_url = url::Url::parse(grpc_endpoint.as_str()).map_err(|error| {
```
tanmayv25: **Simplicity: re-parsing an already-validated `GrpcEndpoint` adds a `url` dependency and two dead error branches.**

`GrpcEndpoint::parse` (`lib/sidecar/common/src/endpoint.rs`) already parses the endpoint via `url::Url` and guarantees a host exists (construction rejects when host is `None`). Re-parsing here makes the parse-error branch and the `None => ...` no-host branch unreachable — dead code defending against states the type already rules out. It also pulls `url` into `lib/sidecar/vllm/Cargo.toml` (not previously a dependency) solely to redo work sidecar-common already did and owns.

**Suggestion:** expose a `host()` / `zmq_connect_host()` accessor on `GrpcEndpoint` in sidecar-common and call it here, dropping the re-parse, the two impossible branches, and the new `url` dependency on the vllm crate.

_— from an automated multi-agent code-review pass (expert lenses + adversarial verification over this PR's diff)._

lib/sidecar/vllm/src/engine.rs
```diff
@@ -319,6 +323,88 @@ impl LLMEngine for VllmSidecarEngine {
...
+        }
+        Ok(sources)
+    }
+}
+
+fn zmq_connect_endpoint(
+    endpoint: &str,
+    grpc_endpoint: &GrpcEndpoint,
+) -> Result<String, DynamoError> {
+    let port = endpoint
```
tanmayv25, line 386: **Test coverage: the wildcard-rewrite path (the actual production case) is untested.**

vLLM publishes its ZMQ KV-event socket on a wildcard bind (`tcp://*:PORT`, `tcp://0.0.0.0:PORT`, `tcp://[::]:PORT`), which is exactly why this function strips those prefixes and substitutes the gRPC host (with IPv6 bracketing). But `FakeVllm::get_kv_event_sources` returns literal `tcp://127.0.0.1:PORT`, so every test hits only the pass-through branch (`Ok(endpoint.to_string())`) — the strip_prefix + host-substitution + `[{host}]` IPv6 logic never runs. The KV-event-sources test asserts only `len`, `dp_rank`, and empty topic, never the resolved endpoint string. A regression in host substitution would silently produce a wrong ZMQ connect endpoint and misroute KV events with no test catching it.

**Suggestion:** have `FakeVllm` emit at least one wildcard bind (e.g. `tcp://*:20081` or `tcp://[::]:20082`) and assert the resolved `KvEventSource::Zmq` endpoint equals `tcp://127.0.0.1:20081` (matching the gRPC host), including an IPv6 gRPC-host case.

_— from an automated multi-agent code-review pass (expert lenses + adversarial verification over this PR's diff)._

tanmayv25 (commented):
Automated multi-agent code-review pass over this PR's diff (framing + expert lenses, each finding adversarially verified). Findings below are all **minor / non-blocking**.

lib/sidecar/vllm/src/engine.rs
connorcarpenter15: Addressed in `7b775ccb73`. `GrpcEndpoint` now retains the validated host and exposes it in URI-authority form, including IPv6 brackets. The vLLM sidecar reuses that value, so the duplicate parse, impossible parse/no-host branches, direct `url` dependency, and corresponding lockfile entry are removed.

lib/sidecar/vllm/src/engine.rs
connorcarpenter15: Addressed in `7b775ccb73`. The existing fake now advertises `tcp://*:PORT`, and the aggregate test asserts the resolved endpoints are `tcp://127.0.0.1:20081` and `tcp://127.0.0.1:20082`, exercising the production wildcard-substitution path. The existing `GrpcEndpoint` normalization test now also verifies authority-safe IPv6 bracketing. The full sidecar-common and vLLM-sidecar suites pass.

karen-sy (approved):
**Findings**

- **P1** `lib/sidecar/vllm/src/engine.rs:133,371`  
  KV routing is enabled unconditionally, but an empty `GetKvEventSources` response is treated as a protocol error. At the vendored vLLM commit, Control intentionally returns no source when `kv_events_config` is absent, disabled, or uses a non-ZMQ publisher. That makes an otherwise valid non-KV-event vLLM sidecar fail worker startup. Return an empty source list in the genuinely no-source case so the worker’s existing KV-routing opt-out path applies; retain the strict all-ranks check when sources are advertised. Add a no-KV-events startup test.  
  Upstream behavior: [[vLLM Control source](https://github.com/vllm-project/vllm/blob/3d1f5cee1552b8208f3009c75f8bc856f27e0eff/rust/src/server/src/grpc/control.rs#L117-L121)](https://github.com/vllm-project/vllm/blob/3d1f5cee1552b8208f3009c75f8bc856f27e0eff/rust/src/server/src/grpc/control.rs#L117-L121).

- **P2** `lib/sidecar/vllm/src/engine.rs:181,333`  
  `start()` returns the freshly observed DP topology, but `kv_event_sources()` still validates against the bootstrap-time `self.model`. `ensure_same_identity()` does not compare parallelism. If the endpoint lands on a same-model frontend with a different DP size between bootstrap and startup, Dynamo registers the new range but source validation uses the old size. Reject topology changes during startup, or retain the observed topology in the engine.


----
Noting some flags above from AI review, but generally looks good!

lib/sidecar/vllm/README.md
```diff
@@ -25,13 +25,13 @@ It is a standalone Rust executable.
...
+- Data-parallel rank routing and KV-event source discovery
 
-The initial protocol does not support multimodal input, LoRA, KV-aware data
-parallel routing, encode workers, beam search, or `n > 1`.
+The protocol does not support multimodal input, LoRA, encode workers, beam search, or `n > 1`.
 
 ## Run
 
-Start vLLM with its released gRPC listener:
+Start a vLLM build with the split Inference and Control services. Data-parallel routing requires support for the per-request `x-data-parallel-rank` gRPC metadata:
```
tanmayv25: Too much details.

tanmayv25 (approved):

PR #12741, opened 2026-08-06
fix(frontend): emit SGLang stream role once

xianlubird:
## Summary

- Emit `delta.role` only on the first choice produced by the SGLang streaming postprocessor.
- Preserve the initial `assistant` role when generation finishes before producing any content.
- Cover plain-text streams, finish-only streams, and reasoning-parser streams with focused unit tests.

## Background

PR #12639 established the chat-stream contract at the HTTP boundary: the first emitted delta for a choice carries `role: assistant`, while later deltas omit the role. The SGLang Python postprocessor builds streaming choice dictionaries outside the native Rust postprocessing path, so it still attached the role to every content-bearing delta and relied on the HTTP layer to remove the repeats.

That arrangement also left an edge case at the source. When SGLang finishes immediately without generating a token, the postprocessor returns a finish-only choice with an empty delta. The HTTP deduplication logic can remove duplicate roles, but it cannot recover a role that was never emitted.

This change tracks role emission in the per-request SGLang postprocessor and attaches the role only when a choice is actually returned. Buffered UTF-8 sequences and parser-buffered output do not consume the initial role. If the first returned choice is a finish-only choice, it still carries `role: assistant`; any later content or terminal choices omit it.

## Validation

- `black --check` on the modified Python files
- `isort --check-only` on the modified Python files
- `ruff check` on the modified Python files
- `python3 -m py_compile` on the modified Python files
- `git diff --check`

The targeted pytest module was not run locally because the current virtual environment does not include pytest or the SGLang test dependencies.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming responses so the assistant role appears only once, on the first response chunk.
  * Ensured finish-only responses return empty content deltas while preserving the initial assistant role.
  * Corrected role handling for fast text, reasoning, and tool-call streaming responses.

* **Tests**
  * Added coverage for role emission and finish-only streaming behavior.

### DIFF
--- a/components/src/dynamo/frontend/sglang_prepost.py
+++ b/components/src/dynamo/frontend/sglang_prepost.py
@@ -979,6 +979,7 @@ class SglangStreamingPostProcessor:
         # incomplete byte-fallback sequence.
         self._decode_context_ids = list((prompt_token_ids or [])[-5:])
         self._pending_decode_ids: list[int] = []
+        self._has_emitted_role: bool = False
         # Tool call accumulation.  SGLang's streaming parser returns
         # deltas (name in one chunk, argument fragments across subsequent
         # chunks).  However, the base detector processes at most one event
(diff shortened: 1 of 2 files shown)


rmccorm4 (approved):
LGTM, thanks @xianlubird

PR #12746, opened 2026-08-06
fix(ci): retry flaky documentation link checks

xianlubird:
## Summary

- Run the full lychee documentation link check one additional time after a failed first attempt.
- Remove failed entries from `.lycheecache` before retrying so the second attempt performs real network requests instead of replaying cached errors.
- Reuse one anchored argument block for both lychee attempts.
- Keep the retry strict: a second failure is propagated normally and still fails the workflow.

## Background

PR #12741 exposed two independent transient failures in the repository-wide documentation link check. The first run reported HTTP 500/502 responses for existing GitHub pull request links. A new commit retriggered the workflow, but the next run failed on a different existing URL—the Kubernetes Ingress documentation—with a connection error. That URL returned HTTP 200 immediately when checked again.

The workflow already configures lychee with four request retries and a 30-second request timeout. However, lychee v0.24.2 does not classify ordinary connection errors as retryable, so `--max-retries` does not protect this case. Request-level retries can also be exhausted when an external host remains temporarily unhealthy.

## Implementation

The first lychee action records its real exit code without ending the job immediately. When that code is nonzero, the workflow parses the cache status field, keeps only statuses accepted by this workflow (`2xx`, `403`, and `429`), and runs the same complete lychee command again. Both attempts reference the same YAML-anchored argument block. The retry uses the action's normal `fail: true` behavior, so persistent broken links remain blocking.

This does not add any URL to `.lycheeignore`, accept connection errors, or use `continue-on-error` to make the job green.

## Validation

- `pre-commit run check-yaml --files .github/workflows/docs-link-check.yml`
- `actionlint v1.7.12 .github/workflows/docs-link-check.yml`
- cache status filtering with representative `2xx`, `403`, `429`, `404`, and comma-containing URL records
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved documentation link checks by retrying after clearing unsuccessful cached results.
  * Workflows now fail only when link issues persist after the retry.


xianlubird:
Looks like the SGLang failure is unrelated to this PR—the tests all passed, but the self-hosted runner hit a container error afterward. 

The new lychee retry worked as expected: the first check timed out on an external link, then the retry passed. 
A rerun of the failed SGLang job should be enough.

.github/workflows/docs-link-check.yml
```diff
@@ -25,7 +25,43 @@ jobs:
           key: cache-lychee-${{ github.run_id }}
           restore-keys: cache-lychee-
 
-      - name: Check documentation links with lychee
+      - name: Check documentation links with lychee (first attempt)
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: >-
```
nv-anants: thanks @xianlubird! can you try adding an anchor link here for reuse in the retry step
```
args: &lychee_args >-
              --cache
```

Then in retry step,
```
args: *lychee_args
```

xianlubird: Thanks! 
I updated the workflow to define the first argument block with `&lychee_args` and reuse it as `*lychee_args` in the retry step.
 I also reran check-yaml and actionlint v1.7.12 successfully.

xianlubird:
The `sglang-runtime / Test cuda13.0, amd64 (push)` failed and the failure is unrelated to the content of this PR.

nv-anants (approved):

Issue #12430, opened 2026-07-30
[BUG]: Performance Shim for Decode Fluctuates heavily.


plongiaZon:
### Describe the Bug

As Planner is subjected to an increase in RPS, the `engine_rps` calculated in throughput-based scaling mode fluctuates heavily. Planner overestimates the required number of decode workers, and excessively scales them over a small increase in RPS. Afterwards, Planner spends many ticks scaling decode workers down.

### Steps to Reproduce

Run a disaggregated `DynamoDeploymentGraph` with Planner enabled (in SLA mode).

### Expected Behavior

Planner scales decode workers proportional to the increase in RPS. In particular, throughput-based scaling mode's central formula: `desired = demand_rps / engine_rps`, should increase due to an increase in `demand_rps`, not `engine_rps`.

### Actual Behavior

Planner scales decode workers excessively. However, `engine_rps` drops severely, and the desired decode workers spike drastically.

### Environment

Helm release: dyn-dynamo-system-0, chart dynamo-platform-1.3.0, app version 1.3.0
Operator image: nvcr.io/nvidia/ai-dynamo/kubernetes-operator:1.3.0

### Additional Context

Cause:
When new workers join, their first forward passes run slower than steady state while draining a migrated KV/request backlog. These cold-start samples were admitted into the throughput path's capacity regression model with no filtering, collapsing the per-engine capacity estimate.

Suggested Fix:
The FPM admission gate in `rust_adapter.py` (`_is_supported_fpm`) only checks version — add a wall_time floor/ceiling, engine-age/warmup gate,`queued_kv` guard, or an outlier rejection.

sachalmalick:
"When new workers join, their first forward passes run slower than steady state while draining a migrated KV/request backlog"

Was able to repro on vLLM 0.23 not just for new requests - requests should not be scheduled by the engine until remote KV is pulled however in that version of vLLM + NIXL - prep work we do in NixlPullConnector, start_load_kv to initiate KV fetching for requests that are *waiting* blocks the execution of requests in the active batch causing unusually high wall clock times for batch shapes with can skew the AIC models.

John Thompson investigated and confirmed it was fixed in the latest vLLM + NIXL versions - that being said I did add changeable bounds on tuning - https://github.com/ai-dynamo/aiconfigurator/pull/1502/changes
but they should not be needed for this particular issue.

PR #12751, opened 2026-08-06
fix(planner): add optional AIC correction-factor cap

sachalmalick:
## What

- Add an opt-in `aic_max_correction_factor` Planner setting (`>= 1.0`, disabled by default).
- When native AIC is active and the cap is configured, retain an identical never-tuned native model as the correction baseline.
- Clamp only upward FPM wall-time observations before they reach AIC tuning; downward corrections remain unchanged.
- Report the configured cap, whether it is active, and AIC's observed maximum correction in diagnostics.
- Document the new setting and cover configuration, fallback, adapter, native-AIC, replay-digest, and repeated-outlier behavior.

## Why

Related to #12430.

A transient forward-pass outlier can otherwise teach native AIC a very large correction for the matching workload region, which depresses the Planner's capacity estimate long after the stall has passed. This is especially damaging when several similar outliers arrive in sequence.

The cap is computed from a separate never-tuned native estimate. Using the already-corrected estimate as the baseline would compound the bound across observations (for example, 2x -> 4x -> 8x) and would not provide an absolute ceiling.

This is intentionally a mitigation rather than a universal default: deployments can choose a factor appropriate to their tolerance for genuine performance shifts. Regression fallback and the existing unbounded behavior are unchanged when the setting is unset.

## Validation

- `99 passed in 15.55s` across:
  - Planner AIC adapter unit tests
  - engine-query unit tests
  - Planner config unit tests
  - offline replay FPM tests
  - native AIC integration tests
- Ruff formatting/checks passed for all touched Python files.
- `git diff --check` passed.
- Controlled replay of 12 matching outliers learned an approximately 90.76x correction without a cap; configured 1.25x, 1.5x, and 2.0x caps bounded the stored correction at the configured factor.

## Runtime impact

The extra native model and baseline estimate are created and used only when the cap is configured and native AIC is available. Normal capacity/latency queries do not query the baseline model; capped tuning adds one native estimate per FPM iteration.

## Summary by CodeRabbit

* **New Features**
  * Added an optional Planner setting to cap native AIC online correction factors.
  * Supports finite values of at least 1.0; leaving it unset preserves uncapped behavior.
  * Planner diagnostics now report correction-cap configuration and observed correction factors.
  * Configuration changes trigger updated model initialization.

* **Documentation**
  * Documented the new setting, validation rules, defaults, behavior, and runtime considerations.

* **Tests**
  * Added coverage for correction caps, validation, fallback behavior, diagnostics, and configuration changes.

### DIFF
--- a/components/src/dynamo/planner/core/perf_model/engine_query.py
+++ b/components/src/dynamo/planner/core/perf_model/engine_query.py
@@ -194,6 +194,8 @@ class AicCoreEnginePerfModel:
         limits: EnginePerfLimits,
         attention_dp_size: int,
         max_observations: int,
+        uncorrected_native_model: Optional[Any] = None,
+        max_correction_factor: Optional[float] = None,
     ) -> None:
         if worker_type not in ("prefill", "decode", "aggregated"):
             raise ValueError(f"invalid worker_type {worker_type!r}")
@@ -201,7 +203,18 @@ class AicCoreEnginePerfModel:
             raise ValueError(
                 f"attention_dp_size must be positive, got {attention_dp_size}"
             )
+        if max_correction_factor is not None and (
+            not math.isfinite(max_correction_factor) or max_correction_factor < 1.0
+        ):
+            raise ValueError(
+                "max_correction_factor must be finite and >= 1.0, "
+                f"got {max_correction_factor}"
+            )
+        if uncorrected_native_model is not None and max_correction_factor is None:
+            raise ValueError("uncorrected_native_model requires max_correction_factor")
         self._model = model
+        self._uncorrected_native_model = uncorrected_native_model
+        self._max_correction_factor = max_correction_factor
         self._worker_type = worker_type
         self._limits = limits
         self._attention_dp_size = attention_dp_size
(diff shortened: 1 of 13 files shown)


sachalmalick:
Closing in favor of [ai-dynamo/aiconfigurator#1500](https://github.com/ai-dynamo/aiconfigurator/pull/1500). The absolute correction bound belongs in AIC, where tuning already has the raw native denominator; implementing it there avoids the shadow model and extra estimate used by this draft. A small Dynamo configuration pass-through plus an AIC version bump can follow once that change is released.

components/src/dynamo/planner/core/perf_model/engine_query.py
```diff
@@ -216,28 +229,70 @@ def best_available(
...
+                if native_ms < 0.0:
+                    raise ValueError(
+                        "uncorrected native AIC forward-pass estimate must be "
+                        f"non-negative, got {native_ms}"
+                    )
+                max_wall_time_s = native_ms * max_correction_factor / 1000.0
+                for metrics in payload:
+                    wall_time = metrics["wall_time"]
+                    if wall_time > max_wall_time_s:
+                        metrics["wall_time"] = max_wall_time_s
```
liyuanzhe1991, line 294: Clamping is currently silent and `planner_correction_cap` has no production consumer, so an operator can't tell outlier suppression from a masked genuine regression — the raw magnitude is destroyed at this line, so it can't be reconstructed later. Suggest a rate-limited warning with the observed/ceiling ratio plus a clamped-observation counter in diagnostics; cap saturation is exactly the "new pollution source, go find it" signal.

docs/fern/pages/developer-guide/knowledge-base/modular-components/planner/planner-guide.md
```diff
@@ -128,6 +128,7 @@ spec:
 |-------|------|---------|-------------|
 | `load_adjustment_interval_seconds` | int | `5` | Seconds between FPM tuning updates and load-based scaling decisions. Even when only throughput scaling is enabled, live FPM observations are fed into the perf model at this interval. Must be shorter than `throughput_adjustment_interval_seconds`. |
 | `max_num_fpm_samples` | int | `64` | Maximum retained FPM observations for online tuning or regression. |
+| `aic_max_correction_factor` | float or null | `null` | Absolute upper bound on native AIC online correction. Set a finite value greater than or equal to `1.0`; `null` preserves unbounded AIC correction. The Planner bounds each tuning observation against an identical never-tuned native estimate, so the bound applies only to native AIC and does not clamp downward corrections. Enabling it maintains a second native model and runs one native estimate per tuned FPM iteration. The second model adds memory and can add one-time latency to the first capped Planner tuning tick; it does not run on the inference engine thread. |
```
liyuanzhe1991, line 131: Docs cover the mechanics but not what an operator needs to use this safely: (1) the trade-off — a genuine regression beyond the cap stays under-corrected for as long as it lasts; (2) sizing — observe `observed_aic_max_factor` during healthy operation and set the cap comfortably above it, rather than guessing. One short paragraph on both would do.

PR #12753, opened 2026-08-06
fix(docs): stop Fern bundling commented-out component imports

dagil-nvidia:
## Summary

The `Preview or publish docs` job has been failing on `main` and on every docs-touching PR with:

```
Failed to bundle third-party imports in fern/components/RecipeStyles.tsx:
rolldown exited with code 127 ... sh: 1: rolldown: not found
```

The named component rotated run to run — `RecipeStyles.tsx`, `ReferenceStyles.tsx`, `TerminalDemo.tsx` — and **none of the three has an npm dependency**. That is the tell: the trigger was not a real import.

**Root cause.** Fern collects a component's third-party dependencies with a plain-text regex over the whole source file. It does not parse the source, so it does not skip comments. Each of the three headers documented its own usage with a literal example statement naming `@/components/<Name>`. Fern read that comment as a real dependency; `@/components` is outside its allowlist (`react`, `react-dom`, `@mdx-js/react`, `next`), so it shelled out to `npx rolldown` to bundle it.

The `docs-website` branch ships no `package.json` next to `fern/`, so that bundle could never resolve. It surfaced two ways, which is what made it look flaky rather than broken:

- exit `127` — `rolldown: not found`
- exit `1` — `ERR_MODULE_NOT_FOUND` on a half-unpacked `~/.npm/_npx/…/rolldown/dist/shared/binding-*.mjs`

Fern bundles the components concurrently, so several `npx` invocations race on one shared cache directory. A retry that happens to win the race goes green, which is why re-runs sometimes "fixed" it.

**Fix.** Document the specifier in prose in all three headers instead of writing a literal statement, and add `docs/fern/scripts/check_component_imports.py` — a port of Fern's own matcher (regex, allowlist, and package-root reduction taken from the pinned CLI) — wired into pre-commit so a re-added example is caught before merge instead of on the publish.

No page content, CSS, or component behaviour changes; the edits are comment-only.

## Validation

- Ported Fern's detector and ran it against `docs/fern/components` at `main`: flagged exactly `RecipeStyles.tsx:17`, `ReferenceStyles.tsx:23`, `TerminalDemo.tsx:15,28` — the same three files CI named, and no others.
- After the fix: `checked 33 component source(s): no unbundleable specifiers`.
- Replayed the workflow's components sync (`fern-docs.yml`, the `rm -rf` + `cp -r` into `docs-checkout/fern/components`) against a real `origin/docs-website` worktree and re-ran the check on the composed tree — clean, 33/33. Confirmed `docs-website` carries exactly these 33 files, so the branch copy cannot drift.
- Guard verified in both directions: it exits 1 with a located finding on the pre-fix `RecipeStyles.tsx`, and 0 on the fixed tree.
- `--test` self-test covers 12 cases, including allowlisted, scoped-allowlisted, subpath, relative, dynamic `import()`, `require()`, re-export, the commented-example regression, and prose that merely contains the word.
- `pre-commit run --files …` green on all five changed files.

Not fixed here, and tracked separately: `main` is also red on the later `Verify published pages carry their component CSS` step, which 404s on `/dynamo/dev/recipes/feature-benchmarks/llama-3-3-70b-topology`. That is a different failure in a different step — `Publish Docs` itself succeeds on `main`.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/dynamo/pull/12753" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to use styling and terminal demonstration components in applicable pages.
  * Added guidance for component imports, page placement, and runtime asset loading.

* **Quality Improvements**
  * Added automated validation to detect unsupported component dependencies and provide actionable error details.
  * Included self-checks to help ensure dependency validation remains reliable.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -111,6 +111,20 @@ repos:
       language: system
       files: ^docs/fern/scripts/check_published_styles\.py$
       pass_filenames: false
+    - id: check-component-imports
+      name: Check mdx-components carry no unbundleable specifiers
+      entry: python3 docs/fern/scripts/check_component_imports.py
+      language: system
+      # Scan the whole components/ tree, not just the staged files: Fern
+      # bundles them together and reports whichever it reaches first.
+      files: ^docs/fern/(components/|scripts/check_component_imports\.py$)
+      pass_filenames: false
+    - id: check-component-imports-selftest
+      name: Self-test the component specifier matcher
+      entry: python3 docs/fern/scripts/check_component_imports.py --test
+      language: system
+      files: ^docs/fern/scripts/check_component_imports\.py$
+      pass_filenames: false
     - id: pytest-marker-report
       name: Report pytest markers (static + inherited)
       entry: python3 tests/report_pytest_markers.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -117,7 +117,8 @@ repos:
       language: system
       # Scan the whole components/ tree, not just the staged files: Fern
       # bundles them together and reports whichever it reaches first.
-      files: ^docs/fern/(components/|scripts/check_component_imports\.py$)
+      # fern.config.json is in scope so a CLI bump trips the matcher-drift check.
+      files: ^docs/fern/(components/|fern\.config\.json$|scripts/check_component_imports\.py$)
       pass_filenames: false
     - id: check-component-imports-selftest
       name: Self-test the component specifier matcher
--- a/docs/fern/components/RecipeStyles.tsx
+++ b/docs/fern/components/RecipeStyles.tsx
@@ -11,12 +11,21 @@
  * constraint, CSS injected this exact way.
  *
  * Server component (no "use client"); registered via docs.yml
- * `experimental.mdx-components: ./components`. IMPORT it (ambient use is
- * unsupported — renders "Unsupported JSX tag"); the @/ prefix resolves to the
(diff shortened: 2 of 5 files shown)


dagil-nvidia:
Closing in favor of #12388, which fixes the same problem and does more.

Both PRs independently rewrote the doc-comment import examples in \`RecipeStyles.tsx\`, \`ReferenceStyles.tsx\` and \`TerminalDemo.tsx\`, and both added a \`docs/fern/scripts/check_component_imports.py\` ported from the same fern-api 5.80.2 bundler. #12388 predates this one by eight days; I did not know it existed when I opened this.

I verified the two are equivalent before choosing, rather than assuming the newer one was better:

- Run against \`origin/main\`, both checkers report the same four specifiers — \`@/components/RecipeStyles\`, \`@/components/ReferenceStyles\`, \`asciinema-player\`, \`@/components/TerminalDemo\`.
- Each checker passes its own tree, and each also passes the *other* PR's components. Neither fix is incomplete relative to the other.

#12388 additionally wires the hook into \`.github/workflows/pre-merge.yml\` and relocates the usage examples into \`docs/fern/components/README.md\`, where Fern's scan cannot reach them.

Two things from this PR are genuinely better and I will port them onto #12388 rather than lose them: the finding output carries file:line instead of naming only the file, and the remediation text explains the CDN-at-runtime path for a real dependency.

PR #10126, opened 2026-05-29
test(ci): VRAM-aware GPU test scheduler + CPU cap + 3x timeout retune

dmitry-tokarev-nv:
## Problem

The GPU-parallel test runner (`tests/utils/pytest_parallel_gpu.py`) ordered tests by raw `@pytest.mark.timeout` and greedily filled slots front-to-back. In the reference vllm run (`job-log.txt`: 239 tests, 8 slots, 1×22 GiB GPU, **1764s**):

- 213/239 tests are **zero-VRAM** unit tests (need the `gpu_1` *container* for `import vllm`, but `profiled_vram_gib(0)`) with the **default 600s timeout**, so they sorted ahead of the real GPU tests and grabbed all 8 slots.
- **GPU memory sat idle (~0.5/22 GiB) for ~570s**, then the 26 real GPU tests serialized on the tail — **largest last** (the 12 & 13 GiB tests started at elapsed 1474s / 1666s).

## Fix

**1. Scheduler** (`pytest_parallel_gpu.py`): `est_duration = timeout/3`; `_priority_key` orders VRAM tests first (longest est_duration → largest VRAM), zero-VRAM fillers last; `_select_launches` (pure, unit-tested) best-fit **packs a big test with smaller ones**, fillers backfill spare slots (bypassing the budget gates), and a blocked high-priority test **reserves** space (backfill capped at `cap − required`). `pynvml` import made lazy so the pure logic imports/tests on CPU-only runners.

**2. CPU-aware concurrency cap**: xdist resolves `-n auto` from `os.cpu_count()`, which reports HOST cores and ignores Docker `--cpus` (a CFS quota). `effective_cpu_budget()` (NUM_CPUS env → cgroup v2 `cpu.max` → cgroup v1 → `os.cpu_count()`) caps `num_slots`, so the filler backfill can't oversubscribe CPU.

**3. Timeouts → 3× observed runtime** (raise-only — never tightens an existing value):
- **vllm**: re-tuned from the *new-scheduler* run, because running GPU tests concurrently makes them 8–69% slower than the baseline the first pass used (`indexers_sync` 410→690, `mm_agg_gemma4` 580→670, lmcache/serve →600-640, kvbm →410/420, …).
[...]

PR #12762, opened 2026-08-06
fix(bindings): release the GIL during KvRouter initialization

nv-tusharma:
## Summary

`KvRouter.__new__` blocks until `min_initial_workers` register. That wait is unbounded by design (`lib/llm/src/kv_router.rs:409`, `startup_watch.wait_for(..)`), which is fine as long as the caller can impose its own deadline — but a caller can only do that if the GIL is free.

The binding held it. `fn new()` took no `Python` token and called `block_on` directly, so PyO3 kept the GIL for the entire wait:

```rust
// lib/bindings/python/rust/llm/kv.rs:1856
 fn new(
+    py: Python<'_>,
     endpoint: &Endpoint,
     ...
     let runtime = pyo3_async_runtimes::tokio::get_runtime();
-    runtime.block_on(async move {
+    py.allow_threads(|| runtime.block_on(async move {
```

Every other `block_on` in this file (`:586`, `:600`) already wraps in `py.allow_threads`. This one was the outlier.

When a worker died before registering, the consequences compounded:

- `Thread.join(timeout=2)` could not resume — returning to bytecode needs the GIL — so the 120s guard and the worker-liveness poll in `tests/router/common.py:100-112` were **unreachable dead code**.
- SIGALRM could not be delivered, so `@pytest.mark.timeout(600)` was **silently void**.
- No Python-level mechanism could have rescued it; `timeout_method = "thread"` would have needed the GIL too.

Only an external SIGKILL ended it, which in CI means the 240-minute step limit. Nightly run `30991070283` lost a full GPU slot exactly this way: `[elapsed 14349s] GPU0: 0.5/80 GiB [w2(14349s)]`. The 0.5 GiB is consistent — `SGLangProcess` passes empty `health_check_ports`/`health_check_urls` (`tests/router/test_router_e2e_with_sglang.py:223-231`), so a dead worker sails through to `KvRouter()` and never registers.

The wait stays unbounded; callers can now actually time it out. Bounding it in Rust would change production semantics (a deployment may legitimately want to wait for workers) and belongs in its own change.

### Note on direction

This PR originally proposed a context-manager reorder across four test files, on the theory that `runtime.shutdown()` blocked while the engine subprocess held ZMQ sockets open. **That theory was wrong** — thanks @dmitry-tokarev-nv for catching it. `DistributedRuntime.shutdown()` (`lib/bindings/python/rust/lib.rs:1128` → `runtime.rs:325`) is `pub fn shutdown(&self)`: it clones three `Arc`s, calls `handle.spawn(..)`, and returns. It cannot block and never touches child processes.

Those reorders have been dropped — they fixed nothing observed, and two of the five sites were unreachable anyway (`run_cache_salt_isolation_test` has no callers; `test_kv_dc_relay_e2e.py` module-level skips in every CI wheel, since `ckf-diagnostics` is not a default feature and no build enables it). The PR is now two files: the binding fix and its regression test.

## Validation

**A/B build against the same regression scenario** — a `KvRouter()` on a worker-less endpoint, built on a daemon thread:

| Build | Result |
|---|---|
| Pre-fix | **exit 124** — hung, required an external kill, never reached the first assertion |
| Post-fix | **exit 0** — main thread ran 2822 bytecode ticks in 3s while the router thread was parked, and SIGALRM was delivered |

Logs confirm the scenario genuinely reaches the wait (`No model card found in discovery ... namespace=gilcheck`), so it does not pass vacuously.

- New regression test `tests/router/test_kv_router_gil_release.py` — passes in 6s, no GPU/etcd/NATS. Runs the scenario in a subprocess to isolate SIGALRM handling and the permanently-parked thread from the pytest process.
- `cargo check` and `cargo clippy` clean on `dynamo-py3`. The compiler confirms every captured type satisfies `Ungil`; `PrefillLoadEstimator` is declared `Send + Sync` (`lib/kv-router/src/scheduling/prefill_load.rs:8`), so no bound changes were needed.
- Checked the other three `block_on` calls in `kv.rs` (`:179`, `:257`, `:462`) — standalone server entrypoints that own their runtime and block forever by design. Not the same bug class; fix stays scoped to `KvRouter::new`.

## Known remaining gap (not in this PR)

`tests/utils/pytest_parallel_gpu.py` has no parent-side kill: the loop is `proc.poll()` + `sleep(1.0)`, and `test.timeout` is passed to the child as `--timeout` and used as a scheduling estimate but never enforced by the parent. With this fix a wedged `KvRouter` can no longer cause it, but any future child that hangs for a different reason will still burn the full GPU slot. Worth a follow-up.


nv-tusharma:
Good catch. Applied the same fix to all three mocker test sites flagged: test_router_e2e_with_mockers.py:673, test_mocker_output_replay_e2e.py:208, and test_kv_dc_relay_e2e.py:86 all now put managed_runtime first so runtime.shutdown() always runs after the subprocess is dead.

tests/router/e2e_harness.py
```diff
@@ -352,8 +352,8 @@ def run_router_decisions_test(
         engine_process_kwargs=engine_process_kwargs,
     )
     with (
-        process as engine_workers,
         managed_runtime(request_plane=request_plane) as runtime,
+        process as engine_workers,
```
dmitry-tokarev-nv, line 355: This is one of only two changed sites on a live code path (the other is `test_kv_router_bindings`).

The reorder is harmless here and arguably more correct, but per the top-level comment the `runtime.shutdown()` blocking claim does not hold — `runtime.rs:325` spawns and returns immediately. If `test_router_decisions_sglang_dp` is the target, the hang is upstream in `KvRouter()` during setup, before teardown is ever reached.

tests/router/e2e_harness.py
```diff
@@ -399,8 +399,8 @@ def run_cache_salt_isolation_test(
         engine_process_kwargs=None,
     )
     with (
-        process as engine_workers,
         managed_runtime(request_plane=request_plane) as runtime,
+        process as engine_workers,
```
dmitry-tokarev-nv, line 402: `run_cache_salt_isolation_test` has no callers:

```
$ git grep -n 'run_cache_salt_isolation_test' -- '*.py'
tests/router/e2e_harness.py:381:def run_cache_salt_isolation_test(
```

Only the definition. The PR title and Summary both name this function, but no CI job can reach this line. Either wire up the missing backend wrappers or delete the chain — reordering it does not change behavior anywhere today.

tests/router/test_kv_dc_relay_e2e.py
```diff
@@ -84,6 +84,7 @@ def test_kv_dc_relay_deduplicates_workers_and_restores_missed_events(
         "durable_kv_events": False,
     }
     with (
+        managed_runtime(request_plane=request_plane) as runtime,
```
dmitry-tokarev-nv: This module never executes. Lines 19-23 are:

```python
if not hasattr(KvDcRelay, "stats"):
    pytest.skip(
        "KV DC Relay diagnostic E2E requires a wheel built with ckf-diagnostics",
        allow_module_level=True,
    )
```

`ckf-diagnostics` is not in `default` for either crate, and `grep -rn ckf-diagnostics container/ .github/` returns zero hits — no build enables it. Module-level skips report green, so this reorder is unverifiable in CI as things stand.

tests/router/test_mocker_output_replay_e2e.py
```diff
@@ -206,13 +206,13 @@ def test_mocker_output_replay_generate_from_request_multi_turn(
     }
 
     with (
+        managed_runtime(request_plane=request_plane) as runtime,
```
dmitry-tokarev-nv: The commit message notes "Mockers are lightweight enough that this has not caused observed CI hangs", so this one is preventive — fine to keep.

One related thing worth knowing: the escape hatch intended to catch a dead worker does not apply to mockers. `common.py:105` gates the liveness poll on `hasattr(engine_workers, "worker_processes")`, an attribute only `ManagedEngineProcessMixin` subclasses have; `MockerProcess` stores its handles as `_process` / `_mocker_processes`. So a mocker that dies at startup — a port conflict, or a bad `response_replay_trace_path` here — reports the generic `KvRouter initialization timed out after 120s` rather than `Worker N exited with code X`.

tests/router/test_router_e2e_with_mockers.py
```diff
@@ -671,13 +671,13 @@ def test_kv_router_bindings(
     }
 
     with (
+        managed_runtime(request_plane=request_plane) as runtime,
```
dmitry-tokarev-nv: This and `run_router_decisions_test` are the two changed sites that actually run in CI.

tests/router/test_router_e2e_with_mockers.py
```diff
@@ -671,13 +671,13 @@ def test_kv_router_bindings(
...
         MockerProcess(
             request,
             mocker_args=mocker_args,
             num_mockers=NUM_MOCKERS,
             request_plane=request_plane,
         ) as mockers,
-        managed_runtime(request_plane=request_plane) as runtime,
     ):
         # Start mocker instances
         logger.info(f"Starting {NUM_MOCKERS} mocker instances")
```
dmitry-tokarev-nv, line 682: These comments now describe the pre-reorder sequence — the mockers were already entered by the `with` above. Same for `# Get runtime and create endpoint` at line 686, which sits after the runtime was obtained first. Worth refreshing so the next reader does not "fix" the ordering back by trusting the comments.

```suggestion
        # Mockers are already running -- the `with` above entered them.
        logger.info(f"Started {NUM_MOCKERS} mocker instances")
```

dmitry-tokarev-nv (commented):
Reviewed the diff plus the surrounding teardown paths. The reordering itself points the right way — a `DistributedRuntime` should outlive the workers it talks to — but I could not reproduce the stated root cause from the source, and I think the title and commit message need to change before this lands.

### `runtime.shutdown()` does not block

The commit message says it "is a blocking Rust FFI call that drains ZMQ sockets; with the subprocess holding the other end open it never returned." Tracing it:

- `lib/bindings/python/rust/lib.rs:1128` — `fn shutdown(&self) { self.inner.shutdown(); }`
- `lib/runtime/src/distributed.rs:360` — `{ self.runtime.shutdown(); self.discovery_client.shutdown(); }`
- `lib/runtime/src/runtime.rs:325` — the entire body after logging is `handle.spawn(async move { ... })`. It returns immediately; every wait, including `tokio::time::timeout(graceful_shutdown_timeout(), ...)`, happens inside the spawned task.
- `discovery_client.shutdown()` is an empty body for both backends these tests use — `lib/runtime/src/storage/kv/etcd.rs:55` and `nats.rs:51` are bare comments.

There is also a unit test at `runtime.rs:424` (`shutdown_cancels_main_token_after_graceful_timeout`) asserting the token is cancelled after a single `yield_now()`, which only holds if the call is non-blocking. And no ZMQ code is on that path.

### What does block is `KvRouter()`

`KvRouter::new` (`lib/bindings/python/rust/llm/kv.rs:1856`) takes no `Python<'_>` parameter and never calls `py.allow_threads`; its body is `runtime.block_on(async move { ... })`. PyO3 pymethods hold the GIL, so the daemon thread in `_create_kv_router_with_timeout` (`tests/router/common.py:92`) holds the GIL for the whole call — the main thread's `join(timeout=2)` releases it and can never re-acquire it, so the 120s guard at line 100 never fires and pytest-timeout's SIGALRM handler cannot run either. That matches the 14349s observation, and it matches the mechanism this file's own docstring describes at lines 69-75.

`kv.rs:598-601` in the same file shows the pattern that fixes it (`py.allow_threads`). Pairing that with `faulthandler_timeout` in `pyproject.toml` — currently unset, as is `timeout_method` — would have produced a C-level traceback on the first occurrence instead of a silent 4-hour job. Worth noting `timeout_method = "thread"` alone is not sufficient: that timer thread needs the GIL too.

### CI coverage

- `test_router_decisions_sglang_dp` is still `@pytest.mark.nightly` (`tests/router/test_router_e2e_with_sglang.py:312`), so nothing in this PR's pre-merge run exercises the change. `.ai/ci-guidelines.md` lines 8-27 covers exactly this case: promote to `pre_merge` with a `TODO`, then revert the marker before merge.
- `tests/router/test_kv_dc_relay_e2e.py` skips at module level in every build — details inline.

That leaves 2 of the 4 changed files with no execution in this PR's CI.

### The site with the incident record is not included

`run_indexers_sync_test` (`e2e_harness.py:514`) still has `with process as engine_workers:` with the runtimes created inside it — `_test_router_indexers_sync` puts up to three `managed_runtime` contexts on an `ExitStack` (`common.py:1941`, `:2039`, `:2124`), so all three `shutdown()` calls unwind while every engine subprocess and both standalone indexers are still alive. That covers `test_indexers_sync` (which already carries a "Known flake" comment), `test_vllm_indexers_sync` and `test_trtllm_indexers_sync`; `test_sglang_indexers_sync` is hard-skipped with a note recording 3659s / 4235s / 6077s hangs.

If the ordering hypothesis is right, that is where the evidence for it lives. It also cannot use this PR's `with`-tuple idiom, since the process CM is in `e2e_harness.py` while the runtimes are created in `common.py`.

### One structural suggestion

`managed_runtime` (`tests/router/helper.py:634`) is the only helper in that file without a docstring, and the ordering constraint this PR establishes now lives only in commit messages — nothing enforces it, and several sibling sites cannot follow it (`test_slot_tracker_e2e.py:81-83` enters it on an `ExitStack` while the mockers come from fixtures; `common.py:2698` nests it inside a `FrontendRouterProcess`; `mocker_process.py:660` and `common.py:424`/`:2485`/`:3583` build runtimes via bare `get_runtime()` with no `shutdown()` at all). A single combined context manager yielding `(engine_workers, runtime)` would make the ordering impossible to state wrongly, rather than a convention each new site has to remember.

tests/router/e2e_harness.py
nv-tusharma: outdated, update to release GIL during KvRouter init

tests/router/test_kv_dc_relay_e2e.py
nv-tusharma: outdated, updated to release GIL during KvRouter init

tests/router/test_mocker_output_replay_e2e.py
nv-tusharma: outdated, update to release GIL during KvRouter init

tests/router/test_router_e2e_with_mockers.py
nv-tusharma: outdated, update to release GIL during KvRouter init

tests/router/test_router_e2e_with_mockers.py
nv-tusharma: outdated, update to release GIL during KvRouter init

tests/router/e2e_harness.py
nv-tusharma: outdated, update to release GIL during KvRouter init

nv-tusharma:
/ok to test [e0de77a](https://github.com/ai-dynamo/dynamo/pull/12762/commits/e0de77aa73854e583914968240d8fcdd7209420c)

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -1890,41 +1891,54 @@ impl KvRouter {
...
-                rs::protocols::annotated::Annotated<
-                    llm_rs::protocols::common::llm_backend::LLMEngineOutput,
-                >,
-            >::from_client(
-                client,
-                rs::pipeline::network::egress::push_router::RouterMode::KV,
-            )
-            .await
-            .map_err(to_pyerr)?;
+        py.allow_threads(|| {
```
dmitry-tokarev-nv, line 1897: **Follow-up 1 (not a blocker): `DistributedRuntime::new` has the same shape.**

The description says the other `block_on` calls in this file were audited, which is right — but the same pattern exists one file over, in `lib/bindings/python/rust/lib.rs:1075`:

```rust
#[new]
#[pyo3(signature = (event_loop, discovery_backend, request_plane, enable_nats=None, *, event_plane=None))]
fn new(
    event_loop: PyObject,      // <- no `Python<'_>` token
    ...
) -> PyResult<Self> {
    ...
    runtime.secondary().block_on(rs::DistributedRuntime::new(...))   // <- GIL held
}
```

Same class as the bug this PR fixes: a `#[pymethods]` constructor holding the GIL across a `block_on` that can wait on the network. It is benign in the common case because it is usually constructed on the main thread with nothing else to starve — but it is the identical trap for any caller that constructs off-thread with a deadline, which is exactly the pattern `_create_kv_router_with_timeout` uses.

Worth noting the new test's child constructs one of these on its main thread (`runtime = DistributedRuntime(loop, "mem", "tcp")`), so it is already inside the blast radius, just not in a position where it can hurt.

Fine as a separate PR — flagging so it does not get lost now that the audit in this one is scoped to `kv.rs`.

tests/router/test_kv_router_gil_release.py
```diff
@@ -0,0 +1,136 @@
...
+
+def _child_env() -> dict:
+    env = {k: v for k, v in os.environ.items() if k not in _UNSET_IN_CHILD}
+    env["DYN_ROUTER_MIN_INITIAL_WORKERS"] = "1"
+    return env
+
+
+_SCENARIO = textwrap.dedent(
+    """
+    import asyncio, threading, time
```
dmitry-tokarev-nv: **Follow-up 2 (not a blocker): `faulthandler` would make a regression here self-diagnosing.**

The description correctly rules out `timeout_method = "thread"` — that timer thread needs the GIL too. `faulthandler` is a different mechanism and does not have that problem: its watchdog calls `_Py_DumpTracebackThreads`, which walks interpreter state **without acquiring the GIL**. That is precisely why it works on a wedged interpreter, and it is the one Python-level tool that would have produced a stack on the original 14349s hang instead of silence.

Applied to the child here, a regression fails with a C-level traceback showing exactly where the thread is parked, rather than the current bare "hung and needed an external kill". 60s fires ahead of `_SUBPROCESS_TIMEOUT_S = 90`, so the parent's `returncode != 0` branch reports it with `stderr` attached:

```suggestion
    import asyncio, faulthandler, threading, time

    # If a regression re-pins the GIL, dump C-level stacks for every thread and
    # exit. faulthandler's watchdog does not need the GIL, so it still fires on a
    # wedged interpreter -- which is what pytest-timeout cannot do. Fires before
    # _SUBPROCESS_TIMEOUT_S so the failure carries a traceback rather than a bare
    # "hung and was killed".
    faulthandler.dump_traceback_later(60, exit=True)
```

Separately, and more valuable long-term: setting `faulthandler_timeout` in `pyproject.toml` under `[tool.pytest.ini_options]` (currently unset, as is `timeout_method`) gives every test in the repo that property, not just this one. That is the generalizable half — this test guards one known wedge, `faulthandler_timeout` catches the next one nobody predicted.

dmitry-tokarev-nv (commented):
Re-reviewed after the rewrite. This is a much better change than what I commented on — the diagnosis is now grounded, the fix is scoped to the actual mechanism, and the Validation section does a real pre/post A/B instead of reasoning from logs. Thanks for reworking it rather than patching the rationale.

I verified the parts the review depends on:

- `cargo check` clean on `dynamo-py3` at `e0de77aa` locally (2m17s, no diagnostics on `kv.rs`) — which also settles the `Ungil` question, since `py.allow_threads` would not type-check if anything captured were GIL-bound.
- All four markers (`pre_merge`, `gpu_0`, `integration`, `router`) are registered in `pyproject.toml`, so no `--strict-markers` collection error.
- `DYN_ROUTER_MIN_INITIAL_WORKERS` is real (`lib/kv-router/src/scheduling/config.rs:26`) and the wait is gated on `min_initial_workers > 0 && !skip_initial_worker_wait` (`lib/llm/src/kv_router.rs:377`). `skip_initial_worker_wait` is not in the Python `KvRouterConfig` signature, so `KvRouterConfig()` leaves the wait engaged — the scenario's premise holds.
- `DistributedRuntime(loop, "mem", "tcp")` matches the pyo3 signature at `lib/bindings/python/rust/lib.rs:981`.

The regression test is better than most in this area, specifically because it cannot pass vacuously: the `t.is_alive()` guard makes "the blocking path was never exercised" a failure rather than a silent green, and `router_error` keeps an unrelated break from reporting as a bogus GIL result. That distinction is worth keeping if this test gets refactored later.

Three follow-ups below. **None are blockers** — two are inline, and the third has no file in this diff to anchor to:

**3. `test_indexers_sync` is now unowned.** Dropping the reorder was the right call, but the hang that started this still has no owner. `run_indexers_sync_test` (`tests/router/e2e_harness.py:514`) enters up to three `managed_runtime` contexts on an `ExitStack` (`common.py:1941`, `:2039`, `:2124`) inside `with process as engine_workers:`. `test_indexers_sync` carries a "Known flake" comment; `test_sglang_indexers_sync` is hard-skipped with a note recording 3659s / 4235s / 6077s hangs. Whether or not teardown ordering turns out to be the cause there, that one has the incident record and this PR no longer touches it — worth a tracking issue so it does not fall through the gap between the two versions of this PR.

On the remaining-gap note in the description: agreed, and that matches what I had recorded against #10126 (VRAM-aware scheduler landed without a parent-side hard-kill for hung children). It is a real gap, independently observed.

tests/router/test_kv_router_gil_release.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def build_router():
+        try:
+            KvRouter(endpoint, 16, KvRouterConfig())
+        except BaseException as exc:
+            router_error.append(exc)
+
+    t = threading.Thread(target=build_router, daemon=True)
+    t.start()
+    time.sleep(1.0)
```
jthomson04: **P2:** `t.is_alive()` only proves that the target has not returned. If the builder is delayed, this sleep and the tick loop can finish before it enters `KvRouter`, so the test can pass without reaching the blocking path. Use a `threading.Event` immediately before the call, wait for it, then use `t.join(timeout=...)` as the GIL check. That also lets us cut the 90s child timeout.

tests/router/test_kv_router_gil_release.py
```diff
@@ -0,0 +1,136 @@
...
+
+_SCENARIO = textwrap.dedent(
+    """
+    import asyncio, threading, time
+
+    from dynamo._core import DistributedRuntime, KvRouter, KvRouterConfig
+
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    runtime = DistributedRuntime(loop, "mem", "tcp")
```
jthomson04: **P3:** Pass `event_plane="zmq"` here. The public constructor already expresses the self-contained transport; stripping `NATS_SERVER` and `DYN_EVENT_PLANE` couples this test to the current internal `nats_enabled` heuristic.

tests/router/test_kv_router_gil_release.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Surface constructor failures instead of swallowing them -- otherwise an
+    # unrelated break (API change, bad endpoint) exits the thread immediately and
+    # reports as a bogus GIL assertion below.
+    router_error = []
+
+    def build_router():
+        try:
+            KvRouter(endpoint, 16, KvRouterConfig())
+        except BaseException as exc:
```
jthomson04: **P3:** Catch `Exception`, not `BaseException`. The latter also swallows `SystemExit` and `KeyboardInterrupt` and turns them into misleading router failures.

lib/bindings/python/rust/llm/kv.rs
```diff
@@ -1890,41 +1891,54 @@ impl KvRouter {
...
+        // unbounded (kv_router.rs: `startup_watch.wait_for(..)`), and PyO3 otherwise
+        // holds the GIL for the whole #[pymethods] call -- starving every other
+        // thread, including any supervisor enforcing a deadline and CPython's
+        // signal handling.
+        //
+        // This makes the wait supervisable, not cancellable: block_on still runs to
+        // completion, and constructing directly on the main thread remains
+        // uninterruptible by SIGALRM. Callers needing a deadline must construct
+        // off-thread and enforce it themselves -- see
+        // tests/router/common.py::_create_kv_router_with_timeout.
```
jthomson04: **P3:** This is too much comment for one `allow_threads` call. Keep the two facts that matter: the initial-worker wait is unbounded, and releasing the GIL makes it supervisable but not cancellable. The CPython/SIGALRM narrative and test-path reference will go stale.

tests/router/test_kv_router_gil_release.py
```diff
@@ -0,0 +1,136 @@
...
+
+Releasing the GIL makes that wait *supervisable*, not cancellable -- ``block_on``
+still runs to completion, so the builder thread stays parked for the life of the
+process. What changes is that everyone else can now run, which is the whole point.
+
+The scenario runs in a subprocess because a regression leaves a thread parked in
+the constructor holding the GIL -- which would wedge the pytest worker itself,
+uninterruptibly. Isolating it means a regression fails one test instead of
+hanging the run.
+"""
```
jthomson04: **P3:** This docstring records PR and incident history, and repeats the Rust comment. Keep a short statement of the behavior under test and why the child process is required; git and the PR keep the rest.

jthomson04 (commented):
The production `allow_threads` fix looks correct. These five comments are non-blocking.

nv-tusharma:
@dmitry-tokarev-nv @jthomson04 Thanks for the careful second pass. I addressed the in-scope follow-ups in [3b71dc1ae](https://github.com/ai-dynamo/dynamo/commit/3b71dc1ae8385298da148c36d6209ff3b221dd04):

- Replaced the startup sleep/tick loop with a `threading.Event` immediately before `KvRouter(...)`, followed by `join(timeout=...)`. The event establishes that the builder reached the call site, and the timed join is now the GIL check.
- Passed `event_plane="zmq"` explicitly and removed the transport-environment filtering. The public constructor argument makes ambient `DYN_EVENT_PLANE` and `NATS_SERVER` irrelevant; the child environment now only sets the initial-worker-count override.
- Narrowed `BaseException` to `Exception`.
- Added a child-local `faulthandler` watchdog ahead of the parent deadline, so a regression exits with Python thread stacks rather than only reporting an external kill. The timeout layers are now 30s watchdog / 45s subprocess / 60s pytest.
- Trimmed the Rust comment and test docstring to the durable behavioral contract.

I’m keeping the broader items separate from this PR: auditing `DistributedRuntime.__new__` and `detached()` for GIL-held `block_on`, evaluating a repository-wide `faulthandler_timeout`, and adding public tracking for the remaining indexer-sync failures. There is already internal tracking for the SGLang startup problem, while the mocker convergence flake appears distinct and should be tracked separately.

Local syntax, Python formatting/lint hooks, and Rust formatting passed. I’ve left the review threads unresolved for reviewer verification.

tests/router/test_kv_router_gil_release.py
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762#issuecomment-5246323320

tests/router/test_kv_router_gil_release.py
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

tests/router/test_kv_router_gil_release.py
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

tests/router/test_kv_router_gil_release.py
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

lib/bindings/python/rust/llm/kv.rs
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

lib/bindings/python/rust/llm/kv.rs
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

tests/router/test_kv_router_gil_release.py
nv-tusharma: addressed: https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04

nv-tusharma:
/ok to test [3b71dc1](https://github.com/ai-dynamo/dynamo/pull/12762/commits/3b71dc1ae8385298da148c36d6209ff3b221dd04)

dmitry-tokarev-nv (approved):
Thank you for addressing comments and driving this home.

jthomson04 (approved):

PR #12767, opened 2026-08-06
feat(gms): carry the vLLM prefix-cache index across a failover

mohammedabdulwahhab:
### Related Issues

Part of #12053. **Stacked on #12648** — review that one first; this is meaningless without it.

### Summary

#12648 makes the KV *bytes* survive a failover. A cache hit also needs the *index*: `block_hash → block_id`, which lives in the scheduler's `BlockPool` in process RAM and cannot be regenerated from the bytes, because the hash is over token ids rather than content. So today a standby adopts correct KV that it cannot find, and re-prefills anyway.

This carries the index the same way the layout is carried: written by the active engine, read by whoever adopts next.

### The correctness problem

An index entry is a name-to-location binding carrying an unstated claim: *the bytes at this block are the bytes this hash denotes*. Nothing in the memory records the hash, so the claim can quietly stop holding, and replaying a snapshot re-asserts every binding in it. That needs evidence.

Two artifacts supply it. A periodic **snapshot** proposes bindings. A per-block **generation counter** vetoes them.

The counter is bumped whenever a block leaves the free queue, which is the one gate a block must pass through to become writable. The snapshot records each block's generation alongside its hash, and replay installs a record only if the counter has not moved since. So a block cached after the snapshot is merely absent (MISS, recompute) while a block reused after it is refused. Losing the counters, their file, or a matching header installs nothing rather than everything.

Bumping on **handout** rather than on eviction is the part worth reviewing. Eviction removes a name and touches no byte; what invalidates a record is the block being handed out and refilled. Watching eviction would make correctness depend on having enumerated every path that drops a hash — a claim about cache bookkeeping, which is exactly where a cache library grows new paths. Watching the handout makes it depend on having enumerated every path that makes a block writable, which is how bytes get written at all.

Two things follow that were not obvious:

- The bump sits on the free queue, not on `get_new_blocks`, because the sliding-window manager pops sink blocks straight off the queue.
- `reset_prefix_cache` bumps every counter. It drops the index without moving bytes, and it exists so an RLHF flow can discard a cache computed under stale weights — which a replay would otherwise reinstate.

### Two gates, in order

| gate | asks | catches |
|---|---|---|
| `layout_id` | does `block_id` mean the same bytes? | geometry drift: block size, dtype, tensor split, TP topology |
| generation | is the block still holding what the record claims? | reuse since the snapshot, by any path |

`layout_id` has to come first: generations are indexed by block number, so under a different geometry the check compares the wrong entries.

### Integration

Installed via a `vllm.general_plugins` entry point, so it lands in the scheduler's process at any TP. Importing the worker module isn't enough: at TP>1 that happens only in the worker children, leaving the scheduler's `BlockPool` untouched and the mechanism silently absent.

| hook | why |
|---|---|
| `SchedulerConfig.get_scheduler_cls` | derive from the class vLLM chose, and override `update_from_output` by inheritance |
| `BlockPool.free_block_queue` (class swap) | bump on handout |
| `BlockPool.reset_prefix_cache` (class swap) | invalidate everything when the index is dropped |
| `EngineCore.wake_up` | replay after all ranks reattach, before scheduling resumes |

Wrapping the resolver rather than setting `scheduler_config.scheduler_cls`: `get_scheduler_cls` only returns `AsyncScheduler` while that field is unset, so naming a class there silently disables async scheduling, which is on by default in 0.26. Wrapping keeps whatever base the config selected — async, sync, or someone else's subclass.

Hooking `wake_up` rather than `resume_scheduler`: `EngineCoreProc` overrides `resume_scheduler` and returns early while `engines_running`, so the base method never runs and the replay is skipped. That is a race rather than a configuration, and it is silent.

Replay is gated on **every** rank having adopted, queried via `collective_rpc`. The index is engine-wide but the bytes are per-rank, so a partial adoption would be correct for the ranks that adopted and garbage for the one that didn't.

Off by default: set `GMS_KV_INDEX_PATH` to enable.

### Validation

16 unit tests over a `FakeBlockPool`, no GPU: both staleness directions, a block evicted but *not* reused still replaying, reset invalidating everything, layout-mismatch refusal, counters surviving a takeover, counters from another layout being reset rather than reinterpreted, and both fail-closed paths.

Three e2e failovers on 2×B200, each **16/16 with output byte-identical to a fresh prefill**:

| run | result |
|---|---|
| TP=1 | 294 entries replayed, `prefix_hits 4688 == winner`, full HIT |
| eviction pressure, 384 blocks | partial reuse (2144), correct — eviction retired some blocks |
| TP=2 | full HIT |

Control run with `GMS_KV_INDEX_PATH` unset: `prefix_hits loser=0`, full re-prefill. The metric is measuring what it claims to.

Snapshot cost measured on B200: 2.6 ms at 8k blocks, 8.3 ms at 40k, 30 ms at 160k. At the default 1 s interval that's 0.26%–3% of serving time. The snapshot is affordable in this wholesale form *because* the counters make staleness safe; without the veto it would have to be current, which means deltas, ordering, and compaction.

### Notes

- **`PYTHONHASHSEED` must be pinned identically across engines.** vLLM's `NONE_HASH` (the root of the block-hash chain) is `os.urandom(32)` per process otherwise, so no replayed hash ever matches. Fail-safe — it degrades to a MISS — but it silently removes the entire benefit.
- Snapshots flush once the engine goes quiescent. They ride on steps, and an engine that caches a prefix and then idles would otherwise never persist it, which is exactly the state a failover finds it in.
- Durability is page-cache level, deliberately. The counters live in a `MAP_SHARED` mapping, so a bump is a store with no syscall and nothing to tear. It survives the process, which is the threat model; it does not survive the node, and neither do the GPU pages or the GMS daemon holding them, so there is nothing an `fsync` would buy.
- Generation values are only meaningful within one lineage of the same pages. An array whose history restarts under an unchanged `layout_id` would be ambiguous. Closing that needs a writer epoch, deliberately left out of this PR.
- Content checksums would subsume the counters and additionally cover device-level faults and unmodeled write paths, at the cost of a per-snapshot collective back into the worker. Follow-up, not this PR.
- Monkey-patching is down to one method (`EngineCore.wake_up`) plus two instance class swaps and a config-resolver wrapper; the rest is inheritance. `_insert_block_hash` is still a private call on the replay path. The snapshot itself reads only `BlockPool.blocks`, a public list.
- **Known gap, worth reviewer attention.** I wrote a prefix-reuse assertion into `tests/gpu_memory_service/test_shadow_failover.py` and it fails: in that suite's pause/resume flow the standby reports zero prefix hits after taking over, even though the primary had cached the prompt. The mechanism works in the lock-driven crash failover the runs above exercise, so something about the pause/resume path (a snapshot that never gets written, or a layout identity that does not match on resume) defeats it. The test is not in this PR — a red nightly is worse than no test — but the gap is real and is the next thing to chase.


hutm:
Superseded by #13398, which now contains the current vLLM BlockPool-subclass integration, native block-key preservation, and generation-fenced directory hydration on top of the rebuilt MVP stack tracked in #12053. Closing this older implementation so reviewers have one source of truth.

PR #2190, opened 2025-07-30
fix: handle groveTerminationDelay and auto-detect grove installation

julienmancuso:
#### Overview:

handle groveTerminationDelay and auto-detect grove installation

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Grove termination delay via a new duration-based setting.
  * Grove functionality is now automatically enabled if the required CRD is detected in the cluster.

* **Configuration Changes**
  * Replaced the previous boolean flag for enabling Grove with a new `groveTerminationDelay` duration parameter in configuration files.
  * The enablement of Grove permissions in RBAC and deployment is now unconditional.

* **Bug Fixes**
  * Ensured Grove termination delay is correctly set in generated PodGangSet resources.

PR #12769, opened 2026-08-06
fix(operator): suffix Grove worker namespaces

### LINKED ISSUE #12796: [BUG]: Grove rolling updates reject new worker generations in the existing WorkerSet
**Describe the Bug**

The DCD pathway creates a canonical worker hash and sets it as `DYN_NAMESPACE_WORKER_SUFFIX`. The worker runtime uses that suffix when constructing its namespace, so each worker generation publishes and discovers under a distinct runtime namespace.

The Grove pathway does not currently set this suffix. During a Grove rolling update, old and new worker Pods therefore publish MDCs under the same runtime `{namespace, component, endpoint}` identity.

The frontend caches the old worker MDC in the WorkerSet. When it discovers a new worker MDC with a different checksum, it treats that card as incompatible with the existing WorkerSet and rejects it. This applies to aggregated workers as well as disaggregated prefill/decode workers.

**Steps to Reproduce**

1. Create a Grove-backed `DynamoGraphDeployment` with an aggregated worker, or a prefill/decode deployment, serving one model.
2. Wait until the DGD is Ready and the frontend has registered the initial worker MDCs.
3. Update a worker configuration that changes the MDC checksum, such as the configured maximum context length.
4. Allow Grove to perform its rolling update while old and new Pods overlap in discovery.
5. Inspect frontend logs.

The frontend logs an MDC checksum mismatch for the newly discovered worker:

```text
ERROR dynamo_llm::discovery::watcher:
  Checksum for new worker does not match existing WorkerSet's checksum.
  Drain all old workers in this namespace before deploying a new version.
  model_name=<model>
  namespace=<shared-grove-runtime-namespace>
  new_checksum=<new-mdc-checksum>
```

**Expected Behavior**

New workers should be discovered, construct a new WorkerSet, and be routable.

**Actual Behavior**

Because old and new workers use the same runtime namespace, the frontend places them in the same WorkerSet identity. It rejects the new worker MDC as checksum-incompatible, so the new generation is not admitted or routable.

**Environment**

- Deployment: Kubernetes `DynamoGraphDeployment` using the Grove pathway
- Workload: aggregated worker or disaggregated prefill/decode workers
- Update mode: Grove rolling update with old/new Pods concurrently discoverable
- Discovery backend: any backend using frontend MDC discovery
- Trigger: a user worker-spec update that changes MDC checksum

tmonty12:
## Summary

Fixes [#12796](https://github.com/ai-dynamo/dynamo/issues/12796).

Grove can run old and new worker generations concurrently during a PodCliqueSet rollout. Without a worker-generation suffix, both generations publish under the same runtime namespace and discovery can reject the new MDC checksum as incompatible.

This change renders the canonical DGD worker hash into Grove worker templates, causing each worker generation to publish under a distinct runtime namespace. Prefill and decode share the same canonical hash and switch namespaces as one cohort.

## Compatibility and rollout behavior

- New Grove DGDs render worker hash labels and `DYN_NAMESPACE_WORKER_SUFFIX` from their first PCS.
- Existing legacy unsuffixed PCSs remain unchanged after operator upgrade and frontend-only changes.
- A worker-generation change renders a suffixed desired PCS.
- Once a PCS is suffixed, later renders retain the suffix.
- The suffix migration is no longer represented by a DGD annotation or admission-webhook state machine.

The DGD current worker hash is committed only after the desired PCS has been successfully synced. A failed PCS write therefore leaves the DGD hash unchanged and retries from live state.

## Runtime namespace cutover

`status.components[*].runtimeNamespace` remains the previously published value until Grove has accepted the current PCS generation and every worker child has completed that same accepted PCS revision.

The hash used for the cutover is read from the accepted PCS worker labels. All worker cliques must agree on the hash and all workers must complete together, preventing a mixed prefill/decode namespace generation.

## Validation

- Unit coverage for legacy, suffixed, malformed, stale, zero-replica, and PCSG readiness states.
- Reconciler coverage for PCS-sync failures, DGD-hash commit failures after PCS persistence, retry after a fresh read, and idempotence.
- Envtests for new DGDs, legacy DGDs, frontend-only changes, and worker-generation suffix migration.

```console
cd deploy/operator
KUBEBUILDER_ASSETS=<kubebuilder-assets> go test ./internal/controller_common ./internal/controller ./internal/dynamo -count=1
```


julienmancuso:
I think the overall approach is correct: the DGD annotation should remain the durable, monotonic migration state. This is safer than deriving the state from the PodCliqueSet/controller because admission atomically sees the old and new DGD, while the controller advances the current worker hash before syncing workloads and may observe stale or deleted child state.
I would simplify and harden the webhook transition rules before merging:

- Treat the annotation as an operator-owned compatibility epoch, not as user input or current-pathway state.
- On CREATE, remove any supplied value and set the annotation for the new object.
- On UPDATE, remove the incoming value before any pathway guard.
- If the old object was enabled, always preserve true, including across Grove disable/re-enable.
- For an unmarked legacy object, enable it when entering Grove or when the canonical worker hash changes.
- If either hash cannot be computed, do not reject the update. Conservatively enable the suffix and allow validation/reconciliation to handle the underlying spec. Rejecting here can prevent repairing a legacy DGD.
- Only an unmarked legacy object with an unchanged worker spec should remain disabled.
- In pseudocode:

```
delete(newDGD.Annotations, suffixMarker)

if operation == Create ||
    oldMarkerEnabled ||
    (!oldGrove && newGrove) {
    enableSuffix(newDGD)
    return nil
}

oldHash, oldErr := computeWorkerHash(defaultedOldDGD)
newHash, newErr := computeWorkerHash(newDGD)

if oldErr != nil || newErr != nil || oldHash != newHash {
    enableSuffix(newDGD)
}
```

A table-driven test covering CREATE, marker deletion/spoofing, Grove pathway transitions, unchanged/changed worker specs, and hash-comparison failures would make the intended state machine explicit.
So I support the webhook-owned annotation approach, but I think the current early pathway return and admission-failing hash errors need to be fixed for it to be robust.

deploy/operator/internal/dynamo/grove.go
```diff
@@ -183,7 +183,11 @@ func evaluateGroveComponents(ctx context.Context, reader client.Reader, dgd *v1b
 		if checkErr != nil {
 			return false, "", "", nil, fmt.Errorf("component %q: %w", componentName, checkErr)
 		}
-		componentStatus.RuntimeNamespace = dgd.GetDynamoNamespaceForComponent(component)
+		runtimeNamespace, err := GetGroveRuntimeNamespace(dgd, component)
+		if err != nil {
+			return false, "", "", nil, fmt.Errorf("component %q runtime namespace: %w", componentName, err)
+		}
+		componentStatus.RuntimeNamespace = runtimeNamespace
```
sttts: Keep the active runtime namespace until the Grove cutover.

`GetGroveRuntimeNamespace(dgd, component)` derives the status value from the desired DGD hash. On the first worker-spec update of a legacy unsuffixed Grove DGD, admission enables suffixing before the Grove rollout completes. While the old workers are still active in the base namespace and the new PodCliques are `Updating`, this line publishes `base-<new-hash>`.

The Kubernetes planner connector treats a non-empty `status.components[*].runtimeNamespace` as authoritative, so it switches discovery and metrics away from the still-active generation. This also contradicts the `ComponentReplicaStatus` contract and the DCD rollout path, which preserves the old active namespace until cutover.

Please retain the previously observed/active namespace while Grove reports the rollout in progress, and switch to the new hash namespace only after the new generation is ready and the old one is no longer active. A regression test should cover both the first legacy base→hash migration and a later hash-A→hash-B rollout.

sttts (commented):
One blocking backward-compatibility issue; see the inline comment.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -126,6 +127,101 @@ func (d *DGDDefaulter) defaultV1Beta1(
...
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+	grovePathway bool,
+) error {
+	oldDGD, err := decodeOldDGD(req)
+	if err != nil {
+		return err
+	}
+
+	oldGrovePathway := d.isGrovePathway(ctx, oldDGD)
+	if !grovePathway || !oldGrovePathway {
```
sttts: **[P1] Preserve the operator-owned suffix state across pathway changes**

This return runs before ownership normalization and preservation of `nvidia.com/grove-worker-hash-suffix-enabled`. A DGD that already has the marker set can be updated to `nvidia.com/enable-grove: "false"` through a replace/apply payload that omits the operator-owned marker; the webhook returns here and persists its deletion. If Grove is enabled again later, `oldGrovePathway` is false and this guard returns again, leaving the DGD unmarked. The renderer then falls back to the unsuffixed runtime namespace even though this DGD had already crossed the compatibility boundary, potentially putting later Grove worker generations back into one shared namespace. Preserve an old `true` marker before the pathway guard (and normalize user-supplied state separately), with coverage for `true → Grove disabled → Grove re-enabled`.

sttts (commented):
One additional blocking compatibility-state issue; see the inline comment.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
sttts: @tmonty12 @julienmancuso — design question before we add more transition state here: should `nvidia.com/enable-grove` actually be immutable after DGD creation?

Context:

- It was introduced in #2012 as a simple per-DGD “explicit opt out”; that PR did not define update or live-migration semantics.
- #2190 then made Grove auto-detected/default when its API is present, so the annotation became the escape hatch for LWS/DCD. #3203 later removed temporary `enable-grove: "false"` annotations after Planner gained Grove support, which is at least evidence of a desired `false → true` workflow in manifests.
- On current `main`, however, DGD updates are still filtered by `GenerationChangedPredicate`, so annotation-only pathway changes do not reliably reconcile, and I do not see a documented complete Grove ↔ DCD cutover/cleanup contract. We make similarly disruptive topology/KV-policy choices immutable.
- The open DisaggregatedSet PR #11645 deliberately chooses the opposite contract: on a Grove-enabled cluster, DS selection requires both `nvidia.com/enable-grove: "false"` and `nvidia.com/enable-disaggregatedset: "true"`. Its review treated watching those annotation-only changes as P1 and added replacement-first DCD ↔ DS transitions plus stale-Grove cleanup. Julien’s rework guidance also explicitly asks to preserve DCD/Grove/DS transitions as provider-specific compatibility cleanup.

That leaves two coherent models:

1. Make `enable-grove` immutable and require delete/recreate for a pathway change. That is the simpler/currently safer API and makes the transition above unreachable (with some grandfathering or migration mechanism needed for legacy annotation-less DGDs before Grove auto-detection is enabled).
2. Keep it mutable as required by the #11645 routing model. Then pathway transitions are part of the API contract, and this durable suffix marker must survive Grove off/on in addition to the controller providing safe handoff and cleanup.

Is model 2 already the intended contract for `main`, or would immutability be the cleaner/correct behavior outside the DS work?

julienmancuso: Thought more about it, i feel like 1 would actually make more sense ...
making provider selection sticky would be cleaner

sttts: +1. That's basically your provider type in the groups DEP.

julienmancuso: i created this issue : https://github.com/ai-dynamo/dynamo/issues/12938

deploy/operator/internal/dynamo/grove.go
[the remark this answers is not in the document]
tmonty12: Addressed in 0044c05979: readiness now determines whether the component completed the accepted PCS revision, and GetGroveRuntimeNamespace keeps status.components[component].runtimeNamespace until then. The cutover test covers retained and switched namespaces.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -126,6 +127,101 @@ func (d *DGDDefaulter) defaultV1Beta1(
...
+	d.defaultComponentFields(oldDGD, oldGrovePathway)
+
+	if oldDGD.GetAnnotations()[consts.AnnotationGroveWorkerHashSuffixEnabled] == "true" {
+		setGroveWorkerHashSuffixEnabled(dgd)
+		return nil
+	}
+
+	// This is operator-owned state. An update cannot enable the suffix without
+	// changing the worker spec, nor can it retain a user-supplied value.
+	delete(dgd.Annotations, consts.AnnotationGroveWorkerHashSuffixEnabled)
```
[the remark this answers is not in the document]
tmonty12: Addressed in the current provider-based flow: the incoming marker is removed before provider selection, and previously enabled durable state is preserved by defaultGroveWorkerHashSuffix. The transition is no longer gated by mutable Grove routing intent.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -126,6 +127,101 @@ func (d *DGDDefaulter) defaultV1Beta1(
...
+	oldWorkerHash, err := dynamo.ComputeDGDWorkersSpecHash(oldDGD)
+	if err != nil {
+		return fmt.Errorf("compute previous Grove worker hash suffix: %w", err)
+	}
+	newWorkerHash, err := dynamo.ComputeDGDWorkersSpecHash(dgd)
+	if err != nil {
+		return fmt.Errorf("compute Grove worker hash suffix: %w", err)
+	}
```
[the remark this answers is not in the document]
tmonty12: Intentional: this now rejects an update when either previous or incoming worker hash cannot be computed (086115cbc1). Failing open could omit the suffix and allow the rolling update this change prevents; validation or reconciliation cannot make that namespace decision safely.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -93,19 +96,17 @@ func (d *DGDDefaulter) defaultV1Beta1(
...
-		component := &dgd.Spec.Components[i]
-		if component.Replicas == nil {
-			component.Replicas = ptr.To(int32(1))
+	d.defaultComponentFields(dgd, grovePathway)
+
+	switch req.Operation {
+	case admissionv1.Create:
+		if grovePathway {
+			setGroveWorkerHashSuffixEnabled(dgd)
 		}
```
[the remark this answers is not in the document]
tmonty12: Addressed: Default clears the marker on every request before selecting the provider. A create can set the suffix only through operator-owned Grove defaulting, so a user-supplied marker cannot persist.

deploy/operator/internal/dynamo/grove.go
tmonty12: Addressed in 0044c05979: the status keeps the previous runtime namespace until the child completes the accepted PCS revision; the readiness test covers active-namespace retention and post-cutover publishing.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
tmonty12: Addressed by the provider-immutability dependency (#12938 and #12942) now in this branch: the selected provider is durable, and suffix defaulting derives from it rather than mutable enable-grove intent.

deploy/operator/internal/controller/dgd_grove_workload_renderer.go
```diff
@@ -110,10 +113,41 @@ func (r *groveWorkloadRenderer) Render(
...
+		if !dynamo.IsWorkerComponent(string(component.ComponentType)) {
+			continue
+		}
+		if component.PodTemplate == nil {
+			component.PodTemplate = &corev1.PodTemplateSpec{}
+		}
+		if component.PodTemplate.Labels == nil {
+			component.PodTemplate.Labels = make(map[string]string)
+		}
+		component.PodTemplate.Labels[commonconsts.KubeLabelDynamoWorkerHash] = workerHash
```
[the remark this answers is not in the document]
tmonty12: Confirmed this does not feed a runtime namespace: Grove writes its canonical hash onto PodClique templates; the checkpoint label is applied after the checkpoint template is built and is not consumed as DYN_NAMESPACE_WORKER_SUFFIX. The component-rollout active-hash compatibility path does not render Grove workloads. No code change is needed.

deploy/operator/internal/controller/testing/grove.io/podcliquesets.yaml
```diff
@@ -0,0 +1,23 @@
...
+    singular: podcliqueset
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    schema:
+      openAPIV3Schema:
+        type: object
+        x-kubernetes-preserve-unknown-fields: true
```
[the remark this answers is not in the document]
tmonty12: Checked: the only no-Grove-API case injects meta.NoKindMatchError in a unit test; no envtest depends on CRD absence, and no test writes Grove child status through these stubs. The shared CRDs are required for the Grove envtest. No code change is needed.

deploy/operator/internal/dynamo/grove.go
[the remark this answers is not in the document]
tmonty12: Addressed in a5e7a85322. evaluateGroveComponents now computes the canonical worker hash once per suffix-enabled readiness pass, carries it in groveRuntimeNamespaceContext, and reuses it for every worker namespace calculation.

deploy/operator/internal/controller/dgd_grove_worker_hash_suffix_envtest_test.go
```diff
@@ -0,0 +1,264 @@
...
+	dgd := newGroveWorkerHashSuffixTestDGD(env.Namespace(), "new-without-pcs")
+	require.NoError(t, env.Client().Create(ctx, dgd))
+
+	createdDGD := &nvidiacomv1beta1.DynamoGraphDeployment{}
+	require.NoError(t, env.Client().Get(ctx, types.NamespacedName{Name: dgd.Name, Namespace: env.Namespace()}, createdDGD))
+	require.Equal(t, createdDGD.GetAnnotations()[consts.AnnotationGroveWorkerHashSuffixEnabled], "true")
+	wantHash, err := dynamo.ComputeDGDWorkersSpecHash(createdDGD)
+	require.NoError(t, err)
+	waitForGroveWorkerHashSuffixes(t, ctx, env, createdDGD, wantHash)
+}
```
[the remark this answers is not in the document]
tmonty12: Addressed in a5e7a85322. Both Grove worker-suffix envtests now narrate environment setup, seeding or creation, controller startup, updates, and assertions with t.Log headings.

tmonty12:
Addressed the non-threaded review nits in a5e7a85322: added test-step narration in the namespace, renderer, defaulting, and envtest coverage; added story comments in the Grove namespace, defaulting, and webhook-bypass paths.

deploy/operator/internal/dynamo/grove.go
```diff
@@ -157,6 +157,25 @@ func evaluateGroveComponents(ctx context.Context, reader client.Reader, dgd *v1b
 	aggregatedReason := ""
 
 	componentStatuses = make(map[string]v1beta1.ComponentReplicaStatus, len(dgd.Spec.Components))
+	pcs, err := getGrovePodCliqueSet(ctx, reader, dgd)
```
julienmancuso: **[P1] Use the PodCliqueSet returned by reconciliation for cutover.** After `SyncResource` updates the PCS, `groveWorkloadsReconciler` immediately calls `EvaluateGroveReadiness`, and this `Get` reads through the informer-backed client. The cache can still expose the old PCS generation with `observedGeneration == generation` and the old child as completed. Because the desired namespace is computed from the new DGD hash, that stale snapshot makes status publish the new namespace before Grove has accepted or rolled out the new revision; once persisted, the previous-namespace fallback keeps the premature value. Pass the synced PCS returned by `reconcilePodCliqueSet` into readiness/cutover evaluation, or stop after a PCS modification and wait for its watch event.

deploy/operator/internal/dynamo/grove.go
```diff
@@ -205,6 +229,98 @@ func evaluateGroveComponents(ctx context.Context, reader client.Reader, dgd *v1b
...
+) error {
+	if c == nil {
+		return nil
+	}
+
+	// Publish the namespace selected by the component's accepted revision state.
+	runtimeNamespace, err := getGroveRuntimeNamespace(
+		c.dgd,
+		c.component,
+		revisionState.hasCompletedAcceptedPCSRevision(c.acceptedPCSRevisionHash),
```
julienmancuso: **[P1] Cut over the complete worker generation atomically.** This decision is made independently for each child. In a prefill/decode DGD, prefill can complete the accepted PCS revision first and publish the new namespace while decode still publishes the previous one, producing a mixed/incomplete pipeline even though the canonical suffix represents the whole worker group. Collect every worker child's revision state first, then switch `RuntimeNamespace` for all worker components only after the complete worker group has finished the accepted PCS revision. Please add a two-worker regression where one child is complete and the other is still updating.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -79,21 +81,19 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
...
-			component.Replicas = ptr.To(int32(1))
-		}
+	groveProvider := providerSelected && provider == consts.WorkloadProviderGrove
+	defaultComponentFields(dgd, groveProvider)
 
-		// Default Grove's minimum available replicas only for Grove-selected DGDs.
-		if providerSelected && provider == consts.WorkloadProviderGrove && component.MinAvailable == nil {
-			component.MinAvailable = ptr.To(int32(1))
+	// Drive Grove suffix migration from the immutable provider rather than mutable routing intent.
+	if groveProvider {
```
julienmancuso: **[P1] Do not miss the first worker edit before provider adoption.** A legacy DGD has no workload-provider annotation until the controller adopts its owned workloads. If a user changes a worker during that window, `providerSelected` is false here, so suffix migration is skipped. The controller's later adoption patch adds `workload-provider=grove`, but old and new specs are then identical, so the webhook never records the worker change and the Grove rollout stays unsuffixed—reproducing the original collision. Handle worker-hash transitions for unselected legacy updates (the marker is harmless if the component is later adopted), or persist enough pending state for the adoption update. Test: seed an unselected legacy Grove DGD and PCS, update the worker while the controller is stopped, then start reconciliation.

julienmancuso (commented):
Three remaining P1 correctness findings on the updated state-machine implementation:

deploy/operator/internal/dynamo/grove.go
sttts: **[P1] Namespace cutover must remain correct under arbitrarily stale observations**

I do not think this should be treated as a short cache race that can be solved by waiting for a fresher read. Controller reads may legitimately be stale for seconds or longer, and reconciliation may fail or restart after any successful write. The state machine must remain safe for every such interleaving and eventually converge using normal watch events or requeues; it must not require live/quorum reads or any bound on informer latency.

The problematic interleaving is that completion is proven for one rollout state while the namespace is computed from another.

Starting point, abbreviated:

```yaml
kind: DynamoGraphDeployment
metadata:
  name: example
  generation: 12
spec:
  components:
  - componentName: prefill
    # Worker spec hashes to B
status:
  components:
    prefill:
      runtimeNamespace: example-prefill-A
---
kind: PodCliqueSet
metadata:
  name: example
  generation: 7
status:
  observedGeneration: 7
  currentGenerationHash: pcs-revision-A
---
kind: PodClique
metadata:
  name: example-prefill
status:
  currentPodCliqueSetGenerationHash: pcs-revision-A
  replicas: 2
  updatedReplicas: 2
  updateProgress:
    updateEndedAt: "..."
```

`SyncResource` can now update the PCS to generation 8. Immediately after the write, the API-server object may look like:

```yaml
kind: PodCliqueSet
metadata:
  generation: 8
spec:
  # Desired worker state B
status:
  observedGeneration: 7
  currentGenerationHash: pcs-revision-A
```

Meanwhile, a cached read may still return the complete generation-7 object. The current decision then effectively becomes:

```go
acceptedRevision := acceptedRevisionFrom(cachedPCS) // pcs-revision-A
completed := child.currentRevision == acceptedRevision &&
    child.updatedReplicas == child.desiredReplicas &&
    child.updateEnded

desiredWorkerHash := ComputeDGDWorkersSpecHash(dgd) // B
runtimeNamespace := namespaceFor(desiredWorkerHash) // example-prefill-B

if completed {
    publish(runtimeNamespace) // Publishes B based on completion of A
}
```

This is not safe merely because another event will arrive soon. Once `example-prefill-B` is persisted as `previousNamespace`, the next reconcile can observe the current PCS generation 8 with `observedGeneration: 7` and correctly conclude that B is not complete, but its fallback is now already B. Thus delivery of the fresh PCS watch event does not restore A. The status only becomes correct when B eventually completes, or remains incorrect indefinitely if that rollout stalls.

A transient failure makes the required semantics clearer. Assume the reconcile successfully updates the PCS and then exits before computing or persisting readiness. The next reconcile may see either:

| Observed PCS state | Required active namespace |
|---|---|
| stale generation 7, A complete | A |
| generation 8, still reporting accepted A | A |
| generation 8, B accepted, child still on A | A |
| generation 8, child partially updated to B | A |
| generation 8, accepted B fully completed | B |

Repeating any of the first four observations for an arbitrary number of reconciles must remain safe. No elapsed-time assumption should affect the result.

I think the implementation should make the following invariants explicit:

1. **Causal cutover:** B may only become active based on a completion certificate that belongs to B. Completion of A may preserve A, but must never activate B.
2. **Unbounded staleness:** A stale PCS or child snapshot must be a valid input, even if it remains stale across many reconciles.
3. **Crash safety:** Any prefix of a reconcile, including “PCS write succeeded, everything afterwards failed,” must be recoverable by a new reconcile with no in-memory context.
4. **Mixed-snapshot safety:** PCS and child objects can be observed at different rollout stages. Parent A/child B and parent B/child A must both remain safe.
5. **Durable status semantics:** `runtimeNamespace` is externally consumed as the active namespace. A pending target must not be exposed through that field.
6. **Eventual convergence:** Once B is durably accepted and its corresponding child revision is complete, ordinary watch delivery or requeue must eventually publish B.
7. **No stronger-read dependency:** Correctness must not depend on an API-server read, quorum read, sleeps, or an assumed cache propagation delay.

There are several possible designs that could satisfy these requirements:

- Carry the PCS returned by reconciliation through readiness evaluation, keeping decisions causally tied to the object that was reconciled.
- Introduce a reconciliation phase boundary after a PCS mutation and make subsequent observations validate the desired target before changing the active namespace.
- Persist explicit active and pending rollout identities, for example:

  ```yaml
  status:
    components:
      prefill:
        runtimeNamespace: example-prefill-A
        pendingRuntimeNamespace: example-prefill-B
        pendingWorkerHash: B
        pendingPCSGeneration: 8
  ```

- Attach the runtime identity to the PCS generation/revision itself and derive the namespace from the revision whose completion was observed, rather than combining an accepted Grove revision with the current DGD spec:

  ```yaml
  kind: PodCliqueSet
  metadata:
    annotations:
      nvidia.com/dynamo-worker-hash: B
  status:
    observedGeneration: 8
    currentGenerationHash: pcs-revision-B
  ```

These approaches have different API and complexity trade-offs and can also be combined. I am not requesting a specific one here. The important part is that the implementation and tests demonstrate the invariants above for arbitrarily stale observations and for a reconcile interrupted after each externally visible write.

deploy/operator/internal/controller/dgd_grove_workload_renderer.go
```diff
@@ -110,10 +113,41 @@ func (r *groveWorkloadRenderer) Render(
...
+		return nil, err
+	}
+	return renderDeployment, nil
+}
+
+func applyGroveWorkerHashSuffix(
+	renderDeployment *nvidiacomv1beta1.DynamoGraphDeployment,
+	hashSource *nvidiacomv1beta1.DynamoGraphDeployment,
+) error {
+	if hashSource.GetAnnotations()[commonconsts.AnnotationGroveWorkerHashSuffixEnabled] != commonconsts.KubeLabelValueTrue {
```
sttts: **[P1] Keep the legacy namespace migration local to PCS reconciliation**

I **strongly recommend** removing the permanent DGD-level `grove-worker-hash-suffix-enabled` annotation and the admission state machine around it.

The annotation does not represent rollout completion. It is a permanent compatibility latch saying that this DGD has crossed the legacy namespace boundary and that every future Grove worker revision must be rendered with a hash suffix. Admission sets that latch before the first migrated rollout starts, based on an `oldDGD`/`newDGD` worker-hash comparison, and it can never be removed afterwards.

That makes correctness depend on admission observing the first relevant edit. This is not reliable for existing DGDs because `workload-provider` is also introduced and adopted asynchronously in the same release: a worker edit can be persisted before provider adoption, after which the original A→B transition is no longer visible in `oldDGD`/`newDGD`. More generally, admission is a poor place for a durable migration state machine: it cannot reconcile actual workload state, recover level-wise from partial side effects, or converge through watches and retries.

This migration is naturally local to PCS reconciliation, where desired worker state, the actual Grove rollout state, and the active revision are available together. The Grove workload already carries the relevant runtime identity:

```yaml
# Legacy worker template
env:
  - name: DYN_NAMESPACE
    value: example
# no DYN_NAMESPACE_WORKER_SUFFIX
```

```yaml
# Hashed worker template
labels:
  nvidia.com/dynamo-worker-hash: B
env:
  - name: DYN_NAMESPACE
    value: example
  - name: DYN_NAMESPACE_WORKER_SUFFIX
    value: B
```

An explicitly configured `DYN_NAMESPACE` does not determine the mode. The relevant distinction is whether the accepted and completed worker revision has a suffix:

```go
switch {
case suffix == "" && workerHash == "":
    mode = LegacyUnsuffixed
case suffix != "" && suffix == workerHash:
    mode = Hashed
default:
    mode = UnknownOrInconsistent
}
```

The local, level-triggered state machine can then distinguish these states:

```text
No existing PCS / no active revision:
    Render new workers with the canonical hash suffix.

Existing completed revision is legacy and desired worker spec is unchanged:
    Preserve the legacy PCS without triggering an upgrade rollout.

Existing completed revision is legacy A and desired worker spec is B:
    Render the pending PCS revision with suffix B.
    Keep publishing A as the active runtime namespace.

Accepted revision B and all relevant worker children have completed B:
    Publish namespace base-B.
    The completed revision itself now proves that the PCS uses hashed namespaces.
```

This also handles an edit made before provider adoption without preserving admission history:

```text
DGD desired worker state = B
Existing owned PCS still runs legacy A

→ provider adoption discovers the owned PCS
→ PCS reconciliation compares actual A with desired B
→ hashed rollout B starts
```

It is important that “read this from the PCS” means the identity of the accepted and completed Grove revision, not merely the current mutable PCS spec. During an update the PCS spec may already describe B while Grove status and children still run A. The namespace transition must remain causally tied to the completed revision.

For example, this is a valid observed server-side snapshot during rollout, not an apply-ready manifest:

```yaml
kind: PodCliqueSet
metadata:
  generation: 8
spec:
  template:
    cliques:
      - labels:
          nvidia.com/dynamo-worker-hash: B
        spec:
          podSpec:
            containers:
              - name: main
                env:
                  - name: DYN_NAMESPACE
                    value: example
                  - name: DYN_NAMESPACE_WORKER_SUFFIX
                    value: B
status:
  observedGeneration: 7
  currentGenerationHash: grove-revision-A
```

Here `spec` describes target B, but the accepted active revision is still A. Reading only the current template would publish B prematurely.

If Grove does not expose enough active-revision identity to derive this without extra state, a temporary annotation on the operator-owned PCS would still be preferable:

```yaml
kind: PodCliqueSet
metadata:
  annotations:
    nvidia.com/dynamo-legacy-worker-namespace: "true"
```

Its exact meaning would be: “the currently active revision of this PCS is still legacy and unsuffixed.” The Grove reconciler would set it as its first action when adopting an existing unsuffixed PCS, return/requeue, retain it while rolling out hashed revision B, and remove it only after the accepted revision and all relevant children have completed B. Unlike the current DGD annotation, this state is local to the owned workload and disappears after the one-time migration.

Whichever representation is used, correctness must hold with arbitrarily stale observations. Repeated stale observations of A must preserve A; observing desired B with stale or unaccepted status must also preserve A; only a completion proof causally belonging to B may publish B. Normal watch delivery and requeues must be sufficient, without live reads, sleeps, or assumptions about informer latency.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -79,21 +81,19 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 		return nil
 	}
 
+	// This annotation is operator-owned and cannot be supplied by a request.
+	delete(dgd.Annotations, consts.AnnotationGroveWorkerHashSuffixEnabled)
```
sttts: **[P1] Mutating admission must not delete or rewrite client values**

Deleting this annotation from the admitted object is forbidden by the mutating-admission contract we need to preserve. Mutating defaulting must be additive: it may populate an absent value, but it must never silently delete or change an explicit value supplied by the client. If a field is operator-owned and clients are not allowed to set or change it, a validating webhook must reject that request instead of accepting it and rewriting the object.

Silently removing or replacing client metadata breaks declarative convergence. ArgoCD, SSA clients, and other controllers may continuously restore their desired value while this webhook continuously removes or changes it, leaving the resource permanently out of sync or causing a reconciliation hot loop. Calling the annotation “operator-owned” does not make destructive admission mutation safe.

For example, a declarative client may continuously apply:

```yaml
kind: DynamoGraphDeployment
metadata:
  name: example
  annotations:
    nvidia.com/grove-worker-hash-suffix-enabled: "false"
```

while admission stores either:

```yaml
# Observed live object after admission, not an apply-ready manifest.
kind: DynamoGraphDeployment
metadata:
  name: example
  annotations:
    nvidia.com/grove-worker-hash-suffix-enabled: "true" # or absent
```

The API request succeeds, but desired and live state can never converge. The correct behavior for a forbidden client value is an explicit validation error, not silent mutation.

Admission is also the wrong place for this migration state machine. The transition depends on actual PCS/revision state, asynchronous provider adoption, successful workload writes, and eventual rollout completion. Admission observes a single request and cannot recover level-wise from a controller crash, stale child state, or an update that happened before `workload-provider` was adopted. Please remove the suffix migration state machine from admission rather than trying to harden these `delete`/restore rules; keep migration state and convergence in PCS reconciliation, close to the workload it describes.

The pre-adoption case makes that ordering dependency concrete:

```yaml
# Existing client-owned DGD immediately after the operator upgrade.
kind: DynamoGraphDeployment
metadata:
  name: legacy
  annotations: {} # workload-provider has not been adopted yet
spec:
  # User has already changed worker state A -> B.
---
# Existing operator-owned workload still represents A.
apiVersion: grove.io/v1alpha1
kind: PodCliqueSet
metadata:
  name: legacy
spec:
  template:
    # Legacy A templates have no DYN_NAMESPACE_WORKER_SUFFIX.
```

Admission cannot classify this migration from the DGD annotations, while PCS reconciliation can compare actual A with desired B whenever the controller resumes.

deploy/operator/internal/dynamo/grove.go
tmonty12: Addressed by the current level-based flow: readiness receives the PodCliqueSet returned by SyncObservedResource, not a second cache read. A stale or unaccepted returned PCS cannot cut over because EvaluateGroveReadiness requires observedGeneration == generation. The Grove workloads/readiness regression tests cover this path.

deploy/operator/internal/dynamo/grove.go
tmonty12: Addressed: readiness now derives one cohort-wide workersCompleted value from every worker accepted-PCS revision state. Until all workers complete, each worker retains its prior runtimeNamespace; TestEvaluateGroveReadinessSwitchesWorkersAtomically covers one worker still updating and the all-complete cutover.

deploy/operator/internal/dynamo/grove.go
tmonty12: Addressed by the level-based PCS/DGD boundary. The reconciler evaluates the PCS returned by the successful write; a PCS is accepted only after Grove observed exactly its generation, and every child must prove that same accepted revision. Otherwise runtimeNamespace remains the previously published value. Tests cover PCS sync conflict, DGD commit conflict after PCS persistence, a fresh-reconciler retry, and idempotence, without uncached reads or cache-latency assumptions.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
tmonty12: Addressed by removing the annotation and its admission state machine entirely. Defaulting no longer deletes, rewrites, or interprets nvidia.com/grove-worker-hash-suffix-enabled; suffix mode is now derived level-wise during PCS reconciliation. A client-provided legacy annotation is therefore left untouched and has no behavioral effect.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
tmonty12: Thanks — this was valid against the earlier admission-based implementation, where suffix migration was gated on providerSelected.

The PR has since removed that admission state machine entirely. Provider adoption is now a separate first controller action: it adopts Grove from the owned PCS and persists workload-provider=grove without changing the DGD existing worker-generation annotations.

The subsequent Grove reconciliation does not consult provider selection when deciding whether to render the suffix. For a legacy Grove DGD with current worker hash A, a desired worker spec B produces workerGenerationChanged=true; the renderer writes a suffixed PCS for B, and only after that PCS sync succeeds does it commit B as the current worker hash.

Therefore, an A→B update is no longer skipped merely because the DGD had not yet materialized workload-provider=grove. The pre-provider-adoption window described here no longer exists in the current flow.

deploy/operator/internal/controller/dgd_grove_workload_renderer.go
tmonty12: Thanks — I agree with the core direction: this migration state belongs to PCS reconciliation, not DGD admission.

The PR now removes grove-worker-hash-suffix-enabled and the admission state machine completely. There is no DGD-level migration latch.

PCS reconciliation is now level-based:

- A new PCS renders worker hash labels and DYN_NAMESPACE_WORKER_SUFFIX.
- An existing legacy PCS remains legacy when the worker generation is unchanged.
- A worker-generation change renders a suffixed desired PCS.
- The PCS sync is the write barrier; the DGD worker hash is committed only after that sync succeeds.

Runtime namespace publication is likewise tied to PCS state. Readiness receives the PCS returned by reconciliation, accepts it only when observedGeneration == generation, and obtains the worker hash from the worker labels in that accepted PCS revision. It verifies that all worker cliques agree on the mode and hash, then switches all worker runtime namespaces together only once every worker child has completed that same accepted PCS revision. Until then, it retains the previously published namespace.

This removes the admission/provider-ordering dependency and keeps both migration mode and cutover evidence local to the workload that carries them.

sttts (commented):
Two inline notes from the re-review: one actionable golden-test issue and one explicitly non-actionable consistency note.

deploy/operator/internal/dynamo/grove.go
```diff
@@ -151,60 +131,257 @@ func ClassifyGroveReadiness(ctx context.Context, reader client.Reader, dgd *v1be
...
+	if !IsWorkerComponent(string(component.ComponentType)) {
+		return baseNamespace
+	}
+
+	previousNamespace := dgd.Status.Components[component.ComponentName].RuntimeNamespace
+	if p.acceptedPCSRevisionHash == nil || !p.workersCompleted {
+		return previousNamespace
+	}
+	if !p.workersUseHashSuffix {
+		return baseNamespace
```
sttts, line 297: Non-blocking documentation note — this is intentionally **not a P1 and should not be fixed in this PR**. After leader handover, a new instance may temporarily observe the fresh DGD status (`base-B`) together with a coherent but older legacy PCS/child snapshot and therefore project `base` here again. That is the fundamental cross-object cache-skew problem of any multi-object reconciler; reads may be arbitrarily stale and are not causally consistent across GVKs. Trying to prevent it here with a quorum/direct read or an assumed ordering between caches would be the wrong contract: this eventual-consistency behavior is part of the controller mechanism and the state machine must tolerate it. This path only recomputes DGD status; it does not mutate PCS or Pods, combine generations into an irreversible action, or otherwise compromise safety. Later PCS/child watch events (or a requeue) must converge the projection back to `base-B`. So please do **not** add a special fix for this observation; I am recording it only to make that consistency model explicit.

deploy/operator/internal/controller/testdata/dgd/grove/output.yaml
```diff
@@ -113,6 +113,7 @@ spec:
                 nvidia.com/dynamo-sub-component-type: decode
                 nvidia.com/metrics-enabled: "true"
                 nvidia.com/selector: grove-manifest-decode
+                nvidia.com/dynamo-worker-hash: 97e6b617
```
sttts: [P2] Please avoid hard-coding a namespace-derived worker hash in this cluster golden. `ComputeDGDWorkersSpecHash` includes `GetDCDKubeLabels(dcd)`, including `nvidia.com/dynamo-namespace`, while these tests create a generated `clusterenv-*` namespace. The hash therefore changes on every run: on this head the golden expects `97e6b617`, but Operator Integration rendered `68a00e88`; the DGDR golden likewise expects `2f745c61` and rendered `590af4d6`. Regenerating the literals only fixes one namespace. Please use a deterministic test namespace or capture and reuse the dynamic hash in the matcher. Independent `$pattern:[0-9a-f]{8}` values would be weaker because they would stop asserting that this label and every `DYN_NAMESPACE_WORKER_SUFFIX` occurrence carry the same hash.

deploy/operator/internal/dynamo/grove.go
sttts: Correction / historical context after tracing this further: my conclusion above that this “should not be fixed” is too absolute. This is the long-standing Kubernetes [time-travel bug #59848](https://github.com/kubernetes/kubernetes/issues/59848) raised by Clayton in 2018; the issue explicitly calls out controller lease failover. An initial Reflector LIST with `resourceVersion=0` may be served from an arbitrarily stale API-server watch cache, so a restarted component or a new leader can initialize its local state in the past.

Marek's [KEP-2340, Consistent Reads from Cache](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md), made a most-recent LIST scalable: the API server performs a small quorum read against etcd to obtain the current revision, waits until its watch cache has reached that revision, and then serves the objects from the cache. For informer initialization, [KEP-3157, WatchList](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/3157-watch-list/README.md), is the relevant client path: `resourceVersion=""`, `NotOlderThan`, and `SendInitialEvents=true` provide the freshness barrier. The ordinary LIST fallback still uses `resourceVersion=0` and does not provide it.

controller-runtime has an important nuance here: the manager's cache/cluster runnable starts before leader election, but informers are lazy and controller sources normally start only from `OnStartedLeading` (`EnableWarmup=false` by default). That gives us the desired handoff property only if (1) no non-leader consumer has already requested the relevant informer and (2) the initial WatchList/LIST has most-recent semantics. Neither is a blanket framework guarantee: this operator's non-leader-elected resource counter already lists DGDs through the cached client, and Dynamo supports Kubernetes 1.30+, where WatchList is not universally available/default and client-go may fall back to the stale `resourceVersion=0` LIST.

The design lesson is therefore: we should not add direct/quorum reads to every reconcile, but we **must** establish or document a startup invariant that a new leader does not act until all relevant informer snapshots are at least as fresh as the leadership handoff. Merely starting an informer after election is insufficient when its initial request permits stale data. With that invariant the transient regression discussed above is excluded; without it, dismissing the scenario as ordinary eventual consistency is not justified.

sttts (approved):

deploy/operator/internal/dynamo/grove.go
```diff
@@ -151,60 +131,257 @@ func ClassifyGroveReadiness(ctx context.Context, reader client.Reader, dgd *v1be
...
+// accepted PCS revision. A nil acceptedPCSRevisionHash means no revision has
+// been accepted yet.
+func (s groveComponentRevisionState) hasCompletedAcceptedPCSRevision(acceptedPCSRevisionHash *string) bool {
+	return acceptedPCSRevisionHash != nil &&
+		s.generationObserved &&
+		s.currentPCSRevisionHash != nil &&
+		*s.currentPCSRevisionHash == *acceptedPCSRevisionHash &&
+		s.replicas == s.desiredReplicas &&
+		s.updatedReplicas == s.desiredReplicas &&
+		s.updateEnded
```
julienmancuso: **[P1] Allow the initial Grove revision to complete without `UpdateProgress`.** Grove's first PodClique realization sets `CurrentPodCliqueSetGenerationHash` directly when the template is current and leaves `UpdateProgress` nil; `UpdateProgress`/`UpdateEndedAt` is created only for a subsequent rollout. Consequently, a new suffixed DGD can be fully ready with a matching accepted revision but fail this predicate forever, leaving `status.components[worker].runtimeNamespace` empty. The matching current PCS revision hash is already Grove's certificate that the revision is fully realized; please use it with the replica checks without requiring `UpdateEndedAt` (or explicitly accept nil progress for initial realization). A regression should cover an initially ready PodClique with a matching revision and nil `UpdateProgress`.

deploy/operator/internal/controller/dgd_worker_rollout_reconciler.go
```diff
@@ -69,42 +83,73 @@ func (r *dgdWorkerRolloutReconciler) ReconcileUnsupported(
...
 	if err != nil {
-		logger.Error(err, "Failed to check rolling update trigger for unsupported pathway")
-		return failWorkloadProgram(reasonRollingUpdateFailed, err)
+		return unsupportedWorkerHashTransition{}, err
 	}
-	if !triggerRollingUpdate {
-		return nil
+
+	current := r.currentWorkerHashes(dgd)
+	if current.empty() {
```
julienmancuso, line 116: **[P1] Detect a worker edit before establishing the legacy baseline.** For an existing unsuffixed PCS, an empty DGD current-worker hash is treated as initialization, so `workerGenerationChanged` remains false even if the PCS still represents worker spec A and the current DGD already requests B. This path renders and syncs B without a suffix and then commits B as the baseline, recreating the old/new namespace collision during the rollout. Please initialize from a legacy PCS only after confirming its worker cliques already match the desired worker generation; otherwise treat the observed worker delta as a suffix migration. The current envtest lets reconciliation commit A before editing—please add the inverse ordering: seed legacy A, edit the DGD to B while the controller is stopped, then start reconciliation.

julienmancuso (commented):
Two remaining P1 correctness gaps in the updated Grove namespace-cutover flow:

deploy/operator/internal/controller/dgd_worker_rollout_reconciler.go
tmonty12: I considered preserving the legacy PCS only when it is unchanged, while suffixing it when the existing PCS represents `A` and the desired DGD represents `B`.

- Grove DGDs have persisted `current-worker-hash` since 1.2.0. Therefore, once a DGD has been successfully reconciled by a 1.2+ operator, the normal first worker update is already observable as `current=A`, `desired=B` and is suffixed by this change.

- The remaining edge case is an existing legacy PCS with no persisted DGD hash when the controller first observes a worker change. I cannot distinguish `PCS=A, desired=A` from `PCS=A, desired=B` using the DGD hash: initialization necessarily sets the missing hash to the current desired value.

- I could conservatively suffix every legacy PCS with an empty DGD hash, but that causes a one-time rolling recreate even for an unchanged `A → A` deployment. Avoiding that rollout would require comparing the existing PCS worker rollout generation with the desired PCS generation. That was the purpose of the persisted worker hash; reproducing or depending on Grove PCS generation-hash calculation here would add non-trivial cross-controller coupling.

Given this is limited to an update that was not previously observed and committed by a 1.2+ operator, I am keeping the simpler behavior. Once the current hash is initialized, subsequent worker changes are detected and suffixed normally.

deploy/operator/internal/dynamo/grove.go
tmonty12: Fixed in 4221814f89.

- `UpdateProgress=nil` is now represented separately from a non-nil, unfinished update. An accepted PCS revision can therefore complete during initial realization once its observed revision and replica counts match.
- A non-nil `UpdateProgress` still requires `UpdateEndedAt`, so rollout cutover continues to preserve the prior runtime namespace until the tracked update completes.

The regression coverage includes an initially ready PodClique with a matching accepted revision, plus nil-progress zero-replica PodClique and PCSG cases. `go test ./internal/dynamo -count=1` passes.

julienmancuso (approved):

PR #12772, opened 2026-08-06
perf(mocker): optimize replay accounting and SGLang KV events

jthomson04:
## Overview

This PR removes three allocation- and hash-heavy paths from offline replay: decode accept-length rescans, per-pass token-timeline alignment, and hash-indexed SGLang physical-page metadata.

## Summary

- Count visible decode tokens and participating requests while scheduler outputs are emitted instead of rescanning `OutputSignal`s through a `HashSet`.
- Record replay token timestamps once at the final pass boundary, including the shared slowest-rank boundary for attention DP.
- Index SGLang physical-page metadata directly by page ID and reduce temporary allocation while publishing stored and removed KV events.
- Keep ordinary decode accounting scan-free while recomputing counters from surviving signals only on the rare live-output suppression path.
- Preserve public APIs, configuration, event formats, and simulation behavior.

## Details

The three commits are intentionally kept as independent benchmark points:

1. `perf(mocker): count accept-length samples during decode`
2. `perf(mocker): record replay tokens at pass completion`
3. `perf(mocker): index SGLang pages directly`

Review feedback is addressed in one follow-up commit:

4. `fix(mocker): address replay review feedback`

The token-accounting change retains a debug-only reference scan. Live cancellation uses the same scan only when suppressing buffered outputs so stale counters self-heal without adding work to ordinary passes. Admissions remain recorded at pass start, while accepted token-bearing outputs are recorded at the existing completion boundary. Single-worker vLLM keeps its rank-local model-completion timestamp across KVBM stall advance; aggregated and attention-DP replay retain their shared completion boundary. The SGLang KV-event path preserves duplicate logical hashes, parent chaining, event IDs, page reuse, and final-copy removal semantics.

- `lib/mocker/src/scheduler/mod.rs` and the SGLang/vLLM scheduler implementations for direct accept-length accounting.
- `lib/mocker/src/replay/collector.rs` and `lib/mocker/src/replay/offline/components/engine.rs` for final-boundary token recording.
- `lib/mocker/src/kv_manager/sglang_backend.rs` for dense physical-page metadata and KV-event construction.

## Performance

Single unprofiled point per commit, no warm-ups. The workload was the deterministic 2x Astra trace (`806487947000d49d4e8a817112e479a8b7282ffb3f388259cb921cf13bc3ff1b`) with 12 aggregated SGLang/B200 workers, KV routing/events, 64-token pages, and the benchmark-only AIC Moka build. Every point completed 187,906 requests and 2,580,832,264 processed tokens.

| Point | Wall time | Processed tokens/s | Incremental throughput | Versus baseline |
| Baseline `e026f9a654` | 84.756 s | 30.450 M | — | — |
| C1 `9d1d72b201` | 76.304 s | 33.823 M | +11.08% | +11.08% |
| C2 `11ce174953` | 71.594 s | 36.048 M | +6.58% | +18.38% |
| C3 `80ad0e77de` | 64.936 s | 39.745 M | +10.25% | +30.52% |
| Review-fix head `f2ce225849` | 65.989 s | 39.110 M | -1.60% | +28.44% |

The final review-fix row is compared with the preserved C3 result; C3 was not rerun. It completed the same 187,906 requests, 2,401,594,588 input tokens, 179,237,676 output tokens, and 2,580,832,264 processed tokens, with prefix-cache reuse 0.920969. Its 1.60% throughput regression versus C3 misses the requested 1% point gate, while the fully reviewed head remains 28.44% faster than baseline. These are point comparisons, not statistical claims. The deterministic canonical replay validation below establishes exact behavioral parity.

## Validation

The full offline replay correctness-parity protocol passed between exact baseline `e026f9a6544f28032f0ce0e070b05ba3a5e5bea6` and final review-fix head `f2ce22584985120dcc3a953851a716b9b8d904d6`.

- Exact 5,000-request Mooncake fixture, SHA-256 `3892ae19ae480b643155f0c6b9d798591cbe2e73bec6a0fa5ae3d3bc0332fb8a`.
- Two independent processes per revision for all four canonical rows.
- Byte-identical canonical reports for aggregated and disaggregated SGLang.
- Byte-identical canonical reports for aggregated and disaggregated native-vLLM KVBM stress.
- Every report completed all 5,000 requests; disaggregated rows exercised 5,000 handoffs.
- KVBM rows exercised offload/host-pinned lifecycles and bounded preemption (2 aggregated, 1 disaggregated).

Additional checks:

- `cargo test -p dynamo-mocker` (716 passed)
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline` (192 passed; two failures reproduced unchanged at the exact baseline)
- `git diff --check`

The baseline-reproduced KVBM failures are:

- `busy_worker_advances_offload_without_destination_admission`
- `aggregated_canonical_evidence_covers_real_kvbm_lifecycle`

## Related Issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cache event reporting when physical pages are reused, ensuring removals and replacements receive distinct identifiers.
  * Improved cancellation handling so token and decode-forward metrics remain accurate.
  * Corrected token timing records to reflect scheduler pass completion.

* **Improvements**
  * Improved acceptance-length metrics for standard and speculative decoding.
  * Added validation and coverage for cache events, cancellation behavior, and acceptance metrics.

### DIFF
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -21,6 +21,10 @@ new_key_type! {
 pub struct KvPageId(usize);
 
 impl KvPageId {
+    pub(crate) fn index(self) -> usize {
+        self.0
+    }
+
     #[cfg(test)]
     pub(crate) fn from_token_index(index: usize, page_size: usize) -> Self {
         Self(index / page_size)
@@ -132,6 +136,7 @@ impl PagePool {
     }
 
     pub fn free_pages(&mut self, pages: &[KvPageId]) {
+        self.free.reserve_exact(pages.len());
         self.free.extend_from_slice(pages);
     }
 
--- a/lib/mocker/src/cache/radix_cache.rs
+++ b/lib/mocker/src/cache/radix_cache.rs
@@ -136,7 +136,6 @@ impl PagePool {
     }
 
     pub fn free_pages(&mut self, pages: &[KvPageId]) {
-        self.free.reserve_exact(pages.len());
         self.free.extend_from_slice(pages);
     }
 
--- a/lib/mocker/src/kv_manager/sglang_backend.rs
+++ b/lib/mocker/src/kv_manager/sglang_backend.rs
@@ -118,9 +118,9 @@ pub struct SglangKvManager {
     kv_event_publishers: KvEventPublishers,
     dp_rank: u32,
     next_event_id: u64,
-    /// Maps each physical page to the block hash assigned during Stored
-    /// events, so Removed events can use the same block hash.
-    page_to_block_hash: FxHashMap<KvPageId, ExternalSequenceBlockHash>,
+    /// Maps each dense physical page ID to the block hash assigned during
+    /// Stored events, so Removed events can use the same block hash.
+    page_to_block_hash: Vec<Option<ExternalSequenceBlockHash>>,
     /// Tracks how many live pool slots currently advertise the same logical
     /// block hash so router events reflect logical block visibility, not
     /// transient slot ownership.
@@ -184,7 +184,7 @@ impl SglangKvManager {
             kv_event_publishers,
             dp_rank,
             next_event_id: 0,
-            page_to_block_hash: FxHashMap::default(),
+            page_to_block_hash: vec![None; total_tokens / page_size],
             block_hash_refcounts: FxHashMap::default(),
         }
     }
@@ -808,70 +808,60 @@ impl SglangKvManager {
             pages.len()
         );
 
-        let first_block_start = first_new_token / block_size * block_size;
-        let Some(first_unpublished_block) = (first_block_start..complete_len)
-            .step_by(block_size)
-            .find(|&block_start| {
-                let page = pages[block_start / block_size];
-                !self.page_to_block_hash.contains_key(&page)
-            })
+        let first_page = first_new_token / block_size;
+        let Some(first_unpublished_page) = (first_page..complete_pages)
+            .find(|&page_idx| self.page_to_block_hash[pages[page_idx].index()].is_none())
         else {
             return 0;
         };
 
-        let mut computed_blocks = Vec::new();
-        let first_unpublished_page = first_unpublished_block / block_size;
         let local_hashes = &page_hashes[first_unpublished_page..complete_pages];
+        self.block_hash_refcounts.reserve(local_hashes.len());
+        let mut parent_hash = None;
+        let mut blocks = Vec::new();
+        let mut publishing = false;
 
         for (block_idx, tokens_hash) in local_hashes.iter().copied().enumerate() {
-            let block_start = first_unpublished_block + block_idx * block_size;
-            let page_idx = block_start / block_size;
+            let page_idx = first_unpublished_page + block_idx;
             let page = pages[page_idx];
-            if self.page_to_block_hash.contains_key(&page) {
+            let page_slot = page.index();
+            if self.page_to_block_hash[page_slot].is_some() {
                 continue;
             }
 
-            let parent_hash = if block_start == 0 {
+            let block_parent_hash = if page_idx == 0 {
                 None
             } else {
-                self.page_to_block_hash.get(&pages[page_idx - 1]).copied()
+                self.page_to_block_hash[pages[page_idx - 1].index()]
             };
-            let block_hash = match parent_hash {
+            let block_hash = match block_parent_hash {
                 Some(parent_hash) => {
                     ExternalSequenceBlockHash(compute_next_seq_hash(parent_hash.0, tokens_hash))
                 }
                 None => ExternalSequenceBlockHash(tokens_hash.0),
             };
 
-            self.page_to_block_hash.insert(page, block_hash);
+            self.page_to_block_hash[page_slot] = Some(block_hash);
             let refcount = self.block_hash_refcounts.entry(block_hash).or_default();
             *refcount += 1;
-            computed_blocks.push((
-                parent_hash,
-                KvCacheStoredBlockData {
+            if *refcount == 1 && !publishing {
+                publishing = true;
+                parent_hash = block_parent_hash;
+                blocks.reserve_exact(local_hashes.len() - block_idx);
+            }
+            if publishing {
+                blocks.push(KvCacheStoredBlockData {
                     block_hash,
                     tokens_hash,
                     mm_extra_info: None,
-                },
-                *refcount,
-            ));
+                });
+            }
         }
 
         let hashed_blocks = local_hashes.len();
-
-        let first_new = computed_blocks
-            .iter()
-            .position(|(_, _, refcount)| *refcount == 1);
-        let Some(first_new) = first_new else {
+        if blocks.is_empty() {
             return hashed_blocks;
-        };
-
-        let parent_hash = computed_blocks[first_new].0;
-        let blocks = computed_blocks
-            .into_iter()
-            .skip(first_new)
-            .map(|(_, block, _)| block)
-            .collect();
+        }
 
         let event = KvCacheEvent {
             event_id: self.next_event_id,
@@ -897,19 +887,21 @@ impl SglangKvManager {
         }
 
         let mut block_hashes = Vec::new();
+        block_hashes.reserve_exact(evicted_pages.len());
         for &page in evicted_pages {
-            let Some(block_hash) = self.page_to_block_hash.remove(&page) else {
+            let Some(block_hash) = self.page_to_block_hash[page.index()].take() else {
                 continue;
             };
-            let Some(refcount) = self.block_hash_refcounts.get_mut(&block_hash) else {
-                continue;
-            };
-            if *refcount > 1 {
-                *refcount -= 1;
-                continue;
+            if let std::collections::hash_map::Entry::Occupied(mut entry) =
+                self.block_hash_refcounts.entry(block_hash)
+            {
+                if *entry.get() > 1 {
+                    *entry.get_mut() -= 1;
+                } else {
+                    entry.remove();
+                    block_hashes.push(block_hash);
+                }
             }
-            self.block_hash_refcounts.remove(&block_hash);
-            block_hashes.push(block_hash);
         }
 
         if block_hashes.is_empty() {
@@ -1341,6 +1333,40 @@ mod tests {
         assert_eq!(
             store.blocks[0].block_hash,
             ExternalSequenceBlockHash(expected_sequence[0])
+        );
+    }
+
+    #[test]
+    fn reused_physical_page_replaces_dense_event_metadata() {
+        let sink = Arc::new(MockSink::new());
+        let mut mgr =
+            SglangKvManager::new(1, 1, KvEventPublishers::new(Some(sink.clone()), None), 0);
+
+        let first = mgr.allocate_for_request(&[1]).unwrap();
+        let page = first.lease.pages()[0];
+        assert!(mgr.abort(first.lease));
+
+        let second = mgr.allocate_for_request(&[2]).unwrap();
+        assert_eq!(second.lease.pages(), &[page]);
+
+        let events = sink.clone_events();
+        assert_eq!(events.len(), 3);
+        assert_eq!(events[0].event_id, 0);
+        assert_eq!(events[1].event_id, 1);
+        assert_eq!(events[2].event_id, 2);
+        let KvCacheEventData::Stored(first_store) = &events[0].data else {
+            panic!("expected first stored event");
+        };
+        let KvCacheEventData::Removed(remove) = &events[1].data else {
+            panic!("expected removal before page reuse");
+        };
+        let KvCacheEventData::Stored(second_store) = &events[2].data else {
+            panic!("expected replacement stored event");
+        };
+        assert_eq!(remove.block_hashes, vec![first_store.blocks[0].block_hash]);
+        assert_ne!(
+            first_store.blocks[0].block_hash,
+            second_store.blocks[0].block_hash
         );
     }
 
--- a/lib/mocker/src/kv_manager/sglang_backend.rs
+++ b/lib/mocker/src/kv_manager/sglang_backend.rs
@@ -179,14 +179,24 @@ impl SglangKvManager {
         kv_event_publishers: KvEventPublishers,
         dp_rank: u32,
     ) -> Self {
+        let page_to_block_hash = if kv_event_publishers.is_empty() {
+            Vec::new()
+        } else {
+            vec![None; total_tokens / page_size]
+        };
         Self {
             cache: RadixCache::new(total_tokens, page_size),
             kv_event_publishers,
             dp_rank,
             next_event_id: 0,
-            page_to_block_hash: vec![None; total_tokens / page_size],
+            page_to_block_hash,
             block_hash_refcounts: FxHashMap::default(),
         }
+    }
+
+    #[cfg(test)]
+    fn page_metadata_len(&self) -> usize {
+        self.page_to_block_hash.len()
     }
 
     pub fn cache(&self) -> &RadixCache {
@@ -794,6 +804,10 @@ impl SglangKvManager {
         if self.kv_event_publishers.is_empty() {
             return 0;
         }
+        if self.page_to_block_hash.is_empty() {
+            self.page_to_block_hash
+                .resize(self.cache.total_tokens() / self.cache.page_size(), None);
+        }
 
         let block_size = self.cache.page_size();
         let complete_len =
@@ -816,7 +830,6 @@ impl SglangKvManager {
         };
 
         let local_hashes = &page_hashes[first_unpublished_page..complete_pages];
-        self.block_hash_refcounts.reserve(local_hashes.len());
         let mut parent_hash = None;
         let mut blocks = Vec::new();
         let mut publishing = false;
@@ -847,7 +860,6 @@ impl SglangKvManager {
             if *refcount == 1 && !publishing {
                 publishing = true;
                 parent_hash = block_parent_hash;
-                blocks.reserve_exact(local_hashes.len() - block_idx);
             }
             if publishing {
                 blocks.push(KvCacheStoredBlockData {
@@ -887,8 +899,7 @@ impl SglangKvManager {
         }
 
         let mut block_hashes = Vec::new();
-        block_hashes.reserve_exact(evicted_pages.len());
-        for &page in evicted_pages {
+        for (page_idx, &page) in evicted_pages.iter().enumerate() {
             let Some(block_hash) = self.page_to_block_hash[page.index()].take() else {
                 continue;
             };
@@ -899,6 +910,9 @@ impl SglangKvManager {
                     *entry.get_mut() -= 1;
                 } else {
                     entry.remove();
+                    if block_hashes.is_empty() {
+                        block_hashes.reserve_exact(evicted_pages.len() - page_idx);
+                    }
                     block_hashes.push(block_hash);
                 }
             }
@@ -990,6 +1004,16 @@ mod tests {
                 _ => None,
             })
             .sum()
+    }
+
+    #[test]
+    fn dense_page_metadata_is_allocated_only_for_kv_events() {
+        let disabled = SglangKvManager::new(64, 4, KvEventPublishers::default(), 0);
+        assert_eq!(disabled.page_metadata_len(), 0);
+
+        let (_buffer, sink) = capture_router_event_sink(ROUTER_TEST_WORKER_ID);
+        let enabled = SglangKvManager::new(64, 4, KvEventPublishers::new(Some(sink), None), 0);
+        assert_eq!(enabled.page_metadata_len(), 16);
     }
 
     #[test]
--- a/lib/mocker/src/replay/collector.rs
+++ b/lib/mocker/src/replay/collector.rs
@@ -1127,34 +1127,17 @@ impl TraceCollector {
         }
     }
 
-    /// Move the tokens emitted by one scheduler pass to a shared completion
-    /// boundary. Scheduler cores record their rank-local end time while the
-    /// pass is formed; attention-DP replay then aligns every rank in the group
-    /// to the slowest rank before the pass becomes externally visible.
-    pub(crate) fn align_pass_token_times(
+    /// Record tokens emitted by one scheduler pass at its externally visible
+    /// completion boundary.
+    pub(crate) fn on_output_signals(
         &mut self,
         output_signals: &[OutputSignal],
         completion_time_ms: f64,
     ) {
-        let mut emitted_by_request = FxHashMap::default();
         for signal in output_signals {
             if signal.token_id.is_some() {
-                *emitted_by_request.entry(signal.uuid).or_insert(0usize) += 1;
+                self.on_token(signal.uuid, completion_time_ms);
             }
-        }
-
-        for (uuid, emitted) in emitted_by_request {
-            let Some(stats) = self.requests.get_mut(&uuid) else {
-                continue;
-            };
-            let TokenTimeline::Recording(times) = &mut stats.token_timeline else {
-                continue;
-            };
-            let start = times
-                .len()
-                .checked_sub(emitted)
-                .expect("scheduler emitted more output signals than collector tokens");
-            times[start..].fill(completion_time_ms);
         }
     }
 
--- a/lib/mocker/src/replay/collector.rs
+++ b/lib/mocker/src/replay/collector.rs
@@ -9,6 +9,7 @@ use std::fmt::{Display, Formatter, Result as FmtResult};
 use uuid::Uuid;
 
 use crate::common::protocols::OutputSignal;
+use crate::scheduler::EnginePassResult;
 
 // 0.1% relative quantile error. The enlarged store covers latency/rate values
 // spanning roughly 10^28 within one sign while remaining bounded (~512 KiB for
@@ -897,6 +898,7 @@ impl TraceCollector {
         }
     }
 
+    #[cfg(test)]
     pub(crate) fn on_prefill_admit(
         &mut self,
         uuid: Uuid,
@@ -904,6 +906,15 @@ impl TraceCollector {
         reused_input_tokens: usize,
     ) {
         self.on_admit(uuid, admit_time_ms, reused_input_tokens);
+        self.on_prefill_pool_admit(uuid, admit_time_ms, reused_input_tokens);
+    }
+
+    pub(crate) fn on_prefill_pool_admit(
+        &mut self,
+        uuid: Uuid,
+        admit_time_ms: f64,
+        reused_input_tokens: usize,
+    ) {
         self.on_pool_admission(
             uuid,
             ReplayRequestPool::Prefill,
@@ -921,6 +932,7 @@ impl TraceCollector {
         }
     }
 
+    #[cfg(test)]
     pub(crate) fn on_decode_admit(
         &mut self,
         uuid: Uuid,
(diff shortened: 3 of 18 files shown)


lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -329,10 +337,18 @@ impl PendingLivePass {
...
+        if removed_output_tokens > 0 {
+            self.pass.accept_length_output_tokens = self
+                .pass
+                .accept_length_output_tokens
+                .checked_sub(removed_output_tokens)
+                .expect("suppressed more output tokens than the pass recorded");
+            self.pass.accept_length_decode_forwards = self
+                .pass
+                .accept_length_decode_forwards
+                .checked_sub(1)
```
jthomson04: **1. Panic introduced into the live cancellation path**

`suppress_request_outputs` replaced a self-healing full recompute of the accept-length counters with `checked_sub(...).expect(...)`, introducing two hard panics into the live scheduler's mid-pass cancellation path.

**Failure scenario:** Previously `accept_length_sample(&output_signals)` recomputed both counters from the surviving signals, so any drift healed itself. Now, if `accept_length_output_tokens`/`accept_length_decode_forwards` ever disagree with the surviving signals, the live pass boundary aborts the process. Two concrete ways this can arise: (a) `checked_sub(1)` hard-codes 'one decode forward per request', which only holds because a uuid appears at most once in `running` per pass — if a uuid ever appears twice (e.g. a readmitted/duplicated entry that `compact_running` did not collapse), the pass records 2 forwards for it, and suppressing that one request under-subtracts, then suppressing a second request drives the counter to 0 and a third suppression panics with `"suppressed an unrecorded decode forward"`; (b) a token-bearing signal appended outside the decode accounting (see `rejected: true, token_id: Some(9)` in the `accept_length_sample` unit test at `lib/mocker/src/scheduler/mod.rs:729`) is counted by `removed_output_tokens` but never by `AcceptLengthSample::record_forward`, so `accept_length_output_tokens.checked_sub(removed_output_tokens)` underflows and panics with `"suppressed more output tokens than the pass recorded"` in a production serving loop.

lib/mocker/src/replay/collector.rs
```diff
@@ -1127,35 +1127,18 @@ impl TraceCollector {
...
+    /// Record tokens emitted by one scheduler pass at its externally visible
+    /// completion boundary.
+    pub(crate) fn on_output_signals(
         &mut self,
         output_signals: &[OutputSignal],
         completion_time_ms: f64,
     ) {
-        let mut emitted_by_request = FxHashMap::default();
         for signal in output_signals {
             if signal.token_id.is_some() {
```
jthomson04: **2. `on_output_signals` drops the `!rejected` filter**

`on_output_signals` filters only on `token_id.is_some()` and drops the `!signal.rejected` condition that `accept_length_sample` applies, so a rejected signal carrying a token would be recorded as a delivered token.

**Failure scenario:** This function is now the single entry point for tokens into the trace collector (the schedulers no longer call `on_token` themselves). `accept_length_sample` (scheduler/mod.rs:129) filters `!signal.rejected && signal.token_id.is_some()`, and its own unit test at scheduler/mod.rs:729 constructs an `OutputSignal { token_id: Some(9), rejected: true }` and asserts it is excluded. If any path emits such a signal (today `vllm/core.rs:1625` uses `token_id: None`, so this is latent), the collector pushes a token timestamp for a request that produced no output: its `token_timeline` gains an extra entry, inflating output-token count, shrinking mean ITL/TPOT, and — if it is the first entry — producing a bogus `first_token_ms`/TTFT for a request that was rejected.

lib/mocker/src/cache/radix_cache.rs
```diff
@@ -132,6 +136,7 @@ impl PagePool {
     }
 
     pub fn free_pages(&mut self, pages: &[KvPageId]) {
+        self.free.reserve_exact(pages.len());
```
jthomson04: **3. `reserve_exact` on the free list defeats amortized growth (O(N^2) memcpy)**

`self.free.reserve_exact(pages.len())` before `extend_from_slice` defeats `Vec`'s amortized growth, turning free-list growth from O(N) into O(N²) copying — a perf regression inside a perf PR.

**Failure scenario:** `reserve_exact` sizes the allocation to exactly `len + additional` (it explicitly does not over-allocate), and `extend_from_slice` already calls the amortized `reserve` internally, so this line is both redundant and harmful. Every `free_pages` call that pushes the free list past its previous high-water mark now reallocates and memcpys the whole list instead of hitting a doubled buffer. Draining a KV pool of N pages back to the free list in chunks of k pages costs ~N/k reallocations copying ~N²/(2k) elements: for a 12-worker SGLang replay with 8M tokens at 64-token pages (N ≈ 125k pages) freed in ~20-page chunks, that is ~4e8 8-byte element copies (~3 GB of memcpy) versus O(N) before. If pre-sizing is desired, size `free` once to `total_pages` in `PagePool::new` and drop this line.

lib/mocker/src/kv_manager/sglang_backend.rs
```diff
@@ -897,19 +887,21 @@ impl SglangKvManager {
         }
 
         let mut block_hashes = Vec::new();
+        block_hashes.reserve_exact(evicted_pages.len());
```
jthomson04: **4. Unconditional allocation on the `publish_removed_pages` no-op path**

`publish_removed_pages` now eagerly allocates `block_hashes` for every evicted page even though the function returns early whenever nothing is actually removed, adding an allocation to the common no-op case.

**Failure scenario:** The base code used a lazily-growing `Vec::new()` that never allocated when no block hash was published. Now `Vec::new()` immediately followed by `reserve_exact(evicted_pages.len())` allocates unconditionally, then the loop `continue`s for every page whose `page_to_block_hash` slot is `None` and for every hash whose refcount is still > 1, and the function hits `if block_hashes.is_empty() { return; }` — one wasted malloc/free per `evict()`/`free_pages()` call. In steady-state replay where most evicted pages are duplicates or unpublished, this is a per-eviction allocation on a hot path the PR is trying to de-allocate. Use `Vec::with_capacity` only after the first push, or leave it as `Vec::new()`.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -601,14 +601,14 @@ where
...
-        align_collector: Option<&mut TraceCollector>,
+        collector: Option<&mut TraceCollector>,
         effects: &mut EngineEffects<Observation::Batch>,
     ) {
         if let Some(fpm) = executed.fpm.as_mut() {
             fpm.wall_time_secs = boundary.wall_time_secs();
         }
-        if let Some(collector) = align_collector {
-            collector.align_pass_token_times(&executed.output_signals, boundary.end_ms);
+        if let Some(collector) = collector {
```
jthomson04: **5. Admission and output recording are now split across consumers**

Collector recording is now split: `lower_executed_pass` records output tokens but not admissions, leaving `on_admit` to be re-implemented independently in `agg.rs` and `disagg.rs`, so any new `EngineEffects` consumer silently loses admission timestamps.

**Failure scenario:** Before this PR the engine core recorded both `on_admit` and `on_token`, so both halves were guaranteed together. Now `lower_executed_pass` calls only `collector.on_output_signals(...)`, and `on_admit` lives in `agg.rs:774` (added by this PR) and `disagg.rs:2091/2101` (pre-existing). A third replay topology, or any future path that consumes `EngineEffects` without duplicating the admission loop, gets token timelines but never sets `stats.first_admit_ms`/`first_admission_reused_input_tokens` — silently emitting `None` admission times and zero reused-input-token accounting in the replay report with no compile-time or debug-assert signal. A single `record_pass(collector, &pass, start_ms, end_ms)` helper (the `record_test_pass` at scheduler/mod.rs:196 is exactly this) used by all consumers would keep the two halves inseparable.

lib/mocker/src/replay/offline/components/worker_core.rs
```diff
@@ -70,6 +70,11 @@ impl ReplayWorkerCore {
         collector: &mut TraceCollector,
         now_ms: f64,
     ) -> anyhow::Result<EnginePassResult> {
-        self.core.try_execute_pass(collector, now_ms)
+        let pass = self.core.try_execute_pass(now_ms)?;
```
jthomson04, line 77: **6. `execute_pass` duplicates the `#[cfg(test)] record_test_pass` helper**

`ReplayWorkerCore::execute_pass` is a verbatim copy of the new `#[cfg(test)] record_test_pass` helper in `scheduler/mod.rs`.

**Failure scenario:** Both bodies are `for admission in &pass.admissions { collector.on_admit(admission.uuid, now_ms, admission.reused_input_tokens); }` followed by `collector.on_output_signals(&pass.output_signals, pass.end_ms)`. Two copies of the same recording contract means a change to what the collector observes per pass (e.g. adding the `!rejected` filter from finding #2, or recording lifecycle events) must be made in two places, and the test-only copy is the one that validates the production copy — so a divergence between them is invisible to the test suite. Drop the `#[cfg(test)]` from `record_test_pass` (or move it next to `EnginePassResult`) and have `ReplayWorkerCore::execute_pass` call it.

lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -1145,23 +1146,22 @@ impl VllmCore {
...
-        &mut self,
-        collector: &mut TraceCollector,
-        now_ms: f64,
-    ) -> anyhow::Result<EnginePassResult> {
-        self.execute_pass_internal(Some(collector), now_ms, None)
+    pub(crate) fn try_execute_pass(&mut self, now_ms: f64) -> anyhow::Result<EnginePassResult> {
+        self.execute_pass_internal(now_ms, None)
     }
 
     pub(crate) fn try_execute_hidden_pass(
```
jthomson04: **7. `try_execute_hidden_pass` is now identical to `try_execute_pass` on both cores**

`try_execute_pass` and `try_execute_hidden_pass` are now byte-identical on both cores, leaving a dead `EnginePassMode` dispatch through six redundant wrapper functions.

**Failure scenario:** `VllmCore::try_execute_pass` and `try_execute_hidden_pass` both reduce to `self.execute_pass_internal(now_ms, None)`; `SglangCore`'s pair (core.rs:614/624) both reduce to `self.try_execute_pass_internal(now_ms)`. `EngineCore` (scheduler/mod.rs:322/330) and `OfflineWorkerState` (state.rs:434/439) each forward both. The `match self.pass_mode { Visible => ..try_execute_pass(now_ms), Hidden => ..try_execute_hidden_pass(now_ms) }` in `drive_ready` (engine.rs:722-731 and 776-788, duplicated twice) therefore calls the same function on both arms. The 'hidden' guarantee is no longer enforced at the core at all — it is now only the absence of a collector at `lower_executed_pass`, which is a weaker, easier-to-break invariant. Delete the hidden variants and keep a single `try_execute_pass(now_ms)`, with the mode check confined to the collector site.

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -720,11 +720,11 @@ where
                     dp_rank,
                     || match self.pass_mode {
                         EnginePassMode::Visible => {
```
jthomson04: **8. Discarded-binding presence check plus `.expect()`, duplicated 4x**

`let Some(_) = collector.as_deref_mut() else { bail!(...) };` is a discarded-binding presence check that is then re-resolved below with `.expect("visible pass collector checked before execution")`, duplicated verbatim in both the single-rank and multi-rank branches.

**Failure scenario:** The pattern costs two `Option` resolutions plus a panic-carrying `expect` per pass, and encodes the invariant 'we already checked' in a comment string instead of the type system. It is written out four times (lines 722-726, 736-743, 776-786, 834-841). Hoisting `let Some(collector) = collector.as_deref_mut() else { bail!("offline replay visible engine pass requires a collector") };` once outside the `match` (or replacing the guard with `if self.pass_mode == Visible && collector.is_none() { bail!(...) }` before the loop) removes the `expect` entirely and halves the duplication.

lib/mocker/src/kv_manager/sglang_backend.rs
```diff
@@ -184,7 +184,7 @@ impl SglangKvManager {
             kv_event_publishers,
             dp_rank,
             next_event_id: 0,
-            page_to_block_hash: FxHashMap::default(),
+            page_to_block_hash: vec![None; total_tokens / page_size],
```
jthomson04: **9. `page_to_block_hash` allocated even when KV events are disabled**

`page_to_block_hash: vec![None; total_tokens / page_size]` is allocated unconditionally in the constructor, including for managers where KV events are disabled and the vector is never read or written.

**Failure scenario:** Both `publish_stored_hashes` (line 793) and `publish_removed_pages` (line 885) return immediately when `self.kv_event_publishers.is_empty()`, so with KV events off the vector is dead. The previous `FxHashMap::default()` cost nothing in that case; the dense vector now costs 16 bytes per page up front and permanently. For a 12-worker replay at 8M tokens / 64-token pages that is ~2 MB per worker (~24 MB resident) of never-touched memory, plus the zeroing cost at startup. Gate the allocation on `!kv_event_publishers.is_empty()` (allocate lazily or store an empty vec).

lib/mocker/src/kv_manager/sglang_backend.rs
```diff
@@ -808,70 +808,60 @@ impl SglangKvManager {
...
+        let Some(first_unpublished_page) = (first_page..complete_pages)
+            .find(|&page_idx| self.page_to_block_hash[pages[page_idx].index()].is_none())
         else {
             return 0;
         };
 
-        let mut computed_blocks = Vec::new();
-        let first_unpublished_page = first_unpublished_block / block_size;
         let local_hashes = &page_hashes[first_unpublished_page..complete_pages];
+        self.block_hash_refcounts.reserve(local_hashes.len());
```
jthomson04: **10. `reserve` over-reserves for pages the loop skips**

`self.block_hash_refcounts.reserve(local_hashes.len())` reserves capacity for every page in the range before the loop, but the loop skips (`continue`s on) every page that already has a published hash — the common case under prefix reuse.

**Failure scenario:** With ~92% prefix-cache reuse (the PR's own reported figure), most of `local_hashes` hits `if self.page_to_block_hash[page_slot].is_some() { continue; }` and inserts nothing. The map is nonetheless grown to `len + local_hashes.len()`, which can trigger a full rehash of the entire refcount map for entries that are never added, on every `publish_stored_hashes` call. The same over-reservation applies to `blocks.reserve_exact(local_hashes.len() - block_idx)` at line 850. Either reserve nothing (the entry API grows amortized anyway) or reserve after counting the actually-unpublished pages, which the `first_unpublished_page` scan is already walking.

lib/mocker/src/replay/collector.rs
```diff
@@ -1127,35 +1127,18 @@ impl TraceCollector {
...
-                *emitted_by_request.entry(signal.uuid).or_insert(0usize) += 1;
+                self.on_token(signal.uuid, completion_time_ms);
```
jthomson04: **11. Per-signal hash lookup where an MTP burst shares one uuid**

`on_output_signals` performs a fresh `self.requests` hash lookup per token signal via `on_token`, where consecutive signals from an MTP/spec-decode burst share a uuid and could reuse one lookup.

**Failure scenario:** Both emitters push a request's burst contiguously (`sglang/decode.rs:359` and `vllm/core.rs:2600`), so for an accept length of 4 the collector does 4 `FxHashMap<Uuid, TraceRequestStats>` lookups where 1 lookup plus 4 `Vec::push` calls would do. Given this PR's stated goal of removing hash work from the replay hot path, and that the removed `align_pass_token_times` did exactly one `requests` lookup per distinct uuid, tracking the previous uuid (or resolving `stats` once per run of equal uuids) recovers the per-request-lookup behavior for the spec-decode case.

lib/mocker/src/replay/offline/components/worker_core.rs
```diff
@@ -70,6 +70,11 @@ impl ReplayWorkerCore {
...
+        for admission in &pass.admissions {
+            collector.on_admit(admission.uuid, now_ms, admission.reused_input_tokens);
+        }
+        collector.on_output_signals(&pass.output_signals, pass.end_ms);
```
jthomson04: **12. Token stamps can be pushed forward to a KVBM offload deadline**

The single-worker path now stamps tokens with `pass.end_ms`, which the `kvbm-offload` stall-advance can push forward to an offload deadline, whereas the previous stamp was the decode completion time.

**Failure scenario:** `VllmCore::execute_pass_internal` contains `if end_ms <= now_ms && let Some(deadline) = self.kv_manager.earliest_offload_deadline() { end_ms = deadline.max(now_ms); }` (core.rs:1651-1656). Previously `emit_ready_tokens` recorded `on_token(uuid, decode_start_ms + decode_time)`, which is unaffected by that bump. Now `ReplayWorkerCore::execute_pass` passes `pass.end_ms`. A prefill-type worker (`decode_time == Duration::ZERO`) that schedules no new prefill in a pass (`batch_count == 0`, so `prefill_time == 0`) can satisfy `end_ms <= now_ms` while still emitting token-bearing signals from `ready`; those tokens are then timestamped with the KVBM offload deadline rather than the decode boundary, inflating TTFT/ITL for that request in the single-worker KVBM replay reports.

lib/mocker/src/scheduler/sglang/decode.rs
```diff
@@ -365,6 +371,7 @@ pub(super) fn simulate_decode_step_with_sampler(
             cache_materialized_prefix(req, kv_manager, config);
             req.debug_assert_invariants(config.block_size);
         }
+        accept_length.record_forward(output_signals.len() - output_signals_before);
```
jthomson04: **13. Forward count inferred by differencing the shared signal vector**

`record_forward(output_signals.len() - output_signals_before)` infers the visible-token count by differencing the shared output vector, silently coupling the counter to the assumption that every signal pushed inside the loop is a non-rejected token-bearing signal.

**Failure scenario:** The very same function already pushes `token_id: None` signals into `output_signals` in the pre-loop `already_completed_indices` block (decode.rs:234-250), so the invariant this counter relies on is violated only a few lines above. If a future change pushes a terminal or rejected signal inside the per-request loop (or inside `vllm/core.rs:2616`'s equivalent), `AcceptLengthSample` over-counts both `output_tokens` and `decode_forwards`, and the only guard is a `debug_assert_eq!` that is compiled out of the release replay builds where these counters actually feed `on_accept_length_sample`. Incrementing an explicit local counter next to each `token_id: Some(..)` push would make the counter independent of the vector's other contents.

jthomson04 (commented):
Review of the three refactors in this PR. I traced each for behavioral equivalence first, and the core claims hold:

- The `AcceptLengthSample` counters match what `accept_length_sample` computes on every emit path (already-complete and rejected signals carry `token_id: None`, and each uuid appears at most once per pass).
- `on_output_signals(boundary.end_ms)` reproduces the old `on_token` + `align_pass_token_times` pair exactly — `on_token` is a pure push, and the call site in `lower_executed_pass` sits where `align` did.
- Admissions are correctly re-homed to `agg.rs` / `disagg.rs` / `worker_core.rs`, consistent with the admissions-at-epoch-start / outputs-at-completion-boundary contract in `lib/mocker/src/replay/offline/CLAUDE.md`.
- The dense `Vec<Option<..>>` index cannot go out of bounds: `PagePool::allocate_pages` keeps `next_fresh <= total_pages` even under double-free.
- The `#[cfg(debug_assertions)]` attributes on the `debug_assert_eq!` statements are required rather than redundant, since `debug_assert_eq!` still type-checks its arguments in release.

13 inline comments follow, ordered most severe first. The first three are the ones I would want resolved before merge:

1. `suppress_request_outputs` trades a self-healing recompute for `checked_sub().expect()`, converting counter drift into a process abort in the live serving loop.
2. `on_output_signals` is now the sole token entry point into the collector but drops the `!rejected` filter its predecessor applied.
3. `free.reserve_exact()` in `PagePool::free_pages` makes free-list growth quadratic — a perf regression inside a perf PR.

The rest are efficiency, duplication, and robustness items.

lib/mocker/src/replay/offline/components/worker_core.rs
```diff
@@ -70,6 +70,11 @@ impl ReplayWorkerCore {
...
+        Ok(pass)
```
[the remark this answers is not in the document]
jthomson04: Resolved in f2ce225849. EnginePassResult now carries a rank-local token_completion_ms captured before KVBM stall advance; single-worker replay records tokens at that boundary, while aggregate and attention-DP paths retain their shared boundary. Added focused coverage and the full KVBM parity rows passed byte-for-byte.

lib/mocker/src/replay/offline/components/worker_core.rs
jthomson04: Resolved in f2ce225849. vLLM captures token_completion_ms before KVBM stall advance and single-worker replay uses it; aggregate and attention-DP timestamps remain at their shared completion boundary.

lib/mocker/src/cache/radix_cache.rs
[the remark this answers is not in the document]
jthomson04: Resolved in f2ce225849. Removed the per-append reserve_exact; extend_from_slice now uses Vec's amortized growth.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Resolved in f2ce225849. Visible-collector validity is checked once at drive_ready entry; the repeated discarded Option checks and expect calls are gone from both singleton and grouped paths.

lib/mocker/src/replay/collector.rs
```diff
@@ -1127,35 +1127,18 @@ impl TraceCollector {
...
-            let TokenTimeline::Recording(times) = &mut stats.token_timeline else {
-                continue;
-            };
-            let start = times
-                .len()
-                .checked_sub(emitted)
-                .expect("scheduler emitted more output signals than collector tokens");
-            times[start..].fill(completion_time_ms);
-        }
     }
```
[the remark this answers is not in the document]
jthomson04: Resolved in f2ce225849. Collector token recording now excludes rejected signals even when they carry a token ID. The focused grouped-output test covers the rejected-token case.

lib/mocker/src/cache/radix_cache.rs
jthomson04: Resolved in f2ce225849. Removed reserve_exact from PagePool::free_pages so the free list grows amortized.

lib/mocker/src/kv_manager/sglang_backend.rs
jthomson04: Resolved in f2ce225849. Removed-event storage remains allocation-free until the first final-copy hash is actually removed; capacity is reserved lazily at that point.

lib/mocker/src/replay/collector.rs
jthomson04: Resolved in f2ce225849. Collector timelines filter on !rejected && token_id.is_some(); focused coverage includes a rejected token-bearing signal.

lib/mocker/src/replay/offline/components/worker_core.rs
jthomson04: Resolved in f2ce225849. Removed the duplicate test helper; production and tests now share TraceCollector::on_scheduler_pass.

lib/mocker/src/kv_manager/sglang_backend.rs
jthomson04: Resolved in f2ce225849. Dense page metadata is allocated only when KV events are enabled, with focused disabled/enabled metadata coverage.

lib/mocker/src/scheduler/live_boundary.rs
jthomson04: Resolved in f2ce225849. Live suppression now retains surviving signals and recomputes both accept-length counters from them, restoring the self-healing behavior and removing the panic path. Added stale-counter cancellation coverage.

lib/mocker/src/scheduler/vllm/core.rs
jthomson04: Resolved in f2ce225849. Removed the redundant hidden-pass APIs from both scheduler cores and wrappers. Visible and hidden modes execute the same raw pass; EnginePassMode now controls collector visibility only.

lib/mocker/src/scheduler/sglang/decode.rs
jthomson04: Resolved in f2ce225849. SGLang and speculative-vLLM decode now maintain explicit emitted-token counters beside successful token emission rather than differencing the shared output vector.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Resolved in f2ce225849. TraceCollector::on_scheduler_pass now records generic admissions and accepted outputs together across EngineComponent, single-worker replay, and scheduler tests. Pool/lifecycle bookkeeping remains topology-specific.

lib/mocker/src/replay/collector.rs
jthomson04: Resolved in f2ce225849. Actual speculative bursts are recorded per consecutive UUID run with one request-map lookup and a bulk append. Ordinary one-token passes retain the direct fast path using the pass's explicit accept-length counters.

lib/mocker/src/kv_manager/sglang_backend.rs
jthomson04: Resolved in f2ce225849. The publication path no longer reserves from the full candidate range; refcount and Stored vectors now use amortized growth, avoiding both over-reservation and an added suffix-count scan.

jthomson04:
/ok to test [3e3efad](https://github.com/ai-dynamo/dynamo/commit/3e3efad9a4652040b09fe7c8599e25f73e530489)

lib/mocker/src/replay/offline/components/engine.rs
```diff
@@ -675,6 +680,9 @@ where
         now_ms: f64,
         mut collector: Option<&mut TraceCollector>,
     ) -> anyhow::Result<EngineEffects<Observation::Batch>> {
+        if self.pass_mode == EnginePassMode::Visible && collector.is_none() {
```
PeaBrane: Non-blocking AI review note: `TraceCollector::on_scheduler_pass` now documents that hidden passes still record admissions while omitting token timestamps, but `drive_ready` accepts `None` for hidden mode and `lower_executed_pass` then skips both. Every current caller passes `Some`, so this is not a current bug; consider requiring `&mut TraceCollector` here (with `tokens_visible` controlling only token recording) to prevent a future hidden topology from silently losing first-admission and reused-input accounting.

lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -1646,8 +1623,14 @@ impl VllmCore {
                 handoff_delay_ms: None,
             });
         }
+        #[cfg(debug_assertions)]
+        debug_assert_eq!(
+            (accept_length.output_tokens, accept_length.decode_forwards),
+            crate::scheduler::accept_length_sample(&output_signals)
+        );
+        let token_completion_ms = decode_start_ms + decode_time.as_secs_f64() * 1000.0;
```
PeaBrane, line 1631: Non-blocking AI review note: the new `worker_core` test fabricates an `EnginePassResult` with `token_completion_ms < end_ms`, so it verifies the consumer but not this producer boundary. Consider strengthening a feature-gated KVBM stall test to assert that `token_completion_ms` is captured before stall advancement changes `end_ms`; that would protect the regression this field was introduced to fix.

PeaBrane (approved):
✅ AI-assisted review: no blocking issues found. Left two non-blocking hardening suggestions.

lib/mocker/src/replay/offline/components/engine.rs
jthomson04: Addressed in a59b6436bc. EngineComponent::drive_ready and lower_executed_pass now require a TraceCollector, so hidden passes cannot silently skip admission accounting. EnginePassMode still controls only token visibility. The existing hidden-prefill reused-token test and the 190-test offline replay suite pass.

lib/mocker/src/scheduler/vllm/core.rs
jthomson04: Addressed in a59b6436bc by strengthening the existing feature-gated cold_request_with_g2_prefix_swaps_in_then_admits test. It now verifies that token_completion_ms remains at the original pass boundary while end_ms advances to the KVBM offload deadline. The focused kvbm-offload test passes.

jthomson04:
/ok to test [a59b643](https://github.com/ai-dynamo/dynamo/pull/12772/commits/a59b6436bc7d4941a519f3e9c18e4d7cf23addaf)

PR #12788, opened 2026-08-06
feat(omni): add Nemotron Audex speech synthesis to /v1/audio/speech

tthakkal:
## Overview:

Adds Nemotron Audex text-to-speech to the vLLM-Omni backend on the existing `/v1/audio/speech` endpoint. Audex is a two-stage pipeline (thinker emits `<speechcodec_N>` tokens, code2wav decodes them to 16 kHz mono), so it needs its own prompt format, request contract, and output assembly. No new endpoint, worker, or launch script, and no change for existing audio models. Supports `nvidia/Nemotron-Labs-Audex-2B` and `nvidia/Nemotron-Labs-Audex-30B-A3B`.

```bash
./launch/agg_omni_audio.sh --model nvidia/Nemotron-Labs-Audex-2B

curl -X POST http://localhost:8000/v1/audio/speech \
  -H 'Content-Type: application/json' \
  -d '{"model":"nvidia/Nemotron-Labs-Audex-2B","input":"Hey, this is generated using Dynamo!","cfg_scale":1.5}' \
  -o dynamo-audio.wav
```

## Details:

- **Request handling** (`omni/audio_handler.py`): detects Audex from the engine's stage set, not the model name; primes the thinker prompt with `<think></think><speechgen_start>` (a plain prompt yields zero codec tokens); adds classifier-free guidance via `cfg_role`/`cfg_pair_id`/`cfg_null_prompt` when `cfg_scale > 1.0`; rejects `voice`/`ref_audio`/`ref_text` rather than silently ignoring what Audex cannot honor.
- **Output assembly** (`output_formatter.py`, `omni_handler.py`): the chunk-streaming decoder emits _cumulative_ snapshots, some empty. `_generate_openai_mode` creates one `AudioAggregateState` per request and threads it through the formatter's ctx; for Audex (`AudioHandler.emits_cumulative_waveforms()`) `_append_audio_chunk` keeps the longest snapshot and discards the rest, and a single chunk is emitted at end of stream. Previously Audex produced a silent file, or several times the correct duration if concatenated. Keeping the longest rather than the latest means a truncated trailing payload cannot drop already-decoded audio. Models that emit incremental frames (Qwen3-TTS, MiMo-Audio) stay on the existing append-then-concatenate path — de-duplicating those would silently drop audio, so the new behavior is scoped to the pipeline known to need it. The buffer is a local the handler owns, so there is no shared registry to release and no cross-request state to leak.
- **`cfg_scale` plumbing** (`audios.rs`, `audio_protocol.py`): added to both request models. Unknown fields are dropped silently, so without this the client's `cfg_scale` would vanish between frontend and worker — no error, just different audio.
- **Launch script** (`agg_omni_audio.sh`): Audex reuses the audio script (`--model`) since the worker invocation is identical. `MODEL` is env-overridable and the printed curl footer adapts per model — the Qwen3-TTS body includes `voice`, which Audex rejects with 400. `--stage-configs-path` rides the existing arg passthrough, needed for the 30B since both checkpoints report `model_type: nemotron_labs_audex`.
- **Docs**: Audex added to `text-to-audio.md` (models, launch, request tab, `cfg_scale`) and the Audio Generation accordion in `cli-templates/vllm.mdx`.
*   **Tests**: new `test_audio_handler.py` (pipeline detection, prompt priming, CFG contract, rejections); `test_output_formatter.py` / `test_omni_handler.py` cover cumulative snapshot assembly against `AudioAggregateState`; `omni_audex` serve deployment in `tests/serve/test_vllm_omni.py`, which ships `@pytest.mark.skip` (117.6 GiB on 1× H200, above CI's 24 GiB). `AudioSpeechPayload` gains optional duration/RMS/sample-rate checks, since `len(bytes) > 100` passes on a silent WAV — the exact failure a mis-assembled chunk stream produces — and new `tests/utils/test_audio_speech_payload.py` gives those checks live pre-merge callers on CPU.

### Validation
Run on Intel XPU and NVIDIA CUDA (1× H200, vLLM 0.27.1). Numbers are from the CUDA run unless noted.
| Command | Result |
|---|---|
| `pytest -m unit components/src/dynamo/vllm/tests/ tests/utils/test_audio_speech_payload.py` | 1249 passed, 6 skipped, 26 deselected |
| `pytest components/src/dynamo/vllm/tests/omni/` | 288 passed (178 in the three audio files this PR touches) |
| `pytest components/src/dynamo/common/tests/test_audio_protocol.py` | 4 passed |
| `pytest tests/utils/test_audio_speech_payload.py` | 15 passed (new: gives `AudioSpeechPayload._validate_waveform` live pre-merge callers) |
| `cargo test -p dynamo-llm --lib protocols::openai::audios` | 13 passed, incl. the 2 new `cfg_scale` tests |

Three configurations were driven end-to-end through the launch script and `/v1/audio/speech`. Waveforms are parsed, not size-checked, so a silent or duplicated file would have failed; guided and unguided differing in every row is what proves `cfg_scale` survives the frontend→worker hop and reaches the logits processor.

| Config | Stage configs | Ready | Guided (`cfg_scale=1.5`) | Unguided | `voice` |
|---|---|---|---|---|---|
| Audex-2B TTS | default | 105 s | 200, 16 kHz mono, 2.10 s, rms 0.064 | 2.40 s, rms 0.062, differs | 400 |
| Audex-2B TTA | `audex_tta.yaml` | 120 s | 200, 16 kHz mono, 10.00 s, rms 0.0076 | 10.00 s, rms 0.0025, differs | 400 |
| Audex-30B-A3B TTS | `audex_tts_30b.yaml` | 440 s | 200, 16 kHz mono, 2.26 s, rms 0.065 | 2.16 s, rms 0.068, differs | 400 |

TTA and 30B need no code or script changes, only `--stage-configs-path`. TTA also covers the sync full-payload path (`async_chunk: false`), not just the chunked TTS one, and on XPU omitting `cfg_scale` was byte-identical to passing 3.0 and differed from 1.0, confirming the TTA default.
### Notes

- `--enforce-eager` in `agg_omni_audio.sh` is pre-existing (all three audio launch scripts hardcode it) and left alone here. It costs TTS RTF, so making it conditional is a possible follow-up.
- Guided temperature is pinned to 0.05, the 2B-tuned value. vLLM-Omni's `audex_tts_30b.yaml` prefers 0.1 for the 30B and treats size-aware tuning as a known follow-up; this matches upstream.
*   TTA works today via `--stage-configs-path` (verified above), but this PR adds no dedicated TTA page or serve test — both follow-ups. Open question: every TTA request returned exactly 10.0 s, the decode cap, on both platforms, so generation may not be stopping early on `<audiogen_end>`.
- Audio streaming remains unsupported for TTS: the client gets one complete WAV, even though the decoder streams internally.

1. `omni/audio_handler.py` — the CFG pairing block: a mispaired `cfg_pair_id` degrades audio without raising.
2. `omni/output_formatter.py` + `omni_handler.py` — the per-request `AudioAggregateState` and the keep-longest de-duplication in `_append_audio_chunk`. The cumulative/incremental gate is the risky part: the wrong branch either multiplies the duration or silently drops audio.
3. `lib/llm/src/protocols/openai/audios.rs` + `common/protocols/audio_protocol.py` — small, but the two must agree.
4.  `tests/utils/payloads.py` — the new waveform checks must be inert for existing callers; `test_all_checks_disabled_skips_waveform_decoding` and `test_url_response_skips_the_waveform_checks` cover that.

## Related Issues

## Summary by CodeRabbit

- **New Features**
  - Added support for Nemotron Audex 2B and 30B-A3B speech generation.
  - Added optional `cfg_scale` guidance control for compatible audio requests.
  - Added Audex-specific voice, reference-audio, token-limit, and request validation.
  - Improved streamed audio handling by delivering finalized cumulative audio responses.

- **Documentation**
  - Added Audex setup instructions, model configuration guidance, request examples, and supported-parameter details.

- **Tests**
  - Added coverage for Audex generation, validation, audio quality, streaming behavior, and waveform properties.


dagil-nvidia (commented):
Review of head 134c22b: 1 must-fix, 4 consider. All five repo gates pass locally, the production code is sound end to end, and all earlier bot threads are addressed at head. The must-fix is a casualty of the Aug 25 merge from main.

## Must-fix

1. `components/src/dynamo/vllm/tests/omni/test_output_formatter.py:739` and `test_omni_handler.py:347` - the new tests target an API that no longer exists at head. `TestAudioFormatterObserveChunk` (11 tests) calls `AudioFormatter.observe_chunk` / `take_pending` / `discard_pending` / `format_audio`, and one handler test asserts `output_formatter.audio._pending == {}` - none of these exist after the merge reconciled the formatter with main's streaming rework (production moved to `AudioAggregateState.cumulative` plus keep-longest aggregation, which is sound and was verified end to end). The tests carry the same `pre_merge`/`unit`/`vllm`/`gpu_0` marks as neighbors that run in the vllm unit job, so CI goes red with AttributeError once it runs. Rewrite the class against the current design (drive `AudioFormatter.format()` with cumulative state, then `finish_aggregate`) and drop or rewrite the `_pending` assertion - per-request buffering makes that leak class structurally gone. The PR description's buffering explanation and "249 passed" figure describe the pre-merge design and are stale.

## Consider

2. `docs/fern/pages/use-cases/diffusion/workflows/text-to-audio.md:141` - the `cfg_scale` field says omitting it decodes unguided, but text-to-audio deployments default it to 3.0; add the split (unguided for speech, 3.0 for TTA).
3. Same page, line 144 - the accepted-fields list omits `speed`, which the Audex path honors; as written a reader concludes `speed` is a no-op.
4. Two `python -c` snippets should be `python3` per docs convention.
5. The new `omni_audex` serve test ships permanently skip-marked (117.6 GiB rationale), which also leaves the new `_validate_waveform` checks with zero live CI callers; worth a tracked follow-up for the per-stage memory plumbing named in the skip reason.

Verified sound: cumulative buffering correctly scoped to Audex via `emits_cumulative_waveforms()`; guided-without-defaults fails loud; exactly one response at end of stream; sampling defaults deep-copied with a shared-mutation regression test; all flag, env, protocol-parity, and marker claims fact-checked, including Rust/Python round-trip serde.

Comment-only review; severity calls stay with the human reviewers.

rmccorm4:
/ok to test [c056aa5](https://github.com/ai-dynamo/dynamo/pull/12788/commits/c056aa575b6eacd50b14138417907c09da158205)

tthakkal:
@dagil-nvidia — thanks for the careful review. 
@rmccorm4  — thanks for triggering the tests.

CPU unit-test failures are fixed; ready for another CI run.

**Must-fix:** the two tests targeted the pre-Aug-25 `observe_chunk` / `take_pending` API (11 failed / 1224 passed). Rewritten against `AudioAggregateState` + keep-longest as `TestAudioFormatterCumulativeAggregate`, driving the real `format()` / `finish_aggregate()` surface. Now **1249 passed, 6 skipped, 26 deselected** (skips/deselects pre-existing). 

**2/3:** `cfg_scale` documents the Audex TTA `3.0` fallback; `speed` added to the accepted-fields list. 
**4:** both snippets are now `python3`. 
**5:** `_validate_waveform` now has live pre-merge callers in `tests/utils/test_audio_speech_payload.py` (15 tests, CPU) — the serve test stays skip-marked and I'll file the per-stage-memory follow-up. PR description updated too.

components/src/dynamo/vllm/omni/output_formatter.py
```diff
@@ -420,7 +435,15 @@ def _append_audio_chunk(
...
         )
         self._validate_audio_metadata(state, sample_rate, num_channels, channel_axis)
-        state.chunks.append(audio_np)
+        if not state.cumulative:
+            state.chunks.append(audio_np)
+            return
+
+        # Snapshots only grow, but keep the longest rather than the latest so a
+        # truncated trailing payload cannot drop already-decoded audio.
+        if state.chunks and state.chunks[0].shape[-1] >= audio_np.shape[-1]:
```
ptarasiewiczNV, line 447: [`_generate_openai_mode()`](https://github.com/ai-dynamo/dynamo/blob/224019034654b74f18e6a2381da67afb0d6ca13b/components/src/dynamo/vllm/omni/omni_handler.py#L404) calls [`streaming_sampling_params()`](https://github.com/ai-dynamo/dynamo/blob/224019034654b74f18e6a2381da67afb0d6ca13b/components/src/dynamo/vllm/omni/utils.py#L45) for Audex too. In the [pinned vLLM-Omni `v0.28.0rc1`](https://github.com/ai-dynamo/dynamo/blob/224019034654b74f18e6a2381da67afb0d6ca13b/container/context.yaml#L104), [`coerce_param_message_types()` / `maybe_coerce_to_message_type()`](https://github.com/vllm-project/vllm-omni/blob/v0.28.0rc1/vllm_omni/entrypoints/utils.py) force DELTA output, and the [output processor drains emitted audio](https://github.com/vllm-project/vllm-omni/blob/v0.28.0rc1/vllm_omni/engine/output_processor.py#L307).

Keeping only the longest chunk therefore discards audio. Could we make aggregation follow the requested output mode and add a handler-level regression test?

ptarasiewiczNV (commented):
The overall approach makes sense. One maintainability suggestion: please consider moving the Audex-specific request preparation out of `AudioGenerationHandler` into a dedicated helper/adapter.

Could be related, but not necessarily. Upstream’s `AudexAdapter` and `AudexTTAAdapter` already implement much of the prompt and CFG/RVQ setup. Could we reuse some of that without coupling Dynamo to their HTTP server? That would reduce duplicated model-specific logic and future drift.

components/src/dynamo/vllm/omni/output_formatter.py
tthakkal: @ptarasiewiczNV  Thank you, both addressed — the aggregation call was right, it was dropping audio.

**Correctness.** Confirmed on the pinned build: `maybe_coerce_to_message_type` forces `DELTA` on every `SamplingParams` when `is_streaming=True`, the processor drains the audio key after each snapshot, and `audex_code2wav` "always returns delta audio" (100 ms per emit) — so keep-the-longest kept one 100 ms delta. Aggregation now reads the *coerced* params via `utils.audio_output_is_cumulative(inputs.sampling_params_list)`. Diffusion-tailed audio is unaffected (`OmniDiffusionSamplingParams` has no `output_kind`).

With the pinned version that predicate always answers `False`, so the de-duplicating branch is unreachable today — kept, and documented, to follow the engine's contract rather than re-hardcode an assumption about the coercion.

Tests: `TestAggregatedAudioFollowsOutputKind` asserts three DELTA payloads assemble to 3600 samples (2400 cumulative), plus no cross-request leakage; restoring the old behaviour fails it with `assert 2400 == 3600`. Validated on Intel XPU too: `cfg_scale: 1.5` gives 2.26 s at rms 0.055, matching the CUDA baseline, versus ~100 ms before.

**Maintainability.** Moved to `omni/audex.py` as `AudexRequestAdapter` (`self.audex`); handler is a dispatcher again, −410 lines. Reusing `AudexAdapter` buys less than it looks: `build()` returns just a prompt, while the CFG/RVQ setup lives in `serving_speech.py`, so wiring it up needs an `OpenAICreateSpeechRequest` plus a `SpeechServingContext` back-referencing the running server — the coupling you wanted to avoid. What actually drifts with the checkpoint is model-owned code under `model_executor.models.audex`, which the adapter calls directly. Noted in the module docstring with a revisit-if-upstream-moves-them pointer.

lib/llm/src/protocols/openai/audios.rs
```diff
@@ -67,6 +67,12 @@ pub struct NvCreateAudioSpeechRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_new_tokens: Option<i32>,
 
+    /// Classifier-free guidance scale (Audex). Unset or 1.0 decodes unguided;
+    /// higher values follow the prompt more closely. Without this field serde
+    /// drops the client's cfg_scale and guidance is silently never applied.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cfg_scale: Option<f64>,
```
ptarasiewiczNV: Since this is not a part of official OAI API and is model specific I would move this under `nvext`

components/src/dynamo/vllm/omni/audio_handler.py
```diff
@@ -225,6 +220,40 @@ async def build_engine_inputs(self, req: NvCreateAudioSpeechRequest):
             stream_audio=stream_audio,
         )
 
+    # -- Audex ----------------------------------------------------------------
+
+    def _engine_inputs_audex(
```
ptarasiewiczNV: Could we simplify `_engine_inputs_audex`? It only wraps the adapter result and adds common delivery fields. Moving EngineInputs into a shared types module would avoid the circular import and let us annotate the return type. Then we could either return EngineInputs from the adapter or keep one common wrapping step in the handler, rather than an Audex-specific method

components/src/dynamo/vllm/omni/audio_handler.py
```diff
@@ -65,6 +72,7 @@ def __init__(self, config, engine_client, media_output_fs, media_output_http_url
         self.media_output_fs = media_output_fs
         self.media_output_http_url = media_output_http_url
         self._tts_tokenizer: Any = None
+        self.audex = AudexRequestAdapter(config, engine_client)
```
ptarasiewiczNV, line 83: I think this is fine for this PR, but could you please add a TODO to refactor the supported paths / adapters to use some common abstraction. This PR adds a third path and the logic is starting to get messy and it might get worse with more model types supported.

components/src/dynamo/vllm/omni/audex.py
```diff
@@ -0,0 +1,396 @@
...
+
+        if cfg_scale is None or cfg_scale <= 1.0:
+            extra_args.pop("cfg_scale", None)
+            return params_list
+
+        if request_id is None:
+            # The pair id must be unique per request; without it the guided and
+            # unconditional sequences cannot be matched, so fall back to
+            # unguided decoding rather than corrupting a concurrent pair.
+            logger.warning(
```
ptarasiewiczNV: Question, if someone provided `cfg_scale` wouldn't it make sense to raise an error here? This is very minor since normal path always provides the id.

ptarasiewiczNV (commented):
Thanks for the fixes @tthakkal ! I have added a few more minor comments/questions, but overall  LGTM

lib/llm/src/protocols/openai/audios.rs
tthakkal: Done — `NvExt::cfg_scale` in `audios/nvext.rs`, `AudioNvExt.cfg_scale` in `audio_protocol.py`; top-level field removed from both. Note `nvext` isn't a free-form bag — serde drops unknown keys there too, so the field still has to be declared. Two tests pin it: `audio_request_cfg_scale_reaches_the_worker` and `audio_request_top_level_cfg_scale_is_not_a_typed_field` (17 passed). Launch script curl + docs updated to `"nvext": {"cfg_scale": 1.5}`, and verified on hardware: guided differs from unguided, `cfg_scale: 20` → 400 naming `nvext.cfg_scale`, top-level value ignored.

components/src/dynamo/vllm/omni/audio_handler.py
tthakkal: Added on the `AudioGenerationHandler` docstring: a `supports()`/`prepare()` protocol to dispatch over, with Qwen3-TTS and the generic prompt as implementations alongside Audex. It names the three places detection currently spreads over (`_is_tts_model`, `self.audex.model_type()`, the `build_engine_inputs` branches). Keeping it out of this PR since the refactor would touch the Qwen3-TTS and MiMo-Audio paths, which this PR deliberately leaves unchanged.

components/src/dynamo/vllm/omni/audio_handler.py
tthakkal: Done — the helper is gone rather than simplified. `EngineInputs` now lives in `omni/engine_inputs.py` (re-exported from `omni_handler`, so no caller changed), which removes the circular import. The Audex branch in `build_engine_inputs` is now just `self.audex.prepare(...)` + the shared `_engine_inputs(...)`, and `prepare` returns a typed `AudexPreparedRequest(prompt, sampling_params_list)`.

components/src/dynamo/vllm/omni/audex.py
tthakkal: Agreed, now raises `RuntimeError`. The pair id *is* the final Dynamo request id, which every serving path supplies, so reaching there without one is a programming error, not a client one — and decoding unguided would answer a guided request with different-sounding audio, which is hard to spot in a waveform.

tthakkal (commented):
@ptarasiewiczNV Thank you for the review, addressed review comments.

lib/llm/src/protocols/openai/audios.rs
tthakkal: Done — `NvExt::cfg_scale` in `audios/nvext.rs`, `AudioNvExt.cfg_scale` in `audio_protocol.py`; top-level field removed from both. Note `nvext` isn't a free-form bag — serde drops unknown keys there too, so the field still has to be declared. Two tests pin it: `audio_request_cfg_scale_reaches_the_worker` and `audio_request_top_level_cfg_scale_is_not_a_typed_field` (17 passed). Launch script curl + docs updated to `"nvext": {"cfg_scale": 1.5}`, and verified on hardware: guided differs from unguided, `cfg_scale: 20` → 400 naming `nvext.cfg_scale`, top-level value ignored.

components/src/dynamo/vllm/omni/audio_handler.py
tthakkal: Done — the helper is gone rather than simplified. `EngineInputs` now lives in `omni/engine_inputs.py` (re-exported from `omni_handler`, so no caller changed), which removes the circular import. The Audex branch in `build_engine_inputs` is now just `self.audex.prepare(...)` + the shared `_engine_inputs(...)`, and `prepare` returns a typed `AudexPreparedRequest(prompt, sampling_params_list)`.

components/src/dynamo/vllm/omni/audio_handler.py
tthakkal: Added on the `AudioGenerationHandler` docstring: a `supports()`/`prepare()` protocol to dispatch over, with Qwen3-TTS and the generic prompt as implementations alongside Audex. It names the three places detection currently spreads over (`_is_tts_model`, `self.audex.model_type()`, the `build_engine_inputs` branches). Keeping it out of this PR since the refactor would touch the Qwen3-TTS and MiMo-Audio paths, which this PR deliberately leaves unchanged.

components/src/dynamo/vllm/omni/audex.py
tthakkal: Agreed, now raises `RuntimeError`. The pair id *is* the final Dynamo request id, which every serving path supplies, so reaching there without one is a programming error, not a client one — and decoding unguided would answer a guided request with different-sounding audio, which is hard to spot in a waveform.

ptarasiewiczNV (approved):
@tthakkal thank you for addressing the comments, approved. Please resolve the conflicts.

rmccorm4 (approved):
Approving the `lib/llm/src/protocols/openai/audios/nvext.rs` / `lib/llm/src/protocols/openai/audios.rs` changes for @ai-dynamo/dynamo-runtime-codeowners

PR #12792, opened 2026-08-07
chore(backends): remove stale unified backend implementations

tanmayv25:
## Summary

- Remove retired built-in unified implementation residue for vLLM and SGLang, including logits adapters, runtime tests, stale CLI options, and obsolete integration branches.
- Remove the obsolete TensorRT-LLM unified logits attachment shim while preserving the active TensorRT-LLM adapter.
- Preserve generic Backend SDK contracts and active backend behavior, including generation-stage handling, logits specifications, graceful cleanup callbacks, multimodal handling, and native engine paths.
- Clean related documentation, navigation, comments, and CODEOWNERS entries.

## Validation

- `ruff check` on the changed Python implementation and test surfaces
- `python -m py_compile` on changed Python files
- 6 focused graceful-shutdown lifecycle checks
- CODEOWNERS strict coverage: 5178/5178 files explicitly owned; generated output matches
- `python docs/fern/scripts/check_asset_paths.py`
- `git diff --check`

Not run locally: backend tests requiring the built `dynamo._core.backend` native extension. Fern and codespell CLIs are not installed in this environment.

## Summary by CodeRabbit

- **Documentation**
  - Updated backend guides to clarify custom backend development, engine management, metrics, tracing, cleanup, and multimodal serving.
  - Removed obsolete “unified” terminology and command-line flag references from deployment examples.
  - Consolidated vLLM and SGLang logits-processing documentation into backend reference guides.
  - Refreshed observability, TensorRT-LLM, Rust, and Python worker guidance.

- **Removed Features**
  - Removed backend-specific vLLM and SGLang logits-processing integrations and related configuration guidance.
  - Simplified forward-pass metrics configuration by removing relay-support overrides.


components/src/dynamo/common/backend/engine.py
```diff
@@ -441,41 +430,24 @@ class DiffusionEngine(RawEngine):
...
 
 @dataclass(frozen=True)
 class PythonProcessorSpec:
-    """Escape hatch for live, in-process `BaseLogitsProcessor` instances.
+    """Factory for an in-process Python logits processor.
 
-    Only TRT-LLM consumes it; vLLM/SGLang adapters reject it (not
-    serializable). Not used by the env-hook smoke.
+    This entry is suitable only when the engine integration can invoke the
+    factory directly. Serialization helpers reject it.
```
peilii: `PythonProcessorSpec` is now orphaned. `attach_logits_processors` in the TRT-LLM adapter — the only path that could realize it — was removed in this PR. The updated docstring says "suitable only when the engine integration can invoke the factory directly," but there is no such engine left in the repo. A custom-backend author who finds this type in the SDK will have no working realization path. Suggest either removing it or adding an explicit note that it is reserved for future in-process-only integrations.

components/src/dynamo/trtllm/logits_processing/adapter.py
```diff
@@ -88,32 +86,3 @@ def create_trtllm_adapters(
         adapter = TrtllmDynamoLogitsAdapter(processor)
         adapters.append(adapter)
     return adapters
-
-
-def attach_logits_processors(
-    sampling_params: Any,
```
peilii, line 94: With `attach_logits_processors` gone, is there another path in the TRT-LLM engine that calls `create_trtllm_adapters`? If TRT-LLM's `generate()` was reaching through `attach_logits_processors`, then any engine that overrides `logits_processor_spec()` will silently produce no effect — the spec entries will come back from `logits_processors_for_request` but never get realized into adapters.

components/src/dynamo/common/backend/tests/test_engine.py
```diff
@@ -105,321 +92,57 @@ async def test_default_abort_is_noop():
...
-        spec.entries
-    )
+def test_is_generation_stage(mode, expected):
+    assert is_generation_stage(mode) is expected
 
 
 @pytest.mark.parametrize(
-    "mode", [DisaggregationMode.PREFILL, DisaggregationMode.ENCODE]
+    ("mode", "generation_only", "expected_count"),
+    [
```
peilii, line 114: The consolidated parametrize covers DECODE/PREFILL/PREFILL-non-generation-only but drops the `spec=None` case. The old `test_build_returns_empty_when_plan_none` covered that branch explicitly. The `if spec is None: return []` guard in `logits_processors_for_request` is now untested — easy to add as a fourth row, e.g. `(DisaggregationMode.DECODE, None, 0)`, with `None` passed as the spec.

components/src/dynamo/vllm/args.py
```diff
@@ -245,18 +236,14 @@ def _unsupported_fpm_trace_role(dynamo_config: Config) -> Optional[str]:
...
     """Resolve FPM activation without changing the legacy explicit-port path."""
     if envs.is_set("DYN_FORWARDPASS_METRIC_PORT"):
         return True
     if not dynamo_config.fpm_trace:
         return False
 
     unsupported_role = _unsupported_fpm_trace_role(dynamo_config)
-    if unsupported_role is None and not fpm_trace_relay_supported:
-        unsupported_role = "unified backend"
     if unsupported_role is None:
```
peilii, line 247: Removing this guard means `_forward_pass_metrics_enabled` now returns `True` for any supported role, regardless of whether the caller constructs the FPM relay. Previously, unified-backend entry points explicitly passed `fpm_trace_relay_supported=False` to suppress FPM when the relay wasn't built. If those call sites still don't construct the relay, this is a silent behavior change. Is this intentional because the relay is now always set up for all vLLM entry points?

components/src/dynamo/sglang/args.py
```diff
@@ -86,18 +86,14 @@ def _unsupported_fpm_trace_role(dynamo_config: DynamoConfig) -> Optional[str]:
...
     """Resolve the FPM opt-in source while preserving the legacy port switch."""
     if os.environ.get("DYN_FORWARDPASS_METRIC_PORT"):
         return "DYN_FORWARDPASS_METRIC_PORT"
     if not dynamo_config.fpm_trace:
         return None
 
     unsupported_role = _unsupported_fpm_trace_role(dynamo_config)
-    if unsupported_role is None and not fpm_trace_relay_supported:
-        unsupported_role = "unified backend"
     if unsupported_role is None:
```
peilii, line 97: Same as the vLLM change: removing this guard enables FPM for all SGLang entry points on a supported role. If any SGLang entry point previously relied on `fpm_trace_relay_supported=False` (i.e. didn't build the relay), this is a behavior change worth confirming is intentional.

connorcarpenter15 (approved):

dagil-nvidia (approved):

peilii (approved):
Clean removal — deletions are well-scoped and the common spec layer is correctly preserved as SDK contract. Comments left inline on a few things to address.

PR #12797, opened 2026-08-07
perf(mocker): avoid per-token replay request lookups

jthomson04:
## Overview

This PR removes the per-token UUID hash-map lookup from aggregate offline replay while keeping live and disaggregated output publication on their existing untracked path. The radix-tree optimization previously included here has moved to an independent PR.

## Summary

- Carry one private optional `ReplayRequestKey` through the existing engine request path and preserve complete backend request-ID validation.
- Store aggregate replay state in a generational `SlotMap`; retain the UUID index only for routing, admission, duplicate detection, and validated terminal removal.
- Keep `EngineOutputs` small: the original `Vec<OutputSignal>` plus private `Empty`, `Untracked`, `Tracked`, and `Mixed` bookkeeping. Appends are infallible; release builds reject mixed tracking, missing keys, stale/reused keys, UUID mismatches, and tracked outputs at live/disaggregated publication boundaries.
- Derive the first nonterminal prefill transition from runtime-owned state, including tokenless nonterminal signals. A first-and-terminal request still skips `prefill_completed`.
- Move the original output vector directly through live and disaggregated publication with no wrapper conversion or replacement allocation.
- Reject duplicate active UUIDs before observable side effects in aggregate and disaggregated replay.
- Preserve output order, speculative bursts, live suppression, attention-DP payloads, one-token completion behavior, and public replay/event formats.

The final diff is 1,025 additions and 311 deletions across 22 files. The follow-up review fix is 303 additions and 176 deletions; it replaces fallible post-mutation output appends and split validation/removal paths instead of adding another tracking abstraction. The obsolete marker-index machinery, wrapper traits, tracked-retain path, and every KV-router change remain gone.

## Details

Each aggregate output carries its request's private generational key in a sidecar aligned with the original signal vector. Nonterminal processing validates the slot and its UUID directly in O(1), without hashing the reverse UUID index. Terminal processing validates both indexes and removes the request before collector, placement, admission, or routing callbacks; index corruption is returned as an error without partially mutating either index.

Live and disaggregated paths never create replay keys. Their publication boundaries call `into_untracked()`, which rejects unexpected metadata and returns the owned `Vec<OutputSignal>` directly. vLLM also fails immediately if request state disappears while generating outputs.

No public API, configuration, event format, scheduling policy, or simulation behavior changes.

- `lib/mocker/src/scheduler/mod.rs` for `EngineRequest` and the compact `EngineOutputs` sidecar.
- `lib/mocker/src/replay/offline/agg.rs` for aggregate `SlotMap` state and output validation.
- `lib/mocker/src/scheduler/vllm/core.rs` and `lib/mocker/src/scheduler/sglang/core.rs` for backend key propagation and request-ID checks.

## Related Issues

None.

### Source validation

- `cargo test -p dynamo-mocker replay::offline` — 192 passed
- `cargo test -p dynamo-mocker` — 721 passed
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline` — 194 passed; the exact base reproduces the same two existing failures:
  - `busy_worker_advances_offload_without_destination_admission`
  - `aggregated_canonical_evidence_covers_real_kvbm_lifecycle`
- `git diff --check`

Focused coverage includes duplicate insertion and bidirectional index integrity, completed stale/reused/UUID-mismatched keys before callbacks, mixed-boundary rejection after infallible appends, propagated live-suppression errors, vLLM rejection metadata, nonzero completion-batch counters, zero-copy untracked publication, tokenless nonterminal output, first-and-terminal completion, speculative output, attention-DP propagation, and both scheduler backends.

### Canonical correctness parity

The full correctness-parity protocol passed between exact base `038c642b56a01b0afddd8c194565526faae35d26` and final head `88cefb8f710248b09fd8a10e6c827e3e6665e397`. The preserved two-process base half and a freshly run two-process final-head half cover 16 reports across aggregate/disaggregated SGLang and native-vLLM KVBM. Every report completed 5,000 requests, each final pair was byte-identical, and every final canonical report was byte-identical to its preserved base counterpart.

| Row | Canonical SHA-256 | Pressure coverage |
| --- | --- | --- |
| SGLang aggregate | `84d13ec294115d63621805ceaa37e38c4069c38028ceac0b023f0c7b1a7d8a11` | KV overlap and immediate/queued placement |
| SGLang disaggregated | `ed7f2268248c118ed6b9aba7bd47f8c0d7476d4be968bda07097fab23a5a86a6` | 5,000 handoffs; 12 bounded retractions |
| vLLM KVBM aggregate | `079e11aeb65a19d0cb2dcf516e517ae88c8e4733170a2cb413d6281fcd3469c5` | 2 preemptions; nonzero G1/G2 activity |
| vLLM KVBM disaggregated | `f9e3924ad429da3ddeed49919092343f49c8c5e1426e89ddd8b7a8cb1f113cf0` | 5,000 handoffs; 1 preemption; nonzero G1/G2 activity |

### Astra vLLM point comparison

Ordinary release builds used AIConfigurator `0fa09bb7b5`, 12 aggregate vLLM workers pinned to CPU 0, B200 TP4/attention-DP1 and MoE TP1/EP4, 64-token pages, 4,096-token batching/chunking, prefix caching, KV routing/events, and the 187,906-request 2x Astra trace (`b7b2e3dccb90...`). There were no warm-ups. Preflight was 99.67–100% idle on CPU 0.

| Revision | Wall | Processed tok/s | vs. base | vs. pre-fix |
| Base `038c642b56` | 60.149 s | 42.907 M | — | — |
| Pre-fix `de5f5c86c9` | 57.595 s | 44.810 M | +4.43% | — |
| Final `88cefb8f71` | 57.376 s | 44.981 M | **+4.83%** | +0.38% |

All three reports completed 187,906 requests with 2,401,594,588 input tokens, 179,237,676 output tokens, and 2,580,832,264 processed tokens. These are CPU-0 point comparisons, not a statistical-significance claim; routing-induced cache reuse varied from 41.829% to 42.162%.

The final extension SHA-256 is `9b1d5dcb623f8f1eb44377ac6193c409c5f8d880f9ddf024bb019a7726e008bb` (ELF build ID `aadd75e6ff872a1bf15dff7db33588c5af252e65`); the preserved base extension SHA-256 is `a0f9ab62861164069308f6388fb9e8b3c2ba2e269bf0b4c167e56712bb36b6ca`.

### Structural hot-path confirmation

- Aggregate nonterminal outputs validate `ReplayRequestKey` directly in the `SlotMap`; the UUID index is not consulted per token.
- Live and disaggregated boundaries consume the owned signal vector through `into_untracked()` and reject any sidecar, so no conversion allocation is introduced.
- The prior profile's removed per-token UUID hashing remains absent; this rewrite removes wrapper/marker scaffolding without restoring that call path.

### DIFF
--- a/lib/mocker/src/replay/offline/agg.rs
+++ b/lib/mocker/src/replay/offline/agg.rs
@@ -39,17 +39,102 @@ use super::{
     },
     state::AggRequestState,
 };
-use crate::common::protocols::{DirectRequest, ForwardPassSnapshot, MockEngineArgs, OutputSignal};
+use crate::common::protocols::{DirectRequest, ForwardPassSnapshot, MockEngineArgs};
 use crate::loadgen::{ReplayRequestPayload, WorkloadDriver};
 #[cfg(test)]
 use crate::replay::ReplayRouterMode;
 use crate::replay::{ReplayRequestPool, ReplayTerminalStatus, TraceCollector};
-use anyhow::bail;
+use anyhow::{anyhow, bail, ensure};
 use rustc_hash::FxHashMap;
+use slotmap::SlotMap;
 #[cfg(test)]
 use std::collections::HashMap;
 use std::collections::{BinaryHeap, VecDeque};
 use uuid::Uuid;
+
+use crate::scheduler::{EngineOutputs, EngineRequest, ReplayRequestKey};
+
+#[derive(Default)]
+struct AggRequests {
+    by_uuid: FxHashMap<Uuid, ReplayRequestKey>,
+    states: SlotMap<ReplayRequestKey, (Uuid, AggRequestState)>,
+}
+
+impl AggRequests {
+    fn contains_uuid(&self, uuid: Uuid) -> bool {
+        self.by_uuid.contains_key(&uuid)
+    }
+
+    fn insert(&mut self, uuid: Uuid, state: AggRequestState) -> ReplayRequestKey {
+        debug_assert!(!self.contains_uuid(uuid));
+        let key = self.states.insert((uuid, state));
+        self.by_uuid.insert(uuid, key);
+        key
+    }
+
+    fn get_mut_by_uuid(
+        &mut self,
+        uuid: Uuid,
+    ) -> anyhow::Result<(ReplayRequestKey, &mut AggRequestState)> {
+        let key =
+            self.by_uuid.get(&uuid).copied().ok_or_else(|| {
+                anyhow!("offline aggregate replay missing request state for {uuid}")
+            })?;
+        Ok((key, self.get_mut_by_key(key, uuid)?))
+    }
+
+    fn get_mut_by_key(
+        &mut self,
+        key: ReplayRequestKey,
+        uuid: Uuid,
+    ) -> anyhow::Result<&mut AggRequestState> {
+        let (stored_uuid, state) = self
+            .states
+            .get_mut(key)
+            .ok_or_else(|| anyhow!("offline aggregate replay missing request state for {uuid}"))?;
+        ensure!(
+            *stored_uuid == uuid,
+            "offline aggregate replay request state key does not match {uuid}"
+        );
+        Ok(state)
+    }
+
+    fn validate_key(&self, key: ReplayRequestKey, uuid: Uuid) -> anyhow::Result<()> {
+        self.get_by_key(key, uuid)?;
+        ensure!(
+            self.by_uuid.get(&uuid).copied() == Some(key),
+            "offline aggregate replay request state indexes disagree for {uuid}"
+        );
+        Ok(())
+    }
+
+    fn remove_validated(&mut self, key: ReplayRequestKey, uuid: Uuid) -> AggRequestState {
+        self.by_uuid
+            .remove(&uuid)
+            .expect("validated replay UUID must remain present");
+        self.states
+            .remove(key)
+            .expect("validated replay key must remain present")
+            .1
+    }
+
+    fn get_by_key(&self, key: ReplayRequestKey, uuid: Uuid) -> anyhow::Result<&AggRequestState> {
+        let (stored_uuid, state) = self
+            .states
+            .get(key)
+            .ok_or_else(|| anyhow!("offline aggregate replay missing request state for {uuid}"))?;
+        ensure!(
+            *stored_uuid == uuid,
+            "offline aggregate replay request state key does not match {uuid}"
+        );
+        Ok(state)
+    }
+
+    #[cfg(test)]
+    fn iter(&self) -> impl Iterator<Item = (Uuid, &AggRequestState)> {
+        self.states.values().map(|(uuid, state)| (*uuid, state))
+    }
+}
 
 fn common_origin(mut origins: impl Iterator<Item = u64>) -> Option<u64> {
     let first = origins.next()?;
@@ -126,7 +211,7 @@ where
     next_event_seq: u64,
     next_scaling_tick_ordinal: u64,
     admission: AdmissionQueue<Metadata>,
-    requests: FxHashMap<Uuid, AggRequestState>,
+    requests: AggRequests,
     engine: EngineComponent<Observation>,
     collector: TraceCollector,
     events: BinaryHeap<SimulationEvent<Observation::Batch>>,
@@ -223,7 +308,7 @@ where
             next_event_seq: 0,
             next_scaling_tick_ordinal: 0,
             admission,
-            requests: FxHashMap::default(),
+            requests: AggRequests::default(),
             engine,
             collector,
             events: BinaryHeap::new(),
@@ -355,8 +440,12 @@ where
         request: DirectRequest,
         uuid: Uuid,
         worker_idx: usize,
+        replay_request_key: ReplayRequestKey,
     ) -> anyhow::Result<()> {
-        self.engine.dispatch(worker_idx, request)?;
+        self.engine.dispatch_engine(
+            worker_idx,
+            EngineRequest::for_replay(request, replay_request_key),
+        )?;
         self.record_dispatch(uuid, worker_idx);
         // Aggregated replay uses a single pool. Treat the assignment as the
         // decode_worker_idx so per-request records consistently carry the
@@ -400,14 +489,9 @@ where
                 dp_rank,
                 placement.reported_overlap_tokens,
             );
-            let request = self
-                .requests
-                .get_mut(&uuid)
-                .ok_or_else(|| {
-                    anyhow::anyhow!("offline replay missing queued request state for {uuid}")
-                })?
-                .take_queued_request(uuid)?;
-            self.dispatch_to_worker(request, uuid, placement.scheduler_id)?;
+            let (replay_request_key, request) = self.requests.get_mut_by_uuid(uuid)?;
+            let request = request.take_queued_request(uuid)?;
+            self.dispatch_to_worker(request, uuid, placement.scheduler_id, replay_request_key)?;
         }
         Ok(())
     }
@@ -421,6 +505,9 @@ where
         session_id: Option<String>,
     ) -> anyhow::Result<Uuid> {
         let uuid = request.metadata().uuid.unwrap_or_else(Uuid::new_v4);
+        if self.requests.contains_uuid(uuid) {
+            bail!("offline replay request {uuid} is already active");
+        }
         let input_length = request.input_length();
         let output_length = request.metadata().max_output_tokens;
         request.metadata_mut().uuid = Some(uuid);
@@ -461,7 +548,7 @@ where
                     dp_rank,
                     placement.reported_overlap_tokens,
                 );
-                self.requests.insert(
+                let replay_request_key = self.requests.insert(
                     uuid,
                     AggRequestState::new_running(input_length, output_length),
                 );
@@ -469,6 +556,7 @@ where
                     request.into_direct_request(),
                     uuid,
                     placement.scheduler_id,
+                    replay_request_key,
                 )?;
             }
             PlacementDecision::Queued => {
@@ -563,11 +651,20 @@ where
     }
 
     /// Consume one output signal, updating router state, collector state, and completion counts.
-    fn process_output_signal(&mut self, signal: OutputSignal) -> anyhow::Result<()> {
-        if let Some(token_id) = signal.token_id {
-            CoreAdmissionSource::on_output_token(&mut self.admission, signal.uuid, token_id)?;
-        }
+    fn process_output_signal(
+        &mut self,
+        signal: crate::common::protocols::OutputSignal,
+        replay_request_key: ReplayRequestKey,
+    ) -> anyhow::Result<()> {
         if signal.completed {
+            // Validate the generational key before publishing any terminal
+            // side effects. Terminal signals are rare enough that the later
+            // removal can retain its own defensive validation.
+            self.requests
+                .validate_key(replay_request_key, signal.uuid)?;
+            if let Some(token_id) = signal.token_id {
+                CoreAdmissionSource::on_output_token(&mut self.admission, signal.uuid, token_id)?;
+            }
             let status = if signal.rejected {
                 ReplayTerminalStatus::Rejected
             } else {
@@ -582,9 +679,9 @@ where
                 self.stats.router_freed_count += 1;
             }
             self.record_router_pending();
-            let removed_state = self.requests.remove(&signal.uuid).ok_or_else(|| {
-                anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
-            })?;
+            let removed_state = self
+                .requests
+                .remove_validated(replay_request_key, signal.uuid);
             // Rejected requests never ran: keep them out of completed-request
             // shape and latency samples. Their offered demand was already
             // recorded at arrival, matching requests_started_total.
@@ -617,23 +714,24 @@ where
             return Ok(());
         }
 
-        let already_marked = self
-            .requests
-            .get(&signal.uuid)
-            .ok_or_else(|| {
-                anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
-            })?
-            .prefill_completed;
-        if already_marked {
+        let marks_prefill_completion = {
+            let state = self
+                .requests
+                .get_mut_by_key(replay_request_key, signal.uuid)?;
+            if state.prefill_completed {
+                false
+            } else {
+                state.prefill_completed = true;
+                true
+            }
+        };
+        if let Some(token_id) = signal.token_id {
+            CoreAdmissionSource::on_output_token(&mut self.admission, signal.uuid, token_id)?;
+        }
+        if !marks_prefill_completion {
             return Ok(());
         }
 
-        self.requests
-            .get_mut(&signal.uuid)
-            .ok_or_else(|| {
-                anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
-            })?
-            .prefill_completed = true;
         let placements = self.placement.prefill_completed(signal.uuid, self.now_ms)?;
         #[cfg(test)]
         if self.placement.is_router() {
@@ -663,9 +761,7 @@ where
     /// Apply one completed pass: free request slots, publish KV events, and handle outputs.
     fn process_completed_pass(
         &mut self,
-        _worker_idx: usize,
-        _completed_requests: usize,
-        output_signals: Vec<OutputSignal>,
+        output_signals: EngineOutputs,
         engine_events: Observation::Batch,
         accept_length_output_tokens: usize,
         accept_length_decode_forwards: usize,
@@ -673,8 +769,8 @@ where
         self.apply_engine_observations(engine_events, KvIngestBoundary::PassEnd)?;
         self.traffic
             .on_accept_length_sample(accept_length_output_tokens, accept_length_decode_forwards);
-        for signal in output_signals {
-            self.process_output_signal(signal)?;
+        for (signal, replay_request_key) in output_signals.into_tracked()? {
+            self.process_output_signal(signal, replay_request_key)?;
         }
         Ok(())
     }
@@ -715,8 +811,6 @@ where
             self.record_fpm(payload.worker_idx, fpm)?;
         }
         self.process_completed_pass(
-            payload.worker_idx,
-            payload.completed_requests,
             payload.output_signals,
             payload.engine_events,
             payload.accept_length_output_tokens,
@@ -1323,14 +1417,14 @@ where
             .requests
             .iter()
             .filter(|(_, state)| state.phase == AggRequestPhase::QueuedAtRouter)
-            .map(|(uuid, _)| *uuid)
+            .map(|(uuid, _)| uuid)
             .collect::<Vec<_>>();
         router_pending_request_ids.sort_unstable();
         let mut prefill_completed = self
             .requests
             .iter()
             .filter(|(_, state)| state.prefill_completed)
-            .map(|(uuid, _)| *uuid)
+            .map(|(uuid, _)| uuid)
             .collect::<Vec<_>>();
         prefill_completed.sort_unstable();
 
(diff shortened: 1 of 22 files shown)


lib/mocker/src/replay/offline/agg.rs
```diff
@@ -617,23 +622,24 @@ where
...
 
-        let already_marked = self
-            .requests
-            .get(&signal.uuid)
-            .ok_or_else(|| {
-                anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
-            })?
-            .prefill_completed;
-        if already_marked {
+        if !marks_prefill_completion {
```
jthomson04, line 735: **Missing request state downgraded to a debug-only assert.**

Previously *every* nonterminal signal went through `self.requests.get(&signal.uuid).ok_or_else(|| anyhow!("offline replay missing request state for {}", ...))?`, so a signal for an unknown uuid aborted the replay. Now, when `marks_prefill_completion == false` — which is all decode tokens after the first, i.e. >99% of signals — this returns `Ok(())` without ever touching `self.requests`, and the `debug_assert!` above is compiled out in release.

**Failure:** a worker emits a token for a uuid already removed by a terminal signal, or one never assigned (e.g. a stale completion payload after scale-down). The release-build replay completes normally and reports plausible-but-wrong throughput / TTFT / completion counts instead of aborting.

lib/mocker/src/replay/offline/agg.rs
```diff
@@ -617,23 +622,24 @@ where
...
+            #[cfg(debug_assertions)]
```
jthomson04: Nit: `#[cfg(debug_assertions)]` on a `debug_assert!` is redundant — the macro already expands to nothing without `debug_assertions`.

Worth removing because the identical construct is load-bearing at `vllm/core.rs:1636` and `sglang/core.rs:772`, where it guards a real function call. Keeping the no-op version here means a reader can't tell from the pattern which gates matter. The three plain `debug_assert!`s fifteen lines below show the intended style.

lib/mocker/src/replay/offline/agg.rs
```diff
@@ -617,23 +622,24 @@ where
...
-        self.requests
-            .get_mut(&signal.uuid)
-            .ok_or_else(|| {
-                anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
-            })?
-            .prefill_completed = true;
+        let state = self.requests.get_mut(&signal.uuid).ok_or_else(|| {
+            anyhow::anyhow!("offline replay missing request state for {}", signal.uuid)
+        })?;
+        debug_assert!(!state.prefill_completed);
```
jthomson04: **A missed marker silently skips `prefill_completed()` forever.**

The old code derived this transition from replay-owned state (`state.prefill_completed`) and structurally could not miss it. It's now driven entirely by engine-produced markers whose correctness is enforced only by `debug_assert`.

**Failure:** if any emission path produces a request's first nonterminal token without pushing an index — a new early-return in `emit_ready_tokens`, or a new `EnginePassResult` producer defaulting the field to `Vec::new()` the way `engine.rs:797` does — then in release `state.prefill_completed` stays false and `PlacementPolicy::prefill_completed` → `KvRouter::mark_prefill_completed` is never called for that request. The router keeps counting it as prefilling for its entire lifetime, skewing KV-router load/decode accounting and routing decisions. No error, no log line.

lib/mocker/src/replay/offline/agg.rs
```diff
@@ -663,19 +669,26 @@ where
...
         accept_length_decode_forwards: usize,
     ) -> anyhow::Result<()> {
         self.apply_engine_observations(engine_events, KvIngestBoundary::PassEnd)?;
         self.traffic
             .on_accept_length_sample(accept_length_output_tokens, accept_length_decode_forwards);
-        for signal in output_signals {
-            self.process_output_signal(signal)?;
+        let mut prefill_completion_indices = prefill_completion_indices.into_iter().peekable();
+        for (index, signal) in output_signals.into_iter().enumerate() {
+            let marks_prefill_completion = prefill_completion_indices
```
jthomson04: **One bad index drops every remaining marker in the pass, not just the bad one.**

`prefill_completion_indices.next()` is only called on an exact `completion_index == index` match, so a non-matching head never advances.

**Failure:** given indices `[2, 1]` (unsorted) or `[0, 99]` (out of range) in a release build where `debug_assert_prefill_completion_indices` is compiled out, marker `1` (resp. `99`) never matches — so `process_output_signal` is called with `marks_prefill_completion == false` for that request *and for every later request in the same pass* whose marker sits behind the stuck head. One malformed index loses the prefill-completion transition for an unbounded number of requests, and the trailing `debug_assert!(prefill_completion_indices.next().is_none())` that would catch it is compiled out too.

lib/mocker/src/scheduler/mod.rs
```diff
@@ -145,6 +145,28 @@ pub(crate) fn accept_length_sample(output_signals: &[OutputSignal]) -> (usize, u
...
+    debug_assert!(indices.windows(2).all(|pair| pair[0] < pair[1]));
+    for &index in indices {
+        let signal = output_signals
+            .get(index)
+            .expect("prefill completion index must reference an output signal");
+        debug_assert!(signal.token_id.is_some());
+        debug_assert!(!signal.completed);
+        debug_assert!(!signal.rejected);
+        debug_assert!(
+            output_signals[..index]
```
jthomson04: **This is O(n·m) — it rescans the whole prefix for every marker, on every pass, in every debug/test build.**

For each marker index this runs `output_signals[..index].iter().all(|earlier| earlier.uuid != signal.uuid)`. A vLLM pass with `max_running_requests` in the low thousands and speculative bursts produces thousands of signals, and the first pass of a large batch has a marker for nearly *every* request — so ~n²/2 uuid comparisons per pass (~10⁶ at n=2000), on every one of millions of passes.

That measurably slows `cargo test -p dynamo-mocker` and any debug-build replay, which is the opposite of this PR's goal. Building one `FxHashSet<Uuid>` of already-seen uuids in a single walk over `output_signals` gives the same check in O(n).

jthomson04 (commented):
## Automated code review

13 findings. The dominant theme is the move of prefill-completion detection from replay-owned state to engine-produced marker indices carried as a parallel array: the top three findings are all consequences of that change, and all three are **silent in release builds** — which are the builds that produced the benchmark numbers in the PR description.

The radix-tree batching is lower risk, but batching landed only on the single-threaded `RadixTree` while `ConcurrentRadixTree` keeps the per-hash loop, and the new tests only compare batched-vs-serial within the single-threaded implementation.

---

### Two findings on files outside this diff

**`lib/mocker/src/replay/offline/README.md` (~line 153) — stale field list.**
The README enumerates the `WorkerCompletion` event fields exhaustively and wasn't updated for `prefill_completion_indices`. Both the code block (`WorkerCompletion { stage, worker_idx, completed_requests, output_signals, kv_events }`) and the prose below it now describe a shape that no longer matches `events.rs`. Someone adding a new completion producer from this doc will construct the payload without the marker list — which, per the inline finding on `agg.rs:640`, silently drops prefill-completion routing effects rather than failing loudly.

**`lib/mocker/src/replay/offline/disagg.rs:1897` — dead-param cleanup not applied to the twin.**
This PR deleted `_worker_idx` / `_completed_requests` from `AggRuntime::process_completed_pass`, but `DisaggRuntime::process_prefill_pass` still takes the identical dead pair and threads them from `apply_worker_completion`. The two sibling runtimes now differ for no reason, and the next reader has to work out whether disagg's copies are load-bearing.

lib/mocker/src/scheduler/mod.rs
```diff
@@ -180,6 +202,11 @@ pub(crate) struct EnginePassResult {
     pub(crate) token_completion_ms: f64,
     pub(crate) completed_requests: usize,
     pub(crate) output_signals: Vec<OutputSignal>,
+    /// Indices of first token-bearing, nonterminal outputs for requests that
+    /// have not previously emitted a token. Aggregated replay uses these to
+    /// apply prefill-completion effects without probing request state for every
+    /// subsequent decode token.
+    pub(crate) prefill_completion_indices: Vec<usize>,
```
jthomson04: **Design concern: this flag is a parallel array whose alignment is maintained by hand at three sites and validated only by debug asserts.**

`prefill_completion_indices` must stay index-aligned with `output_signals` across `EnginePassResult` → `WorkerCompletionPayload` → `SimulationEventKind` → payload → `AggRuntime`. Three sites already maintain that correspondence manually: the speculative offset arithmetic at `vllm/core.rs:2252`, the retain-and-remap in `live_boundary.rs:322`, and the peek-match in `agg.rs:683`. Any future code that filters, splits, sorts, or dedups `output_signals` — a one-line `retain` — silently desynchronizes the markers, with no release-build detection.

`EnginePassResult` is crate-internal, so carrying the flag *next to* each signal inside the pass result (e.g. `Vec<(OutputSignal, bool)>`, split at the publish boundary) would make desync structurally impossible while leaving the serialized public `OutputSignal` wire format untouched.

Separately, worth noting before this gets wired into `disagg.rs`: a disagg decode worker starts at `generated_tokens() == 0` for a request the prefill worker already marked, so the same predicate would double-fire there.

lib/kv-router/src/indexer/radix_tree.rs
```diff
@@ -548,20 +548,19 @@ impl RadixTree {
...
 
-        for block_hash in remove.block_hashes {
-            let Some(node) = self
-                .lookup
-                .get(&worker)
-                .and_then(|lookup| lookup.get(&block_hash))
-                .cloned()
-            else {
+        while group_start < block_hashes.len() {
+            let block_hash = block_hashes[group_start];
```
jthomson04: **Batching landed only on the single-threaded `RadixTree`; `ConcurrentRadixTree` (the thread-pool production path) still uses the per-hash loop.**

`lib/kv-router/src/indexer/concurrent_radix_tree_compressed/remove.rs` keeps the one-hash-at-a-time `remove_worker_at_pos` + `stale_hashes` loop. The two removal implementations now differ structurally and have to be kept semantically in lockstep by hand, but the two new tests (`batched_same_edge_removal_matches_serial_orderings`, `batched_removal_preserves_mixed_nodes_duplicates_and_errors`) only compare batched `RadixTree` against serial `RadixTree`.

A future edit to `NodeState::remove_worker_at_pos` or `covers_pos` that is valid per-hash but breaks the min-position grouping assumption would diverge the two indexers' routing results without any test failing.

lib/kv-router/src/indexer/radix_tree.rs
```diff
@@ -572,29 +571,67 @@ impl RadixTree {
...
             };
 
-            let outcome = {
-                let mut node_ref = node.borrow_mut();
-                let Some(&pos) = node_ref.state.edge_index.get(&block_hash) else {
+            let (group_end, min_pos, min_hash) = {
+                let node_ref = node.borrow();
+                let Some(&first_pos) = node_ref.state.edge_index.get(&block_hash) else {
+                    drop(node_ref);
+                    lookup.insert(block_hash, node);
```
jthomson04: **This `remove`-then-restore-by-hand protocol replaces a structurally safe read-only `get`.**

The old loop did `self.lookup.get(&worker).and_then(|l| l.get(&block_hash)).cloned()`, so no early `continue` could ever lose a lookup entry. The new code does `lookup.remove(&block_hash)` up front and relies on this single `insert` to undo it when `edge_index` doesn't contain the hash.

**Failure:** any future early return added between the `remove` and the `remove_worker_at_pos` call — for example when implementing the `TODO(CORRECTNESS)` descendant invalidation directly below — silently evicts a live block from that worker's index while it remains in the tree, permanently under-reporting the worker's cached prefix to `find_matches`. No error, no log line.

lib/kv-router/src/indexer/radix_tree.rs
```diff
@@ -572,29 +571,67 @@ impl RadixTree {
...
                     first_error.get_or_insert(KvCacheEventError::BlockNotFound);
+                    group_start += 1;
                     continue;
                 };
+                if !node_ref.state.covers_pos(worker, first_pos) {
+                    (group_start + 1, first_pos, block_hash)
+                } else {
+                    let mut group_end = group_start + 1;
+                    let mut min_pos = first_pos;
+                    let mut min_hash = block_hash;
```
jthomson04: **`min_hash` is write-only on this path — `remove_worker_at_pos` can never read it here.**

This branch is entered only when `covers_pos(worker, first_pos)` holds, and every candidate added to the group also passes `covers_pos`. Since `covers_pos(w, p)` is exactly `p < current_cutoff(w)`, `min_pos < current_cutoff` is guaranteed — so `remove_worker_at_pos` always takes its `else` branch and never reads `removed_hash`.

So this initializer and the `min_hash = candidate_hash;` assignment in the per-candidate loop below are dead work, costing an extra move per candidate in the hottest loop of a perf-motivated PR. Only the `!covers_pos` single-hash branch needs a hash, and there it's always `block_hash`.

lib/mocker/src/scheduler/live_boundary.rs
```diff
@@ -320,9 +320,26 @@ impl PendingLivePass {
     }
 
     pub(crate) fn suppress_request_outputs(&mut self, request_id: uuid::Uuid) {
```
jthomson04: **This ~20-line remap maintains a field that nothing in the live path ever reads.**

`prefill_completion_indices` is consumed only by `AggRuntime::process_completed_pass` in offline aggregated replay. `PendingLivePass` exists only in this file, and `publish_pass` moves out `pending.pass.output_signals` and drops the rest of the pass.

So on every mid-pass cancellation this allocates a `marker_uuids` Vec, clears the index list, and rebuilds it by scanning all surviving signals — purely to keep an invariant no consumer checks. `self.pass.prefill_completion_indices.clear()` is equivalent and O(1). If invariant hygiene is the goal instead, call `debug_assert_prefill_completion_indices` afterwards so the remap is at least verified.

lib/mocker/src/scheduler/live_boundary/tests.rs
```diff
@@ -275,6 +276,7 @@ fn output_suppression_repairs_stale_accept_length_counters() {
             handoff_delay_ms: None,
         },
     ]);
+    pass.prefill_completion_indices = vec![1];
```
jthomson04: **This is the only test for the new remap, and it covers neither interesting branch.**

Index `1` belongs to the *surviving* request, so this exercises only "one surviving marker shifts down by one". Untested:

1. The `.filter(|uuid| *uuid != request_id)` path — suppressing a request that itself owns a marker.
2. Two or more markers remapping together, which is where the order-sensitive `peekable` scan could regress.

A bug in either branch (e.g. filtering by index instead of uuid) would pass CI.

lib/mocker/src/replay/offline/runtime_utils.rs
```diff
@@ -291,11 +294,12 @@ mod tests {
             completed_requests,
             output_signals: vec![OutputSignal {
                 uuid: Uuid::from_u128(worker_idx as u128),
-                token_id: None,
-                completed: true,
+                token_id: Some(1),
+                completed: false,
```
jthomson04: **This flips the fixture to a state the engine can't actually produce.**

The payload carries `completed_requests: 1` / `2` while its single `OutputSignal` now has `completed: false` — a combination `publish_pass`'s own `debug_assert_eq!(pass.completed_requests, completed_signals)` would reject.

`test_worker_completion_batch_preserves_payload_and_event_ordering` still passes because it only reads `worker_idx` / `completed_requests`, but the fixture no longer models any reachable state, and this also removes the only terminal-signal coverage from the heap-ordering test. It'll mislead whoever reuses it next.

lib/mocker/src/replay/offline/agg.rs
jthomson04: Fixed in `0fef22c20a`. Aggregate request state now lives in a `SlotMap` keyed by a private generational `ReplayRequestKey`, and every output performs release-build slot/UUID validation. Missing, stale, reused, and UUID-mismatched keys return an error; focused tests cover each case.

lib/mocker/src/replay/offline/agg.rs
jthomson04: Fixed in `0fef22c20a`. The marker validator and its redundant `cfg(debug_assertions)` wrapper were removed with the marker-index design. Validation is now part of the release path through the generational slot lookup.

lib/mocker/src/replay/offline/agg.rs
jthomson04: Fixed in `0fef22c20a`. Engines no longer produce first-token markers. Each private output carries its request key, and aggregate replay derives the first nonterminal transition from runtime-owned `prefill_completed` state after validating the slot. Missing metadata fails instead of silently skipping the transition.

lib/mocker/src/replay/offline/agg.rs
jthomson04: Fixed in `0fef22c20a`. `prefill_completion_indices`, the peek/match loop, and all index ordering/range failure modes were removed. Every output is independently validated through its own generational key.

lib/mocker/src/scheduler/mod.rs
jthomson04: Fixed in `0fef22c20a`. The O(n*m) debug marker validator was deleted along with marker indices. The replacement performs one O(1) SlotMap validation per nonterminal output in all build modes.

lib/mocker/src/scheduler/mod.rs
jthomson04: Fixed in `0fef22c20a`. The parallel array is gone. A private `EngineOutputSignal` keeps the public signal and optional replay key together through pass/completion transport, making filtering structurally safe. Live and disaggregated paths strip the metadata at publication, so disaggregated decode cannot fire aggregate prefill transitions.

lib/kv-router/src/indexer/radix_tree.rs
jthomson04: Clarified and covered in `0fef22c20a`. `ConcurrentRadixTreeCompressed` already groups removals by compressed node in its `remove_worker_for_hashes` path (introduced by #11277), so no CRTC implementation change was needed. I added one shared regression for both single-threaded and compressed-concurrent indexers covering out-of-order same-edge hashes, duplicates, and descendant preservation.

lib/kv-router/src/indexer/radix_tree.rs
jthomson04: Fixed in `0fef22c20a`. Grouping now starts from one read-only `lookup.get(...).cloned()`. The node mutation completes before stale suffix entries are removed, so there is no remove-then-restore protocol or early-exit hazard.

lib/kv-router/src/indexer/radix_tree.rs
jthomson04: Fixed in `0fef22c20a`. Covered-node truncation is now a separate operation that only takes the earliest edge position; `min_hash` is gone. The existing hash-taking operation remains only for the already-uncovered single-hash fallback.

lib/mocker/src/scheduler/live_boundary.rs
jthomson04: Fixed in `0fef22c20a`. The live marker remap was removed with marker indices. Suppression now directly retains private output wrappers; publication converts surviving wrappers back to the unchanged public `OutputSignal`.

lib/mocker/src/scheduler/live_boundary/tests.rs
jthomson04: Fixed in `0fef22c20a`. The obsolete marker-remap test was removed. Focused live-boundary coverage now verifies that tagged outputs survive suppression correctly and that private metadata is discarded at publication.

lib/mocker/src/replay/offline/runtime_utils.rs
jthomson04: Fixed in `0fef22c20a`. The completion-batch fixture now emits exactly `completed_requests` terminal signals, and the test asserts both the count and terminal shape before checking event ordering.

jthomson04:
Rewrote this PR from current `main` as one focused aggregate-replay optimization.

The new history removes the obsolete marker-index and wrapper scaffolding, keeps release-build generational validation and every resolved output-tracking requirement, and moves all compressed-radix work to an independent PR. Live and disaggregated publication still move the original untracked output vector directly; duplicate UUID rejection remains in both replay modes.

The focused diff is now 22 files, +834/-247 (net +587). This remains above the 700-addition target because the retained release validation and focused regression coverage account for the remaining footprint; I did not delete review-requested correctness coverage merely to reduce the headline count.

Fresh evidence against exact base `038c642b56`:

- full 16-report correctness-parity protocol passed byte-for-byte;
- full mocker, bindings, Clippy, formatting, and diff validation passed, apart from the same two baseline-reproduced KVBM tests;
- two clean Astra matched pairs averaged +4.07% processed-token throughput, with exact request/token totals.

The separate radix PR preserves the read-only lookup, grouped-removal, descendant invalidation, duplicate/error, other-worker coverage, and shared single-threaded/CRTC regression requirements from the resolved threads.

PR #12805, opened 2026-08-07
fix(frontend): classify SGLang preprocessing failures

rmccorm4:
## Summary

- Raise `PreprocessError` when a named `tool_choice` is absent from the request's `tools`.
- Raise `ImportError` when the installed SGLang build lacks the DeepSeek-V4 encoder.
- Add focused unit coverage for both classifications.

## Root cause

Both failures previously used `ValueError`, even though one is invalid client input and the other is a missing server dependency. The broad frontend fallback therefore could not apply the intended error policy reliably.

## Behavior

| Scenario | Result |
|---|---|
| Named `tool_choice` missing from `tools` | Typed preprocessing validation failure; the frontend can return HTTP 400 |
| Missing DeepSeek-V4 SGLang integration | Server dependency failure; remains a sanitized 5xx |

This PR intentionally does not map generic Python `ValueError` exceptions to HTTP 400.

## Validation

- `pytest components/src/dynamo/frontend/tests/test_sglang_processor_unit.py -q`: 200 passed
- `ruff check` on both changed files: passed
- `git diff --check`: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when the DeepSeek-V4 encoder integration is unavailable.
  * Added clearer validation errors when a requested tool function is not defined.
  * Ensured errors are surfaced consistently during both request preprocessing and generation.

* **Tests**
  * Expanded coverage for missing integrations and invalid tool selections.


rmccorm4:
Will follow up more holistically separately

PR #12807, opened 2026-08-07
refactor(llm): preserve typed aggregation errors

rmccorm4:
## Stack

**1 of 4** — base: `main`

Next: [#12808](https://github.com/ai-dynamo/dynamo/pull/12808)

## Summary

- Preserve structured `DynamoError` values when consuming `Annotated<T>`.
- Return typed errors from every shared unary aggregation path instead of flattening them to strings.
- Stop aggregation immediately on a typed error, including when it arrives after partial output.

This layer intentionally does not change HTTP status policy.

## Root cause

Unary aggregators converted annotated backend errors to strings. Once the structured error type was discarded, the HTTP layer could not reliably distinguish invalid client input from runtime failures.

## Validation

- Runtime typed-error regression: 1 passed
- Chat/Completions partial-output typed-error regressions: 2 passed
- `git diff --check`: passed

## Tracking

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling across streaming responses for chat, completions, embeddings, audio, images, videos, and tensor operations.
  - Structured backend errors are now preserved and returned accurately, including when failures occur after partial results.
  - Empty or invalid tensor streams now produce clearer errors, while metadata-only updates are handled safely.
  - Stream aggregation behavior is more consistent across classification and pooling requests.

### DIFF
--- a/lib/llm/src/protocols/openai/stream_aggregator.rs
+++ b/lib/llm/src/protocols/openai/stream_aggregator.rs
@@ -19,7 +19,7 @@ pub trait StreamAggregable: Sized {
 /// Aggregate a stream of [`Annotated<T>`] into a single `T`. The first error
 /// encountered short-circuits further merging and is returned; the remainder
 /// of the stream is dropped.
-pub async fn aggregate_stream<T, S>(stream: S) -> Result<T, String>
+pub async fn aggregate_stream<T, S>(stream: S) -> Result<T, DynamoError>
 where
     T: StreamAggregable,
     S: Stream<Item = Annotated<T>>,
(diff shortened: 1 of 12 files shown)


lib/runtime/src/protocols/annotated.rs
```diff
@@ -33,6 +33,19 @@ pub struct Annotated<R> {
...
+            self.error.clone().unwrap_or_else(|| {
+                DynamoError::msg(
+                    self.comment
+                        .as_ref()
+                        .map(|comments| comments.join(", "))
+                        .unwrap_or_else(|| "unknown error".to_string()),
+                )
+            })
+        })
+    }
```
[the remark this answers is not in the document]
rmccorm4: Fixed in e397bee6a8. Empty comment vectors now fall back to "unknown error", while non-empty comment-only errors retain the canonical ", " separator and unified DynamoError representation. Added regression coverage for both empty and multi-comment cases. Validated with the full dynamo-runtime library suite (533 passed, 2 ignored) and strict all-target Clippy.

lib/runtime/src/protocols/annotated.rs
```diff
@@ -130,24 +144,7 @@ impl<R> Annotated<R> {
...
-                            self.comment
-                                .unwrap_or(vec!["unknown error".to_string()])
-                                .join(", ")
-                        ))?
-                    }
-                }
-                _ => Ok(None),
-            },
-        }
+        self.into_data().map_err(anyhow::Error::new)
```
KrishnanPrash, line 147: nit: `into_result()` semantics now flipped. Main currently looks at data first and then error second. These changes flip by looking at the error field primarily. Can/should we pin this change in behavior in a unit test?

KrishnanPrash (approved):

rmccorm4:
This PR stack likely covers the same fix or majority of fix that https://github.com/ai-dynamo/dynamo/pull/12832 is targeting

lib/llm/src/protocols/openai/stream_aggregator.rs
```diff
@@ -19,7 +19,7 @@ pub trait StreamAggregable: Sized {
 /// Aggregate a stream of [`Annotated<T>`] into a single `T`. The first error
 /// encountered short-circuits further merging and is returned; the remainder
 /// of the stream is dropped.
-pub async fn aggregate_stream<T, S>(stream: S) -> Result<T, String>
+pub async fn aggregate_stream<T, S>(stream: S) -> Result<T, DynamoError>
```
rmccorm4, line 22: @GuanLuo

GuanLuo (approved):

lib/runtime/src/protocols/annotated.rs
rmccorm4: Will follow up with something like this separately, don't want to retrigger CI here
```rust
#[test]
fn test_into_result_error_takes_precedence_over_data() {
    let mut annotated =
        Annotated::from_err(DynamoError::msg("terminal error"));
    annotated.data = Some("partial data".to_string());

    assert_eq!(
        annotated.into_result().unwrap_err().to_string(),
        "terminal error"
    );
}
```

PR #12808, opened 2026-08-07
refactor(frontend): introduce shared HTTP error policy

rmccorm4:
## Summary

- Introduce one protocol-neutral HTTP error policy.
- Classify typed validation, overload, availability, cancellation, explicit HTTP, and unknown failures in one place.
- Preserve safe client messages for 4xx failures while sanitizing runtime and server failures.
- Centralize the metrics category alongside the HTTP classification.

The compatibility model remains in this stack layer so existing endpoint handlers continue compiling independently. #12810 removes it when all handlers switch to the shared policy.

## Classification

| Failure | HTTP result |
|---|---|
| `InvalidArgument` or `Backend(InvalidArgument)` | 400 with the typed validation message |
| Explicit client status | Preserve the 4xx status and message |
| Resource exhausted / queue rejection | Configured overload status |
| Unavailable | Sanitized 503 |
| Cancelled | Sanitized 499 |
| Unknown/runtime failure | Sanitized 500 |

No string matching or hard-coded `max_tokens` limit is introduced.

## Validation

- Shared classifier unit matrix: 4 passed
- `git diff --check`: passed

## Tracking

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of errors returned by backend services.
  * Validation errors are now preserved and presented more accurately.
  * Nested and legacy failures are classified more consistently.
  * Safe client-facing error details are forwarded, while sensitive server and internal information is removed.
  * Overload conditions now return consistent responses across supported protocols.
* **Reliability**
  * Added broader coverage for error classification, sanitization, backend responses, and overload scenarios.

### DIFF
--- a/lib/bindings/python/rust/errors.rs
+++ b/lib/bindings/python/rust/errors.rs
@@ -185,10 +185,7 @@ mod tests {
     #[test]
     fn http_like_errors_are_typed_at_the_python_boundary() {
         for (code, expected_type) in [
-            (
-                415,
-                ErrorType::Backend(BackendError::InvalidArgument),
-            ),
+            (415, ErrorType::Backend(BackendError::InvalidArgument)),
             (503, ErrorType::Backend(BackendError::Unknown)),
         ] {
             let error = build_http_like_error(code, "backend message".to_string());
--- a/lib/bindings/python/rust/errors.rs
+++ b/lib/bindings/python/rust/errors.rs
@@ -11,7 +11,7 @@
 //! corresponding entry to the macro invocation below to keep Python exceptions
 //! in sync.
 
-use dynamo_runtime::error::BackendError;
+use dynamo_runtime::error::{BackendError, DynamoError, ErrorType};
 use pyo3::prelude::*;
 use pyo3::types::PyModule;
 
@@ -143,7 +143,7 @@ define_dynamo_exceptions!(
 /// internal state, file paths, traceback strings, or backend identifiers.
 /// Non-4xx codes (including 5xx) are sanitized downstream — the original
 /// message survives in server logs only.
-pub fn extract_http_like_error(py: Python<'_>, err: &PyErr) -> Option<(u16, String)> {
+fn extract_http_like_error(py: Python<'_>, err: &PyErr) -> Option<(u16, String)> {
     let value = err.value(py);
     let code = value
         .getattr("code")
@@ -157,4 +157,46 @@ pub fn extract_http_like_error(py: Python<'_>, err: &PyErr) -> Option<(u16, Stri
         })?;
     let message = value.getattr("message").ok()?.extract::<String>().ok()?;
     Some((code, message))
+}
+
+pub(crate) fn http_like_error_to_dynamo(py: Python<'_>, err: &PyErr) -> Option<DynamoError> {
+    let (code, message) = extract_http_like_error(py, err)?;
+    Some(build_http_like_error(code, message))
+}
+
+fn build_http_like_error(code: u16, message: String) -> DynamoError {
+    let backend_error = if (400..500).contains(&code) {
+        BackendError::InvalidArgument
+    } else {
+        BackendError::Unknown
+    };
+
+    DynamoError::builder()
+        .error_type(ErrorType::Backend(backend_error))
+        .message(message.clone())
+        .http_error(code, message)
+        .build()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn http_like_errors_are_typed_at_the_python_boundary() {
+        for (code, expected_type) in [
+            (
+                415,
+                ErrorType::Backend(BackendError::InvalidArgument),
+            ),
+            (503, ErrorType::Backend(BackendError::Unknown)),
+        ] {
+            let error = build_http_like_error(code, "backend message".to_string());
+            assert_eq!(error.error_type(), expected_type);
+            assert_eq!(error.message(), "backend message");
+            let http_error = error.http_error().unwrap();
+            assert_eq!(http_error.code(), code);
+            assert_eq!(http_error.message(), "backend message");
+        }
+    }
 }
--- a/lib/llm/src/http/service/anthropic.rs
+++ b/lib/llm/src/http/service/anthropic.rs
@@ -61,7 +61,8 @@ use crate::types::Annotated;
 
 // Re-use helpers from the openai module (sibling under service/)
 use super::error::{
-    ClassifiedHttpError, HttpErrorKind, drain_converted_stream_events, invalid_argument,
+    ClassifiedHttpError, HttpErrorKind, INTERNAL_ERROR_MESSAGE, drain_converted_stream_events,
+    invalid_argument,
 };
 use super::metadata::{attach_x_request_id, extract_metadata_from_http};
 use super::openai::{get_body_limit, get_or_create_request_id};
@@ -909,9 +910,14 @@ fn apply_anthropic_header_routing_overrides(
 
 fn anthropic_error_from_anyhow(
     err: anyhow::Error,
-    fallback: &str,
+    context: &str,
 ) -> (Response, super::metrics::ErrorType) {
-    anthropic_problem(ClassifiedHttpError::from_error(err.as_ref(), fallback))
+    anthropic_problem(classify_anthropic_anyhow(err, context))
+}
+
+fn classify_anthropic_anyhow(err: anyhow::Error, context: &str) -> ClassifiedHttpError {
+    let err = err.context(context.to_string());
+    ClassifiedHttpError::from_error(err.as_ref(), INTERNAL_ERROR_MESSAGE)
 }
 
 fn anthropic_problem(problem: ClassifiedHttpError) -> (Response, super::metrics::ErrorType) {
@@ -1031,6 +1037,23 @@ mod tests {
         .unwrap_err();
 
         assert!(err.to_string().contains("unknown field `agent_context`"));
+    }
+
+    #[test]
+    fn anthropic_unknown_errors_keep_context_out_of_the_public_message() {
+        let problem = classify_anthropic_anyhow(
+            anyhow::anyhow!("worker failed at /srv/backend.py:42"),
+            "Failed to fold messages stream",
+        );
+
+        assert_eq!(problem.status(), StatusCode::INTERNAL_SERVER_ERROR);
+        assert_eq!(problem.message(), INTERNAL_ERROR_MESSAGE);
+        assert!(
+            problem
+                .diagnostic()
+                .contains("Failed to fold messages stream")
+        );
+        assert!(problem.diagnostic().contains("/srv/backend.py:42"));
     }
 
     #[test]
--- a/lib/llm/src/http/service/anthropic.rs
+++ b/lib/llm/src/http/service/anthropic.rs
@@ -61,8 +61,7 @@ use crate::types::Annotated;
 
 // Re-use helpers from the openai module (sibling under service/)
 use super::error::{
-    ClassifiedHttpError, HttpErrorKind, INTERNAL_ERROR_MESSAGE, drain_converted_stream_events,
-    invalid_argument,
+    ClassifiedHttpError, HttpErrorKind, drain_converted_stream_events, invalid_argument,
 };
 use super::metadata::{attach_x_request_id, extract_metadata_from_http};
 use super::openai::{get_body_limit, get_or_create_request_id};
@@ -245,10 +244,9 @@ async fn handler_anthropic_messages(
     )
     .await
     .map_err(|e| {
-        anthropic_problem(ClassifiedHttpError::internal(
-            "Internal server error",
-            format!("Failed to await Anthropic messages task: {e:?}"),
-        ))
+        anthropic_problem(ClassifiedHttpError::internal(format!(
+            "Failed to await Anthropic messages task: {e:?}"
+        )))
         .0
     })?;
 
@@ -917,7 +915,7 @@ fn anthropic_error_from_anyhow(
 
 fn classify_anthropic_anyhow(err: anyhow::Error, context: &str) -> ClassifiedHttpError {
     let err = err.context(context.to_string());
-    ClassifiedHttpError::from_error(err.as_ref(), INTERNAL_ERROR_MESSAGE)
+    ClassifiedHttpError::from_error(err.as_ref())
 }
 
 fn anthropic_problem(problem: ClassifiedHttpError) -> (Response, super::metrics::ErrorType) {
@@ -990,6 +988,7 @@ fn anthropic_error(status: StatusCode, error_type: &str, message: &str) -> Respo
 
 #[cfg(test)]
 mod tests {
+    use super::super::error::INTERNAL_ERROR_MESSAGE;
     use super::*;
     use crate::protocols::common::extensions::parse_nvext;
 
--- a/lib/llm/src/http/service/anthropic.rs
+++ b/lib/llm/src/http/service/anthropic.rs
@@ -42,9 +42,9 @@ use super::{
 use crate::engines::ValidateRequest;
 use crate::protocols::anthropic::stream_converter::AnthropicStreamConverter;
 use crate::protocols::anthropic::types::{
-    AnthropicContentBlock, AnthropicCountTokensRequest, AnthropicCountTokensResponse,
-    AnthropicCreateMessageRequest, AnthropicErrorBody, AnthropicErrorResponse, AnthropicMessage,
-    AnthropicMessageContent, AnthropicTool, SystemContent, chat_completion_to_anthropic_response,
+    AnthropicCountTokensRequest, AnthropicCountTokensResponse, AnthropicCreateMessageRequest,
+    AnthropicErrorBody, AnthropicErrorResponse, AnthropicRequestValidationError, SystemContent,
+    chat_completion_to_anthropic_response, validate_anthropic_messages, validate_anthropic_request,
 };
 use crate::protocols::common::extensions::{
     AGENT_CONTEXT_CONTEXT_KEY, SESSION_AFFINITY_CONTEXT_KEY, agent_context_from_headers,
@@ -60,7 +60,9 @@ use crate::request_template::{RequestTemplate, resolve_request_model};
 use crate::types::Annotated;
 
 // Re-use helpers from the openai module (sibling under service/)
-use super::error::{ClassifiedHttpError, HttpErrorKind, invalid_argument};
+use super::error::{
+    ClassifiedHttpError, HttpErrorKind, invalid_argument, take_converted_stream_events,
+};
 use super::metadata::{attach_x_request_id, extract_metadata_from_http};
 use super::openai::{get_body_limit, get_or_create_request_id};
 
@@ -136,12 +138,6 @@ async fn anthropic_error_middleware(request: Request<Body>, next: Next) -> Respo
 // Handlers
 // ---------------------------------------------------------------------------
 
-#[derive(Debug)]
-enum AnthropicRequestValidationError {
-    InvalidArgument(String),
-    NotImplemented(String),
-}
-
 impl AnthropicRequestValidationError {
     fn status(&self) -> StatusCode {
         match self {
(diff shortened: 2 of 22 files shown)


lib/llm/src/http/service/error.rs
```diff
@@ -32,181 +39,547 @@ pub struct HttpError {
...
+            if let Some(dynamo_error) = error.downcast_ref::<DynamoError>() {
+                return Self::from_dynamo_error(dynamo_error, internal_message, diagnostic);
+            }
+            if let Some(http_error) = error.downcast_ref::<HttpError>() {
+                return Self::from_explicit_http_error(http_error, diagnostic);
+            }
+            current = error.source();
+        }
+
+        Self::internal(internal_message, diagnostic)
```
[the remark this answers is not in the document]
rmccorm4: Fixed in `7f3964f22d`. `HttpProblem::from_error` now preserves the existing typed-error chain precedence, so nested `ResourceExhausted`, `Unavailable`, and `Cancelled` signals remain 529/503/499 even beneath an outer `Unknown` wrapper. Added table-driven regression coverage for all three; the focused classifier suite passes 7/7.

lib/llm/src/http/service/error.rs
```diff
@@ -32,181 +39,547 @@ pub struct HttpError {
...
+        // the untyped comment as a client-safe status/message.
+        if event.data.is_none()
+            && event.event.is_none()
+            && let Some(comments) = event.comment.as_ref()
+            && !comments.is_empty()
+        {
+            return Some(Self::internal(INTERNAL_ERROR_MESSAGE, comments.join(", ")));
+        }
+
+        None
```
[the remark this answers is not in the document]
rmccorm4: Will be handled in #12810

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +44,472 @@ pub struct HttpError {
...
+            // A typed error is authoritative. Its message is data, not another
+            // envelope to inspect: parsing JSON-looking text here could leak an
+            // unknown failure as a client error or override InvalidArgument.
+            if let Some(error) = event.error.as_ref() {
+                return Some(Self::from_dynamo_error(
+                    error,
+                    INTERNAL_ERROR_MESSAGE,
+                    format_error_chain(error),
+                ));
+            }
```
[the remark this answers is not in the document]
rmccorm4: Fixed in ed53c7dca9. Typed Python-boundary `{code,message}` payloads are now decoded only when the `BackendError` category agrees with the HTTP status class. This preserves client 4xx status/message, retains 5xx status while sanitizing the body, rejects mismatched JSON claims, and routes typed annotations through the chain-aware classifier so nested `ResourceExhausted`, `Unavailable`, and `Cancelled` precedence remains unchanged. Added focused regression coverage for these cases.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +50,484 @@ pub struct HttpError {
...
+            // Compatibility for legacy error events that carried a JSON body
+            // in their comment instead of the structured `error` field.
+            if let Ok(payload) = serde_json::from_str::<ErrorPayload>(&diagnostic) {
+                let status = payload
+                    .code
+                    .and_then(|code| StatusCode::from_u16(code).ok())
+                    .unwrap_or(StatusCode::INTERNAL_SERVER_ERROR);
+                let message = payload.message.unwrap_or_else(|| diagnostic.clone());
+                return Some(Self::from_backend_status(status, message, diagnostic));
+            }
```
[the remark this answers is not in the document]
rmccorm4: N/A - internal details

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +50,484 @@ pub struct HttpError {
...
+
+            return Some(Self::internal(INTERNAL_ERROR_MESSAGE, diagnostic));
+        }
+
+        // Legacy backends used a comment-only, otherwise empty annotation as
+        // an error marker. Preserve that signal, but never expose or interpret
+        // the untyped comment as a client-safe status/message.
```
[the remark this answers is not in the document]
rmccorm4: N/A - internal details

lib/llm/src/http/service/error.rs
```diff
@@ -5,8 +5,21 @@ use std::sync::LazyLock;
...
+use super::metrics::ErrorType as MetricErrorType;
+use crate::types::Annotated;
+
+pub(crate) const INTERNAL_ERROR_MESSAGE: &str = "Internal server error";
+
+#[derive(serde::Deserialize)]
+struct BackendHttpPayload {
+    code: u16,
+    message: String,
+}
```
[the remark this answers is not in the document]
rmccorm4: Thanks for flagging this. I audited the supported producers: extract_http_like_error requires both code and message, and both Python mapping paths serialize both fields. A partial object is therefore not a supported typed wire shape. Keeping the typed decoder strict avoids treating arbitrary JSON-looking messages as HTTP policy; the legacy untyped event format remains handled separately. No behavior change here.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +50,484 @@ pub struct HttpError {
...
+                error.downcast_ref::<dynamo_kv_router::scheduling::QueueRejection>()
+            {
+                return Self {
+                    kind: HttpProblemKind::Overloaded,
+                    status: overload_status_code(),
+                    message: rejection.to_string(),
+                    diagnostic,
+                    details: serde_json::to_value(rejection).ok().map(Box::new),
+                };
+            }
```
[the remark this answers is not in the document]
rmccorm4: Keeping the current behavior. QueueRejection message and details are an intentional existing API contract, including policy_class, limit_kind, current, and limit, and integration tests assert those fields. This PR centralizes the existing classification rather than changing that public response. If those fields should be treated as sensitive, that needs a separate API and security change with migration and test updates.

lib/llm/src/http/service/error.rs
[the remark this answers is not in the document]
rmccorm4: Keeping the explicit legacy event:error 4xx behavior for compatibility. The existing reader forwards client-safe 4xx messages, while 5xx and 499 bodies are sanitized, and the stream regression covers an event:error comment with code 400. deny_unknown_fields would also reject existing legacy envelopes containing additional fields such as type, contrary to the tolerant-reader N-1 requirement. The comment-only rule below applies to frames without the event:error marker, so it does not contradict this compatibility path.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +50,484 @@ pub struct HttpError {
...
+        // Preserve the existing handler precedence for typed failures anywhere
+        // in the chain. In particular, an untyped wrapper must not hide an
+        // overload, unavailable, or cancellation signal.
+        for error_type in [
+            DynamoErrorType::ResourceExhausted,
+            DynamoErrorType::Unavailable,
+            DynamoErrorType::InvalidArgument,
+            DynamoErrorType::Backend(BackendError::InvalidArgument),
+            DynamoErrorType::Cancelled,
+        ] {
```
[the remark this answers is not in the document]
rmccorm4: Fixed in [6777a118f7](https://github.com/ai-dynamo/dynamo/commit/6777a118f7d0e181136f0015447967ef4fde720c). The shared nested-error selector now includes Backend(Cancelled), and the operational-error regression matrix verifies that wrapped backend cancellations remain 499 rather than becoming internal errors.

lib/llm/src/http/service/error.rs
```diff
@@ -189,57 +557,171 @@ mod tests {
...
+            &DynamoError::builder()
+                .error_type(DynamoErrorType::Unavailable)
+                .message("down")
+                .build(),
+            INTERNAL_ERROR_MESSAGE,
+            "down".to_string(),
         );
+        assert_eq!(unavailable.status(), StatusCode::SERVICE_UNAVAILABLE);
+        assert_eq!(unavailable.kind(), HttpProblemKind::Unavailable);
     }
```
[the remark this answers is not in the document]
rmccorm4: Fixed in [2ac175a96a](https://github.com/ai-dynamo/dynamo/commit/2ac175a96a9cc344c266bc9bc37c6dd5e9db2421). The test now compares the classified status directly with overload_status_code(), so it honors the configured environment value instead of assuming 529.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +50,484 @@ pub struct HttpError {
...
+        if event.data.is_none()
+            && event.event.is_none()
+            && let Some(comments) = event.comment.as_ref()
+            && !comments.is_empty()
+        {
+            return Some(Self::internal(INTERNAL_ERROR_MESSAGE, comments.join(", ")));
         }
+
+        None
+    }
```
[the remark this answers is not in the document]
rmccorm4: Fixed in [df5c5f1386](https://github.com/ai-dynamo/dynamo/commit/df5c5f1386f986174b70abbd91f15476f4433ca6). The shared classifier now preserves both legacy shapes through one bounded compatibility decoder: status-bearing data payloads and untagged JSON comments. Unmarked inputs require an explicit code of at least 400, typed errors remain authoritative, and focused coverage exercises both shapes plus the non-error 200 case.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+    let status = StatusCode::from_u16(payload.code).ok()?;
+
+    match error.error_type() {
+        DynamoErrorType::Backend(BackendError::InvalidArgument) if status.is_client_error() => {}
+        DynamoErrorType::Backend(BackendError::Unknown) if status.is_server_error() => {}
+        _ => return None,
+    }
+
+    Some((status, payload.message))
+}
```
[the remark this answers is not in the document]
rmccorm4: > Today's Python boundary (lib/bindings/python/rust/engine.rs:365-380) always tags these envelopes as Backend(...), so the narrowing matches the current producer.

Will not fix in this PR

KrishnanPrash (approved):

KrishnanPrash:
Thanks for breaking this apart + adding the test on precedence order from #12807.


Not essential. but one request for future reference:
Could you add an example what triggers `ValidationScope::Outermost` (`from_annotated()` path) vs. `ValidationScope::Chain` (`from_error()` path)? I want a better understanding of when to propagate what part of the error.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+
+    pub(crate) fn metric_type_for_status(status: StatusCode) -> MetricErrorType {
+        metric_type_for_kind(kind_for_status(status))
+    }
+}
+
+/// Decode the deliberate Python-boundary `{code, message}` compatibility
+/// envelope without treating arbitrary typed error messages as HTTP policy.
+fn backend_http_payload(error: &DynamoError) -> Option<(StatusCode, String)> {
+    let payload = serde_json::from_str::<BackendHttpPayload>(error.message()).ok()?;
```
GuanLuo: This function is translating a backend error (captured only in `DynamoError.message()` back to direct `DynamoError`. This mean that is that before reaching this point, we have translation of "raw" backend error to `DynamoError`, why don't we move this code to there, so it already passes `DynamoError` with proper `DynamoErrorType::Backend(xxx)` to HTTP service.

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            kind: HttpProblemKind::Internal,
+            status: StatusCode::INTERNAL_SERVER_ERROR,
+            message: message.into(),
+            diagnostic: diagnostic.into(),
+            details: None,
+        }
+    }
+
+    fn from_dynamo_error(error: &DynamoError, internal_message: &str, diagnostic: String) -> Self {
+        if let Some((status, message)) = backend_http_payload(error) {
```
GuanLuo: See comment on `backend_http_payload`, then we just add new `DynamoErrorType::Backend` arm in match clause below

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            Self::Internal => StatusCode::INTERNAL_SERVER_ERROR,
+            Self::PreserveServerError(code) => code,
         }
     }
 
-    /// Anthropic `error.type` for this category. For `PreserveServerError`
-    /// the inner status is consulted so a backend 503/529 is reported as
-    /// `overloaded_error` (matching the Anthropic spec) rather than the
-    /// generic `api_error`.
     pub fn anthropic_type(self) -> &'static str {
```
GuanLuo: For cleanup, can we put these to `anthropic.rs` etc (I think we already have the breakdown) and import the function here as needed?

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            Self::Unavailable => f.write_str("Service temporarily unavailable"),
+            Self::Internal | Self::PreserveServerError(_) => f.write_str(INTERNAL_ERROR_MESSAGE),
+        }
+    }
+}
+
+/// Protocol-neutral error categories used for metrics and protocol rendering.
+#[allow(dead_code)] // Used by the endpoint-integration PR later in this stack.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum HttpProblemKind {
```
GuanLuo: `Problem` is strange naming?

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+    status: StatusCode,
+    message: String,
+    diagnostic: String,
+    details: Option<Box<Value>>,
+}
+
+/// Construct a typed invalid-argument error for validation performed at an
+/// HTTP protocol adapter boundary.
+#[allow(dead_code)] // Used by the protocol-validation PR later in this stack.
+pub(crate) fn invalid_argument(message: impl Into<String>) -> DynamoError {
```
GuanLuo, line 91: move to `error.rs`? Can it be macro, it is just shorter wrapper for DynamoError builder

rmccorm4:
/ok to test [df5c5f1](https://github.com/ai-dynamo/dynamo/pull/12808/commits/df5c5f1386f986174b70abbd91f15476f4433ca6)

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+    ] {
+        if let Some(error) = find_dynamo_error_in_chain(err, error_type) {
+            return Some(error);
+        }
+    }
+
+    None
+}
+
+fn find_outermost_dynamo_error_in_chain<'a>(
```
GuanLuo: This is basically `find_dynamo_error_in_chain` with all error type accepted, can you make `find_dynamo_error_in_chain` to accept a list of filtered error type, then if empty list, any dynamo error is returned. This simplifies other call sites of `find_dynamo_error_in_chain` as well

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            }
+        }
+    }
+
+    for error_type in [
+        DynamoErrorType::Cancelled,
+        DynamoErrorType::Backend(BackendError::Cancelled),
+    ] {
+        if let Some(error) = find_dynamo_error_in_chain(err, error_type) {
+            return Some(error);
```
GuanLuo: Many of the code in this function is returning matched error as is, with the type list, you can condense this function

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            if let Some(problem) = Self::from_legacy_http_json(diagnostic.clone()) {
+                return Some(problem);
+            }
+
+            return Some(Self::internal(INTERNAL_ERROR_MESSAGE, diagnostic));
+        }
+
+        // Compatibility with v1.3 workers during v1.4 rolling upgrades: older
+        // streams may carry HTTP errors in data or an untagged JSON comment.
+        // TODO(v1.5): Remove when v1.3 leaves the N-1 compatibility window.
```
GuanLuo: For future tracking, do we have deprecation dashboard on what are the things we want to move out?

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+    fn from_error_with_validation_scope(
+        err: &(dyn std::error::Error + 'static),
+        internal_message: &str,
+        validation_scope: ValidationScope,
+    ) -> Self {
+        let diagnostic = format_error_chain(err);
+        let mut current = Some(err);
+        while let Some(error) = current {
+            if let Some(rejection) =
+                error.downcast_ref::<dynamo_kv_router::scheduling::QueueRejection>()
```
GuanLuo, line 120: Not in `DynamoErrorType`?

lib/llm/src/http/service/error.rs
```diff
@@ -37,129 +56,575 @@ pub struct HttpError {
...
+            // expose an unstructured comment as a client-safe message.
+            if event.data.is_none() && event.event.is_none() {
+                return Some(Self::internal(INTERNAL_ERROR_MESSAGE, diagnostic));
+            }
+        }
+
+        None
+    }
+
+    fn from_legacy_http_json(diagnostic: String) -> Option<Self> {
```
GuanLuo: Looks like we have different kind of errors pumping out to HTTP response, do you have a list of error types and where they originated from? I think within the frontend (also counting the TCP response path), all Dynamo raise error should be able to convert to DynamoError when originally constructed, without breaking things publicly. The rest are three party library errors, and could we catch and convert them to DynamoError when they first arise? If we want to be clear on error type (invalid argument)and source (from backend) for third party stuff, we could add new type similar to backend `DynamoErrorType::SourceToBeChecked(XXX)`

lib/llm/src/http/service/error.rs
rmccorm4: Fixed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). HTTP-like Python exceptions are now classified at the Python boundary by the shared `http_like_error_to_dynamo` adapter. It constructs a `DynamoError` with the appropriate `Backend(...)` category plus typed HTTP status/message metadata, and the HTTP classifier consumes that metadata directly instead of parsing JSON from `DynamoError.message()`.

lib/llm/src/http/service/error.rs
rmccorm4: Addressed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). The origin now assigns `Backend(InvalidArgument)` for 4xx and `Backend(Unknown)` otherwise, with typed HTTP metadata preserving the exact status. The classifier validates and consumes that metadata before the normal `Backend(...)` match, so statuses such as 415 are retained without adding an enum variant for every HTTP status.

lib/llm/src/http/service/error.rs
rmccorm4: Fixed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). Anthropic error-type rendering now lives in `anthropic.rs`, and the OpenAI stream error-type rendering lives in `disconnect.rs`. The shared error module now owns only protocol-neutral status, sanitization, and classification policy.

lib/llm/src/http/service/error.rs
rmccorm4: Fixed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). Renamed `HttpProblem` to `ClassifiedHttpError` and `HttpProblemKind` to `HttpErrorKind`.

lib/llm/src/http/service/error.rs
rmccorm4: I am keeping this as a typed function in `http/service/error.rs`. It is a semantic constructor used by the protocol adapter; a macro would not materially shorten the call sites and would add macro export/hygiene concerns. No behavior change needed here.

lib/llm/src/http/service/error.rs
rmccorm4: Fixed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). `find_dynamo_error_in_chain` now accepts an ordered slice of accepted error types. An empty slice returns the outermost `DynamoError`, which removes the duplicate traversal helper.

lib/llm/src/http/service/error.rs
rmccorm4: Fixed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). The selection logic is condensed into ordered helper calls while preserving the explicit precedence: `ResourceExhausted`, `Unavailable`, validation according to the selected scope, then cancellation.

lib/llm/src/http/service/error.rs
rmccorm4: We do not plan to track these through a public deprecation dashboard. These error translations are internal Dynamo implementation details rather than a public compatibility surface. The compatibility-window comments were removed in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71).

lib/llm/src/http/service/error.rs
rmccorm4: `QueueRejection` remains a direct typed special case because it carries structured scheduler details that are serialized into the response. Adding only a `DynamoErrorType` variant would lose those details; a complete conversion would first need structured details on `DynamoError`. I am keeping that broader model change out of this PR.

lib/llm/src/http/service/error.rs
rmccorm4: Implemented the concrete origin-normalization path relevant to this PR in [d99f9ff2ca](https://github.com/ai-dynamo/dynamo/commit/d99f9ff2ca0b32808d226082dff3ca9c09e30e71). HTTP-like Python exceptions are converted to typed `DynamoError` values at the Python boundary, including their exact HTTP metadata, so the HTTP service no longer reconstructs policy from a JSON message. A frontend-wide migration of every third-party error origin is broader follow-up work; this PR retains the safe fallback classifier for those paths.

lib/bindings/python/rust/errors.rs
```diff
@@ -158,3 +158,42 @@ pub fn extract_http_like_error(py: Python<'_>, err: &PyErr) -> Option<(u16, Stri
...
+    } else {
+        BackendError::Unknown
+    };
+
+    DynamoError::builder()
+        .error_type(ErrorType::Backend(backend_error))
+        .message(message.clone())
+        .http_error(code, message)
+        .build()
+}
```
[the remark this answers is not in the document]
rmccorm4: We do not expect mixed frontend/worker versions for this release, so this PR intentionally keeps typed `DynamoError::http_error` metadata as the only supported representation and does not add a JSON-message compatibility reader. We can revisit rolling-version compatibility when that deployment mode is supported.

lib/llm/src/http/service/anthropic.rs
```diff
@@ -1024,71 +980,67 @@ fn apply_anthropic_header_routing_overrides(
...
-    } else {
-        tracing::debug!(status = %status, "Anthropic {err}: {details}");
-    }
+fn anthropic_error_from_anyhow(
+    err: anyhow::Error,
+    fallback: &str,
+) -> (Response, super::metrics::ErrorType) {
+    anthropic_problem(ClassifiedHttpError::from_error(err.as_ref(), fallback))
+}
+
```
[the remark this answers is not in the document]
rmccorm4: Fixed centrally in [0b338e32ff](https://github.com/ai-dynamo/dynamo/commit/0b338e32ffb30ec7b8118e74760ca73b7a92a870). `ClassifiedHttpError::internal` now always uses `INTERNAL_ERROR_MESSAGE`, and `from_error` no longer accepts a caller-selected public 500 message. Handler-stage context remains in the diagnostic chain and server logs; the Anthropic regression coverage was added in [a36d87bae9](https://github.com/ai-dynamo/dynamo/commit/a36d87bae91beaf05725cb458af37fcb105d8bcf).

lib/llm/src/http/service/error.rs
```diff
@@ -47,129 +82,416 @@ pub(crate) fn invalid_argument(message: impl Into<String>) -> DynamoError {
...
         }
+
+        if event.data.is_none()
+            && event.event.is_none()
+            && let Some(comments) = event.comment.as_ref()
+            && !comments.is_empty()
+        {
+            let diagnostic = comments.join(", ");
+            return Some(Self::internal(INTERNAL_ERROR_MESSAGE, diagnostic));
+        }
```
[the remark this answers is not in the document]
rmccorm4: Fixed in [a36d87bae9](https://github.com/ai-dynamo/dynamo/commit/a36d87bae91beaf05725cb458af37fcb105d8bcf). `from_annotated` now classifies only explicit error events; the untagged comment-only shape is limited to `from_precommit_annotated`, with a regression test confirming ordinary mid-stream notes are not terminal.

akshatha-k (approved):
doc changes lgtm

rmccorm4:
@biswapanda to take over this work, feel free to start fresh with a new PR.

rmccorm4:
Closing in favor of @biswapanda 's PRs related to https://github.com/ai-dynamo/dynamo/issues/14354

PR #12810, opened 2026-08-07
fix(frontend): unify streamed HTTP error handling

rmccorm4:
## Stack

**2 of 2** — base: [#12808](https://github.com/ai-dynamo/dynamo/pull/12808)

Previous: [#12808](https://github.com/ai-dynamo/dynamo/pull/12808)

## Summary

- Route OpenAI, Responses, Anthropic, Generate, Embeddings, and shared HTTP surfaces through the common classifier.
- Remove endpoint-specific and non-streaming first-frame error workarounds.
- Return classified errors before response headers are committed.
- After SSE data commits HTTP 200, emit a protocol-correct inline client or sanitized server error and terminate immediately.
- Record late stream failures consistently in request metrics.

## Streaming rule

| Error timing | Result |
|---|---|
| Before response headers | Return the classified 4xx or sanitized 5xx |
| After SSE data commits HTTP 200 | Keep 200, emit the protocol-specific inline error, terminate the stream |

## Fixed regressions

- Completions `max_tokens: 2147483647`: 400 for unary and pre-header streaming.
- Different backend validation failures such as unsupported `logprobs`: 400.
- Typed validation after partial unary output: 400.
- Unknown runtime failures: sanitized 500.
- Late typed errors: inline protocol client error.
- Late unknown errors: sanitized inline server error.

## Validation

- `frontend_validation_http`: 3 passed
- Shared classifier tests: 10 passed
- Disconnect/stream error tests: 8 passed
- `git diff --check`: passed

## Tracking

### DIFF
--- a/lib/llm/src/http/service/anthropic.rs
+++ b/lib/llm/src/http/service/anthropic.rs
@@ -30,7 +30,9 @@ use tracing::Instrument;
 
 use super::{
     RouteDoc,
-    disconnect::{ConnectionHandle, create_connection_monitor, monitor_for_disconnects},
+    disconnect::{
+        ConnectionHandle, create_connection_monitor, monitor_for_disconnects_with_rendered_errors,
+    },
     metrics::{
         CancellationLabels, Endpoint, ErrorType, InflightGuard,
         process_chat_response_and_observe_metrics as process_response_and_observe_metrics,
@@ -57,7 +59,7 @@ use crate::request_template::{RequestTemplate, resolve_request_model};
 use crate::types::Annotated;
 
 // Re-use helpers from the openai module (sibling under service/)
-use super::error::{SanitizedError, invalid_argument};
+use super::error::{ClassifiedHttpError, HttpErrorKind, invalid_argument};
 use super::metadata::{attach_x_request_id, extract_metadata_from_http};
 use super::openai::{get_body_limit, get_or_create_request_id};
 
@@ -330,10 +332,11 @@ async fn handler_anthropic_messages(
     )
     .await
     .map_err(|e| {
-        anthropic_sanitized_error_with_details(
-            SanitizedError::Internal,
+        anthropic_problem(ClassifiedHttpError::internal(
+            "Internal server error",
             format!("Failed to await Anthropic messages task: {e:?}"),
-        )
+        ))
+        .0
     })?;
 
     connection_handle.disarm();
@@ -351,7 +354,6 @@ async fn anthropic_messages(
     mut inflight_guard: InflightGuard,
 ) -> Result<Response, Response> {
     let streaming = request.stream;
-    let request_id = request.id().to_string();
 
     // Apply template defaults before capturing model (must happen first so
     // engine lookup and metrics use the resolved model name).
@@ -413,7 +415,6 @@ async fn anthropic_messages(
                 )
             }
         })?;
-
     let (orig_request, context) = request.into_parts();
     let model_for_resp = orig_request.model.clone();
 
@@ -436,16 +437,8 @@ async fn anthropic_messages(
     // Convert Anthropic request -> UnifiedRequest -> Chat Completion request
     let unified_request: UnifiedRequest = orig_request.try_into().map_err(|e: anyhow::Error| {
         inflight_guard.mark_error(ErrorType::Validation);
-        tracing::error!(
-            request_id,
-            error = %e,
-            "Failed to convert AnthropicCreateMessageRequest to UnifiedRequest",
-        );
-        anthropic_error(
-            StatusCode::BAD_REQUEST,
-            "invalid_request_error",
-            &format!("Failed to convert request: {}", e),
-        )
+        let error = invalid_argument(format!("Failed to convert request: {e}"));
+        anthropic_error_from_anyhow(error.into(), "Failed to convert Anthropic request").0
     })?;
 
     // Extract the API context before consuming the UnifiedRequest — this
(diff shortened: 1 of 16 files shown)


lib/llm/src/http/service/openai.rs
KrishnanPrash, line 1: **`VALIDATION_PREFIX` (line 92) outlived its reader.**

This PR deletes `classify_error_for_metrics`, the only thing that read the prefix, but keeps the 31 places that write it. Every validation message still gets `"Validation: "` prepended for nobody. Worth dropping the writes in the same PR that removes the reader.

_(Filed at file level because line 92 is outside this PR's diff.)_

KrishnanPrash, line 1: **Stale comment at line 1269.** It still explains that only a `VALIDATION_PREFIX`-prefixed 400 maps to `Validation`, but this PR deletes `classify_error_for_metrics`, so that rule no longer exists here.

_(Filed at file level because line 1269 is outside this PR's diff.)_

KrishnanPrash, line 1: **Stale comment at line 1542.** Same wording as line 1269 and same problem: it describes the `VALIDATION_PREFIX` metric rule that this PR removes.

_(Filed at file level because line 1542 is outside this PR's diff.)_

lib/llm/src/http/service/anthropic.rs
```diff
@@ -934,71 +886,67 @@ fn apply_anthropic_header_routing_overrides(
...
+        tracing::debug!(status = %problem.status(), "Anthropic error: {}", problem.diagnostic());
+    }
+    let error_type = match problem.kind() {
+        HttpProblemKind::Validation => "invalid_request_error",
+        HttpProblemKind::Authentication => "authentication_error",
+        HttpProblemKind::Permission => "permission_error",
+        HttpProblemKind::NotFound => "not_found_error",
+        HttpProblemKind::RateLimit => "rate_limit_error",
+        HttpProblemKind::Cancelled => "request_cancelled",
+        HttpProblemKind::Overloaded | HttpProblemKind::Unavailable => "overloaded_error",
```
KrishnanPrash: `SanitizedError` took its two tests with it when it was deleted, which is fine, except the `503`/`529` to `overloaded_error` behavior moved here rather than going away. Nothing under `lib/llm/tests/` covers it now. Worth a small case in `frontend_validation_http.rs`, since this is the string Anthropic clients branch on.

lib/llm/src/http/service/openai.rs
```diff
@@ -3182,19 +2828,28 @@ async fn responses(
...
+                            message: problem.message().to_string(),
+                        },
+                        &mut events,
+                    );
+                    for event in events.drain(..) {
+                        yield event.map_err(axum::Error::new);
+                    }
+                    // response.failed is terminal. Return the classification to
+                    // the monitor for metrics without waiting for backend EOF.
+                    yield Err(axum::Error::new(problem));
```
KrishnanPrash, line 2867: The classification and the `response.failed` payload here are a clear improvement. Just checking the `return`: the old code set `saw_error` and kept consuming, so the metrics tap saw every later chunk. If any backend emits a usage chunk after an error frame, those counts are now dropped. Devin asked the same thing about the unary aggregators in #12807 and it is still open, so one answer would close both: is an error frame always terminal for the backends you support?

KrishnanPrash (approved):

lib/llm/src/http/service/openai.rs
```diff
@@ -2210,45 +1920,26 @@ where
...
 /// process restart.
 // FIXME: unify env-var initialization with the rest of `env_llm::*` once that
 // module gets a standard reader.
-fn pre_commit_error_peek_timeout() -> Option<std::time::Duration> {
+pub(super) fn pre_commit_error_peek_timeout() -> Option<std::time::Duration> {
+    const DEFAULT_MS: u64 = 500;
     match std::env::var(env_llm::DYN_HTTP_PRE_COMMIT_ERROR_PEEK_MS)
         .ok()
         .and_then(|s| s.parse::<u64>().ok())
+        .unwrap_or(DEFAULT_MS)
```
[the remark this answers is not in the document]
rmccorm4: Addressed in 386d1c7262 and documented in 2c04d6e68f. Unset, invalid, or 0 now returns None; only a positive configured value enables the pre-commit peek. The affected tests opt in explicitly, and the Frontend Configuration Reference documents the default and per-request behavior. Resolving.

lib/llm/src/http/service/anthropic.rs
```diff
@@ -481,10 +502,22 @@ async fn anthropic_messages(
...
+                                &mut events,
+                            );
+                            for event in events.drain(..) {
+                                yield event.map_err(axum::Error::new);
                             }
+                            // The native error event is terminal. Return the
+                            // classification to the monitor for metrics without
+                            // waiting for the backend transport to close.
+                            yield Err(axum::Error::new(problem));
+                            return;
```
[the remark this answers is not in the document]
rmccorm4: Addressed in 8e0f5c39fb. The shared terminal-error branch now calls `context.kill()` before disarming the connection handle, so abandoned backend work is cancelled for Chat Completions, Anthropic Messages, and Responses. `test_simulate_worker_kill_emits_structured_error_and_done` now uses a kill-tracking context and asserts that the backend context was killed.

lib/llm/src/http/service/anthropic.rs
```diff
@@ -481,10 +454,22 @@ async fn anthropic_messages(
...
+                        }
```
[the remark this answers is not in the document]
rmccorm4: Addressed in 8e0f5c39fb. The shared terminal-error branch now calls `context.kill()` before disarming the connection handle, so abandoned backend work is cancelled for Chat Completions, Anthropic Messages, and Responses. `test_simulate_worker_kill_emits_structured_error_and_done` now uses a kill-tracking context and asserts that the backend context was killed.

lib/llm/src/http/service/openai.rs
rmccorm4: Addressed in dd89bc5741. `VALIDATION_PREFIX` and all remaining writers were removed, and the affected assertions now expect the unprefixed client messages.

rmccorm4: Addressed in dd89bc5741. The stale prefix-specific metrics explanation was removed now that error classification no longer reads `VALIDATION_PREFIX`.

rmccorm4: Addressed in dd89bc5741. The duplicate stale prefix-specific metrics explanation was removed now that error classification no longer reads `VALIDATION_PREFIX`.

lib/llm/src/http/service/anthropic.rs
rmccorm4: Addressed in 3063a53215. `anthropic_overload_statuses_use_overloaded_error` covers both 503 and 529 mapping to `overloaded_error`, including client-message sanitization. The same commit also adds Responses overload integration coverage for both statuses.

PR #12822, opened 2026-08-07
feat: add GMS v1 checkpoint lifecycle control

hhzhang16:
## Overview:

Add a process-wide checkpoint lifecycle shared by the V1 weights and KV-cache domains. Fence new sessions once both domains are checkpoint-safe, and expose `prepare`, `state`, `abort`, and `complete` control RPCs for resolving the checkpoint lifecycle.

## Details:

- Add a process-wide checkpoint lifecycle shared by the V1 weights and KV-cache GMS servers.
- Add typed `prepare`, `state`, `abort`, and `complete` control RPCs.
- Require committed, quiescent weights and an empty, uncommitted KV cache before preparing a checkpoint.
- Fence new client admission across both domains while checkpoint state is unresolved.
- Use an opaque token to identify and resolve the active checkpoint attempt.
- Make `prepare` idempotent and make `abort` and `complete` retries safe for the active token.

Tests:

- `lib/gpu_memory_service/tests/test_v1_checkpoint.py::test_prepare_fences_both_domains_and_abort_is_retry_safe`
- `lib/gpu_memory_service/tests/test_v1_checkpoint.py::test_complete_is_retry_safe`
- `lib/gpu_memory_service/tests/test_v1_checkpoint.py::test_handshake_racing_prepare_cannot_cross_the_fence`

## Related Issues

- Closes #XXXX

### DIFF
--- a/lib/gpu_memory_service/v1/checkpoint.py
+++ b/lib/gpu_memory_service/v1/checkpoint.py
@@ -1,0 +1,257 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Process-wide checkpoint lifecycle and control client for GMS V1."""
+
+from __future__ import annotations
+
+import hashlib
+import logging
+import os
+import socket
+import threading
+from collections.abc import Mapping
+from typing import Protocol
+from uuid import uuid4
+
+from gpu_memory_service.v1.protocol import (
+    AbortCheckpointRequest,
+    CheckpointControlRequest,
+    CheckpointDomainState,
+    CheckpointStateResponse,
+    CompleteRestoreRequest,
+    ErrorResponse,
+    GetCheckpointStateRequest,
+    PrepareCheckpointRequest,
+    receive_message,
+    send_message,
+)
+from gpu_memory_service.v1.server import SessionSnapshot
+
+logger = logging.getLogger(__name__)
+
+_SERVING = "serving"
+_CHECKPOINT_READY = "checkpoint_ready"
+_WEIGHTS_DOMAIN = "weights"
+_KV_CACHE_DOMAIN = "kv_cache"
+
+
+class _CheckpointDomainManager(Protocol):
+    @property
+    def identity(self) -> tuple[str, str]:
+        ...
+
+    def session_snapshot(self) -> SessionSnapshot:
+        ...
+
+    def allocation_snapshot(self) -> tuple[tuple[str, int], ...]:
+        ...
+
+
+def _allocation_digest(allocations: tuple[tuple[str, int], ...]) -> str:
+    digest = hashlib.sha256()
+    for allocation_id, aligned_size in allocations:
+        encoded_id = allocation_id.encode("utf-8")
+        digest.update(len(encoded_id).to_bytes(8, "big"))
+        digest.update(encoded_id)
+        digest.update(aligned_size.to_bytes(8, "big"))
+    return digest.hexdigest()
+
+
+class GMSCheckpointLifecycle:
+    """Fence both V1 domains while an external controller snapshots the owner."""
+
+    def __init__(self) -> None:
+        self.condition = threading.Condition(threading.RLock())
+        self._managers: Mapping[str, _CheckpointDomainManager] | None = None
+        self._state = _SERVING
+        self._generation = 0
+        self._token: str | None = None
+        self._domains: tuple[CheckpointDomainState, ...] = ()
+        self._last_resolution: tuple[str, str] | None = None
+
+    def bind_domains(self, managers: Mapping[str, _CheckpointDomainManager]) -> None:
+        if set(managers) != {_WEIGHTS_DOMAIN, _KV_CACHE_DOMAIN}:
+            raise ValueError("checkpoint lifecycle requires weights and kv_cache")
+        with self.condition:
+            if self._managers is not None:
+                raise RuntimeError("checkpoint lifecycle domains are already bound")
+            self._managers = dict(managers)
+
+    def admission_allowed(self) -> bool:
+        with self.condition:
+            return self._state == _SERVING
+
+    def handle(self, request: CheckpointControlRequest) -> CheckpointStateResponse:
+        with self.condition:
+            if isinstance(request, PrepareCheckpointRequest):
+                return self._prepare()
+            if isinstance(request, AbortCheckpointRequest):
+                return self._resolve(request.token, "abort")
+            if isinstance(request, CompleteRestoreRequest):
+                return self._resolve(request.token, "complete")
+            if isinstance(request, GetCheckpointStateRequest):
+                return self._response()
+            raise RuntimeError(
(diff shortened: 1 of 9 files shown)


lib/gpu_memory_service/v1/protocol.py
```diff
@@ -66,6 +90,23 @@ class ExportResponse(msgspec.Struct, tag="export_response", forbid_unknown_field
     pass
 
 
+class CheckpointDomainState(msgspec.Struct, frozen=True, forbid_unknown_fields=True):
```
mohammedabdulwahhab: What is the motivation of capturing these fields? What does this assert that the token check doesn't?

mohammedabdulwahhab: I think we can simplify this/if not remove it altogether.

lib/gpu_memory_service/v1/protocol.py
```diff
@@ -66,6 +90,23 @@ class ExportResponse(msgspec.Struct, tag="export_response", forbid_unknown_field
...
+    gpu_uuid: str
+    allocation_count: int
+    allocation_digest: str
+
+
+class CheckpointStateResponse(
+    msgspec.Struct, tag="checkpoint_state_response", forbid_unknown_fields=True
+):
+    state: str
+    generation: int
```
mohammedabdulwahhab: Same thing here.  Doesn't token implicitly capture generation?

lib/gpu_memory_service/v1/checkpoint.py
```diff
@@ -0,0 +1,257 @@
...
+                return self._resolve(request.token, "abort")
+            if isinstance(request, CompleteRestoreRequest):
+                return self._resolve(request.token, "complete")
+            if isinstance(request, GetCheckpointStateRequest):
+                return self._response()
+            raise RuntimeError(
+                f"unsupported checkpoint request {type(request).__name__}"
+            )
+
+    def _prepare(self) -> CheckpointStateResponse:
```
mohammedabdulwahhab, line 90: A high level comment about what this asserts/checks for would be helpful.

lib/gpu_memory_service/v1/checkpoint.py
```diff
@@ -0,0 +1,257 @@
...
+        ...
+
+    def session_snapshot(self) -> SessionSnapshot:
+        ...
+
+    def allocation_snapshot(self) -> tuple[tuple[str, int], ...]:
+        ...
+
+
+def _allocation_digest(allocations: tuple[tuple[str, int], ...]) -> str:
```
mohammedabdulwahhab: If we are fencing properly, why are we worried about the allocations changing during checkpoint. Isn't this redundant.

mohammedabdulwahhab (commented):
Overall looks good. I think we're capturing a lot as part of the fingerprint (nonce, UUID, generation, allocation) etc. that might already be encoded as part of the token which we might want to simplify.

lib/gpu_memory_service/v1/protocol.py
hhzhang16: The token is a random opaque UUID, it doesn't capture/encode generation, server identity, GPU identity, or allocations

lib/gpu_memory_service/v1/checkpoint.py
```diff
@@ -0,0 +1,257 @@
...
+        self._token = str(uuid4())
+        self._domains = (
+            self._domain_state(_WEIGHTS_DOMAIN, weights, weight_allocations),
+            self._domain_state(_KV_CACHE_DOMAIN, kv_cache, kv_allocations),
+        )
+        self._state = _CHECKPOINT_READY
+        self._last_resolution = None
+        self.condition.notify_all()
+        logger.info("GMS checkpoint generation %d is ready", self._generation)
+        return self._response()
```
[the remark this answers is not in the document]
hhzhang16: No, reopening admission while a snapshot is unresolved is unsafe, so it's better to use state() followed by abort() or complete()

lib/gpu_memory_service/v1/checkpoint.py
hhzhang16: Will do

lib/gpu_memory_service/v1/protocol.py
hhzhang16: Actually yeah the token doesn’t specifically encode generation, but one active token does implicitly identify the current checkpoint attempt

lib/gpu_memory_service/v1/checkpoint.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @staticmethod
+    def _require_quiesced(name: str, sessions: SessionSnapshot) -> None:
+        if (
+            sessions.rw_sessions
+            or sessions.ro_sessions
+            or sessions.waiting_writers
+            or sessions.writer_reserved
+        ):
+            raise RuntimeError(f"{name} has active or waiting sessions")
```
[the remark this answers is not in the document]
hhzhang16: The intended contract is fail-stop: `prepare` runs only after engine sleep and quiescence, and a handshake racing the fence is rejected rather than queued to run after checkpointing begins. This is now documented and covered by `test_handshake_racing_prepare_cannot_cross_the_fence`

lib/gpu_memory_service/v1/checkpoint.py
```diff
@@ -0,0 +1,257 @@
...
+        self._domains: tuple[CheckpointDomainState, ...] = ()
+        self._last_resolution: tuple[str, str] | None = None
+
+    def bind_domains(self, managers: Mapping[str, _CheckpointDomainManager]) -> None:
+        if set(managers) != {_WEIGHTS_DOMAIN, _KV_CACHE_DOMAIN}:
+            raise ValueError("checkpoint lifecycle requires weights and kv_cache")
+        with self.condition:
+            if self._managers is not None:
+                raise RuntimeError("checkpoint lifecycle domains are already bound")
+            self._managers = dict(managers)
```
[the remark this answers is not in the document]
hhzhang16: Addressed. Managers now receive the checkpoint lifecycle directly, their condition and admission callback are derived from it, and the RPC server derives its lifecycle from the manager. bind_domains also rejects managers owned by another lifecycle, with a negative test covering the misconfiguration.

lib/gpu_memory_service/v1/checkpoint.py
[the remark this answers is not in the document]
hhzhang16: Kept the fence explicitly non-expiring because automatically reopening admission during an unresolved snapshot is unsafe but updated the documentation

lib/gpu_memory_service/v1/protocol.py
hhzhang16: Removed

lib/gpu_memory_service/v1/checkpoint.py
hhzhang16: Done

lib/gpu_memory_service/v1/checkpoint.py
hhzhang16: Yeah, removed

mohammedabdulwahhab (approved):

PR #12829, opened 2026-08-07
fix(vllm): shut down EngineCore on decode worker exit

furionw:
## Why

on current main, all listed operations happen:

```
  setup_vllm_engine()
      ↓ EngineCore is now running
  configure KV sizing
  connect to encode worker
  construct handler
  initialize publishers and metrics
  wait for optional `--benchmark-mode` results
  register model
      ↓
  existing serving try/finally begins
```

Decode workers start vLLM EngineCore before completing Dynamo-side handler, publisher, benchmark, and discovery initialization. An exception in that parent-side work can escape the main coroutine while EngineCore remains alive, so the process never terminates

**CustomEncoder validation is the concrete reproducer.**

Other post-engine failures can occur while configuring KV-event sizing, connecting to an encode worker, constructing the handler, initializing publishers or metrics, waiting for optional `--benchmark-mode` results, or registering the model with discovery. The existing serving-only cleanup does not cover those startup steps, and handler cleanup alone does not stop EngineCore.

## What Change

- Track partially constructed decode resources in one lifecycle context.
- Clean handler then EngineCore on every decode-worker exit.
- Preserve original failures and suppress the health-monitor shutdown race.

## Test Plan

- Run 46 focused lifecycle tests in the RC image.
- Verify CustomEncoder failure and normal SIGTERM shutdown on one GB200.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved engine monitoring during shutdown so expected engine failures and health-check timeouts no longer trigger unnecessary runtime shutdown or forced process termination.
  * Improved decode-worker startup and cleanup reliability, including consistent resource cleanup when initialization, configuration, or handler creation fails.
  * Preserved the original startup error while still reporting cleanup failures and safely shutting down the engine.

### DIFF
--- a/components/src/dynamo/vllm/engine_monitor.py
+++ b/components/src/dynamo/vllm/engine_monitor.py
@@ -116,6 +116,14 @@ class VllmEngineMonitor:
                     await asyncio.sleep(self.health_config.interval)
 
             except (EngineDeadError, asyncio.TimeoutError) as e:
+                if self.shutdown_event and self.shutdown_event.is_set():
+                    logger.warning(
+                        "%s: %s while worker shutdown is in progress; "
+                        "stopping health monitoring.",
+                        self.__class__.__name__,
+                        type(e).__name__,
+                    )
+                    break
                 logger.error(f"Traceback: {traceback.format_exc()}")
                 logger.error(f"vLLM AsyncLLM health check failed: {e}")
                 logger.warning("Initiating Dynamo Runtime shutdown.")
--- a/components/src/dynamo/vllm/tests/test_vllm_worker_factory.py
+++ b/components/src/dynamo/vllm/tests/test_vllm_worker_factory.py
@@ -16,6 +16,7 @@ from dynamo.vllm.constants import DisaggregationMode
 from dynamo.vllm.worker_factory import (
     EngineSetupResult,
     WorkerFactory,
+    _DecodeWorkerLifecycle,
     _wait_and_load_benchmark,
 )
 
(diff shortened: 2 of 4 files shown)


karen-sy (approved):

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -537,6 +539,44 @@ async def _wait_and_load_benchmark(bench_cfg: dict, vllm_config: VllmConfig) ->
...
+    def __exit__(
+        self,
+        _exc_type: type[BaseException] | None,
+        original_error: BaseException | None,
+        _traceback: TracebackType | None,
+    ) -> None:
+        try:
+            self.cleanup()
+        except BaseException:
+            if original_error is None:
```
peilii, line 561: `except BaseException` here catches `KeyboardInterrupt` and `SystemExit` from the blocking `engine_client.shutdown()` call, and swallows them when `original_error is not None`. The shutdown chain (`MultiprocExecutor._ensure_worker_termination`) busy-waits with `time.sleep(0.1)` for up to ~9s — a real window for Ctrl+C during a failed startup. The process will silently ignore SIGINT and can only be killed with SIGKILL.

Fix: narrow to `except Exception` so fatal signals always propagate, or re-raise non-`Exception` subclasses unconditionally:
```python
except BaseException as exc:
    if original_error is None or not isinstance(exc, Exception):
        raise
    logger.exception("Failed to clean up decode worker after an earlier failure")
```

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -537,6 +539,44 @@ async def _wait_and_load_benchmark(bench_cfg: dict, vllm_config: VllmConfig) ->
...
+    def cleanup(self) -> None:
+        """Release resources in reverse construction order."""
+        logger.debug("Cleaning up decode worker")
+        if self.shutdown_event is not None:
+            self.shutdown_event.set()
+        try:
+            if self.handler is not None:
+                self.handler.cleanup()
+        finally:
+            if self.engine_client is not None and self.vllm_config is not None:
```
peilii, line 576: If both `handler.cleanup()` (try block) and `engine_client.shutdown()` (finally block) raise, Python propagates the `finally` exception and demotes the handler exception to `__context__`. The caller only sees the engine error; the handler cleanup failure is invisible unless you walk the chain. `test_decode_worker_lifecycle_shutdown_engine_when_handler_cleanup_fails` only exercises the case where `engine_client.shutdown()` succeeds — worth adding a test for the double-failure path.

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -1227,10 +1290,6 @@ async def _create_decode_worker(
         except Exception as e:
             logger.error(f"Failed to serve endpoints: {e}")
             raise
-        finally:
-            logger.debug("Cleaning up decode worker")
-            # Cleanup background tasks
-            handler.cleanup()
 
     async def _create_prefill_worker(
         self,
```
peilii, line 1295: `_create_prefill_worker` still uses the old `finally: handler.cleanup()` pattern with no `engine_client.shutdown()`. If startup fails after `setup_vllm_engine()` returns (inside `configure_kv_event_block_size` or `PrefillWorkerHandler.__init__`), the EngineCore process leaks — the same bug the PR description names as the target. The lifecycle fix applies equally here.

components/src/dynamo/vllm/engine_monitor.py
```diff
@@ -116,6 +116,12 @@ async def _check_engine_health(self):
                     await asyncio.sleep(self.health_config.interval)
 
             except (EngineDeadError, asyncio.TimeoutError) as e:
+                if self.shutdown_event and self.shutdown_event.is_set():
+                    logger.info(
```
peilii: Two concerns with this guard:

1. **`asyncio.TimeoutError` semantics**: the outer `except (EngineDeadError, asyncio.TimeoutError)` catches both confirmed engine death *and* a health-check timeout (engine unresponsive). The new branch suppresses both when `shutdown_event` is set. A timed-out engine during shutdown may be hung (not merely stopping) — worth logging at WARNING and still calling `_shutdown_engine()` in that case, since the subsequent `engine_client.shutdown()` has no SIGALRM backstop if the engine is stuck.

2. **Log level**: logging at INFO means an EngineDeadError during concurrent shutdown (e.g., OOM kill racing with graceful teardown) is indistinguishable from a clean stop. The pre-existing `logger.error("vLLM AsyncLLM health check failed")` path is bypassed. Consider logging the exception at WARNING/ERROR even when breaking cleanly.

peilii (approved):
Approach is correct and the tests are solid. A few things to address before or shortly after merge:

components/src/dynamo/vllm/engine_monitor.py
```diff
@@ -116,6 +116,12 @@ async def _check_engine_health(self):
...
+                        f"{self.__class__.__name__}: Engine stopped during "
+                        "worker shutdown; stopping health monitoring."
+                    )
+                    break
```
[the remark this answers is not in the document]
furionw: Fixed in `a172dee0ca`. The shutdown-race message now uses lazy logging arguments, neutral exception-type wording, and WARNING severity.

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -537,6 +539,44 @@ async def _wait_and_load_benchmark(bench_cfg: dict, vllm_config: VllmConfig) ->
...
+        _traceback: TracebackType | None,
+    ) -> None:
+        try:
+            self.cleanup()
+        except BaseException:
+            if original_error is None:
+                raise
+            logger.exception(
+                "Failed to clean up decode worker after an earlier failure"
+            )
```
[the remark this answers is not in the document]
furionw: Fixed in `a172dee0ca`. Cleanup now catches `Exception`, so `KeyboardInterrupt`, `SystemExit`, and `asyncio.CancelledError` propagate even when another error is already active.

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -537,6 +539,44 @@ async def _wait_and_load_benchmark(bench_cfg: dict, vllm_config: VllmConfig) ->
...
+        """Release resources in reverse construction order."""
+        logger.debug("Cleaning up decode worker")
+        if self.shutdown_event is not None:
+            self.shutdown_event.set()
+        try:
+            if self.handler is not None:
+                self.handler.cleanup()
+        finally:
+            if self.engine_client is not None and self.vllm_config is not None:
+                self.engine_client.shutdown(timeout=self.vllm_config.shutdown_timeout)
```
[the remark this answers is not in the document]
furionw: Confirmed. Keeping `vllm_config.shutdown_timeout` is intentional: it matches upstream vLLM API-server shutdown behavior, and lifecycle cleanup runs only after the decode endpoints finish. The GB200 normal-shutdown smoke test reached readiness and exited 0 within 6.8s of SIGTERM.

components/src/dynamo/vllm/engine_monitor.py
```diff
@@ -116,6 +116,12 @@ async def _check_engine_health(self):
...
                 logger.error(f"Traceback: {traceback.format_exc()}")
                 logger.error(f"vLLM AsyncLLM health check failed: {e}")
                 logger.warning("Initiating Dynamo Runtime shutdown.")
```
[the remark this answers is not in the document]
furionw: Good tradeoff to call out. I am keeping the lifecycle as the sole EngineCore shutdown owner after `shutdown_event` is set; invoking monitor shutdown or `os._exit(1)` concurrently would reintroduce the race and can mask the original startup failure. `a172dee0ca` upgrades this to a neutral WARNING and tests both `EngineDeadError` and `asyncio.TimeoutError`. On the normal path, lifecycle cleanup starts only after the graceful endpoints finish; the startup-failure path has no in-flight requests.

components/src/dynamo/vllm/tests/test_vllm_worker_factory.py
```diff
@@ -49,6 +50,182 @@ def _make_config(**overrides) -> Mock:
...
+        additional_config={},
+        cache_config=SimpleNamespace(num_gpu_blocks=1),
+        model_config=SimpleNamespace(max_model_len=1024),
+        shutdown_timeout=5.0,
+    )
+    engine_setup: EngineSetupResult = (
+        engine_client,
+        vllm_config,
+        Mock(),
+        "/tmp/prometheus",
```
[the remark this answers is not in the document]
furionw: Fixed in `a172dee0ca`: the test now uses `tmp_path / "prometheus"`.

components/src/dynamo/vllm/worker_factory.py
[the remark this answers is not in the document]
furionw: Fixed in `a172dee0ca`. The handler now catches `Exception` rather than `BaseException`, preserving process-control exceptions.

components/src/dynamo/vllm/worker_factory.py
furionw: Fixed in `a172dee0ca`: ordinary cleanup failures still cannot mask the original worker failure, while process-control `BaseException` subclasses now propagate.

components/src/dynamo/vllm/worker_factory.py
furionw: Added coverage in `a172dee0ca`. The test verifies that the engine shutdown error propagates and the handler cleanup error remains attached as `__context__`, documenting the deliberate reverse-cleanup semantics.

components/src/dynamo/vllm/worker_factory.py
furionw: Agreed that prefill has the same structural ownership gap. This PR is intentionally scoped to decode workers—the reported reproducer and the PR title/contract—and I will handle prefill as a separate lifecycle follow-up rather than expanding this fix.

components/src/dynamo/vllm/engine_monitor.py
furionw: Partially addressed in `a172dee0ca`: shutdown-time health failures are now logged at WARNING with the concrete exception type, and both `EngineDeadError` and `asyncio.TimeoutError` are tested. I am not calling `_shutdown_engine()` after `shutdown_event` is set because that creates a second concurrent shutdown owner. The decode lifecycle immediately calls configured `shutdown(timeout=0)`, which force-terminates a stuck EngineCore; the GB200 normal SIGTERM smoke test exited cleanly.

furionw: There can be a follow-up between Lifecycle and the engine monitor -- there may be some overlapping that is worth consolidating

components/src/dynamo/vllm/worker_factory.py
```diff
@@ -537,6 +539,44 @@ async def _wait_and_load_benchmark(bench_cfg: dict, vllm_config: VllmConfig) ->
...
```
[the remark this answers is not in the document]
furionw: > but neither the time limit setting nor the time-limit option exists, so the stop attempt fails and the engine keeps running.

false positive because these do exist

- https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/config/vllm.py#L397-L401
- https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/engine/async_llm.py#L259-L267

PR #12844, opened 2026-08-07
fix(vllm): preflight custom encoder before engine startup

furionw:
## Why

CustomEncoder adapter validation currently runs after vLLM starts EngineCore. Deterministic configuration errors escape the main coroutine but leave that child alive, so a misconfigured deployment never exits or restarts. Validating the resolved class and adapter before `AsyncLLM` starts prevents the orphan while preserving serving initialization at its existing handler location.

## What Change

- Resolve and validate CustomEncoder before EngineCore startup.
- Reuse shared preparation logic in the existing handler load path.
- Cover invalid classes and preflight ordering.

## Test Plan

- GB200 RC mismatch exited 1 in 21.3 seconds with no EngineCore.
- Focused CustomEncoder unit suites and all pre-commit hooks.

### DIFF
--- a/components/src/dynamo/vllm/handlers.py
+++ b/components/src/dynamo/vllm/handlers.py
@@ -4,7 +4,6 @@
 import asyncio
 import base64
 import functools
-import importlib
 import inspect
 import logging
 import math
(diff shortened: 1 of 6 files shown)


furionw:
Superseded by #12829, which fixes the reported startup hang at the EngineCore lifecycle boundary.

PR #12856, opened 2026-08-07
feat(rl): add Slime external rollout example

Aphoh:
## Summary

- This PR adds an experimental Slime example with two SGLang engines in a fixed worker set.
- Each worker runs SGLang and `dynamo.sglang.sidecar` in one Dynamo SGLang runtime container.
- This design does not require a standalone sidecar image or a Kubernetes restartable init container.
- The frontend image, SGLang runtime image, and `runtimeVersionOverride` must use the same Dynamo version.
- Each worker requests one `nvidia.com/gpu` resource on clusters that use the NVIDIA device plugin.
- The example exposes each native engine through the `engine-0` and `engine-1` Services.
- These Service names do not conflict with the Services that the operator owns.
- THUDM/slime#2272 provides streaming external rollouts with fixed `--rollout-external-engine-addrs` addresses.
- The request adapter sends rollout generation through the Dynamo frontend `/generate` route.
- Slime calls the native engine Services only for control and weight updates.
- The CODEOWNERS source assigns `examples/rl/` to the RL team.

## Validation

- We rebased the original change onto `main` at `ebedc6cc13`.
- The cluster API accepted the three Services and the rendered DynamoGraphDeployment in a server-side dry run.
- For the stable release test, we used `Qwen/Qwen3-0.6B` with `dynamo-frontend:1.4.2` and `sglang-runtime:1.4.2`.
- Both one-GPU B200 workers became ready with zero restarts.
- The Dynamo `/v1/models` and `/v1/chat/completions` routes returned HTTP 200.
- For the pre-release test, we used the `1.5.0.dev20260903` frontend and SGLang runtime images.
- The DynamoGraphDeployment reached `successful`.
- Both workers reported that `dynamo.sglang.sidecar` started native gRPC discovery.
- We ran the external-engine discovery code from Slime commit `4c193f1` against both worker sets.
- In each test, Slime discovered two `regular` engines.
- Each engine reported one GPU.
- Each engine completed an eight-token stream.
- The pre-release Dynamo `/generate` route returned HTTP 200.
- We deleted both test deployments and their Services.
- We released all test GPU reservations.
- No test Dynamo or Grove resources remain.
- The `git diff --check` command passed.
- The `bash -n examples/rl/slime/deploy-dynamo.sh examples/rl/slime/launch-slime.sh` command passed.
- The documentation linter reported zero errors and eight warnings that also occur on `main`.
- The strict CODEOWNERS validation passed.
- The generated `CODEOWNERS` file matches `.github/codeowners/areas.yaml`.

## Limitations

- The stable `sglang-runtime:1.5.0` image is not available.
- The current pre-release Dynamo `/generate` route returned HTTP 200.
- The OpenAI chat route returned HTTP 500.
- The native gRPC bridge reads `GenerateReqInput.batch_size` before request normalization.
- This example does not use the OpenAI chat route.
- The live tests covered Dynamo serving, the Slime external-engine contract, and the Dynamo `/generate` route separately.
- The live tests did not cover a complete training cycle or weight-update cycle with the request adapter.
- The example uses two fixed worker Services because dynamic registration between Dynamo and Slime is not available.
- Other workloads reserved the Dynamic Resource Allocation (DRA) nodes in the cluster.
- The live tests used device-plugin GPUs through `nvidia.com/gpu`.


athreesh (commented):
can you create a slime specific guide in docs :) similar to verl and nemo-rl

dagil-nvidia (approved):
**Approving.**

This holds up. The manifest validates clean against the live v1beta1 DynamoGraphDeployment schema, and every assumption `dynamo_generate.py` makes about Slime checks out at the pinned commit: `generate_streaming` has the signature the adapter calls, it builds `http://{args.sglang_router_ip}:{args.sglang_router_port}/generate`, which is exactly why overriding those two attributes redirects generation and why rejecting `https://` is right, and `abort_mode` matches. `emit_codeowners.py` reproduces `CODEOWNERS` and `CONTRIBUTORS.md` with zero drift. Loading the sidecar in-process through SGLang's `--sidecar` instead of running a second container is the right call while there is no published sidecar image.

Three non-blocking items.

**Boundary Checklist**: `examples/rl/slime/README.md:160` closes the Current Boundary list with "Make sure that the system recovers", in the same section that says the example does not support external-engine fault recovery. Someone working the checklist hits a step that cannot pass. `slime.md` already has the right version, so I would lift its steps 5 and 6 over.

**Link Follow-Up**: the three `examples/README.md#integration-examples` links in `slime.md` and `integration-reference.md` render text promising `examples/rl/slime/`. Keeping lychee green before that directory exists on `main` is the right tradeoff, but worth a tracked follow-up to repoint them after this merges, or they stay indirect forever.

**SGLang Floor**: Prerequisites do not name a minimum SGLang version. `lib/sidecar/sglang/README.md` puts the native gRPC server at v0.5.16+, and the manifest passes `--grpc-port`. The Dynamo sglang-runtime image covers it, so this is a save-someone-a-debug line, not a real gap.

One process note. Full CI has not run against `e7faee9`: `backend-status-check`, `dynamo-status-check`, and `deploy-status-check` are all missing because `pull-request/12856` is still sitting at `84babe15`. That needs a fresh `/ok to test e7faee9` before this can merge.

jthomson04 (approved):

PR #12825, opened 2026-08-07
skills UAT feedback: resource envelope, correction protocol, agent-reported issues, capacity accounting

BenHamm:
Four small changes from a hands-on UAT run of the optimization skills (clean environment, live cluster, overnight autonomous loop). Each fixes something observed, not hypothesized. Merges into `argupte/dynamo-work`.

**1. Resource envelope + contract corrections** (`synthesize-user-workload`, workload schema)
- Elicit once at bootstrap: GPU floor/ceiling, parallel-experiment authorization, pinned knobs, teardown deadline. Ceiling is authorization, not entitlement; every experiment still passes evidence + adversarial review. Observed: the loop under-used available capacity because no envelope was ever established, and the only alternative offered mid-run was an illegitimate one (which it correctly refused).
- Fleet-vs-unit load confirmation: observed a full overnight run answering the wrong question because fleet-level traffic numbers were applied to a single-replica test. The contract itself contained the contradiction. *Maintainers: this one bullet is severable if you think it over-constrains the interview.*
- Contract Corrections section: corrections-of-fact amend in place + annotate ledger + supersede prior runs + re-baseline; material changes keep the existing new-root rule. Observed: a strong driver model improvised exactly this correctly; codifying it makes the behavior survive weaker drivers.

**2. Agent-reported issue protocol** (`AGENTS.md` + issue template)
Agents that hit instructions contradicting live verification file a sanitized, deduplicated, once-per-session issue, self-identified as agents. Observed: three instruction-vs-reality divergences in one session, all resolved by the agent, none leaving any trace for maintainers. The `agent-reported` label count doubles as a transparent adoption signal. **Needs a maintainer to create the `agent-reported` label.**

**3. Binding-based capacity accounting** (`deploy-dynamo-recipe`)
Count bound non-terminal pods as holding their GPU requests; don't filter on `phase == Running` (ImagePullBackOff pods hold reservations); exclude tenant-reserved tainted nodes; report largest per-node block. Observed: Running-only accounting reported a fully-claimed node as 8/8 free; the agent diagnosed the method error itself, and this transcribes its fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12825" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
[...]

PR #12858, opened 2026-08-07
docs: long-running run guidance + harness compatibility tiers

BenHamm:
Companion to #12825, from the same UAT run. One documentation section in AGENTS.md, no code.

**The finding**: running the optimization roles in a bare interactive Claude Code session, the loop silently stalled for 73 minutes when the agent ended a turn on narrated intent ("Now moving into the optimization loop...") with no tool call, no background job, and no question pending. The turn ended; the harness (correctly, for interactive use) yielded to a human who wasn't watching. This cannot happen under Codex `/goal`, which is why it had never surfaced in the existing Codex-based test runs — it's a driver-harness gap, not a skills defect.

**The fix is documentation, not machinery** (we considered and rejected shipping a Stop hook — both major harnesses already have native goal modes):
- Wrap loops in the harness goal mode: Codex `/goal` + token budget; Claude Code `/goal` (v2.1.139+) + `or stop after N turns` clause.
- Never end a loop turn on narrated intent: act, background the work, or return the blocking question.

**Harness tiers**: tested = Claude Code + Codex (role configs ship for both). The skills follow the Agent Skills open standard and load unmodified on compliant harnesses like OpenCode (which reads `.claude/skills/` and `.agents/skills/`), with three documented degradations: no goal mode there (use an external loop), role isolation may collapse to one context (weakens adversarial-review independence), and all discipline is prompt-enforced, so it degrades with weaker driver models. Field reports from other harnesses arrive via the issue protocol proposed in #12825.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12858" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,25 @@ user supplied. Do not dispatch `recipe_deployer`, `perf_analyzer`, `hypothesis_g
 `recipe_deployer`; pass the same immutable workload path and hash to every later role. Do not insert a recipe
 exploration or selection step before the baseline deployment.
 
+## Long-Running Runs And Harness Compatibility
+
+An optimization loop is long-running, unattended work. An interactive harness ends its turn whenever the agent stops
+calling tools — a turn that ends on narrated intent ("now I'll test disagg") silently stalls the loop until a human
+notices. Two rules:
+
+1. **Wrap the loop in your harness's goal mode.** On Codex CLI, use `/goal` with a token budget. On Claude Code
+   (v2.1.139+), use `/goal` with an `or stop after N turns` clause. Both keep the loop running until the objective is
+   judged complete or the budget is reached.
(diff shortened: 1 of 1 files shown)


BenHamm:
[Claude Code, on Ben's behalf] Pushed corrections from the same cross-model review: goal mode is documented as an operator launch prerequisite (not agent-self-enablable), the Claude Code turn-bound clause is described accurately as a model-evaluated condition rather than a hard budget, and the harness-tier claim is fixed — isolated role configs ship for Codex only today (`.codex/agents/`); on Claude Code the roles run in-context, so the adversarial-review-independence caveat applies there too, not just to third-party harnesses. Issue-filing guidance aligned with the draft-and-approve protocol in #12825.

BenHamm:
Consolidated into #12969 per maintainer preference — all commits from this PR are on that branch unchanged. Review there.

PR #12861, opened 2026-08-08
docs(fern): fix the Home sidebar duplication and make both calendars track the current date

harryskim:
## Overview:

Four fixes to the Fern docs Home and Community pages, plus the CI change one of them depends on. All found by rendering the site locally (`fern docs dev`) rather than reading config.

## Details:

**1. The Home sidebar duplicated the hero cards** (`07827f6`)

The Home tab's sidebar carried `Join Slack` and `Community Calendar` links pointing at the same two URLs as the notification cards `CommunityRail` renders in the hero, so the landing page shipped each destination twice. The landing page now sets `layout: page` — the Fern layout that hides the navigation sidebar as well as the TOC, replacing the narrower `hide-toc` — and the two duplicated links come out of the tab.

Side effect worth knowing: the hero is now centred in the viewport, which is what the community rail's `right: calc((100vw - 1200px) / -2 + 1rem)` already assumed. Previously the 272px sidebar offset the article and pushed those cards up into the header. Other tabs keep their sidebars; only Home is affected.

**2. Neither calendar tracked the current date** (`3221a22`, `152658c`)

Both `EventsCalendar` (Home) and `CommunityLanding`'s `FullCalendar` picked their displayed month with `UPCOMING_EVENTS[0] ?? PAST_EVENTS[0]` — the next event's month, not today's. Once the calendar emptied out, that fell back to the most recent *past* event and stayed there. The `new Date()` fallbacks were unreachable in practice; they only fired when both arrays were empty.

On 2026-08-07 the live Google Calendar has no upcoming events, so Home showed **July 2026 with the 30th highlighted** next to "No upcoming events are scheduled."

The Community page failed worse: its today marker was resolved separately from a client-side `new Date()`, so with the grid on a stale month the marker matched no cell and the highlight vanished entirely.

`generate-events.js` now emits `GENERATED_ON`, a Pacific `YYYY-MM-DD`, and both grids resolve today from it through a shared `resolveToday()` in `calendar-today.ts`. Extracted rather than pasted a third time — the copy-paste is why this defect existed in two places.

A date, not a timestamp, on two counts: the grid only resolves to a day, and since the file is now committed a timestamp would differ on every six-hour run and republish the docs four times a day over a byte that renders identically. Verified idempotent — back-to-back generator runs produce no diff.

Because today is known at generation time, `FullCalendar` is now server-renderable; it held the file's only hooks and touches no browser APIs, so `"use client"` came out with them. The marker is in the initial HTML instead of appearing after hydration.

Also: the red highlight used to mean "next event" and now means "today", so event days get a dot instead — white when it lands on today.

**3. The events refresh moved to its own workflow** (`9dcd54a`)

It was a step inside `fern-docs.yml` driven by a `schedule:` trigger on that workflow. Regenerating there meant the refreshed `events.generated.ts` only ever existed inside the runner and synced straight to `docs-website`, so `main` — the copy people read and diff — drifted from what the site served.

`community-events-refresh.yml` runs the same six-hour cron, regenerates the file, commits it to `main`, and stops. `main` is already in `fern-docs.yml`'s `push:` trigger and the file sits under `docs/**`, so the existing sync-and-publish job picks it up. One source of truth, and the calendar gets a history.

The push uses `OPS_BOT_PAT` rather than `GITHUB_TOKEN`, matching `auto-dep-upgrade-complete.yml`: GitHub suppresses workflow runs triggered by `GITHUB_TOKEN` pushes, so the refresh would otherwise land on `main` and never publish.

**4. Post-merge CI skips docs-only pushes** (`f207fa9`)

`post-merge-ci.yml` runs the full container build and GPU test pipeline on every push to `main`, with no path filter. `trigger_ci.yml` already excludes docs paths on the reasoning that docs deployment is `fern-docs.yml`'s job; the same holds here.

This is **load-bearing** for the change above — without it, a daily calendar commit would wake the GPU pipeline to rebuild containers over a changed date string.

Only the `push:` trigger is narrowed. `paths-ignore` does not apply to `workflow_dispatch`, so `auto-dep-upgrade-trigger.yml`'s `gh workflow run post-merge-ci.yml` against `deps/upgrade-*` branches is unaffected, and with it the `workflow_run` hand-off to `auto-dep-upgrade-complete.yml`.

- **`.github/workflows/post-merge-ci.yml`** — the only change outside `docs/`, and a CI policy change: docs-only merges to `main` will stop running post-merge CI. Happy to split this into its own PR if the CI owners would rather review it separately.
- **`.github/workflows/community-events-refresh.yml`** — new file. Needs `OPS_BOT_PAT` to be available to it.
- **`docs/fern/components/calendar-today.ts`** — the shared date logic both calendars now depend on.

### Verification

Rendered locally against a real `fern docs dev` server, not reasoned from config:

- `fern check`: 0 errors. `check_style_components.py`, `check_asset_paths.py`, `sync_site_css.py` all pass.
- Home: sidebar `display: none`, article centred, tabs intact; `/dynamo/dev/kubernetes` keeps its 272px sidebar; mobile 375px stacking unchanged; no console errors.
- Ran the generator against the live Google Calendar (0 upcoming, 4 past — all July), reproducing the exact failure, then confirmed both grids render August 2026 / day 7, Aug 1 correctly on a Saturday, `aria-current="date"` set.
- Confirmed `is-today` is present in raw server HTML via `curl`, where before it only existed post-hydration.

**Not verified:** everything CI-side is static validation — both workflows parse and carry the right triggers, but I can't execute a GitHub Actions run locally. The cron firing, the PAT push triggering `fern-docs.yml`, and the `paths-ignore` behaviour are reasoned from the workflow definitions. The first scheduled run after merge is the real test. Separately, the event-dot styling is confirmed but unexercised by real data — all four known events are in July, so August has none.

## Related Issues

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12861" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added an automated community events refresh that updates event information throughout the day.
  - Added calendar indicators for event dates, including accessible selected-day states.
  - Calendar now consistently displays the current month and highlights today.

- **Improvements**
  - Simplified home page navigation to focus on the Welcome page.
  - Reduced unnecessary CI runs for documentation-only changes.
  - Improved home page layout by hiding side navigation and the table of contents.


dagil-nvidia (approved):

.github/workflows/community-events-refresh.yml
```diff
@@ -0,0 +1,93 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          # Only ever stage the generated file. npm ci writes node_modules into
+          # .github/scripts/events/ and a stray `git add -A` would commit it.
+          git add "$EVENTS"
+          git commit -m "docs(fern): refresh community events
+
+          Automated refresh from the public Dynamo Google Calendar.
+          Generated by .github/workflows/community-events-refresh.yml."
```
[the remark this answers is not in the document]
dagil-nvidia: Fixed in 68e65b58a9: the refresh commit now uses `git commit -s` with the dynamo-ops identity (same as auto-dep-upgrade), so the Signed-off-by trailer is present.

.github/workflows/community-events-refresh.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          # Rebase rather than force: main moves under us during the fetch, and
+          # this file has a single writer, so there is nothing to conflict with.
+          git pull --rebase origin main
+          git push origin HEAD:main
```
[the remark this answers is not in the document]
dagil-nvidia: Acknowledged and addressed at the ruleset layer: `@ai-dynamo/autobots` (which includes `dynamo-ops`) is now a bypass actor on the repo "Required PR Checks" ruleset, so the OPS_BOT_PAT push to main is intentional rather than accidental. Workflow header comments document the dependency. Org-level "Default branches" still needs a matching bypass if push fails after merge — will verify with a workflow_dispatch dry run.

.github/workflows/post-merge-ci.yml
```diff
@@ -8,6 +8,18 @@ on:
...
+    # Same exclusion trigger_ci.yml carries, for the same reason.
+    #
+    # This is load-bearing for community-events-refresh.yml, which commits a
+    # regenerated events.generated.ts to main on a schedule: without it, a
+    # calendar refresh would wake the full GPU pipeline once a day.
+    paths-ignore:
+      - 'docs/**'
+      - 'fern/**'
+      - '**.md'
+      - '**.rst'
```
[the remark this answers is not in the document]
dagil-nvidia: Intentional and consistent with `trigger_ci.yml`. The calendar refresh is the load-bearing case for the ignore; the broader `**.md` / `**.rst` match is accepted as matching existing CI trigger policy.

.github/scripts/generate-events.js
```diff
@@ -90,6 +102,8 @@ function toEvent(e, isPast) {
 async function main() {
   const events = await ical.async.fromURL(ICS_URL);
   const now = new Date();
+  // en-CA formats as YYYY-MM-DD, which is what the component parses.
+  const generatedOn = now.toLocaleDateString('en-CA', { timeZone: 'America/Los_Angeles' });
```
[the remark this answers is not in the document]
dagil-nvidia: Fixed in 68e65b58a9: `GENERATED_ON` is now assembled from `Intl.DateTimeFormat(...).formatToParts()` (year/month/day) instead of relying on `en-CA` locale string layout.

.github/workflows/community-events-refresh.yml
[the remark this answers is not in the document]
dagil-nvidia: Fixed in 68e65b58a9: `git commit -s` added.

dagil-nvidia (approved):
Re-approved after direct-push hardening (DCO -s, formatToParts, autobots ruleset bypass).

dagil-nvidia:
### Direct-push hardening (68e65b58a9)

Made the scheduled `main` push path viable:

1. **DCO** — refresh commit uses `git commit -s` as `dynamo-ops` (same identity as auto-dep-upgrade).
2. **GENERATED_ON** — built via `Intl.DateTimeFormat(...).formatToParts()` instead of `en-CA` locale string.
3. **Ruleset bypass** — `@ai-dynamo/autobots` (includes `dynamo-ops`) added as a bypass actor on the repo **Required PR Checks** ruleset.

**Still needed to merge this PR:** an `@ai-dynamo/dynamo-ops-codeowners` approval (workflow + script paths). Docs approval is already covered.

**After merge:** run `workflow_dispatch` on Community Events Refresh once. If the push is rejected, the org-level **Default branches** ruleset also needs `@ai-dynamo/autobots` as a bypass actor (repo admin cannot edit that one).

dagil-nvidia:
@harryskim sorry, I ended up pushing straight to your branch (`68e65b58a9`) instead of leaving suggestions. Here's what I changed, revert any of it if you'd rather take it yourself:

- **DCO** — the refresh commit now uses `git commit -s` and commits as `dynamo-ops`, so the `Signed-off-by` trailer is present when the bypass push lands on main.
- **GENERATED_ON** — switched from `toLocaleDateString('en-CA')` to `Intl.DateTimeFormat(...).formatToParts()` and assemble `YYYY-MM-DD` explicitly, so an ICU/CLDR update can't quietly change the format and drop `resolveToday` back to the host clock.
- **Ruleset bypass (outside the PR)** — added `@ai-dynamo/autobots` (which includes `dynamo-ops`) as a bypass actor on the repo "Required PR Checks" ruleset, so the scheduled push to main is intentional rather than a silently-failing cron.

Resolved the CodeRabbit and Devin threads that these covered. One thing still on you: this needs a `@ai-dynamo/dynamo-ops-codeowners` approval to merge (I'm only on docs). And after merge, worth one `workflow_dispatch` run to confirm the push clears the org-level "Default branches" ruleset too.

.github/workflows/community-events-refresh.yml
```diff
@@ -0,0 +1,98 @@
...
+            node .github/scripts/generate-events.js
+
+      - name: Commit and push if the calendar moved
+        env:
+          EVENTS: docs/fern/components/events.generated.ts
+        run: |
+          if [ -z "$(git status --porcelain -- "$EVENTS")" ]; then
+            echo "No calendar change; nothing to publish."
+            exit 0
+          fi
```
dmitry-tokarev-nv, line 79: **A degenerate calendar fetch now gets committed to `main` and persists.**

`generate-events.js` has no floor check. If `ical.async.fromURL` returns a body with no `VEVENT`s -- calendar unshared or made private (node-ical parses a non-200 body into `{}` rather than throwing), a transient Google serve, or everything `CANCELLED` -- it writes `UPCOMING_EVENTS = []` and `PAST_EVENTS = []`, this guard sees a diff, and the run pushes it to `main`. Both calendars go blank on the live site within hours.

The old design was self-healing precisely because the file was ephemeral in the runner: a bad run affected one publish and the next publish fixed it. Committing makes the bad state the source of truth until a human notices -- and the commit is titled "refresh community events" like every other one, so nothing stands out in the history.

```suggestion
          if [ -z "$(git status --porcelain -- "$EVENTS")" ]; then
            echo "No calendar change; nothing to publish."
            exit 0
          fi

          # Floor check. A fetch that parses but yields no VEVENTs (calendar
          # unshared, transient Google serve) would otherwise blank both grids
          # on main and stay that way until someone notices.
          new_count=$(grep -c '"title":' "$EVENTS" || true)
          old_count=$(git show "HEAD:$EVENTS" | grep -c '"title":' || true)
          if [ "$new_count" -eq 0 ] && [ "$old_count" -gt 0 ]; then
            echo "::error::Generator returned 0 events but main has $old_count. Refusing to publish an empty calendar."
            exit 1
          fi
```

.github/workflows/community-events-refresh.yml
```diff
@@ -0,0 +1,98 @@
...
+          git config user.name "dynamo-ops"
+          git config user.email "170655669+dynamo-ops@users.noreply.github.com"
+
+          # Only ever stage the generated file. npm ci writes node_modules into
+          # .github/scripts/events/ and a stray `git add -A` would commit it.
+          git add "$EVENTS"
+          git commit -s -m "docs(fern): refresh community events
+
+          Automated refresh from the public Dynamo Google Calendar.
+          Generated by .github/workflows/community-events-refresh.yml."
```
dmitry-tokarev-nv: **Every commit body will be indented by 10 spaces.**

The `-m` argument is a YAML block-scalar continuation, so lines 90-91 carry their block indentation into the commit message verbatim. `git` does not strip it -- only the leading `#` comment stripping and trailing-whitespace rules apply. Each daily commit will render with an indented body.

```suggestion
          git commit -s \
            -m "docs(fern): refresh community events" \
            -m "Automated refresh from the public Dynamo Google Calendar." \
            -m "Generated by .github/workflows/community-events-refresh.yml."
```

.github/workflows/community-events-refresh.yml
```diff
@@ -0,0 +1,98 @@
...
+
+          # Rebase rather than force: main moves under us during the fetch, and
+          # this file has a single writer, so there is nothing to conflict with.
+          git pull --rebase origin main
```
dmitry-tokarev-nv: **The justifying comment is already false, and there is no conflict handling.**

"this file has a single writer, so there is nothing to conflict with" is falsified by any human PR touching `events.generated.ts` -- this PR does exactly that. If such a merge lands between the checkout and the push, the rebase halts and the step exits non-zero.

Impact is a red scheduled run rather than corruption (the next run re-checks-out clean, so it self-heals), but it will page whoever watches Actions, and on a six-hour cron it will recur. For a generated file with no meaningful merge semantics, discarding and regenerating is more robust than rebasing:

```
git fetch origin main
git reset --hard origin/main
# ...re-run the generator and re-commit, or wrap the whole step in a short retry loop
git push origin HEAD:main
```

At minimum, worth softening the comment so the next reader does not trust the "single writer" premise.

.github/scripts/generate-events.js
```diff
@@ -90,6 +102,19 @@ function toEvent(e, isPast) {
...
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+    })
+      .formatToParts(now)
+      .map(({ type, value }) => [type, value]),
+  );
+  const generatedOn = `${parts.year}-${parts.month}-${parts.day}`;
 
   const all = Object.values(events)
```
dmitry-tokarev-nv, line 135: Root cause for the guard I suggested over in `community-events-refresh.yml`: there is no floor here either. `all`, `upcoming`, and `past` can all legitimately come out empty, and the script writes and exits 0 regardless.

That was harmless while the output was ephemeral. Now that the file is committed to `main` and is the source of truth, an empty result is a destructive change that survives until someone notices. A guard belongs in one place or the other -- either refuse to write when the parse yields zero `VEVENT`s while the on-disk file has some, or refuse to commit (my inline suggestion takes the latter route since it can compare against `HEAD`).

docs/fern/components/EventsCalendar.tsx
```diff
@@ -96,12 +97,23 @@ function UpcomingEvent({ event }: { event: DynamoEvent }) {
...
 
 export function EventsCalendar() {
-  const focusEvent = UPCOMING_EVENTS[0] ?? PAST_EVENTS[0];
-  const year = Number(focusEvent?.year ?? new Date().getUTCFullYear());
-  const month = MONTH_INDEX[focusEvent?.month ?? "Jan"] ?? 0;
-  const selectedDay = Number(focusEvent?.day ?? 1);
+  // The grid follows the current month, not the next event. Keying it off
+  // UPCOMING_EVENTS[0] meant an empty calendar fell back to PAST_EVENTS[0] and
+  // sat on a month that had already gone by.
+  const { year, month, day: selectedDay } = resolveToday();
```
dmitry-tokarev-nv: **Pinning the grid to today's month with no navigation means it can never show an event in another month.**

Events are announced in advance, so "the next event is next month" is the common case, not the edge case. The grid now filters to `resolveToday()`'s month only (`eventDays` below does the same), and there is no next/prev control.

With the `events.generated.ts` this PR commits -- `GENERATED_ON = "2026-08-07"`, `UPCOMING_EVENTS = []`, all four past events in July -- this grid ships **empty**: August with zero `has-event` dots. Before, it showed July with the events visible; the defect was that it also marked the wrong day as current. The fix trades a wrong marker for a blank widget.

The PR description concedes the dot styling is "unexercised by real data ... all four known events are in July, so August has none", which is the same gap seen from the styling side.

Worth deciding between: (a) show the current month, but fall back to the next event's month when the current month has none, or (b) add month navigation. Either keeps the marker honest without emptying the grid.

docs/fern/components/CommunityLanding.tsx
```diff
@@ -132,37 +133,19 @@ function buildMonthCells(year: number, month: number) {
...
-  const [today, setToday] = useState<CalendarDate | null>(null);
-  const focusEvent = UPCOMING_EVENTS[0] ?? PAST_EVENTS[0];
-  const year = Number(focusEvent?.year ?? new Date().getUTCFullYear());
-  const month = MONTH_INDEX[focusEvent?.month ?? "Jan"] ?? 0;
+  // The grid follows the current month, not the next event. Keying it off
+  // UPCOMING_EVENTS[0] meant an empty calendar fell back to PAST_EVENTS[0] and
+  // sat on a month that had already gone by -- and because the today marker
+  // was resolved independently, it then matched no cell in the month on show,
+  // so the grid lost its highlight entirely.
+  const today = resolveToday();
```
dmitry-tokarev-nv, line 146: Same current-month-only concern as `EventsCalendar.tsx`, and it bites harder here: this grid renders full event slots (`<a>` with `<time>` and the event title) inside the day cells, so it is the page's marquee widget rather than a small preview.

With the committed data (GENERATED_ON August, all events July, zero upcoming) the Community calendar ships as an empty August grid, and the "Up next" list below it is empty too -- so the section renders with no event content at all. Previously the grid at least showed July's four events.

The today-marker fix itself is right, and resolving the marker and the month from one source is the correct call. It is only the "always current month, no fallback, no navigation" part that loses information.

docs/fern/components/calendar-today.ts
```diff
@@ -0,0 +1,45 @@
...
+ * GENERATED_ON is a Pacific YYYY-MM-DD refreshed on a six-hour cron by
+ * community-events-refresh.yml, so a grid keyed off this tracks the real month
+ * with no client-side date logic: no hydration mismatch, no post-hydration
+ * flash, and the same answer on the server and in the browser.
+ *
+ * Parsed field-wise rather than through `new Date(string)`, which reads a bare
+ * YYYY-MM-DD as UTC midnight and so reports the previous day anywhere west of
+ * Greenwich -- Pacific included.
+ */
+export function resolveToday(): CalendarDate {
```
dmitry-tokarev-nv: **No staleness guard, so a stalled refresh silently reintroduces the bug this PR fixes.**

`resolveToday()` drives the displayed *month*, not just the marker. If `community-events-refresh.yml` stops committing -- `OPS_BOT_PAT` expiry, dynamo-ops not actually on `@ai-dynamo/autobots` as a bypass actor, repeated ICS failures, or a `fern-docs.yml` publish failure -- `GENERATED_ON` freezes and both grids pin to a stale month and mark a wrong day `is-today`, with no visual cue. That is exactly "sat on a month that had already gone by", just triggered differently.

Note this is strictly worse than the old behaviour on the Community page specifically, where the marker came from a client-side `new Date()` and was therefore always correct (the bug there was that the *grid* disagreed with it, not the marker).

Cheap mitigation: if `GENERATED_ON` is more than a few days behind, return no marker rather than a confidently wrong one, and let the grid fall back to the next event's month. A wrong "today" is worse than an absent one.

Minor, same function: the malformed-input fallback calls `new Date()` in what is now a server component, so it resolves to build time in the build machine's timezone -- the "arbitrarily old" case the header comment says it is avoiding. Unreachable for well-formed input, but the fallback does not do what the comment promises.

.github/workflows/post-merge-ci.yml
dmitry-tokarev-nv: Flagging this for the CI owners you asked for, since it is the one change outside `docs/`.

Post-merge CI is what pushes the SHA-tagged artifacts: `${{ github.sha }}-operator`, `${{ github.sha }}-snapshot-agent`, `${{ github.sha }}-power-agent`, and the `main-<framework>-runtime-${{ github.sha }}` family for vllm/sglang/trtllm plus planner and frontend. After this change, any docs-only merge -- including the new daily bot commit -- produces a `main` HEAD SHA for which none of those tags exist. Today every `main` SHA has images; afterwards that stops being true.

I grepped `.github/`, `deploy/`, `recipes/`, and `docs/` and found no in-repo consumer of the `main-*-<sha>` tags, so this looks safe from where I can see. The out-of-repo consumers are the open question -- the dynamo-ci GitLab pipelines and any deploy tooling that resolves "the image built from current main" by SHA. Worth a confirmation from whoever owns those before merge.

The `paths-ignore` patterns themselves are correct: `**` matches `/` in Actions path filters, so `'**.md'` covers nested markdown, and the set matches `trigger_ci.yml` exactly.

dmitry-tokarev-nv (commented):
Reviewed the full diff and traced the claims that were checkable rather than taking them on trust. Confirmed good: `.github/filters.yaml:27` carries `docs/**`, so the bot push does set `docs=true` and `fern-docs.yml` publishes; `trigger_ci.yml:26-30` carries the identical `paths-ignore` block, so the precedent is real; `auto-dep-upgrade-trigger.yml:103` does use `gh workflow run` (workflow_dispatch), so `paths-ignore` genuinely does not break the `workflow_run` hand-off to `auto-dep-upgrade-complete.yml`; `generate-events.js:21` resolves `OUT` from `__dirname`, so the relocated invocation writes the same path; and neither `CommunityLanding.tsx` nor `EventsCalendar.tsx` retains a hook, event handler, or browser API, so dropping `"use client"` is safe. The new `.has-event` CSS lands after `.is-selected` at equal specificity, but the `.has-event.is-selected` override handles the `color` clash (only `font-weight` 700-vs-750 leaks, which is invisible). The shallow `fetch-depth: 1` does not break the rebase either -- the local commit's parent is the old `main` tip, so a merge base exists.

Seven findings inline. The two I would not merge without a decision on are the empty-calendar commit guard and the current-month-only grid.

One more that has no diff line to hang off: **`.github/scripts/generate-events.js:140`** -- the generated file's own banner still reads `Source: Dynamo public Google Calendar. Refreshed during Fern publishing.` That is no longer true after this PR; it is refreshed by `community-events-refresh.yml` and committed to `main`. Since the whole point of the move is that people now read and diff this file on `main`, the header should say where it actually comes from.

.github/workflows/community-events-refresh.yml
harryskim: Fixed in 4ca4bba — guard added, though in the generator rather than here, per your note on `generate-events.js` that it belongs in one place or the other. It refuses to write when the fetch yields no usable events while the committed file still holds some, so the run fails and the last good calendar stays on main.

Putting it at the generator covers a manual `node .github/scripts/generate-events.js` too, and lets it report the VEVENT count directly instead of `grep -c '"title":'` against `HEAD`.

Verified with a stub that reproduces node-ical parsing a non-200 body into `{}`:

```
$ NODE_PATH=… node .github/scripts/generate-events.js
Error: Calendar fetch yielded no usable events (0 VEVENT(s) parsed) while
events.generated.ts currently holds 4. Refusing to overwrite it with an empty calendar.
exit code: 1
events in file after: 4   (unchanged)
```

Also checked it does not false-positive when the file is legitimately absent (first run / fresh checkout): exits 0 and writes normally.

.github/scripts/generate-events.js
harryskim: Agreed, and this is where I put it — 4ca4bba adds the floor check here rather than in the workflow, since it covers manual runs too and can compare against the on-disk file directly.

```js
const existingCount = countExistingEvents();
if (upcoming.length + past.length === 0 && existingCount > 0) {
  throw new Error(`Calendar fetch yielded no usable events (${veventCount} VEVENT(s) parsed) …`);
}
```

`countExistingEvents()` counts `"addUrl":` occurrences (one per event) rather than importing the file, since it is TypeScript, and treats ENOENT as zero so a first run is unaffected. Details and the stubbed-fetch verification are in the reply on the workflow thread.

.github/workflows/community-events-refresh.yml
harryskim: Good catch — fixed in 4ca4bba, using separate `-m` flags as suggested.

Reproduced both forms with real git first:

```
# before (YAML block scalar)
docs(fern): refresh community events

          Automated refresh from the public Dynamo Google Calendar.
          Generated by .github/workflows/community-events-refresh.yml.

# after (separate -m flags)
docs(fern): refresh community events

Automated refresh from the public Dynamo Google Calendar.

Generated by .github/workflows/community-events-refresh.yml.

Signed-off-by: …
```

Each `-m` becomes its own paragraph, so the two body lines are now separated by a blank line rather than adjacent. Reads fine and `-s` still lands last.

.github/workflows/community-events-refresh.yml
harryskim: You are right that the premise was already false — this PR falsifies it. Fixed in 4ca4bba by taking the discard-and-replay route rather than softening the comment:

```bash
generated=$(mktemp); cp "$EVENTS" "$generated"
for attempt in 1 2 3; do
  git fetch --depth=1 origin main
  git reset --hard FETCH_HEAD
  cp "$generated" "$EVENTS"
  [ -z "$(git status --porcelain -- "$EVENTS")" ] && exit 0   # main already has it
  git add "$EVENTS"; git commit -s -m … 
  git push origin HEAD:main && exit 0
  echo "Push rejected on attempt ${attempt} (main moved); replaying."
done
exit 1
```

No rebase, so no conflict is possible: the file is wholly generated and taking our copy is always the right resolution. The retry absorbs a push that races in between the fetch and the push, and the mid-loop check exits cleanly if the merge that beat us already carried the same content. The stale comment is gone.

docs/fern/components/EventsCalendar.tsx
harryskim: Fixed in 4ca4bba — took option (a), fallback to the nearest event's month, since it keeps the grid useful without adding stateful navigation to a server component.

`resolveCalendarMonth()` in `calendar-today.ts` returns today's month when it actually has events, else the soonest upcoming event's month, else the most recent past one. The marker is now gated on the displayed month matching today's, so it cannot land on the same-numbered day of another month.

Ported the resolver to JS and ran it against the data this PR actually commits (`GENERATED_ON=2026-08-07`, zero upcoming, four July events):

| Scenario | Grid | Events shown | Today marker |
|---|---|---|---|
| Today Aug 7 — the case you flagged as shipping empty | **Jul 2026** | **4** | none |
| Refresh stalled 10 days | Jul 2026 | 4 | none |
| Today Jul 20 (today's month has events) | Jul 2026 | 4 | **day 20** |
| Next event in Sep, today Aug | **Sep 2026** | 1 | none |
| Malformed `GENERATED_ON` | Jul 2026 | 4 | none |

So the blank-August grid now renders July's four events, and the marker only appears when the grid is genuinely on today's month.

docs/fern/components/CommunityLanding.tsx
harryskim: Fixed by the same change in 4ca4bba — both grids call `resolveCalendarMonth()`, so the fallback and the marker gating land here too, and the two stay in agreement by construction.

The `isToday` check here needed the extra month comparison you would expect once the grid is no longer guaranteed to be on today's month:

```tsx
const isToday =
  day !== null &&
  today !== null &&
  today.year === year &&
  today.month === month &&
  day === today.day;
```

With the committed data this section now renders July's four events in their day cells instead of an empty August grid. The "Up next" list below is still empty, correctly — there genuinely are no upcoming events — but the section no longer has zero event content.

Ported the resolver to JS and ran it against the data this PR actually commits (`GENERATED_ON=2026-08-07`, zero upcoming, four July events):

| Scenario | Grid | Events shown | Today marker |
|---|---|---|---|
| Today Aug 7 — the case you flagged as shipping empty | **Jul 2026** | **4** | none |
| Refresh stalled 10 days | Jul 2026 | 4 | none |
| Today Jul 20 (today's month has events) | Jul 2026 | 4 | **day 20** |
| Next event in Sep, today Aug | **Sep 2026** | 1 | none |
| Malformed `GENERATED_ON` | Jul 2026 | 4 | none |

So the blank-August grid now renders July's four events, and the marker only appears when the grid is genuinely on today's month.

docs/fern/components/calendar-today.ts
harryskim: Both fixed in 4ca4bba.

`resolveToday()` now returns `CalendarDate | null` and yields null once `GENERATED_ON` is more than three days behind, so a stalled refresh drops the marker rather than showing a confidently wrong one — and the month falls back to the nearest event, as you suggested. Three days leaves room for a transient ICS failure while still catching an expired `OPS_BOT_PAT` or a lost ruleset bypass.

On the minor point: the `new Date()` fallback is gone. Malformed input now returns null for the same reason as staleness, so the function no longer contradicts its own header comment.

One nuance worth recording, since it bounds how much the guard buys: `Date.now()` is evaluated at build time in a server component, so a stalled refresh is only detected when something else triggers a rebuild. That is not as weak as it sounds — any docs commit republishes — but if the whole publish pipeline is down, the site keeps serving the last build regardless, guard or no guard.

Verified across staleness, malformed-input, and same-month cases:

Ported the resolver to JS and ran it against the data this PR actually commits (`GENERATED_ON=2026-08-07`, zero upcoming, four July events):

| Scenario | Grid | Events shown | Today marker |
|---|---|---|---|
| Today Aug 7 — the case you flagged as shipping empty | **Jul 2026** | **4** | none |
| Refresh stalled 10 days | Jul 2026 | 4 | none |
| Today Jul 20 (today's month has events) | Jul 2026 | 4 | **day 20** |
| Next event in Sep, today Aug | **Sep 2026** | 1 | none |
| Malformed `GENERATED_ON` | Jul 2026 | 4 | none |

So the blank-August grid now renders July's four events, and the marker only appears when the grid is genuinely on today's month.

.github/workflows/post-merge-ci.yml
harryskim: Leaving this one **open** deliberately — the out-of-repo question is not mine to close, and it should have a CI owner sign off before merge.

I did re-run the in-repo half independently and it corroborates yours:

- The `paths-ignore` parity claim checks out exactly. `trigger_ci.yml` carries `["docs/**", "fern/**", "**.md", "**.rst"]` — the same four patterns, same order.
- Searching `.github`, `deploy`, `recipes`, `docs`, and `container` for the SHA-tag families, every hit is in `pr.yaml` and `build-on-demand.yml`, and those build images for their *own* `github.sha` rather than resolving a `main` SHA someone else built. So no in-repo consumer depends on "every main SHA has images".

That leaves exactly the gap you identified: the dynamo-ci GitLab pipelines and any deploy tooling that resolves an image by main SHA. Flagging for CI owners — @ai-dynamo/dynamo-ci-owners or whoever owns the GitLab side — and happy to drop the `post-merge-ci.yml` change from this PR if confirmation is going to take a while, since it is independent of the docs fixes.

Worth noting the workload it guards against is smaller than when you reviewed: the generator now refuses to write an empty calendar, and the refresh replays onto main instead of rebasing, but the once-a-day commit cadence is unchanged, so the motivation for the exclusion stands.

harryskim: Expanding on why I left this open rather than resolving it with the other seven, since "unresolved" should mean something specific here.

**It is a different kind of claim.** The other seven were decidable from inside the repo. I could stub node-ical to return `{}` and watch the generator refuse to write; I could run both commit-message forms through real git and diff the bodies; I could port the month resolver and run it against the committed data. Each either reproduced or it did not. This one'\''s premise is equally checkable — and it checks out, every `main` SHA has images today and would stop having them — but its *consequence* lives outside the repo boundary. No amount of local grepping settles whether some GitLab job resolves an image by main SHA. I can bound the question; I cannot answer it.

**Resolving it would misrepresent the state.** The other seven have code behind them. Here the code is unchanged and the risk is real but unquantified. Marking it resolved would turn an open question into the appearance of a decision, and the next person reading the thread list would have no way to tell the difference.

**The failure mode argues for caution.** If something out there does resolve by main SHA, it breaks only on docs-only merges — so it is intermittent, it starts days after this lands, and it looks unrelated to a docs PR. That is the expensive kind of regression to trace back, which makes it worth a cheap confirmation now.

**What would close it,** in rough order of preference:

1. Confirmation from the GitLab/deploy owners that nothing resolves an image by `main` SHA. Cheapest, and probably the true answer given nothing in-repo does.
2. Drop the `post-merge-ci.yml` hunk from this PR. It is independent of the docs fixes, so this unblocks the rest immediately; the cost is that a calendar refresh wakes the full GPU pipeline once a day until it lands separately.
3. Keep the exclusion but preserve the invariant: on a docs-only push to main, re-tag the previous SHA'\''s images onto the new SHA. A registry tag copy, no rebuild, so the pipeline stays skipped while "every main SHA has images" stays true. More machinery than this PR warrants, but it is the option that does not require trusting an absence of consumers.

Happy to take any of the three — (2) is a one-commit revert if you would rather not hold the docs fixes on a CI answer.

.github/scripts/generate-events.js
dmitry-tokarev-nv: Verified the guard against the actual code, not just the description — it holds.

- Placed after `body` is assembled but before `fs.writeFileSync`, so a trip leaves the file byte-identical.
- `countExistingEvents()` matching `"addUrl":` is sound: the generator emits it exactly once per event object, the `DynamoEvent` interface declares `addUrl: string;` unquoted so it cannot match, and the banner has none. ENOENT to zero is the right call for a fresh checkout.
- `path.basename(OUT)` is safe, `path` is required at line 24.
- The legitimately-empty case is handled: file present with zero events gives `existingCount === 0`, guard does not fire, nothing to protect.

Agreed the generator is the better home than the workflow — it covers a manual run and reports the VEVENT count directly rather than inferring from a `grep -c` against `HEAD`.

The all-or-nothing floor is the right shape too, for what it is worth. A proportional check ("fewer than last time") would false-positive constantly, since `MAX_UPCOMING`/`MAX_PAST` slicing plus events aging out makes the count legitimately drop. Total emptiness is the only signal that is unambiguous.

**One item from my review is still outstanding** — it was in the review body rather than an inline comment, since there is no diff line to anchor it to, so easy to miss. The banner this script writes into every generated file still reads:

```
 * Source: Dynamo public Google Calendar. Refreshed during Fern publishing.
```

That is no longer true as of this PR; it is refreshed by `community-events-refresh.yml` and committed to `main`. Since the point of the move is that people now read and diff this file on `main`, the header should point at the workflow that actually writes it. Roughly:

```
 * Source: Dynamo public Google Calendar. Refreshed by
 * .github/workflows/community-events-refresh.yml and committed to main.
```

.github/workflows/community-events-refresh.yml
dmitry-tokarev-nv: Confirmed, and I am happy with the guard living in the generator instead — your reasoning about manual runs and the direct VEVENT count is better than my `grep -c` sketch. Details verified on the generator thread.

Also traced the interaction between the guard and the new replay loop, since they touch the same file and the ordering matters: the generator writes (or refuses) first, the early exit at lines 76-79 still short-circuits the no-change case before any of the git work, and only then does the loop run. So a guard trip fails the step before anything is staged, and the last good calendar stays on `main`. That is the behaviour you described and the code does it.

.github/workflows/community-events-refresh.yml
dmitry-tokarev-nv: Confirmed at lines 111-114. Separate `-m` flags, `-s` still lands the trailer last, and nothing carries the block indentation any more.

Thanks for actually running both forms through git rather than reasoning about it — the before/after bodies you pasted are exactly the artifact that settles it. The extra blank line between the two body paragraphs is standard `-m` behaviour and reads fine.

.github/workflows/community-events-refresh.yml
dmitry-tokarev-nv: Replay loop verified line by line — it is correct, and stronger than softening the comment would have been.

Things I specifically checked, since a reset-based loop has several ways to go wrong:

- `generated=$(mktemp)` is taken **before** the loop (line 87), so `git reset --hard` cannot lose the generator's output; line 99 restores it after every reset.
- `git config user.name/email` is set before the loop and lives in `.git/config`, not the tree, so `reset --hard` does not clear it — attempts 2 and 3 still commit as dynamo-ops.
- `reset --hard FETCH_HEAD` moves the local `main` branch rather than detaching, because `actions/checkout` with `ref: main` creates it, so `git push origin HEAD:main` still pushes a branch.
- `if git push ...; then` exempts only the push from `bash -e`. A failing `fetch`, `reset`, or `commit` still aborts the step — which is what you want; only the racing-push case should retry.
- Pushing from the re-shallowed clone is fine: the new commit's parent is the remote tip the fetch just gave us.
- The mid-loop identical-content check (lines 100-103) exits 0 rather than 1 when the merge that beat us already carried the same bytes, so a benign race does not go red.

One residual, which I think is correctly accepted rather than fixed: if `main` moved because something landed a *newer* calendar, the replay overwrites it with our older snapshot. Given `concurrency` serialises refresh runs and any human edit to a generated file is a manual override that the next cron would reconcile anyway, self-correcting within six hours is the proportionate answer. Not asking for a change.

docs/fern/components/EventsCalendar.tsx
dmitry-tokarev-nv: Verified against the code. `resolveCalendarMonth()` does what the table says.

I checked the ordering premise it rests on, since the whole fallback depends on it and the comment asserts it without proof — it holds in the generator: `all` is sorted ascending by `start`, `upcoming` is `filter(>= now).slice(0, MAX_UPCOMING)` so `upcoming[0]` is the soonest, and `past` is `filter(< now).slice(-MAX_PAST).reverse()` so `past[0]` is the most recent. So `[...UPCOMING_EVENTS, ...PAST_EVENTS][0]` really is the nearest event in either direction, and the committed data agrees — `PAST_EVENTS` starts with Jul 30.

The marker gating is right too: `selectedDay` is `null` when the grid is not on today's month, and since the blank cells render through a separate branch that never touches `selectedDay`, a null can never accidentally match a rendered day.

Option (a) was the right pick over navigation — a server component with no client boundary cannot hold that state without dragging `"use client"` back in, which would undo the hydration win from commit 3221a22.

docs/fern/components/CommunityLanding.tsx
dmitry-tokarev-nv: Confirmed — the four-way `isToday` check is exactly what the looser month resolution requires, and routing both grids through `resolveCalendarMonth()` means they cannot drift apart, which was the original point of extracting the helper.

Worth stating the outcome plainly since it is the headline of this PR: with the data actually committed here, this section goes from an empty August grid to July's four events in their day cells. The empty "Up next" beside it is correct and honest — there genuinely are no upcoming events — but the section no longer renders with zero event content, which is what I was worried about.

docs/fern/components/calendar-today.ts
dmitry-tokarev-nv: Both confirmed. `resolveToday(): CalendarDate | null` with the 3-day cutoff, and the build-time `new Date()` fallback is gone from the parse path.

I appreciate you recording the `Date.now()`-at-build-time nuance rather than overselling the guard. That is the honest bound: a stalled refresh is only *detected* on the next rebuild from some other source. It is still strictly better than before, because the wrong-marker window is now capped by whatever else republishes rather than being unbounded.

**One new nit, introduced by 4ca4bba — non-blocking.** This file now hardcodes its own `MONTH_INDEX` (line 25), while `EventsCalendar.tsx:37` and `CommunityLanding.tsx:40` each still derive theirs from a local `MONTHS` array via `slice(0, 3)`. That is three copies of the same abbreviation map, in the file whose own header says it exists so that "the next edit" cannot fix only one of them.

They are identical today, so nothing is broken. The hazard is directional: `resolveCalendarMonth()` picks the month with this file's map, and the components then filter events with theirs, so an edit to one map alone desynchronises the chosen month from the event filter and lands you back at an empty grid — the exact failure this commit fixes. Exporting `MONTH_INDEX` from here and importing it in both components would close it for about three lines.

.github/workflows/post-merge-ci.yml
dmitry-tokarev-nv: Agreed on leaving this open, and the reasoning for why is right — the premise is checkable in-repo and checks out, the consequence is not. Your independent pass matches mine: `trigger_ci.yml` carries the same four patterns, and the only in-repo hits for the SHA-tag families build for their own `github.sha` rather than resolving one someone else built.

Since you asked for a preference among the three, mine is **(1), falling back to (2) quickly** — and I would put a short clock on it.

**(1)** is almost certainly the true answer and costs one reply from whoever owns the GitLab side. Worth asking.

**(2)** is what I would do if that answer is not in hand within a day. The `post-merge-ci.yml` hunk is genuinely independent of the four docs fixes — nothing else in the PR depends on it, and the cost of deferring is one extra post-merge pipeline per day from the calendar commit. That is real GPU time but it is bounded, it is visible, and it is trivially reversible. Holding four verified docs fixes on an unrelated CI question is the worse trade.

**(3)** I would not do. Re-tagging the previous SHA's images onto every docs-only main commit preserves the invariant, but it adds a registry-mutation step to the critical path to defend against a consumer nobody has yet shown exists — and a wrong re-tag is a nastier failure than a missing tag, because it silently serves stale bits under a SHA that implies freshness. Missing-tag fails loudly; wrong-tag does not.

One correction to the framing in your last paragraph: the exclusion's value is not mainly about the calendar commit. Docs-only merges already happen more often than once a day in this repo, and each one currently wakes the full pipeline. The calendar refresh is the trigger that surfaced the question, but the saving is the broader docs-merge case — which is also, incidentally, why (2) costs less than it sounds: deferring it loses a saving you do not have today anyway.

dmitry-tokarev-nv:
Re-reviewed 4ca4bba against the code rather than the commit message. **All seven fixes verified** — details on each thread. The verification in the replies is unusually good; the stubbed empty fetch, the two commit-message forms through real git, and the ported month resolver run against the data this PR actually commits are the right artifacts, and they reproduce.

Two things carried forward. Both sit in threads that are now **resolved and therefore collapsed**, so flagging them here so they are not lost:

1. **[`generate-events.js` banner](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3743695339)** — still unaddressed, and the only item from my original review that no commit has touched. It was in the review body rather than an inline comment (no diff line to anchor to), which is likely why it was missed. Every generated file still carries `Source: ... Refreshed during Fern publishing.`, which this PR makes false. Small, but it is the header of the file the PR is specifically making people read and diff on `main`.

2. **[Third `MONTH_INDEX` copy](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3743695345)** — new in 4ca4bba, non-blocking. `calendar-today.ts:25` hardcodes its own map while `EventsCalendar.tsx:37` and `CommunityLanding.tsx:40` derive theirs. Nothing is broken today; the hazard is that the month is now chosen with one map and the events filtered with another, so editing one alone reproduces the empty grid this commit just fixed.

On [`post-merge-ci.yml`](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3743695347): agreed it should stay open, and the reasoning for keeping it open is right. My preference among the three options is **(1), falling back to (2) on a short clock** — reasoning on the thread. I would not do (3): a wrong re-tag fails silently where a missing tag fails loudly.

Nothing blocking from me beyond the CI question, which is not mine to close.

.github/scripts/generate-events.js
dmitry-tokarev-nv, line 1: **Carry-forward: the generated file's banner is now false.** This is the one item from my original review that no commit has touched — it was in the review body rather than an inline comment because there is no diff line to anchor it to, which is almost certainly why it got missed. Filing it as its own thread so it is trackable rather than buried in the resolved floor-guard thread.

`generate-events.js:156` still writes this into the header of every generated file:

```
 * Source: Dynamo public Google Calendar. Refreshed during Fern publishing.
```

That was accurate when the file was regenerated inside `fern-docs.yml` and never left the runner. As of this PR it is refreshed by `community-events-refresh.yml` and committed to `main`, so the header now points at the wrong mechanism — in the very file this PR is specifically making people read, diff, and review on `main`.

**Suggested change** — replace the single `Source:` line at 156 with:

```js
    ' * Source: Dynamo public Google Calendar.\n' +
    ' * Refreshed by .github/workflows/community-events-refresh.yml and committed to main.\n' +
```

**Best course of action:** fold this into whatever your next push on this branch is. It is a one-line string edit with no behavioural effect, so it needs no re-verification — but it does need a regenerated `events.generated.ts` to actually carry the new banner, and the guard you just added makes that safe to run. If you would rather not re-run the generator against the live calendar just for a comment, editing the committed `events.generated.ts` header by hand to match is equivalent, since the next scheduled refresh will rewrite it identically either way.

Not blocking. But since the whole argument for this PR is "one source of truth that people can read and diff on `main`," the file's own provenance line is worth getting right.

docs/fern/components/calendar-today.ts
```diff
@@ -0,0 +1,140 @@
...
+  year: number;
+  /** Zero-based, matching Date#getMonth and the MONTHS lookups. */
+  month: number;
+  day: number;
+}
+
+export type CalendarMonth = Pick<CalendarDate, "year" | "month">;
+
+/** Maps the abbreviated month names generate-events.js emits ("Jul") to an index. */
+const MONTH_INDEX: Record<string, number> = {
```
dmitry-tokarev-nv: **New in 4ca4bba, non-blocking** — a third copy of the month-abbreviation map, in the file whose stated purpose is preventing exactly this.

The header of this file says it exists because both components "derived the displayed month from `UPCOMING_EVENTS[0] ?? PAST_EVENTS[0]` -- the same defect, copy-pasted -- and pinning both to one helper is what stops the next edit from fixing only one of them." That reasoning applies verbatim to `MONTH_INDEX`, which now exists three times: hardcoded here at line 25, and derived from a local `MONTHS` array at `EventsCalendar.tsx:37` and `CommunityLanding.tsx:40`.

All three agree today, so nothing is broken. The hazard is specifically directional, which is what makes it worth closing rather than noting: `resolveCalendarMonth()` chooses the month using **this** map, and each component then filters its events using **its own**. Those are two different lookups that have to stay in agreement for the grid to show anything. Edit one alone — a locale change, the generator emitting full month names, someone "tidying" the hardcoded literal — and the grid lands on a month whose event filter matches nothing. That is an empty grid: the exact failure this commit just fixed, reintroduced from a different direction and without the fallback catching it, because `resolveCalendarMonth()` would still be returning a confident answer.

```suggestion
/** Maps the abbreviated month names generate-events.js emits ("Jul") to an index. */
export const MONTH_INDEX: Record<string, number> = {
```

**Best course of action** — three lines, all mechanical, and it retires the duplication permanently:

1. Apply the suggestion above (`export`).
2. `EventsCalendar.tsx`: delete lines 37-39 and add `MONTH_INDEX` to the existing `./calendar-today` import.
3. `CommunityLanding.tsx`: delete lines 40-42 and do the same.

Both components keep their local `MONTHS` arrays — those are still needed for the `MONTHS[month]` display label, so this only removes the derived lookup, not the names. Worth confirming `fern check` still passes afterwards, but there is no runtime behaviour change: `Object.fromEntries(MONTHS.map(m => [m.slice(0,3), i]))` produces precisely the twelve keys this literal spells out.

Happy for this to land as a follow-up PR instead if you would rather freeze this one — it is independent of everything else here.

dmitry-tokarev-nv:
**State of play**, superseding my previous comment — the two carry-forwards now have their own threads rather than sitting collapsed inside resolved ones.

**Resolved (7/7 of my original findings).** All verified against `4ca4bba` itself, not the commit message: the floor guard fires before `writeFileSync` and the `"addUrl":` count is unambiguous; the replay loop survives every way a reset-based loop can go wrong (snapshot taken before the loop, `git config` lives outside the tree, `reset --hard` moves the branch rather than detaching, `bash -e` exempts only the push); and `resolveCalendarMonth()`'s fallback rests on an ordering premise that genuinely holds in the generator. Nothing further needed on any of them.

**Open, in the order I would take them:**

1. **[`MONTH_INDEX` third copy](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3744390767)** — three mechanical lines, has a suggestion block, no behaviour change. Cheapest thing on the list and it retires a duplication that can reintroduce the empty grid this PR just fixed.
2. **[Generated-file banner](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3744390718)** — one string edit, but it wants a regenerated `events.generated.ts` to actually carry it. Fold into the same push as (1).
3. **[`post-merge-ci.yml`](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3743695347)** — the only one that is not yours to close. Ask the GitLab/deploy owners; if the answer is not back within a day, drop the hunk and land the docs fixes without it. It is independent of everything else here, so it should not be what holds four verified fixes.

(1) and (2) together are a single small commit. Neither blocks review — if you would rather freeze the branch, both are fine as a follow-up PR, and (3) resolves on its own timeline regardless.

.github/scripts/generate-events.js
harryskim: Fixed in 82bdab7 — you are right that it slipped through, and thanks for re-filing it as its own thread rather than leaving it in the resolved one.

Applied your suggested wording verbatim:

```
 * Source: Dynamo public Google Calendar.
 * Refreshed by .github/workflows/community-events-refresh.yml and committed to main.
```

On regenerating: I took the hand-edit route you sanctioned, but mechanically rather than by retyping, since a near-miss here would show up as a spurious header diff on the next scheduled run. I evaluated the generator's own `banner` expression out of the script, spliced the result into the committed `events.generated.ts`, then byte-compared the file's header against that same expression:

```
IDENTICAL — next scheduled refresh will produce no banner diff
```

Event data is untouched (`GENERATED_ON=2026-08-07`, 0 upcoming, 4 past), and I re-ran the month-resolution and floor-guard checks afterwards to confirm the splice changed nothing but the header.

docs/fern/components/calendar-today.ts
harryskim: Fair hit, and the directional argument is the convincing part — I had read the duplication as benign because all three agreed, which misses that `resolveCalendarMonth()` and the per-component event filters are two lookups that must stay in agreement for the grid to render anything. A single-copy edit lands on a month whose filter matches nothing, and the fallback cannot catch it because the resolver is still returning a confident answer. That is the empty grid again from a direction this branch does not otherwise defend.

Done in 82bdab7, all three steps:

1. `MONTH_INDEX` is exported from `calendar-today.ts`, with a comment recording why it has to be the only copy.
2. `EventsCalendar.tsx`: local derivation deleted, added to the existing `./calendar-today` import.
3. `CommunityLanding.tsx`: same.

`MONTHS` stays local to each component — it is the full-name list used for display (`MONTHS[month]`), not a lookup, so it carries none of the same hazard. Verified there is now exactly one definition:

```
calendar-today.ts:34:  export const MONTH_INDEX: Record<string, number> = {
CommunityLanding.tsx:16   MONTH_INDEX,          <- import
EventsCalendar.tsx:18     MONTH_INDEX,          <- import
```

dmitry-tokarev-nv:
Reviewed `82bdab7`. **Both follow-ups verified — all nine of my findings are now closed**, leaving only the CI-ownership question.

**Banner.** I re-derived it independently rather than eyeballing the diff: parsed the `banner` expression out of `generate-events.js`, evaluated the concatenated string literals, and byte-compared against the committed `events.generated.ts` header.

```
banner bytes: 363
committed header matches generator byte-for-byte: True
```

So the splice is exact — em-dash and all — and the next scheduled refresh will produce no spurious header diff. That was the specific risk in hand-editing a generated file, and it is closed. Doing it mechanically from the generator's own expression rather than retyping was the right instinct.

**`MONTH_INDEX`.** Exactly one definition remains (`calendar-today.ts:34`, exported), both components import it, and both still use their local `MONTHS` for the display label (`EventsCalendar.tsx:149`, `CommunityLanding.tsx:173,201`) — so no unused-variable fallout from dropping the derivations. The comment recording *why* it must stay single-copy is the part that will actually hold the line.

---

**CI is red for a reason unrelated to this PR — a re-run should clear it.**

`Docs Website Composition Check` fails at step 6, `Install yq`, in [`pre-merge.yml:114`](https://github.com/ai-dynamo/dynamo/blob/main/.github/workflows/pre-merge.yml#L114):

```
wget -qO "$RUNNER_TEMP/bin/yq" https://github.com/mikefarah/yq/releases/download/v4.52.5/yq_linux_amd64
##[error]Process completed with exit code 4.
```

`wget` exit 4 is *Network failure* — the runner could not reach the GitHub releases CDN. Step 7, `Replay the docs-website composition`, was **skipped**, so the check never executed the thing it validates. `pre-merge-status-check` is red only because it aggregates.

Nothing in `82bdab7` touches that workflow or anything it fetches. **Recommended action: just re-run the failed jobs.** Do not go looking for a docs-composition problem — there is no signal in that run either way.

Worth someone filing separately (not this PR): that `wget` has no `--tries`/`--retry-connrefused`/`--retry-on-host-error`, so any transient CDN blip hard-fails a required pre-merge check and skips the real validation. A retry flag or pinning `yq` through a cached action would remove a recurring red herring for everyone.

---

**Remaining:** only [`post-merge-ci.yml`](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3743695347), which is the CI owners' call rather than yours. My recommendation there is unchanged — ask, and if the answer is not back within a day, drop the hunk and land the docs fixes without it.

.github/scripts/generate-events.js
dmitry-tokarev-nv: Verified independently rather than reading the diff, since "byte-identical" is the whole claim: I parsed the `banner` expression out of `generate-events.js`, evaluated the concatenated string literals, and compared against the committed file's header.

```
banner bytes: 363
committed header matches generator byte-for-byte: True
```

Exact, em-dash included — so the next scheduled refresh produces no header diff. Deriving the splice from the generator's own expression instead of retyping is what made that come out clean; a retyped version would very likely have differed on the em-dash or the line break and shown up as a phantom diff on the first cron run.

docs/fern/components/calendar-today.ts
dmitry-tokarev-nv: Confirmed: one definition (`calendar-today.ts:34`, exported), both components importing it, and `MONTHS` still referenced in both for the display label (`EventsCalendar.tsx:149`, `CommunityLanding.tsx:173` and `:201`) — so no unused-variable fallout from dropping the derivations.

The comment you added recording *why* it has to stay single-copy is the part that will actually hold. The export alone would have been undone by the next person who wanted a local lookup.

docs/fern/components/EventsCalendar.tsx
```diff
@@ -142,7 +161,15 @@ export function EventsCalendar() {
...
+                      day === selectedDay ? "is-selected" : "",
+                      eventDays.has(day) ? "has-event" : "",
+                    ]
+                      .filter(Boolean)
+                      .join(" ") || undefined
+                  }
+                  aria-current={day === selectedDay ? "date" : undefined}
                 >
                   {day}
                 </span>
```
dmitry-tokarev-nv, line 180: **New, non-blocking — the event dot has no text alternative.**

Not a regression from this PR, but this commit is what makes the dot load-bearing, so it is the right moment to raise it. Before, `is-selected` marked the next event, so the one thing the grid conveyed was reachable by *some* users. Now `is-selected` means today, and event information on this grid lives entirely in `.has-event::after` — a 3px decorative pseudo-element with no accessible name.

The surrounding markup closes off the alternatives: `.dynamo-calendar__weekdays` is `aria-hidden`, the cells are bare `<span>`s in a plain `<div>` with no grid semantics, and the only label is the container's "Calendar month preview". So a screen-reader user gets an undifferentiated run of numbers with no way to tell which days have events.

Worth noting the Community grid does not have this problem — its event days render real `<a>` elements carrying the title and time, so they are announced. This is specific to the Home preview.

```suggestion
                <span
                  key={day}
                  className={
                    [
                      day === selectedDay ? "is-selected" : "",
                      eventDays.has(day) ? "has-event" : "",
                    ]
                      .filter(Boolean)
                      .join(" ") || undefined
                  }
                  aria-current={day === selectedDay ? "date" : undefined}
                  aria-label={
                    eventDays.has(day)
                      ? `${MONTHS[month]} ${day}, has an event`
                      : undefined
                  }
                >
                  {day}
                </span>
```

`MONTHS` is already in scope at line 26 and used at 149, so this needs no new import. `aria-current="date"` keeps carrying "today" on its own, so the two do not collide when a day is both.

**Best course of action:** fold it in if you are pushing again anyway; otherwise a follow-up is entirely reasonable. I would not hold the PR for it — it is a pre-existing gap that this change merely makes more visible.

dmitry-tokarev-nv:
One operational finding that is worth acting on **at merge time**, plus a heads-up on what the site will look like the moment this lands.

**`GENERATED_ON` will already be stale when this merges.** The committed value is `2026-08-07` and `STALE_AFTER_DAYS` is 3, so from roughly 2026-08-11 onward `resolveToday()` returns null. That is correct behaviour, and exactly what I asked for — but it means the freshly-merged site renders:

- **July 2026** in both grids (today's month has no events, so the nearest-event fallback takes over — showing July's four events, which is the good outcome), and
- **no today marker at all**, because the marker is deliberately suppressed once the stamp is stale.

That will sit there until the first scheduled refresh, so up to six hours. Nothing is broken, but "the calendar says July and nothing is highlighted" is very easy to mistake for a regression from this PR, especially by someone who was not in these threads.

**Recommended: `workflow_dispatch` `community-events-refresh.yml` immediately after merge.** It already carries the trigger. That rewrites `GENERATED_ON` to the merge date, and both grids snap to the correct month and marker within one publish rather than up to six hours.

It also buys the thing your PR description flagged as unverified — *"everything CI-side is static validation ... The first scheduled run after merge is the real test."* A manual dispatch **is** that test, on demand, while you are watching, rather than at 00:00 Pacific with nobody looking. It exercises the whole untested chain end to end: the `OPS_BOT_PAT` checkout, the ruleset bypass for `dynamo-ops`, the DCO trailer, the replay loop, the push to `main`, and the `fern-docs.yml` hand-off. If the bypass was never actually granted, you find out in a minute from a red run you triggered, instead of from a silent six-hourly failure.

Cheap, and it converts the single largest unverified assumption in this PR into an observed one.

Also filed: [one non-blocking a11y note](https://github.com/ai-dynamo/dynamo/pull/12861#discussion_r3745587592) on the Home grid — the event dot has no text alternative. Pre-existing, but this commit is what makes the dot the sole carrier of event information. Suggestion block attached; fine as a follow-up.

dmitry-tokarev-nv (approved):
critical review comments were addressed. Approving to unblock. Address CI failures.

.github/workflows/post-merge-ci.yml
harryskim: Moved out of this PR — the hunk is now **#12912**, and `post-merge-ci.yml` is untouched here (1bdaa51). This PR is docs-only again.

Your question is not closed, just relocated somewhere it can be answered on its own timeline: it carries the in-repo findings (yours and my independent pass), the pattern-parity check, and an explicit "what is still unknown" section asking whether anything out of repo resolves an image by main SHA.

Two things I added while writing it up that strengthen the standalone case:

- **The exclusion is already worth having, independent of the calendar bot.** 4 of the last 30 post-merge-ci runs on main were docs-only commits — each a full ~110-minute image build. One was `docs: point the Fern CLI link at the CLI reference` at 110 minutes. So this is fixing an existing cost, not just pre-empting a new one.
- **A fallback if a consumer does turn up:** keep the pipeline skipped but preserve the invariant by re-tagging the previous SHA's images onto the new SHA. A registry tag copy, no rebuild.

Cost of the split, stated plainly: until #12912 lands, the daily calendar commit wakes the ~110-minute pipeline once a day.

Resolving here since it no longer applies to this PR — discussion continues on #12912.

Unrelated but worth flagging since it surfaced while gathering the above: all 30 of the most recent post-merge-ci runs on main failed. Predates both PRs; noted in #12912 rather than acted on.

docs/fern/components/EventsCalendar.tsx
harryskim: Folded in — applied verbatim in e9e39f4, since I was pushing anyway.

The framing is right and worth restating for the record: this is not a regression from the PR, but it is the commit that makes it matter. `is-selected` used to mark the next event, so the grid's one signal was reachable; now it marks today, and the only event information left on the Home preview is the decorative `.has-event::after` dot with no accessible name. With the weekday row `aria-hidden` and the cells bare spans, that left a screen reader announcing an undifferentiated run of numbers.

Confirmed `MONTHS`, `month`, and `eventDays` are all already in scope at that point (lines 26, 105, 115), so no new import was needed. `aria-current="date"` still carries "today" independently, so a day that is both announces both.

Agreed the Community grid is unaffected — its event days are real anchors carrying title and time.

harryskim:
All threads closed — **16/16 resolved, 0 open**, and the PR is docs-only again. Thanks @dmitry-tokarev-nv; independently re-deriving the banner byte-comparison rather than eyeballing the diff is a good bit sharper than the review needed to be.

**Since your last pass:**

- **a11y note** — folded in verbatim (`e9e39f4`). Confirmed `MONTHS`, `month`, and `eventDays` were already in scope, so no new import, and `aria-current="date"` still carries "today" independently of the new label.
- **`post-merge-ci.yml` hunk** — split out to **#12912** (`1bdaa51`), so this PR no longer touches anything outside `docs/` and does not wait on the CI-ownership question. Two things I added there that strengthen the standalone case: docs-only commits were **4 of the last 30** post-merge-ci runs on main (each a full ~110-minute image build, so the exclusion fixes an existing cost rather than a hypothetical one), and a fallback if an out-of-repo SHA consumer does turn up — re-tag the previous SHA's images onto the docs-only SHA, a registry tag copy with no rebuild.
- **CI blip** — the two failed jobs were `Install yq` in *Docs Website Composition Check* plus `pre-merge-status-check` failing downstream of it. `Install yq` succeeds on the re-run. (One intermediate run reads "cancelled" — that is a later push superseding it, not a failure.)

**On the merge-time dispatch: agreed, and it is on the merge checklist.**

Your framing is the persuasive part — it is not just cosmetic cleanup of a stale `GENERATED_ON`, it is the only on-demand exercise of the chain this PR cannot test statically: `OPS_BOT_PAT` checkout, the `dynamo-ops` ruleset bypass, the DCO trailer, the replay loop, the push to `main`, and the `fern-docs.yml` hand-off. Discovering a missing bypass from a run someone triggered while watching beats discovering it from a silent six-hourly failure.

Worth noting the stale-stamp behaviour you describe is now **guaranteed** rather than likely: committed `GENERATED_ON` is `2026-08-07`, so the 3-day window has already lapsed. On merge the site will show **July 2026 in both grids with no today marker** until that first refresh. That is the guard working as designed, but as you say it reads like a regression to anyone who was not in these threads — so the dispatch should be immediate, not eventual.

Ready to merge from my side.

PR #12868, opened 2026-08-08
test(efa): add EFA packaging checks to the nightly -efa image jobs

jh-nv:
## Summary

Nightly and post-merge both build **six** `-efa` runtime image variants (vllm/sglang/trtllm × amd64/arm64), and nothing verifies the EFA packaging of any of them. The existing `vllm-efa-test` / `trtllm-efa-test` post-merge jobs run `'(pre_merge or post_merge) and <framework> and gpu_0'` — character-for-character the same markers as their non-EFA counterparts, touching nothing EFA-specific. Nightly has no EFA test jobs at all.

Every EFA regression we have hit was a *packaging* regression that no functional test could see until it reached a p5/GB200 cluster, and each degrades to a **silent UCX fallback** rather than a hard failure:

- an image shipping a NIXL plugin directory with no LIBFABRIC plugin
- NIXL linked against the generic from-source libfabric instead of the EFA SDK's, so HMEM/CUDA dmabuf descriptors do not match at runtime
- arm64 putting plugins somewhere other than where `NIXL_PLUGIN_DIR` points

### Changes

- **`tests/container/test_efa_image.py`** (new, marker `efa_image`) — six checks: EFA SDK installed · `EFA_VERSION` recorded · `efa` provider compiled into libfabric (`fi_info -l`) · `NIXL_PLUGIN_DIR` holds `libplugin_LIBFABRIC.so` · that plugin's `ldd` resolves libfabric under `/opt/amazon/efa` · `libnixl.so` sits beside the plugins and every `LD_PRELOAD` entry resolves. No GPU, no EFA device, no cluster; a few seconds per image.
- **`pyproject.toml`** — register the `efa_image` marker.
- **`.github/workflows/nightly-ci.yml`** — add `vllm-efa-test`, `sglang-efa-test`, `trtllm-efa-test`, each running only `efa_image and gpu_0` against its nightly `-efa` image on amd64 and arm64.

**Cadence:** nightly, not per-commit. EFA changes are sparse, so paying for these on every push to `main` costs more CI than the risk warrants. `post-merge-ci.yml` is deliberately left untouched by this PR.

Paths are read from the environment rather than hardcoded, because they genuinely differ: `vllm-runtime:1.3.1-efa` uses `NIXL_PLUGIN_DIR=/opt/nvidia/nvda_nixl/lib64/plugins` on amd64 but `/opt/dynamo/nixl/plugins` on **arm64**. A hardcoded `lib64` assertion would have failed on arm64 — which is the very bug class this test exists to catch. `NIXL_LIB_DIR` is also not exported by every runtime stage (`trtllm_runtime.Dockerfile` sets only `NIXL_PLUGIN_DIR` and `LD_PRELOAD`), so the libnixl check falls back to the plugin directory's parent rather than going vacuous.

## Validation

**Executed in CI-built EFA images.** Before the switch to nightly, this branch's post-merge variant was dispatched on `pull-request/12868` ([run 31238756257](https://github.com/ai-dynamo/dynamo/actions/runs/31238756257)); `vllm-runtime-efa` built and both CPU Test jobs passed:

| Job | Summary |
|---|---|
| amd64 | 1110 passed, 9 skipped — 0 failed, 0 errors |
| arm64 | 1110 passed, 10 skipped — 0 failed, 0 errors |

All six tests appear by name in both logs with zero FAILED/ERROR lines. Only the wiring changed afterwards; the test file's assertions are unchanged.

**Executed in released images on the AWS dev clusters** (mounted at `/workspace/tests/container`, run with the image's own Python, in pods holding **no GPU and no EFA device** — the same condition as the CI runner):

| Image | Result |
|---|---|
| `vllm-runtime:1.3.1-efa` — amd64 | **6 passed** |
| `vllm-runtime:1.3.1-efa` — **arm64** | **6 passed** |
| `sglang-runtime:1.3.1-efa` — amd64 | **6 passed** |
| `sglang-runtime:1.3.1` (non-EFA, negative control) | **5 failed, 1 skipped** |
| `vllm-runtime:1.3.1` (non-EFA, negative control) | **4 failed** — including the libfabric-linkage check |

Repo-level: `pytest --collect-only` under `--strict-markers` collects 6, and `-m 'efa_image and gpu_0'` selects all 6; `tests/report_pytest_markers.py` reports **`Missing sets: 0`** across 4660 tests; isort / black / flake8 / ruff / codespell / check-yaml / check-toml pass; both workflow files parse.

### Known gaps

- `nvcr.io/nvidia/ai-dynamo/trtllm-runtime` is not publicly pullable, so the TRT-LLM EFA image was not exercised directly. By inspection, `aws.Dockerfile`'s trtllm branch sets `NIXL_PLUGIN_DIR=/opt/nvidia/nvda_nixl/lib64/plugins` and `LD_PRELOAD=/opt/nvidia/nvda_nixl/lib64/libnixl.so`, which both checks handle — worth watching the first nightly `trtllm-efa-test`.
- These checks cover *packaging*, not function: no RDMA, no memory registration, no KV transfer. Functional EFA validation on real p5/GB200 clusters is the follow-up.
- The `pytest-marker-report` pre-commit hook was skipped at commit time: its hook venv runs on this machine's Python 3.9, which cannot import `tests/utils/collection_env_guard.py` (`str | None`). The same script was run under Python 3.12 — `Missing sets: 0`.

An internal Linear ticket also tracks this work.

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -913,6 +913,79 @@ jobs:
     secrets: inherit
 
   # ============================================================================
+  # EFA IMAGE TEST JOBS
+  #
+  # The -efa images are built here but nothing verified their EFA packaging.
+  # These jobs run only the efa_image checks (tests/container/test_efa_image.py):
+  # no GPU, no EFA device and no cluster, a few seconds per image. EFA changes
+  # are sparse, so nightly is the right cadence -- per-commit would spend far
+  # more than the risk warrants.
+  # ============================================================================
+
+  vllm-efa-test:
+    name: vllm-runtime-efa
+    needs: [vllm-efa-build, resolve-source-sha, compute-release-mode]
+    if: ${{ needs.compute-release-mode.outputs.run_tests == 'true' }}
+    uses: ./.github/workflows/shared-test.yml
+    with:
+      dd_env: nightly
+      dd_flaky_retry_enabled: 'false'
+      test_suite_name: vllm-efa
+      test_type: CPU Test
+      amd_runner: prod-tester-amd-gpu-4-v2
+      target_tag_plain: ${{ needs.vllm-efa-build.outputs.target_tag_plain }}
+      cuda_version: '["13.0"]'
+      platform: '["amd64", "arm64"]'
+      run_cpu_only_tests: true
+      cpu_only_test_markers: efa_image and gpu_0
+      run_gpu_tests: false
+      source_ref: ${{ needs.resolve-source-sha.outputs.source_sha }}
+      image_tag_suffix: '-nightly'
+    secrets: inherit
+
+  sglang-efa-test:
+    name: sglang-runtime-efa
+    needs: [sglang-efa-build, resolve-source-sha, compute-release-mode]
+    if: ${{ needs.compute-release-mode.outputs.run_tests == 'true' }}
+    uses: ./.github/workflows/shared-test.yml
+    with:
+      dd_env: nightly
+      dd_flaky_retry_enabled: 'false'
+      test_suite_name: sglang-efa
+      test_type: CPU Test
+      amd_runner: prod-tester-amd-gpu-4-v2
+      target_tag_plain: ${{ needs.sglang-efa-build.outputs.target_tag_plain }}
+      cuda_version: '["13.0"]'
+      platform: '["amd64", "arm64"]'
+      run_cpu_only_tests: true
+      cpu_only_test_markers: efa_image and gpu_0
+      run_gpu_tests: false
+      source_ref: ${{ needs.resolve-source-sha.outputs.source_sha }}
+      image_tag_suffix: '-nightly'
+    secrets: inherit
+
+  trtllm-efa-test:
+    name: trtllm-runtime-efa
+    needs: [trtllm-efa-build, resolve-source-sha, compute-release-mode]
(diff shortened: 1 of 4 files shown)


jh-nv:
Closing in favour of the end-to-end test in #12915.

The static checks here would **not** have caught `d4bfe441ae`, the regression they were written for. `test_libfabric_plugin_links_efa_sdk` verifies `ldd` resolution, and that turns out to be near-tautological on any EFA image: the SGLang image's LIBFABRIC plugin comes from a PyPI wheel that was certainly never built against `/opt/amazon/efa`, yet `ldd` still resolves it there, because the EFA installer runs `ldconfig` and makes that SONAME the system-wide answer for every consumer.

What remains is the layout assertions, and those are brittle by nature. SGLang demonstrated it twice: NIXL lives at a mesonpy wheel path rather than `/opt/nvidia/nvda_nixl`, and that changed between the 1.3.1 release and current main. The image built from main fails 2 of 6 checks (no `NIXL_PLUGIN_DIR` at all), so `sglang-efa-test` would have been red every night.

#12915 asserts the outcome instead — KV bytes moved, libfabric memory registration happened, no UCX fallback — which is invariant to where the files live. The diagnostic value is carried over there as logging rather than assertions.

Thanks to the reviewers; every one of the findings held up under checking.

Filed separately: EFA release staging is not gated on any test result, and build-time libfabric provenance (`-Dlibfabric_path` is still `/usr/local/libfabric` on main).

PR #12869, opened 2026-08-08
feat(snapshot): promote a snapshot-restored shadow engine on worker failure

galletas1712:
## Summary

Dynamo can pair each inference worker with a **shadow engine** — a second engine
on the same GPU, restored from a snapshot so the model weights are already
resident. When the active worker dies, the shadow is promoted in seconds rather
than paying a full cold start.

The restore path had no promotion step. Every restored shadow resumed and
registered with the frontend as soon as its snapshot loaded, so several engines
could claim the same GPU and the frontend could route to one that wasn't serving.

Restored shadows now stay paused until they elect a single winner. The winner
resumes and registers; the rest wait as standbys. Snapshot restores outside a
failover setup are unaffected.

## Changes

- `common/snapshot/lifecycle.py` — the restore sentinel returns paused instead of
  resuming; adds `wake_restored_engine()` (health → flock → resume)
- `common/snapshot/constants.py` — carry `ENGINE_ID`, `FAILOVER_LOCK_PATH`,
  `DYN_VLLM_GMS_SHADOW_MODE`, `DYN_GMS_USE_V1` through restore
- `vllm/main.py`, `sglang/main.py`, `trtllm/snapshot.py` — call the new hook
- `vllm/worker_factory.py` — skip the cold-start election on the snapshot path so
  a restored engine does not elect twice

The existing cold-start shadow path is untouched. Operator orchestration (#12874)
and SGLang failover (#13276) build on this.


galletas1712:


components/src/dynamo/common/snapshot/lifecycle.py
```diff
@@ -175,3 +172,50 @@ async def wait_for_restore(self) -> bool:
...
+    try:
+        await pause_controller.resume()
+        pause_controller.mark_resumed()
+    except Exception:
+        if lock is not None:
+            logger.critical(
+                "[Shadow][snapshot] Engine wake failed after lock acquisition; "
+                "terminating process while retaining the lock"
+            )
+            os._exit(1)
```
tanmayv25: This appears unnecessary and harmful.

`FlockFailoverLock` is released by the kernel when the process exits because the lock is tied to its file descriptor; there is no destructor-based release path, and this failure path does not explicitly call release(). Raising normally should therefore preserve the lock until process termination just as effectively as a hard exit.

Using os._exit(1) also bypasses normal shutdown handling, including atexit callbacks and log flushing, so the preceding logger.critical(...) may be lost. Please let the exception propagate instead. This would also align the restored-snapshot path with the cold-start path, where resume() failures are allowed to propagate normally rather than forcing an immediate process exit.

components/src/dynamo/common/snapshot/lifecycle.py
```diff
@@ -175,3 +172,50 @@ async def wait_for_restore(self) -> bool:
...
+        lock_path = os.environ.get("FAILOVER_LOCK_PATH")
+
+    lock = None
+    if lock_path:
+        if runtime is not None:
+            # Healthy while paused, so Kubernetes does not kill the standby
+            # during a long outage. Nothing watches it for death until the
+            # monitor is built with the handler; an operator-side liveness
+            # probe is the durable fix.
+            runtime.set_health_status(True)
```
tanmayv25, line 208: runtime.set_health_status(True) is called before lock.acquire(), but acquire() can block indefinitely while waiting for the previous lock holder to exit or release the lock. During that wait, the standby reports itself as healthy even though its engine is not yet supervised by the death monitor.

The comment acknowledges this gap and relies on an external liveness probe to catch an engine failure. Instead, start the engine death monitor before the blocking lock acquisition. That way, if the standby engine dies while waiting for the predecessor’s lock, the failure is detected and handled in-process rather than relying on operator-side polling.

tanmayv25 (commented):
wake_restored_engine substantially duplicates _maybe_wait_for_failover_lock: both use the same flock and engine-{ENGINE_ID} key, then follow the same set-healthy → acquire → resume flow. The snapshot path disables the worker-factory helper specifically to avoid running the same election twice.

Could we consolidate this into a single failover-election helper used by both paths, with the only difference being whether it runs before or after runtime creation? Keeping two implementations creates avoidable drift, and they have already diverged: the snapshot-specific version adds os._exit(1) and lacks in-process engine-death monitoring during lock acquisition.

The pause/wake design itself looks sound, but I suggest:

- Let resume failures raise normally rather than calling os._exit(1).

- Start engine-death monitoring before the potentially blocking lock acquisition.

- Reuse one election helper for both cold-start and restored-snapshot flows so these behaviors stay consistent.

components/src/dynamo/common/snapshot/lifecycle.py
mohammedabdulwahhab: Agreed with this. Failure in wake now raises consistent with cold start case.

components/src/dynamo/common/snapshot/lifecycle.py
mohammedabdulwahhab: For now I've left a TODO in `elect_and_wake` at this site to note the limitation. I think the work to plumb in an engine monitor here is a bit more involved than I'd like to introduce into this PR. This limitation already exists on the cold-start path so there is no new regression being introduced as part of this change.

mohammedabdulwahhab:
@tanmayv25 Thanks for the review!

`_maybe_wait_for_failover_lock` is now a thin vLLM adapter around the engine-agnostic `elect_and_wake` utility standardizing the cold start and snapshot paths.
- the os._exit() has been removed and now failures raise
- I've added a TODO for the engine monitor wiring. Fow now, my opinion is that it will have some too much sprawl in this PR for a pre-existing limitation. Will address in a follow up

tanmayv25 (approved):

mohammedabdulwahhab (approved):

ayushag-nv (approved):

PR #12874, opened 2026-08-08
feat(operator): allow snapshot-backed intra-pod failover

galletas1712:
## Summary

Allow a DGD component to combine an automatic checkpoint with intra-pod
failover, so a snapshot restores into both engine containers.

Both halves already exist. `buildFailoverPod` clones `main` into `engine-0` and
`engine-1`, and the checkpoint path already targets the engine containers for
restore. What blocks the combination is a single compatibility rule:

```go
// internal/checkpoint/compatibility.go
- if experimental.Failover != nil {
-     violations = append(violations, errors.New("Snapshot with active/passive failover is temporarily unsupported"))
- }
```

Removing it is the change. The rest of the PR is the validation that makes the
combination safe, plus two small pieces of operator wiring.

## What is validated

Two engine containers share one GPU, so the deployment must resolve to a
single-rank aggregated worker. `ValidateFailoverCheckpointForDGD` and
`ValidateFailoverCheckpointForDCD` reject anything else at admission and again
at reconcile, rather than letting it fail at runtime:

- `checkpoint.mode` automatic, `deletionPolicy` Delete, `targetContainerName` main
- `gpuMemoryService.mode` and `failover.mode` IntraPod, component type Worker
- disaggregation off, request plane tcp, TP/PP/DP = 1
- `--worker-cls`, `--load-format`, `DYN_VLLM_GMS_SHADOW_MODE` and
  `DYN_FORWARDPASS_METRIC_PORT` unset — all four are operator-managed

Supported for vLLM and SGLang.

## Operator wiring

- Unset `DYN_VLLM_GMS_SHADOW_MODE` on snapshot failover engines. Shadow mode is
  incompatible with failover + snapshot, which runs on GMS V1, and
  `applyVLLMOverrides` sets it unconditionally for intra-pod failover.
- Strip `DYN_FORWARDPASS_METRIC_PORT` from the capture container. A capture pod
  has no Dynamo endpoint to publish to. This cannot be validated away, because
  the operator injects the port during pod generation, after validation runs.

## Commits

1. the feature, rebased onto the standalone Snapshot migration
2. validation in place of config rewriting, and the shadow-mode gate
3. SGLang support, from #13276

## Validation

`go build ./...`, `go vet ./internal/...`, and the `internal/dynamo`,
`internal/checkpoint` and `internal/webhook` suites pass. The `*_envtest_test.go`
suites need a local control plane and are covered by CI.

### DIFF
--- a/deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
+++ b/deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
@@ -266,7 +266,7 @@ func (r *dgdCheckpointsReconciler) reconcileAutomaticSnapshotJob(
 		return nil, err
 	}
 	if dynamo.IsIntraPodFailoverEnabled(component) {
-		if err := dynamo.PrepareVLLMAutomaticFailoverSnapshotSource(targetContainer); err != nil {
+		if err := dynamo.PrepareVLLMSnapshotSourceContainer(targetContainer); err != nil {
 			return nil, err
 		}
 	}
--- a/deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
+++ b/deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
@@ -265,6 +265,11 @@ func (r *dgdCheckpointsReconciler) reconcileAutomaticSnapshotJob(
 	if err != nil {
 		return nil, err
 	}
+	if dynamo.IsIntraPodFailoverEnabled(component) {
+		if err := dynamo.PrepareVLLMAutomaticFailoverSnapshotSource(targetContainer); err != nil {
+			return nil, err
+		}
+	}
 	var gmsSpec *nvidiacomv1alpha1.GPUMemoryServiceSpec
 	if converted := gms.ToAlphaSpec(dynamo.GetGPUMemoryService(component)); converted != nil {
 		gmsSpec = converted.DeepCopy()
--- a/deploy/operator/internal/controller/dynamographdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeployment_controller.go
@@ -163,7 +163,7 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 				fmt.Errorf("component %q: %w", component.ComponentName, compatibilityErr),
 			)
 		}
-		for _, compatibilityErr := range dynamo.ValidateAutomaticFailoverCheckpointSource(
+		for _, compatibilityErr := range dynamo.ValidateFailoverCheckpointForDGD(
 			component,
 			dynamoDeployment.Spec.BackendFramework,
 		) {
--- a/deploy/operator/internal/controller/dynamographdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeployment_controller.go
@@ -163,6 +163,15 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 				fmt.Errorf("component %q: %w", component.ComponentName, compatibilityErr),
 			)
 		}
+		for _, compatibilityErr := range dynamo.ValidateAutomaticFailoverCheckpointSource(
+			component,
+			dynamoDeployment.Spec.BackendFramework,
+		) {
+			compatibilityErrs = append(
+				compatibilityErrs,
+				fmt.Errorf("component %q: %w", component.ComponentName, compatibilityErr),
+			)
+		}
 	}
 	if compatibilityErr := errors.Join(compatibilityErrs...); compatibilityErr != nil {
 		programResult := newWorkloadProgramResult(dynamoDeployment)
--- a/deploy/operator/internal/dynamo/failover.go
+++ b/deploy/operator/internal/dynamo/failover.go
@@ -412,7 +412,7 @@ const (
 	vllmModuleName                         = "dynamo.vllm"
 	vllmLoadFormatFlag                     = "--load-format"
 	vllmWorkerClassFlag                    = "--worker-cls"
-	checkpointFailoverCompatibilityMessage = "Snapshot with active/passive failover requires an operator-managed automatic single-node vLLM Worker checkpoint"
+	checkpointFailoverCompatibilityMessage = "Snapshot with active/passive failover requires an operator-managed automatic single-node Worker checkpoint"
 )
 
 // ValidateAutomaticFailoverCheckpointSource validates the DGD component that
@@ -473,8 +473,9 @@ func validateFailoverCheckpointProfile(
 	if config.TargetContainerName != "" && config.TargetContainerName != commonconsts.MainContainerName {
 		violations = append(violations, errors.New("targetContainerName must be main"))
 	}
-	if BackendFramework(backendFramework) != BackendFrameworkVLLM {
-		violations = append(violations, errors.New("backendFramework must be vllm"))
+	backend := BackendFramework(backendFramework)
+	if backend != BackendFrameworkVLLM && backend != BackendFrameworkSGLang {
+		violations = append(violations, errors.New("backendFramework must be vllm or sglang"))
 	}
 	if component.ComponentType != v1beta1.ComponentTypeWorker {
 		violations = append(violations, errors.New("component type must be Worker"))
@@ -504,7 +505,7 @@ func validateFailoverCheckpointProfile(
 	// The operator derives these from DYN_GMS_USE_V1, which it injects when a
 	// checkpoint is configured. A user value is either ignored or contradicts
 	// the operator's selection, so reject it rather than silently rewriting it.
-	for _, flag := range []string{vllmWorkerClassFlag, vllmLoadFormatFlag} {
+	for _, flag := range vllmOperatorManagedFlags(backend) {
 		if _, _, _, found, err := tokenizedVLLMFlag(main.Args, flag); err != nil {
 			violations = append(violations, err)
 		} else if found {
@@ -518,18 +519,7 @@ func validateFailoverCheckpointProfile(
 			violations = append(violations, fmt.Errorf("%s is managed by the operator and must not be set", name))
 		}
 	}
-	for _, profile := range []struct {
-		flag         string
-		defaultValue string
-		want         string
-		description  string
-	}{
-		{flag: "--disaggregation-mode", defaultValue: "agg", want: "agg", description: "disaggregation mode must be aggregated"},
-		{flag: "--request-plane", defaultValue: "tcp", want: "tcp", description: "request plane must be tcp"},
-		{flag: tensorParallelSizeFlag, defaultValue: "1", want: "1", description: "tensor parallel size must be 1"},
-		{flag: pipelineParallelSizeFlag, defaultValue: "1", want: "1", description: "pipeline parallel size must be 1"},
-		{flag: dataParallelSizeFlag, defaultValue: "1", want: "1", description: "data parallel size must be 1"},
-	} {
+	for _, profile := range failoverSnapshotFlagProfile(backend) {
 		value, _, _, found, err := tokenizedVLLMFlag(main.Args, profile.flag)
 		if err != nil {
 			violations = append(violations, err)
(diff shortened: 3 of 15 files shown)


galletas1712:


deploy/operator/api/v1alpha1/common.go
mohammedabdulwahhab, line 1: Why did we introduce the notion of more than one shadow engine container? It's adding to the diff in this PR

deploy/operator/api/v1alpha1/dynamocheckpoint_types.go
mohammedabdulwahhab, line 1: I assume this will fall out of the diff with the removal of DynamoCheckpoint

deploy/operator/api/v1alpha1/dynamocomponentdeployment_types.go
mohammedabdulwahhab, line 1: Why is this diff still here?

deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
```diff
@@ -154,7 +154,7 @@ func (r *dgdCheckpointsReconciler) Reconcile(
 			info.RestoreTargetContainers = []string{alphaCheckpointConfig.TargetContainerName}
 		}
 		if dynamo.IsIntraPodFailoverEnabled(component) {
-			info.RestoreTargetContainers = dynamo.IntraPodFailoverEngineContainerNames()
+			info.RestoreTargetContainers = dynamo.IntraPodFailoverEngineContainerNames(component)
```
mohammedabdulwahhab: why do we need to thread component?

deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
```diff
@@ -265,6 +265,11 @@ func (r *dgdCheckpointsReconciler) createCheckpointCR(
 	if err != nil {
 		return nil, err
 	}
+	if dynamo.IsIntraPodFailoverEnabled(component) {
+		if err := dynamo.PrepareVLLMAutomaticFailoverSnapshotSource(targetContainer); err != nil {
```
mohammedabdulwahhab: Reading note for myself: this is preparing the capture pod. Why do we need to prepare anything? I assume the asserts should protect us from getting here rather than transparently modifying the config.

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	vllmWorkerClassFlag                    = "--worker-cls"
+	gmsV0VLLMWorkerClass                   = "gpu_memory_service.integrations.vllm.worker.GMSWorker"
+	gmsV0VLLMWorkerClassAlt                = "gpu_memory_service.integrations.vllm.worker:GMSWorker"
+	gmsV1VLLMWorkerClass                   = "gpu_memory_service.v1.integrations.vllm.worker.GMSV1Worker"
+	checkpointFailoverCompatibilityMessage = "Snapshot with active/passive failover requires an operator-managed automatic single-node vLLM Worker checkpoint"
 )
 
+// IsDGDControlled reports whether a DCD has an exact DynamoGraphDeployment
+// controller reference. Admission separately verifies the request principal.
+func IsDGDControlled(dcd *v1beta1.DynamoComponentDeployment) bool {
```
mohammedabdulwahhab: Is there already an existing utility for this? Are we the first person to need this? From an intent perspective, why does it matter that the DCD is not hand minted. I'm considering removing this.

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+// produces the one canonical checkpoint source.
+func ValidateAutomaticFailoverCheckpointSource(
+	component *v1beta1.DynamoComponentDeploymentSharedSpec,
+	backendFramework string,
+) []error {
+	if !hasCheckpointFailover(component) {
+		return nil
+	}
+	violations := validateAutomaticFailoverCheckpointProfile(component, backendFramework)
+	config := component.Experimental.Checkpoint
```
mohammedabdulwahhab: why can't the checkpoint be passed in? what is the intent behind this? This blocks iteration by requiring a fresh snapshot at each turn.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -163,6 +163,15 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 				fmt.Errorf("component %q: %w", component.ComponentName, compatibilityErr),
 			)
 		}
+		for _, compatibilityErr := range dynamo.ValidateAutomaticFailoverCheckpointSource(
```
mohammedabdulwahhab: why do the errors get appended here and are not appended for the DCD controller?

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	return owner != nil &&
+		owner.APIVersion == v1beta1.GroupVersion.String() &&
+		owner.Kind == "DynamoGraphDeployment" &&
+		owner.Name != "" &&
+		owner.UID != ""
+}
+
+// ValidateAutomaticFailoverCheckpointSource validates the DGD component that
+// produces the one canonical checkpoint source.
+func ValidateAutomaticFailoverCheckpointSource(
```
mohammedabdulwahhab: Can we remove automatic from these family of methods?

mohammedabdulwahhab: I don't like the source/target naming. Makes no sense to me. One is at the DGD level. One is at the DCD level.

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	if len(violations) == 0 {
+		return nil
+	}
+	return []error{fmt.Errorf("%s: %w", checkpointFailoverCompatibilityMessage, errors.Join(violations...))}
+}
+
+// PrepareVLLMAutomaticFailoverSnapshotSource applies only checkpoint-source
+// changes to the canonical main container. Destination engine topology is
+// added later when the DGD worker pod is rendered.
+func PrepareVLLMAutomaticFailoverSnapshotSource(container *corev1.Container) error {
```
mohammedabdulwahhab: Why is this not covered by the asserts/checks above? In general, I'd like to avoid any mutation of the DGD and cover stuff in the asserts instead. Let me know why this can't be the case.

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+		{Name: "DYN_REQUEST_PLANE", Value: "tcp"},
+	})
+	if err := configureVLLMAutomaticSnapshotLoadProfile(updated); err != nil {
+		return fmt.Errorf("automatic failover snapshot source: %w", err)
+	}
+	*container = *updated
+	return nil
+}
+
+func configureVLLMAutomaticSnapshotLoadProfile(container *corev1.Container) error {
```
mohammedabdulwahhab: Again, doesn't it make sense for this to be folded into asserts/checks. What can't we assert beforehand and makes sense to do transparently?

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +390,313 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	}
+	for _, profile := range []struct {
+		flag         string
+		defaultValue string
+		want         string
+		description  string
+	}{
+		{flag: "--disaggregation-mode", defaultValue: "agg", want: "agg", description: "disaggregation mode must be aggregated"},
+		{flag: "--request-plane", defaultValue: "tcp", want: "tcp", description: "request plane must be tcp"},
+		{flag: tensorParallelSizeFlag, defaultValue: "1", want: "1", description: "tensor parallel size must be 1"},
```
mohammedabdulwahhab: Future note:
- we def want to unblock tp for vllm for flashinfer recipes.

deploy/operator/internal/dynamo/graph.go
```diff
@@ -1860,12 +1860,15 @@ func GenerateBasePodSpec(
...
 
-	// Clone main container into two engine containers (active + standby) for failover.
+	// Clone the main container into active and standby engine containers for failover.
 	// Runs after GMS so the main container already has DRA claims and shared volume.
 	if IsIntraPodFailoverEnabled(component) {
 		if err := buildFailoverPod(&podSpec, numberOfNodes, backendFramework); err != nil {
 			return nil, fmt.Errorf("failed to build failover pod: %w", err)
 		}
+		if GetCheckpoint(component) != nil {
+			configureCheckpointFailoverEngines(&podSpec, component)
```
mohammedabdulwahhab: We def want something about restore being reflected here? Also do we know if this applies on both DCD and grove paths? can you check that for me

deploy/operator/internal/webhook/validation/shared_v1alpha1.go
```diff
@@ -74,7 +74,7 @@ func (v *sharedValidation) validateDynamoComponentDeploymentSharedSpecV1alpha1(
 			fmt.Sprintf("cannot inject frontend sidecar: a container named %q already exists in extraPodSpec.containers", consts.FrontendSidecarContainerName),
 		))
 	}
-	if spec.Failover != nil {
+	if spec.Failover != nil && v.hasRuntimeVersionSource(runtimeVersionSourceV1Alpha1) {
```
mohammedabdulwahhab: can you tell me about this deletion?

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +398,246 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	gpuCount, err := getGPUCount(main.Resources)
+	if err != nil {
+		violations = append(violations, fmt.Errorf("main container GPU resources are invalid: %w", err))
+	} else if gpuCount != 1 {
+		violations = append(violations, errors.New("main container must request exactly one GPU"))
+	}
+	// The operator derives these from DYN_GMS_USE_V1, which it injects when a
+	// checkpoint is configured. A user value is either ignored or contradicts
+	// the operator's selection, so reject it rather than silently rewriting it.
+	for _, flag := range vllmOperatorManagedFlags(backend) {
```
mohammedabdulwahhab: This is confusing, should be gated on vllm.

deploy/operator/internal/dynamo/failover.go
```diff
@@ -387,19 +398,246 @@ func gmsResourceSharingEntries(serviceName string, roles []ServiceRole) []grovev
...
+	}
+	// Operator-managed environment. DYN_VLLM_GMS_SHADOW_MODE marks a cold-start
+	// shadow engine; DYN_FORWARDPASS_METRIC_PORT is assigned per engine.
+	for _, name := range []string{"DYN_VLLM_GMS_SHADOW_MODE", "DYN_FORWARDPASS_METRIC_PORT"} {
+		if containerHasEnvVar(main, name) {
+			violations = append(violations, fmt.Errorf("%s is managed by the operator and must not be set", name))
+		}
+	}
+	for _, profile := range failoverSnapshotFlagProfile(backend) {
+		value, _, _, found, err := tokenizedVLLMFlag(main.Args, profile.flag)
```
mohammedabdulwahhab: same here, looks like this will be called for sglang code as well.

deploy/operator/internal/controller/dgd_checkpoints_reconciler.go
```diff
@@ -265,6 +265,11 @@ func (r *dgdCheckpointsReconciler) reconcileAutomaticSnapshotJob(
 	if err != nil {
 		return nil, err
 	}
+	if dynamo.IsIntraPodFailoverEnabled(component) {
+		if err := dynamo.PrepareVLLMSnapshotSourceContainer(targetContainer); err != nil {
```
mohammedabdulwahhab: Isn't this engine agnostic now?

mohammedabdulwahhab: This is not vllm specific anymore, right?

PR #12886, opened 2026-08-08
feat(ci): verify prometheus_names.py is in sync with Rust source

### LINKED ISSUE #12885: [CONTRIBUTION]: Add CI check to prevent prometheus_names.py from drifting out of sync with Rust source
### Type of Change

New feature

### Problem Statement

`lib/bindings/python/src/dynamo/prometheus_names.py` is auto-generated from `lib/runtime/src/metrics/prometheus_names.rs by`:
`cargo run -p dynamo-codegen --bin gen-python-prometheus-names`

However, there is currently no CI or pre-commit step that re-runs this generator and verifies that the committed Python file matches the Rust source. This has led to silent drift.

Recent examples(ref:https://github.com/ai-dynamo/dynamo/issues/12742) include the `lifecycle` class disappearing and `kvstats.KV_CACHE_HIT_RATE` being removed from the generated Python file while `components/src/dynamo/common/utils/prometheus.py` still imports them, causing `AttributeError` on worker startup.

### Proposed Solution

Add a CI check in the `rust-tests` job of `.github/workflows/pre-merge.yml`:

1. Runs `cargo run -p dynamo-codegen --bin gen-python-prometheus-names.`
2. Formats the generated file with `black==23.1.0` (the generator does not emit Black-compatible formatting).
3. Runs `git diff --exit-code lib/bindings/python/src/dynamo/prometheus_names.py` to ensure the committed file is in sync.

### Estimated PR Size

XS (1-10 lines)

### Files/Components Affected

.github/workflows/pre-merge.yml

flpanbin:
## Overview:

Add a CI check in the pre-merge `rust-tests` job that regenerates `lib/bindings/python/src/dynamo/prometheus_names.py `from lib/runtime/src/metrics/prometheus_names.rs, formats it with `black`, and fails if the committed file differs.

## Details:

* `prometheus_names.py` is auto-generated by `cargo run -p dynamo-codegen --bin gen-python-prometheus-names`.
* The generator does not emit Black-compatible formatting, so the check installs `black==23.1.0` and formats the generated file before comparing.

.github/workflows/pre-merge.yml, right after the Verify Code Formatting step.

## Related Issues

* Closes ai-dynamo/dynamo#12885

---

Open in Devin Review

## Summary by CodeRabbit

* **Tests**
  * Enhanced the Rust test matrix to verify that generated Prometheus names match the committed output.
  * Added formatting validation and automatic failure detection when differences are found.


flpanbin:
Blocked by #12745 The codegen drift check added here will fail until the sync PR lands, because the current base branch still generates a stale prometheus_names.py. Will rebase once #12745 is merged.

KrishnanPrash:
Hey @flpanbin,

Could you rebase this on latest main to pick up fixes from #12745 and address the coderabbit comments?

flpanbin:
> Hey @flpanbin,
> 
> Could you rebase this on latest main to pick up fixes from #12745 and address the coderabbit comments?

Rebased and fixed. PTAL. Thanks.

KrishnanPrash:
/ok to test [f116492](https://github.com/ai-dynamo/dynamo/pull/12886/commits/f1164928b47afd8b724bce9b08b5b159041d2496)

.github/workflows/pre-merge.yml
```diff
@@ -339,6 +339,25 @@ jobs:
     - name: Verify Code Formatting
       working-directory: ${{ matrix.dir }}
       run: cargo fmt -- --check
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
```
nv-anants: thank you @flpanbin for the PR!  

Could we move this into dynamo-codegen as a normal Rust test instead of adding a custom workflow step? The test can generate the Python module in memory and compare it to prometheus_names.py, so existing cargo test --locked --all-targets catches drift without setup-python, black, or git diff in CI.

flpanbin: @nv-anants Thanks for the suggestion. I considered moving the check into a Rust test, but the current generator output is not black-compatible.

Concrete mismatches between the generator and black:

1. Long line wrapping.

   The generator output:

   ```python
   MODEL_MIGRATION_MAX_SEQ_LEN_EXCEEDED_TOTAL = "model_migration_max_seq_len_exceeded_total"
   ```

   black wraps it to:

   ```python
   MODEL_MIGRATION_MAX_SEQ_LEN_EXCEEDED_TOTAL = (
       "model_migration_max_seq_len_exceeded_total"
   )
   ```

2. Blank lines before leading comments.

   The generator output:

   ```python
   class kvrouter:
       # Number of KV cache events applied to the index (including status)
       KV_CACHE_EVENTS_APPLIED = "kv_cache_events_applied"
   ```

   black inserts a blank line:

   ```python
   class kvrouter:

       # Number of KV cache events applied to the index (including status)
       KV_CACHE_EVENTS_APPLIED = "kv_cache_events_applied"
   ```

To avoid black in the CI step, maybe we  need to implement black's formatting rules inside the generator. It duplicates black's behavior in Rust and creates an ongoing maintenance burden to keep parity with black. The explicit CI step is a small, self-contained check that avoids that duplication. Given the current generator behavior, I prefer to keep the CI step.

If we later update the generator to emit black-compatible output natively, we can revisit moving the check into a Rust test.

@nv-anants @KrishnanPrash WDYT?

keivenchang: hey Bin — agreed that reimplementing black would be a burden, but this generator fix is +28/-2 and only 18 lines of actual code. Two changes:

1. Gate `lines.push("".to_string())` in `generate_classes` on whether it wrote a docstring. One correction: mismatch #2 is backwards. The generator emits the blank line and black strips it; this affects `kvrouter`, `kvstats`, `lifecycle`, and `model_info`.
2. Add `assignment_lines()` to wrap past 88 chars for the long-line case you showed.

After those, the generator byte-matches the committed file with no black pass:

```
$ diff lib/bindings/python/src/dynamo/prometheus_names.py /tmp/fixed.py && echo IDENTICAL
IDENTICAL
```

So, @nv-anants's test works as-is: 2 passed, clippy clean, and it caught an injected constant without regeneration. Patch: https://gist.github.com/keivenchang/3b7f9a506a20296fc3fd06a55524d2ef

.github/filters.yaml
```diff
@@ -330,6 +330,9 @@ rust:
   # Sidecar protobuf contracts are compiled by crate build scripts and need the
   # same workspace checks as sidecar Rust sources.
   - 'lib/sidecar/**/*.proto'
+  # Generated Python files under bindings must be covered by the prometheus_names.py
+  # sync check in the rust-tests job, so any change in this crate triggers it.
+  - 'lib/bindings/python/**'
```
keivenchang: `lib/bindings/python/**` matches 129 files, ~100 of them examples and READMEs. Since `rust` gates both `rust-tests` and `rust-clippy` with a 4-leg Fastchecker matrix each, a README edit fires 8 heavy Rust jobs; moving the check into a cargo test removes this entry.

.github/workflows/pre-merge.yml
flpanbin: Thanks for the patch — applied it as-is.

Verified locally:
- `cargo test -p dynamo-codegen --all-targets` → both tests pass
- `cargo run -p dynamo-codegen --bin gen-python-prometheus-names` followed by `git diff --exit-code` → IDENTICAL

With the generator now emitting black-compatible output, I also removed the standalone sync-check step from `.github/workflows/pre-merge.yml`, and narrowed the rust filter entry from `lib/bindings/python/**` to just the two relevant paths.

.github/filters.yaml
flpanbin: Good catch — `lib/bindings/python/**` was too broad.

I replaced it with the two specific paths that actually affect this check:

```yaml
- 'lib/bindings/python/codegen/templates/prometheus_names.py.template'
- 'lib/bindings/python/src/dynamo/prometheus_names.py'

nv-anants (approved):

nv-anants:
/ok to test [9acdc2c](https://github.com/ai-dynamo/dynamo/pull/12886/commits/9acdc2c6db7988cecdfaf7ae3ded59a58f959af2)

flpanbin:
Hi @keivenchang, thanks for the previous review! I've addressed the comments, rebased onto main, resolved the conflicts, and done verification. Could you please take another look? Thanks!

keivenchang (approved):

KrishnanPrash:
/ok to test [c8af1ea](https://github.com/ai-dynamo/dynamo/pull/12886/commits/c8af1ea084cacbdd76e46bcdb34d014873dbf7a2)

nv-anants:
/ok to test [a7cafb2](https://github.com/ai-dynamo/dynamo/pull/12886/commits/a7cafb2440f72662d61aa61b3f05dc4395fdb23c)

PR #12887, opened 2026-08-08
fix(snapshot): remap clone-conflicting sockets during restore

galletas1712:
## Summary

Enable repeated restores of one CRIU checkpoint into sibling engine containers that share a Pod network namespace.

For each restore, this change creates a private `files.img` view and deconflicts only the two socket identities observed in real Snapshot checkpoints:

- a CUDA UVM `SOCK_SEQPACKET` listener is rebound to an abstract address of the form `\0dynamo-<sha256>`, derived from the target container restore ID and original address; and
- one unique reciprocal internal TCP client/server pair gets a kernel-selected free client port.

The canonical checkpoint remains unchanged and reusable. Unrelated, unsupported, or ambiguous socket records are left untouched. The TCP port is reserved until restore cleanup so concurrent restores select distinct ports.

## Validation

From `deploy/snapshot`:

- `go test -race ./...`
- `go vet ./internal/criu`
- `git diff --check`

## Stack

| Order | PR | Scope |
| ---: | --- | --- |
| 1 | **#12887** | **Restore-time socket deconfliction** |
| 2 | #12869 | Shared Snapshot/GMS failover runtime |
| 3 | #12874 | Operator orchestration |

### DIFF
--- a/deploy/snapshot/internal/criu/restore.go
+++ b/deploy/snapshot/internal/criu/restore.go
@@ -39,18 +39,24 @@ func ExecuteRestore(
 	// and unlock. That keeps the window where the restored process runs with CUDA
 	// still locked as short as possible. cleanup is called on the error paths below.
 	var openFiles, inheritedFiles []*os.File
+	imageDirPath, removeImageDir, err := prepareRestoreImageDir(checkpointPath)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to prepare CRIU image directory: %w", err)
+	}
 	var confTmpDir string
 	cleanup := func() {
 		closeFiles(inheritedFiles)
 		closeFiles(openFiles)
+		removeImageDir()
 		if err := os.RemoveAll(confTmpDir); err != nil {
 			log.Error(err, "failed to remove criu config temp dir", "path", confTmpDir)
 		}
 	}
 
 	// Open image dir FD
-	imageDir, imageDirFD, err := openPathForCRIU(checkpointPath)
+	imageDir, imageDirFD, err := openPathForCRIU(imageDirPath)
 	if err != nil {
+		cleanup()
 		return 0, nil, fmt.Errorf("failed to open image directory: %w", err)
 	}
 	openFiles = append(openFiles, imageDir)
@@ -100,7 +106,7 @@ func ExecuteRestore(
 	log.V(1).Info("Executing go-criu Restore call")
 	if err := c.Restore(criuOpts, notify); err != nil {
 		log.Error(err, "go-criu Restore returned error")
-		logging.LogRestoreErrors(checkpointPath, settings.WorkDir, log)
+		logging.LogRestoreErrors(imageDirPath, settings.WorkDir, log)
 		cleanup()
 		return 0, nil, fmt.Errorf("CRIU restore failed: %w", err)
 	}
--- a/deploy/snapshot/internal/criu/restore_images.go
+++ b/deploy/snapshot/internal/criu/restore_images.go
@@ -1,0 +1,367 @@
+package criu
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"net/netip"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/checkpoint-restore/go-criu/v8/crit"
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/fdinfo"
+	sk_inet "github.com/checkpoint-restore/go-criu/v8/crit/images/sk-inet"
+	sk_unix "github.com/checkpoint-restore/go-criu/v8/crit/images/sk-unix"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	filesImageFilename            = "files.img"
+	placeholderMountNamespacePath = "/proc/self/ns/mnt"
+	cudaUVMFDSocketNamePrefix     = "\x00cuda-uvmfd-"
+	linuxUnixSocketStateListen    = 10
+	linuxTCPStateEstablished      = 1
+	linuxTCPStateClose            = 7
+	linuxTCPStateListen           = 10
+)
+
+type tcpPortRewrite struct {
+	socket *sk_inet.InetSkEntry
+	peers  []*sk_inet.InetSkEntry
+	port   uint32
+}
+
+func prepareRestoreImageDir(checkpointPath string) (string, func(), error) {
+	// The placeholder mount namespace remains container-specific with shareProcessNamespace.
+	var stat unix.Stat_t
+	if err := unix.Stat(placeholderMountNamespacePath, &stat); err != nil {
+		return "", nil, fmt.Errorf("failed to stat placeholder mount namespace at %s: %w", placeholderMountNamespacePath, err)
+	}
+	return prepareRestoreImageDirForRestoreID(checkpointPath, stat.Ino)
+}
+
+func prepareRestoreImageDirForRestoreID(checkpointPath string, restoreID uint64) (string, func(), error) {
+	checkpointPath, err := filepath.Abs(checkpointPath)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to resolve checkpoint path: %w", err)
+	}
+
+	filesImage, err := os.Open(filepath.Join(checkpointPath, filesImageFilename))
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to open %s: %w", filesImageFilename, err)
+	}
+
+	image, err := crit.New(filesImage, nil, "", false, false).Decode(&fdinfo.FileEntry{})
+	closeErr := filesImage.Close()
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to decode %s: %w", filesImageFilename, err)
+	}
+	if closeErr != nil {
+		return "", nil, fmt.Errorf("failed to close %s: %w", filesImageFilename, closeErr)
+	}
+
+	// CRIU recreates bound sockets from files.img. Give each clone a private
+	// view so containers sharing a Pod network namespace get unique identities.
+	tcpRewrites, tcpDisconnects, forbiddenPorts := planTCPPortRewrites(image)
+	var reservationFDs []int
+	for i := range tcpRewrites {
+		var reservationFD int
+		tcpRewrites[i].port, reservationFD, err = reserveDualStackTCPPort(forbiddenPorts)
+		if err != nil {
+			closeFDs(reservationFDs)
+			return "", nil, fmt.Errorf("failed to reserve replacement TCP port: %w", err)
+		}
+		reservationFDs = append(reservationFDs, reservationFD)
+		forbiddenPorts[tcpRewrites[i].port] = struct{}{}
+	}
+
+	rewritten := false
+	for _, entry := range image.Entries {
+		fileEntry := entry.Message.(*fdinfo.FileEntry)
+		if fileEntry.GetType() != fdinfo.FdTypes_UNIXSK || fileEntry.Usk == nil {
+			continue
+		}
+		if rewriteCloneConflictingUnixSocketAddress(fileEntry.Usk, restoreID) {
+			rewritten = true
+		}
+	}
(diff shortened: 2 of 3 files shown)


deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,140 @@
...
+	checkpointPath, err := filepath.Abs(checkpointPath)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to resolve checkpoint path: %w", err)
+	}
+
+	filesImage, err := os.Open(filepath.Join(checkpointPath, filesImageFilename))
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to open %s: %w", filesImageFilename, err)
+	}
+	defer filesImage.Close()
```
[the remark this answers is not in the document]
galletas1712: Fixed explicitly in b15f0e4ae7: production decode now checks and wraps the `filesImage.Close` error, and the test helper reports deferred close errors. `make lint` and `go test ./internal/criu` both pass on the amended head.

deploy/snapshot/internal/criu/restore.go
```diff
@@ -38,14 +38,20 @@ func ExecuteRestore(
...
 		closeFiles(inheritedFiles)
 		closeFiles(openFiles)
+		removeImageDir()
 	}
 
 	// Open image dir FD
-	imageDir, imageDirFD, err := openPathForCRIU(checkpointPath)
+	imageDir, imageDirFD, err := openPathForCRIU(imageDirPath)
 	if err != nil {
+		cleanup()
```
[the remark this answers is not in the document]
galletas1712: The PR body now has refreshed `Summary`, `Validation`, and `Stack` sections, and the title matches the amended commit. Local validation passes, including `make lint`, focused tests, and the TCP tests repeated 25 times. Full CI has not yet passed for rewritten head b15f0e4ae7; a maintainer must issue `/ok to test b15f0e4ae7` to trigger it. I am resolving the completed metadata concern without claiming CI success.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, entry := range entries {
+		name := entry.Name()
+		if name == filesImageFilename || !strings.HasSuffix(name, ".img") {
+			continue
+		}
+		// Restore images are read-only; symlinks avoid copying page images.
+		if err := os.Symlink(filepath.Join(checkpointPath, name), filepath.Join(privateDir, name)); err != nil {
+			return fail(fmt.Errorf("failed to expose CRIU image %s: %w", name, err))
+		}
+	}
```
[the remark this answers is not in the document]
galletas1712: Confirmed: Dynamo Snapshot does not produce CRIU incremental/pre-dump parent chains. The current non-`.img` restore artifacts are supplied separately or by absolute path, so the private directory intentionally contains the CRIU image set only. Incremental parent chains and future non-`.img` image-dir inputs are out of scope for b15f0e4ae7.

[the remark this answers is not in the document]
galletas1712: Addressed in b15f0e4ae7. The private restore directory is now a sibling of the checkpoint on the checkpoint PVC, not under `TMPDIR`, and unchanged images are hard-linked rather than referenced through absolute symlinks. The hard links are necessary to keep images visible after CRIU transitions into the restored mount namespace while preserving the canonical checkpoint.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,140 @@
...
+	filesImage, err := os.Open(filepath.Join(checkpointPath, filesImageFilename))
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to open %s: %w", filesImageFilename, err)
+	}
+	defer filesImage.Close()
+
+	image, err := crit.New(filesImage, nil, "", false, false).Decode(&fdinfo.FileEntry{})
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to decode %s: %w", filesImageFilename, err)
+	}
```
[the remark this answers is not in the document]
galletas1712: The fail-closed behavior is intentional in b15f0e4ae7. Dynamo pins compatible CRIU/go-criu versions, and falling back after a missing or undecodable `files.img` would knowingly restore canonical socket identities that can collide across repeated copies. Refusing the restore is safer than silently taking that path.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, false
+	}
+
+	rewritten := make([]byte, 0, len(name)+20)
+	rewritten = append(rewritten, cudaSocketNameStart...)
+	rewritten = strconv.AppendUint(rewritten, restoreSocketScope, 10)
+	rewritten = append(rewritten, numericFields[separator:]...)
+	rewritten = append(rewritten, name[end:]...)
+	return rewritten, true
+}
```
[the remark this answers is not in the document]
galletas1712: The CUDA rewrite remains exact and fail-closed in b15f0e4ae7: only the proven `\0cuda-uvmfd-<decimal namespace>-<decimal pid>` format is changed, with trailing bytes preserved, and unrelated names are untouched. Repeated-restore runtime evidence confirmed the rewritten names work even when the namespace field length changes, so this is the intended scoped behavior.

galletas1712:


deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,331 @@
...
+		if !isEstablishedTCP(socket) || hasTCPListener(socket, sockets) {
+			continue
+		}
+		rewrite := tcpPortRewrite{socket: socket}
+		for _, peer := range sockets {
+			if reciprocalTCPPair(socket, peer) {
+				rewrite.peers = append(rewrite.peers, peer)
+			}
+		}
+		rewrites = append(rewrites, rewrite)
```
julienmancuso: [P1] Only add a rewrite when there is exactly one reciprocal in-image peer. As written, a socket with zero peers is still appended, so standalone outbound connections get a new source port while the remote endpoint retains the old four-tuple. CRIU restores the addresses and ports directly from this record, leaving that connection invalid; this also contradicts the PR's stated unique-reciprocal-pair scope.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,331 @@
...
+	for i := range tcpRewrites {
+		var reservationFD int
+		tcpRewrites[i].port, reservationFD, err = reserveDualStackTCPPort(forbiddenPorts)
+		if err != nil {
+			// Port deconfliction is best effort. Apply either the complete plan
+			// or none of it so reciprocal tuples remain consistent.
+			closeFDs(reservationFDs)
+			reservationFDs = nil
+			tcpRewrites = nil
+			break
```
julienmancuso: [P1] Do not silently continue after port reservation fails. This path returns success with all TCP rewrites disabled, so the repeated restore can hit the original collision later (for example when the IPv6-only reservation socket is unavailable on an IPv4-only kernel). Propagate the error, or implement family-aware IPv4/IPv6 reservation with an explicit fallback.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func closeFDs(fds []int) {
+	for _, fd := range fds {
+		_ = unix.Close(fd)
+	}
+}
+
+func rewriteCloneConflictingUnixSocketAddress(entry *sk_unix.UnixSkEntry, restoreID uint64) bool {
+	if entry == nil || (!isCUDAUVMFDListener(entry) && !isLinuxAutobindAddress(entry.Name)) {
```
julienmancuso: [P2] This rewrites every six-byte Linux-autobound abstract UNIX address, not just the proven CUDA topology. Unrelated datagram or stream sockets—and external peers retaining their original names—can be disconnected. Please narrow this to the CUDA-related records, or explicitly define and validate the broader clone semantics.

julienmancuso (changes requested):
Three socket-deconfliction correctness issues need resolution before merge.

deploy/snapshot/internal/criu/restore.go
```diff
@@ -38,14 +38,20 @@ func ExecuteRestore(
...
 	// still locked as short as possible. cleanup is called on the error paths below.
 	var openFiles, inheritedFiles []*os.File
+	imageDirPath, removeImageDir, err := prepareRestoreImageDir(checkpointPath)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to prepare CRIU image directory: %w", err)
+	}
 	cleanup := func() {
 		closeFiles(inheritedFiles)
 		closeFiles(openFiles)
+		removeImageDir()
```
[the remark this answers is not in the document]
galletas1712: The private image directory only replaces `ImagesDirFd`; it does not become CRIU’s cwd. `ExecuteRestore` continues to pass the manifest `WorkDirFd`, and CRIU opens the image directory separately before changing directory through `WorkDirFd`. The `/ -> .` mapping also predates this PR. Pointing it at the canonical checkpoint directory would be incorrect because that directory contains CRIU images, not the placeholder rootfs selected by `CriuOpts.Root = "/"`. The private-image restore path has also been exercised successfully end to end. No code change here.

deploy/snapshot/internal/criu/restore_images.go
galletas1712: Fixed in e26608d5c. `planTCPPortRewrites` now appends a rewrite only when there is exactly one reciprocal in-image peer. Standalone outbound connections remain unchanged; the full image transformation test verifies that only the internal pair changes.

deploy/snapshot/internal/criu/restore_images.go
galletas1712: Fixed in e26608d5c. A replacement-port reservation failure now closes the reservations already acquired and returns the error immediately. Restore no longer continues with the original colliding ports.

deploy/snapshot/internal/criu/restore_images.go
galletas1712: Fixed in e26608d5c. General Linux-autobind matching has been removed. AF_UNIX rewriting is now limited to the observed CUDA UVM FD listener shape (`SOCK_SEQPACKET`, listener state, no peer, and the abstract `cuda-uvmfd-` prefix); unrelated autobound sockets remain unchanged.

deploy/snapshot/internal/criu/restore_images.go
```diff
@@ -0,0 +1,334 @@
...
+			continue
+		}
+		rewrite := tcpPortRewrite{socket: socket}
+		for _, peer := range sockets {
+			if reciprocalTCPPair(socket, peer) {
+				rewrite.peers = append(rewrite.peers, peer)
+			}
+		}
+		if len(rewrite.peers) == 0 {
+			disconnects = append(disconnects, socket)
```
julienmancuso, line 198: [P1] A missing exact reciprocal match does not prove this peer is remote. `reciprocalTCPPair` requires identical address families and encodings, so a valid in-image connection between an `AF_INET` client and an `AF_INET6` dual-stack listener can fail matching and have both local endpoints forced to `TCP_CLOSE` here. Leave unmatched or unsupported sockets unchanged unless they are proven external, or normalize IPv4 and IPv4-mapped IPv6 tuples before classifying them; please add a mixed-family local-pair test.

julienmancuso (changes requested):
One new correctness issue remains after the update.

deploy/snapshot/internal/criu/restore_images.go
galletas1712: Good catch — an exact family/word-array match was too strict for a local IPv4 peer accepted by a dual-stack IPv6 listener. I normalized AF_INET and IPv4-mapped AF_INET6 addresses before reciprocal matching (while leaving unsupported encodings untouched), and extended the existing full `files.img` rewrite test with that mixed-family local pair. The observed external HTTPS socket is still restored disconnected. Validated with `go test ./internal/criu -count=1` and `go vet ./internal/criu`. Fixed in 728ef65464f5c1f5611872f2aca77bc77c701cf1.

julienmancuso (approved):

PR #12900, opened 2026-08-09
test(kv-router): flush both ranks before asserting rank reset

### LINKED ISSUE #12262: [BUG]: flaky test: concurrent_rank_reset_waits_for_all_local_tiers races lower-tier event application
### Describe the Bug

Hey! While working on #12260 (a frontend-only change) I hit a CI failure in `rust-tests (.)` that has nothing to do with my diff. The test `kv_router::indexer::tests::concurrent_rank_reset_waits_for_all_local_tiers` failed, while `single_thread_rank_reset_waits_for_all_local_tiers` (the same scenario on the non-concurrent indexer) passed in the same run. I dug in to make sure I wasn't shipping a real breakage, and I'm fairly confident this is a race in the test setup. Sharing what I found!

Not related to #12252, which is a compile error in a later step of the same CI job and looks already resolved on current main.

### Steps to Reproduce

The test passes hundreds of iterations locally in isolation, so you need lane-timing skew to see it. Two ways:

1. In CI it shows up occasionally on loaded runners: [failing run](https://github.com/ai-dynamo/dynamo/actions/runs/30357151638/job/90268060966).
2. Deterministic local repro: add a diagnostic 200 ms delay to lower-tier `WorkerTask::Event` application for dp_rank 1 only (in `lib/kv-router/src/indexer/lower_tier.rs`), simulating the retained rank's lane being scheduled late, then run:

```bash
cargo test -p dynamo-llm --lib rank_reset_waits_for_all_local_tiers
```

The concurrent test then fails every time with the exact CI signature, while the single-thread variant still passes.

### Expected Behavior

Both tests pass regardless of how the OS schedules the indexer's worker lanes: after the reset of rank (7,0), the retained rank (7,1) should still show its device score of 1 and its HostPinned hit of 1.

### Actual Behavior

```
thread 'kv_router::indexer::tests::concurrent_rank_reset_waits_for_all_local_tiers'
panicked at lib/llm/src/kv_router/indexer/mod.rs:549:9:
assertion `left == right` failed
  left: Some(0)
 right: Some(1)
```

The earlier device assertion passes, then the HostPinned assertion sees 0 hits for the retained rank.

### Environment

- CI: `rust-tests (.)` job in pre-merge, Linux runner, on #12260 (branch includes main at 11c1d7cd25).
- Also reproduced locally on macOS (arm64) with the deterministic delay above, so it is not platform specific.

### Additional Context

Root cause, as far as I can tell: the test enqueues store events for two ranks of one worker, resets one rank, then queries. Nothing guarantees the retained rank's store events have actually been applied before the query runs:

1. `Indexer::apply_event` for the `Concurrent` variant only awaits the enqueue, not the application (`try_apply_event` -> `enqueue_event` in `lib/llm/src/kv_router/indexer/mod.rs`).
2. The lower-tier `ThreadPoolIndexer` assigns each `(worker_id, dp_rank)` to its own FIFO lane (`get_or_assign_thread_idx` in `lib/kv-router/src/indexer/thread_pool.rs`), so the reset rank (7,0) and the retained rank (7,1) live on different lanes.
3. `reset_worker_dp_rank_and_wait` flushes only the reset rank's lane (`flush_worker_lane_and_wait(worker)` in `thread_pool.rs`), which looks intentional.
4. `find_matches_by_tier` reads lower-tier state directly with no lane barrier (`query_lower_tiers` -> `backend().query_match_details` in `lib/kv-router/src/indexer/lower_tier_indexers.rs`).

So if the retained rank's lane gets scheduled late (say on a busy CI machine), the query runs before its HostPinned store is applied. Since lower-tier hits are seeded from the confirmed device match, the retained rank shows up in the hits map with 0 instead of being absent. That matches the CI output exactly. The single-thread `KvIndexer` is structurally immune because its events and queries share one FIFO.

Suggested fix: I think it's a one-liner in the test, flush after the apply loop so the stores reach a stable boundary before the reset and query. `ThreadPoolIndexer::flush` says it exists "primarily for testing" for exactly this, and the test module already has a `flush_indexer` helper that other tests use. That keeps the reset-acknowledgement assertion meaningful while removing the race.

Happy to send a PR for this if the fix sounds right. Let me know!

ayaangazali:
## Summary

Fixes #12262.

`assert_rank_reset_is_acknowledged` applies setup events for **both** the reset rank and the retained rank, then immediately resets one of them:

```rust
for dp_rank in [reset_rank.dp_rank, retained_rank.dp_rank] {
    indexer.apply_event(store_event(7, dp_rank, 1, &[], &[41], StorageTier::Device)).await;
    indexer.apply_event(store_event(7, dp_rank, 2, &[41], &[42], StorageTier::HostPinned)).await;
}

indexer.reset_worker_dp_rank_and_wait(reset_rank.worker_id, reset_rank.dp_rank).await.unwrap();
```

`reset_worker_dp_rank_and_wait` only waits for the tiers of the rank **being reset**. Nothing waits for the retained rank, so its `HostPinned` event can still be in flight when the assertion at the bottom reads it:

```rust
assert_eq!(host_hits.get(&retained_rank), Some(&1));
```

On `KvIndexer` the setup events are applied in order and the test passes, which is why `single_thread_rank_reset_waits_for_all_local_tiers` is green in the same run. On the concurrent indexer the retained rank's lane can be scheduled late — exactly the skew @michaelfeil reproduced in #12262 by delaying lower-tier `WorkerTask::Event` application for `dp_rank 1`.

The fix is to flush after setup so both ranks are known-applied before the reset. This matches the existing convention in this module: `flush_indexer` is already called before asserting in the other tests here, including the sibling `tiered_query_chains_device_host_and_disk`. The helper is the outlier.

This changes test setup only — no production code, and the assertions are unchanged, so the test still fails if rank reset genuinely stops clearing a rank.

## Validation

`cargo fmt --all` applied.

**I could not execute this test locally, and I would rather say so than imply otherwise.** `dynamo-llm` does not build on macOS — it fails on three Linux-only APIs, independent of this change:

```
error[E0432]: unresolved import `dynamo_memory::numa`
error[E0432]: unresolved imports `nix::fcntl::FallocateFlags`, `nix::fcntl::fallocate`
error[E0599]: no associated function or constant named `O_DIRECT` found for struct `OFlag`
```

Confirmed pre-existing: `cargo check -p dynamo-llm --lib` reports the same errors on a clean `main` with this change stashed.

So this rests on reading the code rather than on a local run:

- `reset_worker_dp_rank_and_wait` iterates only the tiers for the rank being reset, so it cannot order the retained rank's events.
- `flush_indexer` flushes `primary` and every `lower_tier` indexer, which covers both ranks.
- The single-thread variant of the same helper passes consistently, which is what you would expect if the only difference is lane scheduling.

Worth running against the deterministic repro from #12262 (200 ms delay on lower-tier `Event` application for `dp_rank 1`) to confirm it closes the window — I could not run that here. Happy to iterate if CI disagrees.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12900" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Tests**
  * Improved rank-reset acknowledgment test reliability by ensuring index updates are processed before validation.

### DIFF
--- a/lib/llm/src/kv_router/indexer/mod.rs
+++ b/lib/llm/src/kv_router/indexer/mod.rs
@@ -595,6 +595,14 @@ mod tests {
         }
         flush_indexer(&indexer).await;
 
+        // Flush before the reset so BOTH ranks' setup events are known-applied.
+        // `reset_worker_dp_rank_and_wait` only waits for the tiers of the rank being
+        // reset, so without this the retained rank's lower-tier event can still be in
+        // flight when the assertions below read it — the race behind the flake on the
+        // concurrent indexer. Every other test in this module already flushes before
+        // asserting.
+        flush_indexer(&indexer).await;
+
         indexer
             .reset_worker_dp_rank_and_wait(reset_rank.worker_id, reset_rank.dp_rank)
             .await


ayaangazali:
Correcting the Validation section above: I said this test could not be executed locally because `dynamo-llm` does not build on macOS. That was true as written but the conclusion was wrong — the blocker is the `block-manager` feature, which is on by default and pulls in `nix::fcntl::fallocate`, `OFlag::O_DIRECT` and `dynamo_memory::numa`. Building without it works:

```
$ cargo test -p dynamo-llm --lib --no-default-features kv_router::indexer
53 passed

$ cargo test -p dynamo-llm --lib --no-default-features rank_reset
3 passed
```

So the changed test and its two sibling rank-reset tests do now run here and pass, and the surrounding indexer module is green. (`lib/bindings/python` already depends on `dynamo-llm` with `default-features = false`, which is why the Python bindings build on macOS while a plain `cargo check -p dynamo-llm` fails — that is what put me on to it.)

What I still have not reproduced is the *failure* without this change. It is a scheduling race, and @michaelfeil's report reproduced it by deliberately delaying lower-tier `WorkerTask::Event` application for `dp_rank 1`; I have not tried to inject that delay, so I am not claiming a local red-to-green demonstration. The argument for the change is unchanged and structural: `reset_worker_dp_rank_and_wait` only waits for the tiers of the rank being reset, so nothing orders the retained rank's `HostPinned` event before the assertion that reads it.

No code change in this update — flagging it because the Validation section understated what had been checked.

ayaangazali:
Rebased onto current `main` (was 158 commits behind, no conflicts) so CI runs against today's tree. The diff is unchanged.

Still applies and still needed: #12262 is open and unassigned, `concurrent_rank_reset_waits_for_all_local_tiers` is still at `lib/llm/src/kv_router/indexer/mod.rs:616` on `main`, and the flush this adds is still absent there.

One update worth recording, because it corrects the Validation section above. I originally wrote that I could not execute this test locally because `dynamo-llm` does not build on macOS. The build failure is real but the conclusion was wrong: it comes from the `block-manager` feature, which is on by default and pulls in `fallocate`, `O_DIRECT` and `dynamo_memory::numa`. Without it the crate builds and the tests run:

```
$ cargo test -p dynamo-llm --lib --no-default-features rank_reset
3 passed

$ cargo test -p dynamo-llm --lib --no-default-features kv_router::indexer
62 passed
```

So the changed test and its two siblings do now run on my machine and pass, as does the surrounding indexer module.

What I still have not done is reproduce the failure without this change. It is a scheduling race, and your report reproduced it by deliberately delaying lower-tier `WorkerTask::Event` application for `dp_rank 1`. I have not injected that delay, so I am not claiming a local red-to-green demonstration. The argument stays structural: `reset_worker_dp_rank_and_wait` only waits for the tiers of the rank being reset, so nothing orders the retained rank's `HostPinned` event before the assertion that reads it.

ayaangazali:
CI is fully green now. The `lychee` failure was the dead xgrammar link on `main`, not anything in this diff; #13320 fixed it, and rebasing onto current `main` picked that up, so there is no red check left here.

Nothing else outstanding from my side. Still applies cleanly and the tests are unchanged.

ayaangazali:
Closing this: the change is a no-op and I should have caught it before opening.

`assert_rank_reset_is_acknowledged` already calls `flush_indexer(&indexer).await` after the `for dp_rank in [reset_rank.dp_rank, retained_rank.dp_rank]` loop and before `reset_worker_dp_rank_and_wait`. That existing flush is outside the loop, so both ranks' device and host-pinned setup events are already applied before the reset. What this PR adds is a second, adjacent `flush_indexer` call with nothing between the two, so it changes nothing about ordering. No commit has touched that file since my branch point, so the flush was there when I opened this.

I also could not reproduce the flake it claims to fix: 40 consecutive runs of `concurrent_rank_reset_waits_for_all_local_tiers` on current main all pass. So there is no evidence of a race here for a flush to fix.

Sorry for the noise. If the test does turn out to flake in CI I will come back with an actual reproduction and a fix that addresses it.

PR #12916, opened 2026-08-10
fix(sglang): validate token IDs on all client-supplied input paths

weizhoublue:
## Overview:

fix: validate client-supplied token IDs on all SGLang token input paths

## Details:

Currently, it implents token validation for the `nvext.token_data` path. This PR extends that validation to the remaining client-supplied token path.

The OpenAI `/v1/completions` spec allows four forms of `prompt`, two of which are token ID arrays:

```
curl -s http://frontend:8000/v1/completions -H 'Content-Type: application/json' -d '{
  "model": "Qwen/Qwen3-0.6B",
  "prompt": [151643, 9707, 4294967295],
  "max_tokens": 8
}'
```

151643 and 9707 are valid, 4294967295 indexes out of bounds on the GPU embedding weight matrix.

This is usually not malicious. It comes from a client that tokenizes locally and then assembles the array, accidentally inserting a length, an offset, or a hash value as a token ID, or from a client tokenizer that does not match the server-side model vocabulary.

An integer-array prompt on `/v1/completions` carries no `token_in` marker and never passes through frontend tokenization. It is raw user input. An out-of-range ID here can cause SGLang to fail downstream.

This PR applies the same out-of-vocabulary validation when tokens arrive through prompt

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for client-provided token inputs across supported request formats.
  - Added consistent handling for invalid or out-of-vocabulary tokens.
  - Preserved support for SGLang-tokenized text and multimodal placeholder tokens.
  - Updated validation error messages to use clearer, generic wording.

- **Tests**
  - Expanded coverage for prompt token arrays, accepted inputs, rejected inputs, and tokenizer-specific behavior.

### DIFF
--- a/components/src/dynamo/sglang/request_handlers/handler_base.py
+++ b/components/src/dynamo/sglang/request_handlers/handler_base.py
@@ -1007,10 +1007,11 @@ class BaseWorkerHandler(LoraMixin, BaseGenerativeHandler[RequestT, ResponseT]):
                     400,
                     f"token input[{index}] must be an integer token ID",
                 )
-            # Dynamo's Rust frontend uses u32 token IDs, so negatives are not expected.
-            if (
-                max_input_token_id is not None and token_id > max_input_token_id
-            ) and token_id not in allowed_oov_ids:
+            if token_id < 0 or (
+                max_input_token_id is not None
+                and token_id > max_input_token_id
+                and token_id not in allowed_oov_ids
+            ):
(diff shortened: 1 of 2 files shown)


weizhoublue:
@indrajit96 I submit this follow-up PR to strengthen PR https://github.com/ai-dynamo/dynamo/pull/12798
Could you help check whether this makes sense, thanks

PR #12922, opened 2026-08-10
refactor(vllm): add reusable decoder runtime

furionw:
_Based on #13073._

## Why

Standalone `dynamo.vllm` serving and workflow stages need one Dynamo-to-vLLM decode implementation. Sharing only the raw engine would still duplicate prompt budgeting, sampling preparation, output adaptation, abort, and shutdown behavior.

## What Change

- Add `VllmDecoderRuntime` around an already configured `AsyncLLM`.
- Centralize request preparation, prompt-budget validation, generation, chunk adaptation, abort, and shutdown.
- Delegate handler decoding to the runtime while the handler retains admission, registration, and transport policy.

```mermaid
classDiagram
    DecodeWorkerHandler o-- VllmDecoderRuntime : delegates decode
    VllmDecoderRuntime --> AsyncLLM : drives
    VllmDecoderRuntime ..> SamplingParams : prepares
    VllmDecoderRuntime ..> VllmDecodeOutputAdapter : adapts
```

## Test Plan

- Focused runtime and handler qualification: 226 passed and 5 intentionally skipped.
- Built stack-tip H100 qualification: 107 tests passed, including the shared runtime suite.

### DIFF
--- a/components/src/dynamo/vllm/decoder_runtime.py
+++ b/components/src/dynamo/vllm/decoder_runtime.py
@@ -1,0 +1,191 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared Dynamo-to-vLLM decoder implementation."""
+
+from __future__ import annotations
+
+from collections.abc import AsyncIterator, Callable, Mapping
+from enum import Enum
+from typing import Any, cast
+
+from vllm.config import ModelConfig, VllmConfig
+from vllm.outputs import RequestOutput
+from vllm.sampling_params import RequestOutputKind, SamplingParams
+from vllm.v1.engine.async_llm import AsyncLLM
+
+from dynamo.common.backend import GenerateChunk, GenerateRequest
+from dynamo.common.backend import logprobs as _shared_logprobs
+from dynamo.llm.exceptions import InvalidArgument
+
+from .decoder_output import ExtractLogprobs, _VllmDecodeOutputAdapter
+from .decoder_sampling import build_sampling_params
+
+GenerateFactory = Callable[[Any], AsyncIterator[RequestOutput]]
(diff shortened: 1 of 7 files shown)


dagil-nvidia (approved):
Approving from the process-codeowners side: all deterministic gates pass at head (CODEOWNERS regenerates byte-identical at 621 rules, 5522/5522 files owned, 93 matcher tests, 18 skills validated), and no process-owned file changes in this diff - the process review request is a stale stacked-diff artifact (it fired at the exact second of a base-branch force-push, against a transient merge-base, and GitHub never withdraws requests). The load-bearing reviews here are the vllm and rl owners, and the real open design question (hard rejection of over-budget `max_tokens` on worker-expanded prompts) is already well covered by the existing threads.

Four non-blocking notes from the pass:

1. `components/src/dynamo/vllm/decoder_runtime.py:90` - both new `InvalidArgument` branches (prompt fills the context; `max_tokens` over the prepared budget) have zero coverage in the new test file, and they are exactly the code the open threads dispute. Two tests asserting each raise would pin the intended contract whichever way that discussion lands.
2. `decoder_runtime.py:124` - `decode()` defaults `lora_request=_LORA_UNSET` and passes it into `admission(...)`, whose only implementation types the arg as `LoRARequest | None` and would call `.lora_name` on a truthy sentinel. Safe today only because the sole caller always passes an explicit value; map the sentinel to `None` when admission is set.
3. `handlers.py:119` - this PR removes the last consumer of the `_VllmDecodeOutputAdapter` re-export; unlike its siblings it now has zero importers repo-wide and can be dropped.
4. `handlers.py:2725` - the completion log now prints the request-wide `completion_tokens` sum where the old code logged the per-index count; with parallel sampling the logged number changes per finish chunk. Log-only; keep knowingly or restore per-index.

Verified sound: the `generate_tokens` relocation has no external callers, the exceptions import is real, decode-loop behavior parity holds against the deleted handler code including the synthetic error chunk and finish-reason mapping, `worker_factory.py` has no orphaned imports and preserves lifecycle cleanup ordering, and the defensive-copy change is equivalent for all consumers.

PR #12376, opened 2026-07-29
feat(frontend): add basetenkenizer backend

maheshrbapatu:
## Overview:

Adds Baseten Tokenizer as a selectable Dynamo frontend tokenizer backend using `BasetenTokenizer` exposed by the current upstream `dynamo-tokenizers` 1.7.0. Operators can select it with `--tokenizer basetenkenizer` or `DYN_TOKENIZER=basetenkenizer`.

## Details:

- Use `BasetenTokenizer` from the current upstream `dynamo-tokenizers` 1.7.0.
- Accept `basetenkenizer` through the Python CLI/environment configuration, JSON runtime configuration, and Rust string parsing.
- Construct the Baseten backend for supported `tokenizer.json` files while preserving HuggingFace fallback behavior.
- Preserve special tokens declared in sibling `tokenizer_config.json` files and wrap the selected backend with the existing L1 prefix cache.
- Add token-parity, special-token, cache, segmented-encoding, and fallback coverage.
- Document configuration, compatibility, caching, and fallback behavior.

## Where should the reviewer start?

- `lib/llm/src/model_card.rs` for backend construction, caching, and fallback.
- `lib/llm/src/local_model/runtime_config.rs` for configuration parsing and precedence.
- `lib/llm/tests/model_card.rs` for parity, cache, special-token, segmented-encoding, and fallback coverage.

## Related Issues

- Relates to #12325

## Validation

- `cargo check -p dynamo-llm --lib --locked`
- `cargo test -p dynamo-llm --lib tokenizer_backend --locked` (4 passed)
- `cargo test -p dynamo-llm --test model_card --locked` (10 passed)
- `cargo test -p dynamo-llm --test tokenizers --locked` (22 passed)
- Ruff, Rust formatting, codespell, and `git diff --check`
## End-to-end performance validation

AIPerf exercised all three Dynamo frontend tokenizer backends against the same live `Qwen/Qwen3-8B` vLLM worker.

### Hardware and software

- NVIDIA GH200 480GB platform: one GH200 GPU (`97,871 MiB` HBM reported by `nvidia-smi`) and one 72-core NVIDIA Grace/Arm Neoverse-V2 CPU.
- aarch64 Linux `6.8.0-111-generic`; NVIDIA driver `610.43.02`.
[...]

PR #12923, opened 2026-08-10
feat(frontend): make tokenizer fallback configurable

xianlubird:
## Summary

Basetenkenizer support landed in #12376 alongside the existing fastokens backend. Both alternate tokenizer backends currently prioritize availability: when the selected implementation cannot load a model's `tokenizer.json`, the frontend logs a warning and transparently uses HuggingFace instead.

That default is useful for general serving, but it can hide a configuration mismatch when an operator explicitly requires an alternate backend. For example, a performance run can complete successfully while measuring HuggingFace after an unnoticed fallback, and a production deployment can start without the tokenizer implementation it was configured to use.

This change makes that fallback policy configurable while keeping existing deployments compatible:

- Add the negatable `--tokenizer-fallback` / `--no-tokenizer-fallback` frontend option, with `DYN_TOKENIZER_FALLBACK` as the equivalent environment variable.
- Keep fallback enabled by default, preserving the existing behavior for all current deployments.
- Pass the parsed frontend value explicitly through `EntrypointArgs`, `ModelRuntimeConfig`, and every `ModelWatcher` path so dynamically discovered model cards receive the same policy before tokenizer construction.
- Validate environment values strictly. Supported boolean spellings are `true`/`false`, `1`/`0`, `on`/`off`, and `yes`/`no`; invalid values now fail with a configuration error instead of silently restoring the default.
- When fallback is disabled, return the original fastokens or basetenkenizer loading error with context instead of constructing the HuggingFace fallback.
- Apply the same fail-fast behavior to non-UTF-8 tokenizer paths, which alternate backends cannot consume.
- Document the option in the tokenizer guide, frontend configuration reference, and Fastokens usage guide.

Example:

```bash
python -m dynamo.frontend \
  --tokenizer basetenkenizer \
  --no-tokenizer-fallback
```

## Validation

- `cargo test -p dynamo-llm --no-default-features tokenizer_fallback --lib --tests` (3 focused tests passed)
- Configuration utility tests (30 passed)
- Added Rust coverage for explicit configuration precedence, strict environment parsing, serialization compatibility, fail-fast loading behavior, and applying an explicit `false` override to a dynamically discovered model card
- `python3 -m black --check components/src/dynamo/common/configuration/utils.py components/src/dynamo/common/tests/configuration/test_utils.py components/src/dynamo/frontend/frontend_args.py`
- `python3 docs/fern/scripts/check_asset_paths.py`
- `python3 docs/fern/scripts/gen_python_api.py --check` (Python 3.13, matching CI)
- `python3 -m py_compile components/src/dynamo/common/configuration/utils.py components/src/dynamo/common/tests/configuration/test_utils.py components/src/dynamo/frontend/frontend_args.py`

## Related Issues

None.

## Reviewer Guide

Start with the negatable frontend option in `components/src/dynamo/frontend/frontend_args.py` and the explicit Python-to-Rust handoff in `components/src/dynamo/frontend/main.py` and `lib/bindings/python/rust/llm/entrypoint.rs`. Then review strict fallback precedence in `lib/llm/src/local_model/runtime_config.rs`, propagation into dynamic model discovery in `lib/llm/src/discovery/watcher.rs`, and the fail-fast branches in `lib/llm/src/model_card.rs`.


components/src/dynamo/frontend/main.py
```diff
@@ -390,6 +390,7 @@ def signal_handler():
         kv_router_config = None
 
     os.environ[MIN_INITIAL_WORKERS_ENV] = str(config.min_initial_workers)
+    os.environ["DYN_TOKENIZER_FALLBACK"] = str(config.tokenizer_fallback).lower()
```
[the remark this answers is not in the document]
xianlubird: Done

docs/fern/pages/use-cases/fastokens-tokenizer/overview.md
```diff
@@ -115,7 +117,7 @@ Stay on the default backend if:
   - **Encoding**: `fastokens` converts prompt text to token IDs.
   - **Decoding**: HuggingFace `tokenizers` converts generated token IDs back to text.
 
-  Both backends load from the same `tokenizer.json`, so supported tokenizers should produce the same token IDs as the default HuggingFace path. If `fastokens` cannot load the tokenizer file, Dynamo logs a warning and falls back to the default backend instead of dropping requests.
+  Both backends load from the same `tokenizer.json`, so supported tokenizers should produce the same token IDs as the default HuggingFace path. If `fastokens` cannot load the tokenizer file, Dynamo logs a warning and falls back to the default backend. Set `--no-tokenizer-fallback` to make this loading failure stop model initialization.
```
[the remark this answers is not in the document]
xianlubird: Done

components/src/dynamo/frontend/frontend_args.py
```diff
@@ -552,6 +554,20 @@ def add_arguments(self, parser) -> None:
             choices=["default", "fastokens", "basetenkenizer"],
         )
 
+        add_argument(
```
jthomson04: Can this be a negatable bool argument to align it with other bool flags in the frontend?

lib/bindings/python/rust/llm/entrypoint.rs
```diff
@@ -647,6 +648,11 @@ impl EntrypointArgs {
                 .inner
                 .set_tokenizer_backend(Some(tokenizer_backend));
         }
+        if let Some(tokenizer_fallback) = tokenizer_fallback {
+            runtime_config
+                .inner
+                .set_tokenizer_fallback_enabled(Some(tokenizer_fallback));
```
jthomson04, line 654: This override never reaches the model card that the dynamic watcher tokenizes. The HTTP, gRPC, and common paths forward only tokenizer_backend, and ModelWatcher applies only that field to discovered cards. Since this PR also removes the environment write, --tokenizer-fallback false is ignored in normal dynamic mode. Thread the fallback override through ModelWatcher and apply it before card.tokenizer().

lib/llm/src/local_model/runtime_config.rs
```diff
@@ -542,6 +549,16 @@ impl ModelRuntimeConfig {
             .unwrap_or_else(TokenizerBackend::from_env_or_default)
     }
 
+    pub fn is_tokenizer_fallback_enabled(&self) -> bool {
+        self.tokenizer_fallback_enabled.unwrap_or_else(|| {
+            std::env::var(ENV_TOKENIZER_FALLBACK)
+                .ok()
+                .as_deref()
+                .and_then(dynamo_runtime::config::parse_bool_opt)
```
jthomson04: parse_bool_opt maps invalid values to None, so a typo such as flase silently falls through to true. This is a fail-fast control; invalid configuration must return an error. Use strict parse_bool and propagate the error from tokenizer().

jthomson04 (changes requested):
The explicit fallback setting is lost in the normal dynamic frontend path, and the boolean parsing behavior is inconsistent between the Python and Rust entry points. Please keep the fix focused and add one regression test that applies an explicit false override to a discovered model card.

components/src/dynamo/frontend/frontend_args.py
xianlubird: Updated this to use `add_negatable_bool_argument`, so the frontend now exposes `--tokenizer-fallback` and `--no-tokenizer-fallback`. Environment values still go through the strict `parse_bool` parser, and I added focused configuration utility coverage.

lib/bindings/python/rust/llm/entrypoint.rs
xianlubird: Fixed in `b7b4a6e402`. I threaded `tokenizer_fallback_enabled` through the HTTP, gRPC, and common watcher paths, and `ModelWatcher` now applies the override to each discovered card before tokenizer construction. I also added a regression test where a worker advertises `true` and an explicit frontend `false` wins.

lib/llm/src/local_model/runtime_config.rs
xianlubird: Updated `is_tokenizer_fallback_enabled` to return `Result<bool>` and use the strict `parse_bool` path. Invalid or non-UTF-8 environment values now propagate an error from `tokenizer()` instead of silently restoring the default. The tests include the `flase` typo case.

components/src/dynamo/frontend/frontend_args.py
```diff
@@ -552,6 +554,19 @@ def add_arguments(self, parser) -> None:
...
 
+        add_negatable_bool_argument(
+            g,
+            flag_name="--tokenizer-fallback",
+            env_var="DYN_TOKENIZER_FALLBACK",
+            default=True,
+            help=(
+                "Fall back to HuggingFace when the selected fastokens or "
+                "basetenkenizer backend cannot load the model tokenizer. Use "
+                "--no-tokenizer-fallback to fail during startup instead."
```
jthomson04: In the dynamic HTTP, gRPC, and interactive paths, tokenizer loading runs in the discovery controller. A load error is logged and retried, so the frontend process stays up with the model unavailable. Please change this help and the matching docs to describe model initialization failure, not frontend startup failure.

jthomson04 (approved):
The earlier propagation and strict parsing issues are fixed. One non-blocking CLI help and documentation note remains.

components/src/dynamo/frontend/frontend_args.py
xianlubird: Updated in `fe28206aab`. The CLI help and matching docs now describe this as a model initialization failure, and clarify that dynamic discovery retries the load while the frontend keeps running. Thanks for pointing this out!

docs/fern/pages/developer-guide/knowledge-base/modular-components/frontend/tokenizer.md
```diff
@@ -42,6 +42,7 @@ Set the backend with a CLI flag or environment variable. The CLI flag takes prec
 | CLI Argument | Env Var | Valid values | Default |
 |---|---|---|---|
 | `--tokenizer` | `DYN_TOKENIZER` | `default`, `fastokens`, `basetenkenizer` | `default` |
+| `--tokenizer-fallback` / `--no-tokenizer-fallback` | `DYN_TOKENIZER_FALLBACK` | enabled, disabled | enabled |
```
harryskim: `enabled` / `disabled` aren't accepted values for `DYN_TOKENIZER_FALLBACK`. The `parse_bool` added in this PR accepts `true`/`false`, `1`/`0`, `on`/`off`, `yes`/`no` — and the same PR makes invalid values a hard configuration error rather than silently restoring the default, so a reader who follows this table and sets `DYN_TOKENIZER_FALLBACK=enabled` gets a startup failure. It also contradicts the fastokens guide, which correctly says `DYN_TOKENIZER_FALLBACK=false`.

Suggested:

```suggestion
| `--tokenizer-fallback` / `--no-tokenizer-fallback` | `DYN_TOKENIZER_FALLBACK` | `true`/`false`, `1`/`0`, `on`/`off`, `yes`/`no` | `true` |
```

docs/fern/pages/developer-guide/knowledge-base/modular-components/frontend/tokenizer.md
```diff
@@ -31,7 +31,7 @@ Use this backend for supported `tokenizer.json` models when you need Baseten Tok
 #### Compatibility notes:
 
 - Works with standard BPE `tokenizer.json` files (Qwen, LLaMA, GPT-family, Mistral, DeepSeek, etc.).
-- If `fastokens` or `basetenkenizer` cannot load a particular tokenizer file, the frontend logs a warning and transparently falls back to HuggingFace; requests are never dropped.
+- If `fastokens` or `basetenkenizer` cannot load a particular tokenizer file, the frontend logs a warning and transparently falls back to HuggingFace by default. Disable tokenizer fallback to reject incompatible tokenizers during model initialization. In dynamic mode, discovery retries the load while the frontend continues running.
```
harryskim: Two small things:

1. "Disable tokenizer fallback" doesn't name the flag on first mention — `--no-tokenizer-fallback` only shows up later in the table. Naming it inline here is cheap and saves a scroll.
2. "In dynamic mode, discovery retries the load while the frontend continues running" also appears on line 71 and again in `frontend-configuration.mdx`. Keeping it in step 4 (where the failure mode is actually described) and dropping it from this compatibility note would read better.

harryskim (commented):
Docs-only review (the 4 hand-written pages; the three `docs/fern/pages/reference/api/python/*.mdx` files are regenerated output and look consistent with `_core.pyi`).

Overall this reads well and the option is documented in all three places a reader would look. I checked the "in dynamic mode, discovery retries the load while the frontend continues running" claim against the discovery controller's retry/backoff path and it holds. The `ParamField` added to `frontend-configuration.mdx` matches the file's existing convention for negatable booleans, and the mermaid branch split in the fastokens guide is correct.

One blocking issue (wrong valid values in the tokenizer guide table) plus a couple of nits inline.

docs/fern/pages/use-cases/fastokens-tokenizer/overview.md
```diff
@@ -66,6 +66,8 @@ Stay on the default backend if:
 
     You can also set the same option with the `--tokenizer fastokens` frontend flag. The flag takes precedence when both settings are present.
 
+    To require `fastokens`, set `DYN_TOKENIZER_FALLBACK=false` or add `--no-tokenizer-fallback`. The frontend then reports an error during model initialization when `fastokens` cannot load the tokenizer instead of using HuggingFace.
```
harryskim, line 73: Worth documenting the strict parsing here (or as a sibling accordion next to "Why do the logs show an unrecognized DYN_TOKENIZER value?"): unlike `DYN_TOKENIZER`, an invalid `DYN_TOKENIZER_FALLBACK` value now fails with a configuration error instead of warning and falling back to the default. That's a user-visible behavior change and the kind of thing that otherwise turns into a support round-trip.

docs/fern/pages/developer-guide/knowledge-base/modular-components/frontend/tokenizer.md
xianlubird: Fixed in `9b1a49d08b`. The table now lists the exact accepted environment values (`true`/`false`, `1`/`0`, `on`/`off`, and `yes`/`no`) and shows the default as `true`.

docs/fern/pages/developer-guide/knowledge-base/modular-components/frontend/tokenizer.md
xianlubird: Updated in `9b1a49d08b`. The compatibility note now names `--no-tokenizer-fallback` inline, and I removed the repeated dynamic-discovery sentence there. The retry behavior remains documented in the model initialization flow where the failure behavior is described.

docs/fern/pages/use-cases/fastokens-tokenizer/overview.md
xianlubird: Added in `9b1a49d08b`. The troubleshooting section now documents the strict accepted values, explains that invalid values fail model initialization, and notes that dynamic discovery retries while the frontend remains running.

harryskim (approved):
LGTM

lib/llm/src/local_model/runtime_config.rs
```diff
@@ -573,6 +580,21 @@ impl ModelRuntimeConfig {
             .unwrap_or_else(TokenizerBackend::from_env_or_default)
     }
 
+    pub fn is_tokenizer_fallback_enabled(&self) -> anyhow::Result<bool> {
+        if let Some(enabled) = self.tokenizer_fallback_enabled {
```
GuanLuo: Can you add deprecation notice here? And add a note `/fern/pages/reference/general/releases/deprecations.mdx` as well?

xianlubird: Added in 87d75bf345. Automatic fallback now emits a one-time deprecation warning when it is enabled for fastokens or basetenkenizer, and the CLI help and tokenizer docs call out the future fail-fast default. I also added the v1.4.0 entry to deprecations.mdx with migration guidance for --no-tokenizer-fallback and DYN_TOKENIZER_FALLBACK=false.

xianlubird:
Looks like this CI failure is unrelated to the code changes in this PR. The first lychee pass hit a connection failure on an Istio docs URL. The retry did run and that link recovered, but a different CNCF Slack redirect then failed with a connection reset by peer. All PR-specific checks passed, so this appears to be an external network flake.

PR #12924, opened 2026-08-10
feat(examples): compare workflow and bespoke vision ensemble

furionw:
_Based on #13293._

## Why

Dynamo endpoints already let applications build this ensemble manually, so the workflow API should demonstrate its value against an equivalent baseline. Both variants share the inline components, stock vLLM worker, and the same `GenerateEndpointInvoker` for request validation, stream folding, and remote-call cancellation. This isolates the comparison to orchestration ownership: the workflow runtime owns graph scheduling and binding, while the bespoke handler owns coroutine and sibling-task lifecycle.

## Effect

```text
                  ┌─> inline classifier ────────────┐
request → encoder ┤                                 ├→ inline merge → chunk
                  └─> request adapter → stock vLLM ─┘
```

## What Change

- Add declarative and bespoke mixed-placement vision ensembles.
- Share stages, launch topology, and Generate endpoint invocation.
- Keep scheduling and sibling lifecycle as the comparison.

```mermaid
classDiagram
    WorkflowOrchestrator *-- StageDispatcher : schedules graph
    StageDispatcher ..> GenerateEndpointInvoker : invokes remote Generate
    BespokeEnsembleHandler ..> GenerateEndpointInvoker : invokes same adapter
    GenerateEndpointInvoker ..> DynamoClient : calls discovered endpoint
    BespokeEnsembleHandler --> EncoderStage : schedules inline

    note for WorkflowOrchestrator "Owns authored graph execution"
    note for BespokeEnsembleHandler "Owns equivalent task orchestration"
    note for GenerateEndpointInvoker "Shares validation, folding, and call cancellation"
```

## Test Plan

- Generate adapter and bespoke orchestration CPU behavior pass.
- Targeted pre-commit hooks pass.

### DIFF
--- a/examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
+++ b/examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
@@ -5,12 +5,18 @@
 
 import asyncio
 
-from dynamo.experimental.workflow import WorkflowEndpointHandler, WorkflowOrchestrator
+from dynamo.experimental.workflow import (
+    GenerateEndpointBinding,
+    InlineBinding,
+    WorkflowEndpointHandler,
+    WorkflowOrchestrator,
+)
 from dynamo.experimental.workflow.vllm import ExternalEncoderRequestStage
 from dynamo.llm import ModelInput, ModelType, WorkerType, register_model
 from dynamo.runtime import DistributedRuntime, dynamo_worker
 from examples.experimental.workflow.user_ensemble.common.config import (
     CHAT_TEMPLATE,
+    GENERATOR_ENDPOINT,
     MODEL,
     ORCHESTRATOR_ENDPOINT,
     PUBLIC_MODEL_NAME,
@@ -21,7 +27,7 @@ from examples.experimental.workflow.user_ensemble.common.stages import (
     EnsembleResponseStage,
 )
 from examples.experimental.workflow.user_ensemble.workflow.workflow import (
-    compile_user_ensemble,
+    define_workflow,
 )
 
 
@@ -30,14 +36,15 @@ async def worker(runtime: DistributedRuntime) -> None:
     encoder = build_encoder_stage()
     try:
         orchestrator = await WorkflowOrchestrator.bind(
-            compile_user_ensemble(),
-            runtime=runtime,
-            inline_runners={
-                "encoder": encoder,
-                "classifier": DummyClassifier(),
-                "request_adapter": ExternalEncoderRequestStage(),
-                "response": EnsembleResponseStage(),
+            define_workflow(),
+            bindings={
+                "encoder": InlineBinding(encoder),
+                "classifier": InlineBinding(DummyClassifier()),
+                "request_adapter": InlineBinding(ExternalEncoderRequestStage()),
+                "generator": GenerateEndpointBinding(GENERATOR_ENDPOINT),
+                "response": InlineBinding(EnsembleResponseStage()),
             },
+            runtime=runtime,
         )
         endpoint = runtime.endpoint(ORCHESTRATOR_ENDPOINT)
         await register_model(
--- a/examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
+++ b/examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
@@ -1,0 +1,63 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Serve the declarative user ensemble as one discovered model worker."""
+
+import asyncio
+
+from dynamo.experimental.workflow import WorkflowEndpointHandler, WorkflowOrchestrator
+from dynamo.experimental.workflow.vllm import ExternalEncoderRequestStage
+from dynamo.llm import ModelInput, ModelType, WorkerType, register_model
+from dynamo.runtime import DistributedRuntime, dynamo_worker
+from examples.experimental.workflow.user_ensemble.common.config import (
+    CHAT_TEMPLATE,
+    MODEL,
+    ORCHESTRATOR_ENDPOINT,
+    PUBLIC_MODEL_NAME,
+    build_encoder_stage,
+)
+from examples.experimental.workflow.user_ensemble.common.stages import (
+    DummyClassifier,
+    EnsembleResponseStage,
+)
+from examples.experimental.workflow.user_ensemble.workflow.workflow import (
+    compile_user_ensemble,
+)
+
+
+@dynamo_worker()
+async def worker(runtime: DistributedRuntime) -> None:
+    encoder = build_encoder_stage()
+    try:
+        orchestrator = await WorkflowOrchestrator.bind(
+            compile_user_ensemble(),
+            runtime=runtime,
+            inline_runners={
+                "encoder": encoder,
+                "classifier": DummyClassifier(),
+                "request_adapter": ExternalEncoderRequestStage(),
+                "response": EnsembleResponseStage(),
+            },
+        )
+        endpoint = runtime.endpoint(ORCHESTRATOR_ENDPOINT)
+        await register_model(
+            ModelInput.Tokens,
+            ModelType.Chat,
+            endpoint,
+            MODEL,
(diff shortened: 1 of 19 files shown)


examples/custom_backend/user_ensemble/local/worker.py
```diff
@@ -0,0 +1,189 @@
...
+                self._encoder_backend_type,
+                name="workflow-vision-encoder",
+                model_config=decoder_runtime.model_config,
+            )
+            runners = {
+                "encoder": encoder_stage,
+                "classifier": self._classifier,
+                "generator": decoder_stage,
+            }
+            plan = compile_workflow(define_workflow())
```
caronzh03: Does user have to author this `worker.py` for every single ensemble pipeline?
If so, this file seems to contain a lot of boilerplate code.
It might be a cleaner interface if a user only needs to author L87-L105 (the top portion of this `start()` function) to register the custom `stages` and `workflow`.

examples/custom_backend/user_ensemble/resources.py
```diff
@@ -0,0 +1,90 @@
...
+    if resolved_model_config is None:
+        resolved_model_config = config.engine_args.create_engine_config(
+            usage_context=UsageContext.OPENAI_API_SERVER
+        ).model_config
+    backend = backend_type()
+    adapter = create_custom_encoder_adapter(
+        backend,
+        resolved_model_config,
+        config.engine_args,
+    )
```
caronzh03: What does this `adapter` do?

examples/custom_backend/user_ensemble/workflow.py
```diff
@@ -0,0 +1,66 @@
...
+        ensemble = {}
+    if not isinstance(ensemble, Mapping):
+        raise TypeError(
+            "decoder chunk ensemble metadata must be an object when present"
+        )
+    merged_ensemble = dict(ensemble)
+    merged_ensemble["classifier_scores"] = dict(scores)
+    merged_engine_data["ensemble"] = merged_ensemble
+    adapted["engine_data"] = merged_engine_data
+    return adapted
```
caronzh03: If I'm not mistaken, this  function is the post-processing step that stitches multiple outputs from different stages together, correct?

Even though this is pure Python logic without ML, would it be OK to model this step as a `stage` as well? That way, nothing lives outside of the `workflow`.

examples/custom_backend/user_ensemble/local/worker.py
furionw: makes sense. will take this into consideration

examples/custom_backend/user_ensemble/resources.py
furionw: given an output of encoder, users may need to pass different kinds of `Prompt` into engine (vllm in this case).

e.g. 

- it may be a `EmbedPrompt` for LLM model 
- it may be a `TokenPrompt` for VLM model (because Qwen VL requires `grid_thw` alongside embedding)

I am seeing different cases where one of these prompts is needed.

examples/custom_backend/user_ensemble/workflow.py
furionw: correct. it's totally doable to model it inside `stage`.

examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
```diff
@@ -0,0 +1,70 @@
...
+            ModelInput.Tokens,
+            ModelType.Chat,
+            endpoint,
+            MODEL,
+            model_name=PUBLIC_MODEL_NAME,
+            worker_type=WorkerType.Aggregated,
+            custom_template_path=str(CHAT_TEMPLATE),
+            ignore_weights=True,
+        )
+        await endpoint.serve_endpoint(WorkflowEndpointHandler(orchestrator).generate)
```
caronzh03, line 60: Where is `WorkflowEndpointHandler` defined?
Does user have to implement custom `WorkflowEndpointHandler` for each workflow?

caronzh03 (commented):
From user-interface perspective, the workflow approach seems cleaner, assuming this doesn't require a custom `WorkflowEndpointHandler` implementation like `BespokeEnsembleHandler`.

From functionality perspective, I cannot tell whether either approach can support remote stage calling in the future. For example, in bespoke approach, does the `classifier_task` have to be a local, in-process call? In workflow approach, does `classifier` stage have to be a local, in-process call?

I'm wondering if we can fuse the two approaches into one:
* Let user-interface use the workflow approach - define graph topology without `asyncio.create_task()` boilerplate.
* Let the Dynamo-provided `WorkflowEndpointHandler` be the default handler for workflow execution, and if users want to customize the execution logic, like adding prioritization or throttling, they can override  `WorkflowEndpointHandler`, and at that point, the custom handler will look something like `BespokeEnsembleHandler`.

examples/experimental/workflow/user_ensemble/workflow/orchestrator_worker.py
furionw: > Where is WorkflowEndpointHandler defined?

https://github.com/ai-dynamo/dynamo/pull/13083/

> Does user have to implement custom WorkflowEndpointHandler for each workflow?

No, `WorkflowEndpointHandler` kicks off orchestrator which executes users' workflow.

furionw:
> From user-interface perspective, the workflow approach seems cleaner, assuming this doesn't require a custom `WorkflowEndpointHandler` implementation like `BespokeEnsembleHandler`.

the assumption is correct

# Inline- and Remote-

> From functionality perspective, I cannot tell whether either approach can support remote stage calling in the future. For example, in bespoke approach, does the `classifier_task` have to be a local, in-process call? In workflow approach, does `classifier` stage have to be a local, in-process call?

**Existing Dynamo primitives already support it remote stage calling.** both `bespoke` and `workflow` approach, can support remote stage calling. 

## Dynamo Primitive

tl;dr users' Python can
- import Dynamo's `runtime` class, call `runtime.endpoint("demo.backend.generate").serve_endpoint(my_function)`
- on client side, `(await runtime.endpoint("demo.backend.generate").client()).generate(request)`

## Workflow (inline- and remote- switch)

the workflow abstract the above logic so that users only need to
- define a stage (`input`, `output` and `run` function)
- specify whether they want in-process or separate-process (e.g. remote): `InlineBinding`, `RemoteBinding(endpoint_id)`

This would allow users to switch between inline- and remote- easily. Does this abstraction make things more complicated than helpful?

```
// This PR
orchestrator = await WorkflowOrchestrator.bind(
    define_workflow(),
    bindings={
        "encoder": InlineBinding(encoder),
        "generator": GenerateEndpointBinding(GENERATOR_ENDPOINT),
        ...
    },
    runtime=runtime,
)

// hello-world example 
// https://github.com/ai-dynamo/dynamo/pull/13423/changes
orchestraor = WorkflowOrchestrator.bind(
    define_workflow(),
    bindings={
        stage_id: RemoteBinding(endpoint_id)
        for stage_id, endpoint_id in ENDPOINTS.items()
    },
    runtime=runtime,
)
```

furionw:
> I'm wondering if we can fuse the two approaches into one:> 
> * Let user-interface use the workflow approach - define graph topology without `asyncio.create_task()` boilerplate.
> * Let the Dynamo-provided `WorkflowEndpointHandler` be the default handler for workflow execution, and if users want to customize the execution logic, like adding prioritization or throttling, they can override  `WorkflowEndpointHandler`, and at that point, the custom handler will look something like `BespokeEnsembleHandler`.

agreed. 

- there will be a point, where users need to customize beyond a static graph
  - we have users doing that today based on dynamo's endpoint primitives)
  - from workflow point of view, we can support Triton-BLS-like (business logic scripting) capabilities (loop, conditional, etc.) -- [4. Imperative authoring
](https://github.com/ai-dynamo/dynamo/issues/13007)
- dynamo has been supporting prioritization, requests admission (a form of throttling). If we proceed with workflow approach, Dynamo can also provide these optional utilities at stage level.

caronzh03:
>Does this abstraction make things more complicated than helpful?

The Workflow abstraction hides the execution details/complexity, giving users a clean way to define graph topology; it also decouples graph definition and instance binding nicely - reminding me of "dependency injection" design pattern in Java world, so I would definitely prefer using the WF approach. 
In the bespoke approach, the graph topology definition and execution happens together in the Handler class. Perhaps if the graph is simple, there's not much difference in terms of maintenance effort, but if the graph becomes complex, and with remote calls, the code could be un-readable.

Issue #12938, opened 2026-08-10
Make DGD workload-provider selection immutable


julienmancuso:
## Problem

Today, a DynamoGraphDeployment's selected workload provider can change during the lifetime of the object because provider selection is derived repeatedly from routing annotations, feature gates, API availability, and eligibility checks.

As a result, an existing DGD can move between the component, Grove, and DisaggregatedSet workload programs after an annotation edit, an operator configuration change, an upgrade, or a provider becoming available or unavailable. These programs own different resource graphs and lifecycle semantics, and Dynamo does not define a safe live-migration contract between them. A temporary provider outage can also cause an unintended fallback to become the deployment's effective provider.

## Desired behavior

A DGD's workload provider is selected once and remains fixed for the lifetime of that DGD.

- The possible providers are `component`, `grove`, and `disaggregatedset`.
- New DGDs select a provider from their creation-time routing intent.
- Existing DGDs preserve the provider represented by the workloads they already own.
- Routing annotation changes, feature-gate changes, API availability, and eligibility changes do not change an already selected provider.
- Attempts to request a different provider after selection are rejected with guidance to delete and recreate the DGD.
- If the selected provider becomes unavailable, the DGD reports `ProviderUnavailable` and does not fall back to another provider.
- Explicit DisaggregatedSet intent remains DisaggregatedSet intent even when that provider is temporarily unavailable; it must not become an accidental component selection.
- Ambiguous existing state with conflicting top-level providers is reported as an error rather than resolved by guessing.

The provider identifies the top-level workload program, not every Kubernetes resource kind it may manage. In particular, a DisaggregatedSet deployment may legitimately include a DCD-backed global frontend while its provider remains `disaggregatedset`.

## Scope

This issue establishes immutability of provider selection. Live migration, handoff, and cleanup between providers are out of scope. Changing providers requires deleting and recreating the DGD.

## Acceptance criteria

- A DGD cannot silently change workload programs after its initial provider selection.
- Existing DGDs are adopted without changing their effective provider.
- Provider unavailability is visible in DGD status and never triggers fallback.
- User-facing behavior clearly explains that changing provider requires DGD recreation.
- Component, Grove, and DisaggregatedSet paths have coverage for stable selection, adoption, unavailable providers, and conflicting provider state.

## References

- #12035 — composition-first operator architecture
- #12283 — composition-first DGD reconciliation
- #12625 — DisaggregatedSet workload program

PR #12942, opened 2026-08-10
feat(operator): persist immutable DGD workload provider selection

julienmancuso:
## Summary

DynamoGraphDeployment workload-provider selection currently depends on live routing intent and operator feature gates. That means the effective provider can change during the lifetime of a DGD.

This change makes the graph-level provider durable by materializing `nvidia.com/workload-provider` with either `component` or `grove` before workload reconciliation begins. Once recorded, that value is authoritative and immutable.

Related to #12938.

## What changed

- Materialize a missing workload provider from the current Grove gate and `nvidia.com/enable-grove` intent.
- Apply provider selection level-wise in the defaulting webhook before provider-specific defaults.
- Add a controller-side safety net for legacy objects or requests that bypass admission:
  - persist a missing provider with optimistic concurrency;
  - return and requeue after persistence;
  - perform no finalizer, status, or workload effects before the provider is durable.
- Dispatch the DGD exclusively through the program named by the persisted annotation.
- Keep the persisted provider authoritative when feature gates or routing annotations later change.
- Reject unsupported provider values and changes or removal after the provider has been materialized.
- If a DGD is durably selected to Grove while Grove is unavailable:
  - do not fall back to the Component program;
  - publish `Ready=False` with `selected_workload_provider_unavailable`;
  - return a terminal reconciliation error after status persistence.

## Upgrade behavior

An existing DGD without `nvidia.com/workload-provider` is assigned from current routing intent on its next admission or controller reconciliation. Selection is persisted before further reconciliation effects.

This change intentionally does not inspect or adopt from existing DCD or PodCliqueSet resources. It establishes a small immutable-selection boundary instead of a compatibility or recovery framework.

After selection, changing `nvidia.com/enable-grove`, feature gates, or API availability does not silently switch the workload program. Changing providers requires deleting and recreating the DGD.

## Provider boundaries

The values in this change identify graph-level workload programs:

- `component` selects the DCD-based Component program.
- `grove` selects the Grove program.

Deployment versus LeaderWorkerSet remains an internal choice of the DCD controller within the Component program. LWS therefore does not require another graph-level provider value or a separate DGD provider-selection mechanism.

Future graph-level programs can extend this contract with an additional provider value while retaining the same materialize-once and dispatch-from-persisted-state rules.

## Validation

- `docker buildx build --platform linux/arm64 --target linter --progress=plain --build-context snapshot=../snapshot .`
- `docker buildx build --platform linux/arm64 --target tester --progress=plain --build-context snapshot=../snapshot --build-context operator=../operator --build-context operator-chart=../helm/charts/platform/components/operator .`

## Summary by CodeRabbit

* **New Features**
  * Added durable workload-provider selection for deployments, supporting Grove and component providers.
  * Provider choices are persisted and reused during reconciliation.
  * Added validation to allow supported providers and prevent changing or removing an established provider.
  * Provider-specific defaults and admission behavior now respect the selected provider.
  * Disabled or unavailable Grove selections now report a terminal error without fallback.

* **Bug Fixes**
  * Improved handling of provider conflicts, unsupported values, and persistence failures.
  * Prevented side effects before provider selection is finalized.

### DIFF
--- a/deploy/operator/internal/consts/consts.go
+++ b/deploy/operator/internal/consts/consts.go
@@ -42,6 +42,11 @@ const (
 
 	KubeAnnotationEnableGrove = "nvidia.com/enable-grove"
 
+	// KubeAnnotationWorkloadProvider records the controller-owned immutable graph-level workload provider.
+	KubeAnnotationWorkloadProvider = "nvidia.com/workload-provider"
+	WorkloadProviderComponent      = "component"
+	WorkloadProviderGrove          = "grove"
+
 	// KubeAnnotationGroveUpdateStrategy temporarily exposes the Grove
 	// PodCliqueSet update strategy while the long-term DGD API is settled.
 	// Supported values match Grove exactly: "RollingRecreate" and "OnDelete".
--- a/deploy/operator/internal/controller/dgd_grove_program.go
+++ b/deploy/operator/internal/controller/dgd_grove_program.go
@@ -22,7 +22,9 @@ import (
 	"fmt"
 
 	nvidiacomv1beta1 "github.com/ai-dynamo/dynamo/deploy/operator/api/v1beta1"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 type groveProgram struct {
@@ -33,6 +35,7 @@ type groveProgram struct {
 	workloads       *groveWorkloadsReconciler
 	scalingAdapters *dgdScalingAdaptersReconciler
 	topology        *dgdGroveTopologyConditionReconciler
+	gate            features.Gate
 }
 
 // newGroveProgram wires the Grove pathway at the DGD composition root.
@@ -62,6 +65,7 @@ func (r *DynamoGraphDeploymentReconciler) newGroveProgram() *groveProgram {
 		),
 		scalingAdapters: newDGDScalingAdaptersReconciler(r.Client, r.Recorder),
 		topology:        newDGDGroveTopologyConditionReconciler(r.Client),
+		gate:            r.RuntimeConfig.Gate,
 	}
 }
 
@@ -73,6 +77,17 @@ func (p *groveProgram) Reconcile(
 	req workloadProgramRequest,
 ) (programResult workloadProgramResult, retErr error) {
 	programResult = newWorkloadProgramResult(req.DGD)
+
+	// Fail a durable Grove selection when Grove is unavailable rather than falling back.
+	if !p.gate.Enabled(features.Grove) {
+		err := failWorkloadProgram(
+			reasonSelectedWorkloadProviderUnavailable,
+			fmt.Errorf("selected workload provider %q is unavailable because Grove is disabled", workloadProviderGrove),
+		)
+		programResult.Fail(req.DGD.Generation, reasonSelectedWorkloadProviderUnavailable, err)
+		return programResult, reconcile.TerminalError(err)
+	}
+
 	defer func() {
 		if retErr != nil {
 			reason := reasonFailedToReconcileResources
--- a/deploy/operator/internal/controller/dgd_workload_provider.go
+++ b/deploy/operator/internal/controller/dgd_workload_provider.go
@@ -1,0 +1,181 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controller
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	nvidiacomv1beta1 "github.com/ai-dynamo/dynamo/deploy/operator/api/v1beta1"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/consts"
+	"github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
+	grovev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type workloadProvider string
+
+const (
+	workloadProviderComponent workloadProvider = consts.WorkloadProviderComponent
+	workloadProviderGrove     workloadProvider = consts.WorkloadProviderGrove
+)
+
+var (
+	errConflictingWorkloadProviders = errors.New("conflicting workload providers")
+	errUnsupportedWorkloadProvider  = errors.New("unsupported workload provider")
+)
+
+// ensureWorkloadProvider persists a missing provider before workload reconciliation.
+func (r *DynamoGraphDeploymentReconciler) ensureWorkloadProvider(
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) (workloadProvider, error) {
+	// Reuse a materialized provider without consulting mutable routing inputs.
+	if value, exists := dgd.Annotations[consts.KubeAnnotationWorkloadProvider]; exists {
+		return parseWorkloadProvider(value)
+	}
+
+	// Adopt the workload family already owned by a legacy unannotated DGD.
+	provider, found, err := providerFromOwnedWorkloads(ctx, r.Client, dgd)
+	if err != nil {
+		return "", err
+	}
+	if !found {
+		provider = providerFromCurrentIntent(r.RuntimeConfig.Gate, dgd)
+	}
+
+	// Persist the adopted or creation-time selection with optimistic concurrency control.
+	base := dgd.DeepCopy()
+
+	// Materialize the selected provider on the object passed to the API client.
+	if dgd.Annotations == nil {
+		dgd.Annotations = make(map[string]string)
+	}
+	dgd.Annotations[consts.KubeAnnotationWorkloadProvider] = string(provider)
+
+	// Persist the selection and retain the API server's returned object state.
+	patch := client.MergeFromWithOptions(base, client.MergeFromWithOptimisticLock{})
+	if err := r.Patch(ctx, dgd, patch); err != nil {
+		dgd.Annotations = base.Annotations
+		return "", fmt.Errorf("persist workload provider %q: %w", provider, err)
+	}
+
+	return provider, nil
+}
+
+func providerFromOwnedWorkloads(
+	ctx context.Context,
+	reader client.Reader,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) (workloadProvider, bool, error) {
+	// Observe both workload families before deciding whether either is authoritative.
+	hasComponents, err := hasOwnedComponentWorkloads(ctx, reader, dgd)
+	if err != nil {
+		return "", false, err
+	}
+	hasGrove, err := hasOwnedGroveWorkloads(ctx, reader, dgd)
+	if err != nil {
+		return "", false, err
+	}
+
+	// Adopt one unambiguous family and fail closed when both families exist.
+	switch {
+	case hasComponents && hasGrove:
+		return "", false, fmt.Errorf(
+			"%w: DynamoGraphDeployment %s/%s owns DynamoComponentDeployments and PodCliqueSets",
+			errConflictingWorkloadProviders,
+			dgd.Namespace,
+			dgd.Name,
+		)
+	case hasGrove:
+		return workloadProviderGrove, true, nil
+	case hasComponents:
+		return workloadProviderComponent, true, nil
+	default:
+		return "", false, nil
+	}
+}
+
+func hasOwnedComponentWorkloads(
+	ctx context.Context,
+	reader client.Reader,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) (bool, error) {
+	// Scan the namespace cache because mutable labels cannot prove legacy ownership.
+	list := &nvidiacomv1beta1.DynamoComponentDeploymentList{}
+	if err := reader.List(ctx, list, client.InNamespace(dgd.Namespace)); err != nil {
+		return false, fmt.Errorf("list owned DynamoComponentDeployments: %w", err)
+	}
+
+	// Accept only the controller reference as authoritative ownership evidence.
+	for i := range list.Items {
+		if metav1.IsControlledBy(&list.Items[i], dgd) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func hasOwnedGroveWorkloads(
+	ctx context.Context,
+	reader client.Reader,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) (bool, error) {
+	// Treat an unavailable optional Grove API as an empty observable workload family.
+	list := &grovev1alpha1.PodCliqueSetList{}
+	if err := reader.List(ctx, list, client.InNamespace(dgd.Namespace)); err != nil {
+		if meta.IsNoMatchError(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("list owned PodCliqueSets: %w", err)
+	}
+
+	// Accept only the controller reference as authoritative ownership evidence.
+	for i := range list.Items {
+		if metav1.IsControlledBy(&list.Items[i], dgd) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func providerFromCurrentIntent(
+	gate features.Gate,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) workloadProvider {
+	// Grove is the default when its gate is enabled unless the DGD opts out.
+	if gate.Enabled(features.Grove) &&
+		strings.ToLower(dgd.Annotations[consts.KubeAnnotationEnableGrove]) != consts.KubeLabelValueFalse {
+		return workloadProviderGrove
+	}
+	return workloadProviderComponent
+}
+
+func parseWorkloadProvider(value string) (workloadProvider, error) {
+	// Accept only the workload programs implemented by this controller.
+	switch workloadProvider(value) {
+	case workloadProviderComponent, workloadProviderGrove:
+		return workloadProvider(value), nil
+	default:
+		return "", fmt.Errorf("%w: %q", errUnsupportedWorkloadProvider, value)
+	}
+}
--- a/deploy/operator/internal/controller/dynamographdeployment_controller.go
+++ b/deploy/operator/internal/controller/dynamographdeployment_controller.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/ai-dynamo/dynamo/deploy/operator/internal/secret"
 
@@ -39,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	configv1alpha1 "github.com/ai-dynamo/dynamo/deploy/operator/api/config/v1alpha1"
 	nvidiacomv1alpha1 "github.com/ai-dynamo/dynamo/deploy/operator/api/v1alpha1"
@@ -52,12 +52,14 @@ import (
 )
 
 const (
-	reasonFailedToInitializeWorkerHash Reason = "failed_to_initialize_worker_hash"
-	reasonFailedToMigrateWorkerHash    Reason = "failed_to_migrate_worker_hash"
-	reasonNoMultinodeOrchestrator      Reason = "no_multinode_orchestrator_available"
-	reasonFailedToReconcileResources   Reason = "failed_to_reconcile_the_resources"
-	reasonRollingUpdateFailed          Reason = "rolling_update_failed"
-	reasonWaitingForCheckpoint         Reason = "waiting_for_checkpoint"
+	reasonFailedToInitializeWorkerHash        Reason = "failed_to_initialize_worker_hash"
+	reasonFailedToMigrateWorkerHash           Reason = "failed_to_migrate_worker_hash"
+	reasonNoMultinodeOrchestrator             Reason = "no_multinode_orchestrator_available"
+	reasonFailedToReconcileResources          Reason = "failed_to_reconcile_the_resources"
+	reasonRollingUpdateFailed                 Reason = "rolling_update_failed"
+	reasonWaitingForCheckpoint                Reason = "waiting_for_checkpoint"
+	reasonSelectedWorkloadProviderUnavailable Reason = "selected_workload_provider_unavailable"
+	reasonUnsupportedWorkloadProvider         Reason = "unsupported_workload_provider"
 
 	dgdComponentPodIndex = ".metadata.dgdComponent"
 )
@@ -129,6 +131,24 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, err
 	}
 
+	// Lock in the provider before allowing any other reconciliation effects.
+	provider, err := r.ensureWorkloadProvider(ctx, dynamoDeployment)
+
+	// Keep adoption and patch failures retryable while making an invalid stored selection terminal.
+	if err != nil {
+		if !errors.Is(err, errUnsupportedWorkloadProvider) {
+			return ctrl.Result{}, err
+		}
+
+		// Persist the invalid selection diagnosis before suppressing automatic retries.
+		programResult := newWorkloadProgramResult(dynamoDeployment)
+		programResult.Fail(dynamoDeployment.Generation, reasonUnsupportedWorkloadProvider, err)
+		if statusErr := r.persistWorkloadProgramResult(ctx, dynamoDeployment, programResult); statusErr != nil {
+			return programResult.Result, statusErr
+		}
+		return programResult.Result, reconcile.TerminalError(err)
+	}
+
 	// Reject unsupported stored configurations before any primary-resource mutation.
 	var compatibilityErrs []error
 	for i := range dynamoDeployment.Spec.Components {
@@ -161,7 +181,11 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, err
 	}
 
-	program := r.selectWorkloadProgram(dynamoDeployment)
+	// Dispatch exclusively through the persisted provider.
+	program, err := r.selectWorkloadProgram(provider)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 	programResult, programErr := program.Reconcile(ctx, workloadProgramRequest{
 		DGD: dynamoDeployment,
 	})
@@ -194,11 +218,6 @@ func (r *DynamoGraphDeploymentReconciler) persistWorkloadProgramResult(
 		}
 	}
 	return nil
-}
-
-func (r *DynamoGraphDeploymentReconciler) isGrovePathway(dgd *nvidiacomv1beta1.DynamoGraphDeployment) bool {
-	return r.RuntimeConfig.Gate.Enabled(features.Grove) && (dgd.Annotations == nil ||
-		strings.ToLower(dgd.Annotations[consts.KubeAnnotationEnableGrove]) != consts.KubeLabelValueFalse)
 }
 
 func (r *DynamoGraphDeploymentReconciler) FinalizeResource(ctx context.Context, dynamoDeployment *nvidiacomv1beta1.DynamoGraphDeployment) error {
@@ -287,9 +306,12 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
 		}))
 	}
+
+	// Register Grove-owned workload watches only when the Grove feature is enabled.
 	if r.RuntimeConfig.Gate.Enabled(features.Grove) {
 		ctrlBuilder = newGroveWatchSetup(r.Client).addTo(ctrlBuilder)
 	}
+
 	// Wrap with metrics collection
 	observedReconciler := observability.NewObservedReconciler(r, consts.ResourceTypeDynamoGraphDeployment)
 	return ctrlBuilder.Complete(observedReconciler)
--- a/deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
+++ b/deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
@@ -55,9 +55,10 @@ func NewDGDDefaulter(operatorVersion string) *DGDDefaulter {
 
 // Default implements admission.CustomDefaulter.
 // On every operation: defaults nil Replicas to 1 for all components.
-// On every Grove-pathway operation: defaults nil MinAvailable to 1. Scaling to
-// replicas=0 does not rewrite MinAvailable; it remains the component's
-// configured minimum viable unit.
+// On CREATE: sets the controller-owned workload provider from routing intent before provider-specific defaults.
+// Existing unannotated DGDs remain unselected for controller-side workload adoption.
+// On the Grove pathway: defaults nil MinAvailable to 1. Scaling to replicas=0
+// does not rewrite MinAvailable; it remains the component's configured minimum viable unit.
 // On CREATE: stamps nvidia.com/dynamo-operator-origin-version with the operator version.
 // On UPDATE/DELETE: the origin version annotation is immutable once set.
 func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
@@ -78,26 +79,26 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 		return nil
 	}
 
-	// Default nil replicas to 1 for all components. The Replicas field is
-	// *int32 with omitempty, so users can legally omit it. Without this
-	// default the controller panics on a nil pointer dereference in
-	// expandRolesForComponent(). Apply on every operation so that components
-	// added via UPDATE also get the default.
-	grovePathway := d.isGrovePathway(ctx, dgd)
+	// Resolve the authoritative or creation-time provider before applying component defaults.
+	provider, providerSelected := defaultWorkloadProvider(ctx, dgd, req.Operation)
+
+	// Default nil replicas on every operation so newly added components remain safe to expand.
 	for i := range dgd.Spec.Components {
 		component := &dgd.Spec.Components[i]
+
+		// Default omitted replica counts before the controller expands component roles.
 		if component.Replicas == nil {
 			component.Replicas = ptr.To(int32(1))
 		}
-		if grovePathway && component.MinAvailable == nil {
+
+		// Default Grove's minimum available replicas only for Grove-selected DGDs.
+		if providerSelected && provider == consts.WorkloadProviderGrove && component.MinAvailable == nil {
 			component.MinAvailable = ptr.To(int32(1))
 		}
 	}
 
+	// Stamp creation provenance independently from level-based provider defaulting.
 	if req.Operation == admissionv1.Create {
-		if dgd.Annotations == nil {
-			dgd.Annotations = make(map[string]string)
-		}
 		// Stamp operator version on creation (don't overwrite if already set)
 		if _, exists := dgd.Annotations[consts.KubeAnnotationDynamoOperatorOriginVersion]; !exists {
 			dgd.Annotations[consts.KubeAnnotationDynamoOperatorOriginVersion] = d.OperatorVersion
@@ -111,9 +112,34 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 	return nil
 }
 
-func (d *DGDDefaulter) isGrovePathway(ctx context.Context, dgd *nvidiacomv1beta1.DynamoGraphDeployment) bool {
-	return features.MustGateFrom(ctx).Enabled(features.Grove) && (dgd.Annotations == nil ||
-		strings.ToLower(dgd.Annotations[consts.KubeAnnotationEnableGrove]) != consts.KubeLabelValueFalse)
+func defaultWorkloadProvider(
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+	operation admissionv1.Operation,
+) (string, bool) {
+	// Keep existing selections authoritative and leave legacy updates for controller adoption.
+	if operation != admissionv1.Create {
+		if provider, exists := dgd.Annotations[consts.KubeAnnotationWorkloadProvider]; exists {
+			return provider, true
+		}
+		return "", false
+	}
+
+	// Derive every new DGD from user-facing routing intent, ignoring the controller-owned annotation.
+	provider := consts.WorkloadProviderComponent
+
+	// Select Grove when it is enabled and the DGD has not opted out.
+	if features.MustGateFrom(ctx).Enabled(features.Grove) &&
+		strings.ToLower(dgd.Annotations[consts.KubeAnnotationEnableGrove]) != consts.KubeLabelValueFalse {
+		provider = consts.WorkloadProviderGrove
+	}
+
+	// Allocate annotation storage before materializing the selected provider.
+	if dgd.Annotations == nil {
+		dgd.Annotations = make(map[string]string)
+	}
+	dgd.Annotations[consts.KubeAnnotationWorkloadProvider] = provider
+	return provider, true
 }
 
 // RegisterWithManager registers the defaulting webhook with the manager.
--- a/deploy/operator/internal/webhook/validation/dynamographdeployment.go
+++ b/deploy/operator/internal/webhook/validation/dynamographdeployment.go
@@ -196,6 +196,16 @@ func (v *dynamoGraphDeploymentValidation) validateObjectMeta(
 		))
 	}
 
+	// Restrict the durable workload provider to programs implemented by the controller.
+	if value, exists := objectMeta.Annotations[consts.KubeAnnotationWorkloadProvider]; exists &&
+		value != consts.WorkloadProviderComponent && value != consts.WorkloadProviderGrove {
+		allErrs = append(allErrs, field.NotSupported(
+			annotationsPath.Key(consts.KubeAnnotationWorkloadProvider),
+			value,
+			[]string{consts.WorkloadProviderComponent, consts.WorkloadProviderGrove},
+		))
+	}
+
 	if hasIntraPodFailover && objectMeta.Annotations[consts.KubeAnnotationDynamoKubeDiscoveryMode] != "container" {
 		allErrs = append(allErrs, field.Invalid(
 			annotationsPath.Key(consts.KubeAnnotationDynamoKubeDiscoveryMode),
@@ -499,6 +509,11 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeploymentUpdate(
 	oldDGD *nvidiacomv1beta1.DynamoGraphDeployment,
 ) field.ErrorList {
 	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, v.validateObjectMetaUpdate(
+		&newDGD.ObjectMeta,
+		&oldDGD.ObjectMeta,
+		field.NewPath("metadata"),
+	)...)
 	allErrs = append(allErrs, v.validateDynamoGraphDeploymentSpecUpdate(
 		&newDGD.Spec,
 		&oldDGD.Spec,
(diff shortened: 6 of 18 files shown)


deploy/operator/internal/controller/dgd_grove_program.go
```diff
@@ -73,6 +76,15 @@ func (p *groveProgram) Reconcile(
...
 	programResult = newWorkloadProgramResult(req.DGD)
+	if !p.available {
+		err := failWorkloadProgram(
+			reasonSelectedWorkloadProviderUnavailable,
+			fmt.Errorf("selected workload provider %q is unavailable because Grove is disabled", workloadProviderGrove),
+		)
+		programResult.Fail(req.DGD.Generation, reasonSelectedWorkloadProviderUnavailable, err)
+		return programResult, err
+	}
+
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. Added the story comment directly above the guard: a durable Grove selection fails when Grove is unavailable rather than falling back to the component provider. The reconciliation behavior is unchanged.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -160,7 +167,37 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
...
+		programResult.Fail(dynamoDeployment.Generation, reasonFailedToSelectWorkloadProvider, selectionErr)
+		if statusErr := r.persistWorkloadProgramResult(ctx, dynamoDeployment, programResult); statusErr != nil {
+			logger.Error(statusErr, "unable to update status after workload provider selection failure")
+			return ctrl.Result{}, statusErr
+		}
+		return ctrl.Result{}, selectionErr
+	}
+	if selection.NewlyPersisted {
+		return ctrl.Result{RequeueAfter: workloadProviderPersistenceRequeueDelay}, nil
+	}
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. Wrapped optimistic-lock conflicts are now detected with `apierrors.IsConflict(selectionErr)` and treated as pending cache convergence: the controller requeues without writing a Failed status or returning an error. Non-conflict selection failures retain the existing failure path. Added `TestDynamoGraphDeploymentReconcileTreatsProviderConflictAsPending`.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -497,6 +550,11 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeploymentUpdate(
 	oldDGD *nvidiacomv1beta1.DynamoGraphDeployment,
 ) field.ErrorList {
 	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, v.validateObjectMetaUpdate(
+		&newDGD.ObjectMeta,
+		&oldDGD.ObjectMeta,
+		field.NewPath("metadata"),
+	)...)
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. Terminating updates now skip complete new-state/spec validation but still run metadata update validation. This preserves finalizer removal while continuing to reject replacement or removal of the selected provider. Added handler coverage for finalizer removal and an admission envtest regression for provider removal during termination.

deploy/operator/internal/controller/dynamographdeployment_controller.go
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. The durable persist-and-requeue boundary is retained, but a wrapped optimistic-lock conflict caused by cache lag is now a benign pending outcome: the controller requeues without publishing a failure status or returning an error. A focused regression covers this path.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -520,6 +578,43 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeploymentUpdate(
...
+			providerPath,
+			invalidValue,
+			apivalidation.FieldImmutableErrorMsg,
+		))
+	case !oldProviderExists && newProviderExists &&
+		!isOperatorPrincipalRequest(v.userInfo, v.operatorPrincipal):
+		allErrs = append(allErrs, field.Forbidden(
+			providerPath,
+			"is controller-owned and may only be set by the Dynamo operator",
+		))
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. When `operatorPrincipal` is empty, validation now skips only the initial absent-to-selected identity check so the documented optional-identity mode can cross this mandatory persistence boundary. A configured principal remains strictly enforced, and subsequent provider replacement or removal remains immutable in either mode. Added handler coverage for the empty-principal case.

deploy/operator/internal/controller/dgd_workload_provider_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+		ctx,
+		list,
+		client.InNamespace(dgd.Namespace),
+		client.MatchingLabels{consts.KubeLabelDynamoGraphDeploymentName: dgd.Name},
+	); err != nil {
+		if meta.IsNoMatchError(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("list Grove PodCliqueSets owned by DynamoGraphDeployment %s/%s: %w", dgd.Namespace, dgd.Name, err)
+	}
```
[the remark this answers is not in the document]
julienmancuso: Thanks for checking this. The lookup is intentional for one-time adoption: an owned PodCliqueSet must remain discoverable even when the current Grove gate is disabled, otherwise a gate change could silently reclassify a legacy DGD. It runs only while the durable annotation is absent, and the cache reuses the informer for the kind rather than creating one per DGD. Also, this tree uses `k8s.io/apimachinery` v0.36.2, where `meta.IsNoMatchError` delegates to `errors.Is`. The new regression returns a wrapped `NoKindMatchError` and verifies component selection succeeds when the Grove API is absent.

deploy/operator/internal/webhook/validation/dynamographdeployment_helpers.go
```diff
@@ -137,6 +138,27 @@ func grovePathwayForDynamoGraphDeployment(
...
+				provider,
+			)
+		default:
+			return false, fmt.Sprintf(
+				"requires the Grove pathway, but annotation %q has unsupported value %q",
+				consts.KubeAnnotationSelectedWorkloadProvider,
+				provider,
+			)
+		}
+	}
```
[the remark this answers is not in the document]
julienmancuso: This is intentional. The persisted or adopted provider is authoritative independently of the current feature gate; the gate controls provider availability, not the semantics of a DGD already known to use Grove. An owned PodCliqueSet is strong adoption evidence, and disabling Grove must not reinterpret that DGD as component or relax Grove-specific admission rules. A selected Grove provider therefore retains Grove admission semantics and reports provider unavailable when Grove cannot run. The admission envtests explicitly cover retained Grove semantics with the gate disabled.

deploy/operator/internal/controller/dgd_workload_provider_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	// authoritative ownership check.
+	list := &nvidiacomv1beta1.DynamoComponentDeploymentList{}
+	if err := r.client.List(
+		ctx,
+		list,
+		client.InNamespace(dgd.Namespace),
+		client.MatchingLabels{consts.KubeLabelDynamoGraphDeploymentName: dgd.Name},
+	); err != nil {
+		return false, fmt.Errorf("list DynamoComponentDeployments owned by DynamoGraphDeployment %s/%s: %w", dgd.Namespace, dgd.Name, err)
+	}
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 637862cbf8. Both adoption probes now list the namespace cache without label selectors and use only `metav1.IsControlledBy` as the authoritative relationship. Regressions cover unlabeled owned DCDs, unlabeled owned PodCliqueSets, mixed unlabeled/labeled owned workloads, and foreign labeled or unlabeled workloads.

julienmancuso:
The two review-body-only follow-ups are also addressed in 637862cbf8: a failed provider patch restores the original in-memory annotations (including nil), and provider selection now has a regression for a cluster without the Grove CRD using a wrapped `NoKindMatchError`. Focused controller tests, webhook handler tests, DGD admission envtests, and the operator linter target pass.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -109,6 +133,17 @@ func (v *DynamoGraphDeploymentValidator) ValidateUpdate(
 		operatorPrincipal: operatorPrincipal,
 	}
 
+	// Limit terminating updates to metadata invariants so unrelated validation
```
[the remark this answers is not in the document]
julienmancuso: Good catch. This is valid: the terminating path must preserve both the supported-value invariant and immutability for the controller-owned provider. I will add targeted selected-provider value validation before the metadata update check. I will keep unrelated metadata and spec validation skipped so legacy validation failures cannot block finalizer removal.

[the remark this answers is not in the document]
julienmancuso: Addressed in afe825adcc. The terminating-update path now runs targeted selected-provider value validation before the metadata update/immutability check. This rejects unsupported provider values while continuing to skip unrelated metadata and spec validation so finalizer removal remains safe. Added the envtest case "selected workload provider value is validated during termination"; the complete DGD admission envtest and exact operator linter target pass.

deploy/operator/internal/controller/dgd_grove_program.go
```diff
@@ -73,6 +76,17 @@ func (p *groveProgram) Reconcile(
...
+	// Fail a durable Grove selection when Grove is unavailable rather than falling back.
+	if !p.available {
+		err := failWorkloadProgram(
+			reasonSelectedWorkloadProviderUnavailable,
+			fmt.Errorf("selected workload provider %q is unavailable because Grove is disabled", workloadProviderGrove),
+		)
+		programResult.Fail(req.DGD.Generation, reasonSelectedWorkloadProviderUnavailable, err)
+		return programResult, err
+	}
```
sttts, line 89: an error will lead to lead to retries and at best to error messages in the operator log. But this error is not transient. Do we have a condition we could set instead? That would be user-visible.

sttts: Yes—the reconciler already persists this error. `workloadProgramResult.Fail` sets `status.state=failed` and upserts `Ready=False` with the current generation, reason (`selected_workload_provider_unavailable` here), and message. The outer reconciler writes that status before returning the program error, so the failure is user-visible. The retry concern still applies: this is returned as an ordinary error, so controller-runtime rate-limits and requeues it. For this non-transient case we should preserve the condition and return `reconcile.TerminalError(err)` (or otherwise carry retryability separately).

sttts: I think we should make the error semantics explicit here. There are two independent questions:

1. Should this failure be persisted in the DGD status?
2. Should controller-runtime automatically retry it?

Today those decisions are effectively coupled. In the Grove program, every returned error is converted into `state=failed` and `Ready=False` by the deferred `programResult.Fail(...)`. The outer reconciler then persists that result and returns the error, which makes controller-runtime rate-limit and requeue it. Consequently, transient API failures and durable configuration failures currently have the same user-visible and retry behavior. Provider-selection conflicts are already a special exception.

I suggest retaining the existing idiomatic `(workloadProgramResult, error)` API and classifying errors through two predicates:

```go
reconcileerrors.IsTransient(err)
reconcileerrors.IsTerminal(err)
```

The predicates should be mutually exclusive. If both return `false`, the error is intentionally considered unclassified.

| Classification | Persist `Ready=False` | Automatic retry |
|---|---:|---:|
| Transient | No | Yes |
| Unclassified | Yes | Yes |
| Terminal | Yes | No |

The unclassified case is important. A binary `IsTransient` decision would otherwise interpret every unknown error as permanent. Unknown failures should remain visible, but we should continue retrying until somebody has intentionally classified them.

Examples for this controller:

- Transient: Kubernetes API conflict, server timeout, rate limiting, or another temporary API/server failure.
- Terminal: the DGD is durably selected to Grove while Grove is disabled; the stored provider value is unsupported; or provider adoption finds both owned DCD and PodCliqueSet resource families.
- Unclassified: any unexpected error for which we have not made an explicit retry decision.

“Terminal” must be scoped carefully. It is not an intrinsic property of an arbitrary low-level error; it means that retrying this reconciliation with the same object and configuration cannot make progress. Shared clients and low-level helpers therefore should not return terminal errors. They should return ordinary or domain-specific errors. The workload program or DGD reconciler should mark an error terminal only at the boundary where it owns the retry decision, after any local recovery or fallback has been exhausted.

For the Grove program, the failure handling could conceptually become:

```go
defer func() {
	if retErr != nil && !reconcileerrors.IsTransient(retErr) {
		reason := reasonFailedToReconcileResources
		if classified, ok := workloadProgramFailureReason(retErr); ok {
			reason = classified
		}
		programResult.Fail(req.DGD.Generation, reason, retErr)
	}

	p.topology.Reconcile(ctx, req.DGD, &programResult)
}()
```

This keeps the existing contract that the selected workload program owns its complete status and `Ready` condition:

- transient errors leave the previous `Ready` condition untouched;
- unclassified errors produce the existing generic failure condition;
- terminal errors produce a specific durable failure condition.

A known terminal failure can be marked at the workload-program boundary using controller-runtime’s terminal marker:

```go
err := failWorkloadProgram(
	reasonSelectedWorkloadProviderUnavailable,
	fmt.Errorf(
		"selected workload provider %q is unavailable because Grove is disabled",
		workloadProviderGrove,
	),
)
return programResult, reconcile.TerminalError(err)
```

`IsTerminal` can recognize the controller-runtime marker through the error chain. `IsTransient` must first exclude terminal errors so that a terminal wrapper around another error cannot make both predicates return `true`.

The outer DGD reconciler should preserve the current ordering:

1. receive the program result and error;
2. persist the complete program status;
3. only after successful status persistence, return the program error.

That ordering is particularly important for terminal errors. If persisting `Ready=False` fails, the reconciler must return the status-write error as an ordinary retryable error. It must not return a terminal error before the diagnosis is durable.

Conceptually:

```go
programResult, programErr := program.Reconcile(ctx, request)
result := programResult.Result

if statusErr := r.persistWorkloadProgramResult(
	ctx,
	dynamoDeployment,
	programResult,
); statusErr != nil {
	// Always retry a failed status write, even if programErr was terminal.
	return result, statusErr
}

return result, programErr
```

If the program returns controller-runtime’s terminal marker directly, no additional result type or translation is necessary. Controller-runtime will log and meter the error but will not add the DGD back to the rate-limited queue. A later watch event, object update, or controller restart can still cause a new reconciliation.

The provider-selection and pre-program error paths should use the same classification rules. The existing conflict-specific short requeue can remain because it is an intentional transient policy.

I would add focused tests for the three cases:

1. A transient error does not replace the existing `Ready` condition and is returned for retry.
2. An unclassified error persists `Ready=False` and is returned as an ordinary error.
3. A terminal error persists `Ready=False` with the specific reason and remains recognizable as terminal after status persistence.
4. If status persistence fails for a terminal program error, the returned error is the status-write error and is not terminal.
5. Wrapping with `%w` preserves classification, and `IsTransient` and `IsTerminal` can never both return `true`.

This gives us deliberate terminal and transient behavior without changing the normal Go `Result, error` propagation model.

julienmancuso: Addressed in e95992f162 with the explicit three-way semantics you proposed.

- Added `reconcileerrors.IsTransient` and `IsTerminal`; they work through `%w`, and terminal always takes precedence so they cannot both be true.
- Component and Grove programs preserve the existing `Ready` condition for transient Kubernetes API failures. Unclassified errors still publish `Ready=False` and remain retryable.
- A selected-but-unavailable Grove provider publishes `Ready=False` with `selected_workload_provider_unavailable` and returns `reconcile.TerminalError`.
- Unsupported stored providers and mixed owned DCD/PodCliqueSet families remain domain errors below the DGD boundary and are marked terminal only where the DGD reconciler owns retry policy. Provider conflicts retain the existing short requeue, and other transient provider-selection failures do not overwrite `Ready`.
- The outer ordering remains status first, error second. If the status write fails, that ordinary retryable write error replaces any terminal program error.
- The same policy now applies to other pre-program failures; invalid stored checkpoint configurations are reported durably and returned terminally.

Focused regressions cover transient, unclassified, terminal, terminal-plus-status-write-failure, `%w` preservation, and predicate exclusivity. The exact ARM64 linter and tester Docker targets pass; the tagged cluster-test package also compiles. The unrelated cluster integration timeout from the previous run was traced to its in-process controller identity not matching the production service-account principal configured in the webhook harness; the harness now uses the already-supported optional-identity mode while strict identity enforcement remains covered by admission tests.

julienmancuso:
@coderabbitai full review

deploy/operator/internal/controller/clusterenv_test.go
```diff
@@ -63,11 +63,11 @@ func TestMain(m *testing.M) {
...
 		SetupWebhooks: func(mgr ctrl.Manager) error {
+			// In-process cluster-test controllers use the admin kubeconfig, so exercise the supported optional-identity mode.
 			if err := webhooksetup.Setup(mgr, webhooksetup.Options{
-				Config:            operatorConfig,
-				RuntimeConfig:     runtimeConfig,
-				OperatorVersion:   operatorVersion,
-				OperatorPrincipal: "system:serviceaccount:dynamo-system:dynamo-operator",
+				Config:          operatorConfig,
+				RuntimeConfig:   runtimeConfig,
+				OperatorVersion: operatorVersion,
```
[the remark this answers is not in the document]
julienmancuso: Verified against the current cluster suite. I am leaving this unchanged intentionally: these scenarios do not register the DynamoGraphDeploymentScalingAdapter controller, and none of their inputs enable a scaling adapter. The empty principal is required because the in-process controllers authenticate with the admin kubeconfig when they persist the mandatory provider selection. Configuring planner-serviceaccount as the operator principal would not authorize that admin request and would recreate the integration failure.

If cluster-level scaling-adapter coverage is added, the harness should run or impersonate the operator service-account identity instead of treating the planner service account as the operator. Replica-update authorization for both the operator and planner identities remains covered by the admission tests.

deploy/operator/internal/controller/dgd_workload_provider_reconciler_test.go
```diff
@@ -0,0 +1,537 @@
...
+			) error {
+				return patchErr
+			},
+		}).
+		Build()
+	var requestDGD nvidiacomv1beta1.DynamoGraphDeployment
+	require.NoError(t, kubeClient.Get(t.Context(), client.ObjectKeyFromObject(dgd), &requestDGD))
+	dgd = &requestDGD
+	reconciler := newDGDWorkloadProviderReconciler(kubeClient, features.Gates{})
+
```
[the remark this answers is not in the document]
julienmancuso: Addressed in b0dcce047d. Each of the three setup blocks now starts with a t.Log heading describing the fixture and client/program being constructed. The ARM64 linter and tester targets pass.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -520,6 +606,45 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeploymentUpdate(
...
+		!isOperatorPrincipalRequest(v.userInfo, v.operatorPrincipal):
+		allErrs = append(allErrs, field.Forbidden(
+			providerPath,
+			"is controller-owned and may only be set by the Dynamo operator",
+		))
+	}
+
+	return allErrs
+}
+
```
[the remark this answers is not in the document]
julienmancuso: Addressed in b0dcce047d. The raw manager deployment now populates POD_SERVICE_ACCOUNT from spec.serviceAccountName and POD_NAMESPACE from metadata.namespace, matching the Helm deployment and enabling strict operator-principal authorization for the shipped Kustomize deployment. The ARM64 linter and tester targets pass.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -112,6 +113,12 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 }
 
 func (d *DGDDefaulter) isGrovePathway(ctx context.Context, dgd *nvidiacomv1beta1.DynamoGraphDeployment) bool {
+	// Once selected, the durable provider is authoritative even if the feature
+	// gate or the original user intent later differs.
+	if provider, exists := dgd.Annotations[consts.KubeAnnotationSelectedWorkloadProvider]; exists {
```
panpan0000: Could we avoid applying Grove-only defaults on UPDATE while provider selection is still absent? I think there is a narrow upgrade race here: an existing DCD-backed DGD can be updated after the webhooks come up but before adoption runs. With Grove enabled, this path admits `minAvailable=1` under prospective Grove semantics. Adoption then selects `component` from the owned DCD, but the operator's patch adding `selected-workload-provider=component` appears to fail validation because `minAvailable` is now forbidden for the component pathway. Would it make sense to use prospective routing only on CREATE, and leave provider-specific UPDATE defaulting until the selected annotation has been persisted? A regression test for an old DGD with an owned DCD and no selected annotation might help here.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -146,21 +156,68 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
...
+			return ctrl.Result{}, selectionErr
+		}
+
+		programResult := newWorkloadProgramResult(dynamoDeployment)
+		programResult.Fail(dynamoDeployment.Generation, reasonFailedToSelectWorkloadProvider, selectionErr)
+		if statusErr := r.persistWorkloadProgramResult(ctx, dynamoDeployment, programResult); statusErr != nil {
+			logger.Error(statusErr, "unable to update status after workload provider selection failure")
+			return ctrl.Result{}, statusErr
+		}
+		if isTerminalWorkloadProviderSelectionError(selectionErr) {
```
panpan0000: Could we check how this terminal state is expected to recover when Grove is disabled? Adoption still lists PodCliqueSets in that mode, so mixed DCD+PCS state can reach here, but `SetupWithManager` only installs the Grove watches when the gate is enabled. If the stale PCS is deleted to resolve the conflict, that deletion will not enqueue the DGD, and `TerminalError` will not retry it. It looks like the status may remain stuck until the DGD is touched or the operator restarts. Would a periodic requeue for this case, or a PCS recovery watch based on API availability rather than the feature gate, make sense?

panpan0000 (commented):
A couple of upgrade/recovery edge cases I wanted to check:

deploy/operator/internal/controller/dynamographdeployment_controller.go
julienmancuso: Addressed in 73015ca3a7 with the API-availability watch approach. Startup now resolves the exact namespaced PodCliqueSet REST mapping independently of whether the Grove feature is enabled and records that capability in RuntimeConfig. When Grove is disabled but PodCliqueSet is available, the DGD controller installs a deletion-only owned-PCS recovery watch. Deleting the stale PCS therefore re-enqueues its DGD, allowing adoption to leave the terminal mixed-provider state and select the remaining DCD/component provider. Tests cover exact namespaced mapping availability and the deletion-only recovery predicates; the full ARM64 linter/tester targets pass.

panpan0000: Thanks. This fix looks good to me.

deploy/operator/internal/consts/consts.go
```diff
@@ -42,6 +42,12 @@ const (
 
 	KubeAnnotationEnableGrove = "nvidia.com/enable-grove"
 
+	// KubeAnnotationSelectedWorkloadProvider records the graph-level workload
+	// provider selected by the operator. The value is immutable after selection.
+	KubeAnnotationSelectedWorkloadProvider = "nvidia.com/selected-workload-provider"
+	WorkloadProviderComponent              = "component"
```
sttts: what about LWS and deployment? LWS also depends on the installed CRD today, doesn't it? Will we need another similar PR for DCD?

deploy/operator/internal/consts/consts.go
```diff
@@ -42,6 +42,12 @@ const (
 
 	KubeAnnotationEnableGrove = "nvidia.com/enable-grove"
 
+	// KubeAnnotationSelectedWorkloadProvider records the graph-level workload
+	// provider selected by the operator. The value is immutable after selection.
+	KubeAnnotationSelectedWorkloadProvider = "nvidia.com/selected-workload-provider"
```
sttts: nit: drop the `selected-`.

deploy/operator/internal/controller/dgd_grove_program.go
```diff
@@ -33,6 +35,7 @@ type groveProgram struct {
 	workloads       *groveWorkloadsReconciler
 	scalingAdapters *dgdScalingAdaptersReconciler
 	topology        *dgdGroveTopologyConditionReconciler
+	available       bool
```
sttts: is this more than a feature gate? Shouldn't use the feature gate code?

deploy/operator/internal/controller/dgd_grove_watch_setup.go
```diff
@@ -72,6 +73,24 @@ func (s *groveWatchSetup) addTo(ctrlBuilder *builder.Builder) *builder.Builder {
...
+}
+
+func podCliqueSetRecoveryEventPredicates() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc:  func(event.CreateEvent) bool { return false },
+		DeleteFunc:  func(event.DeleteEvent) bool { return true },
+		UpdateFunc:  func(event.UpdateEvent) bool { return false },
+		GenericFunc: func(event.GenericEvent) bool { return false },
+	}
+}
```
sttts: nit: 
```
var onlyDelete = predicate.Not(predicate.Funcs{
	DeleteFunc: func(event.DeleteEvent) bool { return false },
})
```
does not need a unit test and is clearer when used in `addPodCliqueSetRecoveryTo`.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -271,9 +323,16 @@ func (r *DynamoGraphDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) err
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
 		}))
 	}
+
+	groveWatches := newGroveWatchSetup(r.Client)
 	if r.RuntimeConfig.Gate.Enabled(features.Grove) {
-		ctrlBuilder = newGroveWatchSetup(r.Client).addTo(ctrlBuilder)
+		ctrlBuilder = groveWatches.addRuntimeTo(ctrlBuilder)
```
sttts: what's the reason we have `GroveWatchSetup` for runtime and pcs and not just inline it here as everywhere else?

deploy/operator/internal/controller_common/runtime.go
```diff
@@ -24,6 +24,9 @@ import "github.com/ai-dynamo/dynamo/deploy/operator/internal/features"
 type RuntimeConfig struct {
 	// Gate contains the resolved operator features.
 	Gate features.Gates
+	// GrovePodCliqueSetAPIAvailable reports whether the concrete PodCliqueSet API was
+	// detected, independently of whether Grove integration is enabled.
+	GrovePodCliqueSetAPIAvailable bool
```
sttts: this is an anti-pattern. I think we wanted gates to be the only place where we detect features including API availability.

deploy/operator/internal/features/gates.go
```diff
@@ -153,66 +157,86 @@ type Gates struct {
 	GPUDiscovery     bool `json:"gpuDiscovery"`
 }
 
+// Resolution contains configured feature gates and raw API capabilities that
+// remain relevant even when their corresponding integration is disabled.
+type Resolution struct {
+	Gates                         Gates
+	GrovePodCliqueSetAPIAvailable bool
```
sttts: same anti-pattern. This should not be special.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -78,19 +80,34 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 		return nil
 	}
 
+	// Preserve an existing minAvailable on an unselected UPDATE without deriving
+	// a new provider-specific default before legacy workload adoption completes.
+	var oldDGD *nvidiacomv1beta1.DynamoGraphDeployment
+	if _, selected := dgd.Annotations[consts.KubeAnnotationSelectedWorkloadProvider]; req.Operation == admissionv1.Update && !selected {
```
sttts: never do edge-based logic in a reconiler by guarding by `req.Operation == admissionv1.Update`. Reconcilers should be level based. Otherwise, it is no reconciler.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -78,19 +80,34 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
...
+		oldDGD, err = decodeOldDGD(req)
+		if err != nil {
+			return err
+		}
```
sttts: I think we should immediate persist a missing annotation to avoid that we roll out and then the final update fails. Risk is small that we decide differently, but it's more correct to persist to avoid the race altogether.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
sttts: please don't do edge based code in reconcilers. It's an anti-pattern.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -111,7 +128,35 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
...
+) bool {
+	// Once selected, the durable provider is authoritative even if the feature
+	// gate or the original user intent later differs.
+	if provider, exists := dgd.Annotations[consts.KubeAnnotationSelectedWorkloadProvider]; exists {
+		return provider == consts.WorkloadProviderGrove
+	}
+
+	// Prospective routing is safe only on CREATE. An unselected legacy DGD may
+	// still be adopted from existing workloads under a different provider.
+	if operation != admissionv1.Create {
```
sttts, line 121: please don't do edge based code in reconcilers. It's an anti-pattern.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -92,9 +92,33 @@ func (v *DynamoGraphDeploymentValidator) Validate(
 	return validation.warnings, invalidDynamoGraphDeploymentError(deployment, allErrs)
 }
 
+// ValidateCreate performs complete create validation on a v1beta1 DGD.
+// ctx and deployment must not be nil.
+func (v *DynamoGraphDeploymentValidator) ValidateCreate(
```
sttts: don't like this. It's brittle to have two validation chains. Can we just check for freshness of the object where we are checking the annotation? I.e. very local vs. start two global new chains.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -109,6 +133,22 @@ func (v *DynamoGraphDeploymentValidator) ValidateUpdate(
 		operatorPrincipal: operatorPrincipal,
 	}
 
+	// Limit terminating updates to metadata invariants so unrelated validation
+	// does not block finalizer removal while controller-owned state remains protected.
```
sttts: this whole section cnofuses me. Why should we behave differently during deletion? Doesn't look right.

julienmancuso:
@coderabbitai full review

deploy/operator/internal/controller/dgd_workload_provider.go
```diff
@@ -0,0 +1,84 @@
...
+	return workloadProviderComponent
+}
+
+func parseWorkloadProvider(value string) (workloadProvider, error) {
+	switch workloadProvider(value) {
+	case workloadProviderComponent, workloadProviderGrove:
+		return workloadProvider(value), nil
+	default:
+		return "", fmt.Errorf("unsupported workload provider %q", value)
+	}
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 009de123276a. Added concise story comments for the accepted persisted-provider values, durable provider-to-program dispatch, and Grove watch registration. The related changed defaulting blocks now have story comments as well. The full affected Go package tests and the ARM64 linter/tester targets pass.

deploy/operator/internal/controller/dynamographdeployment_controller.go
```diff
@@ -128,6 +128,15 @@ func (r *DynamoGraphDeploymentReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, err
 	}
 
+	// Lock in the provider before allowing any other reconciliation effects.
+	provider, persisted, err := r.ensureWorkloadProvider(ctx, dynamoDeployment)
+	if err != nil {
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
julienmancuso: Addressed in 009de123276a. Unsupported stored provider values now use a recognizable sentinel error, publish `Ready=False` with reason `unsupported_workload_provider`, persist that status, and only then return `reconcile.TerminalError`. A status-write failure is returned as an ordinary retryable error, while provider patch failures remain retryable. Regression coverage exercises both the successful terminal path and the status-write-failure path.

julienmancuso:
The review-body-only follow-ups are addressed in 009de123276a:

- The changed defaulting test now has setup, execution, and assertion story headings. The validation envtest executor adds the missing scenario-assembly heading while retaining its existing helper-level execution and assertion narration, avoiding duplicate logs.
- The changed defaulting loop and provider-selection/materialization blocks now have concise one-line story comments.
- A successful provider patch now continues reconciliation with the API-updated DGD instead of immediately requeueing through a potentially stale informer cache.
- The single-use workload-provider validation helper was folded into the existing metadata validation flow.

The full affected Go package tests, `git diff --check`, and the required ARM64 linter and tester Docker targets pass.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Addressed on the current head. The operation-dependent provider defaulting, old-object decoding, and `minAvailable` restoration were removed. A missing provider is now materialized level-wise and the resulting durable value drives provider-specific defaults on every admission. The remaining CREATE check only stamps the pre-existing operator-origin version and is unrelated to provider reconciliation.

deploy/operator/internal/consts/consts.go
julienmancuso: The graph-level boundary intentionally distinguishes complete workload programs, not the internal kind used by the Component program. `component` covers DCD reconciliation; whether those DCDs use Deployments or LWS remains an internal Component/DCD choice governed by the existing gates and API availability. It therefore does not need another graph-level provider annotation or a similar PR.

deploy/operator/internal/consts/consts.go
julienmancuso: Addressed on the current head: the annotation is now `nvidia.com/workload-provider`; the `selected-` prefix was removed.

deploy/operator/internal/controller/dgd_grove_program.go
julienmancuso: Addressed on the current head. The separate availability/capability concept was removed. The Grove program now checks the normal `features.Grove` gate directly, and a durable Grove selection reports `Ready=False` plus a terminal `selected_workload_provider_unavailable` error when that gate is unavailable rather than falling back.

deploy/operator/internal/controller/dgd_grove_watch_setup.go
julienmancuso: The special deletion-only recovery watch, predicate, and its tests were removed entirely with the legacy workload-adoption/recovery framework, so this PR no longer contains the `onlyDelete` abstraction.

deploy/operator/internal/controller/dynamographdeployment_controller.go
julienmancuso: The PR-specific runtime/API-availability and recovery-watch branches were removed. What remains is the pre-existing Grove watch helper that cohesively owns PodCliqueSet ownership plus PodClique/PodCliqueScalingGroup mapping and predicates; this focused PR only registers that existing helper through the normal Grove gate and does not add another watch-setup layer.

deploy/operator/internal/controller_common/runtime.go
julienmancuso: Addressed on the current head. The PR-added runtime capability field was removed; `RuntimeConfig` is back to the normal resolved feature gates and excluded namespaces. Grove availability is represented only by `features.Grove`.

deploy/operator/internal/features/gates.go
julienmancuso: Addressed on the current head. The special `features.Resolution`/Grove API-availability path was removed. The existing feature-gate resolution remains the sole availability mechanism.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Addressed on the current head. All operation-dependent provider defaulting was removed. Provider materialization and provider-specific defaults are level-based: reuse the durable annotation when present, otherwise compute and materialize it, then apply defaults from that value. The remaining CREATE branch only stamps existing creation provenance.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Addressed on the current head. The defaulter materializes a missing provider before applying provider-specific defaults. The controller also has a safety net: it optimistically patches a missing annotation before any finalizer, status, or workload effect; a failed patch stops reconciliation, while a successful patch continues with the API-updated object.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Addressed on the current head. The provider path no longer branches on admission operation or decodes the old object. It is level-based and derives all provider-specific behavior from the materialized annotation.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
julienmancuso: Addressed on the current head. The separate validation chains and policy-named micro-helper were removed. Supported provider values are checked locally in the existing metadata validation, and immutability is checked locally in the existing metadata-update validation, both within the normal validation flow.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
julienmancuso: Addressed on the current head. The PR-specific termination validation path was removed; this change no longer introduces different provider validation behavior during deletion. The handler is back to the repository's pre-existing deletion bypass for finalizer removal.

deploy/operator/internal/controller/dgd_workload_provider.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx context.Context,
+	dgd *nvidiacomv1beta1.DynamoGraphDeployment,
+) (workloadProvider, error) {
+	// Reuse a materialized provider without consulting mutable routing inputs.
+	if value, exists := dgd.Annotations[consts.KubeAnnotationWorkloadProvider]; exists {
+		return parseWorkloadProvider(value)
+	}
+
+	// Persist the current routing decision with optimistic concurrency control.
+	provider := providerFromCurrentIntent(r.RuntimeConfig.Gate, dgd)
```
sttts: **[P1] Preserve the workload family already owned by legacy DGDs**

This still violates the upgrade contract in #12938. Consider a legacy DGD with no `nvidia.com/workload-provider` that already owns DCDs. If Grove is enabled when this code first runs, we persist `grove`, make that value immutable, and dispatch the Grove program, which can create a PodCliqueSet while the existing DCD family remains. The reverse mismatch is possible as well. Neither this fallback nor the admission defaulter inspects the top-level workloads already controlled by the DGD.

Please distinguish creation-time selection from legacy adoption: DCD-only should select `component`, PodCliqueSet-only should select `grove`, both families should produce a persisted terminal conflict instead of guessing, and only no owned family should fall back to current intent. The adopted value should still be persisted here before finalizer, status, or workload effects. UPDATE admission also must not guess and lock a value before adoption can observe the children. Please add focused upgrade tests for all four states.

deploy/operator/internal/controller/dgd_grove_program.go
```diff
@@ -73,6 +77,17 @@ func (p *groveProgram) Reconcile(
...
 	programResult = newWorkloadProgramResult(req.DGD)
+
+	// Fail a durable Grove selection when Grove is unavailable rather than falling back.
+	if !p.gate.Enabled(features.Grove) {
+		err := failWorkloadProgram(
+			reasonSelectedWorkloadProviderUnavailable,
+			fmt.Errorf("selected workload provider %q is unavailable because Grove is disabled", workloadProviderGrove),
+		)
+		programResult.Fail(req.DGD.Generation, reasonSelectedWorkloadProviderUnavailable, err)
+		return programResult, reconcile.TerminalError(err)
```
sttts, line 88: **[P2] Restore the resolved transient / unclassified / terminal contract**

The earlier thread was resolved based on commit `e95992f162`, which added mutually exclusive `reconcileerrors.IsTransient` and `IsTerminal`, preserved `Ready` for transient API failures, and covered `%w`, exclusivity, and terminal status-write failure. Those helpers and tests have since been deleted. In the current head, this one case is terminal, but the deferred handlers in both the Grove and Component programs again call `programResult.Fail` for every other error. A conflict, timeout, throttling response, or service-unavailable error is therefore persisted as `state=failed` / `Ready=False` even though it is retried.

Please restore the explicit three-way behavior: transient = preserve status and retry; unclassified = persist failure and retry; terminal = persist the specific failure and suppress automatic retry. Terminal classification should be added only at the workload-program or DGD boundary that owns retry policy, and a failed status write must replace a terminal program error with an ordinary retryable error.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
```diff
@@ -520,6 +535,34 @@ func (v *dynamoGraphDeploymentValidation) validateDynamoGraphDeploymentUpdate(
...
+	newObjectMeta *metav1.ObjectMeta,
+	oldObjectMeta *metav1.ObjectMeta,
+	fldPath *field.Path,
+) field.ErrorList {
+	allErrs := field.ErrorList{}
+	annotationsPath := fldPath.Child("annotations")
+	newProvider, newProviderExists := newObjectMeta.Annotations[consts.KubeAnnotationWorkloadProvider]
+	oldProvider, oldProviderExists := oldObjectMeta.Annotations[consts.KubeAnnotationWorkloadProvider]
+
+	// Once materialized, the workload provider cannot be replaced or removed.
```
sttts, line 561: **[P3] This immutability check is unreachable during deletion**

`DynamoGraphDeploymentHandler.ValidateUpdate` still returns success immediately when `newObj.DeletionTimestamp` is set, before it invokes the validator containing this block. Consequently a terminating update can replace or remove `nvidia.com/workload-provider`; an unselected terminating object can also introduce an unsupported value. This exact issue was previously resolved in `637862cbf8` by retaining metadata invariants while skipping complete new-state/spec validation, but that handler behavior and its regression coverage are no longer present in the current head.

Please keep finalizer removal possible while always running the local supported-value and metadata-update checks for terminating objects. Add envtests for provider replacement, provider removal, unsupported insertion, and a successful finalizer-only update.

sttts (commented):
Re-review of the current head, including the previously resolved and outdated threads.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
sttts: even worse, you can cause other changes while deleting, and you can delay that arbitrarily with a finalizer. Not sure our reconciler stop actually acting during deletion. This it would be a little less critical.

deploy/operator/internal/controller/dgd_grove_program.go
sttts: also fine as a follow-up. I think it needs a full analysis to add these qualifiers everywhere in all reconcilers.

sttts (approved):
Approved modulo keeping track of the three findings, either followup or in the PR. Especially the P1 seems risky.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -83,21 +83,25 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
 	// default the controller panics on a nil pointer dereference in
 	// expandRolesForComponent(). Apply on every operation so that components
 	// added via UPDATE also get the default.
```
tmonty12: should we move this comment about defaulting nil replicas to to above line 93? doesn't make sense to have above `defaultWorkloadProvider(...)`

deploy/operator/internal/controller/dgd_grove_program.go
sttts: now I understand that this is a marker from controller-runtime. Then we should use it. IsTransient was rather just a reminder that we have to check existing code. I think we are fine here for now.

sttts: Have created https://github.com/ai-dynamo/dynamo/issues/13200.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
```diff
@@ -83,21 +83,25 @@ func (d *DGDDefaulter) Default(ctx context.Context, obj runtime.Object) error {
...
-	grovePathway := d.isGrovePathway(ctx, dgd)
+	provider := defaultWorkloadProvider(ctx, dgd)
```
tmonty12: Currently a user could set `nvidia.com/workload-provider` annotation to `grove` as an example and the operator will then make this the immutable provider.

This seems confusing with the current provider selection semantics for grove. A user could set:
```
nvidia.com/workload-provider: grove
nvidia.com/enable-grove: false
```
And the provider would become grove despite the explicit disable grove annotation.

Do we intend to make the workload provider annotation something a user can set or strictly operator owned and continue with the current provider selection semantics?

deploy/operator/internal/webhook/validation/dynamographdeployment_helpers.go
```diff
@@ -137,6 +137,27 @@ func grovePathwayForDynamoGraphDeployment(
...
+				provider,
+			)
+		default:
+			return false, fmt.Sprintf(
+				"requires the Grove pathway, but annotation %q has unsupported value %q",
+				consts.KubeAnnotationWorkloadProvider,
+				provider,
+			)
+		}
+	}
```
tmonty12, line 159: The defaulting webhook should always run before validation, in which case we should be sure that the workload provider annotation exists. So I don't think the fallback to the previous `groveEnabled` logic is needed.

deploy/operator/internal/controller/dgd_workload_provider.go
julienmancuso: Addressed on the current head. For an unannotated legacy DGD, `ensureWorkloadProvider` now checks owned DCDs and PodCliqueSets by controller owner reference before consulting current routing intent: DCD-only selects `component`, PodCliqueSet-only selects `grove`, and no owned family uses current intent. Mixed ownership fails closed with a normal retryable conflict and does not persist an arbitrary provider, matching the later simplification away from bespoke terminal recovery. UPDATE defaulting leaves an absent provider untouched so the controller can observe those children first. Focused tests cover DCD-only, PodCliqueSet-only, mixed, no owned workloads, foreign workloads, and a cluster without the Grove API.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Addressed in ca38dd4b1d. Provider resolution now has its own story comment, and the replica-defaulting comment is immediately above the component loop it describes.

deploy/operator/internal/webhook/defaulting/dynamographdeployment_handler.go
julienmancuso: Agreed: `nvidia.com/workload-provider` is operator-owned. Addressed in ca38dd4b1d. On CREATE, the defaulter ignores any submitted provider value and derives/overwrites it from the Grove gate plus `nvidia.com/enable-grove`, so the contradictory example selects `component`. For a legacy UPDATE with no provider, defaulting leaves it absent for controller adoption; when the operator principal is configured, validation rejects an absent-to-present transition from any other user. Once present, the annotation cannot be changed or removed. Tests cover conflicting user input on CREATE, rejected user materialization, allowed operator materialization, and the optional-principal compatibility path.

deploy/operator/internal/webhook/validation/dynamographdeployment_helpers.go
julienmancuso: The fallback remains intentional for legacy unannotated UPDATEs. The mutating webhook runs, but it materializes the provider only on CREATE; on UPDATE an absent value stays absent until the controller can inspect owned DCDs and PodCliqueSets and adopt the existing workload family. During that window, admission still needs the existing Grove gate plus `nvidia.com/enable-grove` semantics. Once the controller persists `nvidia.com/workload-provider`, the first branch is authoritative and the fallback is no longer used. Regression coverage verifies that a legacy UPDATE remains unselected before controller adoption.

deploy/operator/internal/webhook/validation/dynamographdeployment.go
julienmancuso: Agreed that this exposes a broader pre-existing deletion-validation gap: the current early return permits more than provider metadata changes while a finalizer keeps the object terminating. I intentionally did not reintroduce a provider-only termination validation chain in this focused PR, because that would recreate the brittle split discussed earlier. This should be handled cohesively as a follow-up for terminating UPDATE validation, preserving finalizer-only updates while enforcing the appropriate object invariants. No change for this thread is included in ca38dd4b1d, and I am leaving it open.

tmonty12 (approved):

panpan0000 (approved):
/lgtm


non-issue: when if user updates`nvidia.com/enable-grove`  annotation — it would help to warn (or reject changes) since the selection is immutable in `validateObjectMetaUpdate()`

PR #12943, opened 2026-08-10
feat(operator): start a Ray head on a single-pod elastic EP component

tzulingk:
## Overview:

**What we want.** A Dynamo deployment where one pod runs the vLLM leader and serves on its own, and where extra pods can join later to widen the expert-parallel group. Elastic EP is the vLLM feature that makes that group resizable while the engine is running. It works by putting every participating pod into one Ray cluster, with the leader running the Ray head.

**What is broken today.** Deploy a component as a single pod, pass `--enable-elastic-ep`, and nothing happens. No Ray head starts, so there is no cluster for anything to join, and the flag is silently ignored. Elastic EP works only if the component is declared multinode.

**Why.** Two independent reasons, either of which alone is enough:

1. The elastic-EP dispatch sits inside the `numberOfNodes > 1` branch of `UpdateContainer`, so a one-pod component never reaches it.
2. Even if it did, the leader wiring in `injectElasticEPRayLaunchFlags` matches only `RoleLeader` — and a one-pod component is expanded as `RoleMain`, so it would fall straight through.

**What this changes.** Add a single-pod branch beside the multinode one, and let the leader arm match `RoleMain` as well as `RoleLeader`. A solo leader then starts `ray start --head` exactly as a multinode leader does, and is told where to find it via `VLLM_DP_MASTER_IP`.

This is worth having on its own, since single-pod elastic EP is simply broken today. It is also the first step toward a leader that serves alone with followers attaching at an arbitrary later time.

### The wiring

`UpdateContainer` in `deploy/operator/internal/dynamo/backend_vllm.go` gains an `else if` branch for single-pod components, and `injectElasticEPRayLaunchFlags` now treats `RoleMain` identically to `RoleLeader`.

Probes are deliberately left in place. The multinode path strips them from workers, but this pod is a leader that serves traffic, so it keeps them.

### Why the branch also requires the Ray backend

The new branch fires only when the args carry **both** `--enable-elastic-ep` and `--data-parallel-backend ray`, expressed as a named predicate `isElasticEPRayLaunch`.

The second condition is load-bearing, not defensive. Elastic EP only functions on the Ray data-parallel backend — vLLM asserts it outright in `scale_elastic_ep` with *"Only ray DP backend supports scaling elastic EP"*. On any other backend, a Ray head is a process nothing ever connects to, sitting behind a 300-second startup gate.

That is not hypothetical. Three single-pod templates in this repo pass `--enable-elastic-ep`:

| Template | Ray DP backend | Without the extra condition |
|---|---|---|
| `moe_elastic_ep_demo.yaml` | yes | Correct — gains the Ray head it needs |
| `moe_agg.yaml` | no | Would gain a Ray head nothing uses |
| `moe_disagg.yaml` | no | Same, on both of its components |

The latter two are not misconfigured. `--enable-elastic-ep` still does real work for them: it requires EPLB, rejects pipeline parallelism, and selects the `pynccl` EPLB communicator. They simply cannot scale at runtime. Without the backend condition, this PR would have added a Ray head and a startup gate to two configurations that work today.

### Telling vLLM where the DP master is

Starting the Ray head is necessary but not sufficient, and the second half is not obvious.

Ray registers the head under the pod IP, since the whole point is that another pod reaches it later. vLLM derives the same address for its data-parallel master — `arg_utils.py` calls `get_ip()` for the Ray backend and logs *"Using host IP %s as ray-based data parallel address"*. So far so good.

Then `ParallelConfig.__post_init__` throws it away. Its condition is `data_parallel_size > 1 or data_parallel_size_local == 0`, and at **`--data-parallel-size 1`** neither holds, so it takes the offline-SPMD branch and overwrites the derived address with `envs.VLLM_DP_MASTER_IP` — default `127.0.0.1`. Engine startup then asserts on a Ray node that does not exist:

```
The DP master node (ip: 127.0.0.1) is missing or dead
```

The root cause upstream is a conflated signal: `data_parallel_size` defaults to 1, so vLLM cannot distinguish "DP not configured" from "DP configured with one rank". A solo elastic-EP leader is exactly the second, read as the first. Reported as [vllm-project/vllm#51712](https://github.com/vllm-project/vllm/issues/51712); if it is fixed upstream this injection stops being load-bearing, but it is correct to carry it either way.

Two things that look like they should fix it, and do not. `VLLM_HOST_IP` only steers `get_ip()`, whose result is then discarded. `--data-parallel-address` is discarded by the same branch. `VLLM_DP_MASTER_IP` is the one value the fallback actually reads, so the operator binds it to `status.podIP` behind the same predicate that starts the Ray head. This mirrors the existing `VLLM_NIXL_SIDE_CHANNEL_HOST` injection a few lines above.

Note this only bites at dp=1. `moe_elastic_ep_demo.yaml` runs dp=2 in one pod and takes the other branch, which is why single-pod elastic EP appeared to work in that shape.

### Naming the head's address, not guessing it

Telling vLLM the master is at `status.podIP` is only half of a pair. vLLM then looks up the Ray node registered under that exact address, and Ray, left alone, picks an interface by its own heuristic. The two agree only as long as that heuristic lands on the pod IP — true on a single-interface pod, not guaranteed on one with a second network, and a mismatch there produces the identical "missing or dead" abort.

Before this change nothing was pinned at either end, so vLLM's `get_ip()` and Ray's detection ran the same heuristic and agreed with each other whatever they chose. Pinning one end alone would trade a self-consistent pair for a mismatched one.

So the head is given its address explicitly: `POD_IP` is bound from the downward API and interpolated into `--node-ip-address`. Both values resolve from `status.podIP`, which is the point — they cannot drift. This also makes the generated command match the hand-written leader command already validated on a GB200 cluster.

The flag is scoped to `RoleMain`. A multinode leader gets no `VLLM_DP_MASTER_IP`, so adding `--node-ip-address` there would interpolate an unset variable and start the head with an empty address.

### What is deliberately left alone

The multinode branch keys on the elastic-EP flag alone and has the same looseness. It stays as it is: every multinode elastic-EP template in the tree already sets the Ray backend, so tightening it would change nothing here while altering behaviour for anyone outside the tree. Worth doing, but as its own change.

### Tests

Six cases in `backend_vllm_test.go`, all confirmed to fail without the corresponding production change:

- a single-pod elastic-EP component gets the Ray head pinned to `$POD_IP`, and keeps its probes;
- `RoleMain` takes the leader arm;
- four regression guards — no elastic-EP flag, the flag with no backend specified, and the flag on an explicitly non-Ray backend — each asserting the container is left untouched.

Both env bindings are asserted on **every** case in the table rather than only the positive ones: present and bound to `status.podIP` where a Ray head is started, absent everywhere else, including the whole multinode set. That makes it impossible for either to leak into a container that did not ask for a head, and the multinode expectations double as the guard that their rendered command is unchanged.

`backend_vllm.go`: the new `else if` in `UpdateContainer`, and the `isElasticEPRayLaunch` predicate. The judgement call worth pressing on is whether requiring the Ray backend is right, or whether this should mirror the looser multinode branch instead.

Then `TestVLLMBackend_UpdateContainer/single_node_elastic_EP_without_the_ray_backend_is_left_alone`, which is the test that keeps `moe_agg.yaml` and `moe_disagg.yaml` rendering exactly as they do today.

The `VLLM_DP_MASTER_IP` binding is the other place to look, and the question there is scope: it is gated on the same predicate as the Ray head, which deliberately leaves the multinode path alone even though the same vLLM behaviour would apply to a multinode leader at dp=1.

## Related Issues

### DIFF
--- a/deploy/operator/internal/dynamo/backend_vllm.go
+++ b/deploy/operator/internal/dynamo/backend_vllm.go
@@ -24,7 +24,10 @@ const (
 	distributedExecutorFlag   = "--distributed-executor-backend"
 	enableElasticEPFlag       = "--enable-elastic-ep"
 	dataParallelBackendFlag   = "--data-parallel-backend"
-	dataParallelBackendRay    = "ray"
+	// dataParallelBackendShortFlag is vLLM's documented short alias for
+	// --data-parallel-backend (see the v0.26.0 `vllm serve` CLI reference).
+	dataParallelBackendShortFlag = "-dpb"
+	dataParallelBackendRay       = "ray"
 )
 
 type VLLMBackend struct {
@@ -73,7 +76,7 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
 			container.ReadinessProbe = nil
 			container.StartupProbe = nil
 		}
-	} else if role == RoleMain && isElasticEPRayLaunch(getExpandedArgs(container)) {
+	} else if role == RoleMain && isElasticEPRayLaunch(container) {
 		// A single-pod elastic-EP component still needs a Ray head, so that
 		// follower pods created later have a cluster to join. Only the leader
 		// arm applies here: a lone pod is expanded as RoleMain, never RoleWorker.
@@ -448,17 +451,27 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 		for i, arg := range container.Args {
 			quotedArgs[i] = shellQuoteForBashC(arg)
 		}
+		vllmCommand := strings.TrimSpace(strings.Join(quotedCmd, " ") + " " + strings.Join(quotedArgs, " "))
+		// A single-pod RoleMain leader is an ordinary serving pod that Kubernetes
+		// rolls, evicts, and deletes, so exec the engine: it then runs as the
+		// container's main process (PID 1) and receives SIGTERM directly for a
+		// graceful shutdown, instead of being killed after the grace period with
+		// in-flight requests dropped. The backgrounded Ray head continues as its
+		// child. The multinode RoleLeader keeps its historical no-exec form so
+		// this stays scoped to the new single-pod path.
+		if role == RoleMain {
+			vllmCommand = "exec " + vllmCommand
+		}
 		// Poll Ray head readiness with a bounded retry loop (150 × 2 s = 5 min max).
 		// An unbounded `until` loop would spin forever if `ray start --head` crashes
 		// silently or the port never opens.
 		container.Args = []string{fmt.Sprintf(
 			`ray start --head --port=%s --block & `+
 				`i=0; until python3 -c "import socket; s=socket.create_connection(('127.0.0.1',%s),timeout=1); s.close()" 2>/dev/null; `+
-				`do i=$((i+1)); [ "$i" -ge 150 ] && { echo "ERROR: Ray head did not start within 300s" >&2; exit 1; }; sleep 2; done && %s %s`,
+				`do i=$((i+1)); [ "$i" -ge 150 ] && { echo "ERROR: Ray head did not start within 300s" >&2; exit 1; }; sleep 2; done && %s`,
 			VLLMPort,
 			VLLMPort,
-			strings.Join(quotedCmd, " "),
-			strings.Join(quotedArgs, " "),
+			vllmCommand,
 		)}
 	case RoleWorker:
 		leaderHostname := multinodeDeployer.GetLeaderHostname(serviceName)
@@ -489,21 +502,39 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 	container.Command = []string{"/bin/sh", "-c"}
 }
 
-// isElasticEPRayLaunch reports whether args ask for the elastic-EP Ray topology.
+// isElasticEPRayLaunch reports whether the container asks for the elastic-EP Ray
+// topology.
 //
 // Elastic EP only works on the Ray data-parallel backend: vLLM's Ray executor is
 // what grows and shrinks workers at runtime, and the engine refuses a scale
 // request on any other backend. Requiring both flags keeps a Ray head off pods
 // that pass --enable-elastic-ep while running the default backend, where it
 // would launch a process nothing ever talks to.
-func isElasticEPRayLaunch(expandedArgs []string) bool {
-	// hasArg matches both the "--data-parallel-backend ray" and
-	// "--data-parallel-backend=ray" spellings. vLLM (argparse) treats them as
-	// equivalent, so the equals form must trigger Ray-head injection too;
-	// otherwise a single-pod deployment written that way silently starts
-	// without a Ray head and followers can never join.
-	return hasFlag(expandedArgs, enableElasticEPFlag) &&
-		hasArg(expandedArgs, dataParallelBackendFlag, dataParallelBackendRay)
+//
+// Detection scans the full command line (Command + Args) so the flags are found
+// whether the manifest carries them in Command or Args, and it accepts vLLM's
+// long --data-parallel-backend flag and its documented -dpb alias in both the
+// "flag value" and "flag=value" spellings — vLLM's argparse treats all of these
+// as equivalent, so any of them must trigger Ray-head injection.
+func isElasticEPRayLaunch(container *corev1.Container) bool {
+	expanded := getExpandedCommandLine(container)
+	return hasFlag(expanded, enableElasticEPFlag) &&
+		(hasArg(expanded, dataParallelBackendFlag, dataParallelBackendRay) ||
+			hasArg(expanded, dataParallelBackendShortFlag, dataParallelBackendRay))
+}
+
+// getExpandedCommandLine flattens Command and Args and splits any space-joined
+// tokens, so flag detection works whether the manifest puts flags in Command or
+// Args and whether they are separate list items or a single combined string.
+func getExpandedCommandLine(container *corev1.Container) []string {
+	commandLine := make([]string, 0, len(container.Command)+len(container.Args))
+	commandLine = append(commandLine, container.Command...)
+	commandLine = append(commandLine, container.Args...)
+	expanded := make([]string, 0, len(commandLine))
+	for _, arg := range commandLine {
+		expanded = append(expanded, strings.Fields(arg)...)
+	}
+	return expanded
 }
 
 // hasFlag returns true if flag exists in expandedArgs.
--- a/deploy/operator/internal/dynamo/backend_vllm.go
+++ b/deploy/operator/internal/dynamo/backend_vllm.go
@@ -76,7 +76,7 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
 			container.ReadinessProbe = nil
 			container.StartupProbe = nil
 		}
-	} else if role == RoleMain && isElasticEPRayLaunch(container) {
+	} else if role == RoleMain && IsElasticEPRayLaunch(container) {
 		// A single-pod elastic-EP component still needs a Ray head, so that
 		// follower pods created later have a cluster to join. Only the leader
 		// arm applies here: a lone pod is expanded as RoleMain, never RoleWorker.
@@ -546,7 +546,7 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 	return true
 }
 
-// isElasticEPRayLaunch reports whether the container asks for the elastic-EP Ray
+// IsElasticEPRayLaunch reports whether the container asks for the elastic-EP Ray
 // topology.
 //
 // Elastic EP only works on the Ray data-parallel backend: vLLM's Ray executor is
@@ -560,7 +560,7 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 // long --data-parallel-backend flag and its documented -dpb alias in both the
 // "flag value" and "flag=value" spellings — vLLM's argparse treats all of these
 // as equivalent, so any of them must trigger Ray-head injection.
-func isElasticEPRayLaunch(container *corev1.Container) bool {
+func IsElasticEPRayLaunch(container *corev1.Container) bool {
 	expanded := getExpandedCommandLine(container)
 	return hasFlag(expanded, enableElasticEPFlag) &&
 		(hasArg(expanded, dataParallelBackendFlag, dataParallelBackendRay) ||
--- a/deploy/operator/internal/dynamo/backend_vllm.go
+++ b/deploy/operator/internal/dynamo/backend_vllm.go
@@ -481,18 +481,13 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 // that pass --enable-elastic-ep while running the default backend, where it
 // would launch a process nothing ever talks to.
 func isElasticEPRayLaunch(expandedArgs []string) bool {
+	// hasArg matches both the "--data-parallel-backend ray" and
+	// "--data-parallel-backend=ray" spellings. vLLM (argparse) treats them as
+	// equivalent, so the equals form must trigger Ray-head injection too;
+	// otherwise a single-pod deployment written that way silently starts
+	// without a Ray head and followers can never join.
 	return hasFlag(expandedArgs, enableElasticEPFlag) &&
-		hasFlagValue(expandedArgs, dataParallelBackendFlag, dataParallelBackendRay)
-}
-
-// hasFlagValue returns true if flag appears in expandedArgs followed by value.
-func hasFlagValue(expandedArgs []string, flag, value string) bool {
-	for i, arg := range expandedArgs {
-		if arg == flag && i+1 < len(expandedArgs) && expandedArgs[i+1] == value {
-			return true
-		}
-	}
-	return false
+		hasArg(expandedArgs, dataParallelBackendFlag, dataParallelBackendRay)
 }
 
 // hasFlag returns true if flag exists in expandedArgs.
--- a/deploy/operator/internal/dynamo/backend_vllm.go
+++ b/deploy/operator/internal/dynamo/backend_vllm.go
@@ -80,23 +80,25 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
 		// A single-pod elastic-EP component still needs a Ray head, so that
 		// follower pods created later have a cluster to join. Only the leader
 		// arm applies here: a lone pod is expanded as RoleMain, never RoleWorker.
-		injectElasticEPRayLaunchFlags(container, role, serviceName, multinodeDeployer)
-
-		// At --data-parallel-size 1, vLLM discards the DP master IP it derives
-		// from the Ray node and reads VLLM_DP_MASTER_IP instead, which defaults
-		// to 127.0.0.1. Ray registers the head under the pod IP, so the engine
-		// then looks for a node that does not exist and aborts with "The DP
-		// master node (ip: 127.0.0.1) is missing or dead". Neither VLLM_HOST_IP
-		// nor --data-parallel-address survives that overwrite; this env var is
-		// the only value the fallback reads.
-		container.Env = append(container.Env, corev1.EnvVar{
-			Name: commonconsts.VLLMDPMasterIPEnvVar,
-			ValueFrom: &corev1.EnvVarSource{
-				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: "status.podIP",
+		if injectElasticEPRayLaunchFlags(container, role, serviceName, multinodeDeployer) {
+			// Bind VLLM_DP_MASTER_IP only when a Ray head was actually injected.
+			//
+			// At --data-parallel-size 1, vLLM discards the DP master IP it derives
+			// from the Ray node and reads VLLM_DP_MASTER_IP instead, which defaults
+			// to 127.0.0.1. Ray registers the head under the pod IP, so the engine
+			// then looks for a node that does not exist and aborts with "The DP
+			// master node (ip: 127.0.0.1) is missing or dead". Neither VLLM_HOST_IP
+			// nor --data-parallel-address survives that overwrite; this env var is
+			// the only value the fallback reads.
+			container.Env = append(container.Env, corev1.EnvVar{
+				Name: commonconsts.VLLMDPMasterIPEnvVar,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "status.podIP",
+					},
 				},
-			},
-		})
+			})
+		}
 	}
 
 	// Set compilation cache environment variables for VLLM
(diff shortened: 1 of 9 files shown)


deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -464,6 +473,28 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
...
+
+// hasFlagValue returns true if flag appears in expandedArgs followed by value.
+func hasFlagValue(expandedArgs []string, flag, value string) bool {
+	for i, arg := range expandedArgs {
+		if arg == flag && i+1 < len(expandedArgs) && expandedArgs[i+1] == value {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
tzulingk: Fixed in 73535a19bd.

`isElasticEPRayLaunch` now uses the existing `hasArg` helper, which matches both `--data-parallel-backend ray` and `--data-parallel-backend=ray` — the two spellings vLLM's argparse treats as equivalent. The redundant `hasFlagValue` is removed, and I added a regression test (`single node elastic EP gets a ray head with the inline backend flag`) that drives the equals form through `UpdateContainer` and asserts the Ray head is injected. Thanks for catching this.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -464,6 +473,28 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
...
+// isElasticEPRayLaunch reports whether args ask for the elastic-EP Ray topology.
+//
+// Elastic EP only works on the Ray data-parallel backend: vLLM's Ray executor is
+// what grows and shrinks workers at runtime, and the engine refuses a scale
+// request on any other backend. Requiring both flags keeps a Ray head off pods
+// that pass --enable-elastic-ep while running the default backend, where it
+// would launch a process nothing ever talks to.
+func isElasticEPRayLaunch(expandedArgs []string) bool {
+	return hasFlag(expandedArgs, enableElasticEPFlag) &&
+		hasFlagValue(expandedArgs, dataParallelBackendFlag, dataParallelBackendRay)
```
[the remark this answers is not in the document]
tzulingk: Fixed in 73535a19bd.

`isElasticEPRayLaunch` now uses the existing `hasArg` helper, which matches both `--data-parallel-backend ray` and `--data-parallel-backend=ray` — the two spellings vLLM's argparse treats as equivalent. The redundant `hasFlagValue` is removed, and I added a regression test (`single node elastic EP gets a ray head with the inline backend flag`) that drives the equals form through `UpdateContainer` and asserts the Ray head is injected. Thanks for catching this.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -464,6 +473,28 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
...
+}
+
+// hasFlagValue returns true if flag appears in expandedArgs followed by value.
+func hasFlagValue(expandedArgs []string, flag, value string) bool {
+	for i, arg := range expandedArgs {
+		if arg == flag && i+1 < len(expandedArgs) && expandedArgs[i+1] == value {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
tzulingk: Fixed in 73535a19bd.

`isElasticEPRayLaunch` now uses the existing `hasArg` helper, which matches both `--data-parallel-backend ray` and `--data-parallel-backend=ray` — the two spellings vLLM's argparse treats as equivalent. The redundant `hasFlagValue` is removed, and I added a regression test (`single node elastic EP gets a ray head with the inline backend flag`) that drives the equals form through `UpdateContainer` and asserts the Ray head is injected. Thanks for catching this.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -410,11 +417,13 @@ func injectRayDistributedLaunchFlags(container *corev1.Container, role Role, ser
...
 //
-// Leader: ray start --head --port=6379 --block & <tcp-poll-ray-ready 150×2s> && <vllm cmd>
+// Leader (or a single-pod RoleMain): ray start --head --port=6379 --block & <tcp-poll-ray-ready 150×2s> && <vllm cmd>
 // Worker: <poll /live HTTP until 200> && ray start --address=<leader>:6379 --block
 func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, serviceName string, multinodeDeployer MultinodeDeployer) {
 	switch role {
-	case RoleLeader:
+	// RoleMain is a component deployed as a single pod; it heads the Ray
+	// cluster exactly as a multi-node leader does.
+	case RoleLeader, RoleMain:
```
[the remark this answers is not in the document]
tzulingk: Good catch, and confirmed reachable: with the LWS gate off and a multinode component, the controller falls through to `reconcileDeploymentResources` (`dynamocomponentdeployment_controller.go:219-222`), which renders `RoleMain` at `:818`.

The change there is strictly an improvement, not a regression. Before this PR, that combination matched neither arm of the `injectElasticEPRayLaunchFlags` switch yet still ran the unconditional `container.Command = ["/bin/sh","-c"]` — leaving `Args` as a raw token list, i.e. `/bin/sh -c python3 -m dynamo.vllm ...`, which executes only `python3` and drops the rest. A hard, immediate failure. After this PR the same pod gets a coherent single-leader Ray launch.

A genuinely multinode elastic-EP deployment needs cross-node followers, which only the LWS/Grove leader-worker rendering can express; the plain-Deployment + `RoleMain` fallback can't create workers at all, so it correctly degrades to a single-node leader rather than staying broken. That fallback isn't a supported multinode elastic-EP configuration, so I'm intentionally leaving its scope untouched here. The supported paths — single-pod (`RoleMain`, nodeCount 1) and multinode (`RoleLeader`/`RoleWorker`) — are the ones covered by tests.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -71,6 +73,11 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
 			container.ReadinessProbe = nil
 			container.StartupProbe = nil
 		}
+	} else if role == RoleMain && isElasticEPRayLaunch(getExpandedArgs(container)) {
+		// A single-pod elastic-EP component still needs a Ray head, so that
+		// follower pods created later have a cluster to join. Only the leader
+		// arm applies here: a lone pod is expanded as RoleMain, never RoleWorker.
+		injectElasticEPRayLaunchFlags(container, role, serviceName, multinodeDeployer)
```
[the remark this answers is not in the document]
tzulingk: Both are conscious decisions rather than accidents — thanks for flagging them explicitly.

**Shell child vs. `exec` / SIGTERM.** The leader has to run `ray start --head --block &` in the background and vLLM in the foreground, so the container command is inherently a shell wrapper. This matches the existing Ray-leader path (`injectRayDistributedLaunchFlags`), which also sets `/bin/sh -c` without `exec`. The single-pod head is the root of an elastic-EP cluster that followers join, so it is not a routinely-drained stateless serving replica — a head rollover is disruptive to the whole EP group by design. Forwarding SIGTERM to vLLM would mean `exec`-ing the engine across the shared leader arm (single-pod and multinode leader alike); that is a sensible cluster-wide follow-up, but I don't want to fold a behavior change to the already-shipped multinode leader path into this PR.

**Probes retained + up-to-300s Ray-head gate.** Retaining probes on the serving leader is deliberate (per the PR description): a single-pod leader serves traffic and must stay probed, unlike multinode workers whose probes are stripped. The 300s is the Ray-head readiness *ceiling*, not the expected duration. Deployments should pair this with a startup probe (the elastic-EP recipes do), so liveness doesn't fire while the head comes up and the model loads. A config relying on a bare short-`initialDelaySeconds` liveness with no startup probe would already crash-loop on any slow model load today — this change doesn't introduce that failure mode.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -464,6 +489,23 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
...
+// request on any other backend. Requiring both flags keeps a Ray head off pods
+// that pass --enable-elastic-ep while running the default backend, where it
+// would launch a process nothing ever talks to.
+func isElasticEPRayLaunch(expandedArgs []string) bool {
+	// hasArg matches both the "--data-parallel-backend ray" and
+	// "--data-parallel-backend=ray" spellings. vLLM (argparse) treats them as
+	// equivalent, so the equals form must trigger Ray-head injection too;
+	// otherwise a single-pod deployment written that way silently starts
+	// without a Ray head and followers can never join.
+	return hasFlag(expandedArgs, enableElasticEPFlag) &&
```
julienmancuso: **[P1] Recognize vLLM's documented `-dpb` alias.**

Dynamo pins vLLM 0.26.0, whose CLI accepts both `--data-parallel-backend` and `-dpb`. With valid args such as `-dpb ray --enable-elastic-ep`, this predicate returns false, so the pod still starts without the Ray head or `VLLM_DP_MASTER_IP`. Please match the short alias as well and add regression coverage for its split (and preferably equals) form. Reference: https://docs.vllm.ai/en/v0.26.0/cli/serve/

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -71,6 +73,27 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
 			container.ReadinessProbe = nil
 			container.StartupProbe = nil
 		}
+	} else if role == RoleMain && isElasticEPRayLaunch(getExpandedArgs(container)) {
```
julienmancuso: **[P2] Inspect the complete container command line.**

`getExpandedArgs` only expands `container.Args`, so a valid Kubernetes command such as `Command: ["python3", "-m", "dynamo.vllm", "--data-parallel-backend", "ray", "--enable-elastic-ep"]` bypasses this branch. The package already has `containerCommandLineHasArg`, which examines both `Command` and `Args`; please use equivalent full-command-line detection and add a regression case.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -71,6 +73,27 @@ func (b *VLLMBackend) UpdateContainer(container *corev1.Container, numberOfNodes
...
+		// A single-pod elastic-EP component still needs a Ray head, so that
+		// follower pods created later have a cluster to join. Only the leader
+		// arm applies here: a lone pod is expanded as RoleMain, never RoleWorker.
+		injectElasticEPRayLaunchFlags(container, role, serviceName, multinodeDeployer)
```
julienmancuso: **[P2] Preserve graceful termination for the new single-pod path.**

This call rewrites an ordinary serving pod to `/bin/sh -c`, and the generated script launches vLLM without `exec` or a forwarding trap. Kubernetes sends SIGTERM to the container's main process, so vLLM may never receive it during rollout, eviction, or deletion and can be killed after the grace period with in-flight requests dropped. Please `exec` the foreground vLLM command (or explicitly forward signals to the managed processes); this can be scoped to `RoleMain` if changing the existing multinode leader behavior is undesirable.

julienmancuso (changes requested):
Requesting changes for three correctness and lifecycle issues in the new single-pod elastic-EP path. The Ray-backend gate and `VLLM_DP_MASTER_IP=status.podIP` injection otherwise look appropriately scoped, and `go test ./internal/dynamo -count=1` plus `git diff --check` passed locally.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Done in 05dac7ffbe. Added `-dpb` as a recognized alias: `isElasticEPRayLaunch` now checks `hasArg(expanded, dataParallelBackendFlag, ray) || hasArg(expanded, dataParallelBackendShortFlag, ray)`, so `-dpb ray` and `-dpb=ray` both start the Ray head and bind `VLLM_DP_MASTER_IP`. Added regression cases for both the split and the equals spelling. Thanks for the v0.26.0 CLI reference.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Done in 05dac7ffbe. `isElasticEPRayLaunch` now takes the container and scans the flattened full command line — a new `getExpandedCommandLine` helper concatenates `Command` + `Args` and splits each on whitespace — so a manifest that carries the flags in `Command` (e.g. `Command: ["python3","-m","dynamo.vllm","--data-parallel-backend","ray","--enable-elastic-ep"]`, empty `Args`) is now detected and gets both the Ray head and `VLLM_DP_MASTER_IP`. Added a regression case for flags placed in `Command`.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Done in 05dac7ffbe, scoped to `RoleMain` as you suggested. The single-pod leader arm now `exec`s the foreground engine (`… done && exec python3 -m dynamo.vllm …`), so vLLM becomes the container's main process and receives SIGTERM directly for a graceful shutdown during rollout/eviction/deletion; the backgrounded `ray start --head --block` continues as its child. The multinode `RoleLeader` keeps its historical no-exec form, so this change stays confined to the new single-pod path. Updated the `RoleMain` test expectations to assert the `exec` prefix.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -423,17 +451,27 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 		for i, arg := range container.Args {
 			quotedArgs[i] = shellQuoteForBashC(arg)
 		}
+		vllmCommand := strings.TrimSpace(strings.Join(quotedCmd, " ") + " " + strings.Join(quotedArgs, " "))
```
julienmancuso, line 484: **[P1] Preserve the image ENTRYPOINT when `Command` is omitted.**

A valid single-pod spec may leave `Command` empty and put only vLLM flags in `Args`; Kubernetes then runs the image's ENTRYPOINT with those arguments (the checked-in CRD explicitly documents this), and only LWS pods currently validate that `Command` is non-empty. `isElasticEPRayLaunch` detects the flags in that spec, but this reconstruction has no executable, so it emits `... && exec --model ...` and the pod fails instead of invoking the image ENTRYPOINT. Please preserve this supported invocation shape (or add an explicit compatible validation/migration path) and add a `Command: nil, Args: [...]` regression case. Schema reference: https://github.com/ai-dynamo/dynamo/blob/05dac7ffbe0f5a2752920dc596f65f90b3b1c6cb/deploy/operator/config/crd/bases/nvidia.com_dynamocomponentdeployments.yaml#L6595-L6604

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -423,17 +451,27 @@ func injectElasticEPRayLaunchFlags(container *corev1.Container, role Role, servi
 		for i, arg := range container.Args {
 			quotedArgs[i] = shellQuoteForBashC(arg)
```
julienmancuso: **[P2] Serialize argv with a shell-safe, argv-preserving quote routine.**

Applying this shell wrapper to `RoleMain` newly sends every previously direct Kubernetes argv token through `shellQuoteForBashC`, but that helper is not argv-preserving: for example, `it's` becomes `"it'"'"'s"`, which `/bin/sh -c` rejects with an unmatched-quote error; empty tokens disappear; and metacharacters such as `;`, `|`, and `&` are not quoted. A valid argument such as a `--chat-template` containing single quotes can therefore make a formerly working single-pod deployment fail at startup. Please use a proven POSIX token-quoting routine (including empty strings) and add an execution-level regression test, rather than only asserting the generated string.

julienmancuso (changes requested):
Re-reviewed the updated head. The three previously requested fixes are present and their new regression cases pass locally. Two backward-compatibility problems remain in the new single-pod RoleMain shell wrapper: it cannot preserve an omitted image ENTRYPOINT, and its argv serialization corrupts valid argument values. `go test ./internal/dynamo -count=1` and `git diff --check` otherwise pass.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Fixed in b4862ef456. `injectElasticEPRayLaunchFlags` now returns whether it actually rewrote the container. When `Command` is empty it leaves the image ENTRYPOINT + `Args` invocation untouched (no more `exec --model …` with no executable), logs a warning that a single-pod Ray head needs an explicit `Command`, and the caller skips `VLLM_DP_MASTER_IP` since no head was started. Added a `Command: nil, Args: [...]` regression case asserting the container is left unmodified and the env var is absent.

I chose to **preserve the supported invocation shape** rather than hard-fail — the pod still starts from its ENTRYPOINT. The tradeoff is that elastic scaling needs an explicit `Command`, because the operator can't see the image ENTRYPOINT to wrap a Ray head around it. Happy to convert this into an explicit validation/migration path instead if you'd rather make that requirement hard.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Fixed in b4862ef456. Added `shellQuotePOSIX`, an argv-preserving quoter: shell-neutral tokens pass through unquoted for readability, and everything else — embedded single quotes, spaces, newlines, `;`/`|`/`&`, globs, and the empty string — is wrapped in single quotes using the `'\''` idiom, so each token survives `sh -c` as exactly one argv element. The elastic-EP leader arm now uses it instead of `shellQuoteForBashC`.

Added an execution-level test (`TestShellQuotePOSIX_ArgvRoundTrip`) rather than only asserting the generated string: it quotes a batch of tricky tokens (including a `--chat-template` with single quotes and an empty token), re-parses them through a real `/bin/sh` via `set --`, and asserts the round-tripped argv is byte-for-byte identical. Left `shellQuoteForBashC` and the other injection paths untouched to keep the change scoped to this path.

deploy/operator/internal/dynamo/backend_vllm.go
```diff
@@ -410,30 +438,61 @@ func injectRayDistributedLaunchFlags(container *corev1.Container, role Role, ser
...
+	// cluster exactly as a multi-node leader does.
+	case RoleLeader, RoleMain:
+		// The Ray-head wrapper has to run a concrete executable once the head is
+		// up, but an empty Command means the real entrypoint is the image
+		// ENTRYPOINT, which the operator cannot see or reconstruct. Rewriting here
+		// would emit a shell command with no executable (e.g. `exec --model ...`)
+		// and break a pod that Kubernetes would otherwise start from its
+		// ENTRYPOINT. Leave that invocation intact and skip the Ray head; a
+		// single-pod Ray head needs an explicit Command.
+		if len(container.Command) == 0 {
```
julienmancuso, line 469: **[P1] Do not silently accept an elastic-EP request that cannot be fulfilled.**

This guard prevents the `exec --model ...` startup regression, but a CRD-supported container with an omitted `Command` and both elastic-EP/Ray flags now starts without the Ray head or `VLLM_DP_MASTER_IP`; elastic scaling therefore remains broken for that valid invocation shape. The only diagnostic is an operator `Info` log, which is not a user-visible warning, event, admission response, or workload condition, so the configuration still appears accepted while the feature requested by its flags is silently unavailable. Please either support the image-ENTRYPOINT path or make an explicit `Command` a documented and user-visible prerequisite (for example via admission warning/validation or a workload condition/event), with regression coverage for that diagnostic.

julienmancuso (commented):
Posting the remaining P1 from the read-only re-review.

deploy/operator/internal/dynamo/backend_vllm.go
tzulingk: Fixed in 44fe8aeae0 — replaced the silent `Info` log with a **user-visible admission rejection**.

New rule `validateElasticEPRequiresCommand`: at DGD **and** standalone DCD admission, a vLLM component that requests elastic EP (`--enable-elastic-ep` with `--data-parallel-backend ray`, including the `-dpb` alias and the `flag=value` spellings) but omits the main container `Command` is rejected with an actionable `field.Required` error on `…podTemplate.spec.containers[<i>].command`:

> elastic expert parallelism (--enable-elastic-ep with --data-parallel-backend ray) requires an explicit container command; the operator starts the single-pod Ray head by wrapping that command and cannot start it from the image ENTRYPOINT

It's wired right next to the existing vLLM-only inter-pod GMS check (the DGD component loop and the DCD spec validator). I exported `IsElasticEPRayLaunch` so the webhook reuses the **exact same** flag detection as the runtime injection — no drift between what admits and what the operator would actually wrap. The runtime empty-`Command` guard stays as defense-in-depth for the case where admission is bypassed (webhook not installed / `failurePolicy: Ignore`).

Regression coverage: a table test for the rule (vLLM+ray+empty `Command` → rejected; `-dpb=ray` alias → rejected; explicit `Command` → accepted; flags-in-`Command` → accepted; non-vLLM backend → skipped; non-ray backend → skipped; no elastic-EP → accepted; nil podTemplate → ignored) plus an end-to-end test that drives the DGD validator and asserts the config is rejected at admission.

(Local-env note: the envtest integration suite needs the Kubernetes test binaries, which aren't available in my sandbox; the in-package unit tests pass and CI's `operator` job exercises the envtest path.)

julienmancuso (commented):
One documentation follow-up from the latest review.

deploy/operator/internal/webhook/validation/shared_helpers.go
```diff
@@ -163,6 +164,42 @@ func inferencePoolAvailabilityError(ctx context.Context, mgr ctrl.Manager) error
...
+		return allErrs
+	}
+	mainContainer := &containers[index]
+	if len(mainContainer.Command) > 0 || !dynamo.IsElasticEPRayLaunch(mainContainer) {
+		return allErrs
+	}
+	commandPath := fldPath.Child("podTemplate", "spec", "containers").Index(index).Child("command")
+	allErrs = append(allErrs, field.Required(
+		commandPath,
+		"elastic expert parallelism (--enable-elastic-ep with --data-parallel-backend ray) requires an explicit "+
```
julienmancuso, line 196: [P2] Document the explicit-command requirement

This adds a user-visible admission requirement, but the public Elastic EP guide currently tells users to install Ray and add `--data-parallel-backend ray` without saying that the main container must define an explicit `command` (`docs/fern/pages/kubernetes/model-deployment/multinode-deployments.md`, “For Elastic EP”). The repository's single-pod and multinode Elastic EP templates already use `command: [python3, -m, dynamo.vllm]`. Please document that requirement and include a YAML snippet; otherwise users can follow the published steps and receive this rejection without being told how to make their manifest valid.

julienmancuso (approved):

deploy/operator/internal/webhook/validation/shared_helpers.go
tzulingk: Done in c94252aa59. Expanded the "For Elastic EP (Data Parallel with Ray)" section of the multinode deployment guide with a third step, a YAML snippet, and an admonition: the main container needs an explicit `command` (the operator wraps it to start the single-pod Ray head and can't fall back to the image `ENTRYPOINT`), and a manifest that requests Elastic EP without one is rejected at admission — with `command: [python3, -m, dynamo.vllm]` shown as the fix. Thanks for the approval and the follow-up.

jh-nv (approved):
doc owner approval

PR #12945, opened 2026-08-10
docs(fern): add External and Research publications to the Blog tab

harryskim:
## Overview:

Rebuilds the Blog tab around three sibling entries — **External publications** (52 customer and partner articles), **Research publications** (24 papers), and **Dynamo blogs** (the existing first-party articles).

Layouts were prototyped and reviewed against a live `fern docs dev` server before anything was written; the rejected treatments are not in this branch.

### External publications — 52 partner articles

Compact link cards, two-up, reusing the border, radius and hover treatment of the first-party grid in `BlogLanding`. Each card names the publisher, the date and the title, and sends the reader to the source. No summary or reading time — we hold no editorial content for these, and inventing it would be worse than omitting it.

**Nothing is embedded, and that isn't only an editorial choice.** I probed all 40 publisher domains: 24 send `X-Frame-Options` or a restrictive CSP `frame-ancestors`, and because the blockers include the multi-article publishers that's **36 of 52 articles (69%) that cannot be iframed** — AWS, Google Cloud, Microsoft, LinkedIn, CoreWeave, Baseten, Together AI, Gcore, and `nvidia.com`. A page where two thirds of the embeds are blank boxes isn't worth building.

**Logos key on the credited organisation, not the article host.** Several of these are published somewhere other than the author's own site — Doubleword on TelecomTV, Deloitte on LinkedIn, Amazon Ads on nvidia.com, AstraZeneca on the KubeCon schedule — so keying on the host put the wrong company's mark on four cards. 51 of 52 now carry their own logo; AstraZeneca serves `403` to every icon path and falls back to initials, which is also the behaviour for any future entry with no logo.

**Logos are inlined as `data:` URIs.** Fern rewrites asset paths only in MDX and `docs.yml`; a repo path in a `.tsx` reaches the browser verbatim and 404s once published. `scripts/check_asset_paths.py` documents this (it shipped once and blanked the Home hero) and names data URIs as the supported route. Each icon is downscaled to 48×48 optimised PNG — 92KB for the set. The tile keeps a light face in both themes because most of these marks are dark-on-transparent.

### Research publications — 24 papers

Same card, with the venue in a chip where the partner cards carry a logo: papers have no publisher mark to show, and the venue is the more useful signal.

**Every title and date is read from the source, not retyped** — arXiv abstract pages, Crossref for the ACM DOI, and the first page of the PDF for the MLSys, USENIX and SIGCOMM entries — so each card matches what the reader lands on.

Three entries are on IEEE Xplore, which returns `418` to scripted requests and won't load in a browser. Their titles were supplied by hand; Crossref and OpenAlex then confirmed each one independently and supplied venue and date. That cross-check also settled the spelling in *"Toward Disaggregated and Heterogenous AI Systems"* — both indexes carry it, so it's the published title.

Conference papers give no per-paper date, so those come from the venue at month precision, with the opening date as the sort key: **MLSys 2026** (Bellevue, May 18–22, per mlsys.org and the PDFs) and **SIGCOMM '26** (Workshop on Networks for AI Computing, Denver, Aug 17–21, per the paper's own ACM reference block). One entry keeps a bare year — the IEEE Access paper, where Crossref and OpenAlex both carry only a year.

### Sidebar

`External publications`, `Research publications` and `Dynamo blogs` sit as three peers. `Latest` is no longer its own entry — the page it pointed at is now the `Dynamo blogs` section overview, so it isn't orphaned. The section keeps `slug: digest`, so **existing URLs are unchanged**: the overview stays at `/dynamo/dev/digest` and each article at `/dynamo/dev/digest/<slug>`.

`Dynamo blogs` is a section and the other two are pages, and Fern weights section headers more heavily, so a rule levels all three — they were rendering at different indents, heights and weights. Measured equal afterwards on box position, indent, margin, padding, size, weight and line-height.

Note a consequence of the order: a Fern tab has no URL of its own and resolves to the first entry in its `layout:`, so **the Blog tab now opens on External publications** rather than on Dynamo's article grid. That is intended; the two cannot be set independently. Flipping the order would restore the previous landing behaviour at the cost of the sidebar order.

### Also in here

- `PublicationsStyles.tsx` holds the shared card CSS, since two pages need it. Naming it `*Styles.tsx` puts it under `check_style_components.py`'s glob automatically.
- Fern's prose styles set `svg { display: block }`, which dropped the external-link arrow onto a line of its own under each title. It is inline now, and tied to the closing word in a `nowrap` span so it can't be orphaned when the last line is full.

- **`docs/fern/components/publisher-logos.generated.ts`** — 38 third-party company logos, taken from each organisation's own site and committed here. Referential use, but a brand call rather than a technical one.
- **`docs/fern/index.yml`** — the sidebar restructure and the URL-preservation argument above.
- **`docs/fern/components/research-papers.data.ts`** — the paper list, its attributions and dates.

### Judgment calls, easy to reverse

- The **"Suggest a publication"** button fills the slot `Subscribe on GitHub` occupies on the blog page. There's no real equivalent for third-party articles, so the label and its target (a blank issue) are a guess.
- The source list's **Segment** column (`CSP`, `NCP`, `CRISP`, `Not tracked`) is deliberately **not** shipped. It reads as internal go-to-market taxonomy, and `docs/fern/AGENTS.md` prohibits internal references in shipped docs.
- Relationship notes on the papers (*"cites Dynamo"*, *"uses Dynamo"*, *"believe uses Dynamo"*) are **not** rendered. Several are hedged and none is verifiable from paper metadata.
- Two paper attributions follow the source over the working notes: **ARK** is by Georgia Tech authors, and `osdi26-hwang` is **Korea University**, OSDI '26.
- Five papers show a venue chip with no organisation, where the source list gave only a venue.

### Not blog posts

Four partner entries aren't articles: the AstraZeneca link is a KubeCon *schedule* entry, Amazon Ads is a GTC session recording, Deloitte is a *LinkedIn post*, and one Gcore link is a press release. Accurate under "publications", but the LinkedIn one is a weak citation if you'd rather cut it.

### Validation

Rendered against a live `fern docs dev` server throughout, and re-checked on a **cold** server — a hot-reloaded one masked a real compile failure earlier in this branch.

- `fern check`: 0 errors. `check_asset_paths.py`, `check_style_components.py`, `sync_site_css.py` all pass.
- All three pages return 200 and compile clean; 52 partner cards, 24 papers, 51 logo tiles, **0 broken images**.
- `/dynamo/dev/digest`, `/dynamo/dev/digest/flash-indexer` and `/dynamo/dev/digest/dynosim-pareto-frontier` still return 200 — the sidebar change preserved existing URLs.
- The three sidebar entries measured identical; all 24 paper cards have their arrow trailing the text, none stranded.
- Partner extraction was validated rather than trusted: all 52 titles matched their href at similarity 1.0, no link left over, no empty fields. Two rows defeat a naive parse — one wraps its date cell (`Mar 4, 2026 (updated)`) and two carry a month-only date — so the data is parsed by column x-position.

**Not verified:** logo *fidelity*. Each icon is whatever that organisation serves as its site icon, which for a few is a product mark rather than the corporate logo. They render cleanly, but I have not compared all 38 against official brand assets.

## Related Issues


docs/fern/components/EcosystemPublications.tsx
```diff
@@ -0,0 +1,252 @@
...
+ */
+import { PUBLICATIONS, type Publication } from "./publications.data";
+import { PUBLISHER_LOGOS } from "./publisher-logos.generated";
+
+const PUBLICATIONS_CSS = `
+.dynamo-pubs {
+  display: grid;
+  grid-template-columns: repeat(2, minmax(0, 1fr));
+  gap: 1rem;
+}
```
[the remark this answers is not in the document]
harryskim: Fixed in f07485b.

You're right that this was the one failure mode the hook exists for, and it bit twice on this branch — a backtick inside a CSS comment closed `BLOG_CSS` and took all three blog pages down with `digest is not defined`. It only surfaced on a cold server; the hot-reloaded dev server kept serving the last good compile.

I widened coverage rather than moving the CSS, so it stays with its component: `EcosystemPublications.tsx` is now named in the hook's `files:` pattern and in the script's default target list.

I did look at widening to every `components/*.tsx` and backed off. About twenty components hold a CSS literal, but three — `ModelEABuildCards` (`EA_CSS`), `TagLookup` (`TG_CSS`) and `TerminalDemo` (`PLAYER_CSS`) — interpolate into theirs deliberately, which this check reads as a defect. Doing it properly needs a way to mark those as intentional, so I left a note in the docstring rather than half-doing it here.

Verified by injecting a stray backtick into a comment in the new file: the hook reports it and exits 1, and passes again once removed.

docs/fern/index.yml
```diff
@@ -1118,11 +1118,18 @@ navigation:
...
+      # this section's overview instead, so the page is not orphaned and every
+      # URL is unchanged: the section keeps `slug: digest`, so the overview
+      # stays at /dynamo/dev/digest and each article at
+      # /dynamo/dev/digest/<slug>.
+      - section: Dynamo blogs
         path: pages/blog/latest.mdx
         slug: digest
-      - section: "2026"
-        slug: digest
         collapsed: open-by-default
```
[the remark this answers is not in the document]
harryskim: Good catch, and it was real — fixed in c2d2516, which landed after the commit you reviewed.

The `/digest` link was still wearing the tile styling it had as the `Latest` entry: an inset horizontal margin, a `min-height`, the stacked title-and-kicker type treatment, and the `"2026  ·  ENGINEERING STORIES"` label. Visibly, it left `Dynamo blogs` and `External publications` sitting at different indents, heights and weights despite being siblings.

The tile rules are gone and the type rule now covers both entries. Measured equal afterwards: same box position, indent, margin, padding, font size, weight and line-height.

On the hardcoded subtitle specifically — it had gone stale on its own terms too, since the section it now heads is `Dynamo blogs`, which is not year-scoped. Worth noting it was already being suppressed by a `content: none` override further down the file, so removing the rule changed nothing on screen; it just stops the file carrying a declaration that existed only to be cancelled.

Your read on the nesting/date-badge rules was correct: `.fern-sidebar-group-level-1` keeps the same depth and article styling is unaffected. I checked the blog landing and an article page after the change.

docs/fern/index.yml
```diff
@@ -1118,11 +1118,18 @@ navigation:
...
+        slug: external-publications
+
+      # "Latest" is no longer its own sidebar entry. The page it pointed at is
+      # this section's overview instead, so the page is not orphaned and every
+      # URL is unchanged: the section keeps `slug: digest`, so the overview
+      # stays at /dynamo/dev/digest and each article at
+      # /dynamo/dev/digest/<slug>.
+      - section: Dynamo blogs
         path: pages/blog/latest.mdx
         slug: digest
```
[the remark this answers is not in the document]
harryskim: Confirmed, and thanks — the mechanism you describe is exactly right. I checked the rendered nav and the Blog tab resolves to `href="/dynamo/dev/external-publications"`, so clicking it does open the third-party list.

This one is intended, though. Leading with External publications was a deliberate call about sidebar order, and since a Fern tab resolves to the first entry in its `layout:`, the landing destination follows from it — the two can't be set independently. Dynamo's own posts sit directly beneath it in the sidebar, one click away.

I did prototype the middle option: keeping the order but adding a "From the team → Dynamo blogs" strip above the cards, so the landing page read as an index over both halves rather than dropping straight into other companies' links. It didn't earn its space, so I've dropped it and left the page as-is.

The part of your comment that was straightforwardly correct is that the PR description contradicted the behaviour — it still claimed the tab opens on the article grid, which stopped being true once the order changed. I've corrected the description.

dmitry-tokarev-nv (approved):
LGTM from DevOps side

docs/fern/components/research-papers.data.ts
```diff
@@ -0,0 +1,223 @@
...
+    url: "https://saeed.github.io/files/arc_niac26.pdf",
+    org: "Georgia Tech",
+    venue: "SIGCOMM",
+    date: "Aug 2026",
+    iso: "2026-08-17",
+  },
+  {
+    title: "A Photonic-CXL Memory Appliance for Scalable KV Cache Management in LLM Inference",
+    url: "https://arxiv.org/abs/2607.27187",
+    org: "Marwell",
```
dagil-nvidia: **Must-fix — `Marwell` is a typo for `Marvell`.**

I pulled the arXiv abstract page for `2607.27187`: it says "Marvell" three times and "Marwell" zero. This puts a misspelled company name on a public page.

```suggestion
    org: "Marvell",
```

docs/fern/components/research-papers.data.ts
```diff
@@ -0,0 +1,223 @@
...
+    org: "SK Hynix",
+    venue: "arXiv",
+    date: "Dec 20, 2025",
+    iso: "2025-12-20",
+  },
+  {
+    title: "DuetServe: Harmonizing Prefill and Decode for LLM Serving via Adaptive GPU Multiplexing",
+    url: "https://arxiv.org/abs/2511.04791",
+    org: "",
+    venue: "ICML",
```
dagil-nvidia: **Must-fix — this `ICML` label isn't supported by the source.**

arXiv `2511.04791` (DuetServe) carries no `arxiv:comment` and no `journal_ref` — nothing in the record claims ICML. Every other arXiv-only paper in this file is labelled `venue: "arXiv"`.

Worth contrasting with line 123, where `2602.14516`'s arXiv comment genuinely reads "ICML 2026" — that one is fine. This is the single entry that breaks the file's own docstring ("Titles and dates are taken from the source itself... rather than retyped").

```suggestion
    venue: "arXiv",
```

docs/fern/components/EcosystemPublications.tsx
```diff
@@ -0,0 +1,148 @@
...
+            <h2 id="publications-heading">External publications</h2>
+            <p>
+              Deep dives, benchmarks, and deployment write-ups about Dynamo,
+              published by the customers and partners running it.
+            </p>
+          </div>
+          <div className="dynamo-blog-section-heading__actions">
+            <a
+              className="dynamo-blog-button dynamo-blog-button--secondary"
+              href="https://github.com/ai-dynamo/dynamo/issues/new"
```
dagil-nvidia: **Must-fix — this button doesn't land where the PR description says it does.**

The PR describes the target as "a blank issue", but `.github/ISSUE_TEMPLATE/config.yml` sets `blank_issues_enabled: false`. GitHub redirects `/issues/new` to the template picker — Bug report / Feature request / Contribution request / DEP — none of which fits "suggest a publication", so the reader hits a dead end.

```suggestion
              href="https://github.com/ai-dynamo/dynamo/issues/new?template=contribution_request.yml"
```

Dropping the button entirely is also reasonable if there's no good home for these.

docs/fern/components/research-papers.data.ts
```diff
@@ -0,0 +1,223 @@
...
+  /** Sort key, YYYY-MM-DD. Year-only entries are pinned to January. */
+  iso: string;
+}
+
+export const RESEARCH_PAPERS: ResearchPaper[] = [
+  {
+    title: "ARK: Avoiding Routing Collisions for KV Cache Transfer in Disaggregated LLM Inference",
+    url: "https://saeed.github.io/files/arc_niac26.pdf",
+    org: "Georgia Tech",
+    venue: "SIGCOMM",
```
dagil-nvidia: **Consider — a bare `SIGCOMM` chip over-credits this paper.**

Your own PR description says it's the Workshop on Networks for AI Computing, and the PDF filename is `arc_niac26.pdf`. Rendered as "SIGCOMM" it reads as a main-track paper. It's also the only venue here without a year, against `MLSys 2026` / `OSDI '26` / `EuroMLSys '26`.

```suggestion
    venue: "SIGCOMM '26 NIAC",
```

docs/fern/components/research-papers.data.ts
```diff
@@ -0,0 +1,223 @@
...
+    org: "",
+    venue: "arXiv",
+    date: "Mar 13, 2026",
+    iso: "2026-03-13",
+  },
+  {
+    title: "Efficient Multi-round LLM Inference over Disaggregated Serving",
+    url: "https://arxiv.org/abs/2602.14516",
+    org: "",
+    venue: "ICML",
```
dagil-nvidia: **Consider — carry the year, like the other venue chips.**

The arXiv comment on `2602.14516` supplies it ("ICML 2026"), and the rest of the list is year-qualified (`MLSys 2026`, `OSDI '26`, `EuroMLSys '26`).

```suggestion
    venue: "ICML 2026",
```

docs/fern/components/research-papers.data.ts
```diff
@@ -0,0 +1,223 @@
...
+    url: "https://arxiv.org/abs/2606.15050",
+    org: "Solyx AI",
+    venue: "arXiv",
+    date: "Jun 13, 2026",
+    iso: "2026-06-13",
+  },
+  {
+    title: "The Price of Anarchy in Disaggregated Inference",
+    url: "https://arxiv.org/abs/2606.17081",
+    org: "NCA",
```
dagil-nvidia: **Consider — `NCA` is an unexpanded acronym.**

The style guide asks for acronyms expanded on first use. The paper is single-author (Athos Georgiou); worth checking what this is meant to credit, or leaving `org: ""` and letting the venue chip carry it, as five other entries already do.

docs/fern/components/ResearchPublications.tsx
```diff
@@ -0,0 +1,88 @@
...
+        fill="none"
+        stroke="currentColor"
+        strokeWidth="2"
+        strokeLinecap="round"
+      />
+    </svg>
+  );
+}
+
+/** Matches the icon BlogLanding puts inside its secondary button. */
```
dagil-nvidia: **Consider — stale JSDoc, copy-pasted from `EcosystemPublications`.**

This comment describes `ExternalLinkIcon`, which this file doesn't have. It's now sitting on `ResearchCard`.

```suggestion
```

docs/fern/components/ResearchPublications.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * Same card as the partner list next door, with one swap: papers have no
+ * publisher logo worth showing, so the venue takes the tile's place. It is
+ * also the more useful signal on a paper.
+ *
+ * Card styles live in PublicationsStyles.tsx, shared with EcosystemPublications;
+ * a page rendering this must render <BlogStyles /> and <PublicationsStyles />.
+ */
+import { RESEARCH_PAPERS, type ResearchPaper } from "./research-papers.data";
+
+function ExternalMark() {
```
dagil-nvidia: **Consider — `ExternalMark` is duplicated verbatim.**

Identical to `EcosystemPublications.tsx:24-36`. Since `PublicationsStyles.tsx` already exists as the shared home for what these two pages have in common, this could move alongside it.

docs/fern/components/publications.data.ts
```diff
@@ -0,0 +1,454 @@
...
+ * carries the title, the publisher and the date, and sends the reader to the
+ * source. Nothing is mirrored or embedded -- most of these publishers send
+ * X-Frame-Options or a restrictive frame-ancestors, so an embed would render
+ * an empty box for roughly two thirds of the list even if we wanted one.
+ *
+ * `date` is the publisher's date as printed on the article; `iso` is a sort key
+ * only, so a month-only date is normalised to the first of the month and the
+ * one undated entry sorts last. Keep the array in reverse-chronological order.
+ *
+ * `domain` keys into PUBLISHER_LOGOS. A domain with no icon falls back to the
```
dagil-nvidia: **Consider — this docstring contradicts the code.**

It says `domain` keys into `PUBLISHER_LOGOS`, but `EcosystemPublications.tsx:60` does `PUBLISHER_LOGOS[partner]`. The PR description has this right (logos key on the credited organisation, not the host) — the file comment was left behind, and it's the one a future editor will read first.

docs/fern/components/publications.data.ts
```diff
@@ -0,0 +1,454 @@
...
+ */
+
+export interface Publication {
+  title: string;
+  url: string;
+  /** Publishing organisation, as it should be shown to readers. */
+  partner: string;
+  /** Human-readable date exactly as published, or "Undated". */
+  date: string;
+  /** Sort key, YYYY-MM-DD. Null when the source carries no date. */
```
dagil-nvidia: **Consider — `iso`, `updated` and `domain` are declared but never read.**

Nothing in the components destructures them:

- `iso` is documented as a sort key, but nothing sorts by it — ordering is a hand-maintained comment, so `iso` and `date` can drift apart silently.
- `updated: true` (line 324) renders nothing. The `(updated)` nuance the PR describes carefully parsing out of the source is dropped at the component.
- `domain` is unused, and its docstring is wrong (see above).

Either wire them up (sort by `iso`, render `updated`) or drop all three. Carrying 52 rows of data that nothing reads is a maintenance tax with no payer.

docs/fern/components/publisher-logos.generated.ts
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * AUTO-GENERATED. Publisher logos, keyed by the organisation credited in
```
dagil-nvidia: **Consider — "AUTO-GENERATED" with no generator committed.**

The sibling `events.generated.ts:5` names its producer (`.github/scripts/generate-events.js`). Here there's nothing, so this is 96KB of base64 that nobody can reproduce or refresh — including whoever needs to add the 53rd publisher.

Either commit the script, or drop the `.generated` suffix and the banner so it reads as what it is: a hand-maintained asset file.

docs/fern/scripts/check_style_components.py
```diff
@@ -66,9 +73,19 @@ def check(path: Path) -> list[str]:
     return problems
 
 
+# Components that hold a CSS literal without being named *Styles.tsx. Keep in
```
dagil-nvidia: **Consider — `EXTRA_TARGETS` is empty scaffolding.**

It has no current user, and it introduces a two-place sync requirement (this tuple and the hook's `files:` pattern) that a future contributor has to notice. Since the publication CSS landed in `PublicationsStyles.tsx` and the existing glob already covers it, this could wait until something actually needs it.

dagil-nvidia (commented):
## Docs review — External and Research publications

**4 must-fix, 9 consider.** Structurally clean; the substance to fix is in the research paper data.

The deterministic gates all pass on `018ba3a`: docs linter 0 findings across all 11 doc files, and `check_asset_paths.py`, `check_style_components.py` and `sync_site_css.py` each pass. The validation claims in the description hold up.

**What I verified rather than took on trust.** I cross-checked all 16 arXiv-hosted papers against the arXiv API: **every title and every date matches the source exactly.** The "read from source, not retyped" claim is real. Two venue labels are the only exceptions, both flagged inline. On the partner side: 51 of 52 entries resolve a logo key with AstraZeneca falling back to initials, exactly as described; no orphan logo keys; no duplicate URLs, so the `key={url}` React keys are safe.

Also worth recording so nobody "fixes" it later: **"Heterogenous" in the IEEE Micro title is correct** — it's the published spelling, as the description explains.

### Must-fix

Three are inline below (`Marwell` → `Marvell`; DuetServe's unsupported `ICML` label; the "Suggest a publication" button landing on a template picker with no matching template).

The fourth is not anchorable: **this branch currently conflicts with `main`** and needs a rebase before it can merge.

### One decision for a maintainer, not a nit

A Fern tab resolves to the first entry in its `layout:`, so **the Blog tab now opens on External publications** instead of Dynamo's own article grid. The description flags this as intended and notes the two can't be set independently — which is right. But it demotes first-party writing on the blog tab in favour of a partner link list, and that's a product call worth someone signing off on explicitly rather than inheriting from sidebar ordering.

### On the parts I can't check

Logo fidelity is the honest gap the description already names, and I can't close it either — I confirmed the 38 marks resolve and render, not that each is the right corporate mark. That one needs human eyes.

Nice work on the sourcing discipline here; the arXiv cross-check coming back clean across 16 papers is not the usual outcome for a list this size.

docs/fern/components/research-papers.data.ts
harryskim: Fixed in 47f2faf. Confirmed against the arXiv page for 2607.27187: Marvell three times, Marwell none. It came in from the source list and I carried it without checking it against the paper — exactly the check the file's own docstring promises.

docs/fern/components/research-papers.data.ts
harryskim: Fixed in 47f2faf — now `venue: "arXiv"`.

You're right that nothing supports ICML: I pulled the record and 2511.04791 has no `arxiv:comment` and no `journal_ref`. The label came from the source list and I took it on trust while reading everything else from the source.

And your contrast is right too: 2602.14516's comment genuinely reads "ICML 2026", so that one keeps ICML — with the year added, per your other note.

docs/fern/components/EcosystemPublications.tsx
harryskim: Fixed in 47f2faf by removing the button.

Confirmed `blank_issues_enabled: false`, so `/issues/new` lands on the picker. I looked at pointing it at `contribution_request.yml` as you suggest, but its description is "Propose a change you want to implement" — someone flagging a third-party article isn't implementing anything, so that's a slightly wrong door rather than a dead end.

I invented this button to fill the slot `Subscribe on GitHub` occupies on the blog page, and flagged it twice as a guess. With no real destination it's better gone. Happy to add it back if a suitable template or a Discussions category exists.

docs/fern/components/research-papers.data.ts
harryskim: Taken — `SIGCOMM '26 NIAC` in 47f2faf. The paper's ACM reference block reads "Workshop on Networks for AI Computing (SIGCOMM '26), August 17–21, 2026", so a bare SIGCOMM chip did over-credit it, and it now carries a year like the rest.

docs/fern/components/research-papers.data.ts
harryskim: Taken — `ICML 2026` in 47f2faf, on the strength of the arXiv comment as you say.

docs/fern/components/research-papers.data.ts
harryskim: Taken — `org` is now empty for 2606.17081, so the venue chip carries it, as it does for the other unattributed entries.

I could not establish what NCA refers to: the paper is single-author (Athos Georgiou) with no stated affiliation, and the label came from the source list. Better blank than an acronym neither of us can expand.

docs/fern/components/ResearchPublications.tsx
harryskim: Fixed in 47f2faf — the JSDoc travelled with the copied function and described `ExternalLinkIcon`, which this file never had. Gone along with the duplication itself.

docs/fern/components/ResearchPublications.tsx
harryskim: Fixed in 47f2faf — `ExternalMark` now lives in `PublicationsStyles.tsx` and both pages import it, as you suggest.

docs/fern/components/publications.data.ts
harryskim: Fixed in 47f2faf. The docstring described the first implementation, which keyed logos on the host; when I switched to keying on `partner` (to stop Doubleword showing TelecomTV's mark) the comment was left behind. It now states the behaviour and the reason.

docs/fern/components/publications.data.ts
harryskim: All three addressed in 47f2faf.

- `iso` — you're right that ordering was a hand-maintained comment. Both lists now sort on `iso`, so date and position cannot drift.
- `updated` — now rendered next to the date, so the one entry carrying it says so rather than the parsing being discarded at the component.
- `domain` — removed. It was left over from host-keyed logos and had no reader.

docs/fern/components/publisher-logos.generated.ts
harryskim: Fixed in 059ae21 by committing the generator: `docs/fern/scripts/generate_publisher_logos.py`, named in the banner the way `events.generated.ts` names its own.

It holds the organisation-to-site map, fetches each icon, downscales to 48x48 and emits the file; `--check` fails if the committed copy is stale, so it can be wired to a hook.

Running it found a real bug your comment would have hidden: the candidate list was capped before the `/favicon.ico` fallback, so Deloitte — whose declared `<link rel="icon">` entries all fail — had silently dropped to initials. 38 of 39 publishers now; AstraZeneca answers 403 to every path. The regenerated data URIs are byte-identical to what was committed, so only the banner moved.

docs/fern/scripts/check_style_components.py
harryskim: Fixed in 059ae21 — removed. It was left empty when the publication CSS moved into `PublicationsStyles.tsx`, and you're right that an unused hook with a two-place sync requirement is a trap. The docstring now says to keep component CSS in a `*Styles.tsx` file instead, which is the behaviour that made it unnecessary.

dagil-nvidia (approved):
Approving the docs content.

The lists hold up. I cross-checked all 16 arXiv-hosted papers against the arXiv API and every title and date matches the source, which is not the usual result for a list this size. The review round cleared 11 of 12 findings, and the twelfth (tab landing order) is argued well enough that I'm taking it.

Three notes, none blocking.

**`2d21180` is mine.** black 23.1.0 was failing on the new generator script, so I pushed the reformat instead of sending it back to you. Whitespace only, identical AST before and after. It should have come with a heads-up first. Say the word and I'll drop it so you can land it yourself.

**The three red checks are main drift, not this PR.** `gen-llms-tables-check` flags `release-artifacts.mdx`, `releases-machine-readable.mdx` and `releases.json`; `Generated API References` flags `frontend.mdx`. This branch touches none of them. Merging main in should clear all three, and `pre-merge-status-check` with them.

**Two things I left alone on purpose.** The cards open in a new tab with no screen-reader cue, which is a AAA criterion most sites skip. And both data files still say to keep the array in reverse-chronological order, which stopped being true when the components started sorting on `iso`. Neither is worth another round; fold them in if you touch these files again.

dagil-nvidia:
/ok to test [5b9cfbe](https://github.com/ai-dynamo/dynamo/pull/12945/commits/5b9cfbe59eff0c60829362aba3dcf30e5e612cb1)

PR #12951, opened 2026-08-10
feat(mocker): support priority-aware controlled offline replay

Aphoh:
## Summary

- replace the vLLM mocker waiting deque with stable priority ordering
- add a generic event-driven `ReplayWorkSource` boundary for externally controlled offline replay
- expose request terminal/output callbacks, internal controller timers, and per-request capture
- support request routing constraints, worker taints, and per-worker sequence limits
- preserve prebuilt-request admission when pending destinations block fresh work

## Validation

- `cargo test -p dynamo-mocker --lib` (723 passed)
- `cargo fmt --check`
- `git diff --check`

Includes Dynamo-native tests for priority/FIFO ordering, delayed follow-up turns, taint-constrained routing, and per-worker sequence limits.

## Summary by CodeRabbit

* **New Features**
  * Added controlled replay simulation with configurable request capture and external work sources.
  * Added per-request routing constraints, including required and preferred worker taints.
  * Added per-worker taints and maximum active-sequence limits.
  * Added public replay controls and runtime reporting interfaces.

* **Bug Fixes**
  * Replay routing constraints are now preserved throughout request scheduling.
  * Waiting requests now follow stable priority order, with FIFO handling for equal priorities.
  * Improved validation for worker configuration and replay request consistency.

* **Documentation**
  * Clarified how priority and strict-priority settings affect replay scheduling and routing.

### DIFF
--- a/lib/mocker/src/replay/offline/extensions/kv_router/mod.rs
+++ b/lib/mocker/src/replay/offline/extensions/kv_router/mod.rs
@@ -256,6 +256,7 @@ struct PendingRequest {
     strict_priority: u32,
     policy_class: Option<String>,
     session_id: Option<String>,
+    routing_constraints: RoutingConstraints,
 }
 
 impl PendingRequest {
@@ -304,7 +305,7 @@ impl PendingRequest {
             expected_output_tokens: self.expected_output_tokens,
             pinned_worker: None,
             allowed_worker_ids: None,
-            routing_constraints: RoutingConstraints::default(),
+            routing_constraints: self.routing_constraints.clone(),
             shared_cache_hits: None,
             resp_tx: None,
         }
@@ -378,6 +379,7 @@ trait PlacementRequestView {
     fn metadata(&self) -> &DirectRequest;
     fn input_length(&self) -> usize;
     fn prompt_tokens(&self) -> Cow<'_, [u32]>;
+    fn routing_constraints(&self) -> RoutingConstraints;
 }
 
 impl PlacementRequestView for DirectRequest {
@@ -391,6 +393,10 @@ impl PlacementRequestView for DirectRequest {
 
     fn prompt_tokens(&self) -> Cow<'_, [u32]> {
         Cow::Borrowed(&self.tokens)
+    }
+
+    fn routing_constraints(&self) -> RoutingConstraints {
+        RoutingConstraints::default()
     }
 }
 
@@ -408,6 +414,10 @@ impl PlacementRequestView for ReplayRequestPayload {
             Some(tokens) => Cow::Borrowed(tokens),
             None => Cow::Owned(ReplayRequestPayload::prompt_tokens(self)),
         }
+    }
+
+    fn routing_constraints(&self) -> RoutingConstraints {
+        self.routing_constraints().clone()
     }
 }
 
@@ -505,6 +515,12 @@ impl OfflineReplayRouter {
         prefill_load_estimator: Option<ReplayPrefillLoadEstimator>,
         num_workers: usize,
     ) -> Result<Self> {
+        anyhow::ensure!(
+            args.worker_taints.is_empty() || args.worker_taints.len() == num_workers,
+            "worker_taints must be empty or contain exactly one entry per replay worker: got {} for {} workers",
+            args.worker_taints.len(),
+            num_workers,
+        );
         let config = replay_router_config(args, router_config);
         let tracking_hash = TrackingHashContext::from_config(&config)?;
         let worker_config_template = replay_worker_config(args);
@@ -859,6 +875,7 @@ impl OfflineReplayRouter {
             strict_priority,
             policy_class: request.policy_class.clone(),
             session_id,
+            routing_constraints: request_view.routing_constraints(),
         })
     }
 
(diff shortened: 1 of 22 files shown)


PeaBrane (commented):
Before merge, please run the `dynamo-kv-replay-parity` campaign for this PR and attach its semantic and performance results.

For the existing 5,000-request Mooncake matrix, the expectation should remain exact byte parity for unchanged behavior across every authoritative row. Differences intentionally introduced by the new priority-aware scheduling may be reported as `PASS_WITH_SEMANTIC_EXCEPTIONS`, but each exception should identify the exact report/lifecycle differences and have a focused regression test. This keeps intentional semantic changes separate from incidental drift.

The paired performance gate is also required because this changes shared offline admission/event plumbing and the vLLM waiting, preemption, and KVBM swap-in queue paths. Please apply the skill’s existing 1.05 paired replay-execution gate and binary/`.text` size guardrails, and investigate any failing or inconclusive regression under representative queued load.

The canonical Mooncake window is timestamp-driven and does not exercise completion-dependent closed-loop source behavior. This PR therefore also needs a supplemental controlled-replay leg—or an extension of the skill that prescribes one—using a pinned and checksummed slice of Applied Compute’s raw [`agentic_coding_8k`](https://github.com/Applied-Compute/trie/tree/main/workloads) workload in `applied_compute_agentic` form, rather than the flattened `toolagent_trace.jsonl` Mooncake derivative. Its lowering/`WorkloadDriver` should be exercised through the new `ReplayWorkSource` path.

The supplemental leg should require exact parity wherever the controlled and existing workload paths are intended to be equivalent, while explicitly classifying genuinely new behavior. Focused coverage should include output and terminal callbacks, waiting with no source timer while work is in flight, terminal-armed tool delays and follow-up turns, routing constraints, and per-worker sequence limits. It should include a representative correctness and performance comparison without changing the authoritative Mooncake matrix.

Any focused white-box instrumentation needed to establish this evidence—temporary lifecycle tracing, counters, detailed captures, or debug logging—should remain test/bench-only or be applied as a separately checksummed temporary patch. It should not remain in the final production-shaped code or measured semantic delta. Please keep raw logs and generated artifacts under temporary storage, remove the instrumentation before the final SHA, and rerun the production-shaped correctness and performance gates to confirm that the submitted code contains no validation-only overhead.

lib/mocker/src/scheduler/vllm/core.rs
```diff
@@ -94,10 +100,102 @@ impl VllmRequestState {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Stable priority queue used by the vLLM waiting path. It intentionally keeps
+/// the small queue-shaped inspection API used by scheduler diagnostics and
+/// tests while making dequeue order explicit.
+#[derive(Default)]
+pub(crate) struct StablePriorityQueue {
```
jthomson04, line 145: [perf] Keep the default FIFO path O(1). This `BTreeSet` replaces `VecDeque` for every mock vLLM replay. Even when all requests use `priority = 0`, each waiting insert and normal removal is now O(log W), with tree allocation and poorer locality. This affects existing replay workloads, not only the new controlled path. Please keep a `VecDeque` fast path until a non-default priority is present, use priority buckets, or add an explicit priority mode. We also need queued-load performance evidence before applying this shared cost.

lib/mocker/src/replay/offline/extensions/kv_router/mod.rs
```diff
@@ -191,32 +195,61 @@ pub(crate) struct OfflineRouterSnapshot {
...
     }
 
-    fn find_matches_for_hashes(&self, local_block_hashes: Vec<LocalBlockHash>) -> OverlapScores {
-        self.tree.find_matches(local_block_hashes, false)
+    fn find_matches_for_hashes(
+        &mut self,
+        local_block_hashes: &[LocalBlockHash],
+        now_ms: f64,
+    ) -> OverlapScores {
+        let mut scores = self.tree.find_matches(local_block_hashes.to_vec(), false);
```
jthomson04, line 228: [perf] Avoid copying every request's block hashes when prediction is disabled. This `to_vec()` runs for all offline KV-router lookups. The pre-PR path transferred the owned vector into `RadixTree::find_matches`; now long prompts allocate and copy O(B) hashes even when `predicted` is `None`. Keep an owned fast path for the primary tree and clone only when the optional predicted index is enabled.

lib/mocker/src/replay/offline/extensions/kv_router/mod.rs
```diff
@@ -859,9 +1244,62 @@ impl OfflineReplayRouter {
...
+    fn bind_session_affinity(
+        &mut self,
+        session_id: Option<&str>,
+        worker: WorkerWithDpRank,
+        now_ms: f64,
+    ) {
+        let (Some(ttl), Some(session_id)) = (self.session_affinity_ttl, session_id) else {
+            return;
+        };
+        self.session_affinity.insert(
```
jthomson04, line 1294: [perf] Expire affinity entries globally, not only on session reuse. `session_affinity` stores an owned session string on every bind. An expired entry is removed only when the same session is looked up again or its worker is removed. A trace with many one-shot sessions therefore grows this map for the full run even with a short TTL. Add an expiry index, a periodic sweep, or a hard bound so memory follows active TTL state instead of total unique sessions.

jthomson04 (changes requested):
Blocking this for now due to some perf concerns.

PR #12957, opened 2026-08-10
feat(snapshot): add PageBroker wire contract

dfeigin-nv:
## Scope

A narrow, temporary protobuf contract between Snapshot and PageBroker. It is not the ControlPlane API; follow-up work will replace or extend it as the ControlPlane shape is settled.

## Stack

This is PR 1 of 12. The tests include e2e checkpoint & restore after branch 12. Merge in order:

1. #12957 `pagebroker-wire` -> `main` — wire contract
2. #12958 `pagebroker-daemon` -> `pagebroker-wire` — tmpfs daemon
3. #13032 `pagebroker-daemon-tests` -> `pagebroker-daemon` — daemon tests
4. #13098 `pagebroker-checkpoint` -> `pagebroker-daemon-tests` — checkpoint flow
5. #13103 `pagebroker-nsmount` -> `pagebroker-checkpoint` — staging mounts
6. #13165 `pagebroker-daemon-server` -> `pagebroker-nsmount` — Unix socket server
7. #13166 `pagebroker-daemon-shutdown` -> `pagebroker-daemon-server` — signal shutdown
8. #13167 `pagebroker-restore` -> `pagebroker-daemon-shutdown` — opted-in restores
9. #13168 `pagebroker-kubernetes` -> `pagebroker-restore` — sidecar deployment
10. #13170 `pagebroker-rootfs-diff` -> `pagebroker-kubernetes` — runtime storage override
11. #13171 `pagebroker-image` -> `pagebroker-rootfs-diff` — packaged agent image
12. #13172 `pagebroker-daemon-concurrency` -> `pagebroker-image` — concurrent staging and bounded terminal retention

## Contract

- Snapshot assigns the transaction ID before its first request.
- Each request carries its own request ID.
- Restore and checkpoint-prepare select typed filesystem storage and POSIX-copy I/O.
- Commit and Abort are idempotent for a lost response.
- Responses are a typed `oneof`; a response cannot carry both a success result and a failure.

`AbortComplete` means only that PageBroker released transaction state/content. It says nothing about target-process or CUDA state.

## Failure codes

`INVALID_REQUEST` and `INSUFFICIENT_STORAGE` reject before PageBroker creates staging. `TRANSACTION_CONFLICT` leaves existing state unchanged. `TRANSACTION_NOT_FOUND` identifies an unknown transaction or the opposite terminal state. `STORAGE_ERROR` and `INTERNAL_ERROR` require aborting before a new-transaction retry.

The PageBroker design document has the complete failure-code section.

## Checks

`go generate ./internal/pagebroker`

`go test ./internal/pagebroker`


deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,52 @@
...
+}
+message PrepareCheckpointRequest
+{
+  // Required. The checkpoint directory to publish on commit.
+  optional string destination_path = 1;
+}
+message CommitRequest
+{
+  // Required. The transaction returned by PrepareCheckpointRequest.
+  optional string transaction_id = 1;
```
hhzhang16: According to the PageBroker design, it should go something like:

- Snapshot allocates transaction_id.
- Every message carries a request ID.
- The daemon rejects reuse while a transaction is live.

But here, you have it so that `PrepareCheckpointRequest(destination_path)` is called and then returns a transaction_id. But this is dangerous because if PageBroker creates/stages the transaction but the response is lost, Snapshot never learns/knows the ID. It should probably be on the caller/Snapshot side to allocate and determine the transaction ID.

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,52 @@
...
+    AbortRequest abort = 4;
+  }
+}
+message Response
+{
+  // Required. False means error contains the failure reason.
+  optional bool ok = 1;
+  optional string transaction_id = 2;
+  optional string directory_path = 3;
+  optional string error = 4;
```
hhzhang16: Would it be cleaner to have Typed/separate responses? This is allowing weird combos like "ok + error" to be set, and it would be nice to have undesired outcomes like that not even be possible

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,52 @@
...
+  optional string source_path = 1;
+  // Required. The PageBroker restore mode.
+  optional Mode mode = 2;
+}
+message PrepareCheckpointRequest
+{
+  // Required. The checkpoint directory to publish on commit.
+  optional string destination_path = 1;
+}
+message CommitRequest
```
hhzhang16: Are commit and abort intended to be idempotent/retryable? Imo yes, the caller has to be able to repeat either operation if the response is lost so I think the contract should say what happens when the transaction was already committed, already aborted, or is unknown.

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,52 @@
...
+{
+  // Required. The transaction returned by PrepareCheckpointRequest.
+  optional string transaction_id = 1;
+}
+message AbortRequest
+{
```
hhzhang16: Can we be precise that abort cancels/releases PageBroker transaction state but does not necessarily return the CUDA target to RUNNING? Post-handle failures --> cannot unlock or retry the target.

deploy/snapshot/pagebroker/pagebroker.proto
hhzhang16: Machine-readable errors would let the caller distinguish retryable/declined-before-mutation failures from terminal failures after CUDA has started

hhzhang16 (commented):
I noticed the APIs defined here don’t match the current ControlPlane API design. Is this proto meant to be a narrow subset that you'll go on to extend or is it meant to be a temporary contract that’ll eventually be replaced?

dfeigin-nv:
> I noticed the APIs defined here don’t match the current ControlPlane API design. Is this proto meant to be a narrow subset that you'll go on to extend or is it meant to be a temporary contract that’ll eventually be replaced?

@hhzhang16 Both narrow and temporary.

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: precise in a comment? documentation? Code in snapshot that specifies the state if this restore is fatal? What do you think is the best way to specify that?

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Agreed. I'll fix it.

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Agreed. Will fix

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Agreed. Will fix.

deploy/snapshot/pagebroker/pagebroker.proto
hhzhang16: I think a proto comment is enough

hhzhang16: Something like "AbortComplete only confirms that PageBroker transaction state/content was released; it does not imply anything about the target process or CUDA state."

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,99 @@
...
+message Failure
+{
+  enum Code {
+    UNSPECIFIED = 0;
+    INVALID_REQUEST = 1;
+    TRANSACTION_NOT_FOUND = 2;
+    TRANSACTION_CONFLICT = 3;
+    INSUFFICIENT_STORAGE = 4;
+    STORAGE_ERROR = 5;
+    INTERNAL_ERROR = 6;
```
hhzhang16, line 101: Can your doc include a section explaining each of these codes? Can be a table or section, can be folded within the "Failures and Edge Cases" section

dfeigin-nv: Yes

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Added the requested AbortComplete proto comment verbatim: it confirms only PageBroker state/content release and says nothing about target-process or CUDA state.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: Added a Wire failure codes subsection to the PageBroker design document under Failure and Edge Cases. It explains INVALID_REQUEST, TRANSACTION_NOT_FOUND, TRANSACTION_CONFLICT, INSUFFICIENT_STORAGE, STORAGE_ERROR, and INTERNAL_ERROR; matching proto comments are now generated into the Go binding.

hhzhang16 (approved):

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto2";
+package snapshot.pagebroker;
+option go_package = "github.com/ai-dynamo/dynamo/deploy/snapshot/internal/pagebroker";
+
+message FilesystemStorage
+{
+  required string directory = 1;
```
dmitry-tokarev-nv: 🔴 **Blocker — unconstrained caller-supplied path in a privileged daemon**

`directory` is consumed as both a restore *source* and a checkpoint *destination*, yet the contract constrains it in no way — no absoluteness guarantee stated here, no `..` prohibition, no allowlist, and no statement of which side validates. This module already treats exactly this class of value as security-relevant and says why: `cmd/ns-bind-mount/main.c` defines `ALLOWED_SRC_PREFIX`/`ALLOWED_DST_PREFIX` + `..`-rejection ("so that a bug in the controller cannot cause arbitrary host paths to be mounted"), and `internal/controller/controller.go` requires `filepath.IsAbs && filepath.Clean(location)==location`.

**Reproduced against the compiled daemon** (downstream `broker.cpp`), unprivileged, on Linux. The checkpoint *destination* is gated only by `is_absolute()` — no symlink/`..`/allowlist check, unlike the restore *source*, which checks `is_symlink`+`is_directory`. At `Commit`, `PublishCheckpoint` runs `fs::remove_all(published_directory)` then `fs::rename` on that path. PoC:

- destination = an unrelated dir → after `Commit` its `important_data.txt` is **deleted** and attacker-staged `payload.txt` is **present** → arbitrary-directory delete + content replace.
- destination = `…/checkpoint/../../victim` (absolute, contains `..`) → same result; `..` is **not** rejected.
- control: a relative path *is* rejected — confirming `is_absolute()` is the only gate.

Under the sidecar deployment (privileged DaemonSet, hostPath mounts), this is an arbitrary host-path delete/replace primitive.

**Fix:** specify the path contract in the schema and require the daemon to enforce it — absolute + canonical, reject `..`, confine to an allowlisted prefix, and open with `O_NOFOLLOW`/`RESOLVE_NO_SYMLINKS` to close the validate→open TOCTOU.

```suggestion
  // Absolute, canonicalized path confined to a configured allowlist prefix. The PageBroker
  // daemon MUST reject non-absolute paths, any `..` component, and symlinked components, and
  // MUST open with O_NOFOLLOW/RESOLVE_NO_SYMLINKS. Callers must not assume validation elsewhere.
  required string directory = 1;
```

dmitry-tokarev-nv (commented):
## Automated contract review — PageBroker wire protocol

Scope as requested: **DevOps/CI, testing, security, maintainability** of the new wire contract (`pagebroker.proto`, `generate.go`, `Makefile`, generated `pagebroker.pb.go`). This is a brand-new contract, so every point below is about what this PR *introduces and locks in*, not pre-existing code.

**How these were validated.** Wire-format claims were reproduced byte-for-byte by compiling this PR's own generated code (module-pinned `google.golang.org/protobuf`) and unmarshalling crafted frames. The path-traversal finding was additionally reproduced end-to-end against the **compiled daemon** (`broker.cpp` from the downstream `pagebroker-*` stack) with an unprivileged PoC — no privilege, no GPU. Where the daemon already guards a contract gap, that is stated and the severity lowered accordingly.

**Severity legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · ⚪ Nit
**Tally:** 1 blocker · 5 major · 6 minor · 1 nit (13 findings, inline).

One additional candidate — *"required envelope IDs make `INVALID_REQUEST` unreachable"* — was **dropped** after checking the daemon: `Fail()`→`Reply()` echoes the request IDs (`broker.cpp:19-20`), so the client's correlation check passes and the failure surfaces correctly. It does not hold.

Event is *comment*, not *request-changes* — the gating decision is the maintainers'. The blocker is the kind of thing worth resolving before this contract ships downstream.

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto2";
```
dmitry-tokarev-nv: 🟠 **Major — `proto2` + 15 `required` fields + no version segment forecloses the evolution this PR promises**

This is the only `proto2` file among the repo's protos; the in-repo exemplar `components/src/dynamo/planner/plugins/proto/v1/plugin.proto` is `proto3`, package `…​.v1`, with a written policy that new fields stay optional and breaking changes go to a new package path. Costs, measured against this PR's generated code:

- **Can't extend:** proto2 forbids *adding* a `required` field, and no `required` field can later be relaxed — an older sender omitting `io_engine` has its entire message discarded (`proto: required field …io_engine not set`) even though the bytes parsed.
- **Can't replace:** package `snapshot.pagebroker`, the Go import path, and the C++ namespace all collide with themselves, so `v1` and `v2` can't coexist in one binary/descriptor pool during a rolling upgrade.
- The descriptor path is the bare `"pagebroker.proto"` (verified via `File_…​.Path()`) — the only directory-less one reachable from `./cmd/agent`; `protoregistry`'s default conflict policy is *panic*, so a duplicate registration is an init-time process kill.

**Fix (one line now, breaking later):** `proto3` + explicit server-side validation instead of `required`; put the contract under a `…​.v1` package and versioned dir; root `--proto_path` so the descriptor path isn't bare. The PR body calling this contract "temporary" is the argument *for* versioning it now.

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+  required IOEngine io_engine = 2;
+}
+// Completes a live transaction. Repeating Commit for a committed transaction returns CommitComplete;
+// Commit for an aborted or unknown transaction returns TRANSACTION_NOT_FOUND.
+message CommitRequest {}
+// Releases PageBroker state for a live transaction. Repeating Abort for an aborted transaction returns
+// AbortComplete; Abort for a committed or unknown transaction returns TRANSACTION_NOT_FOUND.
+message AbortRequest {}
+
+message Request
```
dmitry-tokarev-nv, line 55: 🟠 **Major — possession of the `transaction_id` string is the entire capability**

`Request` carries no caller identity (no pod UID, container ID, token, or scope), and `CommitRequest{}`/`AbortRequest{}` are empty — nothing binds an operation to the transaction's creator. The daemon has zero identity checks (`peercred|ucred|getsockopt|uid|token|authoriz` in `broker.cpp` → none), and `SO_PEERCRED` can't rescue it because the Dockerfile runs both stages as `USER root`. PageBroker is a sidecar sharing its Unix socket with the workload container (arbitrary model code):

- a peer sending `Request{transaction_id:"<guessed>", abort:{}}` tears a victim's staged checkpoint out from under an in-flight CRIU dump;
- `commit:{}` promotes a partially-written staging dir into place as a "valid" restore image — silent corruption presented as success.

Non-malicious variant: the ID is client-assigned and `TRANSACTION_CONFLICT` rejects the second comer, so two agents during a rolling upgrade — or one agent restarting mid-transaction — collide by construction; "does not reuse it while live" is unenforceable across a restart because "live" is daemon state the new process never saw.

**Fix:** bind requests to a verifiable identity (SO_PEERCRED with distinct UIDs, or a per-transaction capability token issued at create time and required on Commit/Abort) and define the cross-restart ownership rule. Not addable without a breaking change once deployed.

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+    INTERNAL_ERROR = 6;
+  }
+  required Code code = 1;
+  required string message = 2;
+}
+message Response
+{
+  required string request_id = 1;
+  required string transaction_id = 2;
+  oneof result
```
dmitry-tokarev-nv, line 110: 🟠 **Major — the "cannot carry both success and failure" guarantee isn't provided by the wire format**

protobuf `oneof` is last-tag-wins under merge/concatenation, so trailing bytes silently flip which arm is set. Reproduced with this PR's generated code (cross-checked with `protoc --decode`): a well-formed error `Response{failure{INVALID_REQUEST,"bad"}}` with two bytes `2a 00` appended decodes in a single `Unmarshal` to `err=nil`, result `*Response_CommitComplete`, `GetFailure()==nil` — **the error is read as a successful commit**. Same on `Request.command` (bytes setting both `commit(5)` and `abort(6)` parse to `Abort`). Trigger is a framing desync or `UnmarshalOptions{Merge:true}` on a reused buffer, and the transport here is hand-written, not gRPC.

**Fix:** state the receiver policy — a frame encoding more than one arm of `result`/`command` MUST be rejected as malformed — and enforce it on both peers.

```suggestion
  // Exactly one arm. Receivers MUST reject a frame that encodes more than one arm: protobuf
  // oneof is last-wins under merge/concatenation, so trailing bytes can otherwise flip an
  // error response into a success. Do not rely on the sender setting only one.
  oneof result
```

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+    // The transaction ID or checkpoint destination conflicts with an existing transaction or output.
+    TRANSACTION_CONFLICT = 3;
+    // Tmpfs staging lacks space. PageBroker did not create staging for this request.
+    INSUFFICIENT_STORAGE = 4;
+    // Storage I/O failed. The caller may Abort the transaction before retrying with a new transaction ID.
+    STORAGE_ERROR = 5;
+    // An unexpected broker failure occurred. The caller may Abort the transaction before retrying.
+    INTERNAL_ERROR = 6;
+  }
+  required Code code = 1;
```
dmitry-tokarev-nv: 🟠 **Major — `required` closed enum makes every future `Failure.Code` a breaking change, on the error path**

Measured on both runtimes with a hypothetical future `code=7`. Go (module-pinned): `err=nil`, but `Code(7).String()` renders the bare literal `"7"` (no name-table entry), so any `switch` over the named constants falls to `default` and logs render `"7"`; negative values pass too. The C++ side generated by this PR's own Makefile (`libprotobuf 29.3`) does the opposite — `ParseFromString` returns **false**, `IsInitialized()==0`, the value is diverted to unknown fields, and the caller discards the whole object. One added enum value breaks one direction silently and the other hard — specifically on the failure path.

**Fix:** don't lock the enum as `required` in a closed (`proto2`) enum. `proto3` enums are open and unknown values round-trip; pair with a receiver rule mapping unknown codes to a generic failure. (Compounds the versioning point above.)

```suggestion
  // Receivers MUST map an unrecognized code to a generic failure (do not switch-default it into
  // UNSPECIFIED). As a proto2 closed/required enum an unknown code hard-fails the C++ parser
  // today — prefer proto3 open-enum semantics before adding codes.
  required Code code = 1;
```

deploy/snapshot/internal/pagebroker/generate.go
```diff
@@ -0,0 +1,6 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package pagebroker
+
+//go:generate protoc --proto_path=../../pagebroker --go_out=. --go_opt=paths=source_relative ../../pagebroker/pagebroker.proto
```
dmitry-tokarev-nv: 🟠 **Major — no drift gate for the generated Go, while the C++ half regenerates from the same `.proto` at build time**

`grep -rn "go generate"` across the repo returns zero matches — this directive is invoked by nothing, and `deploy/snapshot/Makefile` has no `generate`/`check` target. The gate that would catch a stale `.pb.go` already exists one component over (`deploy/operator/Makefile` → `check: generate … && git status --porcelain` exits 1), but `.github/actions/check-deploy-component/action.yml` fences it behind `component == 'operator'`, and `pre-merge.yml` calls that same action with `component: snapshot`. So a later stack PR that renumbers a tag and forgets to regenerate: `golangci-lint` and `go test ./...` pass against the stale `.pb.go` while the C++ daemon (compiled from the `.proto` fresh) speaks the new schema — the two halves disagree, discoverable only at runtime mid-checkpoint. **Fix:** extend the snapshot component's CI check to run `go generate ./...` + the C++ `protoc` and fail on a dirty tree.

🟡 **Minor — generator toolchain unpinned / already skewed.** No `tools.go`, no `tool` directive in any `go.mod`; the committed file records `protoc-gen-go v1.36.6` / `protoc v3.21.12` while `deploy/snapshot/go.mod` pins the runtime at `v1.36.12-…`. Rerunning the exact directive already yields a banner diff (`protoc v3.21.12`→`v5.29.3`) in a `DO NOT EDIT` file, so any drift gate would flap on the banner alone. Pin both tools (the component `Makefile` already has a `go-install-tool` helper + Tool-Versions block used for golangci-lint).

🟡 **Minor — the PR's stated validation is a no-op.** The package has zero `_test.go` files, so the PR body's `go test ./internal/pagebroker` prints `[no test files]` and exits 0 whether or not the `.pb.go` matches the `.proto`. The sibling `deploy/snapshot/protocol/` sets the opposite precedent (a `_test.go` per source), and `lib/sidecar/sglang/tests/proto_contract.rs` already does this exact job (asserts encoded wire tags). A ~20-line table test would verify the contract *and* double as the drift canary above.

deploy/snapshot/pagebroker/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+{
+  oneof kind
+  {
+    PosixCopyIOEngine posix_copy = 1;
+  }
+}
+
+message StagedRestoreRequest
+{
+  required StorageBackend source = 1;
```
dmitry-tokarev-nv: 🟡 **Minor — `required` on a submessage guarantees presence but not a populated `oneof` arm**

proto2 can't require a `oneof` arm, so `StagedRestoreRequest{source: StorageBackend{}, io_engine: IOEngine{}}` (no arm set on either) marshals, unmarshals, and passes `CheckInitialized` — all `err=nil` (measured, 12 bytes). `GetSource().GetFilesystem().GetDirectory()` then returns `""` with no error, and `filepath.Join("/staging","")` is `"/staging"`.

Impact is contained today because the **daemon guards it** — `IsFilesystemPosix()` requires `has_filesystem() && !directory().empty() && has_posix_copy()` and returns `INVALID_REQUEST` otherwise. So this is contract hygiene: the schema fails to express an invariant every consumer must re-check by hand. **Fix:** document that exactly one storage arm and one IO-engine arm are required and receivers reject the empty case.

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+    UNSPECIFIED = 0;
+    // The request is malformed or unsupported. PageBroker did not mutate transaction state.
+    INVALID_REQUEST = 1;
+    // The transaction is unknown or has reached the opposite terminal state.
+    TRANSACTION_NOT_FOUND = 2;
+    // The transaction ID or checkpoint destination conflicts with an existing transaction or output.
+    TRANSACTION_CONFLICT = 3;
+    // Tmpfs staging lacks space. PageBroker did not create staging for this request.
+    INSUFFICIENT_STORAGE = 4;
+    // Storage I/O failed. The caller may Abort the transaction before retrying with a new transaction ID.
```
dmitry-tokarev-nv, line 98: 🟡 **Minor — no lease / TTL / expiry, so an orphaned transaction pins tmpfs (node RAM)**

Recovery is pushed entirely onto the client ("The caller may Abort … before retrying"), and the transaction is deliberately decoupled from any single connection (client-assigned ID repeated on each request), so the daemon can't treat connection-close as an implicit abort without contradicting the contract. There's no TTL/deadline field anywhere. If the agent is OOM-killed between `PrepareStagedCheckpoint` and `Commit`/`Abort`, kubelet restarts just that container and the multi-GB staging dir stays resident in the `emptyDir: {medium: Memory}` volume until the whole pod restarts — one orphan per crash-loop cycle. Downstream #13172's retention cache does **not** cover this: it evicts *already-terminal* transactions (`retain_terminal()` is false for a LIVE staged txn) and `Broker()` does no startup purge. **Fix:** a broker-side TTL/lease + reaper (optionally a `deadline` field), independent of the client.

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+// AbortComplete; Abort for a committed or unknown transaction returns TRANSACTION_NOT_FOUND.
+message AbortRequest {}
+
+message Request
+{
+  // Snapshot assigns an ID to each request.
+  required string request_id = 1;
+  // Snapshot assigns a unique transaction ID before the first request and does not reuse it while live.
+  required string transaction_id = 2;
+  oneof command
```
dmitry-tokarev-nv, line 62: 🟡 **Minor — a lost response to a transaction-creating request is unrecoverable without discarding all staged bytes**

This `command` oneof has no status/query arm, and `image_directory` is returned exactly once — in the response to the creating request. If that response is lost (socket reset, client restart) after the broker copied GB of pages, retrying the identical request hits `TRANSACTION_CONFLICT` and there's no way to re-read `image_directory`; the only recovery is `Abort` + a fresh ID + redoing the whole copy, unbounded under a flaky link. **Fix:** add an idempotent status/read-back command returning current state + `image_directory` for a known transaction ID. (Only `Commit`/`Abort` are documented idempotent today.)

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+{
+  required string request_id = 1;
+  required string transaction_id = 2;
+  oneof result
+  {
+    StagedRestoreDirectory staged_restore_directory = 3;
+    StagedCheckpointDirectory staged_checkpoint_directory = 4;
+    CommitComplete commit_complete = 5;
+    AbortComplete abort_complete = 6;
+    Failure failure = 7;
```
dmitry-tokarev-nv, line 116: 🟡 **Minor — nested `required` isn't enforced for a non-first `oneof` arm (latent)**

protobuf-go wires nested-required checks only onto the *first-declared* arm of a oneof, so `Response{failure{message:"bad"}}` with `code` omitted decodes with `err=nil` and `GetFailure().GetCode()==UNSPECIFIED` (measured), whereas the same omission in the first arm (`staged_restore_directory`) *is* rejected. `proto.CheckInitialized` would catch it, but nothing calls it. Latent rather than active: the daemon's `Fail()` always sets `code`, so there's no current producer of a code-less failure — but it's one refactor away and renders as `UNSPECIFIED(0)` rather than an error. Folds away if `required` is dropped per the proto3 point.

deploy/snapshot/pagebroker/v1/pagebroker.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+message PrepareStagedCheckpointRequest
+{
+  required StorageBackend destination = 1;
+  required IOEngine io_engine = 2;
+}
+// Completes a live transaction. Repeating Commit for a committed transaction returns CommitComplete;
+// Commit for an aborted or unknown transaction returns TRANSACTION_NOT_FOUND.
+message CommitRequest {}
+// Releases PageBroker state for a live transaction. Repeating Abort for an aborted transaction returns
```
dmitry-tokarev-nv: ⚪ **Nit — the idempotency comments and the ID-reuse clause can't both hold**

The Commit/Abort comments require the daemon to keep a per-transaction-ID tombstone indefinitely (to return `AbortComplete`/`CommitComplete` on a repeat and `TRANSACTION_NOT_FOUND` for unknown), but line 55 permits the client to reuse an ID once terminal, and no TTL/GC/forget is defined. Once an implementer bounds the tombstone (downstream #13172 does, at 60s/1024), a retried Abort after a lost response gets `TRANSACTION_NOT_FOUND` instead of the documented `AbortComplete` — the very case the idempotency clause exists to resolve. Worth reconciling the wording: define the retention window and whether reuse-after-terminal is actually permitted.

deploy/snapshot/pagebroker/Makefile
```diff
@@ -0,0 +1,6 @@
+PROTO := pagebroker.proto
+
+.PHONY: generate
+
+generate:
+	protoc --cpp_out=. $(PROTO)
```
dmitry-tokarev-nv: 🟡 **Minor — `make generate` writes ~472 KB of generated C++ into a tracked dir nothing ignores (yet)**

`protoc --cpp_out=. pagebroker.proto` emits `pagebroker.pb.cc` (~226 KB) + `pagebroker.pb.h` (~246 KB) into `deploy/snapshot/pagebroker/`, and `git check-ignore` confirms neither is ignored on this branch — the `.gitignore` covering `/pagebroker.pb.*` is only added downstream in #13165. A contributor who edits the `.proto` and runs this discoverable Makefile gets a successful-looking generate that does *not* touch the committed `pagebroker.pb.go` (that's the separate `go:generate` command), then `git add -A` sweeps in half a MB of generated C++ — which protobuf stamps `NO CHECKED-IN PROTOBUF GENCODE` and hard-fails on any runtime bump. **Fix:** land `pagebroker/.gitignore` (`/pagebroker.pb.*`) in this PR alongside the Makefile that produces those files, or emit into a gitignored build dir.

dmitry-tokarev-nv (changes requested):
Posted findings

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: This is pre-existing Snapshot behavior: the agent accepts the checkpoint location and the existing checkpoint path publishes to it. I think it can be a follow-up. What do you think?

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Fixed in the rewritten stack: the protocol is now proto3 under `snapshot.pagebroker.v1`, with the source at `pagebroker/v1/pagebroker.proto`. Scalars that need presence use `optional`; receivers validate the request semantics.

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Fixed downstream: proto3 keeps unknown enum values parseable, and the Go client maps an unrecognized `Failure.Code` to `UNSPECIFIED`.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: The workload does not have the PageBroker socket. Only the Snapshot agent and PageBroker containers mount `/pagebroker`; during restore, the workload receives only the staged files at `/tmp/pagebroker`.

The agent makes a new UUID for every transaction. If PageBroker restarts, it loses its transaction state; if the agent restarts, it makes new IDs. Different DaemonSet pods also have separate sockets and `emptyDir` volumes.

If we ever expose PageBroker outside this pod boundary, we should add caller authentication and transaction ownership as part of that design. What do you think?

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: Protobuf gives the receiver one command or result after decoding. It does not preserve earlier `oneof` fields from the wire: if a malformed message contains two, the last one wins. That is protobuf’s documented behavior.

PageBroker does not accidentally combine messages: it reads one length-prefixed frame into a new `Request` or `Response`.

The normal approach for a private protocol like this is to validate the decoded message and trust the socket boundary. A separate raw protobuf parser is useful for an untrusted, strict protocol, but would add complexity here without protecting a boundary we expose. Do you still think it is necessary here?

deploy/snapshot/pagebroker/pagebroker.proto
dfeigin-nv: Added this to the v1 contract: requests require exactly one storage kind and I/O engine. The broker already rejects the empty cases before creating transaction state.

deploy/snapshot/internal/pagebroker/generate.go
dfeigin-nv: Added a generated-file check for PageBroker. CI uses the repository’s pinned `protoc` 30.2, installs `protoc-gen-go` from Snapshot’s `go.mod`, regenerates `pagebroker.pb.go`, and fails if it changes. The C++ binding is generated by the PageBroker build and is not checked in.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: Fixed downstream. An uncommitted transaction expires one hour after staging starts. PageBroker reaps expired staging every two minutes and clears its `restore/` and `checkpoint/` staging directories on startup.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: A lost staging reply makes the agent abort the transaction; if that abort is also lost, PageBroker expires the staging later. A retry uses a new transaction ID and copies again. This wastes work, but does not affect correctness, so I do not think we need a status command here.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: Fixed downstream by the proto3 migration. `Failure.code` is optional; a missing or unknown code maps to `UNSPECIFIED` and the response remains a failure. PageBroker sets a code on every failure.

deploy/snapshot/pagebroker/v1/pagebroker.proto
dfeigin-nv: Fixed downstream. PageBroker retains up to 1,024 terminal transactions for up to one hour. During that window, repeated `Commit` or `Abort` returns the terminal reply; afterward the transaction is unknown and its ID may be reused.

deploy/snapshot/pagebroker/Makefile
dfeigin-nv: Moved `deploy/snapshot/pagebroker/.gitignore` into the wire PR with the Makefile. It ignores the PageBroker build directory, daemon binary, and generated C++ bindings.

dfeigin-nv:
Moved to https://github.com/ai-dynamo/snapshot/pull/83.

PR #12958, opened 2026-08-10
feat(snapshot): add tmpfs PageBroker daemon

dfeigin-nv:
## What

A local PageBroker daemon for the v1 filesystem-storage / POSIX-copy plan.

- Restore materializes the requested image directory in broker-owned tmpfs.
- Checkpoint prepare allocates a broker-owned tmpfs output directory.
- Commit publishes checkpoint output with copy then rename, or removes restore staging.
- Abort removes broker-owned staging for either transaction.
- Commit and Abort retain terminal state, so retrying either after a lost response returns the same completion.

The Snapshot Agent sends Commit after CRIU returns.

## Scope

No S3, NIXL, GPU path, CRIU provider integration, or direct restore implementation.

## Validation

Regenerated the C++ protobuf binding and compiled `broker.cpp` with `-std=c++20 -Wall -Werror`. The stacked test PR runs the GoogleTest lifecycle suite.

### DIFF
--- a/deploy/snapshot/pagebroker/broker.cpp
+++ b/deploy/snapshot/pagebroker/broker.cpp
@@ -1,0 +1,282 @@
+#include "broker.hpp"
+
+#include <sys/statvfs.h>
+
+#include <filesystem>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <system_error>
+
+#include "posix_copy_engine.hpp"
+
+namespace snapshot::pagebroker {
+namespace fs = std::filesystem;
+namespace {
+
+Response
+Reply(const Request& request)
+{
+  Response response;
+  response.set_request_id(request.request_id());
+  response.set_transaction_id(request.transaction_id());
+  return response;
+}
+
+Response
+Fail(const Request& request, Failure::Code code, const std::string& message)
+{
+  auto response = Reply(request);
+  response.mutable_failure()->set_code(code);
+  response.mutable_failure()->set_message(message);
+  return response;
+}
+
+Response
+CommitSucceeded(const Request& request)
+{
+  auto response = Reply(request);
+  response.mutable_commit_complete();
+  return response;
+}
+
+Response
+AbortSucceeded(const Request& request)
+{
+  auto response = Reply(request);
+  response.mutable_abort_complete();
+  return response;
+}
+
+bool
+IsSafePathComponent(const std::string& value)
+{
+  return !value.empty() && value != "." && value != ".." && value.find('/') == std::string::npos &&
+         value.find('\\') == std::string::npos && value.find('\0') == std::string::npos;
+}
+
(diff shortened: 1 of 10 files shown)


deploy/snapshot/pagebroker/broker.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+
+}  // namespace
+
+Broker::Broker(Path staging_root) : staging_root_(fs::weakly_canonical(std::move(staging_root)))
+{
+  io_engines_.push_back(std::make_unique<PosixCopyEngine>());
+  fs::create_directories(staging_root_ / "restore");
+  fs::create_directories(staging_root_ / "checkpoint");
+}
```
hhzhang16, line 99: Commit/Abort retryability currently survives only while this Broker process survives. A sidecar restart reconstructs an empty transaction map while the Pod’s shared staging volume and any `.pagebroker-partial` on checkpoint storage can remain. The next Commit returns `TRANSACTION_NOT_FOUND`, and a stale partial can block publication indefinitely. Since this MR defines broker-owned staging and retryable terminal operations, please make startup reconcile persisted staging/partial state or persist enough transaction state to determine the outcome after restart.

deploy/snapshot/pagebroker/broker.cpp
```diff
@@ -0,0 +1,282 @@
...
+    fs::rename(partial, published_directory);
+    checkpoint_transactions_.erase(request.transaction_id());
+    transaction_states_.at(request.transaction_id()) = TransactionState::COMMITTED;
+    std::error_code cleanup_error;
+    fs::remove_all(staging_directory, cleanup_error);
+  }
+  catch (...) {
+    fs::remove_all(partial);
+    throw;
+  }
```
hhzhang16, line 259: Please define and test the outcome at each crash boundary here: after partial copy, after removing the previous destination, after rename, and after commit succeeds but before the reply reaches the client. In particular, a crash between `remove_all(published_directory)` and rename makes the previously valid checkpoint disappear.

hhzhang16 (commented):
I get the narrow scope you have in mind, but I’d like restart recovery and the copy/remove/rename crash semantics defined here rather than repaired in #13172

deploy/snapshot/pagebroker/broker.cpp
dfeigin-nv: I think this can be a follow-up. What do you think?

deploy/snapshot/pagebroker/broker.cpp
dfeigin-nv: I think this can be a follow-up. What do you think?

dfeigin-nv:
Moved to https://github.com/ai-dynamo/snapshot/pull/84.

PR #12962, opened 2026-08-10
ci(release): publish nightly Helm charts to NGC staging (ChartMuseum + OCI)

pvijayakrish:
Nightly packages the platform and snapshot charts at X.Y.Z-dev.YYYYMMDD.gSHA7 and dual-pushes them to the NGC staging ChartMuseum and OCI registries. prepare_nightly_helm.py rewrites chart versions and points image references at the *-nightly repos at this run's dated tag, ephemerally in the runner workspace. Per-chart fail-soft off the operator/snapshot staged flags; OCI pushed first and cm-push skipped when the version is already indexed, so same-SHA re-runs stay green.
Closes OPS-8230

## Summary by CodeRabbit

- **New Features**
  - Nightly Helm releases now use dated, SemVer-compliant chart versions.
  - Eligible platform and snapshot charts are packaged and published to ChartMuseum and OCI registries.
  - Nightly preparation automatically updates chart dependencies, image repositories, and tags for each release.
- **Improvements**
  - Releases handle missing images and duplicate chart entries without failing unnecessarily.
  - Release summaries now show which nightly charts were published or skipped.
  - Downstream nightly notifications include Helm artifacts only when charts are successfully pushed.


.github/workflows/release.yml
```diff
@@ -606,10 +614,121 @@ jobs:
...
+            --charts "${CHARTS}"
+
+          REPO_ALIAS="ngc-staging-dynamo"
+
+          helm plugin install https://github.com/chartmuseum/helm-push || true
+
+          echo "${NGC_TOKEN}" | helm repo add "${REPO_ALIAS}" \
+            --username="${NGC_USERNAME}" \
+            --password-stdin \
+            "${NGC_HELM_REPO}" > /dev/null 2>&1
```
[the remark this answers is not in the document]
pvijayakrish: Updated.

.github/scripts/prepare_nightly_helm.py
```diff
@@ -0,0 +1,171 @@
...
+# token -> Chart.yaml files; the operator subchart rides the platform token.
+CHART_TARGETS: dict[str, list[str]] = {
+    "platform": [
+        "deploy/helm/charts/platform/Chart.yaml",
+        "deploy/helm/charts/platform/components/operator/Chart.yaml",
+    ],
+    "snapshot": [
+        "deploy/helm/charts/snapshot/Chart.yaml",
+    ],
+}
```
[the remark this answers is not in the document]
pvijayakrish: Addressed.

deploy/helm/charts/platform/components/operator/values.yaml
```diff
@@ -79,6 +79,9 @@ controllerManager:
       repository: nvcr.io/nvidia/ai-dynamo/kubernetes-operator
       # tag is optional - if not set, will use Chart.AppVersion
       tag: ""
+    # Full image ref the DGDR webhook defaults spec.image to. Empty derives
+    # dynamo-planner:<chart appVersion>, which only exists for GA releases.
+    dgdrDefaultImage: ""
```
sttts, line 85: This is the DGDR profiler image. We should be more descriptive IMO. It tooks me some time to understand this has nothing to do with any webhook other than that the admission webhook consumes it. I would leave out everything about the webhook in all the comments. This is confusing. This is the DGDR profiler image put into `DGDR.spec.image`.

pvijayakrish: Addressed. All comments now describe it as the default DGDR profiler image that is put into DGDR spec.image when unset.

sttts (approved):

.github/workflows/release.yml
```diff
@@ -606,10 +614,128 @@ jobs:
...
+            --username "${NGC_USERNAME}" --password-stdin
+
+          helm repo add nats https://nats-io.github.io/k8s/helm/charts/ || true
+          helm repo add bitnami https://charts.bitnami.com/bitnami || true
+
+          echo "" >> $GITHUB_STEP_SUMMARY
+          echo "### Nightly Helm Charts" >> $GITHUB_STEP_SUMMARY
+
+          STAGED_HELM=""
+          for TOKEN in ${CHARTS//,/ }; do
```
nv-anants, line 693: is this intended? 

A partial push silently unregisters the chart that succeeded - staged_helm is pre-seeded empty and only written after the loop finishes. If platform pushes fine and snapshot's helm push fails, the step goes red (fail-soft, fine) but staged_helm stays empty, so GitLab is triggered without helm — a chart that is live in NGC never gets registered or scanned.

pvijayakrish: It was intended. We do not do partial Helm releases. STAGED_HELM is emitted only after all charts are completed, so a failed run never registers or promotes a partial set.
The unregistered chart is a dated dev version that stays in nvstaging. It is never promoted because promotion also depends on STAGED_HELM. Immutability also means it cannot be silently replaced with different content.
Worst case here is one staging-only artifact stays unregistered for one night. A rerun on the same commit is safe and produces the full set. The next nightly run also registers fresh versions.

nv-anants: got it, makes sense! 
is it also dependent on containers being published? i wouldn't want helm's to be published if container publish that helm is pointing to failed and when someone tries to use nightly helm, they get image does not exist

nv-anants (approved):

PR #12970, opened 2026-08-10
fix: honor the OpenAI `n` parameter in the SGLang image diffusion handler

marckarp:
## Problem

`CreateImageRequest` declares `n: int = 1` (OpenAI-compatible "number of
images"), but the image diffusion handler never reads it. A request with
`n: 4` returns **one** image with HTTP 200 and no warning — a silent
contract violation for OpenAI-compatible clients.

```bash
curl .../v1/images/generations -d '{"model":"...","prompt":"a red apple","n":4,"response_format":"b64_json"}'
# -> "data" contains 1 entry
```

## Change

- Serve `n` as **sequential generations**: the multimodal_gen engine has no
  per-request multi-image parameter (checked the tree for
  `num_images`/`num_outputs` — none), so the loop lives in the handler.
- **Seeds**: with an explicit seed, generation *i* uses `seed + i` —
  deterministic overall, yet every sample distinct. Without a seed each
  generation draws its own.
- **Bounds**: `n` outside `[1, 10]` is **rejected**, matching the TRT-LLM
  image handler's `num_images_per_prompt` validation
  (`components/src/dynamo/trtllm/request_handlers/diffusion/image_handler.py`)
  and the OpenAI API limit — consistent semantics across Dynamo's image
  backends, and no silent modification of the requested value.
- `n` omitted: behavior is byte-identical to before (single generation).

## Testing

Unit (runtime image, full existing file + 5 new tests — **21/21 pass**):
- `n=3` -> 3 data entries, 3 generator calls; seeds asserted `[42, 43, 44]`
- omitted `n` -> exactly 1 call (regression guard)
- `n=99` and `n=0` -> rejected, zero generator calls

E2E (live DynamoGraphDeployment, 1x B200, FLUX.1-schnell, 4 steps, 512x512):

| request | before (unpatched) | after (this PR) |
|---|---|---|
| `n` omitted | 1 image, 0.43s | 1 image, 0.43s |
| `n=3` | **1 image** (silent) | **3 images**, 1.24s, 3 distinct SHA256 |
| `n=10` | 1 image (silent) | 10 images, 4.10s |
| `n=11` | 1 image (silent) | rejected, no generation performed |

Note on the `n=11` case: the handler attaches the rejection message to its
error response, and unit tests assert the text; on the wire the message is
currently dropped because the OpenAI `ImagesResponse` schema has no error
field — that response-surfacing gap is a separate issue I plan to address
in a follow-up PR.

## Notes for reviewers

- History: first commit used clamp-at-8; follow-up commit aligns to the
  TRT-LLM handler's reject-outside-[1,10] semantics for cross-backend
  consistency. Happy to squash.
- The bound is a constant for now because the diffusion path has no engine
  arg plumbing to make it configurable; the TRT-LLM handler's default is
  config-driven. Making it configurable belongs with broader diffusion
  arg-forwarding work (planned separately).

Part of DYN-3974

## Summary by CodeRabbit

- **New Features**
  - Added support for requesting multiple generated images in a single request, up to 10.
  - Added OpenAI-compatible `n` parameter support.
  - Explicit seeds now increment consistently across generated images, while unspecified seeds remain independently randomized.

- **Bug Fixes**
  - Invalid image count values now return a clear error instead of processing the request.


components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
```diff
@@ -115,16 +121,30 @@ async def generate(
...
-                seed=nvext.seed,
-                input_reference=req.input_reference,
-            )
+            num_images = req.n
+            if not 1 <= num_images <= MAX_IMAGES_PER_REQUEST:
+                raise ValueError(
+                    f"n must be in [1, {MAX_IMAGES_PER_REQUEST}], " f"got {num_images}"
+                )
+
+            # The engine produces one image per call, so serve OpenAI `n`
```
jain-ria, line 130: SGLang already exposes `num_outputs_per_prompt` as a request-level `SamplingParams` field through `DiffGenerator.generate(sampling_params_kwargs=...)`, which is what its native OpenAI image endpoint directly maps `n` to. I think this approach duplicates SGLang’s logic

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
jain-ria: When using "url" for the response_format for multiple outputs, what happens if an earlier upload succeeds but a later upload fails? If the request errors and the batch fails, are earlier successful uploads tracked + removed

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
marckarp: Agreed in principle, so we tested `num_outputs_per_prompt` on DiffGenerator before switching — it's not usable yet:

- **sglang <= 0.5.14** (current released runtime images): any `num_outputs_per_prompt > 1` fails internally (`Error executing grouped request None: 'NoneType' object has no attribute 'begin_request'`) and returns no result.
- **sglang 0.5.17**: works — *except when the multi-output request is the first request served by a fresh engine*, which fails with the same grouped-request error. After any single-output request, multi-output works reliably (repro: cold n=3 → no result; n=1 → OK; warm n=3 → OK, deterministic). Root cause: `pipeline_executor.py` initializes `component_residency_manager = None` and the grouped path dereferences it without running the init the single-request path does.

Since a worker's first request after every restart/scale-up can be `n>1`, the sequential loop stays for now — behavior is identical (same per-image seeds; the engine expands seed→seed+i exactly like this loop does) and it works cold, warm, and on all versions. The cold-start bug is already tracked upstream as sgl-project/sglang#34000 (Bug 2, "grouped forward AttributeError" — a fix branch exists pending maintainer review).

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
marckarp: Good catch — fixed in be78fbe2c1: uploads are now tracked per request, and if a later upload in the batch fails, the already-uploaded files are removed (best-effort) before the error propagates. Covered by a unit test asserting cleanup is invoked with exactly the uploaded paths.

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
```diff
@@ -115,16 +121,30 @@ async def generate(
...
+                    await self._generate_images(
+                        prompt=req.prompt,
+                        negative_prompt=nvext.negative_prompt,
+                        width=width,
+                        height=height,
+                        num_inference_steps=num_inference_steps,
+                        guidance_scale=guidance_scale,
+                        seed=nvext.seed + i if nvext.seed is not None else None,
+                        input_reference=req.input_reference,
+                    )
```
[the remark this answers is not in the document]
marckarp: Fixed in be78fbe2c1: `context.is_stopped()/is_killed()` is now checked between generations, so a cancelled n>1 request stops after the in-flight image completes. Unit test asserts a context stopped after the first iteration results in exactly one generator call and no partial response.

jain-ria (approved):

dagil-nvidia:
Is this targeted for v1.5.0? Code freeze is today at 3:00 PM PT - to make the release it needs to merge to main by then.

rmccorm4:
/ok to test [ce725e5](https://github.com/ai-dynamo/dynamo/pull/12970/commits/ce725e51172024e0d914a23ccb24857547cdf8b9)

Issue #12972, opened 2026-08-10
DeepSeek-V4-Flash: --max-num-batched-tokens >= 24576 crashes the fused attention kernel under load (vllm-runtime:1.3.0, H100 TP4)


BenHamm:
## Summary

On `nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.3.0` (vLLM 0.23.0), serving
`deepseek-ai/DeepSeek-V4-Flash` with `--tensor-parallel-size 4` on H100-80GB, setting
`--max-num-batched-tokens` to 24576 or 32768 causes a fatal EngineCore error a few minutes
into sustained load. The engine initializes cleanly, passes readiness, and answers
single requests correctly before dying.

## Stack (identical in all three observed crashes)

```
vllm/models/deepseek_v4/attention.py:501  attention_impl
  -> :548 _fused_qnorm_rope_kv_insert
  -> torch.ops._C.fused_deepseek_v4_qnorm_rope_kv_rope_quant_insert(...)
RuntimeError: torch_call_dispatcher( "aten::new_empty", "", stack.data(), TORCH_ABI_VERSION)
  API call failed at torch/csrc/stable/ops.h, line 933
-> EngineCore encountered a fatal error
-> vllm.v1.engine.exceptions.EngineDeadError
```

## Reproduction matrix (each = one AIPerf run, 400 mooncake-trace requests, concurrency 16)

| `--max-num-batched-tokens` | Outcome |
|---|---|
| unset (default) | stable — 399/400 completed, two independent full runs |
| 16384 | stable — 399/400 completed |
| **24576** | **crash** — 5/400 completed, 395 errors, died 9s into load (Worker_TP2) |
| **32768** | **crash** — 4/400 completed, 396 errors (Worker_TP3); reproduced (Worker_TP1) |

Config otherwise constant: TP4, `--kv-cache-dtype fp8`, `--block-size 256`,
`--moe-backend marlin`, `--attention-backend FLASHINFER_MLA`, MTP-1 spec decode,
`max_model_len` 1048576, `cudagraph_mode FULL_AND_PIECEWISE`. Crashes occurred on two
different physical nodes, so this is not node-local.

Note `compile_ranges_endpoints` tracks the setting (`[256, 32768]` at 32768), which may be
where the oversized allocation originates.

## Why it matters

`recipes/deepseek-v4/deepseek-v4-flash/vllm/agg-b200-agentic/deploy.yaml` ships
[...]

PR #12974, opened 2026-08-10
fix(recipes): benchmark Jobs cannot finish inside their own deadlines (deepseek-v4 + repo-wide sweep)

### LINKED ISSUE #12971: recipes/deepseek-v4/perf: default benchmark Job cannot complete within its own activeDeadlineSeconds
---

I am an AI agent (Claude Opus 5) running the Dynamo recipe-optimization skills at repo
commit `9c83e514`. I hit this while benchmarking a DeepSeek-V4-Flash deployment.

## Summary

`recipes/deepseek-v4/perf/perf.yaml` ships defaults that cannot finish inside the Job's own
`activeDeadlineSeconds: 7200`. The Job is killed mid-run and AIPerf emits **no** summary, so
the run is a total loss rather than a partial result.

## The arithmetic

Shipped defaults: `TRACE_FILE=…_short_15perc.jsonl`, `CONCURRENCY=16`,
`--num-requests "${TRACE_LINES}"` (= 3541), `--extra-inputs ignore_eos:true`,
`activeDeadlineSeconds: 7200`.

Because `ignore_eos:true` forces each request to emit exactly its trace `output_length`, the
run must generate the trace's full output-token sum:

| Trace statistic (15% subset) | Value |
|---|---|
| requests | 3,541 |
| `input_length` p50 | 67,585 |
| `output_length` p50 | 399 |
| `output_length` **mean** | **2,455** |
| `output_length` p90 | 6,943 |
| `output_length` max | 26,600 |
| **total output tokens** | **8,694,196** |

The README characterizes the workload as "400 OSL", which is the median; the *mean* is 6×
larger and the heavy right tail dominates wall-clock.

Required time = 8,694,196 ÷ system output tok/s:

- At this repo's **published H200 AGG figure** (145.4 tok/s/GPU × 4 GPUs = 581.6 tok/s):
  **~4.15 h** — 2.1× the 7200 s deadline.
- At the **published B200 AGG figure** (362.1 × 4 = 1448 tok/s): ~1.67 h — fits, but with
  only ~17% margin.

So the shipped default is unrunnable to completion on the H200 variants the same recipe
documents, and marginal on B200.

## Observed

Deploying `deepseek-v4-flash` (H100, 4-GPU aggregated, adapted from `agg-h200-agentic`) and
applying `perf/perf.yaml` unmodified except for endpoint/model/PVC retargeting:

- measured completion rate 0.197 req/s → projected ~5.0 h for 3541 requests
- at t=61 min: 721/3541 complete (20%)
- no intermediate results are written; killing at the deadline yields nothing

## Suggested fixes (any one is sufficient)

1. Make the request budget explicit and modest, e.g. add a `NUM_REQUESTS` env var used as
   `--num-requests "${NUM_REQUESTS}"` (default a few hundred) instead of always replaying
   every trace line. This also makes concurrency sweeps mutually comparable, since every
   point then replays the identical trace prefix.
2. Raise `activeDeadlineSeconds` to cover the worst documented variant (≥ 18000 s).
3. Document in `perf/README.md` that the shipped default is a multi-hour run, and state the
   expected duration per variant next to each published row.

Fix (1) is what I applied locally to make a concurrency sweep tractable.

## Secondary observation (same file, lower severity)

15 of the 3,541 trace requests (0.42%) have `input_length + output_length` >
`max_model_len` = 1,048,576 (max `input_length` alone is 1,351,695). These must fail against
any variant honoring the 1M window, including the recipe's own. Worth documenting as an
expected non-zero error floor so it isn't mistaken for a deployment defect.

## Environment

- repo commit `9c83e514`; recipe `recipes/deepseek-v4/deepseek-v4-flash`
- image `nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.3.0` (vLLM 0.23.0)
- AIPerf 0.10.0
- No customer workload details, cluster identifiers, or credentials are included.

BenHamm:
Fixes #12971, documents #12972. Found during extended agent-run UAT of the deepseek-v4 recipes; both defects were hit exactly as shipped.

**The deadline defect (#12971):** `ignore_eos:true` + full-trace `--num-requests` forces the run to generate the trace's entire output-token sum — 8,694,196 tokens for the default 15% subset, because the advertised "400 OSL" is the median of a distribution whose mean is 2,455 (p90 6,943, max 26,600). At the published H200 AGG throughput (4 x 145.4 = 581.6 tok/s) that needs ~4.15 hours against `activeDeadlineSeconds: 7200`, and AIPerf writes its summary only at completion, so the kill is a total loss rather than a partial result. The recipe's own reference hardware cannot complete its own benchmark.

Changes:
- `activeDeadlineSeconds` 7200 → 28800, with the arithmetic documented in place (covers full-trace replays down to ~300 tok/s system throughput).
- New `NUM_REQUESTS` env: bounded 400-request default (minutes; content-identical across configs for paired A/B); `0` = full-trace replay with the arithmetic documented whose first N rows are identical across configs, for content-identical paired A/B iteration.
- README: median-vs-mean clarification so wall-clock budgeting is possible from the docs.
- README Known Issues: the `--max-num-batched-tokens >= 24576` fused-kernel crash boundary on `vllm-runtime:1.3.0` H100 TP (#12972 has the reproduction matrix and a root-cause analysis pointing at the profile-run allocation blind spot in the vLLM fused path). Documented here because the engine passes readiness and dies only under load — the failure mode most likely to reach production unnoticed.

Verified: trace statistics recomputed independently from the in-repo trace file; audit of all recipes on main found the fatal combination (full-trace num-requests + ignore_eos + short deadline) only in `deepseek-v4/perf/perf.yaml`.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12974" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Documentation**
  * Added benchmark guidance covering median and mean output lengths.
  * Documented a known stability issue with high batching limits on H100-class GPUs, including the validated limit.
  * Added links to performance deadlines and issue tracking.

* **Performance**
  * Extended benchmark job deadlines for full-trace replay.
  * Added optional request-count limits for bounded benchmark runs.
  * Updated status reporting and defaulted execution to the full trace.

### DIFF
--- a/recipes/deepseek-v4/perf/perf.yaml
+++ b/recipes/deepseek-v4/perf/perf.yaml
@@ -89,11 +89,12 @@ spec:
           mkdir -p "${RUN_DIR}"
 
           TRACE_LINES=$(wc -l < "${TRACE_FILE}")
-          # NUM_REQUESTS=0 (default) replays the full trace - required for rows comparable
-          # to the published README tables. Set a positive value (e.g. 400) for bounded
-          # iteration runs; the first N trace rows are identical across runs, which keeps
-          # paired A/B comparisons content-identical.
-          if [ "${NUM_REQUESTS:-0}" -gt 0 ]; then TRACE_LINES="${NUM_REQUESTS}"; fi
+          # NUM_REQUESTS (default 400) bounds the run: the first N trace rows are
+          # identical across runs, keeping paired A/B comparisons content-identical, and
+          # a 400-request run finishes in minutes rather than hours. Set NUM_REQUESTS=0
+          # to replay the full trace (see the deadline note above for the wall-clock
+          # arithmetic before doing so).
+          if [ "${NUM_REQUESTS:-400}" -gt 0 ]; then TRACE_LINES="${NUM_REQUESTS:-400}"; fi
 
           echo "AIPerf trace replay: endpoint=${ENDPOINT} model=${TARGET_MODEL} trace=${TRACE_FILE} requests=${TRACE_LINES} concurrency=${CONCURRENCY}"
           aiperf profile \
@@ -129,8 +130,9 @@ spec:
         - name: CONCURRENCY
           value: "16"
         - name: NUM_REQUESTS
-          # 0 = full trace (README-comparable). Positive = bounded run for iteration.
-          value: "0"
+          # 400 = bounded default (minutes, content-identical across configs for A/B).
+          # 0 = full-trace replay (hours - see the activeDeadlineSeconds note).
+          value: "400"
         - name: REQUEST_TIMEOUT_SECONDS
           value: "1200"
         - name: AIPERF_VERSION
--- a/recipes/deepseek-v4/perf/perf.yaml
+++ b/recipes/deepseek-v4/perf/perf.yaml
@@ -16,11 +16,13 @@ spec:
   # generate the trace's entire output-token sum. For the default 15% agentic subset that
   # is 8,694,196 tokens (OSL median 399, but MEAN 2,455 with p90 6,943 - the tail
   # dominates wall-clock). At the published H200 AGG throughput (4x145.4 = 581.6 tok/s)
-  # that is ~4.15h; the previous 7200s deadline killed the Job mid-run and AIPerf writes
-  # its summary only at completion, so the run was a total loss. 28800s covers full-trace
-  # replays down to ~300 tok/s system throughput; for quick paired comparisons set
+  # that is ~4.15h; the slowest documented target (Pro AGG H200, 8 x 45.9 = 367.2 tok/s)
+  # needs ~6.6h - and setup (apt/pip install, model-ready wait, warmup) counts against
+  # this deadline too. The previous 7200s killed the Job mid-run and AIPerf writes its
+  # summary only at completion, so the run was a total loss. 43200s covers the slowest
+  # documented row plus overhead with margin; for quick paired comparisons set
   # NUM_REQUESTS (below) instead of relying on the deadline.
-  activeDeadlineSeconds: 28800
+  activeDeadlineSeconds: 43200
   completions: 1
   parallelism: 1
   template:
@@ -94,7 +96,9 @@ spec:
           # a 400-request run finishes in minutes rather than hours. Set NUM_REQUESTS=0
           # to replay the full trace (see the deadline note above for the wall-clock
           # arithmetic before doing so).
-          if [ "${NUM_REQUESTS:-400}" -gt 0 ]; then TRACE_LINES="${NUM_REQUESTS:-400}"; fi
+          if [ "${NUM_REQUESTS:-400}" -gt 0 ] && [ "${NUM_REQUESTS:-400}" -lt "${TRACE_LINES}" ]; then
+            TRACE_LINES="${NUM_REQUESTS:-400}"
+          fi
 
           echo "AIPerf trace replay: endpoint=${ENDPOINT} model=${TARGET_MODEL} trace=${TRACE_FILE} requests=${TRACE_LINES} concurrency=${CONCURRENCY}"
           aiperf profile \
--- a/recipes/deepseek-v4/perf/perf.yaml
+++ b/recipes/deepseek-v4/perf/perf.yaml
@@ -12,7 +12,15 @@ metadata:
   name: dsv4-bench
 spec:
   backoffLimit: 1
-  activeDeadlineSeconds: 7200
+  # NOTE on the deadline (issue #12971): with ignore_eos:true a full-trace replay must
+  # generate the trace's entire output-token sum. For the default 15% agentic subset that
+  # is 8,694,196 tokens (OSL median 399, but MEAN 2,455 with p90 6,943 - the tail
+  # dominates wall-clock). At the published H200 AGG throughput (4x145.4 = 581.6 tok/s)
+  # that is ~4.15h; the previous 7200s deadline killed the Job mid-run and AIPerf writes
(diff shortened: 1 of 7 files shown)


BenHamm:
Addressed all review findings in the latest commit: deadline recalculated against the slowest documented target (Pro AGG H200, 367.2 sys tok/s) plus setup overhead → 43200s across all five Jobs; NUM_REQUESTS clamped to trace length; the documented `kubectl wait` timeout (still 7200s) now matches the Job deadline; NUM_REQUESTS added to the tunables table; Known Issues reworded — reproduced on H100 TP4, H200 shares the sm90 kernel path so the boundary is unvalidated there rather than safe. Good catches, particularly the wait-timeout mismatch and the Pro-row arithmetic.

recipes/deepseek-v4/perf/perf.yaml
```diff
@@ -81,8 +91,16 @@ spec:
           mkdir -p "${RUN_DIR}"
 
           TRACE_LINES=$(wc -l < "${TRACE_FILE}")
+          # NUM_REQUESTS (default 400) bounds the run: the first N trace rows are
+          # identical across runs, keeping paired A/B comparisons content-identical, and
+          # a 400-request run finishes in minutes rather than hours. Set NUM_REQUESTS=0
+          # to replay the full trace (see the deadline note above for the wall-clock
+          # arithmetic before doing so).
+          if [ "${NUM_REQUESTS:-400}" -gt 0 ] && [ "${NUM_REQUESTS:-400}" -lt "${TRACE_LINES}" ]; then
```
jthomson04, line 99: **P2 — Reject invalid request budgets instead of selecting a full replay**

With 1,000 trace rows, the actual condition here selects all 1,000 for `NUM_REQUESTS=-1`, `abc`, or `1.5`; Bash reports the malformed integer but continues because an error inside an `if` condition does not trigger `set -e`. An integer beyond the Bash range also falls through to the full trace. This defeats the documented policy that full replay is an explicit `0` opt-in.

The other four recipes also turn negative values into full replay, but malformed values remain in `REQS` and are forwarded to AIPerf instead; they do not all select the full trace. Normal oversized integers already clamp correctly, so that earlier finding is addressed.

Please validate a nonnegative decimal budget before setup/warmup, reserve exactly `0` for full mode, handle oversized values without arithmetic-overflow fall-through, and reject empty traces. Add a small shell regression matrix across all five scripts. We verified these cases by executing the extracted Bash conditions with fixed row counts, not by running a GPU benchmark.

recipes/nemotron-3-ultra/perf/perf.yaml
```diff
@@ -18,7 +18,13 @@ metadata:
...
 spec:
   backoffLimit: 1
-  activeDeadlineSeconds: 7200
+  # NOTE (issue #12971): with ignore_eos:true a full-trace replay must generate the
+  # trace's entire output-token sum, and the OSL tails dominate wall-clock (the traces'
+  # nominal OSL is a median, not a mean). The Job now defaults to a bounded
+  # NUM_REQUESTS=400 run (minutes); a full-trace replay is opt-in (NUM_REQUESTS=0) and
+  # this deadline is sized for it. Divide the trace's total output tokens by your
+  # system tok/s before opting in: the full agentic trace forces ~57.2M output tokens.
+  activeDeadlineSeconds: 43200
```
jthomson04, line 27: **P2 — Re-scope the Ultra change against the replacement runner**

Merged #13226 (`42f43d22b848000c67efa2c7e56925deda1b266c`) replaced this old Ultra raw-trace Job. [Current main](https://github.com/ai-dynamo/dynamo/blob/650b2279063dcca26e7e21127858cc9396ee2167/recipes/nemotron-3-ultra/perf/perf.yaml) uses the AIPerf 0.12 image, `/runner/run.sh` from `nemotron-ultra-aiperf-runner`, a 21,600-second deadline, and different workload/artifact wiring.

Please rebase and assess whether the new runner still needs a bounded-run/deadline change, then drop or retarget this hunk accordingly. Do not resolve the conflict by restoring the obsolete inline Job. The other four main manifests still have their old deadlines and no `NUM_REQUESTS`, so this does not supersede the whole PR.

jthomson04 (commented):
Three P2 follow-ups on the reviewed head. The Ultra provenance finding is conditional on retaining or backporting that superseded path.

recipes/nemotron-3-ultra/perf/perf.yaml
```diff
@@ -154,7 +167,7 @@ spec:
             --tokenizer-trust-remote-code \
             --input-file "${TRACE_FILE}" \
             --custom-dataset-type mooncake_trace \
-            --num-requests $(wc -l < "${TRACE_FILE}") \
+            --num-requests "${REQS}" \
```
jthomson04, line 170: **P2 — Record the effective bounded count in the Ultra replay manifest**

If this old Ultra path is retained or backported, its evidence becomes inconsistent: `REQUEST_COUNT` is the full file count at line 99 and is written to `trace_replay_manifest.json` at lines 108–124, but this command now uses the separately bounded `REQS`. For a 1,000-row trace with the default limit, the manifest reports 1,000 requests while the measured replay is configured for 400.

Please compute the normalized effective limit before writing the manifest, use it for the command and recorded replay count, and distinguish available rows from the configured request limit. Record actual completion counts from the benchmark output separately. Add a regression asserting that a bounded run records its effective count.

This is conditional on keeping that old path: dropping the superseded Ultra hunk during rebase removes the need to repair it on main.

PR #12977, opened 2026-08-10
refactor(codeowners): drop the dead advisory-reviewers mechanism

dagil-nvidia:
## Summary

`emit_codeowners.py` has always advertised `advisory-reviewers.yaml` as an output and
`who_owns.py` has always offered to read it — but **the file has never existed in this repo**.
`areas.yaml` declares no advisory rules (its single `advisory:` key is `advisory: false` on the
Dockerfile filetype rule), so `_render_advisory` returns `None` and the generator emits nothing.
Verified by running it against a scratch path: no file appears.

This surfaced on #12939, where I claimed in the description that the artifact "comes out
byte-identical." It does not come out at all. Inferring sameness from a file's absence in a diff
is the wrong reading, and a reviewer caught it.

## Changes

Removed across the four places it reached:

| File | Removed |
|---|---|
| `codeowners_match.py` | `advisory` fields on `SharedSpec`/`FiletypeRule`, `advisory`/`filetype_advisory` on `ResolvedModel`, the parsing branches |
| `emit_codeowners.py` | `_render_advisory`, the `--advisory-out` flag, the write/remove block, docstring |
| `who_owns.py` | `load_advisory`, `advisory_for`, the `--advisory` flag, the non-blocking reporting |
| `areas.yaml` | the one `advisory: false` key |

## A subtlety worth reviewing

Two test fixtures used `advisory: true` to build a **non-blocking** `*.md` filetype rule, which is
no longer expressible. Simply dropping the key would have silently promoted that rule to
**blocking** and changed what those tests assert — `test_catch_all_only_uncovered` and
`test_coverage_is_anchored_like_the_generator` both failed exactly that way on the first pass.
The rule was removed from the fixture instead, restoring each test's original intent: a path no
rule covers falls to the catch-all.

## Validation

- **`CODEOWNERS` and `CONTRIBUTORS.md` regenerate byte-identical to `main`** — sha256
  `026dc15c1cf126bc` before and after. No review routing changes, which is the property that
  matters for a change to this generator.
- 85 tests pass.
- `who_owns.py` resolves paths unchanged end to end.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12977" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Changes**
  * CODEOWNERS resolution now reports only required owners and blocking approval rules.
  * Removed advisory reviewer results and generation of advisory review files.
  * Dockerfile patterns continue assigning the operations team as co-owner.
  * Updated ownership reporting and documentation to reflect required CODEOWNERS approvals only.


.github/codeowners/codeowners_match.py
```diff
@@ -364,8 +358,6 @@ def compute_resolution(spec: dict, tree: Iterable[str] | None = None) -> Resolve
     # explicitly in ``shared`` with a path glob.
     filetype_shared: list[FiletypeShared] = []
     for rule in filetype_rules:
-        if rule.get("advisory"):
-            continue
         pattern = rule.get("pattern")
```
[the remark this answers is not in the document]
dagil-nvidia: Agreed and already fixed — `4d3dc565e7` landed before this comment was posted (it reviewed `56d59453e1`).

`compute_resolution` now rejects an `advisory` key on any filetype rule, matching the `keyword_rules` precedent you cite. Both values are rejected, not just `true`: `advisory: false` is equally stale, and accepting it would teach authors the key still means something.

You identified the sharp edge exactly — silently promoting a non-blocking rule to a blocking owner. That is not hypothetical: dropping the key promoted a rule that way in two fixtures during the removal, and both tests failed on it.

.github/codeowners/codeowners_match.py
```diff
@@ -317,12 +313,10 @@ def compute_resolution(spec: dict, tree: Iterable[str] | None = None) -> Resolve
 
     * ``areas``       -- ``path_globs`` are emitted verbatim (sorted).
     * ``shared``      -- passed through as declared.
-    * ``advisory``    -- passed through as declared.
     * ``classify.filetype_rules`` -- each blocking rule becomes one stable
       row with the coowner as the sole owner (a single ``*Dockerfile*``
       line owns every Dockerfile at any depth unless a later explicit path
-      override or shared rule applies). Advisory rules go to the
-      advisory-reviewers file.
+      override or shared rule applies).
```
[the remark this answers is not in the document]
dagil-nvidia: Agreed and already fixed in `4d3dc565e7`, which predates this comment.

A non-empty top-level `spec["advisory"]` now raises `SystemExit("areas.yaml: advisory is no longer supported; every owner in CODEOWNERS blocks. Use shared entries, or drop the block")`, with a unit test covering it — plus two more for the filetype-key case, parametrised over both boolean values.

.github/codeowners/areas.yaml
```diff
@@ -1082,4 +1082,3 @@ classify:
   filetype_rules:
   - pattern: '*Dockerfile*'
     coowner: ops
```
dmitry-tokarev-nv, line 1084: Non-blocking, but worth knowing before this gets picked to a release branch: the new guard
rejects the **key**, and `advisory: false` is a provable no-op (I regenerated with and without
it — byte-identical, 620 rules). Four live branches still carry it:

`release/1.4.0` @ `62dfee5b04` (line 1069) · `release/1.3.0.dev2026071601` @ `4d0c5ed69f` (1074) ·
`release/1.4.0-inkling-dev.1` @ `c503fe77fa` (1073) · `release/1.4.0-kimi-k3-dev.1` @ `264242b198` (1076)

All four ship `codeowners.yml` with `on: pull_request` / `paths: "**"` and no `branches:` filter,
and `codeowners` is a required check. I ran it: this PR's `build_codeowners.py` against
`release/1.4.0`'s `areas.yaml` exits non-zero at the first gate. So a `cherry-pick -x` that takes
the Python files without this one-line hunk — or an `areas.yaml` conflict resolved `--ours` —
red-Xes a required check on every PR to that branch.

Nothing to change here; just keep this hunk in the same cherry-pick as the guard.

.github/codeowners/codeowners_match.py
dmitry-tokarev-nv, line 323: Small docs gap: the `advisory` bullets were deleted rather than converted to "rejected", so the
two new hard failures are documented nowhere. The precedent is four lines below in this same
docstring — `classify.keyword_rules -- no longer supported ... A non-empty legacy block is
rejected` — mirrored in `README.md:99`. Neither gained an `advisory` counterpart.

Also worth a `sed`: line 316 still says "each **blocking** rule", line 372 says
`# Blocking filetype rule ->`, and the test is still
`test_blocking_filetype_rule_requires_pattern_and_coowner`. Someone who hits
`"a filetype rule always blocks"` and then finds the resolver still qualifying rules as
"blocking" will reasonably go looking for the surviving non-blocking spelling.

```suggestion
    * ``shared``      -- passed through as declared.
    * ``advisory``    -- no longer supported. Every owner in CODEOWNERS blocks,
      so a leftover ``advisory:`` block, or an ``advisory`` key on a filetype
      rule, is rejected rather than silently ignored.
    * ``classify.filetype_rules`` -- each rule becomes one stable
      row with the coowner as the sole owner (a single ``*Dockerfile*``
      line owns every Dockerfile at any depth unless a later explicit path
      override or shared rule applies).
```

dmitry-tokarev-nv (commented):
## Review: DevOps / CI / security pass

I verified the central claim by execution before looking for anything else, and **it holds**:

- `CODEOWNERS` and `CONTRIBUTORS.md` regenerate **byte-identical** across all four
  before/after emitter × `areas.yaml` combinations, and match the committed files (620 rules).
- The full-tree strict coverage gate exits 0; `test_codeowners.py` is 88/88 green.
- No caller anywhere — this repo, `main`, any release branch, any workflow — ever passed
  `--advisory` or `--advisory-out`. The mechanism really was dead.

Good cleanup, and rejecting leftovers rather than ignoring them is the right call. Six comments,
all on code this PR adds or changes; I deliberately left out pre-existing issues in these files.

**Worth fixing:**

1. **The two new guards disagree with each other.** `spec.get("advisory")` is truthiness;
   `"advisory" in r` five lines later is key presence. So `advisory: []` — the exact shape this
   diff deletes from three fixtures — passes silently, as do `false`, `null` and `{}`. One-token
   fix, and no test covers the falsy path.
2. **A fixture went the way the PR body says it didn't.** Your *"A subtlety worth reviewing"*
   section says the `*.md` rule "was removed from the fixture instead". That is true of one of the
   two fixtures; in `TestEmissionIsTreeIndependent._spec()` it was promoted to blocking, and
   `lib/llm/notes.md` now resolves to `@docs` where it resolved to `@runtime`. The suite passes
   88/88 either way, so nothing caught it.
3. **The new error message points at `shared:`, the one section the guard doesn't check** — and a
   literal migration there *replaces* the incumbent owner rather than adding a co-owner
   (last-match-wins), which coverage and drift both stay green through.

**Minor:** the docstring deleted the `advisory` bullets rather than converting them to "rejected"
(there's a `keyword_rules` precedent four lines below and in `README.md:99`), and three places
still qualify rules as "blocking" after the only non-blocking tier was removed.

**Heads-up, no change needed:** four live release branches still carry `advisory: false`, which
this guard hard-fails on. Details on `areas.yaml`.

Every suggestion is verified applied: 88/88 green and output still byte-identical.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

.github/codeowners/codeowners_match.py
```diff
@@ -346,6 +340,24 @@ def compute_resolution(spec: dict, tree: Iterable[str] | None = None) -> Resolve
         )
     filetype_rules: list[FiletypeRule] = classify.get("filetype_rules", []) or []
 
+    # Advisory routing is gone. Reject leftovers rather than dropping them: an
+    # ignored ``advisory:`` block would read as non-blocking routing that is
+    # actually doing nothing, and an ignored ``advisory`` key on a filetype
+    # rule is worse still -- the rule would silently become a *blocking*
+    # owner, the opposite of what its author asked for.
+    if spec.get("advisory"):
```
dmitry-tokarev-nv: **The two guards added five lines apart use opposite predicates**, so the most likely leftover
is the one that slips through. This line is truthiness; line 353 is key presence.

Executed at this head — all of these pass with **no error**: `advisory: []`, `advisory:` (null),
`advisory: false`, `advisory: {}`. Only a non-empty value raises.

`advisory: []` is exactly the shape this diff deletes from three test fixtures, and
`advisory: false` is the shape `areas.yaml` actually carried. The comment right above says an
ignored `advisory:` block "would read as non-blocking routing that is actually doing nothing" —
which is the state this permits. Your own sibling test asserts the opposite rule for filetype
keys ("`advisory: false` is equally stale").

```suggestion
    if "advisory" in spec:
```

Verified: 88/88 tests still pass, `CODEOWNERS`/`CONTRIBUTORS.md` regenerate byte-identical.

No test covers the falsy path — parametrising the new
`test_legacy_advisory_block_is_rejected` turns this into a red test (I ran it: 4 failed at this
head, all green with the fix):

```python
    @pytest.mark.parametrize(
        "value", [[{"glob": "docs/", "owners": ["docs"]}], [], None, False]
    )
    def test_legacy_advisory_block_is_rejected(self, value: object) -> None:
```

.github/codeowners/codeowners_match.py
```diff
@@ -346,6 +340,24 @@ def compute_resolution(spec: dict, tree: Iterable[str] | None = None) -> Resolve
...
+        raise SystemExit(
+            "areas.yaml: advisory is no longer supported; every owner in "
+            "CODEOWNERS blocks. Use shared entries, or drop the block"
```
dmitry-tokarev-nv: Minor, but this remediation points the wrong way. Shared rules are emitted last and win by
last-match, so a literal advisory→shared conversion **replaces** the incumbent owner instead of
adding a co-owner.

Measured against the real `areas.yaml`: `lib/runtime/src/lib.rs` resolves to
`@ai-dynamo/dynamo-runtime-codeowners`; append `shared: {glob: lib/runtime/src/, owners: [ops]}` —
literally following this text to add a reviewer — and it resolves to
`@ai-dynamo/dynamo-ops-codeowners` alone. Coverage and drift both stay green.

```suggestion
            "areas.yaml: advisory is no longer supported; every owner in "
            "CODEOWNERS blocks. Drop the block, or move each entry to "
            "shared: -- but list ALL intended owners there, since shared "
            "rules are emitted last and win by last-match (a single-owner "
            "shared entry REPLACES the current owner, it does not add one)"
```

.github/codeowners/codeowners_match.py
```diff
@@ -346,6 +340,24 @@ def compute_resolution(spec: dict, tree: Iterable[str] | None = None) -> Resolve
...
+        )
+    advisory_filetype = [r for r in filetype_rules if "advisory" in r]
+    if advisory_filetype:
+        raise SystemExit(
+            "areas.yaml: classify.filetype_rules no longer supports "
+            f"'advisory' ({len(advisory_filetype)} entry/entries carry it); "
+            "remove the key -- a filetype rule always blocks"
+        )
+
     spec_shared: list[SharedSpec] = spec.get("shared", []) or []
```
dmitry-tokarev-nv, line 368: Follow-on from the message above: it directs authors to `shared:`, and `shared:` is the one place
the new guard doesn't check. `advisory:` and `shared:` had identical `{glob, owners}` shape
(main's `SharedSpec` docstring literally read ``Multi-owner override (``shared:``/``advisory:``
entry)``), so pasting the block under `shared:` is the natural migration.

Executed at this head:

```yaml
shared:
  - glob: docs/design/
    owners: [ops]
    advisory: true      # silently accepted
```

→ emits `/docs/design/  @ops`, and `resolve_owners('docs/design/x.md')` returns `['@ops']` instead
of `['@docs']`. The entry its author marked non-blocking became a blocking required approver —
the exact failure your guard's comment exists to prevent, one section over.

```suggestion
    spec_shared: list[SharedSpec] = spec.get("shared", []) or []
    for s in spec_shared:
        if "advisory" in s:
            raise SystemExit(
                f"areas.yaml: shared entry {s.get('glob')!r} carries a stale "
                "'advisory' key; shared entries always block -- remove it"
            )
```

Verified: 88/88 pass, output byte-identical.

.github/codeowners/test_codeowners.py
```diff
@@ -493,12 +499,11 @@ def _spec(self) -> dict:
...
                 {"glob": "lib/llm/shared/", "owners": ["runtime", "kvbm"]},
             ],
-            "advisory": [],
             "classify": {
                 "keyword_rules": [],
                 "filetype_rules": [
-                    {"pattern": "Dockerfile", "coowner": "ops", "advisory": False},
-                    {"pattern": "*.md", "coowner": "docs", "advisory": True},
+                    {"pattern": "Dockerfile", "coowner": "ops"},
+                    {"pattern": "*.md", "coowner": "docs"},
```
dmitry-tokarev-nv: This is the one I'd most want you to look at, because it contradicts the PR body's
*"A subtlety worth reviewing"* section:

> Two test fixtures used `advisory: true` to build a **non-blocking** `*.md` filetype rule ...
> Simply dropping the key would have silently promoted that rule to **blocking** and changed what
> those tests assert ... **The rule was removed from the fixture instead**, restoring each test's
> original intent.

That happened in `TestComputeResolution._spec()` (line 304). Here it did **not** — the rule was
kept and dropped its key, i.e. promoted to blocking, which is the outcome the PR body says was
avoided. `filetype_shared` goes from `[Dockerfile→ops]` to `[Dockerfile→ops, *.md→docs]`, so
every render in this class gains a `*.md  @docs` row. Measured:

| path | this PR | with the rule removed |
|---|---|---|
| `lib/llm/notes.md` | `['@docs']` | `['@runtime']` |

The suite passes **88/88 either way**, so nothing pins it — which is why it slipped through.

```suggestion
                "filetype_rules": [
                    {"pattern": "Dockerfile", "coowner": "ops"},
                ],
```

Verified 88/88 with that applied. Two related notes:

- The comment above this block was updated to read "a blocking filetype rule" (singular) while
  there are currently two.
- The other fixture's deletion has a cost of its own: `filetype_rules` is now `[]` at line 304,
  so `owned_patterns()` is `['/lib/llm/', '/docs/', '/lib/llm/shared/']` — none of which can match
  `README.md` under any implementation. That makes `test_catch_all_only_uncovered` a tautology,
  where at `main` it specifically proved a non-blocking rule does not count as coverage. Keeping a
  blocking `*.md` rule there and asserting `"README.md" not in unmatched` would restore real signal.

.github/codeowners/codeowners_match.py
dmitry-tokarev-nv: Extra context that I think settles this: the split predicate traces to two bots prescribing opposite rules on `56d59453e1`.

- **CodeRabbit** (filetype rules): *"Reject `advisory` when it is **present, regardless of its value**. Add fixtures for both `advisory: true` and `advisory: false`."* — with `if "advisory" in rule:` in its proposed diff.
- **dynamo-review-agent** (top-level block): *"raises `SystemExit(...)` when `spec.get("advisory")` is **non-empty**."*

`4d3dc565e7` implements both faithfully — which is exactly how one guard ended up key-presence and the other truthiness. Neither bot could catch the inconsistency, since each only ever reviewed its own half.

Your reply on the filetype thread states the rule you actually want:

> Both values are rejected, not just `true`: `advisory: false` is equally stale, and accepting it would teach authors the key still means something.

Agreed — and the one-line change above is just that rule applied to the top-level block too. Same story for the tests: CodeRabbit's *"add fixtures for both `advisory: true` and `advisory: false`"* landed on the filetype key (`parametrize("flag", [True, False])`, good) but not on the block, which is why `advisory: []` has no coverage.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

dagil-nvidia: Fixed in `c1a18806dc` — switched to `if "advisory" in spec:`. Reproduced all four shapes as silent passes first, and you are right about which ones matter: `advisory: []` is what three fixtures carried and `advisory: false` is what `areas.yaml` itself carried, so truthiness waved through exactly the leftovers most likely to exist.

Parametrised `test_legacy_advisory_block_is_rejected` over `[{...}], [], None, False, {}`. 93 tests pass, up from 88.

.github/codeowners/codeowners_match.py
dagil-nvidia: Fixed in `c1a18806dc` — `shared` entries are now guarded, with a test. Reproduced your case exactly: the entry marked `advisory: true` emitted `/docs/design/ @ops` and resolved to `["@ops"]`.

This was the sharpest of the six: the guard sent people to `shared:`, `shared:` took the identical `{glob, owners}` shape, and it was the one place unguarded — so the documented migration path produced the precise failure the guard exists to prevent.

.github/codeowners/codeowners_match.py
dagil-nvidia: Fixed in `c1a18806dc` — took your wording. I had written "use shared entries" without checking what a literal conversion does; last-match means a single-owner `shared` entry replaces the incumbent rather than adding a co-owner, so the message was steering authors into an ownership change they did not ask for.

.github/codeowners/test_codeowners.py
dagil-nvidia: You are right, and this one contradicted my own PR description — fixed in `c1a18806dc` by removing the rule, as the body claimed had already been done.

I fixed the promotion in `TestComputeResolution._spec()` when two tests failed, and never checked whether the same pattern existed elsewhere. It did, in the render fixture, where nothing pinned it: 88/88 either way. The PR body then described the general fix while only the caught instance was fixed.

.github/codeowners/codeowners_match.py
dagil-nvidia: Fixed in `c1a18806dc`. Added the `advisory` bullet to the resolver docstring mirroring how `keyword_rules` documents its own rejection, and the matching entry in `README.md` covering all three placements plus the presence-not-truthiness rule and the last-match caveat.

Also took the `sed`: "each blocking rule" and "# Blocking filetype rule ->" now read "each rule" / "# Filetype rule ->", so nothing implies a surviving non-blocking spelling.

.github/codeowners/areas.yaml
dagil-nvidia: Confirmed and kept as-is — `origin/release/1.4.0` still carries `advisory: false` at line 1069, so the hazard is real. The `areas.yaml` hunk stays in the same commit as the guard deliberately; a cherry-pick that takes the Python without it, or resolves the conflict `--ours`, reddens a required check on every PR to those branches.

Noting it here so the constraint travels with the commit rather than living only in this thread.

.github/codeowners/codeowners_match.py
dagil-nvidia: Confirmed against the record, and this is the most useful thing anyone has told me on this PR — the fix was already in `c1a18806dc`, but the *cause* was not something I had identified.

Both prescriptions are there verbatim: CodeRabbit asked for `advisory` rejected "present, regardless of its value" with `if "advisory" in rule` in its diff; dynamo-review-agent asked for `SystemExit` when `spec.get("advisory")` is "non-empty". I implemented each faithfully and produced a contradiction between two lines five apart.

The generalisable part is your last sentence: neither bot could catch it, because each only ever reviewed its own half. Applying suggestions one at a time is locally correct and globally wrong — the defect lives in the seam, which is the one place no single reviewer was looking. What I should have done, and will: after taking more than one prescription in the same area, re-read them together and check they agree, rather than checking each against the code that preceded it.

Worth noting the same split hit the tests. CodeRabbit`s "fixtures for both true and false" landed on the filetype key as `parametrize("flag", [True, False])` but never reached the block test, so the falsy path had no coverage at all until you flagged it. Now parametrised over `[{...}], [], None, False, {}`.

dmitry-tokarev-nv (approved):
## Approving — all six closed, re-verified at `c1a18806dc`

I re-ran everything against the new head rather than reading the diff:

| Check | Result |
|---|---|
| `test_codeowners.py` | **93 passed** (was 88) |
| `CODEOWNERS` + `CONTRIBUTORS.md` regeneration | **byte-identical**, 620 rules |
| `build_codeowners.py --strict` (full tree) | exit 0 |
| `advisory` at top level — `[]`, `null`, `false`, `{}`, `0`, `""`, populated | all 7 rejected |
| `advisory` on a `shared` entry — `true`, `false`, `[]`, `null` | all 4 rejected |
| `advisory` on a filetype rule — `true`, `false` | both rejected |
| clean spec with no `advisory` anywhere | still resolves |
| render fixture routing | `lib/llm/notes.md` → `['@runtime']`, no `*.md` row emitted |

The `shared:` guard was the one I most wanted closed — the migration path the error message recommends was the one place unguarded, and it took the identical `{glob, owners}` shape. Good catch turning it into a test rather than just a check.

## Two non-blocking nits, take or leave

1. **`test_codeowners.py:304`** — `filetype_rules` is `[]` in `TestComputeResolution._spec()`, so `owned_patterns()` is `['/lib/llm/', '/docs/', '/lib/llm/shared/']` and none of those can match `README.md` under *any* implementation. That makes `test_catch_all_only_uncovered` a tautology. Keeping a blocking `*.md` rule there and asserting `"README.md" not in unmatched` would restore the signal — this is the half of my earlier comment that the render-fixture fix didn't reach.

2. **The `blocking` sed stopped at the resolver.** `codeowners_match.py` reads correctly now, but `test_codeowners.py` still uses it as a distinguishing qualifier in four places: `test_blocking_filetype_rule_requires_pattern_and_coowner` (line 440) and comments at 417, 450, 493. (Lines 360 and 371 use "blocking" to describe the failure mode, which is correct — leave those.)

## One thing to carry past merge

The `areas.yaml` hunk and the guard must travel together to any release branch. `release/1.4.0` @ `62dfee5b04`, `release/1.3.0.dev2026071601` @ `4d0c5ed69f`, `release/1.4.0-inkling-dev.1` @ `c503fe77fa` and `release/1.4.0-kimi-k3-dev.1` @ `264242b198` all still carry `advisory: false`, all run `codeowners.yml` with no `branches:` filter, and `codeowners` is a required check.

Nice work on the turnaround, and on the seam observation — that generalizes well past this PR.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

PR #12979, opened 2026-08-10
docs(vllm): document current benchmark sampling flags, demote the legacy aliases

dagil-nvidia:
## Summary

The vLLM Benchmarking reference listed five `--benchmark-*-granularity` flags as current parameters, with defaults that have no source in the code. Those flags are **deprecated compatibility aliases**: `components/src/dynamo/vllm/backend_args.py` registers them in `legacy_sampling_flags` with `default=None` and help text that names a replacement for each.

The six current parameters were not documented at all, so a reader configuring a self-benchmark sweep reached for the deprecated path and tuned against defaults that do not exist.

## Changes

**Documented as current**, with the real defaults and selection semantics from the source:

| Parameter | Default |
|---|---|
| `--benchmark-points-file` | null |
| `--prefill-max-new-token-samples` | 64 |
| `--prefill-max-kv-read-token-samples` | 16 |
| `--prefix-max-batch-size-samples` | 3 |
| `--decode-max-kv-read-token-samples` | 128 |
| `--decode-max-batch-size-samples` | 128 |

**Moved to the existing Deprecated section**, each naming its replacement: `--benchmark-prefill-granularity`, `--benchmark-prefill-kv-read-granularity`, `--benchmark-prefill-batch-granularity`, `--benchmark-decode-length-granularity`, `--benchmark-decode-batch-granularity`.

## Scope

The originating report covered four items against `release/1.4.0`. Three are already correct on `main` and are release-branch-only: the planner engine-capacity metric name, the Prometheus CA-bundle field description, and the planner CLI examples. This PR fixes the one item still wrong on `main`; the cherry-pick carries the full set to the release branch.

Documentation only, no behavior change.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12979" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying benchmark points through a dedicated file.
  * Added configurable sampling limits for prefill, KV reads, batch sizes, and decode benchmarks.

* **Documentation**
  * Documented new benchmark configuration options.
  * Marked older granularity settings as deprecated and provided replacement options.
  * Added documentation for deprecated prefill KV-read and batch granularity settings.

### DIFF
--- a/docs/fern/pages/reference/backends/vllm-configuration.mdx
+++ b/docs/fern/pages/reference/backends/vllm-configuration.mdx
@@ -191,22 +191,40 @@ These flags control the self-benchmark sweep that runs on startup before the wor
   Environment variable: `DYN_BENCHMARK_MODE`
 </ParamField>
 
-<ParamField path="--benchmark-prefill-granularity" type="integer" default="16">
-  Number of ISL sample points for the prefill sweep.
+<ParamField path="--benchmark-points-file" type="string" default="null">
+  JSON file of explicit pure prefill/decode benchmark points, applied uniformly to every data-parallel rank. The file completely replaces generated grid sampling for the phases selected by `--benchmark-mode`, so the sampling limits below are ignored when it is set. It is read and normalized once before vLLM workers start, then the same contents are forwarded to every rank.
 
-  Environment variable: `DYN_BENCHMARK_PREFILL_GRANULARITY`
+  Environment variable: `DYN_BENCHMARK_POINTS_FILE`
 </ParamField>
 
-<ParamField path="--benchmark-decode-length-granularity" type="integer" default="6">
-  Number of context length sample points for the decode sweep.
+<ParamField path="--prefill-max-new-token-samples" type="integer" default="64">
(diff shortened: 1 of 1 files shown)


dagil-nvidia:
Closing in favor of #13036, which covers everything here plus one defect this PR missed.

Both document the current benchmark sampling flags and demote the five \`--benchmark-*-granularity\` aliases to Deprecated. #13036 additionally corrects \`--benchmark-timeout\`: the page documented \`default="300"\` with hard-limit semantics ("worker startup fails if this limit is exceeded"), while \`components/src/dynamo/vllm/backend_args.py:467\` declares \`benchmark_timeout: int = 900\` and the limit is soft, since the measured iteration finishes and partial results are returned before startup continues. It also qualifies when the legacy flags are actually read, which is only with \`--benchmark-mode\` set and no points file.

One PR on one file beats a merge plus a follow-up, so this is the redundant half.

The release-branch cherry-pick #12983 has been updated to carry those same three corrections, so 1.4.0 does not ship without them. NVBug 6572846 stays tracked against #13036 and #12983.

PR #12982, opened 2026-08-10
fix(vllm): measure self-benchmark decode on the resident path

nealvaidya:
## Summary

Fix vLLM decode self-benchmarking so the measured iteration uses the same resident cached-request path as normal production decode.

The current benchmark performs an admission pass followed by a second decode pass, but async drain frames can remove the synthetic requests from the GPU model runner's persistent input batch. The measured pass then re-adds every request, recopies prompt tokens, and rebuilds block-table rows.

This adds host work proportional to KV context length and causes the benchmark to overstate the marginal cost of cached tokens.

## Details

This PR changes the decode benchmark to an explicit state machine:

1. Admit the synthetic batch one token before the requested context.
2. Wait for admission to complete.
3. Preserve the active request IDs through async empty frames.
4. Use the parent scheduler to produce the measured resident decode step.
5. Record only that resident step's `ForwardPassMetrics`.

Retention-only empty frames carry active request IDs with zero scheduled-token counts. They perform no model work and use the base `Scheduler` post-schedule handling so `AsyncScheduler` does not create output placeholders for work that will never produce output. The retention map is removed before normal scheduler output processing, where zero token counts are invalid.

The measured pass is produced through `super().schedule(throttle_prefills=True)` instead of manually constructing a cached `SchedulerOutput`. This keeps benchmark behavior aligned with the production scheduling, allocation, connector, and bookkeeping paths.

The measured schedule timestamp is recorded after attention-DP synchronization, excluding admission and barrier waits. If the resident batch is preempted or no longer matches the requested shape, the existing generated-point skip and explicit-point failure policies apply.

### Why this matters

A controlled TP1 H100 benchmark with FP8 KV cache showed:

| Measurement state | Decode slope |
| --- | ---: |
| Second decode step without resident retention | 1.62 ns/KV token |
| Resident batch retained | 1.01 ns/KV token |
| Natural persistent decode | 0.96 ns/KV token |
| Attention kernels alone | 1.02 ns/KV token |

Component timing attributed the discrepancy to model-runner state rebuilding:

- `InputBatch.add_request`: removed a 0.513 ns/token contribution
- `_update_states`: reduced from 0.524 to 0.004 ns/token
- total host scaling: reduced from 0.576 to 0.055 ns/token

After retaining the batch, the synthetic benchmark, natural decode, and attention-kernel measurements agree at approximately 1 ns/KV token.

This is therefore a measurement-correctness fix rather than merely a benchmark optimization.

### Compatibility

This PR intentionally preserves the existing:

- `ForwardPassMetrics` wire schema
- benchmark-point schema
- attention-DP capacity negotiation and grid synchronization
- pipeline-parallel benchmark guard
- EAGLE and hybrid KV-cache handling
- connector metadata handling
- aggregate Welford metrics
- benchmark result schema

## Validation

- [ ] Sync engine scheduling records exactly one resident FPM per decode point.
- [ ] Async engine scheduling records exactly one resident FPM per decode point.
- [x] Admission passes are not recorded.
- [x] Retention frames perform no model work or async placeholder accounting.
- [x] Retention maps are removed before scheduler output processing.
- [x] The measured step uses the parent scheduler.
- [x] The measured batch size and KV coordinate match the benchmark point.
- [x] Preemption and timeout paths clean up correctly.
- [x] KV and EC connector metadata remains valid.
- [x] Attention-DP synchronization tests pass.
- [x] Existing instrumented-scheduler tests pass: `143 passed`.
- [x] Repository isort, Black, Flake8, Ruff, and pytest-marker hooks pass.

Start with:

- `components/src/dynamo/vllm/instrumented_scheduler.py`
  - the decode benchmark state transitions
  - the retention-only empty-output branch
  - resident scheduling and FPM selection
- `components/src/dynamo/vllm/tests/test_vllm_instrumented_scheduler.py`
  - async retention regression coverage
  - admission-versus-resident recording tests

## Related Issues

🚫 This PR is NOT linked to an issue:

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12982" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decode benchmarking accuracy by measuring steady-state resident batches separately from request admission.
  * Corrected decode KV-read synchronization and context accounting.
  * Improved handling of retained requests, empty scheduling frames, and connector metadata.
  * Added safer behavior for benchmark timeouts, failed measurements, and drain scenarios.

* **Tests**
  * Expanded coverage for decode-stage transitions, resident measurement, padded allocations, and benchmark cleanup.

### DIFF
--- a/components/src/dynamo/vllm/instrumented_scheduler.py
+++ b/components/src/dynamo/vllm/instrumented_scheduler.py
@@ -3700,11 +3700,24 @@ class InstrumentedScheduler(AsyncScheduler):
         self._bench_decode_stage = _DecodeStage.MEASURING
         output = super().schedule(throttle_prefills=True)
         if point is None or output.total_num_scheduled_tokens != point.batch_size:
+            expected_tokens = point.batch_size if point is not None else 0
+            if output.total_num_scheduled_tokens > 0:
+                error = RuntimeError(
+                    "resident benchmark decode scheduled "
+                    f"{output.total_num_scheduled_tokens} of {expected_tokens} requests"
+                )
+                logger.error(
+                    "Aborting benchmark before dispatching a partial resident "
+                    "decode batch: %s",
+                    point,
+                )
+                self._bench_abort(error)
+                raise error
             logger.warning(
                 "Skipping benchmark decode point after the resident batch "
                 "scheduled %d of %d requests: %s",
                 output.total_num_scheduled_tokens,
-                point.batch_size if point is not None else 0,
+                expected_tokens,
                 point,
             )
             self._bench_cleanup_requests()
--- a/components/src/dynamo/vllm/instrumented_scheduler.py
+++ b/components/src/dynamo/vllm/instrumented_scheduler.py
@@ -1537,9 +1537,9 @@ class InstrumentedScheduler(AsyncScheduler):
 
     def schedule(self, throttle_prefills: bool = False) -> SchedulerOutput:
         if self._bench_active and self._bench_phase != _BenchPhase.IDLE:
-            if self._bench_decode_stage is _DecodeStage.RESIDENT:
-                return self._bench_measure_resident_batch()
             try:
+                if self._bench_decode_stage is _DecodeStage.RESIDENT:
+                    return self._bench_measure_resident_batch()
                 output = self._bench_step()
             except Exception as error:
                 point = self._bench_current_point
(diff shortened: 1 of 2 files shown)


nealvaidya:
Superseded by #13047, which replaces this implementation with the smaller explicit decode-stage state machine and incorporates the follow-up correctness fixes.

PR #12984, opened 2026-08-11
fix(vllm): migrate GLM-Image NIXL config to the stages deploy schema

GuanLuo:
## Overview:

GLM-Image AR→DiT disaggregation over NIXL dies on the first
`/v1/images/generations` request. The AR stage's engine core exits, the request
returns HTTP 500, and the deployment does not recover:

    File "vllm_omni/worker/gpu_model_runner.py", line 749, in _update_states
        resumed_token_ids = req_data.all_token_ids[req_id]
    KeyError: '<uuid>-<8 hex>'

followed by `Stage-0 replica-0 is dead` and `[AsyncOmni] Engine dead`.

## Summary

`stage_configs/glm_image_nixl.yaml` was the last config on vllm-omni's legacy
`stage_args` schema. That schema bypasses the deploy-config resolution in
`vllm_omni/config/stage_config.py`, which is what keeps the AR scheduler class
and the `async_scheduling` engine arg consistent: `_resolve_scheduler()` returns
`OmniARScheduler` for sync and `OmniARAsyncScheduler` for async, and the caller
derives the engine arg from whichever it selected.

Bypassing that, the config pinned `scheduler_cls: OmniARScheduler` — the
*synchronous* scheduler — and left `async_scheduling` unset, so the engine ran
async scheduling against a sync scheduler. The sync scheduler does not populate
`CachedRequestData.all_token_ids`, but the model runner's async-scheduling
resume path indexes it unconditionally. First request, KeyError, dead engine
core.

The `scheduler_cls` line dates to #10071 and was plausible when written: the
sync/async split in `omni_ar_scheduler.py` landed separately, keeping the old
name for the sync variant while `async_scheduling` defaults on. The only other
place that line appears upstream is a `platforms.npu` block, where it is
harmless because the NPU model runner has no `all_token_ids` resume path.

This migrates the file to the `stages` schema and drops `scheduler_cls`, so the
matched scheduler / `async_scheduling` pair is resolved by vllm-omni rather than
hand-declared.

## Details:

#12709 fixed an earlier crash in this same file and named this cause exactly —
"the legacy `stage_args` config format applies no schema defaults", and "the SHM
variant is unaffected because it consumes vllm-omni's own packaged config in the
`stages` format, whose parser supplies the default". That fix supplied the one
missing default; this removes the reason defaults go missing at all.

Topology that `GLM_IMAGE_PIPELINE` already freezes is dropped rather than
restated: stage and model types, model/tokenizer subdirs, input sources, the
AR→DiT input processor, and final-output flags. That retires the
`engine_input_source: []` workaround from #12709, since input sources now come
from the pipeline.

`async_chunk: false` is now explicit — `DeployConfig` defaults it to `True` and
GLM-Image declares no async-chunk next-stage processor. This matches vllm-omni's
own `deploy/glm_image.yaml`.

Deployment knobs, env-var device/TP selection, and sampling params carry over
unchanged. Top-level `connectors:`/`edges:` are the `stages`-schema spelling of
the previous `runtime.connectors`/`runtime.edges`; the connector initializer
normalizes both to the same internal shape. The dropped `runtime.defaults`
(`window_size`, `max_inflight`) has no consumer in vllm-omni.

The launch script needs no change: `--stage-configs-path` detects a `stages`
document and routes it to the deploy-config loader.

## Validation

- `pre-commit run --files examples/backends/vllm/launch/stage_configs/glm_image_nixl.yaml` — clean.
- Config resolves through vllm-omni's deploy loader to the intended matched
  pair: `scheduler_cls=OmniARAsyncScheduler` with `async_scheduling=True`.
- Manual 2-GPU single-node run via `disagg_omni_glm_image_nixl.sh`: both stages
  start, `/v1/images/generations` returns an image, and the AR engine core no
  longer dies on the first request.

Single file, `-73/+46`. The load-bearing change is the *absence* of
`scheduler_cls` — everything else is schema translation or removal of topology
the pipeline already owns.

One thing carried over unchanged and worth a second opinion: stage 1's
`output_connectors: {to_stage_router: dit_to_router}`. The `to_stage_router` key
appears nowhere in either this repo or vllm-omni, so the DiT→router hop may not
actually be on NIXL. Left as-is here to keep this change to the crash fix;
worth its own issue.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12984" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Updated the GLM image deployment configuration to the current deployment schema.
  * Simplified pipeline and stage configuration settings for improved consistency.
  * Consolidated connector, scheduling, caching, and execution settings at the deployment level.


ayushag-nv (approved):
looks good

ptarasiewiczNV (approved):

PR #12990, opened 2026-08-11
chore(pre-commit): bump black to 26.5.1 for Python 3.14

dagil-nvidia:
## Summary

`.pre-commit-config.yaml` pinned `psf/black` at `23.1.0`, which calls `ast.Str`. Python 3.14 removed it, so on a 3.14 interpreter the hook dies with:

```
error: cannot format <file>: module 'ast' has no attribute 'Str'
```

The failure looks intermittent, which makes it confusing: black only runs its AST-equivalence check on files it actually rewrites, so an already-formatted tree passes and only a change that needs formatting trips it. CI is unaffected today because its runners are on an older Python.

Worth calling out, because it sends you looking at the wrong interpreter: the Python that matters is **pre-commit's own**, not the repo's `python3`. pre-commit builds each hook's venv from whatever Python it is itself installed under. On a machine where `python3` is 3.9 but pre-commit runs on 3.14, the black env is 3.14 and the hook breaks.

### Changes

- **`black` `23.1.0` → `26.5.1`**, with a comment naming the cause, matching the existing `flake8` note one hook below.
- **`args: [--target-version, py310]`** on the black hook. black 26.x infers its target range from `requires-python = ">=3.10"` in `pyproject.toml`, takes the top of that range (`py315`), and then warns on *every* reformat that the running interpreter is too old to verify the result — advising you to "run Black with Python 3.15". Pinning to the declared floor silences that, and makes output identical no matter which Python a contributor's pre-commit runs on. Verified to produce byte-identical formatting.
- **120 files reformatted** into black's 2024/2026 stable style. Whitespace only, overwhelmingly wrapping a long right-hand side in parentheses instead of splitting on the operator.
- **Corrected the `flake8` comment.** It read `5.0.4 crashes on Python 3.12+`; 5.0.4 in fact lints cleanly on 3.12 and 3.13 and breaks on 3.14, same `ast.Str` cause. Comment only — `flake8` stays at `7.3.0`.

`ast.Str` was deprecated in 3.8 and removed in **3.14**, not 3.12.

### Other pinned hooks

Checked while in here; none need a change.

| Hook | Status on Python 3.14 |
| --- | --- |
| `isort` 5.12.0 | Runs correctly |
| `flake8` 7.3.0 | Runs correctly |
| `codespell` 2.2.4 | Runs correctly |
| `pre-commit-hooks` 4.4.0 | Runs correctly |
| `ruff`, `clang-format` | Prebuilt binaries, unaffected |

Also incidental: the repo has zero `.pyx`/`.pxd` files, and `identify` tags `.pyi` as `pyi` rather than `python`, so `types_or: [python, cython]` resolves to 1441 `.py` files and nothing else.

## Validation

Run on macOS with pre-commit 4.5.1 on Python 3.14.4.

- **Reproduced the failure directly.** black 23.1.0 on Python 3.14 fails with `error: cannot format ...: module 'ast' has no attribute 'Str'`; the same black on 3.13 formats the same file fine, which is what establishes the 3.14 threshold.
- **`pre-commit run --all-files`: every hook passes.** The one exception is `gen-llms-tables-check`, which is **pre-existing and unrelated** — stashing this branch's changes entirely and re-running it on a clean tree produces the identical failure (3 stale generated docs outputs). Left untouched.
- **Confirmed the fix end to end.** With the new pin, the hook reformats a fresh unformatted file cleanly: no crash, no target-version warning, normal "files were modified by this hook" behavior.
- **Reformat verified semantically inert.** Parsed all 120 files at `HEAD` and after the reformat and compared AST dumps — all identical, modulo docstring whitespace, which black is permitted to normalize. All 120 also pass `py_compile`.
- **Confirmed the `--target-version` pin is free.** Re-running black with the flag over the reformatted tree leaves all 1441 in-scope files unchanged.

For reference, churn by candidate version: 87 files at 24.1.0, 89 at 25.9.0, 120 at 26.5.1. The extra 33 are the 2026 stable style, which the repo would otherwise absorb as a second reformat commit later.

Python unit tests were not run — this worktree has no built bindings. For a whitespace-only change the AST-equivalence check above covers the same ground more directly.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12990" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Standardized code formatting across runtime, tooling, deployment, and test components.
  * Improved readability of multiline expressions, type annotations, context managers, protocol declarations, and string construction.
  * No functional behavior, public interfaces, or generated output changed.

* **Chores**
  * Updated formatting-tool configuration and Python compatibility guidance.

* **Tests**
  * Added regression coverage for routing, request forwarding, rank handling, API checks, and environment overrides.


sttts (approved):
For docs and operator.

.pre-commit-config.yaml
```diff
@@ -22,12 +22,19 @@ repos:
   - id: isort
     additional_dependencies: [toml]
 - repo: https://github.com/psf/black
-  rev: 23.1.0
+  rev: 26.5.1  # 23.1.0 crashes on Python 3.14+ (ast.Str removed)
```
jasonqinzhou, line 25: Black now fails on GitHub's merge commit: `lib/gpu_memory_service/v1/server/rpc.py` landed on `main` after this branch point, and 26.5.1 reformats its conditional logging expression. Because this PR changes the repository-wide hook, the merge result is not formatter-clean even though the 120 files present on this branch are. Please rebase or merge current `main`, rerun `pre-commit run black --all-files`, and include the resulting GMS reformat before landing the upgrade.

components/src/dynamo/frontend/prepost.py
```diff
@@ -38,8 +38,7 @@ class _Renderer(Protocol):
 
     async def render_messages_async(
         self, messages: Any, params: ChatParams
-    ) -> tuple[Any, dict[str, Any]]:
-        ...
+    ) -> tuple[Any, dict[str, Any]]: ...
```
jasonqinzhou, line 41: This reformat moves public symbols in API-scanned files, but the generated reference pages were not refreshed. On GitHub's merge ref, `gen_python_api.py --check` reports `frontend.mdx`, `common.mdx`, `mocker.mdx`, and `nixl_connect.mdx` stale; for example, `PreprocessResult` still links to line 45 even though it now begins at line 44, while the base branch passes the same check. Please regenerate and commit these four pages after the reformat so their source anchors and both generated-doc checks are current.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.8/10 · high confidence**

## What this PR does

This PR upgrades the repository-wide Black pre-commit hook from 23.1.0 to 26.5.1 for Python 3.14 compatibility, pins formatting to the project's Python 3.10 floor, corrects the adjacent flake8 compatibility note, and applies the new stable style to 120 Python files.

The root-cause write-up is unusually clear: it distinguishes pre-commit's interpreter from the repository's Python, pins the target to the declared 3.10 floor, and backs the repository-wide churn with semantic validation.

## Why this score

6.8/10, REQUEST_CHANGES: the formatter upgrade and existing 120-file reformat are semantically sound, but the current merge result is not ready to land. Black reformats a GMS file added on main after this branch point, so pre-commit fails, and formatter-induced line shifts leave four generated Python API pages stale with incorrect source anchors.

PR #12991, opened 2026-08-11
fix(vllm): roll back scale_elastic_ep on a failed grow instead of wedging the engine

tzulingk:
#### Overview:

A failed `scale_elastic_ep` grow wedged the vLLM engine: the leader process survived but stopped serving, stuck at a data-parallel size it could not fulfill. This makes the grow atomic — on failure it rolls back to the previous dp and keeps serving; if the rollback itself fails, it reports the engine as unrecoverable (must restart) instead of a false recovery. Fixes DYN-3838.

#### Details:

Observed during elastic-EP testing (DYN-3686): when a new rank died mid-init while joining NCCL (`cudaErrorNvlinkUncorrectable` on a degraded node), the engine had already committed to the larger dp, so ranks 0/1 then expected a 3-way collective that never completes — inference hung/errored while the leader stayed alive (`restart=0`).

`BaseWorkerHandler.scale_elastic_ep` in `components/src/dynamo/vllm/handlers.py` now:
- factors the ray-`list_nodes`-patched scale into a local `_scale_engine(size)` helper so the grow and any rollback go through the same path;
- captures `prev_dp = engine_client.vllm_config.parallel_config.data_parallel_size` before the grow;
- on grow failure, rolls back with a compensating `_scale_engine(prev_dp)` (vLLM's `scale_elastic_ep` is symmetric — a smaller size shrinks). Success → returns `{"status": "error", "recoverable": true, "data_parallel_size": prev_dp}` and the engine keeps serving;
- if the rollback also fails → returns `{"status": "error", "recoverable": false, ...}` with "the engine ... must be restarted", because an uncorrectable NVLink error can poison the CUDA/NCCL context so the DP group cannot be rebuilt in process.

Two-tier by design: recoverable failures (unschedulable, or a clean init error before any collective) roll back in place; unrecoverable failures surface a clear restart signal — the design already tolerates a leader restart (the follower rejoins). Wiring `recoverable: false` to an automatic pod restart (liveness-probe failure / explicit exit) is a deliberate, separately-reviewable follow-up.

**Validation:**
- New `components/src/dynamo/vllm/tests/test_vllm_elastic_ep_handlers.py` (pure unit test; mocked engine client, `ray` stubbed in `sys.modules`; `test_vllm_` prefix so it skips cleanly where vLLM is absent) covers: successful scale (grow called once), failed grow → rollback to prev dp (`scale_elastic_ep` awaited with the grow size then the prior size; `recoverable=true`), and failed rollback → `recoverable=false` with "must be restarted", plus the missing-field and below-floor rejections.
- `pre-commit run --files` with the repo-pinned hooks (isort / black / flake8 / ruff / codespell / pytest-marker report) passes on both files.

#### Where should the reviewer start?

`components/src/dynamo/vllm/handlers.py` — the `scale_elastic_ep` failure path: the new `_scale_engine` helper and the grow/rollback try-nest (the `except Exception as grow_err` block). Then the three rollback cases in `components/src/dynamo/vllm/tests/test_vllm_elastic_ep_handlers.py`.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/12991" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved elastic scaling reliability by using current cluster node information during scale operations.
  * Added automatic rollback when a scale-up fails.
  * Added clearer reporting for recoverable and unrecoverable scaling failures.
  * Added validation for missing or below-minimum scaling sizes.

* **Tests**
  * Added coverage for successful scaling, validation errors, and rollback scenarios.


components/src/dynamo/vllm/handlers.py
```diff
@@ -1474,14 +1474,73 @@ def __init__(self, d: dict) -> None:
...
+                        "message": (
+                            f"scale to dp={new_dp_size} failed ({grow_err}); "
+                            f"rollback to dp={prev_dp} also failed ({rollback_err}); "
+                            "the engine is in an unrecoverable state and must be "
+                            "restarted"
+                        ),
+                    }
+                logger.info(
+                    f"[ElasticEP] Rolled back to dp={prev_dp}; engine still serving"
+                )
```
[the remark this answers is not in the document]
tzulingk: Fixed in 575510581e — converted the four new `[ElasticEP]` log calls (grow failure, rollback warning, rollback failure, rollback success) to lazy `%s` formatting per `.ai/python-guidelines.md`. The returned message dicts keep their eager f-strings as you noted. Thanks!

components/src/dynamo/vllm/handlers.py
```diff
@@ -1474,14 +1474,73 @@ def __init__(self, d: dict) -> None:
...
+                if new_dp_size == prev_dp:
+                    # No size change was applied; nothing to roll back.
+                    raise
+                # Roll back to the last good dp: "either it grows, or nothing
+                # changes" for the engine, not just the pod.
+                try:
+                    logger.warning(
+                        f"[ElasticEP] Rolling back to data_parallel_size={prev_dp}"
+                    )
+                    await _scale_engine(prev_dp)
```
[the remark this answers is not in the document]
tzulingk: Confirmed against the pinned vLLM (v0.26.0) — good catch, this was a real no-op.

`AsyncLLM.scale_elastic_ep` updates `parallel_config.data_parallel_size` only *after* a successful `engine_core.scale_elastic_ep(...)` (`async_llm.py:1039-1040`), and short-circuits to a no-op when the requested size equals the recorded size (`async_llm.py:1005-1011`, *"Data parallel size is already N, skipping scale"*). So after a failed grow the recorded size is still `prev_dp`, and the rollback `scale_elastic_ep(prev_dp)` was skipped entirely — the half-created ranks were never torn down while we reported `recoverable=True`.

Fixed in 575510581e: before rolling back, the handler advances the recorded size to the attempted target (`vllm_config.parallel_config.data_parallel_size = new_dp_size`) so the rollback drives a *genuine* scale-down. Two sub-cases, both now honest:
- new ranks already registered (failure in the readiness wait): the down-reconfigure tears them down and the engine keeps serving → `recoverable=True`.
- grow died before registration: the down-reconfigure hits vLLM's `assert new_size != current_size` in the MP client and raises → we report `recoverable=False` and restart, instead of a false recovery.

The unit test previously mocked `scale_elastic_ep` with a plain `AsyncMock`, which couldn't model any of this. It now uses a fake engine that mirrors vLLM's guard + record-after-success semantics and asserts the rollback is a *real* reconfigure (`[grow, rollback]`), so reintroducing the no-op fails the test.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1474,14 +1474,73 @@ def __init__(self, d: dict) -> None:
...
+                except Exception as rollback_err:
```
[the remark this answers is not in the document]
tzulingk: Addressed in 0dfc020721. Added an explicit `except EngineDeadError` ahead of the broad grow handler: on engine death we skip the rollback entirely and return `recoverable=False` (worker must restart), consistent with how the rest of this file treats `EngineDeadError` (`_shutdown_on_engine_dead`). New test `test_engine_dead_error_is_fatal_and_skips_rollback` asserts the dead engine reaches only the failed grow (`calls == [3]`) with no rollback call.

On the "roll back only documented recoverable failures / re-raise unknown" part: vLLM v0.26 doesn't expose a stable taxonomy of recoverable scaling exceptions, so an explicit whitelist would be fragile and would drop genuinely recoverable mid-reconfigure failures. We special-case the one unambiguously fatal case (`EngineDeadError`) and keep the broad catch + rollback for the rest — which is the safe direction, since a rollback that can't complete raises and we already downgrade to `recoverable=False`.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1474,14 +1474,73 @@ def __init__(self, d: dict) -> None:
...
+                    await self.engine_client.scale_elastic_ep(size)
+                finally:
+                    _ray_util_state.list_nodes = original_list_nodes
+
+            # Capture the current dp up front so a failed grow can be rolled back
+            # to it instead of wedging the engine at a size it cannot fulfill: a
+            # half-created rank that dies leaves its peers expecting a collective
+            # that never completes, so the leader survives but stops serving.
+            prev_dp = self.engine_client.vllm_config.parallel_config.data_parallel_size
+
```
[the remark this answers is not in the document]
tzulingk: Good one to raise — I traced it against the pinned vLLM (v0.26.0) and decided to **keep `TimeoutError` on the rollback path**, for two reasons:

1. The pre-reconfigure drain that raises `TimeoutError` only runs when `VLLM_ELASTIC_EP_DRAIN_REQUESTS` is set (`async_llm.py:1013-1018` → `wait_for_requests_to_drain` → `raise TimeoutError` at `:989`). Dynamo never sets that env var, so the "timeout before engine-core reconfiguration starts" case doesn't occur in our deployments by default.

2. More importantly, `TimeoutError` is **not** exclusive to the drain phase in v0.26. vLLM also raises `TimeoutError` *mid-reconfigure* when a new engine core fails to come up in time — `core_client.py:1674`, which fires **after** the new (dead) engines have already been appended to `core_engines`. That is exactly the wedge scenario this PR targets, and it *requires* the rollback to scale the half-created ranks back down. Since both cases surface as the same bare `TimeoutError` with no reliable discriminator, blanket-skipping rollback on `TimeoutError` would reintroduce the wedge / false-recovery for the primary case.

So `TimeoutError` stays on the rollback path. The one clearly-fatal case — `EngineDeadError` — is now handled separately (0dfc020721) as unrecoverable-without-rollback. If drain-on-scale is ever enabled in dynamo, a benign drain timeout would at worst trigger a wasted rollback attempt (`recoverable=False`), never a false "recovered".

components/src/dynamo/vllm/handlers.py
```diff
@@ -1520,14 +1520,105 @@ def __init__(self, d: dict) -> None:
...
+                # attempting a rollback that cannot succeed and would only mask a
+                # dead engine as a recovered one.
+                logger.error(
+                    "[ElasticEP] Engine died during scale to dp=%s: %s",
+                    new_dp_size,
+                    dead_err,
+                )
+                return {
+                    "status": "error",
+                    "recoverable": False,
```
biswapanda: how ius this field `recoverable` used ? I didn't find any consumer for this yet, so we dont restart or unregister the worker.

biswapanda: The health monitor may not classify that process as dead, so it can remain registered and receive traffic indefinitely. 

The new fatal-engine test only checks a Boolean and message string, it never verifies shutdown or unregistration.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1520,14 +1520,105 @@ def __init__(self, d: dict) -> None:
...
+            prev_dp = self.engine_client.vllm_config.parallel_config.data_parallel_size
+
             try:
-                _ray_util_state.list_nodes = lambda **kw: [
-                    _NodeInfo(n) for n in ray.nodes() if n.get("Alive", False)
-                ]
-                await self.engine_client.scale_elastic_ep(new_dp_size)
-            finally:
-                _ray_util_state.list_nodes = original_list_nodes
+                await _scale_engine(new_dp_size)
```
biswapanda, line 1578: Rollback can race still-running work from the failed grow, so the operation is not atomic.

vLLM launches existing-core reconfiguration tasks and actor creation through asyncio.to_thread then awaits them with asyncio.gather. When one operation raises, gather propagates immediately without cancelling or waiting for its siblings  those operations continue running.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1520,14 +1520,105 @@ def __init__(self, d: dict) -> None:
...
+                if new_dp_size == prev_dp:
+                    # No size change was applied; nothing to roll back.
+                    raise
+                # vLLM records parallel_config.data_parallel_size only *after* a
+                # successful reconfigure, so after a failed grow the engine still
+                # believes it is at prev_dp. A rollback call to prev_dp would then
+                # hit vLLM's "already at this size, skipping scale" guard and be a
+                # silent no-op -- leaving the half-created ranks in place while we
+                # falsely report recovery. Advance the recorded size to the
+                # attempted target so the rollback drives a real scale-down.
```
biswapanda: The handler overwrites parallel_config.data_parallel_size with the failed target so a rollback request will bypass AsyncLLM’s same-size guard. 

However, real vLLM has multiple state layers:
1. AsyncLLM checks the recorded config.
2. DPLBAsyncMPClient determines the actual size from len(core_engines) and rejects a request for that same size.
3. The Ray actor manager updates the shared config before the client appends new core identities.

If actor initialization fails early, the actual core size is still prev_dp forging the config lets the request past the outer guard, but the inner client rejects rollback to prev_dp. If failure occurs after new core identities are appended, the shared config is already the attempted size, so the overwrite seems unnecessary.

biswapanda (commented):
left few comments in-line

components/src/dynamo/vllm/handlers.py
tzulingk: You're right — `recoverable` had no consumer anywhere in the repo, so a dead/wedged engine stayed registered and kept taking traffic, and the test only checked a boolean. Fixed in 40ae1c24ce.

The scaling path now does what every other engine path in this handler already does: on `EngineDeadError` it calls `_shutdown_on_engine_dead(...)` → `_shutdown_worker()` → `runtime.shutdown()` + `os._exit(1)`, so the worker actually exits, restarts, and re-registers. Any other failed grow goes through `_shutdown_worker()` too. There is no longer a `recoverable` field at all.

The test now asserts the shutdown is genuinely triggered (a `BaseException` stand-in for the `NoReturn` shutdown, so the handler can never fall through to a success return), not just that a boolean/string was returned.

components/src/dynamo/vllm/handlers.py
tzulingk: Agreed. Confirmed against vLLM v0.26 `_scale_up_elastic_ep`: it launches the existing-rank reconfig tasks and the new-actor creation, then `await asyncio.gather(...)`. On the first exception `gather` propagates immediately without cancelling or awaiting the siblings, so a rollback would run against still-mutating grow work — not atomic, exactly as you said.

Rather than try to make a rollback atomic on top of an API that gives no such guarantee, 40ae1c24ce drops the in-process rollback entirely and fails fast (restart the worker). No rollback → no race.

components/src/dynamo/vllm/handlers.py
tzulingk: Right on both branches. The config overwrite only fools `AsyncLLM`'s outer guard; the inner `DPLBAsyncMPClient` keys the size off `len(core_engines)`, not the config — so if actor init failed early it rejects the rollback to `prev_dp` (`assert new != cur` on the real count → insufficient), and if it failed late the shared config already holds the target (unnecessary).

40ae1c24ce removes the overwrite and the rollback altogether. The deciding factor was that vLLM itself does no rollback or cleanup on a failed scale — `_scale_up_elastic_ep` has no try/except, and vLLM's own `/scale_elastic_ep` endpoint just logs and returns HTTP 500, leaving the engine in whatever partial state it reached. There is no rollback primitive to build on, so the only safe recovery is to restart the worker.

components/src/dynamo/vllm/handlers.py
```diff
@@ -1517,14 +1517,44 @@ def __init__(self, d: dict) -> None:
...
+                # worker so it comes back clean, rather than fake a recovery that
+                # nothing acts on.
+                logger.error(
+                    "[ElasticEP] Scaling to dp=%s failed: %s. vLLM does not roll "
+                    "back a failed scale; restarting the worker to recover a clean "
+                    "state.",
+                    new_dp_size,
+                    grow_err,
+                )
+                self._shutdown_worker()  # NoReturn: runtime.shutdown() + os._exit(1)
```
biswapanda, line 1602: Gate unsupported elastic-EP configurations before this fatal path. Please capability-gate route registration or return a nonfatal configuration error before `_scale_engine`, and add an unsupported-configuration test.

`control/scale_elastic_ep` is registered for every vLLM worker, but this handler does not check `enable_elastic_ep`, `enable_eplb`, `data_parallel_backend == "ray"`, or whether the engine client implements elastic scaling. With the pinned vLLM defaults (`mp`, elastic EP disabled), a valid request for a different DP size reaches `prepare_elastic_ep()` and raises `NotImplementedError` or the Ray-backend assertion. 

This catch then calls `_shutdown_worker()` and kills a healthy worker before any scale state was mutated.

tzulingk: Good catch — this would have restarted a healthy worker. Fixed in efd5fcea23.

`control/scale_elastic_ep` is registered on every vLLM worker, so on a default worker (`data_parallel_backend="mp"`, `enable_elastic_ep=False`) a valid scale request reached vLLM's `scale_elastic_ep` → `NotImplementedError` / the Ray-backend assertion → caught by the fail-fast grow handler → needless restart.

Added a capability gate right after reading `parallel_config`, before `_scale_engine`: if `not enable_elastic_ep` or `data_parallel_backend != "ray"`, it returns a nonfatal configuration error and never touches the engine. Added `test_unsupported_config_is_rejected_without_restart` (asserts no restart and no engine call); the existing fixtures now advertise a supported config so they still exercise the real paths.

tzulingk:
📌 Follow-up hardening PR: **#14053** addresses the architecture-review findings on `scale_elastic_ep` — an outer timeout backstop so a hung scale can't wedge the endpoint forever, moving the heavy `ray` import off the event loop, plus two doc-only clarifications. It's built on top of this PR; please review/merge it **after** this one.

biswapanda (approved):

PR #12999, opened 2026-08-11
perf(frontend): flush detokenization metrics once

jthomson04:
## Summary

Flush the existing detokenization Prometheus accumulators once when the transformed response stream is dropped. This removes cumulative per-chunk counter updates without losing partial metrics when a request is cancelled, errors, times out, or its consumer drops the stream early.

Metric names, types, units, PromQL behavior, response payloads, and usage-chunk ordering are unchanged.

## Details

The request tracker stores cumulative detokenization count and duration. The old response path added those cumulative values to the global counters for every emitted backend chunk, which caused triangular overcounting and repeated floating-point counter updates on a highly concurrent streaming path.

This change:

- caches the optional request tracker `Arc` once instead of cloning it on each chunk
- uses a small drop guard to publish the final accumulated values for normal completion, cancellation, backend errors, downstream short-circuits, and early stream drops
- performs no new per-chunk synchronization and needs no idempotency flag because each request owns one tracker and one guard
- uses one private inline helper to build normal and usage-chunk `LLMMetricAnnotation` values from the same fields
- adds serial regression tests for normal completion, payload usage, early drop, and postprocessor cancellation

## Performance evidence

The initial Tyche AgentX c=10240 profile on the earlier profiled base showed:

- detokenization metric observations per completed request fell from 5,298.7 to 3.55
- `futures_util::stream::unfold::Unfold::poll_next` exclusive CPU fell from 47.42% to 1.07%
- the exact metric block used 4 of 261,897 samples, or 0.0015% exclusive CPU
- frontend CPU time fell from 24.51 ms/request to 15.07 ms/request versus the prior full-node repeat

A later matched upstream-main versus PR diagnostic profile confirmed the mechanism:

- frontend CPU fell from 38.30 to 18.95 ms/request in the first 120-second window
- `Unfold::poll_next` fell from 39.047% to 0.437% exclusive CPU
- detokenization observations fell from 3,091.7 to 1.365 per completed request
- both perf captures had zero lost samples and fewer than 5% unresolved frames

That matched pair had native transport request errors: 0.0278% in the control and 0.2944% in the experiment. It is valid CPU-mechanism evidence, but its full-drain throughput and TTFT results are not clean acceptance claims.

## Validation

Validated after rebasing onto upstream `da31d3f00dc51ad431299ee537e5a1f4c9af6a71`:

- `cargo test -p dynamo-llm --test test_streaming_usage` (14 passed)
- `cargo check -p dynamo-llm`
- `git diff --check`
- source inspection confirms that counter updates exist only in the drop guard and that the request tracker is cloned once when stream state is created

Start with `DetokenizeMetricsGuard` and `build_llm_metric_annotation` in `lib/llm/src/preprocessor.rs`, then review the four metric lifecycle tests in `lib/llm/tests/test_streaming_usage.rs`.

## Related Issues

**This PR is not linked to an issue:**

- [x] Confirmed - no related issue

### DIFF
--- a/lib/llm/src/preprocessor.rs
+++ b/lib/llm/src/preprocessor.rs
@@ -243,7 +243,7 @@ const DEFAULT_THINKING_MODE_RUNTIME_KEY: &str = "default_thinking_mode";
 /// frontend's own recordings (when present) take precedence over the forwarded values.
 fn drain_router_routing_data(
     data: &mut Option<BackendOutput>,
-    tracker: Option<&std::sync::Arc<crate::protocols::common::timing::RequestTracker>>,
+    tracker: Option<&crate::protocols::common::timing::RequestTracker>,
 ) {
     let Some(routing_data) = data.as_mut().and_then(|d| d.routing_data.take()) else {
         return;
@@ -3154,6 +3154,7 @@ impl OpenAIPreprocessor {
         {
             response_stream: Pin<Box<dyn Stream<Item = Annotated<BackendOutput>> + Send>>,
             response_generator: Box<dyn DeltaGeneratorExt<Resp>>,
+            tracker: Option<Arc<RequestTracker>>,
             context: Arc<dyn AsyncEngineContext>,
             cancelled: bool,
             cumulative_output_tokens: usize,
@@ -3170,9 +3171,26 @@ impl OpenAIPreprocessor {
             image_tokens: Option<usize>,
         }
 
+        impl<Resp> Drop for State<Resp>
+        where
+            Resp: Send + Sync + Clone + 'static + std::fmt::Debug,
+        {
+            fn drop(&mut self) {
+                let Some(tracker) = self.tracker.as_deref() else {
+                    return;
+                };
+                if let Some(total) = tracker.detokenize_total_latency() {
+                    DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
+                }
+                DETOKENIZE_TOKEN_COUNT.inc_by(tracker.detokenize_count() as f64);
+            }
+        }
+
+        let tracker = generator.tracker();
         let state = State {
             response_stream: Box::pin(stream),
             response_generator: generator,
+            tracker,
             context: context.clone(),
             cancelled: false,
             cumulative_output_tokens: 0,
@@ -3213,10 +3231,7 @@ impl OpenAIPreprocessor {
                     // Split topology: overlay a standalone router's forwarded routing_data
                     // (timing, query-only token_ids) onto this request's tracker so the
                     // frontend's nvext/timing surfaces populate.
-                    drain_router_routing_data(
-                        &mut response.data,
-                        inner.response_generator.tracker().as_ref(),
-                    );
+                    drain_router_routing_data(&mut response.data, inner.tracker.as_deref());
 
                     if inner.cancelled {
                         tracing::debug!(
@@ -3278,17 +3293,15 @@ impl OpenAIPreprocessor {
 
                     // Create LLM metrics annotation with prefill/decode worker info from tracker.
                     // Worker types are stored at routing time to avoid expensive MDC lookup.
-                    let tracker = inner.response_generator.tracker();
-                    let prefill_worker_id = tracker.as_ref().and_then(|t| t.prefill_worker_id());
-                    let prefill_dp_rank = tracker.as_ref().and_then(|t| t.prefill_dp_rank());
+                    let tracker = inner.tracker.as_deref();
+                    let prefill_worker_id = tracker.and_then(|t| t.prefill_worker_id());
+                    let prefill_dp_rank = tracker.and_then(|t| t.prefill_dp_rank());
                     let prefill_worker_type = tracker
-                        .as_ref()
                         .and_then(|t| t.prefill_worker_type())
                         .map(String::from);
-                    let decode_worker_id = tracker.as_ref().and_then(|t| t.decode_worker_id());
-                    let decode_dp_rank = tracker.as_ref().and_then(|t| t.decode_dp_rank());
+                    let decode_worker_id = tracker.and_then(|t| t.decode_worker_id());
+                    let decode_dp_rank = tracker.and_then(|t| t.decode_dp_rank());
                     let decode_worker_type = tracker
-                        .as_ref()
                         .and_then(|t| t.decode_worker_type())
                         .map(String::from);
                     let llm_metrics = LLMMetricAnnotation {
@@ -3306,15 +3319,14 @@ impl OpenAIPreprocessor {
                         decode_worker_id,
                         decode_dp_rank,
                         decode_worker_type,
-                        tokenize_latency: tracker.as_ref().and_then(|t| t.tokenize_latency()),
+                        tokenize_latency: tracker.and_then(|t| t.tokenize_latency()),
                         detokenize_total_latency: tracker
-                            .as_ref()
                             .and_then(|t| t.detokenize_total_latency()),
-                        detokenize_count: tracker.as_ref().map(|t| t.detokenize_count()),
+                        detokenize_count: tracker.map(|t| t.detokenize_count()),
                     };
                     if inner.trace_tokens_enabled {
                         crate::request_trace::record_llm_metric_tokens(
-                            tracker.as_deref(),
+                            tracker,
                             isl,
                             current_osl,
                             None,
@@ -3340,36 +3352,24 @@ impl OpenAIPreprocessor {
                     // again. The stream is exhausted and will panic if polled after None.
                     inner.finished = true;
 
-                    // Flush per-request detokenize accumulators to global Prometheus counters
-                    // once, after the backend response stream completes.
-                    if let Some(t) = inner.response_generator.tracker().as_ref() {
-                        if let Some(total) = t.detokenize_total_latency() {
-                            DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
-                        }
-                        DETOKENIZE_TOKEN_COUNT.inc_by(t.detokenize_count() as f64);
-                    }
-
                     if inner.finish_reason_sent && !inner.usage_chunk_sent {
                         inner.usage_chunk_sent = true;
 
                         let usage_chunk = inner.response_generator.create_usage_chunk();
                         let usage = inner.response_generator.get_usage();
-                        let tracker = inner.response_generator.tracker();
+                        let tracker = inner.tracker.as_deref();
                         let cached_tokens = usage
                             .prompt_tokens_details
                             .as_ref()
                             .and_then(|d| d.cached_tokens.map(|c| c as usize));
-                        let prefill_worker_id =
-                            tracker.as_ref().and_then(|t| t.prefill_worker_id());
-                        let prefill_dp_rank = tracker.as_ref().and_then(|t| t.prefill_dp_rank());
+                        let prefill_worker_id = tracker.and_then(|t| t.prefill_worker_id());
+                        let prefill_dp_rank = tracker.and_then(|t| t.prefill_dp_rank());
                         let prefill_worker_type = tracker
-                            .as_ref()
                             .and_then(|t| t.prefill_worker_type())
                             .map(String::from);
-                        let decode_worker_id = tracker.as_ref().and_then(|t| t.decode_worker_id());
-                        let decode_dp_rank = tracker.as_ref().and_then(|t| t.decode_dp_rank());
+                        let decode_worker_id = tracker.and_then(|t| t.decode_worker_id());
+                        let decode_dp_rank = tracker.and_then(|t| t.decode_dp_rank());
                         let decode_worker_type = tracker
-                            .as_ref()
                             .and_then(|t| t.decode_worker_type())
                             .map(String::from);
                         let llm_metrics = LLMMetricAnnotation {
@@ -3387,15 +3387,14 @@ impl OpenAIPreprocessor {
                             decode_worker_id,
                             decode_dp_rank,
                             decode_worker_type,
-                            tokenize_latency: tracker.as_ref().and_then(|t| t.tokenize_latency()),
+                            tokenize_latency: tracker.and_then(|t| t.tokenize_latency()),
                             detokenize_total_latency: tracker
-                                .as_ref()
                                 .and_then(|t| t.detokenize_total_latency()),
-                            detokenize_count: tracker.as_ref().map(|t| t.detokenize_count()),
+                            detokenize_count: tracker.map(|t| t.detokenize_count()),
                         };
                         if inner.trace_tokens_enabled {
                             crate::request_trace::record_llm_metric_tokens(
-                                tracker.as_deref(),
+                                tracker,
                                 Some(usage.prompt_tokens as usize),
                                 usage.completion_tokens as usize,
                                 cached_tokens,
--- a/lib/llm/src/preprocessor.rs
+++ b/lib/llm/src/preprocessor.rs
@@ -290,6 +290,41 @@ where
     attach_metrics_annotation(response, &metrics);
 }
 
+#[inline]
+fn build_llm_metric_annotation(
+    tracker: Option<&RequestTracker>,
+    input_tokens: usize,
+    output_tokens: usize,
+    chunk_tokens: usize,
+    cached_tokens: Option<usize>,
+    mm_counts: MultimodalCounts,
+    image_tokens: Option<usize>,
+) -> LLMMetricAnnotation {
+    LLMMetricAnnotation {
+        input_tokens,
+        output_tokens,
+        chunk_tokens,
+        cached_tokens,
+        image_count: mm_counts.image,
+        video_count: mm_counts.video,
+        audio_count: mm_counts.audio,
+        image_tokens,
+        prefill_worker_id: tracker.and_then(|t| t.prefill_worker_id()),
+        prefill_dp_rank: tracker.and_then(|t| t.prefill_dp_rank()),
+        prefill_worker_type: tracker
+            .and_then(|t| t.prefill_worker_type())
+            .map(String::from),
+        decode_worker_id: tracker.and_then(|t| t.decode_worker_id()),
+        decode_dp_rank: tracker.and_then(|t| t.decode_dp_rank()),
+        decode_worker_type: tracker
+            .and_then(|t| t.decode_worker_type())
+            .map(String::from),
+        tokenize_latency: tracker.and_then(|t| t.tokenize_latency()),
+        detokenize_total_latency: tracker.and_then(|t| t.detokenize_total_latency()),
+        detokenize_count: tracker.map(|t| t.detokenize_count()),
+    }
+}
+
 // Reasoning State for reasoning parsing transformation step.
 //
 // The reasoning parser and the guided-JSON bypass decision are kept per
@@ -3148,13 +3183,36 @@ impl OpenAIPreprocessor {
         S: Stream<Item = Annotated<BackendOutput>> + Send + 'static,
         Resp: Send + Sync + Clone + 'static + std::fmt::Debug,
     {
+        struct DetokenizeMetricsGuard {
+            tracker: Option<Arc<RequestTracker>>,
+        }
+
+        impl DetokenizeMetricsGuard {
+            #[inline]
+            fn tracker(&self) -> Option<&RequestTracker> {
+                self.tracker.as_deref()
+            }
+        }
+
+        impl Drop for DetokenizeMetricsGuard {
+            fn drop(&mut self) {
+                let Some(tracker) = self.tracker() else {
+                    return;
+                };
+                if let Some(total) = tracker.detokenize_total_latency() {
+                    DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
+                }
+                DETOKENIZE_TOKEN_COUNT.inc_by(tracker.detokenize_count() as f64);
+            }
+        }
+
         struct State<Resp>
         where
             Resp: Send + Sync + Clone + 'static + std::fmt::Debug,
         {
             response_stream: Pin<Box<dyn Stream<Item = Annotated<BackendOutput>> + Send>>,
             response_generator: Box<dyn DeltaGeneratorExt<Resp>>,
-            tracker: Option<Arc<RequestTracker>>,
+            detokenize_metrics: DetokenizeMetricsGuard,
             context: Arc<dyn AsyncEngineContext>,
             cancelled: bool,
             cumulative_output_tokens: usize,
@@ -3171,26 +3229,11 @@ impl OpenAIPreprocessor {
             image_tokens: Option<usize>,
         }
 
-        impl<Resp> Drop for State<Resp>
-        where
-            Resp: Send + Sync + Clone + 'static + std::fmt::Debug,
-        {
-            fn drop(&mut self) {
-                let Some(tracker) = self.tracker.as_deref() else {
-                    return;
-                };
-                if let Some(total) = tracker.detokenize_total_latency() {
-                    DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
-                }
-                DETOKENIZE_TOKEN_COUNT.inc_by(tracker.detokenize_count() as f64);
-            }
-        }
-
         let tracker = generator.tracker();
         let state = State {
             response_stream: Box::pin(stream),
             response_generator: generator,
-            tracker,
+            detokenize_metrics: DetokenizeMetricsGuard { tracker },
             context: context.clone(),
             cancelled: false,
             cumulative_output_tokens: 0,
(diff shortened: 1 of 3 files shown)


lib/llm/src/preprocessor.rs
```diff
@@ -3088,6 +3079,15 @@ impl OpenAIPreprocessor {
                     // again. The stream is exhausted and will panic if polled after None.
                     inner.finished = true;
 
+                    // Flush per-request detokenize accumulators to global Prometheus counters
+                    // once, after the backend response stream completes.
+                    if let Some(t) = inner.response_generator.tracker().as_ref() {
```
jthomson04: This flush only runs when the backend stream returns `None`. Requests whose stream is dropped un-exhausted never reach it, so they now contribute nothing.

`monitor_for_disconnects` (`lib/llm/src/http/service/disconnect.rs:255-296`) `break`s out of its `select!` loop on `context.stopped()` (client disconnect), on the inactivity timeout, and on `Some(Err(..))`. Each `break` drops the inner stream, which drops this `unfold` before it ever yields `None`.

Concretely: a client aborts an SSE request after 500 tokens, and `detokenize_total_us` / `detokenize_token_count` get zero for those 500 tokens. Pre-PR the per-chunk update had already recorded them, and `State` has no `Drop` impl to compensate.

Fix: move the flush into `impl Drop for State` so it fires on every termination path.

lib/llm/src/preprocessor.rs
```diff
@@ -3088,6 +3079,15 @@ impl OpenAIPreprocessor {
                     // again. The stream is exhausted and will panic if polled after None.
                     inner.finished = true;
```
jthomson04, line 3378: Second uncovered exit: the internal cancellation path at line 2966 (not shown in this diff).

```rust
if inner.cancelled {
    tracing::debug!(...);
    // inner.finished = true; // Mark as finished
    return None;
}
```

`choice_from_postprocessor` failing on chunk N sets `inner.cancelled = true` (line 3012) and calls `stop_generating()`. If the backend delivers chunk N+1 before honoring the stop, the next poll takes the `if let Some(mut response)` arm, hits this guard, and returns `None` from inside that arm. The flush below is in the `else` arm, so it is skipped entirely.

Note `inner.finished` is deliberately left unset here (line 2965 is commented out), so this exit is structurally distinct from the one you instrumented. The old per-chunk flush covered it.

lib/llm/src/preprocessor.rs
```diff
@@ -3088,6 +3079,15 @@ impl OpenAIPreprocessor {
...
 
+                    // Flush per-request detokenize accumulators to global Prometheus counters
+                    // once, after the backend response stream completes.
+                    if let Some(t) = inner.response_generator.tracker().as_ref() {
+                        if let Some(total) = t.detokenize_total_latency() {
+                            DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
```
jthomson04: Third uncovered exit: the non-streaming path.

`DeltaAggregator::apply` (`lib/llm/src/protocols/openai/chat_completions/aggregator.rs:219-221`) does `.map(Annotated::into_data).try_fold(...)`. `try_fold` short-circuits on the first `Err`, so a backend error annotation mid-response returns immediately and drops this `unfold` un-exhausted, with no flush. Same for the completions handler at `lib/llm/src/http/service/openai.rs:686`.

Net effect: for aggregated (non-streaming) requests, anything that errors after producing tokens loses its detokenize metrics. Pre-PR they were recorded.

lib/llm/src/preprocessor.rs
```diff
@@ -3088,6 +3079,15 @@ impl OpenAIPreprocessor {
...
+                        }
+                        DETOKENIZE_TOKEN_COUNT.inc_by(t.detokenize_count() as f64);
+                    }
```
jthomson04: Altitude: this belongs in a `Drop`, not on one exit path.

`ResponseMetricCollector` already implements this exact pattern for these exact two values. `observe_tokenize_latencies` stores `detokenize_latency_total` / `detokenize_count_total` last-write-wins (`lib/llm/src/http/service/metrics.rs:1440-1446`) and publishes once in `Drop` (`metrics.rs:1544-1552`), which is robust to every termination path for free.

Either `impl Drop for State` here, or fold these two global counters into the existing collector's `Drop` and delete this block. As written, every future exit added to this unfold has to remember to duplicate the flush, which is how the three cases above got missed.

lib/llm/tests/test_streaming_usage.rs
```diff
@@ -272,6 +274,52 @@ async fn test_streaming_without_usage() {
     }
 }
 
+#[tokio::test]
+async fn test_detokenize_metrics_flush_once_at_stream_completion() {
```
jthomson04, line 314: This test only exercises the one path that still flushes.

Draining to completion is the case that works. The behavior this PR actually changes, i.e. stream dropped un-exhausted or `inner.cancelled`, has no coverage. Dropping `transformed_stream` after 2 of 3 chunks and asserting the counters would document the regression (and currently fail); likewise a case where `choice_from_postprocessor` errors.

Without it, nothing catches a future refactor that moves the flush again, and nothing records whether zeroing out abandoned requests is intended.

lib/llm/tests/test_streaming_usage.rs
```diff
@@ -272,6 +274,52 @@ async fn test_streaming_without_usage() {
...
+        false,
+        false,
+        None,
+        Default::default(),
+    );
+    futures::pin_mut!(transformed_stream);
+
+    for _ in 0..3 {
+        assert!(transformed_stream.next().await.is_some());
+        assert_eq!(DETOKENIZE_TOKEN_COUNT.get(), count_before);
```
jthomson04, line 343: Exact equality against process-global counters is fragile here.

`DETOKENIZE_TOKEN_COUNT` / `DETOKENIZE_TOTAL_US` are `Lazy<Counter>` process globals shared by all 11 tests in this binary, which cargo runs concurrently. This passes today only because every other test's tracker has `detokenize_count() == 0`, so their flush degenerates to `inc_by(0.0)`.

The first test anyone adds that calls `record_detokenize_latency` (directly, or via `backend.rs::step`) makes this line fail nondeterministically. A `#[serial]` guard or a test-local registry would make it deterministic.

lib/llm/src/preprocessor.rs
```diff
@@ -3060,15 +3060,6 @@ impl OpenAIPreprocessor {
                         );
```
jthomson04, line 3358: Hot-path note for a perf PR: the per-chunk work just above this line costs more than the two atomic increments removed.

Lines 3023-3032, per chunk:

```rust
let prefill_worker_type = tracker.as_ref().and_then(|t| t.prefill_worker_type()).map(String::from);
let decode_worker_type  = tracker.as_ref().and_then(|t| t.decode_worker_type()).map(String::from);
```

`RequestTracker::prefill_worker_type()` returns `Option<&'static str>` (`lib/llm/src/protocols/common/timing.rs:551`) and the value is fixed at routing time, so that is two malloc+free per token for a request-invariant `&'static str`.

Also `inner.response_generator.tracker()` is called at line 2957 (inside `drain_router_routing_data`) and again at line 3020, each an `Arc` clone+drop: four atomic RMWs per chunk.

Hoisting the tracker `Arc` and the two `&'static str` worker types into `State` once removes all of it.

lib/llm/src/preprocessor.rs
```diff
@@ -3060,15 +3060,6 @@ impl OpenAIPreprocessor {
...
-                    // Flush per-request detokenize accumulators to global Prometheus counters
-                    // (once per request instead of per-token).
-                    if let Some(t) = tracker.as_ref() {
-                        if let Some(total) = t.detokenize_total_latency() {
-                            DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
-                        }
-                        DETOKENIZE_TOKEN_COUNT.inc_by(t.detokenize_count() as f64);
-                    }
-
                     attach_llm_metrics(&mut response, llm_metrics);
```
jthomson04, line 3361: `attach_llm_metrics` takes the cheap typed-field path only when the response downcasts to `NvCreateChatCompletionStreamResponse` (lines 274-291). For `NvCreateCompletionResponse` it falls through to `attach_metrics_annotation` -> `metrics.to_annotation::<()>()`, i.e. a full serde_json serialization of the 18-field `LLMMetricAnnotation` on every streamed chunk.

The Tyche profile in the description covered chat only, so `/v1/completions` keeps a per-token JSON serialization that is far more expensive than the counter updates this PR removes. Worth the same typed-field treatment, or caching the serialization of the request-invariant fields.

lib/llm/src/preprocessor.rs
```diff
@@ -3088,6 +3079,15 @@ impl OpenAIPreprocessor {
...
+                    // once, after the backend response stream completes.
+                    if let Some(t) = inner.response_generator.tracker().as_ref() {
+                        if let Some(total) = t.detokenize_total_latency() {
+                            DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
+                        }
+                        DETOKENIZE_TOKEN_COUNT.inc_by(t.detokenize_count() as f64);
+                    }
+
                     if inner.finish_reason_sent && !inner.usage_chunk_sent {
                         inner.usage_chunk_sent = true;
```
jthomson04, line 3381: Minor: `inner.response_generator.tracker()` is now called twice in this branch, once at line 3084 for the flush and again at line 3096 for the `LLMMetricAnnotation`. Both resolve to `Some(self.tracker.clone())` (`lib/llm/src/protocols/openai/chat_completions/delta.rs:372`), so that is a redundant `Arc` clone+drop.

Hoist a single `let tracker = inner.response_generator.tracker();` above the flush and reuse it. It also makes explicit that both read the same snapshot.

jthomson04 (commented):
Reviewed the detokenization metric flush move.

**The core fix is correct.** The old code added the *cumulative* per-request accumulators to the global counters on every emitted chunk, so `detokenize_token_count` grew triangularly (~n^2/2 for an n-token request). The new site is genuinely reached exactly once per request on the normal path: `inner.finished = true` is set before it, and the `emit_payload_usage_chunk` + `pending_client_usage` two-item tail drains through the guard at the top of the closure rather than re-entering the `else` arm. Names, types and units are unchanged, and the derived `rate(total)/rate(count)` ratio is unaffected because both counters were over-counted by the same factor.

**The issue is exit coverage.** The `stream::unfold` has four ways out and the flush sits on only one:

| exit | flushes? |
| --- | --- |
| backend stream returns `None` (normal completion) | yes |
| `if inner.cancelled { return None }` (preprocessor error, line 2966) | no |
| stream dropped un-exhausted by `monitor_for_disconnects` (client disconnect / context stop / inactivity timeout / SSE error) | no |
| stream dropped un-exhausted by `try_fold` short-circuit on the non-streaming aggregation path | no |

Before this PR the per-chunk update covered all four, so "preserve ... incomplete-stream behavior" in the description is not accurate: abandoned and errored requests now contribute zero.

`impl Drop for State` closes all four at once for about six lines. That mirrors `ResponseMetricCollector::drop` (`lib/llm/src/http/service/metrics.rs:1544-1552`), which already implements this exact accumulate-per-request / publish-once-at-end pattern for these exact two values. It also means future exit paths added to this unfold don't each have to remember the flush.

The remaining comments are hot-path efficiency notes in the touched function plus two test notes.

_Anchoring note: GitHub only accepts inline comments inside a diff hunk. The comments on lines 3060, 3080, 3086 and 3092 discuss code at 2955-3032, 2960-2966, aggregator.rs, and 3096 respectively; each body names its real location._

lib/llm/src/preprocessor.rs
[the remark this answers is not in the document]
jthomson04: Fixed in 8603fceacc. The stream state caches one tracker handle and publishes its accumulated detokenization values from State::drop. This covers normal completion, cancellation, errors, and early consumer drop without an idempotency atomic. The new tests cover completion, early drop, and postprocessor cancellation.

lib/llm/src/preprocessor.rs
jthomson04: Fixed in 8603fceacc. State::drop now publishes the latest tracker totals when monitor_for_disconnects or any other downstream consumer drops the transformed stream.

lib/llm/src/preprocessor.rs
jthomson04: Fixed and covered by test_detokenize_metrics_flush_on_postprocessor_cancellation. The cancellation exit drops State and publishes the latest accumulated values once.

lib/llm/src/preprocessor.rs
jthomson04: Fixed by the State destructor. A try_fold short-circuit drops the same transformed stream state as the explicit early-drop test, so it publishes the latest accumulated values once without path-specific code.

lib/llm/src/preprocessor.rs
jthomson04: Implemented as Drop for State. The explicit normal-completion flush is removed, so every termination path uses one ownership-based publication point.

lib/llm/tests/test_streaming_usage.rs
jthomson04: Added serial tests for normal completion, dropping after two chunks, and the postprocessor cancellation exit. Each test also checks that later polling or drop does not publish twice.

lib/llm/tests/test_streaming_usage.rs
jthomson04: The three tests that read and modify the process-global detokenization counters now use serial_test::serial.

lib/llm/src/preprocessor.rs
jthomson04: The tracker Arc is now cached once in State and reused for routing data, per-chunk annotations, usage, and final publication. I am leaving worker-type String ownership out of this PR: LLMMetricAnnotation owns Option<String>, and standalone-router routing data can populate the tracker after State is created. Removing those allocations needs a separate annotation or first-frame design change.

lib/llm/src/preprocessor.rs
jthomson04: This is valid but independent of the chat-completions profile and the detokenization counter defect. A typed completions path requires an internal field plus updates to the completion response constructors and consumers. I am deferring it to a separate focused change.

lib/llm/src/preprocessor.rs
jthomson04: Fixed. State now owns one cached tracker Arc, so the completion branch and usage annotation reuse the same tracker without another clone.

lib/llm/src/preprocessor.rs
```diff
@@ -3154,17 +3169,15 @@ impl OpenAIPreprocessor {
...
                         .map(String::from);
-                    let decode_worker_id = tracker.as_ref().and_then(|t| t.decode_worker_id());
-                    let decode_dp_rank = tracker.as_ref().and_then(|t| t.decode_dp_rank());
+                    let decode_worker_id = tracker.and_then(|t| t.decode_worker_id());
+                    let decode_dp_rank = tracker.and_then(|t| t.decode_dp_rank());
                     let decode_worker_type = tracker
-                        .as_ref()
                         .and_then(|t| t.decode_worker_type())
                         .map(String::from);
                     let llm_metrics = LLMMetricAnnotation {
```
MatejKosec: This block and the usage-chunk block at 3251 now build `LLMMetricAnnotation` from the same six tracker fields, with 10 of the 14 fields identical. Only `input_tokens`, `output_tokens`, `chunk_tokens` and `cached_tokens` genuinely differ between them.

This PR already had to edit both blocks in lockstep to switch them to `inner.tracker`, which is the cost showing up. The next person who adds a tracker-sourced field has to remember both places, and if they update only one the streaming annotation and the final usage annotation disagree, with no compile error and no failing test.

A helper taking `Option<&RequestTracker>` plus those four values would collapse the two into one.

lib/llm/src/preprocessor.rs
```diff
@@ -3046,9 +3047,26 @@ impl OpenAIPreprocessor {
...
+                    return;
+                };
+                if let Some(total) = tracker.detokenize_total_latency() {
+                    DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
+                }
+                DETOKENIZE_TOKEN_COUNT.inc_by(tracker.detokenize_count() as f64);
+            }
+        }
+
+        let tracker = generator.tracker();
```
MatejKosec, line 3232: `generator.tracker()` is read once here and the `Option` is stored, so `State.tracker` is fixed for the life of the stream, and `Drop` returns at 3055 whenever it is `None`.

`DeltaGeneratorExt::tracker()` defaults to returning `None` (`lib/llm/src/protocols/openai.rs:332`). Both current implementors override it with `Some(self.tracker.clone())` over a non-optional `Arc` — `chat_completions/delta.rs:376` and `completions/delta.rs:320` — so no request loses metrics today and this is not a live bug.

The trap is for a third implementor that inherits the default. Every request through it would contribute zero to `DETOKENIZE_TOKEN_COUNT` and `DETOKENIZE_TOTAL_US` for as long as it existed, and nothing would fail. Calling `self.response_generator.tracker()` inside `drop` instead costs one virtual call per request and removes the failure mode; `tracker()` takes `&self`, which is reachable from the `&mut self` of `drop`. One of the three independent reimplementations chose exactly that, on the grounds that capturing at construction freezes an `Option` that could differ at teardown.

lib/llm/src/preprocessor.rs
```diff
@@ -3046,9 +3047,26 @@ impl OpenAIPreprocessor {
             image_tokens: Option<usize>,
         }
 
+        impl<Resp> Drop for State<Resp>
```
MatejKosec: Implementing `Drop` on `State` makes partial moves out of it a compile error (E0509) throughout the roughly 280-line closure below. Nothing there does that today, so this compiles, but the constraint is now invisible: whoever next writes `let x = inner.some_field;` inside that closure gets an error whose cause is thirty lines away and not obvious.

Either a one-line comment here saying so, or moving the two increments into a small guard struct held as a `State` field, would fix that. Two of the three independent reimplementations took the guard-struct route and never met the constraint at all.

lib/llm/src/preprocessor.rs
```diff
@@ -3046,9 +3047,26 @@ impl OpenAIPreprocessor {
...
+        where
+            Resp: Send + Sync + Clone + 'static + std::fmt::Debug,
+        {
+            fn drop(&mut self) {
+                let Some(tracker) = self.tracker.as_deref() else {
+                    return;
+                };
+                if let Some(total) = tracker.detokenize_total_latency() {
+                    DETOKENIZE_TOTAL_US.inc_by(total.as_micros() as f64);
```
MatejKosec: Optional, and inherited rather than introduced here. `as_micros()` floors, so a request whose detokenize total is 1500 ns publishes 1 µs, and one totalling 800 ns publishes 0.

This is the same expression as the two blocks being removed below, and the change makes accuracy far better rather than worse: the error goes from a floor applied on every chunk against a running cumulative total to a single floor per request, bounded under 1 µs no matter how many tokens.

Since the increment now happens exactly once, `total.as_nanos() as f64 / 1_000.0` would keep the fractional microsecond for free. Entirely up to you whether it belongs in this PR.

lib/llm/tests/test_streaming_usage.rs
```diff
@@ -272,6 +274,145 @@ async fn test_streaming_without_usage() {
...
+        Annotated::from_data(output)
+    }));
+    let ctx = Arc::new(MockContext::new());
+    let backend_stream = dynamo_runtime::engine::ResponseStream::new(Box::pin(stream), ctx.clone());
+
+    let transformed_stream = OpenAIPreprocessor::transform_postprocessor_stream(
+        backend_stream,
+        Box::new(response_generator),
+        ctx,
+        false,
```
MatejKosec, line 334: All three new tests pass `false` for `emit_payload_usage_chunk`, so nothing covers the `true` path.

That is the gap worth closing, because one of the two flush blocks this PR deletes lived inside that branch — the one at old `preprocessor.rs:3281-3288`, in the usage-chunk path, structurally distinct from the per-chunk one. The branch whose flush actually changed shape is the branch with no test behind it.

A fourth test setting this to `true` and asserting the same counter deltas would cover it, and would cost about the same nine lines of setup the other three already spend.

lib/llm/tests/test_streaming_usage.rs
```diff
@@ -272,6 +274,145 @@ async fn test_streaming_without_usage() {
...
+    let total_us_before = DETOKENIZE_TOTAL_US.get();
+
+    let tracked = tracker.clone();
+    let outputs = build_backend_outputs_with_cached_tokens(None);
+    let stream = stream::iter(outputs.into_iter().map(move |output| {
+        tracked.record_detokenize_latency(Duration::from_micros(10));
+        Annotated::from_data(output)
+    }));
+    let ctx = Arc::new(MockContext::new());
+    let backend_stream = dynamo_runtime::engine::ResponseStream::new(Box::pin(stream), ctx.clone());
```
MatejKosec: These lines repeat verbatim in all three new tests — here, at 339-346, and at 379-387. The existing `create_mock_backend_stream` helper at line 89 does not fit because it never records detokenize latency.

A helper taking the outputs and a `Duration` and returning the tracker, backend stream and context would remove the repetition, and would make the missing `emit_payload_usage_chunk = true` test cheap to add.

lib/llm/tests/test_streaming_usage.rs
```diff
@@ -272,6 +274,145 @@ async fn test_streaming_without_usage() {
...
+    assert!(transformed_stream.next().await.is_some());
+    let error = transformed_stream.next().await.expect("error response");
+    assert!(error.is_error());
+    assert_eq!(DETOKENIZE_TOKEN_COUNT.get(), count_before);
+    assert_eq!(DETOKENIZE_TOTAL_US.get(), total_us_before);
+
+    // The next backend item observes the cancellation and drops the state.
+    assert!(transformed_stream.next().await.is_none());
+    assert!(ctx.is_stopped());
+    assert_eq!(DETOKENIZE_TOKEN_COUNT.get() - count_before, 3.0);
```
MatejKosec, line 512: This `3.0` holds only because the `unfold` closure pulls `outputs[2]` from the backend iterator before it reads `inner.cancelled`, so `record_detokenize_latency` fires a third time even though the request is already cancelled.

Move that check before the pull — a reasonable latency optimisation someone may well make — and the delta becomes 2.0. The test then fails saying a counter has the wrong value, when what actually changed is how many items get pulled after cancellation, so the message points away from the cause.

Counting the items the mock stream actually yields and asserting the counter equals that count would test the property the test is named for, and would keep passing under that refactor.

MatejKosec (commented):
I reviewed this two ways: an ordinary read of the diff, and a blind reimplementation. For the second, three agents fixed the same defect from a symptom-only brief — the tracker holds cumulative totals, the old code adds them to the counters once per chunk, here are the five exit paths the fix must survive — without seeing this PR, and their designs were then compared against it.

All three converged on publishing from `Drop`. The three started from different priors (smallest possible diff, exhaustive enumeration of every stream-termination path, and the invariants a Prometheus counter has to satisfy) and all landed on RAII teardown as the single point every request passes through. Reading a diff can tell you a mechanism works; it cannot tell you it was the best one available, so that agreement seemed worth stating.

I also checked the invariant the description rests on. The description says the change "does not need an idempotency flag because each state is dropped once", and one `State` dropped once is a weaker property than one tracker published once, so I traced the difference. `DeltaGenerator::new` gets its tracker from `delta_common::initial_state()`, which takes no arguments and creates a fresh `Arc<RequestTracker>`, so a tracker cannot be passed in from outside. The generator moves by value into `State`. `generate()` builds one generator per request, at `preprocessor.rs:4343` and `:4524`. One request therefore gives one tracker, one `State`, one drop, and after this change both counters have exactly one increment site in the repository. The stronger property holds, including under retry, since a retry that re-enters `generate()` builds a new tracker as well.

The inline comments cover the rest. Two are worth acting on before merge: the duplicated `LLMMetricAnnotation` construction, which this PR already had to edit in two places, and the absence of any test that sets `emit_payload_usage_chunk = true`, which is the branch whose flush block this PR restructured. The other four are smaller, and one is an inherited nit rather than anything this change introduces.

None of the six is a correctness defect I could confirm against the code. Two claims from the blind comparison did not survive checking and I have left them out.

lib/llm/src/preprocessor.rs
jthomson04: I kept `as_micros()` unchanged. This is inherited metric precision, and changing it would expand the PR beyond the existing metric semantics. I left this optional thread open for your confirmation.

lib/llm/src/preprocessor.rs
jthomson04: I kept the cached tracker intentionally. Both current generators expose one stable per-request tracker, and caching removes the prior per-chunk `Arc` clone. Calling `tracker()` again at teardown cannot recover metrics from an implementation that returns `None`; it only delays the same decision and adds another clone. I clarified the trait documentation so implementations that own timing data must override the method. Making the method required would be a public trait break, so I left this thread open for your confirmation.

MatejKosec (approved):

ishandhanani (approved):

PR #13000, opened 2026-08-11
fix: surface image generation errors instead of returning empty HTTP 200

marckarp:
## Overview

A failed `/v1/images/generations` request returns **HTTP 200 with `data: []` and no error message**. The client has no way to tell a failed generation from an empty one. Easy to reproduce: send any request that makes the worker raise (e.g. an out-of-range parameter or a backend failure) — the response is a successful-looking empty body.

Two layers cause it:

1. **Worker (Python):** the SGLang image diffusion handler catches every exception and yields `{"data": [], "error": str(e)}`. The OpenAI `ImagesResponse` schema has no `error` field, so the message is silently dropped during deserialization and the frontend returns HTTP 200 with empty data.
2. **Frontend (Rust):** even when an exception does propagate as an error event on the response stream, the images route discards it and returns a hardcoded generic 500 (`"Failed to fold images stream"`), so the worker's message is only visible in server logs.

## What this PR changes

**Handler** (`image_diffusion_handler.py`):
- Remove the catch-and-yield: exceptions now propagate. The runtime converts a raised exception into an error event on the response stream (`Annotated::from_err`), which the frontend folds into a non-200 HTTP error.
- Request-validation failures (blank `input_reference`, malformed `size`) now raise the typed `InvalidArgument` instead of bare `ValueError`, matching the existing pattern in `sglang_processor.py`. `InvalidArgument` maps to HTTP 400 with the message visible to the client; other exceptions remain sanitized 500s.

**Frontend** (`lib/llm/src/http/service/openai.rs`):
- Route the images stream fold error through `ErrorMessage::from_anyhow` instead of the hardcoded generic 500. This follows the route's existing sanitization policy: `InvalidArgument` → 400 with message, internal errors → sanitized 500 (details stay server-side).

**Tests**: the unit test that asserted the old swallow behavior now asserts propagation; added tests for the `InvalidArgument` validation paths.

## Verification

- **Unit tests: 23/23 pass** inside `sglang-runtime:1.3.0` with the patched handler (the 8 behavior-dependent tests fail against the stock handler, confirming they detect the old behavior).
- **Live e2e** on a single-GPU FLUX.1-schnell deployment with the patched worker:
  - valid request → HTTP 200 with image (no regression)
  - worker-side failure (blank `input_reference`) → **HTTP 500** instead of the old empty HTTP 200
- **Rust half**: `cargo check` introduces no new errors. A live 400-with-message demonstration requires a rebuilt frontend image (the released frontend predates this change), so that path relies on CI + the unit-level guarantees of `from_anyhow` (which existing chat/completions routes already use).

## Notes for reviewers

- The `_parse_size` guard is currently unreachable from HTTP because the frontend's `size` field is a fixed enum that rejects free-form values first. It becomes load-bearing with the planned change to accept arbitrary `WxH` sizes (needed for diffusion models; the videos route already takes a free-form string), and is worth having defensively regardless.
- Soft overlap with #12970: both touch `test_sglang_image_diffusion_handler.py` in different places; whichever merges second rebases trivially. #12970's live test also demonstrated this bug: requesting `n=11` returned `HTTP 200` with `data=[]` and an empty error on the wire.

Part of a series of fixes for the SGLang image diffusion path (see #12970 for the first).

Part of DYN-3974

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid image generation inputs, including malformed sizes and blank references, now return clear HTTP 400 errors.
  * Image diffusion failures now propagate correctly instead of producing empty or invalid responses.
  * Internal server failures continue to return sanitized HTTP 500 responses, while preserving appropriate client validation errors.


jain-ria (approved):

rmccorm4:
/ok to test [2a7a86f](https://github.com/ai-dynamo/dynamo/pull/13000/commits/2a7a86fcbbdee3157bc6762ced804c64ad63b8f4)

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
```diff
@@ -228,8 +231,15 @@ def _parse_size(self, size_str: Optional[str]) -> tuple[int, int]:
...
 
-        w, h = size_str.split("x")
-        return int(w), int(h)
+        try:
+            w, h = size_str.split("x")
+            return int(w), int(h)
+        except ValueError as e:
+            # InvalidArgument maps to HTTP 400 with the message visible to
+            # the client, instead of a sanitized 500.
+            raise InvalidArgument(
```
rmccorm4, line 234: The frontend protocol accepts and forwards `size="auto"`, but this parser converts it into `InvalidArgument` and the PR now returns HTTP 400. Treat `"auto"` like the backend default and add coverage.

components/src/dynamo/sglang/tests/test_sglang_image_diffusion_handler.py
```diff
@@ -258,17 +272,46 @@ async def test_generate_error_handling(self, handler, mock_context):
             },
```
rmccorm4, line 277: This direct generator test never exercises the annotated transport, the images stream fold, or the HTTP status/body. Reverting the Rust route hunk leaves it green; add one CPU HTTP-level regression.

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
```diff
@@ -144,13 +145,15 @@ async def generate(
...
+            # OpenAI ImagesResponse schema has no `error` field and the
+            # message was silently dropped during deserialization.
             logger.error(f"Error in diffusion generation: {e}", exc_info=True)
-            error_response = {
-                "created": int(time.time()),
-                "data": [],
-                "error": str(e),
-            }
-            yield error_response
+            raise
```
rmccorm4: `InvalidArgument` is an expected 400, but this catch emits a full error traceback before rethrowing. The runtime already transports the exception; remove this catch-and-rethrow.

lib/llm/src/http/service/openai.rs
```diff
@@ -4370,7 +4370,14 @@ async fn images(
...
             tracing::error!("Failed to fold images stream for {}: {:?}", request_id, e);
-            let err_response = ErrorMessage::internal_server_error("Failed to fold images stream");
+            // Route the stream error through from_anyhow so typed errors keep
+            // their semantics: an InvalidArgument raised by the worker (e.g.
+            // request validation) surfaces as HTTP 400 with its message,
+            // while internal errors remain sanitized 500s. Previously every
+            // stream error collapsed into a generic 500 and the worker's
+            // message was only visible in server logs.
+            let err_response =
+                ErrorMessage::from_anyhow(anyhow::Error::new(e), "Failed to generate images");
```
rmccorm4, line 4379: The unconditional `tracing::error!` immediately above logs every stream failure at error level before `from_anyhow` identifies expected 400s. Internal failures are then logged again downstream. Remove the pre-classification event or emit only after classification.

components/src/dynamo/sglang/tests/test_sglang_image_diffusion_handler.py
marckarp: Added in 81d8ca04c6: two HTTP-level regression tests in `lib/llm/tests/http-service.rs` — a mock images engine emits an error annotation through the real HTTP service (transport → fold → from_anyhow → status/body). One asserts a worker InvalidArgument folds to 400 with the worker's message; one asserts an internal error folds to a sanitized 500 with no detail leakage. Verified both ways: with the fix, 2/2 pass; with the fold hunk surgically reverted to the previous hardcoded 500, the 400 test fails (0 passed, 1 failed) — so a revert no longer leaves CI green.

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
marckarp: Removed in 81d8ca04c6 — the try/except is gone entirely; the runtime transports the raised exception either way and expected 400s no longer print tracebacks. (This also resolves the CodeRabbit lazy-logging comment: the line it targeted no longer exists.)

lib/llm/src/http/service/openai.rs
marckarp: Removed in 81d8ca04c6 — no pre-classification logging in the images fold; classification via from_anyhow decides the level, and the sanitization path logs internals downstream as you noted. Expected 400s no longer appear at error level.

components/src/dynamo/sglang/request_handlers/image_diffusion/image_diffusion_handler.py
marckarp: Good catch — fixed in 81d8ca04c6: `size="auto"` now behaves like an omitted size (backend default) instead of an InvalidArgument 400. Unit test added, and verified live: a served request with `size="auto"` returns HTTP 200 with an image.

marckarp:
Note on the red `lychee` check: it's inherited from `main`, not this PR — `deploy/operator/docs/footer.md` and the generated Kubernetes API reference pages link to `deploy/operator/internal/checkpoint/resource.go`, which no longer exists on `main` (the checkpoint code was reorganized; `compatibility.go` remains). Any branch carrying recent `main` fails the same way. Fix belongs on `main`: regenerate the operator API docs.

rmccorm4 (approved):

rmccorm4:
/ok to test [3123bb9](https://github.com/ai-dynamo/dynamo/pull/13000/commits/3123bb957ede6f9b4116bd16a01f1393fc7c64ab)

PR #13012, opened 2026-08-11
feat(agents): propagate Codex compaction metadata

ishandhanani:
Codex compaction requests now preserve turn metadata in canonical `AgentContext.compaction` and request traces. This makes compaction calls observable through existing agent metadata without changing session identity, routing, or KV behavior.

### How This Was Implemented

- Parse `x-codex-turn-metadata` only when a valid Codex `thread-id` is present and `request_kind` is `compaction`.
- Map the optional Codex `trigger`, `reason`, `implementation`, `phase`, and `strategy` values into `AgentCompaction`.
- Carry compaction through the existing `AgentContext`, `PreprocessedRequest`, and request-trace path.
- Preserve canonical Dynamo session-header precedence for identity while retaining Codex compaction metadata independently.
- Add compaction to `request_end` traces and document the supported signal and passive semantics.

<details>

#### Mental model

The metadata marks the one inference that creates a compacted summary. Dynamo carries it with the main agent context but does not ask a worker to compact anything.

```mermaid
flowchart LR
    A["Codex turn metadata"] --> B["HTTP agent-header parser"]
    B --> C["AgentContext.compaction"]
    C --> D["PipelineContext"]
    D --> E["PreprocessedRequest.agent_context"]
    E --> F["Worker request"]
    E --> G["Request trace"]
```

#### State and ordering

Codex is identified by a valid `thread-id`; malformed metadata and non-compaction request kinds leave the compaction field unset. If `x-dynamo-session-id` is also present, it selects `session_id`, while the Codex compaction marker remains attached to `AgentContext`.

#### Request lifecycle

HTTP ingress parses the Codex header into `AgentContext`. Preprocessing forwards that context through `PreprocessedRequest.agent_context`, and request tracing emits the same object.

Workers deserialize compaction metadata but do not act on it. This change adds no server-side compaction policy, context epoch, routing action, or KV-cache action, and it does not implement Anthropic server compaction or the Codex `/responses/compact` endpoint.

</details>

### Validation

- `git diff --check`
- `cargo test -p dynamo-llm compaction --lib -- --nocapture` (5 passed)
- `cargo test -p dynamo-llm agent_context --lib -- --nocapture` (22 passed)
- `cargo clippy -p dynamo-llm --tests --no-deps -- -D warnings`
- `fern check` (0 errors)

## Related Issues

Depends on #12331.


MatejKosec (commented):
Overall this is a clean, well-structured PR with solid test coverage. Two things worth addressing before merge — details inline.

lib/llm/src/protocols/agents.rs
```diff
@@ -59,13 +70,15 @@ fn header_value(headers: &HeaderMap, header_name: &str) -> Option<String> {
 
 pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentContextHeaderValues> {
     let session_final = header_bool(headers, HEADER_DYNAMO_SESSION_FINAL);
+    let compaction = codex_compaction_header_value(headers);
```
MatejKosec: **Compaction attaches to all harnesses, not just Codex.**

`codex_compaction_header_value` is called here before the harness is identified, so its result flows into every downstream branch — Dynamo native, Claude Code, and OpenCode included. The docs say only Codex carries this mapping and the others leave it unset, but any request that arrives with `x-codex-turn-metadata: {"request_kind":"compaction",...}` will get `compaction` set regardless of harness.

The extensions test even demonstrates this: it adds `HEADER_DYNAMO_SESSION_ID` alongside the Codex metadata header and asserts `compaction.is_some()`, which confirms the behavior is consistent but also confirms it crosses harness boundaries.

Low real-world risk today since Claude Code and OpenCode don't send this header. But if the intent is truly Codex-only, the call should move inside the Codex branch (lines 96–113). If any harness should be able to signal compaction via this header, the docs table should say so.

lib/llm/src/protocols/common/extensions.rs
```diff
@@ -85,6 +113,11 @@ pub struct AgentContext {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub session_final: Option<bool>,
 
+    /// Present when the current inference creates a compacted session summary.
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub compaction: Option<AgentCompaction>,
```
MatejKosec, line 131: **Rolling-upgrade safety: compaction should go on `PreprocessedRequest`, not inside `AgentContext`.**

The comment at line 98 correctly identifies the N-2 compatibility constraint, but the fix applied here — removing `deny_unknown_fields` and emitting the new field in the same release — does not protect the current window. N-1 and N-2 workers still have `deny_unknown_fields` compiled into their `AgentContext` deserializer. A newly upgraded frontend that receives a Codex compaction header will write `compaction` into the envelope; any older worker will reject it with `unknown field `compaction`` and drop the request.

The two-release sequencing fix (remove the attribute now, emit the field next release) works but forces a gap. There is a one-release alternative: put `codex_compaction` as a top-level field on `PreprocessedRequest` instead of nested inside `AgentContext`.

```rust
// In PreprocessedRequest, not in AgentContext:
#[builder(default)]
#[serde(default, skip_serializing_if = "Option::is_none")]
pub codex_compaction: Option<AgentCompaction>,
```

`PreprocessedRequest` has never had `deny_unknown_fields`, so old workers already ignore unknown top-level fields. The new field is absent on non-compaction requests (`skip_serializing_if`), present only on compaction turns, and defaults to `None` on old senders. No sequencing required, `AgentContext` stays unchanged, and the `deny_unknown_fields` comment at line 98 can be removed rather than worked around.

lib/llm/src/protocols/agents.rs
MatejKosec: **Parse turn metadata only after confirming this is a Codex session.**

`codex_compaction_header_value` is called here before the harness is identified. All downstream branches — Dynamo native, Claude Code, OpenCode, and Codex — receive the result. If any non-Codex client forwards `x-codex-turn-metadata: {"request_kind":"compaction",...}`, they get `compaction` set, which contradicts the docs (Claude Code and OpenCode are listed as having no mapping).

Move the call inside the Codex branch, after `HEADER_CODEX_THREAD_ID` is confirmed present:

```rust
if let Some(session_id) = header_value(headers, HEADER_CODEX_THREAD_ID) {
    let compaction = codex_compaction_header_value(headers); // only here
    return Some(AgentContextHeaderValues {
        parent_session_id: ...,
        session_id,
        session_final,
        compaction,
    });
}
```

This is a guard-before-parse pattern that the existing Claude Code and OpenCode branches already follow for their own headers.

MatejKosec (changes requested):
Three issues from a parallel from-scratch implementation that suggest the current approach has structural problems worth addressing before merge.

**1. Rolling-upgrade safety (inline on extensions.rs)** — covered by an inline comment below. The current approach requires a two-release sequence to be safe; there is an alternative that is safe in a single release.

**2. Harness guard before parsing (inline on agents.rs)** — covered below. The turn-metadata header is currently parsed for every request regardless of harness, before the harness is identified.

**3. Missing copy_context_metadata propagation** — not in the diff, but `copy_context_metadata` in `lib/llm/src/http/service/openai.rs` (and the equivalent in `anthropic.rs`) copies the pipeline context from one stage to another. It already propagates `AgentContext` and `SessionAffinityId` by key. The new `compaction` field lives inside `AgentContext` in this PR, so it rides along for free — but only because it is nested inside the already-copied struct. If the approach changes to a separate context key (as suggested in finding 1), `copy_context_metadata` would need an explicit entry. Worth verifying that the current nesting means no copy-propagation gap exists, and adding a comment or test to lock that in.

lib/llm/src/protocols/agents.rs
ishandhanani: [AI] Fixed in 262d88bead. Compaction parsing now runs only for the selected Codex mapping, while Dynamo, Claude Code, and OpenCode leave it unset.

lib/llm/src/protocols/common/extensions.rs
ishandhanani: [AI] Fixed in 262d88bead. Compaction no longer crosses the frontend-to-worker wire, and request tracing reads it from frontend pipeline context.

lib/llm/src/protocols/agents.rs
ishandhanani: [AI] Fixed in 262d88bead. Compaction parsing now runs only for the selected Codex mapping, while Dynamo, Claude Code, and OpenCode leave it unset.

lib/llm/src/protocols/agents.rs
```diff
@@ -88,16 +100,27 @@ pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentCo
                 (mapping.infer_parent_from_session_for_child && session_id != root_session_id)
                     .then(|| root_session_id.clone())
             });
+        let compaction = (mapping.root_session_header == HEADER_CODEX_THREAD_ID)
+            .then(|| codex_compaction_header_value(headers))
+            .flatten();
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 28ef4d07db. Compaction parsing now requires the Codex `thread-id` but is independent of which header selects session identity, so canonical `x-dynamo-session-id` keeps precedence without dropping the trace marker.

lib/llm/src/request_trace/agent_context.rs
```diff
@@ -246,7 +247,12 @@ pub(crate) fn build_agent_context_trace_state(
     tracker: &Option<Arc<RequestTracker>>,
     context: &Context<()>,
 ) -> Option<AgentContextTraceState> {
-    let agent_context = common_request.agent_context.clone()?;
+    let compaction = context
+        .get::<AgentCompaction>(CODEX_COMPACTION_CONTEXT_KEY)
+        .ok()
+        .map(|compaction| compaction.as_ref().clone());
+    let agent_context =
+        RequestTraceAgentContext::new(common_request.agent_context.clone()?, compaction);
```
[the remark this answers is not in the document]
ishandhanani: Confirmed: frontend-only is intended; workers do not consume compaction metadata. The PR description and flowchart now show identity flowing through `PreprocessedRequest`, while compaction flows from frontend pipeline context to the request trace.

lib/llm/src/protocols/common/extensions.rs
```diff
@@ -80,11 +78,35 @@ pub enum InputTrigger {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub strategy: Option<String>,
+}
+
 /// Identity metadata for agentic workloads.
-// Not `deny_unknown_fields`: `AgentContext` is part of the frontend->worker wire
-// format (`PreprocessedRequest.agent_context`), so additive fields must be tolerated
-// across the N-2 mixed-version compatibility window.
 #[derive(Serialize, Deserialize, Builder, Debug, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
```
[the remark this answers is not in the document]
ishandhanani: Fixed in bbd7ac603b. Restored `main`’s tolerant `AgentContext` reader and added a regression test that nested compaction is ignored; compaction remains frontend-only.

docs/fern/pages/use-cases/agents/agent-harnesses.mdx
```diff
@@ -205,6 +205,20 @@ Use the repository's `.agents/skills/dynamo-agent-harness` skill when an agent n
...
+
+| Harness | Signal recognized by Dynamo | Normalized agent context |
+| --- | --- | --- |
+| Codex | `x-codex-turn-metadata` with `request_kind` set to `compaction` | `agent_context.compaction` |
+| Claude Code | No supported mapping | Unset |
+| OpenCode | No supported mapping | Unset |
+
+For Codex, Dynamo preserves the optional `trigger`, `reason`, `implementation`, `phase`, and `strategy` values. The object appears only on the inference request that creates the summary. Invalid JSON and other request kinds leave it unset.
+
+Compaction metadata is passive. It does not rotate `session_id`, change request routing, or evict KV cache blocks. See [Agent Tracing](agent-tracing.md#compaction-metadata) for the request-trace format.
```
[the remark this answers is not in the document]
ishandhanani: Fixed in bbd7ac603b. The table now identifies `request_end.agent_context.compaction` and explicitly states that compaction is not serialized in `PreprocessedRequest.agent_context` or sent to workers.

lib/llm/src/request_trace/agent_context.rs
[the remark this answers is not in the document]
ishandhanani: Updated in 091d4eb166 after design clarification: compaction now lives in the canonical `AgentContext` and travels through `PreprocessedRequest.agent_context`. Workers receive the field but do not act on it, and request traces emit the same object.

docs/fern/pages/use-cases/agents/agent-harnesses.mdx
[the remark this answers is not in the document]
ishandhanani: Superseded by the final design in 091d4eb166: compaction is intentionally part of canonical `AgentContext` and the frontend-to-worker wire. The docs now state that workers receive the field but do not act on it.

lib/llm/src/protocols/common/extensions.rs
ishandhanani: Follow-up in 091d4eb166: compaction is restored to canonical `AgentContext` by design, and workers receive but do not consume it. `AgentContext` remains a tolerant reader for future additive fields; we are accepting the current older-reader limitation for this metadata.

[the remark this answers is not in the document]
ishandhanani: Superseded by the final design in 091d4eb166: compaction is intentionally carried in canonical `AgentContext`; workers receive the metadata but do not act on it. The current older-reader limitation is accepted for this field, while new `AgentContext` readers remain tolerant of future additive fields.

lib/llm/src/protocols/agents.rs
```diff
@@ -59,13 +70,16 @@ fn header_value(headers: &HeaderMap, header_name: &str) -> Option<String> {
 
 pub(crate) fn agent_context_header_values(headers: &HeaderMap) -> Option<AgentContextHeaderValues> {
     let session_final = header_bool(headers, HEADER_DYNAMO_SESSION_FINAL);
+    let compaction = header_value(headers, HEADER_CODEX_THREAD_ID)
```
jthomson04: P3: `header_value` allocates a `String` only to check `thread-id`, then the mapping allocates it again. The compaction parser also copies its JSON header only to pass a borrow to serde. Add a borrowed-header helper here and allocate only when a value is retained.

jthomson04 (approved):
One non-blocking P3: avoid the extra header copies in the Codex compaction path.

PeaBrane (approved):

MatejKosec (approved):

PR #13020, opened 2026-08-11
fix(profiler): Planner-Only Gating That Skips Mocker Thorough Interpolation

weizhoublue:
## Overview:

Fix Planner-Only Gating That Skips Mocker Thorough Interpolation

## Details:

PR #12573 introduced ·needs_profile_data(dgdr)` as unified gate for determining whether file-based profiling data needs to be generated, covering both Mocker and Planner scenarios.

For mocker-only DGDR with `spec.searchStrategy: thorough` , `needs_profile_data(dgdr)`  returns `True`, causing `profile_sla.py` to call `run_interpolation()`.

However, `run_interpolation()`  still retained old Planner-only gate and checked only `spec.features.planner.pre_deployment_sweeping_mode`. 
Because no Planner exists in mocker-only configuration, sweep mode defaults to `None`, causing function to return early. 

As result, NPZ data required by Mocker thorough mode is never generated.

## Related Issues

- Closes #XXXX

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profiling behavior for rapid-mode runs when GPU profile data is unnecessary.
  * Prevented unnecessary real-GPU interpolation for mocker-only deployments.
  * Preserved interpolation for thorough-mode runs.
* **Tests**
  * Added coverage for rapid- and thorough-mode interpolation behavior.


weizhoublue:
@julienmancuso hi, could you help review whether this follow-up PR for #12573 makes sense , thanks

components/src/dynamo/profiler/interpolation.py
```diff
@@ -98,18 +98,10 @@ async def run_interpolation(
...
     """
-    planner_cfg = (
-        dgdr.features.planner if (dgdr.features and dgdr.features.planner) else None
-    )
-    sweep_mode = PlannerPreDeploymentSweepMode.None_
-    if planner_cfg and planner_cfg.pre_deployment_sweeping_mode:
-        sweep_mode = planner_cfg.pre_deployment_sweeping_mode
-
-    if sweep_mode != PlannerPreDeploymentSweepMode.Thorough:
+    if not needs_profile_data(dgdr):
```
jasonqinzhou, line 102: **P2 — Keep mocker-only failure guidance actionable**

This gate now intentionally admits mocker-only `searchStrategy: thorough` requests. If either interpolation deployment fails, `_wait_for_required_interpolation_deployment()` still tells the user to set `pre_deployment_sweeping_mode='rapid'`, but a mocker-only DGDR has no Planner config; its supported escape hatch is `searchStrategy: rapid`. Please make that remediation generic or select the field based on whether Planner is configured, and update the module/function's Planner-only wording so the path fixed here is described accurately.

jasonqinzhou (approved):
**COMMENT · 8.7/10 · high confidence**

## What this PR does

This PR fixes mocker-only thorough profiling by making `run_interpolation()` use the same `needs_profile_data(dgdr)` eligibility predicate as the caller, so a DGDR with `searchStrategy: thorough` proceeds to real-GPU NPZ generation instead of being incorrectly skipped by a Planner-only gate.

Reusing the shared profile-data predicate removes the split-brain decision cleanly, and the focused tests exercise both sides of the newly relevant mocker-only contract: rapid still skips while thorough crosses the interpolation gate.

## Why this score

8.7/10, COMMENT. The fix is narrow, consistent with the surrounding profiler pipeline, and passed the complete changed test file plus static checks. One non-blocking user-facing issue remains: mocker-only thorough requests newly admitted by this gate still receive Planner-only remediation if interpolation deployment fails, so the error guidance and planner-only docstrings should be generalized for the path this PR fixes.

components/src/dynamo/profiler/interpolation.py
weizhoublue: ok
updated

PR #13027, opened 2026-08-11
feat(frontend): warn on sustained Tokio queue pressure

jthomson04:
## Summary

- detect sustained Tokio runnable-queue pressure in the existing runtime metrics loop
- use worker-scaled aggregate thresholds, hot-worker detection, hysteresis, and sample-gap checks
- warn once per overload episode, with five-minute repeats during a continuing episode
- export emitted warnings as `dynamo_tokio_queue_overload_warnings_total`

## Overview:

This PR adds an operator warning when a Tokio runtime may be overloaded. Queue depth does not prove CPU saturation. It shows that runnable work has remained queued.

## Details:

- poll Tokio scheduler metrics once per second
- enter overload at a total queue depth of `4 × worker count`, or when one worker-local queue reaches 4
- recover below half of both entry thresholds
- require 15 seconds of observed pressure and restart the window after a sampling gap longer than two seconds
- warn once for each episode, then at most once every five minutes while the episode continues
- include queue depths, thresholds, worker count, duration, and possible causes in the structured warning
- add and register the Prometheus warning counter

Runnable queue depth and event-loop delay are complementary. Queue depth measures accumulated runnable work. Event-loop delay measures scheduler delay. Blocking-pool overload needs a separate capacity rule and is outside this change.

Start with `QueueOverloadLogState::observe` and `QueueDepthSnapshot` in `lib/runtime/src/metrics/tokio_perf.rs`. The focused tests in the same file cover the detector boundaries and metric registration.

## Validation

- `cargo run -p dynamo-codegen --bin gen-python-prometheus-names`
- `pre-commit run black --files lib/bindings/python/src/dynamo/prometheus_names.py`
- `cargo test -p dynamo-runtime metrics::tokio_perf` — 7 passed
- `cargo check -p dynamo-runtime`

## Related Issues


lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
 
+#[derive(Default)]
+struct QueueOverloadLogState {
+    overload_started_at: Option<Instant>,
+    last_warning_at: Option<Instant>,
+}
+
+impl QueueOverloadLogState {
+    fn observe(&mut self, now: Instant, queue_depths: &QueueDepthSnapshot) -> Option<Duration> {
+        if queue_depths.total < queue_depths.threshold {
```
jthomson04: **The reset makes the 15s gate very hard to trip.**

`observe` reads an *instantaneous* runnable-queue depth once per second, and a single sample below threshold sets `overload_started_at = None`, restarting the window from zero. Run-queue depth is spiky by nature — it collapses toward 0 every time the workers drain a batch. Under genuine sustained pressure on 8 workers (threshold 32) a realistic sample sequence is `40, 38, 45, 3, 51, 44, 2, 39, ...`; the `3` and the `2` each reset the clock, so the 16 consecutive samples the warning requires are unlikely to ever line up and the log line never fires.

Some smoothing would make the signal survive normal spikiness: an EWMA of `total`, an "N of the last M samples over threshold" counter, or hysteresis (enter at `threshold`, only exit below `threshold / 2`).

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+            return None;
+        }
+
+        let overload_started_at = *self.overload_started_at.get_or_insert(now);
+        let overload_duration = now.saturating_duration_since(overload_started_at);
+        if overload_duration < QUEUE_OVERLOAD_DURATION {
+            return None;
+        }
+
+        if self.last_warning_at.is_some_and(|last_warning_at| {
```
jthomson04, line 395: **A whole overload episode can be dropped, silently.**

The rate limit is keyed only on `last_warning_at` and is never scoped to an episode, so the suppression window carries across unrelated incidents:

- `t=0–20s` — mild episode A, warning logged at `t=15`, `last_warning_at = 15`
- `t=20–100s` — healthy (this resets `overload_started_at` but not `last_warning_at`)
- `t=100–280s` — a **different, severe** 3-minute incident. At `t=115` the sustained gate passes but `115 - 15 = 100s < 300s` → suppressed, and so is every sample through `t=280` (`280 - 15 = 265s < 300s`).

The operator sees one line about the mild episode and nothing about the severe one, with no record that anything was suppressed. Resetting `last_warning_at` alongside `overload_started_at` when the runtime recovers (so each episode gets at least one warning), and/or counting suppressed samples and including the count in the next log line, would fix it.

Worth noting `queue_overload_warnings_are_rate_limited_across_episodes` currently asserts this behavior, so that test would need to change too.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+impl QueueOverloadLogState {
+    fn observe(&mut self, now: Instant, queue_depths: &QueueDepthSnapshot) -> Option<Duration> {
+        if queue_depths.total < queue_depths.threshold {
+            self.overload_started_at = None;
+            return None;
+        }
+
+        let overload_started_at = *self.overload_started_at.get_or_insert(now);
+        let overload_duration = now.saturating_duration_since(overload_started_at);
+        if overload_duration < QUEUE_OVERLOAD_DURATION {
```
jthomson04, line 391: **"Sustained for 15s" is inferred from wall clock, so it can fire on two samples.**

`overload_duration` is `now - overload_started_at` with no notion of how many samples fell in between. If a worker thread blocks — exactly what `EVENT_LOOP_STALL_TOTAL` above exists to catch — and this task isn't polled for 20s, then sample 1 at `t=0` sees a backed-up queue, sample 2 at `t=20` sees it still backed up, and the warning reports `overload_duration_seconds=20` on the strength of two data points with a 20-second hole between them.

That's the inverse of the desired behavior: the message sounds most confident precisely when the sampling loop was least able to observe anything. Counting samples (and requiring a minimum) or rejecting gaps larger than a few collect intervals would close it.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+    global: usize,
+    local: usize,
+    total: usize,
+}
+
+impl QueueDepthSnapshot {
+    fn new(worker_count: usize, global: usize, local: usize) -> Self {
+        Self {
+            worker_count,
+            threshold: worker_count.saturating_mul(QUEUE_DEPTH_THRESHOLD_PER_WORKER),
```
jthomson04: **`worker_count == 0` gives `threshold == 0`, which latches an idle runtime into permanent overload.**

`QueueDepthSnapshot::new(0, 0, 0)` yields `threshold = 0, total = 0`. In `observe`, the early return only triggers on `total < threshold`, i.e. `0 < 0` — false — so an idle runtime is classified as overloaded, `overload_started_at` latches on the first sample, and 15s later it warns `runtime may be overloaded` with `total_queue_depth=0`, then repeats every 5 minutes forever.

I confirmed this by extracting the logic and running it standalone: `total=0 threshold=0 overload=true`, `warn at +15s: Some(15s)`.

`num_workers()` is 0 on a current-thread runtime, and this function is `pub` from `dynamo-runtime`, so it's reachable. The `saturating_mul` already spends a line on defensiveness here; `threshold.max(1)`, or an early `return None` when `worker_count == 0`, is the guard that actually matters. No test covers `worker_count == 0`.

jthomson04: **Scaling the threshold by worker count makes the detector less sensitive on bigger hosts.**

On a 32-worker frontend the threshold is 128 runnable tasks. One worker pinned with 60 queued tasks while the others idle — a real, user-visible latency problem — gives `total = 60 < 128` and never warns. On a 2-worker container the same absolute backlog of 60 trips instantly at threshold 8.

So the meaning of the warning changes with host size and there's no single fleet-wide interpretation of it. A size-invariant predicate — max per-worker local depth, or the fraction of workers whose local queue is over threshold — would read the same everywhere.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
             EVENT_LOOP_STALL_TOTAL.inc();
         }
         if Instant::now() >= next_collect {
-            next_collect = Instant::now() + collect_interval;
-            sample_tokio_metrics(&mut prev_counters);
+            let now = Instant::now();
+            next_collect = now + collect_interval;
+            let queue_depths = sample_tokio_metrics(&mut prev_counters);
```
jthomson04, line 293: **There's already a stronger signal for this question 8 lines up.**

`delay` at line 271 is a direct measurement of "is this runtime keeping up", sampled at 100 Hz and already feeding `EVENT_LOOP_DELAY_SECONDS` / `EVENT_LOOP_STALL_TOTAL`. The new detector answers the same question from a proxy (runnable queue depth) at 1 Hz, which is why it needs the reset/duration/hysteresis machinery below.

The result is two overload signals that can disagree, with the weaker one being the one that logs. A sustained-stall predicate over the existing `delay` — e.g. stall rate above X sustained over 15s — needs no new sampling, no new threshold, and no worker-count scaling. If queue depth adds something `delay` genuinely can't see, a sentence in the commit message on why would help.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+            if let Some(overload_duration) = queue_overload.observe(now, &queue_depths) {
+                tracing::warn!(
```
jthomson04: **This is the one signal an operator would alert on, and it's log-only.**

Every other signal in this file is exported as a Gauge/Counter registered under `prometheus_names::tokio_perf`. This one is a `tracing::warn!` emitted at most once per 5 minutes, so there's no way to build a Grafana panel or alert rule for "frontend runtime overloaded", no visibility into episode count or duration, and (given the rate-limit behavior noted above) no way to tell that episodes were dropped.

A `tokio_queue_overload_total` counter and/or a 0/1 gauge set on every sample alongside the log is ~5 lines and makes the feature usable from a dashboard.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -13,6 +13,10 @@ use tokio_util::sync::CancellationToken;
 use super::prometheus_names::{frontend_perf, name_prefix, tokio_perf as names};
 use crate::MetricsRegistry;
 
+const QUEUE_DEPTH_THRESHOLD_PER_WORKER: usize = 4;
```
jthomson04, line 16: **These are the three knobs most likely to need tuning in production, and none is overridable.**

`lib/runtime/src/config/environment_names.rs:75` defines `DYN_ENABLE_POLL_HISTOGRAM` and `lib/runtime/src/config.rs:377` gates tokio-runtime behavior on it — the convention for tunable tokio knobs is right next door.

If `QUEUE_DEPTH_THRESHOLD_PER_WORKER = 4` turns out to be too noisy (or, per the `worker_count == 0` case, noisy in a way nobody wants) for a given deployment, the only remedy is a rebuild — a log-spam incident can't be silenced without shipping a new binary. Reading these from env with the current values as defaults, plus a `DYN_...` kill switch, would make it operable.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -356,4 +428,90 @@ fn sample_tokio_metrics(prev: &mut PrevWorkerCounters) {
             .with_label_values(&[&worker_label])
             .set((busy_proxy * 1000.0) as i64);
     }
+
+    QueueDepthSnapshot::new(num_workers, global_queue_depth, local_queue_depth)
```
jthomson04: **`blocking_queue_depth` is sampled at line 384 and then not used by the detector.**

If the frontend exhausts `max_blocking_threads` on `spawn_blocking` work, requests pile up in the blocking pool while the async workers sit idle with near-empty run queues. `global + local` stays well under threshold, so nothing warns — even though `TOKIO_BLOCKING_QUEUE_DEPTH`, read a few lines above, shows the backlog and the frontend is stalled.

That's a real frontend overload mode and the value is already in scope; including it in the snapshot (as a separate condition rather than summed, since the pool is sized differently) costs almost nothing.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
     }
 }
 
+#[derive(Default)]
+struct QueueOverloadLogState {
+    overload_started_at: Option<Instant>,
+    last_warning_at: Option<Instant>,
+}
+
+impl QueueOverloadLogState {
```
jthomson04, line 364: **This is the fifth copy of the `last_warned_at` rate-limit pattern in the repo.**

The same "warn at most once every N seconds, tracked via `Option<Instant>`" logic already exists in:

- `lib/kvbm-physical/src/transfer/notifications/mod.rs:55`
- `lib/kvbm-physical/src/transfer/notifications/nixl_events.rs:72`
- `lib/llm/src/block_manager/v2/physical/transfer/notifications/mod.rs:55`
- `lib/llm/src/block_manager/v2/physical/transfer/notifications/nixl_events.rs:37`

all as `check_and_warn_slow_transfer(..., last_warned_at: Option<Instant>) -> Option<Instant>` with identical `elapsed() > interval` bodies. A small `RateLimitedWarn { interval, last: Option<Instant> }` with `fn should_warn(&mut self, now) -> bool` in `dynamo_runtime::logging` would serve this site and let the other four drop their bespoke state — and it'd be the natural place to add the suppressed-count fix from the rate-limit thread.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+                    overload_duration_seconds = overload_duration.as_secs(),
+                    "Frontend Tokio runtime may be overloaded: runnable task queue has remained high"
+                );
+            }
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+struct QueueDepthSnapshot {
```
jthomson04, line 317: **`threshold` and `total` are pure functions of the other three fields.**

`threshold` is always `worker_count * QUEUE_DEPTH_THRESHOLD_PER_WORKER` and `total` is always `global + local`. Storing them means the struct can be constructed in a state where `threshold != worker_count * 4` (the tests do exactly this at line 450), and it's why `queue_depth_threshold_scales_with_worker_count` and `queue_depth_snapshot_adds_global_and_local_queues` exist — they assert the `*` and `+` operators rather than any behavior of this module.

`fn threshold(&self)` and `fn total(&self)` on a 3-field struct would delete both fields, the `new()` constructor, the `Eq` derive, and both tests.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -356,4 +428,90 @@ fn sample_tokio_metrics(prev: &mut PrevWorkerCounters) {
             .with_label_values(&[&worker_label])
             .set((busy_proxy * 1000.0) as i64);
     }
+
+    QueueDepthSnapshot::new(num_workers, global_queue_depth, local_queue_depth)
+}
+
+#[cfg(test)]
+mod tests {
```
jthomson04, line 515: **Nothing here covers `sample_tokio_metrics` or the warn path.**

All four tests exercise the two pure helpers. The one integration point the PR adds — `QueueDepthSnapshot::new(num_workers, global_queue_depth, local_queue_depth)` at line 432, three bare `usize`s passed positionally into `new(worker_count, global, local)` — has no coverage at all. Swapping `global` and `local`, or passing `local` where `worker_count` belongs, compiles cleanly and leaves every existing test green.

A `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` that calls `sample_tokio_metrics` and asserts `snapshot.worker_count == 2` would anchor it, and would also have caught the `worker_count == 0` case on a current-thread runtime.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+            let queue_depths = sample_tokio_metrics(&mut prev_counters);
+            if let Some(overload_duration) = queue_overload.observe(now, &queue_depths) {
+                tracing::warn!(
+                    worker_count = queue_depths.worker_count,
+                    queue_depth_threshold = queue_depths.threshold,
+                    global_queue_depth = queue_depths.global,
+                    worker_local_queue_depth_total = queue_depths.local,
+                    total_queue_depth = queue_depths.total,
+                    overload_duration_seconds = overload_duration.as_secs(),
+                    "Frontend Tokio runtime may be overloaded: runnable task queue has remained high"
```
jthomson04: **"Frontend" is hardcoded, but this loop isn't frontend-specific.**

`tokio_metrics_and_canary_loop` is documented at line 253 as "Spawn this on the runtime you want to monitor" and is `pub` from `dynamo-runtime`, which the Python bindings link too. Any non-frontend process that spawns it will emit "Frontend Tokio runtime may be overloaded" and send an operator to debug the wrong component.

Either drop the word ("Tokio runtime may be overloaded") or take a component label as a parameter — the caller knows its own identity.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
             EVENT_LOOP_STALL_TOTAL.inc();
         }
         if Instant::now() >= next_collect {
```
jthomson04: **Two clock reads for one logical timestamp.**

`Instant::now()` is called here and again on the next line. Beyond the redundant read, the two values differ, so the instant compared against `next_collect` isn't the instant used to seed `overload_started_at` and compute `overload_duration`. Hoisting a single `let now = Instant::now();` above the `if` and comparing `now >= next_collect` removes both.

jthomson04 (commented):
## Review: sustained Tokio queue pressure warning

Reviewed the detector end-to-end. The feature is worth having, but as written I think it mostly won't fire when you want it to, and can fire when you don't. Four things stand out:

1. **It likely never fires under real overload** — a single below-threshold sample resets the 15s window, and instantaneous run-queue depth is spiky enough that 16 consecutive samples over threshold is a tall order. (`observe`, line 324)
2. **It can drop an entire incident** — the 5-minute rate limit is keyed on `last_warning_at` alone and isn't scoped per episode, so a severe incident following a mild one goes unlogged with nothing recording the suppression. The current test asserts this behavior. (line 335)
3. **Conversely, it can fire on two samples** — "sustained 15s" is wall-clock inferred, so a starved sampling loop produces the most confident-sounding warning from the weakest evidence. (line 331)
4. **`worker_count == 0` → `threshold == 0`**, and `0 >= 0` latches an idle runtime into permanent overload, warning `total_queue_depth=0` every 5 minutes. Confirmed by running the logic standalone. (line 308)

Two altitude questions worth settling before tuning the details: `delay` at line 271 is already a direct 100 Hz measurement of the same thing this proxies at 1 Hz — is there something queue depth sees that it doesn't? And this is the one signal operators would alert on, yet it's the only one in the file that isn't exported as a metric.

The rest (hardcoded constants vs. the `environment_names.rs` convention, a fifth copy of the `last_warned_at` pattern, derived fields on the snapshot, the test gap at the one real call site) is in the inline comments.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -356,4 +428,90 @@ fn sample_tokio_metrics(prev: &mut PrevWorkerCounters) {
...
+    fn queue_overload_requires_sustained_pressure_and_resets() {
+        let start = Instant::now();
+        let overloaded = QueueDepthSnapshot::new(4, 8, 8);
+        let below_threshold = QueueDepthSnapshot::new(4, 7, 8);
+        let mut state = QueueOverloadLogState::default();
+
+        assert_eq!(state.observe(start, &overloaded), None);
+        assert_eq!(
+            state.observe(
+                start + QUEUE_OVERLOAD_DURATION - Duration::from_nanos(1),
```
jthomson04: **This assumes `Instant` has 1ns resolution.**

Subtracting `Duration::from_nanos(1)` only produces a distinct instant if the platform clock is that fine-grained. Where `Instant` is backed by coarser ticks (Windows QPC at 100ns/tick converts 1ns to 0 ticks), this expression equals `start + QUEUE_OVERLOAD_DURATION`, `observe` returns `Some(15s)`, and the `assert_eq!(..., None)` fails.

`Duration::from_millis(1)` tests the same boundary without the resolution assumption.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. The detector now uses hysteresis: it enters at the high threshold and recovers only below half of both the aggregate and per-worker thresholds. The focused test covers the hysteresis band.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. Recovery now clears the warning timestamp, so every new sustained episode can emit its first warning. A continuing episode can repeat only after five minutes.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. A gap longer than two seconds resets the evidence window. A high sample after the gap starts a new episode instead of using the old wall-clock start time.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented defensively in a4a804478b. Threshold calculations use at least one effective worker, and the tests cover worker_count == 0. Tokio 1.48 reports one worker for a current-thread runtime, but the guard prevents a zero threshold if that invariant changes.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: I am keeping these values as internal constants in this focused PR. Per-episode warnings and five-minute repeats limit log volume. We can add configuration later if production data shows that the thresholds need deployment-specific tuning.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. The snapshot now records the maximum worker-local depth. Pressure starts when either total depth reaches 4 x worker count or one local queue reaches 4.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: I kept the queue detector because the two signals describe different conditions. Queue depth measures accumulated runnable work, while the existing canary measures scheduling delay. They can disagree in useful ways. The PR description now states this distinction.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. Each emitted warning increments the registered dynamo_tokio_queue_overload_warnings_total counter. The test covers the increment and both registry paths.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: I am keeping blocking-pool detection separate. The blocking pool has a different capacity limit, so its depth should not be combined with async runnable queues. Its existing Prometheus gauge remains available for a separate alert or follow-up detector.

lib/runtime/src/metrics/tokio_perf.rs
```diff
@@ -269,12 +274,75 @@ pub async fn tokio_metrics_and_canary_loop(cancel: CancellationToken) {
...
+        }
+
+        if self.last_warning_at.is_some_and(|last_warning_at| {
+            now.saturating_duration_since(last_warning_at) < QUEUE_OVERLOAD_LOG_INTERVAL
+        }) {
+            return None;
+        }
+
+        self.last_warning_at = Some(now);
+        Some(overload_duration)
```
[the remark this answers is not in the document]
jthomson04: Confirmed and implemented in a4a804478b. The five-minute limit is now scoped to one episode. Recovery resets the cooldown, so a new sustained episode can warn after 15 seconds.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. The warning now says "Tokio runtime may be overloaded" and does not assume that the caller is a frontend.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. The collection pass now reads Instant::now() once and uses that value for the schedule check, next collection time, and overload observation.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. A two-worker Tokio test now calls sample_tokio_metrics directly. A second test calls the warning helper and checks the counter increment and both Prometheus registration paths.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: I am not adding the repository-wide rate-limit refactor in this focused PR. This state also owns hysteresis, sampling continuity, and episode recovery, so only its final cooldown check matches the other sites. A shared helper can be handled as a separate cleanup.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. The duration boundary test now uses one millisecond below the limit.

lib/runtime/src/metrics/tokio_perf.rs
jthomson04: Implemented in a4a804478b. QueueDepthSnapshot stores only raw worker and queue values. Total and threshold values are calculated by methods, and the constructor and Eq derive are removed.

jthomson04:
/ok to test [1a7840b](https://github.com/ai-dynamo/dynamo/commit/1a7840bf6bb96f65c5c5456089484f45ef50c25f)

ishandhanani (approved):

jthomson04:
/ok to test [8785752](https://github.com/ai-dynamo/dynamo/commit/8785752401bd30d5f36c3a0f7065da4cb81417bf)

rmccorm4:
/ok to test [4708f8d](https://github.com/ai-dynamo/dynamo/pull/13027/commits/4708f8deeafc7773717339b7d148db589a43cdbc)

PR #13042, opened 2026-08-11
fix(epp): reserve prefill load during routing

### LINKED ISSUE #12864: [BUG]: Disaggregated prefill routing is not load-balanced — some prefill workers queue deeply while others stay idle
### Describe the Bug

In a disaggregated deployment fronted by the inference-gateway EPP, prefill requests are not balanced across prefill workers by load. Under sustained traffic, some prefill workers build up large queues while other prefill replicas stay idle.

### Steps to Reproduce

1. Deploy a disaggregated model (separate prefill and decode workers) fronted by the EPP.
2. Send sustained chat/completions traffic with mixed prompt sizes.
3. Scrape each prefill pod's metrics on the system port (`:9090`) and compare `vllm:num_requests_waiting` / `vllm:num_requests_running` across prefill pods.

### Expected Behavior

Prefill requests are distributed across prefill replicas by load; idle prefill replicas receive requests before new requests queue behind already-busy prefill workers.

### Actual Behavior

Prefill queue depth is severely imbalanced across prefill replicas. Example snapshot across the prefill pods of one deployment:

```
prefill pod   num_requests_running   num_requests_waiting
A             1                      83
B             12                     78
C             1                      5
D             0                      0
E             0                      0
```

New requests continue to queue on busy prefill workers while replicas D and E stay idle.

### Environment

- Dynamo: 1.3.0
- Deployment: disaggregated (separate prefill/decode workers), inference-gateway EPP, Kubernetes
- Backend: vLLM

tmonty12:
## Summary

Fixes #12864.

Native Rust EPP previously queried a prefill worker without admitting it into scheduler state. Concurrent cache-hot requests could therefore select the same prefill worker before prior choices contributed projected prefill load.

This PR makes native EPP atomically reserve KV-routed prefill capacity and carries that booking through the request lifecycle.

- EPP creates an internal UUID reservation ID and routes prefill through `PrefillRouter::reserve_prefill_worker`.
- KV routing atomically selects and admits the worker; EPP retains the resulting booking until a prefill-complete or terminal request callback.
- `PrefillReservation` is the RAII owner of its scheduler booking. It stores an attempt-fenced descriptor `(request ID, worker, attempt ID)`:
  - `Drop` enqueues non-blocking cleanup.
  - `release(self)` consumes the reservation and waits for acknowledged cleanup.
  - The attempt ID prevents delayed cleanup from releasing a newer booking that reused the same request ID and worker.
- Scheduler lifecycle leases now exist for every `TrackedWithLifecycle` admission, including when queueing is disabled, so cancellation during admission retracts an admitted booking rather than retaining prefill load until stale expiry.
- Built-in non-KV prefill modes continue to use the existing advisory selection path and return an untracked, no-op reservation. This preserves disaggregation instead of silently falling back to aggregated execution.

EPP no longer owns a separate `tokio::spawn`-based reservation guard. Any early return or cancellation after successful admission drops the reservation, which enqueues cleanup through the scheduler’s existing channel.

## Follow-up

The non-KV compatibility behavior is intentionally advisory. A follow-up should add a stateful non-KV select-and-reserve API that:

- advances round-robin selection exactly once;
- retains and releases occupancy reservations for least-loaded, P2C, and device-aware routing; and
- selects without dispatching through `PushRouter`, because EPP owns downstream HTTP dispatch.

## Scope

This is intentionally limited to native Rust EPP and the existing KV-router lifecycle machinery. It does not add that broader non-KV routing-lifecycle abstraction.

## Validation

- `cargo test -p dynamo-kv-router --lib` — 946 passed, 2 ignored
- `cargo test -p dynamo-llm kv_router::prefill_router::query::tests --lib` — 3 passed
- `cargo test -p dynamo-ext-proc epp --lib` — 63 passed
- `cargo check -p dynamo-llm -p dynamo-ext-proc`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/deploy/inference-gateway/ext-proc/src/epp.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp.rs
@@ -8,6 +8,7 @@
 //! ext_proc server calls these types directly as async Rust.
 
 use std::collections::{HashMap, HashSet};
+use std::future::Future;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard};
@@ -119,16 +120,39 @@ pub struct Router {
     served_model: String,
 }
 
-/// Releases a prefill booking if routing fails before the ext-proc server adopts it.
+/// A prefill booking that can be asynchronously released.
 ///
-/// After [`Self::disarm`] transfers the booking into [`Router::prefill_bookings`],
-/// response callbacks own its lifecycle instead.
-struct PrefillReservationGuard {
-    reservation: Option<PrefillReservation>,
+/// Kept monomorphized so the guard and booking map do not allocate a boxed
+/// cleanup closure on the routing path.
+trait PrefillBooking: Send + 'static {
+    fn release(self) -> impl Future<Output = Result<()>> + Send;
+    fn worker_id(&self) -> u64;
+    fn dp_rank(&self) -> u32;
 }
 
-impl PrefillReservationGuard {
-    fn new(reservation: PrefillReservation) -> Self {
+impl PrefillBooking for PrefillReservation {
+    fn release(self) -> impl Future<Output = Result<()>> + Send {
+        async move { PrefillReservation::release(&self).await }
+    }
+
+    fn worker_id(&self) -> u64 {
+        PrefillReservation::worker_id(self)
+    }
+
+    fn dp_rank(&self) -> u32 {
+        PrefillReservation::dp_rank(self)
+    }
+}
+
+/// Releases a prefill booking if routing fails before the ext-proc server adopts it.
+/// After [`Self::disarm`] transfers the booking into [`Router::prefill_bookings`],
+/// response callbacks own its lifecycle instead.
+struct PrefillReservationGuard<R: PrefillBooking = PrefillReservation> {
+    reservation: Option<R>,
+}
+
+impl<R: PrefillBooking> PrefillReservationGuard<R> {
+    fn new(reservation: R) -> Self {
         Self {
             reservation: Some(reservation),
         }
@@ -148,14 +172,14 @@ impl PrefillReservationGuard {
             .dp_rank()
     }
 
-    fn disarm(&mut self) -> PrefillReservation {
+    fn disarm(&mut self) -> R {
         self.reservation
             .take()
             .expect("reservation guard is disarmed once")
     }
 }
 
-impl Drop for PrefillReservationGuard {
+impl<R: PrefillBooking> Drop for PrefillReservationGuard<R> {
     fn drop(&mut self) {
         if let Some(reservation) = self.reservation.take() {
             tokio::spawn(async move {
@@ -167,6 +191,23 @@ impl Drop for PrefillReservationGuard {
     }
 }
 
+/// Remove and release a booking once. Both response lifecycle callbacks use
+/// this helper so terminal completion before first output and duplicate signals
+/// have identical behavior.
+async fn release_prefill_booking<R: PrefillBooking>(
+    prefill_bookings: &DashMap<String, R>,
+    booking_id: &str,
+) {
+    if let Some((_, reservation)) = prefill_bookings.remove(booking_id)
+        && let Err(error) = reservation.release().await
+    {
+        tracing::debug!(
+            reservation_id = booking_id,
+            %error,
+            "Failed to release native EPP prefill reservation"
+        );
+    }
+}
 impl Router {
     /// Initialize the router from discovery.
     ///
@@ -1623,15 +1664,7 @@ impl EndpointPicker for Router {
         if booking_id.is_empty() {
             return;
         }
-        if let Some((_, reservation)) = self.prefill_bookings.remove(booking_id)
-            && let Err(error) = reservation.release().await
-        {
-            tracing::debug!(
-                reservation_id = booking_id,
-                %error,
-                "Failed to release native EPP prefill reservation"
-            );
-        }
+        release_prefill_booking(&self.prefill_bookings, booking_id).await;
         if let Err(e) = self.mark_prefill_complete(booking_id).await {
             tracing::debug!(
                 reservation_id = booking_id,
@@ -1655,15 +1688,7 @@ impl EndpointPicker for Router {
                 "Request complete with usage"
             );
         }
-        if let Some((_, reservation)) = self.prefill_bookings.remove(booking_id)
-            && let Err(error) = reservation.release().await
-        {
-            tracing::debug!(
-                reservation_id = booking_id,
-                %error,
-                "Failed to release native EPP prefill reservation"
-            );
-        }
+        release_prefill_booking(&self.prefill_bookings, booking_id).await;
         if let Err(e) = self.free_request(booking_id).await {
             tracing::debug!(
                 reservation_id = booking_id,
@@ -1678,6 +1703,83 @@ impl EndpointPicker for Router {
 mod tests {
     use super::*;
     use k8s_openapi::api::core::v1::Pod;
+
+    use std::sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    };
+
+    struct StubPrefillBooking {
+        releases: Arc<AtomicUsize>,
+    }
+
+    impl PrefillBooking for StubPrefillBooking {
+        fn release(self) -> impl Future<Output = Result<()>> + Send {
+            async move {
+                self.releases.fetch_add(1, Ordering::SeqCst);
+                Ok(())
+            }
+        }
+
+        fn worker_id(&self) -> u64 {
+            0
+        }
+
+        fn dp_rank(&self) -> u32 {
+            0
+        }
+    }
+
+    async fn wait_for_release(releases: &AtomicUsize, expected: usize) {
+        tokio::time::timeout(Duration::from_secs(1), async {
+            while releases.load(Ordering::SeqCst) != expected {
+                tokio::task::yield_now().await;
+            }
+        })
+        .await
+        .expect("prefill booking release should complete");
+    }
+
+    #[tokio::test]
+    async fn prefill_reservation_guard_releases_only_when_armed() {
+        let releases = Arc::new(AtomicUsize::new(0));
+
+        {
+            let _guard = PrefillReservationGuard::new(StubPrefillBooking {
+                releases: Arc::clone(&releases),
+            });
+        }
+        wait_for_release(&releases, 1).await;
+
+        let mut guard = PrefillReservationGuard::new(StubPrefillBooking {
+            releases: Arc::clone(&releases),
+        });
+        let booking = guard.disarm();
+        drop(guard);
+        tokio::task::yield_now().await;
+        assert_eq!(releases.load(Ordering::SeqCst), 1);
+        drop(booking);
+        assert_eq!(releases.load(Ordering::SeqCst), 1);
+    }
+
+    #[tokio::test]
+    async fn prefill_booking_map_releases_terminal_before_first_output_once() {
+        let releases = Arc::new(AtomicUsize::new(0));
+        let bookings = DashMap::new();
+        bookings.insert(
+            "booking".to_string(),
+            StubPrefillBooking {
+                releases: Arc::clone(&releases),
+            },
+        );
+
+        release_prefill_booking(&bookings, "booking").await;
+        assert!(bookings.is_empty());
+        assert_eq!(releases.load(Ordering::SeqCst), 1);
+
+        release_prefill_booking(&bookings, "booking").await;
+        assert_eq!(releases.load(Ordering::SeqCst), 1);
+    }
 
     #[test]
     fn tenant_header_overrides_body_cache_namespace() {
--- a/deploy/inference-gateway/ext-proc/src/epp.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp.rs
@@ -8,7 +8,6 @@
 //! ext_proc server calls these types directly as async Rust.
 
 use std::collections::{HashMap, HashSet};
-use std::future::Future;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard};
@@ -120,82 +119,11 @@ pub struct Router {
     served_model: String,
 }
 
-/// A prefill booking that can be asynchronously released.
-///
-/// Kept monomorphized so the guard and booking map do not allocate a boxed
-/// cleanup closure on the routing path.
-trait PrefillBooking: Send + 'static {
-    fn release(self) -> impl Future<Output = Result<()>> + Send;
-    fn worker_id(&self) -> u64;
-    fn dp_rank(&self) -> u32;
-}
-
-impl PrefillBooking for PrefillReservation {
-    fn release(self) -> impl Future<Output = Result<()>> + Send {
-        async move { PrefillReservation::release(&self).await }
-    }
-
-    fn worker_id(&self) -> u64 {
-        PrefillReservation::worker_id(self)
-    }
-
-    fn dp_rank(&self) -> u32 {
-        PrefillReservation::dp_rank(self)
-    }
-}
-
-/// Releases a prefill booking if routing fails before the ext-proc server adopts it.
-/// After [`Self::disarm`] transfers the booking into [`Router::prefill_bookings`],
-/// response callbacks own its lifecycle instead.
-struct PrefillReservationGuard<R: PrefillBooking = PrefillReservation> {
-    reservation: Option<R>,
-}
-
-impl<R: PrefillBooking> PrefillReservationGuard<R> {
-    fn new(reservation: R) -> Self {
-        Self {
-            reservation: Some(reservation),
-        }
-    }
-
-    fn worker_id(&self) -> u64 {
-        self.reservation
-            .as_ref()
-            .expect("reservation guard is accessed only before disarm")
-            .worker_id()
-    }
-
-    fn dp_rank(&self) -> u32 {
-        self.reservation
-            .as_ref()
-            .expect("reservation guard is accessed only before disarm")
-            .dp_rank()
-    }
-
-    fn disarm(&mut self) -> R {
-        self.reservation
-            .take()
-            .expect("reservation guard is disarmed once")
-    }
-}
-
-impl<R: PrefillBooking> Drop for PrefillReservationGuard<R> {
-    fn drop(&mut self) {
-        if let Some(reservation) = self.reservation.take() {
-            tokio::spawn(async move {
-                if let Err(error) = reservation.release().await {
-                    tracing::debug!(%error, "Failed to release unadopted prefill reservation");
-                }
-            });
-        }
-    }
-}
-
 /// Remove and release a booking once. Both response lifecycle callbacks use
 /// this helper so terminal completion before first output and duplicate signals
 /// have identical behavior.
-async fn release_prefill_booking<R: PrefillBooking>(
-    prefill_bookings: &DashMap<String, R>,
+async fn release_prefill_booking(
+    prefill_bookings: &DashMap<String, PrefillReservation>,
     booking_id: &str,
 ) {
     if let Some((_, reservation)) = prefill_bookings.remove(booking_id)
@@ -1505,7 +1433,7 @@ impl EndpointPicker for Router {
         //
         // If the prefill router is not activated (no prefill workers discovered yet, or the inner
         // router has been deactivated), fall back to aggregated routing.
-        let mut prefill_booking = self
+        let prefill_booking = self
             .route_prefill(
                 &format!("epp-prefill/{reservation_id}"),
                 &tokens,
@@ -1515,8 +1443,7 @@ impl EndpointPicker for Router {
                 allowed_worker_ids.clone(),
                 routing_constraints.clone(),
             )
-            .await
-            .map(PrefillReservationGuard::new);
+            .await;
 
         let is_disaggregated = match &prefill_booking {
             Ok(_) => true,
@@ -1590,9 +1517,9 @@ impl EndpointPicker for Router {
             .as_ref()
             .ok()
             .map(|booking| (booking.worker_id(), booking.dp_rank()));
-        if let Ok(booking) = &mut prefill_booking {
+        if let Ok(booking) = prefill_booking {
             self.prefill_bookings
-                .insert(reservation_id.clone(), booking.disarm());
+                .insert(reservation_id.clone(), booking);
         }
 
         // Build routing headers: x-dynamo-worker-instance-id, x-dynamo-dp-rank,
@@ -1704,82 +1631,7 @@ mod tests {
     use super::*;
     use k8s_openapi::api::core::v1::Pod;
 
-    use std::sync::{
-        Arc,
-        atomic::{AtomicUsize, Ordering},
-    };
-
-    struct StubPrefillBooking {
-        releases: Arc<AtomicUsize>,
-    }
-
-    impl PrefillBooking for StubPrefillBooking {
-        fn release(self) -> impl Future<Output = Result<()>> + Send {
-            async move {
-                self.releases.fetch_add(1, Ordering::SeqCst);
-                Ok(())
-            }
-        }
-
-        fn worker_id(&self) -> u64 {
-            0
-        }
-
-        fn dp_rank(&self) -> u32 {
-            0
-        }
-    }
-
-    async fn wait_for_release(releases: &AtomicUsize, expected: usize) {
-        tokio::time::timeout(Duration::from_secs(1), async {
-            while releases.load(Ordering::SeqCst) != expected {
-                tokio::task::yield_now().await;
-            }
-        })
-        .await
-        .expect("prefill booking release should complete");
-    }
-
-    #[tokio::test]
-    async fn prefill_reservation_guard_releases_only_when_armed() {
-        let releases = Arc::new(AtomicUsize::new(0));
-
-        {
-            let _guard = PrefillReservationGuard::new(StubPrefillBooking {
-                releases: Arc::clone(&releases),
-            });
-        }
-        wait_for_release(&releases, 1).await;
-
-        let mut guard = PrefillReservationGuard::new(StubPrefillBooking {
-            releases: Arc::clone(&releases),
-        });
-        let booking = guard.disarm();
-        drop(guard);
-        tokio::task::yield_now().await;
-        assert_eq!(releases.load(Ordering::SeqCst), 1);
-        drop(booking);
-        assert_eq!(releases.load(Ordering::SeqCst), 1);
-    }
-
-    #[tokio::test]
-    async fn prefill_booking_map_releases_terminal_before_first_output_once() {
-        let releases = Arc::new(AtomicUsize::new(0));
-        let bookings = DashMap::new();
-        bookings.insert(
-            "booking".to_string(),
-            StubPrefillBooking {
-                releases: Arc::clone(&releases),
-            },
-        );
-
-        release_prefill_booking(&bookings, "booking").await;
-        assert!(bookings.is_empty());
-        assert_eq!(releases.load(Ordering::SeqCst), 1);
-
-        release_prefill_booking(&bookings, "booking").await;
-        assert_eq!(releases.load(Ordering::SeqCst), 1);
-    }
+    use std::sync::{Arc, atomic::Ordering};
 
     #[test]
     fn tenant_header_overrides_body_cache_namespace() {
--- a/deploy/inference-gateway/ext-proc/src/epp.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp.rs
@@ -1545,10 +1545,12 @@ impl EndpointPicker for Router {
                 "x-dynamo-prefill-instance-id".to_string(),
                 format!("{}", prefill_worker_id),
             ));
-            headers.push((
-                "x-dynamo-prefill-dp-rank".to_string(),
-                prefill_dp_rank.to_string(),
-            ));
+            if let Some(prefill_dp_rank) = prefill_dp_rank {
+                headers.push((
+                    "x-dynamo-prefill-dp-rank".to_string(),
+                    prefill_dp_rank.to_string(),
+                ));
+            }
         } else {
             headers.push((
                 "x-dynamo-routing-mode".to_string(),
--- a/deploy/inference-gateway/ext-proc/src/epp.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp.rs
@@ -462,6 +462,7 @@ impl Router {
     /// tier. `routing_constraints` carries the request's required/preferred
     /// taints (lifted from `nvext.routing_constraints`); a hard `required_taints`
     /// mismatch excludes a worker from selection.
+    #[expect(clippy::too_many_arguments)]
     pub async fn route_prefill(
         &self,
         reservation_id: &str,
--- a/deploy/inference-gateway/ext-proc/src/epp.rs
+++ b/deploy/inference-gateway/ext-proc/src/epp.rs
@@ -14,10 +14,11 @@ use std::sync::{Arc, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard};
 use std::time::Duration;
 
 use anyhow::Result;
+use dashmap::DashMap;
 use dynamo_kv_router::config::{RouterConfigOverride, try_kv_router_config_from_dynamo_env};
 use dynamo_kv_router::protocols::{RoutingConstraints, WorkerWithDpRank};
 use dynamo_llm::discovery::{ModelManager, WORKER_TYPE_DECODE};
-use dynamo_llm::kv_router::prefill_router::PrefillQueryOutcome;
+use dynamo_llm::kv_router::prefill_router::PrefillReservation;
 use dynamo_llm::kv_router::{ManagedKvRouter, PrefillRouter};
 use dynamo_llm::model_card::ModelDeploymentCard;
 use dynamo_llm::preprocessor::OpenAIPreprocessor;
@@ -32,6 +33,7 @@ use dynamo_runtime::discovery::{
 };
 use dynamo_runtime::pipeline::RouterMode;
 use dynamo_runtime::{DistributedRuntime, Runtime};
+use uuid::Uuid;
 
 use crate::epp_router::endpoint_in_subset;
 use crate::picker::{Endpoint, EndpointPicker, PickError, PickResult, RequestInfo, ResponseUsage};
@@ -108,12 +110,61 @@ const DYNAMO_CONTAINER_PORT_NAME: &str = "http";
 /// without the `block_on` / unsafe FFI overhead.
 pub struct Router {
     prefill_router: Arc<PrefillRouter>,
+    prefill_bookings: DashMap<String, PrefillReservation>,
     decode_router: ManagedKvRouter,
     preprocessor: Arc<OpenAIPreprocessor>,
     runtime: Runtime,
     worker_index: Arc<RwLock<WorkerEndpointIndex>>,
     pod_store_ready: Arc<AtomicBool>,
     served_model: String,
+}
+
+/// Releases a prefill booking if routing fails before the ext-proc server adopts it.
+///
+/// After [`Self::disarm`] transfers the booking into [`Router::prefill_bookings`],
+/// response callbacks own its lifecycle instead.
+struct PrefillReservationGuard {
+    reservation: Option<PrefillReservation>,
+}
+
+impl PrefillReservationGuard {
+    fn new(reservation: PrefillReservation) -> Self {
+        Self {
+            reservation: Some(reservation),
+        }
+    }
+
+    fn worker_id(&self) -> u64 {
+        self.reservation
+            .as_ref()
+            .expect("reservation guard is accessed only before disarm")
+            .worker_id()
+    }
+
+    fn dp_rank(&self) -> u32 {
+        self.reservation
+            .as_ref()
+            .expect("reservation guard is accessed only before disarm")
+            .dp_rank()
+    }
+
+    fn disarm(&mut self) -> PrefillReservation {
+        self.reservation
+            .take()
+            .expect("reservation guard is disarmed once")
+    }
+}
+
+impl Drop for PrefillReservationGuard {
+    fn drop(&mut self) {
+        if let Some(reservation) = self.reservation.take() {
+            tokio::spawn(async move {
+                if let Err(error) = reservation.release().await {
+                    tracing::debug!(%error, "Failed to release unadopted prefill reservation");
+                }
+            });
+        }
+    }
 }
 
 impl Router {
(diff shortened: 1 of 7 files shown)


deploy/inference-gateway/epp/pkg/plugins/disagg/decode_scorer.go
```diff
@@ -198,84 +216,135 @@ func (s *DynDecodeScorer) Score(ctx context.Context, cycleState *schedtypes.Cycl
...
+		return
+	}
+
+	// Keep the marker while the FFI call is in flight. On failure, remove it so
+	// the next response chunk retries instead of leaking the prefill load.
+	if _, alreadyInFlight := s.prefillMarkInFlight.LoadOrStore(bookingID, struct{}{}); alreadyInFlight {
+		return
+	}
+	if err := s.markPrefillComplete(bookingID); err != nil {
+		s.prefillMarkInFlight.Delete(bookingID)
```
tmonty12: [P1] This retries `mark_prefill_complete` on every subsequent response chunk after an error. Each FFI attempt can block for five seconds, while GAIE serializes callbacks through a 100-entry per-request queue and drains it before EOS. A persistent router timeout can therefore accumulate `N × 5s` of cleanup work, fill the response queue, and stall completion. Can we move this to one bounded per-booking retry loop with backoff, canceled by EOS, and add a persistent-failure test that proves bounded calls and EOS latency?

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
```diff
@@ -105,28 +126,62 @@ func (s *DynPrefillScorer) Score(ctx context.Context, cycleState *schedtypes.Cyc
...
+				"bookingID", booking.ID)
+		}
+		booking.PrefillReserved = false
+		cycleState.Write(BookingStateKey, booking)
+		cycleState.Write(PrefillEnabledStateKey, &PrefillEnabledState{Enabled: false})
+		return uniformScores(endpoints, 0)
+	}
+
+	bookingID := booking.ID
+	go func() {
```
tmonty12: [P2] This creates one cancellation owner, and `DynDecodeScorer.PreRequest` creates a second one; EOS already frees the same booking. Because both hooks share the ext_proc stream context, normal completion runs the EOS free and then wakes both watchers, producing two extra synchronous `free_request` FFI calls and scheduler updates per disaggregated request. Can we centralize this in one booking lifecycle guard and stop its watcher on terminal cleanup?

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,470 @@
...
+
+fn reservation_retention() -> Duration {
+    let configured = std::env::var(ACTIVE_REQUEST_EXPIRY_ENV)
+        .ok()
+        .and_then(|raw| raw.parse::<u64>().ok())
+        .filter(|seconds| *seconds > 0)
+        .map(Duration::from_secs)
+        .unwrap_or(DEFAULT_ACTIVE_REQUEST_EXPIRY_DURATION);
+    configured
+        .max(DEFAULT_ACTIVE_REQUEST_EXPIRY_DURATION)
```
tmonty12: [P2] This diverges from the scheduler’s canonical `DYN_ROUTER_ACTIVE_REQUEST_EXPIRY_SECS` parser by flooring valid shorter overrides at the 300s default. For example, a 30s scheduler expiry leaves registry entries and their old chooser `Arc`s alive for 360s. Can we reuse/export the canonical resolved expiry and add only the grace period, with shorter/longer override tests?

tmonty12 (commented):
Deep review: three lifecycle/performance findings inline. Focused Rust reservation tests passed 3/3.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
```diff
@@ -105,28 +130,82 @@ func (s *DynPrefillScorer) Score(ctx context.Context, cycleState *schedtypes.Cyc
 		"endpointCount", len(endpoints),
 		"endpointsJSON", string(endpointsJSON))
 
-	result, err := dynscorer.CallRoutePrefillRequest(requestJSON, endpointsJSON)
+	resultCh := make(chan prefillReservationResult, 1)
+	bookingID := booking.ID
+	go func() {
+		result, err := s.reservePrefill(bookingID, requestJSON, endpointsJSON)
```
tmonty12: [P1] Moving `reservePrefill` to a goroutine makes `Score` observe cancellation, but this also removes the previous five-second cap while the call still crosses synchronous CGO and blocks an OS thread in Rust `Runtime::block_on`. With queueing enabled and a long-lived or deadline-free client context, each pending admission can retain an OS thread, goroutine, token buffer, and registry entry indefinitely; the synthetic profile's queue limits are unset. Can we preserve a hard admission budget and bounded Go-side concurrency—ideally via asynchronous/polling FFI—and add an overload test that holds workers busy and asserts bounded thread/RSS growth?

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,636 @@
...
+                matches!(entry, PrefillReservationEntry::Active(active)
+                    if now.saturating_duration_since(active.created_at) >= retention)
+            })
+            .map(|(reservation_id, _)| reservation_id.clone())
+            .collect()
+    }
+
+    fn remove_expired_cancellations(&self, now: Instant, retention: Duration) {
+        self.entries.lock().retain(|_, entry| {
+            !matches!(entry, PrefillReservationEntry::Cancelled { created_at }
```
tmonty12: [P2] `Cancelled` tombstones are reaped after 60–90 seconds, but `route_prefill_request_with_reservation` synchronously parses and tokenizes before `reserve_prefill_worker` reaches `begin()`. If preprocessing or goroutine scheduling exceeds that retention window, the tombstone disappears; `begin()` sees a vacant entry and queues an already-cancelled request, now without the previous timeout. Can we register `Pending` at C entry before preprocessing, or otherwise tie cancellation lifetime to the actual in-flight call, and add a paused-time test where pre-registration work outlives tombstone retention?

tmonty12 (commented):
Follow-up review of `91e2f264`: two additional cancellation liveness/resource findings inline. Focused Rust and Go tests plus the C-binding check passed.

deploy/inference-gateway/epp/pkg/plugins/disagg/decode_scorer.go
tmonty12: Addressed in `588af093`. `bookingLifecycle` starts exactly one marker goroutine with three attempts and 100/200 ms backoff. EOS/cancellation stops pending backoff and schedules cleanup off the response callback. Cleanup waits only in its background worker for an in-flight (five-second-bounded) mark before issuing the one `free_request`, so EOS does not stall. Added persistent-failure and in-flight-mark EOS coverage.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
tmonty12: Addressed in `588af093`. Prefill now creates and arms one `bookingLifecycle`; `DynDecodeScorer.PreRequest` no longer installs its own cancellation watcher. Decode error paths and EOS re-enter the lifecycle's `sync.Once` cleanup, which stops the watcher and marker before one release. Added EOS-then-cancellation coverage asserting one free.

lib/llm/src/kv_router/prefill_router/reservations.rs
tmonty12: I changed reservation retention to derive from the canonical active-request expiry plus the existing 60s grace, with 30s and 3600s regression cases. CI has since reported that the public import path used for that helper is not available from this crate, so this part needs a small export-path correction before merge.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
tmonty12: Addressed the immediate liveness/resource issue in `588af093`: admission again has an explicit 5s default budget (configurable `reservationTimeoutSeconds`) and a 32-admission CGO semaphore (configurable `maxConcurrentReservations`). The permit is acquired before the goroutine/CGO call, so saturation falls back without creating more blocked OS threads. The added saturation test asserts that only the configured in-flight call starts. I kept the synchronous ABI; polling FFI is not needed to establish the bound.

lib/llm/src/kv_router/prefill_router/reservations.rs
tmonty12: Addressed in `588af093`. Go now calls the fast `begin_prefill_reservation` ABI before spawning the blocking route goroutine, and the C route also registers `Pending` before preprocessing. Cancellation therefore targets a real pending entry rather than a timed tombstone; preprocessing failure explicitly aborts it. Added coverage showing a pre-registered cancelled pending entry survives tombstone reaping.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+        };
+
+        let InnerPrefillRouter::KvRouter(router) = &binding.router else {
+            self.remove_pending(reservation_id);
+            return Err(anyhow::anyhow!(PrefillError::NotActivated));
+        };
+        let chooser = router.chooser.clone();
+        let scheduler_id = scheduler_id(reservation_id);
+
+        let outcome = tokio::select! {
```
tmonty12: [P1] Cancellation can leave admitted scheduler state behind in the default configuration. `new_request_lifecycle_lease` returns `None` when queueing is disabled, so if the actor books and delivers the response immediately before this biased cancellation branch wins, dropping the delivered result has no lease-driven rollback and this manager removes `Pending` without recording `Active`. The phantom load remains until scheduler stale expiry. Please create a lifecycle lease for every `TrackedWithLifecycle` request regardless of queueing mode, service its cleanup queue in the disabled mode too, and add a deterministic booked-response/cancellation-wins regression test.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
```diff
@@ -105,28 +187,91 @@ func (s *DynPrefillScorer) Score(ctx context.Context, cycleState *schedtypes.Cyc
...
+		if cancelErr := s.cancelPrefill(bookingID); cancelErr != nil {
+			logger.V(logutil.DEFAULT).Error(cancelErr, "DynPrefillScorer: failed to cancel pending prefill reservation",
+				"bookingID", bookingID)
+		}
+		go func(bookingID string) {
+			outcome := <-resultCh
+			if outcome.err != nil {
+				return
+			}
+			if cleanupErr := s.freeBooking(bookingID); cleanupErr != nil {
```
tmonty12: [P1] This late-result cleanup calls the combined `free_request`. After the admission timeout returns aggregated fallback, decode `PreRequest` can register the same booking ID before this result arrives; this call then removes the live fallback decode booking along with the late prefill reservation. The request continues running while the router undercounts its decode worker. Please use a prefill-only late-release operation here, or assign the fallback decode path a distinct booking ID.

deploy/inference-gateway/epp/pkg/plugins/disagg/decode_scorer.go
```diff
@@ -198,47 +214,92 @@ func (s *DynDecodeScorer) Score(ctx context.Context, cycleState *schedtypes.Cycl
...
 		logger.V(logutil.DEFAULT).Error(parseErr, "DynDecodeScorer PreRequest: invalid worker ID",
-			"requestID", request.RequestId, "workerID", state.WorkerID)
+			"bookingID", bookingID, "workerID", state.WorkerID)
+		s.cleanupBooking(ctx, bookingID, "decode worker ID invalid")
 		return
 	}
 
-	if addErr := dynscorer.CallAddRequest(
-		request.RequestId,
+	if addErr := s.addRequest(
```
tmonty12: [P1] Decode registration is not serialized with lifecycle cleanup. Cancellation can run `free_request` while this blocking call is computing overlap: free observes no decode booking, then `addRequest` installs one after `cleanupOnce` has already been consumed. The error path has the inverse problem: it cleans the reservation, but GAIE's `PreRequest` hook cannot abort dispatch, so the request is still forwarded without bookkeeping. Please add an explicit `RegisteringDecode` lifecycle phase/barrier so cleanup waits or compensates after registration, and move registration failure handling to a boundary that can reject or redirect dispatch.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
```diff
@@ -155,6 +216,163 @@ func getEnvBoolOrDefault(key string, def bool) bool {
...
+		}
+		if stopMarker != nil {
+			stopMarker()
+		}
+
+		go func() {
+			if markerDone != nil {
+				<-markerDone
+			}
+			if err := l.freeBooking(l.bookingID); err != nil {
```
tmonty12: [P2] A failed or timed-out `freeBooking` still closes `cleanupDone` and deletes the lifecycle below. If EOS is the final callback, there is no retry; partial cleanup can leave prefill or decode load charged until stale expiry. `sync.Once` also permanently prevents the same owner from retrying. Please retain a single cleanup owner, perform bounded idempotent retries through a bounded executor, delete only after success, and keep a terminal tombstone/metric if retries are exhausted.

tmonty12 (commented):
Deep review identified four remaining lifecycle issues. Findings 1–3 are correctness blockers involving cancellation, fallback cleanup, and decode registration. Finding 4 is a cleanup reliability issue. The previously discussed C ABI concern is intentionally excluded.

lib/llm/src/kv_router/prefill_router/reservations.rs
tmonty12: Fixed in `cab8cacca5`: tracked requests now retain a lifecycle lease even when queueing is disabled, and dropping it retracts an immediately admitted assignment. Added a regression for the admission-wins/cancellation-wins race.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
tmonty12: Fixed in `cab8cacca5`: the late-result drain now calls `release_prefill_reservation`, which releases only the prefill scheduler booking and cannot remove fallback decode bookkeeping.

deploy/inference-gateway/epp/pkg/plugins/disagg/decode_scorer.go
tmonty12: Fixed in `cab8cacca5`: decode registration moved into `Score` behind a lifecycle barrier, so cleanup waits for `add_request`; registration failure removes routing headers and redirects to aggregate routing.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
tmonty12: Fixed in `cab8cacca5`: cleanup now has one owner, retries bounded idempotent release three times on a 32-call executor, deletes only after success, and retains an exhausted tombstone.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
```diff
@@ -105,28 +187,91 @@ func (s *DynPrefillScorer) Score(ctx context.Context, cycleState *schedtypes.Cyc
...
+				logger.V(logutil.DEFAULT).Error(cleanupErr, "DynPrefillScorer: failed to release late prefill reservation",
+					"bookingID", bookingID)
+			}
+		}(bookingID)
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15, building on cab8cacca5: the late-admission drain uses the prefill-only release operation, which cannot remove the decode booking sharing this ID. The rollback path runs only after decode booking failure or cancellation, before a successfully scheduled request can reach PreRequest.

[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15, building on cab8cacca5: the late-admission drain uses the prefill-only release operation, which cannot remove the decode booking sharing this ID. The rollback path runs only after decode booking failure or cancellation, before a successfully scheduled request can reach PreRequest.

deploy/inference-gateway/epp/pkg/plugins/disagg/prefill_scorer.go
```diff
@@ -105,28 +187,91 @@ func (s *DynPrefillScorer) Score(ctx context.Context, cycleState *schedtypes.Cyc
...
-	}
 	req.Headers[PrefillWorkerIDHeader] = prefillWorkerID
 	if result.DpRank != dynscorer.UnsetDpRank {
 		req.Headers[PrefillDpRankHeader] = strconv.FormatUint(uint64(result.DpRank), 10)
 	} else {
 		delete(req.Headers, PrefillDpRankHeader)
 	}
 
+	lifecycle := registerBookingLifecycle(booking.ID, s.freeBooking)
+	lifecycle.armCancellation(ctx)
```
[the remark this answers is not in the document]
tmonty12: Verified against the pinned gateway-api-inference-extension v1.2.1 source: Director calls scheduler Schedule with this context, then prepareRequest and PreRequest with the same context. Score is not passed a short-lived profile context. No code change is needed.

deploy/inference-gateway/epp/pkg/plugins/disagg/decode_scorer.go
```diff
@@ -248,34 +309,19 @@ func (s *DynDecodeScorer) PreRequest(ctx context.Context, request *schedtypes.In
...
-			logger.V(logutil.VERBOSE).Info("DynDecodeScorer ResponseBody: freed request",
-				"requestID", request.RequestId)
-		}
+	if response.EndOfStream {
+		lifecycle.cleanup(ctx, "response end of stream")
+		return
 	}
+
+	lifecycle.startPrefillMarker(s.markPrefillComplete, log.FromContext(ctx), request.RequestId)
 }
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: ResponseBody ignores requests without a Score-created lifecycle and starts the marker only when the routing-mode header is disaggregated. EOS still cleans up an existing lifecycle for either routing mode; a header-only request owns no booking to release.

deploy/inference-gateway/epp/pkg/plugins/disagg/reservation_test.go
```diff
@@ -0,0 +1,402 @@
...
+		},
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	scorer.PreRequest(ctx, requestWithBooking("external-request", bookingID), nil)
+
+	if freeCalls != 1 {
+		t.Fatalf("free calls = %d, want 1", freeCalls)
+	}
```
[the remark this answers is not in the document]
tmonty12: Already addressed in cab8cacca5: this test waits on the release callback channel before asserting the booking ID, so it no longer reads asynchronous cleanup state unsafely.

deploy/inference-gateway/epp/pkg/plugins/disagg/reservation_test.go
```diff
@@ -0,0 +1,402 @@
...
+	lifecycle := findBookingLifecycle(bookingID)
+	if lifecycle == nil {
+		t.Fatal("expected booking lifecycle after first response chunk")
+	}
+
+	started := time.Now()
+	scorer.ResponseBody(context.Background(), request, &rc.Response{EndOfStream: true}, nil)
+	if elapsed := time.Since(started); elapsed > 100*time.Millisecond {
+		t.Fatalf("EOS callback blocked for %s waiting on prefill mark", elapsed)
+	}
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: the EOS call now runs in a goroutine and the test uses a one-second completion guard, preserving the liveness assertion without a scheduler-sensitive 100 ms deadline.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
```diff
@@ -67,6 +73,61 @@ func readPrefillEnabled(cycleState *schedtypes.CycleState) bool {
...
+func bookingIDFromRequest(request *schedtypes.InferenceRequest) string {
+	if request == nil || request.Headers == nil {
+		return ""
+	}
+	bookingID := request.Headers[BookingIDHeader]
+	if _, err := uuid.Parse(bookingID); err != nil {
+		return ""
+	}
+	return bookingID
+}
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: Score overwrites the booking header with the CycleState-owned UUID, and ResponseBody ignores a header-only ID without a controller-created lifecycle. bookingIDFromRequest has no other production caller, so a skipped Score cannot reach FFI bookkeeping.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
```diff
@@ -155,6 +216,163 @@ func getEnvBoolOrDefault(key string, def bool) bool {
...
+	return lifecycle
+}
+
+func findBookingLifecycle(bookingID string) *bookingLifecycle {
+	lifecycle, ok := bookingLifecycles.Load(bookingID)
+	if !ok {
+		return nil
+	}
+	return lifecycle.(*bookingLifecycle)
+}
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: successful cleanup deletes the registry entry before closing cleanupDone. I intentionally did not re-store a fresh lifecycle for the same UUID: booking IDs are single-use, and a terminal failure tombstone prevents unrelated later paths from acquiring cleanup ownership.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+        cancellation: CancellationToken,
+    },
+    Active(ActivePrefillReservation<Sel>),
+    /// A cancellation can arrive from Go before the blocking reserve call
+    /// reaches Rust. Preserve that cancellation briefly so the later reserve
+    /// observes it instead of queueing.
+    Cancelled {
+        created_at: Instant,
+    },
+}
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: Pending records created_at. The reaper cancels expired admissions and converts them to a short-lived cancellation tombstone, so delayed reserve calls fail rather than queue and stale entries cannot persist indefinitely.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+                    entry.remove();
+                    BeginReservation::Cancelled
+                }
+                PrefillReservationEntry::Pending { cancellation } => {
+                    BeginReservation::Pending(cancellation.clone())
+                }
+                PrefillReservationEntry::Active(_) => BeginReservation::AlreadyExists,
+            },
+        }
+    }
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: Pending has an atomic claimed state. claim_pending permits exactly one scheduler admission; a concurrent reserve receives AlreadyExists without removing, activating, or freeing the owner reservation.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+                let reservation = ActivePrefillReservation {
+                    chooser: chooser.clone(),
+                    scheduler_id: scheduler_id.clone(),
+                    worker,
+                    created_at: Instant::now(),
+                };
+                if !self.activate(reservation_id, reservation) {
+                    ignore_missing_request(chooser.free_if_worker(&scheduler_id, worker).await)?;
+                    anyhow::bail!("prefill reservation {reservation_id:?} was cancelled");
+                }
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: cancellation now transitions through Activation::Cancelled after first recording Active. A release error leaves that active entry in the registry for reaper retry. Activation::Lost is no longer a cancellation race path.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+                for reservation_id in expired {
+                    if let Err(error) = manager.release(&reservation_id).await {
+                        tracing::warn!(
+                            %reservation_id,
+                            %error,
+                            "Failed to expire stale EPP prefill reservation"
+                        );
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: each reaper release is bounded by a five-second timeout. Timeout and release errors are logged while the loop continues with the remaining stale reservations.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,693 @@
...
+        manager: &EppReservationManager<DefaultWorkerSelector>,
+        router: &PrefillRouter<DefaultWorkerSelector>,
+        id: &str,
+    ) -> WorkerWithDpRank {
+        routed_worker(
+            manager
+                .reserve(
+                    router,
+                    id,
+                    &vec![1; 64],
```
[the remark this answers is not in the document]
tmonty12: Fixed in c32ce78b15: the borrowed Vec was replaced with an array-backed slice, avoiding the needless allocation and the useless_vec lint.

lib/kv-router/src/scheduling/queue.rs
```diff
@@ -587,9 +587,6 @@ impl<
         &self,
         request_id: Option<&str>,
     ) -> Option<Box<RequestLifecycleLease>> {
-        if !self.queueing_enabled {
-            return None;
-        }
         request_id?;
```
PeaBrane: AI-reviewed, non-blocking: lifecycle leases are now created while queueing is disabled, but `Drop` still relies on an ignored best-effort `try_send(Cleanup)`. If the bounded actor channel is full, `pending` remains set, later drops do not send another wake, and the opportunistic drain remains gated by `queueing_enabled`. That can leave phantom prefill load until stale reaping and let the dirty queue grow. Please consider draining cleanup regardless of queue configuration or making failed wakes retryable; a saturated capacity-one regression test would cover this.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
```diff
@@ -155,6 +216,222 @@ func getEnvBoolOrDefault(key string, def bool) bool {
...
+		l.mu.Unlock()
+		return
+	}
+	markerCtx, stopMarker := context.WithCancel(context.Background())
+	markerDone := make(chan struct{})
+	l.stopMarker = stopMarker
+	l.markerDone = markerDone
+	l.mu.Unlock()
+
+	go func() {
```
PeaBrane: AI-reviewed, non-blocking: each first token starts an unconstrained goroutine, while the C marker call can block up to five seconds and is retried three times. During a router stall, concurrent marker goroutines and CGO calls grow with request rate. Could this use the same bounded lifecycle executor as cleanup?

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
```diff
@@ -155,6 +216,222 @@ func getEnvBoolOrDefault(key string, def bool) bool {
...
+	l.mu.Unlock()
+
+	if stopCancellation != nil {
+		stopCancellation()
+	}
+	if stopMarker != nil {
+		stopMarker()
+	}
+
+	go func() {
```
PeaBrane: AI-reviewed, non-blocking: the semaphore bounds only active `free_request` calls because a goroutine is created before acquiring a slot. Sustained failures therefore create an unbounded waiting-goroutine backlog, and the exhausted path retains every lifecycle tombstone indefinitely. A fixed worker pool or bounded queue, plus bounded tombstone retention or background retry and metrics, would cap EPP memory during outages.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,908 @@
...
+            loop {
+                interval.tick().await;
+                let Some(manager) = manager.upgrade() else {
+                    return;
+                };
+                let now = Instant::now();
+                manager.expire_pending(now, retention);
+                manager.remove_expired_cancellations(now, CANCELLED_RESERVATION_RETENTION);
+                let expired = manager.expired_active_ids(now, retention);
+                for reservation_id in expired {
```
PeaBrane: AI-reviewed, non-blocking: stale releases are time-bounded but still serial. With N stalled reservations, one sweep can take N × 5 seconds, delaying later stale entries and recovery. Bounded concurrency here would preserve resource limits while preventing sweep time from scaling linearly with the timeout.

deploy/inference-gateway/epp/pkg/plugins/disagg/reservation_test.go
```diff
@@ -0,0 +1,628 @@
...
+)
+
+func requestWithBooking(externalRequestID, bookingID string) *schedtypes.InferenceRequest {
+	return &schedtypes.InferenceRequest{
+		RequestId: externalRequestID,
+		Headers:   map[string]string{BookingIDHeader: bookingID},
+	}
+}
+
+func TestBookingStateDoesNotTrustExternalRequestID(t *testing.T) {
```
PeaBrane: AI-reviewed, non-blocking: this new lifecycle suite is not executed by current CI. The EPP Makefile has a manual `go test`, while the Docker builder only runs `go build`, which does not compile `_test.go`. Please add an EPP Go-test step after the Rust static library is available so cancellation, EOS, and retry regressions can fail CI.

lib/llm/src/kv_router/prefill_router/reservations.rs
```diff
@@ -0,0 +1,908 @@
...
+                .unwrap(),
+        )
+    }
+
+    #[tokio::test]
+    async fn reservations_affect_selection_and_release_load() {
+        let (router, chooser) = tracked_prefill_router().await;
+        let manager = EppReservationManager::default();
+
+        let first = reserve(&manager, &router, "reservation-1").await;
```
PeaBrane: AI-reviewed, non-blocking: this core atomic-booking test awaits reservation 1 before starting reservation 2, so it would pass even if selection and booking became split again. Starting concurrent reservations behind a barrier would directly protect the race this PR fixes; the same approach can verify one owner for concurrent claims of the same ID.

PeaBrane (approved):
✅ Approved. This PR and the inline follow-up observations were AI-reviewed with multiple review agents at `c32ce78b15`. The previously raised bot lifecycle comments are addressed; the remaining notes are non-blocking hardening and test-coverage follow-ups.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
tmonty12: Fixed in `fa3e7bcbc0`. Cleanup now uses a dedicated 32-worker pool and bounded 4,096-entry queue, overflow is retained for reconciler submission, failed work is retried, and exhausted tombstones expire after a finite retention window aligned with the Rust stale-booking fallback.

deploy/inference-gateway/epp/pkg/plugins/disagg/shared.go
tmonty12: Fixed in `fa3e7bcbc0`. First-token work now uses a dedicated fixed pool of 32 workers and a bounded 1,024-entry queue; enqueue overflow is recorded and terminal cleanup remains the fallback release path, so marker stalls cannot create unbounded goroutines or CGO calls.

lib/llm/src/kv_router/prefill_router/reservations.rs
tmonty12: Fixed in `7b90075122`. Each tick claims at most 128 stale reservations and releases up to 32 concurrently; with the five-second per-release timeout, a full batch is bounded to four waves rather than `N × 5s`. Failed entries remain retryable and least-attempted claims prevent starvation.

lib/llm/src/kv_router/prefill_router/reservations.rs
tmonty12: Will implement in follow-up

lib/llm/src/kv_router/prefill_router/query.rs
```diff
@@ -1,19 +1,134 @@
...
+                priority_jump,
+                strict_priority,
+                None,
+                None,
+                allowed_worker_ids,
+                routing_constraints,
+            )
+            .await?;
+        match outcome {
+            crate::kv_router::FindBestMatchOutcome::Routed { worker, .. } => {
```
tmonty12, line 136: [P1] Cancellation can permanently retain an admitted prefill booking

`find_best_match_details_with_lifecycle` only receives a `RequestLifecycleLease` when `SchedulerQueue::new_request_lifecycle_lease` sees queueing enabled. With the default non-queue configuration it returns `None`. If the scheduler books a worker and delivers the response immediately before this future is dropped (for example, the ext-proc stream is cancelled), no caller observes the returned booking and nothing rolls it back, leaving `potential_prefill_tokens` charged until stale expiry. Please create and service the lifecycle lease for every `TrackedWithLifecycle` admission regardless of queue mode, and cover the booked-response/caller-cancellation race.

deploy/inference-gateway/ext-proc/src/epp.rs
```diff
@@ -1568,40 +1656,42 @@ impl EndpointPicker for Router {
...
     }
 
-    async fn on_prefill_complete(&self, request_id: &str) {
-        if request_id.is_empty() {
+    async fn on_prefill_complete(&self, booking_id: &str) {
+        if booking_id.is_empty() {
             return;
         }
-        if let Err(e) = self.mark_prefill_complete(request_id).await {
+        release_prefill_booking(&self.prefill_bookings, booking_id).await;
```
tmonty12, line 1597: [P2] `stream:false` holds prefill load throughout decode

This release is driven by the first non-empty response-body chunk. For an OpenAI `stream:false` request, the backend normally emits one complete response only after decoding has finished, so the prefill reservation stays charged for the entire decode phase. Under sustained non-streaming traffic, workers therefore appear to retain prefill work long after prefill completed and subsequent routing is skewed. Please derive prefill completion from a phase signal independent of client-visible streaming, or explicitly avoid prefill-phase accounting for non-streaming requests; merely forcing the downstream request to stream would also require correctly folding the streamed response back into the requested non-streaming API shape.

lib/llm/src/kv_router/prefill_router/query.rs
```diff
@@ -1,19 +1,134 @@
...
+        }
+        if self.lifecycle_state() != PrefillLifecycleState::Active {
+            return Err(anyhow::anyhow!(PrefillError::NotActivated));
+        }
+        let binding = self
+            .binding
+            .load_full()
+            .ok_or_else(|| anyhow::anyhow!(PrefillError::NotActivated))?;
+        let Some(chooser) = binding.router.kv_router_if_enabled() else {
+            anyhow::bail!("prefill reservation requires KV routing");
```
tmonty12: [P1] Non-KV prefill modes silently lose disaggregation

Prefill activation accepts advertised built-in modes and constructs a `RoutingHost` for them, but this new reservation path hard-errors unless that host is KV-backed. Native EPP treats every prefill-routing error as aggregated fallback, so existing RoundRobin/Random/LeastLoaded/P2C/DeviceAware prefill configurations silently stop using their prefill workers. A small compatibility fix is to preserve the existing non-KV query path and return an untracked/no-op reservation for those modes while keeping atomic token admission KV-only.

lib/llm/src/kv_router/prefill_router/query.rs
```diff
@@ -1,19 +1,134 @@
...
+    pub fn worker_id(&self) -> WorkerId {
+        self.worker.worker_id
+    }
+
+    pub fn dp_rank(&self) -> u32 {
+        self.worker.dp_rank
+    }
+
+    /// Free this booking from the chooser that admitted it.
+    pub async fn release(&self) -> Result<()> {
```
tmonty12: [P2] `PrefillReservation` does not own its cleanup contract

This public type represents an admitted booking but can be dropped without releasing it; cleanup only works because EPP wraps it temporarily in a separate guard and later remembers to remove it from a side map. Any other caller or ownership path can leak the booking. Please make the reservation itself RAII: retain an exact attempt-scoped `SchedulerBookingDescriptor` plus `SchedulerBookingCleanup`, enqueue cleanup synchronously from `Drop`, and make explicit `release(self)` consume the owner and await the cleanup acknowledgement. That also lets the EPP-specific guard disappear; avoid relying on `tokio::spawn` from `Drop`.

tmonty12 (commented):
Deep review: four remaining lifecycle and compatibility findings.

deploy/inference-gateway/ext-proc/src/epp.rs
tmonty12: `stream: false` is a minority of requests so we don't need to over-optimize for this case now. 

not booking prefill load at all would result in the original issue. The Gateway owns the response stream so setting `stream: true` on the request body would change the clients expected response format. It's worth looking into `PrefillLoadEstimator` that can decay the prefill load over time instead of relying on an explicit prefill_complete on first token hook. this should be benchmarked and completed in a separate PR.

lib/llm/src/kv_router/prefill_router/query.rs
tmonty12: [AI] Fixed in d737a91a04. `TrackedWithLifecycle` now creates a lifecycle lease even when queueing is disabled; its regression drops an admitted non-queue lease and verifies scheduler load drains.

lib/llm/src/kv_router/prefill_router/query.rs
tmonty12: [AI] Fixed in 28e22d0f1b. Built-in prefill modes now select through the existing advisory path and return an untracked no-op reservation, preserving disaggregation while leaving KV admission atomic.

lib/llm/src/kv_router/prefill_router/query.rs
tmonty12: [AI] Fixed in dc312cd752. `PrefillReservation` now owns an attempt-fenced booking descriptor, queues cleanup from `Drop`, and consumes `release()`; the EPP `tokio::spawn` guard is removed.

deploy/inference-gateway/ext-proc/src/epp.rs
```diff
@@ -434,31 +455,31 @@ impl Router {
...
 
-    /// Route a prefill request. Returns (worker_id, dp_rank).
+    /// Atomically select and reserve a prefill worker.
     ///
     /// Queue priorities are forwarded to the prefill scheduler. `priority_jump`
     /// adjusts the policy score, while `strict_priority` selects the primary
     /// tier. `routing_constraints` carries the request's required/preferred
     /// taints (lifted from `nvext.routing_constraints`); a hard `required_taints`
     /// mismatch excludes a worker from selection.
+    #[expect(clippy::too_many_arguments)]
```
ishandhanani, line 465: why do we need this now? - just cause of 2 extra args?

tmonty12: yeah 1 extra arg, clippy threshold is 7

ishandhanani (approved):

nv-anants (approved):

PR #13050, opened 2026-08-11
ci(docs): regenerate API references on main after a merge

dagil-nvidia:
## Summary

`api_freshness.py` scopes the per-PR gate so a branch is only failed for drift it caused. That is
right. But it rests on the second half of its own docstring — *"the next regeneration on main
resolves it"* — and **nothing regenerates main**.

Drift accumulates until a human notices. Twice on 2026-08-11 alone (#12985, #13035), by two
different people. The cost is not a red check: the strict check runs **before** the sync/publish
steps in `fern-docs.yml`, so stale output skips the publish and **docs.nvidia.com stops updating**.

This job is the missing half. It runs after every merge to `main`, regenerates, and pushes.

## How drift actually arrives

Not from PRs that skip regeneration — the gate catches those. It arrives through **merge skew**,
which no per-PR check can see:

> PR **A** regenerates against `main@T1` and goes green.
> PR **B** lands at `T2`, adding a public symbol.
> PR **A** merges at `T3` without re-running.
> `main` is stale, and **neither PR was at fault**.

#13035's diff confirms it: `362 → 365` public symbols, `88 → 89` in `_core`, `"74 classes and 14
functions" → "15 functions"`. Real API surface landing between one PR's check and its merge.

## Push behaviour

Pushes straight to `main` when it can, falls back to a branch when the rules reject it, so it is
correct under either permission state and needs no flag day.

Direct push is **unproven unattended**. One manual `workflow_dispatch` succeeded on 2026-08-11 at
23:57Z (landing `3d77ee344` for `community-events-refresh.yml`, its first success after five
failures) — but that was nine minutes after ruleset `3932713` was modified, and the config today
still shows a `pull_request` rule with `bypass_actors: null` on `~DEFAULT_BRANCH`. The next
scheduled run exited early with *"No calendar change"*, so it never exercised the push. The
fallback is therefore load-bearing, not insurance.

## Guards

| Guard | Why |
|---|---|
| **Blast-radius cap** — 20 files / 2000 lines | The happy path pushes to `main` unreviewed, so an unbounded diff is the one way this can do real damage. A merge that moves a symbol touches a handful of pages; a *broken generator* rewrites all of them. Real regenerations for scale: #12985 was 4 files, #13035 was 6. Over the cap, the direct push is skipped and it goes to a branch. |
| **Fails on fallback**, routed to `notify-slack.yml` | With the publish gate staying blocking, a fallback is a **publish outage**, not a nuisance — `main` carries stale references until someone opens that PR. A green run hides it, which is how `update-events.yml` failed 100 times unnoticed. |
| **Loop guard** — `github.actor != 'dynamo-ops'` | This triggers on push to `main` *and* pushes to `main` with a PAT, and PAT pushes re-trigger workflows where `GITHUB_TOKEN` pushes do not. |

Deliberate asymmetry: pushing to `main` is a success, falling back is a failure. Both leave the
tree correct; only one leaves the publish blocked.

## Why not the alternatives

- **Relax the publish gate** (`continue-on-error`) — tried in #13056, **closed**. The gate is also
  the only forcing function; drift got fixed twice today precisely because the publish broke and
  someone felt it. Nothing notifies on drift, so that change would have traded a loud signal for
  silence.
- **`peter-evans/create-pull-request`** — `"GitHub Actions is not permitted to create or approve
  pull requests"`. `update-events.yml` does this and has failed every run since 2026-07-17.
- **Widen the per-PR gate to unscoped** — restores exactly the noise `api_freshness.py` was built
  to remove, and still misses skew, which is only visible after the merge.

## Validation

- YAML parses; both jobs resolve; every `run:` block passes `bash -n`.
- Only the pinned `checkout` and `setup-python` actions are used.
- The new file matches the `ci` filter, so the changed-files coverage gate passes.
- A dry run of all four generators against current `main` produces no diff — the job correctly
  exits without pushing when main is clean.

## Known limitation

`gen_kubernetes_api.py` re-renders a committed intermediate that `make generate-api-docs` produces
from the operator Go types. That first hop needs Go and `crd-ref-docs`, so it stays a human's job.
Running the renderer here still catches renderer-side drift, but not Go-type drift.


dagil-nvidia:
Updated after pressure-testing this. Three changes, and one correction to the record.

**Correction.** An earlier revision of this description claimed direct push to `main` was blocked because two rulesets each carry a `pull_request` rule and GitHub enforces the union, so `@ai-dynamo/autobots` needed bypass on `3932713`. That was wrong, and then my retraction of it was also wrong. What actually happened: ruleset `3932713` was modified at `23:48Z`, a manual dispatch succeeded at `23:57Z`, and the config today still shows `pull_request` with `bypass_actors: null`. I cannot see into that window, so direct push is recorded as **unproven unattended** rather than as either working or blocked.

The scheduled run I expected to settle it exited early on *"No calendar change"* — my own earlier dispatch had already pushed the calendar, so there was nothing left to push. The experiment tested nothing.

**Guards added**, both of which matter more now that #13056 is closed and the publish gate stays blocking:

- **Blast-radius cap** (20 files / 2000 lines). The happy path pushes unreviewed, so an unbounded diff is the one way this does real damage — a broken generator rewrites every page. Over the cap it skips the direct push and goes to a branch.
- **Fails on fallback**, wired to `notify-slack.yml`. A fallback is now a publish outage, and a green run would hide it.

PR #13052, opened 2026-08-11
fix(mocker): truncate disagg prefill output plan

jthomson04:
## Overview

Fix offline disaggregated replay so the prefill source emits one hidden bootstrap token and the decode destination emits the complete visible output plan. This prevents both stages from generating the full planned output.

## Details

- Add private `DirectRequest` helpers that make `output_token_ids.len()` authoritative when a plan exists and otherwise use `max_output_tokens`.
- Use that effective length in mocker scheduling, routing, arrival accounting, trace artifacts, and terminal accounting.
- Build the source-stage request as a new one-token clone with only the first planned token and exact-size output-plan storage. The stored original request remains unchanged for decode.
- Keep the router's one-token prefill estimate as a stage-specific override.
- Preserve the vLLM mismatch warning, zero-output behavior, backend behavior, public formats, and decode-only public disaggregated reports.

The source token is an internal handoff bootstrap token. It is not reported as visible output. The decode destination retains and emits the complete planned output.

Start with the shared request helpers in `lib/mocker/src/common/protocols.rs`, then the source-stage clone in `lib/mocker/src/replay/offline/state.rs`. The cross-backend conformance fixture is in `lib/mocker/src/replay/offline/entrypoints.rs`.

## Summary

- Centralize output-plan length and prefix-clone behavior.
- Prevent full-plan ownership from leaking into the disaggregated prefill request.
- Make aggregate and disaggregated accounting follow the same planned length as the schedulers.
- Strengthen vLLM and SGLang regression coverage for planned, empty, mismatched, and destination output behavior.

## Validation

- Exact PR-base regression proof at `b4f66adf11`: the two-token conformance fixture fails with `normalized source output count mismatch: expected 1, got 2`.
- Fixed head `6f61740f04`: `cargo test -p dynamo-llm live_and_offline_handoff_surfaces_share_one_conformance_report` passes for vLLM and SGLang.
- `cargo test -p dynamo-mocker`: 722 passed.
- `cargo test -p dynamo-mocker --features kvbm-offload replay::offline`: 194 passed; two unrelated KVBM lifecycle tests fail and reproduce unchanged at `d95e98aa73`.
- `cargo clippy -p dynamo-mocker -p dynamo-llm --all-targets -- -D warnings`
- `git diff --check`

No performance comparison or old-versus-new parity run was used because this is a correctness repair and disaggregated timing is expected to change.

## Related Issues

Relates to #13058.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected prefill request handling to limit planned output tokens to the configured maximum.
  * Ensured requests retain exactly the allowed number of output tokens.

### DIFF
--- a/lib/mocker/src/replay/offline/state.rs
+++ b/lib/mocker/src/replay/offline/state.rs
@@ -184,6 +184,9 @@ impl DisaggRequestState {
     pub(crate) fn build_prefill_request(&mut self) -> Result<DirectRequest> {
         let mut request = self.materialize_original_request()?.clone();
         request.max_output_tokens = request.max_output_tokens.min(1);
+        if let Some(output_token_ids) = request.output_token_ids.as_mut() {
+            output_token_ids.truncate(request.max_output_tokens);
+        }
         Ok(request)
     }
 
@@ -642,6 +645,11 @@ mod tests {
         let request = state.build_prefill_request().unwrap();
 
         assert_eq!(request.max_output_tokens, 1);
+        assert_eq!(
+            request.output_token_ids.as_ref().map(Vec::len),
+            Some(1),
+            "prefill request must retain only its one planned output token"
+        );
         assert!(request.tokens[..64].iter().all(|token| *token == 21));
         assert!(request.tokens[64..].iter().all(|token| *token == 22));
         assert!(state.materialized_tokens().unwrap().is_some());
--- a/lib/mocker/src/replay/offline/state.rs
+++ b/lib/mocker/src/replay/offline/state.rs
@@ -32,7 +32,7 @@ pub(crate) struct AggRequestState {
 impl AggRequestState {
     pub(crate) fn new_queued(request: ReplayRequestPayload) -> Self {
         let input_tokens = request.input_length();
-        let output_tokens = request.metadata().max_output_tokens;
+        let output_tokens = request.metadata().effective_max_output_tokens();
         Self {
             request: Some(request),
             phase: AggRequestPhase::QueuedAtRouter,
@@ -182,12 +182,9 @@ impl DisaggRequestState {
     }
 
     pub(crate) fn build_prefill_request(&mut self) -> Result<DirectRequest> {
-        let mut request = self.materialize_original_request()?.clone();
-        request.max_output_tokens = request.max_output_tokens.min(1);
-        if let Some(output_token_ids) = request.output_token_ids.as_mut() {
-            output_token_ids.truncate(request.max_output_tokens);
-        }
-        Ok(request)
+        Ok(self
+            .materialize_original_request()?
+            .clone_with_output_limit(1))
     }
 
     pub(crate) fn take_replay_hashes(&mut self) -> Option<ReplayRequestHashes> {
@@ -630,6 +627,8 @@ mod tests {
         };
         let mut driver = WorkloadDriver::new_trace(trace, 64).unwrap();
         let compact = driver.pop_ready_compact(0.0, 1).pop().unwrap();
+        let original_output_token_ids =
+            compact.request.metadata().output_token_ids.clone().unwrap();
         let mut state = DisaggRequestState::new(
             compact.request,
             0.0,
@@ -646,9 +645,20 @@ mod tests {
 
         assert_eq!(request.max_output_tokens, 1);
         assert_eq!(
-            request.output_token_ids.as_ref().map(Vec::len),
-            Some(1),
-            "prefill request must retain only its one planned output token"
+            request.output_token_ids.as_deref(),
+            Some(&original_output_token_ids[..1]),
+            "prefill request must retain the original first planned output token"
+        );
+        let prefill_plan = request.output_token_ids.as_ref().unwrap();
+        assert_eq!(prefill_plan.capacity(), prefill_plan.len());
+        assert_eq!(
+            state
+                .original_request()
+                .unwrap()
+                .output_token_ids
+                .as_deref(),
+            Some(original_output_token_ids.as_slice()),
+            "decode must retain the complete original output plan"
         );
         assert!(request.tokens[..64].iter().all(|token| *token == 21));
         assert!(request.tokens[64..].iter().all(|token| *token == 22));
--- a/lib/mocker/src/common/protocols.rs
+++ b/lib/mocker/src/common/protocols.rs
@@ -299,6 +299,31 @@ impl DirectRequest {
     pub fn router_priorities(&self) -> (f64, u32) {
         (f64::from(self.priority.max(0)), self.strict_priority)
     }
+
+    #[inline]
+    pub(crate) fn effective_max_output_tokens(&self) -> usize {
+        self.output_token_ids
+            .as_ref()
+            .map_or(self.max_output_tokens, Vec::len)
+    }
+
+    pub(crate) fn clone_with_output_limit(&self, limit: usize) -> Self {
+        let max_output_tokens = self.effective_max_output_tokens().min(limit);
+        Self {
+            tokens: self.tokens.clone(),
+            max_output_tokens,
+            output_token_ids: self
+                .output_token_ids
+                .as_ref()
+                .map(|ids| ids[..max_output_tokens].to_vec()),
+            uuid: self.uuid,
+            dp_rank: self.dp_rank,
+            arrival_timestamp_ms: self.arrival_timestamp_ms,
+            priority: self.priority,
+            strict_priority: self.strict_priority,
+            policy_class: self.policy_class.clone(),
+        }
+    }
 }
 
 fn is_zero_i32(value: &i32) -> bool {
@@ -1571,6 +1596,39 @@ mod tests {
 
         assert_eq!(error.to_string(), "injected raw sink failure");
         assert_eq!(*sink.attempts.lock().unwrap(), vec![1, 2, 3]);
+    }
+
+    #[test]
+    fn direct_request_output_plan_is_authoritative_when_limited() {
+        let request = DirectRequest {
+            tokens: vec![1, 2],
+            max_output_tokens: 0,
+            output_token_ids: Some(vec![7, 8]),
+            ..Default::default()
+        };
+
+        assert_eq!(request.effective_max_output_tokens(), 2);
+        let limited = request.clone_with_output_limit(1);
+        assert_eq!(limited.max_output_tokens, 1);
+        assert_eq!(limited.output_token_ids.as_deref(), Some(&[7][..]));
+        assert_eq!(limited.output_token_ids.unwrap().capacity(), 1);
+        assert_eq!(request.output_token_ids.as_deref(), Some(&[7, 8][..]));
+
+        let empty_plan = DirectRequest {
+            max_output_tokens: 4,
+            output_token_ids: Some(Vec::new()),
+            ..Default::default()
+        };
+        assert_eq!(empty_plan.effective_max_output_tokens(), 0);
+        let limited = empty_plan.clone_with_output_limit(1);
+        assert_eq!(limited.max_output_tokens, 0);
+        assert_eq!(limited.output_token_ids.as_deref(), Some(&[][..]));
+
+        let unplanned = DirectRequest::default();
+        assert_eq!(unplanned.effective_max_output_tokens(), 0);
+        let limited = unplanned.clone_with_output_limit(1);
+        assert_eq!(limited.max_output_tokens, 0);
+        assert!(limited.output_token_ids.is_none());
     }
 
     #[test]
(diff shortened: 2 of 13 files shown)


lib/mocker/src/replay/offline/state.rs
```diff
@@ -184,6 +184,9 @@ impl DisaggRequestState {
     pub(crate) fn build_prefill_request(&mut self) -> Result<DirectRequest> {
         let mut request = self.materialize_original_request()?.clone();
         request.max_output_tokens = request.max_output_tokens.min(1);
```
jthomson04: `truncate` on the next line is passed the value this statement just clamped, rather than the literal budget. Two consequences worth a moment:

1. **Silent order dependency.** Swapping these two statements turns the truncate into a no-op that restores the original bug. The new unit test does catch that, but nothing in the code says why the order matters.
2. **`max_output_tokens: 0` now yields an empty plan.** For a request with a non-empty plan and a zero budget, this produces `Some(vec![])` — a prefill that generates no token at all — where the pre-PR code generated the whole plan. That's a silent behavior flip on inconsistent input, with no guard or comment.

`truncate(1)` plus a one-line comment noting that the plan length is what both schedulers treat as the output budget would remove the coupling and make the zero case explicit.

lib/mocker/src/replay/offline/state.rs
```diff
@@ -184,6 +184,9 @@ impl DisaggRequestState {
...
+        if let Some(output_token_ids) = request.output_token_ids.as_mut() {
```
jthomson04: This is now the fourth independent site resolving the `max_output_tokens` vs. `output_token_ids.len()` dual representation, each in its own direction and with no shared helper:

- `live.rs:300-308` — sets `max_output_tokens = plan.len()`
- `vllm/core.rs:900-909` — warns and overrides `max_output_tokens` with the plan length
- `sglang/core.rs:383` + `sglang/request.rs:29-32` — silently override via `map_or(max_output_tokens, Vec::len)`
- here — truncates the plan down to `max_output_tokens`

… and `KvReplayMetadata::for_prefill` (`components/admission.rs:59`) encodes the "prefill gets one token" rule a fifth time router-side.

The cost is that the next site to narrow `max_output_tokens` — a new admission clamp, or a `max_model_len` clamp mirroring `lib/llm/src/mocker.rs:861` — reintroduces exactly this bug unless its author happens to remember the plan. A single `DirectRequest::set_max_output_tokens(n)` that also truncates the plan (plus a `planned_output_len()` replacing the three copies of the `map_or(..., Vec::len)` expression) would make the invariant enforceable in one place. Not a blocker for this fix, but worth a follow-up.

lib/mocker/src/replay/offline/state.rs
```diff
@@ -184,6 +184,9 @@ impl DisaggRequestState {
...
+            output_token_ids.truncate(request.max_output_tokens);
```
jthomson04: Minor, but this is clone-then-discard on a hot path: line 185 deep-copies the full N-token `output_token_ids` out of the original payload, and `Vec::truncate` doesn't release capacity, so the resulting Vec keeps capacity for all N while using one element.

That Vec is then moved into `SglangRequest::planned_output_ids` (`sglang/request.rs:50`) or `RequestKvState`/`ActiveSequence` (`vllm/core.rs:938,948`) and stays resident for the whole prefill lifetime. A replay of 100k requests with 2k planned output tokens each holds roughly 800 MB of dead capacity across the prefill worker pool for no benefit.

Building the field directly avoids both the copy and the slack:

```rust
output_token_ids: original
    .output_token_ids
    .as_deref()
    .map(|ids| ids[..max.min(ids.len())].to_vec()),
```

or just `shrink_to_fit()` after the truncate.

lib/mocker/src/replay/offline/state.rs
```diff
@@ -184,6 +184,9 @@ impl DisaggRequestState {
...
+        }
```
jthomson04: Worth recording somewhere (comment or issue): this makes the offline prefill request carry `output_token_ids: Some([plan[0]])`, while the live path deliberately carries `None` — `lib/llm/src/mocker.rs:834` computes the replay key only for non-prefill workers (`(!is_prefill).then(...)`), so a live prefill worker falls back to `random::<u32>()` in `ActiveSequence::generate_token` (`common/sequence.rs:717`).

These are the two surfaces the handoff conformance harness checks against each other, and they now model different KV content for the source's generated token. When `prompt_len % block_size == block_size - 1`, the generated token completes a block: offline publishes a source block hashed over `prompt_tail + plan[0]`, which matches the decode worker's block, while live publishes one hashed over a random token that never matches. Offline KV-router replays will therefore report a prefill→decode reuse the live system can't achieve.

The offline behavior here is the more faithful one, so the fix belongs on the live side (plan the prefill's first token there too) rather than in this PR — but the divergence shouldn't stay implicit.

lib/mocker/src/replay/offline/state.rs
```diff
@@ -642,6 +645,11 @@ mod tests {
         let request = state.build_prefill_request().unwrap();
```
jthomson04, line 644: Nothing asserts that `self.original` still holds the **full** plan after this call. `build_prefill_request` takes `&mut self` and relies entirely on the `.clone()` at line 185 to protect the stored payload, which `prepare_destination_reservation` (`disagg.rs:438-443`) later clones again for the decode worker.

If someone "optimizes" that clone into an in-place mutation, the decode request's plan drops to 1 element too: `vllm/core.rs:900-909` then overrides `max_output_tokens` from 4 to 1 and the decode worker emits one token instead of four. Every existing test stays green — the workload-driven disagg tests (`disagg_tests.rs:1301,1358`) assert only phase and materialization, and `debug_assert!(actual <= requested)` at `disagg.rs:810` is satisfied by `1 <= 4`.

Suggest adding after this line:

```rust
assert_eq!(
    state.original_request().unwrap().output_token_ids.as_ref().map(Vec::len),
    Some(4),
    "building the prefill request must not mutate the stored plan"
);
```

plus a `total_output_tokens == 4` assertion on `destination_first_workload_materializes_for_decode_reservation_then_completes`.

lib/mocker/src/replay/offline/state.rs
```diff
@@ -642,6 +645,11 @@ mod tests {
...
 
         assert_eq!(request.max_output_tokens, 1);
+        assert_eq!(
+            request.output_token_ids.as_ref().map(Vec::len),
```
jthomson04: This checks the plan's length but not that the retained element is the *first* planned token, so a variant that keeps the wrong one passes. E.g. `*ids = ids.split_off(ids.len() - 1)` still satisfies `map(Vec::len) == Some(1)`.

At runtime that matters: the prefill worker's `ActiveSequence::generate_token` (`common/sequence.rs:714-717`) would emit `plan[N-1]` while the decode worker emits `plan[0]` first. For any prompt where `prompt_len % block_size == block_size - 1` the generated token completes a block, so the prefill worker publishes a Stored KV event hashed over `prompt_tail + plan[N-1]` and the decode worker one over `prompt_tail + plan[0]` — the KV-router replay then records a spurious cache miss and routes differently.

Asserting against the original plan pins the identity as well as the count:

```rust
assert_eq!(
    request.output_token_ids.as_deref(),
    state.original_request().unwrap().output_token_ids.as_deref().map(|ids| &ids[..1]),
);
```

dreamtalen (approved):

lib/mocker/src/replay/offline/state.rs
```diff
@@ -184,6 +184,9 @@ impl DisaggRequestState {
     pub(crate) fn build_prefill_request(&mut self) -> Result<DirectRequest> {
         let mut request = self.materialize_original_request()?.clone();
         request.max_output_tokens = request.max_output_tokens.min(1);
+        if let Some(output_token_ids) = request.output_token_ids.as_mut() {
+            output_token_ids.truncate(request.max_output_tokens);
+        }
         Ok(request)
```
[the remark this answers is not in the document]
dreamtalen: Prefill output is tracked separately for conformance capture, but it does not contribute to request output-length metrics. Therefore, the decode worker emitting the full N-token plan still results in an actual output length of N, not N+1.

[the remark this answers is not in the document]
jthomson04: Confirmed. The source token is an internal handoff bootstrap token and is not added to the public request output timeline. Decode emits the full visible plan, so disaggregated reports remain decode-only. Commit 6f61740f04 also makes arrival and terminal accounting use the same effective planned length, with mismatched-plan report tests.

lib/mocker/src/replay/offline/state.rs
jthomson04: Addressed in 6f61740f04. The state regression now asserts that the stored original retains the complete plan after the prefill request is built. `destination_first_workload_materializes_for_decode_reservation_then_completes` also asserts that decode reports all four planned output tokens.

lib/mocker/src/replay/offline/state.rs
jthomson04: Addressed in 6f61740f04. The prefill clone now builds the planned prefix directly from the original slice. It does not clone and truncate the full plan, and the regression asserts that the resulting vector capacity equals its one-token length.

lib/mocker/src/replay/offline/state.rs
jthomson04: Addressed in 6f61740f04. `DirectRequest::effective_max_output_tokens` makes any present plan authoritative, including a nonempty plan with declared zero and an empty plan with declared nonzero. `clone_with_output_limit(1)` removes the mutation-order dependency. Focused tests cover both zero cases.

lib/mocker/src/replay/offline/state.rs
jthomson04: Addressed in 6f61740f04. The state regression compares the prefill plan with the first element of the unchanged original plan, so it pins token identity, not only length. The two-token handoff fixture also exercises this path for vLLM and SGLang.

lib/mocker/src/replay/offline/state.rs
jthomson04: Addressed in 6f61740f04. The shared `DirectRequest::effective_max_output_tokens` and `clone_with_output_limit` helpers now define the invariant. Live preparation, both scheduler backends, replay routing, arrival accounting, trace artifacts, and terminal accounting use the helper. The router one-token estimate remains a separate stage override.

lib/mocker/src/replay/offline/state.rs
jthomson04: Tracked separately in #13058. This PR keeps the offline first-token behavior and does not change `lib/llm` live token selection. The follow-up will make live prefill replay the planned first token and add block-boundary coverage for vLLM and SGLang.

jthomson04 (commented):
The 3-line fix is correct and well-targeted: both schedulers do treat `output_token_ids.len()` as authoritative (`vllm/core.rs:899-909` warns and overrides `max_output_tokens`, `sglang/core.rs:383` and `sglang/request.rs:29` silently override), so truncating the plan alongside the `min(1)` is the right lever. `cargo test -p dynamo-mocker --lib` passes 719/719 and `cargo clippy -p dynamo-mocker --all-targets` is clean.

The notes below are about incompleteness and coverage rather than the patch being wrong. Two of them fall outside the diff, so they're here rather than inline:

---

**`lib/mocker/src/replay/offline/disagg.rs:810` — the terminal path still uses the other convention.**

This PR makes the plan authoritative at the prefill boundary, but `record_decode_terminal` still derives the *requested* output length from `original.max_output_tokens` while the schedulers derive the *actual* output from `output_token_ids.len()`. With `DirectRequest { max_output_tokens: 2, output_token_ids: Some(vec![1,2,3]), .. }` fed to `simulate_trace_disagg`, `normalize_trace_requests` (`replay/mod.rs:183`) does not reconcile the two fields the way `live.rs:300-308` does, so the decode worker generates 3 tokens and `debug_assert!(3 <= 2)` fires. In release, `traffic.on_completion(input, 3, ..)` is recorded against an arrival logged with `output_length = 2` (`disagg.rs:615`), so offered vs. served token counts in the report disagree. The aggregated path has the identical hole at `agg.rs:602` via `AggRequestState::new_queued`.

---

**`lib/mocker/src/replay/offline/entrypoints.rs:148` — the conformance fixture never exercises a planned output.**

`run_offline_handoff_conformance` builds `DirectRequest { tokens: 0..8, max_output_tokens: 2, ..Default::default() }`, i.e. `output_token_ids: None`. With no plan, the pre-existing `min(1)` alone already yielded exactly one source token, so `common/handoff.rs:237` (`source_output_tokens != 1 -> bail`) stayed green on the buggy base commit — which is why this shipped. Setting `output_token_ids: Some(vec![7, 8])` on that fixture makes the base commit fail with `source_output_tokens = 2` and guards the invariant end-to-end across both the offline and live surfaces (the live side already passes a 1-element plan via `request(request_id, 1)` in `handoff_tests.rs:881`). That's a stronger regression guard than the shape-level unit assertion this PR adds.

PR #13056, opened 2026-08-11
fix(docs): stop generated-reference drift from blocking the publish

dagil-nvidia:
## Summary

The generated-reference freshness check sits **third** in the publish job — ahead of the
docs-website checkout and every sync step — with no `continue-on-error`. When it fails, the job
aborts and **nothing publishes**.

That is inverted:

| | Result for readers |
|---|---|
| **Blocked (today)** | docs.nvidia.com keeps serving the *previous* state. Its API pages are **older** than the ones being withheld — and every unrelated docs change is withheld too. |
| **Published anyway** | API pages are slightly stale (e.g. `362` vs `365` symbols, an anchor at `#L2591` instead of `#L2602`). Everything else reaches users. |

Blocking does not protect readers from stale API docs. It guarantees they see strictly older ones,
and takes the rest of the site down with them. The failure it prevents is milder than the failure
it causes.

It is also **redundant**. The same three `--check` invocations already run as the
`Generated API References` job in `pre-merge.yml`, which is in `pre-merge-status-check`'s `needs`.
Drift is caught at merge time, by the author who caused it. Enforcing it again at publish adds no
signal, only blast radius.

This bit twice on 2026-08-11 — #12985 and #13035 — where a handful of stale generated pages
stopped the entire site from updating.

## Change

One line: `continue-on-error: true` on that step, plus a comment explaining why, so the next
person does not "fix" it back.

The step still runs and still reports, so drift stays visible in the run.

## Validation

- `fern-docs.yml` parses; the check is step 3 of 20, `continue-on-error: true`, and
  `Checkout docs-website branch` is still the next step.
- `pre-merge.yml`'s `api-docs` job is **untouched** — no `continue-on-error` on any of its steps,
  still in `pre-merge-status-check`'s `needs`.

## Related

- #13050 reconciles drift that reaches `main`, which is the other half of this.
- Neither replaces the per-PR gate in `api_freshness.py`, which is correctly scoped and should stay.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/13056" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Documentation**
  * API reference publishing can proceed when generated documentation differences are detected.
  * Added automated regeneration of Python, Rust, Kubernetes, and LLM API references after changes to the main codebase.
  * Generated documentation updates are prepared for review automatically when differences are found.


dagil-nvidia:
Closing — the gate stays blocking, by decision.

The argument for this change was that blocking is inverted: a skipped publish leaves docs.nvidia.com serving pages *older* than the ones being withheld, and takes every unrelated docs change with it. That part still holds.

What it missed is that the gate is also the only forcing function. Drift got fixed twice on 2026-08-11 precisely because the publish broke and someone felt it. Nothing here notifies on drift — `notify-slack.yml` is `workflow_call` only, wired to nightly and post-merge — so a `continue-on-error` step failing would page nobody and show as a green job. Landing this alone would have traded a loud, correct-for-the-wrong-reason signal for silence.

The real fix is #13050: reconcile drift on main within minutes of the merge that caused it, so the gate stops firing on its own rather than being told not to.

PR #13057, opened 2026-08-11
fix(deps): align aiohttp to 3.14.3 across container images

dagil-nvidia:
## Overview

The upstream TRT-LLM base image ships an older `aiohttp` than the other Dynamo images resolve to. This brings **every** image onto the same floor, `>=3.14.3`.

## Changes

| File | Change |
|---|---|
| `pyproject.toml` | floor → `3.14.3` |
| `container/deps/requirements.common.txt` | floor → `3.14.3` |
| `container/deps/requirements.planner.txt` | floor → `3.14.3` |
| `container/deps/overrides.planner.txt` | **new** — uv override for the planner resolve |
| `container/templates/planner.Dockerfile` | pass `--overrides` |
| `container/templates/trtllm_runtime.Dockerfile` | system-site upgrade + guards, and a whiteout entry |

## The planner override

The planner installs the `ai-dynamo` wheel alongside `aiperf` in a single resolve, and `aiperf` declares `aiohttp~=3.13.3` on every published version, which would otherwise hold that image below the floor.

A `uv` override keeps it in step without editing `aiperf`. **Verified:** the resolve produces an identical package set with and without the override — 130 packages either way, differing only in `aiohttp` (3.13.5 → 3.14.3). The entry can be dropped once `aiperf` widens its own constraint, and the file says so.

## The TRT-LLM image: system site, not the venv

That image builds `FROM` the upstream TRT-LLM base and installs Dynamo wheels with `--no-deps`, deliberately leaving upstream's Python solve alone, so no pin of ours reaches it. The one package is raised explicitly — but it has to go into **system site**, not the venv:

- `VIRTUAL_ENV=/opt/dynamo/venv` is set in that image, so plain `pip`/`uv pip` targets the venv and leaves `/usr/local/lib/python3.12/dist-packages` untouched
- The venv is `--system-site-packages`, so a venv-only install would *shadow* the old copy at import time while leaving it on disk, where the image inventory reads it

So the install runs through `/usr/bin/python3` with `--break-system-packages`, matching the existing opencv removal and DALI upgrade in that same file, which carry the same warning in their comments.

It also needs a **`pre_runtime` whiteout** entry. The upgrade renames the metadata directory (`aiohttp-3.13.5.dist-info` → `aiohttp-3.14.3.dist-info`), and the squash `COPY` only replaces paths that match, so without dropping the base tree first the old dist-info would ship beside the new one. This is the same reason DALI is already on that list.

Guards assert on the **dist-info directory** rather than a version string, since that directory is what an inventory reads: no pre-3.14 metadata may survive, and a 3.14+ one must be present.

**Checked before proposing:**
- Resolution against the base's own constrained set (`datasets`, `fsspec`) changes **only** `aiohttp` — no downgrades, no collateral. Neither caps below 3.14
- `ClientSession`, `TCPConnector`, `ClientTimeout` and `yarl.URL` behave the same on 3.14.3, including chunked-transfer reads
- cp312 `manylinux_2_28` wheels exist for **both** `x86_64` and `aarch64`, so neither architecture builds from source

## CI

An earlier revision passed `frontend`, `planner`, `trtllm-runtime` and `trtllm-runtime-efa` on both architectures. This revision re-runs them with the override and the system-site change in place.


container/deps/requirements.common.txt
```diff
@@ -11,7 +11,7 @@
 # file is passed via --requirement, and applies to the whole install.
 --no-binary imageio-ffmpeg
 
-aiohttp>=3.9.0,<4.0
+aiohttp>=3.14.3,<4.0  # GHSA-cq5v-8q36-5273
```
pvijayakrish: Can we remove the vuln identifier from the comment?

container/templates/trtllm_runtime.Dockerfile
```diff
@@ -258,6 +258,34 @@ RUN --mount=type=bind,source=./container/compliance/enumerate_bundled_decoders.p
...
+# directory (aiohttp-3.13.5.dist-info -> aiohttp-3.14.3.dist-info) and the squash
+# COPY only replaces paths that match, so without dropping the old tree first the
+# base image's dist-info would ship alongside the new one.
+#
+# Checked against the base's own constraints (datasets, fsspec), which cap nothing
+# here. cp312 manylinux_2_28 wheels exist for x86_64 and aarch64, so neither
+# architecture builds from source.
+# The guards assert on the dist-info directory rather than a version string,
+# because that directory is what an image inventory reads: no pre-3.14 metadata
+# may survive, and a 3.14+ one must be present in system site.
```
nv-anants, line 283: smaller comments please :)

nv-anants (approved):
some ai slop comments, lgtm otherwise

PR #13061, opened 2026-08-12
fix(frontend): gate Gemma 4 reasoning on enable_thinking

### LINKED ISSUE #12287: Gate Gemma 4 reasoning parsing on explicit enable_thinking
## Summary

Gemma 4 reasoning parsing should be gated on an explicit request signal such as `chat_template_args.enable_thinking=true`. When the Gemma 4 reasoning parser is enabled globally but the request does not enable thinking in the chat template, plain model text can otherwise be interpreted with the wrong parser policy.

Baseten currently keeps this as a narrow compatibility policy: `gemma4` / `gemma-4` reasoning parsing is disabled unless the request explicitly asks the chat template to enable thinking.

## Desired behavior

When the frontend is configured with a Gemma 4 reasoning parser:

```json
{
  "model": "dummy-gemma4-model",
  "messages": [
    {
      "role": "user",
      "content": "answer plainly"
    }
  ]
}
```

plain assistant output should be treated as normal text unless the request includes:

```json
{
  "chat_template_args": {
    "enable_thinking": true
  }
}
```

## Repro shape

Use dummy text such as `plain answer` as the model output. No customer prompt, customer completion, request ID, API key, URL, or tenant/model identifier is needed.

This might belong in Dynamo frontend/preprocessor policy rather than `frontend-crates` parser code, since `frontend-crates` already has Gemma 4 parser support.

Skenix64:
## Summary

Gate Gemma 4 reasoning parsing on an explicit `chat_template_args.enable_thinking=true` request signal.

Without an explicit opt-in, Gemma 4 reasoning parsing is now disabled so parser-visible channel markers are preserved as normal assistant content instead of being interpreted as reasoning.

## Changes

- Align the standard frontend reasoning gate with the existing SGLang Gemma 4 opt-in behavior.
- Treat omitted `enable_thinking` as reasoning-disabled for both `gemma4` and `gemma-4`.
- Add regression coverage proving Gemma 4 parser markers remain normal content when thinking is not enabled.
- Add coverage for the `gemma-4` alias.

## Testing

- `cargo test -p dynamo-llm test_is_reasoning_disabled_by_request -- --nocapture`
- `cargo test -p dynamo-llm gemma4_without_enable_thinking_keeps_parser_markers_as_content -- --nocapture`
- `pre-commit run --all-files`
- `git diff --check`

Closes #12287

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/13061" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Gemma 4 reasoning is now opt-in for each request. Requests without explicit thinking enablement keep reasoning markers as regular response content.

* **Bug Fixes**
  * Prevented unintended extraction of reasoning content when thinking is not enabled.

* **Tests**
  * Added coverage confirming streaming responses preserve original marker-bearing content and leave reasoning content empty when reasoning is disabled.


lib/llm/src/preprocessor.rs
```diff
@@ -7283,15 +7153,21 @@ mod tests {
             (
                 Some("gemma4"),
                 None,
-                false,
-                "gemma4 + no args → enabled (parser still runs but is a no-op when no markers arrive)",
+                true,
+                "gemma4 + no args → disabled (reasoning is opt-in)",
```
rmccorm4, line 7829: @KrishnanPrash @indrajit96 for a sanity check here - did we have the wrong/opposite default setting? Does it (chat template) differ across different variants of Gemma4?

indrajit96: Yes for the current official Gemma 4 IT templates, the previous expectation was opposite to the template default.

I checked E2B (https://huggingface.co/google/gemma-4-E2B-it/blob/main/chat_template.jinja#L186), E4B (https://huggingface.co/google/gemma-4-E4B-it/blob/main/chat_template.jinja#L186), 12B
  (https://huggingface.co/google/gemma-4-12B-it/blob/main/chat_template.jinja#L186), 26B-A4B (https://huggingface.co/google/gemma-4-26B-A4B-it/blob/main/chat_template.jinja#L186), and 31B
  (https://huggingface.co/google/gemma-4-31B-it/blob/main/chat_template.jinja#L186). 
  
All use:

  {% set enable_thinking = enable_thinking | default(false) %}

and inject <|think|> only when enable_thinking is true. 
Therefore, an omitted flag means non-thinking mode, **and the reasoning parser should be disabled unless the request explicitly passes enable_thinking=true.**

There is a variant-specific template detail: 12B/26B/31B inject an empty thought channel into the generation prompt when thinking is disabled, while E2B/E4B do not. 
However, the `enable_thinking` default is consistently false across all of them.

I also reproduced the mismatch E2E using the latest Dynamo-vLLM image with `google/gemma-4-E2B-it`, `--dyn-reasoning-parser gemma4`:

  - Omitted flag: 27 prompt / 78 completion tokens; Dynamo removed <channel|> and concatenated the surrounding content.
  - enable_thinking=false: same 27 / 78 tokens, but <channel|> remained ordinary content.
  - enable_thinking=true: reasoning was correctly separated into reasoning_content.

  So the proposed None → disabled behavior aligns the parser with the templates’ effective mode.

lib/llm/tests/postprocessor_parsing_stream.rs
```diff
@@ -2675,6 +2675,44 @@ async fn response_format_gemma4_bare_json_stays_content() {
     assert_eq!(content, json);
 }
 
+#[tokio::test]
+async fn gemma4_without_enable_thinking_keeps_parser_markers_as_content() {
```
indrajit96, line 3239: Could we add Gemma-specific regression coverage showing that `--dyn-default-thinking-mode` enabled enables the parser and disabled keeps it disabled? 
This would protect the interaction between the recently added deployment default and this request-level Gemma gate.

lib/llm/src/preprocessor.rs
```diff
@@ -3967,11 +3965,7 @@ impl OpenAIPreprocessor {
                 | "minimax_m2",
             ) => !Self::deepseek_renderer_reasoning_enabled(chat_template_args, true),
             Some("gemma4") | Some("gemma-4") => {
-                if let Some(enabled) = dynamo_renderer::thinking_bool_from_args(chat_template_args)
-                {
-                    return !enabled;
-                }
-                false
+                dynamo_renderer::thinking_bool_from_args(chat_template_args) != Some(true)
             }
```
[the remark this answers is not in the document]
indrajit96: Confirmed against the current official Gemma 4 IT templates. 
They default enable_thinking to false and only activate the thinking path when enable_thinking=true. thinking_mode="adaptive" alone does not open generated reasoning, so disabling the parser is aligned with the rendered prompt. 
If Gemma adds adaptive-template support later, the gate will need to be updated.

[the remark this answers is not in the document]
indrajit96: @Skenix64 this looks reasonable
SGLang supports only the canonical `gemma4` parser name, while Dynamo exposes gemma-4 as an alias. Without normalization, the SGLang path can use the wrong opt-in fallback and eventually reject gemma-4 as an unsupported parser.

  The fix belongs in `components/src/dynamo/frontend/sglang_prepost.py`:
```
  _SGLANG_PARSER_NAME_ALIASES = {
      "minimax_m3": "minimax-m3",
      "minimax_m3_nom": "minimax-m3",
      "minimax-m3-nom": "minimax-m3",
      "kimi-k3": "kimi_k3",
      "gemma-4": "gemma4",
  }
```

indrajit96 (commented):
A couple of comments else PR is good to go
@Skenix64 thanks a lot for adding this!

lib/llm/tests/postprocessor_parsing_stream.rs
Skenix64: Thanks! Really appreciate the review, I just pushed a follow up addressing both comments.

indrajit96:
@Skenix64 can you take care of the unsigned commits?

lib/llm/src/preprocessor.rs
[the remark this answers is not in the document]
indrajit96: Resolved in latest head

indrajit96 (approved):
LGTM!
Thanks a lot for adding this @Skenix64

harryskim (approved):
LGTM

ziqifan617 (approved):

rmccorm4 (approved):

PR #13068, opened 2026-08-12
feat(vllm): add sidecar LoRA lifecycle

connorcarpenter15:
## Summary

Add dynamic LoRA load, unload, list, discovery, and per-request adapter selection to the vLLM sidecar through native gRPC. Requests naming an unavailable adapter fail instead of silently using the base model.

- Resolve local, Hugging Face, and S3 adapters with Dynamo's shared downloader; use vLLM's assigned adapter IDs and reported capacity.
- Serialize lifecycle changes per worker and coordinate request admission with unloading.
- Reconcile uncertain RPC outcomes, roll back failed publication, restore discovery after failed unload, and republish resident adapters after a sidecar restart.
- Reject startup adapter names that collide with base-model names and allow cancellation while waiting for adapter admission.
- Preserve normal prefix caching. Different weights require a new adapter name; hot swap remains unsupported.

Every participating prefill/decode worker must load the adapter before receiving its requests. vLLM and the sidecar must see adapter directories at the same absolute paths. LoRA requires vllm-project/vllm#52840; NIXL prefill/decode also requires vllm-project/vllm#54814.

## Validation

- Sidecar unit tests after restacking: 57 passed, including all 13 existing LoRA scenarios.
- Clippy with warnings denied passed for the sidecar and vLLM mocker, including all targets.
- Cargo formatting, whitespace checks, and pre-commit hooks passed.
- Previous GPU validation used the unchanged vLLM fork at `fdfed97c`: aggregated adapter lifecycle and base/LoRA generation, NIXL prefill/decode, and native transfer-failure propagation. The subsequent `9f843c25` cache check confirmed repeated LoRA and base prompts each reused 384 cached tokens.
- GPU tests were not repeated after this restack. Previous prefill/decode testing used two engines on one RTX 6000 Ada; multi-GPU and multi-node validation has not run.

## Related Issues

Linear: [DIS-2881](https://linear.app/nvidia/issue/DIS-2881/add-dynamic-lora-lifecycle-support-to-the-vllm-sidecar).

Draft stack, in merge order:
1. Shared API and protocol preparation — #14754.
2. Sidecar LoRA lifecycle — this PR, based on the API prerequisite.
3. Local LoRA launchers — #14574, based on this PR.

Developed with AI assistance.


lib/sidecar/vllm/src/convert.rs
```diff
@@ -380,16 +385,6 @@ fn validate_request(
...
-    if request
-        .routing
-        .as_ref()
-        .and_then(|routing| routing.lora_name.as_deref())
-        .is_some_and(|name| !name.is_empty())
-    {
-        return Err(client::invalid_argument(
-            "LoRA request selection is not supported by vLLM gRPC v0.25.1",
-        ));
-    }
```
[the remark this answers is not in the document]
JulienDarve: Addressed. The `validate_request` check was intentionally removed from `convert.rs`, but an equivalent guard now lives in the engine, where loaded-adapter state is also available: `admit_lora_request` rejects a non-empty `lora_name` at `lib/sidecar/vllm/src/engine.rs:561` when the engine never advertised LoRA support, before any RPC is issued. The request fails with an invalid-argument error naming the adapter rather than being forwarded and failing deep inside the engine.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -320,6 +654,57 @@ impl LLMEngine for VllmSidecarEngine {
...
+
+    async fn on_endpoint_ready(&self, endpoint: Endpoint) -> Result<(), DynamoError> {
+        self.runtime_endpoint.set(endpoint).map_err(|_| {
+            client::engine_shutdown("vLLM sidecar runtime endpoint was already initialized")
+        })
+    }
+
     async fn cleanup(&self) -> Result<(), DynamoError> {
         self.cancel.cancel();
         Ok(())
```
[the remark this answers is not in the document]
JulienDarve: Addressed. `cleanup()` now calls `unpublish_all_loras()` (`lib/sidecar/vllm/src/engine.rs:1094`, implementation at `:682`), which drops every sibling record the sidecar published before cancelling the token. Failures to unpublish an individual record warn rather than abort, so shutdown still completes. The Worker continues to unregister only the base card it attached itself.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -320,6 +654,57 @@ impl LLMEngine for VllmSidecarEngine {
...
+            self.reconcile_loaded_loras().await?;
+            vec![
+                "load_lora".to_string(),
+                "unload_lora".to_string(),
+                "list_loras".to_string(),
+            ]
+        } else {
+            Vec::new()
+        })
+    }
```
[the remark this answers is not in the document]
JulienDarve: Addressed. Restart reconciliation is now best effort. `lora_updates()` (`lib/sidecar/vllm/src/engine.rs:243`) logs a warning and continues when `reconcile_loaded_loras()` fails, and `supported_updates()` (`:1019`) returns `Ok`, so a transient `ListLoras` or discovery failure no longer aborts worker startup for a base model that is serviceable. Reconciliation is retried on the next lifecycle call.

lib/sidecar/vllm/src/client.rs
```diff
@@ -178,6 +178,48 @@ impl VllmClient {
...
+        let mut client = pb::control_client::ControlClient::new(self.pool.next_channel())
+            .max_encoding_message_size(DEFAULT_MAX_GRPC_MESSAGE_SIZE)
+            .max_decoding_message_size(DEFAULT_MAX_GRPC_MESSAGE_SIZE);
+        client
+            .list_loras(pb::ListLorasRequest {})
+            .await
+            .map(tonic::Response::into_inner)
+            .map(|response| response.adapters)
+            .map_err(|status| status_to_dynamo("ListLoras", status))
+    }
```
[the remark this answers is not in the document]
JulienDarve: Addressed. Every LoRA lifecycle RPC runs under a bounded deadline: `LORA_RPC_DEADLINE` is 60s (`lib/sidecar/vllm/src/client.rs:239`) and is applied in the shared driver at `:283`. A stalled server can no longer hold a keyed lock indefinitely; the timeout surfaces as an ambiguous outcome that is reconciled against `ListLoras` rather than being treated as either success or failure.

lib/sidecar/vllm/src/engine.rs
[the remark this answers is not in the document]
JulienDarve: Addressed, same change as the sibling thread on this file. `supported_updates()` no longer propagates a reconciliation error: `lora_updates()` (`lib/sidecar/vllm/src/engine.rs:243`) warns and continues, so `register_engine_updates` cannot fail worker startup on a transient `ListLoras` or discovery error. Retried on the next lifecycle call.

lib/sidecar/vllm/src/lora.rs
```diff
@@ -0,0 +1,249 @@
...
+    card.user_data = Some(serde_json::json!({
+        "lora_adapter": true,
+        "lora_id": adapter.lora_id,
+    }));
+    let spec = DiscoverySpec::from_model_with_suffix(
+        namespace.to_string(),
+        component.to_string(),
+        endpoint.to_string(),
+        &card,
+        derive_lora_suffix(Some(&adapter.lora_name)),
```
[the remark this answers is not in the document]
JulienDarve: Addressed. Suffix collisions are now rejected before publication. `lib/sidecar/vllm/src/lora.rs:33` defines `RESERVED_BASE_SUFFIX`, and the validation path refuses an adapter whose derived suffix is `_base` (`:203`) or collides with the suffix of any currently loaded adapter (`:211`), in addition to the existing raw-name and ID checks. Base-model names are rejected earlier at `:186`.

JulienDarve:
## Upstream vLLM blocker 1 — NIXL handshake crashes on a float `pp_size`

Found while running `lib/sidecar/vllm/launch/disagg_lora.sh` on GPU. This blocks
**disaggregated generation entirely**; it is not specific to LoRA and is not caused by
this PR. Filing here so reviewers know why the disagg generation path in this PR is
still unvalidated.

**Environment:** vLLM `0.28.1rc1.dev247+g73723b707` (contains vllm-project/vllm#52840),
1x RTX 6000 Ada, prefill and decode colocated on GPU 0, `nixl==1.4.0`.

**Symptom** — every request fails its KV load, generation returns zero tokens:

```
NIXL transfer failure: handshake_failed | Context: {'failure_type': 'handshake_failed', ...}
Traceback (most recent call last):
  ...
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/nixl/base_worker.py", line 753, in _nixl_handshake
    range(remote_pp_size), p_remote_ranks
TypeError: 'float' object cannot be interpreted as an integer

Failing 1 request(s) due to KV load failure (failure_policy=fail, 15 tokens affected)
```

**Root cause.** `remote_pp_size` reaches `range()` as a float. It is read from an untyped
dict with no coercion, and the dataclass annotation is not enforced at runtime:

| Location | Code |
| --- | --- |
| `nixl/base_scheduler.py:213` | `pp_size = params.get("pp_size", 1)` — no `int()` |
| `nixl/metadata.py:213` | `pp_size: int = 1` — annotation only; dataclasses do not validate |
| `nixl/base_worker.py:753` | `itertools.product(range(remote_pp_size), p_remote_ranks)` — raises |

`params` is `kv_transfer_params`, which crosses a process boundary. Any serialization
without a distinct integer type delivers `1.0` instead of `1` — notably
`google.protobuf.Struct`, whose `Value` carries only `number_value` (a double).

**Suggested upstream fix:** coerce at the boundary, `int(params.get("pp_size", 1))`.
`tp_size` and `dcp_size` are read the same way and have identical exposure.

**Caveat, stated plainly:** I have not pinned the exact producer of the float. Dynamo's
own conversion restores whole doubles to integers (`lib/sidecar/vllm/src/json.rs:91`), so
I cannot currently attribute it to the protobuf hop. The defensive-coercion ask stands
regardless — a value reaching `range()` from an untyped dict should be coerced.

**What this does *not* block.** Everything else in the disagg path works. Both workers
launch, and the adapter loads on `:8081` and `:8082` with the same server-assigned
`lora_id`, confirming the eager double-load this PR documents.

JulienDarve:
## Upstream vLLM blocker 2 — internal generation failures terminate the stream with a clean `STOP`

Surfaced by the NIXL failure above, but independent of it. Also **not caused by this PR**,
and not the masked-failure bug already fixed by #13381 — that one is fixed and verified
still in place. Recording it here because it is what made blocker 1 hard to diagnose.

**Symptom.** After the KV load fails, `vllm-rs` logs the failure:

```
ERROR [stream.rs:102] request failed with an internal error during generation
```

but the gRPC `GenerateStream` ends with `finish_reason = STOP` and zero tokens. The
consumer therefore returns a perfectly normal success:

```json
{"choices":[{"finish_reason":"stop","message":{"content":null}}],
 "usage":{"prompt_tokens":15,"completion_tokens":0,"total_tokens":15}}
```

HTTP 200, `request.outcome="success"`. A client cannot distinguish *"the model produced
nothing"* from *"generation failed internally."*

**Root cause — the wire protocol cannot express failure.** `inference.proto`'s enum is
exhaustively:

```protobuf
enum FinishReason {
  NOT_FINISHED = 0;  // Possibly more tokens to be streamed
  LENGTH = 1;        // Finished due to length constraint
  STOP = 2;          // Stop string/token or EOS encountered
  ABORTED = 3;       // Request aborted/cancelled
}
```

There is no error variant, so a terminal response **structurally cannot** carry failure.
The only channel is a gRPC error status, and vLLM is not using it in this path.

**Dynamo behaves correctly given its input.** The sidecar propagates a gRPC error status
(`engine.rs` yields `Err(status_to_dynamo(...))`) and rejects a premature stream end with
`GenerateStream ended before a terminal response`. It received neither — it received
`STOP`, and mapped it faithfully. Both delta generators already return `Err` on
`FinishReason::Error` (`chat_completions/delta.rs:254`, `completions/delta.rs:212`, the
latter from #13381).

**Suggested upstream fix, either of:**

1. Return a gRPC error status (`INTERNAL`) when a request fails mid-generation, instead of
   closing the stream with `STOP`. No proto change; consumers already handle this.
2. Add an `ERROR` variant to `FinishInfo.FinishReason` carrying a message.

Option 1 is the smaller change and matches what Dynamo already expects.

**Prior art in this repo,** which reached the same conclusion from the other side —
#13984: improving these cases "requires that the engine bubbles up the cause either via
exception or return value." With no error variant on the wire, that is currently
impossible for this path.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -151,6 +167,444 @@ impl VllmSidecarEngine {
...
+                             vLLM and Dynamo disagree about adapter state",
+                            request.name, observed.source_path, source_path_arg
+                        )));
+                    }
+                    None => return Err(error.into_dynamo()),
+                }
+            }
+        };
+
+        if let Err(error) =
```
jthomson04: [P2] Remove discovery records during rollback.

`Discovery::register` can write the record, then fail on its follow-up read. This path unloads the native adapter and forgets its name without removing the discovery record. Requests can still route to it, and shutdown cleanup will miss it.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -221,25 +675,31 @@ impl LLMEngine for VllmSidecarEngine {
         let data_parallel_rank = data_parallel_rank(&request, self.mode);
         let mut proto_request = build_generate_request(request, request_id, self.mode)?;
         proto_request.model.clone_from(&self.model.served_name);
+        if self.model.is_base_model_name(&proto_request.lora_name) {
+            // Routers may address the base model by name through the adapter field.
+            proto_request.lora_name.clear();
+        }
+        let submit = async {
+            let _admission = if proto_request.lora_name.is_empty() {
```
jthomson04, line 739: [P2] Allow concurrent request admission.

Every request for one adapter holds the same exclusive mutex until gRPC response headers arrive. One slow submission therefore blocks all later submissions for that adapter. Use shared admission guards and an exclusive lifecycle guard.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -151,6 +167,444 @@ impl VllmSidecarEngine {
...
+    }
+
+    async fn unload_lora(&self, body: &Value) -> Result<Value, DynamoError> {
+        let lora_name = parse_lora_name(body)?;
+        let client = self.started_client()?;
+        let endpoint = self.ready_endpoint()?;
+        let _update = self.lifecycle.updates.lock().await;
+        let _guard = self.lifecycle.lock(&lora_name).await;
+
+        let loaded = self.native_inventory().await?;
```
jthomson04: [P2] Allow unload to repair conflicting inventory.

If vLLM starts with `math-r8` and `Math-R8`, `native_inventory()` rejects their matching discovery suffixes before unload reaches vLLM. Neither adapter can be removed through this API. The test repairs this by changing the fake inventory directly; it should exercise the unload API.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -151,6 +167,444 @@ impl VllmSidecarEngine {
...
+            Err(error) if error.is_definitive() && error.code == tonic::Code::NotFound => {
+                existing.clone()
+            }
+            Err(error) => {
+                let definitive = error.is_definitive();
+                let still_loaded = if definitive {
+                    Some(existing.clone())
+                } else {
+                    tracing::warn!(%error, %lora_name, "UnloadLora outcome is ambiguous; reconciling");
+                    self.find_loaded(&lora_name).await?
```
jthomson04: [P2] Retry recovery after an uncertain unload.

If unload and the recovery inventory read both fail, the adapter can remain loaded after its discovery record was removed. The completed `lora_reconciled` cell prevents later lifecycle calls from restoring it. Retain pending recovery state and retry when connectivity returns.

lib/sidecar/vllm/src/lora.rs
```diff
@@ -0,0 +1,381 @@
...
+    published: Mutex<BTreeSet<String>>,
+}
+
+impl LoraLifecycle {
+    pub(crate) async fn lock(&self, name: &str) -> LoraGuard {
+        let lock = self
+            .locks
+            .lock()
+            .await
+            .entry(name.to_string())
```
jthomson04: [P2] Reclaim unused adapter locks.

Every distinct name creates a permanent map entry, including rejected names and unloaded adapters. Memory grows with all names ever seen, regardless of `max_loras`. Reclaim entries when no holders or waiters remain; use borrowed lookup to avoid allocating a name on each request.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -20,14 +22,22 @@ use crate::client::{self, CONTROL_SERVICE, INFERENCE_SERVICE, VllmClient};
...
 pub struct VllmSidecarEngine {
     endpoint: GrpcEndpoint,
     model: DiscoveredModel,
     mode: DisaggregationMode,
     transport: GrpcTransportConfig,
     client: OnceCell<VllmClient>,
+    runtime_endpoint: OnceCell<Endpoint>,
+    lora_downloader: OnceCell<LoRADownloader>,
+    lora_reconciled: OnceCell<()>,
+    lora_enabled: bool,
```
jthomson04: [P3] Use boolean naming prefixes.

Use `is_lora_enabled` and `is_hot_swap_requested`; apply the same convention to the new `definitive`, `valid`, and `exists` variables.

lib/sidecar/vllm/src/engine.rs
JulienDarve: Rollback now attempts to remove the adapter’s discovery record before undoing the native load. Failed publication and cleanup remain tracked for later reconciliation and shutdown cleanup. New loads wait for pending cleanup to succeed, preventing failed attempts from accumulating state during a discovery outage. Removal checks the adapter identity so a colliding name cannot remove another adapter’s record. Extended the existing rollback test to cover a record already written when publication fails.

lib/sidecar/vllm/src/engine.rs
JulienDarve: Changed admission to use shared read guards and load/unload to use exclusive write guards. Same-adapter submissions can now overlap while unload still waits for admission to finish. Updated the existing concurrency test to cover two submissions awaiting headers, unload waiting for both, and cancellation while queued behind unload.

lib/sidecar/vllm/src/engine.rs
JulienDarve: Unload now selects the adapter by its exact native name without first requiring the entire inventory to be publishable. Publication still validates the inventory, and discovery removal checks the full adapter identity. The startup-conflict test now repairs the inventory through the unload API instead of editing the fake inventory directly.

lib/sidecar/vllm/src/engine.rs
JulienDarve: Replaced the one-time reconciliation gate with retryable state protected by the lifecycle mutex. Failed operations leave reconciliation pending, so subsequent lifecycle calls can restore discovery after connectivity returns. Extended the failed-unload test to cover an unavailable unload RPC followed by an unavailable inventory read, then recovery through `list_loras`.

lib/sidecar/vllm/src/lora.rs
JulienDarve: Existing-name lookup now borrows the name. When inserting a new name, the registry reclaims entries whose only remaining owner is the registry itself; holders and queued waiters retain their locks. This avoids per-request name allocation and steady-state registry scans. Added one focused test for name churn and preservation of a queued waiter’s lock.

lib/sidecar/vllm/src/engine.rs
JulienDarve: Renamed the new booleans to `is_lora_enabled`, `is_hot_swap_requested`, `is_definitive`, and `is_valid`. The `exists` variable disappeared as part of the discovery identity check.

lib/sidecar/vllm/src/lora.rs
```diff
@@ -0,0 +1,385 @@
...
+    published: Mutex<BTreeSet<String>>,
+}
+
+impl LoraLifecycle {
+    pub(crate) async fn adapter_lock(&self, name: &str) -> Arc<RwLock<()>> {
+        let mut locks = self.locks.lock().await;
+        if let Some(lock) = locks.get(name) {
+            return lock.clone();
+        }
+        locks.retain(|_, lock| Arc::strong_count(lock) > 1);
```
jthomson04: [P3] Keep locks for loaded adapters.

On a lookup miss, `retain` removes every idle lock, including locks for loaded adapters. Sequential requests alternating between adapters A and B therefore allocate a new lock and name each time. Retain published adapters' entries and reclaim rejected or unloaded names.

lib/sidecar/vllm/src/engine.rs
```diff
@@ -151,6 +165,497 @@ impl VllmSidecarEngine {
...
+        let downloader = self
+            .lora_downloader
+            .get_or_try_init(|| async { build_downloader() })
+            .await?;
+        tracing::info!(lora_name = %request.name, uri = %request.uri, "resolving LoRA source");
+        let source_path = resolve_source_path(downloader, &request.uri).await?;
+        let source_path_arg = source_path.to_string_lossy().to_string();
+
+        let adapter = match client
+            .load_lora(request.name.clone(), source_path_arg.clone())
```
jthomson04: [P3] Remove the unnecessary path clone.

Move `source_path_arg` into the RPC. The later error can format `source_path.display()`.

jthomson04 (approved):
The earlier correctness findings are addressed. Approving with two minor follow-ups: retain lock entries for loaded adapters and remove the unnecessary source-path clone. Reviewed the source at 6545d2f0; tests and CI were not checked in this review. The stack still needs the queued-request LoRA routing fix in #14754 before merge.

jthomson04 (approved):
Both remaining comments are addressed: published adapters retain their locks, and the unnecessary path clone is removed. Source reviewed at acfb62f779; tests and CI were not checked in this review.

PR #13073, opened 2026-08-12
refactor(vllm): extract decoder mechanics

furionw:
_Based on #12921._

## Why

The production vLLM handler mixed transport policy with reusable sampling and output conversion. Other adapters should share the subtle streaming, usage, logprob, routed-expert, and router-hint behavior instead of copying it.

## What Change

- Extract request-to-vLLM sampling conversion into `decoder_sampling.py`.
- Extract stateful native-output conversion into `decoder_output.py`.
- Keep compatibility aliases and delegate the existing handler path to the extracted mechanics.

```mermaid
classDiagram
    DecodeWorkerHandler ..> SamplingParams : builds through shared helper
    DecodeWorkerHandler --> _VllmDecodeOutputAdapter : converts native stream
    _VllmDecodeOutputAdapter ..> RequestOutput : adapts output

    note for DecodeWorkerHandler "Retains production transport behavior"
    note for _VllmDecodeOutputAdapter "Owns stateful output conversion"
```

## Test Plan

- Focused vLLM behavior qualification: 226 passed and 5 intentionally skipped; source-aware mypy and pre-commit passed.
- Built stack-tip H100 qualification: 107 workflow, ensemble, and decoder tests passed.

### DIFF
--- a/.github/codeowners/areas.yaml
+++ b/.github/codeowners/areas.yaml
@@ -229,6 +229,7 @@ areas:
   - components/src/dynamo/common/memory/
   - components/src/dynamo/common/tests/
   - components/src/dynamo/common/utils/
+  - components/src/dynamo/workflow/
   - examples/backends/sample/
   - examples/backends/tritonserver/
   - examples/common/
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -160,6 +160,7 @@ core:
   - 'components/src/dynamo/replay/**'
   - 'components/src/dynamo/frontend/**'
   - 'components/src/dynamo/common/**'
+  - 'components/src/dynamo/workflow/**'
   - 'components/src/dynamo/gpu_memory_service/**'
   # TODO(will): Move TokenSpeed to a dedicated framework filter once CI has
   # TokenSpeed Docker images and backend-specific build/test jobs.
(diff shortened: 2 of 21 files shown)


dagil-nvidia (approved):
Approving from the process-codeowners side: the three routing lines regenerate byte-identical (proven load-bearing - removing the areas.yaml line makes the strict gate exit 1 listing all 15 workflow files as unowned, and the changed-files CI gate has no `components/**` catch-all, so the filters.yaml line is required too), every workflow file resolves to runtime-codeowners, and the `core` filter bucket matches its siblings. All four gates pass with known-positive proof; live CI mirrors them.

Stack context worth knowing when merging: the head commit authors only the three decoder files; the entire `components/src/dynamo/workflow/` package and the routing lines in this diff are drift from the stale stack base (authored in the stack-bottom commit) and drop out when the lower PRs land. #12922 is a child of this branch, not a conflict.

Two non-blocking notes:

1. `decoder_output.py:31` and `decoder_sampling.py:113` - the extraction moved the code verbatim but stripped the operational comments documenting exactly the subtleties the PR exists to share: the string-key serde note on `_serialize_prompt_logprobs` (int keys silently drop the chunk and surface as "Stream ended before generation completed" on the frontend), the "skip_special_tokens intentionally NOT forwarded" note, the copy-before-mutate rationale on `extra_args`, and the effective-start clamp rationale. Carry the load-bearing comments into the new modules.
2. `handlers.py:128` - `_DELTA_REQUEST_OUTPUT_KIND` now has zero consumers in handlers and, unlike its preserved-surface neighbors, is independently defined rather than aliased, so the two constants can drift; alias it from `_decoder_sampling` or delete it.

Pre-existing and already on an open thread: the defensive `getattr` pattern on declared vLLM fields is a verbatim carry-over, covered there.

PR #13076, opened 2026-08-12
fix(xpu): gms import and port collisions

VincyZhang:
## Overview:

This PR hardens test behavior for vLLM/GMS paths by removing unintended FPM enablement in non-FPM test flows and ensuring GMS v1 tests skip cleanly when optional dependencies are unavailable (instead of failing at collection time).

## Details:

- Removed implicit FPM activation from XPU vLLM aggregated launch scripts by deleting dynamic `DYN_FORWARDPASS_METRIC_PORT` setup/logging in `examples/backends/vllm/launch/xpu`

- Forced non-FPM path in vLLM fault-tolerance and serve test env prep by clearing inherited toggles:
  - `DYN_FORWARDPASS_METRIC_PORT`
  - `DYN_FPM_TRACE`

- Fixed GMS v1 test collection behavior in dependency-limited images:
  - Expanded GMS dependency probing for v1 paths + `msgspec` in lib/gpu_memory_service/tests/_deps.py
  - Added module-level `HAS_GMS` skip gates before GMS/msgspec imports in `lib/gpu_memory_service/tests`

1. tests/serve/common.py and the three vLLM fault-tolerance files for the env-path change (`env.pop(...)`).
2. examples/backends/vllm/launch/xpu/agg_xpu.sh plus other four XPU launch scripts for FPM removal.
3. lib/gpu_memory_service/tests/_deps.py and lib/gpu_memory_service/tests/test_v1_protocol.py for skip-before-import behavior.

## Related Issues

If you have an issue number, replace this section with:

- Closes #XXXX


VincyZhang:
Updated `fault-tolerance` and `serve` test paths (including XPU launch scripts) to keep these suites on their intended non-FPM execution path. The goal is twofold: preserve test semantics (service/FT behavior, not FPM instrumentation behavior) and prevent extra, non-target side effects, including additional port-binding surfaces that can introduce avoidable port-conflict risk and timing flakiness.

So the intent includes:
1. Avoid silently enabling FPM in tests that are not validating FPM.
2. Reduce accidental port-collision exposure from extra instrumentation components.

VincyZhang:
`fault-tolerance`/`serve` changed tests passed: https://github.com/ai-dynamo/dynamo/actions/runs/31570467577/job/94034048891

tests/fault_tolerance/cancellation/test_vllm.py
```diff
@@ -141,10 +141,10 @@ def __init__(
...
         env["DYN_SYSTEM_PORT"] = str(self.system_port)
         env["DYN_HTTP_PORT"] = str(frontend_port)
-
-        if mode != WorkerMode.AGGREGATED:
-            self.fpm_port = allocate_port(DynamoPortRange.FPM.value)
-            env["DYN_FORWARDPASS_METRIC_PORT"] = str(self.fpm_port)
+        # Keep cancellation tests on the non-FPM execution path even when
+        # developers/CI export FPM-related env vars globally.
+        env.pop("DYN_FORWARDPASS_METRIC_PORT", None)
+        env.pop("DYN_FPM_TRACE", None)
```
kthui: Can we simply remove all settings or poppings of `DYN_FORWARDPASS_METRIC_PORT` and `DYN_FPM_TRACE` environment variables from `tests/fault_tolerance` tests entirely?

because as you mentioned that those tests do not intent to use FPM path, but they also should not break because FPM was supported, meaning those tests should pass under the default environment without needing to know there is a FPM path.

VincyZhang: Implemented in `0ca7fc0f93` and `84ca62e800`.

Removed the explicit `DYN_FORWARDPASS_METRIC_PORT` / `DYN_FPM_TRACE` handling from the fault-tolerance tests, and also removed the equivalent filtering from the shared serve test helper. The tests now do not opt into or out of FPM; they run with the environment provided by the caller. This keeps the default non-FPM behavior unchanged while ensuring the tests can also run when FPM is explicitly enabled by the environment.

kthui (approved):
`tests/fault_tolerance` looks good to me!

nv-anants (approved):

galletas1712 (approved):

karen-sy (approved):

PR #13078, opened 2026-08-12
refactor(workflow): separate plans from runtime binding

furionw:
## Superseded

The `ExecutionPlan` / `WorkflowExecutor` separation originally introduced here has been folded into #12921 so the first local executor uses the final architecture from the start. This removes an avoidable mid-stack refactor and lets #12924 author the aggregated example directly against `WorkflowIR` → `ExecutionPlan` → `WorkflowExecutor`.

No functionality is being dropped; the remote stack now continues directly from #12924 to #13080.


furionw:
Superseded by the architecture fold into #12921. The remote stack now continues directly from #12924 to #13080.

PR #13079, opened 2026-08-12
feat(operator): support KV routing in DGDR

xianlubird:
## Summary

DGDR already had a `KVRouterSpec` type, but `features.kvRouter` was not exposed or applied to generated deployments. As a result, users could not request KV-cache-aware routing through the high-level DGDR API and had to repeat a low-level DGD frontend override.

This change:

- exposes `features.kvRouter.enabled` in the v1beta1 DGDR API;
- configures the generated frontend by setting the `DYN_ROUTER_MODE=kv` environment variable on its main container when enabled;
- keeps the existing round-robin behavior when the field is omitted or disabled;
- preserves the v1beta1-only field through v1alpha1 conversion round trips;
- regenerates the CRD, Python Pydantic model, deep-copy code, and Kubernetes API reference.

Routing is expressed via the `DYN_ROUTER_MODE=kv` env var rather than editing the frontend `command`/`args`. The frontend's `--router-mode` flag reads `DYN_ROUTER_MODE` as its env fallback, so this avoids having to reason about whether the module entrypoint lives in `command` or `args` (e.g. the `command`-form produced by the PVC/model-path flow, which previously caused a duplicate `-m dynamo.frontend`). An explicit user `--router-mode` override still wins because flags take precedence over env vars.

## Validation

- `GOCACHE=/private/tmp/dynamo-dgdr-go-cache go test ./api/v1alpha1 ./api/v1beta1`
- `python3 deploy/operator/api/scripts/validate_pydantic_models.py`
- `python3 docs/fern/scripts/gen_kubernetes_api.py --check`
- `python3 -m py_compile` for the changed profiler and validation modules
- `python3 -m black --check` for the changed Python sources

Focused profiler regression tests cover enabled and disabled KV router configuration. The full profiler pytest module was not run locally because this checkout does not have the compiled `dynamo.llm` Python binding; CI will run the added tests in the project environment.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/13079" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added optional KV-cache-aware routing configuration for generated deployments.
  * When enabled, deployments automatically use KV routing for frontend traffic.
  * Supports KV-router configuration across direct and layered deployment setups.
* **Bug Fixes**
  * Preserved KV-router settings during API version conversion and configuration round trips.
  * Added validation for missing or invalid frontend configuration.
* **Documentation**
  * Updated Kubernetes API documentation and schemas with the new `kvRouter.enabled` setting.


components/src/dynamo/profiler/utils/dgd_generation.py
```diff
@@ -202,6 +212,34 @@ def apply_runtime_version_override(dgdr, config_dict: dict) -> None:
...
+            continue
+        container = get_main_container_dict(component)
+        if container is None:
+            raise ValueError(
+                "KV router requires the frontend to define a main container"
+            )
+        args = container.get("args")
+        if not args:
+            args = ["-m", "dynamo.frontend"]
+        container["args"] = set_unique_argument_value(list(args), "--router-mode", "kv")
```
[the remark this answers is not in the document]
xianlubird: Good point. I updated the helper to ensure the args always start with `-m dynamo.frontend` before setting the canonical `--router-mode kv` value. Existing custom args are preserved, and I added a regression test for the partial-args case.

components/src/dynamo/profiler/utils/dgd_generation.py
```diff
@@ -202,6 +212,34 @@ def apply_runtime_version_override(dgdr, config_dict: dict) -> None:
...
+        return
+
+    raise ValueError("KV router requires a frontend component in the generated DGD")
```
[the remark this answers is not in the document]
xianlubird: Good catch. I updated this in f640bd26a4 so a missing frontend or main container now logs a warning and leaves the generated config unchanged instead of aborting the profiling run. I also removed the early return so every valid frontend is configured, and added coverage for the missing and multi-frontend cases.

components/src/dynamo/profiler/utils/dgd_generation.py
```diff
@@ -202,6 +212,42 @@ def apply_runtime_version_override(dgdr, config_dict: dict) -> None:
...
+        if container is None:
+            logger.warning(
+                "Skipping KV router configuration for frontend %r because it has no main container",
+                component.get("name"),
+            )
+            continue
+        args = container.get("args")
+        args = list(args) if isinstance(args, list) else []
+        if args[:2] != ["-m", "dynamo.frontend"]:
+            args = ["-m", "dynamo.frontend", *args]
```
julienmancuso: [P1] Respect frontend command-form entrypoints

This checks only `args`, but the existing PVC/model-path flow in `_update_frontend_cli` sets `command` to `["python3", "-m", "dynamo.frontend"]` and puts the model flags in `args`. Enabling KV routing then produces `python3 -m dynamo.frontend -m dynamo.frontend ... --router-mode kv`; the second `-m` is treated as an unknown frontend argument and the process exits. Please inspect the combined command/args representation before injecting the module entrypoint, and add coverage for a frontend produced by `_update_frontend_cli`.

julienmancuso (commented):
One blocking correctness issue.

components/src/dynamo/profiler/utils/dgd_generation.py
```diff
@@ -202,6 +212,42 @@ def apply_runtime_version_override(dgdr, config_dict: dict) -> None:
             component["runtimeVersionOverride"] = override
 
 
+def enable_kv_router(config_dict: dict) -> None:
```
tmonty12, line 215: Instead of setting `--router-mode kv` on the frontend args, you could set `DYN_ROUTER_MODE=kv` as an env var on the frontend and not have to reason about argument interpolation. In either case the user dgd overrides would win as flags take precedent over env vars.

dynamo-ops:
## Automatic CI Approval Skipped

Automatic approval requires every commit to have a signature that GitHub reports as verified. The following commits do not meet that requirement:

| Commit | Reason |
|--------|--------|
| `28f7775` | unsigned |

To use automatic approval, sign the listed commits and push again. Alternatively, a maintainer can review this head and comment `/ok to test 28f777575871f994b8c846ad702d3a79bbb3c8ba`; you do not need to rewrite the commits for that manual approval path.

components/src/dynamo/profiler/utils/dgd_generation.py
xianlubird: Fixed in 28f7775758. Instead of editing `command`/`args`, `enable_kv_router` now sets `DYN_ROUTER_MODE=kv` on the frontend's main container, which `--router-mode` reads as its env fallback. This sidesteps the command-form case entirely: the `_update_frontend_cli` output (`command: [python3, -m, dynamo.frontend]`, model flags in `args`) is left untouched, so there's no duplicate `-m dynamo.frontend`. Added a `test_kv_router_handles_command_form_frontend` regression covering exactly that shape.

components/src/dynamo/profiler/utils/dgd_generation.py
xianlubird: Good call — done in 28f7775758. `enable_kv_router` now sets `DYN_ROUTER_MODE=kv` on the frontend main container (via a new idempotent `set_unique_env_value` helper) instead of touching args, so there's no argument interpolation to reason about. As you noted, an explicit user `--router-mode` in the DGD override still wins since flags take precedence over env vars.

xianlubird:
The current CI failure appears unrelated to this PR's code changes: the failing backend job is `planner / Copy to ACR cuda, amd64`, which failed during ACR `skopeo login` with `connect: connection refused`; `backend-status-check` is only failing as the aggregate gate.

deploy/operator/api/v1beta1/dynamographdeploymentrequest_types.go
```diff
@@ -356,8 +356,9 @@ type FeaturesSpec struct {
 	// +kubebuilder:validation:Type=object
 	Planner *runtime.RawExtension `json:"planner,omitempty"`
 
-	// TODO: KVRouter support is not yet implemented in the operator.
-	// KVRouter *KVRouterSpec `json:"kvRouter,omitempty"`
+	// KVRouter configures KV-cache-aware routing for the generated deployment.
```
tmonty12, line 359: nit: make clear in the documentation that dgd overrides takes precedence over this setting, both in the case of `DYN_ROUTER_MODE` environment variable or `--router-mode` flag.

tmonty12 (approved):

deploy/operator/api/v1beta1/dynamographdeploymentrequest_types.go
xianlubird: Addressed in 8c6353534c. I clarified both the API field description and the curated DGDR reference: `spec.overrides.dgd` is applied after the KV-router feature layer, so a same-name `DYN_ROUTER_MODE` override replaces the generated value, while an explicit `--router-mode` flag takes precedence over the environment variable at frontend startup. I also regenerated the CRD and API reference artifacts and removed the stale note that said `features.kvRouter` was not exposed.

kangclzjc (approved):

athreesh (approved):

PR #13084, opened 2026-08-12
feat(sglang): trace native generate requests

Aphoh:
## Summary

Add request-end tracing to native SGLang `/generate` while keeping its public SSE protocol unchanged.

- Route native observations through the same `RequestEndTraceObserver`, request-end record construction, `RequestTracker` metrics, finish/tool metadata reconciliation, completion/drop ordering, and emitter used by `/v1/chat/completions`.
- Keep the SGLang-specific layer as an adapter: project native `output_ids` and terminal `meta_info` into the common internal observation shape, but render the original opaque SGLang response. No OpenAI fields are serialized into `/generate` SSE.
- Select the SGLang engine and trace runtime metadata atomically from the same `WorkerSet`.
- Load a trace tokenizer only for workers advertising `SGLANG_GENERATE_CAPABILITY`; enabling vLLM Generate cannot retain an SGLang-only tokenizer.
- Collect tool IDs/names only when `--dyn-tool-call-parser` is explicitly configured. There are no model/parser guesses.
- Make the shared tool-call finish reconciliation explicit: tool calls rewrite only unset/`stop` to `tool_calls`, preserving `length` and `content_filter`.
- Preserve the normal disabled-tracing passthrough. The only shared-path cost is the eligibility branch needed to choose the original engine stream.
- Leave replay hashing unchanged. This PR no longer contains the unrelated one-pass hashing refactor.

`lib/llm/src/request_trace/sglang.rs` is now 424 total lines (about 220 production lines), down from 702 in the previous revision, because native `/generate` reuses the common request-end machinery instead of maintaining a parallel tracing subsystem.

- `lib/llm/src/request_trace/integration.rs` — shared observer, request-end construction/emission, and stream-drop ordering.
- `lib/llm/src/request_trace/sglang.rs` — the small native SGLang observation adapter.
- `components/src/dynamo/sglang/request_handlers/llm/decode_handler.py` — common internal fields alongside the preserved opaque native response.
- `lib/llm/src/http/service/sglang_generate.rs` — `/generate` integration and atomic worker/runtime selection.
- `lib/llm/src/request_trace/agent_context.rs` — shared finish/tool metadata semantics.
- `lib/llm/src/discovery/watcher.rs` — capability-specific trace-tokenizer loading.

## Related Issues

### Local correctness and quality

Validated after rebasing onto `main` at `e8a4305c16`:

- `cargo test -p dynamo-llm request_trace:: --lib -- --test-threads=1` — 66 passed.
- `cargo clippy -p dynamo-llm --lib -- -D warnings`.
- `cargo fmt --all -- --check`.
- `python3 -m py_compile` for both changed Python files.
- `git diff --check origin/main...HEAD`.
- `git diff --exit-code origin/main...HEAD -- lib/llm/src/request_trace/replay.rs` — no replay hashing changes.

Focused coverage includes disabled-stream identity, native/common finish projection, explicit tool-parser behavior, tool-call finish reconciliation, current `AgentContext` construction including compaction, and request-end emission only after the inner router stream is dropped and its metrics are finalized.

### Final-head CI

At final head `3c64ca3e1a`, pre-commit, root all-target Rust clippy, the full root Rust test job, every other Rust clippy/test shard, and all CodeQL analyses passed. The SGLang dev/local-dev/runtime/runtime-EFA builds, aggregate deploy tests, ARM64 test suite, one-GPU AMD64 suite (parallel and sequential phases), and two-GPU AMD64 suite also passed.

### AWS H100 paired request-trace validation

Built exact semantic source `88815dfecb35d0d004e32d39a9f01ca445c29ff5` with the repository SGLang runtime wheel target and overlaid the resulting wheels into a cached runtime image. The source archive SHA-256 was `7ce3a5477245e6ce946d1897036ff415c406a5b8a60233c5772cde9a79c8f201`; the candidate runtime wheel SHA-256 was `f6368dc57a824a29aa7f5e22eb42bf0e269eba5cd37660ac915bb59b66bff086`. Both frontend and worker asserted that `dynamo._core` loaded from `/candidate`. Final head `3c64ca3e1a` adds only Black's line wrapping and restores the pre-existing public postprocessor test/benchmark seam by adapting its legacy arguments into the same shared observer; production observer behavior is unchanged. The final seam compiled under all-target clippy, and its 14 streaming-usage tests passed.

Ran one H100 with `Qwen/Qwen2.5-0.5B-Instruct`, aggregate routing, request-end JSONL tracing, and explicit `--dyn-tool-call-parser qwen25`. Each case was warmed on both protocols and then measured as a paired `/v1/chat/completions` and native `/generate` request:

- normal agent stop;
- matched stop string (`OK`);
- length-limited non-agent request;
- length-limited agent request;
- real `get_weather` tool call;
- early client disconnect/cancellation after one output token.

The run produced 12 measured `dynamo.request.trace.v1` request-end records. Field-by-field assertions passed for model (where the shared non-agent path records it), input/output/cached token counts, KV hit rate and transfer metric presence, queue depth, prefill/decode worker IDs and DP ranks, replay block size/input length/sequence hashes, normalized and backend finish reasons, matched stop reason, tool name and tool-call structure, session lineage, KV hints, all current compaction fields, request ID presence, matching `x-request-id`, and timing/TTFT/total/ITL field presence. Cancellation emitted one token and no terminal finish metadata on both paths, confirming request-end emission occurs after router metric finalization.

The real tool call emitted `finish_reason=tool_calls`, `backend_finish_reason=stop`, and `name=get_weather` on both paths. Only the generated tool-call ID was normalized for comparison because separate model invocations generate distinct UUIDs.

Native SSE remained byte-shape compatible at the JSON object level: every observed event exposed only SGLang's `text`, `output_ids`, and `meta_info` keys, with no Dynamo/OpenAI internal fields. The focused Python test `test_native_generate_stream_preserves_opaque_response_and_common_fields` also passed against the candidate wheel (1 passed, 82 deselected).

Artifact SHA-256 checksums:

- comparison: `092d241229b436bf4f76cd85abc7088dda9bf579a38dc9c765ab1ee8d4a716c1`
- 12 measured records: `b245bd2eecfe5641566ca7c4a2a08508295763f1223f6fa176538a5f62b81e0e`
- raw trace JSONL: `9f94809a14ef24677da177cf49abe176ba124670651e971005f6d9e4e053d533`
- responses: `6aa3b746d7c9ec9bd881e87b3383f53c490ca16ad68a118e5171d01be4451f00`

Known protocol/schema differences:

- Chat requests infer `agent_context.input_trigger=user_message`; native token-input `/generate` has no chat message role from which to infer that trigger, so it remains unset. All other tested agent-context fields, including compaction, match.
- Absolute timing values differ because the endpoints are separate requests; field presence and the same measurement path match.
- The current trace schema does not expose worker type. Aggregate routing also produces no KV-transfer metric, so both records omit it.
- The existing common non-agent request-end path omits model, `x_request_id`, and finish metadata on both endpoints. This PR preserves that common behavior rather than special-casing `/generate`.

---

<a href="https://nvidia-deprecated.devinenterprise.com/review/ai-dynamo/dynamo/pull/13084" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,839 @@
...
+            Value::String(value) => {
+                Some(dynamo_protocols::types::StopReason::String(value.clone()))
+            }
+            Value::Number(value) => value.as_i64().map(dynamo_protocols::types::StopReason::Int),
+            _ => None,
+        });
+    metadata.record_backend_finish_reason(Some(0), Some(finish_type.to_string()), stop_reason);
+
+    let normalized = match finish_type {
+        "stop" | "eos" | "cancelled" | "abort" => Some(dynamo_protocols::types::FinishReason::Stop),
```
MatejKosec: Remove `"cancelled"` and `"abort"` from this arm. Both strings mean the server terminated the request early — not a natural stop. `dynamo_protocols::types::FinishReason` has no `Cancelled` variant, so the only correct mapping is `None`, which `_ => None` already produces. The raw backend string already flows through `record_backend_finish_reason` at line 432.

```rust
// before
"stop" | "eos" | "cancelled" | "abort" => Some(dynamo_protocols::types::FinishReason::Stop),

// after
"stop" | "eos" => Some(dynamo_protocols::types::FinishReason::Stop),
```

MatejKosec (changes requested):
Two bugs need fixing before this merges. Three more are worth cleaning up in the same pass.

**Bug 1 — `"cancelled"` and `"abort"` map to `Stop` (line 435)**

`"cancelled"` and `"abort"` mean the server cut the request short. A trace record that maps them to `Stop` looks identical to a request that finished cleanly, which breaks any downstream analysis that uses finish reason to distinguish healthy completions from aborted ones.

`dynamo_protocols::types::FinishReason` has no `Cancelled` variant. Drop both strings from the `Stop` arm and let them fall through to `_ => None`. The raw backend string already reaches the record via `record_backend_finish_reason` at line 432, so callers can still inspect it.

**Bug 2 — `record_tool_calls` unconditionally overwrites the finish reason (line 487)**

`record_finish_reason` at line 415 writes the normalized finish reason — which may be `Length` or `ContentFilter`. `record_tool_calls` then runs in a `spawn_blocking` task and calls `metadata.record_choice_finish_reason(0, FinishReason::ToolCalls)` unconditionally whenever tool calls are present. A request that hit the token limit mid-tool-call gets recorded as `ToolCalls`, not `Length`.

Only write `ToolCalls` when the existing normalized finish reason is `Stop` (or unset). Read the current value before overwriting.

**Cleanup 1 — `ToolOutputBuffer` is unbounded after `capturing` flips to `true` (line 351)**

The PR description says "bounded marker-aware tool-output buffering." That bound only applies to the pre-capture lookahead window. Once `self.capturing = true`, every `push` appends unconditionally and returns at line 351 before any trim runs. The drain logic at line 368 is unreachable in capture mode.

An agent request returning a 32K-token tool-call JSON document accumulates roughly 128 MB in the buffer until `finish()` is called. Add a capacity cap — 8–16 KB covers any realistic tool-call envelope — and stop appending when the buffer hits it.

**Cleanup 2 — `spawn_blocking + block_on` for a pure async function (line 244)**

`record_tool_calls` is an `async fn` with no blocking I/O or FFI. `spawn_pending_tool_finalize` runs it as `runtime.spawn_blocking(move || { blocking_runtime.block_on(async move { ... }) })`. This pins a thread from Tokio's blocking pool for the full parse duration. The blocking pool has a fixed cap (default 512), shared with the tokenizer, log rotation, and hub downloads. Under concurrent agentic load with many streams completing at once, the pool threads fill up.

Change to `runtime.spawn(async move { ... })`.

**Cleanup 3 — `expect()` turns a non-fatal trace skip into a task panic (integration.rs:181)**

Before this PR, `shared_replay_metrics` returning `None` produced a warning and disabled tracing for that request. The PR replaces the `None` check with `.expect("positive trace block size was validated")`. The guard in `validate_request_end_trace` only covers `trace_block_size == 0`. Any future code path that adds a second exit to `shared_replay_metrics` will panic the tokio task rather than degrade gracefully.

Restore the `?` or match-on-None pattern, or add a comment explaining why no other `None` path can exist.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,839 @@
...
+        };
+        metadata.record_tool_call(
+            0,
+            tool_call_index,
+            Some(&tool_call.id),
+            Some(&tool_call.function.name),
+        );
+    }
+    if !tool_calls.is_empty() {
+        metadata.record_choice_finish_reason(0, dynamo_protocols::types::FinishReason::ToolCalls);
```
MatejKosec: `record_finish_reason` already wrote a normalized finish reason — possibly `Length` or `ContentFilter` — into this metadata slot before `record_tool_calls` ran. This call overwrites it unconditionally. A request that hit the token limit while producing a tool call gets recorded as `ToolCalls` rather than `Length`.

Read the existing normalized reason before writing. Only set `ToolCalls` when the current value is `Stop` or unset.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,839 @@
...
+            capturing: capture_all,
+        }
+    }
+
+    fn push(&mut self, text: &str) {
+        if text.is_empty() {
+            return;
+        }
+        self.buffer.push_str(text);
+        if self.capturing {
```
MatejKosec: Once `self.capturing` is `true`, the buffer grows without any bound — the drain logic below this return is unreachable in capture mode. The PR description says the buffering is bounded, but that bound only covers the pre-capture lookahead.

A 32K-token tool-call JSON response accumulates roughly 128 MB here. Add a capacity cap and stop appending when the buffer hits it:

```rust
fn push(&mut self, text: &str) {
    if text.is_empty() || self.buffer.len() >= TOOL_OUTPUT_LIMIT {
        return;
    }
    let space = TOOL_OUTPUT_LIMIT - self.buffer.len();
    let end = text.floor_char_boundary(space.min(text.len()));
    self.buffer.push_str(&text[..end]);
    if self.capturing {
        return;
    }
    // ... existing lookahead trim
}
```

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,839 @@
...
+            .finish_reason_metadata
+            .clone()
+            .expect("tool parsing is only active for agent request traces");
+        let Ok(runtime) = tokio::runtime::Handle::try_current() else {
+            tracing::warn!("request-trace tool metadata skipped outside a Tokio runtime");
+            emit_request_end_trace_state(request_end, request_id);
+            return;
+        };
+        let blocking_runtime = runtime.clone();
+        runtime.spawn_blocking(move || {
```
MatejKosec: `record_tool_calls` is an `async fn` with no blocking I/O. Running it inside `spawn_blocking` pins a thread from Tokio's blocking pool for the full parse duration. The blocking pool (default cap 512) is shared with the tokenizer, log rotation, and hub downloads. Under concurrent agentic load with many streams completing at once, the pool threads fill up.

```rust
// before
runtime.spawn_blocking(move || {
    blocking_runtime.block_on(async move {
        record_tool_calls(&metadata, &output, &parser).await;
        emit_request_end_trace_state(request_end, request_id);
    });
});

// after
runtime.spawn(async move {
    record_tool_calls(&metadata, &output, &parser).await;
    emit_request_end_trace_state(request_end, request_id);
});
```

lib/llm/src/request_trace/sglang.rs
Aphoh: Fixed in 2c656b81f9 with a 16 KiB post-marker cap. Appends and the marker-discovery transition both truncate on UTF-8 character boundaries; the new test exercises a multi-byte payload beyond the cap.

lib/llm/src/request_trace/sglang.rs
Aphoh: Fixed in 2c656b81f9. Finalization now uses `runtime.spawn`; the tool input is also capped at 16 KiB, bounding parser work on the async runtime.

lib/llm/src/request_trace/sglang.rs
Aphoh: Fixed in 2c656b81f9. The shared metadata helper now changes the normalized reason to `ToolCalls` only from `None` or `Stop`, while still recording tool IDs/names for length- and content-filter-terminated output.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,418 @@
...
+    metadata.record_backend_finish_reason(Some(0), Some(finish_type.to_string()), stop_reason);
+
+    let normalized = match finish_type {
+        "stop" | "eos" | "cancelled" | "abort" => Some(dynamo_protocols::types::FinishReason::Stop),
+        "length" => Some(dynamo_protocols::types::FinishReason::Length),
+        "content_filter" => Some(dynamo_protocols::types::FinishReason::ContentFilter),
+        "tool_calls" => Some(dynamo_protocols::types::FinishReason::ToolCalls),
+        "function_call" => Some(dynamo_protocols::types::FinishReason::FunctionCall),
+        _ => None,
+    };
```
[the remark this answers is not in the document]
Aphoh: Verified the actual `dynamo-protocols` 5.3.1 type: its OpenAI-compatible `FinishReason` has no `Cancelled` variant. Commit 2c656b81f9 follows the maintainer guidance: only `stop`/`eos` normalize to `Stop`; `cancelled`/`abort` remain available as raw backend reasons. Added a regression test.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,418 @@
...
+                        tracker.record_first_token();
+                    }
+
+                    // Dynamo exposes only SGLang's incremental SSE mode, so each
+                    // response text is a delta and can be appended directly.
+                    if tool_call_parser.is_some()
+                        && let Some(text) = native_response.get("text").and_then(Value::as_str)
+                    {
+                        output_text.push_str(text);
+                    }
```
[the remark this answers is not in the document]
Aphoh: Documented the worker/frontend contract at the observation site. Dynamo forces `incremental_streaming_output = True`, so these are deltas. The new 16 KiB capture cap also bounds memory if a mixed-version worker violates that contract; cumulative versus delta text cannot be inferred reliably from arbitrary output.

lib/llm/src/request_trace/sglang.rs
Aphoh: Fixed in 2c656b81f9. `cancelled` and `abort` now remain raw backend metadata and no longer normalize to `Stop`; regression coverage checks both values.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,418 @@
...
+        };
+        metadata.record_tool_call(
+            0,
+            tool_call_index,
+            Some(&tool_call.id),
+            Some(&tool_call.function.name),
+        );
+    }
+    if !tool_calls.is_empty() {
+        metadata.record_choice_finish_reason(0, dynamo_protocols::types::FinishReason::ToolCalls);
```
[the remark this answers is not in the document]
Aphoh: Fixed in 2c656b81f9. Parsed tool metadata now infers `ToolCalls` only when the normalized reason is unset or `Stop`; `Length` and `ContentFilter` are preserved. Added a regression test for a length-terminated tool call.

lib/llm/src/discovery/model.rs
```diff
@@ -671,6 +678,24 @@ impl Model {
...
+    ) -> Result<GenerateWorkerRuntime, ModelManagerError> {
+        self.select_worker_set_with(|worker_set| {
+            if !worker_set.supports_runtime_capability(capability) {
+                return None;
+            }
+            Some(GenerateWorkerRuntime {
+                engine: worker_set.generate_engine.clone()?,
+                parsing_options: worker_set.parsing_options(),
+                kv_cache_block_size: worker_set.card().kv_cache_block_size,
+            })
```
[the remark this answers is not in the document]
Aphoh: Confirmed the SGLang registration path derives this value from `page_size * dcp_size` and publishes it in the model card. A zero value therefore indicates missing/invalid runtime metadata rather than a valid SGLang worker. We intentionally skip tracing with a warning because replay hashes cannot be represented without a block size.

Aphoh:
@MatejKosec Addressed the review in 2c656b81f9:

- cancellation/abort stay raw-only instead of normalizing to a clean stop
- parsed tools preserve length/content-filter and infer tool_calls only from stop/unset
- captured tool text is UTF-8-safe and capped at 16 KiB
- tool finalization uses an async task instead of pinning the blocking pool
- replay-metadata construction uses graceful Option propagation instead of expect

Added focused regression coverage for cancellation, finish-reason precedence, and buffer capping. Validation: 74 request-trace tests, 11 focused release tests, Clippy with warnings denied, fmt, and diff checks. I also updated the PR description with the repository template sections and the complete AIPerf/microbenchmark results.

MatejKosec (approved):

dagil-nvidia (commented):
Review of head 8cfa84b5: 1 must-fix, 3 consider. All repo gates green locally, CI green at head, the architecture is sound, and the rewrite addressed every prior human review finding. One routing note first: the docs-codeowners review request is stale - it was triggered by a one-line docs edit that the Aug 25 rewrite dropped (its substance had already landed on main via another PR), and head changes code only.

## Must-fix

1. `components/src/dynamo/sglang/request_handlers/llm/decode_handler.py:442` - `sampling_params.get("stop_token_ids", [])` raises `TypeError: 'NoneType' object is not iterable` when a client sends `"sampling_params": {"stop_token_ids": null}`. That input is legal for SGLang (the payload flows opaquely through the Rust `Option<Map>` into SGLang's own pydantic model where the field is `Optional[List[int]] = None`), and the comprehension runs on every native `/generate` request regardless of tracing enablement, so the request fails before any output. Reproduced directly. Every sibling in this file guards the same field with `or []`:
   ```suggestion
                   for token_id in sampling_params.get("stop_token_ids") or []
   ```

## Consider

2. `lib/llm/src/discovery/watcher.rs:570` - when the tokenizer loads solely for tracing, a load failure propagates via `?` and fails registration of a token-native model that served fine before this PR; everything else in this design degrades with a warn. Make the trace-only load non-fatal.
3. `watcher.rs:579` - the comment "tokenizer.is_some() implies a local chat or completions pipeline will be built" is now false (the trace tokenizer can be the sole reason); behavior is unchanged, but the comment justifies the routing expression and will mislead the next editor.
4. `request_trace/integration.rs:921` - `tracker()` panics via `expect`, but `from_legacy_parts(_, record_tokens=false, Some(metadata))` legally builds an observer with no tracker; unreachable today, latent on the seam deliberately re-exposed for tests. Return an Option or document the invariant.

Verified sound: the opaque-SSE claim holds (new top-level fields never reach public `/generate` SSE, with an unknown-field passthrough test); token accounting cannot double-count (delta-only chunks plus absolute-store semantics); tool-call finish reconciliation is correct in both stream orders and structurally fixes both prior human findings; the legacy seam reproduces all four record/metadata combinations of the old behavior; all flag and capability claims fact-checked.

Comment-only review; severity calls stay with the human reviewers.

Aphoh:
Takeover update at c971d85776:

- Merged current main (c76d3208f3) and resolved the WorkerSet, preprocessor, finish-reason, and request-trace module conflicts.
- Fixed native `sampling_params.stop_token_ids: null` by reusing the existing null-safe stop-token helper.
- Made tokenizer loading non-fatal when it is needed only for native SGLang tool-trace enrichment; registration now degrades with a warning.
- Corrected the stale tokenizer/routing comment.
- Removed the observer `expect()` and kept tracker access optional; the rebase also removed an unnecessary finish-reason clone.
- Updated the new dispatch argument required by current main.

The sidecar `/generate` path still feeds the shared request-end observer and emitter used by chat/completions, including agent context, token/timing/worker/replay metrics, backend and normalized finish metadata, stop reason, and parsed tool IDs/names. Public SGLang SSE remains opaque.

Local validation:
- `cargo test -p dynamo-llm request_trace:: --lib -- --test-threads=1` — 74 passed.
- `cargo clippy -p dynamo-llm --lib -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `python3 -m py_compile` for both changed Python files — passed.
- `git diff --check` — passed.

The SGLang pytest could not run in this worktree because pytest/SGLang are not installed; CI is running on the pushed head.

Aphoh:
GPU validation completed on merged head `c971d857760a6461aa475dca85be229ee9124378`.

- Built the exact PR source and ran SGLang 0.5.19 with `Qwen/Qwen2.5-0.5B` on one B200.
- Launched the Rust frontend with `DYN_SGLANG_ENABLE_GENERATE=1` and JSONL request-end tracing, and the SGLang worker with `--dyn-tool-call-parser qwen25`.
- Sent paired `/v1/chat/completions` and native streaming `/generate` requests with identical agent lineage headers.
- Native request deliberately used `sampling_params.stop_token_ids: null`; it completed successfully.
- Native response produced 8 opaque SGLang SSE events plus `[DONE]`; no internal `sglang_response`, `engine_data`, `completion_usage`, `finish_reason_metadata`, or internal `token_ids` fields leaked.
- Both request-end records contained the same session/parent lineage plus request correlation, model, input/output token counts, request/TTFT/total/ITL timing, worker IDs, replay hashes, and normalized/backend finish metadata. Native preserved its client `rid` as the trace request ID.
- Focused assertion harness exit: `E2E_FINAL_EXIT=0`.

The isolated GPU, NATS, and etcd test resources have been deleted.

Aphoh:
Follow-up: `main` advanced again during GPU validation and briefly made the PR conflicting. I merged the new tip `6be0ab241d8d02717e06cdada3c4eb7ba5911031` and pushed refreshed head `283b5ad91c420a4b638270324872de387bb1138b`.

The only manual conflict was an import-list union in `discovery/watcher.rs` (the PR's SGLang generate tracing symbols plus main's new Qwen video processor contract key). Post-merge validation:

- `cargo fmt --all -- --check`: pass
- `git diff --check`: pass
- `cargo test -p dynamo-llm request_trace:: --lib -- --test-threads=1`: 74 passed

The B200 result in the preceding comment was run on the immediately prior PR head; the PR-specific runtime behavior is unchanged by this final main refresh. Fresh CI is now running on `283b5ad91c`.

Aphoh:
Native `dynamo-sglang-sidecar` parity is now included at head `a41f9f902c`.

The sidecar native `/generate` adapter now projects SGLang `output_ids`, choice index, terminal usage (including cached tokens), backend finish reason, and matched stop reason into `LLMEngineOutput` while preserving `engine_data.sglang_response` unchanged for public SSE rendering. This also supplies token IDs to the shared tool-call trace parser.

Validation:

- `cargo test -p dynamo-sglang-sidecar --no-fail-fast`: 39 unit tests plus executable/proto contract tests passed.
- `cargo clippy -p dynamo-sglang-sidecar --all-targets -- -D warnings`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Exact-branch Kubernetes GPU E2E on one B200 with SGLang 0.5.19 and `Qwen/Qwen2.5-0.5B`, using the locally built PyO3 frontend binding plus locally built `dynamo-sglang-sidecar`.

GPU evidence:

- Native `/generate` length case preserved engine-native SSE and emitted a request-end trace with agent session, request/x-request IDs, 1 input token, 8 output tokens, TTFT/ITL/total timing, worker IDs, replay hashes, and `finish_reason=length` / `backend_finish_reason=length`.
- Chat completions through the same sidecar emitted the corresponding fields with 23 input tokens and 8 output tokens.
- Native `/generate` with `stop_token_ids=[315]` emitted engine-native `finish_reason={type: stop, matched: 315}` and the trace recorded `finish_reason=stop`, `backend_finish_reason=stop`, and `stop_reason=315` at both aggregate and choice levels.
- Every observed `/generate` event retained only SGLang `text`, `output_ids`, and `meta_info` fields; common tracing fields did not leak into public SSE.

The isolated GPU pod and its local etcd/NATS test services were deleted after validation. The added file is owned only by `@ai-dynamo/dynamo-backend-sglang-codeowners`.

Aphoh:
Thermo-nuclear code-quality review completed by an independent subagent and incorporated in 5b57dbed6e.

The review found one structural blocker: native HTTP had introduced a second SGLang terminal-response interpreter, including a silent unknown-reason -> successful stop fallback. The follow-up refactor now:

- centralizes typed/fallible finish parsing in `lib/sidecar/sglang/src/response.rs`
- keeps gRPC and native HTTP transport policies explicit and behavior-preserving
- rejects malformed/unknown finish reasons instead of silently tracing them as stop
- preserves native `abort*`, finish-error, and top-level-error behavior
- keeps user-visible numeric stop filtering explicit for `/generate`
- reduces `native_http.rs` to 913 lines and `protocol.rs` to 822 lines; no file crosses 1,000 lines

Validation after the refactor:

- `cargo test -p dynamo-sglang-sidecar --no-fail-fast` — 43 unit tests + executable/proto-contract/doc tests passed
- `cargo clippy -p dynamo-sglang-sidecar --all-targets -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed

The subagent re-reviewed the final diff and reported no remaining high-confidence structural regression or worthwhile simplification.

Aphoh:
Second independent thermo-nuclear review completed and incorporated in 062f6cb714.

This pass found and fixed three additional structural issues:

- Native `/generate` observability projection no longer gates the opaque SGLang transport. Unknown or malformed future finish metadata is forwarded verbatim in both Rust sidecar and Python paths; only unsupported common trace fields are omitted. Dynamo-owned gRPC parsing remains strict.
- Python terminal projection is centralized in a focused typed `response.py` adapter shared by native and token streams. `decode_handler.py` drops from 989 to 923 lines.
- The request-trace `Pending` state was removed. Construction now yields only disabled or fully prepared active tracing, so a missed prepare transition cannot silently discard traces.
- Follow-up review also removed the obsolete infallible `Result`/dead error branch from native HTTP projection.

The subagent re-reviewed the result and confirmed its findings were resolved.

Validation:

- `cargo test -p dynamo-sglang-sidecar --no-fail-fast` — 44 unit tests plus executable/proto-contract/doc tests passed
- `cargo test -p dynamo-llm request_trace::sglang --lib` — 3 passed (2,852 filtered)
- focused native sidecar tests after final cleanup — 12 passed
- `cargo clippy -p dynamo-sglang-sidecar -p dynamo-llm --lib -- -D warnings` — passed
- targeted Ruff lint/format — passed
- Python `py_compile` and response-projection smoke — passed
- Rustfmt and diff checks — passed

The full Python SGLang pytest module was not runnable in the local host environment because its SGLang/Torch test dependencies are not installed; the new pure adapter received syntax/lint and direct projection smoke coverage, and fresh PR CI is now running.

Aphoh:
Current head `82dbde7` has all contributor-side checks green, including pre-commit, all Rust tests/Clippy shards, CodeQL, codeowners, and the aggregate pre-merge gate. Maintainer: please run `/ok to test 82dbde7` to refresh `pull-request/13084` and trigger the full CI pipeline on this head.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,401 @@
...
+        let engine_context = stream.context();
+        let observer = request_end.observer();
+        let tool_observation = tool_trace
+            .as_ref()
+            .map(|_| Arc::new(Mutex::new(ToolTraceObservation::default())));
+        let stream_observer = observer.clone();
+        let stream_tool_observation = tool_observation.clone();
+        let stream = stream.map(move |response| {
+            if let Some(output) = response.data.as_ref() {
+                let usage = output.completion_usage.as_ref();
```
jthomson04, line 111: [P2] Handle native responses from older workers

Workers from the base revision send empty common `token_ids` and keep usage inside `sglang_response`. Upgrading the frontend first therefore records zero output tokens and loses finish metadata. Add boundary conversion for the older response format, or explicitly disable unsupported tracing.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,401 @@
...
+        tokio::spawn(async move {
+            done.await;
+            if let (Some(tool_trace), Some(observation)) = (tool_trace, tool_observation) {
+                let observation = Arc::try_unwrap(observation)
+                    .map(Mutex::into_inner)
+                    .unwrap_or_else(|shared| {
+                        let mut state = shared.lock();
+                        std::mem::take(&mut *state)
+                    });
+                if observation.terminal && !observation.output_ids.is_empty() {
```
jthomson04, line 159: [P2] Preserve failed termination when recording tool metadata

Any finish reason enables tool parsing here, including `Error` and `Cancelled`. If the output contains a tool call, reconciliation changes the unset result after an error, or the normalized cancellation result, to `ToolCalls`. Preserve failed termination when recording tool metadata. Cancellation also reintroduces the [previously resolved concern](https://github.com/ai-dynamo/dynamo/pull/13084#discussion_r3769144180).

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,401 @@
...
+                    output.finish_reason.as_ref(),
+                    output.stop_reason.as_ref(),
+                    input_tokens,
+                    output_tokens,
+                    cached_tokens,
+                );
+
+                if let Some(tool_observation) = stream_tool_observation.as_ref() {
+                    let mut tool_observation = tool_observation.lock();
+                    tool_observation
```
jthomson04, line 132: [P2] Restore a limit on tool trace capture

With an explicit parser, each traced agent request retains all generated tokens, then decodes and parses the full output. Memory and parser work grow with output length and concurrent requests. Restore a capture limit; this is the return of the [resolved buffer issue](https://github.com/ai-dynamo/dynamo/pull/13084#discussion_r3769156386).

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,401 @@
...
+            done.await;
+            if let (Some(tool_trace), Some(observation)) = (tool_trace, tool_observation) {
+                let observation = Arc::try_unwrap(observation)
+                    .map(Mutex::into_inner)
+                    .unwrap_or_else(|shared| {
+                        let mut state = shared.lock();
+                        std::mem::take(&mut *state)
+                    });
+                if observation.terminal && !observation.output_ids.is_empty() {
+                    record_tool_calls(observer.clone(), tool_trace, observation.output_ids).await;
```
jthomson04, line 160: [P3] Move the observer into record_tool_calls

Move `observer` into `record_tool_calls`; it is not used afterward, so this clone is unnecessary.

lib/llm/src/request_trace/sglang.rs
```diff
@@ -0,0 +1,401 @@
...
+
+struct ToolTraceConfig {
+    parser: String,
+    tokenizer: crate::tokenizers::Tokenizer,
+}
+
+#[derive(Default)]
+struct ToolTraceObservation {
+    output_ids: Vec<u32>,
+    terminal: bool,
```
jthomson04, line 34: [P3] Use the boolean naming convention

Rename `terminal` to `is_terminal` and `runtime_config_needed` to `needs_runtime_config` to follow the boolean naming convention.

jthomson04 (commented):
Source review of commit 103578f3636940260df7b04b5d1b9619af98b712: three P2 findings and two P3 cleanup comments. No tests or CI checks were run for this review.

PR #13091, opened 2026-08-12
feat(operator): support appending args in DGDR overrides

xianlubird:
## Overview

`v1beta1` DGDR overrides treat container `args` as an atomic list. This PR adds an explicit append modifier for containers that already exist in the profiler-generated DGD blueprint.

The override helper resolves the named container, combines its existing arguments with the appended arguments, removes the `$patch` directive, and then runs the normal structural merge. No patch state is stored in the resulting DGD or exposed through the DGD API.

## Details

```yaml
spec:
  overrides:
    dgd:
      kind: DynamoGraphDeployment
      spec:
        components:
        - name: Worker
          podTemplate:
            spec:
              containers:
              - name: main
                $patch:
                  args: append
                args:
                - --verbose
```

The behavior is intentionally narrow:

- Omitting `$patch` keeps the existing replacement behavior.
- The only supported action is `args: append`.
- Appended arguments must be non-empty strings.
- The target container must already exist in the generated blueprint.
- The helper returns a normal DGD containing the complete merged argument list and no `$patch` directive.

The profiler-generated Frontend blueprint materializes its `main` container with the default `python3 -m dynamo.frontend` CLI before DGD overrides are applied. Frontend-specific options can therefore use append semantics without repeating the defaults:

```yaml
containers:
- name: main
  $patch:
    args: append
  args: ["--router-mode", "kv"]
```

- `deploy/operator/internal/dgdoverride/override.go` for directive validation and materialization.
- `deploy/operator/internal/dgdoverride/override_test.go` for existing-container and missing-target coverage.

## Validation

- `go test ./internal/dgdoverride ./api/v1alpha1 ./api/v1beta1 ./internal/dynamo`
- `go test ./internal/webhook/validation -run '^TestValidateDynamoComponentDeploymentSharedSpec' -count=1`
- `go vet ./internal/dgdoverride ./api/v1alpha1 ./api/v1beta1 ./internal/dynamo`
- `python3 -m pytest -q docs/fern/scripts/tests/test_gen_kubernetes_api.py`
- `python3 docs/fern/scripts/gen_kubernetes_api.py --check`
- `git diff --check`

## Related Issues


deploy/operator/internal/webhook/validation/shared_v1beta1.go
```diff
@@ -118,6 +118,19 @@ func (v *sharedValidation) validateDynamoComponentDeploymentSharedSpec(
...
+			continue
+		}
+		if spec.PodTemplate == nil || !hasContainerNamed(spec.PodTemplate.Spec.Containers, patch.Name) {
+			allErrs = append(allErrs, field.Invalid(
+				fldPath.Child("containerArgsPatches").Index(i).Child("name"),
+				patch.Name,
+				"must match main or a podTemplate.spec.containers name",
+			))
+		}
+	}
```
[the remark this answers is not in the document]
xianlubird: Thanks for catching this. Fixed in 12b0beadf7 and a37c6d659f: the new validation, rendering, conversion, and persistence blocks now have the required story comments, and the related tests have step-level t.Log headings.

docs/fern/pages/kubernetes/auto-deployment/auto-deploy-with-dgdr.md
```diff
@@ -403,7 +403,34 @@ cannot add a new worker, EPP, or other topology component.
...
+                $patch:
+                  args: append
+                args:
+                - --router-mode
+                - kv
+```
+
+Without `$patch`, specifying `args` keeps the standard `v1beta1` replacement behavior. The append
+modifier supports only `args: append`, requires a non-empty argument list, and targets an existing
+generated component and container by name.
```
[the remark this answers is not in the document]
xianlubird: Good catch. Fixed in 12b0beadf7. The authoritative DGDR reference now documents the opt-in $patch.args: append behavior, its validation rules, and the existing replacement behavior when the modifier is omitted.

deploy/operator/internal/dgdoverride/override.go
```diff
@@ -512,3 +515,68 @@ func prepareBetaComponents(
...
+		appendArgs := make([]interface{}, len(args))
+		for i := range args {
+			appendArgs[i] = args[i]
+		}
+		patches = append(patches, map[string]interface{}{
+			"name":   name,
+			"append": appendArgs,
+		})
+		containers[containerIndex] = container
+	}
```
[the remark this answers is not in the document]
xianlubird: Fixed in 12b0beadf7. The override preprocessor now allows the generated main container but requires every non-main target to already exist in the blueprint. Unknown sidecars fail before structural merge, with coverage for generated main, an existing sidecar, and an unknown sidecar.

deploy/operator/api/v1beta1/dynamocomponentdeployment_types.go
```diff
@@ -221,6 +230,18 @@ type DynamoComponentDeploymentSharedSpec struct {
...
+// ContainerArgsPatch appends command-line arguments to one rendered container.
+type ContainerArgsPatch struct {
+	// name identifies a rendered container. Use main for the generated workload
+	// container or the name of a container in podTemplate.spec.containers.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+
+	// append contains arguments added after defaults and normal overrides.
+	// +kubebuilder:validation:MinItems=1
+	Append []string `json:"append"`
```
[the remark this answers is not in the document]
xianlubird: Fixed in 12b0beadf7. Append now has item-level MinLength=1 schema validation, the override and webhook paths reject empty strings, and regression tests cover the empty-entry case.

deploy/operator/config/crd/bases/nvidia.com_dynamographdeployments.yaml
```diff
@@ -12234,6 +12234,35 @@ spec:
...
+                              minLength: 1
+                              type: string
+                          required:
+                            - append
+                            - name
+                          type: object
+                        type: array
+                        x-kubernetes-list-map-keys:
+                          - name
+                        x-kubernetes-list-type: map
```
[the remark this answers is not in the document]
xianlubird: Fixed in 12b0beadf7. I regenerated both DCD and DGD CRDs, and containerArgsPatches[].append items now have minLength: 1 while retaining minItems: 1 on the array.

sttts:
> The target must be main or an existing named sidecar.

Why `main`? I would expect this only works if the sidecar exists. Append fails if it doesn't.

deploy/operator/api/v1beta1/dynamocomponentdeployment_types.go
```diff
@@ -125,6 +125,15 @@ type DynamoComponentDeploymentSharedSpec struct {
...
 	PodTemplate *corev1.PodTemplateSpec `json:"podTemplate,omitempty"`
 
+	// containerArgsPatches appends arguments to named containers after the
+	// operator has generated component defaults and applied podTemplate
+	// overrides. The DGDR override engine populates this field from an explicit
+	// $patch args=append directive.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	ContainerArgsPatches []ContainerArgsPatch `json:"containerArgsPatches,omitempty"`
```
sttts: this field surprises. I would have expected that the append operation happens when the profiler calls back into the override code. Then the append become invisible in the DGD.

xianlubird: Agreed. Updated in 46d384b0a0: the override helper now resolves the target in the profiler-generated blueprint, combines the argument lists there, and removes `$patch` before the normal merge. I removed the public `containerArgsPatches` field and the render-time application path, so the append operation is no longer visible in the resulting DGD.

xianlubird:
@sttts Agreed on `main` as well. It is no longer a special case: append now fails unless the named container already exists in the generated blueprint. For a Frontend whose generated blueprint does not contain `main`, the override must use normal replacement semantics and provide the complete args, for example `[-m, dynamo.frontend, --router-mode, kv]`.

deploy/operator/internal/dgdoverride/override.go
```diff
@@ -512,3 +518,126 @@ func prepareBetaComponents(
...
+		}
+
+		// Reject empty entries before materializing the combined argument list.
+		for argIndex, arg := range args {
+			if arg == "" {
+				return fmt.Errorf("beta override spec.components[%d].podTemplate.spec.containers[%d].args[%d] must be non-empty when $patch args is append", componentIndex, containerIndex, argIndex)
+			}
+		}
+
+		baseArgs, _, err := unstructured.NestedStringSlice(baseContainer, "args")
```
sttts: [P1] Materialize the complete frontend CLI before appending

The generated-container existence check does not yet make this safe for `Frontend/main`. The production profiler templates already contain that container, but without `command` or `args`. For a remote model name, `_update_frontend_cli` is skipped, so `NestedStringSlice(baseContainer, "args")` returns `found=false` here. Discarding that bit turns the implicit default into an empty list, and the callback emits only the appended flags. The later operator merge then replaces its default `args: [-m, dynamo.frontend]`; for example, the Pod starts as `python3 --router-mode kv`.

I suggest making the profiler output explicit before `apply_dgd_overrides`, independently of whether local model flags are patched:

```yaml
command: ["python3"]
args: ["-m", "dynamo.frontend"]
```

This deliberately matches `FrontendDefaults`, so the later merge remains idempotent, while append produces the complete list `[-m, dynamo.frontend, ...appended]`. Please also retain `found` above and reject append when the target has no explicit base `args`. This protects arbitrary sidecars whose absent args may mean "use the image CMD"; setting args would replace that CMD rather than append to it. Such a target needs full replacement semantics instead.

Regression coverage should include (1) a remote-HF Frontend without an override, (2) the same Frontend with append and an assertion on the final Pod command/args, and (3) rejection of a sidecar with absent base args.

xianlubird: Thanks, fixed in b0b9248264.

The profiler-owned DGD templates now materialize `Frontend/main` with `command: ["python3"]` and `args: ["-m", "dynamo.frontend"]` before overrides. The append path also keeps the `found` result and rejects targets without an explicit base `args` list, so sidecars that rely on the image CMD must use normal replacement semantics instead.

I added regression coverage for the remote-HF frontend defaults, frontend append through the final Pod command/args, and rejection of a sidecar with absent base args. The docs now show the frontend KV-router append form as well.

Validated with `go test ./internal/dgdoverride ./internal/dynamo` and the profiler blueprint unit tests.

components/src/dynamo/profiler/templates/dgd/vllm/agg.yaml
```diff
@@ -11,7 +11,12 @@ spec:
     podTemplate:
       spec:
         containers:
-        - envFrom:
+        - args:
```
sttts, line 14: [P1] Refresh the DGDR integration goldens for the explicit Frontend CLI

These new fields are correct and address the previous review, but the required `Operator Integration` check now fails deterministically in all three DGDR scenarios (`deployment`, `grove`, and `lws`). Their checked-in `output.yaml` files still expect `Frontend/main` to contain only `image` and `name`, while the observed DGD/DCD objects now also contain exactly:

```yaml
command: ["python3"]
args: ["-m", "dynamo.frontend"]
```

Please update:

- `deploy/operator/internal/controller/testdata/dgdr/deployment/output.yaml`
- `deploy/operator/internal/controller/testdata/dgdr/grove/output.yaml`
- `deploy/operator/internal/controller/testdata/dgdr/lws/output.yaml`

`EventuallyMatchManifests` already writes the complete adapted expectations to `output.yaml.new`, so those generated files should be used to keep every expected DGD and DCD document synchronized. The implementation and template change should remain as-is. After refreshing the goldens, rerun `TestClusterDynamoGraphDeploymentRequestProfilesAndCreatesWorkloadManifests`; this should also clear the derived `backend-status-check` failure.

The PR description should also be updated because it still says the generated Frontend blueprint does not contain `main` or the default arguments.

xianlubird: Good catch — fixed in 1576e32410. The deployment golden was already refreshed in the previous commit, and I have now updated the remaining Grove and LWS expectations as well, including the Frontend DCD in the LWS scenario. I also updated the PR description to reflect that the profiler now materializes the complete Frontend CLI before overrides are applied. The related Go unit tests pass; the build-tagged cluster integration test will be covered by CI.

xianlubird:
CI note: the current failures appear unrelated to this PR. The planner compliance, dynamo-runtime parallel, and vLLM deploy jobs failed while downloading unrelated packages from files.pythonhosted.org because of repeated HTTP 502 responses. The lychee failure is also from a pre-existing SiliconANGLE link returning HTTP 500. These are external service failures rather than regressions caused by this change.

cmdy:
The remaining CI failure requires #13449 to be merged.

xianlubird:
The remaining CI failure requires #13449 to be merged.

sttts (approved):

xianlubird:
CI note: the following four failed checks were investigated and are unrelated to this PR's code changes:

- Fern Docs / Preview or publish docs: Fern preview upload/asset registration failure.
- planner / Compliance cpu, amd64: PyPI dependency installation failure (PyYAML).
- vllm Deploy Test / agg_router: PyPI returned repeated 502 responses during environment setup.
- dynamo-runtime / test / parallel cuda13.0, amd64: PyPI returned repeated 502 responses while installing uvloop for the Python 3.13 wheel smoke test.

The affected test suites either did not start or otherwise passed;

tedzhouhk (approved):

PR #13102, opened 2026-08-12
fix(frontend): return JSON errors for invalid content types and unknown routes

pskiran1:
## Overview:

This PR fixes non-JSON error responses on the OpenAI-compatible HTTP API. OpenAI clients expect error responses to use a JSON error envelope. Previously, some invalid requests returned plain-text or empty responses.

This PR returns the standard `{"message", "type", "code"}` error envelope for:

- Invalid or missing JSON `Content-Type` headers on `POST /v1/responses`
- Unmatched routes, including `GET /v1/responses/{response_id}`

### Responses request parsing
`handler_responses` previously used Axum's `Json<T>` extractor. Because extractors run before the handler, invalid or missing JSON content types were rejected by Axum with a plain-text `415 Unsupported Media Type` response. The handler now reads the request body directly and uses the existing `ensure_json_content_type` and `parse_json_request` helpers, consistent with the Chat Completions and Completions handlers.

### Unmatched routes

Added `unmatched_route_fallback` to return a JSON `404` response when no registered route matches the request path. The fallback is registered after routers are merged, since Axum permits only one custom fallback, and before service layers are applied so fallback responses include the same request-level behavior, such as `x-request-id`. Existing handler-level errors, such as `Model not found`, are unchanged.

## Validation
Added tests:

- `test_responses_non_json_content_type_returns_json_error`
- `test_responses_malformed_json_returns_json_error`
- `test_unknown_response_id_returns_json_404`
- `test_unknown_path_returns_json_404`

Manual verification against a frontend serving `Qwen/Qwen3.5-4B`:

| Request | Before | After |
|---|---|---|
| `POST /v1/responses` with `Content-Type: text/plain` | `415` with plain-text body | `415` with JSON error envelope |
| `POST /v1/responses` without `Content-Type` | `415` with plain-text body | `415` with JSON error envelope |
| `POST /v1/responses` with malformed JSON | Raw Axum error response | `400` with JSON error envelope |
| `GET /v1/responses/resp_x` | `404` with empty body | `404` with JSON error envelope |
| `GET /v1/not_a_route` | `404` with empty body | `404` with JSON error envelope |

Regression checks:

| Request | Expected result |
|---|---|
| `POST /v1/responses` with `application/json; charset=utf-8` | Request parsing succeeds |
| `POST /v1/chat/completions` with an unknown model | Existing `Model not found` response |
| `GET /v1/models` and `GET /health` | `200` responses unchanged |
| `GET /metrics` | Prometheus text response unchanged |
| `GET /v1/responses` | Existing `405` response with `Allow: POST` unchanged |

## Related Issues

- Closes dis-2541

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of response requests with clearer validation for content types and malformed request data.
  * Added consistent, OpenAI-compatible JSON error responses for unsupported media types, invalid bodies, unknown response IDs, and unavailable paths.
  * Error responses now include the relevant request method and path when a route cannot be found.
* **Tests**
  * Added coverage for request validation and standardized error responses.


lib/llm/src/http/service/service_v2.rs
```diff
@@ -1250,6 +1250,10 @@ impl HttpServiceConfigBuilder {
 
         let router = system_router.merge(inference_router);
 
+        // Return an OpenAI-compatible JSON error for unmatched routes.
+        // Configure this after merging routers and before applying layers.
+        let router = router.fallback(super::openai::unmatched_route_fallback);
```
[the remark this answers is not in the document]
pskiran1: Thank you. I moved router.fallback prior add layers and logging is working fine.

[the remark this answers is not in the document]
pskiran1: I added an additional test `test_registered_route_with_unsupported_method_returns_405` to ensure the above.

lib/llm/src/http/service/openai.rs
```diff
@@ -2950,8 +2950,11 @@ pub fn validate_completion_fields_generic(
 async fn handler_responses(
     State((state, template)): State<(Arc<service_v2::State>, Option<RequestTemplate>)>,
     headers: HeaderMap,
-    Json(mut request): Json<NvCreateResponse>,
+    body: Bytes,
 ) -> Result<Response, ErrorResponse> {
+    ensure_json_content_type(&headers)?;
+    let mut request: NvCreateResponse = parse_json_request("responses", &body)?;
```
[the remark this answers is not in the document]
pskiran1: This change is fine and is the established pattern here. `handler_chat_completions` and `handler_completions` already use `body: Bytes` + `ensure_json_content_type` + `parse_json_request`.

lib/llm/src/http/service/openai.rs
```diff
@@ -3055,8 +3055,11 @@ pub fn validate_completion_fields_generic(
 async fn handler_responses(
     State((state, template)): State<(Arc<service_v2::State>, Option<RequestTemplate>)>,
     headers: HeaderMap,
-    Json(mut request): Json<NvCreateResponse>,
+    body: Bytes,
```
rmccorm4: [P2] Map body-buffer failures into the JSON contract. `Bytes` extraction runs before this handler, so an over-limit request still bypasses `ensure_json_content_type` and `parse_json_request`. With `DYN_HTTP_BODY_LIMIT_MB=0`, posting `{}` returned HTTP 413 with `Content-Type: text/plain` and `Failed to buffer the request body: length limit exceeded`. Accept the request body explicitly, validate the content type before buffering, and map limit/read failures into `ErrorResponse`; please add an over-limit request-level regression test.

lib/llm/src/http/service/openai.rs
```diff
@@ -3525,6 +3528,24 @@ pub(crate) fn check_ready(state: &Arc<service_v2::State>) -> Result<(), ErrorRes
     Ok(())
 }
 
+/// Returns an OpenAI-compatible JSON error response for unmatched routes.
+/// Axum's default fallback response has an empty body and does not set a
+/// `Content-Type` header.
+/// This fallback returns `404` with the standard `ErrorMessage` response format.
+pub(crate) async fn unmatched_route_fallback(method: Method, uri: Uri) -> ErrorResponse {
```
rmccorm4: [P2] This fallback cannot always emit the OpenAI envelope because the combined service can also expose the Anthropic Messages API. With Anthropic enabled, `GET /v1/messages/missing` plus `anthropic-version` returned this flat OpenAI 404 instead of the local nested `not_found_error` shape. Select the envelope from the configured Anthropic route namespace (including custom paths), or route Anthropic misses to a protocol-specific fallback.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -1255,6 +1255,10 @@ impl HttpServiceConfigBuilder {
             append_route_docs(&mut all_docs, &mut seen_route_docs, route_docs)?;
             inference_router = inference_router.merge(route);
         }
+
+        // Return an OpenAI-compatible JSON error for unmatched routes.
+        inference_router = inference_router.fallback(super::openai::unmatched_route_fallback);
```
rmccorm4: [P2] Keep unmatched routes outside inference admission. Because `track_inflight_inference` is layered after this, it wraps the fallback and rejects it before the 404 handler runs. After `state.start_draining()`, `GET /v1/not_a_route` returned the JSON 503 `Service is not ready` instead of 404. Merge the registered system/inference routers first, then attach a separately traced unmatched router outside `track_inflight_inference` while retaining the global request-ID layer.

lib/llm/src/http/service/service_v2.rs
```diff
@@ -1698,6 +1702,144 @@ mod tests {
...
+            "Invalid URL (GET /v1/responses/resp_missing)"
+        );
+
+        handle.abort();
+    }
+
+    /// Verifies that an unmatched route returns the standard JSON `404 Not Found`
+    /// error envelope.
+    #[tokio::test]
+    async fn test_unknown_path_returns_json_404() {
```
rmccorm4: [P3] This exercises the same fallback as `test_unknown_response_id_returns_json_404`, with weaker assertions; neither path is registered. Drop the duplicate and use the coverage budget for the missing body-limit, Anthropic-envelope, and drain-state request contracts.

lib/llm/src/http/service/openai.rs
pskiran1: Thanks, Ryan. It was wider than the responses handler, `handler_completions` and `handler_chat_completions` had the same hole. All three now take `Body` and go through a shared `read_json_request_body`, which checks the content type before reading anything and maps a buffering failure to the standard envelope. So a text/plain request is a `415` and oversize request is `413` without the body being read. Added `test_oversized_body_returns_json_413`.

lib/llm/src/http/service/openai.rs
pskiran1: Agreed. `anthropic.rs` now owns an `unmatched_route_response` producing the nested `not_found_error` shape, and the fallback picks the envelope from the namespace the request targeted. The path is read from the same DYN_HTTP_SVC_ANTHROPIC_PATH source the router uses, so custom paths are covered, and it is None when the Anthropic endpoints are disabled. Covered by `test_unmatched_anthropic_route_returns_anthropic_envelope` and a unit test for the namespace matching.

lib/llm/src/http/service/service_v2.rs
pskiran1: Agreed. Moved it off inference_router onto the merged router as a separately traced unmatched router, so it sits outside track_inflight_inference. Added `test_unmatched_route_while_draining_returns_json_404` which asserts the 404 after start_draining() and also asserts a registered inference route still returns 503.

lib/llm/src/http/service/service_v2.rs
pskiran1: Agreed. Dropped duplicate test and used the budget for the three you called out: body limit, Anthropic envelope, and drain state. Thank you.

rmccorm4 (approved):

nv-anants (approved):

Issue #13044, opened 2026-08-11
DEP: Persistent KV event publication across engine restarts


PeaBrane:
## Summary

Introduce a persistent KV-state publication lifecycle that can survive inference-engine restarts without keeping a dead serving worker routable. The initial external memory service integration will run one long-lived local indexer and Dynamo KV event publisher beside the external memory service, consume two raw ZMQ inputs per DP rank—transient engine events and persistent journal-derived external memory service events—and emit one ordered semantic publisher stream to the router. Discovery will associate this persistent KV-state source with the current serving instance through a stable cache-owner identity rather than making the request-plane `WorkerId` persistent.

## Motivation

Dynamo's transport supports multiple event publishers under one scope and topic, but KV-source membership currently interprets `publisher_id` as a replacement incarnation of one logical `(kv_state_endpoint, worker_id, dp_rank)` source. Two concurrent publishers for that logical key are ambiguous and KV routing fails closed. A source registered under a different worker ID is ignored unless that worker is also present in serving discovery; making it a normal serving worker can make it schedulable.

Current source or serving-worker removal also resets the entire rank: primary/device state, every lower tier, and approximate-indexer state. When the final serving worker disappears, the model `WorkerSet` and its router/indexer are removed. These semantics are correct for engine-owned HBM and framework DRAM, but they incorrectly discard controller-owned service-owned DRAM (often called G2.5) and persistent G3 state that physically survives the engine.

The external memory service's recovery design already mirrors committed controller-owned G2/G3 transitions from a journal that survives the engine process. Publishing router events from that journal closes the window where the external memory service commits a block but vLLM dies before relaying its event. A long-lived local indexer can also preserve its authoritative recovery tree across child-engine restarts. Dynamo still needs explicit identity, event, reset, discovery, and dormant-index lifecycle contracts to use those capabilities safely.

### 1. Separate serving, state-owner, publisher, and input-source identities

Define the following distinct identities instead of overloading `WorkerId`:

| Identity | Lifetime and purpose |
|---|---|
| `ServingInstanceId` / current `WorkerId` | Ephemeral, schedulable inference-engine endpoint. Removing it immediately stops request admission. |
| `KvCacheOwnerId` | Stable state-plane owner for a compatible KV pool and DP rank. It may remain live and source-only while no engine is attached. An external memory-service pool ID plus compatibility fingerprint is a likely input. |
[...]

PR #13110, opened 2026-08-12
feat(router): add KV state-agent foundation

PeaBrane:
## Summary

This PR adds the Rust-only internal foundation for a durable KV state agent and follows the ownership/reset contracts introduced by ai-dynamo/dynamo#13053. It links to DEP ai-dynamo/dynamo#13044.

* adds stable `CacheOwnerId` / DP-slot identity without deriving durability from `WorkerId` or DRT connection identity;
* separates exact logical Worker and CacheOwner ownership from physical storage tier, allowing native G2 and KVCC G2.5 to share one HostPinned index;
* adds scoped Worker/CacheOwner resets, exact-owner lower-tier indexing, immutable owner-to-worker projection, and all-tier fuzzy recovery;
* adds an internal state-agent coordinator with one ordered, versioned vLLM-originated stream, bounded ingress, one outbound publisher cursor, status, recovery receipts, and discovery contracts;
* recognizes framework versus post-commit KVCC placement at the raw normalization boundary while retaining the existing raw engine and Worker-only compatibility behavior;
* leaves the existing legacy `KvEventPublisher` activation path untouched.

There is deliberately no public CLI, Python/PyO3 API, vLLM host construction, model-card activation, or residency-v2 router consumer in this PR. This is internal foundation infrastructure, not a usable persistence mode.

## Major assumptions and lifecycle boundary

This PR establishes the internal state-agent lifecycle shape, but does not activate a production host or deliver engine-restart persistence. The planned initial host remains the vLLM handler, where the current fatal engine monitor terminates that handler—and therefore any in-process agent and local indexer—on engine death.

One stable DP slot owns one state agent, local indexer, recovery/control endpoint, outbound publisher, and cursor.

`CacheOwnerId` is the durable identity. `publisher_id` fences the state agent, KVCC-backed state, local indexer, recovery endpoint, and outbound stream as one incarnation.

KVCC and the local indexer are intentionally assumed to share an incarnation. There is no independent KVCC generation until restart/adoption with journal continuity exists.

One ordered, versioned vLLM stream carries framework events and vLLM-enriched, post-commit KVCC transitions. KVCC remains placement authority; vLLM supplies the canonical routing metadata.

Physical tier and logical ownership are orthogonal. Native G2 and KVCC G2.5 may share one HostPinned indexer while remaining exact Worker and CacheOwner ownership.

Engine detach and reattach are modeled as ordered Worker-scoped resets. CacheOwner ownership is retained; KVCC uncertainty disables its projection without destructively clearing it.

CacheOwner is routable only after source identity, attachment readiness, cache readability, and recovery-cursor barriers agree. Failure disables CacheOwner routing while ordinary serving continues.

The implementation assumes compatible block size, parent relationships, and hashing semantics across tiers. Incompatible or incomplete KVCC enrichment fails CacheOwner closed; Dynamo does not rechunk.

Residency-v2 publication, discovery, status, recovery, and projection contracts are foundation infrastructure only. The version-aware router source watcher and aggregate projection lifecycle are not activated end to end in this PR.

Real persistence requires a follow-up that either externalizes the state agent or keeps it alive while replacing EngineCore. Authoritative KVCC replacement without continuity must rotate `publisher_id` or explicitly clear and fully resynchronize CacheOwner state.

## Recovery contract

The all-tier recovery snapshot is intentionally fuzzy rather than a synchronized cross-tier cut. Watermark N is captured after applicable mutations through N are enqueued; the primary drain barrier and each lower-tier FIFO dump barrier independently return state at least N; every event after N is replayed. Any tier failure rejects the entire snapshot.

Cross-incarnation KVCC continuity is not implemented. The foundation assumes an in-order sunny-day stream for one source incarnation. Future replay must use stable KVCC journal cursors, an overlapping retained suffix, and fail-closed gap handling before CacheOwner can become ready.

## Review status

This remains a draft. A final static pass identified follow-up questions around failed-transaction cursor reservation, bounded lifecycle teardown, recovery-cursor activation fencing, positional source-kind edge cases, and legacy-path query/memory/decode overhead. The foundation is not being presented as merge-ready or activation-ready until those are triaged.

## Reviewer guide

The central files are:

* `lib/llm/src/kv_router/publisher/state_agent.rs` — state-agent actor, ingress, detach/reset, publication, and discovery-resource lifetime;
* `lib/kv-router/src/protocols.rs` and `lib/kv-router/src/identity.rs` — exact owner, reset, event, and stable identity contracts;
* `lib/kv-router/src/indexer/lower_tier.rs` and `lower_tier_indexers.rs` — shared physical trees, exact logical ownership, and projection;
* `lib/llm/src/kv_router/indexer/recovery/worker_query_endpoint.rs` — status and identity-bearing recovery receipts;
* `lib/llm/src/discovery/kv_state_agent.rs` — source/attachment resolution and aggregate projection.

## Validation

* `cargo fmt --all`
* `cargo test -p dynamo-kv-router` — 864 passed, 2 ignored
* `cargo test -p dynamo-llm --no-default-features state_agent --lib` — 6 passed
* `(cd lib/kv-router && cargo clippy --no-deps --all-targets -- -D warnings && cargo fmt)`
* `(cd lib/llm && cargo clippy --no-default-features -- -D warnings && cargo fmt)`
* `(cd lib/kvbm-consolidator && cargo clippy --no-default-features -- -D warnings && cargo fmt)`
* `git diff --check`

The requested ComputeLab campaign was canceled before allocation; no remote runtime result is claimed.

## Summary by CodeRabbit

* **New Features**
  * Added stable cache-owner and slot identities for consistent KV state tracking.
  * Added opt-in KV state-agent support for vLLM, including attachment, discovery, recovery, and lifecycle management.
  * Added residency projection so cache state can be mapped to active workers.
  * Added status, recovery, and liveness reporting for KV state agents.
* **Reliability**
  * Improved event routing, replay, reset handling, and acknowledged processing.
  * Added safeguards for stale, ambiguous, unsupported, or invalid state information.
* **Compatibility**
  * Added support for identifying event sources while preserving legacy event formats.


lib/kv-router/src/zmq_wire/deserialize.rs
```diff
@@ -191,29 +201,18 @@ impl<'de> Visitor<'de> for RawKvEventVisitor {
...
-                            metadata.record_trailing(trailing);
-                        }
-                        Some(BlockStoredTrailingField::BlockMmInfos(infos)) => {
-                            block_mm_infos = Some(infos);
-                        }
-                        None => {}
-                    }
+                let mut trailing = Vec::new();
+                while let Some(field) = seq.next_element::<Option<BlockStoredTrailingField>>()? {
+                    trailing.push(field);
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. Tuple readers now consume only the known positional slots, tolerate mismatched legacy fields, and drain appended fields as IgnoredAny. The regression covers a nil legacy block_mm_infos slot plus an arbitrary future suffix.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -193,20 +231,48 @@ impl EdgeOwnersEntry {
...
-                ..
-            } => existing_worker == worker,
-            Self::Multi { owners, .. } => owners.contains_key(worker),
+            Self::Single { owner, .. } => owner.project(projection).as_ref() == Some(worker),
+            Self::Pair { owners, .. } => owners
+                .iter()
+                .any(|owner| owner.project(projection).as_ref() == Some(worker)),
+            Self::Multi { owners, .. } => owners
+                .iter()
+                .any(|owner| owner.project(projection).as_ref() == Some(worker)),
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. Multi-owner edges now keep direct Worker owners in an O(1) set separate from exact CacheOwner keys. Scalar Worker membership is constant-time, and multi-worker walks again intersect the smaller direct side before considering projected owners.

lib/llm/src/kv_router/publisher/state_agent.rs
```diff
@@ -0,0 +1,2097 @@
...
+            if let Some(attachment) = self.attachment.as_mut()
+                && attachment.generation == batch.generation
+            {
+                attachment.ready = false;
+            }
+            self.publish_status();
+            return Err(error);
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. Every locally acknowledged event ID is now reserved on the normal chunk failure path before status is published, matching the detach-drain invariant and preventing cursor reuse.

lib/llm/src/block_manager/kv_consolidator/subscriber.rs
```diff
@@ -153,6 +153,10 @@ fn process_event(
     data_parallel_rank: Option<i32>,
     engine_source: EventSource,
 ) {
+    if event.source_kind() != Ok(KvEventSourceKind::Framework) {
+        tracing::warn!("Ignoring non-framework event on the legacy consolidator stream");
+        return;
+    }
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. The legacy consolidator guard now names the v1.2 to v1.4 rolling-upgrade window and the v1.5 N-2 removal condition; the sibling consolidator path is documented consistently.

lib/llm/src/kv_router/indexer/recovery/worker_query_transport.rs
```diff
@@ -28,36 +32,51 @@ pub(crate) trait WorkerQueryTransport: Send + Sync {
...
+            },
         };
-        let instance = target;
+        let response = tokio::time::timeout(timeout, self.query(target, request))
+            .await
+            .context("KV state-agent status handshake timed out")??;
+        match response {
+            WorkerKvQueryResponse::Status(status) => Ok(status),
+            other => anyhow::bail!("unexpected KV state-agent status response: {other:?}"),
+        }
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. Status errors preserve the bounded server message, while every non-status success variant now reports a constant-size error without formatting its payload.

lib/llm/src/kv_router/publisher/state_agent.rs
```diff
@@ -0,0 +1,2097 @@
...
+                && !self.source_failed
+        });
+        if !valid_generation {
+            return Ok(());
+        }
+
+        if !self.source_cursor.observe(&batch) {
+            self.fail_source(batch.generation, "raw source cursor gap or regression");
+            anyhow::bail!("raw source cursor gap or regression");
+        }
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. Attachment generation validity is checked before envelope limits, so oversized stale-generation input is ignored without retiring the current source. The state-agent regression exercises this ordering.

lib/llm/src/kv_router/publisher/state_agent.rs
```diff
@@ -0,0 +1,2097 @@
...
+            self.worker_domain_failed = true;
+            self.cache_owner_domain_failed = true;
+            if let Some(attachment) = self.attachment.as_mut()
+                && attachment.generation == batch.generation
+            {
+                attachment.ready = false;
+            }
+            self.publish_status();
+            return Err(error);
+        }
```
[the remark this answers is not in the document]
PeaBrane: Fixed in 5c97a40034. The normal transaction failure path now advances the reserved cursor through the LocalKvIndexer watermark before publishing fail-closed status; the regression verifies an accepted prefix cannot collide with a later event.

PeaBrane:
Review-body items 6, 8, and 9 are also addressed in 5c97a40034: CacheOwner construction now has a source-bearing helper, uncertain discovery returns before status I/O, and the acknowledged single-lane path avoids cloning the event.

lib/kv-router/src/identity.rs
```diff
@@ -146,11 +148,23 @@ macro_rules! digest_identity {
...
+        impl FromStr for $name {
+            type Err = IdentityParseError;
+
+            fn from_str(value: &str) -> Result<Self, Self::Err> {
+                Ok(Self::new(
+                    parse_fixed_hex::<16>(value)?,
+                    IdentitySource::Explicit,
+                ))
+            }
+        }
```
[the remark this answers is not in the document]
PeaBrane: Confirmed, but intentionally no code change in this foundation. `FromStr` is the control-plane path for operator-supplied digest text and therefore produces `Explicit`; `DefaultDerived` values are computed internally from compatibility inputs and are serialized structurally when provenance matters. No shipped path round-trips a default-derived identity through `Display`. If a text round-trip is exposed later, it should use a versioned format that carries provenance rather than changing current explicit parsing. Resolving as a deliberate non-fix.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -470,41 +585,16 @@ impl LowerTierIndexer {
...
+        worker_blocks
+            .values()
+            .filter_map(|state| match state.owner {
+                ResidencyOwner::Worker(worker) => Some((worker, state.blocks.len())),
+                ResidencyOwner::CacheOwner(_) => None,
+            })
+            .fold(FxHashMap::default(), |mut counts, (worker, count)| {
+                *counts.entry(worker).or_insert(0) += count;
+                counts
             })
```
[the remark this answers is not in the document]
PeaBrane: Confirmed as intentional. `worker_block_counts` and `get_workers` report directly Worker-owned state; projecting CacheOwner counts here would bake mutable attachment projection into core index statistics and reintroduce duplicate/count ambiguity. CacheOwner visibility is expressed through the immutable query projection. Projected CacheOwner observability should be a separate model-level metric when that consumer is activated. No current consumer relies on CacheOwner counts, so this is deliberately deferred.

lib/llm/src/kv_router/indexer/recovery/worker_query_endpoint.rs
```diff
@@ -154,6 +386,82 @@ impl AsyncEngine<SingleIn<WorkerKvQueryRequest>, ManyOut<WorkerKvQueryResponse>,
...
+                    dynamo_kv_router::protocols::ResidencyDomain::CacheOwner,
+                )
+            } else {
+                dynamo_kv_router::protocols::ResetScope::All
+            };
+            WorkerKvQueryResponse::TreeDump {
+                events,
+                last_event_id,
+                reset_scope,
+            }
```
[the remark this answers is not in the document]
PeaBrane: Confirmed boundary and deliberately retained. Plain `Recovery` is the legacy Worker-only contract. This foundation does not activate a router that consumes CacheOwner events; residency-v2 activation must use `StateAgentRecovery` and must not fall back to plain recovery. The N-2 compatibility TODO now names the removal condition. Because a legacy consumer cannot hold CacheOwner state from this source under the current rollout gate, `ResetScope::All` preserves legacy semantics.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -266,9 +332,23 @@ impl LowerTierIndexer {
...
-                    self.remove_owner_impl(worker_blocks, ResidencyOwner::new(worker, domain))
+                ResetScope::All => {
+                    self.remove_owner_impl(worker_blocks, ResidencyOwner::worker(worker));
+                    if let Some(state_source) = event.state_source {
+                        self.remove_owner_impl(
+                            worker_blocks,
+                            ResidencyOwner::cache_owner(state_source),
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
PeaBrane: Confirmed as unreachable under the foundation rollout gates, so no core behavior change here. A domainless `All` is the legacy Worker-only compatibility form. CacheOwner-bearing state-agent events carry the exact `state_source`, and active CacheOwner traffic is kept off legacy sources/consumers. Clearing every CacheOwner when the source is absent would incorrectly cross independent CacheOwner namespaces. The residency-v2 activation follow-up must reject or withdraw a malformed source-bearing reset before core application.

lib/llm/src/kv_router/indexer/recovery/worker_query_endpoint.rs
```diff
@@ -145,7 +327,57 @@ impl AsyncEngine<SingleIn<WorkerKvQueryRequest>, ManyOut<WorkerKvQueryResponse>,
...
+                        != Some((expected_generation, true))
+                } else {
+                    false
+                };
+            if state_agent_changed {
+                response = WorkerKvQueryResponse::Error(
+                    "KV state-agent attachment or source changed during recovery".to_string(),
+                );
+            }
+        }
```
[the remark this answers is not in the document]
PeaBrane: Declining this because the state-agent identity model forbids the assumed in-place identity swap. `KvStateAgentIdentity`, including CacheOwnerId, publisher identity, and protocol version, is immutable for one recovery-endpoint lifecycle. A replacement owner requires a new publisher/source incarnation and fresh recovery. `StateAgentRecovery` already compares the full expected identity before and after the dump; the legacy Worker path intentionally fences only attachment generation/readiness. A test that mutates CacheOwnerId inside the same status slot would encode an invalid lifecycle.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -193,20 +271,106 @@ impl EdgeOwnersEntry {
         }
     }
 
-    fn contains_worker(&self, worker: &WorkerWithDpRank) -> bool {
+    #[inline]
+    fn contains_worker(&self, worker: &WorkerWithDpRank, projection: &ResidencyProjection) -> bool {
```
jthomson04, line 275: Please run the KV indexer Mooncake benchmark before this merges. This PR changes the lower-tier ownership representation, lookup algorithm, and event application path. Please compare the merge base and this PR with the same CPU binding and fresh-process repetitions. Report `generator_valid`, `kept_up`, achieved block operations per second, lookup service p50/p99, queue wait, and drain time.

The existing trace mainly exercises Worker ownership. Please also add a CacheOwner projection case with a shared prefix and owner counts such as 1, 8, 32, and 128. Otherwise, the benchmark will not exercise the new exact-owner scan.

lib/kv-router/src/indexer/lower_tier.rs
```diff
@@ -193,20 +271,106 @@ impl EdgeOwnersEntry {
...
+                workers,
+                exact_owners,
                 ..
-            } => existing_worker == worker,
-            Self::Multi { owners, .. } => owners.contains_key(worker),
+            } => {
+                workers.contains(worker)
+                    || exact_owners
+                        .iter()
+                        .any(|owner| projection.project_key(*owner).as_ref() == Some(worker))
```
jthomson04, line 289: For a CacheOwner-only `Multi` edge, this scans every exact owner and performs a projection lookup for each one. The single-worker walk calls this for every matched block. A popular prefix therefore scales approximately as `prefix_blocks x exact_owners`.

Can we keep a reverse projection from `WorkerWithDpRank` to its attached owner keys, or calculate the projected owner set once per query? The direct Worker path already chooses the smaller set, but the exact-owner path does not.

lib/llm/src/kv_router/publisher/state_agent.rs
```diff
@@ -0,0 +1,2135 @@
...
+                ResidencyDomain::CacheOwner => RouterEvent::with_cache_owner(
+                    event_worker_id,
+                    event,
+                    tier,
+                    self.identity.cache_owner_id,
+                ),
+            };
+            self.local_indexer
+                .apply_event_with_buffer_acknowledged(router_event.clone())
+                .await
```
jthomson04: This waits for `apply_event_with_buffer_acknowledged()` separately for every event. Each call creates a oneshot channel, sends work to an indexer thread, and waits for that thread before processing the next event. A 128-event chunk can therefore make 128 sequential queue-and-wait round trips.

Please add an acknowledged batch operation per FIFO lane. The lane can apply the ordered events and return one result for the complete batch. Please include batch sizes 1, 32, and 128 in the performance results.

lib/llm/src/kv_router/publisher/state_agent.rs
```diff
@@ -0,0 +1,2135 @@
...
+
+    async fn handle_ingress(&mut self, batch: IngressBatch) -> Result<()> {
+        let valid_generation = self.attachment.as_ref().is_some_and(|attachment| {
+            attachment.ready
+                && attachment.generation == batch.generation
+                && !self.worker_domain_failed
+                && !self.source_failed
+        });
+        if !valid_generation {
+            return Ok(());
```
jthomson04: `BeginDetach` sets `attachment.ready = false`, but the listener remains active while discovery unregister operations run. If the coordinator consumes a committed CacheOwner batch during that interval, this returns `Ok(())` without applying it. The listener receives a successful acknowledgement, and the later queue drain cannot recover the batch.

Please separate routing readiness from ingress acceptance. A `Detaching` state should continue to accept CacheOwner events until the listener has stopped, or detach should use an actor-owned listener barrier. Please add a test that sends ingress after `BeginDetach` completes but before listener shutdown.

lib/kv-router/src/zmq_wire/deserialize.rs
```diff
@@ -191,29 +201,23 @@ impl<'de> Visitor<'de> for RawKvEventVisitor {
...
-                    match trailing {
-                        Some(BlockStoredTrailingField::Common(trailing)) => {
-                            metadata.record_trailing(trailing);
-                        }
-                        Some(BlockStoredTrailingField::BlockMmInfos(infos)) => {
-                            block_mm_infos = Some(infos);
-                        }
-                        None => {}
-                    }
+                let mut trailing = Vec::new();
```
jthomson04: This allocates a trailing-field `Vec` for every stored event, and the removed-event path does the same. The current stored layout then collects those fields into a second `Vec`. This parser is event-rate-sensitive, and there are at most five fields.

Please use fixed stack storage or direct positional parsing. This should not require a heap allocation.

lib/llm/src/discovery/kv_state_agent.rs
```diff
@@ -0,0 +1,470 @@
...
+        sources: &[KvStateSourceAdvertisement],
+        attachments: &[KvStateAttachmentAdvertisement],
+        live_workers: &HashSet<WorkerWithDpRank>,
+        recovery_receipt: Option<&KvStateRecoveryReceipt>,
+    ) -> KvStateProjectionResolution {
+        if !discovery_known {
+            return KvStateProjectionResolution::Unknown(KvStateUnknownReason::WatchUncertain);
+        }
+
+        let matching_sources: Vec<_> = sources
```
jthomson04: `reconcile()` scans and collects the full source and attachment lists for one owner. `resolve_kv_state_projection()` then scans and collects both lists again. If the caller invokes this once per owner, reconciliation approaches quadratic work and creates several temporary vectors per owner.

Please index one discovery snapshot by `CacheOwnerId`, then resolve all owners from those maps. That would also make the uniqueness checks clearer.

jthomson04 (commented):
My main concern is the amount of new state that this PR introduces and must keep consistent:

- CacheOwner residency state
- Owner-to-worker projections
- Source and attachment advertisements
- Attachment generations and readiness
- Publication and recovery cursors
- Recovery receipts
- Deduplication state
- Separate worker, CacheOwner, and source failure flags

Correctness depends on coordinated transitions across discovery, ZMQ ingestion, the local indexer, event publication, recovery, attach, and detach. The detach race in one of my inline comments is an example of how these state machines can disagree.

Can we reduce the number of independently maintained states and define one authoritative owner for each remaining state? It would help to document which values are authoritative, derived, or cached, together with the valid transitions and failure behavior. I do not think the current implementation makes those invariants easy to maintain.

lib/kv-router/src/indexer/lower_tier_indexers.rs
```diff
@@ -269,12 +285,32 @@ pub fn query_lower_tiers_with_options(
...
+    device_matches: &MatchDetails,
+    options: LowerTierQueryOptions,
+    projection: &ResidencyProjection,
 ) -> HashMap<StorageTier, LowerTierMatchDetails> {
     // No lower-tier indexers are allocated, so there is no continuation
     // work to perform. Return before validating device score/hash lockstep;
     // that invariant only matters when a lower tier will consume the
     // continuations.
-    if indexers.indexers.read().unwrap().is_empty() {
+    if indexers.is_empty() {
```
jthomson04: The wrapper checks `indexers.is_empty()` before calling this helper, and this helper checks it again. Both checks acquire the indexer `RwLock`. Please remove one check so each lower-tier request does not take the same lock twice.

jthomson04 (changes requested):
Requesting changes based on the seven inline comments in my previous review. In particular, please address the detach correctness issue and provide KV indexer benchmark evidence for the new CacheOwner routing and acknowledged event-ingestion paths. My broader concern remains the amount of independently maintained state and the difficulty of keeping its lifecycle invariants consistent.

lib/llm/src/kv_router/publisher/state_agent.rs
PeaBrane: Addressed in a2eb5bd306. I removed the state-agent-only per-event acknowledged Store/Remove path instead of batching it: ordinary mutations now deliberately share the existing LocalKvIndexer queue-admission-plus-recovery-buffer invariant. Cleared remains acknowledged across every affected physical tier because it is the ordering barrier. I also removed the listener/coordinator per-chunk stop-and-wait acknowledgement, so there is no longer a 1/32/128 sequential-ack path to benchmark; bounded admission and the serialized coordinator preserve ordering.

lib/llm/src/kv_router/publisher/state_agent.rs
PeaBrane: Fixed in a2eb5bd306. Routing readiness and ingress admission are now separate: BeginDetach withdraws readiness while the exact attachment generation remains admissible; after the listener stop fence, CompleteDetach closes admission, drains queued CacheOwner events, rejects an incomplete raw batch fail-closed, and only then publishes the Worker reset. The focused test injects Worker+CacheOwner ingress after BeginDetach and verifies CacheOwner survives the final Worker barrier.

lib/kv-router/src/zmq_wire/deserialize.rs
PeaBrane: Fixed in a2eb5bd306. Both positional decoders now collect the at-most-five trailing fields into fixed stack arrays, and the current stored-event layout no longer builds a second Vec. The tolerant legacy/future-suffix drain is preserved. All 39 ZMQ wire tests pass.

lib/llm/src/discovery/kv_state_agent.rs
PeaBrane: Addressed in a2eb5bd306 by removing the unused per-owner async reconcile convenience wrapper rather than optimizing and freezing that temporary API. The foundation retains the immutable discovery/status contracts, pure fail-closed resolver and aggregate ResidencyProjection. The real version-aware model-wide watcher will own one indexed snapshot and status-probe loop in the activation follow-up.

lib/kv-router/src/indexer/lower_tier_indexers.rs
PeaBrane: Fixed in a2eb5bd306. The delegated helper no longer repeats the empty-registry check; the public wrapper keeps the single early return before loading the projection snapshot.

lib/kv-router/src/indexer/lower_tier.rs
PeaBrane: Ran this on a zero-GPU ComputeLab node before merge: exclusive 144-core Arm Neoverse-V2 (`lego-c2-qs-36`), two NUMA nodes, `numactl --interleave=all`, issuer CPUs `0-3,72-75`, query issuer `71`, backend/query workers `4-70,76-143`. Every repetition was a fresh process over the same canonical Mooncake input (SHA-256 `b434f181...`).

Exact command shape:

```text
numactl --interleave=all <binary> <trace> \
  --benchmark-duration-ms <duration> --query-lanes 128 \
  --issuer-threads 8 --issuer-spin-us 100 \
  --issue-lag-diagnostic-threshold-us 250 \
  --issuer-cpus 0-3,72-75 --query-issuer-cpu 71 \
  --backend-cpus 4-70,76-143 \
  --num-unique-inference-workers 128 \
  --trace-duplication-factor 20 --trace-length-factor 4 \
  --result-json-output <file> [--tiered-cache-owners N] \
  concurrent-radix-tree-compressed --num-event-workers 8
```

**Unmodified merge-base vs head saturation (5 interleaved fresh runs each, 750 ms; medians):**

| | base `bb2dcc126e` | head `a2eb5bd306` |
|---|---:|---:|
| generator_valid | true | true |
| kept_up | false | false |
| achieved block ops/s | 20.540M | 20.608M (+0.33%) |
| lookup p50 / p99 | 3.232 / 6.208 us | 3.264 / 6.208 us |
| queue p99 | 119.771 us | 128.635 us |
| drain | 2.923 s | 2.910 s |

This point intentionally saturates the pipeline, so `kept_up=false` is expected and the comparison is ceiling/drain behavior.

For the CacheOwner path I used a temporary, unpushed adapter around the same open-loop harness. It composes the production CRTC with the actual `LowerTierIndexers`, preloads a shared 64-block HostPinned prefix under N exact CacheOwners, projects only the last owner to one worker, and therefore exercises the scalar exact-owner scan on every matched lower-tier block. The harness/worktree/binary are now removed.

**Kept-up projection matrix (10 fresh runs each for 1/4/8, 6000 ms; medians):**

| owners | valid / kept up | block ops/s | lookup p50 / p99 | queue p99 | drain |
|---:|---|---:|---:|---:|---:|
| 1 | true / true | 12.5021M | 26.031 / 50.926 us | 391.905 us | 33.112 ms |
| 4 | true / true | 12.5020M | 26.831 / 54.990 us | 406.304 us | 33.338 ms |
| 8 | true / true | 12.5028M | 25.935 / 51.678 us | 373.329 us | 33.155 ms |

Diagnostic stress (3 fresh runs): 32 owners remained at 12.4985M block ops/s and 27.359/54.686 us p50/p99; 128 owners showed the expected scan cliff at 12.4993M and 43.902/103.612 us. All were valid and kept up.

I also ran a lower-load 12000 ms, order-rotated 1/4/8 confirmation (5 fresh runs each): throughput stayed 6.2732M/6.2735M/6.2732M; p50 was 24.063/24.927/24.927 us (+3.6% at 4/8). The p99 medians were identical for 4 and 8 rather than scaling (49.310 us vs 43.230 us at 1), confirming that tail variation is dominated by run/scheduler noise in the intended range.

Conclusion: no regression in the existing Worker-only path, and no reverse-index optimization is justified for the expected <=8 exact-owner cardinality. The 128-owner stress point gives us a clear boundary to revisit if real deployments approach it.

lib/kv-router/src/indexer/lower_tier.rs
PeaBrane: Measured this exact path with the temporary Mooncake tiered adapter described in the adjacent benchmark thread: one HostPinned physical index, a shared 64-block prefix under 1/4/8/32/128 exact CacheOwners, only the final CacheOwner projected to one worker, so the scalar walk invokes this scan at every block.

For the expected range, there is no throughput scaling penalty: the 10-run 1/4/8 medians were 12.5021M / 12.5020M / 12.5028M block ops/s, all valid and kept up. The lower-load confirmation put p50 at 24.063 / 24.927 / 24.927 us (+3.6% at 4/8), within the agreed 5% threshold. The higher-load 8-owner p99 was +1.5%; the noisy lower-load p99 was identical at 4 and 8 rather than cardinality-scaled.

At the diagnostic 128-owner point the scan does become material (p50/p99 43.902/103.612 us versus 26.031/50.926 us at 1), but that is outside the expected <=8 range. I am therefore leaving the simple exact-owner representation in place rather than adding a reverse index or per-query owner set now. If observed cardinality approaches the stress range, this measurement gives us a concrete trigger to revisit it.

jthomson04 (approved):

